PR #101, opened 2015-05-19
Update DRF to latest version.

feanil:
Changelog can be found here: http://www.django-rest-framework.org/topics/release-notes/#31x-series


clintonb:
:+1:

PR #102, opened 2015-05-19
Added can_approve and can_deny properties to Refund model

clintonb:
These properties can be used to determine if the Refund can be approved or denied.

XCOM-226

@rlucioni @jimabramson


rlucioni:
Thanks for splitting this off from the rest of the backend work. :+1:

PR #104, opened 2015-05-20
Added Support for Processing Refunds

clintonb:
Seeking preliminary feedback. Tests forthcoming.

XCOM-338

@rlucioni @jimabramson @Nickersoft

PR #105, opened 2015-05-21
Created a new processors package

clintonb:
This reorganization should keep the files smaller and a bit more maintainable.

XCOM-338

Splitting out work from #104.

@jimabramson @rlucioni


ecommerce/extensions/payment/tests/test_processors.py
```diff
@@ -83,14 +84,14 @@ class CybersourceTests(CybersourceMixin, PaymentProcessorTestCaseMixin, TestCase
     PI_DAY = datetime.datetime(2015, 3, 14, 9, 26, 53)
     UUID = u'UUID'
 
-    processor_class = processors.Cybersource
+    processor_class = cybersource.Cybersource
```
rlucioni, line 87: Why not just `Cybersource`? You're importing it directly.

clintonb, line 87: See `cybersource.datetime` below.

rlucioni, line 87: I see it. I understand the need there when mocking a module imported by `cybersource`, but it's not clear to me why you can't access `Cybersource` directly on this line. This is admittedly a minor point, I'm just wondering why you made this choice.

rlucioni:
:+1:

jimabramson:
:+1:

PR #106, opened 2015-05-21
Updated ecommerce API requirement from 0.4.0 to 1.0.0

Nickersoft:
@clintonb @rlucioni FYI: @jimabramson @wedaly

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -17,4 +17,3 @@ paypalrestsdk==1.9.0
 pycountry==1.10
 requests==2.6.0
 
-git+https://github.com/edx/ecommerce-api-client.git@0.4.0#egg=ecommerce-api-client==0.4.0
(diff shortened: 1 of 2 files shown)


requirements/base.txt
```diff
@@ -17,4 +17,4 @@ paypalrestsdk==1.9.0
 pycountry==1.10
 requests==2.6.0
 
-git+https://github.com/edx/ecommerce-api-client.git@0.4.0#egg=ecommerce-api-client==0.4.0
+git+https://github.com/edx/ecommerce-api-client.git@1.0.0#egg=ecommerce-api-client==1.0.0
```
clintonb: Please move this to `test.txt` since this is only needed for tests.

clintonb:
:+1: (after moving)

requirements/base.txt
Nickersoft: @clintonb The entire line, or just the updated line?

rlucioni: @Nickersoft this requirement (the ecommerce API client).

rlucioni:
:+1: once the build passes.

jimabramson:
what they said. :+1:

PR #108, opened 2015-05-21
Updated Coverage Usage

clintonb:
This instantiation of coverage ensures coverage is loaded before any code being tested. This eliminates the need for to instruct coverage to omit numerous files, and ensures coverage reporting is accurate.

@rlucioni @jimabramson

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,36 +1,10 @@
 [run]
 data_file = .coverage
 omit = ecommerce/settings*
-    # Custom code in the user app is covered, but something about the way Django 1.7
-    # loads models causes the rest of the app to appear uncovered. The result is
-    # skewed coverage stats, which is why the app is omitted from coverage reports.
-    ecommerce/user*
-
-    # Custom code in the order models app is covered, but the same issue described
-    # above causes all other code in the model to appear uncovered, resulting in
-    # skewed coverage stats.
-    ecommerce/extensions/order/models*
-    ecommerce/extensions/payment/constants*
-    ecommerce/extensions/payment/models*
-
-    ecommerce/extensions/refund/exceptions*
-    ecommerce/extensions/refund/models*
-    ecommerce/extensions/refund/status*
-
-    # The fulfillment app's status module only contains constants, which don't require
-    # test coverage.
-    ecommerce/extensions/fulfillment/status.py
-
-    # These files are loaded before coverage, thus they are falsely marked as not covered.
-    ecommerce/extensions/checkout/apps.py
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -1,32 +1,6 @@
...
-    ecommerce/extensions/refund/exceptions*
-    ecommerce/extensions/refund/models*
-    ecommerce/extensions/refund/status*
-
-    # The fulfillment app's status module only contains constants, which don't require
-    # test coverage.
-    ecommerce/extensions/fulfillment/status.py
-
-    # These files are loaded before coverage, thus they are falsely marked as not covered.
-    ecommerce/extensions/checkout/apps.py
```
rlucioni, line 25: I would leave this here. App configuration (specifically `ready()`) is applied before tests are loaded.

rlucioni, line 25: Speaking which, why is the module named `apps.py` instead of `config.py`? I'm surprised we missed that.

rlucioni, line 25: The contents of [`checkout/apps.py`](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/checkout/apps.py) should just be moved to [`checkout/config.py`](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/checkout/config.py).

rlucioni:
I'm in favor of making this change, but I'm not sure what should be done about the holes in our coverage it's exposed. Should this be merged despite the failing build, or should the holes be patched up as part of this PR to keep us at 100%?

rlucioni:
Also, one of the rules in our [`.coveragerc`](https://github.com/edx/ecommerce/pull/108/files#diff-6f1f859c6ca5c4cae7310924786cf661R9), `*conf*`, is causing `config.py` modules to be excluded from coverage.

.coveragerc
clintonb, line 25: Changing code is out of scope for this commit. If `ready()` is not covered, we should either add tests for it or ignore it via `#pragma: no cover`. This commit removes the need to omit entire files.

rlucioni, line 25: I agree. I think targeted applications of `#pragma: no cover` are a good bet, coupled with a dev story to come back and clean them up.

clintonb:
@rlucioni updated
FYI @benpatterson

rlucioni:
From offline discussion: we'll decrease the coverage threshold above which builds will pass to 95%. This should serve as a sufficient reminder of what's missing coverage.

rlucioni:
:+1: once the coverage threshold is reduced and you can get a passing build.

PR #109, opened 2015-05-21
Removed Unnecessary Requirements

clintonb:
XCOM-340

@jimabramson @rlucioni

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,10 +1,7 @@
 analytics-python==1.0.3
 Django==1.7.7
 django-appconf==0.6
-django_compressor==1.4
 django_extensions==1.5.5
-django-libsass==0.2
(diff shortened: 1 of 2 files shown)


jimabramson:
:+1:

requirements/base.txt
```diff
@@ -1,10 +1,7 @@
 analytics-python==1.0.3
 Django==1.7.7
 django-appconf==0.6
-django_compressor==1.4
```
rlucioni, line 4: Isn't this required by the refund list and detail views?

rlucioni, line 4: https://github.com/edx/ecommerce/blob/master/ecommerce/settings/base.py#L244

rlucioni, line 4: Resolved offline. django_compressor is included with django-oscar.

PR #110, opened 2015-05-22
Updated CSRF cookie name in JS

clintonb:
This was missed when the CSRF_COOKIE_NAME setting was changed. The settings changes have been moved to base.py so that they work across all environments, and we avoid the need to modify JS solely for local usage.

@jimabramson @rlucioni


rlucioni:
Nice catch. :+1:

PR #111, opened 2015-05-22
JS Cleanup

clintonb:
- Minor copy update
- Corrected variable declaration
- Added variable to cache DOM lookup

FYI @rlucioni @jimabramson


rlucioni:
Taking advantage of the no-thumb policy, huh? :+1:

clintonb:
Taking advantage of the "oh crap, I need to demo this, and don't want to deal with cherry-picking" policy.

PR #112, opened 2015-05-22
Fix issue removing all refund actions after a successfully processed refund

rlucioni:
The selector being used to identify buttons to remove from the DOM after successfully processing a refund was too broad. This change narrows the selector to only those buttons with the correct refund ID. Addresses [XCOM-355](https://openedx.atlassian.net/browse/XCOM-355).


jimabramson:
:+1:

clintonb:
:+1:

PR #114, opened 2015-05-27
Refund analytics

rlucioni:
Adds a receiver of the post_refund signal which emits a business intelligence event if Segment is enabled. I revised the existing post_checkout receiver as well as the tests to make emitting and testing business intelligence events easier. We also had two different RefundTestMixins, which I've consolidated.

@jimabramson, could you please review this? @clintonb and @Nickersoft, FYI.

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -1,0 +1,55 @@
+from functools import wraps
+import logging
+
+from django.conf import settings
+
+
+logger = logging.getLogger(__name__)
+
+
+def is_segment_configured():
+    """Returns a Boolean indicating if Segment has been configured for use."""
+    return bool(settings.SEGMENT_KEY)
+
+
+def parse_tracking_context(user):
+    """Extract user and client IDs from a user's tracking context.
+
+    Arguments:
+        user (User): An instance of the User model.
+
+    Returns:
+        Tuple of strings, user_tracking_id and lms_client_id
+    """
+    tracking_context = user.tracking_context or {}
+
+    user_tracking_id = tracking_context.get('lms_user_id')
+    if user_tracking_id is None:
+        # Even if we cannot extract a good platform user ID from the context, we can still track the
+        # event with an arbitrary local user ID. However, we need to disambiguate the ID we choose
+        # since there's no guarantee it won't collide with a platform user ID that may be tracked
+        # at some point.
+        user_tracking_id = 'ecommerce-{}'.format(user.id)
+
+    lms_client_id = tracking_context.get('lms_client_id')
+
+    return user_tracking_id, lms_client_id
+
+
+def log_exceptions(msg):
+    """Log exceptions (avoiding clutter/indentation).
+
+    Exceptions are still raised. This module assumes that signal receivers are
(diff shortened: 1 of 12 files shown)


ecommerce/extensions/analytics/utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+from django.conf import settings
+
+
+logger = logging.getLogger(__name__)
+
+
+def is_segment_configured():
+    """Returns a Boolean indicating if Segment has been configured for use."""
+    if settings.SEGMENT_KEY is None:
```
clintonb: `return settings.SEGMENT_KEY is not None`

clintonb:
:+1:

ecommerce/extensions/analytics/utils.py
rlucioni: Good call; updated.

rlucioni:
Thanks for the review @clintonb. Go enjoy your vacation!

ecommerce/extensions/analytics/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+from django.conf import settings
+
+
+logger = logging.getLogger(__name__)
+
+
+def is_segment_configured():
+    """Returns a Boolean indicating if Segment has been configured for use."""
+    return settings.SEGMENT_KEY is not None
```
clintonb: Even better: `return bool(settings.SEGMENT_KEY)`. This will account for empty string.

rlucioni: Nice. Updated.

ecommerce/extensions/refund/signals.py
```diff
@@ -1,6 +1,46 @@
...
+        return
+
+    user_tracking_id, lms_client_id = parse_tracking_context(refund.user)
+
+    # Ecommerce transaction reversal, performed by emitting an event which is the inverse of an
+    # order completion event emitted previously.
+    # See: https://support.google.com/analytics/answer/1037443?hl=en
+    analytics.track(
+        user_tracking_id,
+        'Completed Order',
```
jimabramson, line 25: s/Completed/Refunded/  ?   the linked help document doesn't seem to spell it out, but 'Refunded Order' is definitely a segment ecommerce event, as mentioned here https://segment.com/docs/integrations/google-analytics/

rlucioni, line 25: The docs aren't very helpful on this, but we've made this mistake before in the LMS. Segment will map `Refunded Order` to an "Enhanced Ecommerce" event which isn't currently supported by their Python client. The resulting event won't surface within Google Analytics' ecommerce reports.

As a workaround, we want to track an inverted `Completed Order` event, which will have an impact on GA ecommerce reports. For more information, [see here](https://support.google.com/analytics/answer/1037443?hl=en).

jimabramson, line 25: got it - i read up on the inverted order event, just not sure the event should be named 'completed' or 'refunded' - if you've confirmed, I withdraw the concern.

rlucioni, line 25: Cool; I have confirmed.

jimabramson:
:+1:

PR #115, opened 2015-05-29
fix dashboard page path in acceptance tests

jimabramson:
XCOM-378

this change is necessary because requests for / no longer redirect to /dashboard.

@rlucioni please review


rlucioni:
:+1:

PR #116, opened 2015-06-01
in acceptance tests, make LMS’ payment buttons more findable.

jimabramson:
the LMS cybersource and paypal buttons were once implemented as A elements, but recently switched to BUTTON elements.  The change in the PR, instead of expecting any particular DOM element tag, just uses css classes and IDs.

@rlucioni or @clintonb please review

### DIFF
--- a/acceptance_tests/test_payment.py
+++ b/acceptance_tests/test_payment.py
@@ -68,7 +68,7 @@ class VerifiedCertificatePaymentTests(EcommerceApiMixin, EnrollmentApiMixin, Lms
         """ Completes the checkout process via CyberSource. """
 
         # Click the payment button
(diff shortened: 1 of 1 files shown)


clintonb:
Minor: HTML ID elements are supposed to be unique, so we _should_ be able to simply use `#paypal` or `#cybersource` without the class. Regardless, :+1: .

rlucioni:
:+1: @clintonb makes a good point.

jimabramson:
@clintonb @rlucioni it works fine with just the ID so I changed it accordingly.

rlucioni:
:+1:

clintonb:
:+1:

PR #117, opened 2015-06-02
Added Migration Command

clintonb:
Early preview. Tests forthcoming. @rlucioni @jimabramson @Nickersoft

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -1,0 +1,41 @@
+import logging
+
+from django.db import models
+from oscar.core.loading import get_model
+
+logger = logging.getLogger(__name__)
+ProductClass = get_model('catalogue', 'ProductClass')
+
+
+class Course(models.Model):
+    id = models.CharField(null=False, max_length=255, primary_key=True, verbose_name='ID')
+    name = models.CharField(null=False, max_length=255)
+
+    @classmethod
+    def is_mode_verified(cls, mode):
+        """ Returns True if the mode is verified, otherwise False. """
+        return mode.lower() in ('verified', 'professional', 'credit')
+
+    @classmethod
+    def certificate_type_for_mode(cls, mode):
+        mode = mode.lower()
+
+        if mode == 'no-id-professional':
+            return 'professional'
+
+        return mode
+
+    @property
+    def seat_products(self):
+        """ Returns a list of course seat Products related to this course. """
+        seat_product_class = ProductClass.objects.get(slug='seat')
+        products = set()
+
+        for product in self.products.all():
+            if product.get_product_class() == seat_product_class:
+                if product.is_parent:
+                    products.update(product.children.all())
+                else:
+                    products.add(product)
+
+        return list(products)
--- a/ecommerce/courses/tests.py
+++ b/ecommerce/courses/tests.py
@@ -1,0 +1,67 @@
+import ddt
+from django.test import TestCase
+from django_dynamic_fixture import G
+
+from ecommerce.courses.models import Course
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+
+
+@ddt.ddt
+class CourseTests(CourseCatalogTestMixin, TestCase):
+    def test_seat_products(self):
+        """
+        Verify the method returns a list containing purchasable course seats.
+
+        These seats should be the child products.
+        """
+        # Create a new course and verify it has no existing products.
+        course = G(Course)
+        self.assertEqual(course.products.count(), 0)
+        self.assertEqual(len(course.seat_products), 0)
+
+        # Create the seat products
+        seats = self.create_course_seats(course.id, ('honor', 'verified'))
+
+        # Associate the parent and child products with the course. The method should be able to filter out the parent.
+        parent = seats.values()[0].parent
+        parent.course = course
+        parent.save()
+
+        for seat in seats.itervalues():
+            seat.course = course
+            seat.save()
+
+        self.assertEqual(course.products.count(), 3)
+
+        # The method should return only the child seats.
+        # We do not necessarily care about the order, but must sort to check equality.
+        expected = seats.values().sort()
+        self.assertEqual(course.seat_products.sort(), expected)
+
+    @ddt.data(
+        ('verified', True),
+        ('credit', True),
+        ('professional', True),
+        ('honor', False),
+        ('no-id-professional', False),
+        ('audit', False),
+        ('unknown', False),
+    )
+    @ddt.unpack
+    def test_is_mode_verified(self, mode, expected):
+        """ Verify the method returns True only for verified modes. """
+        self.assertEqual(Course.is_mode_verified(mode), expected)
+
+    @ddt.data(
+        ('Verified', 'verified'),
+        ('credit', 'credit'),
+        ('professional', 'professional'),
+        ('honor', 'honor'),
+        ('no-id-professional', 'professional'),
+        ('audit', 'audit'),
+        ('unknown', 'unknown'),
+    )
+    @ddt.unpack
+    def test_certificate_type_for_mode(self, mode, expected):
+        """ Verify the method returns the correct certificate type for a given mode. """
+        self.assertEqual(Course.certificate_type_for_mode(mode), expected)
--- a/ecommerce/extensions/catalogue/management/commands/migrate_course.py
+++ b/ecommerce/extensions/catalogue/management/commands/migrate_course.py
@@ -1,0 +1,214 @@
+import logging
+from optparse import make_option
+from urlparse import urljoin
+
+from django.conf import settings
+from django.core.management import BaseCommand
+from django.db import transaction
+from django.utils.text import slugify
+from oscar.core.loading import get_model
+import requests
+
+from ecommerce.courses.models import Course
+from ecommerce.extensions.catalogue.utils import generate_sku
+
+logger = logging.getLogger(__name__)
+
+Category = get_model('catalogue', 'Category')
+Partner = get_model('partner', 'Partner')
+Product = get_model('catalogue', 'Product')
+ProductCategory = get_model('catalogue', 'ProductCategory')
+ProductClass = get_model('catalogue', 'ProductClass')
+StockRecord = get_model('partner', 'StockRecord')
+
+
+class MigratedCourse(object):
+    def __init__(self, course_id):
+        # Ensure this value is Unicode to avoid issues with slugify.
+        self.course_id = unicode(course_id)
+
+        self.parent_seat = None
+        self.child_seats = {}
+        self._unsaved_stock_records = {}
+
+        try:
+            self.course = Course.objects.get(id=self.course_id)
+        except Course.DoesNotExist:
+            self.course = Course(id=self.course_id)
+
+    @transaction.atomic
+    def save(self):
+        self.course.save()
+
+        self.parent_seat.course = self.course
+        self.parent_seat.save()
+        category = Category.objects.get(name='Seats')
+        ProductCategory.objects.get_or_create(category=category, product=self.parent_seat)
+
+        for product in self.child_seats.values():
+            product.parent = self.parent_seat
+            product.course = self.course
+            product.save()
+
+        for seat_type, stock_record in self._unsaved_stock_records.iteritems():
+            stock_record.product = self.child_seats[seat_type]
+            stock_record.save()
+
+        self._unsaved_stock_records = {}
+
+    def load_from_lms(self):
+        """
+        Loads course products from the LMS.
+
+        Loaded data is NOT persisted until the save() method is called.
+        """
+        name, modes = self._retrieve_data_from_lms()
+        self.course.name = name
+        self._get_products(name, modes)
+
+    def _retrieve_data_from_lms(self):
+        """
+        Retrieves the course name and modes from the LMS.
+        """
+        headers = {
+            'Accept': 'application/json',
+            'X-Edx-Api-Key': settings.EDX_API_KEY
+        }
+
+        # Get course name from Course Structure API
+        url = urljoin(settings.LMS_URL_ROOT, 'api/course_structure/v0/courses/{}/'.format(self.course_id))
+        response = requests.get(url, headers=headers)
+        data = response.json()
+        logger.debug(data)
+        course_name = data['name']
+
+        # TODO Handle non-200 responses and other errors
+
+        # Get modes and pricing from Enrollment API
+        url = urljoin(settings.LMS_URL_ROOT, 'api/enrollment/v1/course/{}'.format(self.course_id))
+        response = requests.get(url, headers=headers)
+        data = response.json()
+        logger.debug(data)
+        modes = data['course_modes']
+
+        # TODO Handle non-200 responses and other errors
+
+        return course_name, modes
+
+    def _get_product_name(self, course_name, mode):
+        name = u'Seat in {course_name} with {certificate_type} certificate'.format(
+            course_name=course_name,
+            certificate_type=Course.certificate_type_for_mode(mode))
+
+        if Course.is_mode_verified(mode):
+            name += u' (and ID verification)'
+
+        return name
+
+    def _get_products(self, course_name, modes):
+        """
+        Creates course seat products.
+
+        Returns:
+            seats (dict):  Mapping of seat types to seat Products
+            stock_records (dict):  Mapping of seat types to StockRecords
+        """
+
+        course_id = self.course_id
+
+        seats = {}
+        stock_records = {}
+        slug = u'parent-cs-{}'.format(slugify(course_id))
+        partner = Partner.objects.get(code='edx')
+
+        try:
+            parent = Product.objects.get(slug=slug)
+            logger.info(u'Retrieved parent seat product for [%s] from database.', course_id)
+        except Product.DoesNotExist:
+            product_class = ProductClass.objects.get(slug='seat')
+            parent = Product(slug=slug, is_discountable=True, structure=Product.PARENT, product_class=product_class)
+            logger.info(u'Parent seat product for [%s] does not exist. Instantiated a new instance.', course_id)
+
+        parent.title = u'Seat in {}'.format(course_name)
+        parent.attr.course_key = course_id
+
+        # Create the child products
+        for mode in modes:
+            seat_type = mode['slug']
+            slug = u'child-cs-{}-{}'.format(seat_type, slugify(course_id))
+            try:
+                seat = Product.objects.get(slug=slug)
+                logger.info(u'Retrieved [%s] course seat child product for [%s] from database.', seat_type, course_id)
+            except Product.DoesNotExist:
+                seat = Product(slug=slug)
+                logger.info(u'[%s] course seat product for [%s] does not exist. Instantiated a new instance.',
+                            seat_type, course_id)
+
+            seat.parent = parent
+            seat.is_discountable = True
+            seat.structure = Product.CHILD
+            seat.title = self._get_product_name(course_name, seat_type)
+            seat.attr.certificate_type = seat_type
+            seat.attr.course_key = course_id
+            seat.attr.id_verification_required = Course.is_mode_verified(seat_type)
+
+            seats[seat_type] = seat
+
+            try:
+                stock_record = StockRecord.objects.get(product=seat, partner=partner)
+                logger.info(u'Retrieved [%s] course seat child product stock record for [%s] from database.',
+                            seat_type, course_id)
+            except StockRecord.DoesNotExist:
+                partner_sku = generate_sku(seat)
+                stock_record = StockRecord(product=seat, partner=partner, partner_sku=partner_sku)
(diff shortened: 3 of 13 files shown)


ecommerce/extensions/catalogue/models.py
```diff
@@ -1 +1,204 @@
...
+        super(MigratedCourse, self).save()
+
+        for seat_type, stock_record in self._unpersisted_stock_records.iteritems():
+            stock_record.product = self.seats[seat_type]
+            stock_record.save()
+
+        self._unpersisted_stock_records = {}
+
+    @classmethod
+    def load_from_lms(cls, course_key):
```
rlucioni: For clarity and consistency with the way this method is used, the argument name here should be `course_id`. Course keys are typically objects, whereas course IDs are the strings we're used to passing around.

ecommerce/extensions/catalogue/models.py
```diff
@@ -1 +1,204 @@
...
+
+    @classmethod
+    def _retrieve_data_from_lms(cls, course_key):
+        """
+        Retrieves the course name and modes from the LMS.
+        """
+        headers = {'X-Edx-Api-Key': settings.EDX_API_KEY}
+
+        # Ensure course_id is unicode to avoid issues with slugify.
+        course_key = unicode(course_key)
```
rlucioni: Unicode conversion is already performed in `load_from_lms`. It doesn't hurt to do it again here, but I'm not sure you need it.

ecommerce/extensions/catalogue/models.py
```diff
@@ -1 +1,204 @@
...
+
+        # Get course name from Course Structure API
+        url = get_lms_url('api/course_structure/v0/courses/{}/'.format(course_key))
+        response = requests.get(url, headers=headers)
+        response = response.json()
```
rlucioni: Minor: For the sake of readability, I'd avoid overwriting the value of response, instead making this:

``` python
data = response.json()
```

ecommerce/extensions/catalogue/models.py
```diff
@@ -1 +1,204 @@
...
+        course_name = response['name']
+
+        # TODO Handle non-200 responses and other errors
+
+        # Get modes and pricing from Enrollment API
+        url = get_lms_url('api/enrollment/v1/course/{}'.format(course_key))
+        response = requests.get(url, headers=headers).json()
```
rlucioni: Minor: For the sake of readability, I'd separate the response from the response data, making this:

``` python
response = requests.get(url, headers=headers)
data = response.json()
```

ecommerce/extensions/catalogue/models.py
```diff
@@ -1 +1,204 @@
...
+        return u'{}-{}'.format(partner.code, sha).upper()
+
+    @classmethod
+    def _get_products(cls, course_id, course_name, modes):
+        """
+        Creates course seat products.
+
+        Returns
+            seats (dict)            --  Mapping of seat types to seat Products
+            stock_records (dict)    --  Mapping of seat types to StockRecords
```
rlucioni: Minor: Please stick to our [Google-style](http://google-styleguide.googlecode.com/svn/trunk/pyguide.html#Comments) (Napoleon Sphinx) format for docstrings.

rlucioni: Also, I think it's worth documenting the arguments this method accepts, in addition to the values it returns.

ecommerce/extensions/catalogue/models.py
```diff
@@ -1 +1,204 @@
...
+                logger.info(u'Retrieved [%s] course seat child product for [%s] from database.', seat_type, course_id)
+            except Product.DoesNotExist:
+                seat = Product(slug=slug)
+                logger.info(u'[%s] course seat product for [%s] does not exist. Instantiated a new instance.',
+                            seat_type, course_id)
+
+            seat.parent = parent
+            seat.is_discountable = True
+            seat.structure = 'child'
+            seat.title = u'{} Seat in {}'.format(seat_type.capitalize(), course_name)
```
rlucioni: This will give us products with names we may not want users to see, such as "No-id-professional Seat in Course Name."

ecommerce/extensions/catalogue/models.py
jimabramson: for even more clarity let's call it `course_key_string`.

rlucioni: Whatever name is chosen, it should be consistent throughout this module. I see `course_id` and `course_key` used interchangeably, and I wouldn't want to see a third variant of the name.

ecommerce/extensions/catalogue/models.py
```diff
@@ -1 +1,204 @@
...
+        Generates a SKU for the given partner and and course seat product.
+
+        Generated SKU is of the format [Partner Code]-[SHA1 Hash].
+
+        Example: EDX-76E4E71
+        """
+        sha = u' '.join((seat.attr.certificate_type, seat.attr.course_key))
+        sha = sha1(sha)
+        sha = sha.hexdigest()[:7]
+        return u'{}-{}'.format(partner.code, sha).upper()
```
rlucioni: 1) I like the direction this has moved since our discussion last night. However, I'm uneasy about using the human-readable partner code as a prefix, for the same reasons I raised in our discussion. If my understanding is correct, you've included this to ensure SKU uniqueness across different partners. Could you instead include this in the string you pass to the hashing function?

2) Truncating the hash increases the probability of a collision. To avoid issues in the unlikely event of a collision, it might be worth checking to make sure that the generated hash isn't already present in the system.

ecommerce/extensions/catalogue/models.py
```diff
@@ -1 +1,204 @@
...
+logger = logging.getLogger(__name__)
+Partner = get_model('partner', 'Partner')
+StockRecord = get_model('partner', 'StockRecord')
+
+
+def get_lms_url(path):
+    return urljoin(settings.LMS_URL_ROOT, path)
+
+
+class Course(object):
```
jimabramson: why not make this a proper Model?

ecommerce/extensions/catalogue/models.py
```diff
@@ -1 +1,204 @@
...
+
+        Arguments
+            key (string) -- Course key
+
+        Raises
+            DoesNotExistError, if no course-related products are found in the database.
+        """
+        # TODO Filter out parent products, by ensuring certificate_type is not blank (empty string or null)
+        products = Product.objects.filter(attribute_values__attribute__code='course_key',
+                                          attribute_values__value_text='edX/DemoX/Demo_Course').filter(
```
jimabramson: assuming the demo course reference is temporary...

clintonb:
@jimabramson @rlucioni this is ready for final review.

ecommerce/courses/tests.py
```diff
@@ -0,0 +1,43 @@
...
+        course = Course.objects.create(id=course_id)
+        self.assertEqual(unicode(course), course_id)
+
+    def test_seat_products(self):
+        """
+        Verify the method returns a list containing purchasable course seats.
+
+        These seats should be the child products.
+        """
+        # Create a new course and verify it has not existing products.
```
rlucioni: s/not/no

ecommerce/courses/models.py
```diff
@@ -0,0 +1,35 @@
+import logging
+
+from django.db import models
+from oscar.core.loading import get_model
+
+logger = logging.getLogger(__name__)
+Partner = get_model('partner', 'Partner')
+Product = get_model('catalogue', 'Product')
+ProductClass = get_model('catalogue', 'ProductClass')
+StockRecord = get_model('partner', 'StockRecord')
```
rlucioni: It looks like `ProductClass` is the only one of these models which is actually used in this module.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -0,0 +1,214 @@
...
+        return course_name, modes
+
+    def _get_product_name(self, course_name, certificate_type):
+        # TODO How to handle "no-id-professional"?
+        name = u'Seat in {course_name} with {certificate_type} certificate'.format(
+            course_name=course_name,
+            certificate_type=certificate_type.lower())
+
+        if certificate_type in VERIFIED_MODES:
+            name += u' (with ID verification)'
```
rlucioni: Do we want users to see products with names such as "Seat in Programming with C# with verified certificate with ID verification" or "Seat in Marketing for Non-Marketers with no-id-professional certificate"? I know these are nitpicks, but this will eventually become user-facing copy. Maybe it's worth speaking with Docs about structuring these names?

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -0,0 +1,214 @@
...
+    def _get_products(self, course_name, modes):
+        """
+        Creates course seat products.
+
+        Returns:
+            seats (dict)            --  Mapping of seat types to seat Products
+            stock_records (dict)    --  Mapping of seat types to StockRecords
+        """
+
+        # Ensure this value is Unicode to avoid issues with slugify.
```
rlucioni: Move this comment up to the Unicode conversion performed in `__init__`.

clintonb: Done.

ecommerce/courses/models.py
clintonb: Updated.

ecommerce/courses/tests.py
clintonb: Updated.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -0,0 +1,214 @@
...
+            seat.attr.id_verification_required = (seat_type in VERIFIED_MODES)
+
+            seats[seat_type] = seat
+
+            try:
+                stock_record = StockRecord.objects.get(product=seat, partner=partner)
+                logger.info(u'Retrieved [%s] course seat child product stock record for [%s] from database.',
+                            seat_type, course_id)
+            except StockRecord.DoesNotExist:
+                partner_sku = generate_sku(partner, seat)
```
rlucioni: Minor: In its current state, this won't reassign SKUs to products which already have SKUs. However, for the sake of consistency across products, what do you think of moving this outside of the exception handler and setting the SKU like currency and price?

ecommerce/extensions/catalogue/management/commands/migrate_course.py
clintonb: Updated.

ecommerce/extensions/catalogue/utils.py
```diff
@@ -0,0 +1,17 @@
...
+    Generated SKU is of the format [Partner Code]-[Hash].
+
+    Example: EDX-76E4E71
+    """
+    # Note: This currently supports seats. In the future, this should
+    # be updated to accommodate other product classes.
+    _hash = u' '.join((product.attr.certificate_type.lower(), product.attr.course_key.lower()))
+    _hash = md5(_hash)
+    _hash = _hash.hexdigest()[-7:]
+    return u'{}-{}'.format(partner.code, _hash).upper()
```
rlucioni: I'm still not convinced about prefixing the partner code. I'd be happy to talk about it offline, if you'd like.

ecommerce/extensions/catalogue/tests/test_migrate_course.py
```diff
@@ -0,0 +1,162 @@
+import json
+import logging
+from urlparse import urljoin
+
+from django.conf import settings
+from django.core.management import call_command
+from django.test import TestCase
+import httpretty
+
```
rlucioni: Nitpick: unnecessary newline.

ecommerce/extensions/catalogue/tests/test_migrate_course.py
```diff
@@ -0,0 +1,162 @@
...
+        self.assertEqual(seat.title, expected_title)
+        self.assertEqual(seat.attr.certificate_type, seat_type)
+        self.assertEqual(seat.attr.course_key, self.course_id)
+        self.assertEqual(seat.attr.id_verification_required, (seat_type in VERIFIED_MODES))
+
+    def assert_course_migrated(self):
+        """ Verify the course was migrated and saved to the database. """
+        course = Course.objects.get(id=self.course_id)
+        seats = course.seat_products
+        self.assertEqual(len(seats), 2)
```
rlucioni: This assumes that the test course will have honor and verified modes. More generally, what about testing our other mode configurations (e.g., honor-only, professional, no-id-professional)?

rlucioni:
In its current state, the management command will create complete products within the ecommerce service, but doesn't set its generated SKUs on the LMS. Manually keeping SKUs in sync across the LMS and ecommerce is going to get out of hand quickly for ≈1000 products. Will we extend the enrollment API or create a new API to allow this management command (and the course admin tool) to set SKUs?

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -0,0 +1,214 @@
...
+        self.child_seats = {}
+        self._unsaved_stock_records = {}
+
+        try:
+            self.course = Course.objects.get(id=self.course_id)
+        except Course.DoesNotExist:
+            self.course = Course(id=self.course_id)
+
+    def save(self):
+        # TODO Use a transaction
```
rlucioni: Any reason not to take care of this now by tacking on `@transaction.atomic`?

clintonb:
LMS synchronization is covered by XCOM-312.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
clintonb: If an existing product has a SKU we should avoid modifying it in case that SKU has been published elsewhere (e.g. LMS).

ecommerce/extensions/catalogue/management/commands/migrate_course.py
clintonb: Updated.

ecommerce/courses/models.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return mode
+
+    @property
+    def seat_products(self):
+        """ Returns a list of course seat Products related to this course. """
+        seat_product_class = ProductClass.objects.get(slug='seat')
+        products = set()
+
+        for product in self.products.all():
```
jimabramson: `self.products` doesn't seem to be defined

jimabramson: d'oh never mind.

ecommerce/courses/models.py
```diff
@@ -0,0 +1,44 @@
...
+            if product.get_product_class() == seat_product_class:
+                if product.is_parent:
+                    products.update(product.children.all())
+                else:
+                    products.add(product)
+
+        return list(products)
+
+    def __unicode__(self):
```
jimabramson: i find this kind of thing can be confusing / misleading when debugging.  is there a specific need for it?

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -0,0 +1,214 @@
...
+        self.course_id = unicode(course_id)
+
+        self.parent_seat = None
+        self.child_seats = {}
+        self._unsaved_stock_records = {}
+
+        try:
+            self.course = Course.objects.get(id=self.course_id)
+        except Course.DoesNotExist:
+            self.course = Course(id=self.course_id)
```
jimabramson, line 37: why not just `get_or_create` ?

ecommerce/courses/models.py
clintonb: To avoid confusion when debugging. Actually, this might not be necessary since Django models print the `id` by default.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
clintonb, line 37: That will immediately persist the data to the database. There was a request for a dry-run option (which I call commit, the inverse). Data is only saved if the commit flag is set on the command line.

jimabramson, line 37: understood

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -0,0 +1,214 @@
...
+
+        try:
+            parent = Product.objects.get(slug=slug)
+            logger.info(u'Retrieved parent seat product for [%s] from database.', course_id)
+        except Product.DoesNotExist:
+            product_class = ProductClass.objects.get(slug='seat')
+            parent = Product(slug=slug, is_discountable=True, structure=Product.PARENT, product_class=product_class)
+            logger.info(u'Parent seat product for [%s] does not exist. Instantiated a new instance.', course_id)
+
+        parent.title = u'Seat in {}'.format(course_name)
```
jimabramson, line 132: prepending "Seat in " to all the product titles seems like unnecessary noise.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -0,0 +1,214 @@
...
+        data = response.json()
+        logger.debug(data)
+        modes = data['course_modes']
+
+        # TODO Handle non-200 responses and other errors
+
+        return course_name, modes
+
+    def _get_product_name(self, course_name, mode):
+        name = u'Seat in {course_name} with {certificate_type} certificate'.format(
```
jimabramson, line 99: (as commented elsewhere) "Seat in " seems superfluous

ecommerce/extensions/catalogue/management/commands/migrate_course.py
clintonb, line 132: Any suggestions for a better naming scheme?

rlucioni, line 132: Simply drop "Seat in" and begin with the course title.

jimabramson, line 132: I think it's fine as you have it, except that the opening 'Seat in ' part feels redundant to me, since the products have a 'seat' class / category and there's no other kind of product right now anyway.

clintonb, line 132: Keep in mind the title is intended to be displayed on the receipt and in other user-visible areas. `Seat in French Calculus with honor certificate` more accurately describes the product than `French Calculus with honor certificate`. It also reads better.

jimabramson, line 132: That's a good point.  I was mainly thinking about internal-facing stuff, e.g. Google Analytics.  I'm comfortable leaving it the way it is, given that perspective.

rlucioni, line 132: I was forgetting that we'll have products other than Seats in the future. Given that, I'm also comfortable leaving it as it is.

clintonb:
@rlucioni @jimabramson Any other feedback? Approvals?

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -0,0 +1,214 @@
...
+
+        return name
+
+    def _get_products(self, course_name, modes):
+        """
+        Creates course seat products.
+
+        Returns:
+            seats (dict)            --  Mapping of seat types to seat Products
+            stock_records (dict)    --  Mapping of seat types to StockRecords
```
rlucioni: Nitpick: `--` should be replaced with colons to align with Google-style.

clintonb: Done.

ecommerce/extensions/catalogue/utils.py
rlucioni: @clintonb and I discussed this offline. I'm still not sold that the partner code prefix is necessary - I've been using Amazon's [ASIN](http://en.wikipedia.org/wiki/Amazon_Standard_Identification_Number) as my reference for how our SKUs should appear - but this isn't important enough to keep arguing about. Let the record show that I was against the partner code prefix.

rlucioni:
I'm still not cool with the partner code SKU prefix, but :+1: despite that.

ecommerce/extensions/catalogue/utils.py
rlucioni: I kept thinking about this on the way home today and want to raise two more things to consider here. Bear with me.
1. The partner code prefix makes these SKUs look very similar to our order numbers (i.e., `EDX-{string}`). This has the potential to confuse users (e.g., if the SKU were to appear in an email along with an order number) and future maintainers (who may not understand what motivated the decision to include the partner code in the SKU).
2. [Yagni](http://martinfowler.com/bliki/Yagni.html). That we _might_ want to include the SKU in an email _someday_ really isn't a great reason to introduce unnecessary complexity to our codebase, however small.

I'd love to see the SKU be nothing more than a 10-character hash of partner code, course key, and certificate type. Simple, distinct from order numbers, and yielding slightly cleaner code.

PR #121, opened 2015-06-03
Add Refund class method for creating targeted Refunds

rlucioni:
Given an order and order lines, creates a Refund with corresponding RefundLines. Only creates RefundLines for unrefunded order lines. Meets the requirements of [XCOM-356](https://openedx.atlassian.net/browse/XCOM-356).

### DIFF
--- a/ecommerce/extensions/refund/models.py
+++ b/ecommerce/extensions/refund/models.py
@@ -14,6 +14,7 @@ from ecommerce.extensions.payment.helpers import get_processor_class_by_name
 from ecommerce.extensions.refund.exceptions import InvalidStatus
 from ecommerce.extensions.refund.status import REFUND, REFUND_LINE
 
+
 logger = logging.getLogger(__name__)
 
 post_refund = get_class('refund.signals', 'post_refund')
@@ -29,13 +30,12 @@ class StatusMixin(object):
         return getattr(settings, self.pipeline_setting)
 
     def available_statuses(self):
-        """ Returns all possible statuses that this object can move to. """
+        """Returns all possible statuses that this object can move to."""
         return self.pipeline.get(self.status, ())
 
     # pylint: disable=access-member-before-definition,attribute-defined-outside-init
     def set_status(self, new_status):
-        """
-        Set a new status for this object.
+        """Set a new status for this object.
 
         If the requested status is not valid, then ``InvalidStatus`` is raised.
         """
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/refund/models.py
```diff
@@ -68,12 +68,49 @@ class Refund(StatusMixin, TimeStampedModel):
...
+
+        Arguments:
+            order (order.Order): The order to which the newly-created refund corresponds.
+            lines (list of order.Line): Order lines to be refunded.
+
+        Returns:
+            None: If no unrefunded order lines have been provided.
+            Refund: With RefundLines corresponding to each given unrefunded order line.
+        """
+        unrefunded_lines = [line for line in lines if line.refund_lines.count() == 0]
```
clintonb: Use `line.refund_lines.exists()` to avoid unnecessary loading of data.

ecommerce/extensions/refund/models.py
```diff
@@ -68,12 +68,49 @@ class Refund(StatusMixin, TimeStampedModel):
...
+        if unrefunded_lines:
```
clintonb, line 89: This looks a lot like the body of `refund.api.create_refunds`. That function should be updated to use this method.

ecommerce/extensions/refund/tests/test_models.py
```diff
@@ -72,6 +73,28 @@ def test_all_statuses(self):
         """ Refund.all_statuses should return all possible statuses for a refund. """
         self.assertEqual(Refund.all_statuses(), self.pipeline.keys())
 
+    def test_create_with_lines(self):
+        """
+        Given an order and order lines that have not been refunded, Refund.create_with_lines
+        should create a Refund with corresponding RefundLines.
+        """
+        order = self.create_order(user=UserFactory())
+        refund = Refund.create_with_lines(order, [order.lines.first()])
```
clintonb: What about multiple lines?

ecommerce/extensions/refund/models.py
rlucioni: [`count()`](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.count) doesn't load data either, but I agree that `exists()` is more concise. Thanks for the suggestion.

ecommerce/extensions/refund/models.py
rlucioni, line 89: Nice call, I missed this. Updated.

ecommerce/extensions/refund/tests/test_models.py
rlucioni: Updated.

ecommerce/extensions/refund/tests/test_models.py
```diff
@@ -72,6 +73,29 @@ def test_all_statuses(self):
...
         self.assertEqual(Refund.all_statuses(), self.pipeline.keys())
 
+    @ddt.data(False, True)
+    def test_create_with_lines(self, multiple_lines):
+        """
+        Given an order and order lines that have not been refunded, Refund.create_with_lines
+        should create a Refund with corresponding RefundLines.
+        """
+        order = self.create_order(user=UserFactory(), multiple_lines=multiple_lines)
+        refund = Refund.create_with_lines(order, [line for line in order.lines.all()])
```
clintonb: `[order.lines.all()]` should suffice. No need for the loop.

clintonb:
:+1:

ecommerce/extensions/refund/tests/test_models.py
rlucioni: Thanks, good catch; `order.lines.all()` actually suffices; `create_with_lines` doesn't know what to do with a list containing a QuerySet, which doesn't make much sense, anyway.

ecommerce/extensions/refund/tests/test_models.py
```diff
@@ -72,6 +73,29 @@ def test_all_statuses(self):
...
         self.assertEqual(Refund.all_statuses(), self.pipeline.keys())
 
+    @ddt.data(False, True)
+    def test_create_with_lines(self, multiple_lines):
+        """
+        Given an order and order lines that have not been refunded, Refund.create_with_lines
+        should create a Refund with corresponding RefundLines.
+        """
+        order = self.create_order(user=UserFactory(), multiple_lines=multiple_lines)
+        refund = Refund.create_with_lines(order, order.lines.all())
```
clintonb: Ah. I meant `list(order.lines.all())`.

PR #122, opened 2015-06-03
Updated Order Detail view to enable refund creation

clintonb:
Users can now select line items to refund from the dashboard's order detail view.

XCOM-357

@rlucioni @jimabramson

### DIFF
--- a/ecommerce/extensions/dashboard/orders/tests.py
+++ b/ecommerce/extensions/dashboard/orders/tests.py
@@ -1,0 +1,89 @@
+from django.contrib.messages import constants as MSG
+from django.core.urlresolvers import reverse
+from django.test import TestCase
+from oscar.core.loading import get_model
+
+from ecommerce.extensions.refund.tests.mixins import RefundTestMixin
+from ecommerce.tests.mixins import UserMixin
+
+Refund = get_model('refund', 'Refund')
+
+
+class OrderDetailViewTests(UserMixin, RefundTestMixin, TestCase):
+    def setUp(self):
+        super(OrderDetailViewTests, self).setUp()
+
+        # Staff permissions are required for this view
+        self.user = self.create_user(is_staff=True)
+
+    def assert_message_equals(self, response, msg, level):  # pylint: disable=unused-argument
+        """ Verify the latest message matches the expected value. """
+        messages = []
+        for context in response.context:
+            messages += context.get('messages', [])
+
+        message = messages[0]
+        self.assertEqual(message.level, level)
+        self.assertEqual(message.message, msg)
+
+    def _request_refund(self, order):
+        """POST to the view."""
+
+        # Login
+        self.client.login(username=self.user.username, password=self.password)
+
+        # POST to the view, creating the Refund
+        data = {
+            'line_action': 'create_refund',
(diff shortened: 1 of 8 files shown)


ecommerce/extensions/dashboard/orders/views.py
```diff
@@ -0,0 +1,30 @@
+from django.contrib import messages
+from django.core.urlresolvers import reverse
+from django.utils.safestring import mark_safe
+from django.utils.translation import ugettext_lazy as _
+from oscar.apps.dashboard.orders.views import OrderDetailView as CoreOrderDetailView
```
jimabramson, line 5: trivial: `BaseOrderDetailView` ?

ecommerce/extensions/dashboard/orders/views.py
```diff
@@ -0,0 +1,30 @@
...
+        refund = Refund.create_with_lines(order, lines)
+
+        if refund:
+            data = {
+                'link_start': '<a href="{}" target="_blank">'.format(
+                    reverse('dashboard:refunds:detail', kwargs={'pk': refund.pk})),
+                'link_end': '</a>',
+                'refund_id': refund.pk
+            }
+            message = _('%(link_start)sRefund #%(refund_id)d%(link_end)s created!') % data
```
jimabramson: That I could click on the confirmation message and go to the page for the new refund was not at all obvious (though now that I know, I know).  Did you consider 302-ing directly to that page upon success?  That would also get rid of the need to cobble together so much markup here.

jimabramson: alternatively, spell it out in the message, e.g. "Refund created!  Click here to view or process the refund."

ecommerce/extensions/dashboard/orders/tests.py
```diff
@@ -0,0 +1,82 @@
...
+
+Refund = get_model('refund', 'Refund')
+
+
+class OrderDetailViewTests(UserMixin, RefundTestMixin, TestCase):
+    def setUp(self):
+        super(OrderDetailViewTests, self).setUp()
+
+        # Staff permissions are required for this view
+        self.user = self.create_user(is_staff=True)
```
jimabramson, line 17: is blocking access to this view for non-staff already covered by tests elsewhere?

ecommerce/extensions/dashboard/orders/views.py
rlucioni: Minor: I would also encourage the use of `format()` here. As a translator, I'd be more likely to know that `{refund_id}` is a placeholder than `%(refund_id)d`.

rlucioni:
Looks good. My minor comment aside, :+1: .

ecommerce/extensions/dashboard/orders/views.py
clintonb: I've addressed both of your comments.

ecommerce/extensions/dashboard/orders/tests.py
clintonb, line 17: This should be covered by tests in Oscar.

jimabramson:
:+1:

PR #123, opened 2015-06-03
Pass mode when revoking course seat fulfillment.

jimabramson:
XCOM-388

This change solves HTTP 400 errors when attempting to revoke a course seat in a course that has no honor mode configured.

@rlucioni @clintonb please review.


clintonb:
:+1:

rlucioni:
:+1:

PR #124, opened 2015-06-04
Boosted Code Coverage

clintonb:
- Ignoring files that cannot be covered
- Removed dead code

XCOM-193

@rlucioni @jimabramson


ecommerce/extensions/order/models.py
```diff
@@ -15,34 +15,6 @@ def can_retry_fulfillment(self):
         """ Returns a boolean indicating if order is eligible to retry fulfillment. """
         return self.status == ORDER.FULFILLMENT_ERROR
 
-    @classmethod
-    def check_order_total(cls, order_num, auth_amount, auth_currency):
```
rlucioni, line 19: Thanks for killing this.

PR #126, opened 2015-06-05
Recover from 400 on enrollment revocation when due to mode mismatch.

jimabramson:
XCOM-396

related to https://github.com/edx/edx-platform/pull/8379 - this avoids an error in otto if we approve a refund request for one course mode but the student has already re-enrolled in another course mode.

@rlucioni @clintonb please have a look

### DIFF
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -216,9 +216,16 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
                 logger.info('Successfully revoked line [%d].', line.id)
                 return True
             else:
+                # check if the error / message are something we can recover from.
                 data = response.json()
                 detail = data.get('message', '(No details provided.)')
-                logger.error('Failed to revoke fulfillment of Line [%d]: %s', line.id, detail)
+                if response.status_code == 400 and "Enrollment mode mismatch" in detail:
+                    # The user is currently enrolled in different mode than the one
+                    # we are refunding an order for.  Don't revoke that enrollment.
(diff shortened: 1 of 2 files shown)


clintonb:
Tests are needed. Otherwise, this is fine.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -215,6 +215,11 @@ def revoke_line(self, line):
             if response.status_code == status.HTTP_200_OK:
                 logger.info('Successfully revoked line [%d].', line.id)
                 return True
+            elif response.status_code == 400 and "Enrollment mode mismatch" in response.json().get('message', ''):
+                # The user is currently enrolled in different mode than the one
+                # we are refunding an order for.  Don't revoke that enrollment.
+                logger.warning('Skipping deactivation: %s', response.json()['message'])
```
rlucioni: This log message could benefit from being a little more explicit: `'Skipping enrollment deactivation: '`

rlucioni: Even better would be something more general than enrollments and consistent with the INFO-level message above:

``` python
logger.warning('Skipping revocation for line [%d].', line.id)
```

jimabramson:
updated: changed log message, slight refactor, added test.

rlucioni:
:+1:

clintonb:
:+1:

jimabramson:
test this please

jimabramson:
not sure why we didn't get the call back from coveralls, but the check passed (0% change in coverage) - https://coveralls.io/builds/2758565

merging.

PR #127, opened 2015-06-08
Automatically approve refunds corresponding to a total credit of $0

rlucioni:
Also prevents completed and denied refunds from appearing in the default refund list view, and allows filtering of refunds by multiple statuses.

@jimabramson or @clintonb, could you please review this when you're able?

### DIFF
--- a/ecommerce/extensions/dashboard/refunds/forms.py
+++ b/ecommerce/extensions/dashboard/refunds/forms.py
@@ -2,10 +2,20 @@ from django import forms
 from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_model
 
+from ecommerce.extensions.refund.status import REFUND
+
+
 Refund = get_model('refund', 'Refund')
 
 
 class RefundSearchForm(forms.Form):
     id = forms.IntegerField(required=False, label=_('Refund ID'))
-    status_choices = (('', '---------'),) + tuple([(status, status) for status in Refund.all_statuses()])
-    status = forms.ChoiceField(choices=status_choices, label=_("Status"), required=False)
+    status_choices = tuple([(status, status) for status in Refund.all_statuses()])
+    status = forms.MultipleChoiceField(choices=status_choices, label=_("Status"), required=False)
+
+    def clean(self):
+        cleaned_data = super(RefundSearchForm, self).clean()
+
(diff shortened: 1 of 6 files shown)


ecommerce/extensions/refund/models.py
```diff
@@ -128,11 +132,11 @@ def can_deny(self):
 
     def _issue_credit(self):
         """Issue a credit to the purchaser via the payment processor used for the original order."""
-
-        # TODO Update this if we ever support multiple payment sources for a single order.
-        source = self.order.sources.first()
-        processor = get_processor_class_by_name(source.source_type.name)()
-        processor.issue_credit(source, self.total_credit_excl_tax, self.currency)
+        if self.total_credit_excl_tax > 0:
```
clintonb: If we ever (for some odd reason) decide to process $0 orders, we should also refund them. I am also leery of injecting too much business logic into the model. Instead of the if statement, let this code run, and account for the (most-likely) case where the order has no sources:

``` python
try:
  source = self.order.sources.first()
  processor = get_processor_class_by_name(source.source_type.name)()
  processor.issue_credit(source, self.total_credit_excl_tax, self.currency)
except Source.DoesNotExist:
  log.info('No payments to refund for order/refund...')
```

ecommerce/extensions/dashboard/refunds/forms.py
```diff
@@ -2,10 +2,20 @@
...
 
+from ecommerce.extensions.refund.status import REFUND
+
+
 Refund = get_model('refund', 'Refund')
 
 
 class RefundSearchForm(forms.Form):
     id = forms.IntegerField(required=False, label=_('Refund ID'))
     status_choices = (('', '---------'),) + tuple([(status, status) for status in Refund.all_statuses()])
```
clintonb: Remove the default blank status; otherwise, it will be selectable in the UI.

clintonb:
:+1: (once comments are addressed)

ecommerce/extensions/dashboard/refunds/forms.py
rlucioni: Updated.

ecommerce/extensions/refund/models.py
rlucioni: Updated.

PR #129, opened 2015-06-09
Tracking Product Change History

clintonb:
All changes to the StockRecord, Product and ProductAttributeValue models are now recorded in the database.

XCOM-401

@jimabramson @rlucioni

### DIFF
--- a/ecommerce/extensions/catalogue/admin.py
+++ b/ecommerce/extensions/catalogue/admin.py
@@ -1,10 +1,20 @@
 from oscar.apps.catalogue.admin import *  # pylint: disable=unused-import,wildcard-import,unused-wildcard-import
+from simple_history.admin import SimpleHistoryAdmin
 
 
-class ProductAdminExtended(ProductAdmin):
-    list_display = ('get_title', 'upc', 'get_product_class', 'structure',
-                    'attribute_summary', 'date_created', 'course')
+class ProductAdminExtended(SimpleHistoryAdmin):
+    list_display = ('get_title', 'upc', 'get_product_class', 'structure', 'attribute_summary', 'date_created', 'course')
+    prepopulated_fields = {"slug": ("title",)}
+    inlines = [AttributeInline, CategoryInline, ProductRecommendationInline]
 
 
 admin.site.unregister(Product)
 admin.site.register(Product, ProductAdminExtended)
+
+
+class ProductAttributeValueAdminExtended(SimpleHistoryAdmin):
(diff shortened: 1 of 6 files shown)


ecommerce/extensions/catalogue/models.py
```diff
@@ -1,10 +1,34 @@
...
+    def _history_user(self, value):  # pragma: no cover
+        self.changed_by = value
+
+
+class ProductAttributeValue(AbstractProductAttributeValue):
+    changed_by = models.ForeignKey('user.User', null=True, blank=True)
+    history = HistoricalRecords()
+
+    @property
+    def _history_user(self):  # pragma: no cover
```
rlucioni, line 26: Why is this property necessary?

rlucioni, line 26: Never mind, I answered my own question. ([Reference](https://django-simple-history.readthedocs.org/en/latest/advanced.html#recording-which-user-changed-a-model))

rlucioni:
:+1:

jimabramson:
@clintonb the migration is missing for stockrecord.

ecommerce/extensions/catalogue/admin.py
```diff
@@ -1,10 +1,20 @@
...
+from simple_history.admin import SimpleHistoryAdmin
 
 
-class ProductAdminExtended(ProductAdmin):
-    list_display = ('get_title', 'upc', 'get_product_class', 'structure',
-                    'attribute_summary', 'date_created', 'course')
+class ProductAdminExtended(SimpleHistoryAdmin):
+    list_display = ('get_title', 'upc', 'get_product_class', 'structure', 'attribute_summary', 'date_created', 'course')
+    prepopulated_fields = {"slug": ("title",)}
+    inlines = [AttributeInline, CategoryInline, ProductRecommendationInline]
```
jimabramson, line 8: why is product recommendation in there?

clintonb, line 8: Copied from the original admin.

jimabramson, line 8: ¯_(ツ)_/¯

clintonb, line 8: My only intent is to add history support to the admin, not change/remove any other existing functionality. Although we do not use the ProductRecommendation model currently, I see no need to remove it from the admin UI.

clintonb:
@jimabramson Migration added.

jimabramson:
:+1:

PR #130, opened 2015-06-09
Exposing User full_name field in admin site

clintonb:
User.full_name is now visible and editable in the admin site.

FYI @jimabramson @rlucioni

### DIFF
--- a/ecommerce/user/migrations/0004_auto_20150609_1450.py
+++ b/ecommerce/user/migrations/0004_auto_20150609_1450.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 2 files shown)


ecommerce/user/migrations/0004_auto_20150609_1450.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
```
clintonb, line 7: This is a pointless migration generated by Django. Modifying any kwarg on a field will result in the need for a new migration. In this case the `blank` kwarg was added a while ago, but no migration was generated.

PR #131, opened 2015-06-10
Unenrolling after running acceptance tests

clintonb:
The user is unenrolled after each test is run.

@jimabramson @rlucioni


rlucioni:
:+1:

PR #132, opened 2015-06-10
Don’t let LMS take us down during health checks.

jimabramson:
Inspired by the brief outage we just had.
1. calls to the LMS' health page from Otto's healthcheck should not block on an unresponsive LMS and hang the system.
2. don't remove Otto from the LB (by failing the healthcheck) when the LMS is unresponsive.  Otto should remain available to handle messages from payment processors and perform tasks that do not depend on the LMS.  (We still report LMS status in the "detailed health" output however.)

@feanil @clintonb @rlucioni agree?

### DIFF
--- a/ecommerce/health/views.py
+++ b/ecommerce/health/views.py
@@ -48,7 +48,7 @@ def health(_):
         database_status = Status.UNAVAILABLE
 
     try:
-        response = requests.get(LMS_HEALTH_PAGE)
+        response = requests.get(LMS_HEALTH_PAGE, timeout=1)
 
         if response.status_code == status.HTTP_200_OK:
             lms_status = Status.OK
@@ -59,7 +59,7 @@ def health(_):
         logger.critical(UnavailabilityMessage.LMS)
         lms_status = Status.UNAVAILABLE
 
-    overall_status = Status.OK if (database_status == lms_status == Status.OK) else Status.UNAVAILABLE
+    overall_status = Status.OK if (database_status == Status.OK) else Status.UNAVAILABLE
 
     data = {
         'overall_status': overall_status,
--- a/ecommerce/health/tests/test_views.py
+++ b/ecommerce/health/tests/test_views.py
@@ -1,16 +1,18 @@
 """Tests of the service health endpoint."""
 import json
-import logging
 
+from django.core.urlresolvers import reverse
+from django.db import DatabaseError
+from django.test import TestCase
 import mock
 from requests import Response
 from requests.exceptions import RequestException
 from rest_framework import status
-from django.test import TestCase
-from django.db import DatabaseError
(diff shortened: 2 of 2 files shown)


ecommerce/health/views.py
```diff
@@ -48,7 +48,7 @@ def health(_):
         database_status = Status.UNAVAILABLE
 
     try:
-        response = requests.get(LMS_HEALTH_PAGE)
+        response = requests.get(LMS_HEALTH_PAGE, timeout=1)
```
rlucioni, line 51: We have an `ENROLLMENT_FULFILLMENT_TIMEOUT` setting, set to five seconds by default. I wonder if it's worth generalizing that to `LMS_REQUEST_TIMEOUT` and using it here.

jimabramson, line 51: @rlucioni I would be ok with using the same setting but think 5s is not aggressive enough.

rlucioni:
I agree with this on the grounds that Otto is capable of handling the LMS' unavailability gracefully. Users may still wish to view orders and refunds, create and/or process refunds, and configure products, actions that are either possible without the LMS, or can be easily retried once the LMS recovers.

@jimabramson could you please update the tests?

ecommerce/health/views.py
rlucioni, line 51: @jimabramson in retrospect, I don't think we gain anything by making this configurable, so feel free to ignore me. Is one second too aggressive? What are you basing the timeout on?

jimabramson, line 51: In general we want a heartbeat page to respond as quickly as possible since it will be called constantly and frequently.  The 95th percentile response time of the edx.org heartbeat is around 30ms, and 99th percentile is less than 1s.  If anything 1s seems like it might be too lenient, but in practice it means we could handle getting hit up to 1x/sec while the LMS is unresponsive, without dragging down the system.

rlucioni, line 51: @jimabramson great, thanks. I just wanted to make sure we'd looked at the data before concluding one second was right.

clintonb:
:+1: (once tests updated)

feanil:
@jimabramson generally agree, but there are apis  endpoints that otto should not try to use if the LMS is unhealthy, is there short-circuiting to prevent that right now or will timeouts/errors from the API requests be handled correctly?

rlucioni:
@feanil the `ENROLLMENT_FULFILLMENT_TIMEOUT` setting I mentioned above is used in conjunction with appropriate error handling to recover in case the LMS in unavailable.

feanil:
:+1:  once tests are updated

jimabramson:
fixed tests.  they may not work on first pass, as I seem to be doing something wrong with LogCapture.

PR #134, opened 2015-06-10
Removed changed_by fields from all models

clintonb:
This field unnecessarily duplicates behavior from simple history, and prevents us from using the middleware to get the user making the change.

XCOM-193

@jimabramson @rlucioni

### DIFF
--- a/ecommerce/extensions/catalogue/migrations/0005_auto_20150610_1355.py
+++ b/ecommerce/extensions/catalogue/migrations/0005_auto_20150610_1355.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0004_auto_20150609_0129'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='historicalproduct',
+            name='changed_by',
+        ),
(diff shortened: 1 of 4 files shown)


rlucioni:
The `_history_user` property seems to recommended by simple history's [documentation](https://django-simple-history.readthedocs.org/en/latest/advanced.html#recording-which-user-changed-a-model). For my own edification, could you elaborate on why it's not necessary to record which user is changing the model?

clintonb:
The documentation in question assumes the model already has a field being used to track the user. If the model doesn't have an `_history_user` property, simple-history will fallback to its request middleware, which is what we want. The overwhelming majority of our changes to these models will arrive via a request, so using this middleware is the desired behavior.

rlucioni:
Thank you. :+1:

jimabramson:
@clintonb how does this affect our ability to display in some view, for example, the user who last modified some object or another?

jimabramson:
:+1: to code, apart from my question

clintonb:
We are not currently displaying that information, so technically there is no effect. That being said, any model being tracked has an `history` manager that can be used to locate previous data:

``` python
# Returns a Product object
p = Product.objects.first()

# Returns a ProductHistory object
ph = p.history.latest()

# Returns a User
ph.history_user
```

clintonb:
FYI @mulby @jab5569 @brianhw I stand corrected on that statement about having no plans to remove columns.

clintonb:
:+1:

PR #135, opened 2015-06-10
Added view to run migrate_course management command

clintonb:
This view can be used by superusers to migrate courses without needing to SSH into the machine.

XCOM-193

@rlucioni @jimabramson 

Review #134 first.


clintonb:
FYI @feanil

rlucioni:
Nice. :+1:

ecommerce/courses/views.py
```diff
@@ -0,0 +1,59 @@
...
+
+        course_ids = course_ids.split(',')
+
+        call_command('migrate_course', *course_ids, commit=commit, stdout=out, stderr=err)
+
+        # Format the output for display
+        out.seek(0)
+        err.seek(0)
+        log.seek(0)
+        output = 'STDOUT\n{out}\n\nSTDERR\n{err}\n\nLOG\n{log}'.format(out=out.read(), err=err.read(), log=log.read())
```
jimabramson: can't you just use [getvalue](https://docs.python.org/2/library/stringio.html#StringIO.StringIO.getvalue) ?

clintonb: I can now that I know it exists. Updated.

ecommerce/courses/views.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # Remove the log capture handler
+        root_logger.removeHandler(log_handler)
+
+        return HttpResponse(output, content_type='text/plain')
```
jimabramson, line 57: @clintonb I know there's a hurry to get this out but did you consider using a [StreamingResponse](https://docs.djangoproject.com/en/1.8/ref/request-response/#django.http.StreamingHttpResponse)

jimabramson, line 57: ...in order to immediately stream out the feedback instead of storing it up in memory and then dumping it all at the end

clintonb, line 57: No _major_ hurry, so don't feel obligated to rush the review. I did consider that, and would _love_ it if we could live stream the log! However, `call_command` does not return until the management command is complete. If we end up wanting to adopt this idea of self-service management commands, I definitely want to find a way to stream the output.

jimabramson:
:+1:   let's implement streaming in a subsequent PR.

PR #136, opened 2015-06-10
Updated Course Migration Command and View

clintonb:
- Command now raises an exception and provides more detail when calls to external APIs fail
- View now specifies settings to be used by the command

XCOM-193

FYI @rlucioni @jimabramson

### DIFF
--- a/ecommerce/courses/views.py
+++ b/ecommerce/courses/views.py
@@ -1,5 +1,6 @@
 from StringIO import StringIO
 import logging
+import os
 
 from django.core.management import call_command
 from django.http import Http404, HttpResponse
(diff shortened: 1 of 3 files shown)


rlucioni:
:+1:

ecommerce/courses/views.py
```diff
@@ -38,14 +39,17 @@ def get(self, request, *_args, **_kwargs):
...
+        user = request.user
+        logger.info(msg, user.username, course_ids)
 
         if not course_ids:
             return HttpResponse('No course_ids specified.', status=400)
 
         course_ids = course_ids.split(',')
 
-        call_command('migrate_course', *course_ids, commit=commit, stdout=out, stderr=err)
+        print user.access_token
```
jimabramson, line 50: is that intentional?

clintonb, line 50: Most definitely not.

rlucioni, line 50: Nice save @jimabramson.

PR #137, opened 2015-06-10
Relay order number to payment processors

rlucioni:
We've been sending basket IDs to our payment processors instead of order numbers. The changes proposed in this PR (pursuant to [XCOM-358](https://openedx.atlassian.net/browse/XCOM-358)) address this by using the order number as the invoice or "reference" number.

@jimabramson and @clintonb, could you please review this when you're able?

### DIFF
--- a/ecommerce/extensions/order/tests/test_utils.py
+++ b/ecommerce/extensions/order/tests/test_utils.py
@@ -1,6 +1,5 @@
 """Test Order Utility classes """
-from unittest import TestCase
-from django.test import override_settings
+from django.test import TestCase, override_settings
 from oscar.test.newfactories import BasketFactory
 
 from ecommerce.extensions.order.utils import OrderNumberGenerator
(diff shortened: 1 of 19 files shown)


ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -65,6 +66,7 @@ def get_transaction_parameters(self, basket, request=None):
                 a payment from being created.
         """
         return_url = urljoin(self.ecommerce_url_root, reverse('paypal_execute'))
+        order_number = OrderNumberGenerator().order_number(basket)
```
clintonb: These calls to `OrderNumberGenerator` could be cleaned up by adding an `order_number` property to `Basket`:

``` python
class Basket(AbstractBasket):
  @property
  def order_number(self):
    if self.order:
      return self.order.number

```

clintonb:
:+1:

clintonb:
@jab5569 @mulby @brianhw This change corrects the information sent to the payment processors; however, our own `PaymentProcessorResponse` class is still linked to baskets. Let us know if this will be an issue.

ecommerce/extensions/order/utils.py
```diff
@@ -7,22 +7,38 @@ class OrderNumberGenerator(object):
 
     We need this as the order number is often required for payment
     which takes place before the order model has been created.
-
     """
+    OFFSET = 100000
```
jimabramson, line 11: what's the reason for the offset?

ecommerce/extensions/payment/processors/paypal.py
jimabramson: i wouldn't even bother checking `self.order` - why trouble the database.

ecommerce/extensions/payment/tests/mixins.py
```diff
@@ -381,4 +385,4 @@ def _create_api_url(self, path):
         mode = settings.PAYMENT_PROCESSOR_CONFIG['paypal']['mode']
         root = u'https://api.sandbox.paypal.com' if mode == 'sandbox' else u'https://api.paypal.com'
 
-        return root + path
+        return urljoin(root, path)
```
jimabramson, line 383: good call

ecommerce/extensions/payment/views.py
```diff
@@ -160,6 +162,9 @@ def get(self, request):
 
         try:
             user = basket.owner
+            # Given a basket, order number generation is idempotent. Although we've already
+            # generated this order number once before, it's faster to generate it again
+            # than to retrieve an invoice number from PayPal.
             order_number = self.generate_order_number(basket)
```
jimabramson: what's the guideline for using this method vs. using `OrderNumberGenerator().order_number()` directly ?

jimabramson:
@rlucioni looks solid - just had a few questions inline

ecommerce/extensions/order/utils.py
rlucioni, line 11: @jimabramson it's arbitrary. It was inspired by the [default behavior](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/order/utils.py#L30) of Oscar's `OrderNumberGenerator`. I think the idea is to slightly obfuscate basket database IDs.

ecommerce/extensions/payment/views.py
rlucioni: `generate_order_number` is a method on Oscar's `OrderPlacementMixin`. If your class is responsible for placing orders, I'd subclass our `EdxOrderPlacementMixin` and use this method. If you just need an order number, I think calling `OrderNumberGenerator().order_number()` directly suffices.

ecommerce/extensions/payment/processors/paypal.py
rlucioni: Thanks for the suggestions @clintonb and @jimabramson. This is updated.

ecommerce/extensions/payment/views.py
rlucioni: @jimabramson reconsidering what I said here in light of the new `order_number` property added to `Basket`. I think it'd be cleanest to prefer use of `basket.order_number` over either of the approaches discussed above.

ecommerce/extensions/order/tests/test_utils.py
```diff
@@ -9,15 +8,26 @@
 class UtilsTest(TestCase):
     """Unit tests for the order utility functions and classes. """
 
-    ORDER_NUMBER_PREFIX = "Zoidberg"
+    ORDER_NUMBER_PREFIX = 'FOO'
```
jimabramson, line 11: wait - why not zoidberg?!

jimabramson:
:+1:

ecommerce/extensions/order/tests/test_utils.py
rlucioni, line 11: Good one!

PR #139, opened 2015-06-12
Suppress emission of business intelligence events for free orders

rlucioni:
Pursuant to [XCOM-412](https://openedx.atlassian.net/browse/XCOM-412); this is a change requested by marketing, which consumes these events.

### DIFF
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -12,7 +12,7 @@ post_checkout = get_class('checkout.signals', 'post_checkout')
 @log_exceptions("Failed to emit tracking event upon order completion.")
 def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unused-argument
     """Emit a tracking event when an order is placed."""
-    if not is_segment_configured():
+    if not (is_segment_configured() and order.total_excl_tax > 0):
         return
 
     user_tracking_id, lms_client_id = parse_tracking_context(order.user)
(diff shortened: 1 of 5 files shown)


ecommerce/extensions/checkout/signals.py
```diff
@@ -12,7 +12,7 @@
 @log_exceptions("Failed to emit tracking event upon order completion.")
 def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unused-argument
     """Emit a tracking event when an order is placed."""
-    if not is_segment_configured():
+    if not is_segment_configured() or order.total_excl_tax == 0:
```
clintonb: This would be slightly easier to read: `if not (is_segment_configured() and order.total_excl_tax > 0)`.

ecommerce/extensions/refund/signals.py
```diff
@@ -12,7 +12,7 @@
 @log_exceptions("Failed to emit tracking event upon refund completion.")
 def track_completed_refund(sender, refund=None, **kwargs):  # pylint: disable=unused-argument
     """Emit a tracking event when a refund is completed."""
-    if not is_segment_configured():
+    if not is_segment_configured() or refund.total_credit_excl_tax == 0:
```
clintonb: (Same note as above.)

clintonb:
:+1:

ecommerce/extensions/checkout/signals.py
rlucioni: Good call, updated.

ecommerce/extensions/refund/signals.py
rlucioni: Updated.

PR #140, opened 2015-06-12
Nonatomic payment processing

rlucioni:
Allows payment processing views to commit to the database at each step along the way to order placement. These views perform three major functions: 1) recording raw processor responses, 2) recording successful payments by creating payment events and sources, and 3) placing orders. If any of these operations fails, we don't want any prior operations to be rolled back.

Additionally, if a failure occurs at any point in this process, dirty data shouldn't be introduced to the system. Processor responses, payment events and sources, and orders can each be introduced independently, one after the other. We don't rely on the atomic creation of these data; in fact, it's exactly what we'd like to avoid.

This addresses [XCOM-415](https://openedx.atlassian.net/browse/XCOM-415). @jimabramson or @clintonb, please review.

### DIFF
--- a/ecommerce/extensions/payment/tests/test_views.py
+++ b/ecommerce/extensions/payment/tests/test_views.py
@@ -12,6 +12,7 @@ from oscar.apps.payment.exceptions import PaymentError
 from oscar.core.loading import get_class, get_model
 from oscar.test import factories
 from oscar.test.contextmanagers import mock_signal_receiver
+from testfixtures import LogCapture
 
 from ecommerce.extensions.fulfillment.status import ORDER
 from ecommerce.extensions.payment.processors.cybersource import Cybersource
@@ -47,7 +48,7 @@ class CybersourceNotifyViewTests(CybersourceMixin, PaymentEventsMixin, TestCase)
         self.processor = Cybersource()
         self.processor_name = self.processor.NAME
 
-    def assert_payment_data_recorded(self, notification):
+    def _assert_payment_data_recorded(self, notification):
         """ Ensure PaymentEvent, PaymentProcessorResponse, and Source objects are created for the basket. """
 
         # Ensure the response is stored in the database
@@ -63,6 +64,34 @@ class CybersourceNotifyViewTests(CybersourceMixin, PaymentEventsMixin, TestCase)
         # Validate that PaymentEvents exist
         paid_type = PaymentEventType.objects.get(code='paid')
         self.assert_payment_event_exists(self.basket, paid_type, reference, self.processor_name)
+
+    def _assert_processing_failure(self, notification, status_code, error_message):
+        """Verify that payment processing operations fail gracefully."""
+        logger_name = 'ecommerce.extensions.payment.views'
+        with LogCapture(logger_name) as l:
+            response = self.client.post(reverse('cybersource_notify'), notification)
+
+            self.assertEqual(response.status_code, status_code)
+
+            self.assert_processor_response_recorded(
+                self.processor_name,
+                notification[u'transaction_id'],
+                notification,
+                basket=self.basket
+            )
+
+            l.check(
+                (
+                    logger_name,
+                    'INFO',
+                    'Received CyberSource merchant notification for transaction [{transaction_id}], '
+                    'associated with basket [{basket_id}].'.format(
+                        transaction_id=notification[u'transaction_id'],
+                        basket_id=self.basket.id
+                    )
+                ),
+                (logger_name, 'ERROR', error_message)
+            )
 
     # Disable the normal signal receivers so that we can verify the state of the created order.
     @mute_signals(post_checkout)
@@ -104,7 +133,7 @@ class CybersourceNotifyViewTests(CybersourceMixin, PaymentEventsMixin, TestCase)
         self.assertEqual(response.status_code, 200)
 
         # Validate the payment data was recorded for auditing
-        self.assert_payment_data_recorded(notification)
+        self._assert_payment_data_recorded(notification)
 
         # The basket should be marked as submitted. Refresh with data from the database.
         basket = Basket.objects.get(id=self.basket.id)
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/payment/tests/test_views.py
```diff
@@ -270,9 +295,17 @@ def test_payment_error(self):
         """
         with mock.patch.object(PaypalPaymentExecutionView, 'handle_payment',
                                side_effect=PaymentError) as fake_handle_payment:
-            self._assert_execution_redirect()
+            creation_response, _ = self._assert_execution_redirect()
```
clintonb: Use a double underscore to avoid potential issues with `ugettext`.

ecommerce/extensions/payment/views.py
```diff
@@ -81,6 +83,12 @@ def post(self, request):
             order_number = cybersource_response.get('req_reference_number')
             basket_id = OrderNumberGenerator().basket_id(order_number)
 
+            logger.info(
+                'Received CyberSource merchant notification for basket [%d], to be associated with order [%s].',
```
clintonb: This is redundant information. Choose one. Also, it is crucial that the CyberSource transaction ID be logged in the event of another failure to store this information in the database.

ecommerce/extensions/payment/tests/test_views.py
rlucioni: Updated.

clintonb:
You've disabled atomicity for the views as a whole, but you have not restored atomicity inside the view functions (after the payment processor response is recorded).

ecommerce/extensions/payment/views.py
```diff
@@ -29,6 +30,7 @@
 PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
 
 
+@transaction.non_atomic_requests
```
clintonb: This should decorate the post method.

jimabramson: I agree based on a look at the [docs](https://docs.djangoproject.com/en/1.8/topics/db/transactions/#django.db.transaction.non_atomic_requests).  The fact that the class is decorated, but the tests still pass, is confusing to me.

rlucioni: My mistake, I've corrected this. @jimabramson decorating the class appears to disable atomicity for every view method within, including the `post` method.

ecommerce/extensions/payment/views.py
rlucioni: Thanks, good catch. I've updated this and added tests to cover it. The tests have exposed some holes in our logging, but patching those is out of scope for this work and should be addressed by [XCOM-381](https://openedx.atlassian.net/browse/XCOM-381).

rlucioni:
@clintonb I am intentionally refraining from restoring atomicity within the (now correctly decorated) view functions for reasons described above in the description of this PR, and in the comment thread for [XCOM-415](https://openedx.atlassian.net/browse/XCOM-415). In a nutshell, we want these views to be able to commit to the database at each step along the way to order placement.

jimabramson:
@rlucioni if the process of placing or fulfilling the order involves more than one database insert/update it must be done atomically.  Otherwise a failure in between those operations will leave a mess.

clintonb:
What @jimabramson said. We don't want dirty data in the DB.

rlucioni:
@jimabramson and @clintonb you're right; I was thinking of order placement and fulfillment as an atomic operation, when it is in fact not. I'll update to make sure those operations are handled atomically.

rlucioni:
Partial atomicity has made its triumphant return, along with tests. Ready for re-review.

clintonb:
:+1: (Awaiting second review from @jimabramson)

ecommerce/extensions/payment/views.py
```diff
@@ -156,28 +172,34 @@ def get(self, request):
...
+                        billing_address=None,
+                        order_total=order_total
+                    )
+                except UnableToPlaceOrder:
+                    logger.exception('Payment was executed, but an order was not created for basket [%d].', basket.id)
+
+                return redirect(receipt_url)
+        except:  # pylint: disable=bare-except
+            logger.exception('Payment was received, but attempts to create an order for basket [%d] failed.', basket.id)
+            return redirect(receipt_url)
```
jimabramson, line 218: so when order creation fails, the user will get the checkout receipt page, which will time out after 5 attempts to fetch the order and draw a receipt (right)?    not saying this is better/worse than the present state, just confirming the expectation.

i doubt it's worth it, but we could consider sending the user to the receipt without a basket_id parameter.   this would at least have the effect of skipping the 5 attempts at retrieving the order.

ecommerce/extensions/payment/views.py
```diff
@@ -103,27 +111,34 @@ def post(self, request):
...
+                except UnableToPlaceOrder:
+                    logger.exception('Payment was received, but an order was not created for basket [%d].', basket.id)
+                    # Ensure we return, in case future changes introduce post-order placement functionality.
+                    return HttpResponse()
 
-        return HttpResponse()
+                return HttpResponse()
+        except:  # pylint: disable=bare-except
+            logger.exception('Payment was received, but attempts to create an order for basket [%d] failed.', basket.id)
+            return HttpResponse()
```
jimabramson: is there any reason not to respond with status 500 here?   it's not user facing, but i think it would increase the likelihood of noticing problems quickly.

ecommerce/extensions/payment/views.py
rlucioni, line 218: @jimabramson yes, that's the expected behavior. I don't think the added complexity of sending the user to the receipt without a basket ID is worth it. One way or another, they're going to encounter an error message.

ecommerce/extensions/payment/views.py
rlucioni: Good idea. Updated.

jimabramson:
shouldn't the call to `handle_payment` (https://github.com/edx/ecommerce/blob/renzo/nonatomic-payment-processing/ecommerce/extensions/payment/views.py#L103) also get wrapped in `atomic`?   it's doing two inserts.

rlucioni:
@jimabramson yes, nice catch. Sources and payment events shouldn't be created separately here. Updated.

ecommerce/extensions/payment/views.py
```diff
@@ -92,38 +100,54 @@ def post(self, request):
...
+                billing_address = self._get_billing_address(cybersource_response)
+
+                try:
+                    user = basket.owner
+                    self.handle_order_placement(order_number, user, basket, None, shipping_method, shipping_charge,
+                                                billing_address, order_total)
+                except UnableToPlaceOrder:
+                    logger.exception('Payment was received, but an order was not created for basket [%d].', basket.id)
+                    # Ensure we return, in case future changes introduce post-order placement functionality.
+                    return HttpResponse()
```
jimabramson: sorry for missing this before, but I think this should be a 500 as well.

ecommerce/extensions/payment/views.py
```diff
@@ -92,38 +100,54 @@ def post(self, request):
...
+                        ppr.id
+                    )
+                    return HttpResponse(status=400)
+                except PaymentError:
+                    logger.exception(
+                        'CyberSource payment failed for basket [%d]. The payment response was recorded in entry [%d].',
+                        basket.id,
+                        ppr.id
+                    )
+                    return HttpResponse()
```
jimabramson, line 119: same here, 500

jimabramson, line 119: also, one more here https://github.com/edx/ecommerce/blob/renzo/nonatomic-payment-processing/ecommerce/extensions/payment/views.py#L96.

if you prefer to make those changes in another PR i'm ok with that

jimabramson:
@rlucioni :+1: whether you decide to add those 500s or not - it's not strictly in the scope of this change.

ecommerce/extensions/payment/views.py
rlucioni: Updated.

ecommerce/extensions/payment/views.py
rlucioni, line 119: Updated both. I've made the latter a 400, since CyberSource shouldn't be sending us invalid payments.

ecommerce/extensions/payment/views.py
rlucioni: This actually shouldn't be a 500, since we're handling CyberSource's response appropriately. Reverted to 200.

rlucioni: Scratch that, I was referring to the [PaymentError response](https://github.com/edx/ecommerce/pull/140#discussion-diff-32354343R119).

rlucioni:
@jimabramson I've updated status codes where appropriate. Thanks to you both for the sharp eyes.

PR #141, opened 2015-06-12
BI events contain product's certificate_type in the 'sku' field

rlucioni:
For backwards-compatibility with older events, Segment's `sku` field is (ab)used to store the product's `certificate_type`, while the `id` field holds the product's SKU. Marketing is aware that this approach will not scale once we start selling products other than courses, and that it will need to change in the future.

Meets the requirements of [XCOM-413](https://openedx.atlassian.net/browse/XCOM-413). @jimabramson or @clintonb, please review.

### DIFF
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -26,8 +26,12 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
             'currency': order.currency,
             'products': [
                 {
-                    'id': line.upc,
-                    'sku': line.partner_sku,
+                    # For backwards-compatibility with older events the `sku` field is (ab)used to
+                    # store the product's `certificate_type`, while the `id` field holds the product's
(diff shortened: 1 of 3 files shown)


jimabramson:
this is kind of unfortunate IMO, but :+1:

rlucioni:
I agree about it being unfortunate. However, since we're unable to use GA's "Enhanced Ecommerce" product from within our Python code, we don't have another good way to relay this kind of categorical information to GA.

clintonb:
Why are we unable to use the Enhanced Ecommerce feature?

rlucioni:
@clintonb as explained in [XCOM-414](https://openedx.atlassian.net/browse/XCOM-414)'s comment thread, Segment's server-side libraries [do not currently support GA's Enhanced Ecommerce product](https://segment.com/docs/integrations/google-analytics/#enhanced-e-commerce). I'm not aware of a Python SDK for GA, published by Google or anyone else, which is compatible with Enhanced Ecommerce, nor have I seen documentation for an Enhanced Ecommerce API.

PR #142, opened 2015-06-12
Added JS/SASS pipeline

clintonb:
TODO
- Move django_rjs to a separate repo (to be shared with Insights)
- Thoughts on a common SASS lib (installable via bower) @frrrances?

XCOM-367


clintonb:
FYI @jimabramson @rlucioni. This isn't quite yet ready to merge, due to the outstanding TODO items.

Makefile
```diff
@@ -45,6 +51,12 @@ quality:
 
 validate: test_python quality
 
+static:
```
rlucioni, line 54: Should the `static` target be run as part of the `validate` target?

ecommerce/django_rjs/models.py
```diff
@@ -0,0 +1 @@
+# This app does not require any models.
```
rlucioni: As of [Django 1.7](https://docs.djangoproject.com/en/1.7/releases/1.7/#app-loading-refactor), you can omit `models.py` if an app doesn't have any models.

ecommerce/django_rjs/templatetags/rjs.py
```diff
@@ -0,0 +1,45 @@
...
+
+register = template.Library()
+
+
+def get_rjs_path(path):
+    if getattr(settings, 'RJS_OPTIMIZATION_ENABLED', False):
+        rjs_output_dir = getattr(settings, 'RJS_OUTPUT_DIR')
+
+        if rjs_output_dir:
+            return '{0}/{1}'.format(rjs_output_dir, path)
```
rlucioni: Nitpick: you don't need these formatting indices.

ecommerce/static/sass/_base.scss
```diff
@@ -0,0 +1,462 @@
+// app styling: base
```
rlucioni, line 1: Where is this Sass from? Insights? I'm nervous about maintaining base styles across two or more separate repos.

rlucioni, line 1: Ignore me, I see this is one of your TODO items.

frrrances:
@clintonb Up to speed now... I don't think we should dupe the Insights sass into this project as it was done as a quick custom job for the particular Insights needs. Ideally we want to be using our pattern library set up and standards: 

https://github.com/edx/ux-pattern-library/wiki/Styleguide:-General

We can help support work with those standards. If the decision for bootstrap has already been made, I would recommend you use Bootstrap out of the box and use their tools for theming/customization, though we won't have time to customize for special cases at this time. Let me know which route you choose and how I can help.

clintonb:
@frrrances I personally like Bootstrap; but, I am not opposed to exploring alternatives. What did you have in mind.

Will bring the layout up to standard shortly.

Makefile
clintonb, line 54: That is not necessary for local testing (where `COMPRESS_ENABLED` is generally set to `False`). The command is run for Travis builds. Also, it's slow.

ecommerce/settings/base.py
```diff
@@ -103,6 +103,7 @@
 
 # See: https://docs.djangoproject.com/en/dev/ref/contrib/staticfiles/#std:setting-STATICFILES_DIRS
 STATICFILES_DIRS = (
+    normpath(join(DJANGO_ROOT, 'static', 'build')),  # Check the r.js output directory first
```
clintonb, line 106: FYI @dsjen: I was able to eliminate the need for the `django_rjs` app with this change. A similar update should work for Insights.

clintonb:
Barring any objections from @frrrances, this is ready for review/merge.

ecommerce/settings/base.py
dsjen, line 106: thanks for letting me know!

frrrances:
Thanks for the cleanup @clintonb. No objections from me.

build.js
```diff
@@ -0,0 +1,28 @@
+/* jshint asi: true, expr:true */
```
jimabramson: what's the motivation for relaxing these?

jimabramson: nit: close whitespace: `asi:true`

build.js
```diff
@@ -0,0 +1,28 @@
...
+        {
+            name: 'js/common'
+        },
+        {
+            name: 'js/config'
+        },
+        //{
+        //    name: 'js/engagement-content-main',
+        //    exclude: ['js/common']
+        //}
```
jimabramson, line 29: please uncomment, remove, or add explanation.

ecommerce/settings/test.py
```diff
@@ -107,3 +107,8 @@ def __getitem__(self, item):
     },
 }
 # END PAYMENT PROCESSING
+
+# Use production settings for asset compression so that asset compilation can be tested on the CI server.
+COMPRESS_ENABLED = True
+COMPRESS_OFFLINE = True
+RJS_OPTIMIZATION_ENABLED = True
```
jimabramson: does this imply a penalty when running tests locally?

ecommerce/templates/edx/base.html
```diff
@@ -0,0 +1,42 @@
+{# Base template for edX-specific pages. #}
```
jimabramson, line 1: how does this play out in terms of theming?  out of the box, IMO, we should provide the 'openedx' theme - if edx.org requires domain-specific branding, that should be done via configuration.

build.js
clintonb, line 29: Done.

ecommerce/settings/test.py
clintonb: The RJS setting has been removed as it is no longer needed. The penalty is incurred when running on Travis since it calls `make static`. This is not necessary for local Django tests since they do not actually execute JS/CSS.

build.js
clintonb: Updated. More on jshint at http://jshint.com/docs/options/. `build.js` is "special" in that using a semi-colon or assignment will fail when r.js is called.

ecommerce/templates/edx/base.html
clintonb, line 1: The OpenEdx theme used by platform is not compatible with Bootstrap. We'll need to work with @frrrances and @AlasdairSwan to create a compatible theme. "edX-specific" here means OpenEdX, not edX.org. Overriding `ecommerce/templates/base.html` causes issues for the dashboard that I would rather not solve right now.

PR #143, opened 2015-06-15
log incoming payments and their application to orders.

jimabramson:
XCOM-381

@rlucioni @clintonb pls check it.   I will follow up with a fixup commit for tests.

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -53,3 +53,26 @@ def log_exceptions(msg):
                 raise
         return wrapper
     return decorator
+
+
+def log_payment_received(processor_name, reference, amount, basket_id):
+    """Emit a log entry indicating that we were notified about a successful payment.
+
+    This should be called immediately after the notification has been received and interpreted (i.e.
+    after `handle_payment`).
+    """
+    logger.info(
+        'payment_received: processor_name="%s", reference="%s", amount="%s", basket_id="%s"',
+        processor_name,
+        reference,
+        amount,
+        basket_id,
+    )
+
+
+def log_payment_applied(amount, basket_id, user_id):
+    """Emit a log entry indicating that a payment was successfully applied to an order.
+
+    This should be called immediately after the order has been saved to db (i.e. after `handle_order_placement`).
+    """
+    logger.info('payment_applied: amount="%s", basket_id="%s", user_id="%s"', amount, basket_id, user_id)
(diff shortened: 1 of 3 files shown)


ecommerce/extensions/analytics/utils.py
```diff
@@ -53,3 +53,26 @@ def wrapper(*args, **kwargs):  # pylint: disable=missing-docstring
...
+
+
+def log_payment_received(processor_name, reference, amount, basket_id):
+    """Emit a log entry indicating that we were notified about a successful payment.
+
+    This should be called immediately after the notification has been received and interpreted (i.e.
+    after `handle_payment`).
+    """
+    logger.info(
+        'payment_received: processor="%s", ref="%s", amount="%s", basket="%s"',
```
rlucioni: Minor, but for the sake of clarity and consistency with payment events, could we use:
- "processor_name" instead of "processor"
- "reference" instead of "ref"
- "basket_id" instead of "basket"

jimabramson:
updated

ecommerce/extensions/checkout/mixins.py
```diff
@@ -35,8 +36,10 @@ def handle_payment(self, response, basket):
 
         self.add_payment_source(source)
         self.add_payment_event(payment_event)
+        log_payment_received(payment_event.processor_name, payment_event.reference, payment_event.amount, basket.id)
 
     def handle_successful_order(self, order):
         # Send a signal so that receivers can perform relevant tasks (e.g. fulfill the order).
         post_checkout.send_robust(sender=self, order=order)
+        log_payment_applied(order.total_incl_tax, order.basket.id, order.user.id)
```
rlucioni: Also minor, but we've been using `total_excl_tax` instead of `total_incl_tax`; your tests use `total_excl_tax`.

rlucioni:
My second minor comment aside, :+1: .

ecommerce/extensions/checkout/tests/test_mixins.py
```diff
@@ -22,6 +28,37 @@ def setUp(self):
...
+            'ecommerce.extensions.checkout.mixins.EdxOrderPlacementMixin.payment_processor',
+            mock_payment_processor
+        ):
+            with LogCapture(LOGGER_NAME) as l:
+                EdxOrderPlacementMixin().handle_payment(Mock(), Mock(id='test-basket-id'))
+                l.check(
+                    (
+                        LOGGER_NAME,
+                        'INFO',
+                        'payment_received: processor_name="{}", reference="{}", amount="{}", basket_id="{}"'.format(
```
clintonb, line 53: Minor: If you aren't going to share variables/constants, there is no need to use `format`.

clintonb:
:+1:

PR #144, opened 2015-06-15
Fixed Unicode bug

clintonb:
Some of the data being logged contains Unicode, thus it should be logged as such.

XCOM-336

@jimabramson @rlucioni


rlucioni:
:+1:

jimabramson:
:+1:

PR #145, opened 2015-06-15
Fixed URL decoding bug

clintonb:
Django is being too helpful when decoding params. We want to keep the '+' instead of converting it to a space.

XCOM-336

@jimabramson @rlucioni This is the last merge needed to finalize the migration.


jimabramson:
@clintonb why can't you just urlencode the course ids prior to making the HTTP call?

clintonb:
@jimabramson We do NOT want any encoding. The downstream APIs expect an unencoded course ID.

clintonb:
Closing since we can/should URL-encode the params.

PR #146, opened 2015-06-15
Ensure that newly-created enrollments are always active

rlucioni:
The old LMS `shoppingcart` app creates _inactive_ "honor" enrollments in courses for which the user is trying to purchase a certificate. These default "honor" enrollments could be left in the system between migrating a course from `shoppingcart` to Otto. Since Otto didn't previously specify an activation status, only a mode, the enrollment API would update the mode of the user's existing enrollment, but not the activation status.


clintonb:
:+1:

jimabramson:
:+1:

PR #147, opened 2015-06-16
Storing Historical Course Edits

clintonb:
XCOM-196

@jimabramson @rlucioni 

@Nickersoft once this is merged you will be able to access course edit dates as follows:

``` python
course = ...
course.history.latest().date_updated
```

### DIFF
--- a/ecommerce/courses/admin.py
+++ b/ecommerce/courses/admin.py
@@ -1,10 +1,30 @@
+from django.conf.urls import patterns, url
 from django.contrib import admin
+from simple_history.admin import SimpleHistoryAdmin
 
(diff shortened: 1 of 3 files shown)


rlucioni:
This will require that we run the history population command during the next deployment, before modifying any course products or creating new ones. Could you create a ticket to remind us to do this?

Otherwise, :+1: .

jimabramson:
:+1:

clintonb:
FYI @Nickersoft. Rebase against master to get access to historical course data. You'll also need to run migrations and the following script: `python manage.py populate_history --auto`.

PR #149, opened 2015-06-16
Awais786/ecom 1714 checkout page

awais786:
End point for checkout page.

### DIFF
--- a/ecommerce/credit/tests/test_views.py
+++ b/ecommerce/credit/tests/test_views.py
@@ -1,0 +1,41 @@
+"""
+Tests for the checkout page.
+"""
+from django.core.urlresolvers import reverse
+from django.test import TestCase
+
+import mock
+from ecommerce.credit import views
+
+
+from ecommerce.tests.mixins import UserMixin
+
+
+class CheckoutPageTest(UserMixin, TestCase):
+    """Test for Checkout page"""
+    def setUp(self):
+        super(CheckoutPageTest, self).setUp()
+        user = self.create_user(is_superuser=False)
+        self.client.login(username=user.username, password=self.password)
+        self.course_id = u'edx/Demo_Course/DemoX'
+
+    def test_get_with_enabled_flag(self):
+        """
+        Test checkout page accessibility. Page will appear only if feature
+        flag is enabled.
+        """
+        with mock.patch.object(views.waffle, 'flag_is_active') as mock_flag:
+            mock_flag.return_value = True
+            response = self.client.get(reverse('credit:checkout', args=[self.course_id]))
+
+            self.assertEqual(response.status_code, 200)
+
+    def test_get_with_disabled_flag(self):
+        """
+        Test checkout page accessibility. Page will return 404 if no flag is defined
+        of it is disabled.
+        """
+
+        response = self.client.get(reverse('credit:checkout', args=[self.course_id]))
+
+        self.assertEqual(response.status_code, 404)
--- a/ecommerce/credit/urls.py
+++ b/ecommerce/credit/urls.py
@@ -1,0 +1,11 @@
+from django.conf import settings
+from django.conf.urls import patterns, url
+
+from ecommerce.credit.views import Checkout
+
+
+urlpatterns = patterns(
+    '',
+    url(r'^checkout/{course}/$'.format(course=settings.COURSE_ID_PATTERN),
+        Checkout.as_view(), name='checkout'),
+)
--- a/ecommerce/credit/views.py
+++ b/ecommerce/credit/views.py
@@ -1,0 +1,24 @@
+from django.http import Http404
+from django.views.generic import TemplateView
+
+import waffle
+
+
+class Checkout(TemplateView):
+    """Checkout page that describes the product the user is buying
+    and displays the data of the institution offering credit.
+    """
+    template_name = 'edx/credit/checkout.html'
+
+    def get_context_data(self, **kwargs):
+        context = super(Checkout, self).get_context_data(**kwargs)
+        context['request'] = self.request
+        context['user'] = self.request.user
+
+        return context
+
+    def get(self, request, *args, **kwargs):
+        """Get method for checkout page."""
+        if not waffle.flag_is_active(request, 'ENABLE_CREDIT_APP'):
+            raise Http404
+        return super(Checkout, self).get(request, args, **kwargs)
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -410,3 +410,6 @@ SESSION_COOKIE_NAME = 'ecommerce_sessionid'
 CSRF_COOKIE_NAME = 'ecommerce_csrftoken'
 LANGUAGE_COOKIE_NAME = 'ecommerce_language'
 # END COOKIE CONFIGURATION
+
+# Standard regex for course_id.
+COURSE_ID_PATTERN = r'(?P<course_id>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'
--- a/ecommerce/templates/edx/credit/checkout.html
+++ b/ecommerce/templates/edx/credit/checkout.html
@@ -1,0 +1,7 @@
+{% extends 'edx/base.html' %}
+{% load i18n %}
+{% block title %}{% trans "Credit Checkout" %}{% endblock title %}
+
+{% block content %}
+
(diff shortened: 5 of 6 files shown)


ecommerce/credit/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import patterns, url
+
+from ecommerce.credit.views import Checkout
+
+urlpatterns = patterns(
+    '',
+    url(r'^checkout/$', Checkout.as_view(), name='check_out'),
```
clintonb: The name should be "checkout".

ecommerce/credit/tests/test_views.py
```diff
@@ -0,0 +1,22 @@
...
+    """Test for Checkout page"""
+    def setUp(self):
+        super(CheckoutPageTest, self).setUp()
+        user = self.create_user(is_superuser=False)
+        self.client.login(username=user.username, password=self.password)
+
+    def test_get_method(self):
+        """Test while accessing the url valid checkout page appear."""
+        response = self.client.get(
+            '/credit/checkout/'
```
clintonb: No hard-coded URLs. Use `reverse('credit:checkout')`.

ecommerce/credit/views.py
```diff
@@ -0,0 +1,17 @@
+from django.shortcuts import render_to_response
+from django.views.generic import View
+
+
+class Checkout(View):
```
clintonb: Subclass [`TemplateView`](https://docs.djangoproject.com/en/1.8/ref/class-based-views/base/#templateview).

ecommerce/credit/views.py
```diff
@@ -0,0 +1,17 @@
...
+    """Checkout page that describes the product the user is buying
+    and displays the name of the institution offering credit.
+    """
+
+    def get(self, request):
+        """get method for checkout page."""
+        context = {
+            'request': request,
```
clintonb: This context variable is not used.

ecommerce/templates/oscar/dashboard/credit/checkout.html
```diff
@@ -0,0 +1,28 @@
+{% extends 'dashboard/layout.html' %}
```
clintonb: This is a customer-facing page. Extend `edx/base.html`.

ecommerce/templates/oscar/dashboard/credit/checkout.html
```diff
@@ -0,0 +1,28 @@
...
+{% load currency_filters %}
+{% load i18n %}
+
+{% block body_class %}{{ block.super }} users{% endblock %}
+
+{% block title %}{{ customer.email }} | {{ block.super }}{% endblock %}
+
+{% block breadcrumbs %}
```
clintonb: None of this is necessary.

awais786:
@zubair-arbi please review.

ecommerce/settings/test.py
```diff
@@ -107,7 +107,8 @@ def __getitem__(self, item):
     },
 }
 # END PAYMENT PROCESSING
-
+COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'
```
zubair-arbi: Since we are importing settings from `base.py` so its redundant. Remove it.

ecommerce/templates/oscar/dashboard/credit/checkout.html
```diff
@@ -0,0 +1 @@
+{% extends 'edx/base.html' %}
```
zubair-arbi: nitpick: add newline at the end of file.

zubair-arbi:
Changes looks good :+1:

ecommerce/credit/views.py
```diff
@@ -0,0 +1,19 @@
...
+    template_name = "dashboard/credit/checkout.html"
+
+    def get_context_data(self, **kwargs):
+        context = super(Checkout, self).get_context_data(**kwargs)
+        context['request'] = self.request
+        context['user'] = self.request.user
+
+        return context
+
+    def get(self, request, *args, **kwargs):
```
clintonb, line 20: This override is unnecessary and should be removed.

ecommerce/credit/views.py
```diff
@@ -0,0 +1,19 @@
+from django.views.generic import TemplateView
+
+
+class Checkout(TemplateView):
+    """Checkout page that describes the product the user is buying
+    and displays the data of the institution offering credit.
+    """
+    template_name = "dashboard/credit/checkout.html"
```
clintonb: None of these pages should be in the dashboard directory. Also, please use single quotes.

ecommerce/credit/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.conf import settings
+from django.conf.urls import patterns, url
+
+from ecommerce.credit.views import Checkout
+
+
+urlpatterns = patterns(
+    '',
+    url(r'^checkout/{course}/$'.format(course=settings.COURSE_ID_PATTERN),
```
clintonb, line 9: I am hesitant to add the course_key to the URL. This led to problems with the receipt page in LMS. However, I will discuss this with @wedaly later today.

ecommerce/credit/tests/test_views.py
```diff
@@ -0,0 +1,22 @@
...
+
+class CheckoutPageTest(UserMixin, TestCase):
+    """Test for Checkout page"""
+    def setUp(self):
+        super(CheckoutPageTest, self).setUp()
+        user = self.create_user(is_superuser=False)
+        self.client.login(username=user.username, password=self.password)
+        self.course_id = u'edx/Demo_Course/DemoX'
+
+    def test_get_method(self):
```
clintonb: Just `test_get` is sufficient.

ecommerce/settings/base.py
```diff
@@ -410,3 +410,7 @@ def get_lms_url(path):
 CSRF_COOKIE_NAME = 'ecommerce_csrftoken'
 LANGUAGE_COOKIE_NAME = 'ecommerce_language'
 # END COOKIE CONFIGURATION
+
+# These are standard regexes for pulling out info like course_ids.
+COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'
+COURSE_ID_PATTERN = COURSE_KEY_PATTERN.replace('course_key_string', 'course_id')
```
clintonb: This is unnecessary and will lead to confusion. Just go with `COURSE_ID_PATTERN = r'(?P<course_id>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'` and remove the rest.

jimabramson: @awais786 this looks like it's only meant to match old slash-separated course IDs, not CourseKeys based on opaque_keys.  Is that intentional?

jimabramson:
I notice this is being added in ecommerce/credit rather than ecommerce/extensions/credit.  I think that makes sense, assuming the convention is to put things in extensions only when they override things that are built in to Oscar.  But, would like to confirm that and make it explicit.   @awais786 @clintonb

clintonb:
@jimabramson Yes, this is the same behavior I initiated with the `courses` app.

ecommerce/urls.py
```diff
@@ -79,3 +81,11 @@ def handler403(_):
             '',
             url(r'^__debug__/', include(debug_toolbar.urls)),
         )
+
+
+# Credit app
+if waffle.switch_is_active('ENABLE_CREDIT_APP'):
```
awais786: @clintonb  Urls.py might not pickup the django admin flags effect. May be need to move this into view ? but that will expose the url.

clintonb: Modify the view itself:

``` python
def get(...):
    raise Http404
  return super(...)
```

awais786:

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'edx/base.html' %}
+{% load i18n %}
+{% block title %}{% trans "Credit Checkout" %}{% endblock title %}
+
+{% block content %}
+
+{% endblock content %}
```
clintonb: Add a new line.

clintonb:
:+1:

PR #151, opened 2015-06-17
Added edX styling to Bootstrap

clintonb:
The default, primary, and link buttons have been updated according to the edX pattern library. A demo page is included to demonstrate the new styling.

XCOM-367

@frrrances @jimabramson @rlucioni

### DIFF
--- a/ecommerce/static/sass/_base.scss
+++ b/ecommerce/static/sass/_base.scss
@@ -1,3 +1,56 @@
 // app styling: base
 // ====================
 // NOTE: used for any basic styling or overriding that isn't covered by Bootstrap or theming
+body {
+  // Make room for the navbar.
+  padding-top: $navbar-height + $navbar-margin-bottom;
+}
+
+.btn-default {
+  // Invert the background and foreground colors on active/focus/hover
+  &:hover,
+  &:focus,
+  &.focus,
+  &:active,
+  &.active,
+  .open > &.dropdown-toggle {
+    color: $btn-default-bg;
+    background-color: $btn-default-color;
+    border-color: $btn-default-color;
+  }
+
+  // Use edX disabled state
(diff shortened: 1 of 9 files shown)


ecommerce/urls.py
```diff
@@ -70,6 +70,8 @@ def handler403(_):
         url(r'^403/$', handler403, name='403'),
         url(r'^404/$', 'django.views.defaults.page_not_found', name='404'),
         url(r'^500/$', 'django.views.defaults.server_error', name='500'),
+        url(r'^bootstrap/$', TemplateView.as_view(template_name='bootstrap-demo.html')),
```
jimabramson, line 82: would it be worth only adding this when DEBUG is True ?

ecommerce/static/sass/_base.scss
```diff
@@ -1,3 +1,56 @@
 // app styling: base
 // ====================
 // NOTE: used for any basic styling or overriding that isn't covered by Bootstrap or theming
+body {
```
rlucioni, line 4: Is this the best way to use the pattern library? What if something in the pattern library changes? We'd need to update it here, separately.

clintonb, line 4: For now, yes. As we discussed in grooming, there is no perfect integration of the edX pattern library with Bootstrap.

ecommerce/urls.py
clintonb, line 82: It is. See line 66 (which Github is hiding).

ecommerce/static/sass/_base.scss
rlucioni, line 4: Thanks; I hadn't seen this before grooming.

rlucioni:
:+1:

PR #154, opened 2015-06-17
Running compress before tests

clintonb:
XCOM-367

FYI @jimabramson @rlucioni

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,7 @@ clean:
 	coverage erase
 
 test_python: clean
+	python manage.py compress --settings=ecommerce.settings.test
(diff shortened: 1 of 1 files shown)


rlucioni:
For my own edification, could you explain why this is necessary?

clintonb:
This specific change is only needed for local usage. In general, we want to ensure asset compilation works as expected _before_ we build an AMI and attempt to deploy. We had a few issues with assets when deploying Insights. Testing this functionality ensures we don't waste time later.

rlucioni:
Cool, thanks.

PR #155, opened 2015-06-17
Display and filter by username for orders and refunds

rlucioni:
Customer usernames are now displayed on order list and detail pages, refund list and detail pages, and user detail pages. Order and refund lists can also be filtered by username. Pursuant to [XCOM-390](https://openedx.atlassian.net/browse/XCOM-390).

@clintonb and @jimabramson, please review when able. @Nickersoft, FYI.

### DIFF
--- a/ecommerce/extensions/dashboard/orders/views.py
+++ b/ecommerce/extensions/dashboard/orders/views.py
@@ -2,10 +2,26 @@ from django.contrib import messages
 from django.core.urlresolvers import reverse
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
-from oscar.apps.dashboard.orders.views import OrderDetailView as CoreOrderDetailView
+from oscar.apps.dashboard.orders.views import (
+    OrderListView as CoreOrderListView, OrderDetailView as CoreOrderDetailView
+)
 from oscar.core.loading import get_model
 
+
 Refund = get_model('refund', 'Refund')
+
+
+class OrderListView(CoreOrderListView):
+    def get_queryset(self):
+        queryset = super(OrderListView, self).get_queryset()
+
+        form = self.form_class(self.request.GET)
+        if form.is_valid():
+            for field, value in form.cleaned_data.iteritems():
+                if field == 'username' and value:
+                    queryset = queryset.filter(user__username__istartswith=value)
+
+        return queryset
 
 
 class OrderDetailView(CoreOrderDetailView):
--- a/ecommerce/templates/oscar/dashboard/orders/order_detail.html
+++ b/ecommerce/templates/oscar/dashboard/orders/order_detail.html
@@ -30,31 +30,22 @@
   {% block customer_information %}
       <table class="table  table-bordered table-hover">
           <caption><i class="icon-group icon-large"></i>{% trans "Customer Information" %}</caption>
-        {% if order.guest_email %}
-            <tr>
-                <th>{% trans "Name" %}</th>
-                <td>
-                  {% trans "Customer checked out as a guest." %}
-                </td>
-            </tr>
-            <tr>
-                <th>{% trans "Email address" %}</th>
-                <td>{{ order.email }}</td>
-            </tr>
-          {% elif order.user %}
-            <tr>
-                <th>{% trans "Name" %}</th>
-                <th>{% trans "Email address" %}</th>
-            </tr>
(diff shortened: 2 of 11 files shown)


clintonb:
:+1: 

Minor: Link to the ticket in the commit message.

ecommerce/extensions/dashboard/orders/views.py
```diff
@@ -2,12 +2,29 @@
...
+class OrderListView(CoreOrderListView):
+    def get_queryset(self):
+        queryset = super(OrderListView, self).get_queryset()
+
+        form = self.form_class(self.request.GET)
+
+        if form.is_valid():
+            for field, value in form.cleaned_data.iteritems():
+                if field == 'username' and value:
+                    queryset = queryset.filter(user__username=value)
```
clintonb: It would make me super-happy if you used [`istartswith`](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#istartswith) to perform a case-sensitive inexact match (e.g. search). At a minimum we should use [`iexact`](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#iexact) since we have usernames with upper- and lower-case letters.

ecommerce/templates/oscar/dashboard/orders/order_detail.html
```diff
@@ -30,31 +30,22 @@
...
+                  <th>{% trans "Email address" %}</th>
+              </tr>
+              <tr>
+                  <td><a href="{% url 'dashboard:user-detail' pk=order.user.id %}">{{ order.user.username|default:"-" }}</a></td>
+                  <td>{{ order.user.get_full_name|default:"-" }}</td>
+                  <td>{{ order.user.email|default:"-" }}</td>
+              </tr>
+          {% else %}
+              <tr>
+                  <td>{% trans "Customer has deleted their account." %}</td>
```
jimabramson, line 46: does this not need `colspan="3"` ?

jimabramson, line 46: never mind, i see it does not.

ecommerce/templates/oscar/dashboard/refunds/refund_list.html
```diff
@@ -114,7 +114,11 @@ <h3 class="pull-left"><i class="icon-repeat icon-large icon-flip-horizontal"></i
                     <td>{{ refund.num_items }}</td>
                     <td class="refund-status">{{ refund.status }}</td>
                     <td>
-                        <a href="{% url 'dashboard:user-detail' pk=refund.user.id %}">{{ refund.user.get_full_name|default:"-" }}</a>
+                        {% if refund.user %}
+                            <a href="{% url 'dashboard:user-detail' pk=refund.user.id %}">{{ refund.user.username|default:"-" }}</a>
```
jimabramson: isn't username required?

ecommerce/templates/oscar/dashboard/refunds/refund_detail.html
```diff
@@ -46,14 +46,22 @@
...
+        {% if refund.user %}
+            <tr>
+                <th>{% trans "Username" %}</th>
+                <th>{% trans "Full name" %}</th>
+                <th>{% trans "Email address" %}</th>
+            </tr>
+            <tr>
+                <td><a href="{% url 'dashboard:user-detail' pk=refund.user.id %}">{{ refund.user.username|default:"-" }}</a></td>
+                <td>{{ refund.user.get_full_name|default:"-" }}</td>
+                <td>{{ refund.user.email|default:"-" }}</td>
```
jimabramson: there aren't all required?

jimabramson:
:+1: assuming that you remove the `default` bits in the templates iff they are not necessary.

ecommerce/templates/oscar/dashboard/refunds/refund_list.html
rlucioni: Yes. Updated.

ecommerce/templates/oscar/dashboard/refunds/refund_detail.html
rlucioni: Not the full name field, but the others, yes. Updated.

rlucioni: Correction: none of these should ever be `None`. Good call.

PR #158, opened 2015-06-18
Revert "Fixed lingering issues with the CourseListView"

Nickersoft:
Until I can figure out where these errors are originating from, this change should not have been merged into the codebase.

@clintonb


clintonb:
:+1:

PR #163, opened 2015-06-18
Displaying User Enrollments on Dashboard

clintonb:
User with enrollments:
user-enrollments

User without enrollments:
user-no-enrollments

@rlucioni @jimabramson A bit of playing around.

### DIFF
--- a/ecommerce/extensions/dashboard/users/tests/test_views.py
+++ b/ecommerce/extensions/dashboard/users/tests/test_views.py
@@ -1,0 +1,85 @@
+import json
+
+from django.conf import settings
+from django.contrib.messages import constants as MSG
+from django.core.urlresolvers import reverse
+from django.test import TestCase
+import httpretty
+import mock
+from requests import Timeout
+from testfixtures import LogCapture
+from waffle import Switch
+
+from ecommerce.extensions.dashboard.tests import DashboardViewTestMixin
+from ecommerce.extensions.dashboard.users.views import UserDetailView
+from ecommerce.tests.mixins import UserMixin
+
+LOGGER_NAME = 'ecommerce.extensions.dashboard.users.views'
+
+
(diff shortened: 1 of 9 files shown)


jimabramson:
@clintonb I think we should handle an enrollment api timeout (or other error) gracefully, displaying something like "unable to retrieve enrollment information" instead of the table when there's a problem.

clintonb:
@jimabramson Updated. I'll add tests for this later (before merging). It isn't in scope for this sprint, so I'm not going to rush it.

clintonb:
@jimabramson I went ahead and wrote tests and created a ticket.

clintonb:
XCOM-347

rlucioni:
Nice. :+1:

ecommerce/extensions/dashboard/users/tests/test_views.py
```diff
@@ -0,0 +1,85 @@
...
+from ecommerce.extensions.dashboard.users.views import UserDetailView
+from ecommerce.tests.mixins import UserMixin
+
+LOGGER_NAME = 'ecommerce.extensions.dashboard.users.views'
+
+
+class UserDetailViewTests(DashboardViewTestMixin, UserMixin, TestCase):
+    def setUp(self):
+        super(UserDetailViewTests, self).setUp()
+        self.switch, _created = Switch.objects.get_or_create(name='user_enrollments_on_dashboard', active=True)
```
jimabramson: nitpick: is it more idiomatic to use `__` for the unused var?

jimabramson:
@clintonb nitpick is optional. :+1:   thanks for adding the failure handling for the enrollment call.

ecommerce/extensions/dashboard/users/tests/test_views.py
clintonb: Updated.

PR #164, opened 2015-06-19
Merged OrderListVIew test classes

clintonb:
I discovered the overlapping test classes while working on #163.

FYI @rlucioni @jimabramson


rlucioni:
Nice catch, thanks. :+1:

PR #165, opened 2015-06-19
Course List Cleanup

clintonb:
- Removed template_name override, and updated template name
- Minor spacing cleanup
- Updated page header
- Corrected date sorting

XCOM-196

@Nickersoft @rlucioni @jimabramson


rlucioni:
:+1:

PR #166, opened 2015-06-19
Handle malformed/empty payer_info in Paypal execution response

jimabramson:
XCOM-418

@rlucioni @clintonb please review.

### DIFF
--- a/ecommerce/extensions/payment/processors/paypal.py
+++ b/ecommerce/extensions/payment/processors/paypal.py
@@ -176,7 +176,10 @@ class Paypal(BasePaymentProcessor):
         currency = payment.transactions[0].amount.currency
         total = Decimal(payment.transactions[0].amount.total)
         transaction_id = payment.id
+        # payer_info.email may be None, see:
+        # http://stackoverflow.com/questions/24090460/paypal-rest-api-return-empty-payer-info-for-non-us-accounts
         email = payment.payer.payer_info.email
+        label = 'PayPal ({})'.format(email) if email else 'PayPal Account'
 
         source = Source(
             source_type=source_type,
@@ -184,7 +187,7 @@ class Paypal(BasePaymentProcessor):
             amount_allocated=total,
             amount_debited=total,
             reference=transaction_id,
-            label=email,
+            label=label,
             card_type=None
         )
 
--- a/ecommerce/extensions/payment/tests/mixins.py
+++ b/ecommerce/extensions/payment/tests/mixins.py
@@ -181,6 +181,20 @@ class PaypalMixin(object):
     PAYMENT_ID = u'PAY-123ABC'
     PAYMENT_CREATION_STATE = u'created'
     PAYMENT_EXECUTION_STATE = u'approved'
+    PAYER_INFO = {
+        u'email': EMAIL,
+        u'first_name': u'test',
+        u'last_name': u'buyer',
+        u'payer_id': u'123ABC',
+        u'shipping_address': {
+            u'city': u'San Jose',
+            u'country_code': u'US',
+            u'line1': u'1 Main St',
+            u'postal_code': u'95131',
+            u'recipient_name': u'test buyer',
+            u'state': u'CA'
+        }
+    }
     RETURN_DATA = {
         u'paymentId': PAYMENT_ID,
         u'PayerID': PAYER_ID
@@ -278,7 +292,9 @@ class PaypalMixin(object):
 
         return payment_creation_response
 
-    def mock_payment_execution_response(self, basket, state=PAYMENT_EXECUTION_STATE):
+    def mock_payment_execution_response(self, basket, state=PAYMENT_EXECUTION_STATE, payer_info=None):
+        if payer_info is None:
+            payer_info = self.PAYER_INFO
         total = unicode(basket.total_incl_tax)
         payment_execution_response = {
             u'create_time': u'2015-05-04T15:55:27Z',
@@ -290,20 +306,7 @@ class PaypalMixin(object):
                 u'rel': u'self'
             }],
             u'payer': {
-                u'payer_info': {
-                    u'email': self.EMAIL,
-                    u'first_name': u'test',
-                    u'last_name': u'buyer',
-                    u'payer_id': u'123ABC',
-                    u'shipping_address': {
-                        u'city': u'San Jose',
-                        u'country_code': u'US',
-                        u'line1': u'1 Main St',
-                        u'postal_code': u'95131',
-                        u'recipient_name': u'test buyer',
-                        u'state': u'CA'
-                    }
-                },
+                u'payer_info': payer_info,
                 u'payment_method': u'paypal'
             },
             u'redirect_urls': {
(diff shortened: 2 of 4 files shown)


ecommerce/extensions/payment/tests/mixins.py
```diff
@@ -290,20 +304,7 @@ def mock_payment_execution_response(self, basket, state=PAYMENT_EXECUTION_STATE)
...
-                    u'shipping_address': {
-                        u'city': u'San Jose',
-                        u'country_code': u'US',
-                        u'line1': u'1 Main St',
-                        u'postal_code': u'95131',
-                        u'recipient_name': u'test buyer',
-                        u'state': u'CA'
-                    }
-                },
+                u'payer_info': payer_info if payer_info is not None else self.PAYER_INFO,
```
rlucioni: This would be easier to reason about if you defaulted the `payer_info` argument to `PAYER_INFO`, as is done for the `state` argument.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -176,7 +176,9 @@ def handle_processor_response(self, response, basket=None):
         currency = payment.transactions[0].amount.currency
         total = Decimal(payment.transactions[0].amount.total)
         transaction_id = payment.id
-        email = payment.payer.payer_info.email
+        # payer_info.email may be None, see:
+        # http://stackoverflow.com/questions/24090460/paypal-rest-api-return-empty-payer-info-for-non-us-accounts
+        label = payment.payer.payer_info.email or payment.id
```
rlucioni: A comment in the Oscar models recommends that `label` be a "customer-friendly label for the source." Since we have access to the user's `basket` at this stage, what about using the email address on the user associated with the `basket` as a fallback, instead of `payment.id`?

clintonb: Using that address will lead to confusion if people use separate accounts for edX and PayPal. Think back to the case where a student used his parent's account.

rlucioni: Good point. With that in mind, would it make sense to use `'PayPal'` or `'PayPal account'` as the default label? A payment ID won't make it clear to a customer that a PayPal account was used.

jimabramson: That may be the best we can do.  PayPal isn't sending anything else of
particular use.  When I get to updating this, I'll go with your suggestion
@renzo.

clintonb: Another alternative: set `card_type` to "paypal" and use the payment ID as the `label`.

clintonb:
Have we confirmed that subsequent calls to lookup the payment do not include shopper info?

jimabramson:
@clintonb I'm not sure what you mean, so, I don't think so.   could you elaborate?

ecommerce/extensions/payment/tests/mixins.py
jimabramson: I did not do that because [mutable default arguments are bad](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments).  The `payer_info` dict is not modified in this function so it's probably OK, but it's still a known antipattern.  I can push the default/fallback computation of this value up to the top of the function which would make it a little more obvious what's happening, does that sound like an acceptable compromise?

clintonb:
@jimabramson I am wondering if this is simply a problem with the API not sending the data, or PayPal not storing the data. We know that we do not currently receive the data as expected. If we query the API for the information, can we get it that way?

rlucioni:
@clintonb the first operation performed by our PayPal processor is to look up the payment. Performing the same operation a second time would likely expose us to the same issue, meaning we'd still require this kind of fallback logic.

ecommerce/extensions/payment/tests/mixins.py
rlucioni: I agree that mutable default arguments are bad. With that in mind, I'd just leave this as it is.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -176,7 +176,10 @@ def handle_processor_response(self, response, basket=None):
         currency = payment.transactions[0].amount.currency
         total = Decimal(payment.transactions[0].amount.total)
         transaction_id = payment.id
+        # payer_info.email may be None, see:
+        # http://stackoverflow.com/questions/24090460/paypal-rest-api-return-empty-payer-info-for-non-us-accounts
         email = payment.payer.payer_info.email
+        label = 'PayPal ({})'.format(email) if email else 'PayPal Account'
```
jimabramson, line 182: @clintonb @rlucioni what do you think of this solution?

ecommerce/extensions/payment/tests/test_processors.py
```diff
@@ -386,14 +386,20 @@ def test_approval_url_missing(self):
         self.assertRaises(GatewayError, self.processor.get_transaction_parameters, self.basket, request=self.request)
         self.assert_processor_response_recorded(self.processor.NAME, self.PAYMENT_ID, response, basket=self.basket)
 
-    @httpretty.activate
+    # the test case mixin defines this test, but since we're using ddt for it (see next method), the test runner
+    # tries to run the base test method (with name unmangled name) and raises NotImplemented.
     def test_handle_processor_response(self):
+        pass
```
jimabramson: i know this is gross - any ideas for alternative approaches welcome

ecommerce/extensions/payment/tests/test_views.py
```diff
@@ -315,7 +316,10 @@ def _assert_execution_redirect(self):
         self.processor.get_transaction_parameters(self.basket, request=self.request)
 
         creation_response = self.mock_payment_creation_response(self.basket, find=True)
-        execution_response = self.mock_payment_execution_response(self.basket)
+        execution_response = self.mock_payment_execution_response(
+            self.basket,
+            payer_info=payer_info if payer_info is not None else self.PAYER_INFO
```
jimabramson: just realized the `if` condition is no longer needed, will simplify

ecommerce/extensions/payment/processors/paypal.py
clintonb, line 182: Fine.

rlucioni, line 182: Works for me.

ecommerce/extensions/payment/tests/test_processors.py
clintonb: One alternative is to use a for loop inside the test:

``` python
def test_handle_processor_response(self):
  for payer_info in payer_infos:
    # Log payer_info being tested in case of failure
    # Test stuff
```

Either solution is acceptable to me.

clintonb:
:+1:

ecommerce/extensions/payment/tests/test_processors.py
rlucioni: I like this.

rlucioni:
:+1: once the tests are cleaned up.

jimabramson:
cleaned up tests, will merge once tests finish.

PR #167, opened 2015-06-22
Increase enrollment fulfillment timeout

rlucioni:
It's [considered good practice](http://docs.python-requests.org/en/latest/user/advanced/#advanced) to set connection timeouts to slightly larger than a multiple of 3, the default TCP packet retransmission window. Bumping the connection timeout from 5 to 7 will get us an extra retry we didn't have before.

This change is meant to alleviate enrollment revocation timeouts described in [XCOM-424](https://openedx.atlassian.net/browse/XCOM-424). I'll be following this up with LMS changes. @jimabramson or @clintonb, please review.


jimabramson:
:+1:

clintonb:
:+1:

clintonb:
Will configuration changes be needed? Is this value overridden?

rlucioni:
@clintonb no, this setting is not present in edx/configuration.

PR #168, opened 2015-06-22
Course list view now requires staff permissions to view

Nickersoft:
The list view of the course admin tool now requires that you be logged in and have staff member permissions in order to view. If you do not have staff permissions, you are redirected to a 404 page. If you are not logged in, you are redirected to the login page.

@clintonb @rlucioni FYI: @wedaly @jimabramson

### DIFF
--- a/ecommerce/courses/tests/test_views.py
+++ b/ecommerce/courses/tests/test_views.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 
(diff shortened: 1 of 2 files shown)


wedaly:
Not a blocker for this PR, but I'm concerned about using `is_staff` to include "PMs" in this context.  Do `is_staff` users also have access to financial data?

Nickersoft:
@wedaly That would be a question for @clintonb

clintonb:
@wedaly `is_staff` is a placeholder to temporarily secure this view. XCOM-368 will do this properly.

clintonb:
:+1:

rlucioni:
:+1:

PR #169, opened 2015-06-22
Upgraded to django-oscar 1.1

clintonb:
Did Discovery on a fresh installation, and existing. There don't appear to be any issues with the upgrade.

XCOM-439


clintonb:
@rlucioni @jimabramson @Nickersoft please review, but let's hold off on merging until we resolve outstanding production issues.

jimabramson:
@clintonb why don't we need the button styles anymore?

clintonb:
Oscar is no longer overriding the Bootstrap defaults, so we no longer need
to  "reset" them.

## 

Clinton Blackburn

rlucioni:
Looks good to me, :+1: . Onwards to Django 1.8!

jimabramson:
@clintonb one other thing, did you need to perform the category migration as noted [here](http://django-oscar.readthedocs.org/en/latest/releases/v1.1.html#backwards-incompatible-changes-in-1-1) ?  It seemed to imply that was mandatory - if so we should leave a note for ourselves to remember to run it perhaps.

clintonb:
@jimabramson Yes, I had to run an included migration. I did not, as I initially suspected, have to modify our existing category.

jimabramson:
i'm totally :+1: with this, assuming that `make migrate` will pick up anything that needs to be done.   could you stick "requires migration" at the end of the commit message or something?

PR #170, opened 2015-06-22
Upgrade to Django 1.8

clintonb:
XCOM-441

FYI @rlucioni @jimabramson @Nickersoft this is purely discovery.

### DIFF
--- a/ecommerce/extensions/payment/tests/test_processors.py
+++ b/ecommerce/extensions/payment/tests/test_processors.py
@@ -23,7 +23,6 @@ from paypalrestsdk.resource import Resource
 from ecommerce.core.constants import ISO_8601_FORMAT
 from ecommerce.courses.models import Course
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
-from ecommerce.extensions.payment import processors
 from ecommerce.extensions.payment.exceptions import (InvalidSignatureError, InvalidCybersourceDecision,
                                                      PartialAuthorizationError)
 from ecommerce.extensions.payment.models import PaypalWebProfile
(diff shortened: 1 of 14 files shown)


rlucioni:
@clintonb [relevant](https://docs.djangoproject.com/en/1.8/releases/1.8/#django-conf-urls-patterns).

TL;DR: `urlpatterns` should now be a list of `django.conf.urls.url()` instances.

clintonb:
Ready for review @rlucioni.

ecommerce/extensions/payment/tests/test_processors.py
```diff
@@ -101,9 +100,7 @@ def test_get_transaction_parameters(self):
         """ Verify the processor returns the appropriate parameters required to complete a transaction. """
 
         # Patch the datetime object so that we can validate the signed_date_time field
-        with mock.patch.object(processors.cybersource.datetime, u'datetime',
-                               mock.Mock(wraps=datetime.datetime)) as mocked_datetime:
```
rlucioni, line 105: Any specific reason to remove this? You're so against mocking `datetime`!

rlucioni:
:shipit: @clintonb .

ecommerce/extensions/payment/tests/test_processors.py
clintonb, line 105: I honestly don't care whether we mock `datetime` or not. Doing so is simply inconvenient. In this case, your mock caused issues with Django's sqlite3 library. Specifically, the mocked value was not returning a proper type value. This was the simplest workaround.

rlucioni, line 105: Cool, thanks for the clarification.

PR #171, opened 2015-06-23
Avoid revoking lines corresponding to $0 refunds

rlucioni:
Refunds corresponding to a total credit of $0 require no revocation. This change also prevents deadlocking with the LMS which occurs when Otto attempts to revoke an automatically-approved refund. For more context, please see [XCOM-424](https://openedx.atlassian.net/browse/XCOM-424).

### DIFF
--- a/ecommerce/extensions/fulfillment/tests/test_api.py
+++ b/ecommerce/extensions/fulfillment/tests/test_api.py
@@ -90,6 +90,20 @@ class FulfillmentApiTests(FulfillmentTestMixin, TestCase):
         self.assertEqual(refund.status, REFUND.PAYMENT_REFUNDED)
         self.assertEqual(set([line.status for line in refund.lines.all()]), {REFUND_LINE.COMPLETE})
 
+    @override_settings(FULFILLMENT_MODULES=[])
+    def test_suppress_revocation_for_zero_dollar_refund(self):
+        """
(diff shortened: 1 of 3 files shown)


ecommerce/extensions/fulfillment/tests/test_api.py
```diff
@@ -90,6 +90,19 @@ def test_revoke_fulfillment_for_refund(self):
         self.assertEqual(refund.status, REFUND.PAYMENT_REFUNDED)
         self.assertEqual(set([line.status for line in refund.lines.all()]), {REFUND_LINE.COMPLETE})
 
+    def test_suppress_revocation_for_zero_dollar_refund(self):
+        """
+        Verify that the function does not require use of fulfillment modules to mark lines in a refund
+        corresponding to a total credit of $0 as complete.
```
jimabramson, line 97: to explicitly reinforce what the docstring says, could you override settings in this method with `FULFILLMENT_MODULES=[]` ?

jimabramson:
@rlucioni lgtm, only request is the settings override.

ecommerce/extensions/fulfillment/tests/test_api.py
rlucioni, line 97: Updated.

jimabramson:
:+1:

PR #173, opened 2015-06-23
Display refund lines inline with their parent refund in the admin

rlucioni:
This is convenient, and is also required to fully manually correct refunds in a revocation error state.


clintonb:
:+1:

PR #174, opened 2015-06-23
Exercise CyberSource payment flow with domestic and international addresses

rlucioni:
We should be testing international addresses in addition to domestic addresses. Until now, we've only been testing domestic addresses; this is how the issue with the missing state field slipped past us. [XCOM-435](https://openedx.atlassian.net/browse/XCOM-435).


clintonb:
:+1:

PR #175, opened 2015-06-24
Added notes and command to aid local SASS development

clintonb:
XCOM-367

FYI @Nickersoft @rlucioni @jimabramson


rlucioni:
:+1:

PR #176, opened 2015-06-24
Improve performance of refund and processor response admin views

rlucioni:
Formerly, these views rendered `<select>` elements for every basket, user, and order line in the system. These changes avoid that overhead, and come with the added benefit of making sensitive fields which shouldn't be edited read-only.

### DIFF
--- a/ecommerce/extensions/payment/admin.py
+++ b/ecommerce/extensions/payment/admin.py
@@ -8,6 +8,8 @@ PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
 class PaymentProcessorResponseAdmin(admin.ModelAdmin):
     list_display = ('id', 'processor_name', 'transaction_id', 'basket', 'created')
 
+    readonly_fields = ('processor_name', 'transaction_id', 'basket', 'response')
+
 
 admin.site.register(PaymentProcessorResponse, PaymentProcessorResponseAdmin)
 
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/payment/admin.py
```diff
@@ -8,6 +8,8 @@
 class PaymentProcessorResponseAdmin(admin.ModelAdmin):
     list_display = ('id', 'processor_name', 'transaction_id', 'basket', 'created')
 
+    readonly_fields = ('processor_name', 'transaction_id', 'basket', 'response')
```
clintonb, line 11: @rlucioni Can you find a way to prettify the response? As is, it appears as a wall of text.

clintonb:
:+1: (would be nice to prettify response JSON)

ecommerce/extensions/payment/admin.py
rlucioni, line 11: @clintonb I agree that it would be nice to prettify the response. I looked into it briefly but decided the wall of text was passable for now. Since it bothers you, too, I'll take another quick look.

rlucioni, line 11: Turns out this isn't easy. I think I'm wasting my time on it now, so I'm going to go ahead and merge without it. If I can find a good way to nicely format the JSON, I'll address it in a separate PR.

PR #177, opened 2015-06-24
Add support for PayPal web experience profiles.

jimabramson:
XCOM-319

@rlucioni @clintonb FYI @Nickersoft 

I still need to add tests, but would like to start getting feedback on the overall approach.  PR consists of:
- Management command:
  - wrappers around PayPal API for list/create/show/update/delete.  Initially these seemed important as manually testing directly via the REST API is cumbersome.   However, most aren't actually part of any deployment workflow.  Since there's a substantial amount of code around this, I think it should have tests, but alternatively, we can just gut it and keep this a lot smaller.
  - `generate`, `enable`, and `disable` commands.  These, respectively: 1) generate and store with PayPal a web experience profile based on values in configuration; 2) save the profile in a django model that we use when setting up a PayPal payment; 3) delete the profile from that model.
- Add settings to store the brand name and logo image under a named profile key.
- Add the PayPalWebProfile model and migration.   Very simple.

There will be a configuration PR shortly that corresponds to this change.

### DIFF
--- a/ecommerce/extensions/payment/management/commands/paypal_profile.py
+++ b/ecommerce/extensions/payment/management/commands/paypal_profile.py
@@ -1,0 +1,172 @@
+import json
+import logging
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db.utils import IntegrityError
+from paypalrestsdk import WebProfile
+
+from ecommerce.extensions.payment.models import PaypalWebProfile
+
+
+log = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+
+    help = """Manage PayPal Web Experience Profiles.
+
+    Supported actions:
+        list                Print a JSON-formatted list of existing web profiles.
+        create [json]       Create a new profile using the specified JSON string.
+        show [id]           Print the contents of an existing web profile.
+        update [id] [json]  Update an existing profile using the specified JSON string.
+        delete [id]         Delete an existing profile. (Use -d to automatically disable when deleting.)
+        enable [id]         Enable the web profile in this Django application (send it in PayPal API calls).
+        disable [id]        Disable the web profile in this Django application (don't send in PayPal API calls).
+
+    The 'enable' and 'disable' actions are idempotent so it is safe to run them repeatedly in the same environment.
+    """
+    args = "action [id] [json]"
+
+    @staticmethod
+    def _get_argument(args, variable_name, action_name):
+        """
+        DRY helper.  Tries to pop the topmost value from `args` and raises a CommandError
+        with a formatted message in case of failure.  This function mutates `args` in place.
+        """
+        try:
+            return args.pop(0)
+        except IndexError:
+            raise CommandError("Action `{}` requires a {} to be specified.".format(action_name, variable_name))
+
+    def print_json(self, data):
+        self.stdout.write(json.dumps(data, indent=1, ensure_ascii=True))
+
+    def handle(self, *args, **options):
+        """
+        Main dispatch.
+        """
+        args = list(args)
+        action = None
+        try:
+            action = args.pop(0)
+            handler = getattr(self, 'handle_{}'.format(action))
+        except IndexError:
+            raise CommandError("no action specified.")
+        except AttributeError:
+            raise CommandError("unrecognized action: {}".format(action))
+        return handler(args)
+
+    def _do_create(self, profile_data):
+        """
+        Creates a new profile in the PayPal account with the specified id, using the specified data.
+        """
+        profile = WebProfile(profile_data)
+        result = profile.create()
+        if not result:
+            raise CommandError("Could not create web profile: {}".format(profile.error))
+        else:
+            log.info('Created profile "%s" (id=%s).', profile.name, profile.id)
+        return profile
+
+    def _do_update(self, profile_id, profile_data):
+        """
+        Updates the existing profile in the PayPal account with the specified id, replacing
+        all data with the specified data.
+        """
+        profile = WebProfile.find(profile_id)
+        result = profile.update(profile_data)
+        if not result:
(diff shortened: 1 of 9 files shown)


ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def handle(self, *args, **options):
+        """
+        Main dispatch.
+        """
+        args = list(args)
+        action = None
+        try:
+            action = args.pop(0)
+            handler = getattr(self, 'handle_{}'.format(action))
```
clintonb: A help message for these actions would be...helpful. :)

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -31,6 +32,7 @@ class Paypal(BasePaymentProcessor):
     """
 
     NAME = u'paypal'
+    WEB_PROFILE_NAME = 'openedx_default'
```
clintonb: Should this be a setting? There doesn't seem to be a way to override it.

jimabramson: There isn't.  At this stage I think there is no need for multiple profiles and so a hardcoded name would be fine.  It's partially a hedge against the fact that we may want to dynamically generate these in the future, or just pre-configure a limited amount of them - I wanted to avoid committing to either direction.

clintonb: Okay.

ecommerce/extensions/payment/models.py
```diff
@@ -25,5 +25,11 @@ class Meta(object):
 class Source(AbstractSource):
     card_type = models.CharField(max_length=255, choices=CARD_TYPE_CHOICES, null=True, blank=True)
 
+
+class PayPalWebProfile(models.Model):
```
rlucioni: Minor: for consistency with our `Paypal` payment processor class, I'd name this `PaypalWebProfile`.

jimabramson:
updated and ready for re-review.

ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -0,0 +1,229 @@
...
+    help = """Manage PayPal Web Experience Profiles.
+
+    Supported actions:
+        list                Print a JSON-formatted list of existing web profiles.
+        create [json]       Create a new profile using the specified JSON string.
+        show [id]           Print the contents of an existing web profile.
+        update [id] [json]  Update an existing profile using the specified JSON string.
+        delete [id]         Delete an existing profile. (Use -d to automatically disable when deleting.)
+        enable [id]         Enable the web profile in this Django application (send it in PayPal api calls).
+        disable [id]        Disable the web profile in this Django application (don't send in PayPal api calls).
```
rlucioni: Minor, but: s/api/API

rlucioni:
:+1:

jimabramson:
@rlucioni thanks, fixed caps with 'api'
@clintonb when you have time, lmk if the amended help string in the command module seems sufficient, or if you have additional concerns.

clintonb:
:+1:

jimabramson:
@clintonb pls review.

added: PaypalProfileAdminView
removed: settings related to profiles and the `generate` command intended to automate this stuff.  (see the last fixup commit to isolate what's changed since your last thumbsup)

jimabramson:
last commit fixed the quality issue

ecommerce/extensions/payment/views.py
```diff
@@ -216,3 +219,59 @@ def get(self, request):
         except:  # pylint: disable=bare-except
             logger.exception('Payment was received, but attempts to create an order for basket [%d] failed.', basket.id)
             return redirect(receipt_url)
+
+
+class PaypalProfileAdminView(View):
```
clintonb, line 224: It would be great if this were a proper REST API, but I won't push the matter.

ecommerce/extensions/payment/views.py
```diff
@@ -216,3 +219,59 @@ def get(self, request):
...
+
+    ACTIONS = ('list', 'create', 'show', 'update', 'delete')
```
clintonb: What about enable/disable?

ecommerce/extensions/payment/views.py
```diff
@@ -216,3 +219,59 @@ def get(self, request):
...
+        return super(PaypalProfileAdminView, self).dispatch(request, *args, **kwargs)
+
+    def get(self, request, *_args, **_kwargs):
+
+        # Capture all output and logging
+        out = StringIO()
+        err = StringIO()
+        log = StringIO()
+
+        root_logger = logging.getLogger()
```
clintonb: Use the existing global, `logger`.

ecommerce/extensions/payment/views.py
jimabramson, line 224: thanks :-)

it's already somewhat surreal that this exists at all, but following the example of the course migration, doing everything with GET really heightens the convenience factor.

ecommerce/extensions/payment/views.py
jimabramson: those can be done via django admin, so it seemed redundant to expose them here as well.

clintonb: My concern here is that we will still be reliant on DevOps by not exposing
the enable/disable actions.

## 

Clinton Blackburn

jimabramson: OK, updated.

ecommerce/extensions/payment/views.py
jimabramson: fixed.

clintonb:
:+1: (after squashing)

PR #178, opened 2015-06-25
Publishing course mode info to LMS

clintonb:
After a Course is saved, a call is made to the LMS to update the course modes table with Oscar's product info.

XCOM-315

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -1,8 +1,12 @@
 import logging
 
-from django.db import models
+from django.db import models, transaction
 from oscar.core.loading import get_model
 from simple_history.models import HistoricalRecords
+import waffle
+from ecommerce.courses.exceptions import PublishFailed
+
+from ecommerce.courses.publishers import LMSPublisher
 
 logger = logging.getLogger(__name__)
 ProductClass = get_model('catalogue', 'ProductClass')
@@ -12,6 +16,17 @@ class Course(models.Model):
     id = models.CharField(null=False, max_length=255, primary_key=True, verbose_name='ID')
     name = models.CharField(null=False, max_length=255)
     history = HistoricalRecords()
+
+    @transaction.atomic
+    def save(self, force_insert=False, force_update=False, using=None, update_fields=None):
+        super(Course, self).save(force_insert, force_update, using, update_fields)
+
+        if waffle.switch_is_active('publish_course_modes_to_lms'):
+            if not LMSPublisher().publish(self):
+                # Raise an exception to force a rollback
+                raise PublishFailed('Failed to publish {}'.format(self.id))
+        else:
+            logger.debug('Course mode publishing is not enabled. Commerce changes will not be published!')
 
     @classmethod
     def is_mode_verified(cls, mode):
--- a/ecommerce/courses/publishers.py
+++ b/ecommerce/courses/publishers.py
@@ -1,0 +1,64 @@
+import json
+import logging
+
+from django.conf import settings
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+class LMSPublisher(object):
+    def serialize_seat_for_commerce_api(self, seat):
+        """ Serializes a course seat product to a dict that can be further serialized to JSON. """
+        stock_record = seat.stockrecords.first()
+        return {
+            'name': seat.attr.certificate_type,
+            'currency': stock_record.price_currency,
+            'price': int(stock_record.price_excl_tax),
+            'sku': stock_record.partner_sku
+        }
+
+    def publish(self, course):
+        """ Publish course commerce data to LMS.
+
+        Uses the Commerce API to publish course modes, prices, and SKUs to LMS.
+
+        Args:
+            course (Course): Course to be published.
+
+        Returns:
+            True, if publish operation succeeded; otherwise, False.
+        """
+
+        if not settings.COMMERCE_API_URL:
+            logger.error('COMMERCE_API_URL is not set. Commerce data will not be published!')
+            return
+
+        modes = [self.serialize_seat_for_commerce_api(seat) for seat in course.seat_products]
+        course_id = course.id
+        data = {
+            'id': course_id,
+            'modes': modes
+        }
+
+        url = '{}/courses/{}/'.format(settings.COMMERCE_API_URL.rstrip('/'), course_id)
+        timeout = settings.COMMERCE_API_TIMEOUT
+
+        headers = {
+            'Content-Type': 'application/json',
+            'X-Edx-Api-Key': settings.EDX_API_KEY
+        }
+
+        try:
+            response = requests.put(url, data=json.dumps(data), headers=headers, timeout=timeout)
+            status_code = response.status_code
+            if status_code in (200, 201):
+                logger.info(u'Successfully published commerce data for [%s].', course_id)
+                return True
+            else:
+                logger.error(u'Failed to publish commerce data for [%s] to LMS. Status was [%d]. Body was [%s].',
+                             course_id, status_code, response.content)
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
+
+        return False
--- a/ecommerce/courses/tests/test_publishers.py
+++ b/ecommerce/courses/tests/test_publishers.py
@@ -1,0 +1,116 @@
+from decimal import Decimal
+import json
+
+import ddt
+from django.conf import settings
+from django.test import TestCase, override_settings
+from django_dynamic_fixture import G
+import httpretty
+import mock
+from oscar.test import factories
+from requests import Timeout
+
+from testfixtures import LogCapture
+
+from ecommerce.courses.models import Course
+from ecommerce.courses.publishers import LMSPublisher
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+
+EDX_API_KEY = 'edx'
+JSON = 'application/json'
+LOGGER_NAME = 'ecommerce.courses.publishers'
+
+
+@ddt.ddt
+@override_settings(COMMERCE_API_URL='http://example.com/commerce/api/v1/', EDX_API_KEY=EDX_API_KEY)
+class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
+    def setUp(self):
+        super(LMSPublisherTests, self).setUp()
+        self.course = G(Course)
+        self.create_course_seats(self.course.id, ('honor', 'verified'))
+        self.publisher = LMSPublisher()
+
+    def _mock_commerce_api(self, status, body=None):
+        self.assertTrue(httpretty.is_enabled, 'httpretty must be enabled to mock Commerce API calls.')
+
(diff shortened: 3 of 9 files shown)


ecommerce/courses/signals.py
```diff
@@ -0,0 +1,60 @@
...
+        return
+
+    modes = [serialize_seat_for_commerce_api(seat) for seat in instance.seat_products]
+    course_id = instance.id
+    data = {
+        'id': course_id,
+        'modes': modes
+    }
+
+    url = '{}/courses/{}/'.format(settings.COMMERCE_API_URL.rstrip('/'), course_id)
```
rlucioni: Can we not use `urljoin()` here because the regex used the match course IDs on the LMS can't deal with URL-encoded course IDs?

ecommerce/settings/base.py
```diff
@@ -231,6 +231,10 @@ def get_lms_url(path):
 # URL to which enrollment requests should be made
 ENROLLMENT_API_URL = None
 
+# Commerce API settings used for publishing information to LMS.
+COMMERCE_API_TIMEOUT = 7
```
rlucioni, line 235: This is essentially the same as the `ENROLLMENT_FULFILLMENT_TIMEOUT` setting. We could simplify by replacing both with a single `LMS_REQUEST_TIMEOUT` setting.

ecommerce/courses/signals.py
clintonb: `urljoin` will also remove any existing path info. I do not like `urljoin` as it introduces more issues than it resolves.

ecommerce/courses/signals.py
```diff
@@ -0,0 +1,60 @@
+import json
+import logging
+
+from django.conf import settings
+
```
rlucioni: Nitpick: Unnecessary newline.

ecommerce/courses/tests/test_signals.py
```diff
@@ -0,0 +1,128 @@
+from decimal import Decimal
+import json
+
+import ddt
+from django.conf import settings
+
```
rlucioni: Nitpick: Unnecessary newline.

ecommerce/courses/tests/test_signals.py
```diff
@@ -0,0 +1,128 @@
...
+import httpretty
+import mock
+from oscar.test import factories
+from requests import Timeout
+from testfixtures import LogCapture
+from waffle import Switch
+
+from ecommerce.courses.models import Course
+from ecommerce.courses.signals import publish_to_lms, serialize_seat_for_commerce_api
+
```
rlucioni: Nitpick (last one, I promise): Unnecessary newline.

ecommerce/settings/base.py
clintonb, line 235: Let me think about that. There are advantages to keeping these separate.

ecommerce/courses/signals.py
rlucioni: Could you give me an example of the behavior you're describing? I've found `urljoin` to be pretty helpful, and I'd like to know when to avoid it.

ecommerce/courses/tests/test_signals.py
```diff
@@ -0,0 +1,126 @@
...
+    @ddt.data('', None)
+    def test_commerce_api_url_not_set(self, setting_value):
+        """ If the Commerce API is not setup, the method should log an INFO message and return """
+        with override_settings(COMMERCE_API_URL=setting_value):
+            with LogCapture(LOGGER_NAME) as l:
+                publish_to_lms(Course, self.course)
+                l.check((LOGGER_NAME, 'INFO', 'COMMERCE_API_URL is not set. Commerce changes will not be published!'))
+
+    @httpretty.activate
```
rlucioni: Does `httpretty` need to be active here? `requests.put` is being patched.

ecommerce/courses/signals.py
clintonb: `COMMERCE_API_URL` includes a path in addition to a hostname (`http://example.com/commerce/api/v1/`). `urljoin` strips the path, and only retains the host.

``` python
api_url = 'http://example.com/commerce/api/v1/'
new_url = urljoin(api_url, 'courses/')
print url    # http://example.com/courses/
```

ecommerce/courses/tests/test_signals.py
clintonb: Removed.

ecommerce/courses/signals.py
rlucioni: TIL, thanks.

rlucioni:
:+1: . Let me know what you decide to do about the timeout settings.

ecommerce/courses/signals.py
```diff
@@ -0,0 +1,59 @@
...
+        'currency': stock_record.price_currency,
+        'price': int(stock_record.price_excl_tax),
+        'sku': stock_record.partner_sku
+    }
+
+
+@receiver(post_save, sender=Course)
+def publish_to_lms(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    if not waffle.switch_is_active('publish_course_modes_to_lms'):
+        logger.debug('Course mode publishing is not enabled. Commerce changes will not be published!')
```
jimabramson: should this be INFO ?

ecommerce/settings/local.py
```diff
@@ -78,6 +78,8 @@
 LMS_DASHBOARD_URL = get_lms_url('/dashboard')
 
 OAUTH2_PROVIDER_URL = get_lms_url('/oauth2')
+
+COMMERCE_API_URL = get_lms_url('commerce/api/v1/')
```
jimabramson: though it's functionally equivalent, i think it would be clearer and more consistent to lead with a slash like the others do.

ecommerce/courses/signals.py
```diff
@@ -0,0 +1,59 @@
...
+    try:
+        response = requests.put(url, data=json.dumps(data), headers=headers, timeout=timeout)
+        status_code = response.status_code
+        if status_code in (200, 201):
+            logger.info(u'Successfully published commerce data for [%s].', course_id)
+        else:
+            logger.error(u'Failed to publish commerce data for [%s] to LMS. Status was [%d]. Body was [%s].',
+                         course_id, status_code, response.content)
+    except Exception:  # pylint: disable=broad-except
+        logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
```
jimabramson: If we can't publish the mode because of an error, it doesn't seem like the user will find out.  It seems like the impact of Otto and LMS being out of sync could be bad - should we let the exception re-raise and fail the whole save?  If not, how do we recover / retry the push?

ecommerce/courses/signals.py
clintonb: Not really. That would get noisy for installations that actually want the functionality disabled.

ecommerce/courses/signals.py
clintonb: Many hours later (due to distractions)...this has been updated. I have replaced signals with an override of the model's save method to ensure atomicity.

ecommerce/settings/local.py
clintonb: Updated.

ecommerce/courses/publishers.py
```diff
@@ -0,0 +1,64 @@
...
+            'price': int(stock_record.price_excl_tax),
+            'sku': stock_record.partner_sku
+        }
+
+    def publish(self, course):
+        """ Publish course commerce data to LMS.
+
+        Uses the Commerce API to publish course modes, prices, and SKUs to LMS.
+
+        Argus:
```
jimabramson: was that meant to be Args?

ecommerce/courses/publishers.py
```diff
@@ -0,0 +1,64 @@
...
+            course (Course): Course to be published.
+
+        Returns:
+            True, if publish operation succeeded; otherwise, False.
+        """
+
+        if not settings.COMMERCE_API_URL:
+            logger.info('COMMERCE_API_URL is not set. Commerce changes will not be published!')
```
jimabramson: if we get here, it means the waffle switch is enabled but we aren't configured to do what we need to.  should this be a WARN or ERROR in that case?

ecommerce/courses/publishers.py
clintonb: Fixed.

ecommerce/courses/publishers.py
clintonb: Updated to error.

ecommerce/courses/tests/test_publishers.py
```diff
@@ -0,0 +1,116 @@
...
+        httpretty.register_uri(httpretty.PUT, url, status=status, body=json.dumps(body),
+                               content_type=JSON)
+
+    @ddt.data('', None)
+    def test_commerce_api_url_not_set(self, setting_value):
+        """ If the Commerce API is not setup, the method should log an INFO message and return """
+        with override_settings(COMMERCE_API_URL=setting_value):
+            with LogCapture(LOGGER_NAME) as l:
+                self.publisher.publish(self.course)
+                l.check((LOGGER_NAME, 'INFO', 'COMMERCE_API_URL is not set. Commerce changes will not be published!'))
```
jimabramson: s/Commerce changes/course changes/ ?

ecommerce/courses/models.py
```diff
@@ -13,6 +16,15 @@ class Course(models.Model):
     name = models.CharField(null=False, max_length=255)
     history = HistoricalRecords()
 
+    @transaction.atomic
```
jimabramson, line 20: The publish() function catches all exceptions, so this function will always commit regardless of whether publishing succeeded or not.  Is that the intention or am I reading it wrong?

clintonb, line 20: Oops. The purpose of `publish()` returning a boolean is to trigger an exception in `Courses.save()`. Updated.

ecommerce/courses/models.py
```diff
@@ -13,6 +17,17 @@ class Course(models.Model):
...
+    def save(self, force_insert=False, force_update=False, using=None, update_fields=None):
+        super(Course, self).save(force_insert, force_update, using, update_fields)
+
+        if waffle.switch_is_active('publish_course_modes_to_lms'):
+            if not LMSPublisher().publish(self):
+                # Raise an exception to force a rollback
+                raise PublishFailed
```
rlucioni: Should this exception be handled? As it stands, it looks like it will bubble up and result in a 500.

clintonb: The model doesn't care how you handle the exception. That exception handling should be done by the consuming views (which don't yet exist).

rlucioni: Got it, thanks.

jimabramson: @clintonb cool. optional: adding a message to the exception that contains the course_id might be helpful for diagnosing an exception after the fact.

jimabramson:
@clintonb :+1: whether or not you decide to add an exception message.

ecommerce/courses/models.py
clintonb: Updated.

PR #179, opened 2015-06-26
Improve logging of transaction lifecyle

rlucioni:
Adds standardized audit logging for successful basket freezing, payment receipt, order placement, line fulfillment, refund creation, payment credit, and line revocation. [XCOM-426](https://openedx.atlassian.net/browse/XCOM-426).

@jimabramson or @clintonb, please review when able.

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 from functools import wraps
 import logging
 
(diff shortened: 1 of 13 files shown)


ecommerce/extensions/analytics/utils.py
```diff
@@ -55,24 +56,27 @@ def wrapper(*args, **kwargs):  # pylint: disable=missing-docstring
...
+        Indefinite. Keyword arguments are strung together as comma-separated key-value
+        pairs alphabetically ordered by key in the resulting log message.
 
-    This should be called immediately after the order has been saved to db (i.e. after `handle_order_placement`).
+    Returns:
+        None
     """
-    logger.info('payment_applied: amount="%s", basket_id="%s", user_id="%s"', amount, basket_id, user_id)
+    d = OrderedDict(sorted(kwargs.items(), key=lambda i: i[0]))
+    payload = u''.join(['{k}="{v}", '.format(k=k, v=v) for k, v in d.iteritems()]).strip(', ')
```
clintonb: What is this supposed to do?

ecommerce/extensions/checkout/mixins.py
```diff
@@ -36,10 +37,28 @@ def handle_payment(self, response, basket):
...
+            basket_id=basket.id,
+            currency=source.currency,
+            processor_name=payment_event.processor_name,
+            reference=payment_event.reference,
+            user_id=basket.owner.id
+        )
 
     def handle_successful_order(self, order):
-        # Send a signal so that receivers can perform relevant tasks (e.g. fulfill the order).
+        """Send a signal so that receivers can perform relevant tasks (e.g., fulfill the order)."""
```
clintonb, line 52: The comma is not necessary.

ecommerce/extensions/checkout/tests/test_mixins.py
```diff
@@ -28,33 +28,45 @@ def setUp(self):
         self.user = UserFactory()
         self.order = self.create_order()
 
-    def test_handle_payment_logging(self, __mock_track):
+    def test_handle_payment_logging(self, _):
```
clintonb: Use a double-underscore to avoid potential conflicts with the ugettext alias.

ecommerce/extensions/analytics/utils.py
```diff
@@ -55,24 +56,27 @@ def wrapper(*args, **kwargs):  # pylint: disable=missing-docstring
     return decorator
 
 
-def log_payment_received(processor_name, reference, amount, basket_id):
-    """Emit a log entry indicating that we were notified about a successful payment.
+def audit_log(name, **kwargs):
+    """Variadic DRY helper used to emit an INFO-level log message.
```
clintonb: It's great that we know the definition of this word. Let's simplify it, please. I shouldn't need to a dictionary to read a docstring.

ecommerce/extensions/checkout/mixins.py
rlucioni, line 52: Five of six major style guides recommend the use of a comma after "i.e." and "e.g." [Reference](http://english.stackexchange.com/a/16215).

ecommerce/extensions/analytics/utils.py
rlucioni: This joins keys and values from the sorted dictionary above with an "=", wraps each value in quotes, and separates each pair with a comma and a space. The trailing comma and space are stripped. I'll add a comment.

ecommerce/extensions/checkout/tests/test_mixins.py
rlucioni: Updated.

ecommerce/extensions/analytics/utils.py
clintonb: Join on `u', '` and you don't have to strip anything:

`````` python
u', '.join(['{key}="{value}"'.format(key=key, value=value) for key, value in d.iteritems()])```
``````

ecommerce/extensions/analytics/utils.py
rlucioni: I'm going to push back on this. Describing the function as variadic here is the most correct and concise way to describe its behavior. The docstring later explains that the number of keyword arguments accepted is indefinite. I'm generally in favor of simplifying, but not when doing so would result in a less concise, more roundabout description. If the reader doesn't recognize the word, they'll benefit from looking it up and learning how it's meant to be used.

ecommerce/extensions/analytics/utils.py
rlucioni: Nice, I overlooked that. Updated.

ecommerce/extensions/analytics/utils.py
rlucioni: Resolved offline. I have my convictions, but I'll concede for the sake of clarity here.

clintonb:
:+1:

PR #180, opened 2015-06-26
Gracefully handling checkout errors during basket creation

clintonb:
If an exception is raised during checkout initiation, the view will log it and return HTTP status 500. This ensures that all payment processor responses are recorded, while not having to sacrifice the benefits of autocommit.

XCOM-418


rlucioni:
:+1:

PR #181, opened 2015-06-26
Added edX header w/ dropdown to list view for course admin tool

Nickersoft:
@clintonb @jimabramson @AlasdairSwan FYI: @rlucioni @wedaly @zubair-arbi


clintonb:
Squash the commits.

ecommerce/core/tests.py
```diff
@@ -0,0 +1,21 @@
...
+
+class ContextProcessorTests(UserMixin, TestCase):
+
+    def test_get_settings(self):
+        """ Verify that the correct settings are returned from the processor. """
+
+        user = self.create_user()
+
+        request = RequestFactory().get('/courses')
+        request.user = user
```
clintonb, line 17: You need to also test for anonymous users.

ecommerce/settings/base.py
```diff
@@ -151,10 +151,12 @@
 # TEMPLATE CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#template-context-processors
 TEMPLATE_CONTEXT_PROCESSORS = (
+    'core.context_processors.get_settings',
     'django.contrib.auth.context_processors.auth',
     'django.core.context_processors.debug',
     'django.core.context_processors.i18n',
     'django.core.context_processors.media',
+    'django.core.context_processors.request',
```
clintonb: Why is this necessary?

ecommerce/static/sass/_base.scss
```diff
@@ -1,19 +1,103 @@
...
+
+// Header
+// -----------
+
+.nav {
+
+    .open > a,
+    .open > a:hover,
+    .open > a:focus {
+        border-color:#C8C8C8;
```
clintonb: Where did all of this come from? Colors should be variables.

ecommerce/static/sass/_base.scss
```diff
@@ -51,6 +135,112 @@ body {
   }
 }
 
-.navbar {
-  border-bottom: 4px solid $navbar-default-border;
+// Conditional Stylings
+// --------------------
+
+// Small Screens
+@media only screen and (min-width: 320px) {
```
clintonb: Use bootstrap's variables (e.g. `$screen-xs`).

ecommerce/static/sass/main.scss
```diff
@@ -1,3 +1,7 @@
+// web
+// --------------------
+@import url(http://fonts.googleapis.com/css?family=Open+Sans:700,300,600,800,400);
```
clintonb: This won't work for China, which blocks Google. Follow the behavior of Insights (https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/static/sass/_config-assets.scss).

ecommerce/templates/edx/base.html
```diff
@@ -19,8 +19,46 @@
...
+        <span class="sr-only">Toggle navigation</span>
+        <span class="icon-bar"></span>
+        <span class="icon-bar"></span>
+        <span class="icon-bar"></span>
+      </button>
+      <div class="navbar-brand">
+        <img class="site-logo image-style-none" itemprop="logo" typeof="foaf:Image" src="https://www.edx.org/sites/default/files/theme/edx-logo-header.png" alt="edX">
+      </div>
+    </div>
+    <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
```
clintonb: Remove this ID.

ecommerce/templates/edx/base.html
```diff
@@ -19,8 +19,46 @@
...
+  <div class="container">
+    <div class="navbar-header">
+      <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#bs-example-navbar-collapse-1" aria-expanded="false">
+        <span class="sr-only">Toggle navigation</span>
+        <span class="icon-bar"></span>
+        <span class="icon-bar"></span>
+        <span class="icon-bar"></span>
+      </button>
+      <div class="navbar-brand">
+        <img class="site-logo image-style-none" itemprop="logo" typeof="foaf:Image" src="https://www.edx.org/sites/default/files/theme/edx-logo-header.png" alt="edX">
```
clintonb: The logo should never be hardcoded. Use an asset link.

ecommerce/static/sass/_config-variables.scss
```diff
@@ -26,7 +28,8 @@ $border-radius-base: 3px;
 $border-radius-large: 3px;
 $border-radius-small: 3px;
 
-$navbar-default-bg: white;
-
-// TODO Use the correct color
+// Navigation bar
 $navbar-default-border: $brand-primary-accent;
+$navbar-default-bg: rgb(255, 255, 255);
```
clintonb: Just use `white`.

ecommerce/static/sass/_base.scss
clintonb: Also, where is all of this code coming from? Did you copy-paste from Bootstrap?

Nickersoft: @clintonb Nope, these were just the required overrides necessary to create the desired style. I noticed a lot of the provided variable overrides in _config-variables didn't modify what was needed (most only modified general colors), so I had to code it by hand. I will add all of the colors into new variables though. I just hadn't because I thought _config-variables was reserved strictly for Bootstrap SASS variables that were being overwritten.

Nickersoft: @clintonb Also seeing now that a few variables that do cover what I coded were overlooked, so I'll remove the corresponding CSS and move the attribute to a Bootstrap variable.

ecommerce/templates/edx/base.html
```diff
@@ -19,8 +19,46 @@
...
+                  <i class="icon fa fa-home"></i>
+                  <div>{{ username }}</div>
+              </div>
+              <div class="arrow">
+                  <span class="caret"></span>
+              </div>
+          </a>
+          <ul class="dropdown-menu">
+            <li><a href="{{ lms_dashboard_url }}">Dashboard</a></li>
+            <li><a href="{% url 'customer:logout' %}">Logout</a></li>
```
jimabramson: i think it's expected, but clicking this puts me on the oscar dashboard and doesn't log me out of anything.  If I go to the lms dashboard first and then do the logout, it logs me out of LMS, but i'm still logged into Otto and can still access /courses/ there until I logout there a second time.

jimabramson:
@Nickersoft code lgtm apart from issues noted by @clintonb.  Would like to confirm the expected behavior of logout.

The mobile-style dropdown toggle seems to kick in fairly early when i resize the window width, and the open state needs to be expanded separately, but i assume this is expected as well.

screen shot 2015-06-28 at 9 59 05 am

Nickersoft:
@jimabramson Yes it is, as this behavior is controlled by Bootstrap directly. I'll make the other changes then let you guys know when it's ready for a second review.

ecommerce/templates/edx/base.html
Nickersoft: @clintonb I changed this ID  to something more suitable instead of removing it. It is required for the mobile dropdown to work properly.

ecommerce/templates/edx/base.html
Nickersoft: @jimabramson Is there a recommended fix for this? I was under the impression this logs the user out of both. @clintonb can you offer any insight?

ecommerce/settings/base.py
Nickersoft: @clintonb Nothing was added. I just sorted the list alphabetically. All of the context processors listed here are required as far as I know. context_processors.request was included here when I first started working on the codebase.

clintonb: Please revert this change, and avoid unnecessary sorts. It's not a big deal in this case; but, had you sorted middleware for example, you could have introduced wanted issues.

ecommerce/templates/edx/base.html
clintonb: Remove this for now. We will restore it once we have single logout.

ecommerce/core/tests.py
```diff
@@ -0,0 +1,29 @@
...
+
+        user = self.create_user()
+        anon = AnonymousUser()
+
+        request = RequestFactory().get('/courses')
+        request.user = user
+
+        settings_dict = get_settings(request)
+
+        self.assertEqual(settings_dict['username'], user.username)
```
clintonb: Use `assertDictEqual`.

ecommerce/core/tests.py
```diff
@@ -0,0 +1,29 @@
...
+        self.assertEqual(settings_dict['lms_dashboard_url'], settings.LMS_DASHBOARD_URL)
+
+        request.user = anon
```
clintonb: `request.user = AnonymousUser()`

ecommerce/core/tests.py
```diff
@@ -0,0 +1,29 @@
...
+
+class ContextProcessorTests(UserMixin, TestCase):
+
+    def test_get_settings(self):
+        """ Verify that the correct settings are returned from the processor. """
+
+        user = self.create_user()
+        anon = AnonymousUser()
+
+        request = RequestFactory().get('/courses')
```
clintonb: Use `.get('/')` since this app does not depend on the courses app.

ecommerce/static/sass/main.scss
```diff
@@ -1,3 +1,7 @@
+// web
```
clintonb: This comment isn't very helpful. Any reason why you opted not to use config-assets as we do in edx-analytics dashboard?

ecommerce/core/tests.py
clintonb, line 17: @Nickersoft Remember to actually acknowledge that you addressed feedback; otherwise, things slip through the cracks.

ecommerce/static/sass/main.scss
Nickersoft: @clintonb That was left-over from the previous Google fonts implementation. I'll change it to correspond with the edx-analytics dashboard.

Nickersoft:
@clintonb @jimabramson Updated

ecommerce/core/tests.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_get_settings(self):
+        """ Verify that the correct settings are returned from the processor. """
+
+        user = self.create_user()
+
+        request = RequestFactory().get('/')
+        request.user = user
+
+        settings_dict = get_settings(request)
```
clintonb: actual

ecommerce/core/tests.py
```diff
@@ -0,0 +1,30 @@
...
+        expected_dict = {
```
clintonb: expected

ecommerce/static/sass/_config-variables.scss
```diff
@@ -26,7 +30,27 @@ $border-radius-base: 3px;
 $border-radius-large: 3px;
 $border-radius-small: 3px;
 
+// Navigation bar
 $navbar-default-bg: white;
-
-// TODO Use the correct color
 $navbar-default-border: $brand-primary-accent;
+$navbar-default-link-color: rgb(60, 60, 60);
```
clintonb: Where are you getting all of these values? There seem to be a number of duplicates. Is it your intention to change the colors for _all_ dropdowns?

ecommerce/static/sass/_base.scss
```diff
@@ -1,19 +1,98 @@
 // app styling: base
 // ====================
 // NOTE: used for any basic styling or overriding that isn't covered by Bootstrap or theming
-body {
-  // Make room for the navbar.
-  padding-top: $navbar-height + $navbar-margin-bottom;
+@import "mixins";
```
clintonb: Mixins should be imported in main.scss.

ecommerce/static/sass/_base.scss
```diff
@@ -51,6 +130,104 @@ body {
   }
 }
 
-.navbar {
-  border-bottom: 4px solid $navbar-default-border;
+// Conditional Stylings
+// --------------------
+
+// Small Screens
+@media only screen and (max-width: $screen-sm-min) {
```
clintonb: You've added a lot of styles. Please move them to a separate file, _navbar.scss.

ecommerce/static/sass/_config-assets.scss
```diff
@@ -0,0 +1,9 @@
+// config: assets
+// ====================
+// NOTE: these are external assets needed for rendering (typefaces, sprites, images, etc.)
+
+// + typefaces
```
clintonb: Unnecessary.

Nickersoft:
@clintonb @jimabramson Updated

ecommerce/settings/base.py
clintonb: @Nickersoft this needs to be addressed.

ecommerce/static/sass/_base.scss
```diff
@@ -1,19 +1,21 @@
 // app styling: base
 // ====================
 // NOTE: used for any basic styling or overriding that isn't covered by Bootstrap or theming
-body {
-  // Make room for the navbar.
-  padding-top: $navbar-height + $navbar-margin-bottom;
+
+* {
```
clintonb: What does this do?

ecommerce/static/sass/_config-variables.scss
```diff
@@ -26,7 +30,27 @@ $border-radius-base: 3px;
...
 $border-radius-small: 3px;
 
+// Navigation bar
 $navbar-default-bg: white;
-
-// TODO Use the correct color
 $navbar-default-border: $brand-primary-accent;
+
+// Navigation bar (mobile)
+$navbar-mobile-border: rgb(229, 232, 234);
```
clintonb: Why does the mobile navbar need a special border?

ecommerce/static/sass/_config-variables.scss
```diff
@@ -26,7 +30,27 @@ $border-radius-base: 3px;
...
+
+// Dropdown button
+$dropdown-btn-bg: white;
+$dropdown-btn-border: rgb(200, 200, 200);
```
clintonb, line 39: For the third time: where are these values coming from? Did you just make up the colors?

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,178 @@
+// app styling: navbar
+// ====================
+// Styling used for the navbar
+
+// Base Styling
+// ---------------
+
+.nav {
+    &.navbar-default {
+        padding:20px 10px;
```
clintonb, line 10: As stated previously, these values need to be captured as variables. Is there a reason you've deviated from the default Bootstrap values/variables?

clintonb, line 10: The primary advantage of using Bootstrap is that we can quickly restyle it by changing a few variables. This seems to break that behavior for downstream users.

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,178 @@
...
+                font-size:16px;
+                padding:3px 5px 4px;
+                border:1px solid transparent;
+                margin: 5px 0px;
+            }
+        }
+    }
+}
+
+@media (max-width: ($screen-md-min - 1)) {
```
clintonb: What type of screen is this supposed to cover? What doesn't the medium screen definition cover it?

ecommerce/static/sass/_config-variables.scss
Nickersoft, line 39: @clintonb All of these colors were taken directly from the current implementation of the edX dropdown, found at [https://courses.edx.org/dashboard](https://courses.edx.org/dashboard).

clintonb:
@frrrances please review

ecommerce/static/sass/_navbar.scss
Nickersoft, line 10: @clintonb The current implementation of the dropdown on the edX dashboard goes against Bootstrap conventions. In order to make the style work, the HTML DOM hierarchy needed to be reordered, and as a result, custom styles need to be put in to place. Certain attributes, like navbar padding, either do not exist or are configured in a which such that changing them has no affect on the highly-modified UI. With the case of navbar padding, this seems to be the latter. I initially were going to make all of these variables, but there did not seem to be enough code reuse to call for variable declarations, plus I didn't want to clutter the _config-variables file.

ecommerce/static/sass/_navbar.scss
Nickersoft: @clintonb I can change it. I put it this way to model it after the current implementation, in which the media query used a min of 768px, and the max used 787px. Not sure why this was the case, but I kept it for accuracy's sake and to prevent anything from breaking. I can remove the '-1' though.

ecommerce/static/sass/_config-variables.scss
Nickersoft: @clintonb Because the default behavior is to take the wide blue bottom border of the header and shrink it down to 1px wide, and leave it as a divider between the dropdown and the rest of the header. In other words, it didn't stay at the bottom of the mobile header, it cut right through it. Thinking that a bright blue line dividing a header in half wasn't that aesthetically appealing, I changed the color to a light gray so that it would appear as more of a divider. Actually, this might be the only custom color on the whole page. I can rewrite it to match the default divider color of the dropdown instead.

ecommerce/static/sass/_base.scss
Nickersoft: @clintonb This prevents Chrome and other webkit browsers from drawing their own blue glowing outline around buttons, text boxes, etc. A single asterisk (*) represents every element on the page.

clintonb: Why would you intentionally disable that? That functionality is quite useful for determining which elements have focus, especially if users prefer keyboards over mice.

frrrances: This is not a great way to do this and it breaks accessibility. Please leave the outline and scope changes to particular elements/classes (preferably classes).

Nickersoft: @clintonb @frrrances Usually it is overwritten when a custom focus styling is in place and has been I problem I've run into in the past, so I just wanted to avoid the issue. Looking at it now, however, I see that this was implemented to suppress the default hover action of Bootstrap on the dropdown, which was to set a dotted outline around the element. I'll narrow the scope of this style to be specific to the dropdown.

ecommerce/settings/base.py
Nickersoft: @clintonb Somehow I missed your respond to my comment. I'll get right on it.

Nickersoft:
@clintonb @frrrances @jimabramson Updated

clintonb:
@AlasdairSwan can you also review, please?

AlasdairSwan:
@clintonb as Q4 rapidly nears an end I am a tad busy. Can the PR wait until tomorrow?

clintonb:
@AlasdairSwan yes.

jimabramson:
I'm wondering if the logo image belongs in an edx-theme subdirectory, and there should be a default-theme subdirectory with the generic logo image.  I'm not sure if that's our long-term theming approach but it is consistent with LMS.

jimabramson:
whether or not you implement it in this PR, here are recommendations from @singingwolfboy about how to organize the assets in a theming-ready manner.

create a directory tree in your project: themes/edx.org/static and put the edX logo in that directory, with exactly the same name as the default, generic logo that you use in the core static directory
then, you can have a setting that determines what theme, if any, you are using
and if you're using the edx.org theme, insert that themes/edx.org/static directory at the front of the static file lookup path
if there's a way to do that
it's not great, because that theme is specific to your application, but it's better than nothing, and goes along with the ideas behind the comprehensive theming project

AlasdairSwan:
The Sass folder needs to be structured appropriately. I would suggest following the Pattern Library structure of base, components and utilities folders (and possibly adding a vendor one too) https://github.com/edx/ux-pattern-library/tree/master/_src/pattern-library/sass

ecommerce/static/sass/_base.scss
```diff
@@ -12,8 +11,7 @@ body {
   &:focus,
   &.focus,
   &:active,
-  &.active,
-  .open > &.dropdown-toggle {
+  &.active {
```
AlasdairSwan, line 14: What about hover?

ecommerce/static/sass/_base.scss
```diff
@@ -50,7 +48,3 @@ body {
     opacity: 0.50;
   }
 }
```
AlasdairSwan, line 50: This should be moved to its own _buttons.scss partial in the components folder

ecommerce/static/sass/_config-assets.scss
```diff
@@ -0,0 +1,7 @@
+// config: assets
+// ====================
+// NOTE: these are external assets needed for rendering (typefaces, sprites, images, etc.)
+
+// typefaces
+// --------------------
+@import url(https://files.edx.org/fonts/edx-fonts/open-sans/open-sans.css);
```
AlasdairSwan, line 7: I suggest moving this to a _fonts.scss partial in the base folder

ecommerce/static/sass/_config-variables.scss
```diff
@@ -26,7 +30,24 @@ $border-radius-base: 3px;
...
+$dropdown-bg: white;
+$dropdown-border: rgb(100, 100, 100);
+$dropdown-divider: $dropdown-btn-border;
+$dropdown-fallback-border: $dropdown-border;
+$dropdown-link-color: rgb(0, 120, 176);
+$dropdown-link-hover-color: rgb(60, 60, 60);
+$dropdown-link-hover-bg: rgba(0, 0, 0, 0);
+$dropdown-link-active-color: $dropdown-link-hover-color;
+$dropdown-link-active-bg: $dropdown-link-hover-bg;
+$dropdown-shadow: $dropdown-border;
```
AlasdairSwan, line 53: Is this partial overriding Bootstrap variables?

ecommerce/static/sass/_navbar.scss
AlasdairSwan, line 10: I recommend using the JS @clrux recently wrote for the nav dropdown for accessibility and consistency reasons, but this would require writing Sass to style it. If you are sticking with Bootstrap for this PR then I don't see any value in making the padding values variables.

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+        border-bottom: {
+            width:4px;
+            style: solid;
+        }
+        .navbar-brand {
+            width:70px;
+            height:auto;
+            margin:auto;
+            float:none;
+            padding:0px;
```
AlasdairSwan, line 20: For Sass/CSS declarations that are 0 you don't need a unit (micro-optimization)

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+    }
+    .navbar-nav {
+        & > .dropdown {
+            border:1px solid $dropdown-btn-border;
+            border-radius:$border-radius-base;
+            margin-top:4px;
+            & > .dropdown-toggle {
+                padding:0px !important;
+                color:$dropdown-btn-color;
+                &:hover, &:focus {
```
AlasdairSwan, line 43: What about :active ?

ecommerce/static/sass/_base.scss
downzer0, line 14: Ideally, `hover` and `focus` styles should be the same.

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+                    outline:none;
+                }
+                & > .text {
+                    padding:3px 12px 4px 12px;
```
AlasdairSwan, line 47: nitpick: this is the same as 

```
padding: 3px 12px 4px;
```

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+                    font-weight:bold;
+                    display:inline-block;
+                    vertical-align:middle;
+                    background-color:transparent !important;
```
AlasdairSwan, line 51: Why the !important ?

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+                        float:left;
+                    }
+                    div {
+                        float:right;
+                        margin-top:-2px;
+                    }
+                }
+                & > .arrow {
+                    border-left:1px solid $dropdown-btn-border;
+                    padding:3px 8px 4px 8px;
```
AlasdairSwan, line 65: Nitpick: is the same as

```
padding: 3px 8px 4px;
```

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+                    display:inline-block;
+                    .caret {
+                        margin:0px;
+                        border-width:5px 5px 5px 5px;
```
AlasdairSwan, line 69: Why 4 of the same value?

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+                        float:right;
+                        margin-top:-2px;
+                    }
+                }
+                & > .arrow {
+                    border-left:1px solid $dropdown-btn-border;
+                    padding:3px 8px 4px 8px;
+                    display:inline-block;
+                    .caret {
+                        margin:0px;
```
AlasdairSwan, line 68: don't need the px

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+}
+
+
+// Conditional Styling
+// ----------------------
+
+// Small Screens
+@media only screen and (max-width: $screen-sm-min) {
```
AlasdairSwan, line 85: The preferred approach is mobile first so you are only ever using min-width. Are you forced to use this approach because Bootstrap or was it a conscious decision?

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+
+    .nav.navbar-default {
+        padding:20px 10px 10px 10px;
```
AlasdairSwan, line 88: Please use the appropriate format based on the use case (1, 2, 3, or 4 params)

```
padding: 20px 10px 10px;
```

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    .dropdown {
+        border:none !important;
```
AlasdairSwan, line 92: Why the !important ?

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+        &>.dropdown-toggle>.arrow {
+            border:none !important;
```
AlasdairSwan, line 94: Why the !important ?

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    .navbar-default .navbar-nav > .open > a,
+    .navbar-default .navbar-nav > .open > a:hover,
+    .navbar-default .navbar-nav > .open > a:focus {
+        border-radius:$border-radius-base;
```
AlasdairSwan, line 101: This can be written more cleanly so you're not repeating so many selectors

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+      padding:17px 10px;
+    }
+
+    .navbar-nav > li > .dropdown-menu {
+        border-radius:$border-radius-base;
+    }
+
+    .navbar-default .navbar-nav > .open > a,
+    .navbar-default .navbar-nav > .open > a:hover,
+    .navbar-default .navbar-nav > .open > a:focus {
```
AlasdairSwan, line 124: This can be written more cleanly so you're not repeating so many selectors.
What about :active ?

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+        // Tooltip caret
+        &:before {
+            @include rotate(-45deg);
+            border: {
+                top: 6px solid $dropdown-bg;
+                right: 6px solid $dropdown-bg;
+                bottom: 6px solid rgba(0, 0, 0, 0);
+                left: 6px solid rgba(0, 0, 0, 0);
+            };
+            background: rgba(0, 0, 0, 0);
```
AlasdairSwan, line 142: Let's create a _colors.scss partial and create a variable for $transparent which can be used here (and anywhere else you are using rgba(0, 0, 0, 0)

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+            box-shadow: 1px 0 0 0 $dropdown-shadow, 0 -1px 0 0 $dropdown-shadow;
+            content: "";
+            display: block;
+            width: 0px;
```
AlasdairSwan, line 146: no need for px

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+            height: 0px;
```
AlasdairSwan, line 147: no need for px

ecommerce/static/sass/_navbar.scss
```diff
@@ -0,0 +1,177 @@
...
+        &>li {
+            border-top: 1px solid $dropdown-divider;
+            &:first-child {
+                border-top:none;
+            }
+            a {
+                font-size:16px;
+                padding:3px 5px 4px;
+                border:1px solid transparent;
+                margin: 5px 0px;
```
AlasdairSwan, line 162: no need for px for 0

AlasdairSwan:
@Nickersoft is there a sandbox in which I can see this code live?

ecommerce/templates/edx/base.html
```diff
@@ -19,8 +19,45 @@
...
 <body role="document">
+
+<nav class="nav navbar-default" role="banner" aria-label="Main">
+  <div class="container">
+    <div class="navbar-header">
+      <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#edx-navbar-collapse" aria-expanded="false">
+        <span class="sr-only">Toggle navigation</span>
+        <span class="icon-bar"></span>
+        <span class="icon-bar"></span>
+        <span class="icon-bar"></span>
```
AlasdairSwan, line 33: Is this how the hamburger menu is being created? If so could you not use glyphicon glyphicon-menu-hamburger from Bootstrap?

ecommerce/templates/edx/base.html
```diff
@@ -19,8 +19,45 @@
...
     {% endblock %}
   {% endcompress %}
+
 </head>
 <body role="document">
+
+<nav class="nav navbar-default" role="banner" aria-label="Main">
+  <div class="container">
+    <div class="navbar-header">
+      <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#edx-navbar-collapse" aria-expanded="false">
```
AlasdairSwan, line 29: Does focus move to the navbar dropdown on button submit?

ecommerce/static/sass/_config-variables.scss
Nickersoft, line 53: @AlasdairSwan It's a mix between Bootstrap variables and custom variables. I added a few custom variables upon the recommendation of @clintonb, due to his feeling that the CSS contains too many hard-coded variables.

ecommerce/static/sass/_navbar.scss
Nickersoft, line 43: @AlasdairSwan Most of these selector overrides, such as :active, :focus, and :hover, were taken directly from the CSS rules found in Chrome developer tools when inspecting certain elements of the Bootstrap DOM. In a case like this, such as with the previous case, the rule affected .dropdown-toggle:hover and dropdown-toggle:focus only, so I only included those in the SASS. I can add :active, just figured I'd explain my reasoning first.

ecommerce/static/sass/_navbar.scss
Nickersoft, line 51: @AlasdairSwan In this particular case, I'm not exactly sure. It may have been leftover from a previous change in the DOM hierarchy. I'll investigate, and if the !important, or the line altogether, is not needed, then I'll remove.

ecommerce/static/sass/_navbar.scss
Nickersoft, line 69: @AlasdairSwan I think perhaps previously it had been 4 different values, but when they were all adjusted to be the same, I just forgot to group them back into a single value. I'll fix.

ecommerce/static/sass/_config-variables.scss
AlasdairSwan, line 53: Can you please separate out the bootstrap variables into their own partial. Then we should have a colors partial for custom color variables and other more specific partials if there are more variables.

ecommerce/static/sass/_navbar.scss
AlasdairSwan, line 43: Can you test what the element looks like when :active and decide based on that.

ecommerce/static/sass/_navbar.scss
Nickersoft, line 85: @AlasdairSwan It was a conscious decision. I believe it had to do with the fact that the dropdown/navbar requires much more styling by default than it does when displayed on mobile. Instead of moving all of the default styling into a media query, I figured I'd leave it as a default and then remove the mobile styling after it surpassed the breakpoint. I'll use min instead.

ecommerce/static/sass/_navbar.scss
Nickersoft, line 88: @AlasdairSwan I was not aware of the three parameter option. I'll use it when applicable from now on.

ecommerce/static/sass/_navbar.scss
Nickersoft, line 92: @AlasdairSwan CSS selectors tend to prefer the most specific rule possible. Because .dropdown fell under more specific rules, the !important was needed to ensure that this particular rule be preferred. Although, I will experiment with all of the included !important and remove the nonessentials. On some, !important may have been needed at one time but are no longer needed.

ecommerce/templates/edx/base.html
Nickersoft, line 29: @AlasdairSwan This hasn't been tested. Will be investigated further.

ecommerce/templates/edx/base.html
Nickersoft, line 33: @AlasdairSwan This was directly from the Bootstrap documentation.

See: [http://getbootstrap.com/components/#navbar](http://getbootstrap.com/components/#navbar)

ecommerce/static/sass/_navbar.scss
AlasdairSwan, line 92: It is better practice to make your rule more specific than using !important. !important should be used very rarely (mainly when overriding CMS generated junk)

frrrances, line 92: I second that. !important trumps everything else so you should only use it in very, very scoped cases and very rarely.

clintonb:
Closing in favor of #189.

PR #183, opened 2015-06-29
Audit logging tweaks

rlucioni:
Simplify audit logging utility, log order_placed message prior to order fulfillment, and include course ID and certificate type in messages logged by the EnrollmentFulfillmentModule. XCOM-426.

@jimabramson or @clintonb, I'd appreciate a review when you get a chance.

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -1,4 +1,3 @@
-from collections import OrderedDict
 from functools import wraps
 import logging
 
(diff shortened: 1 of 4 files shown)


jimabramson:
@rlucioni code lgtm; my one question is whether the course key and cert type attribute lookups are incurring database hits that would otherwise not be.  if that's the case i think we should be sure the additional attributes in the logs justify the additional work.

rlucioni:
@jimabramson I've made a slight change so that including these additional attributes in the log messages results in no additional work. The course key and certificate type attributes are retrieved once and cached in process, during both fulfillment and revocation.

jimabramson:
:+1: good stuff

clintonb:
:+1:

PR #186, opened 2015-07-01
ECOM-1718: Added the enrollment_attributes of provider_id in credit c…

aamir-khan:
@awais786 @ahsan-ul-haq @zubair-arbi Please review this PR.

### DIFF
--- a/ecommerce/extensions/catalogue/tests/mixins.py
+++ b/ecommerce/extensions/catalogue/tests/mixins.py
@@ -52,7 +52,7 @@ class CourseCatalogTestMixin(object):
             seat.course = course
             seat.save()
 
-    def create_course_seats(self, course_id, certificate_types):
+    def create_course_seats(self, course_id, certificate_types, provider=None):
         title = 'Seat in {}'.format(course_id)
         parent_product = factories.ProductFactory(structure='parent', title=title,
                                                   product_class=self.seat_product_class)
@@ -65,6 +65,7 @@ class CourseCatalogTestMixin(object):
 
             seat.attr.certificate_type = certificate_type
             seat.attr.course_key = course_id
+            seat.attr.credit_provider = provider
             seat.save()
 
             factories.StockRecordFactory(product=seat)
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -159,6 +159,11 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
                 logger.error("Supported Seat Product does not have required attributes, [certificate_type, course_key]")
                 line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
                 continue
+            try:
+                provider = line.product.attr.credit_provider
+            except AttributeError:
+                logger.info("Supported Seat Product does not have the provider. Setting it to None")
+                provider = None
 
             data = {
                 'user': order.user.username,
@@ -166,9 +171,17 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
                 'mode': certificate_type,
                 'course_details': {
                     'course_id': course_key
-                }
+                },
+                'enrollment_attributes': []
             }
-
+            if provider:
+                data['enrollment_attributes'].append(
+                    {
+                        'namespace': 'credit',
+                        'name': 'provider_id',
+                        'value': provider
+                    }
+                )
             try:
                 response = self._post_to_enrollment_api(data)
 
@@ -182,7 +195,8 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
                         product_class=line.product.get_product_class().name,
                         course_id=course_key,
                         certificate_type=certificate_type,
-                        user_id=order.user.id
+                        user_id=order.user.id,
+                        credit_provider=provider,
                     )
                 else:
                     try:
--- a/ecommerce/extensions/fulfillment/tests/test_modules.py
+++ b/ecommerce/extensions/fulfillment/tests/test_modules.py
@@ -32,10 +32,10 @@ class EnrollmentFulfillmentModuleTests(CourseCatalogTestMixin, FulfillmentTestMi
     """Test course seat fulfillment."""
     course_id = 'edX/DemoX/Demo_Course'
     certificate_type = 'test-certificate-type'
+    # provider = 'MIT'
 
     def setUp(self):
         user = UserFactory()
-
         seats = self.create_course_seats(self.course_id, (self.certificate_type,))
         self.seat = seats[self.certificate_type]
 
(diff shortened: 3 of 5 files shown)


ahsan-ul-haq:
lgtm :+1:

clintonb:
This needs tests before it can be merged.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -159,6 +159,11 @@ def fulfill_product(self, order, lines):
                 logger.error("Supported Seat Product does not have required attributes, [certificate_type, course_key]")
                 line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
                 continue
+            try:
+                provider_id = line.product.attr.provider_id
+            except AttributeError:
+                logger.info("Supported Seat Product does not have the provider_id. Setting it to None")
+                provider_id = None
```
wedaly: I think this will cause issues downstream, since the LMS needs this to know which provider to request credit from.  The best case scenario is that the user will see "an unexpected error occurred, please contact support".  Could we log this with log level "error" so we get notified early if this happens?

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -218,7 +231,22 @@ def revoke_line(self, line):
...
+
+            if provider_id:
+                data['enrollment_attributes'].append(
+                    {
+                        'namespace': 'credit',
+                        'name': 'provider_id',
+                        'value': provider_id
+                    }
+                )
```
wedaly: Do we need this when revoking the order?  Shouldn't unenrolling the user reset their enrollment attributes?

ecommerce/extensions/fulfillment/modules.py
aamir-khan: @wedaly I did this in separate try except block because in the case when the payment is for only `verified` mode then this attribute will not be set. If the every mode has this attribute set then we can log an error for notification.

ecommerce/settings/local.py
```diff
@@ -128,3 +128,10 @@
 
 ENABLE_AUTO_AUTH = True
 LOGGING = get_logger_config(debug=DEBUG, dev_env=True, local_loglevel='DEBUG')
+
+#####################################################################
+# Lastly, see if the developer has any local overrides.
+try:
+    from .private import *      # pylint: disable=import-error
```
clintonb, line 135: Thank you! I've been meaning to do this for a while now.

ecommerce/courses/tests/test_models.py
```diff
@@ -30,7 +33,17 @@ def test_seat_products(self):
         self.assertEqual(len(course.seat_products), 0)
 
         # Create the seat products
-        seats = self.create_course_seats(course.id, ('honor', 'verified'))
+        seats = self.create_course_seats(course.id, ('honor', 'verified'), self.provider)
+
+        # Associate the parent and child products with the course. The method should be able to filter out the parent.
+        parent = seats.values()[0].parent
```
clintonb: I updated `create_course_seats` last week to do the seat-course association. You no longer need these lines.

ecommerce/extensions/catalogue/tests/mixins.py
```diff
@@ -52,7 +52,7 @@ def _link_course_and_seats(self, course_id, seats):
             seat.course = course
             seat.save()
 
-    def create_course_seats(self, course_id, certificate_types):
+    def create_course_seats(self, course_id, certificate_types, provider):
```
clintonb: `provider` should be a kwarg that defaults to `None`.

ecommerce/extensions/catalogue/tests/mixins.py
```diff
@@ -65,6 +65,7 @@ def create_course_seats(self, course_id, certificate_types):
 
             seat.attr.certificate_type = certificate_type
             seat.attr.course_key = course_id
+            seat.attr.provider = provider
```
clintonb: Shouldn't this be for credit seats only? Theoretically, yes, we could support a third-party "provider" for non-credit seats; but, we should be testing the intended usecase: credit seats.

ecommerce/extensions/catalogue/tests/test_utils.py
```diff
@@ -7,11 +7,15 @@
 
 
 class UtilsTests(CourseCatalogTestMixin, TestCase):
+
```
clintonb: Once `provider` is made an optional field for `create_course_seats`, these changes can be reverted.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -166,9 +171,17 @@ def fulfill_product(self, order, lines):
                 'mode': certificate_type,
                 'course_details': {
                     'course_id': course_key
-                }
+                },
+                'enrollment_attributes': []
```
clintonb, line 175: This is not necessary.

clintonb, line 175: Eh. Nevermind. It could be useful in the future if we ever have other attributes.

ecommerce/extensions/payment/tests/test_processors.py
```diff
@@ -51,11 +51,12 @@ class PaymentProcessorTestCaseMixin(RefundTestMixin, CourseCatalogTestMixin, Pay
 
     COURSE_KEY = 'test-course-key'
     CERTIFICATE_TYPE = 'test-certificate-type'
+    provider = 'MIT'
```
clintonb: Once `provider` is made an optional field for `create_course_seats`, these changes can be reverted.

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -88,6 +89,13 @@ def test_enrollment_module_fulfill(self):
             'course_details': {
                 'course_id': self.course_id,
             },
+            'enrollment_attributes': [
```
clintonb: There should be a separate test for credit seats. We want to make sure non-credit seats continue to work as expected (e.g. enrollment_attributes is an empty list).

ecommerce/extensions/fulfillment/modules.py
aamir-khan: Agreed. Removing this block of code.

ecommerce/extensions/fulfillment/modules.py
aamir-khan: @wedaly We are checking on the LMS side about the required params of the given mode (as per your comment https://github.com/edx/edx-platform/pull/8580#discussion_r33729743). Please let me know if we need this validation on the ecommerece side also?

ecommerce/extensions/catalogue/tests/mixins.py
aamir-khan: @clintonb for non-credit seats this attribute value will be set to `None` but the attribute name will exists. Do we need to remove this `provider` attribute for the non-credit seats?

clintonb: No, that is not necessary.

aamir-khan:
@wedaly @clintonb Please review this PR

aamir-khan:
jenkins test this please

aamir-khan:
test this please

wedaly:
:+1: Nice work!

PR #187, opened 2015-07-01
ECOM-1807 new attribute for the credit-provider.

awais786:
Add thumbnail url field for the credit provider.

### DIFF
--- a/ecommerce/credit/tests/test_views.py
+++ b/ecommerce/credit/tests/test_views.py
@@ -42,6 +42,7 @@ class CheckoutPageTest(UserMixin, CourseCatalogTestMixin, TestCase):
 
         seat = self.seats['credit']
         seat.attr.credit_provider = self.provider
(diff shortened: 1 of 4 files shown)


awais786:

clintonb:
@wedaly we discussed making this a provider attribute (instead of product). Is that no longer the case?

wedaly:
@clintonb No, it's still the case -- it's just that we had that discussion yesterday afternoon :)

@awais786 After talking this over with Clinton, we think it makes more sense to store the image URL with the credit provider in the LMS.  This will allow us to re-use the same image for all credit seats from the same provider; we also get the benefit of using an asset pipeline / CDN that's already set up.

Unfortunately, I think this is a scope change.  Could you please move the subtask you created into its own story so we can point it at our next grooming meeting?  My apologies for the miscommunication.

awais786:
@wedaly, @clintonb closing this PR.

PR #188, opened 2015-07-02
Awais786/ecom 1720 checkout processor params

awais786:


awais786:
WIP @clintonb Please review this. Its not completed. paymentData.payment_form_data is empty and when the form submit from handleCreateOrderResponse I am getting csrf issue. working on these issues.

ecommerce/static/js/views/credit_checkout_view.js
```diff
@@ -0,0 +1,78 @@
...
+                dataType: "json",
+                method:'post',
+                data: JSON.stringify(
+                    {
+                        'products': [{'sku': sku}],
+                        'checkout': true,
+                        'payment_processor_name': processor_name
+                    }
+                ),
+                headers: {'X-CSRFToken':  $("[name='csrfmiddlewaretoken']").val()},
```
clintonb: Use jQuery.cookie: `$.cookie('ecommerce_csrftoken')`. You'll need to require `jquery-cookie` in your definitions above.

ecommerce/static/js/views/credit_checkout_view.js
```diff
@@ -0,0 +1,78 @@
...
+//        submitForm: function( form ) {
+//            form.submit();
+//        },
+
+        handleCreateOrderError: function( xhr ) {
+            alert('error');
+        },
+
+        ajaxcall: function(sku, processor_name){
+            $.ajax({
```
clintonb: Use `$.post`.

ecommerce/static/js/views/credit_checkout_view.js
```diff
@@ -0,0 +1,78 @@
...
+                url: '/api/v2/baskets/',
+                contentType: "application/json",
```
clintonb: This isn't necessary.

ecommerce/static/js/views/credit_checkout_view.js
```diff
@@ -0,0 +1,78 @@
...
+                dataType: "json",
+                method:'post',
+                data: JSON.stringify(
```
clintonb: No need to stringify this.

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -1,10 +1,12 @@
 {% extends 'edx/base.html' %}
+{% load staticfiles %}
 {% load i18n %}
 {% block title %}{% trans "Credit Checkout" %}{% endblock title %}
 
 {% block content %}
     Course Name : {{ course.name }}
     <img src="{{ course.thumbnail_url }}" alt="{{ course.name }}">
+    {% csrf_token %}
```
clintonb: This can be removed when you use jquery-cookie.

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -19,7 +21,20 @@
             <td>{{ credit_seat.stockrecords.first.price_excl_tax }}</td>
           </tr>
         </table>
+
     {% endfor %}
 
+    <div id="payment-buttons">
+            <button id="btn-paypal" data-processor-name="paypal" data-sku="{{ credit_seats.0.stockrecords.first.partner_sku }}" class="btn btn-primary payment-button">Checkout with PayPal</button>
```
clintonb: You should iterate over all payment processors instead of declaring these buttons statically. Additionally, the SKU should only be set once per provider.

ecommerce/static/js/views/credit_checkout_view.js
```diff
@@ -0,0 +1,78 @@
...
+
+            // Marketing needs a way to tell the difference between users
+            // leaving for the payment processor and users dropping off on
+            // this page. A virtual pageview can be used to do this.
+//            window.analytics.page( 'payment', 'payment_processor_step' );
+
+            form.submit();
+        },
+
+//        submitForm: function( form ) {
```
clintonb: Delete this if it won't be used.

ecommerce/static/js/views/credit_checkout_view.js
```diff
@@ -0,0 +1,78 @@
...
+
+        return Backbone.View.extend({
+
+        events: {
+            'click .payment-button': 'clicked'
+        },
+        initialize: function() {
+            _.bindAll(this, 'clicked');
+        },
+        clicked: function (event) {
```
clintonb: Use a more descriptive name (e.g. checkout).

ecommerce/static/js/views/credit_checkout.js
```diff
@@ -0,0 +1,12 @@
+require([
+        'jquery',
+        'underscore',
```
clintonb: Neither underscore nor Backbone is needed in this file.

ecommerce/static/js/views/credit_checkout_view.js
```diff
@@ -0,0 +1,78 @@
+define([
```
clintonb: Nothing here is credit-specific. This view can be generalized to "checkout_view.js".

ecommerce/static/js/views/credit_checkout.js
```diff
@@ -0,0 +1,12 @@
...
+        'backbone',
+        'js/views/credit_checkout_view'
+    ],
+    function( $, _, Backbone, CreditView ) {
+        var view1 = new CreditView({
+           el: $( '#payment-buttons')
+        });
+    }
+);
```
clintonb: Add a new line.

ecommerce/static/js/views/credit_checkout_view.js
clintonb: Actually "payment_button_view.js" is more accurate.

ecommerce/static/js/views/credit_checkout_view.js
awais786: Without stringify data is not reaching in proper format.

```
<QueryDict: {u'payment_processor_name': [u'paypal'], u'checkout': [u'true'], u'products[0][sku]': [u'credit-variant']}>
```

This is correct format
{u'checkout': True, u'products': [{u'sku': u'credit-variant'}], u'payment_processor_name': u'paypal'}

ecommerce/static/js/views/payment_button_view.js
```diff
@@ -0,0 +1,96 @@
...
+        'jquery',
+        'underscore',
+        'backbone',
+        'jquery-cookie'
+    ],
+    function( $, _, Backbone, cookie ) {
+
+        return Backbone.View.extend({
+
+        events: {
```
clintonb: Add newlines between each of these declarations.

ecommerce/static/js/views/payment_button_view.js
```diff
@@ -0,0 +1,96 @@
...
+            'click .payment-button': 'checkout'
+        },
+        initialize: function() {
+            _.bindAll(this, 'checkout');
+        },
+        checkout: function (event) {
+            this.ajaxcall($(event.currentTarget).data('sku'),$(event.currentTarget).data('processor-name'));
```
clintonb: Why not combine these two functions? There doesn't seem to be a need to keep them separate.

ecommerce/static/js/views/credit_checkout.js
```diff
@@ -0,0 +1,10 @@
+require([
+        'jquery',
+        'js/views/payment_button_view'
+    ],
+    function( $, CreditView ) {
```
clintonb: Update the variable name.

awais786:
@rlucioni Please check analytics implementation. In segment-io.html I am using code from https://segment.com/docs/libraries/analytics.js/quickstart/ ( step 1). If I try to chane the analytics.js path from bower-component then it is giving me error. Reason is default implementation generates file on cdn including write-key e.g 
http://cdn.segment.com/analytics.js/v1/mykey/analytics.min.js

ecommerce/templates/edx/segment-io.html
```diff
@@ -0,0 +1,16 @@
+<!-- begin Segment.io -->
```
rlucioni: I recognize this template from the LMS. Let's take this opportunity to clean it up. Please rename the file `analytics.html`. The "begin Segment.io" and "end Segment.io" tags aren't necessary; please remove them.

ecommerce/templates/edx/segment-io.html
```diff
@@ -0,0 +1,16 @@
...
+  analytics.load("{{ SEGMENT_KEY }}");
+  analytics.page()
+  }}();
+</script>
+{% if user.is_authenticated %}
+    <script type="text/javascript">
+      analytics.identify("{{user.id}}", {
+        email: "{{user.email}}",
+        username: "{{user.username}}"
+    });
```
rlucioni: Align this with line 10.

ecommerce/static/js/views/payment_button_view.js
```diff
@@ -0,0 +1,102 @@
...
+        },
+
+        checkout: function (event) {
+
+            var sku = $(event.currentTarget).data('sku') ;
+            var processor_name = $(event.currentTarget).data('processor-name');
+
+            analytics.track('ecommerce.credit.course.checkout', {
+              category: $(event.currentTarget).data('course-id'),
+              label:processor_name
```
rlucioni: Nitpick: add a space between the colon and the variable name.

ecommerce/static/js/views/payment_button_view.js
```diff
@@ -0,0 +1,102 @@
...
+        initialize: function() {
+            _.bindAll(this, 'checkout');
+        },
+
+        checkout: function (event) {
+
+            var sku = $(event.currentTarget).data('sku') ;
+            var processor_name = $(event.currentTarget).data('processor-name');
+
+            analytics.track('ecommerce.credit.course.checkout', {
```
rlucioni: Business intelligence event names use a standardized format developed by the analytics team. We also want the event's name to make the action the event corresponds to clear. I suggest changing the name to `edx.bi.ecommerce.credit.payment_selected`.

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -35,13 +36,29 @@ <h3 class="title">{{ credit_seats.0.attr.credit_provider }}</h3>
         <div class="row center-block text-right total-price">
             <div class="col-sm-12">Total: ${{ credit_seats.0.stockrecords.first.price_excl_tax }}</div>
         </div>
-        <div class="row center-block checkout-controls">
+        <div id="payment-buttons" class="row center-block checkout-controls">
           <div class="col-xs-6 ">
-            <button class="btn btn-primary payment-button">Checkout with PayPal</button>
+            <button data-processor-name="paypal" data-sku="{{ credit_seats.0.stockrecords.first.partner_sku }}"
```
rlucioni: Why are these payment buttons being hardcoded? They should be populated dynamically based on the configured payment processors.

awais786:
@rlucioni May i remove the bower analytics stuff from the PR ?

ecommerce/templates/edx/segment-io.html
```diff
@@ -0,0 +1,16 @@
...
+<script type="text/javascript">
+  !function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","group","track","ready","alias","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t){var e=document.createElement("script");e.type="text/javascript";e.async=!0;e.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(e,n)};analytics.SNIPPET_VERSION="3.0.1";
+  analytics.load("{{ SEGMENT_KEY }}");
+  analytics.page()
+  }}();
+</script>
+{% if user.is_authenticated %}
+    <script type="text/javascript">
+      analytics.identify("{{user.id}}", {
```
rlucioni: This user ID needs to be replaced with the LMS ID stored in the user's tracking context. Using the Otto ID will cause problems for those using Google Analytics to track user flows.

rlucioni:
@awais786 I'd like to try harder to use Bower and RequireJS for this. The way the LMS loads Segment is not ideal, and Segment does support use via Bower and RequireJS.

I don't fully understand the difficulties you're experiencing. Here's what I'd do, some of which you already appear to have done successfully:
1. Use Bower to install the latest version of Segment's `analytics` package.
2. In `config.js`, use `require.config()` to configure the path to use for `analytics`. I'd point it to the minified version of the library.
3. For any Backbone views that need to emit events, require `analytics`. In the view's initialization method, initialize `analytics` using `load()` and our API key, then call the `analytics` methods you need to emit events.

Have you tried this? If so, where in the process do things break down?

ecommerce/templates/edx/segment-io.html
rlucioni: Since we're trying to take the Bower/RequireJS route, feel free to ignore this for now.

awais786:
@rlucioni https://github.com/edx/ecommerce/pull/188/files#diff-8efaeacdc41ccb27fc87c5da2eefc30bR4
it generates the file on cdn. http://cdn.segment.com/analytics.js/v1/mykey/analytics.min.js
[redacted] is the write-key.  The current js script has this line 

```
e.src("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";
```

I was trying to replace this line with {% static "bower_components/analytics/analytics.js" %}" .

rlucioni:
@awais786 I've updated your comment to remove the write key you pasted. Please don't paste API keys in public places.

ecommerce/templates/edx/segment-io.html
awais786: @rlucioni You mean this.
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/analytics/utils.py#L36
lms_client_id

rlucioni: No, I'm referring to the `user_tracking_id`.

ecommerce/static/js/views/credit_checkout.js
```diff
@@ -0,0 +1,10 @@
+require([
+        'jquery',
+        'js/views/payment_button_view'
+    ],
+    function( $, PaymentView ) {
```
clintonb: PaymentButtonView

awais786:
@clintonb PR is ready.

ecommerce/credit/tests/test_views.py
```diff
@@ -74,9 +76,15 @@ def test_get_with_disabled_flag(self):
...
         """
         Test checkout page with credit course showing all information.
-        course name , thumbnail url , provider name , price.
+        course name , thumbnail url , provider name , price , payment processors.
         """
         response = self.client.get(reverse('credit:checkout', args=[self.course.id]))
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.context['course'], self.course)
         self.assertEqual(response.context['credit_seats'][0], self.seats['credit'])
+        self.assertEqual(
```
clintonb, line 85: You can use `assertDictEqual` to check all of context in one command.

ecommerce/credit/views.py
```diff
@@ -28,8 +32,14 @@ def get_context_data(self, **kwargs):
             seat for seat in course.seat_products if seat.attr.certificate_type == self.CREDIT_MODE
         ]
 
+        # Fetch the list of payment processor classes based on Django settings
```
clintonb: You can clean these lines up a bit:

``` python
context.update({
  'course': course,
  'credit_seats': ...,
})
```

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -35,13 +36,31 @@ <h3 class="title">{{ credit_seats.0.attr.credit_provider }}</h3>
...
-          <div class="col-xs-6 ">
-            <button class="btn btn-primary payment-button">Checkout</button>
-          </div>
+        <div id="payment-buttons" class="row center-block checkout-controls">
+          {% for processor in payment_processors %}
+            <div class="col-xs-6 ">
+                <button data-processor-name="{{ processor.NAME|lower }}" data-sku="{{ credit_seats.0.stockrecords.first.partner_sku }}"
+                        data-course-id="{{ course.id }}"
+                        class="btn btn-primary payment-button">
+                    {% ifequal processor.NAME|lower 'cybersource'  %}
```
clintonb: Move this logic into the view.

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -35,13 +36,31 @@ <h3 class="title">{{ credit_seats.0.attr.credit_provider }}</h3>
...
+          {% endfor %}
         </div>
     </div>
+
+    <form id="payment-processor-form">
+    </form>
 {% endblock content %}
+
+{% block javascript %}
+    <script src="{% static 'js/views/credit_checkout.js' %}"></script>
```
clintonb: Wrap in a compressor block, and add this script to build.js. See https://github.com/edx/ecommerce#asset-pipeline.

ecommerce/credit/tests/test_views.py
awais786, line 85: Both objects being compare are lists, is it necessary to compare all list objects individually ?
Otherwise can compare with iterating the lists.

awais786:
@clintonb please review this again.

build.js
```diff
@@ -23,6 +23,9 @@
         {
             name: 'js/views/course_list_view',
             exclude: ['js/common']
+        },
+        {
+            name: 'js/views/payment_button_view'
```
clintonb: You also need `exclude: ['js/common']` otherwise, that data will be loaded twice for this page.

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -35,13 +37,25 @@ <h3 class="title">{{ credit_seats.0.attr.credit_provider }}</h3>
...
         </div>
-        <div class="row center-block checkout-controls">
-          <div class="col-xs-6 ">
-            <button class="btn btn-primary payment-button">Checkout with PayPal</button>
-          </div>
-          <div class="col-xs-6 ">
-            <button class="btn btn-primary payment-button">Checkout</button>
-          </div>
+        <div id="payment-buttons" class="row center-block checkout-controls">
+          {% for processorname, buttontext in payment_processors.items %}
```
clintonb: processor_name
button_text

clintonb:
:+1: (once build.js and tempalte updated)

PR #189, opened 2015-07-02
Added theming support

clintonb:
- Styles (esp. logo) can be overridden with a theme-specific CSS file
- Added primary navigation bar

XCOM-367

@andywaldrop @frrrances @AlasdairSwan 
FYI @awais786 @zubair-arbi @wedaly @rlucioni @jimabramson @Nickersoft


clintonb:
**Signed Out**
screen shot 2015-07-02 at 1 58 04 pm

screen shot 2015-07-02 at 1 58 28 pm

**Signed In**
desktop
responsive

ecommerce/static/sass/_buttons.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.btn-default {
+  // Invert the background and foreground colors on active/focus/hover
+  &:hover,
+  &:focus,
+  &.focus,
+  &:active,
+  &.active,
+  .open > &.dropdown-toggle {
+    color: $btn-default-bg;
```
downzer0: NIT: property order (we go from out to in, so border, background, then color)

ecommerce/static/sass/_buttons.scss
```diff
@@ -0,0 +1,48 @@
...
+    background-color: $btn-default-color;
+    border-color: $btn-default-color;
+  }
+
+  // Use edX disabled state
+  &.disabled,
+  &[disabled],
+  fieldset[disabled] & {
+    color: $gray-light;
+    border-color: $gray-light;
```
downzer0, line 22: NIT: property order

ecommerce/static/sass/_buttons.scss
```diff
@@ -0,0 +1,48 @@
...
+.btn-primary {
+  // Highlight active/focus/hover
+  &:hover,
+  &:focus,
+  &.focus,
+  &:active,
+  &.active,
+  .open > &.dropdown-toggle {
+    background-color: $brand-primary-accent;
+    border-color: $brand-primary-accent;
```
downzer0, line 35: NIT: property order

ecommerce/static/sass/_buttons.scss
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  // Use edX disabled state
+  &.disabled,
+  &[disabled],
+  fieldset[disabled] & {
+    color: $gray-darker;
```
downzer0: NIT: property order

ecommerce/static/sass/_mixins.scss
```diff
@@ -0,0 +1,25 @@
...
+  background: url($brand-logo-path) no-repeat;
+  background-size: contain;
+
+  @if $scaled-height != "" {
+    $scale: $scaled-height / $height;
+    $width: $width * $scale;
+    $height: $scaled-height;
+  }
+
+  width: $width;
```
downzer0, line 10: NIT: property order

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
   {% endcompress %}
 </head>
 <body role="document">
+{% block navbar %}
+    <nav class="navbar navbar-default">
+        <div class="container-fluid">
+            <!-- Brand and toggle get grouped for better mobile display -->
+            <div class="navbar-header">
+                <button type="button" class="navbar-toggle collapsed" data-toggle="collapse"
```
downzer0, line 35: Should we add `aria-pressed` here @cptvitamin?

ecommerce/static/sass/_mixins.scss
clintonb, line 10: What's the correct order here?

ecommerce/static/sass/_buttons.scss
clintonb: Updated.

ecommerce/static/sass/_buttons.scss
clintonb, line 22: Updated.

ecommerce/static/sass/_buttons.scss
clintonb, line 35: Updated.

ecommerce/static/sass/_buttons.scss
clintonb: Updated.

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
...
+                        data-target="#main-navbar-collapse" aria-expanded="false">
+                    <span class="sr-only">{% trans "Toggle navigation" %}</span>
+                    <span class="icon-bar"></span>
```
downzer0, line 38: Three of these `icon-bar`'s?

clintonb, line 38: Yes. They create the handle for the responsive menu.

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
...
+                              <i class="icon fa fa-home" aria-hidden="true"></i>
+                              <span class="sr-only">{% trans "Dashboard for:" %}</span>
+                            {{ user.username }}
+                          </button>
+                          <button type="button" class="btn btn-default dropdown-toggle hidden-xs" data-toggle="dropdown"
+                                  aria-haspopup="true" aria-expanded="false">
+                              <span class="caret"></span>
+                              <span class="sr-only">{% trans "Toggle Dropdown" %}</span>
+                          </button>
+                          <a href="#" class="dropdown-toggle visible-xs" data-toggle="dropdown" role="button"
```
downzer0: Is it possible to just use a `button` here, instead of changing the link's role? (cc @cptvitamin)

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
...
+                             aria-haspopup="true" aria-expanded="false"><i
+                                  class="icon fa fa-home"></i> {{ user.username }} <span class="caret"></span></a>
+                          <ul class="dropdown-menu">
+                              <li>
+                                  <a href="{% settings_value 'LMS_DASHBOARD_URL' %}" target="_blank">
```
downzer0: Wondering if we should use an `aria-label` here to let screenreader users know they'll be going to a new browser window. (cc @cptvitamin)

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
...
+                                    {% trans "Student Dashboard" %}
+                                  </a>
+                              </li>
+                              <li><a href="{% url 'courses_list' %}">{% trans "Course Admin Tool" %}</a></li>
+                              <li><a href="{% url 'dashboard:index' %}">{% trans "E-Commerce Dashboard" %}</a></li>
+                              <li role="separator" class="divider"></li>
```
downzer0: Remove the `role` here as it serves no purpose.

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
...
+                              <li><a href="{% url 'logout' %}">{% trans "Sign Out" %}</a></li>
+                          </ul>
+                      </li>
+                  {% else %}
+                      <a class="btn btn-primary navbar-btn hidden-xs" href="{% url 'login' %}" role="button">
```
downzer0: Can we use a `button` instead of an `a` with an altered role?

ecommerce/static/sass/_mixins.scss
downzer0, line 10: https://github.com/edx/ux-pattern-library/wiki/Styleguide:-Sass-&-CSS#property-and-declaration-order

ecommerce/templates/edx/base.html
downzer0, line 38: Do they need to be interacted with, or are they just visual?

ecommerce/templates/edx/base.html
clintonb: Removed.

ecommerce/templates/edx/base.html
clintonb: Role removed.

ecommerce/templates/edx/base.html
clintonb, line 38: Visual only.

ecommerce/static/sass/_mixins.scss
clintonb, line 10: Updated.

frrrances:
@clintonb Can you give some context on what is going on in this PR? The nav items, nav interaction patterns, and header layout don't seem to match anything we have existing. Is this only for the internal ecommerce tool? Will this persist on the Credit Checkout page (and future ecommerce pages)? Once I have that context, we can give some feedback. 

Aside from that, the UI is off from our Pattern Library and/or LMS styles - button styles, font face, typography, etc. @andywaldrop provided some feedback in the credit checkout PR that also applies here - or he can review this independently. (Or, let me know if that is not part of this PR.)

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
   {% endcompress %}
 </head>
 <body role="document">
+{% block navbar %}
+    <nav class="navbar navbar-default">
```
cptvitamin: This `<nav>` (and all `<nav>`'s) needs to have have an `aria-label` on it that describes the type of navigation contained within (do not include the word "navigation" as it is implied/redundant.)

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
   {% endcompress %}
 </head>
 <body role="document">
```
cptvitamin: role declaration here is not necessary.  document is implied.

ecommerce/templates/edx/base.html
cptvitamin, line 35: I don't think its necessary if we are properly managing `aria-expanded`.  I like to reserve `aria-pressed` for when we are using a button like a radio button, to indicate a state.

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
...
+                              <i class="icon fa fa-home" aria-hidden="true"></i>
+                              <span class="sr-only">{% trans "Dashboard for:" %}</span>
+                            {{ user.username }}
+                          </button>
+                          <button type="button" class="btn btn-default dropdown-toggle hidden-xs" data-toggle="dropdown"
+                                  aria-haspopup="true" aria-expanded="false">
+                              <span class="caret"></span>
+                              <span class="sr-only">{% trans "Toggle Dropdown" %}</span>
+                          </button>
+                          <a href="#" class="dropdown-toggle visible-xs" data-toggle="dropdown"
```
cptvitamin: What does this hyperlink do?  isn't it the same as the button that precedes it?  '#' is an invalid value for `href`.  Whenever I see it used, 99% of the time it indicates that a `<button>` should be used.

ecommerce/templates/edx/base.html
cptvitamin: an aria-label would override the default accessible label.  I don't think its necessary to do that here, but it does beg the question: Why are we opening a new browser window?  I'd rather see that behavior removed than anything else.

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
...
+                <ul class="nav navbar-nav navbar-right">
+                  {% if user.is_authenticated %}
+                      <li class="btn-group user-menu">
+                          <button type="button" class="btn btn-default hidden-xs main-btn"
+                                  onclick="window.open('{% settings_value 'LMS_DASHBOARD_URL' %}');">
+                              <i class="icon fa fa-home" aria-hidden="true"></i>
+                              <span class="sr-only">{% trans "Dashboard for:" %}</span>
+                            {{ user.username }}
+                          </button>
+                          <button type="button" class="btn btn-default dropdown-toggle hidden-xs" data-toggle="dropdown"
```
cptvitamin: Just a reminder that clicking on this button should move focus to the first menu item and focus should remain trapped in the menu until the user clicks escape, toggles the button, or clicks outside the menu.

ecommerce/templates/edx/base.html
cptvitamin: Actually, its a link to the login page its OK to leave it a as a link, but get rid of the role="button".  The general rule is Links go somewhere, Buttons do something.

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
...
+                              <li><a href="{% url 'dashboard:index' %}">{% trans "E-Commerce Dashboard" %}</a></li>
+                              <li class="divider"></li>
+                              <li><a href="{% url 'logout' %}">{% trans "Sign Out" %}</a></li>
+                          </ul>
+                      </li>
+                  {% else %}
+                      <a class="btn btn-primary navbar-btn hidden-xs" href="{% url 'login' %}" role="button">
+                        {% trans "Login" %}
+                      </a>
+                      <li class="visible-xs"><a href="{% url 'login' %}">{% trans "Login" %}</a></li>
```
cptvitamin: isn't this the same as the element that precedes it?

ecommerce/core/templatetags/core_extras.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@register.simple_tag
+def settings_value(name):
+    """
+    Retrieve a value from settings.
+
+    If setting is not found, None is returned.
+    """
+    return getattr(settings, name)
```
jimabramson, line 16: If the setting doesn't exist I think `getattr` will raise `AttributeError` unless you explicitly pass `None` as the third param. 

EDIT: based on subsequent tests it looks like AttributeError is expected, and it's the docstring that's incorrect.

ecommerce/core/tests/test_templatetags.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class CoreExtrasTests(TestCase):
+    def test_settings_value(self):
+        template = Template(
+            "{% load core_extras %}"
+            "{% settings_value \"FAKE_SETTING\" %}"
+        )
+
+        # If setting is not found, tag returns None.
```
jimabramson: this inline comment is incorrect

jimabramson:
sanity check: having just finished the work of building a reusable header across marketing & courseware apps, was it impossible to leverage that here as well?

ecommerce/templates/edx/base.html
```diff
@@ -21,6 +27,66 @@
...
+                             aria-haspopup="true" aria-expanded="false"><i
+                                  class="icon fa fa-home"></i> {{ user.username }} <span class="caret"></span></a>
+                          <ul class="dropdown-menu">
+                              <li>
+                                  <a href="{% settings_value 'LMS_DASHBOARD_URL' %}" target="_blank">
+                                    {% trans "Student Dashboard" %}
+                                  </a>
+                              </li>
+                              <li><a href="{% url 'courses_list' %}">{% trans "Course Admin Tool" %}</a></li>
+                              <li><a href="{% url 'dashboard:index' %}">{% trans "E-Commerce Dashboard" %}</a></li>
```
jimabramson: the above two links would presumably be hidden for users without the necessary authorization - is that something we don't need to worry about for now?

ecommerce/urls.py
```diff
@@ -57,6 +58,7 @@ def handler403(_):
         ),
         name='login'
     ),
+    url(r'^accounts/logout/$', logout, name='logout'),
```
jimabramson, line 61: does this actually work?  (i assume i'll still be logged into the platform beyond this point; if i want to log back in to otto will it work normally?)

ecommerce/core/tests/test_templatetags.py
clintonb: Updated.

ecommerce/core/templatetags/core_extras.py
clintonb, line 16: Updated.

ecommerce/templates/edx/base.html
clintonb: Yes, this will be updated when we have proper permissions.

ecommerce/urls.py
clintonb, line 61: You will be logged out of Otto. We don't yet have single logout.

ecommerce/templates/edx/base.html
clintonb: Updated with `aria-label="main"`.

ecommerce/templates/edx/base.html
clintonb: Removed.

ecommerce/templates/edx/base.html
clintonb: This element serves the same purpose as the button above—toggle the dropdown menu—however, it is only visible for mobile clients. I have removed the `href` attribute. Changing the anchor to a button will require some other styling changes, which are feasible but I'd rather not do unless absolutely necessary.

ecommerce/templates/edx/base.html
clintonb: This is handled by [bootstrap-accessibility-plugin](https://github.com/paypal/bootstrap-accessibility-plugin).

ecommerce/templates/edx/base.html
clintonb: The functionality is the same, yes. This particular `<li>` element is only displayed for mobile clients, while the `<a>` above is displayed for non-mobile clients.

ecommerce/templates/edx/base.html
clintonb: Removed.

ecommerce/templates/edx/base.html
clintonb: Now opening in same tab/window.

ecommerce/templates/edx/base.html
clintonb, line 35: Bootstrap updates `aria-expanded` for collapse elements by default.

PR #191, opened 2015-07-08
XCOM-281: JWT issuer and expiration date are now validated

Nickersoft:
In order to enable this feature in production, the following settings must be set in base.py under JWT_AUTH:

'JWT_ISSUER': <issuer passed in LMS>
'JWT_VERIFY_EXPIRATION': True

@clintonb @rlucioni @wedaly @jimabramson

### DIFF
--- a/ecommerce/extensions/api/tests/test_authentication.py
+++ b/ecommerce/extensions/api/tests/test_authentication.py
@@ -1,16 +1,35 @@
 # -*- coding: UTF-8 -*-
+from datetime import datetime
 import json
-from django.conf import settings
 
 import httpretty
+from django.conf import settings
+from django.conf.urls import patterns
+from django.http import HttpResponse
 from django.test import TestCase, override_settings, RequestFactory
 from oscar.test import factories
+from rest_framework import permissions
 from rest_framework.exceptions import AuthenticationFailed
+from rest_framework.views import APIView
+from rest_framework_jwt import utils
+from rest_framework_jwt.settings import api_settings, DEFAULTS
 
 from ecommerce.extensions.api.authentication import BearerAuthentication, JwtAuthentication
-
+from ecommerce.tests.mixins import JwtMixin, UserMixin
 
 OAUTH2_PROVIDER_URL = 'https://example.com/oauth2'
+
+
+class MockView(APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def get(self, request):  # pylint: disable=unused-variable, unused-argument
+        return HttpResponse()
+
+urlpatterns = patterns(
+    '',
+    (r'^jwt/$', MockView.as_view(authentication_classes=[JwtAuthentication]))
+)
 
 
 class AccessTokenMixin(object):
@@ -104,12 +123,64 @@ class BearerAuthenticationTests(AccessTokenMixin, TestCase):
         self.assertEqual(self.auth.authenticate(request), (user, self.DEFAULT_TOKEN))
 
 
-class JwtAuthenticationTests(TestCase):
+class JwtAuthenticationTests(JwtMixin, UserMixin, TestCase):
+    urls = 'ecommerce.extensions.api.tests.test_authentication'
+
+    def setUp(self):
+        api_settings.JWT_ISSUER = 'http://example.com/oauth'
+        api_settings.JWT_VERIFY_EXPIRATION = True
+
+    def tearDown(self):
+        api_settings.JWT_ISSUER = DEFAULTS['JWT_ISSUER']
+        api_settings.JWT_VERIFY_EXPIRATION = DEFAULTS['JWT_VERIFY_EXPIRATION']
+
+    def assert_jwt_status(self, issuer=None, expires=None, status_code=200):
+        """ Assert that the payload has a valid issuer and has not expired. """
+
+        staff_user = self.create_user(is_staff=True)
+        issuer = issuer or api_settings.JWT_ISSUER
+
+        payload = {
+            'iss': issuer,
+            'username': staff_user.username,
+            'email': staff_user.email,
+            'full_name': staff_user.full_name
+        }
+
+        if expires:
+            payload['exp'] = expires
+
+        token = utils.jwt_encode_handler(payload)
+
+        auth = 'JWT {0}'.format(token)
+        response = self.client.get('/jwt/', HTTP_AUTHORIZATION=auth)
+
+        self.assertEqual(response.status_code, status_code)
+
+    def test_valid_issuer(self):
+        """ Verify payloads with valid issuers are validated. """
+        self.assert_jwt_status()
+
+    def test_invalid_issuer(self):
+        """ Verify payloads with invalid issuers are NOT validated. """
+        self.assert_jwt_status(issuer='some-invalid-issuer', status_code=401)
+
+    def test_valid_expiration(self):
+        """ Verify payloads with valid expiration dates are validated. """
+        valid_timestamp = int(datetime.max.strftime('%s'))
+        self.assert_jwt_status(expires=valid_timestamp)
+
+    def test_invalid_expiration(self):
+        """ Verify payloads with invalid expiration dates are NOT validated. """
+        self.assert_jwt_status(expires=1, status_code=401)
+
     def test_authenticate_credentials_user_creation(self):
         """ Test whether the user model is being assigned fields from the payload. """
+
         full_name = 'Ｇｅｏｒｇｅ Ｃｏｓｔａｎｚａ'
         email = 'gcostanza@gmail.com'
         username = 'gcostanza'
+
         payload = {'username': username, 'email': email, 'full_name': full_name}
         user = JwtAuthentication().authenticate_credentials(payload)
         self.assertEquals(user.username, username)
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -15,4 +15,4 @@ pylint==1.4.1
 selenium==2.45.0
 testfixtures==4.1.2
 
-git+https://github.com/edx/ecommerce-api-client.git@1.0.0#egg=ecommerce-api-client==1.0.0
+git+https://github.com/edx/ecommerce-api-client.git@1.1.0#egg=ecommerce-api-client==1.1.0


ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -1,17 +1,27 @@
...
+from datetime import date
+from time import mktime
 
 import httpretty
+from django.core.urlresolvers import reverse
 from django.test import TestCase, override_settings, RequestFactory
 from oscar.test import factories
 from rest_framework.exceptions import AuthenticationFailed
+from rest_framework_jwt.settings import api_settings
+from rest_framework_jwt.settings import DEFAULTS
```
clintonb: Combine into a single line.

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -1,17 +1,27 @@
...
 
-from ecommerce.extensions.api.authentication import BearerAuthentication, JwtAuthentication
-
+from ecommerce.extensions.api.authentication import BearerAuthentication
+from ecommerce.extensions.api.authentication import JwtAuthentication
```
clintonb: Combine into a single line.

ecommerce/extensions/api/tests/test_authentication.py
Nickersoft: @clintonb But wouldn't that go against http://google.github.io/styleguide/pyguide.html#Imports_formatting?

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -1,17 +1,27 @@
...
+from ecommerce.tests.mixins import JwtMixin, UserMixin
 
 OAUTH2_PROVIDER_URL = 'https://example.com/oauth2'
 
+FULL_NAME = 'Ｇｅｏｒｇｅ Ｃｏｓｔａｎｚａ'
```
clintonb: These variables don't seem to be used anywhere else in the file except in the method from which you moved them. Please move them back.

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +114,86 @@ def test_authenticate(self):
...
+    def setUp(self):
+        api_settings.JWT_ISSUER = 'http://example.com/oauth'
+        api_settings.JWT_VERIFY_EXPIRATION = True
+
+    def _mock_payload_response(self, payload):
+        auth_header = 'JWT ' + self.generate_token(payload)
+        response = self.client.get(reverse('api:v2:orders:list'), HTTP_AUTHORIZATION=auth_header)
+        return response
+
+    def test_validate_issuer(self):
```
clintonb: Split the valid/invalid tests into two separate tests.

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +114,86 @@ def test_authenticate(self):
...
+        """ Test whether authentication fails if the JWT has an invalid issuer. """
+
+        staff_user = self.create_user(is_staff=True)
+        self.client.login(username=staff_user.username, password=self.password)
```
clintonb: There should be no reason to login if you are testing JWT auth.

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +114,86 @@ def test_authenticate(self):
...
+            'email': staff_user.email,
+            'full_name': staff_user.full_name
+        }
+
+        response = self._mock_payload_response(payload)
+        self.assertEqual(response.status_code, 401)
+
+        # Non-expired token
+        # -----------------
+        valid_date = date(2084, 1, 24)
```
clintonb: Use `datetime`.

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +114,86 @@ def test_authenticate(self):
...
+            'username': staff_user.username,
+            'email': staff_user.email,
+            'full_name': staff_user.full_name
+        }
+
+        response = self._mock_payload_response(payload)
+        self.assertEqual(response.status_code, 401)
+
+        # Valid issuer
+        # -------------
```
clintonb: Remove the bar.

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +114,86 @@ def test_authenticate(self):
...
+        self.assertEqual(response.status_code, 200)
+
+    def test_validate_expiration(self):
+        """ Test whether authentication fails if the JWT token has expired. """
+
+        staff_user = self.create_user(is_staff=True)
+        self.client.login(username=staff_user.username, password=self.password)
+
+        # Expired token
+        # -------------
```
clintonb: Remove the bar.

requirements/test.txt
```diff
@@ -15,4 +15,4 @@ pylint==1.4.1
 selenium==2.45.0
 testfixtures==4.1.2
 
-git+https://github.com/edx/ecommerce-api-client.git@1.0.0#egg=ecommerce-api-client==1.0.0
+git+https://github.com/edx/ecommerce-api-client.git@7cd154828ff1fbcd12e72ea96b051c5c2f179a15#egg=ecommerce-api-client==7cd154828ff1fbcd12e72ea96b051c5c2f179a15
```
clintonb: This should be the release version, once your changes are merged.

ecommerce/extensions/api/tests/test_authentication.py
clintonb: You can also use `datetime.max`.

requirements/test.txt
Nickersoft: @clintonb Will do. Still waiting on the client PR

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +114,86 @@ def test_authenticate(self):
...
+        self.assertEqual(response.status_code, 401)
+
+        # Non-expired token
+        # -----------------
+        valid_date = date(2084, 1, 24)
+        valid_timestamp = mktime(valid_date.timetuple())
+
+        payload['exp'] = valid_timestamp
+
+        response = self._mock_payload_response(payload)
```
clintonb: You don't need to load a view. You should instead copy the behavior of `test_authenticate_credentials_user_creation` and simply verify the authorized user.

ecommerce/extensions/api/tests/test_authentication.py
clintonb: Nope.

Nickersoft: Fixed

ecommerce/extensions/api/tests/test_authentication.py
Nickersoft: Fixed

ecommerce/extensions/api/tests/test_authentication.py
Nickersoft: Fixed

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +123,67 @@ def test_authenticate(self):
...
+        token = utils.jwt_encode_handler(payload)
+
+        auth = 'JWT {0}'.format(token)
+        response = self.client.get('/jwt/', HTTP_AUTHORIZATION=auth)
+
+        self.assertEqual(response.status_code, status_code)
+
+    def test_valid_issuer(self):
+        """ Verify payloads with valid issuers are validated. """
+        self.assert_jwt_status(issuer=api_settings.JWT_ISSUER, status_code=200)
```
jimabramson: arguments to the assertion function equal its defaults, so unnecessary.

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +123,67 @@ def test_authenticate(self):
...
+
+    def test_invalid_issuer(self):
+        """ Verify payloads with invalid issuers are NOT validated. """
+        self.assert_jwt_status(issuer='some-invalid-issuer', status_code=401)
+
+    def test_valid_expiration(self):
+        """ Verify payloads with valid expiration dates are validated. """
+        valid_timestamp = calendar.timegm(datetime.max.timetuple())
```
jimabramson: FYI: `datetime.datetime.max.strftime('%s')` will give the same value (as a string)

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +123,67 @@ def test_authenticate(self):
         self.assertEqual(self.auth.authenticate(request), (user, self.DEFAULT_TOKEN))
 
 
-class JwtAuthenticationTests(TestCase):
+class JwtAuthenticationTests(JwtMixin, UserMixin, TestCase):
+    urls = 'ecommerce.extensions.api.tests.test_authentication'
+
+    def setUp(self):
+        api_settings.JWT_ISSUER = 'http://example.com/oauth'
+        api_settings.JWT_VERIFY_EXPIRATION = True
```
jimabramson, line 131: would [`override_settings`](https://docs.djangoproject.com/en/1.7/topics/testing/tools/#django.test.override_settings) work for this?

clintonb, line 131: No. DRF does some manipulation of settings when the app loads/tests start: https://github.com/tomchristie/django-rest-framework/blob/fb4233736f3eb780b48f124a42df571902d16c56/rest_framework/settings.py#L215. Modifying `api_settings` is what is done in the DRF tests.

requirements/test.txt
jimabramson: sidetrack: given the findings in https://github.com/edx/edx-platform/pull/8836#discussion_r34200702 do we even need to fuss with the `@` part?

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -1,17 +1,36 @@
...
-
+from ecommerce.tests.mixins import JwtMixin, UserMixin
 
 OAUTH2_PROVIDER_URL = 'https://example.com/oauth2'
 
+class MockView(APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def get(self, request):
+        return HttpResponse({'a': 1, 'b': 2, 'c': 3})
```
jimabramson: considering the response content is never checked, could this be removed or simplified?

ecommerce/extensions/api/tests/test_authentication.py
jimabramson, line 131: understood, thanks for the detail.

ecommerce/extensions/api/tests/test_authentication.py
clintonb: Yep. The arguments can be removed entirely.

requirements/test.txt
clintonb: Yes. We need to tell pip exactly which version of the code to download. The bit after the `==` should generally conform to the version of the package being installed, not the commit hash.

Nickersoft:
@rlucioni @clintonb @jimabramson Updated

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -1,18 +1,38 @@
...
-
+from ecommerce.tests.mixins import JwtMixin, UserMixin
 
 OAUTH2_PROVIDER_URL = 'https://example.com/oauth2'
 
 
+class MockView(APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    # pylint: disable=W0612,W0613
```
clintonb: Use the names, not the codes. Also, this should be on the same line as the method definition.

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +124,66 @@ def test_authenticate(self):
...
         email = 'gcostanza@gmail.com'
         username = 'gcostanza'
+
         payload = {'username': username, 'email': email, 'full_name': full_name}
         user = JwtAuthentication().authenticate_credentials(payload)
         self.assertEquals(user.username, username)
         self.assertEquals(user.email, email)
         self.assertEquals(user.full_name, full_name)
+
+    def tearDown(self):
```
clintonb: Move this below `setUp`.

clintonb:
:+1: (once feedback addressed)

ecommerce/extensions/api/tests/test_authentication.py
```diff
@@ -104,14 +124,66 @@ def test_authenticate(self):
...
+        token = utils.jwt_encode_handler(payload)
+
+        auth = 'JWT {0}'.format(token)
+        response = self.client.get('/jwt/', HTTP_AUTHORIZATION=auth)
+
+        self.assertEqual(response.status_code, status_code)
+
+    def test_valid_issuer(self):
+        """ Verify payloads with valid issuers are validated. """
+        self.assert_jwt_status(issuer=api_settings.JWT_ISSUER)
```
jimabramson: this is effectively the default issuer in the assertion function - is there a reason not to remove the function argument?

jimabramson:
:+1: with just one nitpick

PR #192, opened 2015-07-09
Name and Description of provider added in checkout page

tasawernawaz:
ECOM-1858

@awais786 @aamir-khan @zubair-arbi @ahsan-ul-haq 

Updated for more than one providers, and also separate method added that will responsible for communicating with LMS and returns JSON response.
Context of checkout page updated for provider info. Display name and description of provider added in checkout page.

edX-platform related ticket: https://github.com/edx/edx-platform/pull/8853

### DIFF
--- a/ecommerce/credit/views.py
+++ b/ecommerce/credit/views.py
@@ -1,5 +1,9 @@
+from urlparse import urljoin
+
 from django.http import Http404
 from django.views.generic import TemplateView
+from django.conf import settings
+import requests
 import waffle
 
 from ecommerce.courses.models import Course
@@ -27,6 +31,14 @@ class Checkout(TemplateView):
         context['credit_seats'] = [
             seat for seat in course.seat_products if seat.attr.certificate_type == self.CREDIT_MODE
         ]
+
+        if context['credit_seats']:
+            provider_id = context['credit_seats'][0].attr.credit_provider
+            url = urljoin(settings.LMS_URL_ROOT, '/api/credit/v1/provider/info/{}'.format(provider_id))
+            response = requests.get(url)
+            if response.status_code != 200:
+                response.raise_for_status()
+            context['provider_info'] = response.json()
 
         return context
 
--- a/ecommerce/credit/views.py
+++ b/ecommerce/credit/views.py
@@ -1,6 +1,6 @@
-from urlparse import urljoin
-import requests
 import logging
+import requests
+from urlparse import urljoin
 
 from django.http import Http404
 from django.views.generic import TemplateView
@@ -31,13 +31,14 @@ class Checkout(TemplateView):
 
         context['course'] = course
 
-        context['credit_seats'] = [
+        credit_seats = [
             seat for seat in course.seat_products if seat.attr.certificate_type == self.CREDIT_MODE
         ]
 
-        if context['credit_seats']:
-            provider_id = [seat.attr.credit_provider for seat in context['credit_seats']]
-            context['providers'] = self.get_providers_from_lms(provider_id)
+        if credit_seats:
+            provider_id = [seat.attr.credit_provider for seat in credit_seats]
+            context['providers'] = self.get_providers_detail(provider_id)
+            context['credit_seats'] = credit_seats
 
         return context
 
@@ -47,17 +48,14 @@ class Checkout(TemplateView):
             raise Http404
         return super(Checkout, self).get(request, args, **kwargs)
 
-    def get_providers_from_lms(self, provider_id):
-        """
+    def get_providers_detail(self, provider_id):
+        """Helper method for getting provider info from LMS.
 
-        helper method for getting provider info from LMS with graceful
-        error handling
+        Arguments:
+            provider_id: provider_id will be list containing provider ids
 
-        **Parameter**
-            *provider_id: provider_id will be list containing provider ids
-
-        **Returns**
-            * get response from LMS as json containing list of providers and
+        Returns:
+            Get response from LMS as json containing list of providers and
             returns
 
         """
@@ -67,7 +65,7 @@ class Checkout(TemplateView):
         try:
             response = requests.get(url)
         except requests.exceptions.RequestException as ex:
-            logging.error(ex)
+            logging.exception(ex)
             return {}
 
         if response.status_code != 200:
--- a/ecommerce/credit/views.py
+++ b/ecommerce/credit/views.py
@@ -1,12 +1,15 @@
 from urlparse import urljoin
+import requests
+import logging
 
 from django.http import Http404
 from django.views.generic import TemplateView
 from django.conf import settings
-import requests
 import waffle
 
 from ecommerce.courses.models import Course
+
+logger = logging.getLogger(__name__)
 
 
 class Checkout(TemplateView):
@@ -33,12 +36,8 @@ class Checkout(TemplateView):
         ]
 
         if context['credit_seats']:
-            provider_id = context['credit_seats'][0].attr.credit_provider
-            url = urljoin(settings.LMS_URL_ROOT, '/api/credit/v1/provider/info/{}'.format(provider_id))
-            response = requests.get(url)
-            if response.status_code != 200:
-                response.raise_for_status()
-            context['provider_info'] = response.json()
+            provider_id = [seat.attr.credit_provider for seat in context['credit_seats']]
+            context['providers'] = self.get_providers_from_lms(provider_id)
 
         return context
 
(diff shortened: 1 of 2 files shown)


ecommerce/credit/views.py
```diff
@@ -1,5 +1,9 @@
+from urlparse import urljoin
+
 from django.http import Http404
 from django.views.generic import TemplateView
+from django.conf import settings
+import requests
```
zubair-arbi: Please move this import in separate block as third party python import: https://github.com/edx/edx-platform/wiki/Python-Guidelines#imports-order

ecommerce/credit/views.py
```diff
@@ -28,6 +32,14 @@ def get_context_data(self, **kwargs):
             seat for seat in course.seat_products if seat.attr.certificate_type == self.CREDIT_MODE
         ]
 
+        if context['credit_seats']:
+            provider_id = context['credit_seats'][0].attr.credit_provider
```
zubair-arbi: Please add comment the why you are just getting first credit provider?
Also add a `TODO` to to handle this for all credit providers.

ecommerce/credit/views.py
```diff
@@ -28,6 +32,14 @@ def get_context_data(self, **kwargs):
...
+            url = urljoin(settings.LMS_URL_ROOT, '/api/credit/v1/provider/info/{}'.format(provider_id))
```
zubair-arbi: I would be great if you create a separate method e.g., `get_provider_info_from_lms` which will return a json response and move this cross site call there.

ecommerce/credit/views.py
```diff
@@ -28,6 +32,14 @@ def get_context_data(self, **kwargs):
...
+            response = requests.get(url)
```
clintonb: Wrap this in an try-catch block. If there is a timeout, or some other exception, we need to handle it gracefully.

ecommerce/credit/views.py
```diff
@@ -28,6 +32,14 @@ def get_context_data(self, **kwargs):
...
+            if response.status_code != 200:
+                response.raise_for_status()
+            context['provider_info'] = response.json()
```
clintonb: Why not simply "provider"?

ecommerce/credit/views.py
clintonb: Has this API already been approved? The URL doesn't exactly conform to our normal patterns. I would expect the detail URL for a provider to be "/api/credit/v1/providers/:id/".

ecommerce/credit/views.py
zubair-arbi: @wedaly If we handle all credit providers separately then we will have to call LMS api individually for all providers which will add additional calls overhead.
For compatibility with future work, should we create a separate api in LMS `get_credit_providers_info` which will return list of all credit providers info?

ecommerce/credit/views.py
zubair-arbi: @clintonb that api is being built as part of this PR: https://github.com/edx/edx-platform/pull/8853/files

ecommerce/credit/views.py
wedaly: Yes, I think that makes sense.  From a RESTful design standpoint, I'd imagine we'd want something like:

```
GET /api/credit/v1/providers?provider_id=asu,hogwarts
```

which would respond with a 200 OK and a JSON-encoded list of provider dictionaries.

wedaly: https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions

tasawernawaz:
@wedaly Please review it.

ecommerce/credit/views.py
```diff
@@ -1,9 +1,16 @@
+from urlparse import urljoin
```
clintonb: Alphabetize each group of imports.

ecommerce/credit/views.py
```diff
@@ -28,6 +35,10 @@ def get_context_data(self, **kwargs):
             seat for seat in course.seat_products if seat.attr.certificate_type == self.CREDIT_MODE
         ]
 
+        if context['credit_seats']:
```
clintonb: Create a `credit_seats` variable instead of retrieving it from context multiple times.

ecommerce/credit/views.py
```diff
@@ -35,3 +46,31 @@ def get(self, request, *args, **kwargs):
...
+        try:
+            response = requests.get(url)
+        except requests.exceptions.RequestException as ex:
+            logging.error(ex)
+            return {}
+
+        if response.status_code != 200:
+            response.raise_for_status()
+
+        return response.json()
```
clintonb: Add a newline.

ecommerce/credit/views.py
```diff
@@ -35,3 +46,31 @@ def get(self, request, *args, **kwargs):
...
+            returns
+
+        """
+
+        provider_id = ",".join(provider_id)
+        url = urljoin(settings.LMS_URL_ROOT, '/api/credit/v1/providers/?provider_id={}'.format(provider_id))
+        try:
+            response = requests.get(url)
+        except requests.exceptions.RequestException as ex:
+            logging.error(ex)
```
clintonb: Use `logging.exception`

ecommerce/credit/views.py
```diff
@@ -35,3 +46,31 @@ def get(self, request, *args, **kwargs):
         if not waffle.switch_is_active('ENABLE_CREDIT_APP'):
             raise Http404
         return super(Checkout, self).get(request, args, **kwargs)
+
+    def get_providers_from_lms(self, provider_id):
+        """
+
```
clintonb: Remove the unnecessary newlines.

ecommerce/credit/views.py
```diff
@@ -35,3 +46,31 @@ def get(self, request, *args, **kwargs):
         if not waffle.switch_is_active('ENABLE_CREDIT_APP'):
             raise Http404
         return super(Checkout, self).get(request, args, **kwargs)
+
+    def get_providers_from_lms(self, provider_id):
```
clintonb: get_provider_detail

ecommerce/credit/views.py
```diff
@@ -35,3 +46,31 @@ def get(self, request, *args, **kwargs):
...
+        """
+
+        helper method for getting provider info from LMS with graceful
+        error handling
+
+        **Parameter**
```
clintonb: This should be `Arguments:`. Here is an example of our documentation standard: https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/data.py#L47.

ecommerce/credit/views.py
```diff
@@ -24,10 +31,15 @@ def get_context_data(self, **kwargs):
 
         context['course'] = course
 
-        context['credit_seats'] = [
+        credit_seats = [
             seat for seat in course.seat_products if seat.attr.certificate_type == self.CREDIT_MODE
         ]
 
+        if credit_seats:
+            provider_id = [seat.attr.credit_provider for seat in credit_seats]
```
zubair-arbi: This will be a list of provider id's so change variable name to `provider_ids`

ecommerce/credit/views.py
```diff
@@ -24,10 +31,15 @@ def get_context_data(self, **kwargs):
...
+            context['providers'] = self.get_providers_detail(provider_id)
```
aamir-khan: This will raise exception if credit_seats list is empty and credit_provider will not bi initialized

ecommerce/credit/views.py
```diff
@@ -35,3 +47,28 @@ def get(self, request, *args, **kwargs):
         if not waffle.switch_is_active('ENABLE_CREDIT_APP'):
             raise Http404
         return super(Checkout, self).get(request, args, **kwargs)
+
+    def get_providers_detail(self, provider_id):
```
aamir-khan: provider_id should be provider_ids as it is a list

ecommerce/credit/views.py
```diff
@@ -35,3 +47,28 @@ def get(self, request, *args, **kwargs):
...
+        provider_id = ",".join(provider_id)
+        url = urljoin(settings.LMS_URL_ROOT, '/api/credit/v1/providers/?provider_id={}'.format(provider_id))
+        try:
+            response = requests.get(url)
+        except requests.exceptions.RequestException as ex:
+            logging.exception(ex)
+            return {}
+
+        if response.status_code != 200:
+            response.raise_for_status()
```
aamir-khan: this should also be documented what exception will be raised if status_code is not 200

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -13,18 +13,19 @@ <h3 class="title">
                     Purchasing credit for:
                     <span class="course-title text-nowrap">{{ course.name }}</span>
                 </h3>
-                Congratulations, you are eligible to purchase credit from {{ credit_seats.0.attr.credit_provider }} by Feb 05, 2013.
+                Congratulations, you are eligible to purchase credit from {{ providers.0.display_name }} by Feb 05, 2013.
```
aamir-khan, line 16: This date "Feb 05, 2013" should not be hard coded.

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -13,18 +13,19 @@ <h3 class="title">
...
             </div>
         </div>
 
+        {# TODO in future we have to deal with more than one credit provider on checkout page #}
+
         <div class="row center-block provider-details">
             <div class="col-sm-3 hidden-xs">
-                <img alt="{{ credit_seats.0.attr.credit_provider }}" class="provider-image" src="http://img2.findthebest.com/sites/default/files/10/media/images/Arizona_State_University_at_the_West_Campus_226828.jpg">
+                <img alt="{{ providers.0.display_name }}" class="provider-image" src="http://img2.findthebest.com/sites/default/files/10/media/images/Arizona_State_University_at_the_West_Campus_226828.jpg">
```
aamir-khan, line 24: the src attribute value will be in providers attribute not the hard coded value

zubair-arbi:
Closing this PR in favor of: https://github.com/edx/ecommerce/pull/202

PR #195, opened 2015-07-09
Added expiration datetime to Product model

clintonb:
XCOM-476

### DIFF
--- a/ecommerce/extensions/partner/strategy.py
+++ b/ecommerce/extensions/partner/strategy.py
@@ -1,0 +1,35 @@
+from django.utils import timezone
+
+from oscar.apps.partner import availability, strategy
+from oscar.core.loading import get_model
+
+
+class CourseSeatAvailabilityPolicyMixin(strategy.StockRequired):
+    """
+    Availability policy for Course seats.
+
+    Child seats are only available if the current date is not beyond the seat's enrollment close date.
+    Parent seats are never available.
+    """
+
+    @property
+    def seat_class(self):
+        ProductClass = get_model('catalogue', 'ProductClass')
+        return ProductClass.objects.get(slug='seat')
+
+    def availability_policy(self, product, stockrecord):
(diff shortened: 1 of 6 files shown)


ecommerce/extensions/partner/strategy.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+
+    @property
+    def seat_class(self):
+        ProductClass = get_model('catalogue', 'ProductClass')
+        return ProductClass.objects.get(slug='seat')
+
+    def availability_policy(self, product, stockrecord):
+        """ A seat is unavailable if the current date is beyond the seat's expiration date. """
+        if product.expires and timezone.now() > product.expires:
```
rlucioni, line 22: Is our implementation of Oscar timezone-aware? Is the expiration datetime specified in UTC? If the latter is true, why not use `datetime`'s `utcnow()`?

clintonb, line 22: Yes, Django is setup to be timezone-aware. I prefer to use `timezone.now()` because it is easier to mock. However, in this case I am not doing any mocking, so I am changing this to use datetime.

clintonb, line 22: Actually, because our implementation is timezone-aware, `timezone.now()` is the correct call. `datetime.utcnow()` does not have `tzinfo` attached, so comparing it to the timezone-aware `product.expires` will raise an error.

rlucioni, line 22: Cool, makes sense. Do you know if `product.expires` is specified in EDT? I'm concerned about timezone confusion when specifying expiration times.

rlucioni:
:+1:

ecommerce/extensions/partner/strategy.py
clintonb, line 22: Data is stored in UTC and presented in the timezone set in settings.

## 

Clinton Blackburn

jimabramson:
:+1:

PR #196, opened 2015-07-09
Updated Course Migrator and Publisher

clintonb:
- Migrator now reads and stores the course mode expiration datetime
- Publisher now publishes the course mode expiration datetime

XCOM-478

Review #195 first

### DIFF
--- a/ecommerce/courses/publishers.py
+++ b/ecommerce/courses/publishers.py
@@ -15,7 +15,8 @@ class LMSPublisher(object):
             'name': seat.attr.certificate_type,
             'currency': stock_record.price_currency,
             'price': int(stock_record.price_excl_tax),
-            'sku': stock_record.partner_sku
+            'sku': stock_record.partner_sku,
+            'expires': seat.expires.isoformat() if seat.expires else None,
         }
 
     def publish(self, course):
--- a/ecommerce/courses/tests/test_models.py
+++ b/ecommerce/courses/tests/test_models.py
@@ -1,6 +1,6 @@
 import ddt
 from django.test import TestCase
-from django_dynamic_fixture import G
+from django_dynamic_fixture import G, N
 import mock
 from testfixtures import LogCapture
 from waffle import Switch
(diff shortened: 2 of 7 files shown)


ecommerce/courses/publishers.py
```diff
@@ -15,7 +15,8 @@ def serialize_seat_for_commerce_api(self, seat):
             'name': seat.attr.certificate_type,
             'currency': stock_record.price_currency,
             'price': int(stock_record.price_excl_tax),
-            'sku': stock_record.partner_sku
+            'sku': stock_record.partner_sku,
+            'expires': seat.expires.isoformat() if seat.expires else None,
```
jimabramson, line 19: is that going to come out as UTC, or whatever django decides the requestor's timezone is ?

clintonb, line 19: It should be UTC; however, on the slim chance it is not, `isoformat()` includes the timezone offset (e.g. +00:00).

rlucioni:
:+1:

ecommerce/courses/publishers.py
jimabramson, line 19: don't mean to be a pain, but isoformat doesn't include the offset unless the datetime is timezone-aware, is that guaranteed?

clintonb, line 19: Pain leads to good code...I hope. Yes, our dates are timezone-aware.

jimabramson:
:+1:

ecommerce/courses/tests/test_models.py
```diff
@@ -104,3 +104,12 @@ def test_save_with_publish_failure(self):
         # Reload the course from the database
         course = Course.objects.get(id=course.id)
         self.assertEqual(course.name, orignal_name)
+
+    def test_save_without_publish(self):
+        """ Verify the Course is not published to LMS is the publish kwarg is set to False. """
```
clintonb: Fixing a typo here.

PR #198, opened 2015-07-13
ECOM-1716 Adding analytics.js into config.

awais786:
Google Analytics events on the checkout page.


awais786:
@rlucioni I am getting error while running the make static file.  Screenshot is attached.

screen shot 2015-07-13 at 4 58 32 pm

awais786:
This file is is causing the issue /static/bower_components/analytics/analytics.js

```
(function umd(require){
  if ('object' == typeof exports) {
    module.exports = require('1');
  } else if ('function' == typeof define && (define.amd || define.cmd)) {
    define(function(){ return require('1'); });
  } else {
    this['analytics'] = require('1');
  }
```

ecommerce/static/js/views/payment_button_view.js
```diff
@@ -2,9 +2,10 @@ define([
         'jquery',
         'underscore',
         'backbone',
-        'jquery-cookie'
+        'jquery-cookie',
+        'analytics'
```
zubair-arbi: Adding `analytics` here cause `make static` command to throw `1.js` error:
screen shot 2015-07-13 at 4 58 32 pm
If we remove `analytics` everything works fine. We are adding `payment_button_view` in build `modules`: https://github.com/edx/ecommerce/blob/master/build.js#L28
@rlucioni @jimabramson @AlasdairSwan Can you please help?

rlucioni:
@awais786 and @zubair-arbi, sorry about the delay in getting to this. I did some digging last night, and I'm now having second thoughts about using Bower and Require for this. It wasn't clear to me until now why Segment's quickstart recommends using their [minified snippet](https://segment.com/docs/libraries/analytics.js/quickstart/#step-1-copy-the-snippet).

I imagine you've seen Segment's "integrations" panel. It looks like this:



Segment uses this UI to allow us to configure which third-party analytics providers we interface with. When the quickstart's snippet is executed, Segment takes these selections into account, dynamically generates a custom `analytics.js` file for us which only includes the code required for the providers we've selected, and places it on their CDN. This brings the size of the minified library from about 250KB:



down to about 10KB:



Trying to prepare this file locally might prevent us from using Segment's UI/app altogether. Doing this kind of customization ourselves also looks like it would require an unreasonable amount of effort; developer-oriented documentation for Segment's library doesn't seem to be available.

This being the case, and now that we understand what Segment's doing on our behalf, I would encourage you to load and initialize Segment using the quickstart's snippet. For reference, the edX analytics team uses a [custom Backbone view](https://github.com/edx/edx-analytics-dashboard/blob/89a83992ee1d4a91ee565a55fc368213bd302144/analytics_dashboard/static/js/views/tracking-view.js) to load and initialize `analytics.js`.

wedaly:
@awais786 @zubair-arbi Based on @rlucioni 's comment above, I believe the right approach is to use the analytics.js snippet directly.  We don't need to use Bower/RequireJS to do this.

You can look at the example from Insights that @rlucioni linked to for an example.

Please let me know if there is anything else we can do to get this PR unblocked.  Thank you!

awais786:
@rlucioni thanks for the detailed reply. I am working on the issue according the above guide lines. I will let you know If i need any help. @wedaly I have moved the ticket into progress.

awais786:
@rlucioni Please check this PR. In analytics_view.js it is pointing towards old version '2.0.9'.

ecommerce/static/js/views/analytics_view.js
```diff
@@ -0,0 +1,151 @@
...
+
+                if (_.isUndefined(self.segment)) {
+                    // This is taken directly from https://segment.io/docs/tutorials/quickstart-analytics.js/.
+
+                    /* jshint ignore:start */
+                    // jscs:disable
+                    window.analytics=window.analytics||[],window.analytics.methods=['identify','group','track','page','pageview','alias','ready','on','once','off','trackLink','trackForm','trackClick','trackSubmit'],window.analytics.factory=function(t){return function(){var a=Array.prototype.slice.call(arguments);return a.unshift(t),window.analytics.push(a),window.analytics}};for(var i=0;i<window.analytics.methods.length;i++){var key=window.analytics.methods[i];window.analytics[key]=window.analytics.factory(key)}window.analytics.load=function(t){if(!document.getElementById('analytics-js')){var a=document.createElement('script');a.type='text/javascript',a.id='analytics-js',a.async=!0,a.src=('https:'===document.location.protocol?'https://':'http://')+'cdn.segment.io/analytics.js/v1/'+t+'/analytics.min.js';var n=document.getElementsByTagName('script')[0];n.parentNode.insertBefore(a,n)}},window.analytics.SNIPPET_VERSION='2.0.9';
```
awais786: @rlucioni this is older version. When I try to replace this with https://segment.com/docs/libraries/analytics.js/quickstart/ ( latest minified ) then it is throwing error while running make static.

awais786:

ecommerce/static/js/views/analytics_view.js
```diff
@@ -0,0 +1,151 @@
...
+
+        /**
+         * This 'view' doesn't display anything, but rather sends tracking
+         * information in response to 'segment:track' events triggered by the
+         * model.
+         *
+         * Actions will only be tracked if segmentApplicationId is set in the
+         * model.
+         */
+        var TrackingView = Backbone.View.extend({
```
clintonb: No need for the variable. Just return the view.

ecommerce/static/js/views/clickable-view.js
```diff
@@ -0,0 +1,36 @@
+define(['underscore', 'backbone'],
+    function (_, Backbone) {
+        'use strict';
+
+        /**
+         * Use this for triggering track events when an element is clicked.
+         * 'segment:track' and an event are fired when the element is clicked.
+         */
+        var ClickableView = Backbone.View.extend({
```
clintonb: No need for variable. Just return.

awais786:
screen shot 2015-07-23 at 12 56 37 am
 Getting this error message in console.

ecommerce/static/js/views/analytics_view.js
clintonb: Instead of doing all of this inline, create a new file `/static/vendor/analytics.js` with the contents of analytics.js (or use the bower version) and load the script in `templates/edx/base.html` before loading require.js.

clintonb: Simpler change: If you only copy the first line of the minified quickstart JS, you will have issues due to a missing `}`. Additionally, removing the `analytics` variable will cause problems because that variable is used throughout the minified script. Here is the corrected block:

``` javascript
if (_.isUndefined(self.segment)) {
    // This is taken directly from https://segment.io/docs/tutorials/quickstart-analytics.js/.

    /* jshint ignore:start */
    // jscs:disable
    var analytics = window.analytics = window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","group","track","ready","alias","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t){var e=document.createElement("script");e.type="text/javascript";e.async=!0;e.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(e,n)};analytics.SNIPPET_VERSION="3.0.1";}
    // jscs:enable
    /* jshint ignore:end */

    // shortcut to segment.io
    self.segment = analytics;
}
```

awais786: @clintonb I have replace the line but new version is not sending the button click tracking events.
It is working fine with older version. Although new version identify the user.

clintonb: Can you follow up with Segment tech. support, forums, etc. to determine
what's going on, please? The API probably changed between version 2 and
version 3, so your code may need to change.

awais786: I will followup also checking the code.

ecommerce/static/js/views/analytics_view.js
```diff
@@ -0,0 +1,133 @@
...
+         * Actions will only be tracked if segmentApplicationId is set in the
+         * model.
+         */
+        return Backbone.View.extend({
+
+            /**
+             * Reference to segment.io analytics library.  This is set after
+             * loading.
+             */
+            segment: undefined,
```
clintonb: Remove.

ecommerce/static/js/views/analytics_view.js
```diff
@@ -0,0 +1,133 @@
...
+
+            events: {
+                'shown.bs.tooltip': 'trackElementEvent'
```
clintonb: Remove.

ecommerce/static/js/views/analytics_view.js
```diff
@@ -0,0 +1,133 @@
...
+                    // now segment has been loaded, we can track events
+                    this.listenTo(this.model, 'segment:track', this.track);
+                }
+            },
+
+            /**
+             * This emits an event to our external tracking systems when an
+             * event bubbles up from a DOM element.
+             */
+            trackElementEvent: function (ev) {
```
clintonb: Remove.

ecommerce/static/js/views/credit_checkout.js
```diff
@@ -1,10 +1,47 @@
...
+        'js/views/analytics_view',
+        'js/views/payment_button_view',
+        'js/utils/utils'
     ],
-    function( $, PaymentButtonView ) {
+    function( $,Backbone, ClickableView, TrackingView, PaymentButtonView, Utils ) {
         var paybtnview = new PaymentButtonView({
            el: $( '#payment-buttons')
         });
+        var TrackingModel = Backbone.Model.extend({
```
clintonb: Move to models directory.

ecommerce/static/js/views/credit_checkout.js
```diff
@@ -1,10 +1,47 @@
...
+                var self = this,
+                    trackId = self.get('segmentApplicationId');
+                return !_(trackId).isUndefined() && !_(trackId).isNull();
+            }
+        });
+
+        var trackingModel = new TrackingModel();
+        trackingModel.set(initModelData.tracking);
+
+        var UserModel = Backbone.Model.extend({});
```
clintonb: Move to models directory.

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -56,6 +61,13 @@ <h3 class="title">{{ credit_seats.0.attr.credit_provider }}</h3>
 
     <form id="payment-processor-form">
     </form>
+
+    {% if analytics_data %}
+        <script type="text/javascript">
+            initModelData = {{ analytics_data|safe }};
```
clintonb: var

build.js
```diff
@@ -27,6 +27,10 @@
         {
             name: 'js/views/payment_button_view',
             exclude: ['js/common']
+        },
+        {
+            name: 'js/views/analytics_view',
```
clintonb: Only your top-level views/pages need to be declared here. In this case, you only need `js/views/credit_checkout`.

ecommerce/static/js/views/credit_checkout.js
```diff
@@ -1,10 +1,47 @@
 require([
```
clintonb: We have moved top-level views to the pages directory. This view qualifies and should be moved there.

ecommerce/static/js/views/credit_checkout.js
```diff
@@ -1,10 +1,47 @@
...
+        });
+
+        var trackingModel = new TrackingModel();
+        trackingModel.set(initModelData.tracking);
+
+        var UserModel = Backbone.Model.extend({});
+        userModel = new UserModel();
+        userModel.set(initModelData.user);
+
+        var analyticsView = new TrackingView({model: trackingModel, userModel: userModel});
```
clintonb: This variable is never used. Also, please add a comment explaining what's going on here. It's still not clear to me how events eventually trigger `TrackingView.track()`.

ecommerce/static/js/pages/credit_checkout.js
```diff
@@ -0,0 +1,53 @@
+require([
+        'jquery',
+        'backbone',
+        'js/models/user-model',
+        'js/models/tracking-model',
+        'js/models/course-model',
+        'js/views/clickable-view',
+        'js/views/analytics_view',
```
clintonb, line 8: Either change the variable `TrackingView` to `AnalyticsView` or change the filename to `tracking_view`. This is why I was confused earlier and didn't understand how tracking was called.

clintonb:
:+1: (once naming issue addressed)

PR #199, opened 2015-07-13
Merged health and users apps into core

clintonb:
This single app houses the code that, as its name implies, is core to this project. This is an organizational change. No functionality has been modified.

XCOM-461


clintonb:
FYI the new migration will automatically be faked by Django.

ecommerce/core/views.py
```diff
@@ -1,19 +1,23 @@
...
 import requests
 from requests.exceptions import RequestException
 from rest_framework import status
-from django.conf import settings
 from django.db import transaction, connection, DatabaseError
 from django.http import JsonResponse
+from django.conf import settings
+from django.contrib.auth import get_user_model, login, authenticate
+from django.http import Http404
+from django.shortcuts import redirect
```
rlucioni: Nitpick: unnecessary newline.

ecommerce/core/admin.py
```diff
@@ -2,7 +2,7 @@
 from django.contrib.auth.admin import UserAdmin
 from django.utils.translation import ugettext_lazy as _
 
-from ecommerce.user.models import User
+from ecommerce.core.models import User
```
rlucioni, line 5: Could we use `get_user_model()` instead of importing the user model directly?

rlucioni:
Looks fine to me. I'd prefer it if the `user` and `health` apps were turned into subapps of the `core` app, as we've done within the `extensions` app. I don't think the heartbeat and auto-auth views really belong together, let alone in the same app as the user model.

Regardless, :+1: .

ecommerce/core/admin.py
clintonb, line 5: We could, but given that this particular admin is specifically for `ecommerce.core.models.User`, it makes sense to use that model explicitly.

ecommerce/core/views.py
clintonb: Removed.

ecommerce/core/admin.py
rlucioni, line 5: Cool.

ecommerce/core/migrations/0001_initial.py
```diff
@@ -27,6 +28,8 @@ class Migration(migrations.Migration):
                 ('is_staff', models.BooleanField(default=False, help_text='Designates whether the user can log into this admin site.', verbose_name='staff status')),
                 ('is_active', models.BooleanField(default=True, help_text='Designates whether this user should be treated as active. Unselect this instead of deleting accounts.', verbose_name='active')),
                 ('date_joined', models.DateTimeField(default=django.utils.timezone.now, verbose_name='date joined')),
+                ('full_name', models.CharField(max_length=255, null=True, verbose_name='Full Name', blank=True)),
+                ('tracking_context', jsonfield.fields.JSONField(null=True, blank=True)),
```
jimabramson, line 32: what's the deployment impact of simultaneously moving and changing this migration?

clintonb, line 32: This migration will be faked. No tables are changing, so there should be no impact on any running code.

jimabramson:
I don't understand how the migration will be automatically faked, but assuming we don't need to worry about that or prepare any special dance moves for ops, :+1:

clintonb:
@jimabramson I ran the migration locally, expecting an error, and Django knew to fake it.
It's :sparkles:magic:sparkles:!

clintonb:
@rlucioni I am opting not to split these into two sub-apps primarily because there isn't really a need to do so. The views are related in the sense that they are necessary to run/test the system. If the the size of core/views.py grows at some point, which I really don't see happening, we should split the file. For now, the views can coexist.

rlucioni:
Thanks for the context @clintonb.

PR #200, opened 2015-07-13
Added theming support to the course admin tool

Nickersoft:
Bootstrap 3 was mixed with the pattern library to add theming support to the course admin tool. Based off  of #189. 

A couple things to note:
- Before importing a theme, base/main.scss must be imported first. 
- The theme can be changed via the THEME_SCSS setting.

@clintonb @frrrances @clrux @andy-armstrong @jimabramson @AlasdairSwan 
### LMS Button Styles

cat-button
### LMS Search Field

cat-text
### Bootstrap Dropdown Menu

**_(will eventually be styled to look like LMS)**_
cat-menu
### Responsive Menu

cat-mobile


ecommerce/static/sass/base/_main-ltr.scss
```diff
@@ -0,0 +1 @@
...media (min-width:1024px){.pldoc-wrapper-content-supplemental .pldoc-nav{display:block}}.pldoc-wrapper-content-supplemental .pldoc-footer-site{display:none}@media (min-width:1024px){.pldoc-wrapper-content-supplemental .pldoc-footer-site{display:block}}.pldoc-element-header{max-width:73.125rem;margin-left:auto;margin-right:auto}.pldoc-element-header::after{clear:both;content:"";display:table}.pldoc-element-title{margin-bottom:.625rem}.pldoc-element-meta .meta{display:block}@media (min-width:768px){.pldoc-element-title{width:48.57143%;float:left;margin-right:2.85714%;margin-bottom:0}.pldoc-element-meta{width:48.57143%;float:right;margin-right:0;text-align:right}.pldoc-element-meta .meta{display:inline-block;vertical-align:middle;margin-right:.625rem}}.pldoc-element-meta .meta-updated{margin-bottom:.625rem}@media (min-width:768px){.pldoc-element-meta .meta-updated{margin-bottom:0}}.pldoc.view-index .pldoc-wrapper-link-top{max-width:73.125rem;margin-left:auto;margin-right:auto;padding:0 2.5rem 1.25rem}.pldoc.view-index .pldoc-wrapper-link-top::after{clear:both;content:"";display:table}.pldoc.view-index .pldoc-link-top{width:100%;float:left;margin-left:0;margin-right:0;text-align:center}
```
clintonb: Use the uncompressed styles from the pattern library. Compressor will compress them.

ecommerce/static/sass/base/main-rtl.scss
```diff
@@ -0,0 +1 @@
...}@media (min-width:1024px){.pldoc-wrapper-content-supplemental .pldoc-nav{display:block}}.pldoc-wrapper-content-supplemental .pldoc-footer-site{display:none}@media (min-width:1024px){.pldoc-wrapper-content-supplemental .pldoc-footer-site{display:block}}.pldoc-element-header{max-width:73.125rem;margin-left:auto;margin-right:auto}.pldoc-element-header::after{clear:both;content:"";display:table}.pldoc-element-title{margin-bottom:.625rem}.pldoc-element-meta .meta{display:block}@media (min-width:768px){.pldoc-element-title{width:48.57143%;float:left;margin-right:2.85714%;margin-bottom:0}.pldoc-element-meta{width:48.57143%;float:right;margin-right:0;text-align:left}.pldoc-element-meta .meta{display:inline-block;vertical-align:middle;margin-left:.625rem}}.pldoc-element-meta .meta-updated{margin-bottom:.625rem}@media (min-width:768px){.pldoc-element-meta .meta-updated{margin-bottom:0}}.pldoc.view-index .pldoc-wrapper-link-top{max-width:73.125rem;margin-left:auto;margin-right:auto;padding:0 2.5rem 1.25rem}.pldoc.view-index .pldoc-wrapper-link-top::after{clear:both;content:"";display:table}.pldoc.view-index .pldoc-link-top{width:100%;float:left;margin-left:0;margin-right:0;text-align:center}
```
clintonb: Use the uncompressed styles from the pattern library. Compressor will compress them.

ecommerce/static/sass/themes/default.scss
```diff
@@ -0,0 +1,30 @@
+// local variables
+// --------------------
+$brand-logo-path: '/static/images/default-logo.png' !default;
+$brand-logo-width: 72px !default;
+$brand-logo-height: 48px !default;
+
+// utilities
+// --------------------
+@import "../utilities/variables";
```
clintonb: What variables are you using here?

clintonb:
Remove the file _sticky-footer.scss. It is not used.

ecommerce/static/sass/themes/default.scss
Nickersoft: @clintonb I believe that this import was required at one point.. I see now that it's not. Will remove.

ecommerce/static/sass/themes/default.scss
```diff
@@ -0,0 +1,30 @@
...
+@import "../utilities/mixins";
+
+// overrides
+// --------------------
+html {
+    font-size:16px;
+}
+
+body {
+    background-color:rgb(248, 248, 248);
```
clintonb: The difference between this color and white seems too subtle to actually matter. Why was this color selected?

Nickersoft: @clintonb This is the default background color of the Drupal site. The white background (also taken from the Drupal site) was included in an effort to distinguish the page header from the rest of the page body.

clintonb: Any copied styles should come from either LMS or the pattern library, not Drupal. The background color for `<body>` in LMS is `#fafafa` or `rgb(250, 250, 250)` which, again, is so subtly different from the default color of `white` that it doesn't seem worth pointing out.

Nickersoft: @clintonb These colors were used to preserve consistency with the LMS stylings. If the background were to be changed to the color of the header, it would go against LMS convention.

ecommerce/static/sass/base/main-rtl.scss
Nickersoft: @clintonb The pattern library's gulpfile compresses the CSS during its build process by default. The only other option I can think of is importing the raw SASS components manually in main.scss. Would that work?

Nickersoft: However, I will say that I have had trouble navigating the pattern library's directory structure in the past, so if there is a central unminified CSS or SASS file I'm overlooking, please let me know

clintonb: Run the gulp task with minification disabled.

clintonb:
ecommerce/static/sass/base/main-rtl.scss is not used and should be removed.

ecommerce/static/sass/base/main-rtl.scss
clintonb: Remove this file. We will deal with RTL support later.

ecommerce/static/sass/components/_buttons.scss
```diff
@@ -1,8 +1,10 @@
...
-// NOTE: used for any basic styling or overriding that isn't covered by Bootstrap or theming
-.container {
-  background-color: white;
+//
+// Buttons
+// --------------------------------------------------
+
+.btn:focus {
+    // Disable Bootstrap default dotted outline
+    outline:none;
```
clintonb: Remove this. Outlines are useful for accessibility.

ecommerce/static/sass/themes/edx.scss
```diff
@@ -0,0 +1,7 @@
+$brand-logo-path: '/static/images/edx-logo.png';
+$brand-logo-width: 93px;
+$brand-logo-height: 47px;
+
+/* We only need to update the logo. The variables above will override those in the Open edX theme (which will handle
+  defining the position of the logo). */
+@import "default";
```
clintonb, line 7: themes/default

ecommerce/static/sass/components/_buttons.scss
Nickersoft: @clintonb My goal here is to disable the custom outline of Bootstrap and have it default to the browser's outline. The dotted outline Bootstrap provides clashes with all existing visual styles, and it would be best if all buttons use the browser's default outline. I will experiment with it to ensure that including this does not disable all outlining, however, from what I remember the button still becomes highlighted when tabbed to (using the browser's default outline) when this line is in place.

ecommerce/static/sass/themes/edx.scss
Nickersoft, line 7: @clintonb Isn't edx.scss in the same directory as default.scss?

clintonb, line 7: Ah. Yes.

ecommerce/static/sass/components/_buttons.scss
clintonb: `outline:default`

Nickersoft: Fixed

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,72 @@
+// Navbars
+// --------------------------------------------------
+
+.nav {
+    > li > a {
+        &:hover, &:focus {
+            outline:none;
```
downzer0: Outlines are critical for keyboard users. If we're removing this outline then we need to make sure there is a very noticeable hover and focus state for these links. I generally advise to leave the outline as certain users expect it.

jimabramson:
python- / template-wise, LGTM

ecommerce/static/sass/components/_navbar.scss
Nickersoft: @clrux As mentioned before, this was to originally intended to prevent Bootstrap outlines, not outlines altogether. I see now that this was not the best approach. As per recommendation of @clintonb, this will be changed to `outline: default`.

ecommerce/static/sass/base/main.scss
```diff
@@ -0,0 +1,26 @@
...
+// --------------------
+@import '../../bower_components/bootstrap-sass/assets/stylesheets/bootstrap';
+@import '../../bower_components/bootstrapaccessibilityplugin/plugins/css/bootstrap-accessibility';
+@import '../../vendor/dataTables/dataTables.bootstrap';
+@import '../../bower_components/fontawesome/scss/font-awesome';
+@import '../../vendor/dataTables/dataTables.fontAwesome';
+
+// pattern library
+// --------------------
+@import './_pattern_library';
```
clintonb: `@import 'pattern_library';`

ecommerce/static/sass/components/_buttons.scss
```diff
@@ -1,8 +1,10 @@
...
-// NOTE: used for any basic styling or overriding that isn't covered by Bootstrap or theming
-.container {
-  background-color: white;
+//
+// Buttons
+// --------------------------------------------------
+
+.btn:focus {
+    // Disable Bootstrap default dotted outline
+    outline:default;
```
clintonb: `inherit` is the correct value here.

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,72 @@
+// Navbars
+// --------------------------------------------------
+
+.nav {
+    > li > a {
+        &:hover, &:focus {
+            outline:default;
```
clintonb: s/default/inherit

ecommerce/static/sass/themes/default.scss
clintonb: Set the value of `$body-bg` in variables.scss and Bootstrap will do this for you.

ecommerce/static/sass/components/_navbar.scss
Nickersoft: Fixed

ecommerce/static/sass/components/_buttons.scss
Nickersoft: Fixed

ecommerce/static/sass/themes/default.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.container {
+    background-color: white;
```
clintonb: This isn't necessary since it's the default.

Nickersoft: @clintonb The default is transparent.

Nickersoft:
I would like to point out that the build is failing because the changes to the pattern library required for this to run have not been merged yet. The PR for this feature is here:

https://github.com/edx/ux-pattern-library/pull/123

clintonb:
@Nickersoft point bower to your branch until it is merged to master.

ecommerce/static/sass/themes/default.scss
clintonb: And...the body's background is already white.

ecommerce/static/sass/components/_buttons.scss
```diff
@@ -35,21 +37,17 @@
...
   &.active,
   .open > &.dropdown-toggle {
-    background-color: $brand-primary-accent;
     border-color: $brand-primary-accent;
+    background-color: $brand-primary-accent;
   }
 
   // Use edX disabled state
   &.disabled,
   &[disabled],
```
downzer0: Our pattern library should generate styles for disabled elements automatically. If using the `disabled` attribute doesn't do it, try adding a `.is-disabled` class. No need to add additional styling here.

ecommerce/static/sass/components/_buttons.scss
```diff
@@ -35,21 +37,17 @@
...
   fieldset[disabled] & {
```
downzer0: Our pattern library should include styles for disabled field elements as well. If using the `disabled` attribute doesn't work, try adding a `.is-disabled` class. no need for additional styling here.

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,72 @@
+// Navbars
```
downzer0: If possible, I recommend using variables (sizes and colors) from our pattern library to bring in some consistency. Check our `_variables.scss` file for a complete list.

ecommerce/static/sass/components/_buttons.scss
Nickersoft: @clrux Actually, this brings up a good point. @clintonb implemented these button styles before this branch was created in an attempt to _merge_ Bootstrap with the pattern library. This branch uses the pattern library alongside Bootstrap. Should all of these button styles simply be removed?

ecommerce/static/sass/themes/default.scss
downzer0: It's best practice too specify background colors, especially when foreground colors are specified. Otherwise, automated accessibility tests may fail.

ecommerce/static/sass/utilities/_variables.scss
```diff
@@ -31,5 +34,15 @@ $border-radius-small: 3px;
 
 $navbar-default-bg: white;
 
-// TODO Use the correct color
 $navbar-default-border: $brand-primary-accent;
+$navbar-height: 74px;
+$navbar-border-bottom-width: 4px;
+
+// Grays
```
downzer0: I recommend using our Pattern Library's thoughtful palettes, found in `_variables.scss`. You can see these visually at http://ux.edx.org for classnames, hex, and rgba values.

ecommerce/templates/courses/course_list.html
```diff
@@ -7,13 +7,13 @@
 {% block content %}
     <div class="container">
         <div class="page-header">
-            <h1>
+            <h1 class="hd-1 emphasized">
               {% trans "Courses" %}
-                <button class="btn btn-primary btn-xs">{% trans "Add New" %}</button>
+                <button class="btn btn-primary btn-small">{% trans "Add New" %}</button>
```
downzer0: I recommend adding context to this button. A screen reader only sees "Add New" but there's no context as to what is being added. Suggest: "Add New Course".

ecommerce/templates/courses/course_list.html
```diff
@@ -7,13 +7,13 @@
...
             </h1>
         </div>
 
-        <table id="courseTable" class="table table-striped table-bordered" cellspacing="0">
+        <table id="courseTable" class="copy copy-base table table-striped table-bordered" cellspacing="0">
```
downzer0, line 16: Recommend adding a [table `caption`](http://webaim.org/techniques/tables/data#caption).

ecommerce/templates/edx/base.html
```diff
@@ -1,26 +1,81 @@
...
+<body>
+{% block navbar %}
+    <nav class="navbar navbar-default" aria-label="main">
+        <div class="container">
+            <!-- Brand and toggle get grouped for better mobile display -->
+            <div class="navbar-header">
+                <button type="button" class="navbar-toggle collapsed" data-toggle="collapse"
+                        data-target="#main-navbar-collapse" aria-expanded="false">
+                    <span class="sr-only">{% trans "Toggle navigation" %}</span>
+                    <span class="icon-bar"></span>
```
downzer0: I would add `aria-hidden="true"` to the three of these. They don't provide any useful context.

ecommerce/static/sass/components/_buttons.scss
downzer0: I would use as much of our pattern library as possible. Doing so now will save efforts later, as we'll eventually be syncing up the LMS and CMS.

ecommerce/static/sass/components/_navbar.scss
Nickersoft: Fixed

ecommerce/static/sass/utilities/_variables.scss
Nickersoft: Fixed

ecommerce/templates/courses/course_list.html
Nickersoft: Fixed

ecommerce/templates/courses/course_list.html
Nickersoft, line 16: @clrux What should the text of this caption say? If I add it, it would have to be beneath the table, otherwise it will interfere with the search field.

ecommerce/static/sass/components/_buttons.scss
Nickersoft: Fixed

ecommerce/static/sass/components/_buttons.scss
Nickersoft: Fixed

ecommerce/static/sass/themes/default.scss
Nickersoft: @clintonb Actually, it is rgb(148, 148, 148).

ecommerce/templates/edx/base.html
Nickersoft: Fixed

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,72 @@
...
+    // Display the user menu as a .navbar-btn when not collapsed to handle vertical centering.
+    .user-menu {
+      @include navbar-vertical-align($input-height-base);
+    }
+  }
+}
+
+.user-menu {
+  > button {
+    border-color: $grayLight;
```
downzer0: For these colors, I recommend using variables from our library.

Nickersoft: Fixed

ecommerce/templates/courses/course_list.html
downzer0, line 16: The caption should briefly describe the contents of the table. It should go at the top, but you can give it a class of `.sr-only` to have it hidden visually.

Nickersoft, line 16: @clrux Oh, okay. Will do.

Nickersoft, line 16: Fixed

Nickersoft:
@clrux @clintonb Updated

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,73 @@
+// Navbars
+// --------------------------------------------------
+
+.nav {
+    > li > a {
```
downzer0: We prefer to use classnames instead of targeting elements. Can you make sure that:
- there are classnames on each of these `li`'s and `a`'s, and
- these styles reflect that change

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,73 @@
...
+        &:hover, &:focus {
```
downzer0: Please see our [Sass/CSS Styleguide](https://github.com/edx/ux-pattern-library/wiki/Styleguide:-Sass-&-CSS#sass-specifically) for proper code formatting.

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,73 @@
...
+  .navbar {
+    // Display the user menu as a .navbar-btn when not collapsed to handle vertical centering.
+    .user-menu {
+      @include navbar-vertical-align($input-height-base);
+    }
+  }
+}
+
+.user-menu {
+  > button {
```
downzer0: We prefer to use classnames instead of targeting elements. Can you make sure that:
- there are classnames on each of these li's and a's, and
- these styles reflect that change

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,73 @@
...
+    border-color: palette(grayscale, light);
+    color: palette(grayscale, dark);
+
+    &:hover, &:focus {
```
downzer0: Please check the formatting here. We prefer to use a new line for each rule. I.e.,

```
&:hover,
&:focus
```

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,73 @@
...
+      float: left;
+      margin: 2px 6px -3px 3px;
+      font-size: 1.2em;
+      color: palette(grayscale, base);
+    }
+  }
+}
+
+.dropdown-menu {
+    > li > a {
```
downzer0: We prefer to use classnames instead of targeting elements. Can you make sure that:
- there are classnames on each of these li's and a's, and
- these styles reflect that change

ecommerce/static/sass/components/_navbar.scss
```diff
@@ -0,0 +1,73 @@
...
+        // Disables default link behavior of pattern library on menu items
+        transition:none;
+        &:hover, &:focus {
```
downzer0: Please correct the formatting here.

ecommerce/static/sass/utilities/_mixins.scss
```diff
@@ -0,0 +1,25 @@
...
+  display: inline-block;
+
+  @if $scaled-height != "" {
+    $scale: $scaled-height / $height;
+    $width: $width * $scale;
+    $height: $scaled-height;
+  }
+  width: $width;
+  height: $height;
+
```
downzer0: Please remove the extra line break here.

ecommerce/templates/courses/course_list.html
```diff
@@ -33,6 +33,7 @@
                 </tr>
             {% endfor %}
             </tbody>
+            <caption class="sr-only">Current Courses</caption>
```
downzer0: I believe this should go directly beneath the `table` element. I.e.,

```
<table>
<caption>
...
```

ecommerce/templates/courses/menu_options.html
```diff
@@ -0,0 +1,12 @@
+{% load core_extras %}
+{% load i18n %}
+
+<li class="{{ additional_class }}">
+    <a href="{% settings_value 'LMS_DASHBOARD_URL' %}">
```
downzer0: Please add a semantic classname to this `a` element for clean targeting in the Sass.

ecommerce/static/sass/components/_navbar.scss
Nickersoft: Fixed

ecommerce/static/sass/components/_navbar.scss
Nickersoft: Fixed

ecommerce/static/sass/components/_navbar.scss
Nickersoft: Fixed

ecommerce/static/sass/components/_navbar.scss
Nickersoft: Fixed

ecommerce/static/sass/components/_navbar.scss
Nickersoft: Fixed

ecommerce/static/sass/components/_navbar.scss
Nickersoft: Fixed

ecommerce/static/sass/utilities/_mixins.scss
Nickersoft: Fixed

ecommerce/templates/courses/course_list.html
Nickersoft: Fixed

ecommerce/static/sass/components/_navbar.scss
downzer0: I still see `button` here. Any way to use a classname instead?

ecommerce/static/sass/components/_navbar.scss
downzer0: I'm not seeing a line break here.

ecommerce/static/sass/components/_navbar.scss
downzer0: I'm not seeing classes here.

ecommerce/static/sass/components/_navbar.scss
downzer0: I'm still not seeing a line break here.

ecommerce/templates/courses/menu_options.html
Nickersoft: Fixed

Nickersoft:
@clrux Updated

downzer0:
@Nickersoft is there any way for me to preview this locally? I keep getting distracted by the animated gifs above where the buttons don't quite look right, among other things.

Nickersoft:
@clrux This can be tested locally by starting the ecommerce server, logging in, then navigating to localhost:8002/courses

downzer0:
@Nickersoft I'm seeing a few accessibility issues:

**Languages**

The "Go" button has no context. I suggest the following:

```
<form>
    <label for="language" id="label-language" class="sr-only">Change language</label>
    <select>...</select>
    <input ... aria-labelledby="label-language" value="Go" />
</form>
```

Or maybe even better/simpler:

```
...
<input ... value="Change language" />
...
```

**View basket button**

The combo button, while keyboard accessible, also has no context. The button itself just says "button" when it has focus with a screen reader. I recommend something like the following:

```
<form>
    <button>Selected button</button>
    <button class="dropdown" ... aria-expanded="false" aria-haspopup="true" aria-controls="cart-basket-menu"><span class="sr-only">Open menu</span></button
    <ul class="dropdown" id="cart-basket-menu" tabindex="-1">...</ul>
</form>
```

Additionally, you'll want to programmatically send focus to the dropdown menu once it opens. Be sure and maintain all keyboard functional controls, such as Esc to close, up and down arrow keys, etc.

**Browse Store menu**

Need to add an `aria-label` to the `ul` that has the Browse Store items. Right now it only says "list", but there's no context as to what the list is.

cc @cptvitamin

Nickersoft:
@clrux Will get right on this. That is one disclaimer I should have made, accessibility testing has not been run on the course list view yet. Thanks for pointing these issues out.

Nickersoft:
@clrux Looking closer at these issues, I realized that this is outside of the scope of the PR. This PR is specific to the theming of the course admin tool. @clintonb or @jimabramson, might either of you know the correct course of action to take regarding these accessibility changes?

jimabramson:
@clintonb may have a different opinion, but if this PR introduces or changes UI elements with accessibility issues, IMO those should be resolved herein.  The only other related upcoming ticket I see is XCOM-337, but I don't think these changes would be considered in scope for that.

ecommerce/templates/edx/base.html
```diff
@@ -1,26 +1,81 @@
...
+    {% compress css %}
+        {# This block is separated to better support browser caching. #}
+        {% block stylesheets %}
+        {% endblock %}
+    {% endcompress %}
 </head>
-<body role="document">
+<body>
+{% block navbar %}
+    <nav class="navbar navbar-default" aria-label="main">
```
cptvitamin: @Nickersoft the aria-label="main" here makes it sound like this is the app's main navigation, which it is not.  I think `aria-label="User"` would be more appropriate here?

Nickersoft:
@jimabramson The focus of this PR is to theme the course admin tool. The accessibility issues mentioned above occur on the Oscar homepage and are unrelated to the theming changes in this ticket.

ecommerce/templates/edx/base.html
Nickersoft: @cptvitamin Would you mind elaborating on the choice of the label `User` over `main`? I'm not familiar with the conventions edX uses in naming its the accessibility attributes of elements.

clintonb:
@clrux @jimabramson The page described by @clrux is not a page we intend users to actually see. Please limit the accessibility review to `/courses/` for the time being.

ecommerce/static/sass/themes/default.scss
```diff
@@ -0,0 +1,25 @@
...
+$brand-logo-width: 72px !default;
+$brand-logo-height: 48px !default;
+
+// utilities
+// --------------------
+@import "../utilities/mixins";
+
+// overrides
+// --------------------
+html {
```
clintonb: Why are the `html` and `.container` overrides overrides? Why not declare these values in main.scss or one of its children?

ecommerce/static/sass/utilities/_variables.scss
```diff
@@ -0,0 +1,25 @@
+// config: variables
+// ====================
+// NOTE: these variables override Bootstrap-centric variables.
+
+// FontAwesome font directory
+$fa-font-path: "/static/bower_components/fontawesome/fonts";
+
+// UX Pattern Library font directory
+$ux-pattern-library-font-path: '/static/bower_components/edx-ux-pattern-library/_src/pattern-library/fonts';
```
clintonb: This should be `$font-path` because...silly reasons.

bower.json
```diff
@@ -18,6 +18,7 @@
     "underscore": "~1.8.2",
     "bootstrapaccessibilityplugin": "~1.0.4",
     "datatables": "~1.10.7",
-    "fontawesome": "~4.3.0"
+    "fontawesome": "~4.3.0",
+    "edx-ux-pattern-library": "https://github.com/edx/ux-pattern-library.git"
```
clintonb: Specify a commit hash/version.

ecommerce/static/sass/themes/default.scss
Nickersoft: Fixed

ecommerce/static/sass/utilities/_variables.scss
Nickersoft: Fixed

ecommerce/templates/edx/base.html
cptvitamin: "Main" would be the navigation in the banner (How it works, Courses, Schools & Partners).  We also have "Course Material" which is the main navigation within a course, (Courseware, Course Info, Syllabus...)  , then there is "Course" which is the Weekly topics, and "Sequence" which are the individual sequences in a Topic.  I would expect a list of options for managing my account, logging in and logging out etc, would be labelled something like "User"  as in my User menu.  Account would also be acceptable. If you need to find out conventions, you can just "View Source". in the app.

Nickersoft: @cptvitamin Alright. Will make this change.

Nickersoft: Fixed

clintonb:
:+1: (once tests pass)

downzer0:
Hi @Nickersoft, I found a few more things in the `/courses` view:
- Add new button needs context (add `<span class="sr-only">Course</span>` in the button text)
- The filter needs a label. It may have the `"sr-only"` class

Nickersoft:
@clrux Updated

ecommerce/static/js/views/course_list_view.js
```diff
@@ -1,5 +1,8 @@
 require(['jquery', 'dataTablesBootstrap'], function ($) {
     $(document).ready(function () {
+
+        var $empty_label = $('<label>').addClass('sr-only').html('Search');
```
downzer0: Hmm, does this filter or search? This label should describe what this mechanism does.

Nickersoft: @clrux Technically it filters. I corrected the label to be more appropriate to the widget's functionality.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -1,5 +1,8 @@
 require(['jquery', 'dataTablesBootstrap'], function ($) {
     $(document).ready(function () {
+
+        var $empty_label = $('<label>').addClass('sr-only').html('Filter');
```
clintonb: Why not make this text exactly match the placeholder text?

Nickersoft: Good call @clintonb. I think that because it's a label I was aiming for brevity. I guess that doesn't matter if the text is invisible anyway. I'll make the change.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -1,5 +1,8 @@
 require(['jquery', 'dataTablesBootstrap'], function ($) {
     $(document).ready(function () {
+
+        var $empty_label = $('<label>').addClass('sr-only').html('Filter by org or course ID');
```
clintonb: The text still needs to be translated. Create a variable and share it in both locations.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -8,6 +11,11 @@ require(['jquery', 'dataTablesBootstrap'], function ($) {
             }
         });
 
-        $('#courseTable_filter input').attr('placeholder', gettext('Filter by org or course ID'));
+        $('#courseTable_filter label').prepend($empty_label);
```
clintonb, line 16: If a label already exists, just call `addClass` and `html` on it.

clintonb:
The hover state for links has a double underline. This is not desirable.

ecommerce/static/sass/base/_base.scss
```diff
@@ -0,0 +1,10 @@
+// overrides
```
clintonb: These aren't really overrides. They are the default values.

ecommerce/static/js/views/course_list_view.js
Nickersoft: Fixed

ecommerce/static/js/views/course_list_view.js
Nickersoft, line 16: @clintonb There is, but it wraps the input field. If I add `sr-only` to it, it hides the entire search box. As a result, I need to add a new label that is a child element.

Nickersoft:
@clintonb I am not seeing this. What are the steps to reproduce?

clintonb:
@Nickersoft create a normal link (not a navbar link) on the page. Observe the odd decision to add a bottom border to `a:hover`.

ecommerce/static/sass/base/_base.scss
Nickersoft: @clintonb They override bootstrap rules that conflict with the pattern library.

clintonb: Not really. This is all new. Regardless, this file is where the base styles are stored, overrides are not.

Nickersoft:
@clintonb I see that now. Apparently Bootstrap was interfering with the pattern library. It's fixed now.

clintonb:
@Nickersoft make sure you rebase before merging. The courses URLs have changed a small bit.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -1,5 +1,8 @@
 require(['jquery', 'dataTablesBootstrap'], function ($) {
     $(document).ready(function () {
+
+        var $empty_label = $('<label>').addClass('sr-only').html(gettext('Filter by org or course ID'));
```
clintonb: Again, use a variable for the text.

Nickersoft: Fixed

ecommerce/static/js/views/course_list_view.js
```diff
@@ -8,6 +13,11 @@ require(['jquery', 'dataTablesBootstrap'], function ($) {
             }
         });
 
-        $('#courseTable_filter input').attr('placeholder', gettext('Filter by org or course ID'));
+        $('#courseTable_filter label').prepend($empty_label);
+
+        $('#courseTable_filter input')
+            .attr('placeholder', gettext('Filter by org or course ID'))
```
clintonb: `sr_filter_label`

ecommerce/templates/courses/menu_options.html
```diff
@@ -0,0 +1,8 @@
+{% load core_extras %}
+{% load i18n %}
+
+<li class="{{ additional_class }}"><a class="nav-link" href="{% settings_value 'LMS_DASHBOARD_URL' %}">{% trans "Student Dashboard" %}</a></li>
+<li class="{{ additional_class }}"><a class="nav-link" href="{% url 'courses_list' %}">{% trans "Course Admin Tool" %}</a></li>
```
clintonb: Make sure you rebase and run the service. This should be `courses:list`

Nickersoft: Fixed

clintonb:
Merging. @clrux if you have additional feedback (esp. regarding accessibility), please continue commenting here. We will address it as we continue developing the list view.

PR #203, opened 2015-07-15
BUG: Fixed Unicode support for migration command and view

clintonb:
The migration command now uses Unicode exclusively, and the view has been updated to use io.StringIO which also uses Unicode.

### DIFF
--- a/ecommerce/extensions/catalogue/tests/test_migrate_course.py
+++ b/ecommerce/extensions/catalogue/tests/test_migrate_course.py
@@ -1,3 +1,5 @@
+# coding=utf-8
+from __future__ import unicode_literals
 import datetime
 import json
 import logging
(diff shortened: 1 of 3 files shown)


rlucioni:
:+1:

rlucioni:
A test case verifying that the management command is able to handle Unicode would be great.

clintonb:
@rlucioni Tests updated.

ecommerce/extensions/catalogue/tests/test_migrate_course.py
```diff
@@ -64,7 +66,7 @@ def assert_seat_valid(self, seat, seat_type):
         """ Verify the given seat is configured correctly. """
         certificate_type = Course.certificate_type_for_mode(seat_type)
 
-        expected_title = u'Seat in A Test Course with {} certificate'.format(certificate_type)
+        expected_title = u'Seat in A Tést Côurse with {} certificate'.format(certificate_type)
```
rlucioni: Minor: Since you've imported `unicode_literals`, you could drop the `u''` prefixes from these strings.

rlucioni:
Nice, thanks @clintonb.

ecommerce/extensions/catalogue/tests/test_migrate_course.py
clintonb: Updated.

jimabramson:
:+1: (side question: how / when / how much do we want to be using `unicode_literals`?  [tricky subject apparently](http://python-future.org/unicode_literals.html).)

clintonb:
@jimabramson this is definitely a file where we _want_ to support Unicode for nearly all strings since course names can contain Unicode values. If we just upgraded to Python 3 we wouldn't have this problem. :stuck_out_tongue:

PR #209, opened 2015-07-17
Unified Course Seat Creation

clintonb:
The creation of all course seats (for tests, via migration, etc.) is now handled via Course.create_or_update_seat(). Consolidating this logic into a single location removes duplicate code and ensures we are fully testing the actual code paths used in production.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ clean:
 test_python: clean
 	python manage.py compress --settings=ecommerce.settings.test
 	DISABLE_MIGRATIONS=True coverage run --branch --source=ecommerce ./manage.py test ecommerce \
-	--settings=ecommerce.settings.test --with-ignore-docstrings
+	--settings=ecommerce.settings.test --with-ignore-docstrings --logging-level=DEBUG
 	coverage report
 
 quality:
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -1,15 +1,23 @@
+from __future__ import unicode_literals
 import logging
 
 from django.db import models, transaction
+from django.utils.text import slugify
 from oscar.core.loading import get_model
 from simple_history.models import HistoricalRecords
 import waffle
 
 from ecommerce.courses.exceptions import PublishFailed
 from ecommerce.courses.publishers import LMSPublisher
+from ecommerce.extensions.catalogue.utils import generate_sku
 
 logger = logging.getLogger(__name__)
+Category = get_model('catalogue', 'Category')
+Partner = get_model('partner', 'Partner')
+Product = get_model('catalogue', 'Product')
+ProductCategory = get_model('catalogue', 'ProductCategory')
 ProductClass = get_model('catalogue', 'ProductClass')
+StockRecord = get_model('partner', 'StockRecord')
 
 
 class Course(models.Model):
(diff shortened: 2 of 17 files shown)


clintonb:

Makefile
```diff
@@ -44,7 +44,7 @@ clean:
 test_python: clean
 	python manage.py compress --settings=ecommerce.settings.test
 	DISABLE_MIGRATIONS=True coverage run --branch --source=ecommerce ./manage.py test ecommerce \
-	--settings=ecommerce.settings.test --with-ignore-docstrings
+	--settings=ecommerce.settings.test --with-ignore-docstrings --logging-level=DEBUG
```
Nickersoft, line 47: Is this necessary for production?

clintonb, line 47: Makefiles aren't used in production. This is necessary to ensure we capture all logs should tests fail, and ease debugging.

ecommerce/courses/models.py
```diff
@@ -49,19 +81,83 @@ def certificate_type_for_mode(cls, mode):
...
+        Returns:
+            Product:  The seat that has been created or updated.
+        """
+
+        certificate_type = certificate_type.lower()
+        course_id = unicode(self.id)
+        slug = 'child-cs-{}-{}'.format(certificate_type, slugify(course_id))
+
+        if id_verification_required:
+            slug += 'id-verified'
```
jimabramson: should there be a leading `'-'` here?

ecommerce/courses/tests/test_models.py
```diff
@@ -118,3 +125,50 @@ def test_save_without_publish(self):
...
+    def test_create_or_update_seat(self):
+        """ Verify the method creates or updates a seat Product. """
+        course = Course.objects.create(id='a/b/c', name='Test Course')
+
+        # Test creation
+        certificate_type = 'honor'
+        id_verification_required = True
+        price = 0
+        course.create_or_update_seat(certificate_type, id_verification_required, price)
+        self.assertEqual(course.products.count(), 2)
```
jimabramson, line 166: this is counterintuitive - it's 2 because of the parent product?

ecommerce/courses/models.py
clintonb: Yes. Updated.

ecommerce/courses/tests/test_models.py
clintonb, line 166: Yes. Added comments to this assertion and the one below.

jimabramson:
:+1: 

optional: consider using constants instead of literals for the certificate type values e.g. 'honor', 'verified'

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -220,26 +221,11 @@ def request_callback(_method, _uri, _headers):
 
 @ddt.ddt
 @override_settings(EDX_API_KEY='foo')
-class EnrollmentFulfillmentCreditSeatModuleTests(CourseCatalogTestMixin, FulfillmentTestMixin, TestCase):
+class EnrollmentFulfillmentCreditSeatModuleTests(EnrollmentFulfillmentTestMixin, TestCase):
```
rlucioni, line 224: As part of the clean up work you're doing here, consider removing the `EnrollmentFulfillmentCreditSeatModuleTests` class and moving the single test contained within into the `EnrollmentFulfillmentModuleTests`. Credit Seats are fulfilled by the `EnrollmentFulfillmentModule`, just like any other Seat, and keeping a whole test class around just for testing fulfillment of a credit Seat seems messy.

rlucioni:
:+1:, regardless of whether you choose to address my comment about the tests.

ecommerce/extensions/fulfillment/tests/test_modules.py
clintonb, line 224: That's a bit more than I want to touch with this PR. If you feel it's important, feel free to create a ticket and either do the work or add it to the backlog.

rlucioni, line 224: Cool.

EDIT: [XCOM-503](https://openedx.atlassian.net/browse/XCOM-503)

PR #210, opened 2015-07-20
Added Course and Product API Endpoints

clintonb:
XCOM-505

@jimabramson @rlucioni @wedaly 

FYI @AlasdairSwan @Nickersoft

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -1,8 +1,10 @@
 """Serializers for order and line item data."""
-# pylint: disable=abstract-method
-from oscar.core.loading import get_model
-from rest_framework import serializers
 
+from oscar.core.loading import get_model, get_class
+from rest_framework import serializers
+from rest_framework.reverse import reverse
+
+from ecommerce.courses.models import Course
 from ecommerce.extensions.payment.constants import ISO_8601_FORMAT
 
 BillingAddress = get_model('order', 'BillingAddress')
@@ -11,6 +13,10 @@ Order = get_model('order', 'Order')
 Product = get_model('catalogue', 'Product')
 ProductAttributeValue = get_model('catalogue', 'ProductAttributeValue')
 Refund = get_model('refund', 'Refund')
+Selector = get_class('partner.strategy', 'Selector')
+
+COURSE_DETAIL_VIEW = 'api:v2:course-detail'
+PRODUCT_DETAIL_VIEW = 'api:v2:product-detail'
 
 
 class BillingAddressSerializer(serializers.ModelSerializer):
@@ -22,30 +28,52 @@ class BillingAddressSerializer(serializers.ModelSerializer):
         fields = ('first_name', 'last_name', 'line1', 'line2', 'postcode', 'state', 'country', 'city')
 
 
-class ProductAttributeValueSerializer(serializers.Serializer):
+class ProductAttributeValueSerializer(serializers.ModelSerializer):
     """ Serializer for ProductAttributeValue objects. """
     name = serializers.SerializerMethodField()
-    value = serializers.CharField()
-    type = serializers.SerializerMethodField()
+    value = serializers.SerializerMethodField()
 
     def get_name(self, instance):
         return instance.attribute.name
 
-    def get_type(self, instance):
-        return instance.attribute.type
+    def get_value(self, obj):
+        return obj.value
 
     class Meta(object):
         model = ProductAttributeValue
-        fields = ('name', 'value', 'type')
+        fields = ('name', 'value',)
 
 
-class ProductSerializer(serializers.ModelSerializer):
+class ProductSerializer(serializers.HyperlinkedModelSerializer):
     """ Serializer for Products. """
-    attribute_values = ProductAttributeValueSerializer(many=True)
+    attribute_values = ProductAttributeValueSerializer(many=True, read_only=True)
+    product_class = serializers.SerializerMethodField()
+    price = serializers.SerializerMethodField()
+    is_available_to_buy = serializers.SerializerMethodField()
+
+    def get_product_class(self, product):
+        return product.get_product_class().name
+
+    def get_price(self, product):
+        info = self._get_info(product)
+        if info.availability.is_available_to_buy:
+            return serializers.DecimalField(max_digits=10, decimal_places=2).to_representation(info.price.excl_tax)
+        return None
+
+    def _get_info(self, product):
+        info = Selector().strategy().fetch_for_product(product)
+        return info
+
+    def get_is_available_to_buy(self, product):
+        info = self._get_info(product)
+        return info.availability.is_available_to_buy
 
     class Meta(object):
         model = Product
-        fields = ('attribute_values',)
+        fields = ('id', 'url', 'product_class', 'title', 'price', 'expires', 'attribute_values', 'is_available_to_buy',)
+        extra_kwargs = {
+            'url': {'view_name': PRODUCT_DETAIL_VIEW},
+        }
 
 
 class LineSerializer(serializers.ModelSerializer):
@@ -68,7 +96,7 @@ class OrderSerializer(serializers.ModelSerializer):
         fields = ('number', 'date_placed', 'status', 'currency', 'total_excl_tax', 'lines', 'billing_address')
 
 
-class PaymentProcessorSerializer(serializers.Serializer):
+class PaymentProcessorSerializer(serializers.Serializer):   # pylint: disable=abstract-method
     """ Serializer to use with instances of processors.BasePaymentProcessor """
 
     def to_representation(self, instance):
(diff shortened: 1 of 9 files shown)


ecommerce/extensions/api/serializers.py
```diff
@@ -22,30 +28,45 @@ class Meta(object):
...
     class Meta(object):
         model = ProductAttributeValue
-        fields = ('name', 'value', 'type')
+        fields = ('name', 'value',)
 
 
-class ProductSerializer(serializers.ModelSerializer):
+class ProductSerializer(serializers.HyperlinkedModelSerializer):
     """ Serializer for Products. """
+    course = serializers.CharField
```
rlucioni: Looking towards a future where we might want to sell products other than Seats, does it make sense to include a link to a course detail view when serializing products? Would the `course_key` attribute alone be sufficient? Alternatively, could we include a nested `links` object, which contains any links relevant to the serialized product?

clintonb: `course_key` assumes clients know how to get the course details, an assumption we want to avoid. Having dealt with nested `_links` dicts for the past couple weeks, I would rather avoid that baggage. In this case if a product has no associated course, the value will simply be `null`.

ecommerce/extensions/api/v2/urls.py
```diff
@@ -49,3 +50,10 @@
     url(r'^payment/', include(PAYMENT_URLS, namespace='payment')),
     url(r'^refunds/', include(REFUND_URLS, namespace='refunds')),
 )
+
+router = ExtendedSimpleRouter()
+router.register(r'courses', views.CourseViewSet)\
```
jimabramson, line 55: nitpick: is that `\` mandatory?

clintonb, line 55: The alternative is to create a variable and call `register()` on it.

ecommerce/settings/base.py
```diff
@@ -422,4 +422,5 @@ def get_lms_url(path):
 # END COOKIE CONFIGURATION
 
 # Standard regex for course_id.
-COURSE_ID_PATTERN = r'(?P<course_id>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'
+COURSE_ID_REGEX = r'[^/+]+(/|\+)[^/+]+(/|\+)[^/]+'
```
jimabramson, line 425: Unrelated to this PR, but this will not match `course-v1` IDs.  That only seems to be used in the credit app - @wedaly is this a concern?

ecommerce/extensions/api/serializers.py
```diff
@@ -22,30 +28,45 @@ class Meta(object):
...
+    price = serializers.SerializerMethodField()
+
+    def get_product_class(self, product):
+        return product.get_product_class().name
+
+    def get_price(self, product):
+        info = Selector().strategy().fetch_for_product(product)
+        if info.availability.is_available_to_buy:
+            return serializers.DecimalField(max_digits=10, decimal_places=2).to_representation(info.price.excl_tax)
+        return None
```
jimabramson, line 61: are clients to assume that products with a price of None are unavailable, otherwise available?

ecommerce/settings/base.py
wedaly, line 425: Good catch!  That is definitely a concern, since users going to the checkout page for a course with a new-style course ID would presumably see a 404.

FYI: @awais786 @zubair-arbi

ecommerce/extensions/api/serializers.py
jimabramson: is it at all possible to have the serializer field named `course_url` instead of `course` ?   same question for the product(_url) in the course serializer.

ecommerce/settings/base.py
clintonb, line 425: This regex is exactly the same as that used in edx-platform: https://github.com/edx/edx-platform/blob/80cf4d6ecbcf27805f9c0f9cedc237329fb8b770/lms/envs/common.py#L590.

ecommerce/extensions/api/serializers.py
clintonb, line 61: That is correct. It might also be useful to expose an `is_available_to_buy` field.

ecommerce/extensions/api/serializers.py
clintonb: Removed the course field, since it's not being used (and it is easier to add a field later than it is to remove/modify one). Updated `products` to `products_url`.

ecommerce/extensions/api/serializers.py
jimabramson, line 61: offhand, it seems like it'd be better to be explicit about availability.

ecommerce/settings/base.py
jimabramson, line 425: Sorry, I was wrong about that.  Looking more closely it should work.

ecommerce/extensions/api/serializers.py
clintonb, line 61: Added `is_available_to_buy` field.

jimabramson:
LGTM, but is there a need for coverage of the nested product views?

clintonb:
@jimabramson adding tests for `ProductViewSet` as we speak/type/communicate.

clintonb:
Tests added for `ProductViewSet`.

ecommerce/extensions/api/v2/tests/views/test_courses.py
```diff
@@ -0,0 +1,211 @@
...
+class ProductViewSetTests(TestServerUrlMixin, CourseCatalogTestMixin, UserMixin, TestCase):
+    maxDiff = None
+
+    def setUp(self):
+        super(ProductViewSetTests, self).setUp()
+        self.user = self.create_user(is_staff=True)
+        self.client.login(username=self.user.username, password=self.password)
+        self.course = Course.objects.create(id='edX/DemoX/Demo_Course', name='Test Course')
+
+        # TODO Update the expiration date by 2099-12-31
```
jimabramson: please create a JIRA ticket.

clintonb: https://openedx.atlassian.net/browse/XCOM-512

ecommerce/extensions/api/v2/tests/views/test_courses.py
```diff
@@ -0,0 +1,211 @@
...
+
+        path = reverse('api:v2:product-detail', kwargs={'pk': self.seat.id})
+        response = self.client.put(path, json.dumps(data), JSON_CONTENT_TYPE)
+        self.assertEqual(response.status_code, 200, response.content)
+
+        product = Product.objects.get(id=self.seat.id)
+        self.assertEqual(product.title, data['title'])
+        self.assertDictEqual(json.loads(response.content), self.serialize_product(product))
+
+    def test_list_for_course(self):
```
jimabramson: not sure if this is overkill but it seems like testing that another course's product(s) are not returned here would be appropriate.

ecommerce/extensions/api/v2/tests/views/test_courses.py
jimabramson: haha, i was kidding!

jimabramson:
:+1:

ecommerce/extensions/api/v2/tests/views/test_courses.py
clintonb: Updated. Also, I moved the this class to test_products.py.

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def serialize_product(self, product):
+        """ Serializes a Product to a Python dict. """
+        attribute_values = [{'name': av.attribute.name, 'value': av.value} for av in product.attribute_values.all()]
+        data = {
+            'id': product.id,
+            'url': self.get_full_url(reverse('api:v2:product-detail', kwargs={'pk': product.id})),
+            'product_class': unicode(product.get_product_class()),
+            'title': product.title,
+            'expires': product.expires.isoformat()[:-6] + 'Z' if product.expires else None,
```
rlucioni, line 41: Minor: We keep the `ISO_8601_FORMAT` constant around in the payment processor constants, so this could be a little cleaner as:

``` python
from ecommerce.extensions.payment.constants import ISO_8601_FORMAT
...
product.expires.strftime(ISO_8601_FORMAT)
```

requirements/base.txt
```diff
@@ -18,3 +18,6 @@ python-dateutil==2.4.2
 pytz==2015.4
 requests==2.6.0
 suds==0.4
+
+# TODO https://github.com/chibisov/drf-extensions/pull/87
```
rlucioni, line 22: Would you mind creating a dev story for updating this? The maintainer claims he'll merge this soon.

ecommerce/extensions/api/v2/tests/views/test_products.py
clintonb, line 41: Good call. I want to move `ISO_8601_FORMAT` out of `payments` and into `core`. I will take care of both tasks in a follow-on PR.

rlucioni:
:+1:

PR #213, opened 2015-07-21
Moved ISO_8601_FORMAT to core app

clintonb:
This constant is shared across multiple apps, not just payment, so should be in a more centralized location.

XCOM-505

Follow up from #210


rlucioni:
:+1:

PR #214, opened 2015-07-21
Course Detail View

clintonb:
XCOM-507

### DIFF
--- a/ecommerce/static/js/collections/drf_pageable_collection.js
+++ b/ecommerce/static/js/collections/drf_pageable_collection.js
@@ -1,0 +1,34 @@
+define([
+        'backbone',
+        'underscore',
+        'backbone.paginator'
+    ],
+    function (Backbone, _) {
+        'use strict';
+
+        return Backbone.PageableCollection.extend({
+            state: {
+                pageSize: 20
+            },
+
+            parseRecords: function (resp, options) {
+                return resp.results;
+            },
+
+
+            parseState: function (resp, queryParams, state, options) {
+                return {
+                    totalRecords: resp.count
+                };
+            },
+
+            parseLinks: function (resp, options) {
+                return {
+                    first: null,
+                    next: resp.next,
+                    prev: resp.previous
+                };
+            }
+        });
+    }
+);
--- a/ecommerce/static/js/config.js
+++ b/ecommerce/static/js/config.js
@@ -1,16 +1,23 @@
 require.config({
     baseUrl: '/static/',
     paths: {
-        backbone: 'bower_components/backbone/backbone',
-        bootstrap: 'bower_components/bootstrap-sass/assets/javascripts/bootstrap',
-        bootstrap_accessibility: 'bower_components/bootstrapaccessibilityplugin/plugins/js/bootstrap-accessibility',
-        dataTables: 'bower_components/datatables/media/js/jquery.dataTables',
-        dataTablesBootstrap: 'vendor/dataTables/dataTables.bootstrap',
-        jquery: 'bower_components/jquery/dist/jquery',
+        'backbone': 'bower_components/backbone/backbone',
+        'backbone.paginator': 'bower_components/backbone.paginator/lib/backbone.paginator',
+        'bootstrap': 'bower_components/bootstrap-sass/assets/javascripts/bootstrap',
+        'bootstrap_accessibility': 'bower_components/bootstrapaccessibilityplugin/plugins/js/bootstrap-accessibility',
+        'collections': 'js/collections',
+        'dataTables': 'bower_components/datatables/media/js/jquery.dataTables',
+        'dataTablesBootstrap': 'vendor/dataTables/dataTables.bootstrap',
+        'jquery': 'bower_components/jquery/dist/jquery',
         'jquery-cookie': 'bower_components/jquery-cookie/jquery.cookie',
-        requirejs: 'bower_components/requirejs/require',
-        underscore: 'bower_components/underscore/underscore',
-        views: 'js/views'
+        'models': 'js/models',
+        'moment': 'bower_components/moment/moment',
+        'requirejs': 'bower_components/requirejs/require',
+        'templates': 'templates',
+        'text': 'bower_components/text/text',
+        'underscore': 'bower_components/underscore/underscore',
+        'underscore.string': 'bower_components/underscore.string/dist/underscore.string',
+        'views': 'js/views'
     },
     shim: {
         bootstrap: {
@@ -27,6 +34,6 @@ require.config({
         },
         'jquery-cookie': {
             deps: ['jquery']
-        },
+        }
     }
 });
--- a/ecommerce/static/js/models/course_model.js
+++ b/ecommerce/static/js/models/course_model.js
@@ -1,0 +1,35 @@
+define([
+        'backbone',
+        'underscore',
+        'collections/product_collection',
+        'models/course_seat_model'
+    ],
+    function (Backbone, _, ProductCollection, CourseSeatModel) {
+        'use strict';
+
+        return Backbone.Model.extend({
+            urlRoot: '/api/v2/courses/',
+            defaults: {
+                name: ''
+            },
+
+            getProducts: function () {
+                if (_.isUndefined(this._products)) {
+                    this._products = new ProductCollection();
+                    this._products.url = this.get('products_url');
+                    return this._products.getFirstPage({fetch: true});
+                }
+
+                return this._products;
+            },
+
+            getSeats: function () {
+                // Returns the seat products
+                return this.getProducts().filter(function (product) {
+                    // Filter out parent products since there is no need to display or modify.
+                    return (product instanceof CourseSeatModel) && product.get('structure') !== 'parent';
+                });
+            }
+        });
+    }
+);
--- a/ecommerce/static/js/views/course_detail_view.js
+++ b/ecommerce/static/js/views/course_detail_view.js
@@ -1,0 +1,78 @@
+define([
+        'jquery',
+        'backbone',
+        'underscore',
+        'underscore.string',
+        'moment',
+        'models/course_model',
+        'text!templates/course_detail.html',
+        'text!templates/_course_seat.html'
+    ],
+    function ($,
+              Backbone,
+              _,
+              _s,
+              moment,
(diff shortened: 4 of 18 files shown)


clintonb:
**List View**
list

**Detail View**
detail

ecommerce/static/js/views/course_detail_view.js
```diff
@@ -0,0 +1,48 @@
...
+             CourseDetailTemplate) {
+    var CourseDetailView = Backbone.View.extend({
+        initialize: function () {
+            var self = this,
+                course_id = self.$el.data('course-id');
+
+            self.course = new CourseModel({id: course_id});
+            self.course.fetch({
+                success: function (course) {
+                    // TODO Make product retrieval more synchronous
```
clintonb: @AlasdairSwan please advise on options to bundle the fetching of the course and its products. What I have here doesn't seem right, especially sense subsequent calls to `course.getProducts()` will return a collection instead of a jqXHR.

clintonb:
Note that this PR _does not_ address client-side routing. That comes later.

ecommerce/static/js/views/course_detail_view.js
AlasdairSwan: @clintonb let's meet today to discuss the approach, I would have expected the course list view to fetch the collection and then loop through it passing the models to the course details views. But when I look at the course list view it isn't a view at all! I think I need to understand the current implementation more fully first and we can take it from there.

ecommerce/static/js/views/course_detail_view.js
```diff
@@ -0,0 +1,48 @@
...
+                seats: self.course.getSeats()
+            };
+            html = _.template(CourseDetailTemplate)(templateData);
+            self.$el.html(html)
+        }
+    });
+
+    $('[data-view=course-detail]').each(function () {
+        new CourseDetailView({el: $(this)})
+    });
```
AlasdairSwan: Why isn't this in a function within the view?

ecommerce/static/js/views/course_detail_view.js
```diff
@@ -0,0 +1,48 @@
...
+                    // TODO Make product retrieval more synchronous
+                    course.getProducts().done(function () {
+                        self.render();
+                    });
+                }
+            })
+        },
+
+        render: function () {
+            var html, templateData, self = this;
```
AlasdairSwan: Why are you aliasing this? You shouldn't ever need to, and doing so makes the code confusing to read.

ecommerce/static/js/models/course_model.js
```diff
@@ -0,0 +1,54 @@
+define(['backbone', 'underscore', 'collections/product_collection'], function (Backbone, _, ProductCollection) {
+    return Backbone.Model.extend({
+        urlRoot: '/api/v2/courses/',
+        defaults: {
+            name: ''
+        },
+
+        getProducts: function () {
+            var self = this;
```
AlasdairSwan: Why not just use this?

ecommerce/static/sass/views/_course_detail.scss
```diff
@@ -0,0 +1,26 @@
+[data-view=course-detail] {
```
AlasdairSwan: Can we use a class instead of a data-attribute here to be more performant? Or is that a reason not to?

ecommerce/static/js/models/course_model.js
clintonb: Updated.

ecommerce/static/js/views/course_detail_view.js
clintonb: Updated.

ecommerce/static/sass/views/_course_detail.scss
clintonb: Updated.

ecommerce/static/js/models/course_model.js
```diff
@@ -0,0 +1,51 @@
...
+
+        getType: function () {
+            // Returns the type of Course (e.g. honor, verified, credit) represented by this model.
+            var seats = this.getSeats();
+
+            if ('credit' in seats) {
+                return 'Credit';
+            } else if ('verified' in seats) {
+                return 'Verified'
+            }
```
jimabramson, line 32: why isn't no-id-professional included here?

clintonb, line 32: This has been cleaned up and moved to the CourseSeatModel.

ecommerce/static/js/views/course_detail_view.js
clintonb: Updated.

ecommerce/static/js/views/course_detail_view.js
```diff
@@ -0,0 +1,69 @@
...
+            el: '[data-view=course-detail]',
+
+            initialize: function () {
+                var self = this,
+                    course_id = self.$el.data('course-id');
+
+                this.course = new CourseModel({id: course_id});
+                this.course.fetch({
+                    success: function (course) {
+                        // TODO Make product retrieval more synchronous
```
clintonb: Reminder for me and @AlasdairSwan to discuss this later.

ecommerce/static/js/views/course_detail_view.js
```diff
@@ -0,0 +1,77 @@
...
+            el: '[data-view=course-detail]',
+
+            initialize: function () {
+                var self = this,
+                    course_id = self.$el.data('course-id');
+
+                this.course = new CourseModel({id: course_id});
+                this.course.fetch({
+                    success: function (course) {
+                        self.render();
```
clintonb, line 31: @AlasdairSwan I believe I gleaned enough info from your previous comment to fix this. After the course is loaded I am rendering the course template with a placeholder for the seats. Once that is rendered I retrieve the products and render a template for each.

ecommerce/static/js/views/course_detail_view.js
```diff
@@ -0,0 +1,77 @@
...
+            },
+
+            renderSeats: function () {
+                var html,
+                    $seatHolder = $('.course-seats', this.$el);
+
+                this.getSeats().forEach(function (seat) {
+                    html = _.template(CourseSeatTemplate)({seat: seat, moment: moment});
+                    $seatHolder.append(html);
+                });
```
AlasdairSwan, line 72: There is a performance hit every time the DOM is updated so it would be better to keep adding to the html variable and moving $seatHolder.appned(html) to after the each loop.

ecommerce/static/js/views/course_detail_view.js
```diff
@@ -0,0 +1,77 @@
...
+              _,
+              _s,
+              moment,
+              CourseModel,
+              CourseDetailTemplate,
+              CourseSeatTemplate) {
+        'use strict';
+
+        return Backbone.View.extend({
+            el: '[data-view=course-detail]',
```
AlasdairSwan: Is there a reason a data attribute is being used instead of a class? It's more performant to use classes as selectors.

clintonb: Updated to `.course-detail-view`.

ecommerce/static/js/views/course_detail_view.js
clintonb, line 72: Updated.

clintonb:
Any other feedback? Merging this now would help streamline #230.

ecommerce/static/js/collections/drf_pageable_collection.js
```diff
@@ -0,0 +1,34 @@
...
+    function (Backbone, _) {
+        'use strict';
+
+        return Backbone.PageableCollection.extend({
+            state: {
+                pageSize: 20
+            },
+
+            parseRecords: function (resp, options) {
+                return resp['results'];
```
AlasdairSwan: dot notation is nicer 

```
 resp.results
```

ecommerce/static/js/collections/drf_pageable_collection.js
```diff
@@ -0,0 +1,34 @@
...
+            },
+
+
+            parseState: function (resp, queryParams, state, options) {
+                return {
+                    totalRecords: resp['count']
```
AlasdairSwan: dot notation is nicer

ecommerce/static/js/collections/drf_pageable_collection.js
```diff
@@ -0,0 +1,34 @@
...
+                };
+            },
+
+            parseLinks: function (resp, options) {
+                return {
+                    first: null,
+                    next: resp['next'],
+                    prev: resp['previous']
```
AlasdairSwan: dot notation is nicer

ecommerce/static/js/config.js
```diff
@@ -2,14 +2,21 @@ require.config({
     baseUrl: '/static/',
     paths: {
         backbone: 'bower_components/backbone/backbone',
+        'backbone.paginator': 'bower_components/backbone.paginator/lib/backbone.paginator',
```
AlasdairSwan, line 5: are quotes necessary for special characters?

ecommerce/static/js/collections/drf_pageable_collection.js
clintonb: Updated.

ecommerce/static/js/collections/drf_pageable_collection.js
clintonb: Updated.

ecommerce/static/js/collections/drf_pageable_collection.js
clintonb: Updated.

ecommerce/static/js/config.js
clintonb, line 5: Yes. Bad things happen without quotes.

AlasdairSwan, line 5: Nitpick: Then maybe it would look cleaner to wrap all of the keys in quotes.

AlasdairSwan:
Once tests pass :+1:

ecommerce/static/js/config.js
clintonb, line 5: Lazy Clinton says, "no". OCD Clinton says, "updated".

jimabramson:
:+1:

PR #216, opened 2015-07-22
Less verbose output from make static

jimabramson:
These commands emit thousands of lines of output, which makes Travis console output hard to work with.

@clintonb @rlucioni @Nickersoft  objections?


rlucioni:
None from me. :+1: .

clintonb:
:+1:

jimabramson:
merging without waiting for coveralls.

PR #218, opened 2015-07-22
Avoid publishing Courses to the LMS during migration

rlucioni:
Noticed this while working on the Course publication endpoint.

@clintonb or @jimabramson, could you please review this when you're able?


clintonb:
:+1:

PR #221, opened 2015-07-23
Include nodeenv in base requirements

rlucioni:
Facilitates inclusion of ecommerce in devstack.

@clintonb, any objections?


clintonb:
Strange. Why aren't you using production settings for devstack/sandboxes?

rlucioni:
The ecommerce Ansible role [installs dev requirements](https://github.com/edx/configuration/blob/master/playbooks/roles/ecommerce/tasks/main.yml#L46) - local.txt - when bringing up a devstack.

clintonb:
:+1:

PR #223, opened 2015-07-23
WIP: Braintree Integration

clintonb:
This integration is functional, but needs additional work before being production-ready:
- Unit/acceptance tests
- Error handling
- Processor response logging
### How it works

Configuration is the same as with our existing processors, albeit a bit simpler. The bigger changes happen on the front end.
1. Instead of POSTing to `/api/v2/baskets/` and redirecting, LMS POSTS, and uses the data to initialize a payment form (displayed on the track selection page). 
2. The user submits this form and the data is POSTed to Otto `/payments/braintree/checkout/` where the sale is completed and logged.
3. After processing, Otto redirects the user to the receipt page.

FYI @rlucioni @jimabramson @wedaly @Nickersoft @griffresch @pbaruah @rock345

### DIFF
--- a/ecommerce/extensions/api/authentication.py
+++ b/ecommerce/extensions/api/authentication.py
@@ -1,4 +1,5 @@
 """JWT authentication scheme for use with DRF."""
+import logging
 from django.conf import settings
 from django.contrib.auth import get_user_model
 import requests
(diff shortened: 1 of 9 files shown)


rlucioni:
:cry:

PR #224, opened 2015-07-24
Updated Course Migration and Publishing

clintonb:
- XCOM-514: Course publishing has been decoupled from course saving (to DB).
- XCOM-515: The migration management command once again publishes data (esp. SKU) back to LMS
- When publishing, professional seats NEVER have an expiration date

@rlucioni @jimabramson

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -5,9 +5,7 @@ from django.db import models, transaction
 from django.utils.text import slugify
 from oscar.core.loading import get_model
 from simple_history.models import HistoricalRecords
-import waffle
 
-from ecommerce.courses.exceptions import PublishFailed
 from ecommerce.courses.publishers import LMSPublisher
 from ecommerce.extensions.catalogue.utils import generate_sku
 
@@ -48,19 +46,10 @@ class Course(models.Model):
         else:
             logger.debug('Parent seat [%d] already exists for [%s].', parent.id, self.id)
 
-    # pylint: disable=arguments-differ
     @transaction.atomic
-    def save(self, force_insert=False, force_update=False, using=None, update_fields=None, publish=True):
+    def save(self, force_insert=False, force_update=False, using=None, update_fields=None):
         super(Course, self).save(force_insert, force_update, using, update_fields)
-
         self._create_parent_seat()
-
-        if publish and waffle.switch_is_active('publish_course_modes_to_lms'):
-            if not self.publish_to_lms():
-                # Raise an exception to force a rollback
-                raise PublishFailed('Failed to publish {}'.format(self.id))
-        else:
-            logger.debug('Course mode publishing is not enabled. Commerce changes will not be published!')
 
     def publish_to_lms(self):
         """ Publish Course and Products to LMS. """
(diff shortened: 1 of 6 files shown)


ecommerce/courses/models.py
```diff
@@ -125,13 +114,22 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
...
         slug = 'child-cs-{}-{}'.format(certificate_type, slugify(course_id))
 
+        # Note (CCB): Our previous method of slug generation did not account for ID verification. By using a list
+        # we can update these seats. This should be removed after the courses have been re-migrated.
+        if certificate_type == 'verified':
+            slugs.append(slug)
+
         if id_verification_required:
             slug += '-id-verified'
+        slugs.append(slug)
```
jimabramson, line 128: nitpick: could you append the suffix inline rather than overwriting the variable?  I feel like this may lead to subtle bugs down the road.

jimabramson:
one nitpick. LGTM otherwise

ecommerce/courses/publishers.py
```diff
@@ -14,6 +14,12 @@ def mode_for_seat(self, seat):
             return 'no-id-professional'
         return seat.attr.certificate_type
 
+    def get_seat_expiration(self, seat):
```
rlucioni, line 17: Should this validation occur sooner, at the time of migration? As is, we're allowing bad data to get into our system, but then not letting it leave. It seems preferable to prevent bad data from entering our system in the first place, then correct the LMS' data at publication time.

ecommerce/courses/tests/test_publishers.py
```diff
@@ -115,3 +115,25 @@ def test_serialize_seat_for_commerce_api(self):
...
         self.assertDictEqual(actual, expected)
+
+    @ddt.unpack
+    @ddt.data(
+        (True, 'professional'),
+        (False, 'no-id-professional'),
+    )
+    def test_serialize_seat_for_commerce_api_with_professional(self, is_verified, expected_mode):
+        """
+        Verify that (a) professional seats NEVER have an expiration date and (b) the name/mode is properly set for
```
rlucioni, line 126: Could you please pass an expiration date to `create_or_update_seat()` to exercise the validation logic being introduced?

ecommerce/courses/models.py
clintonb, line 128: I don't quite follow you. We need a list of slugs for verified seats, not a single one.

ecommerce/courses/publishers.py
clintonb, line 17: I see your point. We know for a fact that we want to avoid publishing expiration dates for professional seats to LMS. Going forward we have not thought out if that same logic applies to Otto, and I'd rather not deal with that at the moment.

ecommerce/courses/tests/test_publishers.py
clintonb, line 126: Good catch. Updated.

rlucioni:
:+1:

ecommerce/courses/models.py
clintonb, line 128: @jimabramson You have DND setup on HipChat. Going to merge; but, if you have major objections, let me know. Note that this is a minor hack that will be removed once the existing 12 migrated courses are re-migrated.

jimabramson, line 128: sorry, just caught this.  my suggestion was trivial, and as a minor hack, not worth going back to fuss over.

`slugs.append(slug + '-id-verified')`

PR #227, opened 2015-07-25
Devstack-specific settings module

rlucioni:
Settings module to be used when running the ecommerce service within devstack.

@feanil @clintonb @jimabramson


clintonb:
:+1:

jimabramson:
:+1:

PR #229, opened 2015-07-27
Added last edited date to courses API

Nickersoft:
@clintonb @jimabramson FYI: @rlucioni @wedaly

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -114,6 +114,10 @@ class RefundSerializer(serializers.ModelSerializer):
 
 class CourseSerializer(serializers.HyperlinkedModelSerializer):
     products_url = serializers.SerializerMethodField()
+    last_edited = serializers.SerializerMethodField()
+
+    def get_last_edited(self, obj):
+        return obj.history.latest().history_date.strftime(ISO_8601_FORMAT)
 
     def get_products_url(self, obj):
         return reverse('api:v2:course-product-list', kwargs={'parent_lookup_course_id': obj.id},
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/api/serializers.py
```diff
@@ -121,8 +125,8 @@ def get_products_url(self, obj):
 
     class Meta(object):
         model = Course
-        fields = ('id', 'url', 'name', 'type', 'products_url',)
-        read_only_fields = ('type',)
+        fields = ('id', 'url', 'name', 'type', 'products_url', 'last_edited')
+        read_only_fields = ('type')
```
clintonb: Revert this line.

jimabramson:
tests?  also, does `obj.history.latest()` ever not have a value?

ecommerce/extensions/api/serializers.py
Nickersoft: @clintonb Should `last_edited` be added to read_only_fields instead? If it is omitted, an error occurs

Nickersoft:
@jimabramson The only time there shouldn't be a value for `obj.history.latest()` is when the course is first created. Then I believe it returns an empty string. Tests were overlooked... I'll append a test for `last_edited` to the current courses API tests.

ecommerce/extensions/api/serializers.py
clintonb: Revert the deletion of the comma. The `fields` declaration is fine as you have it.

Nickersoft: @clintonb Oh, okay. I didn't notice that. Will revert.

clintonb:
@jimabramson @Nickersoft There will always be a last-edited date.

Nickersoft:
@clintonb Created date is included in `last_edited`? I thought there was a previous occasion were the course had to be updated in order for `last_edited` to have a value. It's good to know, nevertheless.

Nickersoft:
@clintonb @jimabramson Added tests

ecommerce/extensions/api/serializers.py
Nickersoft: Fixed

clintonb:
:+1: (once tests pass)

Nickersoft:
@jimabramson ?

rlucioni:
Looks good, :+1: .

Nickersoft:
@jimabramson Closing due to two :+1:

PR #230, opened 2015-07-27
Course list view now populates using API over AJAX

Nickersoft:
@clintonb @jimabramson @rlucioni @AlasdairSwan


bower.json
```diff
@@ -20,5 +20,8 @@
     "datatables": "~1.10.7",
     "fontawesome": "~4.3.0",
     "edx-ux-pattern-library": "https://github.com/edx/ux-pattern-library.git#82fa1c823bc322ba8b0742f0c546a89b0c69e952"
+  },
+  "devDependencies": {
+    "requirejs-text": "~2.0.14"
```
clintonb: Use 'text' and save it to the production dependencies. It will be used by the course detail view.

ecommerce/static/js/models/course_list_model.js
```diff
@@ -0,0 +1,35 @@
+define([
+  'underscore',
+  'backbone',
+], function (_, Backbone) {
+
+  return Backbone.Model.extend({
+
+    initialize: function (options) {
```
clintonb: This method is unnecessary. Just declare url as you would any other attribute or function: `url: 'api/v2/courses/'`.

ecommerce/static/js/models/course_list_model.js
```diff
@@ -0,0 +1,35 @@
+define([
+  'underscore',
+  'backbone',
+], function (_, Backbone) {
+
+  return Backbone.Model.extend({
```
clintonb: You should use a `Collection` rather than making a `Model` specific to this view.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,29 +13,61 @@ define([
 
             el: '#course-list-view',
 
-            initialize: function () {
+            template: _.template(tpl),
+
+            initialize: function (options) {
+
```
clintonb, line 26: Why so much whitespace?

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,29 +13,61 @@ define([
...
-                    info: false,
-                    paging: false,
-                    oLanguage: {
-                        sSearch: ''
-                    }
-                });
+                this.$el.html(this.template);
+
+                this.model.fetch({
+                    success: function () {
```
clintonb: Move this to its own function (e.g. `renderCourseTable`).

ecommerce/static/js/models/course_list_model.js
Nickersoft: @clintonb Could you please elaborate? I just want to make sure I understand correctly.

ecommerce/static/js/views/course_list_view.js
Nickersoft, line 26: @clintonb Force of habit.. I was just trying to ensure readability. I can cut down on the whitespace

ecommerce/static/js/models/course_list_model.js
clintonb: http://liquidmedia.org/blog/2011/01/backbone-js-part-1/

clintonb: Note that you might need to bring in `drf_pageable_collection` from #214 to deal with pagination.

Nickersoft: @clintonb I probably should have specified... I know what a Collection is, I was wondering how the collection would be used throughout the Backbone application. Would each model in the collection correspond to a row in the table?

clintonb: The collection has one model for each course so, yes.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,29 +13,61 @@ define([
...
+                        var results = self.model.get('results'),
+                            tableData = [],
+                            filterPlaceholder = gettext('Filter by org or course ID'),
+                            $emptyLabel = '<label class="sr">' + filterPlaceholder + '</label>';
 
-                this.$el.find('#courseTable_filter input')
-                    .attr('placeholder', filterPlaceholder)
-                    .addClass('field-input input-text')
-                    .removeClass('form-control input-sm');
+                        _.each(results, function (value) {
```
clintonb: You should be able to call `toJSON()` on your collection and use field/column mapping.

ecommerce/static/js/models/course_list_model.js
```diff
@@ -0,0 +1,35 @@
...
+        $.ajax({
+            type: 'GET',
+            url: this.url,
+            contentType: 'application/json; charset=utf-8',
+            crossDomain: true,
+            context: this,
+            success: function( data ) {
+                this.set( data );
+                this.trigger( 'sync' );
+                if (options.success) options.success();
```
AlasdairSwan: if statements should not be in a single line and should use brackets i.e.

```
if ( options.success ) {
    options.success();
}
```

ecommerce/static/js/models/course_list_model.js
```diff
@@ -0,0 +1,35 @@
...
+            },
+            error: function( jqXHR ) {
+                this.trigger( 'error', jqXHR );
+                if (options.error) options.error();
```
AlasdairSwan: if statements should not be in a single line and should use brackets

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,29 +13,61 @@ define([
...
-                this.$el.find('#courseTable').DataTable({
-                    info: false,
-                    paging: false,
-                    oLanguage: {
-                        sSearch: ''
-                    }
-                });
+                this.$el.html(this.template);
+
+                this.model.fetch({
```
AlasdairSwan: The model should be fetched before the render function is called. Render should just be taking the model and rendering the DOM with it.

clintonb:
@Nickersoft rebase on master to get the existing course collection.

Nickersoft:
@clintonb Do you mean Backbone collection?

clintonb:
@Nickersoft yes. There is no other type of collection.

Nickersoft:
@clintonb I didn't know if you meant collection in the general sense of the term.. anyway, I rebased and don't see a preexisting collection. Which directory is it under?

Nickersoft:
@clintonb My mistake, I hadn't rebased yet. I see it now.

clintonb:
@Nickersoft I stand corrected. I only created a product collection. You'll need to create a course collection based on it. The code should be mostly the same, except `model` can be hardcoded to `CourseModel`.

Nickersoft:
@clintonb Okay

ecommerce/static/js/models/course_list_model.js
Nickersoft: Fixed

ecommerce/static/js/models/course_list_model.js
Nickersoft: Fixed

ecommerce/static/js/views/course_list_view.js
Nickersoft: Fixed

Nickersoft:
@AlasdairSwan @clintonb Updated

ecommerce/static/js/collections/course_collection.js
```diff
@@ -0,0 +1,51 @@
...
+
+        $.ajax({
+          type: 'GET',
+          url: this.url,
+          contentType: 'application/json; charset=utf-8',
+          crossDomain: true,
+          context: this,
+          success: function( data ) {
+
+              results = data.results;
```
AlasdairSwan: instead of line 15 results should be scoped here because this function starts at line 25. It would be a little cleaner to move this code to its own function.

ecommerce/static/js/pages/course_list_page.js
```diff
@@ -1,9 +1,19 @@
...
+        'collections/course_collection',
         'views/course_list_view'
     ],
-    function (CourseListView) {
+    function (CourseCollection, CourseListView) {
 
-        return new CourseListView();
+        var courseCollection = new CourseCollection(),
+            courseListView = new CourseListView({
+                models: courseCollection
```
AlasdairSwan: it would be better to pass in the collection attribute

```
collections: courseCollection
```

Nickersoft: @AlasdairSwan Was not aware of this attribute. The more you know I guess! I'll change it.

ecommerce/static/js/collections/course_collection.js
Nickersoft: Fixed

Nickersoft:
@AlasdairSwan @clintonb Updated

ecommerce/static/js/collections/course_collection.js
```diff
@@ -0,0 +1,51 @@
+define([
+  'underscore',
+  'backbone',
+  'models/course_model'
+], function (_, Backbone, CourseModel) {
+
+  return Backbone.Collection.extend({
+
```
clintonb, line 6: Same note about whitespace.

ecommerce/static/js/collections/course_collection.js
```diff
@@ -0,0 +1,51 @@
...
+        _.each( data, function ( attributes ) {
+            courseModel = new CourseModel( attributes );
+            courseModels.push( courseModel );
+        });
+
+        this.reset( courseModels );
+    },
+
+    fetch: function (options) {
```
clintonb: Why is this necessary? `DrfPageableCollection` handles all of this for you.

ecommerce/static/js/collections/course_collection.js
```diff
@@ -0,0 +1,51 @@
...
+  'models/course_model'
+], function (_, Backbone, CourseModel) {
+
+  return Backbone.Collection.extend({
+
+    model: CourseModel,
+
+    url: '/api/v2/courses/',
+
+    updateModels: function (data) {
```
clintonb: Why is this necessary?

ecommerce/static/js/config.js
```diff
@@ -13,7 +13,7 @@ require.config({
         'models': 'js/models',
         'moment': 'bower_components/moment/moment',
         'requirejs': 'bower_components/requirejs/require',
-        'templates': 'templates',
+        'templates': 'js/templates',
```
clintonb: Revert.

ecommerce/static/js/models/course-model.js
```diff
@@ -1,10 +0,0 @@
-define(['backbone'], function(Backbone) {
```
clintonb, line 1: You've removed this without updating the consuming code.

ecommerce/static/js/pages/course_list_page.js
```diff
@@ -1,9 +1,19 @@
...
-    function (CourseListView) {
+    function (CourseCollection, CourseListView) {
 
-        return new CourseListView();
+        var courseCollection = new CourseCollection(),
+            courseListView = new CourseListView({
+                collection: courseCollection
+            });
+
+        courseCollection.fetch({
```
clintonb: This should be part of the `render` method in `CourseListView`. The page should not care about fetching data.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,22 +13,46 @@ define([
...
+                        paging: false,
+                        oLanguage: {
+                            sSearch: ''
+                        },
+                        columns: [
+                            {
+                                title: gettext('ID'),
+                                data: 'id',
+                                fnCreatedCell: function (nTd, sData, oData, iRow, iCol) {
+                                    $(nTd).html("<a href='/courses/"+oData.id+"'>"+oData.id+"</a>");
```
clintonb: Clean this up with [`_s.sprintf()`](http://gabceb.github.io/underscore.string.site/#sprintf).

ecommerce/static/js/views/course_list_view.js
```diff
@@ -36,6 +61,26 @@ define([
...
 
+            },
+
+            render: function () {
+
+                var tableData = [];
+
+                this.$el.html(this.template);
+
+                this.collection.each(function (value) {
```
clintonb: Replace this with `Collection.fetch()` which in turn calls the `renderCourseTable` method.

ecommerce/static/js/models/course-model.js
Nickersoft, line 1: @clintonb I removed all references to this file. There were two course models, and I left the more robust one. Is there an additional step I should have taken?

ecommerce/static/js/config.js
Nickersoft: @clintonb This would then point to /static/templates, which are all Django templates. Underscore templates are in `js/templates`. Why would this be necessary?

ecommerce/static/js/pages/course_list_page.js
Nickersoft: @clintonb This issue is addressed [here](https://github.com/edx/ecommerce/pull/230#discussion_r35570304)

ecommerce/static/js/views/course_list_view.js
Nickersoft: @clintonb https://github.com/edx/ecommerce/pull/230#discussion_r35570304

ecommerce/static/js/collections/course_collection.js
Nickersoft, line 6: Fixed

ecommerce/static/js/config.js
clintonb: Check again. `static/templates` contains underscore templates.

ecommerce/static/js/models/course-model.js
clintonb, line 1: Search again for usages of "course-model".

ecommerce/static/js/pages/course_list_page.js
clintonb: @AlasdairSwan please chime in. It seems weird that the page is concerned with the collection consumed by the view.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -2,9 +2,10 @@ define([
         'jquery',
         'underscore',
         'backbone',
+        'text!templates/course_list.underscore',
         'dataTablesBootstrap'
       ],
-      function ($, _, Backbone) {
+      function ($, _, Backbone, tpl) {
```
jimabramson: I think a specific name would be better than tpl

ecommerce/static/js/collections/course_collection.js
Nickersoft: @clintonb Does DrfPageableCollection pass in the `results` attribute of the JSON response automatically? That is the primary reason this function exists, to ensure the correct data is being assigned to each course model.

ecommerce/static/js/collections/course_collection.js
Nickersoft: @clintonb To clean up the `fetch()` function under the recommendation of @AlasdairSwan.

ecommerce/static/js/views/course_list_view.js
Nickersoft: Fixed

ecommerce/static/js/collections/course_collection.js
clintonb: Yes. As I mentioned in a previous comment, all you need to do is copy the behavior of the `ProductCollection` and make a few modifications.

ecommerce/static/js/models/course-model.js
Nickersoft, line 1: @clintonb I stand corrected. I see that now. Didn't check for usage in credit_checkout. Will update.

ecommerce/static/js/config.js
Nickersoft: @clintonb Not in my copy of the repo image

Nickersoft: Or in https://github.com/edx/ecommerce/tree/master/ecommerce/static/templates

ecommerce/static/js/views/course_list_view.js
Nickersoft: Fixed

ecommerce/static/js/config.js
clintonb: Again, those are both underscore templates. This is getting frustrating. I _created_ the directory! I know what's in there. Take a moment to actually look at the contents of the files before making assumptions.

Nickersoft: @clintonb I'm sorry for the confusion. I think I was getting `static/templates` and `ecommerce/templates` mixed up. By convention, underscore templates generally have `.underscore` as a file extension, such as on the [marketing site](https://github.com/edx/edx-mktg/tree/master/docroot/sites/all/themes/atedx/scripts/templates), which is why I was confused. The template file for the course list view is currently in `js/templates`, which I had created after I didn't see any underscore files in the static directory. I'll move it to the new templates folder. Should I rename your templates to have a `.underscore` file extension to prevent further confusion?

clintonb: No. Keep the scope of your PR small. Change your templates to match. We can change them all later if we decide to do so.

Nickersoft:
@clintonb @AlasdairSwan @rlucioni Updated

ecommerce/static/js/pages/course_list_page.js
AlasdairSwan: I agree with @clintonb the collection should be instantiated from within the courseListView

Nickersoft: @clintonb This issue has been addressed. Because the collection is not reused, the fetch logic was moved back into the view. See https://github.com/edx/ecommerce/pull/230/files#diff-86eaaaf53981e4a0820194ebe63e6dc1R72.

ecommerce/static/js/collections/course_collection.js
```diff
@@ -0,0 +1,25 @@
...
+  'underscore',
+  'collections/drf_pageable_collection',
+  'models/course_model'
+], function (_, DrfPageableCollection, CourseModel) {
+
+  return DrfPageableCollection.extend({
+    model: CourseModel,
+    url: '/api/v2/courses/',
+
+    updateModels: function (data) {
```
clintonb: Why is this necessary?

ecommerce/static/js/pages/course_list_page.js
```diff
@@ -1,9 +1,14 @@
 require([
+        'collections/course_collection',
         'views/course_list_view'
     ],
-    function (CourseListView) {
+    function (CourseCollection, CourseListView) {
 
-        return new CourseListView();
+        var courseCollection = new CourseCollection();
```
clintonb: The variable isn't necessary since the value isn't shared.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -1,10 +1,12 @@
 define([
         'jquery',
         'underscore',
+        'underscore.string',
         'backbone',
+        'text!templates/course_list.html',
         'dataTablesBootstrap'
       ],
-      function ($, _, Backbone) {
+      function ($, _, s, Backbone, courseListViewTemplate) {
```
clintonb: _s

ecommerce/static/js/views/course_list_view.js
```diff
@@ -36,6 +63,33 @@ define([
                     .addClass('field-input input-text')
                     .removeClass('form-control input-sm');
 
+            },
+
+            render: function () {
+
+                var self = this;
+
+                this.collection.fetch({
```
clintonb: It would be ideal to keep the focus of your methods specific. `render` shouldn't care about fetching. It should be focused solely on rendering. Call `this.collection.fetch()` in `initialize()`.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -36,6 +63,33 @@ define([
...
+                    success: function () {
+
+                        var tableData = [];
+
+                        self.$el.html(self.template);
+
+                        self.collection.each(function (value) {
```
clintonb: Move this into `renderCourseTable`. `render` shouldn't care about the data for the table it is not directly rendering.

Nickersoft:
@clintonb @AlasdairSwan @rlucioni Updated

ecommerce/static/js/views/course_list_view.js
Nickersoft: Fixed

ecommerce/static/js/views/course_list_view.js
Nickersoft: Fixed

ecommerce/static/js/views/course_list_view.js
Nickersoft: Fixed

ecommerce/static/js/pages/course_list_page.js
Nickersoft: Fixed

ecommerce/static/js/collections/course_collection.js
Nickersoft: Fixed

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,29 +14,78 @@ define([
...
+                this.listenTo(this.collection, 'add remove change', this.render);
                 this.render();
             },
 
-            render: function () {
+            renderCourseTable: function () {
+
+                var self = this;
+
+                this.collection.fetch({
```
clintonb: Why are you fetching here?

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,29 +14,78 @@ define([
 
             el: '#course-list-view',
 
-            initialize: function () {
+            template: _.template(courseListViewTemplate),
+
+            initialize: function (options) {
+                this.listenTo(this.collection, 'add remove change', this.render);
```
clintonb, line 21: Call `this.collection.fetch()` and remove the call to `render()` below.

ecommerce/static/js/pages/course_list_page.js
```diff
@@ -1,9 +1,12 @@
 require([
+        'collections/course_collection',
         'views/course_list_view'
     ],
-    function (CourseListView) {
+    function (CourseCollection, CourseListView) {
 
-        return new CourseListView();
+        return courseListView = new CourseListView({
```
jimabramson: i don't think the local variable assignment is necessary.

ecommerce/static/js/pages/credit_checkout.js
```diff
@@ -3,7 +3,7 @@ require([
         'backbone',
         'js/models/user-model',
         'js/models/tracking-model',
-        'js/models/course-model',
+        'js/models/course_model',
```
jimabramson, line 6: maybe a separate PR, but we should be consistent in our use of `_` vs `-` in these module names.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,29 +14,78 @@ define([
...
+                                ]
+                            });
+                        }
+
+                        self.$el.find('#courseTable_filter label').prepend($emptyLabel);
+
+                        self.$el.find('#courseTable_filter input')
+                            .attr('placeholder', filterPlaceholder)
+                            .addClass('field-input input-text')
+                            .removeClass('form-control input-sm');
```
jimabramson: minor - i assume you're already working on this with @clintonb, and i don't have much experience working with this data table library, but it seems slightly weird to re-render the filter inputs/label each time the collection is refreshed.

ecommerce/static/js/views/course_list_view.js
Nickersoft: @clintonb I misunderstood your previous comment regarding table data. I'll fix it.

ecommerce/static/js/pages/course_list_page.js
Nickersoft: @jimabramson That was leftover from back when it was a local variable. Will remove.

ecommerce/static/js/pages/credit_checkout.js
Nickersoft, line 6: @jimabramson Agreed. Right now there seems to be inconsistencies in the naming conventions of files.

ecommerce/static/js/views/course_list_view.js
Nickersoft: @jimabramson I agree. This was overlooked... I'll move it into the above if block, which only get calls once when the table is initialized for the first time.

Nickersoft: Fixed

ecommerce/static/js/pages/course_list_page.js
Nickersoft: Fixed

ecommerce/static/js/views/course_list_view.js
Nickersoft, line 21: Fixed

ecommerce/static/js/views/course_list_view.js
Nickersoft: Fixed

Nickersoft:
@clintonb @AlasdairSwan @jimabramson @rlucioni Updated

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,29 +14,72 @@ define([
...
-                    paging: false,
-                    oLanguage: {
-                        sSearch: ''
-                    }
+                this.collection.each(function (value) {
+                    tableData.push(
+                        {
+                            id: value.get('id'),
+                            name: value.get('name'),
+                            last_edited: new Date(Date.parse(value.get('last_edited'))).toLocaleDateString()
```
clintonb: This should be date and time (`July 15, 2015, 3:09 p.m.`). Right now it is only showing the date (`7/15/2015`).

clintonb: Consider using [Moment.js](http://momentjs.com/), if necessary to handle the parsing and formatting. The library is already installed.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -12,29 +14,72 @@ define([
...
+                        }
+                    );
                 });
 
-                this.$el.find('#courseTable_filter label').prepend($emptyLabel);
+                if (!$.fn.dataTable.isDataTable('#courseTable')) {
+
+                    this.$el.find('#courseTable').DataTable({
+                        data: tableData,
```
clintonb, line 45: Add `autoWidth: false,` to remove the width attributes set by Datatables. These widths make the table look weird and are not updated when the window is resized.

clintonb:
:+1: (once cosmetic issues addressed)

ecommerce/static/js/views/course_list_view.js
Nickersoft: Fixed

ecommerce/static/js/views/course_list_view.js
Nickersoft, line 45: Fixed

Nickersoft:
@AlasdairSwan @jimabramson @rlucioni Updated

AlasdairSwan:
Looks good :+1:

jimabramson:
@Nickersoft my only request, if not addressing the `_` vs `-` inconsistency in this PR, is to create a separate ticket to address that.  Otherwise :+1:

Nickersoft:
@jimabramson Will do

PR #231, opened 2015-07-27
Atomic Course publication

rlucioni:
Facilitates atomic saving and publication of Courses and Products associated with them. XCOM-506.

@clintonb and @jimabramson, tests are in flight, but the rest of the implementation is solidified enough that it could benefit from extra eyes, if you have the time.

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -1,11 +1,20 @@
-"""Serializers for order and line item data."""
+"""Serializers for data manipulated by ecommerce API endpoints."""
+from decimal import Decimal
+import logging
 
+from dateutil.parser import parse
+from django.db import transaction
+from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_model, get_class
 from rest_framework import serializers
 from rest_framework.reverse import reverse
+import waffle
 
-from ecommerce.core.constants import ISO_8601_FORMAT
+from ecommerce.core.constants import ISO_8601_FORMAT, COURSE_ID_REGEX
 from ecommerce.courses.models import Course
+
+
+logger = logging.getLogger(__name__)
 
 BillingAddress = get_model('order', 'BillingAddress')
 Line = get_model('order', 'Line')
@@ -113,6 +122,7 @@ class RefundSerializer(serializers.ModelSerializer):
 
 
 class CourseSerializer(serializers.HyperlinkedModelSerializer):
+    id = serializers.RegexField(COURSE_ID_REGEX, max_length=255)
     products_url = serializers.SerializerMethodField()
     last_edited = serializers.SerializerMethodField()
 
@@ -130,3 +140,104 @@ class CourseSerializer(serializers.HyperlinkedModelSerializer):
         extra_kwargs = {
             'url': {'view_name': COURSE_DETAIL_VIEW}
         }
+
+
+class AtomicPublicationSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    """Serializer for saving and publishing a Course and associated products.
+
+    Using a ModelSerializer for the Course data makes it difficult to use this serializer to handle updates.
+    The automatically applied validation logic rejects course IDs which already exist in the database.
+    """
+    id = serializers.RegexField(COURSE_ID_REGEX, max_length=255)
+    name = serializers.CharField(max_length=255)
+    products = serializers.ListField()
+
+    def validate_products(self, products):
+        """Validate product data."""
+        for product in products:
+            # Verify that each product is intended to be a Seat.
+            product_class = product.get('product_class')
+            if product_class != 'Seat':
+                raise serializers.ValidationError(
+                    _(u"Invalid product class [{product_class}] requested.".format(product_class=product_class))
+                )
+
+            # Verify that attributes required to create a Seat are present.
+            attrs = self._flatten(product['attribute_values'])
+            if attrs.get('certificate_type') is None:
+                raise serializers.ValidationError(_(u"Products must have a certificate type."))
+            elif attrs.get('id_verification_required') is None:
+                raise serializers.ValidationError(_(u"Products must indicate whether ID verification is required."))
+
+            # Verify that a price is present.
+            if product.get('price') is None:
+                raise serializers.ValidationError(_(u"Products must have a price."))
+
+        return products
+
+    def save(self):
+        """Save and publish Course and associated products."
+
+        Returns:
+            tuple: A Boolean indicating whether the Course was created, an Exception,
+                if one was raised (else None), and a message for the user, if necessary (else None).
+        """
+        course_id = self.validated_data['id']
+        course_name = self.validated_data['name']
+        products = self.validated_data['products']
+
+        try:
+            # Explicitly delimit operations which will be rolled back if an exception is raised.
+            with transaction.atomic():
+                course, created = Course.objects.get_or_create(id=course_id, defaults={'name': course_name})
+
+                for product in products:
+                    attrs = self._flatten(product['attribute_values'])
+
+                    # Extract arguments required for Seat creation, deserializing as necessary.
+                    certificate_type = attrs['certificate_type']
+                    id_verification_required = (attrs['id_verification_required'] == 'True')
+                    price = Decimal(product['price'])
+
+                    # Extract arguments which are optional for Seat creation, deserializing as necessary.
+                    expires = product.get('expires')
+                    expires = parse(expires) if expires else None
+                    credit_provider = attrs.get('credit_provider')
+                    credit_hours = attrs.get('credit_hours')
+                    credit_hours = int(credit_hours) if credit_hours else None
+
+                    course.create_or_update_seat(
+                        certificate_type,
+                        id_verification_required,
+                        price,
+                        expires=expires,
+                        credit_provider=credit_provider,
+                        credit_hours=credit_hours,
+                    )
+
+                if waffle.switch_is_active('publish_course_modes_to_lms'):
+                    published = course.publish_to_lms()
+                    if published:
+                        return created, None, None
+                    else:
+                        message = (
+                            u'An error occurred while publishing [{course_id}] to LMS. '
+                            u'No data has been saved or published.'
+                        ).format(course_id=course_id)
+                        raise Exception(message)
+                else:
+                    message = (
+                        u'Course [{course_id}] was not published to LMS '
+                        u'because the switch [publish_course_modes_to_lms] is disabled. '
+                        u'Data has been saved, but not published.'
+                    ).format(course_id=course_id)
+                    logger.info(message)
+
+                    return created, None, message
+        except Exception as e:  # pylint: disable=broad-except
+            logger.exception(u'Failed to save and publish [%s]: [%s]', course_id, e.message)
+            return False, e, e.message
+
+    def _flatten(self, attrs):
+        """Transform a list of attribute names and values into a dictionary keyed on the names."""
+        return {attr['name']: attr['value'] for attr in attrs}
--- a/ecommerce/extensions/api/v2/tests/views/test_publication.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_publication.py
@@ -1,0 +1,228 @@
+from datetime import datetime
+from copy import deepcopy
+from decimal import Decimal
+import json
+
+from django.core.urlresolvers import reverse
+from django.test import TestCase
+import mock
+import pytz
+from waffle import Switch
+
+from ecommerce.core.constants import ISO_8601_FORMAT
+from ecommerce.courses.models import Course
+from ecommerce.courses.publishers import LMSPublisher
+from ecommerce.extensions.api.v2.tests.views import JSON_CONTENT_TYPE
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.tests.mixins import UserMixin
+
+
+EXPIRES = datetime(year=1992, month=4, day=24, tzinfo=pytz.utc)
+EXPIRES_STRING = EXPIRES.strftime(ISO_8601_FORMAT)
+
+
+class AtomicPublicationTests(CourseCatalogTestMixin, UserMixin, TestCase):
+    def setUp(self):
+        super(AtomicPublicationTests, self).setUp()
+
+        self.course_id = 'BadgerX/B101/2015'
+        self.course_name = 'Dances with Badgers'
+        self.create_path = reverse('api:v2:publication:create')
+        self.update_path = reverse('api:v2:publication:update', kwargs={'course_id': self.course_id})
+        self.data = {
+            'id': self.course_id,
+            'name': self.course_name,
+            'products': [
+                {
+                    'product_class': 'Seat',
+                    'expires': None,
+                    'price': 0.00,
+                    'attribute_values': [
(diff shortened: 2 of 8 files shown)


ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,70 @@ class Meta(object):
         extra_kwargs = {
             'url': {'view_name': COURSE_DETAIL_VIEW}
         }
+
+
+class AtomicPublicationSerializer(serializers.Serializer):  # pylint: disable=abstract-method
```
clintonb, line 145: Why not use a `ModelSerializer` for the Course?

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,70 @@ class Meta(object):
...
+    """Serializer for saving and publishing a Course and its Products."""
+    id = serializers.CharField()
+    name = serializers.CharField(required=False)
+    products = serializers.ListField()
+
+    def validate_products(self, products):
```
clintonb, line 155: This isn't really necessary given that we have no immediate plans to deal with non-seat products.

ecommerce/extensions/api/serializers.py
```diff
@@ -1,12 +1,21 @@
 """Serializers for order and line item data."""
```
jimabramson: minor: update docstring

ecommerce/extensions/api/serializers.py
jimabramson, line 155: We've been pretty consistent in avoiding the assumption in Otto that all products are course seats; I'm in favor of keeping this.

ecommerce/extensions/api/serializers.py
rlucioni, line 145: No reason, I just blanked on it. Updated; thanks for the tip.

ecommerce/extensions/api/serializers.py

ecommerce/extensions/api/serializers.py
rlucioni, line 155: @clintonb I agree that it's not strictly necessary, but it's a relatively cheap way to validate our assumptions about the data we expect this endpoint to receive. Unless you strongly object, I'm in favor of keeping this around, too.

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,70 @@ class Meta(object):
...
+                            u'No data has been saved or published.'
+                        )
+                        raise Exception(msg.format(course_id=course_id))
+                else:
+                    msg = (
+                        u'Course [{course_id}] was not published to LMS '
+                        u'because the switch [publish_course_modes_to_lms] is disabled. '
+                        u'No data has been saved or published.'
+                    )
+                    raise Exception(msg.format(course_id=course_id))
```
jimabramson: i feel like i've seen the preceding if/else block in another part of the code, is there a reason to duplicate that here instead of reusing?

rlucioni: It is very similar to the conditional found in the `CourseViewSet`'s `publish` action. We could reuse the conditional by moving the switch checking into the `Course` model; I'll look into it.

ecommerce/extensions/api/v2/tests/views/test_publication.py
```diff
@@ -0,0 +1,125 @@
...
+from ecommerce.courses.publishers import LMSPublisher
+from ecommerce.extensions.api.v2.tests.views import JSON_CONTENT_TYPE
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.tests.mixins import UserMixin
+
+
+Product = get_model('catalogue', 'Product')
+
+
+class AtomicPublicationTests(CourseCatalogTestMixin, UserMixin, TestCase):
```
jimabramson, line 24: please add coverage to the test for behavior when one of the expected product fields / attr+value pairs is missing

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,72 @@ class Meta(object):
...
+                if course_name is not None:
+                    course.name = course_name
+                    course.save()
+
+                for product in products:
+                    # Transform the list of attribute names and values into a dictionary keyed on names.
+                    kwargs = {attr['name']: attr['value'] for attr in product['attribute_values']}
+
+                    # Strip out the 'course_key' attribute; it's not necessary for Seat creation.
+                    kwargs.pop('course_key', None)
```
jimabramson: so, I'm thinking about what happens if a client mistakenly POSTs instead of PUTting for an existing course.  The result of that operation would be a duplicated course and products, but we can defend against that by ensuring the course_id (if present) matches the course key here, and 400-ing if not.   Let me know if you think I'm being overly paranoid about this.

jimabramson:
@rlucioni aside from comments, LGTM so far

ecommerce/extensions/api/serializers.py
rlucioni: @jimabramson it's not clear to me how this could result in duplicated courses and products. The save method gets-or-creates a Course and gets-or-creates corresponding Seats. If one were to POST information for an existing course, the effect should be the same as PUTing for an existing course.

jimabramson: @rlucioni sorry, let me clarify: I'm not seeing a problem with the
server implementation but an opportunity to recognize and defend
against the result of a client mistake.  If the client posts course ids in
their product list that don't match the parent course id, we can probably
deduce that the client is making a mistake of some sort and refuse to
process the data.  If the course id matches or is empty then we could just
ignore.

rlucioni: @jimabramson got it. I agree that this is a good way to protect against the unexpected; I'll look into incorporating this kind of validation.

ecommerce/extensions/api/serializers.py
rlucioni, line 145: @clintonb I've reverted to using a plain `Serializer`. Using a `ModelSerializer` for the Course data makes it difficult to use this serializer to handle updates. The automatically applied validation logic rejects course IDs which already exist in the database.

I've seen Tom Christie (maintainer of DRF) opt to use a regular `Serializer` in situations like these, so I've done the same.

ecommerce/extensions/api/serializers.py
rlucioni: @jimabramson I'd like to leave this as-is. Keeping this switch-checking logic external to `publish_to_lms()` gives us some valuable flexibility, and I'm not sure there's an awful lot which can be reused effectively in this situation. We need the flexibility to log context-sensitive messages when we check the switch, and the ability to take a range of actions when the switch is disabled.

jimabramson: understood @rlucioni, thanks for the consideration and explanation.

rlucioni:
@clintonb @jimabramson this is ready for review.

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,103 @@ class Meta(object):
...
+        return products
+
+    def validate(self, data):
+        """Perform validation spanning multiple fields."""
+        course_id = data['id']
+        products = data['products']
+
+        for product in products:
+            attrs = self._flatten(product['attribute_values'])
+            if attrs.get('course_key') != course_id:
```
clintonb: Let's just ignore this attribute. It isn't necessary for submission because `Course. create_or_update_seat` will handle the association.

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,103 @@ class Meta(object):
...
+    def save(self):
+        """Save and publish Course and associated products."""
+        course_id = self.validated_data['id']
+        course_name = self.validated_data['name']
+        products = self.validated_data['products']
+
+        try:
+            # Explicitly delimit operations which will be rolled back if an exception is raised.
+            with transaction.atomic():
+                course, created = Course.objects.get_or_create(id=course_id)
```
clintonb: Only set the name on creation (e.g. via `defaults` kwarg). I don't think there are plans to support editing course name here.

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,103 @@ class Meta(object):
...
+                course.save()
+
+                for product in products:
+                    attrs = self._flatten(product['attribute_values'])
+
+                    # Strip the 'course_key' attribute; it's not necessary for Seat creation.
+                    attrs.pop('course_key', None)
+
+                    # Convert the value of 'id_verification_required' to a Boolean.
+                    attrs['id_verification_required'] = True if attrs['id_verification_required'] == 'True' else False
```
clintonb: `attrs['id_verification_required'] = (attrs['id_verification_required'] == 'True')`

clintonb:
:+1:

ecommerce/extensions/api/serializers.py

ecommerce/extensions/api/serializers.py
rlucioni: Nice, updated.

ecommerce/extensions/api/serializers.py

ecommerce/extensions/api/serializers.py
rlucioni: @jimabramson following up here: I added this validation, but then removed it in light of the fact that the course admin tool will not be including course IDs as part of the serialized products.

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,86 @@ class Meta(object):
...
+        for product in products:
+            # Verify that each product is intended to be a Seat.
+            product_class = product.get('product_class')
+            if product_class != 'Seat':
+                raise serializers.ValidationError(
+                    _(u"Invalid product class [{product_class}] requested.".format(product_class=product_class))
+                )
+
+            # Verify that a price is present.
+            if product.get('price') is None:
```
clintonb: Honor seats have no price.

clintonb: Disregard. I'll set it to 0.

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,86 @@ class Meta(object):
...
+                course, created = Course.objects.get_or_create(id=course_id, defaults={'name': course_name})
+
+                for product in products:
+                    attrs = self._flatten(product['attribute_values'])
+
+                    # Convert the value of 'id_verification_required' to a Boolean.
+                    attrs['id_verification_required'] = (attrs['id_verification_required'] == 'True')
+
+                    # Extract additional metadata required for Seat creation.
+                    expires = product['expires']
```
clintonb: This field is not relevant to honor modes.

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,86 @@ class Meta(object):
...
+                    attrs['expires'] = dateutil.parser.parse(expires) if expires else None
```
clintonb: This fails with `AttributeError: 'module' object has no attribute 'parser'`. Fixed it by updating the import statement `from dateutil.parser import parse` and call `parse()`.

ecommerce/extensions/api/serializers.py
clintonb: An info message should be logged, but an exception should not be raised here. If we decide to consciously disable publishing (for whatever reason), users should still be able to save data.

ecommerce/extensions/api/serializers.py
rlucioni: Odd, I haven't seen that failure. Updated.

ecommerce/extensions/api/serializers.py
rlucioni: Good catch. Updated.

ecommerce/extensions/api/v2/views.py
```diff
@@ -521,3 +521,35 @@ class ProductViewSet(NestedViewSetMixin, NonDestroyableModelViewSet):
...
+        if course_id is not None:
+            data['id'] = course_id
+
+        serializer = self.get_serializer(data=data)
+        is_valid = serializer.is_valid(raise_exception=True)
+        if is_valid:
+            published, created, error = serializer.save()
+            if published:
+                if created:
+                    return Response(status=status.HTTP_201_CREATED)
```
clintonb: The serialized course/products should be returned.

clintonb: `return Response(serializer.data, status=status.HTTP_201_CREATED if created else status.HTTP_200_OK)`

rlucioni: Updated.

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,88 @@ class Meta(object):
...
+            attrs = self._flatten(product['attribute_values'])
+            if attrs.get('certificate_type') is None:
+                raise serializers.ValidationError(_(u"Products must have a certificate type."))
+            elif attrs.get('id_verification_required') is None:
+                raise serializers.ValidationError(_(u"Products must indicate whether ID verification is required."))
+
+        return products
+
+    def save(self):
+        """Save and publish Course and associated products."""
```
jimabramson: mind adding a description of the return value to the docstring?

jimabramson:
other than docstring request, :+1:

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,88 @@ class Meta(object):
...
+
+                    # Convert the value of 'id_verification_required' to a Boolean.
+                    attrs['id_verification_required'] = (attrs['id_verification_required'] == 'True')
+
+                    # Extract additional metadata required for Seat creation.
+                    expires = product.get('expires')
+                    attrs['expires'] = parse(expires) if expires else None
+                    attrs['price'] = Decimal(product['price'])
+
+                    course.create_or_update_seat(**attrs)
```
clintonb: List out the kwargs. While the creation calls aren't sending a `course_key` field, the edit calls are; and, I'd rather not have to strip out all irrelevant attributes if we later decide to expose additional attributes (e.g. credit_provider).

ecommerce/extensions/api/serializers.py
rlucioni: @jimabramson good call, all set.

ecommerce/extensions/api/serializers.py
rlucioni: @clintonb updated.

ecommerce/extensions/api/serializers.py
```diff
@@ -126,3 +135,104 @@ class Meta(object):
...
         }
+
+
+class AtomicPublicationSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    """Serializer for saving and publishing a Course and associated products.
+
+    Using a ModelSerializer for the Course data makes it difficult to use this serializer to handle updates.
+    The automatically applied validation logic rejects course IDs which already exist in the database.
+    """
+    id = serializers.CharField(max_length=255)
```
clintonb: Use a [`RegexField`](http://www.django-rest-framework.org/api-guide/fields/#regexfield) to ensure the course ID/key is valid.

clintonb:
:+1: (once id field updated)

ecommerce/extensions/api/serializers.py
clintonb: Would be great if you can update `CourseSerializer`, too.

rlucioni: Done (both).

PR #233, opened 2015-07-28
Dropped the usage of dashes in static file names in preference of underscores

Nickersoft:
@jimabramson @clintonb


clintonb:
:+1:

jimabramson:
:+1:   thanks for taking care of this @Nickersoft

Nickersoft:
@jimabramson Np

PR #234, opened 2015-07-29
Corrected require modules

clintonb:
Only top-level pages should be declared as modules.

FYI @awais786 @Nickersoft @jimabramson @rlucioni


rlucioni:
Looks like an ephemeral Selenium failure. :+1:

PR #235, opened 2015-07-31
Allow course name to be updated by CAT

rlucioni:
Facilitates the modification of course names in the CAT edit view. XCOM-533.

@clintonb @jimabramson


clintonb:
:+1:

PR #236, opened 2015-07-31
WIP: Course Admin Tool

clintonb:
This will be split into multiple PRs later today.

FYI @jimabramson @wedaly @rlucioni @AlasdairSwan 

XCOM-520
XCOM-507


ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,156 @@
+require([
+    'backbone',
+    'collections/course_collection',
+    'models/course_model',
+    'views/course_create_edit_view',
+    'views/course_detail_view',
+    'views/course_list_view'
+], function (Backbone,
```
AlasdairSwan: Indentation inconsistent with preferred approach (see ecommerce/static/js/models/course_model.js)

ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,156 @@
...
+             CourseCollection,
+             Course,
+             CourseCreateEditView,
+             CourseDetailView,
+             CourseListView) {
+    'use strict';
+
+    var $app,
+        courseApp = new (Backbone.Router.extend({
```
AlasdairSwan: I highly recommend moving the Router to its own file.

ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,156 @@
...
+
+            notFound: function (path) {
+                alert(path + ' is invalid.');
+            },
+
+            index: function () {
+                document.title = gettext('Courses');
+
+                var collection = new CourseCollection(),
+                    view = new CourseListView({collection: collection});
```
AlasdairSwan: var declarations should be at the top of the function.

ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,156 @@
...
+
+            new: function () {
+                document.title = gettext('Create New Course');
+
+                var model = new Course({}),
+                    view = new CourseCreateEditView({model: model});
+
+                this.currentView = view;
+                view.render();
+                $('#app').html(view.el);
```
AlasdairSwan: It seems that the router is doing too much work here, I would pretty much just expect it to load a page view (maybe with a little code around checking whether view already exists and destroying current view to avoid the dreaded zombie views). I may be easier to discuss in person.

ecommerce/static/js/collections/course_collection.js
```diff
@@ -1,12 +1,14 @@
 define([
-  'underscore',
-  'collections/drf_pageable_collection',
-  'models/course_model'
-], function (_, DrfPageableCollection, CourseModel) {
+    'underscore',
+    'collections/drf_pageable_collection',
+    'models/course_model'
+], function (_,
```
AlasdairSwan: Indentation inconsistent with preferred approach (see ecommerce/static/js/models/course_model.js)

ecommerce/static/js/common.js
```diff
@@ -1,8 +1,12 @@
 require([
     'jquery',
     'backbone',
+    'backbone-super',
     'bootstrap',
     'bootstrap_accessibility',
     'underscore'
 ], function () {
```
AlasdairSwan: Indentation inconsistent with preferred approach (see ecommerce/static/js/models/course_model.js)

ecommerce/static/js/views/alert_view.js
```diff
@@ -0,0 +1,26 @@
...
+
+            initialize: function (options) {
+                this.level = options.level || 'info';
+                this.title = options.title || '';
+                this.message = options.message || '';
+            },
+
+            render: function () {
+                var body = this.template({title: this.title, message: this.message});
+                this.$el.addClass('alert-' + this.level).attr('role', 'alert').html(body);
```
AlasdairSwan, line 21: className can be a function so that should be able to set the alert-level class.

ecommerce/static/js/views/course_create_edit_view.js
```diff
@@ -0,0 +1,53 @@
...
+                this.editing = options.editing;
+            },
+
+            remove: function(){
+                if(this.formView) {
+                    this.formView.remove();
+                    this.formView = null;
+                }
+
+                this._super()
```
AlasdairSwan: Missing ;

ecommerce/static/js/views/course_create_edit_view.js
```diff
@@ -0,0 +1,53 @@
...
+            },
+
+            render: function () {
+                var $html,
+                    data = this.model.attributes;
+
+                this.formView = this.formView || new CourseFormView({editing: this.editing, model: this.model});
```
AlasdairSwan, line 38: It would be cleaner to add a postRender function that inits a formView and passes in the container element it should be added to (.course-form-outer).

ecommerce/static/js/views/course_form_view.js
```diff
@@ -0,0 +1,353 @@
...
+                    message = _s.sprintf('Validation failed:<br><ul>%s</ul>', errorMessages.join('<br>'));
+                    this.clearAlerts();
+                    this.renderAlert('danger', message);
+                    return;
+                }
+
+                // Get the product data
+                var data,
+                    products = [],
+                    activeSeatTypes = this.courseTypeSeatMapping[this.model.get('type')];
```
AlasdairSwan: vars should be declared at the top of the function

ecommerce/static/js/views/course_form_view.js
```diff
@@ -0,0 +1,353 @@
...
+            },
+
+            submit: function (e) {
+                e.preventDefault();
+
+                // Validate the input and display a message, if necessary.
+                var validation = this.model.validate();
+                if (validation) {
+                    var errorMessages,
+                        message;
```
AlasdairSwan: vars should be declared at the top of the function

ecommerce/static/js/views/course_form_view.js
```diff
@@ -0,0 +1,353 @@
...
+
+                }, this);
+
+                data = {
+                    id: this.getFieldValue('id'),
+                    name: this.getFieldValue('name'),
+                    products: products
+                };
+
+                $.ajax({
```
AlasdairSwan, line 338: There should be a model which handles the ajax call.

ecommerce/static/js/views/course_form_view.js
```diff
@@ -0,0 +1,353 @@
...
+                    url: '/api/v2/publication/',
+                    success: function (data, textStatus, jqXHR) {
+                        this.goTo(data.id);
+                    },
+                    error: function (jqXHR, textStatus, errorThrown) {
+                        this.renderAlert('danger', jqXHR.responseJSON.error);
+                        this.$el.animate({scrollTop: 0}, 'slow');
+                    }
+                });
+            }
```
AlasdairSwan, line 354: It seems like this function should be calling 3 smaller functions
- this.validate
- this.getData
- model.submit

ecommerce/extensions/api/serializers.py
```diff
@@ -198,7 +206,7 @@ def save(self):
 
                     # Extract arguments required for Seat creation, deserializing as necessary.
                     certificate_type = attrs['certificate_type']
-                    id_verification_required = (attrs['id_verification_required'] == 'True')
+                    id_verification_required = attrs['id_verification_required']
```
rlucioni, line 221: When is `id_verification_required` converted to a Boolean?

clintonb, line 221: In the bowels of the serializer, there is a call to `to_native()` for each field. I'm 95% certain that's where the conversion takes place.

clintonb, line 221: FYI, this will be split into a separate PR.

rlucioni, line 221: Okay, cool. Since `id_verification_required` isn't one of the fields on the serializer (it's part of the `products` list), I don't think that `to_native()` call is applied to it.

ecommerce/static/js/apps/course_admin_app.js
clintonb: Updated.

ecommerce/static/js/apps/course_admin_app.js
clintonb: Updated.

ecommerce/static/js/collections/course_collection.js
clintonb: Updated.

ecommerce/static/js/common.js
clintonb: Updated.

ecommerce/static/js/views/alert_view.js
clintonb, line 21: `className` is called before `initialize`, so `this.level` is undefined. Let me know if you are aware of a workaround.

ecommerce/static/js/views/course_create_edit_view.js
clintonb: Fixed.

ecommerce/static/js/views/course_form_view.js
clintonb: Updated.

ecommerce/static/js/views/course_form_view.js
clintonb: Updated.

ecommerce/static/js/views/course_form_view.js
clintonb, line 338: Hmm...let's chat about this one. I want to make sure I'm using the nested views/models correctly.

PR #237, opened 2015-07-31
Never save seats if LMS publication is not able to complete

rlucioni:
Avoids the possibility of ghost SKUs on the LMS. XCOM-534.

@clintonb @jimabramson

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -226,19 +226,24 @@ class AtomicPublicationSerializer(serializers.Serializer):  # pylint: disable=ab
                             u'An error occurred while publishing [{course_id}] to LMS. '
                             u'No data has been saved or published.'
                         ).format(course_id=course_id)
+
                         raise Exception(message)
                 else:
                     message = (
                         u'Course [{course_id}] was not published to LMS '
                         u'because the switch [publish_course_modes_to_lms] is disabled. '
-                        u'Data has been saved, but not published.'
+                        u'To avoid ghost SKUs, data has not been saved.'
                     ).format(course_id=course_id)
-                    logger.info(message)
 
-                    return created, None, message
+                    raise Exception(message)
         except Exception as e:  # pylint: disable=broad-except
             logger.exception(u'Failed to save and publish [%s]: [%s]', course_id, e.message)
-            return False, e, e.message
+
+            user_message = (
+                u'Publication of course data to the LMS failed. '
+                u'To avoid checkout failures, this data has NOT been saved.'
+            )
+            return False, e, user_message
 
     def _flatten(self, attrs):
         """Transform a list of attribute names and values into a dictionary keyed on the names."""
--- a/ecommerce/extensions/api/v2/tests/views/test_publication.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_publication.py
@@ -155,12 +155,11 @@ class AtomicPublicationTests(CourseCatalogTestMixin, UserMixin, TestCase):
 
     def test_create(self):
         """Verify that a Course and associated products can be created and published."""
-        # If LMS publication is disabled, the view should return a 201 and data should be saved.
+        # If LMS publication is disabled, the view should return a 500 and data should NOT be saved.
         response = self.client.post(self.create_path, json.dumps(self.data), JSON_CONTENT_TYPE)
-        self.assertEqual(response.status_code, 201)
-        self._assert_course_saved(self.course_id, expected=self.data)
+        self.assertEqual(response.status_code, 500)
+        self._assert_course_saved(self.course_id)
 
-        self._purge_courses()
         self._toggle_publication(True)
 
         with mock.patch.object(LMSPublisher, 'publish') as mock_publish:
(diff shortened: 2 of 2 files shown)


ecommerce/extensions/api/serializers.py
```diff
@@ -226,16 +226,16 @@ def save(self):
...
                     message = (
                         u'Course [{course_id}] was not published to LMS '
                         u'because the switch [publish_course_modes_to_lms] is disabled. '
-                        u'Data has been saved, but not published.'
+                        u'To avoid ghost SKUs, data has not been saved.'
                     ).format(course_id=course_id)
-                    logger.info(message)
 
-                    return created, None, message
+                    raise Exception(message)
```
clintonb, line 238: This message is perfect for logging; but, needs to be updated for the exception message returned to the user: `Publication of course data to LMS failed. This data has NOT been saved to avoid checkout failures.` @srpearce can spruce this up later.

ecommerce/extensions/api/v2/tests/views/test_publication.py
```diff
@@ -155,12 +155,11 @@ def _assert_course_saved(self, course_id, expected=None):
 
     def test_create(self):
         """Verify that a Course and associated products can be created and published."""
-        # If LMS publication is disabled, the view should return a 201 and data should be saved.
+        # If LMS publication is disabled, the view should return a 500 and data should NOT be saved.
         response = self.client.post(self.create_path, json.dumps(self.data), JSON_CONTENT_TYPE)
-        self.assertEqual(response.status_code, 201)
-        self._assert_course_saved(self.course_id, expected=self.data)
+        self.assertEqual(response.status_code, 500)
+        self._assert_course_saved(self.course_id)
```
clintonb, line 161: The course should _not_ be saved in this case.

ecommerce/extensions/api/v2/tests/views/test_publication.py
```diff
@@ -181,13 +180,11 @@ def test_update(self):
         self._purge_courses(recreate=True)
         updated_data = self._update_data()
 
-        # If LMS publication is disabled, the view should return a 200 and data should be saved.
+        # If LMS publication is disabled, the view should return a 500 and data should NOT be saved.
         response = self.client.put(self.update_path, json.dumps(updated_data), JSON_CONTENT_TYPE)
-        self.assertEqual(response.status_code, 200)
-        self._assert_course_saved(self.course_id, expected=updated_data)
+        self.assertEqual(response.status_code, 500)
+        self._assert_course_saved(self.course_id, expected=self.data)
```
clintonb, line 186: Do you test that the data is not updated anywhere?

rlucioni, line 186: I do. When given `expected` data, `_assert_course_saved()` verifies that the data in the system matches the `expected` data. In this case, `self.data` is the original, unmodified data which we would expect to find in the database, as opposed to the `updated_data`.

ecommerce/extensions/api/v2/tests/views/test_publication.py
rlucioni, line 161: Correct. When given no `expected` data (i.e, `expected` is `None`), `_assert_course_saved()` verifies that the course has _not_ been saved.

clintonb:
:+1:

ecommerce/extensions/api/serializers.py
rlucioni, line 238: Updated. In the event that publication is not possible (for whatever reason), detailed exception messages are still logged, but the message returned to the user reads: "Publication of course data to the LMS failed. To avoid checkout failures, this data has NOT been saved."

jimabramson:
@rlucioni  @clintonb  merged already, but I have some questions about this when you have a chance.
- neither the JIRA ticket nor the PR explain what a ghost SKU is, or what scenario causes them to occur.  SKUs appear to be immutable as far as a given course/mode/cert type, so the worst scenario I can think of is where the LMS holds a SKU that is different from the one in Otto, but it doesn't seem like this change would help that.
- in terms of the code, why are we waiting til the very end of serializer logic to decide to throw an error - why not check the waffle switch and abort right away when then user is modifying course products?
- what is the (operational) use case for the waffle switch at this point?

rlucioni:
@jimabramson thanks, those are good points. I'm sorry I merged this before you had a chance to see it; I was trying to get this issue taken care of quickly &mdash; perhaps too quickly &mdash; so we could focus on other obstacles.
- I've added detail to the JIRA ticket explaining what a ghost SKU is, which I'll copy here.

> Allowing users of the CAT to save their data to Otto when publication is disabled may cause SKUs in Otto to differ from the SKUs stored in the CourseModes table. Preventing users from saving data when LMS publication is not an option (for whatever reason) avoids the possibility of these "ghost" SKUs in the LMS.

Say a user of the CAT were to configure a course as professional (with ID verification), then return to the CAT at a later time when publication was disabled and attempt to change their course to professional without ID verification. Since SKU generation currently relies on attributes which can be modified by the CAT, including whether a product requires ID verification, a SKU would be generated and saved which differed from the SKU stored on the LMS. The result would be failed Otto product lookups. If we add to the set of attributes hashed during SKU generation (e.g., credit provider(s)), it becomes more likely than an attribute used for SKU generation may be modified by a CAT user.

This change prevents the issue from occurring by preventing any data from being saved when publication to the LMS isn't possible. A more sophisticated solution would be to add an additional stockrecord associated with the new SKU, but that would require additional development time.
- Great point. I've opened #238 making your suggested optimization.
- The value of the switch is that it gives us the ability to limit access to the CourseModes table. Perhaps we might need to perform some operation requiring that the CourseModes table not be modified for some period of time, or someone might start using the CAT to make changes they shouldn't. Having the switch as an option would make dealing with either of these situations easier.

jimabramson:
@rlucioni thanks for the detailed explanation (and the rapid response in the first place).

I had assumed that product SKUs were immutable and strongly bound to the course id + cert type + verification - i.e. you wouldn't change a prof-ed course into a no-id-prof-ed course, but rather create the new one (and potentially retire or soft-delete the old one).  Maybe we can consider moving in this direction, if possible, once we get clear of the immediate issues.

rlucioni:
@jimabramson I agree that we should think harder about how we approach this issue. We'll run into it again when we work on supporting "audit" courses: a CAT user might change an honor seat into an honor seat without a certificate.

PR #238, opened 2015-08-01
Optimize publication endpoint by checking Waffle switch earlier.

rlucioni:
It doesn't make sense to wait until the very end of serializer logic to decide whether a user should be allowed to save their data. This change allows the serializer to avoid wasting time doing unnecessary work.

@jimabramson, as you suggested in #237. @clintonb, FYI.

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -187,6 +187,15 @@ class AtomicPublicationSerializer(serializers.Serializer):  # pylint: disable=ab
         products = self.validated_data['products']
 
         try:
+            if not waffle.switch_is_active('publish_course_modes_to_lms'):
+                message = (
+                    u'Course [{course_id}] was not published to LMS '
+                    u'because the switch [publish_course_modes_to_lms] is disabled. '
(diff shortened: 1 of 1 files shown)


clintonb:
:+1:

ecommerce/extensions/api/serializers.py
```diff
@@ -187,37 +187,45 @@ def save(self):
...
-                if waffle.switch_is_active('publish_course_modes_to_lms'):
+            if not waffle.switch_is_active('publish_course_modes_to_lms'):
+                message = (
+                    u'Course [{course_id}] was not published to LMS '
+                    u'because the switch [publish_course_modes_to_lms] is disabled. '
+                    u'To avoid ghost SKUs, data has not been saved.'
+                ).format(course_id=course_id)
+
+                raise Exception(message)
+            else:
```
jimabramson: minor: there's no way out of the `if` block above except an exception, so this can be dedented a level (without `else`)

jimabramson:
:+1:  whether or not you decide to dedent.   Assume there isn't anything else using create_or_update_seat in a way that requires us to add this check there as well?

ecommerce/extensions/api/serializers.py

PR #239, opened 2015-08-02
Add option to run Python unit tests in parallel

rlucioni:
This is meant to help accelerate local development. We can take advantage of nose's [multiprocess plugin](http://nose.readthedocs.org/en/latest/doc_tests/test_multiprocess/multiprocess.html) to bring the execution time of our Python unit tests down from about 95 seconds to about 19 seconds. The multiprocess plugin and Selenium don't play nicely together, so any tests requiring a browser are disabled when running tests in parallel.

@jimabramson @clintonb, impressions?


jimabramson:
run python faster and skip selenium?   count me in!   :+1: :+1:

clintonb:
:+1: Improvement: add another command that only runs Selenium tests. That way we can use this on Travis.

PR #215, opened 2015-07-22
Order Notification

ahsan-ul-haq:
ECOM-1885
This Pull request include the work for sending email notifications to the users which will be triggered at different places.
- Basic functionality for sending email notifications `send_notification`.
- Added a base email notification html `email_base.html` (with default styling).
- Added basic template for purchase order `commtype_course_purchased_body.html`
- Added new app `premailer==2.9.2` for add inline styling in email notifications.
- Updated README to include instructions for creating and sending a specific email notification.
- Added myself to AUTHORS

PR #240, opened 2015-08-03
Credit Payment - Email Receipt

ahsan-ul-haq:
ECOM-1796

### DIFF
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -1,14 +1,21 @@
+import logging
+
 import analytics
-import waffle
 from django.dispatch import receiver
-from oscar.core.loading import get_class
+import waffle
 
 from ecommerce.courses.utils import mode_for_seat
 from ecommerce.extensions.analytics.utils import is_segment_configured, parse_tracking_context, log_exceptions
+from ecommerce.extensions.checkout.utils import get_provider_data
 from ecommerce.notifications.notifications import send_notification
+from ecommerce.settings.base import get_lms_url
+from oscar.core.loading import get_class
 
 
 post_checkout = get_class('checkout.signals', 'post_checkout')
+logger = logging.getLogger(__name__)
+# Number of orders currently supported for the email notifications
+ORDER_LINE_COUNT = 1
 
 
 @receiver(post_checkout, dispatch_uid='tracking.post_checkout_callback')
@@ -55,6 +62,30 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
 def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable=unused-argument
     """Send course purchase notification email when a course is purchased."""
     if waffle.switch_is_active('ENABLE_NOTIFICATIONS'):
-        # Here we assume that order type will always be a 'Seat' i.e. course
-        if len(order.lines.all()) == 1:
-            send_notification(order.user, 'COURSE_PURCHASED', {'course_title': order.lines.first().product.title})
+        # We do not currently support email sending for orders with more than one item.
+        if len(order.lines.all()) == ORDER_LINE_COUNT:
+            product = order.lines.first().product
+            provider_id = getattr(product.attr, 'credit_provider', None)
+            if not provider_id:
+                logger.error(
+                    'Failed to send credit receipt notification. Credit seat product [%s] has not provider.', product.id
+                )
+                return
+            elif product.get_product_class().name == 'Seat':
+                provider_data = get_provider_data(provider_id)
+                if provider_data:
+                    send_notification(
+                        order.user,
+                        'CREDIT_RECEIPT',
+                        {
+                            'course_title': product.title,
+                            'receipt_page_url': get_lms_url(
+                                '/commerce/checkout/receipt/?basket_id={}'.format(order.basket.id)
+                            ),
+                            'credit_hours': product.attr.credit_hours,
+                            'credit_provider': provider_data['display_name'],
+                        }
+                    )
+
+        else:
+            logger.info('Currently support receipt emails for order with one item.')
--- a/ecommerce/extensions/checkout/tests/test_signals.py
+++ b/ecommerce/extensions/checkout/tests/test_signals.py
@@ -1,0 +1,60 @@
+import httpretty
+
+from django.conf import settings
+from django.core import mail
+from django.test import TestCase
+
+from waffle import Switch
+
+from ecommerce.courses.models import Course
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.extensions.checkout.signals import send_course_purchase_email
+from ecommerce.settings.base import get_lms_url
+from oscar.test import factories
+from oscar.test.newfactories import BasketFactory, UserFactory
+
+
+class SignalTests(CourseCatalogTestMixin, TestCase):
+    @httpretty.activate
+    def test_post_checkout_callback(self):
+        """
+        When the post_checkout signal is emitted, the receiver should attempt
+        to fulfill the newly-placed order and send receipt email.
+        """
+        httpretty.register_uri(
+            httpretty.GET, get_lms_url('api/credit/v1/providers/ASU'),
+            body='{"display_name": "Hogwarts"}',
+            content_type="application/json"
+        )
+        Switch.objects.get_or_create(name='ENABLE_NOTIFICATIONS', active=True)
+        user = UserFactory()
+        course = Course.objects.create(id='edX/DemoX/Demo_Course', name='Demo Course')
+        seat = course.create_or_update_seat('credit', False, 50, 'ASU', None, 2)
+
+        basket = BasketFactory()
+        basket.add_product(seat, 1)
+        order = factories.create_order(number=1, basket=basket, user=user)
+        send_course_purchase_email(None, order=order)
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertEqual(mail.outbox[0].subject, 'Order Receipt')
+        self.assertEqual(
+            mail.outbox[0].body,
+            '\nReceipt Confirmation for: {course_name}'
+            '\n\nHi {full_name},\n\n'
+            'Thank you for purchasing {credit_hour} credit hours from {provider_name} for {course_name}.'
+            ' The charge below will appear on your next credit or debit card statement with a '
+            'company name of {platform_name}.\n\nYou can see the status the status of your credit request or '
+            'complete the credit request process on your {platform_name} dashboard\nTo browse other '
+            'credit-eligible courses visit the edX website. More courses are added all the time.\n\n'
+            'Thank you and congratulation on your achievement. We hope you enjoy the course!\n\n'
+            'To view receipt please visit the link below'
+            '\n\n{receipt_url}\n\n'
+            '{platform_name} team\n\nThe edX team\n'.format(
+                course_name=order.lines.first().product.title,
+                full_name=user.get_full_name(),
+                credit_hour=2,
+                provider_name='Hogwarts',
+                platform_name=settings.PLATFORM_NAME,
+                receipt_url=get_lms_url('/commerce/checkout/receipt/?basket_id={}'.format(order.basket.id))
+            )
+        )
--- a/ecommerce/extensions/checkout/utils.py
+++ b/ecommerce/extensions/checkout/utils.py
@@ -1,0 +1,40 @@
+import logging
+import requests
+
+from django.conf import settings
+
+from ecommerce.settings.base import get_lms_url
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_provider_data(provider_id):
+    """Get the provider information for provider id provider.
+
+    Args:
+        provider_id(str): Identifier for the provider
+
+    Returns: dict
+    """
+    provider_info_url = get_lms_url('api/credit/v1/providers/{}'.format(provider_id))
+    timeout = settings.PROVIDER_DATA_PROCESSING_TIMEOUT
+    headers = {
+        'Content-Type': 'application/json',
+        'X-Edx-Api-Key': settings.EDX_API_KEY
+    }
+    try:
+        response = requests.get(provider_info_url, headers=headers, timeout=timeout)
+        if response.status_code == 200:
+            return response.json()
+        else:
+            logger.error(
+                'Failed retrieve provider information for %s provider. Provider API returned status code %d. Error: %s',
+                provider_id, response.status_code, response.text)
+            return None
+    except requests.exceptions.ConnectionError:
+        logger.exception('Connection error occurred during getting data for %s provider', provider_id)
+        return None
+    except requests.Timeout:
+        logger.exception('Failed to retrieve data for %s provider, connection timeout', provider_id)
+        return None
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -236,6 +236,9 @@ ENROLLMENT_API_URL = None
 COMMERCE_API_TIMEOUT = 7
 COMMERCE_API_URL = None
 
+# PROVIDER DATA PROCESSING
+PROVIDER_DATA_PROCESSING_TIMEOUT = 15  # Value is in seconds.
+
 # OAuth2 provider URL used for OAuth2 transactions (e.g. validating access tokens)
 OAUTH2_PROVIDER_URL = None
 # END URL CONFIGURATION
--- a/ecommerce/templates/customer/email_base.html
+++ b/ecommerce/templates/customer/email_base.html
@@ -11,13 +11,53 @@
         .ReadMsgBody { width: 100%; background-color: #ebebeb;}
         .ExternalClass { width: 100%; background-color: #ebebeb;}
         .ExternalClass, .ExternalClass p, .ExternalClass span, .ExternalClass font, .ExternalClass td, .ExternalClass div {line-height:100%;}
-        body {-webkit-text-size-adjust:none;
-            -ms-text-size-adjust:none;
+        body {
             margin:0;
             padding:10px 0;
         }
         table {border-spacing:0;}
         table td {border-collapse:collapse;}
+        .container-pre-header {
+            background-color: #ffffff;
+            padding-left: 30px;
+            padding-right: 30px;
+            font-size: 14px;
+            line-height: 20px;
+            font-family: Open Sans, sans-serif;
+            color: #333;
+            border-top: 10px solid #1AA1E0;
+            border-radius: 5px;
+        }
+        .container-header {
+            padding-left: 30px;
+            padding-right: 30px;
+            font-size: 14px;
+            line-height: 20px;
+            font-family: Open Sans, sans-serif;
+            color: #333;
+            background-color: #f3f3f3;
+            border-top: 1px solid #cccccc;
+            border-bottom: 1px solid #eeeeee;
+        }
+        .container-body{
+            background-color: #ffffff;
+            padding-left: 30px;
+            padding-right: 30px;
+            font-size: 13px;
+            line-height: 20px;
+            font-family: Open Sans, sans-serif;
+            color: #333;
+            border-radius: 5px;
+        }
+        .container-footer{
+            padding-left: 30px;
+            padding-right: 30px;
+            font-size: 13px;
+            line-height: 20px;
+            font-family: Open Sans, sans-serif;
+            color: #002D40;
+            background-color: #f3f3f3;
+        }
         .yshortcuts a {border-bottom: none !important;}
         /* Constrain email width for small screens */
         @media screen and (max-width: 600px) {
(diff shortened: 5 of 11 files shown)


ecommerce/extensions/checkout/signals.py
```diff
@@ -1,14 +1,17 @@
 import analytics
+import logging
+import requests
```
zubair-arbi: Please separate python imports, django imports and third party local imports

ecommerce/extensions/checkout/signals.py
```diff
@@ -1,14 +1,17 @@
...
+from django.dispatch import receiver
 
 from ecommerce.extensions.analytics.utils import is_segment_configured, parse_tracking_context, log_exceptions
 from ecommerce.notifications.notifications import send_notification
-
+from ecommerce.settings.base import get_lms_url
 
 post_checkout = get_class('checkout.signals', 'post_checkout')
-
+logger = logging.getLogger(__name__)
```
zubair-arbi, line 16: two newlines before and after static variables.

ecommerce/extensions/checkout/signals.py
```diff
@@ -57,3 +60,32 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
         # Here we assume that order type will always be a 'Seat' i.e. course
         if len(order.lines.all()) == 1:
             send_notification(order.user, 'COURSE_PURCHASED', {'course_title': order.lines.first().product.title})
+
+@receiver(post_checkout, dispatch_uid='send_order_receipt_email')
+@log_exceptions("Failed to send order receipt email.")
+def send_order_receipt_email(sender, order=None, **kwargs):  # pylint: disable=unused-argument
+    """Send order receipt email after checkout"""
+    import pudb; pu.db
```
zubair-arbi: Remove debugger breakpoint

ecommerce/settings/local.py
```diff
@@ -104,6 +104,9 @@
 EDX_API_KEY = 'replace-me'
 # END ORDER PROCESSING
 
+# PROVIDER DATA PROCESSING
+PROVIDER_DATA_PROCESSING_TIMEOUT = 15
```
zubair-arbi: Please add some unit for value e.g., ms (miliseconds)

ahsan-ul-haq:
@aamir-khan @zubair-arbi @awais786 @tasawernawaz please review.

ecommerce/extensions/checkout/signals.py
```diff
@@ -1,13 +1,19 @@
+import logging
+
 import analytics
-import waffle
-from django.dispatch import receiver
+from django.conf import settings
```
zubair-arbi: Please separate python, django and local imports

ecommerce/extensions/checkout/signals.py
```diff
@@ -57,3 +63,31 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
...
+            'CREDIT_RECEIPT',
+            {
+                'course_title': order.lines.first().product.title,
+                'receipt_page_url': get_lms_url('/commerce/checkout/receipt/?basket_id={}'.format(order.basket.id)),
+                'credit_hours': order.lines.first().product.attr.credit_hours,
+                'credit_provider': provider_data['display_name'],
+            }
+        )
+    else:
+        logger.error('Failed to retrieve data for %s provider', order.lines.first().product.attr.credit_provider)
```
zubair-arbi: replace `order.lines.first().product.attr.credit_provider` with `provider_id`

ecommerce/settings/local.py
```diff
@@ -104,6 +104,9 @@
 EDX_API_KEY = 'replace-me'
 # END ORDER PROCESSING
 
+# PROVIDER DATA PROCESSING
+PROVIDER_DATA_PROCESSING_TIMEOUT = 15 # 15 second because devstack respond slow!
```
zubair-arbi: `PROVIDER_DATA_PROCESSING_TIMEOUT = 15   # Value is in seconds. Devstack respond slow!`

zubair-arbi:
Just minor nitpicks. Once build passes :+1:

ahsan-ul-haq:
@clintonb @wedaly Please review this PR.

ecommerce/extensions/checkout/signals.py
```diff
@@ -57,3 +63,38 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
         # Here we assume that order type will always be a 'Seat' i.e. course
         if len(order.lines.all()) == 1:
             send_notification(order.user, 'COURSE_PURCHASED', {'course_title': order.lines.first().product.title})
+
+
+@receiver(post_checkout, dispatch_uid='send_order_receipt_email')
+@log_exceptions("Failed to send order receipt email.")
+def send_order_receipt_email(sender, order=None, **kwargs):  # pylint: disable=unused-argument
+    """Send order receipt email after checkout"""
+    if len(order.lines.all()) == 1:
```
wedaly: 1) Please add a comment explaining that we do not currently support orders with more than one item.
2) Please add an `else` statement that logs an error if we try to send a receipt email for an order with `> 1` line.

ecommerce/extensions/checkout/signals.py
```diff
@@ -57,3 +63,38 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
...
+            provider_info_url = get_lms_url('api/credit/v1/providers/{}'.format(provider_id))
+            timeout = settings.PROVIDER_DATA_PROCESSING_TIMEOUT
+            headers = {
+                'Content-Type': 'application/json',
+                'X-Edx-Api-Key': settings.EDX_API_KEY
+            }
+
+            response = requests.get(provider_info_url, headers=headers, timeout=timeout)
+            if response.status_code == 200:
+                provider_data = response.json()
```
wedaly: Please move the code to retrieve provider information into a helper method.

ecommerce/extensions/checkout/signals.py
```diff
@@ -57,3 +63,38 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
...
+                        'course_title': order.lines.first().product.title,
+                        'receipt_page_url': get_lms_url(
+                            '/commerce/checkout/receipt/?basket_id={}'.format(order.basket.id)
+                        ),
+                        'credit_hours': order.lines.first().product.attr.credit_hours,
+                        'credit_provider': provider_data['display_name'],
+                    }
+                )
+            else:
+                logger.error('Email sending failed cannot retrieve data for %s provider.', provider_id)
```
wedaly: Please add the response code and content to the error message.  This will be helpful to debug what went wrong.

ecommerce/extensions/checkout/signals.py
```diff
@@ -57,3 +63,38 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
...
+        if hasattr(order.lines.first().product.attr, 'credit_provider'):
+            provider_id = order.lines.first().product.attr.credit_provider
+            provider_info_url = get_lms_url('api/credit/v1/providers/{}'.format(provider_id))
+            timeout = settings.PROVIDER_DATA_PROCESSING_TIMEOUT
+            headers = {
+                'Content-Type': 'application/json',
+                'X-Edx-Api-Key': settings.EDX_API_KEY
+            }
+
+            response = requests.get(provider_info_url, headers=headers, timeout=timeout)
```
wedaly: If the request times out, does this raise an exception?  If so, we should be catching it and logging an error.

ecommerce/extensions/checkout/signals.py
```diff
@@ -57,3 +63,38 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
         # Here we assume that order type will always be a 'Seat' i.e. course
         if len(order.lines.all()) == 1:
             send_notification(order.user, 'COURSE_PURCHASED', {'course_title': order.lines.first().product.title})
+
+
+@receiver(post_checkout, dispatch_uid='send_order_receipt_email')
+@log_exceptions("Failed to send order receipt email.")
+def send_order_receipt_email(sender, order=None, **kwargs):  # pylint: disable=unused-argument
```
wedaly: Could you explain the difference between `send_order_receipt_email` and `send_completed_order_email` (received on line 58)?  I believe the expected behavior is to send exactly one email to the user for a credit course -- is there any chance this implementation will send two?

wedaly: Also, line 62 uses a feature flag to disable notifications.  Do we want to do the same thing here?

ecommerce/extensions/checkout/signals.py
clintonb: Also, the response body, assuming the API returns helpful error messages.

ecommerce/extensions/checkout/signals.py
```diff
@@ -57,3 +63,38 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
...
+    """Send order receipt email after checkout"""
+    if len(order.lines.all()) == 1:
+        if hasattr(order.lines.first().product.attr, 'credit_provider'):
```
clintonb: This can be cleaned up a bit:

``` python
product = order.lines.first().product
provider_id = getattr(product.attr, 'credit_provider', None)

if provider_id:
 ...
```

ecommerce/settings/local.py
```diff
@@ -104,6 +104,9 @@
 EDX_API_KEY = 'replace-me'
 # END ORDER PROCESSING
 
+# PROVIDER DATA PROCESSING
+PROVIDER_DATA_PROCESSING_TIMEOUT = 15  # Value is in seconds. Devstack respond slow!
```
clintonb: This needs to be in `base.py` as well; otherwise, an error will be raised in non-local environments.

clintonb: The fact that this wasn't caught by tests means there also need to be tests for this functionality.

ecommerce/templates/customer/emails/commtype_credit_receipt_body.html
```diff
@@ -0,0 +1,36 @@
...
+<!-- Message Body -->
+<tr>
+    <td class="container-padding" bgcolor="#ffffff" style="background-color: #ffffff; padding-left: 30px; padding-right: 30px; font-size: 13px; line-height: 20px; font-family: Open Sans, sans-serif; color: #333; border-radius: 5px;" align="left">
+        <br>
+        <!--/message HTML content -->
+        <p>{% blocktrans %}Hi {{full_name}},{% endblocktrans %}</p>
+        <br>
+        <p>{% blocktrans with credit_hours=credit_hours credit_provider=credit_provider course_title=course_title platform_name=platform_name %}Thank you for purchasing {{credit_hours}} credit hours from {{credit_provider}} for {{course_title}}. The charge below will appear on your next credit or debit card statement with a company name of {{platform_name}}.{% endblocktrans %}</p>
+        <br>
+        <p>{% blocktrans %}You can see the status the status of your credit request or complete the credit request process on your edX dashboard{% endblocktrans %}</p>
```
clintonb: Replace "edX" with a variable that pulls from `settings.PLATFORM_NAME`.

ecommerce/templates/customer/emails/commtype_credit_receipt_body.html
```diff
@@ -0,0 +1,36 @@
...
+        <p>{% blocktrans %}To browse other credit-eligible courses visit the edX website. More courses are added all the time.{% endblocktrans %}</p>
```
clintonb: Same as above regarding "edX".

ecommerce/templates/customer/emails/commtype_credit_receipt_body.html
```diff
@@ -0,0 +1,36 @@
+{% extends 'customer/email_base.html' %}
```
clintonb, line 1: @marcotuts please advise regarding the content of this email. The text is referencing a "charge below" that doesn't exist.

ecommerce/templates/customer/emails/commtype_credit_receipt_body.html
clintonb: It looks like you already have a `platform_name` variable. Please use it.

ecommerce/templates/customer/emails/commtype_credit_receipt_body.txt
```diff
@@ -0,0 +1,19 @@
+{% load i18n %}
+{% trans "Receipt Confirmation for: " %}{{course_title}}
+
+{% blocktrans %}Hi {{full_name}},{% endblocktrans %}
+
+{% blocktrans with credit_hours=credit_hours credit_provider=credit_provider course_title=course_title platform_name=platform_name %}Thank you for purchasing {{credit_hours}} credit hours from {{credit_provider}} for {{course_title}}. The charge below will appear on your next credit or debit card statement with a company name of {{platform_name}}.{% endblocktrans %}
+
+{% blocktrans %}You can see the status the status of your credit request or complete the credit request process on your edX dashboard{% endblocktrans %}
```
clintonb: Same as above regarding "edX".

ecommerce/extensions/checkout/signals.py
clintonb: Each grouping should also be alphabetized.

ecommerce/extensions/checkout/signals.py
ahsan-ul-haq: Actually when implemented https://github.com/edx/ecommerce/pull/215 we need a way to check whether order notifications is implemented properly or not so I added this. We could remove this as receipt notification is implemented here and only one email need to be send on order completion.

wedaly: Okay, I'd recommend combining this function with `send_course_purchase_email` above to ensure that only one email gets sent.

wedaly:
I agree with @clintonb 's concern about test cases.  We need a test that the credit email is sent on purchase of a credit seat.

ecommerce/templates/customer/emails/commtype_credit_receipt_body.html
marcotuts, line 1: Happy to help here but I'm missing a bit of context unfortunately. Is there a wiki page (or any other resource) anywhere with feature details for this I could read through? I can also check with my team to see if anybody else worked on this in parallel to this request.

clintonb, line 1: This goes back to my earlier questions about the usefulness of this email given that it includes absolutely no order details.

ecommerce/extensions/checkout/signals.py
```diff
@@ -55,6 +61,28 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
 def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable=unused-argument
     """Send course purchase notification email when a course is purchased."""
     if waffle.switch_is_active('ENABLE_NOTIFICATIONS'):
-        # Here we assume that order type will always be a 'Seat' i.e. course
+        # We do not currently support email sending for orders with more than one item.
         if len(order.lines.all()) == 1:
```
awais786: Instead on 1 we can use constant.e.g ORDER_COUNT = 1

ecommerce/extensions/checkout/signals.py
```diff
@@ -55,6 +61,28 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
...
-            send_notification(order.user, 'COURSE_PURCHASED', {'course_title': order.lines.first().product.title})
+            product = order.lines.first().product
+            provider_id = getattr(product.attr, 'credit_provider', None)
+            if provider_id:
```
awais786: In stead of if else some thing like this 
if not provider_id:
logger.error('Failed to send mail! No provider for given course.')
return same thing for nested if

ecommerce/extensions/checkout/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    headers = {
+        'Content-Type': 'application/json',
+        'X-Edx-Api-Key': settings.EDX_API_KEY
+    }
+    try:
+        response = requests.get(provider_info_url, headers=headers, timeout=timeout)
+    except requests.exceptions.ConnectionError:
+        logger.error('Connection error  occurred during getting data for %s provider', provider_id)
+        return
+    if response.status_code == 200:
```
awais786: Move this inside the above try block.

ahsan-ul-haq:
@clintonb @wedaly I have added the test please review.

ecommerce/extensions/checkout/signals.py
```diff
@@ -1,14 +1,21 @@
+import logging
+
+from django.dispatch import receiver
```
clintonb: Alphabetize with analytics and waffle.

ecommerce/extensions/checkout/signals.py
```diff
@@ -1,14 +1,21 @@
...
 from ecommerce.extensions.analytics.utils import is_segment_configured, parse_tracking_context, log_exceptions
+from ecommerce.extensions.checkout.utils import get_provider_data
 from ecommerce.notifications.notifications import send_notification
+from ecommerce.settings.base import get_lms_url
+from oscar.core.loading import get_class
 
 
 post_checkout = get_class('checkout.signals', 'post_checkout')
+logger = logging.getLogger(__name__)
+ORDER_COUNT = 1
```
clintonb: `ORDER_LINE_COUNT`

Also, please add a comment explaining the purpose of this constant.

ahsan-ul-haq:
@clintonb make the changes you suggested please take a look

ecommerce/extensions/checkout/signals.py
```diff
@@ -55,6 +62,28 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
...
     """Send course purchase notification email when a course is purchased."""
     if waffle.switch_is_active('ENABLE_NOTIFICATIONS'):
-        # Here we assume that order type will always be a 'Seat' i.e. course
-        if len(order.lines.all()) == 1:
-            send_notification(order.user, 'COURSE_PURCHASED', {'course_title': order.lines.first().product.title})
+        # We do not currently support email sending for orders with more than one item.
+        if len(order.lines.all()) == ORDER_LINE_COUNT:
+            product = order.lines.first().product
+            provider_id = getattr(product.attr, 'credit_provider', None)
+            if provider_id:
```
clintonb: You should also check to see if the product is actually a credit seat.

ecommerce/extensions/checkout/signals.py
```diff
@@ -55,6 +62,28 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
...
+                provider_data = get_provider_data(provider_id)
+                if provider_data:
```
clintonb, line 76: What happens if there is no data?

ecommerce/extensions/checkout/signals.py
```diff
@@ -55,6 +62,28 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
...
+                            'receipt_page_url': get_lms_url(
+                                '/commerce/checkout/receipt/?basket_id={}'.format(order.basket.id)
+                            ),
+                            'credit_hours': product.attr.credit_hours,
+                            'credit_provider': provider_data['display_name'],
+                        }
+                    )
+
+            else:
+                logger.error('Failed to send mail! No provider for given course.')
```
clintonb: `Failed to send credit receipt notification. Credit seat product [%s] has not provider.`

Set `%s` to the `product.id`.

ecommerce/extensions/checkout/signals.py
```diff
@@ -55,6 +62,28 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
...
+
+        else:
+            logger.error('Currently support receipt emails for order with one item.')
```
clintonb: I'm not certain we care that the notification was not sent in this case. If we do, it's certainly not worthy of an error-level message. This is info-level at best.

ecommerce/extensions/checkout/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    headers = {
+        'Content-Type': 'application/json',
+        'X-Edx-Api-Key': settings.EDX_API_KEY
+    }
+    try:
+        response = requests.get(provider_info_url, headers=headers, timeout=timeout)
+        if response.status_code == 200:
+            return response.json()
+        else:
+            logger.error('Failed retrieve provider information for %s provider. Provider API returned status code %d',
```
clintonb: You also want to log the body of the response, as it _should_ contain a useful error message.

ecommerce/extensions/checkout/utils.py
```diff
@@ -0,0 +1,35 @@
...
+                         provider_id, response.status_code)
+        return
+    except requests.exceptions.ConnectionError:
+        logger.error('Connection error  occurred during getting data for %s provider', provider_id)
```
clintonb: Use `logger.exception`. Does this class also cover `Timeout`? Any other exceptions that need to be handled?

ecommerce/extensions/checkout/utils.py
```diff
@@ -0,0 +1,35 @@
...
+        'X-Edx-Api-Key': settings.EDX_API_KEY
+    }
+    try:
+        response = requests.get(provider_info_url, headers=headers, timeout=timeout)
+        if response.status_code == 200:
+            return response.json()
+        else:
+            logger.error('Failed retrieve provider information for %s provider. Provider API returned status code %d',
+                         provider_id, response.status_code)
+        return
```
clintonb: Provide an explicit return value of `None`.

ecommerce/extensions/checkout/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    except requests.exceptions.ConnectionError:
+        logger.error('Connection error  occurred during getting data for %s provider', provider_id)
+        return
```
clintonb: Provide an explicit return value of `None`.

ecommerce/extensions/fulfillment/tests/test_mail.py
```diff
@@ -0,0 +1,50 @@
...
+from oscar.core.loading import get_class
+from oscar.test import factories
+from oscar.test.newfactories import UserFactory, BasketFactory
+from requests import Response
+from rest_framework import status
+from waffle import Switch
+
+from ecommerce.courses.models import Course
+from ecommerce.extensions.fulfillment.tests.mixins import FulfillmentTestMixin
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
```
clintonb: Alphabetize.

ecommerce/extensions/fulfillment/tests/test_mail.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+post_checkout = get_class('checkout.signals', 'post_checkout')
+
+
+@mock.patch('requests.get')
+class SignalTests(FulfillmentTestMixin, CourseCatalogTestMixin, TestCase):
```
clintonb: Test code should be placed in the same app as the code being tested (when possible). The code being tested lives in `ecommerce/extensions/checkout/signals.py`. The tests should live in the `checkout` app in `ecommerce/extensions/checkout/tests/test_signals.py`.

ecommerce/settings/base.py
```diff
@@ -236,6 +236,9 @@ def get_lms_url(path):
 COMMERCE_API_TIMEOUT = 7
 COMMERCE_API_URL = None
 
+# PROVIDER DATA PROCESSING
+PROVIDER_DATA_PROCESSING_TIMEOUT = 15  # Value is in seconds. Devstack respond slow!
```
clintonb: Remove the bit about devstack.

ecommerce/settings/local.py
clintonb: Unless you are overriding this value with a value different from that in `base.py`, this can be deleted.

ecommerce/templates/customer/email_base.html
```diff
@@ -128,7 +127,7 @@
                                 <table border="0" cellspacing="0" cellpadding="10" width="100" align="right" class="col-2">
                                     <tr>
                                         <td align="center" valign="middle" style="font-size:13px; font-family: Open Sans, Arial, sans-serif; background-color: #127eb1; border-radius: 5px; box-shadow: 0 2px #002D40;">
-                                            <a href="{{course_url}}" style="color: #ffffff; text-decoration: none;">{% trans "View Receipt" %}</a>
+                                            <a href="{{receipt_page_url}}" style="color: #ffffff; text-decoration: none;">{% trans "View Receipt" %}</a>
```
clintonb: This content is very specific to the receipt email, yet is in the base template. If you need content in this location, make a `{% block %}` in the base template, and override it in your credit receipt-specific template.

ecommerce/templates/customer/emails/commtype_credit_receipt_body.html
```diff
@@ -0,0 +1,36 @@
...
+{% load i18n %}
+{% block body %}
+<!-- Message Body -->
+<tr>
+    <td class="container-padding" bgcolor="#ffffff" style="background-color: #ffffff; padding-left: 30px; padding-right: 30px; font-size: 13px; line-height: 20px; font-family: Open Sans, sans-serif; color: #333; border-radius: 5px;" align="left">
```
clintonb: Is this specific to your email, or generalizable enough that it should be part of the base template?

ecommerce/templates/customer/emails/commtype_credit_receipt_body.html
```diff
@@ -0,0 +1,36 @@
...
+        <br>
+        <p>{% trans "Thank you and congratulation on your achievement. We hope you enjoy the course!" %}</p>
+        <p>{% blocktrans %}{{platform_name}} team{% endblocktrans %}</p>
+        <!--/message HTML content -->
+        <br><br>
+    </td>
+</tr>
+
+<tr>
+    <td class="footer-padding" style="padding-left: 30px; padding-right: 30px; font-size: 13px; line-height: 20px; font-family: Open Sans, sans-serif; color: #002D40; background-color: #f3f3f3;" align="left">
```
clintonb: Is this specific to your email, or generalizable enough that it should be part of the base template?

ecommerce/extensions/fulfillment/tests/test_mail.py
```diff
@@ -0,0 +1,50 @@
...
+
+from ecommerce.courses.models import Course
+from ecommerce.extensions.fulfillment.tests.mixins import FulfillmentTestMixin
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+
+
+post_checkout = get_class('checkout.signals', 'post_checkout')
+
+
+@mock.patch('requests.get')
```
clintonb: Use [`httpretty`](http://falcao.it/HTTPretty/).

ecommerce/extensions/fulfillment/tests/test_mail.py
```diff
@@ -0,0 +1,50 @@
...
+            'Receiver not connected to post_checkout signal!',
+        )
+        user = UserFactory()
+        course = Course.objects.create(id='edX/DemoX/Demo_Course', name='Demo Course')
+        seat = course.create_or_update_seat('verified', False, 50, 'ASU', None, 2)
+
+        basket = BasketFactory()
+        basket.add_product(seat, 1)
+        order = factories.create_order(number=1, basket=basket, user=user)
+        post_checkout.send(sender=self.__class__, order=order)
```
clintonb: You don't need any of the fulfillment or post_checkout code. Just call `send_course_purchase_email` directly.

ecommerce/extensions/fulfillment/tests/test_mail.py
```diff
@@ -0,0 +1,50 @@
...
+        self.assert_order_fulfilled(order)
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertEqual(mail.outbox[0].subject, 'Order Receipt')
```
clintonb: What about the email content?

clintonb:
Please attach a copy of the email. Will this PR also setup the configuration necessary to send an email? I don't think SES is setup for Otto.

ecommerce/extensions/checkout/signals.py
ahsan-ul-haq, line 76: error or exception will be logged inside `get_provider_data(provider_id)`

clintonb, line 76: It would be preferable to log an error here, so that when debugging we see exactly where/why the message was not sent. By logging only in `get_provider_data`, we don't know specifically who called `get_provider_data`.

ahsan-ul-haq:
@clintonb please review. Here is the copy of email
screen shot 2015-08-10 at 6 57 58 pm

clintonb:
@ahsan-ul-haq have you confirmed that emails are being sent by Otto? I do not believe this functionality has yet been setup. Is there a plan to do this?

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -0,0 +1,49 @@
...
+from ecommerce.settings.base import get_lms_url
+from oscar.core.loading import get_class
+from oscar.test import factories
+from oscar.test.newfactories import BasketFactory, UserFactory
+
+
+post_checkout = get_class('checkout.signals', 'post_checkout')
+
+
+class SignalTests(FulfillmentTestMixin, CourseCatalogTestMixin, TestCase):
```
clintonb: You don't need any of the fulfillment logic.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+        Switch.objects.get_or_create(name='ENABLE_NOTIFICATIONS', active=True)
+        self.assertIn(
+            'fulfillment.post_checkout_callback',
+            [receiver[0][0] for receiver in post_checkout.receivers],
+            'Receiver not connected to post_checkout signal!',
+        )
+        user = UserFactory()
+        course = Course.objects.create(id='edX/DemoX/Demo_Course', name='Demo Course')
+        seat = course.create_or_update_seat('verified', False, 50, 'ASU', None, 2)
```
clintonb: Shouldn't this be a credit seat?

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -0,0 +1,49 @@
...
+
+        basket = BasketFactory()
+        basket.add_product(seat, 1)
+        order = factories.create_order(number=1, basket=basket, user=user)
+        send_course_purchase_email(sender=self.__class__, order=order)
```
clintonb: Set `sender` to `None` since we know it isn't being used. If it is used later, we can easily track down why the test is broken.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertEqual(mail.outbox[0].subject, 'Order Receipt')
```
clintonb, line 39: It would be nice if you checked the content of the message—links, course name—as well.

ecommerce/extensions/checkout/utils.py
```diff
@@ -0,0 +1,39 @@
...
+import requests
+
+from django.conf import settings
+
+from ecommerce.settings.base import get_lms_url
+
+logger = logging.getLogger(__name__)
+
+
+def get_provider_data(provider_id):
```
clintonb, line 12: This needs tests.

clintonb:
@ahsan-ul-haq I am almost ready to approve. I would like to see tests for `get_provider_data` and a plan/ticket for actually enabling email sending.

ahsan-ul-haq:
@clintonb I have added the tests for the `get_provider_data` could you please check them and confirm the changes.

clintonb:
:+1:

PR #242, opened 2015-08-03
Update publication endpoint

clintonb:
The id_verification_required attribute is being handled as a proper boolean, instead of a string.

XCOM-506


jimabramson:
:+1:

rlucioni:
:+1:

PR #244, opened 2015-08-03
XCOM-519: Added Karma tests for Course Details View

Nickersoft:
Resulted in 100% test coverage

@clintonb @rlucioni @jimabramson


rlucioni:
@Nickersoft sounds like this has been pulled into #247.

clintonb:
@rlucioni @Nickersoft this code indeed has been pulled into #247.

PR #245, opened 2015-08-03
Support migration and publication of course verification deadlines

rlucioni:
XCOM-537.

@clintonb @jimabramson

### DIFF
--- a/ecommerce/extensions/catalogue/management/commands/migrate_course.py
+++ b/ecommerce/extensions/catalogue/management/commands/migrate_course.py
@@ -2,7 +2,7 @@ from __future__ import unicode_literals
 import logging
 from optparse import make_option
 
-import dateutil.parser
+from dateutil.parser import parse
 from django.conf import settings
 from django.core.management import BaseCommand
 from django.db import transaction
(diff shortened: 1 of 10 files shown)


clintonb:
`CourseSerializer` needs to be updated.

rlucioni:
@clintonb updated.

clintonb:
:+1:

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -35,40 +38,65 @@ def _build_lms_url(self, path):
...
+            ))
 
         data = response.json()
         logger.debug(data)
+
         course_name = data['name']
+        if course_name is None:
+            message = u'Aborting migration. No name is available for {}.'.format(self.course.id)
+            logger.error(message)
+            raise Exception(message)
```
jimabramson, line 65: question: if migrating a bunch of courses, and this happens somewhere midway through, will all the successfully migrated courses up til that point still be committed in Otto?

jimabramson:
:+1: other than my question about transaction scope / exception handling

ecommerce/extensions/catalogue/management/commands/migrate_course.py
rlucioni, line 65: My understanding is that yes, those successfully migrated courses will be committed. We're using `transaction.atomic()` to explicitly delimit a block of operations that will be rolled back in the event of an exception. That block of operations is attempted repeatedly in the case of a bulk migration, once for each course; successful runs are committed, while those which encounter exceptions are not.

PR #246, opened 2015-08-03
Initial course app creation and list view migration

clintonb:
- Created Backbone app/router
- Updated list view to operate via new router

XCOM-309, XCOM-337

Breaking down #236 (Part 1 of n)

@AlasdairSwan @Nickersoft @rlucioni @jimabramson


ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,81 @@
...
+
+                    this.currentView = view;
+                    collection.fetch({
+                        success: function () {
+                            view.render();
+                        }
+                    });
+                    $app.html(view.el);
+                }
+            }));
```
AlasdairSwan: It makes sense to move the router to its own file

ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,81 @@
...
+
+                    document.title = gettext('Courses');
+
+                    this.currentView = view;
+                    collection.fetch({
+                        success: function () {
+                            view.render();
+                        }
+                    });
+                    $app.html(view.el);
```
AlasdairSwan: It would be cleaner to have a CourseListPage (passing the element into it) which instantiated the collection and passed it to the CourseListView (which would listen for collection sync and render the view based off that)

ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,81 @@
...
+                        this.currentView.remove();
+                        this.currentView = null;
+                    }
+
+                    Backbone.Router.prototype.navigate.apply(this, arguments);
+                },
+
+                notFound: function (path) {
+                    // TODO Render something!
+                    alert(path + ' is invalid.');
```
AlasdairSwan: A generic error page view would be good

ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,81 @@
...
+                } else {
+                    courseApp.navigate(url, {trigger: true});
+                }
+            };
+
+            $app = $('#app');
+            courseApp.start();
+
+            // Handle navbar clicks.
+            $('a.navbar-brand').on('click', navigate)
```
AlasdairSwan: Where is the navbar? Click handlers should be attached to the view that the DOM element is added with

clintonb: The navbar currently lives outside of the app. It's the main navbar used by the entire site. I'll look into moving it in.

ecommerce/templates/courses/course_app.html
```diff
@@ -0,0 +1,74 @@
...
+
+{% block content %}
+    <div id="app" class="container"></div>
+{% endblock %}
+
+{% block footer %}
+    <footer class="footer">
+        <div class="container">
+            <div class="row">
+                <div class="col-xs-12 text-right"><em>edX Course Administration Tool</em></div>
```
rlucioni: This string should be marked for translation, too.

rlucioni:
:+1:

ecommerce/templates/courses/course_app.html
clintonb: Just a placeholder for now, until we get a real footer.

rlucioni: Okay. It might be a good idea to make a note of that in the template.

ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,81 @@
...
+
+        $(function () {
+            var navigate = function (event) {
+                var url = $(this).attr('href').replace(courseApp.root, '');
+
+                if (event.ctrlKey || event.shiftKey || event.metaKey || event.which == 2) {
+                    return true;
+                }
+
+                event.preventDefault();
```
AlasdairSwan: Is this not above the if statement because it is really the else? If so it could make sense to move to else for clarity.

ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -0,0 +1,81 @@
...
+                    });
+                    $app.html(view.el);
+                }
+            }));
+
+        $(function () {
+            var navigate = function (event) {
+                var url = $(this).attr('href').replace(courseApp.root, '');
+
+                if (event.ctrlKey || event.shiftKey || event.metaKey || event.which == 2) {
```
AlasdairSwan: could event.which be a string (i.e. '2')? If not it would be better to use === instead of ==

ecommerce/static/sass/base/_base.scss
```diff
@@ -2,20 +2,34 @@
...
-    }
+  &:hover,
+  &:focus {
+    text-decoration: none;
+  }
 }
 
 .container {
-    background-color: white;
+  background-color: white;
```
AlasdairSwan: All colours should be variables (even the ones that are words)

ecommerce/static/sass/components/_footer.scss
```diff
@@ -0,0 +1,18 @@
...
+  /* The html and body elements cannot have any padding or margin. */
+}
+
+/* Set the fixed height of the footer here */
+footer.footer {
+  height: $footer-height;
+  margin-top: $footer-margin;
+  border-top: $navbar-border-bottom-width solid palette(secondary, base);
+  padding-top: 8px;
+  background-color: white;
```
AlasdairSwan: All colours should be variables

ecommerce/static/js/apps/course_admin_app.js
clintonb: `event.which` should always be an integer.

ecommerce/static/js/apps/course_admin_app.js
clintonb: Yes, everything below is part of the else statement.

ecommerce/static/js/apps/course_admin_app.js
clintonb: Done.

ecommerce/static/js/apps/course_admin_app.js
clintonb: Done.

ecommerce/static/sass/base/_base.scss
clintonb: Done.

ecommerce/static/sass/components/_footer.scss
clintonb: Done.

ecommerce/templates/courses/course_app.html
clintonb: Updated.

PR #247, opened 2015-08-04
Updated course detail view

clintonb:
- Moved view into course admin app
- Updated styling

XCOM-507

Breaking down #236 (Part 2 of n)
Depends on #246 


AlasdairSwan:
Once comments have been addressed (or will be addressed in the future) :+1:

PR #250, opened 2015-08-04
Ignore audit modes during migration

rlucioni:
XCOM-541.

@jimabramson @clintonb. FYI @wedaly.


clintonb:
:+1:

rlucioni:
@clintonb would you mind taking another look? ~~Based on configuration, credit modes may now also be ignored.~~

rlucioni:
If one edits a course which has a credit mode on the LMS but no credit product on Otto, when changes are published, that credit mode will be nuked. In light of this, I've reverted changes which allow credit modes to be ignored during migration.

rlucioni:
Extending on my comment above: the same will be true for audit modes. Since we can't have audit modes in which many users are enrolled disappearing, I'm going to close this PR and work on supporting migration of legacy audit modes.

PR #251, opened 2015-08-04
Support migration of legacy audit modes

rlucioni:
If a ProductAttribute is saved with a value of None or the empty string, the ProductAttribute is deleted ([reference](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/catalogue/abstract_models.py#L870-L871)). As a consequence, Seats derived from a migrated "audit" mode do not have a certificate_type attribute. XCOM-541.

@clintonb @jimabramson

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -72,13 +72,16 @@ class Course(models.Model):
 
         if mode == 'no-id-professional':
             return 'professional'
+        elif mode == 'audit':
+            # Historically, users enrolled in an 'audit' mode have not received a certificate.
+            return ''
 
         return mode
 
     @property
     def type(self):
         """ Returns the type of the course (based on the available seat types). """
-        seat_types = [(seat.attr.certificate_type or '').lower() for seat in self.seat_products]
+        seat_types = [getattr(seat.attr, 'certificate_type', '').lower() for seat in self.seat_products]
         if 'credit' in seat_types:
             return 'credit'
         elif 'professional' in seat_types or 'no-id-professional' in seat_types:
@@ -100,12 +103,13 @@ class Course(models.Model):
 
     def _get_course_seat_name(self, certificate_type, id_verification_required):
         """ Returns the name for a course seat. """
-        name = 'Seat in {course_name} with {certificate_type} certificate'.format(
-            course_name=self.name,
-            certificate_type=certificate_type)
+        name = u'Seat in {}'.format(self.name)
 
-        if id_verification_required:
-            name += ' (and ID verification)'
+        if certificate_type != '':
+            name += u' with {} certificate'.format(certificate_type)
+
+            if id_verification_required:
+                name += u' (and ID verification)'
 
         return name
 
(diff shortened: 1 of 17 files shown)


rlucioni:
As part of merging this, I'd like to delete the audit seat currently associated with the migrated Linux course. Since this PR changes the way we derive certificate types from course modes &mdash; specifically, audit modes no longer correspond to an "audit" certificate &mdash; the seat in question won't be updated by repeated migration. Objections?

clintonb:
Geez! That escalated quickly!

:+1:

ecommerce/courses/models.py
```diff
@@ -72,13 +72,16 @@ def certificate_type_for_mode(cls, mode):
 
         if mode == 'no-id-professional':
             return 'professional'
+        elif mode == 'audit':
+            # Historically, users enrolled in an 'audit' mode have not received a certificate.
+            return ''
```
jimabramson, line 77: does this have to be empty string instead of null ?

ecommerce/courses/models.py
```diff
@@ -100,12 +103,13 @@ def seat_products(self):
...
     def _get_course_seat_name(self, certificate_type, id_verification_required):
         """ Returns the name for a course seat. """
-        name = 'Seat in {course_name} with {certificate_type} certificate'.format(
-            course_name=self.name,
-            certificate_type=certificate_type)
+        name = u'Seat in {}'.format(self.name)
 
-        if id_verification_required:
-            name += ' (and ID verification)'
+        if certificate_type is not '':
```
jimabramson: nitpick: I'm not used to seeing identity comparison with strings in Python, though I understand this works.  is there a reason you're not using `==` here?

ecommerce/courses/models.py
```diff
@@ -149,7 +158,11 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
         seat.structure = Product.CHILD
         seat.title = self._get_course_seat_name(certificate_type, id_verification_required)
         seat.expires = expires
+
+        # If a ProductAttribute is saved with a value of None or the empty string, the ProductAttribute is deleted.
```
jimabramson, line 162: ah, now i get it

jimabramson:
@rlucioni LGTM but just a sanity check, would this be cleaner if we went with an explicit string value like "none" or "n/a" or "no-certificate" ?   Would there be a downside to that?

ecommerce/courses/models.py
rlucioni, line 77: @jimabramson as you saw below, it can be either; both result in the ProductAttribute being deleted. I went with the empty string for consistency with the default value used elsewhere. Using the empty string as the default value allows us to perform string operations without needing to convert the value of certificate_type first.

ecommerce/courses/models.py
rlucioni: I was using None as the certificate_type value, then switched to the empty string and forgot to update this. Fixed.

rlucioni:
@jimabramson I think deferring to Oscar's default behavior here is actually pretty clean: if a seat has no certificate, it has no certificate_type attribute. Setting the field to an explicit string value seems less clear to me. As it stands, if a seat has a certificate_type, we know that the value of the attribute is the type of the certificate.

jimabramson:
@rlucioni works for me.   :+1: 

Regarding the Linux course will ping with a question offline

PR #252, opened 2015-08-05
Make submission of the course verification deadline optional

rlucioni:
Verification deadline should only be required if the course actually requires verification. XCOM-537.

@clintonb @jimabramson

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -158,7 +158,8 @@ class AtomicPublicationSerializer(serializers.Serializer):  # pylint: disable=ab
     """
     id = serializers.RegexField(COURSE_ID_REGEX, max_length=255)
     name = serializers.CharField(max_length=255)
-    verification_deadline = serializers.DateTimeField()
+    # Verification deadline should only be required if the course actually requires verification.
+    verification_deadline = serializers.DateTimeField(required=False, allow_null=True)
     products = serializers.ListField()
 
     def validate_products(self, products):
@@ -193,7 +194,7 @@ class AtomicPublicationSerializer(serializers.Serializer):  # pylint: disable=ab
         """
         course_id = self.validated_data['id']
         course_name = self.validated_data['name']
-        course_verification_deadline = self.validated_data['verification_deadline']
+        course_verification_deadline = self.validated_data.get('verification_deadline')
         products = self.validated_data['products']
 
         try:
--- a/ecommerce/extensions/api/v2/tests/views/test_publication.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_publication.py
@@ -125,7 +125,7 @@ class AtomicPublicationTests(CourseCatalogTestMixin, UserMixin, TestCase):
             course = Course.objects.get(id=course_id)
             self.assertEqual(course.name, expected['name'])
 
-            verification_deadline = EXPIRES if expected['verification_deadline'] else None
+            verification_deadline = EXPIRES if expected.get('verification_deadline') else None
             self.assertEqual(course.verification_deadline, verification_deadline)
 
             # Validate product structure.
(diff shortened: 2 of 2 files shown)


clintonb:
The field should be required if a course contains a verified course seat.

ecommerce/extensions/api/v2/tests/views/test_publication.py
```diff
@@ -238,3 +238,14 @@ def test_missing_product_price(self):
         response = self.client.post(self.create_path, json.dumps(self.data), JSON_CONTENT_TYPE)
         self.assertEqual(response.status_code, 400)
         self._assert_course_saved(self.course_id)
+
+    def test_verification_deadline_optionalz(self):
```
jimabramson: s/optionalz/optional  (or, update all other tests to also end with a z)

jimabramson:
@clintonb to clarify - is it always required if id verification is required, or only required with verified cert?  also, should it be allowed to be set if verification is not required, since irrelevant?

rlucioni:
@jimabramson I'm implementing it as if it's the former.

ecommerce/extensions/api/v2/tests/views/test_publication.py
rlucioni: Updated; resisted the temptation to end all other tests with a z.

rlucioni:
@clintonb updated.

clintonb:
@jimabramson @rlucioni the updated validation follows my understanding which is that if any seats require ID verification, a verification deadline must be set. Non-verified courses do not require it.

clintonb:
:+1:

wedaly:
@rlucioni Really sorry to bring this up so late in the game.  While it's true that currently all courses with a verified seat have a verification deadline, the primary motivation for separating the dates was so that we could disable the deadline for credit courses.  The thought is that in-course reverification can take the place of a hard verification deadline.

I'm not sure if this is a blocker or not, since we're configuring the first run of credit courses manually.

wedaly:
In sum, I think the validation rules should be:
- Only courses with a verified seat are allowed to have a verification deadline.
- For courses with a verified seat, the verification deadline is optional.
- If a course has both an upgrade and verification deadline set, then the verification deadline must be after the upgrade deadline.

where "verified seat" currently means either a "verified" or "professional" (with verification!) course mode.

ecommerce/extensions/api/serializers.py
```diff
@@ -158,7 +158,8 @@ class AtomicPublicationSerializer(serializers.Serializer):  # pylint: disable=ab
     """
     id = serializers.RegexField(COURSE_ID_REGEX, max_length=255)
     name = serializers.CharField(max_length=255)
-    verification_deadline = serializers.DateTimeField()
+    # Verification deadline should only be required if the course actually requires verification.
+    verification_deadline = serializers.DateTimeField(required=False)
```
clintonb: Also include `allow_null=True`.

rlucioni:
@wedaly I'm not comfortable rushing to implement validation rules that I don't fully understand. Given that 1) it's late in the game, 2) the validation you're describing is already in place on edx-platform, and 3) adding these validation rules to the CAT isn't critical to the migration, would it be okay if we address this kind of validation in a separate story?

Since this validation is present on edx-platform (specifically, within the Commerce API), a user won't be able to, for example, set a verification deadline earlier than a seat's upgrade deadline. If this were attempted, publication to the LMS would fail and the data wouldn't be persisted.

wedaly:
@rlucioni That's fine as long as we have a way to disable the verification deadline (and ensure it stays disabled) for credit courses.

clintonb:
:+1:

jimabramson:
:+1:

rlucioni:
FYI, I've created [XCOM-544](https://openedx.atlassian.net/browse/XCOM-544) to track the addition of the validation rules described above.

PR #253, opened 2015-08-05
Using edX API Key for Course Migration

clintonb:
The migration command now uses the edX API key header for authorization instead of the logged-in user's access token. This ensures we do not encounter API throttling issues.

XCOM-543

@jimabramson @rlucioni


jimabramson:
:+1:

rlucioni:
:+1:

PR #255, opened 2015-08-05
XCOM-501: Advanced search on order/refund list views now displays correctly

Nickersoft:
Due to some unnecessary Bootstrap classes, advanced search modals in Oscar's order/refund list views were not displaying when "Advanced Search" was clicked. These classes were removed and the DOM was updated to reflect the Bootstrap 3 standard.

@rlucioni @clintonb @jimabramson


clintonb:
:+1:

rlucioni:
:+1:

PR #256, opened 2015-08-05
Fall back to the Course Structure API when data is unavailable from the Commerce API

rlucioni:
The migration command can't rely solely on the Commerce API for course name and verification. The Commerce API queries CourseModes, but some courses which exist in the modulestore don't have course modes. XCOM-537.

@clintonb and/or @jimabramson, please review. @feanil, FYI; coming in hot.

### DIFF
--- a/ecommerce/extensions/catalogue/management/commands/migrate_course.py
+++ b/ecommerce/extensions/catalogue/management/commands/migrate_course.py
@@ -11,6 +11,7 @@ import waffle
 
 from ecommerce.courses.models import Course
 
+
 logger = logging.getLogger(__name__)
 
 
@@ -18,13 +19,13 @@ class MigratedCourse(object):
     def __init__(self, course_id):
         self.course, _created = Course.objects.get_or_create(id=course_id)
 
-    def load_from_lms(self):
+    def load_from_lms(self, access_token):
         """
         Loads course products from the LMS.
 
         Loaded data is NOT persisted until the save() method is called.
         """
-        name, verification_deadline, modes = self._retrieve_data_from_lms()
+        name, verification_deadline, modes = self._retrieve_data_from_lms(access_token)
 
         self.course.name = name
         self.course.verification_deadline = verification_deadline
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -94,7 +115,15 @@ def _retrieve_data_from_lms(self):
             'X-Edx-Api-Key': settings.EDX_API_KEY
         }
 
-        course_name, course_verification_deadline = self._query_commerce_api(headers)
+        try:
+            course_name, course_verification_deadline = self._query_commerce_api(headers)
+        except Exception:  # pylint: disable=broad-except
+            course_structure_headers = {
```
clintonb: It would be cleaner to move this inside `_query_course_structure_api` where it is actually used.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -94,7 +115,15 @@ def _retrieve_data_from_lms(self):
             'X-Edx-Api-Key': settings.EDX_API_KEY
         }
 
-        course_name, course_verification_deadline = self._query_commerce_api(headers)
+        try:
+            course_name, course_verification_deadline = self._query_commerce_api(headers)
+        except Exception:  # pylint: disable=broad-except
```
clintonb: Log a warning with the body of the exception, so that we know exactly why the Course Structure API is called.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
rlucioni: Nice call. Updated.

clintonb:
The view needs to be updated to pass the access token.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
rlucioni: Done.

clintonb:
Disregard view comment. Looks like I forgot to remove it earlier.

:+1:

jimabramson:
:+1:

PR #258, opened 2015-08-06
Updated Course List View

clintonb:
backbone.paginator does not meet our needs. It has been replaced, and the CourseCollection has been updated to simply fetch the remaining pages of data when parsing. The view have also been updated to properly add the new data as it is received.

@AlasdairSwan 

FYI @jimabramson @rlucioni


ecommerce/static/js/collections/course_collection.js
```diff
@@ -1,16 +1,26 @@
...
 
-        return DrfPageableCollection.extend({
+        return Backbone.Collection.extend({
             model: CourseModel,
-            url: '/api/v2/courses/'
+            url: '/api/v2/courses/',
+
+            parse: function(response){
+                // Continue retrieving the remaining data
+                if(response.next){
```
rlucioni: Minor: Missing space between if and parentheses.

rlucioni:
Looks good to me. It'd be great if we could get this into the AMIs we're gearing up to deploy.

ecommerce/static/js/collections/course_collection.js
clintonb: Updated.

ecommerce/static/js/pages/course_list_page.js
```diff
@@ -14,8 +14,8 @@ define([
             initialize: function () {
                 this.collection = new CourseCollection();
                 this.view = new CourseListView({collection: this.collection});
-                this.listenTo(this.collection, 'reset', this.render);
-                this.collection.fetch({reset: true});
+                this.render();
```
AlasdairSwan, line 17: Rendering before the collection has returned?

AlasdairSwan, line 17: Shouldn't it at least be listenToOnce so it doesn't render until the first lot of data has returned?

clintonb, line 17: The page only needs to render once. The nested view (CourseListView, in this case) will first render the table. This view listens for collection add events and adds data to the table.

AlasdairSwan, line 17: But doesn't this require 2 DOM updates (add table, populate table), whereas waiting until the collection returns you could do both at once, and thus limiting it to 1 DOM update?

clintonb, line 17: The first call will add the table and any data that happens to exist in the collection. The call to `collection.fetch()` will trigger an `add` event for every item it fetches. `CourseListView` will listen for these events and use the [DataTables API](https://datatables.net/examples/api/add_row.html) to render the rows.

Later, I would like to optimize this to insert the added data as a batch (instead of individually); but, we are in the middle of the migration and I don't have time to figure out that additional optimization.

AlasdairSwan:
:+1:  it would be great if you could create a ticket in Jira for the optimization discussed in the comments so it doesn't get lost @clintonb

ecommerce/static/js/pages/course_list_page.js
clintonb, line 17: [XCOM-549](https://openedx.atlassian.net/browse/XCOM-549) added to track the optimization.

ecommerce/static/js/views/course_list_view.js
```diff
@@ -105,13 +99,23 @@ define([
...
+                this.refreshTableData();
 
                 return this;
-            }
+            },
+
+            /**
+             * Refresh the data table with the collection's current information.
+             */
+            refreshTableData: function () {
```
clintonb, line 112: @AlasdairSwan figured out the right way to do this. I am using the `update` event which is fired once per call to `collection.fetch()`.

I am currently using a page size of 50 courses, so this method will be called about 19 times in stage (~930 courses) and 24 times in production (~1200 courses). Again, this is not ideal. Let me know if you have suggestions regarding further optimizations (aside from increasing the page size).

PR #259, opened 2015-08-06
Request that the Enrollment API include expired course modes in its responses

rlucioni:
XCOM-537.

Demo: Without [`include_expired`](https://courses.edx.org/api/enrollment/v1/course/TUMx/AUTONAVx/2T2014) vs. with [`include_expired=1`](https://courses.edx.org/api/enrollment/v1/course/TUMx/AUTONAVx/2T2014?include_expired=1).

@clintonb @jimabramson


clintonb:
:+1:

PR #260, opened 2015-08-07
Update outdated slugs for existing products

rlucioni:
Our previous method of slug generation did not account for ID verification. This change allows us to correct the structure of existing slugs during migration.

@clintonb and/or @jimabramson, please review. We may want to deploy this change prior to the migration, but it shouldn't be a blocker. In order to update old slugs post-migration, we can re-migrate only the products with outdated slugs.

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -139,6 +139,8 @@ class Course(models.Model):
 
         try:
             seat = Product.objects.get(slug__in=slugs)
+            seat.slug = slug
+
             logger.info(
                 'Retrieved course seat child product with certificate type [%s] for [%s] from database.',
                 certificate_type,


clintonb:
:+1:

clintonb:
The demo seat needs to be updated manually.

rlucioni:
@clintonb right, as do the French pre-calc course and Marketing for Non-Marketers, if I'm understanding correctly. All three were manually migrated and have outdated slugs which won't be picked up during migration. I'm taking care of this now.

rlucioni:
Slugs for all seven manually migrated products (three parent products and four child products) have been updated as follows:
1. parent-cs-edxdemox12014
2. child-cs-honor-edxdemox12014
3. parent-cs-course-v1epflxtrigoexpx2t2015
4. child-cs-honor-course-v1epflxtrigoexpx2t2015
5. child-cs-verified-course-v1epflxtrigoexpx2t2015-id-verified
6. parent-cs-course-v1ubcxmarketing5501x2t2015
7. child-cs-no-id-professional-course-v1ubcxmarketing5501x2t2015

These products should be discovered during migration, associated with Course instances, and modified as necessary (e.g., title correction).

PR #261, opened 2015-08-07
Added support for the display and editing of audit modes

clintonb:
Users can interact with existing audit seats, but will not be able to create new ones.

XCOM-541

@AlasdairSwan @jimabramson @rlucioni @Nickersoft

### DIFF
--- a/ecommerce/static/js/models/course_model.js
+++ b/ecommerce/static/js/models/course_model.js
@@ -89,10 +89,10 @@ define([
              * Mapping of course type to an array of course seat types.
              */
             courseTypeSeatMapping: {
-                honor: ['honor'],
-                verified: ['honor', 'verified'],
+                honor: ['audit', 'honor'],
+                verified: ['audit', 'honor', 'verified'],
                 professional: ['professional'],
-                credit: ['honor', 'verified', 'credit']
+                credit: ['audit', 'honor', 'verified', 'credit']
             },
 
             initialize: function () {
@@ -146,6 +146,7 @@ define([
                         return (product instanceof CourseSeat) && (product.seatType === seatType);
                     });
 
+                // Do NOT create new audit seats
                 if (!seat) {
                     seatClass = CourseUtils.getCourseSeatModel(seatType);
                     seat = new seatClass();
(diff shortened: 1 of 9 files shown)


ecommerce/static/js/models/course_model.js
```diff
@@ -89,10 +89,10 @@ define([
              * Mapping of course type to an array of course seat types.
              */
             courseTypeSeatMapping: {
-                honor: ['honor'],
-                verified: ['honor', 'verified'],
+                honor: ['audit', 'honor'],
```
rlucioni, line 92: This looks like an invalid mode mapping. Historically, courses have had an honor mode only, or audit, honor, and verified modes, but not audit and honor modes. Am I misunderstanding how this mapping is used?

clintonb, line 92: This is intended only for existing courses. It's a blanket targeting for all potential places where audit might be used, regardless of whether it's actually used. Given what our data looks like, I'm leaving nothing to chance.

rlucioni, line 92: Should the mapping in course_form_view.js be updated or removed?

clintonb, line 92: Which mapping? Everything is as it should be for the time being.

rlucioni, line 92: https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/views/course_form_view.js#L86-L92

ecommerce/static/templates/audit_course_seat_form_field.html
```diff
@@ -0,0 +1,13 @@
+<div class="col-sm-12">
+    <div class="seat-type"><%= gettext('Audit') %></div>
+</div>
+<div class="col-sm-4">
+    <label class="price-label"><%= gettext('Price (in USD)') %>:</label> <span class="seat-price">$0.00</span>
```
AlasdairSwan: Labels should have a for property.

AlasdairSwan:
Minor comment re. a label, once this is addressed :+1:

ecommerce/static/js/models/course_model.js
clintonb, line 92: Deleted.

rlucioni:
:+1:

ecommerce/static/templates/audit_course_seat_form_field.html
clintonb: This shouldn't be a label since there are no form fields. Changed to `<span>` and updated styling of `.price-label`.

PR #262, opened 2015-08-07
Strip whitespace from course names during migration

rlucioni:
XCOM-552. Once this is deployed, we'll want to remigrate `RiceX/RELI157x/1T2015` and `RiceX/BIOC300.2x/1T2015`.

@clintonb @jimabramson


clintonb:
:+1:

jimabramson:
:+1:

PR #263, opened 2015-08-07
Fixed display of verification deadline on course form

clintonb:
The HTML5 standard only supports datetime-local fields. We need UTC always. We work around this by removing the timezone component when parsing data from the server, and restoring the timezone component when writing back to the server.

  This workaround is already in place for the seat expiration/upgrade deadline. This commit implements the behavior for the course verification deadline.

XCOM-520

### DIFF
--- a/ecommerce/static/js/test/specs/utils/utils_spec.js
+++ b/ecommerce/static/js/test/specs/utils/utils_spec.js
@@ -1,0 +1,30 @@
+define([
+        'utils/utils'],
+    function (Utils) {
+        describe('stripTimezone', function(){
+            it('should return the input value if the input is empty', function(){
+                expect(Utils.stripTimezone('')).toEqual('');
+                expect(Utils.stripTimezone(null)).toBeNull();
+                expect(Utils.stripTimezone(undefined)).toBeUndefined();
+            });
+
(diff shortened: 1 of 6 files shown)


clintonb:

AlasdairSwan:
:+1:

ecommerce/static/js/test/specs/utils/utils_spec.js
```diff
@@ -0,0 +1,30 @@
...
+        });
+
+        describe('restoreTimezone', function(){
+            it('should return the input value if the input is empty', function(){
+                expect(Utils.restoreTimezone('')).toEqual('');
+                expect(Utils.restoreTimezone(null)).toBeNull();
+                expect(Utils.restoreTimezone(undefined)).toBeUndefined();
+            });
+
+            it('should return the datetime without the timezone component', function(){
```
rlucioni: s/without/with

rlucioni:
:+1: once typo corrected.

ecommerce/static/js/test/specs/utils/utils_spec.js
clintonb: Fixed.

PR #265, opened 2015-08-08
Reporting JS coverage alongside Python

clintonb:


clintonb:

rlucioni:
If this is merged before #267, our builds will fail until we can bump coverage past 95% again. Just something to keep in mind. :+1: otherwise.

PR #266, opened 2015-08-10
Displaying average paid order cost on dashboard

clintonb:
Given that most of our orders are free orders, the current "average order cost" value is mostly useless. This commit adds an additional row to the table—average paid order cost—that will be more useful.

XCOM-558


clintonb:

rlucioni:
Looks good, although I'm not sure about the test failure. :+1: once tests pass.

PR #268, opened 2015-08-10
Avoid product slug collisions

rlucioni:
Avoids product slug collisions by retaining information about periods during slug generation. Periods would normally be stripped entirely. XCOM-562.

@jimabramson @clintonb

### DIFF
--- a/ecommerce/settings/_oscar.py
+++ b/ecommerce/settings/_oscar.py
@@ -152,6 +152,7 @@ OSCAR_REFUND_LINE_STATUS_PIPELINE = {
 }
 # END REFUND PROCESSING
 
+
 # DASHBOARD NAVIGATION MENU
 OSCAR_DASHBOARD_NAVIGATION = [
     {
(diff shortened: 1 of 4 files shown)


ecommerce/settings/_oscar.py
```diff
@@ -229,6 +230,15 @@
 ]
 # END DASHBOARD NAVIGATION MENU
 
+
+# SLUG SETTINGS
+# Mapping applied before slug generation. Use to replace characters which would normally be stripped.
+OSCAR_SLUG_MAP = {
+    '.': '_',
```
jimabramson, line 237: I don't think this will be sufficient since `_` is a legal character in course ids; we'd still be exposed to collisions e.g. 'org/course_1x/123' and 'org/course.1x/123'.

jimabramson:
Since slugification is inherently lossy and therefore prone to collision, I think our fix needs to ensure that we never try to treat the slug as a unique selector.

jimabramson:
@rlucioni can we add an additional course_id criteria in this line?    https://github.com/edx/ecommerce/blob/master/ecommerce/courses/models.py#L141

that might actually make it unnecessary to change slugification.

rlucioni:
@jimabramson we can add a `course` kwarg to that query, which should resolve the issue we're seeing. I'll give it a shot.

clintonb:
@jimabramson @rlucioni Now that all products are linked to courses, we don't need slugs at all. (course_id, certificate_type) should be a valid unique key on which we can locate products.

rlucioni:
@clintonb good call. I'll open a new PR which removes our dependence on slugs.

PR #269, opened 2015-08-10
Eliminate use of product slugs as unique selectors

rlucioni:
Avoids product slug collisions by eliminating our reliance on slugs altogether. XCOM-562.

@jimabramson and @clintonb, please review. This is the successor to https://github.com/edx/ecommerce/pull/268.

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -2,7 +2,7 @@ from __future__ import unicode_literals
 import logging
 
 from django.db import models, transaction
-from django.utils.text import slugify
+from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_model
 from simple_history.models import HistoricalRecords
@@ -35,15 +35,14 @@ class Course(models.Model):
 
     def _create_parent_seat(self):
         """ Create the parent seat product if it does not already exist. """
-        slug = 'parent-cs-{}'.format(slugify(unicode(self.id)))
-        defaults = {
-            'is_discountable': True,
-            'structure': Product.PARENT,
-            'product_class': ProductClass.objects.get(slug='seat')
-        }
-        parent, created = self.products.get_or_create(slug=slug, defaults=defaults)
+        parent, created = self.products.get_or_create(
+            course=self,
+            structure=Product.PARENT,
+            product_class=ProductClass.objects.get(slug='seat'),
+        )
         ProductCategory.objects.get_or_create(category=Category.objects.get(name='Seats'), product=parent)
         parent.title = 'Seat in {}'.format(self.name)
+        parent.is_discountable = True
         parent.attr.course_key = self.id
         parent.save()
 
@@ -98,8 +97,8 @@ class Course(models.Model):
 
     @property
     def seat_products(self):
-        """ Returns a list of course seat Products related to this course. """
-        return list(self.parent_seat_product.children.all().prefetch_related('stockrecords'))
+        """ Returns a queryset of course seat Products related to this course. """
+        return self.parent_seat_product.children.all().prefetch_related('stockrecords')
 
     def _get_course_seat_name(self, certificate_type, id_verification_required):
         """ Returns the name for a course seat. """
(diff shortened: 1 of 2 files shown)


ecommerce/courses/models.py
```diff
@@ -35,13 +35,12 @@ def __unicode__(self):
...
     def _create_parent_seat(self):
         """ Create the parent seat product if it does not already exist. """
-        slug = 'parent-cs-{}'.format(slugify(unicode(self.id)))
         defaults = {
             'is_discountable': True,
             'structure': Product.PARENT,
             'product_class': ProductClass.objects.get(slug='seat')
         }
-        parent, created = self.products.get_or_create(slug=slug, defaults=defaults)
+        parent, created = self.products.get_or_create(course=self, structure=Product.PARENT, defaults=defaults)
```
clintonb: `product_class` should be part of your criteria here. We are quickly approaching a world in which we will sell more than seats.

ecommerce/courses/models.py
```diff
@@ -35,13 +35,12 @@ def __unicode__(self):
 
     def _create_parent_seat(self):
         """ Create the parent seat product if it does not already exist. """
-        slug = 'parent-cs-{}'.format(slugify(unicode(self.id)))
         defaults = {
             'is_discountable': True,
             'structure': Product.PARENT,
```
clintonb: This isn't necessary since you are already searching for it.

ecommerce/courses/models.py
```diff
@@ -124,22 +123,25 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
...
+            )
 
         try:
-            seat = Product.objects.get(slug__in=slugs)
-            seat.slug = slug
+            seat = Product.objects.filter(
+                certificate_type_query,
+                course=self,
+                structure=Product.CHILD
+            ).get(
```
clintonb: Why split this out?

ecommerce/courses/models.py
```diff
@@ -124,22 +123,25 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
...
+                attribute_values__value_text=certificate_type
+            )
 
         try:
-            seat = Product.objects.get(slug__in=slugs)
-            seat.slug = slug
+            seat = Product.objects.filter(
+                certificate_type_query,
+                course=self,
+                structure=Product.CHILD
```
clintonb: Include `product_class`.

ecommerce/courses/models.py
```diff
@@ -124,22 +123,25 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
...
+            # Yields a match if attribute with name 'certificate_type' matches provided value.
+            certificate_type_query = Q(
+                attributes__name='certificate_type',
+                attribute_values__value_text=certificate_type
+            )
 
         try:
-            seat = Product.objects.get(slug__in=slugs)
-            seat.slug = slug
+            seat = Product.objects.filter(
```
clintonb: `Product` can be replaced with `course.products`, or `course.seat_products` if you update that property to return a `QuerySet` instead of a `list`.

ecommerce/courses/models.py
```diff
@@ -124,22 +123,25 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
...
+                certificate_type_query,
+                course=self,
+                structure=Product.CHILD
+            ).get(
+                attributes__name='id_verification_required',
```
clintonb: Why not use `Q` and combine these with the query above?

ecommerce/courses/models.py
rlucioni: Done.

ecommerce/courses/models.py
rlucioni: Good catch, updated.

ecommerce/courses/models.py
rlucioni: Updated.

rlucioni: Modified. Oscar's child products [use the `product_class` assigned to their parent](http://django-oscar.readthedocs.org/en/latest/topics/modelling_your_catalogue.html#parent-and-child-products), but don't have a `product_class` themselves. Instead, I'll filter seats using the parent seat product.

ecommerce/courses/models.py
rlucioni: Updated to use `seat_products`.

ecommerce/courses/models.py
rlucioni: @clintonb I've modified this to use `Q`, but I still need to perform the `filter` and `get` separately because the `Q` objects may both contain queries across `attributes__name`. In this case, applying both `Q` objects simultaneously yields an empty queryset.

rlucioni: Revisiting this, there might be a cleaner way to do it by querying the through model (AttributeValues).

ecommerce/courses/models.py
```diff
@@ -35,13 +35,12 @@ def __unicode__(self):
...
-            'is_discountable': True,
-            'structure': Product.PARENT,
-            'product_class': ProductClass.objects.get(slug='seat')
-        }
-        parent, created = self.products.get_or_create(slug=slug, defaults=defaults)
+        parent, created = self.products.get_or_create(
+            course=self,
+            structure=Product.PARENT,
+            product_class=ProductClass.objects.get(slug='seat'),
+            defaults={'is_discountable': True},
```
clintonb: Since you still need to eventually call `parent.save()`, you might as well set `is_discountable` after you've retrieved/created `parent`.

clintonb:
:+1:

ecommerce/courses/models.py
rlucioni: Updated.

ecommerce/courses/models.py
```diff
@@ -99,7 +98,7 @@ def parent_seat_product(self):
     @property
     def seat_products(self):
         """ Returns a list of course seat Products related to this course. """
```
jimabramson: docstring: looks like this no longer returns a list.

rlucioni: Updated.

ecommerce/courses/models.py
```diff
@@ -124,22 +123,23 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
...
-        slugs = set(slugs)
+        id_verification_required_query = Q(
+            attributes__name='id_verification_required',
+            attribute_values__value_boolean=id_verification_required
+        )
 
         try:
-            seat = Product.objects.get(slug__in=slugs)
-            seat.slug = slug
+            seat = self.seat_products.filter(certificate_type_query).get(id_verification_required_query)
```
jimabramson, line 142: does it not work to just pass both criteria to `.get()` ?

rlucioni, line 142: No, unfortunately it doesn't; see my answer to @clintonb above about separating the `filter` and `get`.

jimabramson:
:+1:

PR #272, opened 2015-08-11
Truncate product titles used when creating PayPal payments

rlucioni:
PayPal requires that item names be at most 127 characters long. XCOM-564.

@clintonb @jimabramson

### DIFF
--- a/ecommerce/extensions/payment/utils.py
+++ b/ecommerce/extensions/payment/utils.py
@@ -1,0 +1,39 @@
+from django.utils.translation import ugettext_lazy as _
+
+
+def middle_truncate(string, chars):
+    """Truncate the provided string, if necessary.
+
+    Cuts excess characters from the middle of the string and replaces
+    them with a string indicating that truncation has occurred.
+
+    Arguments:
+        string (unicode or str): The string to be truncated.
+        chars (int): The character limit for the truncated string.
+
+    Returns:
+        Unicode: The truncated string, of length less than or equal to `chars`.
+            If no truncation was required, the original string is returned.
+
+    Raises:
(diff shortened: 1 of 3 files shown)


clintonb:
:+1:

ecommerce/extensions/payment/utils.py
```diff
@@ -0,0 +1,45 @@
...
+            the truncation indicator.
+    """
+    if len(string) <= chars:
+        return string
+
+    indicator_length = len(indicator)
+    if chars < indicator_length:
+        raise ValueError
+
+    # Calculate the size of the capture groups.
```
jimabramson: (nitpick) is there a reason to do things with regexps instead of using simple string slices?

jimabramson:
:+1: assuming this will internationalize OK, and product is on board.

clintonb:
Our product names aren't translated.

## 

Clinton Blackburn

jimabramson:
@clintonb not all our course titles are English, and not all openedx instances are run in English, so the concern is just to ensure that form of truncation/ellipsis makes sense for those cases.

clintonb:
Got it.

ecommerce/extensions/payment/utils.py
rlucioni: There isn't, other than me thinking too hard at 7 PM. Updated.

rlucioni:
@jimabramson I've attempted to address your i18n concerns by pulling the truncation indicator string into the function and marking it for translation. I've included a translator's comment explaining what the string is used for.

jimabramson:
@rlucioni the i18n approach makes sense to me, thanks!  :+1:

clintonb:
:+1:

PR #274, opened 2015-08-12
Caching Node and Bower Dependencies

clintonb:
This should cut Travis CI build times by at least a minute.

FYI @jimabramson @rlucioni @Nickersoft

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,11 @@ sudo: false
 
 # Cache the pip directory. "cache: pip" doesn't work due to install override. See https://github.com/travis-ci/travis-ci/issues/3239.
 cache:
-  directories:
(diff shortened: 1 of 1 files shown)


rlucioni:
This seems to have added a minute to the build time.

clintonb:
@rlucioni I don't believe so. Build times have been around 9 minutes for the last couple days. In any case, only _subsequent_ builds will take advantage of the caching.

rlucioni:
:+1:

jimabramson:
:+1:

clintonb:
We got some speedups; but, there is room to improve. Specifically, we need to figure out how to cache lxml, pillow (which should really be removed), and libsass.

PR #276, opened 2015-08-12
Using pip wheel to speed up builds

clintonb:


clintonb:
Build times dropped to 4.5 minutes! :trophy: 

FYI @rlucioni @jimabramson @Nickersoft

rlucioni:
:tada:

PR #277, opened 2015-08-13
Sending Course ID to Segment/Google Analytics

clintonb:
Marketing requests that the course ID be sent instead of the product title.

ECOM-2005


clintonb:
@jimabramson @rlucioni

rlucioni:
`ecommerce/extensions/refunds/signals.py` should also be updated to send the course ID instead of the product title.

clintonb:
@rlucioni updated.

rlucioni:
A heavy-hearted :+1: .

PR #278, opened 2015-08-13
Added additional JS tests

clintonb:
XCOM-523


clintonb:
FYI @Nickersoft @rlucioni @jimabramson

PR #281, opened 2015-08-14
Update Course model to allow creation of multiple credit seats

rlucioni:
Courses may have multiple credit seats associated with them. To make this possible, the query used to locate existing seats has been extended to include credit provider in the query. Our SKU generation utility has also been modified to include the credit provider in the hash. XCOM-574.

@clintonb @jimabramson, please review when able.

### DIFF
--- a/ecommerce/courses/tests/test_models.py
+++ b/ecommerce/courses/tests/test_models.py
@@ -107,31 +107,56 @@ class CourseTests(CourseCatalogTestMixin, TestCase):
         """ Verify the method creates or updates a seat Product. """
         course = Course.objects.create(id='a/b/c', name='Test Course')
 
-        # Test creation
-        certificate_type = 'honor'
+        # Test seat creation
+        certificate_type = 'verified'
         id_verification_required = True
-        price = 0
+        price = 5
         course.create_or_update_seat(certificate_type, id_verification_required, price)
 
-        # Two seats: one honor, the other the parent seat product
+        # Two seats: one verified, the other the parent seat product
         self.assertEqual(course.products.count(), 2)
         seat = course.seat_products[0]
         self.assert_course_seat_valid(seat, course, certificate_type, id_verification_required, price)
 
-        # Test update
-        price = 100
-        credit_provider = 'MIT'
-        credit_hours = 2
-        course.create_or_update_seat(
-            certificate_type, id_verification_required, price, credit_provider, credit_hours=credit_hours
-        )
+        # Test seat update
+        price = 10
+        course.create_or_update_seat(certificate_type, id_verification_required, price)
 
         # Again, only two seats with one being the parent seat product.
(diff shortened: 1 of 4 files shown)


ecommerce/courses/tests/test_models.py
```diff
@@ -107,31 +107,50 @@ def test_create_or_update_seat(self):
...
         # Again, only two seats with one being the parent seat product.
         self.assertEqual(course.products.count(), 2)
         seat = course.seat_products[0]
-        self.assert_course_seat_valid(
-            seat, course, certificate_type, id_verification_required, price, credit_provider, credit_hours=credit_hours
-        )
+        self.assert_course_seat_valid(seat, course, certificate_type, id_verification_required, price)
+
+        # Verify that the course can have multiple credit seats added to it
+        credit_data = {'MIT': 2, 'Harvard': 0.5}
```
clintonb, line 133: This one method has grown rather significantly. That's a code smell. Please split the credit changes into a separate test method.

ecommerce/extensions/catalogue/utils.py
```diff
@@ -12,7 +12,8 @@ def generate_sku(product):
     _hash = u' '.join((
         getattr(product.attr, 'certificate_type', '').lower(),
         product.attr.course_key.lower(),
-        unicode(product.attr.id_verification_required)
+        unicode(product.attr.id_verification_required),
+        getattr(product.attr, 'credit_provider', '').lower()
```
clintonb: Instead of calling `.lower()` on each component, call it once on the value of `_hash` after the join.

clintonb:
:+1:

ecommerce/courses/tests/test_models.py
rlucioni, line 133: Updated. Things should smell nicer now.

ecommerce/extensions/catalogue/utils.py
rlucioni: Updated.

ecommerce/extensions/catalogue/utils.py
```diff
@@ -10,11 +10,12 @@ def generate_sku(product):
     # Note: This currently supports seats. In the future, this should
     # be updated to accommodate other product classes.
     _hash = u' '.join((
-        getattr(product.attr, 'certificate_type', '').lower(),
+        getattr(product.attr, 'certificate_type', ''),
         product.attr.course_key.lower(),
```
jimabramson: nitpick: get rid of this `lower()` as well ?

jimabramson:
:+1:  addressing the nitpick is optional

ecommerce/extensions/catalogue/utils.py
rlucioni: Whoops, yes. Good catch. Updated.

PR #282, opened 2015-08-17
Update CAT detail view to support display of credit seats

rlucioni:
XCOM-510. Here's a screenshot of how this appears for a course with two credit seats:

@clintonb and/or @jimabramson, please review when able.

### DIFF
--- a/ecommerce/static/js/test/specs/utils/course_utils_spec.js
+++ b/ecommerce/static/js/test/specs/utils/course_utils_spec.js
@@ -43,5 +43,33 @@ define([
                 });
             });
         });
+
+        describe('filterSeats', function () {
+            it('should filter CourseSeats of the requested type', function () {
+                var honor = new HonorSeat(),
+                    verified = new VerifiedSeat(),
+                    professional = new ProfessionalSeat(),
+                    expected = [
+                        {
+                            input: [honor, verified, professional],
+                            output: {
+                                filtered: [verified],
+                                residual: [honor, professional]
+                            }
+                        },
+                        {
+                            input: [honor, verified, verified, professional],
+                            output: {
+                                filtered: [verified, verified],
+                                residual: [honor, professional]
+                            }
+                        },
+                    ];
+
+                _.each(expected, function (expected) {
+                    expect(CourseUtils.filterSeats(expected.input, 'verified')).toEqual(expected.output);
+                });
+            });
+        });
     }
 );
--- a/ecommerce/static/js/utils/course_utils.js
+++ b/ecommerce/static/js/utils/course_utils.js
@@ -55,7 +55,26 @@ define([
                 return _.sortBy(seatTypes, function (seatType) {
                     return this.seatSortObj[seatType];
                 }, this);
-            }
+            },
+
+            /**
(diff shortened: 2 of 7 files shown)


rlucioni:
I'll be adding tests for the course detail view changes.

rlucioni:
Rudimentary test added.

ecommerce/static/js/test/specs/utils/course_utils_spec.js
```diff
@@ -41,5 +41,33 @@ define([
...
+                            input: [honor, verified, verified, professional],
+                            output: {
+                                filteredSeats: [verified, verified],
+                                residualSeats: [honor, professional]
+                            }
+                        },
+                    ];
+
+                _.each(data, function (data) {
+                    expect(CourseUtils.filterSeats(data.input, 'verified')).toEqual(data.output);
```
jimabramson: check residualSeats too?

jimabramson:
@rlucioni LGTM apart from one question on the js test

ecommerce/static/js/utils/course_utils.js
```diff
@@ -55,7 +55,37 @@ define([
...
+             * @param {String} seatType
+             * @returns {Object}
+             */
+            filterSeats: function (seats, seatType) {
+                var i,
+                    seat,
+                    filteredSeats = [],
+                    residualSeats = [];
+
+                for (i = 0; i < seats.length; i++) {
```
clintonb: [`_.groupBy`](http://underscorejs.org/#groupBy) can do this for you.

ecommerce/static/templates/_course_credit_seats.html
```diff
@@ -0,0 +1,37 @@
...
+                                print(moment.utc(expires).format('lll z'));
+                            }
+                        %></td>
+                    </tr>
+                <% }); %>
+            </tbody>
+        </table>
+    </div>
+
+    <div class="col-sm-4 seat-certificate-type">
```
clintonb: Maintain the consistency of certificate type being in the middle by moving this block above the table block, and [utilizing offsets](http://getbootstrap.com/css/#grid-offsetting) to keep it on its own row:

```
<div class="col-sm-4 col-sm-offset-4 seat-certificate-type">...</div>
<div class="col-sm-4"></div>
<div class="col-md-8">...</div>
```

clintonb:
:+1:  (see notes about groupby and columns)

ecommerce/static/js/test/specs/utils/course_utils_spec.js
rlucioni: @jimabramson both `filteredSeats` and `residualSeats` are checked. The utility returns an object with keys for both, and the test data includes the expected object.

ecommerce/static/js/utils/course_utils.js
rlucioni: Mind blown. Thanks for the tip. Updated.

ecommerce/static/templates/_course_credit_seats.html
rlucioni: Updated.

rlucioni:
Updated screenshot:

PR #283, opened 2015-08-17
Credit Cleanup

clintonb:
- Corrected path to requirements
- Corrected styling
- Updated collections to use initializer and url methods for URL construction
- Removed global variable, lmsUrlRoot
- Updated views to make proper use of events and collections

@AlasdairSwan @zubair-arbi @awais786 @ahsan-ul-haq 

FYI @wedaly @jimabramson @rlucioni


zubair-arbi:
Changes looks good :+1:

ecommerce/static/js/pages/credit_checkout.js
```diff
@@ -18,27 +18,39 @@ require([
...
               Utils,
-              ProviderView,
+              CreditProviderView,
               CreditEligibilityView) {
         'use strict';
 
+        var $container = $('.credit-checkout'),
+            $courseDetails = $container.find('#course-name'),
+            $providerDetails = $container.find('.provider-details'),
+            lmsRootUrl = $container.data('lms-root-url');
```
AlasdairSwan, line 34: It may be worthwhile to be a little more defensive here (and later in the code) so you are checking that the $ variables actually exist on the page before getting their data elements (so it will fail a little more gracefully if they're not there for some reason). This may be going too far 'tho, what do you think?

ecommerce/static/js/views/credit_eligibility_view.js
```diff
@@ -1,48 +1,28 @@
...
-                    // TODO Use moment.format(). See http://momentjs.com/docs/#/displaying/.
-                    formattedDate = deadline.toLocaleString('en-us', {month: 'long'}) + ' ' + deadline.getDay() + ',' +
-                        deadline.getFullYear();
-                    $('.eligibility-details').find('.deadline-date').text(formattedDate);
+                if (this.collection.length) {
+                    deadline = moment.utc(this.collection.at(0).get('deadline'));
+                    $('.eligibility-details').find('.deadline-date').text(deadline.format('LL'));
                     this.toggleProviderContent(true);
                 } else {
                     this.toggleProviderContent(false);
```
AlasdairSwan, line 28: @clintonb this view appears to be updating content that is generated by another view, is this correct? If so, why isn't the view that is rendering the elements with the following classes handling this functionaliy
- eligibility-details
- provider-panel
- error-message

ecommerce/static/js/views/credit_provider_view.js
```diff
@@ -0,0 +1,44 @@
...
+
+            toggleProviderContent: function (isEnabled) {
+                // On request failure hide provider panel and show error message.
+                $('.provider-panel').toggleClass('hide', !isEnabled);
+                $('.error-message').toggleClass('hide', isEnabled);
+            }
+        });
+    }
+);
+
```
AlasdairSwan, line 44: Same comment as credit_eligibility_view.js why is the view that renders the title not handling the update. It looks like toggleProviderContent is error messaging shared by several views, it would be preferable to have a single view that handles the display of error messaging that is listening to changes in an errorMsgModel to update the DOM.

ecommerce/static/js/pages/credit_checkout.js
clintonb, line 34: I'd rather see more tests to make sure those elements are always on the page. I see your point; but, such testing could be a slippery slope where you're testing for every element before you interact with it.

That being said, are you aware of any alternative ways to bootstrap pages/views that may be more reliable?

AlasdairSwan, line 34: I guess it's more that the View isn't adding the DOM elements that it relies on so just a slight nervousness around that (whereas any elements it adds you can be more confident that they are there)

ecommerce/static/js/views/credit_eligibility_view.js
clintonb, line 28: I didn't architect this page and views. I initially set out to fix a few small issues I identified while setting up for a demo. That grew a little bit out of hand into what you see here. I created ECOM-2119 to track a more robust cleanup.

ecommerce/static/js/views/credit_provider_view.js
clintonb, line 44: (See previous note.)

AlasdairSwan:
As @clintonb mentioned ECOM-2119 is going to address my comments this PR is good to merge once the comments have been added to said ticket :+1:

PR #284, opened 2015-08-18
Added test of Course model's product collection

clintonb:


clintonb:
FYI @rlucioni @jimabramson

rlucioni:
:+1:

PR #285, opened 2015-08-18
Properly wrapping tests

clintonb:
Updated offending tests with a describe block representing the object being tested (e.g. course model). This will result in a proper grouping of specs when Jasmine displays the results.

FYI @rlucioni @jimabramson


rlucioni:
Nice. :+1:

PR #286, opened 2015-08-18
Fixed Audit Bug

clintonb:
Previously the publication serializer expected a certificate_type value. This value does not exist for audit seats. The serializer has been updated to use a default value (empty string).

ECOM-2122

@rlucioni @jimabramson


rlucioni:
:+1:

PR #288, opened 2015-08-18
Corrected nesting for spec

clintonb:
FYI @rlucioni @jimabramson


rlucioni:
Whoops, thanks for fixing this.

PR #289, opened 2015-08-19
Price Validation and Display Fixes

clintonb:
- Removed price placeholders
- Displaying price validation errors 

FYI @jimabramson @rlucioni 

ECOM-2075


rlucioni:
:+1:

PR #290, opened 2015-08-19
Validating course verification deadline

clintonb:
This field is validated to ensure the verification deadline always occurs after the course seats' upgrade deadlines.

XCOM-581

@rlucioni @jimabramson @AlasdairSwan

### DIFF
--- a/ecommerce/static/js/models/course_model.js
+++ b/ecommerce/static/js/models/course_model.js
@@ -65,6 +65,24 @@ define([
                 type: {
                     required: true,
                     msg: gettext('You must select a course type.')
+                },
+                verification_deadline: function (value) {
+                    var invalid;
+
+                    // No validation is needed for empty values
+                    if (_.isEmpty(value)) {
+                        return;
+                    }
+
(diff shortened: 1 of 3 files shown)


ecommerce/static/js/models/course_model.js
```diff
@@ -65,6 +65,22 @@ define([
...
                     msg: gettext('You must select a course type.')
+                },
+                verification_deadline: function (value) {
+                    // No validation is needed for empty values
+                    if (_.isEmpty(value)) {
+                        return;
+                    }
+
+                    // Find seats where the verification deadline occurs before the upgrade deadline.
+                    var invalid = _.some(this.seats(), function (seat) {
```
AlasdairSwan: var should be declared at the top of the function.
Also, should this be its own function, i.e. would it be reused elsewhere?

clintonb: I cannot think of where else we might perform this validation.

Declaration updated.

AlasdairSwan:
Once tests pass :+1:

rlucioni:
Posthumous :+1: .

PR #292, opened 2015-08-19
Removed browser references from spec-runner

clintonb:
The removed code paths never seem to be run.

@dsjen we initially copied spec-runner from edx/edx-analytics-dashboard. Are you aware of the `isBrowser` paths having ever been used? Any idea how to trigger them? Running gulp as single run or multi-run always works with this code deleted.

FYI @AlasdairSwan


dsjen:
`isBrowser` is used when the jasmine tests are run in the browser, as opposed to running via gulp/karma.  Good luck!

PR #293, opened 2015-08-20
Added validation for all nested products on the course model

clintonb:
This validation, run before saving, will ensure that all product fields are valid before submitting data to the server.

ECOM-2075


clintonb:
@rlucioni @AlasdairSwan

rlucioni:
:+1:

PR #295, opened 2015-08-24
Fixed product validation bug

clintonb:
Skipping product validation when retrieving data from the server after saving.

ECOM-2075

@rlucioni 

FYI @AlasdairSwan let me know if you know of a better way of skipping validation for server-side data.

### DIFF
--- a/ecommerce/static/js/models/course_model.js
+++ b/ecommerce/static/js/models/course_model.js
@@ -81,7 +81,9 @@ define([
                     }
                 },
                 products: function (value) {
-                    if (!value.isValid()) {
+                    // NOTE (CCB): When syncing from the server, the value is an array. We can safely ignore
+                    // validation in this case since the values from the server should be valid.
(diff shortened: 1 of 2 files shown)


ecommerce/static/js/models/course_model.js
```diff
@@ -81,6 +81,12 @@ define([
                     }
                 },
                 products: function (value) {
+                    // NOTE (CCB): When syncing from the server, the value is an array. We can safely ignore
+                    // validation in this case since the values from the server should be valid.
+                    if (_.isArray(value)) {
+                        return;
+                    }
+
```
rlucioni: Why not `if (!_.isArray(value) && !value.isValid())`, on the same line?

clintonb: Updated.

rlucioni:
Merge conflicts, smh. :+1:

PR #297, opened 2015-08-24
Requirements updates

clintonb:
Review #296 first.

@rlucioni

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -7,8 +7,8 @@ django-libsass==0.3
 django-oscar==1.1
 django_simple_history==1.6.1
 django-waffle==0.10.1
-djangorestframework==3.1.2
-djangorestframework-jwt==1.5.0
+djangorestframework==3.1.3
+djangorestframework-jwt==1.7.1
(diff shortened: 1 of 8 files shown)


requirements/base.txt
```diff
@@ -6,9 +6,9 @@ django_extensions==1.5.5
 django-libsass==0.3
 django-oscar==1.1
 django_simple_history==1.6.1
-django-waffle==0.10
-djangorestframework==3.1.2
-djangorestframework-jwt==1.5.0
+django-waffle==0.10.1
+djangorestframework==3.1.3
```
rlucioni, line 10: 3.2.3 is the latest version of DRF. Is 3.1.3 the earliest version to include support for Django 1.8?

clintonb, line 10: DRF 3.2.x is not fully-compatible with DRF 3.1.x. That upgrade will happen
with XCOM-569. DRF has had Django 1.8 support for a while. This is just the
first opportunity we've taken to update all of our requirements.

rlucioni:
:+1:

PR #298, opened 2015-08-24
Fixed catalog migration test

clintonb:
This test was taking 75s to complete because it was calling the Course Structure API, and timing out. The test has been updated to mock the API.

XCOM-441

@rlucioni


rlucioni:
Nice catch, :+1:. ~~I don't understand why the test didn't fail with a `ConnectionError`. Do you?~~ I'm also questioning that 75 second delay; build times don't seem to have come down significantly as a result of this.

rlucioni:
Ignore my question; I see that any `ConnectionError` would have been caught and handled by the call to `assertRaises`. The migration command's exceptions are too broad...

PR #299, opened 2015-08-24
Update CAT create/edit view to support creation and modification of credit courses

rlucioni:
XCOM-511

@clintonb and @AlasdairSwan, this is still a WIP, but I'd appreciate an early look, if you have the time. I'm still trying to get the table validation to work properly, and I still need to write tests.


ecommerce/static/js/views/course_form_view.js
```diff
@@ -263,19 +261,18 @@ define([
...
                             viewClass,
                             view = this.courseSeatViews[seatType];
 
                         if (!view) {
-                            model = this.model.getOrCreateSeat(seatType);
+                            seats = new ProductCollection(this.model.getOrCreateSeats(seatType));
                             viewClass = this.courseSeatViewMappings[seatType];
 
-                            if (viewClass && model) {
+                            if (viewClass && seats.length >= 1) {
```
AlasdairSwan: Nitpick: > 0 would mean you can reduce the code by 1 character

ecommerce/static/js/views/course_form_view.js
```diff
@@ -311,6 +308,8 @@ define([
                 this.$el.find('.alerts').append(view.el);
                 this.alertViews.push(view);
 
+                $('body').animate({ scrollTop: 0 }, 'fast');
```
AlasdairSwan: This should really be handled at the page level because a view should only interact with its own elements. Is scrolling to the top of the page a common occurrence?

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
```diff
@@ -35,55 +41,110 @@ define([
             },
 
             className: function () {
-                return 'row course-seat ' + this.seatType;
+                var className = 'row ' + this.seatType;
+
+                if (this.seatType != 'credit') {
```
AlasdairSwan: Please use !==

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
```diff
@@ -35,55 +41,110 @@ define([
...
+                // Instantiate new Views handling data binding for each Model in the Collection.
+                this.collection.each( function (seat) {
+                    row = new self.rowView({
+                        model: seat,
+                        isRemovable: false,
+                        parent: self.$el.find('tbody'),
+                        course: self.course,
+                        seatType: self.seatType
+                    });
+                });
```
AlasdairSwan: You can pass context in here so you don't need to alias this to the self var.

```
this.collection.each( function(seat) {
    ...
}, this);
```

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
```diff
@@ -35,55 +41,110 @@ define([
...
                 return this;
             },
 
-            cleanIdVerificationRequired: function (val) {
-                return _s.toBoolean(val);
-            },
+            /**
+             * Renders a table of course seats sharing a common seat type.
+             */
+            renderSeatTable: function () {
```
AlasdairSwan: There should not be a space between function and (
It should be

```
renderSeatTable: function( {
```

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
```diff
@@ -35,55 +41,110 @@ define([
...
+
+                // Inject the wrapper template.
+                this.$el.html(this.template());
+
+                // Instantiate new Views handling data binding for each Model in the Collection.
+                this.collection.each( function (seat) {
+                    row = new self.rowView({
+                        model: seat,
+                        isRemovable: false,
+                        parent: self.$el.find('tbody'),
```
AlasdairSwan: You are using this selector multiple times so it makes sense to assign it to a variable before you go into the each loop

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
```diff
@@ -35,55 +41,110 @@ define([
...
-             * Return the input data from the form fields.
+            /**
+             * Adds a new row to the seat table.
+             *
+             * @param e
              */
-            getData: function () {
-                var data = {},
-                    fields = ['certificate_type', 'id_verification_required', 'price', 'expires'];
+            addSeatTableRow: function (e) {
```
AlasdairSwan: e isn't being used, please remove

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
```diff
@@ -0,0 +1,77 @@
...
+
+                // Add seat model to course product collection. This operation
+                // is idempotent.
+                this.course.get('products').add(this.model);
+
+                this.render();
+            },
+
+            render: function () {
+                this.$el.html(this.template(_.extend({}, this.model.attributes, {isRemovable: this.isRemovable})));
```
AlasdairSwan: Please break this line down, there's too much going on for a single line.

ecommerce/static/templates/credit_course_seat_form_field_row.html
```diff
@@ -0,0 +1,28 @@
+<td class="credit-provider">
+    <div class="input-group">
+        <div class="input-group-addon"><i class="fa fa-building-o" aria-hidden="true"></i></div>
+        <input type="text" id="credit_provider" name="credit_provider" class="form-control" value="">
```
AlasdairSwan: All inputs should have associated labels (for a11y)

ecommerce/static/templates/credit_course_seat_form_field_row.html
```diff
@@ -0,0 +1,28 @@
...
+    </div>
+</td>
+<td class="expires">
+    <div class="input-group">
+        <div class="input-group-addon"><i class="fa fa-calendar" aria-hidden="true"></i></div>
+        <input type="datetime-local" id="expires" name="expires" class="form-control" value="">
+    </div>
+</td>
+<% if (isRemovable) { %>
+<td><span class="table-remove fa fa-trash-o"></span></td>
```
AlasdairSwan: Is this a CTA? If so it should be a button. Need to add screen reader only text explaining the action

ecommerce/static/templates/credit_course_seat_form_field_row.html
```diff
@@ -0,0 +1,28 @@
+<td class="credit-provider">
+    <div class="input-group">
+        <div class="input-group-addon"><i class="fa fa-building-o" aria-hidden="true"></i></div>
```
AlasdairSwan: Is this a CTA? If so it should be a button. Need to add screen reader only text explaining the action. This comment is true of all subsequent input-group-addon instances in the rest of the file.

ecommerce/static/js/models/course_seats/credit_seat.js
```diff
@@ -0,0 +1,32 @@
...
+                        required: true,
+                        msg: gettext('All credit seats must have a credit provider.')
+                    },
+                    credit_hours: {
+                        required: true,
+                        msg: gettext('All credit seats must designate a number of credit hours.')
+                    }
+                }
+            )
+        }, {seatType: 'credit'});
```
AlasdairSwan, line 32: Why is this a separate object?

ecommerce/static/js/views/course_form_view.js
AlasdairSwan: You should also move focus to the top of the page for a11y

ecommerce/static/templates/credit_course_seat_form_field.html
```diff
@@ -0,0 +1,31 @@
...
+                    <th><%= gettext('Price (USD)') %></th>
+                    <th><%= gettext('Credit Hours') %></th>
+                    <th><%= gettext('Upgrade Deadline') %></th>
+                    <th><span class="table-add fa fa-plus"></span></th>
+                </tr>
+            </thead>
+            <tbody></tbody>
+        </table>
+        <!-- NOTE: This help-block is here for validation messages. -->
+        <span class="help-block"></span>
```
AlasdairSwan: You should have an aria-live property so screenreaders catch any validation messages (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)

ecommerce/static/js/models/course_seats/course_seat.js
```diff
@@ -10,6 +10,8 @@ define([
                 expires: null,
                 id_verification_required: null,
                 price: 0,
+                credit_provider: null,
+                credit_hours: 0,
```
clintonb: This should default to null, otherwise non-credit seats will be saved with 0 credit hours.

ecommerce/static/js/models/course_seats/credit_seat.js
```diff
@@ -0,0 +1,32 @@
...
+    ],
+    function (CourseSeat) {
+        'use strict';
+
+        return CourseSeat.extend({
+            defaults: _.extend({}, CourseSeat.prototype.defaults,
+                {
+                    certificate_type: 'credit',
+                    id_verification_required: true,
+                    price: null,
```
clintonb: Price should default to 0.

ecommerce/static/js/models/course_seats/credit_seat.js
clintonb, line 32: This was a workaround for the fact that I could not create a static/class variable. It didn't work, and should be removed.

clintonb, line 32: Correction: it does work. Leave it for now, until I can remember why I didn't use it. Weird.

clintonb, line 32: Some background: https://taurenmills.wordpress.com/2011/10/08/backbone-js-with-class-properties/

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
```diff
@@ -19,6 +21,10 @@ define([
             seatType: null,
             template: null,
 
+            events: {
```
clintonb: Any particular reason this wasn't moved to a separate view class?

ecommerce/static/js/views/course_form_view.js
rlucioni: Updated.

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
rlucioni: Updated to use `_.isEqual()` so we don't have to think about `!=` as opposed to `!==`.

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
```diff
@@ -0,0 +1,77 @@
...
+                this.parent.append(this.$el);
```
clintonb: The parent should control the children. The parent should call this view's render method and handle the appending. The parent should also call this view's remove method to handle cleanup.

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
rlucioni: Thanks for the tip. Updated.

ecommerce/static/templates/credit_course_seat_form_field.html
```diff
@@ -0,0 +1,31 @@
...
+
+    <div class="form-group">
+        <table class="table table-striped credit-seats">
+            <thead>
+                <tr>
+                    <th><%= gettext('Credit Provider') %></th>
+                    <th><%= gettext('Price (USD)') %></th>
+                    <th><%= gettext('Credit Hours') %></th>
+                    <th><%= gettext('Upgrade Deadline') %></th>
+                    <th><span class="table-add fa fa-plus"></span></th>
```
clintonb: Use `<button>` instead of `<span>` and you won't need to deal with changing the CSS pointer. Also, accessibility.

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
rlucioni: Every function definition in ecommerce currently puts a space between `function` and the opening parentheses. Updating this now will make the diff really dirty. Is it okay if we address this in a separate PR?

AlasdairSwan: Sure

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
rlucioni: Updated.

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
rlucioni: Updated.

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
rlucioni: Updated.

ecommerce/static/templates/credit_course_seat_form_field_row.html
rlucioni: Updated.

ecommerce/static/templates/credit_course_seat_form_field.html
rlucioni: Updated.

ecommerce/static/templates/credit_course_seat_form_field_row.html
rlucioni: Updated.

ecommerce/static/templates/credit_course_seat_form_field_row.html
rlucioni: This is not a CTA; it's a decorative icon. Each of these input groups is now labelled by using `aria-labelledby` to link input elements to the headers. Is screen reader-only text still required?

AlasdairSwan: Nope, that should be good.

ecommerce/static/templates/credit_course_seat_form_field.html
rlucioni: Updated with `aria-live="assertive"`.

AlasdairSwan: It should be polite, because manners are important (but also it will wait until what is being currently read is finished before alerting the user)

ecommerce/static/js/models/course_seats/course_seat.js
rlucioni: Nice catch. Updated.

ecommerce/static/js/models/course_seats/credit_seat.js
rlucioni: Updated.

ecommerce/static/templates/credit_course_seat_form_field.html
rlucioni: Cool, I wasn't sure which was preferred for validation messages.

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
rlucioni: Given that this view is responsible for rendering and otherwise controlling tables, it seemed appropriate here.

clintonb: The base CourseSeatFormFieldView doesn't need to know about credit. It would be cleaner to make a CreditCourseSeatFormFieldView (or a more appropriate name).

rlucioni: The base CourseSeatFormFieldView doesn't actually know anything about credit. However, it does know how to manage course seat tables in general. I take advantage of this by extending the base CourseSeatFormFieldView to make a CreditCourseSeatFormFieldView which can render a table of credit seats.

clintonb: The base class has no notion of tables or multiple seats. This should be moved to the credit view class since that's the only place where it is used.

ecommerce/static/js/views/course_form_view.js
rlucioni: Updated.

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
rlucioni: @clintonb updated. Logic for handling seat tables is now only present in the CreditCourseSeatFormFieldView.

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
rlucioni: Updated so that the parent calls its children's render methods and handles appending to the DOM. In light of our earlier conversation, the children still handle their own deletion and cleanup.

rlucioni:
@clintonb and @AlasdairSwan, this is ready for final review.

ecommerce/static/js/utils/utils.js
```diff
@@ -81,6 +85,43 @@ define([
...
+            /**
+             * Bind the provided view for form validation.
+             *
+             * @param {Backbone.View} view
+             */
+            bindValidation: function (view) {
+                /* istanbul ignore next */
+                Backbone.Validation.bind(view, {
+                    valid: function (view, attr) {
+                        var $el = view.$('[name=' + attr + ']'),
```
AlasdairSwan: Shouldn't this be

```
view.$el.find('[name=' + attr + ']'),
```

ecommerce/static/js/utils/utils.js
```diff
@@ -81,6 +85,43 @@ define([
...
+                            className = 'invalid-' + attr,
+                            $msg = $helpBlock.find('.' + className);
+
+                        $msg.remove();
+
+                        $group.removeClass('has-error');
+                        $helpBlock.addClass('hidden');
+                    },
+                    invalid: function (view, attr, error) {
+                        var $el = view.$('[name=' + attr + ']'),
```
AlasdairSwan: Shouldn't this be

```
view.$el.find('[name=' + attr + ']'),
```

ecommerce/static/js/views/course_form_view.js
```diff
@@ -306,11 +310,19 @@ define([
              * @param {Sring} message - Message to display to the user.
              */
             renderAlert: function (level, message) {
-                var view = new AlertView({level: level, title: gettext('Error!'), message: message});
+                var view = new AlertView({level: level, title: gettext('Error!'), message: message}),
+                    $alerts = this.$el.find('.alerts');
```
AlasdairSwan: If there are multiple alerts possible without re-rendering the view it could be worthwhile setting a view variable this.$alerts after render to avoid having to create the jQuery object every time an alert has to be rendered

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
```diff
@@ -0,0 +1,69 @@
...
+                this.course = options.course;
+                this.isRemovable = options.isRemovable;
+
+                this.render();
+            },
+
+            render: function () {
+                var context = _.extend({}, this.model.attributes, {isRemovable: this.isRemovable});
+
+                this.$el.html(this.template(context));
```
AlasdairSwan, line 50: Doesn't passing an extension of this.model instead of the actual model mess-up the two-way data-binding that stickit is doing?

ecommerce/static/js/utils/utils.js
rlucioni: Updated.

ecommerce/static/js/utils/utils.js
rlucioni: Updated.

ecommerce/static/js/views/course_form_view.js
rlucioni: Updated.

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
rlucioni, line 50: @AlasdairSwan I don't think so. By default, stickit uses `view.bindings` and `view.model` to set up bindings. This view has its model set during initialization. Why would passing additional context to the template mess up the data binding?

ecommerce/static/js/models/course_seats/credit_seat.js
```diff
@@ -0,0 +1,32 @@
...
+            ),
+
+            validation: _.extend({}, CourseSeat.prototype.validation,
+                {
+                    credit_provider: {
+                        required: true,
+                        msg: gettext('All credit seats must have a credit provider.')
+                    },
+                    credit_hours: {
+                        required: true,
```
clintonb, line 25: You should also validate that the number is greater than 0.

ecommerce/static/js/test/specs/models/course_model_spec.js
```diff
@@ -17,8 +17,8 @@ define([
 
         var model,
             honorSeat = {
-                id: 9,
```
clintonb, line 20: Why the ID change?

ecommerce/static/js/test/specs/views/course_seat_form_fields/professional_course_seat_form_field_view_spec.js
```diff
@@ -1,38 +0,0 @@
-define([
```
clintonb, line 1: Why remove these specs? Having a spec corresponding to the class seems ideal for keeping files organized and relatively short.

ecommerce/static/js/utils/utils.js
clintonb: `_s.sprintf` will help avoid bugs here.

ecommerce/static/js/views/course_form_view.js
```diff
@@ -287,7 +291,7 @@ define([
                 }
 
                 // Retrieve these after any new renderings.
-                $courseSeats = $courseSeatsContainer.find('.course-seat');
```
clintonb, line 290: Any particular reason for this change? We may eventually get rid of Bootstrap, so using the semantic class name is preferable in this case.

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
```diff
@@ -35,55 +37,22 @@ define([
             },
 
             className: function () {
-                return 'row course-seat ' + this.seatType;
+                return 'row ' + this.seatType + ' course-seat';
```
clintonb, line 40: Why the change given that the output is the same?

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
```diff
@@ -0,0 +1,69 @@
...
+                return this;
+            },
+
+            /**
+             * Removes the selected row from the seat table.
+             */
+            removeSeatTableRow: function () {
+                // Remove deleted seat from course product collection.
+                this.course.get('products').remove(this.model);
+                this.undelegateEvents();
```
clintonb: `remove()` handles this for you: http://backbonejs.org/#View-remove.

ecommerce/static/templates/credit_course_seat_form_field_row.html
```diff
@@ -0,0 +1,43 @@
...
+            min="5" step="1" pattern="\d+" value="<%= price %>">
+    </div>
+</td>
+<td class="credit-hours">
+    <div class="input-group">
+        <div class="input-group-addon">
+            <i class="fa fa-clock-o" aria-hidden="true"></i>
+        </div>
+        <input type="number" class="form-control" id="credit_hours" name="credit_hours"
+            aria-labelledby="credit-hours-label" min="1" max="6" step="1" pattern="\d+" value="">
```
clintonb: Why the max of 6?

clintonb: Also, the `value` attribute can be removed.

ecommerce/static/templates/credit_course_seat_form_field_row.html
```diff
@@ -0,0 +1,43 @@
...
+    </div>
+</td>
+<td class="expires">
+    <div class="input-group">
+        <div class="input-group-addon">
+            <i class="fa fa-calendar" aria-hidden="true"></i>
+        </div>
+        <input type="datetime-local" class="form-control" id="expires" name="expires"
+            aria-labelledby="expires-label" value="">
```
clintonb: This, and any other empty value attributes should be removed.

ecommerce/static/templates/credit_course_seat_form_field_row.html
```diff
@@ -0,0 +1,43 @@
...
+    </div>
+</td>
+<% if (isRemovable) { %>
+<td>
+    <button class="remove-seat btn btn-danger">
+        <span class="sr">Delete course seat</span>
```
clintonb: Mark for translation.

ecommerce/static/templates/credit_course_seat_form_field_row.html
clintonb: There is no "expires-label".

ecommerce/static/templates/credit_course_seat_form_field.html
```diff
@@ -0,0 +1,37 @@
...
+        <table class="table table-striped credit-seats">
+            <thead>
+                <tr>
+                    <th id="credit-provider-label"><%= gettext('Credit Provider') %></th>
+                    <th id="price-label"><%= gettext('Price (USD)') %></th>
+                    <th id="credit-hours-label"><%= gettext('Credit Hours') %></th>
+                    <th id="expires-label"><%= gettext('Upgrade Deadline') %></th>
+                    <th>
+                        <button class="add-seat btn btn-primary">
+                            <span class="sr">Add course seat</span>
```
clintonb: Mark for translation.

ecommerce/static/js/models/course_seats/credit_seat.js
rlucioni, line 25: Updated.

ecommerce/static/js/test/specs/models/course_model_spec.js
rlucioni, line 20: Just a minor reordering so that the IDs proceed more logically in the test data (parent = 7, honor = 8, verified = 9, credit = 10).

ecommerce/static/js/test/specs/views/course_seat_form_fields/professional_course_seat_form_field_view_spec.js
rlucioni, line 1: The specs were testing dead code, which has also been removed.

ecommerce/static/js/views/course_form_view.js
rlucioni, line 290: It's no longer guaranteed that the seat "sections" will all have the `course-seat` class at the top level. Each row in the credit seat table is assigned this class, but the credit seat "section" doesn't have the `course-seat` class. The intention here is to retrieve and hide these seat "sections", not individual course seats.

ecommerce/static/js/views/course_seat_form_fields/course_seat_form_field_view.js
rlucioni, line 40: It's cosmetic, to be consistent with the way the credit seat "section" is assigned its [className](https://github.com/edx/ecommerce/pull/299/files#diff-50d977ab761972b23eb11e40b102324aR29).

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_view.js
```diff
@@ -0,0 +1,92 @@
...
+
+                // Instantiate new Views handling data binding for each Model in the Collection.
+                this.collection.each( function (seat) {
+                    row = new this.rowView({
+                        model: seat,
+                        isRemovable: false,
+                        course: this.course
+                    });
+
+                    row.render();
```
AlasdairSwan, line 63: It looks like the rowView already calls this on initialize (https://github.com/edx/ecommerce/blob/renzo/credit-create-edit/ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js#L46)

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_view.js
```diff
@@ -0,0 +1,92 @@
...
+                    $tableBody.append(row.el);
```
AlasdairSwan: It would be more performant to store all of the elements in a variable and append them in a single function outside the each loop.

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_view.js
```diff
@@ -0,0 +1,92 @@
...
+                    seat = new seatClass(),
+                    /*jshint newcap: true */
+                    row = new this.rowView({
+                        model: seat,
+                        isRemovable: true,
+                        course: this.course
+                    }),
+                    $tableBody = this.$el.find('tbody');
+
+                row.render();
```
AlasdairSwan, line 87: It looks like the rowView already calls this on initialize (https://github.com/edx/ecommerce/blob/renzo/credit-create-edit/ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js#L46)

ecommerce/static/templates/course_form.html
```diff
@@ -1,4 +1,4 @@
-<div class="alerts"></div>
+<div class="alerts" tabindex="-1"></div>
```
AlasdairSwan: There should be an aria-live="polite" attribute here

ecommerce/static/templates/credit_course_seat_form_field_row.html
```diff
@@ -0,0 +1,43 @@
...
+        <i class="fa fa-trash" aria-hidden="true"></i>
+        Delete
```
AlasdairSwan: Mark for translation. Also, with the sr text above this will read as 'Delete course seat Delete". I would recommend moving the sr span above to just after this Delete and updating so line 40 will be

```
Delete<span class="sr"> course seat</span>
```

Translations will make it more interesting than that though :dancers:

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
rlucioni: Removed.

ecommerce/static/templates/credit_course_seat_form_field_row.html
rlucioni: One of the acceptance criteria on the parent story, [XCOM-325](https://openedx.atlassian.net/browse/XCOM-325), is that the user should be able to associate up to 6 credit hours with a credit seat. It seems arbitrary, but I was going by what was written there.

ecommerce/static/templates/credit_course_seat_form_field_row.html
rlucioni: Removed.

ecommerce/static/templates/credit_course_seat_form_field_row.html
rlucioni: Nice catch. Marked.

ecommerce/static/templates/credit_course_seat_form_field.html
rlucioni: Marked.

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_view.js
rlucioni, line 63: Whoops, missed that. `rowView` no longer does that.

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_view.js
rlucioni: Updated.

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_view.js
rlucioni, line 87: Fixed above; `rowView` no longer does that.

ecommerce/static/templates/course_form.html
rlucioni: Added.

ecommerce/static/templates/credit_course_seat_form_field_row.html
rlucioni: @AlasdairSwan I addressed this by removing the unnecessary "Delete" and marking the screenreader text for translation; I did the same for the seat addition button. The icons are sufficient for sighted users.

ecommerce/static/templates/credit_course_seat_form_field_row.html
rlucioni: Updated in light of comments on the ticket.

rlucioni:
@clintonb or @AlasdairSwan, any further comments? I've addressed everything up to this point.

clintonb:
:+1:

rlucioni:
@AlasdairSwan if you have further comments, let me know and I'll address them with a follow-up PR.

PR #300, opened 2015-08-25
Added search to course admin view

clintonb:
Admin users can now search for courses by ID or name.


clintonb:
FYI @rlucioni

rlucioni:
Right off of my todo list. :+1:

PR #301, opened 2015-08-26
Django 1.8 Fixes

clintonb:
ECOM-2158

@rlucioni

### DIFF
--- a/ecommerce/extensions/api/authentication.py
+++ b/ecommerce/extensions/api/authentication.py
@@ -1,4 +1,6 @@
 """JWT authentication scheme for use with DRF."""
+import logging
+
 from django.conf import settings
 from django.contrib.auth import get_user_model
 import requests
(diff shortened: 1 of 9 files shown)


ecommerce/extensions/api/authentication.py
```diff
@@ -59,8 +61,10 @@ def authenticate_credentials(self, payload):
                         is_update = True
                 if is_update:
                     user.save()
-            except:  # pragma: no cover
-                raise exceptions.AuthenticationFailed('User retrieval failed.')
+            except:
```
rlucioni, line 64: You may want to tell pylint to ignore this broad `except`.

clintonb, line 64: Doesn't seem to be an issue.

rlucioni:
:+1: once tests pass.

PR #302, opened 2015-08-27
Update acceptance tests to select country during LMS registration

rlucioni:
With this fix in place, acceptance tests run successfully against stage. The username being created by the `LmsUserMixin` was too long and truncated by Otto, causing later API calls to Otto checking for the user's order/enrollment status to fail.

@jimabramson and @clintonb, FYI.

### DIFF
--- a/acceptance_tests/pages.py
+++ b/acceptance_tests/pages.py
@@ -3,6 +3,7 @@ import urllib
 
 from bok_choy.page_object import PageObject
 from bok_choy.promise import EmptyPromise
+from selenium.webdriver.support.select import Select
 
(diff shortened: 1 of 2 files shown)


acceptance_tests/pages.py
```diff
@@ -83,6 +84,10 @@ def register_and_login(self, username, name, email, password):
         self.q(css='input#register-name').fill(name)
         self.q(css='input#register-email').fill(email)
         self.q(css='input#register-password').fill(password)
+
+        select = Select(self.browser.find_element_by_css_selector('select#register-country'))
+        select.select_by_value('US')
```
jimabramson, line 89: optional: consider selecting a country other than 'US' which might be good for shaking out edge cases.

jimabramson:
:+1:  whether or not you decide to switch the Country.

PR #303, opened 2015-08-28
Adding new strategy for admin users.

awais786:
CAT displays $0 prices for expired products.
To fix this issue I have add new seat-policy-mixin class for admin and also new admin strategy. 
During serilazation assign the admin strategy if user is staff.

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -70,8 +70,9 @@ class ProductSerializer(serializers.HyperlinkedModelSerializer):
         return None
 
     def _get_info(self, product):
-        info = Selector().strategy().fetch_for_product(product)
-        return info
+        return Selector().strategy(
+            request=self.context.get('request')
+        ).fetch_for_product(product)
 
     def get_is_available_to_buy(self, product):
         info = self._get_info(product)
--- a/ecommerce/extensions/partner/strategy.py
+++ b/ecommerce/extensions/partner/strategy.py
@@ -18,11 +18,16 @@ class CourseSeatAvailabilityPolicyMixin(strategy.StockRequired):
         return ProductClass.objects.get(slug='seat')
 
     def availability_policy(self, product, stockrecord):
-        """ A seat is unavailable if the current date is beyond the seat's expiration date. """
-        if product.expires and timezone.now() > product.expires:
-            return availability.Unavailable()
+        """ A product is unavailable for non-admin users if the current date is
+        beyond the product's expiration date. Products are always available for admin users.
+        """
 
-        return super(CourseSeatAvailabilityPolicyMixin, self).availability_policy(product, stockrecord)
+        is_staff = getattr(self.user, 'is_staff', False)
+        is_available = product.expires is None or (product.expires >= timezone.now())
+        if is_staff or is_available:
+            return super(CourseSeatAvailabilityPolicyMixin, self).availability_policy(product, stockrecord)
+        else:
+            return availability.Unavailable()
 
 
 class DefaultStrategy(strategy.UseFirstStockRecord, CourseSeatAvailabilityPolicyMixin,
@@ -32,4 +37,4 @@ class DefaultStrategy(strategy.UseFirstStockRecord, CourseSeatAvailabilityPolicy
 
 class Selector(object):
     def strategy(self, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
-        return DefaultStrategy()
+        return DefaultStrategy(request if hasattr(request, 'user') else None)
--- a/ecommerce/extensions/partner/tests/test_strategy.py
+++ b/ecommerce/extensions/partner/tests/test_strategy.py
@@ -1,15 +1,18 @@
 import datetime
+import ddt
 
-from django.test import TestCase
+from django.test import TestCase, RequestFactory
 from oscar.apps.partner import availability
 import pytz
 
 from ecommerce.courses.models import Course
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.extensions.partner.strategy import DefaultStrategy, Selector
+from ecommerce.tests.mixins import UserMixin
 
 
-class DefaultStrategyTests(CourseCatalogTestMixin, TestCase):
+@ddt.ddt
+class DefaultStrategyTests(CourseCatalogTestMixin, UserMixin, TestCase):
     def setUp(self):
         super(DefaultStrategyTests, self).setUp()
         self.strategy = DefaultStrategy()
@@ -39,6 +42,28 @@ class DefaultStrategyTests(CourseCatalogTestMixin, TestCase):
         stock_record = product.stockrecords.first()
         actual = self.strategy.availability_policy(product, stock_record)
         self.assertIsInstance(actual, availability.Unavailable)
+
+    @ddt.unpack
+    @ddt.data(
+        (True, availability.Available),
+        (False, availability.Unavailable),
+    )
+    def test_expired_seats_availability_for_users(self, is_staff, available):
+        """ A product is unavailable for students if the current date
+        is beyond the product's expiration date. But for Admin products
+        are always available.
+        """
+        self.assert_expired_product_availability(is_staff, available)
+
+    def assert_expired_product_availability(self, is_staff, available):
+        request = RequestFactory()
+        request.user = self.create_user(is_staff=is_staff)
+        strategy = DefaultStrategy(request)
+        product = self.honor_seat
+        product.expires = pytz.utc.localize(datetime.datetime.min)
+        stock_record = product.stockrecords.first()
+        actual = strategy.availability_policy(product, stock_record)
+        self.assertIsInstance(actual, available)
 
 
 class SelectorTests(TestCase):


ecommerce/extensions/api/serializers.py
```diff
@@ -70,8 +70,10 @@ def get_price(self, product):
         return None
 
     def _get_info(self, product):
-        info = Selector().strategy().fetch_for_product(product)
-        return info
+        if self.context['request'].user.is_staff:
```
zubair-arbi: Please add comment that why we need to change the Selectors?

ecommerce/extensions/partner/strategy.py
```diff
@@ -30,6 +30,31 @@ class DefaultStrategy(strategy.UseFirstStockRecord, CourseSeatAvailabilityPolicy
...
+class AdminCourseSeatAvailabilityPolicyMixin(strategy.StockRequired):
+    """
+    Availability policy for Course seats to admin.
+
+    Child seats are always available to admin tool.
+    Parent seats are never available.
+    """
+    @property
+    def seat_class(self):
+        ProductClass = get_model('catalogue', 'ProductClass')
```
zubair-arbi: Please pass keyword arguments for better understanding of values being passed to the method `get_model`

awais786: This is built in method of oscar.

awais786:
@rlucioni Please review this PR. I have made new strategy for this fix.

ecommerce/extensions/partner/strategy.py
```diff
@@ -30,6 +30,31 @@ class DefaultStrategy(strategy.UseFirstStockRecord, CourseSeatAvailabilityPolicy
...
+        return super(AdminCourseSeatAvailabilityPolicyMixin, self).availability_policy(product, stockrecord)
+
+
+class AdminStrategy(strategy.UseFirstStockRecord, AdminCourseSeatAvailabilityPolicyMixin,
+                    strategy.NoTax, strategy.Structured):
+    pass
+
+
 class Selector(object):
     def strategy(self, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
```
rlucioni, line 39: Note the `user` kwarg. Take advantage of this kwarg to add a conditional to the existing `strategy` method which is responsible for selecting the appropriate strategy to instantiate.

ecommerce/extensions/partner/strategy.py
```diff
@@ -30,6 +30,31 @@ class DefaultStrategy(strategy.UseFirstStockRecord, CourseSeatAvailabilityPolicy
...
 
+class AdminCourseSeatAvailabilityPolicyMixin(strategy.StockRequired):
+    """
+    Availability policy for Course seats to admin.
+
+    Child seats are always available to admin tool.
+    Parent seats are never available.
+    """
+    @property
+    def seat_class(self):
```
rlucioni: Subclass `CourseSeatAvailabilityPolicyMixin` so that you don't need to duplicate this property.

ecommerce/extensions/partner/strategy.py
rlucioni, line 39: Please also add tests verifying 1) that expired products are not available to non-staff users, and 2) that they are to staff users.

ecommerce/extensions/partner/strategy.py
```diff
@@ -30,6 +30,31 @@ class DefaultStrategy(strategy.UseFirstStockRecord, CourseSeatAvailabilityPolicy
...
+        ProductClass = get_model('catalogue', 'ProductClass')
+        return ProductClass.objects.get(slug='seat')
+
+    def availability_policy(self, product, stockrecord):
+        """ A seat is always available."""
+        return super(AdminCourseSeatAvailabilityPolicyMixin, self).availability_policy(product, stockrecord)
+
+
+class AdminStrategy(strategy.UseFirstStockRecord, AdminCourseSeatAvailabilityPolicyMixin,
```
rlucioni: Oscar's source tends to name strategies and availability policies after what they do, which makes things a bit more readable/maintainable. For clarity, consider naming this `AlwaysAvailableStrategy`.

ecommerce/extensions/partner/strategy.py
```diff
@@ -30,6 +30,31 @@ class DefaultStrategy(strategy.UseFirstStockRecord, CourseSeatAvailabilityPolicy
     pass
 
 
+class AdminCourseSeatAvailabilityPolicyMixin(strategy.StockRequired):
```
rlucioni: In the same vein, `AlwaysAvailableCourseSeatAvailabilityPolicyMixin` would be a bit clearer.

ecommerce/extensions/partner/strategy.py
clintonb: Just use Oscar's default strategy: https://django-oscar.readthedocs.org/en/releases-1.1/topics/prices_and_availability.html?highlight=strategy#default-strategy. There is no need to create a new `Strategy` or `Selector`.

ecommerce/extensions/api/v2/tests/views/__init__.py
```diff
@@ -19,17 +19,18 @@ def url(self):
     def setUp(self):
         super(OrderDetailViewTestMixin, self).setUp()
 
-        user = self.create_user()
```
clintonb, line 22: Why are any of these changes necessary?

ecommerce/extensions/api/serializers.py
```diff
@@ -70,8 +70,16 @@ def get_price(self, product):
...
     def _get_info(self, product):
-        info = Selector().strategy().fetch_for_product(product)
-        return info
+        """
+        If user is admin then show complete course information despite availability. That's
+        why use the admin-strategy while fetching the product. In case of student fetch the default
+        strategy which will restrict the user's in case of non-availability or if the current date
+        is beyond the seat's expiration date
+        """
+        if self.context['request'].user.is_staff:
```
clintonb: This logic should reside in `Selector.strategy`.

ecommerce/extensions/partner/strategy.py
rlucioni: +1, good call. Oscar's default strategy does exactly what is needed here.

ecommerce/extensions/api/v2/tests/views/__init__.py
awais786, line 22: Following test case was failing. That's why i updated it.

```
ERROR: test_get_order (ecommerce.extensions.api.v2.tests.views.test_orders.OrderDetailViewTests)
  File "/Users/apple/workspace/ecommerece_repo/ecomm/lib/python2.7/site-packages/oscar/apps/partner/strategy.py", line 54, in __init__
    if request and request.user.is_authenticated():
AttributeError: 'WSGIRequest' object has no attribute 'user'
```

clintonb, line 22: Instead of updating the test, update your code to account for the fact that
a request will not always have an attached user (or the user may be
anonymous).

## 

Clinton Blackburn

awais786, line 22: I will recheck the code. But please check this test case.
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/v2/tests/views/__init__.py#L30
I think in this test case request should have user.

ecommerce/extensions/partner/strategy.py
```diff
@@ -32,4 +38,6 @@ class DefaultStrategy(strategy.UseFirstStockRecord, CourseSeatAvailabilityPolicy
 
 class Selector(object):
     def strategy(self, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
+        if request and hasattr(request, 'user'):
```
awais786: This condition is needed here otherwise one of the test case fails. Because that test has a request but without user.

ecommerce/extensions/api/v2/tests/views/__init__.py
awais786, line 22: Error is coming from inside oscar file.site-packages/oscar/apps/partner/strategy.py
But I have added the check.

ecommerce/extensions/partner/strategy.py
```diff
@@ -18,8 +18,14 @@ def seat_class(self):
         return ProductClass.objects.get(slug='seat')
 
     def availability_policy(self, product, stockrecord):
-        """ A seat is unavailable if the current date is beyond the seat's expiration date. """
-        if product.expires and timezone.now() > product.expires:
+        """ For student seat is unavailable if the current date is beyond the seat's
```
zubair-arbi: nit:

```
""" For a student, seat is unavailable if the current date is beyond the seat's
expiration date. For the admin, seat is always available.
"""
```

ecommerce/extensions/partner/strategy.py
```diff
@@ -18,8 +18,14 @@ def seat_class(self):
...
+        expiration date. For admin seat is always available.
+        """
+
+        # If request is valid and have a user object with is_staff=true than seat is available.
```
zubair-arbi: nit: `# A seat will be available only If the request is valid and it has a staff user object.`

zubair-arbi:
Just small nits, once build passes :+1:

awais786:
@clintonb please review this.

ecommerce/extensions/partner/strategy.py
```diff
@@ -18,8 +18,14 @@ def seat_class(self):
         return ProductClass.objects.get(slug='seat')
 
     def availability_policy(self, product, stockrecord):
-        """ A seat is unavailable if the current date is beyond the seat's expiration date. """
-        if product.expires and timezone.now() > product.expires:
+        """ For a student, seat is unavailable if the current date is beyond the seat's
```
clintonb: A product is unavailable for non-admin users if the current date is beyond the products's expiration date. Products are always available for admin users.

ecommerce/extensions/partner/tests/test_strategy.py
```diff
@@ -40,6 +41,32 @@ def test_availability_policy_expired(self):
         actual = self.strategy.availability_policy(product, stock_record)
         self.assertIsInstance(actual, availability.Unavailable)
 
+    def test_expired_seats_availability_for_staff_users(self):
+        """ For staff users If the course seat's expiration date has passed,
+        even than seat should be available.
+        """
+        request = RequestFactory()
```
clintonb: DRY: Create an assertion method that combines the logic of these two tests:

``` python
def assert_expired_product_availability(self, is_staff, available):
 ...
```

ecommerce/extensions/api/serializers.py
```diff
@@ -70,8 +70,12 @@ def get_price(self, product):
         return None
 
     def _get_info(self, product):
-        info = Selector().strategy().fetch_for_product(product)
-        return info
+        """ Passing request object to strategy. It will be use in side default strategy to check
```
clintonb: This comment shouldn't be a docstring. The docstring for a method should describe the methods inputs and outputs, not necessarily its inner workings. Either update the docstring or remove it.

ecommerce/extensions/partner/strategy.py
clintonb: Why not use the user attribute?

awais786:
@clintonb it's ready now.

clintonb:
:+1:

ecommerce/extensions/partner/tests/test_strategy.py
```diff
@@ -40,6 +41,28 @@ def test_availability_policy_expired(self):
...
+        """ Products are always available to admin even if the current
+        date is beyond the product's expiration date.
+        """
+        self.assert_expired_product_availability(True, availability.Available)
+
+    def test_availability_policy_expired_for_students(self):
+        """ A product is unavailable for students if the current date
+        is beyond the product's expiration date.
+        """
+        self.assert_expired_product_availability(False, availability.Unavailable)
```
rlucioni: You can use ddt to fold these test cases into one.

clintonb: I disagree. `ddt` hides the explicit cases for which we are testing. Tests are one place where I think verbosity is good to ensure that what is being tested is properly communicated.

ecommerce/extensions/partner/strategy.py
```diff
@@ -32,4 +38,6 @@ class DefaultStrategy(strategy.UseFirstStockRecord, CourseSeatAvailabilityPolicy
 
 class Selector(object):
     def strategy(self, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
+        if request and hasattr(request, 'user'):
+            return DefaultStrategy(request)
         return DefaultStrategy()
```
rlucioni: @awais786 you can fold this into a single line:

``` python
return DefaultStrategy(request if hasattr(request, 'user') else None)
```

ecommerce/extensions/partner/strategy.py
```diff
@@ -18,8 +18,14 @@ def seat_class(self):
...
 
     def availability_policy(self, product, stockrecord):
-        """ A seat is unavailable if the current date is beyond the seat's expiration date. """
-        if product.expires and timezone.now() > product.expires:
+        """ A product is unavailable for non-admin users if the current date is
+        beyond the product's expiration date. Products are always available for admin users.
+        """
+
+        # A seat will be available only If the request is valid and it has a staff user object.
+        if self.request and self.request.user.is_staff:
```
rlucioni: You can take advantage of the work Oscar's [base strategy class](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/partner/strategy.py#L51-L55) does for you by accessing `self.user`:

``` python
if getattr(self.user, 'is_staff', False):
```

awais786:
@rlucioni please check this PR.

awais786:
@rlucioni it's ready.

ecommerce/extensions/api/serializers.py
```diff
@@ -70,8 +70,9 @@ def get_price(self, product):
         return None
 
     def _get_info(self, product):
-        info = Selector().strategy().fetch_for_product(product)
-        return info
+        return Selector().strategy(
+            request=self.context.get('request', None)
```
rlucioni: `get()` returns `None` by default if the provided key is not found; you don't need to provide it explicitly.

ecommerce/extensions/partner/strategy.py
```diff
@@ -18,8 +18,13 @@ def seat_class(self):
...
-        """ A seat is unavailable if the current date is beyond the seat's expiration date. """
-        if product.expires and timezone.now() > product.expires:
+        """ A product is unavailable for non-admin users if the current date is
+        beyond the product's expiration date. Products are always available for admin users.
+        """
+
+        if getattr(self.user, 'is_staff', False):
+            return super(CourseSeatAvailabilityPolicyMixin, self).availability_policy(product, stockrecord)
+        elif product.expires and timezone.now() > product.expires:
             return availability.Unavailable()
```
rlucioni, line 30: You can tighten this up by reversing the inequality:

``` python
if getattr(self.user, 'is_staff', False) or (product.expires and timezone.now() <= product.expires):
    return super(CourseSeatAvailabilityPolicyMixin, self).availability_policy(product, stockrecord)
else:
    return availability.Unavailable()
```

This allows you to delete L30.

awais786, line 30: @rlucioni we need to allow products without expires also here.

```
if getattr(self.user, 'is_staff', False) or not product.expires or (
    product.expires and timezone.now() <= product.expires):
    return super(CourseSeatAvailabilityPolicyMixin, self).availability_policy(product, stockrecord)
else:
    return availability.Unavailable()
```

rlucioni, line 30: @awais786 good point. In that case, I'd abstract that conditional to a couple of variables:

``` python
is_staff = getattr(self.user, 'is_staff', False)
is_available = product.expires is None or (product.expires >= timezone.now())
if is_staff or is_available:
    return super(CourseSeatAvailabilityPolicyMixin, self).availability_policy(product, stockrecord)
else:
    return availability.Unavailable()
```

awais786:
@rlucioni it's ready.

rlucioni:
:+1:

PR #304, opened 2015-08-28
wait longer for alert modal in acceptance test

jimabramson:
`ecommerce.extensions.dashboard.orders.tests.OrderListViewTests.test_retry_fulfillment` fails consistently on my local system without a longer wait.   i don't have the same problem with any other tests.

@rlucioni any objections?


rlucioni:
@jimabramson none, :shipit: .

PR #305, opened 2015-08-28
avoid exception when `paypal_profile list` has no results

jimabramson:
not bothering with a deep investigation of the internal paypal sdk behavior here; the frequency of this use case is extremely low.

@rlucioni FYI

### DIFF
--- a/ecommerce/extensions/payment/management/commands/paypal_profile.py
+++ b/ecommerce/extensions/payment/management/commands/paypal_profile.py
@@ -106,7 +106,13 @@ class Command(BaseCommand):
     def handle_list(self, args):  # pylint: disable=unused-argument
         """Wrapper for paypalrestsdk List operation."""
         profiles = WebProfile.all()
-        self.print_json([profile.to_dict() for profile in profiles])
+        result = []
+        try:
+            result = [profile.to_dict() for profile in profiles]
+        except KeyError:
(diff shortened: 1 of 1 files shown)


jimabramson:
XCOM-471

ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -106,7 +106,13 @@ def _do_enable(self, profile_id, profile_name):
     def handle_list(self, args):  # pylint: disable=unused-argument
         """Wrapper for paypalrestsdk List operation."""
         profiles = WebProfile.all()
-        self.print_json([profile.to_dict() for profile in profiles])
+        result = []
+        try:
+            result = [profile.to_dict() for profile in profiles]
+        except KeyError:
+            # weird internal paypal sdk behavior; it means the result was empty.
+            pass
```
rlucioni, line 114: Would it be worth at least logging an error message?

jimabramson, line 114: i don't think so - `profiles` iterates just fine if it's got at least one item in it; i basically consider this a quirk on paypal's side and not ours, so there's nothing to worry the end user about with an error message.

rlucioni:
:+1:

PR #306, opened 2015-08-31
Awais786/xcom 595 update ecom api

awais786:
Verify that a helpful error message is displayed (e.g., "Course does not exist") when the Commerce API responds in case of non-existent courses.


ecommerce/courses/publishers.py
```diff
@@ -68,13 +70,31 @@ def publish(self, course):
...
             status_code = response.status_code
+
             if status_code in (200, 201):
                 logger.info(u'Successfully published commerce data for [%s].', course_id)
-                return True
+                return True, None
             else:
-                logger.error(u'Failed to publish commerce data for [%s] to LMS. Status was [%d]. Body was [%s].',
-                             course_id, status_code, response.content)
+                data = json.loads(response.content)
```
clintonb, line 146: This needs to be wrapped in a `try/except` block in case the content is not JSON.

ecommerce/courses/publishers.py
```diff
@@ -68,13 +70,31 @@ def publish(self, course):
...
+                if 'id' in data:
+                    error_message = data['id'][0]
```
clintonb: Why is this an array?

ecommerce/courses/publishers.py
```diff
@@ -68,13 +70,31 @@ def publish(self, course):
...
+                            status_code=status_code,
+                            response=response.content
+                        )
+                logger.error(error_message)
+
         except Exception:  # pylint: disable=broad-except
-            logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
 
-        return False
+            error_message = (
```
clintonb: It would be nice to include the exception's message when displaying to the user.

ecommerce/extensions/api/serializers.py
```diff
@@ -236,24 +236,16 @@ def save(self):
...
-                published = course.publish_to_lms()
+                published, resp_message = course.publish_to_lms()
                 if published:
                     return created, None, None
                 else:
-                    message = (
-                        u'An error occurred while publishing [{course_id}] to LMS. '
-                        u'No data has been saved or published.'
-                    ).format(course_id=course_id)
+                    raise Exception(resp_message)
```
clintonb: Use a custom exception class.

clintonb:
Why are there no new tests for this new functionality?

awais786:
@clintonb the PR was in internal review. I am working on test cases also.

ecommerce/courses/publishers.py
```diff
@@ -68,13 +70,31 @@ def publish(self, course):
...
+                u'Failed to publish commerce data for [{course_id}] to LMS.').format(
```
zubair-arbi: nit: `u'Failed to publish commerce data for course '{course_id}' to LMS.').format(`

ecommerce/courses/publishers.py
zubair-arbi, line 146: @awais786 I think you will have to refactor the code a bit so that the `try catch` block will have just `put` `request` and move the `if else` code outside of the `try catch`

clintonb:
Please prefix internal/work-in-progress reviews with "WIP" to avoid future confusion.

awais786:
WIP (internal/work-in-progress)

awais786:
@rlucioni , @clintonb Please review this PR.

ecommerce/extensions/api/exceptions.py
```diff
@@ -29,3 +29,8 @@ class ProductNotFoundError(ApiError):
 
 class BadRequestException(APIException):
     status_code = status.HTTP_400_BAD_REQUEST
+
+
+class LMSPublisherException(ApiError):
+    """Raised when the publish course to LMS fails."""
```
clintonb: Raised when LMSPublisher fails to publish a course.

ecommerce/extensions/api/serializers.py
```diff
@@ -241,24 +242,15 @@ def save(self):
...
                 else:
-                    message = (
-                        u'An error occurred while publishing [{course_id}] to LMS. '
-                        u'No data has been saved or published.'
-                    ).format(course_id=course_id)
+                    raise LMSPublisherException(resp_message)
 
-                    raise Exception(message)
-        except Exception as e:  # pylint: disable=broad-except
+        except (Exception, LMSPublisherException) as e:  # pylint: disable=broad-except
```
clintonb: `Exception` will also cover `LMSPublisherException`.

ecommerce/extensions/api/v2/tests/views/test_publication.py
```diff
@@ -215,7 +215,7 @@ def test_create(self):
             self._assert_course_saved(self.course_id)
 
             # If publication succeeds, the view should return a 201 and data should be saved.
-            mock_publish.return_value = True
```
clintonb, line 218: You also need to test publication failure, and ensure the error message is conveyed via API.

ecommerce/courses/publishers.py
```diff
@@ -95,22 +97,32 @@ def publish(self, course, access_token=None):
...
                         logger.info(u'Successfully published CreditCourse for [%s] to LMS.', course_id)
                     else:
-                        logger.error(
-                            u'Failed to publish CreditCourse for [%s] to LMS. Status was [%d]. Body was [%s].',
-                            course_id,
-                            response.status_code,
-                            response.content
+                        error_message = (
+                            u'Failed to publish CreditCourse for [{course_id}] to LMS. '
+                            u'Status was [{status}]. Body was [{body}].'.format(
```
rlucioni: @awais786 I'm not comfortable with this approach to implementing error messages.

Messages like these, originally intended as developer-facing log messages, will not be helpful to users of the course admin tool. We want to help our users by giving them a brief, plain-English explanation of why publication failed (e.g., "Publication of course data to the LMS failed. The course you're attempting to publish does not exist on the LMS.").

You shouldn't be hijacking these log messages and turning them into user messages. Does the Commerce API return a useful status code when a course does not exist on the LMS? If it were to return a 410, you could use that information to return a simple, high-level message explaining why publication failed.

awais786: In case of course does not exist LMS  raises serializers.ValidationError which returns 400 as status code also with error message. 'Course {course_id} does not exist.'

rlucioni: @awais786 consider modifying the LMS to return a 404 or a 410 so that you can return a more intelligible user message. Alternatively, consider employing the error message returned by the LMS as a user message here.

awais786:
@rlucioni I removed all my modified code and try to return only generic messages.

ecommerce/courses/publishers.py
awais786: @rlucioni I have added new subtask for this purpose.
https://openedx.atlassian.net/browse/XCOM-598

ecommerce/extensions/api/exceptions.py
```diff
@@ -29,3 +29,8 @@ class ProductNotFoundError(ApiError):
 
 class BadRequestException(APIException):
     status_code = status.HTTP_400_BAD_REQUEST
+
+
+class LMSPublisherException(ApiError):
```
jimabramson: minor: suggest renaming to `LMSPublisherError` since we are based on `ApiError` and not `APIException`.

jimabramson:
@awais786 a cleaner alternative might be, instead of returning a tuple from `LMSPublisher.publish`, to simply return None when it succeeds and the error message when some failure has occurred.  Then callers could use it like this

```
error_message = publisher.publish()
succeeded = (error_message is None)
...
```

ecommerce/courses/publishers.py
```diff
@@ -78,11 +78,16 @@ def publish(self, course, access_token=None):
             True, if publish operation succeeded; otherwise, False.
```
jimabramson: this docstring needs to be updated for the new return value.

ecommerce/courses/publishers.py
```diff
@@ -101,16 +106,17 @@ def publish(self, course, access_token=None):
...
                             response.content
                         )
-                        return False
-                except:  # pylint: disable=bare-except
+                        return False, error_message
+                except Exception as e:  # pylint: disable=bare-except
                     logger.exception(u'Failed to publish CreditCourse for [%s] to LMS.', course_id)
-                    return False
+                    error_message = e.message
+                    return False, error_message
```
jimabramson: no need for the `error_message` local var, can just return `e.message`

jimabramson:
messages intended for display to a user should be translated.

PR #307, opened 2015-08-31
Support publication of Courses with a credit seat

rlucioni:
Creates or updates a CreditCourse object on the LMS when publishing a Course with a credit seat. XCOM-585.

@clintonb and/or @jimabramson, please review.

### DIFF
--- a/ecommerce/courses/publishers.py
+++ b/ecommerce/courses/publishers.py
@@ -5,11 +5,15 @@ from django.conf import settings
 import requests
 
 from ecommerce.courses.utils import mode_for_seat
+from ecommerce.settings.base import get_lms_url
+
 
 logger = logging.getLogger(__name__)
 
 
 class LMSPublisher(object):
+    timeout = settings.COMMERCE_API_TIMEOUT
+
     def get_seat_expiration(self, seat):
         if not seat.expires or 'professional' in getattr(seat.attr, 'certificate_type', ''):
             return None
@@ -30,13 +34,45 @@ class LMSPublisher(object):
             'expires': self.get_seat_expiration(seat),
         }
 
-    def publish(self, course):
+    def _publish_creditcourse(self, course_id, access_token):
+        """Creates or updates a CreditCourse object on the LMS."""
+        url = get_lms_url('api/credit/v1/courses/')
+
+        data = {
+            'course_key': course_id,
+            'enabled': True
+        }
+
+        headers = {
+            'Content-Type': 'application/json',
+            'Authorization': 'Bearer ' + access_token
+        }
+
+        kwargs = {
+            'url': url,
+            'data': json.dumps(data),
+            'headers': headers,
+            'timeout': self.timeout
+        }
+        response = requests.post(**kwargs)
+        if response.status_code == 400:
+            # The CreditCourse already exists. Try updating it.
+            kwargs['url'] += course_id.strip('/') + '/'
+            response = requests.put(**kwargs)
+
+        return response
+
+    def publish(self, course, access_token=None):
         """ Publish course commerce data to LMS.
 
-        Uses the Commerce API to publish course modes, prices, and SKUs to LMS.
+        Uses the Commerce API to publish course modes, prices, and SKUs to LMS. Uses
+        CreditCourse API endpoints to publish CreditCourse data to LMS when necessary.
 
-        Args:
+        Arguments:
             course (Course): Course to be published.
+
+        Keyword Arguments:
+            access_token (str): Access token used when publishing CreditCourse data to the LMS.
 
         Returns:
             True, if publish operation succeeded; otherwise, False.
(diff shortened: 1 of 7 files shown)


ecommerce/courses/publishers.py
```diff
@@ -50,6 +87,36 @@ def publish(self, course):
...
         verification_deadline = self.get_course_verification_deadline(course)
         modes = [self.serialize_seat_for_commerce_api(seat) for seat in course.seat_products]
+
+        has_credit = 'credit' in [mode['name'] for mode in modes]
+        if has_credit:
+            if access_token is not None:
+                try:
+                    response = self._publish_creditcourse(course_id, access_token)
+
+                    if response.status_code == 400:
```
clintonb: Move this logic into `_publish_creditcourse`.

ecommerce/courses/tests/test_publishers.py
```diff
@@ -37,6 +38,31 @@ def _mock_commerce_api(self, status, body=None):
         httpretty.register_uri(httpretty.PUT, url, status=status, body=json.dumps(body),
                                content_type=JSON)
 
+    def _mock_credit_api(self, creation_status, update_status):
+        self.assertTrue(httpretty.is_enabled, 'httpretty must be enabled to mock Credit API calls.')
+
+        url = '{host}/{path}'.format(
```
clintonb: Use `get_lms_url`.

ecommerce/courses/publishers.py
```diff
@@ -30,14 +32,49 @@ def serialize_seat_for_commerce_api(self, seat):
...
+        }
+
+        kwargs = {
+            'url': url,
+            'data': json.dumps(data),
+            'headers': headers,
+            'timeout': self.timeout
+        }
+        if update:
+            kwargs['url'] += course_id
```
clintonb: Missing the trailing slash.

ecommerce/courses/tests/test_publishers.py
```diff
@@ -139,3 +165,59 @@ def test_serialize_seat_for_commerce_api_with_professional(self, is_verified, ex
...
+
+    @httpretty.activate
+    @ddt.data(
+        ('access_token', 201, None, 201, True),
+        ('access_token', 400, 200, 200, True),
+        ('access_token', 400, 418, 200, False),
+        (None, None, None, None, False)
+    )
+    @ddt.unpack
+    def test_credit_publication(self, access_token, creation_status, update_status, commerce_status, success):
```
clintonb: There is a lot of magic going on here. This test suite would be easier to understand if this test was broken into multiple tests that validate the explicit error states.

ecommerce/extensions/api/v2/views.py
```diff
@@ -538,6 +539,11 @@ class AtomicPublicationView(generics.CreateAPIView, generics.UpdateAPIView):
     permission_classes = (IsAuthenticated, IsAdminUser,)
     serializer_class = serializers.AtomicPublicationSerializer
 
+    def get_serializer_context(self):
+        context = super(AtomicPublicationView, self).get_serializer_context()
+        context['access_token'] = getattr(self.request.user, 'access_token', None)
```
clintonb: Why not use `request.user.access_token`, which returns `None` if no token exists?

rlucioni: I didn't realize the property returns `None` if no token exists. Updated.

ecommerce/courses/publishers.py
rlucioni: Moved.

ecommerce/courses/tests/test_publishers.py
rlucioni: Updated.

ecommerce/courses/publishers.py
rlucioni: Added.

ecommerce/courses/publishers.py
```diff
@@ -30,14 +32,49 @@ def serialize_seat_for_commerce_api(self, seat):
...
-        Uses the Commerce API to publish course modes, prices, and SKUs to LMS.
+        Uses the Commerce API to publish course modes, prices, and SKUs to LMS. Uses
+        CreditCourse API endpoints to publish CreditCourse data to LMS when necessary.
 
-        Args:
+        Arguments:
             course (Course): Course to be published.
 
+        Keyword Arguments:
+            access_token (str): Access token used when publishing CreditCourse data to the LMS.
```
jimabramson, line 75: why is access_token required for CreditCourse publishing but not for other types / modes?

rlucioni, line 75: The Credit API uses OAuth2 for authentication, not an API key, like the Commerce API. Only publication of Courses with a credit seat requires making calls to the Credit API.

ecommerce/courses/tests/test_publishers.py
rlucioni: @clintonb broken up.

ecommerce/courses/publishers.py
clintonb, line 75: Correct. I have a preference toward OAuth2 and any other authentication scheme that explicitly associates a request with a user. Our server-to-server authentication does not provide this, meaning we have to create minor hacks to deal with permissions.

clintonb:
:+1:

jimabramson:
:+1:

PR #308, opened 2015-08-31
Running migrations for Travis builds

clintonb:
We run migrations to ensure no migrations are missing, and they work on fresh installs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
     - pip install coveralls
     - make requirements.js
 script:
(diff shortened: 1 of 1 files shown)


clintonb:
@jimabramson @rlucioni @zubair-arbi

clintonb:
Build will pass once #347 is merged and this branch rebased.

rlucioni:
:+1: Reminds me that we should get rid of the `DisableMigrations` hack.

PR #309, opened 2015-09-01
Add setup.py

rlucioni:
In preparation for asynchronous order fulfillment. Allows ecommerce to be installed on multitenant workers.

@jimabramson @clintonb

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,32 @@
+"""
+This setup module is only intended for use when installing Celery tasks, NOT the Django application itself.
+Any interaction with the application should rely solely on the HTTP-based RESTful APIs exposed by the application.
+"""
+from setuptools import setup
+
+
+with open('worker/README.rst') as readme, open('AUTHORS') as authors:
+    long_description = '{}\n\n{}'.format(readme.read(), authors.read())
+
+setup(
+    name='edx-ecommerce-tasks',
+    version='0.1.0',
+    description='Celery tasks supporting the operations of edX\'s ecommerce service',
+    long_description=long_description,
+    classifiers=[
+        'Development Status :: 1 - Planning',
+        'License :: OSI Approved :: GNU Affero General Public License v3',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 2.7',
+        'Topic :: Internet',
+        'Intended Audience :: Developers',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -0,0 +1,28 @@
...
+        'Topic :: Internet',
+        'Intended Audience :: Developers',
+        'Environment :: Web Environment',
+    ],
+    keywords='edx ecommerce',
+    url='https://github.com/edx/ecommerce',
+    author='edX',
+    author_email='oscm@edx.org',
+    license='AGPL',
+    packages=find_packages(exclude=['acceptance_tests', '*.tests']),
```
clintonb: Only export the tasks sub-package.

clintonb:
:+1: (once the package is updated)

setup.py
```diff
@@ -0,0 +1,28 @@
+from setuptools import setup, find_packages
```
jimabramson: i suggest a module level docstring with a brief note explaining that this setup is only used for installing tasks and not the django application.

setup.py
jimabramson: @clintonb tasks may wish to make use of things like constants and other (stateless) utilities in the main webapp repo.  I think we would have to move those things into a third, shared package if we want to not install the webapp, which is a bigger change than we might want to make right now.

clintonb: We don't have to move those dependencies out now, but we should do so to avoid the possibility of mistakenly calling non-task code from the workers. I realize I may be a bit pedantic; but, I want to make it hard for future developers to screw up.

rlucioni: @clintonb acknowledged and agreed. Moving those dependencies seems like work that should happen when writing the tasks themselves. I can make a TODO note here indicating that the exported packages should only be the future worker and utilities packages.

jimabramson: @rlucioni on that score it seems like it would be correct to blacklist all but the worker code now.  (if that's even possible, because that will exclude basically everything.)

rlucioni: Updated.

setup.py
rlucioni: Added.

rlucioni:
@jimabramson @clintonb revised in light of our discussion.

jimabramson:
:+1:

PR #312, opened 2015-09-02
Enable Email Sending

ahsan-ul-haq:
ECOM-2078

@zubair-arbi @clintonb @feanil please review

### DIFF
--- a/ecommerce/settings/production.py
+++ b/ecommerce/settings/production.py
@@ -15,6 +15,9 @@ from ecommerce.settings.logger import get_logger_config
 COMPRESS_ENABLED = True
 COMPRESS_OFFLINE = True
 
+# Email configuration
+EMAIL_BACKEND = 'django_ses.SESBackend'
+
 # Minify CSS
 COMPRESS_CSS_FILTERS += [
     'compressor.filters.cssmin.CSSMinFilter',
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -1,6 +1,7 @@
 # Packages required in a production environment
 -r base.txt
 
+django-ses==0.7.0
 MySQL-python==1.2.5
 PyYAML==3.11
 gunicorn==19.2.1


ecommerce/extensions/checkout/signals.py
```diff
@@ -62,31 +62,30 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
 @log_exceptions("Failed to send order completion email.")
 def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable=unused-argument
     """Send course purchase notification email when a course is purchased."""
-    if waffle.switch_is_active('ENABLE_NOTIFICATIONS'):
```
clintonb, line 65: What was this removed? During grooming we stated that no _new_ switches would be added. The existing switch should remain.

requirements/base.txt
```diff
@@ -1,10 +1,12 @@
 analytics-python==1.0.3
+boto==2.38.0
```
clintonb: Unless you are calling `boto` directly, it should not be listed here. `pip` will handle finding the `django-ses` dependencies.

ecommerce/settings/local.py
```diff
@@ -17,8 +17,7 @@
 # END DEBUG CONFIGURATION
 
 # EMAIL CONFIGURATION
-# See: https://docs.djangoproject.com/en/dev/ref/settings/#email-backend
-EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
+EMAIL_BACKEND = 'django_ses.SESBackend'
```
clintonb: SES is not needed locally. It's needed for production.

requirements/base.txt
```diff
@@ -5,6 +5,7 @@ django-compressor==1.5
 django_extensions==1.5.5
 django-libsass==0.3
 django-oscar==1.1
+django-ses==0.7.0
```
feanil: Shouldn't this go into the production requirements.txt file at `requirements/production.txt`?

ahsan-ul-haq:
@feanil @clintonb I have checked email sending locally by setting up django-ses. Following is the screen shot of email sent.

screen shot 2015-09-07 at 6 06 56 pm

clintonb:
:+1:

Please coordinate with @feanil to ensure the sender's address (billing@edx.org) is verified and properly set in configuration. Nice work!

feanil:
Is the sender e-mail address hard coded?  Can we pull it out into configuration in case we want to change it?

clintonb:
The address is a setting that needs to be added to configuration.

## 

Clinton Blackburn

ahsan-ul-haq:
@feanil Default sender email address needed to be added in the settings as `OSCAR_FROM_EMAIL`. Currently it used default value which is `oscar@example.com`. I wasn't sure which email address to be added so once you guys let me know I will add that as well.

FYI: @clintonb

feanil:
@clintonb @ahsan-ul-haq That seems fine as a default, let's assume we'll override it in config_secure.  Please open a ticket for adding this change to configuration_secure.  Please mention that the e-mail addresses should be verified by SES as well.

clintonb:
@ahsan-ul-haq https://github.com/ahsan-ul-haq please open a ticket,
requesting the address be set to billing@edx.org.

## 

Clinton Blackburn

ahsan-ul-haq:
@clintonb @feanil I have opened a ticket [DEVOPS-2784](https://openedx.atlassian.net/browse/DEVOPS-2784). So should I merge this one?

clintonb:
@ahsan-ul-haq merge away.

PR #314, opened 2015-09-02
Updated Settings

clintonb:
- Cleaned template settings, taking advantage of Django 1.8 changes
- Removed unnecessary permissions scope from OIDC login requests
- Moved get_lms_url to ecommerce.settings module

ECOM-2178

@jimabramson @rlucioni


rlucioni:
:+1:

PR #317, opened 2015-09-04
Updated BasketAdmin

clintonb:
Added read-only inline view of associated PaymentProcessor objects.

@jimabramson @rlucioni @zubair-arbi @awais786 @ahsan-ul-haq @tasawernawaz @aamir-khan


jimabramson:
:+1:

PR #319, opened 2015-09-05
Updated translations

clintonb:
FYI @jimabramson @rlucioni


rlucioni:
@clintonb is the ecommerce project listed on Transifex? If not, it's unclear to me what this PR has added.

clintonb:
Yes, the project is on Transifex. This PR added the updated translations. They must be periodically updated.

rlucioni:
@clintonb nice. It might be worth adding something to the release process notes indicating how and when in the process to update translations.

PR #320, opened 2015-09-07
mark strings in javascript files for translation

zubair-arbi:
XCOM-343
@awais786 @aamir-khan @ahsan-ul-haq @tasawernawaz 
- Verify that all strings presented to the user via JavaScript are marked for translation.
- Include django javascript translation catalog `jsi18n` file in templates which extend from `'dashboard/layout.html'` instead of local `'edx/base.html'` (which already have this file).

### DIFF
--- a/ecommerce/static/vendor-extensions/oscar/js/refund_list.js
+++ b/ecommerce/static/vendor-extensions/oscar/js/refund_list.js
@@ -3,7 +3,8 @@ $(document).ready(function () {
     var processRefund = function (e) {
         var $btn = $(e.target),
             refund_id = $btn.data('refund-id'),
-            decision = $btn.data('decision');
+            decision = $btn.data('decision'),
+            message = '';
 
         // Disable button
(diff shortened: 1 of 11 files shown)


clintonb:
:+1:

zubair-arbi:
@rlucioni Please review.

ecommerce/static/vendor-extensions/oscar/js/refund_list.js
```diff
@@ -27,10 +32,15 @@ $(document).ready(function () {
             // refund, even when the refund has entered an error state during processing.
             $('tr[data-refund-id=' + refund_id + '] .refund-status').text('Error');
```
rlucioni: This "Error" message must be marked for translation.

zubair-arbi:
@rlucioni Please review again.

rlucioni:
:+1: once commits are squashed and tests pass.

PR #321, opened 2015-09-08
Awais786/xcom 595 update ecom api

awais786:
**Old related PR:** https://github.com/edx/ecommerce/pull/306 _(This PR was closed in favor of better implementation approach)_

This PR follows the new implementation approach as suggested by @jimabramson.
- Return the error message instead of a tuple `(boolean status [True, False], message)` as response in case of an `exception`, otherwise just return `None` in case of success. A sample usage:

```
error_message = publisher.publish()
succeeded = (error_message is None)
```

> @rlucioni  LMS is sending the error messages so I think we don't need to send the status codes.

### DIFF
--- a/ecommerce/courses/publishers.py
+++ b/ecommerce/courses/publishers.py
@@ -1,8 +1,9 @@
 import json
 import logging
+import requests
 
 from django.conf import settings
-import requests
+from django.utils.translation import ugettext_lazy as _
 
 from ecommerce.courses.utils import mode_for_seat
 from ecommerce.settings import get_lms_url
@@ -75,14 +76,18 @@ class LMSPublisher(object):
             access_token (str): Access token used when publishing CreditCourse data to the LMS.
 
         Returns:
-            True, if publish operation succeeded; otherwise, False.
+            None, if publish operation succeeded; otherwise, error message.
         """
+
+        course_id = course.id
+        error_message = _(u'Failed to publish commerce data for {course_id} to LMS.').format(
+            course_id=course_id
+        )
 
         if not settings.COMMERCE_API_URL:
             logger.error('COMMERCE_API_URL is not set. Commerce data will not be published!')
-            return False
+            return error_message
 
-        course_id = course.id
         name = course.name
         verification_deadline = self.get_course_verification_deadline(course)
         modes = [self.serialize_seat_for_commerce_api(seat) for seat in course.seat_products]
@@ -101,16 +106,17 @@ class LMSPublisher(object):
                             response.status_code,
                             response.content
                         )
-                        return False
-                except:  # pylint: disable=bare-except
+
+                        return self._parse_error(response, error_message)
+                except Exception:  # pylint: disable=broad-except
                     logger.exception(u'Failed to publish CreditCourse for [%s] to LMS.', course_id)
-                    return False
+                    return error_message
             else:
                 logger.error(
                     u'Unable to publish CreditCourse for [%s] to LMS. No access token available.',
                     course_id
                 )
-                return False
+                return error_message
 
         data = {
             'id': course_id,
@@ -131,11 +137,45 @@ class LMSPublisher(object):
             status_code = response.status_code
             if status_code in (200, 201):
                 logger.info(u'Successfully published commerce data for [%s].', course_id)
-                return True
+                return
             else:
                 logger.error(u'Failed to publish commerce data for [%s] to LMS. Status was [%d]. Body was [%s].',
                              course_id, status_code, response.content)
+
+                return self._parse_error(response, error_message)
+
         except Exception:  # pylint: disable=broad-except
             logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
 
-        return False
+        return error_message
+
+    def _parse_error(self, response, default_error_message):
+        """When validation errors occur during publication, the LMS is expected
+         to return an error message.
+
+        Arguments:
+            response (Response): A 'Response' object which contains json error message.
+            default_error_message (str) : default error message in case of exception.
+
+        Returns:
+            string: Returns the error message extracted from response.content
+            along with default message. If no message is available in response
+            then default message will be return.
+
+        """
+        message = None
+        try:
+            data = response.json()
+            if isinstance(data, basestring):
+                message = data
+            elif isinstance(data, dict) and len(data) > 0:
+                message = data.values()[0]
+            if isinstance(message, list):
+                message = message[0]
+        except Exception:  # pylint: disable=broad-except
+            pass
+
+        if message:
+            return ' '.join([default_error_message, message])
+        else:
+            return default_error_message
--- a/ecommerce/courses/tests/test_publishers.py
+++ b/ecommerce/courses/tests/test_publishers.py
@@ -30,6 +30,9 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
         self.course.create_or_update_seat('honor', False, 0)
         self.course.create_or_update_seat('verified', True, 50)
         self.publisher = LMSPublisher()
+        self.error_message = u'Failed to publish commerce data for {course_id} to LMS.'.format(
+            course_id=self.course.id
+        )
 
     def _mock_commerce_api(self, status, body=None):
         self.assertTrue(httpretty.is_enabled, 'httpretty must be enabled to mock Commerce API calls.')
@@ -39,7 +42,7 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
         httpretty.register_uri(httpretty.PUT, url, status=status, body=json.dumps(body),
                                content_type=JSON)
 
-    def _mock_credit_api(self, creation_status, update_status):
+    def _mock_credit_api(self, creation_status, update_status, body=None):
         self.assertTrue(httpretty.is_enabled, 'httpretty must be enabled to mock Credit API calls.')
 
         url = get_lms_url('api/credit/v1/courses/')
@@ -47,7 +50,7 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
             httpretty.POST,
             url,
             status=creation_status,
-            body='{}',
+            body=json.dumps(body),
             content_type=JSON
         )
 
@@ -57,7 +60,7 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
                 httpretty.PUT,
                 url,
                 status=update_status,
-                body='{}',
+                body=json.dumps(body),
                 content_type=JSON
             )
 
@@ -66,29 +69,55 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
         """ If the Commerce API is not setup, the method should log an INFO message and return """
         with override_settings(COMMERCE_API_URL=setting_value):
             with LogCapture(LOGGER_NAME) as l:
-                self.publisher.publish(self.course)
+                response = self.publisher.publish(self.course)
                 l.check((LOGGER_NAME, 'ERROR', 'COMMERCE_API_URL is not set. Commerce data will not be published!'))
+                self.assertIsNotNone(response)
+                self.assertEqual(
+                    response,
+                    self.error_message.format(
+                        course_id=self.course.id
+                    )
+                )
 
     def test_api_exception(self):
         """ If an exception is raised when communicating with the Commerce API, an ERROR message should be logged. """
-        with mock.patch('requests.put', side_effect=Timeout):
+        error = 'time out error'
+        with mock.patch('requests.put', side_effect=Timeout(error)):
             with LogCapture(LOGGER_NAME) as l:
-                self.publisher.publish(self.course)
+                response = self.publisher.publish(self.course)
                 l.check(
-                    (LOGGER_NAME, 'ERROR', 'Failed to publish commerce data for [{}] to LMS.'.format(self.course.id)))
+                    (
+                        LOGGER_NAME, 'ERROR',
+                        u'Failed to publish commerce data for [{course_id}] to LMS.'.format(
+                            course_id=self.course.id
+                        )
+                    )
+                )
+                self.assertIsNotNone(response)
+                self.assertEqual(self.error_message, response)
 
     @httpretty.activate
-    @ddt.data(401, 403, 404, 500)
-    def test_api_bad_status(self, status):
+    @ddt.unpack
+    @ddt.data(
+        (400, {'non_field_errors': ['deadline issue']}, 'deadline issue'),
+        (404, 'page not found', 'page not found'),
+        (401, {'detail': 'Authentication'}, 'Authentication'),
+        (401, {}, ''),
+    )
+    def test_api_bad_status(self, status, error_msg, expected_msg):
         """ If the Commerce API returns a non-successful status, an ERROR message should be logged. """
-        body = {u'error': u'Testing!'}
-        self._mock_commerce_api(status, body)
-
+        self._mock_commerce_api(status, error_msg)
         with LogCapture(LOGGER_NAME) as l:
-            self.publisher.publish(self.course)
-            l.check((LOGGER_NAME, 'ERROR',
-                     u'Failed to publish commerce data for [{}] to LMS. Status was [{}]. Body was [{}].'.format(
-                         self.course.id, status, json.dumps(body))))
+            response = self.publisher.publish(self.course)
+            l.check(
+                (
+                    LOGGER_NAME, 'ERROR',
+                    u'Failed to publish commerce data for [{}] to LMS. Status was [{}]. Body was [{}].'.format(
+                        self.course.id, status, json.dumps(error_msg))
+                )
+            )
+
+            self.assert_response_message(response, expected_msg)
 
     @httpretty.activate
     @ddt.data(200, 201)
@@ -97,7 +126,10 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
         self._mock_commerce_api(status)
 
         with LogCapture(LOGGER_NAME) as l:
-            self.publisher.publish(self.course)
+
+            response = self.publisher.publish(self.course)
+            self.assertIsNone(response)
+
             l.check((LOGGER_NAME, 'INFO', 'Successfully published commerce data for [{}].'.format(self.course.id)))
 
         last_request = httpretty.last_request()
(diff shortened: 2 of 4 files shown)


awais786:
https://github.com/edx/ecommerce/pull/306
Old PR

ecommerce/courses/publishers.py
```diff
@@ -101,16 +106,16 @@ def publish(self, course, access_token=None):
                             response.status_code,
                             response.content
                         )
-                        return False
-                except:  # pylint: disable=bare-except
+                        return response.content
+                except Exception as e:  # pylint: disable=broad-except
                     logger.exception(u'Failed to publish CreditCourse for [%s] to LMS.', course_id)
-                    return False
+                    return str(e.message)
```
aamir-khan: make it to `unicode(e.message)`

make sure we are returning unicode strings everywhere.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +136,13 @@ def publish(self, course, access_token=None):
             status_code = response.status_code
             if status_code in (200, 201):
                 logger.info(u'Successfully published commerce data for [%s].', course_id)
-                return True
+                return None
             else:
                 logger.error(u'Failed to publish commerce data for [%s] to LMS. Status was [%d]. Body was [%s].',
                              course_id, status_code, response.content)
-        except Exception:  # pylint: disable=broad-except
+                return response.content
```
awais786: response.content is coming in this format.

```
'{"non_field_errors": ["Verification deadline must be after the course mode upgrade deadlines."]}'
'{"id": ["Course arb/420/2015_t211 does not exist."]}'
```

It can be as extracted like this

```
data = json.loads(response.content)
err  = ",".join([v[0] for k, v in data.items()])
```

awais786: @jimabramson  what is your suggestion about this ? Do we need to extract or just show the response.content.

awais786:

awais786:
@rlucioni please review this PR.

ecommerce/courses/publishers.py
clintonb: I second this. You can also simply return `e.message` which should already be a Unicode value.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +136,13 @@ def publish(self, course, access_token=None):
             status_code = response.status_code
             if status_code in (200, 201):
                 logger.info(u'Successfully published commerce data for [%s].', course_id)
-                return True
+                return None
```
rlucioni: Simply `return` is sufficient.

ecommerce/courses/publishers.py
rlucioni: @awais786 parse a clean error message from the response content. A user won't want to see a Python dictionary as their error message. I'd suggest appending it to the existing `error_message`.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +136,13 @@ def publish(self, course, access_token=None):
...
             else:
                 logger.error(u'Failed to publish commerce data for [%s] to LMS. Status was [%d]. Body was [%s].',
                              course_id, status_code, response.content)
-        except Exception:  # pylint: disable=broad-except
+                return response.content
+        except Exception as e:  # pylint: disable=broad-except
             logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
+            error_message = str(e.message)
```
rlucioni: Either append `e.message` to the existing `error_message`, or simply return `error_message`. I would avoid returning `e.message` alone. It will be something oriented towards developers (e.g., "HTTPConnectionPool(host='github.com', port=80): Request timed out. (timeout=1)") and will most likely not be helpful to a user.

ecommerce/courses/tests/test_publishers.py
```diff
@@ -181,8 +199,8 @@ def test_credit_publication_success(self, creation_status, update_status, commer
         self._mock_commerce_api(commerce_status)
 
         access_token = 'access_token'
-        published = self.publisher.publish(self.course, access_token=access_token)
-        self.assertTrue(published)
+        error_message = self.publisher.publish(self.course, access_token=access_token)
+        self.assertTrue((error_message is None))
```
rlucioni: Use `assertIsNone`.

ecommerce/extensions/api/serializers.py
```diff
@@ -242,24 +243,17 @@ def save(self):
...
+                published = (resp_message is None)
+
                 if published:
                     return created, None, None
                 else:
-                    message = (
-                        u'An error occurred while publishing [{course_id}] to LMS. '
-                        u'No data has been saved or published.'
-                    ).format(course_id=course_id)
+                    raise LMSPublisherError(resp_message)
```
rlucioni: I wouldn't bother with this custom exception. It's not being used by the `LMSPublisher`.

ecommerce/courses/publishers.py
jimabramson: agree with Renzo.  Since the status/response is unknown I would ensure that if the attempt to extract error messages fails (because it didn't return any JSON at all, for example), we fall back to something generic like "Unknown error" rather than triggering additional errors trying to generate the message.

ecommerce/extensions/api/serializers.py
awais786: It was suggested by @clintonb to use a custom exception class.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,28 @@ def publish(self, course, access_token=None):
...
         except Exception:  # pylint: disable=broad-except
             logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
 
-        return False
+        return error_message
+
+    def parse_response_content(self, default_error_message, response_content):
+        """In case of any validation error from api returns the error message.
+        Arguments:
+            content (str): A string contains json error message.
```
ahsan-ul-haq: Please add all arguments in the Doc String

ecommerce/courses/tests/test_publishers.py
```diff
@@ -97,7 +123,10 @@ def test_api_success(self, status):
         self._mock_commerce_api(status)
 
         with LogCapture(LOGGER_NAME) as l:
-            self.publisher.publish(self.course)
+
+            response = self.publisher.publish(self.course)
+            self.assertTrue((response is None))
```
ahsan-ul-haq: use `assertIsNone` instead

awais786:
@rlucioni please check this PR again.

ecommerce/extensions/api/serializers.py
rlucioni: It's no longer relevant. The exception's docstring implies that it should be raised by the LMSPublisher when publication fails. Since your implementation has changed, the exception is no longer being used in that way. If you insist on using a custom exception, use something like `AtomicPublicationException`. However, I don't see the benefit of a custom exception if all you're trying to do is bubble a message up to the caller.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,30 @@ def publish(self, course, access_token=None):
...
+
+    def parse_response_content(self, default_error_message, response_content):
+        """In case of any validation error from api returns the error message.
+        Arguments:
+            default_error_message (str) : default error message in case of exception.
+            response_content (str): A string contains json error message.
+        Returns:
+            string: response.content along with default error message.
+        """
+        data = json.loads(response_content)
```
rlucioni: This may fail if the response content does not contain valid JSON. You shouldn't need to use the `json` module to decode the response content, anyway. Pass the response to this helper function and call `response.json()`. Wrap it with an exception handler in case the response doesn't contain valid JSON.

rlucioni: In the event that the response doesn't contain valid JSON, return the default error message.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,30 @@ def publish(self, course, access_token=None):
...
+        if isinstance(data, dict):
+            message = "".join(data.values()[0])
```
rlucioni: Join the message contained in the response with the default message. This should read something like "Failed to publish commerce data for {course_id} to LMS. Course does not exist."

ecommerce/courses/publishers.py
```diff
@@ -75,14 +76,18 @@ def publish(self, course, access_token=None):
             access_token (str): Access token used when publishing CreditCourse data to the LMS.
 
         Returns:
-            True, if publish operation succeeded; otherwise, False.
+            None, if publish operation succeeded; otherwise, error message.
         """
 
+        course_id = course.id
+        error_message = _(u'Failed to publish commerce data for [{course_id}] to LMS.').format(
```
rlucioni: Remove the brackets around {course_id}; these should only be used for log messages.

ecommerce/courses/publishers.py
rlucioni: I see you're doing this below. The `join` here is not necessary.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,30 @@ def publish(self, course, access_token=None):
...
+            default_error_message (str) : default error message in case of exception.
+            response_content (str): A string contains json error message.
+        Returns:
+            string: response.content along with default error message.
+        """
+        data = json.loads(response_content)
+        if isinstance(data, dict):
+            message = "".join(data.values()[0])
+        else:
+            message = data
```
rlucioni: What if data is `None`? You don't want to be doing this in that case; the `join` with the base error message will fail. To be safe and avoid throwing additional errors when parsing this message, check if the data is a string here. If it's not, just return the base error message.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,34 @@ def publish(self, course, access_token=None):
...
+
         except Exception:  # pylint: disable=broad-except
             logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
 
-        return False
+        return error_message
+
+    def parse_response_content(self, default_error_message, response):
+        """In case of any validation error api will returns the error message as json.
+        Arguments:
```
zubair-arbi, line 156: nit: newline before the blocks `Arguments` and `Returns`

ecommerce/extensions/api/exceptions.py
```diff
@@ -29,3 +29,8 @@ class ProductNotFoundError(ApiError):
 
 class BadRequestException(APIException):
     status_code = status.HTTP_400_BAD_REQUEST
+
+
+class LMSPublisherError(ApiError):
```
zubair-arbi: I think the name `LMSPublishError` makes more sense than `LMSPublisherError` since the class `LMSPublisher` don't directly raise this exception.

ecommerce/courses/publishers.py
```diff
@@ -2,6 +2,7 @@
 import logging
 
 from django.conf import settings
+from django.utils.translation import ugettext_lazy as _
```
zubair-arbi, line 6: Please update the import order

ecommerce/extensions/api/exceptions.py
awais786: I removed this custom exception.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,36 @@ def publish(self, course, access_token=None):
...
+                return self.parse_response_content(error_message, response)
+
         except Exception:  # pylint: disable=broad-except
             logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
 
-        return False
+        return error_message
+
+    def parse_response_content(self, default_error_message, response):
+        """In case of any validation error api will returns the error message as json.
```
rlucioni: > When validation errors occur during publication, the LMS is expected to return an error message.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,36 @@ def publish(self, course, access_token=None):
...
+        """
+        message = ''
+        try:
+            data = response.json()
+            if isinstance(data, basestring):
+                message = data
+            elif isinstance(data, dict) and len(data) > 0:
+                message = "".join(data.values()[0])
+        except Exception:  # pylint: disable=broad-except
+            message = response.content
```
rlucioni: In this case, we do not want to return the response content at all, since we can't format it nicely for the user. Just return  the default error message.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,36 @@ def publish(self, course, access_token=None):
...
+
+        return " ".join([default_error_message, message])
```
rlucioni: When `message` is empty, we don't want an extra space at the end of the error message.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,36 @@ def publish(self, course, access_token=None):
...
+        Returns:
+            string: response.content along with default error message.
+        """
+        message = ''
+        try:
+            data = response.json()
+            if isinstance(data, basestring):
+                message = data
+            elif isinstance(data, dict) and len(data) > 0:
+                message = "".join(data.values()[0])
```
rlucioni: Nitpick: stick with single quotes.

awais786:
@rlucioni it's ready now.

zubair-arbi:
lgtm :+1:

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,37 @@ def publish(self, course, access_token=None):
...
+        Returns:
+            string: response.content along with default error message.
+        """
+        message = ''
+        try:
+            data = response.json()
+            if isinstance(data, basestring):
+                message = data
+            elif isinstance(data, dict) and len(data) > 0:
+                message = ''.join(data.values()[0])
```
rlucioni: Joining with the empty string doesn't seem necessary. Can you just let `message = data.values()[0]`?

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,37 @@ def publish(self, course, access_token=None):
...
+
         except Exception:  # pylint: disable=broad-except
             logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
 
-        return False
+        return error_message
+
+    def parse_response_content(self, default_error_message, response):
+        """When validation errors occur during publication, the LMS is expected
+         to return an error message..
```
rlucioni: Nitpick: Extra period at the end of this sentence.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,37 @@ def publish(self, course, access_token=None):
...
+
+        Arguments:
+            default_error_message (str) : default error message in case of exception.
+            response_content (str): A string contains json error message.
+
+        Returns:
+            string: response.content along with default error message.
```
rlucioni: Please update this to describe the function's current behavior.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,37 @@ def publish(self, course, access_token=None):
...
+        try:
+            data = response.json()
+            if isinstance(data, basestring):
+                message = data
+            elif isinstance(data, dict) and len(data) > 0:
+                message = ''.join(data.values()[0])
+        except Exception:  # pylint: disable=broad-except
+            return default_error_message
+
+        return ' '.join([default_error_message, message])
```
rlucioni: We don't want an extra space at the end of the error message if no message can be extracted from the response.

```
if message:
    return ' '.join([default_error_message, message])
else:
    return default_error_message
```

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,37 @@ def publish(self, course, access_token=None):
...
+
+                return self.parse_response_content(error_message, response)
+
         except Exception:  # pylint: disable=broad-except
             logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
 
-        return False
+        return error_message
+
+    def parse_response_content(self, default_error_message, response):
```
rlucioni: Nitpick: You could make this function's name more precise, to reflect what it's actually doing. For example, you might update the signature to `parse_error(response, default_error_message)`.

ecommerce/courses/publishers.py
rlucioni: Were you trying to account for the possibility that the value could be a list? If so, please be more explicit about that. The contents of `message` must be a string in order for the later `join` to work.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,44 @@ def publish(self, course, access_token=None):
...
+
+                return self.parse_error(response, error_message)
+
         except Exception:  # pylint: disable=broad-except
             logger.exception(u'Failed to publish commerce data for [%s] to LMS.', course_id)
 
-        return False
+        return error_message
+
+    def parse_error(self, response, default_error_message):
```
zubair-arbi: Since this method is used only for class `LMSPublisher` internally so I think it makes more sense to name this method `_parse_error`

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,44 @@ def publish(self, course, access_token=None):
...
+        """When validation errors occur during publication, the LMS is expected
+         to return an error message.
+
+        Arguments:
+            default_error_message (str) : default error message in case of exception.
```
zubair-arbi, line 158: nit: order parameters in `Arguments` just in same way as they are passed to this method.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,44 @@ def publish(self, course, access_token=None):
...
+            response_content (str): A string contains json error message.
```
zubair-arbi: change it to `response`

awais786:
@rlucioni it's ready for final review.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,44 @@ def publish(self, course, access_token=None):
...
 
-        return False
+        return error_message
+
+    def _parse_error(self, response, default_error_message):
+        """When validation errors occur during publication, the LMS is expected
+         to return an error message.
+
+        Arguments:
+            response (str): A string contains json error message.
```
rlucioni: This is not a string. It's a `Response` object.

ecommerce/courses/publishers.py
```diff
@@ -131,11 +137,44 @@ def publish(self, course, access_token=None):
...
+            default_error_message (str) : default error message in case of exception.
+
+        Returns:
+            string: Returns the error message extracted from response.content
+            along with default message. If no message is available in response
+            then default message will be return.
+        """
+        message = ''
```
rlucioni: `message = None`

rlucioni:
:+1: once you make those last two changes, commits are squashed, and tests pass.

PR #322, opened 2015-09-08
XCOM-503: Consolidated the test cases

aamir-khan:
- Removed the class `EnrollmentFulfillmentCreditSeatModuleTests`
- Merged the `EnrollmentFulfillmentCreditSeatModuleTests` test into `EnrollmentFulfillmentModuleTests`

@awais786 @zubair-arbi @ahsan-ul-haq  @tasawernawaz Please review this PR

### DIFF
--- a/ecommerce/extensions/fulfillment/tests/test_modules.py
+++ b/ecommerce/extensions/fulfillment/tests/test_modules.py
@@ -27,27 +27,44 @@ ProductAttribute = get_model("catalogue", "ProductAttribute")
 User = get_user_model()
 
 
-class EnrollmentFulfillmentTestMixin(CourseCatalogTestMixin, FulfillmentTestMixin):
-    certificate_type = 'test-certificate-type'
+@ddt.ddt
+@override_settings(EDX_API_KEY='foo')
+class EnrollmentFulfillmentModuleTests(CourseCatalogTestMixin, FulfillmentTestMixin, TestCase):
+    """Test course seat fulfillment."""
+
     course_id = 'edX/DemoX/Demo_Course'
+    certificate_type = 'test-certificate-type'
     provider = None
 
     def setUp(self):
-        super(EnrollmentFulfillmentTestMixin, self).setUp()
+        super(EnrollmentFulfillmentModuleTests, self).setUp()
 
-        user = UserFactory()
-        course = Course.objects.create(id=self.course_id, name='Demo Course')
-        self.seat = course.create_or_update_seat(self.certificate_type, False, 100, self.provider)
+        self.user = UserFactory()
+        self.course = Course.objects.create(id=self.course_id, name='Demo Course')
+
+        self.seat = self.course.create_or_update_seat(self.certificate_type, False, 100, self.provider)
 
         basket = BasketFactory()
         basket.add_product(self.seat, 1)
-        self.order = factories.create_order(number=1, basket=basket, user=user)
+        self.order = factories.create_order(number=1, basket=basket, user=self.user)
 
+    # pylint: disable=attribute-defined-outside-init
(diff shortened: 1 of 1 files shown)


aamir-khan:
@rlucioni @clintonb Please review this PR.

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -28,20 +28,25 @@
...
     def setUp(self):
         super(EnrollmentFulfillmentTestMixin, self).setUp()
 
-        user = UserFactory()
-        course = Course.objects.create(id=self.course_id, name='Demo Course')
-        self.seat = course.create_or_update_seat(self.certificate_type, False, 100, self.provider)
+        self.user = UserFactory()
+        self.course = Course.objects.create(id=self.course_id, name='Demo Course')
+
+    def create_seat_and_order(self, certificate_type='test-certificate-type', provider=None):
```
clintonb: Add a docstring explaining what this method does.

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -28,20 +28,25 @@
...
+        self.certificate_type = certificate_type  # pylint: disable=attribute-defined-outside-init
```
clintonb: Can this be disabled at the block level?

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -51,12 +56,16 @@ class EnrollmentFulfillmentModuleTests(EnrollmentFulfillmentTestMixin, TestCase)
 
     def test_enrollment_module_support(self):
         """Test that we get the correct values back for supported product lines."""
+        # Create the certificate type and order for the default types.
+        self.create_seat_and_order()
```
clintonb: This is getting called with every test. Move it to `setUp`.

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -28,20 +28,25 @@
 
 
 class EnrollmentFulfillmentTestMixin(CourseCatalogTestMixin, FulfillmentTestMixin):
-    certificate_type = 'test-certificate-type'
+
```
clintonb, line 34: This mixin is no longer necessary. It is only used in the class below. Merge them.

ecommerce/extensions/fulfillment/tests/test_modules.py
aamir-khan: I created this function because the last test `test_credit_enrollment_module_fulfill` uses different type of certificate and also the different provider. So this function can not be moved to `setUp`.

clintonb: You can still use `setUp`. In your credit test,  call `create_seat_and_order` as you are now with the certificate type and provider. Yes, we will have some extraneous data; but, I'll rather have extraneous fake data than extraneous real code.

clintonb:
:+1:

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -27,27 +27,44 @@
...
 
+    # pylint: disable=attribute-defined-outside-init
+    def create_seat_and_order(self, certificate_type='test-certificate-type', provider=None):
+        """ Create the certificate of given type and seat of given provider.
 
-@ddt.ddt
-@override_settings(EDX_API_KEY='foo')
-class EnrollmentFulfillmentModuleTests(EnrollmentFulfillmentTestMixin, TestCase):
-    """Test course seat fulfillment."""
+        Params:
```
rlucioni: `Arguments:`

rlucioni:
:+1: once the docstring is fixed and tests pass.

PR #324, opened 2015-09-10
Scope and Template Settings Fix

clintonb:
- Restored the permissions scope, which is essential for assigning staff permissions to new users.
- Removed old TEMPLATE_DEBUG setting

ECOM-2178

@rlucioni @jimabramson 

FYI @mjfrey


rlucioni:
:+1:

PR #325, opened 2015-09-11
Update Partner Model

ahsan-ul-haq:
WIP. Update partner model to add field which will save Partner code for the products.

ECOM-2235

### DIFF
--- a/ecommerce/extensions/partner/models.py
+++ b/ecommerce/extensions/partner/models.py
@@ -1,9 +1,23 @@
-from oscar.apps.partner.abstract_models import AbstractStockRecord
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from oscar.apps.partner.abstract_models import AbstractPartner, AbstractStockRecord
 from simple_history.models import HistoricalRecords
 
 
 class StockRecord(AbstractStockRecord):
     history = HistoricalRecords()
+
+
+class Partner(AbstractPartner):
+    # short_code is the unique identifier for the 'Partner'
+    short_code = models.CharField(max_length=8, unique=True, null=True)
+
+    class Meta(object):
+        # Model name that will appear in the admin panel
+        verbose_name = _('Partner')
+        verbose_name_plural = _('Partners')
+
 
 # noinspection PyUnresolvedReferences
 from oscar.apps.partner.models import *  # noqa pylint: disable=wildcard-import,unused-wildcard-import
--- a/ecommerce/extensions/partner/migrations/0007_auto_20150914_0841.py
+++ b/ecommerce/extensions/partner/migrations/0007_auto_20150914_0841.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+def add_short_code_data(apps, schema_editor):
+    """Get the value of the 'code' field from the Partner table and add it to
+    the field 'short_code'.
+    """
+    Partner = apps.get_model('partner', 'Partner')
+    partners = Partner.objects.all()
+    for partner in partners:
+        partner.short_code = partner.code
+        partner.save()
+
+
+def reverse_short_code_data(apps, schema_editor):
+    """Backward data migration for field 'short_code'."""
+    pass
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('partner', '0006_auto_20150709_1205'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='partner',
+            options={'verbose_name': 'Partner', 'verbose_name_plural': 'Partners'},
+        ),
+        migrations.AddField(
+            model_name='partner',
+            name='short_code',
+            field=models.CharField(max_length=8, unique=True, null=True),
+        ),
+        migrations.RunPython(add_short_code_data, reverse_short_code_data),
+    ]


ecommerce/extensions/partner/models.py
```diff
@@ -1,9 +1,15 @@
...
+from oscar.apps.partner.abstract_models import AbstractStockRecord, AbstractPartner
 from simple_history.models import HistoricalRecords
 
 
 class StockRecord(AbstractStockRecord):
     history = HistoricalRecords()
 
+
+class Partner(AbstractPartner):
+    product_code = models.CharField(_("Partner's Product Code"), max_length=8, unique=True, null=True)
```
clintonb: The name is simply "code". This value is simply a short name for the Partner. It is not specifically tied to orders or products.

Remove the label for now. I still can't think of a quick way to describe this field. Help text might be a better option.

ahsan-ul-haq: `AbstractPartner` already have a field name code https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/partner/abstract_models.py#L22. Therefore I named it `product_code` so we could either use that one or may be you could suggest some other name as we could not have `code`

clintonb: Ah. Didn't realize that. Let's go with `short_code` instead. Please add a comment explaining this issue so that we can avoid this confusion in the future.

ecommerce/extensions/partner/migrations/0007_partner_product_code.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('partner', '0006_auto_20150709_1205'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='partner',
+            name='product_code',
+            field=models.CharField(max_length=8, unique=True, null=True, verbose_name="Partner's Product Code"),
```
clintonb: This migration will fail if run on a machine that already has a Partner. Include a data migration that copies the value from the existing `code` field.

ecommerce/extensions/partner/models.py
```diff
@@ -1,9 +1,19 @@
...
 
 
 class StockRecord(AbstractStockRecord):
     history = HistoricalRecords()
 
+
+class Partner(AbstractPartner):
+    product_code = models.CharField(max_length=8, unique=True, null=True, help_text=_("Partner's Product Code"))
+
+    class Meta:
```
clintonb: This isn't really necessary. Those are the default values.

ahsan-ul-haq: I have done this as in admin panel this model appears as the name `Fulfillment partner` so I changed it here so it appears as `Partner`

clintonb: Ah. Right. Again, please add a comment explaining.

ecommerce/extensions/partner/models.py
```diff
@@ -1,9 +1,23 @@
...
 from simple_history.models import HistoricalRecords
 
 
 class StockRecord(AbstractStockRecord):
     history = HistoricalRecords()
 
+
+class Partner(AbstractPartner):
+    # As AbstractPartner model already have code field this field will be used
+    # as prefix with order number
```
clintonb: This is not _only_ for order numbers. The field could be used for multiple purposes in the future.

ecommerce/extensions/partner/models.py
```diff
@@ -1,9 +1,23 @@
...
+    short_code = models.CharField(max_length=8, unique=True, null=True,
```
zubair-arbi: Since this field `short_code` will be used as prefix so I think it would make sense to restrict the user to use alphanumeric characters

ahsan-ul-haq:
@clintonb I have added data migration could you please review.

ecommerce/extensions/partner/models.py
```diff
@@ -1,9 +1,22 @@
...
 
 
 class StockRecord(AbstractStockRecord):
     history = HistoricalRecords()
 
+
+class Partner(AbstractPartner):
+    # short_code is the unique identifier for the 'Partner'
+    short_code = models.CharField(max_length=8, unique=True, null=True,
+                                  help_text=_("Partner's code that will appears as prefix with order"))
```
clintonb: Stop referencing orders.

clintonb: Just remove the help text for now.

clintonb:
:+1: (once the help text is removed)

ecommerce/extensions/partner/models.py
ahsan-ul-haq: so could I add Partner's unique identifier

ecommerce/extensions/partner/models.py
```diff
@@ -1,9 +1,21 @@
-from oscar.apps.partner.abstract_models import AbstractStockRecord
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+from oscar.apps.partner.abstract_models import AbstractStockRecord, AbstractPartner
```
zubair-arbi: nit: please order the imports

zubair-arbi:
lgtm :+1:

ecommerce/extensions/partner/models.py
```diff
@@ -1,9 +1,23 @@
...
 from simple_history.models import HistoricalRecords
 
 
 class StockRecord(AbstractStockRecord):
     history = HistoricalRecords()
 
+
+class Partner(AbstractPartner):
+    # short_code is the unique identifier for the 'Partner'
+    short_code = models.CharField(max_length=8, unique=True, null=True)
```
clintonb, line 14: @ahsan-ul-haq Please create a follow-up PR and make this non-nullable.

PR #326, opened 2015-09-11
Update basket model for multi-tenancy

awais786:
Adding foreign key from the Basket model pointing to the Partner model. Due to this change dashboard was crashing also lot of test cases were failing too. Reason was oscar default functionality were creating basket without any partners. To fix these I have created new middleware and new factory methods.


clintonb:
The model changes are missing.

ecommerce/extensions/basket/migrations/0006_partner_key.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def add_partner_reference_id(apps, schema_editor):
+    """ Updating existing stock records. Adding relation between stock record
+    and Partner exd.
+    """
+    Basket = apps.get_model('basket', 'Basket')
+    Partner = apps.get_model('partner', 'Partner')
+
+    partner, __ = Partner.objects.get_or_create(code='edx')
```
clintonb: Never use `.get_or_create()` in a migration. You run the risk of creating data unexpectedly. Just use `.get()`.

ecommerce/extensions/basket/models.py
```diff
@@ -6,6 +8,9 @@
 
 
 class Basket(AbstractBasket):
+
+    partner = models.ForeignKey('partner.Partner', null=True, blank=True, related_name='partner')
```
clintonb: `null=False`
`blank=False`
`related_name='baskets'`

You will need to update your migration to do three things:
1. Add the column as nullable.
2. Update the existing data.
3. Make the column non-nullable.

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -15,3 +16,12 @@ def test_order_number_generation(self):
         """Verify that an instance of Basket can generate its own order number."""
         expected = OrderNumberGenerator().order_number(self.basket)
         self.assertEqual(self.basket.order_number, expected)
+
+    def test_basket_relation_with_partner(self):
```
clintonb: This test isn't useful. You're testing the fact that Django can create foreign keys, which we already know to be true.

ecommerce/extensions/api/data.py
```diff
@@ -6,21 +6,24 @@
 
 Basket = get_model('basket', 'Basket')
 Product = get_model('catalogue', 'Product')
+Partner = get_model('partner', 'Partner')
```
clintonb: This isn't used anywhere.

ecommerce/extensions/api/v2/views.py
```diff
@@ -126,7 +126,6 @@ def create(self, request, *args, **kwargs):
                 }
             }
         """
-        basket = data_api.get_basket(request.user)
```
clintonb: Revert. Get the partner from the current site's configuration. See https://docs.djangoproject.com/en/1.8/ref/contrib/sites/#django.contrib.sites.shortcuts.get_current_site.

FYI @zubair-arbi 

``` python
from django.contrib.sites.shortcuts import get_current_site
site = get_current_site(request)
partner = site.site_configuration.partner
```

ecommerce/extensions/api/data.py
```diff
@@ -6,21 +6,24 @@
...
-def get_basket(user):
-    """Retrieve the basket belonging to the indicated user.
+def get_basket(user, partner):
+    """Retrieve the basket belonging to the indicated user and partner.
 
     If no such basket exists, create a new one. If multiple such baskets exist,
     merge them into one.
     """
-    editable_baskets = Basket.objects.filter(owner=user, status__in=Basket.editable_statuses)
+    # Partner.objects.get()
```
clintonb: Remove this comment.

ecommerce/extensions/api/v2/views.py
awais786: @clintonb Just a question why we are picking partner from config. Instead of seat associated partner ?

clintonb: Let's say the product has multiple stock records associated with multiple partners. How do you know which partner is the right partner to use?

You will also need to update the code to ensure the stock record associated with the SKU is associated with the correct partner. Again, we want to avoid brand leakage where a user for one site can purchase products from another site.

ecommerce/extensions/api/v2/views.py
```diff
@@ -126,7 +128,11 @@ def create(self, request, *args, **kwargs):
...
+
+        site = get_current_site(request)
+        partner = site.site_configuration.partner
```
awais786: This code needs another PR in progress from @zubair-arbi

awais786:
@clintonb There is an exception while accessing the dashboard.
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/basket/middleware.py#L136

NOT NULL constraint failed: basket_basket.partner_id

clintonb:
I have updated the middleware. Now working on test cases. Following method is used every where in test cases and all are failing.
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/test/factories/__init__.py#L142 

```
basket = Basket.objects.create()
```

I have created my own method custom_create_order and updating the test cases.  Is there any better approach for this ?

ecommerce/extensions/api/v2/views.py
```diff
@@ -3,6 +3,8 @@
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.contrib.sites.shortcuts import get_current_site
+
```
clintonb: Remove the newline.

ecommerce/extensions/basket/migrations/0006_partner_key.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def add_partner_reference_id(apps, schema_editor):
+    """ Updating existing stock records. Adding relation between stock record
+    and Partner exd.
+    """
+    Basket = apps.get_model('basket', 'Basket')
+    Partner = apps.get_model('partner', 'Partner')
+
+    partner = Partner.objects.get(code='edx')
```
clintonb: Use `id=1`;

ecommerce/extensions/basket/migrations/0007_auto_20150914_1032.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('basket', '0006_partner_key'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
clintonb: Merge this with the migration above.

awais786:
@clintonb Please check the approach of new middleware and methods. 
At the moment there are few quality issues i will fix tomorrow.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -70,6 +71,8 @@ def setUp(self):
             stockrecords__price_excl_tax=Decimal('240000.00'),
         )
 
+        self.partner = Partner.objects.first()
```
clintonb: Create a `Partner` rather than assuming one exists.

ecommerce/extensions/api/v2/views.py
```diff
@@ -126,7 +128,13 @@ def create(self, request, *args, **kwargs):
                 }
             }
         """
-        basket = data_api.get_basket(request.user)
+
+        # The multi-tenant implementation has one site per partner
+
+        site = get_current_site(request)
+        partner = site.siteconfiguration_set.first().partner
```
clintonb: `site.siteconfiguration.partner`

ecommerce/extensions/basket/middleware.py
```diff
@@ -0,0 +1,66 @@
+from django.contrib.sites.shortcuts import get_current_site
+from oscar.apps.basket.middleware import BasketMiddleware
+
```
clintonb: Remove the newline.

ecommerce/extensions/basket/middleware.py
```diff
@@ -0,0 +1,66 @@
...
+from oscar.core.loading import get_model
```
clintonb, line 3: Merge with the line below.

ecommerce/extensions/basket/middleware.py
```diff
@@ -0,0 +1,66 @@
...
+from oscar.core.loading import get_class
+
+Applicator = get_class('offer.utils', 'Applicator')
+Basket = get_model('basket', 'basket')
+Selector = get_class('partner.strategy', 'Selector')
+
+selector = Selector()
+
+
+class CustomBasketMiddleware(BasketMiddleware):
```
clintonb: Call it `PartnerBasketMiddleware`. Also, add a docstring explaining the need for this middleware.

ecommerce/tests/mixins.py
```diff
@@ -6,11 +6,15 @@
 import jwt
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.contrib.sites.models import Site
 from django.core.cache import cache
 from django.core.urlresolvers import reverse
 from mock import patch
+
```
clintonb: Remove the newline.

ecommerce/tests/mixins.py
```diff
@@ -6,11 +6,15 @@
...
+from oscar.apps.partner import strategy
 from oscar.test import factories
 from oscar.core.loading import get_model, get_class
+from ecommerce.core.models import SiteConfiguration
```
clintonb: Group this with the imports below.

ecommerce/tests/mixins.py
```diff
@@ -62,7 +67,75 @@ def generate_token(self, payload, secret=None):
         return token
 
 
-class BasketCreationMixin(JwtMixin):
+class BasketWithPartnerMixin(object):
```
clintonb: This class isn't really necessary. Instead of updating the tests, update the factories and other methods that create data to either allow a partner to be specified or create one.

ecommerce/tests/mixins.py
```diff
@@ -62,7 +67,75 @@ def generate_token(self, payload, secret=None):
...
+        SiteConfiguration.objects.create(
+            site=site,
+            partner=self.partner,
+            lms_url_root='/lms/root/',
+            theme_scss_path='/css/path/',
+            payment_processors='paypal'
+        )
+
+
+def custom_create_order(number=None, basket=None, user=None, shipping_address=None,
```
clintonb: Similar to `oscar.test.factories` we should have an `ecommerce.extenstions.test.factories` where these methods belong. Get rid of the `custom_` prefix.

awais786:
@clintonb Please review this PR again.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -11,6 +12,8 @@
 from ecommerce.extensions.checkout.signals import send_course_purchase_email
 from ecommerce.settings import get_lms_url
 
+Partner = get_model('partner', 'Partner')
```
clintonb: This isn't used anywhere.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -29,8 +32,11 @@ def test_post_checkout_callback(self):
         course = Course.objects.create(id='edX/DemoX/Demo_Course', name='Demo Course')
         seat = course.create_or_update_seat('credit', False, 50, 'ASU', None, 2)
 
-        basket = BasketFactory()
+        basket = BasketFactory(partner=self.partner)
```
clintonb: Where is `self.partner` declared?

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -29,8 +32,11 @@ def test_post_checkout_callback(self):
...
         basket.add_product(seat, 1)
+        basket.partner = self.partner
```
clintonb: Isn't the purpose of updating `BasketFactory` to have it set the partner?

ecommerce/extensions/dashboard/tests.py
```diff
@@ -1,6 +1,9 @@
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 from oscar.test.factories import OrderFactory, UserFactory
+from oscar.core.loading import get_model
```
clintonb: Move this up a line to keep the imports alphabetized.

ecommerce/tests/mixins.py
```diff
@@ -94,6 +97,17 @@ def setUp(self):
             stockrecords__price_excl_tax=D('0.00'),
         )
 
+        site = Site.objects.get(domain='example.com')
```
clintonb: This `Site` is not guaranteed to exist. Use `get_or_create`.

ecommerce/extensions/test/factories.py
```diff
@@ -0,0 +1,66 @@
+# -*- coding: utf-8 -*-
+"""Broadly-useful factory methods for use in automated tests."""
+from decimal import Decimal as D
+
+from oscar.apps.partner import strategy
+from oscar.test import factories
```
clintonb: Consider importing both Oscar-provided factories, and overriding them as necessary:

``` python
form oscar.test.factories import *
from oscar.test.newfactories import *

def create_order(...):
  ...

# Not sure if this is possible.
class BasketFactory(BasketFactory):
  ...
```

ecommerce/extensions/checkout/tests/test_signals.py
awais786: It is available in CourseCatalogTestMixin class.

ecommerce/extensions/test/factories.py
awais786: I did not understand the purpose of Importing both factories.

clintonb: If you can determine how to combine the factories into a single file, you
can avoid having to import 2-3 different factory modules.

## 

Clinton Blackburn

ecommerce/extensions/api/v2/tests/views/__init__.py
```diff
@@ -1,16 +1,20 @@
+from django.contrib.sites.models import Site
 from django.core.urlresolvers import reverse
 from django.test import RequestFactory
-from oscar.core.loading import get_class
-from oscar.test import factories
+from oscar.core.loading import get_class, get_model
 from oscar.test.newfactories import ProductAttributeValueFactory
 
+
```
zubair-arbi, line 4: nit: remove extra line

ecommerce/tests/mixins.py
```diff
@@ -94,6 +97,17 @@ def setUp(self):
             stockrecords__price_excl_tax=D('0.00'),
         )
 
+        site, __ = Site.objects.get_or_create(domain='example.com')
+        partner, __ = Partner.objects.get_or_create(short_code='edx', name='edx')
+
+        SiteConfiguration.objects.create(
+            site=site,
+            partner=partner,
+            lms_url_root='/lms/root/',
```
zubair-arbi: this should be a url, chage it to `https://courses.stage.edx.org`

zubair-arbi:
@awais786 two small changes otherwise :+1:
@clintonb the build has passed but the `coveralls` is not responding on all PR's

ecommerce/extensions/refund/tests/factories.py
```diff
@@ -3,10 +3,10 @@
 from django.conf import settings
 import factory
 from oscar.core.loading import get_model
-from oscar.test import factories
-from oscar.test.newfactories import UserFactory
+from ecommerce.extensions.test.factories import UserFactory
```
awais786: @clintonb this is what you were suggesting ? Importing things from newly developed factory.

clintonb: Yes.

clintonb: Although, you're already importing `ecommerce.extensions.test.factories` below. No need for both.

ecommerce/extensions/basket/migrations/0006_partner_key.py
```diff
@@ -0,0 +1,47 @@
...
+    dependencies = [
+        ('partner', '0006_auto_20150709_1205'),
+        ('basket', '0005_auto_20150709_1205'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='basket',
+            name='partner',
+            field=models.ForeignKey(related_name='baskets', blank=True, to='partner.Partner', null=True),
```
clintonb: I was reminded that you can do all of the migrations in a single step:

``` python
migrations.AddField(
            model_name='basket',
            name='partner',
            field=models.ForeignKey(default=1, to='partner.Partner'),
            preserve_default=False,
        ),
```

Delete this migration and re-run `makemigrations`. Specify 1 as the default.

ecommerce/extensions/basket/models.py
```diff
@@ -1,3 +1,5 @@
+from django.db import models
+
```
clintonb: No need for the newline.

ecommerce/extensions/basket/models.py
```diff
@@ -6,6 +8,9 @@
 
 
 class Basket(AbstractBasket):
+
```
clintonb: No need for the newline.

ecommerce/extensions/test/factories.py
```diff
@@ -0,0 +1,69 @@
...
+"""Broadly-useful factory methods for use in automated tests."""
+import factory
+
+from oscar.test.factories import *  # pylint: disable=unused-wildcard-import, function-redefined, wildcard-import
+from oscar.test.newfactories import *  # pylint: disable=unused-wildcard-import, function-redefined, wildcard-import
+
+
+class PartnerFactory(factory.DjangoModelFactory):  # pylint: disable=function-redefined
+    name = "Gardners"
+    short_code = factory.Sequence(lambda n: 'code' + '-{0}'.format(n))
```
clintonb: Keep this simple and use a single number.

ecommerce/extensions/test/factories.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+"""Broadly-useful factory methods for use in automated tests."""
+import factory
+
+from oscar.test.factories import *  # pylint: disable=unused-wildcard-import, function-redefined, wildcard-import
+from oscar.test.newfactories import *  # pylint: disable=unused-wildcard-import, function-redefined, wildcard-import
+
+
+class PartnerFactory(factory.DjangoModelFactory):  # pylint: disable=function-redefined
+    name = "Gardners"
```
clintonb: Single quotes.

ecommerce/extensions/test/factories.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+"""Broadly-useful factory methods for use in automated tests."""
+import factory
+
+from oscar.test.factories import *  # pylint: disable=unused-wildcard-import, function-redefined, wildcard-import
+from oscar.test.newfactories import *  # pylint: disable=unused-wildcard-import, function-redefined, wildcard-import
+
+
+class PartnerFactory(factory.DjangoModelFactory):  # pylint: disable=function-redefined
```
clintonb: What is the Pylint disable for?

ecommerce/extensions/refund/tests/factories.py
```diff
@@ -18,7 +17,7 @@
 
 class RefundFactory(factory.DjangoModelFactory):
```
clintonb: Move this to the new factories module.

ecommerce/extensions/catalogue/tests/mixins.py
```diff
@@ -23,7 +23,7 @@ def setUp(self):
 
         # Force the creation of a seat ProductClass
         self.seat_product_class  # pylint: disable=pointless-statement
-        self.partner, _created = Partner.objects.get_or_create(code='edx')
+        self.partner, _created = Partner.objects.get_or_create(short_code='edx', name='edx')
```
clintonb: Use the factory.

ecommerce/extensions/dashboard/tests.py
```diff
@@ -18,6 +22,8 @@ def assert_message_equals(self, response, msg, level):  # pylint: disable=unused
 class ExtendedIndexViewTests(TestCase):
     def test_average_paid_order_costs(self):
         """ Verify the stats contain average_paid_order_costs. """
+        partner, __ = Partner.objects.get_or_create(short_code='edx', name='edx')
```
clintonb: Use the factory.

ecommerce/extensions/dashboard/tests.py
```diff
@@ -26,6 +32,6 @@ def test_average_paid_order_costs(self):
         actual = response.context['average_paid_order_costs']
         self.assertEqual(actual, 0)
 
-        order = OrderFactory()
+        order = OrderFactory(basket__partner=partner)
```
clintonb: Do this in the factory. Ideally, unless you are testing for an explicit partner, you should not need to create a partner in order to create a basket or order. The factories should do all of that for you.

ecommerce/extensions/test/factories.py
awais786: Because I am declaring same class name factory which already exist.
This is the error message.
E:  9, 0: class already defined line 5 (function-redefined)

awais786:
@clintonb please take a look on this PR. Few of the test cases failing I am trying to fix.

ecommerce/tests/mixins.py
```diff
@@ -111,6 +125,7 @@ def create_basket(self, skus=None, checkout=None, payment_processor_name=None, a
             request_data[AC.KEYS.PAYMENT_PROCESSOR_NAME] = payment_processor_name
 
         if auth:
+
```
clintonb: Remove.

ecommerce/extensions/catalogue/tests/test_migrate_course.py
```diff
@@ -43,7 +44,8 @@ class CourseMigrationTestMixin(CourseCatalogTestMixin):
     commerce_api_url = '{}/courses/{}/'.format(settings.COMMERCE_API_URL.rstrip('/'), course_id)
     course_structure_url = urljoin(settings.LMS_URL_ROOT, 'api/course_structure/v0/courses/{}/'.format(course_id))
     enrollment_api_url = urljoin(settings.LMS_URL_ROOT, 'api/enrollment/v1/course/{}'.format(course_id))
-    partner_short_code = 'edx'
+    partner = PartnerFactory(name='edx')
```
clintonb: This, along with the line below, needs to be moved into a setup method. Your tests are failing because, when the mixins are loaded, they are attempting to create a Partner. However, the table doesn't exist until the tests at this time during execution.

awais786:
@clintonb The coverage is missing. Here are the test cases link. We can try to make some thing like this for coverage.
https://github.com/django-oscar/django-oscar/blob/master/tests/unit/basket/middleware_tests.py

clintonb:
@awais786 you can use that as a starting point, yes.

awais786:
@clintonb , @rlucioni Following are stats on my branch
Total records in basket before test = 2096727.
Users = 10
Hatch Rate = 2
- Start Executing the tests.
- Execute the migration. ( it took 2 min to complete. )
- For initial request there are following errors.

OperationalError: (1054, "Unknown column 'basket_basket.partner_id' in 'field list'")
Total records in basket after test = 2096771.

So total baskets created = 44
screen shot 2015-10-07 at 8 27 23 pm
requests
time

awais786:
On average migration is taking 2-3 mins but for one instance it took 7 mins also.

awais786:
Post migration test cases.

Users = 10
Hatch Rate = 2
screen shot 2015-10-07 at 8 40 32 pm

awais786:
These are the stats with master branch.
screen shot 2015-10-07 at 8 40 32 pm

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -21,6 +19,7 @@
 from ecommerce.extensions.fulfillment.tests.mixins import FulfillmentTestMixin
 from ecommerce.tests.mixins import PartnerMixin
 
+from ecommerce.extensions.test.factories import UserFactory, BasketFactory, create_order
```
clintonb, line 21: Group this with the other imports.

ecommerce/tests/mixins.py
```diff
@@ -96,6 +99,17 @@ def setUp(self):
             stockrecords__price_excl_tax=D('0.00'),
         )
 
+        site, __ = Site.objects.get_or_create(domain='example.com')
+        partner, __ = Partner.objects.get_or_create(short_code='edx', name='edx')
+
+        SiteConfiguration.objects.create(
+            site=site,
+            partner=partner,
+            lms_url_root='https://courses.stage.edx.org',
```
clintonb: Avoid using real servers. Use 'https://lms.example.com' instead.

clintonb:
Squash your commits and rebase on master.

awais786:
@clintonb any feedback on load stats ? specially on the error.
OperationalError: (1054, "Unknown column 'basket_basket.partner_id' in 'field list'")

clintonb:
@awais786 I'm not sure what that error means. What investigation have you done?

awais786:
@clintonb we run the load test first and then run the migration. And migration took 2 mins to complete. And i think at the time field was not in sql but django-orm was looking for the field.

clintonb:
@awais786 have you run this in the load test environment?

awais786:
@clintonb I run the test case from local and target was my sandbox
awais786.sandbox.edx.org

awais786:
@clintonb this is the traceback.

```
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/utils/decorators.py", line 30, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "/edx/app/ecommerce/ecommerce/ecommerce/extensions/api/v2/views/baskets.py", line 36, in dispatch
    return super(BasketCreateView, self).dispatch(request, *args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/oscar/apps/checkout/session.py", line 68, in dispatch
    request, *args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/views.py", line 466, in dispatch
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/views.py", line 463, in dispatch
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/generics.py", line 192, in post
    return self.create(request, *args, **kwargs)
  File "/edx/app/ecommerce/ecommerce/ecommerce/extensions/api/v2/views/baskets.py", line 131, in create
    basket = data_api.get_basket(request.user, partner)
  File "/edx/app/ecommerce/ecommerce/ecommerce/extensions/api/data.py", line 23, in get_basket
    if len(editable_baskets) == 0:
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 144, in __len__
    self._fetch_all()
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 238, in iterator
    results = compiler.execute_sql()
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 840, in execute_sql
    cursor.execute(sql, params)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/utils.py", line 97, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/backends/mysql/base.py", line 124, in execute
    return self.cursor.execute(query, args)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/MySQLdb/cursors.py", line 205, in execute
    self.errorhandler(self, exc, value)
    raise errorclass, errorvalue
OperationalError: (1054, "Unknown column 'basket_basket.partner_id' in 'field list'")
```

awais786:
Now i am splitting the PR into two.
One will contain only migration and model file.
2nd will contain the new middleware and updated test cases.

ahsan-ul-haq:
LGTM :+1:

rlucioni:
@cpennington FYI

ecommerce/extensions/basket/models.py
```diff
@@ -6,6 +7,8 @@
 
 
 class Basket(AbstractBasket):
+    partner = models.ForeignKey('partner.Partner', null=False, blank=False, related_name='baskets')
```
cpennington: This is going to cause a problem, going to production.

The migration will run before the code is deployed to the running cluster. So, the existing cluster will be writing data to a column that isn't nullable, and will error out. We either need to::
1. create this column as nullable, and then, as a second deployment, convert it to non-nullable, or
2. create this column with a default value.

If we chose option 1, then we should consider making the process that converts `null` to `Provider 1`, as is done in the migration currently, a background task that can be executed before running the migration that marks the column as non-nullable.

In any of these cases, we should rerun the loadtests to see the effect the migration has active creation requests.

awais786:
@cpennington I go with 2nd option. Add the default value in model too. If it looks ok i will run the load test tomorrow.

clintonb:
@awais786 please share your plan for splitting this PR. It's not clear why #385 was closed. As I see it you need to do this work in four phases:
1. Update basket table with nullable partner_id field that defaults to the value of 1.
2. Update basket creation code to set the partner foreign key.
3. Run a SQL script (not a migration) that updates the existing basket rows.
4. Remove the default value, and make the field nullable.

zubair-arbi:
@clintonb @rlucioni @cpennington 
We have done loadtesting with following configuration.

```
Users: 10
Hatch rate: 2
min_wait = 30 * 1000
max_wait = 120 * 1000
```

**Steps:**
1. Run loadtests with old code and the first migration (partner `null` with default `None`). 
2. Restart server and Run loadtests with new code.
**Results:**
There were only two or three request failing with the error `OperationalError: (1213, 'Deadlock found when trying to get lock; try restarting transaction')`. A full traceback of this error can be found here:
https://gist.github.com/zubair-arbi/c7bf7d63115a2ba15e12
Discussed with @rlucioni and we think that this amount of errors are tolerable.

The first migration is created with `deault=None` since with `default=1` requests were failing with the error `OperationalError: (1364, "Field 'partner_id' doesn't have a default value")`. Full traceback of this error can be found here:
https://gist.github.com/zubair-arbi/7b08cdf81d0c990936df

**Next Plan**
1. Do a release with new code **(this PR)**and first migration (with partner null).
2. Do a second release with second migration (with partner `not` null and `default` `1`) https://github.com/edx/ecommerce/pull/392.

rlucioni:
@zubair-arbi why were the minimum and maximum wait times set so high during the load test? Is that a typo? If you bring them down to a more conventional range (3 to 5 seconds) you're likely to see many more errors when running the first migration due to the baskets table being locked. How long does the first migration take to complete with 2.5 million records in the table?

zubair-arbi:
@rlucioni This configuration was suggested by @cpennington since migration was taking average 2 min to complete. We also were suggested to increase the number of user spawn per minute but the LMS service in sandbox crashes.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -122,8 +123,12 @@ def create(self, request, *args, **kwargs):
         # atomic() context managers restore atomicity at points where we are modifying data
         # (baskets, then orders) to ensure that we don't leave the system in a dirty state
         # in the event of an error.
+        # The multi-tenant implementation has one site per partner
```
clintonb, line 122: Put the comments in the right order. The atomic comment belongs above the atomic transaction.

clintonb:
@feanil it seems that no matter what we do an `ALTER` statement will lock the table. Any recommendations? Do we have access to [Percona Toolkit](https://www.percona.com/doc/percona-toolkit/2.1/pt-online-schema-change.html), or similar tools, so that we can make these changes without having to deal with locking?

feanil:
@clintonb we don't have percona toolkit on the RDS instances.  What about not creating a new column and having a new table that tracks the tenancy data?

PR #327, opened 2015-09-11
Restored offers navigation link

clintonb:
Users with dashboard access will now see links to manage offers and vouchers.

@rlucioni @mjfrey


rlucioni:
:+1:

PR #328, opened 2015-09-14
add model 'SiteConfiguration' and django admin for custom sites

zubair-arbi:
ECOM-2233
@awais786 @aamir-khan @ahsan-ul-haq @tasawernawaz 
- Add model `SiteConfiguration` which extends [Django “sites” framework](https://docs.djangoproject.com/en/1.8/ref/contrib/sites/).
- Add foreign key to model `Partner` for multi-tenant implementation with one site per partner.
- Add migration for app `core` which has new custom sites model `SiteConfiguration`.
- Add django admin for model `SiteConfiguration` in app `core` admin.

### DIFF
--- a/ecommerce/core/admin.py
+++ b/ecommerce/core/admin.py
@@ -2,7 +2,12 @@ from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
 from django.utils.translation import ugettext_lazy as _
 
-from ecommerce.core.models import User
+from ecommerce.core.models import SiteConfiguration, User
+
+
+class SiteConfigurationAdmin(admin.ModelAdmin):
+    list_display = ('site', 'partner', 'lms_url_root', 'theme_scss_path', 'payment_processors')
+    search_fields = ['site__name']
 
 
 class EcommerceUserAdmin(UserAdmin):
@@ -16,4 +21,5 @@ class EcommerceUserAdmin(UserAdmin):
     )
 
 
+admin.site.register(SiteConfiguration, SiteConfigurationAdmin)
 admin.site.register(User, EcommerceUserAdmin)
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -1,7 +1,42 @@
 from django.contrib.auth.models import AbstractUser
-from jsonfield.fields import JSONField
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
+from jsonfield.fields import JSONField
+
+
+class SiteConfiguration(models.Model):
+    """Custom Site model for custom sites/microsites.
+
+    This model will enable the basic theming and payment processor
+    configuration for each custom site.
+    The multi-tenant implementation has one site per partner.
+    """
+
+    site = models.ForeignKey('sites.Site', null=False, blank=False)
+    partner = models.ForeignKey('partner.Partner', null=False, blank=False)
+    lms_url_root = models.URLField(
+        verbose_name=_('LMS base url for custom site/microsite'),
+        help_text=_("Root URL of this site's LMS (e.g. https://courses.stage.edx.org)"),
+        null=False,
+        blank=False
+    )
+    theme_scss_path = models.CharField(
+        verbose_name=_('Path to custom site theme'),
+        help_text=_('Path to scss files of the custom site theme'),
+        max_length=255,
+        null=False,
+        blank=False
+    )
+    payment_processors = models.CharField(
+        verbose_name=_('Payment processors'),
+        help_text=_("Comma-separated list of processor names: 'cybersource,paypal'"),
+        max_length=255,
+        null=False,
+        blank=False
+    )
+
+    class Meta(object):
+        unique_together = ('site', 'partner')
 
 
 class User(AbstractUser):
--- a/ecommerce/core/migrations/0003_auto_20150914_1120.py
+++ b/ecommerce/core/migrations/0003_auto_20150914_1120.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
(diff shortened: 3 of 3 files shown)


aamir-khan:
:+1:

zubair-arbi:
@clintonb Please review.
There is one confusion about the relation between the `SiteConfiguration` and the `Partner`. In the `SiteConfiguration` model in the wiki `Otto+Multi-tenancy` there is no relationship shown for the `partner`.

ecommerce/core/models.py
```diff
@@ -1,8 +1,33 @@
...
+    """Custom Site model for custom sites/microsites.
+
+    This model will enable the basic theming and payment processor
+    configuration for each custom site.
+    """
+
+    site = models.ForeignKey(Site)
+    lms_url_root = models.URLField(
+        verbose_name=_('LMS base url for custom site/microsite'),
+        help_text=_('The multi-tenant implementation has one site per partner: '
```
clintonb: Root URL of this site's LMS (e.g. https://courses.stage.edx.org)

zubair-arbi:
I have added the foreign key to model 'partner' in model 'SiteConfiguration'. Please review.

ecommerce/core/models.py
```diff
@@ -1,8 +1,36 @@
...
+
+class SiteConfiguration(models.Model):
+    """Custom Site model for custom sites/microsites.
+
+    This model will enable the basic theming and payment processor
+    configuration for each custom site.
+    The multi-tenant implementation has one site per partner.
+    """
+
+    site = models.ForeignKey(Site)
```
clintonb: Add `null=False` and `blank=False` to all fields.

zubair-arbi:
@clintonb there is one more thing. Do we need a unique togather constraint for `site` and `partner`?

ecommerce/core/models.py
```diff
@@ -1,8 +1,36 @@
...
+    partner = models.ForeignKey(Partner)
```
clintonb: Use `partner.Partner` and get rid of the import. For future reference, use `get_model()` to load Oscar models: https://django-oscar.readthedocs.org/en/releases-1.1/topics/class_loading_explained.html.

ecommerce/core/admin.py
```diff
@@ -1,8 +1,14 @@
...
 from django.utils.translation import ugettext_lazy as _
+from django.contrib.sites.admin import SiteAdmin
 
-from ecommerce.core.models import User
+from ecommerce.core.models import SiteConfiguration, User
+
+
+class SiteConfigurationAdmin(SiteAdmin):
+    list_display = ('site', 'partner', 'lms_url_root', 'theme_scss_path', 'payment_processors')
+    search_fields = ['site__name', 'payment_processors']
```
clintonb: We don't need to search by payment processor.

ecommerce/core/models.py
zubair-arbi: @clintonb aren't these attributes set by default?

ecommerce/core/models.py
```diff
@@ -1,8 +1,36 @@
 from django.contrib.auth.models import AbstractUser
-from jsonfield.fields import JSONField
+from django.contrib.sites.models import Site
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
+from jsonfield.fields import JSONField
```
clintonb, line 4: Let's stick with our three import groups for now: native, third-party, project.

clintonb:
@zubair-arbi Nice catch. Yes, we do need that constraint.

ecommerce/core/models.py
zubair-arbi: I have fetch model 'Partner' explicitly at top:

``` python
Partner = get_model('partner', 'Partner')
```

zubair-arbi:
@clintonb Please review again

ecommerce/core/admin.py
```diff
@@ -1,8 +1,14 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
 from django.utils.translation import ugettext_lazy as _
+from django.contrib.sites.admin import SiteAdmin
 
-from ecommerce.core.models import User
+from ecommerce.core.models import SiteConfiguration, User
+
+
+class SiteConfigurationAdmin(SiteAdmin):
```
clintonb: This should inherit from `ModelAdmin`.

ecommerce/core/models.py
clintonb: I don't know without digging into the code. It doesn't hurt to be explicit.

zubair-arbi: https://docs.djangoproject.com/en/1.8/ref/models/fields/#field-options

ecommerce/core/models.py
clintonb: You can avoid the extra code altogether by replacing `Partner` with `'partner.Partner'`. The same for `Site`.

zubair-arbi:
@clintonb its ready for final review.

clintonb:
:+1:

ecommerce/core/models.py
```diff
@@ -1,7 +1,42 @@
...
+
+class SiteConfiguration(models.Model):
+    """Custom Site model for custom sites/microsites.
+
+    This model will enable the basic theming and payment processor
+    configuration for each custom site.
+    The multi-tenant implementation has one site per partner.
+    """
+
+    site = models.ForeignKey('sites.Site', null=False, blank=False)
```
clintonb, line 15: @zubair-arbi can you update this to be a [`OneToOneField`](https://docs.djangoproject.com/en/1.8/ref/models/fields/#ref-onetoone), please? I just noticed my mistake in recommending a `ForeignKey`.

zubair-arbi, line 15: @clintonb sure. I will create a new PR.

zubair-arbi, line 15: @clintonb I have add a new PR: https://github.com/edx/ecommerce/pull/334

PR #329, opened 2015-09-14
WIP update order model

tasawernawaz:
ECOM-2236
- Update the model `Order` and the order number generation class `OrderNumberGenerator` logic. 
- Add migration for the change in the model `Order`
- Update related tests.

This PR is dependent on (ECOM-2232): https://github.com/edx/ecommerce/pull/326
Docs: https://github.com/edx/ecommerce/pull/341

### DIFF
--- a/ecommerce/extensions/order/models.py
+++ b/ecommerce/extensions/order/models.py
@@ -9,6 +9,7 @@ from ecommerce.extensions.fulfillment.status import ORDER
 
 class Order(AbstractOrder):
     history = HistoricalRecords()
+    partner = models.ForeignKey('partner.Partner', related_name='orders')
 
     @property
     def is_fulfillable(self):
--- a/ecommerce/extensions/order/tests/test_utils.py
+++ b/ecommerce/extensions/order/tests/test_utils.py
@@ -8,9 +8,6 @@ from ecommerce.extensions.order.utils import OrderNumberGenerator
 class UtilsTest(TestCase):
     """Unit tests for the order utility functions and classes. """
 
-    ORDER_NUMBER_PREFIX = 'FOO'
-
-    @override_settings(ORDER_NUMBER_PREFIX=ORDER_NUMBER_PREFIX)
     def test_order_number_generation(self):
         """
         Verify that order numbers are generated correctly, and that they can
@@ -22,8 +19,10 @@ class UtilsTest(TestCase):
         first_order_number = OrderNumberGenerator().order_number(first_basket)
         second_order_number = OrderNumberGenerator().order_number(second_basket)
 
-        self.assertIn(self.ORDER_NUMBER_PREFIX, first_order_number)
-        self.assertIn(self.ORDER_NUMBER_PREFIX, second_order_number)
+        # TODO add partner code instead of hard coded value
+
+        # self.assertIn(first_basket.short_code, first_order_number)
+        # self.assertIn(second_basket.short_code, second_order_number)
         self.assertNotEqual(first_order_number, second_order_number)
 
         first_basket_id = OrderNumberGenerator().basket_id(first_order_number)
--- a/ecommerce/extensions/order/utils.py
+++ b/ecommerce/extensions/order/utils.py
@@ -1,5 +1,5 @@
 """Order Utility Classes. """
-from django.conf import settings
+from oscar.core.loading import get_model
 
 
 class OrderNumberGenerator(object):
@@ -23,7 +23,15 @@ class OrderNumberGenerator(object):
 
         """
         order_id = basket.id + self.OFFSET
-        order_number = u'{prefix}-{order_id}'.format(prefix=settings.ORDER_NUMBER_PREFIX, order_id=order_id)
+
+        # TODO add partner code instead of hard coded value
+        # after updating basket modal
+        # import here to avoid circular imports between 'order' and 'basket' apps
+        # Basket = get_model('basket', 'Basket')
+        # partner_code = Basket.objects.get(pk=basket.id).partner.short_code
+
+        partner_code = 'edx'
+        order_number = u'{partner_code}-{order_id}'.format(partner_code=partner_code, order_id=order_id)
 
         return order_number
 
@@ -38,7 +46,8 @@ class OrderNumberGenerator(object):
         Returns:
             int: The basket ID used to generate the provided order number.
         """
-        order_id = int(order_number.lstrip(u'{prefix}-'.format(prefix=settings.ORDER_NUMBER_PREFIX)))
+
+        order_id = int(order_number.rsplit('-', 1)[1])
         basket_id = order_id - self.OFFSET
 
         return basket_id
--- a/ecommerce/extensions/order/migrations/0010_auto_20150918_0529.py
+++ b/ecommerce/extensions/order/migrations/0010_auto_20150918_0529.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('partner', '0008_auto_20150914_1057'),
+        ('order', '0009_auto_20150709_1205'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='historicalorder',
+            name='partner',
+            field=models.ForeignKey(related_name='+', on_delete=django.db.models.deletion.DO_NOTHING, db_constraint=False, blank=True, to='partner.Partner', null=True),
+        ),
+        migrations.AddField(
+            model_name='order',
+            name='partner',
+            field=models.ForeignKey(related_name='orders', default=1, to='partner.Partner'),
+            preserve_default=False,
+        ),
+    ]
--- a/ecommerce/extensions/payment/tests/test_views.py
+++ b/ecommerce/extensions/payment/tests/test_views.py
@@ -218,7 +218,7 @@ class CybersourceNotifyViewTests(CybersourceMixin, PaymentEventsMixin, TestCase)
 
     def test_invalid_basket(self):
         """ When payment is accepted for a non-existent basket, log an error and record the response. """
-        order_number = '{}-{}'.format(settings.ORDER_NUMBER_PREFIX, 101986)
+        order_number = '{}-{}'.format('edx', 101986)
 
         notification = self.generate_notification(
             self.processor.secret_key,
--- a/ecommerce/settings/_oscar.py
+++ b/ecommerce/settings/_oscar.py
@@ -36,11 +36,6 @@ OSCAR_APPS = [
 ])
 # END APP CONFIGURATION
 
-
-# ORDER PROCESSING
-# Prefix appended to every newly created order number.
-ORDER_NUMBER_PREFIX = 'OSCR'
-
 # The initial status for an order, or an order line.
 OSCAR_INITIAL_ORDER_STATUS = ORDER.OPEN
 OSCAR_INITIAL_LINE_STATUS = LINE.OPEN


ecommerce/extensions/order/models.py
```diff
@@ -2,6 +2,7 @@
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 from oscar.apps.order.abstract_models import AbstractOrder, AbstractPaymentEvent, AbstractLine
+from oscar.apps.partner.models import Partner
```
zubair-arbi: Please order the imports

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def add_partner_and_update_order_number(apps, schema_editor):
+    """ Updating existing orders. Adding relation between order and partner.
+    """
+
+    Order = apps.get_model('order', 'Order')
+    Partner = apps.get_model('partner', 'Partner')
+
+    partner, __ = Partner.objects.get_or_create(code=PARTNER_CODE)
```
clintonb: You should NOT be creating a partner. All you need to do here is update `Order.partner` for existing orders.

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
```diff
@@ -0,0 +1,66 @@
...
+
+    orders = Order.objects.all()
+
+    # Update all existing basket with edx partner
+    orders.update(partner=partner)
```
clintonb: @feanil do we have a way to test a query that updates all 1M+ order rows?

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Update all orders to add PARTNER_CODE as a prefix to order numbers
+    for order in orders:
```
clintonb: The order numbers already have a prefix.

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
```diff
@@ -0,0 +1,66 @@
...
+        order.number = '{partner_code}-{order_number}'.format(partner_code=PARTNER_CODE,
+                                                              order_number=order.number.rsplit('-')[-1])
+        order.save()
+
+
+def remove_partner_and_update_order_number(apps, schema_editor):
+    # This method is for backward compatibility, It will remove partner reference_id
+    # and add ORDER_NUMBER_PREFIX to the order numbers
+
+    Order = apps.get_model('order', 'Order')
```
clintonb: This should be a no-op.

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
```diff
@@ -0,0 +1,66 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='historicalorder',
+            name='partner',
+            field=models.ForeignKey(related_name='+', on_delete=django.db.models.deletion.DO_NOTHING, db_constraint=False, blank=True, to='partner.Partner', null=True),
+        ),
+        migrations.AddField(
+            model_name='order',
+            name='partner',
+            field=models.ForeignKey(default=1, to='partner.Partner'),
```
clintonb: This is not a valid default. There should NOT be a default value.

ecommerce/extensions/order/models.py
```diff
@@ -1,14 +1,18 @@
 # noinspection PyUnresolvedReferences
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
-from oscar.apps.order.abstract_models import AbstractOrder, AbstractPaymentEvent, AbstractLine
```
clintonb, line 4: Revert this change. Imports should be should be grouped (native, third-party, project) and alphabetized.

ecommerce/extensions/order/models.py
```diff
@@ -1,14 +1,18 @@
...
+
 from simple_history.models import HistoricalRecords
 
+from oscar.apps.order.abstract_models import AbstractOrder, AbstractPaymentEvent, AbstractLine
+from oscar.apps.partner.models import Partner
```
clintonb: Do not import models in this manner. See https://django-oscar.readthedocs.org/en/releases-1.1/topics/class_loading_explained.html?highlight=get_class.

ecommerce/extensions/order/models.py
```diff
@@ -1,14 +1,18 @@
...
+
 from ecommerce.extensions.fulfillment.status import ORDER
 
 
 class Order(AbstractOrder):
     history = HistoricalRecords()
+    partner = models.ForeignKey(Partner)
```
clintonb: Use `partner.Partner`.

ecommerce/extensions/order/models.py
zubair-arbi, line 4: @clintonb the import order was changed as suggested by this guideline: https://github.com/edx/edx-platform/wiki/Python-Guidelines#imports-order

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
tasawernawaz: @clintonb Here we need to add partner code as prefix instead of our previous prefix 'OSCAR'.
Similarly for backward compatibility I am adding again 'OSCAR' and removing partner code.

clintonb:
ECOM-2232 (basket model updates) must be finished before work can be completed on the order model.

ecommerce/extensions/order/models.py
clintonb, line 4: That's a rather strange interpretation of PEP8. We aren't that strict. Django and Oscar are both third-party libraries and should be grouped together. I'll bring this up in a later architecture meeting.

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
clintonb: We don't use `OSCAR`. We have an override for that setting in stage and production, where it is set to `EDX`.

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
tasawernawaz: I am using here `get_or_create` method, because there must be at least one partner that will be referenced for existing orders.
Do we assume that there will be partner in Partner modal ? In that case we simple update orders.

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
tasawernawaz: @clintonb For backward compatibility I am adding 'OSCAR' and removing partner code. Do you think we should keep existing order numbers without updating them for partner code as prefix. In that case we can simply pass this method.

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
tasawernawaz: @clintonb Thanks. So we can remove forward and backward data migration for `order.numbers`.

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
clintonb: We already have an edX partner. Use `Partner.objects.get(id=1)`.

ecommerce/extensions/order/migrations/0010_auto_20150911_0905.py
feanil: Loadtest environment with a loadtest running against the ecommerce app.

clintonb: Thanks @feanil. @tasawernawaz let me know if you need help with this.

zubair-arbi:
@tasawernawaz change looks good. But we will need to wait for the PR: https://github.com/edx/ecommerce/pull/326 to merge first.

clintonb:
@tasawernawaz remember to squash your commits.

ecommerce/extensions/order/migrations/0011_auto_20150915_0336.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('order', '0010_auto_20150915_0332'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
clintonb: Merge this with the migration above.

ecommerce/extensions/order/models.py
```diff
@@ -9,6 +9,7 @@
 
 class Order(AbstractOrder):
     history = HistoricalRecords()
+    partner = models.ForeignKey('partner.Partner', related_name='partner')
```
clintonb: related_name should be `'orders'`.

ecommerce/extensions/order/tests/test_utils.py
```diff
@@ -22,8 +19,11 @@ def test_order_number_generation(self):
         first_order_number = OrderNumberGenerator().order_number(first_basket)
         second_order_number = OrderNumberGenerator().order_number(second_basket)
 
-        self.assertIn(self.ORDER_NUMBER_PREFIX, first_order_number)
-        self.assertIn(self.ORDER_NUMBER_PREFIX, second_order_number)
+        # TODO add partner code instead of hard coded value
```
clintonb, line 22: Do not merge until this is fixed.

ecommerce/extensions/order/utils.py
```diff
@@ -38,7 +46,10 @@ def basket_id(self, order_number):
         Returns:
             int: The basket ID used to generate the provided order number.
         """
-        order_id = int(order_number.lstrip(u'{prefix}-'.format(prefix=settings.ORDER_NUMBER_PREFIX)))
+
+        # here we are removing partner code from order number and converting back into integer
+        # for getting basket_id, e.g order_number = EDX-100001
+        order_id = int(order_number.rsplit('-')[-1])
```
clintonb: This will be slightly more legible if you use `[1]` as your index instead of `[-1]`.

ecommerce/extensions/order/utils.py
```diff
@@ -38,7 +46,10 @@ def basket_id(self, order_number):
         Returns:
             int: The basket ID used to generate the provided order number.
         """
-        order_id = int(order_number.lstrip(u'{prefix}-'.format(prefix=settings.ORDER_NUMBER_PREFIX)))
+
+        # here we are removing partner code from order number and converting back into integer
```
clintonb: The docstring already explains what's going on in this method. The extra comment is really necessary.

ecommerce/settings/local.py
```diff
@@ -12,9 +12,6 @@
 # DEBUG CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#debug
 DEBUG = True
-
-# See: https://docs.djangoproject.com/en/dev/ref/settings/#template-debug
-TEMPLATE_DEBUG = DEBUG
```
clintonb: Make sure you've rebased on master. This setting was removed sometime last week.

ecommerce/extensions/order/utils.py
tasawernawaz: I have used this because a partner code may contain '-', in that we case may get unexpected result.

clintonb: The partner's code should only be alphanumeric characters.

tasawernawaz: @clintonb I have confirmed this from #325. We have partner.short_code as Char field.

ecommerce/extensions/order/migrations/0010_auto_20150915_0332.py
```diff
@@ -0,0 +1,53 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='historicalorder',
+            name='partner',
+            field=models.ForeignKey(related_name='+', on_delete=django.db.models.deletion.DO_NOTHING, db_constraint=False, blank=True, to='partner.Partner', null=True),
+        ),
+        migrations.AddField(
+            model_name='order',
+            name='partner',
+            field=models.ForeignKey(related_name='orders', to='partner.Partner', null=True, blank=True),
```
clintonb: I was reminded that you can do all of the migrations in a single step. Delete this migration and re-run `makemigrations`. Specify 1 as the default.

PR #330, opened 2015-09-14
Add Non Nullable short_code field migrations

ahsan-ul-haq:
@clintonb please review

### DIFF
--- a/ecommerce/extensions/partner/models.py
+++ b/ecommerce/extensions/partner/models.py
@@ -11,7 +11,7 @@ class StockRecord(AbstractStockRecord):
 
 class Partner(AbstractPartner):
     # short_code is the unique identifier for the 'Partner'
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/partner/models.py
```diff
@@ -11,7 +11,7 @@ class StockRecord(AbstractStockRecord):
 
 class Partner(AbstractPartner):
     # short_code is the unique identifier for the 'Partner'
-    short_code = models.CharField(max_length=8, unique=True, null=True)
+    short_code = models.CharField(max_length=8, unique=True, null=False)
```
clintonb: Also `blank=False`

clintonb:
:+1: (once blank kwarg updated)

zubair-arbi:
lgtm :+1:

PR #336, opened 2015-09-15
Asynchronous order fulfillment

rlucioni:
Prepares Otto for asynchronous order fulfillment by initializing a Celery app on startup; asynchronous order fulfillment is controlled with a new setting. Also modifies the order fulfillment API endpoint to allow initial fulfillment attempts, not just retries. ECOM-2242.

Todo:
- [x] Address potential for race condition created by calling tasks before committing orders to the database.
- [x] Use `CELERY_ALWAYS_EAGER` during development to execute tasks locally (synchronously) instead of sending them to the queue.
- [x] Add ecommerce worker as a dependency, once available (preferably from PyPI).

Related:
- [x] Add new settings to the ecommerce Ansible role, and communicate with DevOps about secure/production values (https://github.com/edx/configuration/pull/2292).

@jimabramson @clintonb @feanil


rlucioni:
@jimabramson @clintonb @feanil this is ready for review. Uses the task from https://github.com/edx/ecommerce-worker/pull/1.

requirements/base.txt
```diff
@@ -22,3 +23,6 @@ suds==0.4
 
 # TODO https://github.com/chibisov/drf-extensions/pull/87
 git+https://github.com/edx/drf-extensions.git@regex-fix#egg=drf-extensions==0.2.7
+
+# TODO (RFL): Update once ecommerce-worker uploaded to PyPI.
+git+https://github.com/edx/ecommerce-worker.git@c0efe6d823e41a97a79732707ac649257a12610f#egg=ecommerce_worker
```
feanil: This should be pinned to a version.

clintonb: It should also be on PyPI.

rlucioni: Agreed. I'll take care of this once https://github.com/edx/ecommerce-worker/pull/1 is merged.

feanil:
:+1:

ecommerce/celery_app.py
```diff
@@ -0,0 +1,13 @@
...
+
+try:
+    from celery import Celery
+
+    # Set the default configuration module, if one is not aleady defined.
+    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'ecommerce.settings.local')
+
+    app = Celery('ecommerce')
+    app.config_from_object('django.conf:settings')
+except ImportError:
```
jimabramson: since celery and the worker app are in the base requirements, i'm confused why there's handling for ImportErrors.

rlucioni: Good catch. This is left over from when Celery was only a production requirement. Updated.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -59,6 +95,13 @@ def handle_successful_order(self, order):
...
 
-        post_checkout.send_robust(sender=self, order=order)
+        if waffle.switch_is_active('async_order_fulfillment'):
+            # Always commit transactions before sending tasks depending on state from the current transaction!
+            # There's potential for a race condition here if the task starts executing before the active
+            # transaction has been committed; the necessary order doesn't exist in the database yet.
+            # See http://celery.readthedocs.org/en/latest/userguide/tasks.html#database-transactions.
+            fulfill_order.delay(order.number)
+        else:
+            post_checkout.send_robust(sender=self, order=order)
```
jimabramson: i know it's not introduced in this PR, but i notice that the post_checkout signal gets called with `_robust` here, but without that in the `OrderFulfillView`.   Do you recall why?

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -143,19 +139,24 @@ def test_change_permissions_required(self):
         self.user.user_permissions.add(permission)
         self.assertNotEqual(403, self._put_to_view().status_code)
 
-    @ddt.data(ORDER.OPEN, ORDER.COMPLETE)
-    def test_order_fulfillment_error_state_required(self, order_status):
-        """ If the order is not in the Fulfillment Error state, the view must return an HTTP 406. """
+    @ddt.data(ORDER.COMPLETE)
```
jimabramson: might as well remove ddt here.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -48,6 +52,38 @@ def handle_payment(self, response, basket):
...
+                               shipping_method,
+                               shipping_charge,
+                               billing_address,
+                               order_total,
+                               **kwargs):
+        """
+        Place an order and mark the corresponding basket as submitted.
+
+        Differs from the superclass' method by not implicitly handling
+        successfully placed orders.
```
jimabramson: instead of removing the call to `handle_successful_order` in this override, i'm wondering if it might be a little cleaner to move the atomic wrapper around the place_order / submit calls in here.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -143,19 +139,24 @@ def test_change_permissions_required(self):
...
-        attempt to fulfill the order. The view should return HTTP 200.
+        If the user is authenticated/authorized, and the order is in the Open or Fulfillment Error
+        states, the view should attempt to fulfill the order. The view should return HTTP 200.
         """
-        self.assertEqual(ORDER.FULFILLMENT_ERROR, self.order.status)
+        self.order.status = order_status
+        self.order.save()
+
+        if order_status == ORDER.FULFILLMENT_ERROR:
+            self.order.lines.all().update(status=LINE.FULFILLMENT_CONFIGURATION_ERROR)
```
jimabramson: does this actually have any effect on the test?

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -176,7 +177,10 @@ def handle_shipping_event(order, _event_type, _lines, _line_quantities, **_kwarg
 
     def test_fulfillment_failed(self):
         """ If fulfillment fails, the view should return HTTP 500. """
-        self.assertEqual(ORDER.FULFILLMENT_ERROR, self.order.status)
+        self.order.status = ORDER.FULFILLMENT_ERROR
+        self.order.save()
+        self.order.lines.all().update(status=LINE.FULFILLMENT_CONFIGURATION_ERROR)
```
jimabramson: ditto

ecommerce/extensions/checkout/mixins.py
rlucioni: @jimabramson good question. It may simply have been an oversight. `send_robust()` guarantees that all receivers are notified of the emitted signal, preventing program flow from being interrupted by an exception in a receiver. However, it also swallows these exceptions, and we don't inspect the values returned by the receivers to check for errors.

@clintonb, you originally suggested that we use `send_robust()` here. Would you agree with reverting to `send()`? Being noisy about errors will make debugging easier for us. Plus, we only have a single receiver attached to the `post_checkout` signal at the moment, so I don't really see the benefit of `send_robust()`.

ecommerce/extensions/api/v2/tests/views/test_orders.py
rlucioni: Yup. Removed.

ecommerce/extensions/checkout/mixins.py
clintonb: Yes, that's fine.

ecommerce/extensions/checkout/mixins.py
rlucioni: Good call, updated.

ecommerce/extensions/checkout/mixins.py
rlucioni: Updated.

ecommerce/extensions/api/v2/tests/views/test_orders.py
rlucioni: Since fulfillment is mocked, no, it does not. Good catch. Removed.

ecommerce/extensions/api/v2/tests/views/test_orders.py
rlucioni: Not quite the same as the above, but since our ultimate goal here is to trigger a fulfillment failure, this is also not necessary. Removed.

ecommerce/extensions/analytics/utils.py
```diff
@@ -36,12 +36,15 @@ def parse_tracking_context(user):
     return user_tracking_id, lms_client_id
 
 
-def log_exceptions(msg):
-    """Log exceptions (avoiding clutter/indentation).
+def silence_exceptions(msg):
```
rlucioni, line 39: @jimabramson heads up. The motivation for this was the move away from `send_robust()`. We don't want non-essential receivers like analytics event emitters to interrupt critical program flow when they raise exceptions.

ecommerce/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Ensures that the initialized Celery app is loaded when Django starts.
+# Allows Celery tasks to bind themselves to an initialized instance of the Celery library.
+from ecommerce import celery_app
```
clintonb: Any reason why this cannot live in the AppConfig for the core app? See https://docs.djangoproject.com/en/1.8/ref/applications/#django.apps.AppConfig.ready.

ecommerce/extensions/api/v2/views.py
```diff
@@ -42,6 +44,10 @@ class BasketCreateView(EdxOrderPlacementMixin, generics.CreateAPIView):
     """
     permission_classes = (IsAuthenticated,)
 
+    @method_decorator(transaction.non_atomic_requests)
```
clintonb, line 51: What's going on in this class? You call non-atomic here; but, later call atomic.

jimabramson:
:+1:

ecommerce/__init__.py
rlucioni: Nope; good call. I forgot about AppConfig. Moved.

ecommerce/extensions/api/v2/views.py
rlucioni, line 51: Right. I first disable atomicity for the view. Otherwise, we'd be unable to commit to the database until the request had concluded; Django will refuse to commit when an `atomic()` block is active, since that would break atomicity. The `atomic()` handlers then restore atomicity at points where we are modifying data (baskets, then orders) to ensure that we don't leave the system in a dirty state in the event of an error.

clintonb, line 51: Please add that in a comment on the view and the method.

clintonb:
:+1:

ecommerce/extensions/api/v2/views.py
rlucioni, line 51: Done.

PR #338, opened 2015-09-16
Update SKU Generation

ahsan-ul-haq:
ECOM-2237
@zubair-arbi @awais786 @aamir-khan @tasawernawaz

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -112,8 +112,8 @@ class Course(models.Model):
 
         return name
 
-    def create_or_update_seat(self, certificate_type, id_verification_required, price, credit_provider=None,
-                              expires=None, credit_hours=None):
+    def create_or_update_seat(self, certificate_type, id_verification_required, price, partner,
+                              credit_provider=None, expires=None, credit_hours=None):
         """
         Creates course seat products.
 
@@ -191,8 +191,6 @@ class Course(models.Model):
 
         seat.save()
 
-        # TODO Expose via setting
-        partner = Partner.objects.get(code='edx')
         try:
             stock_record = StockRecord.objects.get(product=seat, partner=partner)
             logger.info(
@@ -201,7 +199,7 @@ class Course(models.Model):
                 course_id
             )
         except StockRecord.DoesNotExist:
-            partner_sku = generate_sku(seat)
+            partner_sku = generate_sku(seat, partner)
             stock_record = StockRecord(product=seat, partner=partner, partner_sku=partner_sku)
             logger.info(
                 'Course seat product stock record with certificate type [%s] for [%s] does not exist. '
--- a/ecommerce/courses/tests/test_publishers.py
+++ b/ecommerce/courses/tests/test_publishers.py
@@ -14,6 +14,7 @@ from ecommerce.courses.models import Course
 from ecommerce.courses.publishers import LMSPublisher
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.settings import get_lms_url
+from ecommerce.tests.mixins import PartnerMixin
 
 
 EDX_API_KEY = 'edx'
@@ -23,12 +24,13 @@ LOGGER_NAME = 'ecommerce.courses.publishers'
 
 @ddt.ddt
 @override_settings(COMMERCE_API_URL='http://example.com/commerce/api/v1/', EDX_API_KEY=EDX_API_KEY)
-class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
+class LMSPublisherTests(CourseCatalogTestMixin, PartnerMixin, TestCase):
     def setUp(self):
         super(LMSPublisherTests, self).setUp()
         self.course = G(Course)
-        self.course.create_or_update_seat('honor', False, 0)
-        self.course.create_or_update_seat('verified', True, 50)
+        self.partner = self.create_partner('edx')
+        self.course.create_or_update_seat('honor', False, 0, self.partner)
+        self.course.create_or_update_seat('verified', True, 50, self.partner)
         self.publisher = LMSPublisher()
         self.error_message = u'Failed to publish commerce data for {course_id} to LMS.'.format(
             course_id=self.course.id
@@ -184,7 +186,9 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
         Verify that (a) professional seats NEVER have an expiration date and (b) the name/mode is properly set for
         no-id-professional seats.
         """
-        seat = self.course.create_or_update_seat('professional', is_verified, 500, expires=datetime.datetime.utcnow())
+        seat = self.course.create_or_update_seat(
+            'professional', is_verified, 500, self.partner, expires=datetime.datetime.utcnow()
+        )
         stock_record = seat.stockrecords.first()
         actual = self.publisher.serialize_seat_for_commerce_api(seat)
         expected = {
@@ -207,7 +211,7 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
         Verify that course publication succeeds if the Credit API responds
         with 2xx status codes when publishing CreditCourse data to the LMS.
         """
-        self.course.create_or_update_seat('credit', True, 100, credit_provider='Harvard', credit_hours=1)
+        self.course.create_or_update_seat('credit', True, 100, self.partner, credit_provider='Harvard', credit_hours=1)
 
         self._mock_credit_api(creation_status, update_status)
         self._mock_commerce_api(commerce_status)
@@ -250,7 +254,7 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
         Verify that course publication fails if the Credit API does not respond
         with 2xx status codes when publishing CreditCourse data to the LMS.
         """
-        self.course.create_or_update_seat('credit', True, 100, credit_provider='Harvard', credit_hours=1)
+        self.course.create_or_update_seat('credit', True, 100, self.partner, credit_provider='Harvard', credit_hours=1)
 
         self._mock_credit_api(400, 418, error_message)
 
@@ -262,7 +266,7 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
         Verify that course publication fails if no access token is provided
         when publishing CreditCourse data to the LMS.
         """
-        self.course.create_or_update_seat('credit', True, 100, credit_provider='Harvard', credit_hours=1)
+        self.course.create_or_update_seat('credit', True, 100, self.partner, credit_provider='Harvard', credit_hours=1)
 
         response = self.publisher.publish(self.course, access_token=None)
         self.assertIsNotNone(response)
@@ -273,7 +277,7 @@ class LMSPublisherTests(CourseCatalogTestMixin, TestCase):
         Verify that course publication fails if an exception is raised
         while publishing CreditCourse data to the LMS.
         """
-        self.course.create_or_update_seat('credit', True, 100, credit_provider='Harvard', credit_hours=1)
+        self.course.create_or_update_seat('credit', True, 100, self.partner, credit_provider='Harvard', credit_hours=1)
 
         with mock.patch.object(LMSPublisher, '_publish_creditcourse') as mock_publish_creditcourse:
             mock_publish_creditcourse.side_effect = Exception(self.error_message)
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -20,6 +20,7 @@ BillingAddress = get_model('order', 'BillingAddress')
 Line = get_model('order', 'Line')
 Order = get_model('order', 'Order')
 Product = get_model('catalogue', 'Product')
+Partner = get_model('partner', 'Partner')
 ProductAttributeValue = get_model('catalogue', 'ProductAttributeValue')
 Refund = get_model('refund', 'Refund')
 Selector = get_class('partner.strategy', 'Selector')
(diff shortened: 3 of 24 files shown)


ecommerce/courses/models.py
```diff
@@ -192,7 +192,7 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
         seat.save()
 
         # TODO Expose via setting
```
clintonb: Remove this comment.

ecommerce/extensions/catalogue/utils.py
```diff
@@ -1,7 +1,7 @@
 from hashlib import md5
 
 
-def generate_sku(product):
+def generate_sku(product, prefix=None):
```
clintonb: Update this to accept a Partner model, and make the prefix `partner.short_code.upper()`.

clintonb:
@jimabramson @rlucioni now's the chance to fix SKU generation.

ecommerce/extensions/catalogue/tests/test_utils.py
```diff
@@ -14,9 +18,10 @@ def test_generate_sku_for_course_seat(self):
         course = Course.objects.create(id=course_id, name='Test Course')
         certificate_type = 'honor'
         product = course.create_or_update_seat(certificate_type, False, 0)
+        partner, _ = Partner.objects.get_or_create(short_code='edx', name='edx')
 
         _hash = u'{} {} {} {}'.format(certificate_type, course_id, 'False', '')
         _hash = md5(_hash.lower()).hexdigest()[-7:]
-        expected = _hash.upper()
-        actual = generate_sku(product)
+        expected = partner.short_code.upper() + '-' + _hash.upper()
```
zubair-arbi: nit: add comment:

``` python
# verify that generated sku has partner 'short_code' as prefix
```

zubair-arbi:
Just one small nit, otherwise :+1:

ecommerce/courses/models.py
```diff
@@ -113,7 +113,7 @@ def _get_course_seat_name(self, certificate_type, id_verification_required):
         return name
 
     def create_or_update_seat(self, certificate_type, id_verification_required, price, credit_provider=None,
-                              expires=None, credit_hours=None):
+                              expires=None, credit_hours=None, partner_short_code="edx"):
```
rlucioni: Avoid hardcoding references to edX into our software. Consider replacing this default with the lower-cased `ORDER_NUMBER_PREFIX` defined in settings.

ecommerce/extensions/catalogue/utils.py
```diff
@@ -17,5 +17,6 @@ def generate_sku(product):
     ))
     _hash = md5(_hash.lower())
     _hash = _hash.hexdigest()[-7:]
+    sku = partner.short_code.upper() + '-' + _hash.upper()
```
rlucioni: Why not include the partner code in the hash? I'm nervous about including human-readable data in our SKUs. It makes it easy for future developers to "do the wrong thing" and parse data directly from the SKU.

ecommerce/courses/models.py
clintonb: I agree that we should not hardcode "edx". Instead pass the partner in as an argument. The `ORDER_NUMBER_PREFIX` setting should be considered deprecated.

rlucioni: @ahsan-ul-haq please add a comment in the base Oscar settings module marking `ORDER_NUMBER_PREFIX` as deprecated.

ahsan-ul-haq: @clintonb just to confirm one thing that the partner would be required to pass here and could not have any default value right?

clintonb: That is correct. We should be identifying the partner earlier (e.g. from the site/site configuration).

ahsan-ul-haq:
@clintonb @rlucioni please review

ecommerce/courses/tests/test_publishers.py
```diff
@@ -9,6 +9,7 @@
 import mock
 from requests import Timeout
 from testfixtures import LogCapture
+from oscar.core.loading import get_model
```
clintonb: Alphabetize.

ecommerce/extensions/api/v2/tests/views/test_publication.py
```diff
@@ -108,6 +112,16 @@ def setUp(self):
         self.error_message = u'Failed to publish commerce data for [{course_id}] to LMS.'.format(
             course_id=self.course_id
         )
+        self.partner, __ = Partner.objects.get_or_create(short_code='edx', name='edx')
+        self.site, __ = Site.objects.get_or_create(domain='example.com')
+
+        SiteConfiguration.objects.create(
```
clintonb: Where is this used?

ecommerce/extensions/api/v2/views.py
```diff
@@ -558,10 +561,15 @@ def _save_and_publish(self, data, course_id=None):
...
-            created, failure, message = serializer.save()
+            created, failure, message = serializer.save(partner)
             if failure:
                 return Response({'error': message}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
             else:
                 content = serializer.data
                 content['message'] = message if message else None
                 return Response(content, status=status.HTTP_201_CREATED if created else status.HTTP_200_OK)
+
+    def _get_partner_for_site(self, request):
```
clintonb: Move this to a new module, `ecommerce.extensions.partner.shortcuts`, so that it can be used elsewhere. Also, add tests.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -141,13 +143,16 @@ def _retrieve_data_from_lms(self, access_token):
 
     def _get_products(self, modes):
         """ Creates/updates course seat products. """
+        partner, __ = Partner.objects.get_or_create(short_code='edx', name='edx')
```
clintonb: Use a partner passed in from the command line. The only place where the edX partner should be used is in tests.

ecommerce/extensions/api/v2/views.py
```diff
@@ -558,10 +561,15 @@ def _save_and_publish(self, data, course_id=None):
         serializer = self.get_serializer(data=data)
         is_valid = serializer.is_valid(raise_exception=True)
         if is_valid:
-            created, failure, message = serializer.save()
+            created, failure, message = serializer.save(partner)
```
clintonb: It would be ideal if you passed the partner on the serializer's constructor instead of adding arguments to `save()`.

ecommerce/extensions/catalogue/tests/test_utils.py
```diff
@@ -1,22 +1,28 @@
...
         """Verify the method generates a SKU for a course seat."""
         course_id = 'sku/test/course'
         course = Course.objects.create(id=course_id, name='Test Course')
+        partner, _ = Partner.objects.get_or_create(short_code='edx', name='edx')
         certificate_type = 'honor'
-        product = course.create_or_update_seat(certificate_type, False, 0)
+        product = course.create_or_update_seat(certificate_type, False, 0, partner)
 
-        _hash = u'{} {} {} {}'.format(certificate_type, course_id, 'False', '')
+        _hash = u'{} {} {} {} {}'.format(certificate_type, course_id, 'False', '', partner.short_code)
```
clintonb: Use `partner.id` as it is guaranteed to never change.

ecommerce/extensions/catalogue/utils.py
```diff
@@ -13,7 +13,8 @@ def generate_sku(product):
         getattr(product.attr, 'certificate_type', ''),
         product.attr.course_key,
         unicode(product.attr.id_verification_required),
-        getattr(product.attr, 'credit_provider', '')
+        getattr(product.attr, 'credit_provider', ''),
+        partner.short_code
```
clintonb: `partner.id`

clintonb:
For future reference: If you find yourself duplicating a lot of test code to create a model, make a factory instead. We have access to both [factory_boy](https://factoryboy.readthedocs.org/en/latest/) and [django-dynamic-fixture](http://django-dynamic-fixture.readthedocs.org/en/latest/index.html) for this purpose.

ahsan-ul-haq:

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -165,20 +170,30 @@ class Command(BaseCommand):
                     default=False,
                     help='Save the migrated data to the database. If this is not set, '
                          'migrated data will NOT be saved to the database.'),
+        make_option('--partner_short_code',
+                    action='store',
+                    dest='partner_short_code',
+                    default=None,
+                    help='Partner\'s short code that provides the course'),
```
rlucioni: "Short code for the partner providing the course."

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -165,20 +170,30 @@ class Command(BaseCommand):
...
     )
 
     def handle(self, *args, **options):
         course_ids = args
         access_token = options.get('access_token')
+        partne_short_code = options.get('partner_short_code')
```
rlucioni: This should be `partner_short_code`.

ahsan-ul-haq:
@rlucioni @clintonb made the suggested changes please review

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -165,20 +170,30 @@ class Command(BaseCommand):
...
     def handle(self, *args, **options):
         course_ids = args
         access_token = options.get('access_token')
+        partner_short_code = options.get('partner_short_code')
         if not access_token:
             logger.error('Courses cannot be migrated if no access token is supplied.')
             return
 
+        if not partner_short_code:
+            logger.error('Courses cannot be migrated if partner\'s short code is not provided.')
```
rlucioni: `Courses cannot be migrated without providing a partner short code.`

ecommerce/extensions/catalogue/tests/test_migrate_course.py
```diff
@@ -286,9 +291,35 @@ def test_handle_with_commit(self):
...
 
             # Verify that the migrated course was published back to the LMS
             self.assertTrue(mock_publish.called)
 
         self.assert_course_migrated()
+
+    @httpretty.activate
+    def test_handle_with_no_partner(self):
+        """ Verify the management command didn't run if partner's short code
+        is not provided.
```
rlucioni: `Verify the management command does not run if no partner short code is provided`.

ecommerce/extensions/api/serializers.py
```diff
@@ -167,6 +168,7 @@ def __init__(self, *args, **kwargs):
         super(AtomicPublicationSerializer, self).__init__(*args, **kwargs)
 
         self.access_token = kwargs['context'].pop('access_token')
+        self.partner = kwargs['context'].pop('partner')
```
rlucioni: Without updating the CAT, this will raise a `KeyError`. Make this more robust by providing a default value (i.e., `None`), then validating the partner in a separate validation method (e.g., `validate_partner`). If the partner is invalid, raise a `ValidationError`.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -165,20 +170,30 @@ class Command(BaseCommand):
                     default=False,
                     help='Save the migrated data to the database. If this is not set, '
                          'migrated data will NOT be saved to the database.'),
+        make_option('--partner_short_code',
```
clintonb: Trim this down to just `partner`. You can call it short_code internally.

ecommerce/extensions/api/serializers.py
rlucioni: This brings a more serious concern to light, that being that without updating the CAT first, this change will break course publication. The CAT doesn't currently pass a partner code to the server. 

We can get around this issue in the short-term by defaulting to the first partner in the system when no partner code is provided. You can do this in your custom validation method.

ahsan-ul-haq:
@clintonb code coverage for the project is failing but the coverage for my changes/patch is 100%. what should I do to pass the build

clintonb:
@ahsan-ul-haq See [the suggestions](https://codecov.io/github/edx/ecommerce/features/suggestions?ref=49e3f848cf6f23e8cfba681bb8037e7229eee1a8). The simplest change would be to increase the coverage in `migrate_course.py`.

Note that you did nothing wrong given that your patch is at 100% coverage. You can opt to ignore the project coverage; but, the team benefits from maintaining a high level of coverage.

ecommerce/tests/mixins.py
```diff
@@ -217,3 +218,11 @@ def assert_correct_event_payload(self, instance, event_payload, order_number, cu
         else:
             # Payload validation is currently limited to order and refund events
             self.fail()
+
+
+class PartnerMixin(object):
+    """Provides utility methods for creating partners in test cases."""
+    def create_partner(self, short_code, name=None):
+        """Create Partner object and return it"""
+        partner, __ = Partner.objects.get_or_create(short_code=short_code, name=name)
```
clintonb: If name is `None`, it should match the short code.

clintonb: You could also simplify the `create_partner` method and simply take a `name` argument. Use the one argument to populate both fields.

clintonb:
:+1: (once mixin updated)

ecommerce/extensions/api/serializers.py
```diff
@@ -189,6 +191,14 @@ def validate_products(self, products):
 
         return products
 
+    def validate_partner(self):
```
rlucioni: This is the correct way to perform field-level [validation](http://www.django-rest-framework.org/api-guide/serializers/#validation) when using DRF. However, since `partner` isn't a serializer field, it would be more appropriate to name this method `get_partner`.

PR #340, opened 2015-09-17
Displaying provider name instead of ID on credit checkout page 

clintonb:
ECOM-2172

@rlucioni @jimabramson

### DIFF
--- a/ecommerce/static/templates/credit_provider_details.html
+++ b/ecommerce/static/templates/credit_provider_details.html
@@ -2,7 +2,7 @@
     <img alt="<%= display_name %>" class="img-responsive provider-image" src="<%= thumbnail_url %>">
 </div>
 <div class="col-sm-10">
-    <h3 class="title"><%= interpolate(gettext('About Credit at %(provider)s'), {provider:id}, true) %></h3>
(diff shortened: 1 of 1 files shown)


rlucioni:
:+1:

jimabramson:
:+1:  no objections, but as an aside, curious why the text needs to be upper'd.

clintonb:
@jimabramson 
screencapture-ecommerce-stage-edx-org-credit-checkout-course-v1-3aasux-2bcreditdemo-2bdemo_course-1442501728182

wedaly:
Why are we showing the provider ID here rather than the provider's display name?

jimabramson:
@clintonb thanks for the visual explanation :-)

clintonb:
@wedaly I'm just fixing an error I saw. Product is out, and this needs to hit stage before 1 PM. Any preference on ID vs name? I can defend either.

wedaly:
@clintonb If it's not additional work, I'd prefer the name.  Uppercasing the ID could bite us down the road if we add an additional provider.

clintonb:
Agreed. Consider it done.
\

clintonb:
Updated to use display name instead of ID.

rlucioni:
:+1:

PR #347, opened 2015-09-18
Removed unnecessary migration

clintonb:
We should no longer create partners and products in the migration. This should be managed in Django admin and the Course Administration Tool.

ECOM-2235

@jimabramson @rlucioni @zubair-arbi @ahsan-ul-haq


rlucioni:
What about the manual migrations in the `catalogue` app? They're responsible for creation of the product whose stockrecord was created by these migrations.

clintonb:
@rlucioni good catch. Updated.

rlucioni:
:+1:

PR #348, opened 2015-09-18
Removed migration hack for tests

clintonb:
Prior to Django 1.8 we had a hack in place to prevent migrations from being run for tests. While this decreases test run times, it also prevents us from testing our migrations and ensuring all model changes have associated migrations. This has bitten us in the past and even resulted in an outage. We can live with the extra test time if it means greater protection from outages.

Closes #308

@rlucioni @jimabramson

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,14 +50,14 @@ validate_js:
 
 validate_python: clean
 	python manage.py compress --settings=ecommerce.settings.test -v0
-	DISABLE_MIGRATIONS=True coverage run --branch --source=ecommerce ./manage.py test ecommerce \
+	REUSE_DB=1 coverage run --branch --source=ecommerce ./manage.py test ecommerce \
 	--settings=ecommerce.settings.test --with-ignore-docstrings --logging-level=DEBUG
 	coverage report
 	make quality
 
 fast_validate_python: clean
 	python manage.py compress --settings=ecommerce.settings.test -v0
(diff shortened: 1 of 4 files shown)


jimabramson:
is there any usefulness to leaving `DisableMigrations` around so that it can be manually set by the developer in the env?

clintonb:
@jimabramson leaving it would cut down on time, but also leaves devs open to committing without testing migrations. The migrations run in less than 30 seconds, which I think is a bit long. We can trim this time (once all the multi-tenancy models are complete) by [squashing migrations](https://docs.djangoproject.com/en/1.8/topics/migrations/#squashing-migrations).

jimabramson:
@clintonb what i meant was to keep the proposed changes to the Makefile, so migrations are not disabled by default, but allow the ability to override.  The use case I have in mind is when working on one or a few specific tests and running them repeatedly ala TDD; I'd hate to have to wait 30 seconds for every run of a 1-second test, so I'd want to be able to cut that step out.   If the migrations only run when executing the complete suite (I'm starting to suspect that this is the case) then I'm fine with proceeding with this change.

rlucioni:
@clintonb what about taking advantage of `REUSE_DB=1`, as we do in the IDA cookiecutter?

clintonb:
Added `REUSE_DB=1`

rlucioni:
:+1: 

@clintonb if you have access to our codecov config, it might be worth dropping the required coverage to 95% so we don't get otherwise fine builds failing on us.

PR #349, opened 2015-09-21
Using Codecov for coverage reporting

clintonb:


clintonb:
FYI @jimabramson @rlucioni

rlucioni:
@clintonb some context for those less familiar with the motivations behind this change would be nice.

clintonb:
We are trying out codecov in place of coveralls.

rlucioni:
@clintonb do we want to continue using coveralls in the meantime for the other repos we own?

clintonb:
Yes.

jimabramson:
good call.

PR #354, opened 2015-09-23
Removed devstack overrides

clintonb:
- Retrieving base settings from configuration
- Allowing for local overrides in private.py

### DIFF
--- a/ecommerce/settings/devstack.py
+++ b/ecommerce/settings/devstack.py
@@ -7,8 +7,7 @@ from ecommerce.settings import get_lms_url
 from ecommerce.settings.base import *
 from ecommerce.settings.logger import get_logger_config
 
-
-LOGGING = get_logger_config()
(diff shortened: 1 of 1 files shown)


clintonb:
@feanil @jibsheet 

FYI @rlucioni @jimabramson @bderusha @peter-fogg

rlucioni:
@clintonb what's the impetus for removing the JWT and API key overrides? The way it's currently structured allows these settings to be configured by using the configuration file on disk.

clintonb:
@rlucioni `JWT_AUTH` is specified in configuration so there is no need for it to be updated/redefined. If overrides are necessary they should be in `private.py`.

rlucioni:
Got it, thanks. Good catch on the logging configuration, too. :+1:

feanil:
:+1:

PR #355, opened 2015-09-23
Removed docs compilation from Travis

clintonb:
The docs compilation breaks coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,11 +24,13 @@ install:
     - pip install -r requirements/docs.txt
     - make requirements.js
 script:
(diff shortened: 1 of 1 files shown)


clintonb:
FYI @rlucioni @jimabramson @bderusha @peter-fogg

rlucioni:
:+1:

peter-fogg:
:+1:, but why does it break coverage?

clintonb:
@peter-fogg I don't know. However, I have updated the commit to do docs compilation before validation. I suspect something in docs compilation is deleting the coverage output. Let's see if that is true.

PR #356, opened 2015-09-24
Added script to create publish_course_modes_to_lms switch

clintonb:
This switch is needed for a proper integration with LMS. Creating it via migration means one less step for setup.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_courses.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_courses.py
@@ -5,9 +5,9 @@ from django.core.urlresolvers import reverse
 from django.test import TestCase
 import mock
 from oscar.core.loading import get_model, get_class
-from waffle.models import Switch
 
 from ecommerce.core.constants import ISO_8601_FORMAT
+from ecommerce.core.tests import toggle_switch
 from ecommerce.courses.models import Course
 from ecommerce.courses.publishers import LMSPublisher
 from ecommerce.extensions.api.v2.tests.views import JSON_CONTENT_TYPE, ProductSerializerMixin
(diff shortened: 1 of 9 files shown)


clintonb:
@rlucioni @bderusha @peter-fogg @jimabramson

peter-fogg:
:+1:

ecommerce/extensions/catalogue/tests/test_migrate_course.py
```diff
@@ -130,7 +130,9 @@ def assert_lms_api_headers(self, request, bearer=False):
 class MigratedCourseTests(CourseMigrationTestMixin, TestCase):
     def setUp(self):
         super(MigratedCourseTests, self).setUp()
-        Switch.objects.get_or_create(name='publish_course_modes_to_lms', active=True)
+        switch = Switch.objects.get_or_create(name='publish_course_modes_to_lms', defaults={'active': True})
```
rlucioni: A tuple is currently being assigned to `switch`, which is causing your test failures. You want `switch, __ = ...`.

ecommerce/extensions/api/v2/tests/views/test_courses.py
```diff
@@ -141,7 +141,11 @@ def test_publish(self):
         path = reverse('api:v2:course-publish', kwargs={'pk': course_id})
 
         # Method should return a 500 if the switch is inactive
-        switch, _created = Switch.objects.get_or_create(name='publish_course_modes_to_lms', active=False)
+        switch, _created = Switch.objects.get_or_create(name='publish_course_modes_to_lms',
```
rlucioni: To be consistent, please use `__` instead of `_created`.

PR #357, opened 2015-09-25
Correct OAuth and and Celery settings for devstack.

peter-fogg:
@renzo @clintonb @rock345 @bderusha


clintonb:
The OAUTH and JWT settings were removed so as not to conflict with configuration. Why are they being restored?

peter-fogg:
@clintonb I didn't realize they'd been removed at any point. I wasn't able to get devstack working without them.

clintonb:
ecommerce and configuration have been updated and these overrides are not
necessary. Either reprovision or add these changes to your private.py file.

## 

Clinton Blackburn

PR #358, opened 2015-09-25
Use JWT auth for ecommerce workers.

peter-fogg:
~~This isn't ready yet; it (at least) needs to be updated with the correct ecommerce worker version. That can wait until https://github.com/edx/ecommerce-worker/pull/4 has merged. I'm just putting this up for initial review.~~

@rlucioni @clintonb @jimabramson @rock345 @bderusha


ecommerce/extensions/api/v2/views.py
```diff
@@ -384,11 +384,18 @@ class OrderByBasketRetrieveView(OrderRetrieveView):
...
-    queryset = Order.objects.all()
     serializer_class = serializers.OrderSerializer
 
+    def get_queryset(self):
+        """
+        Returns a queryset consisting of only the authenticated user's orders.
+
+        This ensures we do not allow one user to fulfill another's orders.
+        """
+        return self.request.user.orders
```
rlucioni: If the user attached to the request is a superuser, return a queryset containing all orders:

``` python
def get_queryset(self):
    user = self.request.user
    qs = user.orders

    if user.is_superuser:
        qs = Order.objects.all()

    return qs
```

rlucioni: Context: this endpoint is also used when attempting to retry fulfillment of orders on behalf of our users. Anyone with superuser permissions should be able to fulfill any other user's order.

peter-fogg:
@rlucioni Added the superuser check.

jimabramson:
@peter-fogg @rlucioni looking at this in code, I am concerned about breaking the design principle we established for sticking with scalar, immutable arguments in messages.

Can we reconsider the concept of a system user instead?  It seems to me this could be done fairly cleanly:
- a data migration in the ecommerce webapp creates a user with hardcoded username (e.g. 'system') and superuser bit set
- instead of the worker generating a JWT based on user info passed in the message, it generates a JWT with the 'system' username.

I don't see any need to synchronize additional credentials across systems/configs under this approach; we also still get to use JWT instead of adding a different authentication mechanism.  Am I missing some downside or additional effort needed to go this route?

rlucioni:
@jimabramson the downside/additional effort is the need to create a service user on ecommerce and add additional configuration to the worker app. Is the concern that the the email address or full name might change while a task is sitting on the queue? If so, I'd also back the service user approach.

peter-fogg:
@jimabramson I'd experimented with this approach briefly last week, and it will work well enough. I agree with @rlucioni that the need to create/configure a service user is a small downside, but it's not a huge burden.

jimabramson:
@rlucioni the additional concern is that we want to keep task signatures simple and immutable, and passing dictionaries around creates a major backdoor there.  this is somewhat related to the original concern where we're passing a blob of user data that may be stale by the time it's being processed, and introducing id tokens into the system based on non-authoritative data.  I know I'm being kind of a hardliner here - it's not so much that these are unacceptable risks for the present use case but want to avoid establishing too lenient a precedent for how similar stuff gets built in the future.

jimabramson:
@peter-fogg regarding the service user, is there more to it than inserting a row into the auth user table with a hard-coded / reserved username, via migrations?

rlucioni:
@jimabramson that makes sense. I think it's worth noting that in this case, if the full name or email address for a user were stale by the time they were processed, the user's name or email address would actually be changed on the ecommerce service. We clearly don't want that, so I can see the appeal of the service user.

peter-fogg:
@jimabramson @rlucioni I've switched over to the service user approach now.

rlucioni:
@peter-fogg still a WIP?

peter-fogg:
@rlucioni I'm going to leave the WIP tag up until the worker side of things merges and the tests can run against the correct version of that repo.

jimabramson:
might not hurt to namespace the username, e.g. 'edx-service'.   otherwise looks good to me.

ecommerce/settings/base.py
```diff
@@ -312,6 +312,9 @@
     'JWT_ISSUER': None
 }
 
+# Service user for worker processes.
+SERVICE_USERNAME = 'service'
```
rlucioni: Could you make this more explicit? `ECOMMERCE_WORKER_SERVICE_USERNAME` and `ecommerce_worker_service`, for example.

rlucioni:
@jimabramson @peter-fogg Does it make sense for the service user to have a password set on their account?

ecommerce/extensions/checkout/migrations/0001_add_service_user.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
rlucioni: I think it might be more appropriate to include this migration in the `core` app.

jimabramson:
@rlucioni I don't envision using password authentication with this user, but I'm in favor of generating something at random to ensure it can't be guessed, or simply disabling password auth for the user if that is possible somehow.

rlucioni:
@jimabramson @peter-fogg if disabling password auth is the right thing to do for the service user, we can take advantage of [`set_unusable_password()`](https://docs.djangoproject.com/en/1.8/ref/contrib/auth/#django.contrib.auth.models.User.set_unusable_password).

peter-fogg:
Made those changes.

ecommerce/core/migrations/0006_add_service_user.py
```diff
@@ -0,0 +1,29 @@
...
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    def add_service_user(apps, schema_editor):
+        User = apps.get_model('checkout', 'User')
```
rlucioni: Use `get_user_model()` from `django.contrib.auth` to get the `User` model.

ecommerce/settings/base.py
```diff
@@ -312,6 +312,9 @@
     'JWT_ISSUER': None
 }
 
+# Service user for worker processes.
+ECOMMERCE_SERVICE_WORKER_USERNAME = 'edx-ecommerce-service-worker'
```
rlucioni: `edx` shouldn't be hardcoded into anything that might end up being used by someone in our open-source community. I think `ecommerce_worker` (same as the string used by the ecommerce worker to identify itself) is probably sufficient.

ecommerce/core/migrations/0006_add_service_user.py
peter-fogg: To my knowledge the reason for using `apps.get_model` is that it will be sure to pull in the right version of the class in case of migrations changing the data. Presumably the user model won't change often, but this seems like the correct way to do it anyway.

ecommerce/core/migrations/0006_add_service_user.py
```diff
@@ -0,0 +1,29 @@
...
+        service_user = User.objects.create(
+            username=settings.SERVICE_USERNAME,
```
rlucioni: Don't forget to update this with the new variable name from settings.

rlucioni:
@peter-fogg @jimabramson given the way we replicate users from the LMS on the ecommerce service, what if someone happens to register a user with the same username as the ecommerce worker's service user, then places an order? I don't think this would break anything, but I wanted to raise it as a concern.

peter-fogg:
@rlucioni I suppose we could register a user on edx.org with that username. To my knowledge there's no way to programmatically create a user with a username that can't be created by a user, so without reserving the name for ourself we can't guarantee that we'll avoid that situation.

jimabramson:
i think reserving the username in the LMS makes sense.  You could basically use the exact same migration as the one here.  One caveat though, rather than having to do this repeatedly, maybe it would make more sense to use a more generic username like @peter-fogg initially proposed, so it could be reused systemwide.

rlucioni:
Using a migration like this on edx-platform makes sense. I think it's more reasonable than expecting our open source users to manually reserve service users on their instances.

With regards to a generic username: my understanding is that we should prefer more specific users tied to individual services, since doing so makes us more flexible in the event that we need to limit permissions for one of these services.

ecommerce/core/migrations/0006_add_service_user.py
```diff
@@ -0,0 +1,29 @@
...
+        service_user = User.objects.create(
+            username=settings.ECOMMERCE_SERVICE_WORKER_USERNAME,
+            is_superuser=True
+        )
+        service_user.set_unusable_password()
+        service_user.save()
+
+    def remove_service_user(apps, schema_editor):
+        User = apps.get_model('checkout', 'User')
+        User.objects.get(username=settings.SERVICE_USERNAME).delete()
```
rlucioni: Update this, as well.

ecommerce/core/migrations/0006_add_service_user.py
rlucioni: In that case, it seems like the `app_label` should be `core`, not `checkout`.

jimabramson:
@rlucioni your argument about the service-specific user, rather than a generic one, makes sense to me.  we can always use additional migrations to provision additional reserved platform users in the future.

rlucioni:
@peter-fogg since we're going with the service user approach, something else to account for here is rate limiting. We currently use DRF's default [UserRateThrottle](http://www.django-rest-framework.org/api-guide/throttling/#userratethrottle) to limit the rate of requests against our endpoints to 40/minute. I think we'll want to exempt superusers (or just specific service users) from that rate limit.

peter-fogg:
@rlucioni I've pushed up some fixes, as well as a new throttling class to allow the service user through. I'm only using for the specific view that we need now and thus keeping it scoped to the api app, but we could move it around and add it to other classes later if it's necessary.

ecommerce/core/migrations/0006_add_service_user.py
peter-fogg: Turns out that using `apps.get_model` doesn't return a class with the correct methods (it's missing `set_unusable_password`), so I've switched to using `get_user_model`.

ecommerce/extensions/api/v2/views.py
```diff
@@ -14,6 +14,7 @@
 from rest_framework_extensions.mixins import NestedViewSetMixin
 import waffle
 
+from .throttles import ServiceUserThrottle
```
rlucioni: Could you bump the `throttles` module up one level, to the `api` directory? Like `authentication` and `exceptions`, the contents of `throttles` don't need to be versioned.

rlucioni:
Looking good @peter-fogg, thanks for making those changes. Have you been able to look into the handling of multiple issuers?

ecommerce/extensions/api/v2/views.py
peter-fogg: Done.

jimabramson:
@peter-fogg changes LGTM too but seems coverage needs to be placated.

peter-fogg:
@rlucioni @jimabramson I've pushed up the work to handle multiple issuers. You can the meat of it [here](https://github.com/edx/ecommerce/pull/358/files#diff-58d19e84277bc98f830d9fd1d5049c8dR8) -- I'm overriding the function which is responsible for decoding the payload and verifying the issuer, trying each configured issuer in turn.

ecommerce/settings/test.py
```diff
@@ -12,7 +12,7 @@
 
 TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
 
-LOGGING = get_logger_config(debug=DEBUG, dev_env=True, local_loglevel='DEBUG')
+# LOGGING = get_logger_config(debug=DEBUG, dev_env=True, local_loglevel='DEBUG')
```
rlucioni: Restore this.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -94,13 +94,14 @@ def test_super_user(self):
...
 @ddt.ddt
+@override_settings(ECOMMERCE_SERVICE_WORKER_USERNAME='test-service-user')
 class OrderFulfillViewTests(UserMixin, TestCase):
 
     def setUp(self):
         super(OrderFulfillViewTests, self).setUp()
         ShippingEventType.objects.get_or_create(name=SHIPPING_EVENT_NAME)
 
-        self.user = self.create_user(is_superuser=True)
+        self.user = self.create_user(is_superuser=True, username='test-service-user')
```
jimabramson, line 106: under the async model is there a configuration/mode in which fulfillments can still run in process?  if so, do these tests need to cover both service user and original end user?

ecommerce/settings/test.py
```diff
@@ -50,6 +50,8 @@
 ENABLE_AUTO_AUTH = True
 
 JWT_AUTH['JWT_SECRET_KEY'] = 'insecure-secret-key'
+
+JWT_ISSUERS = (None,)
```
rlucioni: I don't think you need to perform this override. The empty tuple from base should be fine, considering you override this setting in tests where you need it.

ecommerce/settings/base.py
```diff
@@ -309,9 +309,14 @@
     'JWT_SECRET_KEY': None,
     'JWT_ALGORITHM': 'HS256',
     'JWT_VERIFY_EXPIRATION': True,
-    'JWT_ISSUER': None
+    'JWT_DECODE_HANDLER': 'ecommerce.extensions.api.handlers.jwt_decode_handler',
 }
 
+JWT_ISSUERS = ()
```
rlucioni, line 315: Please open a PR against the configuration repo adding this setting to the ecommerce role.

ecommerce/extensions/api/v2/tests/views/test_orders.py
rlucioni, line 106: @jimabramson it is possible to configure fulfillment such that it continues to be run in process. However, this test covers the order fulfillment endpoint, which an original end user will never contact.

@peter-fogg is it necessary to override the service user username for this test? It seems like any superuser should work just fine for the purposes of the test.

ecommerce/settings/test.py
peter-fogg: It actually is necessary -- the empty tuple will cause any JWT-protected endpoint to raise an `InvalidIssuerError`, but `None` signals to not check the issuer on any JWTs in tests.

ecommerce/extensions/api/v2/tests/views/test_orders.py
peter-fogg, line 106: It's not necessary for the test to pass, but we need coverage of `allow_request` in `ecommerce/extensions/api/throttles.py`, so it makes sense to use the superuser there.

ecommerce/settings/test.py
peter-fogg: Whoops, done.

ecommerce/settings/test.py
clintonb: Set a fake issuer. We will always have an issuer, so should our tests.

ecommerce/extensions/api/v2/tests/views/test_orders.py
rlucioni, line 106: @peter-fogg got it. Would you mind adding a comment to that effect where you're overriding the service user's username? It'll make the purpose of the override clearer to future maintainers.

ecommerce/settings/base.py
peter-fogg, line 315: https://github.com/edx/configuration/pull/2319

ecommerce/settings/test.py
peter-fogg: Done.

ecommerce/extensions/api/v2/tests/views/test_orders.py
peter-fogg, line 106: Done.

ecommerce/extensions/api/handlers.py
```diff
@@ -0,0 +1,37 @@
...
+def jwt_decode_handler(token):
+    """Attempt to decode the given token with each of the configured JWT issuers.
+
+    Args:
+        token (str): the JWT to decode.
+
+    Returns:
+        dict: The JWT's payload.
+
+    Raises:
```
rlucioni, line 17: Looks like this is incomplete.

```
Raises:
    InvalidIssuerError, if the issuer in the provided token doesn't match a known issuer.
```

rlucioni:
@peter-fogg just one last note about a docstring, otherwise :+1:.

jimabramson:
:+1:

rlucioni:
:fireworks:

ecommerce/core/migrations/0006_add_service_user.py
```diff
@@ -0,0 +1,31 @@
...
+
+User = get_user_model()
+
+
+class Migration(migrations.Migration):
+
+    def add_service_user(apps, schema_editor):
+        service_user = User.objects.create(
+            username=settings.ECOMMERCE_SERVICE_WORKER_USERNAME,
+            is_superuser=True
```
clintonb, line 17: This user should have permissions to fulfill orders, but should not be a superuser.

PR #359, opened 2015-09-25
Improve the acceptance testing experience

rlucioni:
I got fed up with the state of our acceptance tests while trying to see if asynchronous order fulfillment had broken them. This PR cleans up and updates our acceptance testing documentation, gives environment variables better names, fixes an issue preventing LMS registration and enrollment from being tested locally, and makes testing of the CyberSource payment flow optional. Prelude to ECOM-2248.

@clintonb @jimabramson @bderusha @peter-fogg

PR #360, opened 2015-09-26
Make acceptance tests compatible with async fulfillment

rlucioni:
Acceptance tests check a configurable number of times for an order's completion before failing. Review https://github.com/edx/ecommerce/pull/359 first. ECOM-2248.

@clintonb @jimabramson @bderusha @peter-fogg


peter-fogg:
:+1:, though you'll either need to just merge this PR and not #359, or do some rebasing.

PR #362, opened 2015-09-28
Move acceptance testing documentation to the docs directory

rlucioni:
@clintonb as requested.


clintonb:
Don't have the time/attention for a complete review of the content; but, I assume it's copied from the old location. Regardless, :+1:, and thanks for responding so quickly!

rlucioni:
No problem!

PR #363, opened 2015-09-29
Clean up README

rlucioni:
Create new docs sections for asset pipeline and email notifications, extend existing sections with content from README.

@clintonb


clintonb:
:+1:

PR #368, opened 2015-10-02
Add IP to Segment tracking calls

bderusha:
Depends on the following PR https://github.com/edx/edx-platform/pull/9979/files#diff-f17399a99e052ec3802476d5a5508f15R15 where `lms_ip` is added the to context manager passed to ecom.

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -19,7 +19,7 @@ def parse_tracking_context(user):
         user (User): An instance of the User model.
 
     Returns:
-        Tuple of strings, user_tracking_id and lms_client_id
+        Tuple of strings: user_tracking_id, lms_client_id, lms_ip
     """
     tracking_context = user.tracking_context or {}
 
@@ -32,8 +32,9 @@ def parse_tracking_context(user):
         user_tracking_id = 'ecommerce-{}'.format(user.id)
 
     lms_client_id = tracking_context.get('lms_client_id')
+    lms_ip = tracking_context.get('lms_ip')
 
-    return user_tracking_id, lms_client_id
+    return user_tracking_id, lms_client_id, lms_ip
 
 
 def silence_exceptions(msg):
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -13,7 +13,7 @@ def is_segment_configured():
 
 
 def parse_tracking_context(user):
-    """Extract user and client IDs from a user's tracking context.
+    """Extract user ID, client ID, and IP address from a user's tracking context.
 
     Arguments:
         user (User): An instance of the User model.
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -26,7 +26,7 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
     if not (is_segment_configured() and order.total_excl_tax > 0):
         return
 
-    user_tracking_id, lms_client_id = parse_tracking_context(order.user)
+    user_tracking_id, lms_client_id, lms_ip = parse_tracking_context(order.user)
 
     analytics.track(
         user_tracking_id,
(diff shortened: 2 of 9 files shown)


bderusha:
Ready for review @jimabramson @clintonb @rlucioni @peter-fogg 

FYI @rock345

ecommerce/extensions/analytics/utils.py
```diff
@@ -19,7 +19,7 @@ def parse_tracking_context(user):
         user (User): An instance of the User model.
```
rlucioni, line 19: Could you update the docstring above to include mention of the IP?

rlucioni:
Looking good @bderusha. The failing tests inspect the context passed to Segment API calls. Since you've added extra context that's not present in the test data, the tests are failing.

We don't need to solve this as part of this PR, but if you're able to find a quick way to include the user's IP in the fulfillment call made to the Enrollment API, we'd take care of [ECOM-2175](https://openedx.atlassian.net/browse/ECOM-2175). However, solving that might require additional work on the LMS.

bderusha:
This now has tests and is ready for final review.

/CC @peter-fogg @rlucioni @clintonb @jimabramson

ecommerce/courses/tests/test_publishers.py
```diff
@@ -122,14 +122,20 @@ def test_api_bad_status(self, status, error_msg, expected_msg):
             self.assert_response_message(response, expected_msg)
 
     @httpretty.activate
-    @ddt.data(200, 201)
-    def test_api_success(self, status):
+    @ddt.data(
+        (200, None),
+        (200, '11.22.33.44'),
+        (201, None),
+        (201, '11.22.33.44')
```
rlucioni: This is a Cartesian product! You can do this using `itertools`:

``` python
@ddt.data(*itertools.product(('200', '201'), (None, '11.22.33.44')))
@ddt.unpack
```

ecommerce/extensions/analytics/utils.py
```diff
@@ -13,13 +13,13 @@ def is_segment_configured():
 
 
 def parse_tracking_context(user):
-    """Extract user and client IDs from a user's tracking context.
+    """Extract user, client IDs, and IP address from a user's tracking context.
```
rlucioni: `"""Extract user ID, client ID, and IP address...`

rlucioni:
@bderusha I agree with the changes you've made to include the client IP in requests to the Enrollment API. That said, I don't think it's necessary to do this for requests issued during Course publication. We're only after the IP in cases where enrollments are being made on behalf of students. Is there a reason to include the IP in _all_ requests to the LMS?

bderusha:
@rlucioni I guess my thinking was that I don't want to run into an issue like this again.  There is an API endpoint in ecom that will then make a request to publish a course in lms, why wouldn't we want to know the originator of that request?  If no context is passed from LMS (or whatever service is hitting that endpoint) then it will work as it works today, but if we are passed a context we should honor it.

jimabramson:
I'm a little concerned about establishing a pattern of appending individual function arguments all through the call stack like this.  One thing I'd be more comfortable with would be passing all of the tracking context as a dictionary, since that would at least allow for changing the payload over time without having to modify all the functions through which it's getting passed.  An even better but admittedly more complex solution would be to use a middleware to tie this context to the request and fetch it only when needed (exactly as is done in edx-platform for this use case,  e.g. see https://github.com/edx/edx-platform/blob/master/lms/djangoapps/commerce/__init__.py)

jimabramson:
@bderusha OTOH, to @rlucioni's point, I'd also double-check the necessity for tracking context in this case.  I have very little context (no pun intended) here, but BI tracking, I would assume, is really not interested in the administrative actions of internal users.  For the purposes of auditing and diagnostics I think straight-up logging is the way to go.

bderusha:
@rlucioni @jimabramson I've removed the IP passing from the course publish paths. Ready for re-review.

rlucioni:
:+1:

bderusha:
@jimabramson based on our hipchat conversation I've made a change to move the context parsing into the `_post_to_enrollment_api` rather than doing it twice outside the method and passing the contents in. please have a look

jimabramson:
:+1:  once tests fixed

PR #371, opened 2015-10-03
Compressing JavaScript with Closure Compiler

clintonb:
Closure Compiler is slower, but results in a 27% decrease in the size (340KB to 246KB) of our JS output.

ECOM-2468


clintonb:
Compiled output doesn't load properly. Closing for now.

PR #375, opened 2015-10-05
Quieted noisy logging statement

clintonb:
The majority of our course seats will not have a credit provier set. This logging statement is only useful when developing for credit locally. It's severity has been appropriately decreased.

ECOM-2477

FYI @rlucioni @jimabramson @peter-fogg @bderusha @cpennington


rlucioni:
:+1:

PR #376, opened 2015-10-05
Add migration creating a Waffle switch controlling async order fulfillment

rlucioni:
The switch defaults to False. Saves time when bringing up a fresh ecommerce instance.

@clintonb @jimabramson @peter-fogg @bderusha

### DIFF
--- a/ecommerce/core/migrations/0007_auto_20151005_1333.py
+++ b/ecommerce/core/migrations/0007_auto_20151005_1333.py
@@ -5,20 +5,20 @@ from django.db import migrations
 
 
 def create_switch(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


peter-fogg:
:+1:

jimabramson:
@rlucioni should probably implement a reverse for the migration.

clintonb:
Yes, reverse migrations are needed. Please correct the issue for the publication switch as well. (Sorry.)

clintonb:
:+1: (once reverse migration added)

rlucioni:
@jimabramson good catch. Reverse operations added.

jimabramson:
:+1:

PR #377, opened 2015-10-06
Remove unnecessary setting from base settings.

rlucioni:
djangorestframework-jwt uses HS256 by default for cryptographic signing.

@clintonb @peter-fogg @jimabramson @bderusha


clintonb:
What if that default changes? It costs us nothing to leave this in place and be explicit about how we are signing.

rlucioni:
Fair. I'll extend my configuration PR to include this setting.

PR #380, opened 2015-10-07
Enrollmentcodes app

vkaracic:


clintonb:
@vkaracic please do not merge code without review. Additionally, @mjfrey is already working on a design for enrollment codes that properly uses Oscar's models. This implementation does not and will be reverted.

vkaracic:
@clintonb We are part of @mjfrey 's team and this is the feature branch which we are using to create a proof of concept that we are tasked to do. 
This is related to: https://openedx.atlassian.net/browse/SOL-1178
Additionally, this is a work in progress.

clintonb:
So sorry about that. Please carry on.

## 

Clinton Blackburn

PR #384, opened 2015-10-07
Upgraded to django-nose 1.4.2

clintonb:


clintonb:
FYI @rlucioni @jimabramson

PR #392, opened 2015-10-13
Creating new migration with setting default value to 1

awais786:
This is the 2nd phase of migration and follows previous work https://github.com/edx/ecommerce/pull/326 to add `partner` to model `basket`.
- Set the `default` to value `1` and also set `null=False`.

_**Note:** This need to rebase with master after the previous work https://github.com/edx/ecommerce/pull/326 is merged._

### DIFF
--- a/ecommerce/extensions/basket/migrations/0007_auto_20151013_1209.py
+++ b/ecommerce/extensions/basket/migrations/0007_auto_20151013_1209.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('basket', '0006_basket_partner'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='basket',
+            name='partner',
+            field=models.ForeignKey(related_name='baskets', default=1, to='partner.Partner'),
+        ),
+    ]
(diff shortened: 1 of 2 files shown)


awais786:
@clintonb This 2nd migration sets the default value to `1` for all existing partners with value `None`.
fyi: @rlucioni @cpennington 
- After running first migration for field `partner` and model `basket` from https://github.com/edx/ecommerce/pull/326

```
mysql> select count(*) from basket_basket where partner_id is  NULL;
+----------+
| count(*) |
+----------+
|  2110922 |
+----------+
1 row in set (3.96 sec)
```
- After running this second migration for field `partner` and model `basket` (after the first migration for partner)

```
mysql> select count(*) from basket_basket where partner_id is  NULL;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.00 sec)

select count(*) from basket_basket where partner_id = 1;
mysql> select count(*) from basket_basket where partner_id =1;
+----------+
| count(*) |
+----------+
|  2111138 |
+----------+
1 row in set (0.44 sec)
```

clintonb:
This should be done via SQL so that we can batch the updates and avoid locking the table. The migrations in #326 should set the default value.

awais786:
@clintonb after setting the default=1 it gives error with old code. That's why we set default=None.
https://gist.github.com/zubair-arbi/7b08cdf81d0c990936df

If possible would you please try to check this at ur machine or sandbox (mysql only).

clintonb:
@awais786 what did you do to debug?

awais786:
@clintonb when the migration executes with default=1 it does not set the default value at schema level. So when new order places with old code than mysql throws error 

```
    raise errorclass, errorvalue
OperationalError: (1364, "Field 'partner_id' doesn't have a default value")
```

clintonb:
@awais786 I understand that you see an error; but, what have you done to fix it? Have you looked at the docs or Stack Overflow?

PR #395, opened 2015-10-16
Updated dashboard orders list view

clintonb:
Removing the display and querying of address values drastically improves query speed.

ECOM-2501


clintonb:
@edx/ecommerce

rlucioni:
Nice, thanks for taking care of this! :100:

PR #396, opened 2015-10-16
ECOM-2521: Added the support to check for the eligibility from LMS on…

aamir-khan:
… on checkout page

@zubair-arbi @awais786 @tasawernawaz @ahsan-ul-haq Please review this PR.


ecommerce/settings/base.py
```diff
@@ -217,6 +217,10 @@
 # URL to which enrollment requests should be made
 ENROLLMENT_API_URL = None
 
+# URL to check the user eligibility on LMS.
+ELIGIBILITY_API_URL = None
```
clintonb: `CREDIT_API_URL`

ecommerce/credit/views.py
```diff
@@ -75,4 +80,46 @@ def get(self, request, *args, **kwargs):
...
+            True if the user is eligible otherwise False
+        """
+        is_eligible = False
+        eligibility_api_url = settings.ELIGIBILITY_API_URL + '?username={}&course_key={}'.format(
+            user.username, course_key
+        )
+        timeout = settings.ELIGIBILITY_FULFILLMENT_TIMEOUT
+        headers = {
+            'Content-Type': 'application/json',
+            'X-Edx-Api-Key': settings.EDX_API_KEY
```
clintonb, line 108: Please discontinue the use of the edX API key. We should be using JWT or OAuth2. If the API doesn't support either of these, update the API to comply with our API guidelines.

ecommerce/credit/views.py
```diff
@@ -75,4 +80,46 @@ def get(self, request, *args, **kwargs):
...
+        }
+
+        __, client_id, ip = parse_tracking_context(user)
```
clintonb: None of this tracking info is necessary for an API call.

ecommerce/credit/tests/test_views.py
```diff
@@ -63,8 +94,44 @@ def test_get_with_disabled_flag(self):
...
+        """
+        Test checkout page accessibility. Page will return 404 if user eligibility deadline has
+        expired for given course.
+        """
+        httpretty.register_uri(httpretty.GET, settings.ELIGIBILITY_API_URL, status=200, content_type=JSON, body='[]')
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 404)
+
+    @httpretty.activate
+    @ddt.data(500, 404, 408)
```
clintonb: Raise an error to simulate a timeout. The HTTP status code will never be seen. See http://docs.python-requests.org/en/latest/user/quickstart/#timeouts for more info.

ecommerce/credit/tests/test_views.py
```diff
@@ -63,8 +94,44 @@ def test_get_with_disabled_flag(self):
 
         self.assertEqual(response.status_code, 404)
 
+    @httpretty.activate
+    def test_get_with_no_eligibility(self):
+        """
+        Test checkout page accessibility. Page will return 404 if user is not eligible for the
```
clintonb: Who made the decision to show a 404? Our 404 page hasn't been built for Otto.

@marcotuts any reason we aren't displaying a helpful error message? I guarantee this will come back to bite us when we need to debug. At a minimum, if we don't display the message to the user, we need to log why the 404 was raised.

ecommerce/credit/views.py
```diff
@@ -75,4 +80,46 @@ def get(self, request, *args, **kwargs):
...
+
+        if client_id:
+            headers['X-Edx-Ga-Client-Id'] = client_id
+
+        if ip:
+            headers['X-Forwarded-For'] = ip
+
+        response = requests.get(eligibility_api_url, headers=headers, timeout=timeout)
```
clintonb: Add logging statements to this method:
1. `Retrieving credit eligibility for user [%s] and course [%s]...`
2. `User [%s] IS eligible for credit for course [%s].`
3. `User [%s] is NOT eligible for credit for course [%s].`

ecommerce/credit/views.py
```diff
@@ -39,7 +43,8 @@ def get_context_data(self, **kwargs):
                 processors_dict[processor] = 'Checkout with {}'.format(processor)
 
         credit_seats = [
-            seat for seat in course.seat_products if getattr(seat.attr, 'certificate_type', '') == self.CREDIT_MODE
```
clintonb, line 42: Why the change?

aamir-khan:
@clintonb @rlucioni In previous implementations there was an ajax call that check the eligibility once the checkout page has loaded. But now I am making a server to server call to check the eligibility for the given course before rendering the checkout page. Should I remove the old ajax call because because we don't need it now as we are checking before rendering the page?

ecommerce/credit/tests/test_views.py
zubair-arbi: Please update docstring to reflect the changes in response.

ecommerce/credit/tests/test_views.py
```diff
@@ -63,8 +87,45 @@ def test_get_with_disabled_flag(self):
...
+        httpretty.register_uri(httpretty.GET, settings.CREDIT_API_URL, status=200, content_type=JSON, body='[]')
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, "You are not eligible to buy this course.")
+
+    @httpretty.activate
+    @ddt.data(500, 404)
+    def test_get_with_eligibility_api_failure(self, status):
+        """
+        Test checkout page for credit eligibility. Page will return 404 if the credit eligibility
```
zubair-arbi: Please update docstring to reflect the changes in response.

ecommerce/credit/tests/test_views.py
```diff
@@ -63,8 +87,45 @@ def test_get_with_disabled_flag(self):
...
+        """
+        body = json.dumps([{"course_key": unicode(self.course.id), "deadline": self.get_past_date()}])
+        httpretty.register_uri(httpretty.GET, settings.CREDIT_API_URL, status=status, content_type=JSON, body=body)
+        response = self.client.get(self.path)
+        self.assertContains(response, "You are not eligible to buy this course.")
+
+    @mock.patch('requests.get', mock.Mock(side_effect=Timeout))
+    def test_get_with_eligibility_api_timeout(self):
+        """
+        Test checkout page for credit eligibility. Page will return 404 if the credit eligibility
```
zubair-arbi: Please update docstring to reflect the changes in response.

clintonb:
@aamir-khan make as few calls as possible.

ecommerce/credit/views.py
```diff
@@ -75,4 +83,50 @@ def get(self, request, *args, **kwargs):
...
         if not waffle.switch_is_active('ENABLE_CREDIT_APP'):
             raise Http404
+
+        is_eligible, expired = self._check_credit_eligibility(request.user, kwargs.get('course_id'))
+        kwargs["is_eligible"] = is_eligible
+
         return super(Checkout, self).get(request, args, **kwargs)
+
+    def _check_credit_eligibility(self, user, course_key):
+        """ Check is the user is eligible for credit.
```
zubair-arbi: nit: `Check that the user is eligible for credit.`

ecommerce/credit/tests/test_views.py
marcotuts: I don't know that an explicit decision or request to handle this as a 404 was made necessarily. My understanding is that users won't get to this page (excepting manual url matching & triggering a hidden button's action) so it doesn't seem critical to me to add the 404 page to this story.

@clintonb -
If I'm mischaracterizing this or missing something let me know. My tentative suggestion is to exclude any 404 page from this story.

ecommerce/settings/local.py
```diff
@@ -99,6 +99,9 @@
 ENROLLMENT_API_URL = get_lms_url('/api/enrollment/v1/enrollment')
 ENROLLMENT_FULFILLMENT_TIMEOUT = 15  # devstack is slow!
 
+CREDIT_API_URL = get_lms_url('/api/credit/v1/eligibility/')
+CREDIT_FULFILLMENT_TIMEOUT = 15  # devstack is slow!
```
rlucioni: These settings should be moved to the section containing URL configuration.

ecommerce/settings/test.py
```diff
@@ -43,8 +43,11 @@
 LMS_DASHBOARD_URL = get_lms_url('/dashboard')
 
 COMMERCE_API_URL = get_lms_url('/api/commerce/v1/')
+
+CREDIT_API_URL = get_lms_url('/api/credit/v1/eligibility/')
 # END URL CONFIGURATION
 
+CREDIT_FULFILLMENT_TIMEOUT = 15
```
rlucioni, line 50: Minor: please be consistent with the way you've presented this in other settings modules.

ecommerce/credit/views.py
```diff
@@ -75,4 +83,58 @@ def get(self, request, *args, **kwargs):
...
+            course_key(str): The course Key
+        Returns:
+            True if the user is eligible otherwise False
+        """
+        logger.info("Retrieving credit eligibility for user [%s] and course [%s]", user.username, course_key)
+
+        is_eligible = False
+        deadline = None
+        api_failure = False
+        eligibility_api_url = settings.CREDIT_API_URL + '?username={}&course_key={}'.format(
```
rlucioni: Don't construct this URL by hand. Requests allows you to specify [`params`](http://docs.python-requests.org/en/latest/user/quickstart/#passing-parameters-in-urls).

ecommerce/credit/views.py
```diff
@@ -75,4 +83,58 @@ def get(self, request, *args, **kwargs):
         """Get method for checkout page."""
         if not waffle.switch_is_active('ENABLE_CREDIT_APP'):
             raise Http404
+
+        is_eligible, deadline, api_failure = self._check_credit_eligibility(request.user, kwargs.get('course_id'))
+        kwargs["is_eligible"] = is_eligible
```
rlucioni: Please stick to single quotes.

ecommerce/credit/views.py
```diff
@@ -75,4 +83,58 @@ def get(self, request, *args, **kwargs):
...
+                    deadline = eligibilities[0]['deadline']
+                    deadline = parse(deadline)
+                    logger.info("User [%s] is eligible for credit for course [%s].", user.username, course_key)
+                else:
+                    logger.info("User [%s] is not eligible for course [%s].", user.username, course_key)
+            else:
+                api_failure = True
+                logger.error("Credit API request failed for user [%s] for course [%s].", user.username, course_key)
+        except ConnectionError:
+            logger.error(
```
rlucioni: Use `logger.exception` to log the exception frame.

ecommerce/credit/views.py
```diff
@@ -75,4 +83,58 @@ def get(self, request, *args, **kwargs):
...
+                "Credit API request failed due to network error for user [%s] for course [%s].",
+                user.username, course_key
+            )
+            api_failure = True
+        except Timeout:
+            logger.error("Credit API request timeout for user [%s] for course [%s].", user.username, course_key)
```
rlucioni: Use `logger.exception` to log the exception frame.

ecommerce/credit/views.py
```diff
@@ -75,4 +83,58 @@ def get(self, request, *args, **kwargs):
...
+            'X-Edx-Api-Key': settings.EDX_API_KEY
+        }
+
+        try:
+            response = requests.get(eligibility_api_url, headers=headers, timeout=timeout)
+            if response.status_code == 200:
+                eligibilities = response.json()
+                if eligibilities:
+                    is_eligible = True
+                    deadline = eligibilities[0]['deadline']
```
rlucioni: Is it safe to indiscriminately access the first element in the list like this? @clintonb do you have context here?

clintonb: The length should be checked first.

ecommerce/credit/views.py
```diff
@@ -75,4 +83,58 @@ def get(self, request, *args, **kwargs):
...
+            user(User): User object for which checking the eligibility.
+            course_key(str): The course Key
+        Returns:
+            True if the user is eligible otherwise False
+        """
+        logger.info("Retrieving credit eligibility for user [%s] and course [%s]", user.username, course_key)
+
+        is_eligible = False
+        deadline = None
+        api_failure = False
```
rlucioni: I don't think you need this variable. Instead, you can leave `is_eligible` or `deadline` as `None` to signify a failure.

ecommerce/templates/edx/credit/checkout.html
```diff
@@ -58,12 +68,13 @@ <h3 class="title">
                     <div class="col-sm-12">
                         {# date will be added with ajax call #}
                         <small>
-                            {% blocktrans with date='<span class="deadline-date"></span>' %}
+                            {% blocktrans with date=deadline %}
```
zubair-arbi, line 71: We might need to use [moment.js](https://github.com/moment/moment) for date time internationalization.
Reference: https://groups.google.com/forum/#!topic/openedx-translation/M9ijURi0pOI

ecommerce/credit/views.py
aamir-khan, line 42: There was a pylint indentation warning on this line and fixed that.

ecommerce/credit/views.py
aamir-khan: There will always be one record if user is eligible because user will be eligible for once in a given course. The length will automatically checked in above check `if eligibilities:` if the list is empty then above check will fail.

ecommerce/templates/edx/credit/checkout.html
aamir-khan, line 71: This is rendered by the python backend and moment.js is used for front end rendering.

ecommerce/credit/views.py
rlucioni: @aamir-khan are there cases in which there might be more than one record? I know @clintonb dealt with an issue recently where an assumption like this bit us. Since `eligibilities` is a list, it can presumably contain more than one element. Is it possible (or necessary) to check that the retrieved eligibility object is the one we want?

aamir-khan: As per current implementation there is no chance for more than one record as the API query the `CreditEligibility` model for the given `course_key` and `username` and the model has uniqueness constraint `unique_together = ('username', 'course')` but to be on safe side I can check the list to match the `course_key` of the response with the requested params.

rlucioni: @aamir-khan I think being defensive would be a good idea. At a minimum, please add your note above as a comment here to avoid this confusion in the future.

ecommerce/credit/views.py
aamir-khan, line 108: @clintonb Can you please give me any reference where this JWT or OAuth2 authentication is being used for LMS API. I think on LMS JWT authentication is not implemented and I am not suer about OAuth2 authentication in LMS.

ecommerce/credit/views.py
```diff
@@ -75,4 +82,54 @@ def get(self, request, *args, **kwargs):
...
+            True if the user is eligible otherwise False
+        """
+        logger.info("Retrieving credit eligibility for user [%s] and course [%s]", user.username, course_key)
+
+        is_eligible = None
+        deadline = None
+        params = {'username': user.username, 'course_key': course_key}
+        headers = {
+            'Content-Type': 'application/json',
+            'X-Edx-Api-Key': settings.EDX_API_KEY
```
clintonb, line 108: :disappointed: none of the Credit API endpoints is secure. This needs to be fixed first. FYI @marcotuts @rock345.

ecommerce/credit/views.py
aamir-khan: I added the validation for the api response to check the returned data has the requested `course_key` in the list.

ecommerce/credit/views.py
aamir-khan, line 108: @clintonb For the authentication you have created a ticket [ECOM-2609](https://openedx.atlassian.net/browse/ECOM-2609) so the authentication will be added in separate PR?

ecommerce/credit/views.py
clintonb, line 108: @aamir-khan not sure why my original comment is hidden. Authentication/authorization needs to be fixed _before_ this change is merged. Otherwise, we will have to coordinate the releases of platform and ecommerce to make sure we don't have a period during which this page is broken.

PR #397, opened 2015-10-16
Log the repr() of response content when CreditCourse publication fails

rlucioni:
@edx/ecommerce


clintonb:
Please add a comment to the commit message and/or code explaining this change.

clintonb:
:+1:

jimabramson:
nitpick: square brackets are not going to be needed to demarcate the value of the body.

otherwise :+1:    thanks for squeezing this in

rlucioni:
@clintonb @jimabramson I've made both suggested changes. Thanks for the feedback.

jimabramson:
:+1:

PR #398, opened 2015-10-20
Updated Makefile for easier testing

clintonb:
- Removed static asset compression from the validate target
- Forcing compression when calling the static target (so that the target can be used locally)
- Added the static target to the help text

ECOM-2594

FYI @edx/ecommerce

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ addons:
 before_install:
     - "export DISPLAY=:99.0"
     - "sh -e /etc/init.d/xvfb start"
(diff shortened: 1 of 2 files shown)


cpennington:
I notice in the build that the asset compression happens both during `make static` and `make validate`. I'm not sure how long that step takes, but getting rid of that duplication might help build speed.

clintonb:
:facepunch: I'll fix that.

rlucioni:
@clintonb did removing the duplicated asset compression step give us another significant speed-up?

clintonb:
@rlucioni Compression takes about 10-15 seconds, so we should see that drop in our build times.

PR #399, opened 2015-10-20
Squashed migrations for multiple apps

clintonb:
Migration time dropped from `1m22.536s` to `0m53.173s`.
Test time dropped from `3m16.765s` to `2m38.807s`.

ECOM-2594

FYI @edx/ecommerce


rlucioni:
:+1:

Per the Django docs, the recommended process is to squash, keeping the old migrations, commit and release, wait until all systems are upgraded with the new release, and then remove the old files, commit and do a second release. @clintonb can you create a cleanup ticket to delete the old migrations? There shouldn't be a need to keep them around once we release these to production.

PR #400, opened 2015-10-20
Associating baskets with sites

clintonb:
ECOM-2232

@edx/ecommerce I noticed yesterday, while squashing migrations, that orders already have a foreign key to the `Site` model. This means we can skip adding a foreign key to the `Partner` model since every site links to a partner. I have decided to carry this linkage to the `Basket` model and link it to a site instead of a partner.

### DIFF
--- a/ecommerce/extensions/basket/models.py
+++ b/ecommerce/extensions/basket/models.py
@@ -1,14 +1,19 @@
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
 from oscar.apps.basket.abstract_models import AbstractBasket
 from oscar.core.loading import get_class
-
 
 OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
 
 
 class Basket(AbstractBasket):
+    site = models.ForeignKey('sites.Site', verbose_name=_("Site"), null=True, blank=True, default=None,
+                             on_delete=models.SET_NULL)
+
     @property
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/basket/models.py
```diff
@@ -1,14 +1,18 @@
...
+from django.utils.translation import ugettext_lazy as _
 from oscar.apps.basket.abstract_models import AbstractBasket
 from oscar.core.loading import get_class
 
-
 OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
 
 
 class Basket(AbstractBasket):
+    site = models.ForeignKey('sites.Site', verbose_name=_("Site"), null=True, on_delete=models.SET_NULL)
```
rlucioni: Would it be appropriate to set a default value for `site`?

clintonb: No. In a multi-tenant situation, there is no default. The code will updated
later to actually set a value when creating new baskets.

rlucioni:
@clintonb I'm concerned about locking the baskets table to add this new column. This seems to carry the same risks as adding a foreign key to the `Partner` model.

clintonb:
Yep, that's why we are load testing.

rlucioni:
Got it, thanks. :+1: pending results from the load test.

PR #403, opened 2015-10-23
Pretty print PaymentProcessorResponses in the Django admin

rlucioni:
I finally figured out how to do this in a minimally invasive way!

@clintonb @edx/ecommerce


clintonb:
:+1:

cpennington:
:+1:

jimabramson:
nice :+1:

PR #404, opened 2015-10-23
WAIT-LMS: Updated CreditCourse creation

clintonb:
The credit endpoint has been updated to use PUT to create new CreditCourse objects. The publisher has been updated to use this method, greatly simplifying the code necessary to create these objects.

ECOM-2524

### DIFF
--- a/ecommerce/courses/publishers.py
+++ b/ecommerce/courses/publishers.py
@@ -1,13 +1,15 @@
+from __future__ import unicode_literals
 import json
 import logging
-import requests
 
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
+from edx_rest_api_client.client import EdxRestApiClient
+from edx_rest_api_client.exceptions import SlumberHttpBaseException
+import requests
 
 from ecommerce.courses.utils import mode_for_seat
 from ecommerce.settings import get_lms_url
-
 
 logger = logging.getLogger(__name__)
 
@@ -37,31 +39,19 @@ class LMSPublisher(object):
 
     def _publish_creditcourse(self, course_id, access_token):
         """Creates or updates a CreditCourse object on the LMS."""
-        url = get_lms_url('api/credit/v1/courses/')
+
+        api = EdxRestApiClient(
+            get_lms_url('api/credit/v1/'),
+            oauth_access_token=access_token,
+            timeout=self.timeout
+        )
 
         data = {
             'course_key': course_id,
             'enabled': True
         }
 
-        headers = {
-            'Content-Type': 'application/json',
-            'Authorization': 'Bearer ' + access_token
-        }
-
-        kwargs = {
-            'url': url,
-            'data': json.dumps(data),
-            'headers': headers,
-            'timeout': self.timeout
-        }
-        response = requests.post(**kwargs)
-        if response.status_code == 400:
-            # The CreditCourse already exists. Try updating it.
-            kwargs['url'] += course_id.strip('/') + '/'
-            response = requests.put(**kwargs)
-
-        return response
+        api.courses(course_id).put(data)
 
     def publish(self, course, access_token=None):
         """ Publish course commerce data to LMS.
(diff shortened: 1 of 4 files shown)


clintonb:
@edx/ecommerce

rlucioni:
Looks like unit tests need to be updated.

ecommerce/courses/publishers.py
```diff
@@ -94,31 +89,17 @@ def publish(self, course, access_token=None):
...
-                        )
-
-                        return self._parse_error(response, error_message)
-                except Exception:  # pylint: disable=broad-except
-                    logger.exception(u'Failed to publish CreditCourse for [%s] to LMS.', course_id)
+            try:
+                if self._publish_creditcourse(course_id, access_token):
+                    logger.info(u'Successfully published CreditCourse for [%s] to LMS.', course_id)
+                else:
+                    # Note that %r is used to log the repr() of the response content, which may sometimes
```
peter-fogg: Is this comment still relevant? Looks like the response content isn't getting logged anymore, and I don't see `%r`.

ecommerce/courses/publishers.py
```diff
@@ -37,31 +39,24 @@ def serialize_seat_for_commerce_api(self, seat):
...
-            'headers': headers,
-            'timeout': self.timeout
-        }
-        response = requests.post(**kwargs)
-        if response.status_code == 400:
-            # The CreditCourse already exists. Try updating it.
-            kwargs['url'] += course_id.strip('/') + '/'
-            response = requests.put(**kwargs)
+        try:
+            api.courses(course_id).put(data)
```
bderusha: Just want to confirm that this will always raise an error for non-200 responses.  Is that true?

ecommerce/courses/publishers.py
```diff
@@ -37,31 +39,24 @@ def serialize_seat_for_commerce_api(self, seat):
...
+            return True
+        except (SlumberBaseException, Timeout):
+            logger.exception('Failed to create or update CreditCourse [%s].', course_id)
 
-        return response
+        return False
```
cpennington: Returning `True` or `False` to indicate an failure state seems suboptimal to me. Would it be better to just let the exceptions raise, and then catch them in the calling method?

ecommerce/courses/publishers.py
rlucioni: @bderusha see [here](https://github.com/samgiles/slumber/blob/master/slumber/exceptions.py); 4xx and 5xx responses will cause Slumber to raise an exception. @clintonb it's nice to see the API client here!

ecommerce/courses/publishers.py
rlucioni: Not presently, but it might be helpful to log the response body if Slumber raises an exception.

rlucioni:
:+1:

ecommerce/courses/publishers.py
```diff
@@ -37,31 +39,24 @@ def serialize_seat_for_commerce_api(self, seat):
...
-        response = requests.post(**kwargs)
-        if response.status_code == 400:
-            # The CreditCourse already exists. Try updating it.
-            kwargs['url'] += course_id.strip('/') + '/'
-            response = requests.put(**kwargs)
+        try:
+            api.courses(course_id).put(data)
+            return True
+        except (SlumberBaseException, Timeout):
+            logger.exception('Failed to create or update CreditCourse [%s].', course_id)
```
jimabramson: debugging 4xx/5xx exceptions from slumber can be a major PITA because the underlying HTTP response is not output.  However, based on my read of [the code](https://github.com/samgiles/slumber/blob/master/slumber/__init__.py#L99-L103), you can get the response content using the `content` attribute of the raised exception.  I'd strongly recommend logging this.

ecommerce/courses/publishers.py
clintonb: My goal in this change is to isolate concerns. The calling method should not care about the exceptions raised. It should only care if the publish action finished.

cpennington: There's only one user of this method (`publish`).

Also, I'm not wild about the fact that this will result in two messages in the logs, one marking that the publish failed, the other marking that the create/update failed, the latter of which will have a stack trace. There won't be anything to connect the two failures (which are really the same failure).

`publish` is already catching all exceptions, and is logging the same message in that case that it does in the case of `_publish_creditcourse` failing.

clintonb:
@cpennington @jimabramson feedback implemented.

ecommerce/courses/publishers.py
```diff
@@ -94,32 +84,23 @@ def publish(self, course, access_token=None):
...
-                logger.error(
-                    u'Unable to publish CreditCourse for [%s] to LMS. No access token available.',
-                    course_id
+            try:
+                self._publish_creditcourse(course_id, access_token)
+                logger.info(u'Successfully published CreditCourse for [%s] to LMS.', course_id)
+            except SlumberHttpBaseException as e:
+                # Note that %r is used to log the repr() of the response content, which may sometimes
+                # contain non-ASCII Unicode. %s would call str() on the response content, sometimes
+                # resulting in a UnicodeDecodeError.
```
cpennington: Let's change the second sentence to: "We don't know (or want to guess) the encoding, so using %r will log the raw bytes of the message, freeing us from the possibility of encoding errors"

clintonb: Updated.

jimabramson:
:+1: once coverage issue addressed

PR #406, opened 2015-10-25
WAIT-LMS: Using order number for receipt page

clintonb:
The updated receipt page relies on order numbers, not basket IDs.

ECOM-2653


clintonb:
Depends on edx/edx-platform#10349.

jimabramson:
nitpick:  why the switch to camelCase ?

clintonb:
@jimabramson personal preference. Don't get too attached. This page will ultimately die once we move it to Otto.

bderusha:
:+1: lgtm

cpennington:
Do we have a consistent style for url parameters? We should pick one, and stick with it.

rlucioni:
FWIW, we tend to use underscores instead of camelCase elsewhere. :+1: otherwise. Let's make sure this isn't deployed before https://github.com/edx/edx-platform/pull/10349.

clintonb:
Neither the team nor edX has a style for query string parameters. Ultimately, I don't even want to use querystring parameters. I want `https://ecommerce.edx.org/orders/EDX-1234/`.

PR #407, opened 2015-10-25
Added management command to delete ordered baskets

clintonb:
This command can be used to remove baskets linked to orders.

ECOM-2655

### DIFF
--- a/docs/maintenance.rst
+++ b/docs/maintenance.rst
@@ -1,0 +1,17 @@
+Maintenance
+===========
+Most installations should not require any maintenance beyond patching; however, there are a few exceptions.
+
+Baskets
+~~~~~~~
+As more baskets and orders are created, the baskets table can grow pretty large. Depending on your database backend, a
+large table can become difficult to manage and migrate. Once an order is placed, the corresponding basket doesn't hold
+much value and can be deleted. The `delete_ordered_baskets` management command performs this action.
+
+.. code-block:: bash
+
+    # Display number of baskets eligible for deletion
+    $ ./manage.py delete_ordered_baskets
+
+    # Delete all baskets that have been ordered
+    $ ./manage.py delete_ordered_baskets --commit
--- a/ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
+++ b/ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
@@ -1,0 +1,34 @@
+"""
+Management command that deletes baskets associated with orders.
+
+These baskets don't have much value once the order is placed, and unnecessarily take up space.
+"""
+from __future__ import unicode_literals
+from django.core.management import BaseCommand
+from oscar.core.loading import get_model
+
+Basket = get_model('basket', 'Basket')
+
+
+class Command(BaseCommand):
+    help = 'Delete baskets for which orders have been placed.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--commit',
+                            action='store_true',
+                            dest='commit',
+                            default=False,
+                            help='Actually delete the baskets.')
+
+    def handle(self, *args, **options):
+        queryset = Basket.objects.filter(order__isnull=False)
+        count = queryset.count()
+
+        if options['commit']:
+            self.stderr.write('Deleting [{}] baskets...'.format(count))
+            queryset.delete()
+            self.stderr.write('Done.')
+        else:
+            msg = 'This is a dry run. Had the --commit flag been included, [{}] baskets would have been deleted.'. \
+                format(count)
+            self.stderr.write(msg)
--- a/ecommerce/extensions/basket/tests/test_commands.py
+++ b/ecommerce/extensions/basket/tests/test_commands.py
@@ -1,0 +1,52 @@
+from __future__ import unicode_literals
+from StringIO import StringIO
+
+from django.core.management import call_command
+from django.test import TestCase
+from oscar.core.loading import get_model
+from oscar.test import factories
+
+Basket = get_model('basket', 'Basket')
+
+
+class DeleteOrderedBasketsCommandTests(TestCase):
+    command = 'delete_ordered_baskets'
+
+    def setUp(self):
+        super(DeleteOrderedBasketsCommandTests, self).setUp()
+
+        # Create baskets with and without orders
+        self.orders = [factories.create_order() for __ in range(0, 2)]
+        self.unordered_baskets = [factories.BasketFactory() for __ in range(0, 3)]
+
+    def test_without_commit(self):
+        """ Verify the command does not delete baskets if the commit flag is not set. """
+        expected = Basket.objects.count()
+
+        # Call the command with dry-run flag
+        out = StringIO()
+        call_command(self.command, commit=False, stderr=out)
+
+        # Verify no baskets deleted
+        self.assertEqual(Basket.objects.count(), expected)
+
+        # Verify the number of baskets expected to be deleted was printed to stderr
+        expected = 'This is a dry run. Had the --commit flag been included, [{}] baskets would have been deleted.'.\
+            format(len(self.orders))
+        self.assertEqual(out.getvalue().strip(), expected)
+
+    def test_with_commit(self):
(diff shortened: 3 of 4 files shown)


clintonb:
@edx/ecommerce

ecommerce/extensions/basket/tests/test_commands.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def test_basket_deletion(self):
+        """ Verify the command deletes baskets with orders. """
+        # Verify we have baskets with orders
+        self.assertEqual(Basket.objects.filter(order__isnull=False).count(), len(self.orders))
+
+        # Call the command without the dry-run flag
+        out = StringIO()
+        call_command(self.command, stdout=out)
+        print out.getvalue()
```
peter-fogg: Does this need to be here?

peter-fogg:
One question, otherwise :+1:

ecommerce/extensions/basket/tests/test_commands.py
clintonb: Removed.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
```diff
@@ -0,0 +1,33 @@
...
+                            dest='dry_run',
+                            default=False,
+                            help='Show how many baskets will be deleted without actually deleting them')
+
+    def handle(self, *args, **options):
+        queryset = Basket.objects.filter(order__isnull=False)
+        count = queryset.count()
+
+        if options['dry_run']:
+            self.stdout.write(
```
jimabramson: i think it's good practice to output user-facing messages on stderr instead of stdout.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
```diff
@@ -0,0 +1,33 @@
...
+                'This is a dry run. Had this been a real run, [{}] baskets would have been deleted.'.format(count)
+            )
+        else:
+            self.stdout.write('Deleting [{}] baskets...'.format(count))
+            queryset.delete()
```
jimabramson, line 29: do you know if this is going to issue a single query, one per row, or something in between?

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
clintonb: Why? Both output to the same place.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
clintonb, line 29: One query.

ecommerce/extensions/basket/tests/test_commands.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_basket_deletion(self):
+        """ Verify the command deletes baskets with orders. """
+        # Verify we have baskets with orders
+        self.assertEqual(Basket.objects.filter(order__isnull=False).count(), len(self.orders))
+
+        # Call the command without the dry-run flag
+        call_command(self.command)
+
+        # Verify baskets with orders deleted
+        self.assertListEqual(list(Basket.objects.all()), self.unordered_baskets)
```
rlucioni: You can just call `assertEqual` here. Type-specific methods are automatically used by `assertEqual`.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class Command(BaseCommand):
+    help = 'Delete baskets for which orders have been placed.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--dry-run',
+                            action='store_true',
+                            dest='dry_run',
+                            default=False,
```
rlucioni, line 20: Since this command will destroy data, I'd default this argument to `True`. Making it a little more work to actually delete the data could help prevent users from accidentally running this command.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
jimabramson: That's true only when you're running in a console and not redirecting program output somewhere else.  http://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
jimabramson, line 20: I'm fine with `--dry-run` as is, it's a pretty standard construct for shell scripts.  But if we want to be more defensive, I would suggest doing a `--commit` option instead, like we had in the course migration tool.  In that case, you could actually run the delete query, but then roll it back, which might be more useful as a test run anyway.

jimabramson:
Is the plan to run this on an ongoing schedule?  Or are there app changes to automatically delete the baskets once the orders are created?  (the latter seems like a better idea)

clintonb:
@jimabramson I plan to delete baskets after orders are created (ECOM-2677); however, that might be more easily accomplished when we get rid of the create baskets endpoint that simultaneously does checkout.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
clintonb: You can still pipe `stdout` if necessary. What am I missing?

ecommerce/extensions/basket/tests/test_commands.py
clintonb: Updated.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
jimabramson: There's a well-established convention: stderr is for diagnostic messages output while the program is running, stdout is for outputting the actual results of the program.  The former is what is being emitted here; the program does not actually generate any useful "results".

Is it ever going to matter, for this particular command?  Dunno.  If this is only going to ever be run interactively, then probably not.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
clintonb, line 20: Updated to use `commit` to actually perform the deletion. Not running the query when `commit` is omitted, as that could take a while.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
clintonb: Updated.

docs/maintenance.rst
```diff
@@ -0,0 +1,17 @@
+Maintenance
```
clintonb, line 1: @srpearce can you review this at your leisure? I will create a follow-on PR with any recommended changes.

rlucioni:
:+1:

jimabramson:
:+1:

docs/maintenance.rst
srpearce, line 1: Didn't have time today (catching up from last week's vacation), but will do this tomorrow.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
```diff
@@ -0,0 +1,34 @@
...
+    def handle(self, *args, **options):
+        queryset = Basket.objects.filter(order__isnull=False)
+        count = queryset.count()
+
+        if options['commit']:
+            self.stderr.write('Deleting [{}] baskets...'.format(count))
+            queryset.delete()
+            self.stderr.write('Done.')
+        else:
+            msg = 'This is a dry run. Had the --commit flag been included, [{}] baskets would have been deleted.'. \
```
srpearce, line 32: This is a practice command. To delete baskets, include the --commit flag.

This has been an example operation. If the --commit flag had been included, the command would have deleted [{ }] baskets.

Maybe?

clintonb, line 32: I'll go with the latter.

PR #408, opened 2015-10-26
Updated baskets order endpoint

clintonb:
The /baskets/:id/order/ endpoint now retrieves orders based on the order number instead of the order's foreign key to the basket. This allows orders to be retrieved even if the basket has been deleted.

ECOM-2654

### DIFF
--- a/ecommerce/extensions/api/v2/views/orders.py
+++ b/ecommerce/extensions/api/v2/views/orders.py
@@ -10,7 +10,6 @@ from ecommerce.extensions.api import serializers
 from ecommerce.extensions.api.constants import APIConstants as AC
 from ecommerce.extensions.api.throttles import ServiceUserThrottle
 
-
 logger = logging.getLogger(__name__)
 
 Order = get_model('order', 'Order')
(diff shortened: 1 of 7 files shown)


clintonb:
@edx/ecommerce

jimabramson:
:+1:

optional: it's somewhat confusing that the view defined in orders is tested in test_baskets.  i'm not sure if that's easy to fix in this PR, but i think it would be an improvement.

ecommerce/extensions/api/v2/views/orders.py
```diff
@@ -87,7 +87,10 @@ class OrderByBasketRetrieveView(OrderRetrieveView):
     Works exactly the same as OrderRetrieveView, except that orders are looked
     up via the id of the related basket.
     """
-    lookup_field = 'basket_id'
+    def dispatch(self, request, *args, **kwargs):
```
rlucioni: Given that the LMS receipt page now looks up orders by order number instead of basket ID, is this endpoint still necessary?

clintonb: Yes. We still have receipt page links that rely on this endpoint. This endpoint will continue to exist until we deprecate, and remove, v2 of this API.

rlucioni:
:+1:

clintonb:
@jimabramson view moved.

PR #409, opened 2015-10-26
Batching basket deletions

clintonb:
Due to the on_delete constraints, this deletion is not a single SQL call as originally understood. The deletions are now being batched to avoid creating large transactions and timing out.

ECOM-2655

### DIFF
--- a/ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
+++ b/ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
@@ -5,6 +5,7 @@ These baskets don't have much value once the order is placed, and unnecessarily 
 """
 from __future__ import unicode_literals
 from django.core.management import BaseCommand
+from django.db import transaction
 from oscar.core.loading import get_model
 
 Basket = get_model('basket', 'Basket')
@@ -14,6 +15,12 @@ class Command(BaseCommand):
     help = 'Delete baskets for which orders have been placed.'
 
     def add_arguments(self, parser):
+        parser.add_argument('-b', '--batch-size',
+                            action='store',
+                            dest='batch_size',
+                            default=1000,
+                            type=int,
+                            help='Size of each batch of baskets to be deleted.')
         parser.add_argument('--commit',
                             action='store_true',
                             dest='commit',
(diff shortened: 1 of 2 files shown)


clintonb:
Related to #407.

clintonb:
@jimabramson looks like I was wrong about the deletion being a single query.

cpennington:
What things are being deleted when we delete the baskets?

clintonb:
@cpennington We delete the basket and associated line items. Orders and payment processor responses (the only models referencing baskets) are left in tact; however, the basket_id column is set to NULL.

cpennington:
I thought the order kept a reference to the line items. Or are the line items copied into the order?

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
```diff
@@ -25,9 +32,19 @@ def handle(self, *args, **options):
...
-            self.stderr.write('Done.')
+            if count:
+                self.stderr.write('Deleting [{}] baskets...'.format(count))
+                batch_size = options['batch_size']
+                max_id = queryset.order_by('-id')[0].id
+                for start in range(0, max_id, batch_size):
+                    end = min(start + batch_size, max_id)
+                    self.stderr.write('...deleting baskets [{start}] through [{end}]...'.format(start=start, end=end))
+                    with transaction.atomic():
+                        queryset.filter(pk__gte=start, pk__lte=end).delete()
```
jimabramson, line 43: nitpick:  is `pk__gte` needed?  presumably the records where id < start have been deleted already.

jimabramson:
other than a small nitpick, :+1:

clintonb:
@cpennington Baskets have their own representation of line items separate from order's line items. Order line items will not be touched.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
clintonb, line 43: That assumption should be safe; but, the costs of testing it (me regenerating a few million baskets) outweigh the benefits. I'm anxious to be done with this, so I will leave the query as it is.

cpennington:
:+1:

PR #410, opened 2015-10-26
Finished squashing refund migrations

clintonb:
This is the final step in the squash.

ECOM-2594


clintonb:
@edx/ecommerce

clintonb:
Relates to #399 and #405.

rlucioni:
:+1:

PR #412, opened 2015-10-27
Send order number as an enrollment_attribute

bderusha:
[ECOM-2595](https://openedx.atlassian.net/browse/ECOM-2595)

@edx/ecommerce

### DIFF
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -180,7 +180,13 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
                 'course_details': {
                     'course_id': course_key
                 },
-                'enrollment_attributes': []
+                'enrollment_attributes': [
+                    {
+                        'namespace': 'order',
+                        'name': 'order_number',
+                        'value': order.number
+                    }
+                ]
             }
             if provider:
                 data['enrollment_attributes'].append(
--- a/ecommerce/extensions/fulfillment/tests/test_modules.py
+++ b/ecommerce/extensions/fulfillment/tests/test_modules.py
@@ -114,7 +114,13 @@ class EnrollmentFulfillmentModuleTests(CourseCatalogTestMixin, FulfillmentTestMi
             'course_details': {
                 'course_id': self.course_id,
             },
-            'enrollment_attributes': []
+            'enrollment_attributes': [
+                {
+                    'namespace': 'order',
+                    'name': 'order_number',
+                    'value': self.order.number
+                }
+            ]
         }
 
         expected_headers = {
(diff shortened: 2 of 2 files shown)


peter-fogg:
@bderusha You'll have some failing tests which don't expect `enrollment_attributes` to contain the order number.

jimabramson:
@bderusha is the plan to store order_number with the course enrollment somewhere in the LMS ?

bderusha:
@peter-fogg Thanks.  Looking into them now.
@jimabramson Yes.  We are storing it as an enrollment attribute so we can directly link enrollments to orders.

jimabramson:
@bderusha great, thanks.  that should be broadly useful.

rlucioni:
@bderusha I have doubts about whether it's appropriate for this information to be stored with course enrollments on the LMS. Can't you get the same information by querying the ecommerce service for a user's recent orders, filtered by SKU? It seems like we should try to keep these systems and their data as separate as possible.

clintonb:
:+1:

jimabramson:
@rlucioni i think that's a fair point, and i'm interjecting here as i am not completely aware of the context, but IMO: storing a reference to the order for an enrollment - if one exists - is worth it and possibly even inevitable, because it allows us to have solid ecommerce logic that determines whether or not a particular enrollment is eligible for refund, upgrade, etc.  We've been making these connections using leaps of faith up til now.

That said, there is probably more than one way to do this, and I'm curious if other approaches have been discussed.

rlucioni:
@jimabramson I can see how it's a reasonable counterpart to the SKU on course modes. I'm just trying to prevent us from letting too much of the ecommerce system leak into the LMS.

bderusha:
@rlucioni @jimabramson I think the proper long term solution is to actually expose the order history and allow users to directly request refunds for orders.  As it stands right now, the unenrollment action is in charge of everything, so it makes sense for it to have a link to the object it needs to act on.

I would love to pull all of the refund logic encapsulated in the EnrollmentCourse object and give it to Otto.  Unfortunately that is a little bigger than what we are trying to ship right now.

rlucioni:
@bderusha @jimabramson this feels like debt to me. I understand if we need to introduce it in order to hit a milestone. Would it be appropriate to add a story to the self-paced epic to capture the long-term solution referred to above?

jimabramson:
@rlucioni I am fine with that (even if I'm not totally convinced this is debt - under the long-term solution we still need to be able to correlate an enrollment record with an order right?)

meanwhile, for this PR, :+1:   I assume the Enrollment API can handle calls with these additional attributes, or that we won't merge until the needed changes in platform are ready.

rlucioni:
@jimabramson the long-term solution proposed by @bderusha doesn't require associating enrollments and orders on the LMS. Bill's proposed approach would allow users to request refunds by making requests directly to Otto, which then handles unenrollment when the refund is approved. This also means that logic governing when refunds can be requested could live in the ecommerce service, where it belongs. The route we're taking with this PR feels like debt to me because there seems to be a cleaner, more correct way to approach the task at hand.

@bderusha assuming this can be deployed independently of changes to edx-platform, :+1: . If you still feel it's the right way to go, please add a story to the self-paced epic detailing the long-term solution you described.

PR #414, opened 2015-10-27
Updated basket creation logic to use site

clintonb:
This ensures baskets are tied to a single site/tenant, and we avoid leakage across sites.

ECOM-2583

### DIFF
--- a/ecommerce/extensions/partner/shortcuts.py
+++ b/ecommerce/extensions/partner/shortcuts.py
@@ -1,12 +1,3 @@
-from django.contrib.sites.shortcuts import get_current_site
-from django.contrib.sites.models import Site
-
-
 def get_partner_for_site(request):
-    """Get the partner for the requested site"""
-    try:
-        site = get_current_site(request)
-    except Site.DoesNotExist:
-        return None
-
-    return site.siteconfiguration.partner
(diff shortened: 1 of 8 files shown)


clintonb:
@cpennington @rlucioni please review.

ecommerce/extensions/partner/shortcuts.py
```diff
@@ -1,12 +1,3 @@
...
 def get_partner_for_site(request):
-    """Get the partner for the requested site"""
-    try:
-        site = get_current_site(request)
-    except Site.DoesNotExist:
-        return None
-
-    return site.siteconfiguration.partner
+    """ Returns the Partner associated with the request. """
+    return request.site.siteconfiguration.partner
```
cpennington, line 3: Does `request` always have a `site`? If so, why were we using `get_current_site` before?

clintonb, line 3: I added the [`site middleware`](https://docs.djangoproject.com/en/1.8/ref/contrib/sites/#site-middleware). The previous developer was probably unaware of the middleware.

clintonb:
Merging so that we can keep this project moving, and deploy today.

PR #416, opened 2015-10-28
Fixed admin page

clintonb:
- Overrode Basket model's Unicode method to include basket ID
- Payment processor response detail page now loads without issue
- Order detail page loads more efficiently without loading all baskets

ECOM-2732


clintonb:
FYI @edx/ecommerce

PR #417, opened 2015-10-28
Moved get_basket function

clintonb:
Moved the get_basket function from api/data to a class method on the Basket model.

ECOM-2583

### DIFF
--- a/ecommerce/extensions/basket/models.py
+++ b/ecommerce/extensions/basket/models.py
@@ -4,6 +4,7 @@ from oscar.apps.basket.abstract_models import AbstractBasket
 from oscar.core.loading import get_class
 
 OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
+Selector = get_class('partner.strategy', 'Selector')
 
 
 class Basket(AbstractBasket):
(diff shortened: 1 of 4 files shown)


clintonb:
FYI @edx/ecommerce

ecommerce/extensions/basket/models.py
```diff
@@ -14,6 +15,28 @@ class Basket(AbstractBasket):
...
         return OrderNumberGenerator().order_number(self)
 
+    @classmethod
+    def get_basket(cls, user):
+        """Retrieve the basket belonging to the indicated user.
+
+        If no such basket exists, create a new one. If multiple such baskets exist,
+        merge them into one.
+        """
+        editable_baskets = Basket.objects.filter(owner=user, status__in=Basket.editable_statuses)
```
peter-fogg: This and the other lines in this classmethod (and all classmethods) shouldn't mention the class name explicitly; use `cls.objects.filter(...)`.

clintonb: Good catch. Fixed.

peter-fogg:
:+1:

rlucioni:
:+1:

bderusha:
:+1:

PR #418, opened 2015-10-28
Added management command to add site to baskets without one

clintonb:
ECOM-2584


clintonb:
@edx/ecommerce

peter-fogg:
:+1:

rlucioni:
If it's appropriate, it would be great to see something added to the partner config docs about this management command. That aside, :+1: .

PR #420, opened 2015-10-29
Setting order site from basket site

clintonb:
If a basket is associated with a site, the order created from the basket will have the same site.

ECOM-2587

### DIFF
--- a/ecommerce/extensions/order/tests/test_utils.py
+++ b/ecommerce/extensions/order/tests/test_utils.py
@@ -1,11 +1,17 @@
 """Test Order Utility classes """
 from django.test import override_settings
 from oscar.core.loading import get_class
+from oscar.test.factories import create_basket as oscar_create_basket
 from oscar.test.newfactories import BasketFactory
 
+from ecommerce.extensions.fulfillment.status import ORDER
+from ecommerce.tests.factories import SiteConfigurationFactory
 from ecommerce.tests.testcases import TestCase
 
+NoShippingRequired = get_class('shipping.methods', 'NoShippingRequired')
+OrderCreator = get_class('order.utils', 'OrderCreator')
 OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
+OrderTotalCalculator = get_class('checkout.calculators', 'OrderTotalCalculator')
 
 
 class UtilsTest(TestCase):
@@ -34,3 +40,51 @@ class UtilsTest(TestCase):
 
         self.assertEqual(first_basket_id, first_basket.id)
         self.assertEqual(second_basket_id, second_basket.id)
+
+
+class OrderCreatorTests(TestCase):
+    order_creator = OrderCreator()
+
+    def setUp(self):
+        super(OrderCreatorTests, self).setUp()
+        self.user = self.create_user()
+
+    def create_order_model(self, basket):
+        """ Call the create_order_model method to create an Order from the given Basket. """
+        shipping_method = NoShippingRequired()
+        shipping_charge = shipping_method.calculate(basket)
+        total = OrderTotalCalculator().calculate(basket, shipping_charge)
+        return self.order_creator.create_order_model(self.user, basket, None, shipping_method, shipping_charge,
+                                                     None, total, basket.order_number, ORDER.OPEN)
+
+    def create_basket(self, site):
+        """ Returns a new Basket with the specified Site. """
+        basket = oscar_create_basket()
+        basket.site = site
+        basket.save()
+        return basket
+
+    def test_create_order_model_default_site(self):
+        """
+        Verify the create_order_model method associates the order with the default site
+        if the basket does not have a site set.
+        """
+        # Create a basket without a site
+        basket = self.create_basket(None)
+
+        # Ensure the order's site is set to the default site
+        order = self.create_order_model(basket)
+        self.assertEqual(order.site, self.site)
+
+    def test_create_order_model_basket_site(self):
+        """ Verify the create_order_model method associates the order with the basket's site. """
+        # Create a non-default site
+        site_configuration = SiteConfigurationFactory(site__domain='star.fake', partner__name='star')
+        site = site_configuration.site
+
+        # Associate the basket with the non-default site
+        basket = self.create_basket(site)
+
+        # Ensure the order has the non-default site
+        order = self.create_order_model(basket)
+        self.assertEqual(order.site, site)
--- a/ecommerce/extensions/order/utils.py
+++ b/ecommerce/extensions/order/utils.py
@@ -1,6 +1,7 @@
 """Order Utility Classes. """
 from __future__ import unicode_literals
 from django.conf import settings
+from oscar.apps.order.utils import OrderCreator as OscarOrderCreator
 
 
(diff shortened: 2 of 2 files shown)


ecommerce/tests/mixins.py
```diff
@@ -235,7 +235,21 @@ def create_partner(self, name):
...
+        """ Create a new Site, and link it to the specified Partner. """
+        site = Site.objects.create(domain=domain)
+
+        SiteConfiguration.objects.create(
+            site=site,
+            partner=partner,
+            lms_url_root='https://lms.{}'.format(domain),
+            theme_scss_path='/css/path/',
+            payment_processors='fake_processor'
+        )
```
cpennington: Should this use `factory_boy`?

ecommerce/tests/mixins.py
```diff
@@ -244,10 +258,12 @@ def setUp(self):
...
         # Create and configure the default site
-        Site.objects.all().delete()
-        self.site = Site.objects.create(id=1, domain=domain)
+        self.partner = self.create_partner('edx')
 
-        self.partner, _created = Partner.objects.get_or_create(short_code='edx', name='edx')
+        Site.objects.all().delete()
+        self.site = self.create_site(domain, self.partner)
+        self.site.id = 1
+        self.site.save()
```
cpennington: Why does this have to change the id of the site?

ecommerce/tests/mixins.py
clintonb: Yes, that would make this so much simpler.

ecommerce/extensions/order/utils.py
```diff
@@ -42,3 +43,23 @@ def basket_id(self, order_number):
...
+        """
+
+        # Pull the order's site from the basket, if the basket has a site and
+        # a site is not already being explicitly set.
+        if basket.site and 'site'not in extra_order_fields:
+            extra_order_fields['site'] = basket.site
+
+        return super(OrderCreator, self).create_order_model(user, basket, shipping_address, shipping_method,
+                                                            shipping_charge, billing_address, total, order_number,
+                                                            status, **extra_order_fields)
```
cpennington: Nit: I think we tend to drop arguments down on the next line (when they're going to be multiline).

ecommerce/extensions/order/tests/test_utils.py
```diff
@@ -1,8 +1,15 @@
 """Test Order Utility classes """
 from django.test import TestCase, override_settings
+from oscar.core.loading import get_class
+from oscar.test.factories import create_basket
 from oscar.test.newfactories import BasketFactory
 
-from ecommerce.extensions.order.utils import OrderNumberGenerator
+from ecommerce.extensions.fulfillment.status import ORDER
+from ecommerce.extensions.order.utils import OrderNumberGenerator, OrderCreator
```
cpennington: If we load this using the oscar loader, then if we ever accidentally misconfigure the `OrderCreator`, this test will fail and let us know.

ecommerce/tests/mixins.py
clintonb: The default site ID (defined by the setting `SITE_ID`) is 1. I want to make sure we have the default site set properly. I'm going to update this to use `factory_boy` as well.

cpennington: Shouldn't we then use `SITE_ID`, rather than hardcoding 1?

clintonb: (facepalm) Yes.

ecommerce/extensions/order/utils.py
```diff
@@ -42,3 +43,23 @@ def basket_id(self, order_number):
...
+        """
+        Create an order model.
+
+        This override ensures the order's site is set to that of the basket. If the basket has no site, the default
+        site is used. The site value can be overridden by setting the `site` kwarg.
+        """
+
+        # Pull the order's site from the basket, if the basket has a site and
+        # a site is not already being explicitly set.
+        if basket.site and 'site'not in extra_order_fields:
```
jimabramson: `'site' not in`

jimabramson: why wouldn't a basket be associated with a site?

ecommerce/extensions/order/tests/test_utils.py
```diff
@@ -31,3 +38,51 @@ def test_order_number_generation(self):
...
+    def setUp(self):
+        super(OrderCreatorTests, self).setUp()
+        self.user = self.create_user()
+
+    def create_order_model(self, basket):
+        """ Call the create_order_model method to create an Order from the given Basket. """
+        shipping_method = NoShippingRequired()
+        shipping_charge = shipping_method.calculate(basket)
+        total = OrderTotalCalculator().calculate(basket, shipping_charge)
+        return self.order_creator.create_order_model(self.user, basket, None, NoShippingRequired(), shipping_charge,
```
jimabramson: you already have a `shipping_method`, why not just pass that?

ecommerce/extensions/order/utils.py
clintonb: Pre-multi-tenancy baskets have no site. Until this field is made non-nullable, we cannot assume a site exists.

ecommerce/extensions/order/tests/test_utils.py
clintonb: Fixed.

ecommerce/extensions/order/utils.py
clintonb: Fixed.

jimabramson:
BTW, +1 for factory boy

ecommerce/extensions/order/utils.py
jimabramson: understood.

ecommerce/extensions/order/utils.py
clintonb: Updated.

ecommerce/extensions/order/tests/test_utils.py
clintonb: Updated.

clintonb:
Review and merge #421 first.

ecommerce/extensions/order/tests/test_utils.py
```diff
@@ -34,3 +40,51 @@ def test_order_number_generation(self):
...
+        """ Call the create_order_model method to create an Order from the given Basket. """
+        shipping_method = NoShippingRequired()
+        shipping_charge = shipping_method.calculate(basket)
+        total = OrderTotalCalculator().calculate(basket, shipping_charge)
+        return self.order_creator.create_order_model(self.user, basket, None, shipping_method, shipping_charge,
+                                                     None, total, basket.order_number, ORDER.OPEN)
+
+    def create_basket(self, site):
+        """ Returns a new Basket with the specified Site. """
+        basket = create_basket()
```
rlucioni: It took me a few takes to understand what was going on here. Maybe import `create_basket` as `oscar_create_basket`?

rlucioni:
:+1:

ecommerce/extensions/order/tests/test_utils.py
clintonb: Done.

PR #421, opened 2015-10-29
Updated tests

clintonb:
- Using a custom TestCase that creates a partner and site for every test
- Removed PartnerMixin

ECOM-2587

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/__init__.py
+++ b/ecommerce/extensions/api/v2/tests/views/__init__.py
@@ -6,14 +6,14 @@ from oscar.test.newfactories import ProductAttributeValueFactory
 from ecommerce.core.constants import ISO_8601_FORMAT
 from ecommerce.extensions.api.serializers import OrderSerializer
 from ecommerce.extensions.test import factories
-from ecommerce.tests.mixins import UserMixin, ThrottlingMixin
+from ecommerce.tests.mixins import ThrottlingMixin
 
 JSON_CONTENT_TYPE = 'application/json'
 Product = get_model('catalogue', 'Product')
 Selector = get_class('partner.strategy', 'Selector')
 
 
-class OrderDetailViewTestMixin(ThrottlingMixin, UserMixin):
+class OrderDetailViewTestMixin(ThrottlingMixin):
     def url(self):
         raise NotImplementedError
 
@@ -28,12 +28,15 @@ class OrderDetailViewTestMixin(ThrottlingMixin, UserMixin):
 
         self.token = self.generate_jwt_token_header(user)
 
+    def serialize_order(self, order):
+        request = RequestFactory(SERVER_NAME=self.site.domain).get('/')
+        return OrderSerializer(order, context={'request': request}).data
+
     def test_get_order(self):
         """Test successful order retrieval."""
-        request = RequestFactory().get(self.url)
         response = self.client.get(self.url, HTTP_AUTHORIZATION=self.token)
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.data, OrderSerializer(self.order, context={'request': request}).data)
+        self.assertEqual(response.data, self.serialize_order(self.order))
 
     def test_order_wrong_user(self):
         """Test scenarios where an order should return a 404 due to the wrong user."""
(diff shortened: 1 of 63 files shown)


clintonb:
@edx/ecommerce I changed many lines here, so the more eyes, the better.

ecommerce/extensions/dashboard/orders/tests.py
```diff
@@ -15,14 +15,14 @@
...
-from ecommerce.tests.mixins import UserMixin
+from ecommerce.tests.testcases import LiveServerTestCase, TestCase
 
 Order = get_model('order', 'Order')
 Refund = get_model('refund', 'Refund')
 ShippingEventType = get_model('order', 'ShippingEventType')
 
 
-class OrderViewTestsMixin(UserMixin):
+class OrderViewTestsMixin(object):
```
bderusha, line 25: So did we need `UserMixin` at all before?  Is that why this is inheriting from `object` instead of `TestCase`?

bderusha:
so many changes! :+1:

ecommerce/extensions/dashboard/orders/tests.py
clintonb, line 25: Yes. `UserMixin` has the `create_user` method, which will now be on `TestCase`. We should remove some of these mixins later.

ecommerce/extensions/api/v2/tests/views/__init__.py
```diff
@@ -46,7 +49,7 @@ def test_order_wrong_user(self):
 class TestServerUrlMixin(object):
```
peter-fogg: Could this be replaced by the new `ecommerce.tests.mixins.TestServerUrlMixin`?

clintonb: Yes, that was the plan. Nice catch.

ecommerce/extensions/fulfillment/tests/mixins.py
```diff
@@ -1,10 +1,9 @@
...
 
 from ecommerce.extensions.fulfillment.status import ORDER, LINE
-from ecommerce.tests.mixins import UserMixin
 
 
-class FulfillmentTestMixin(UserMixin):
+class FulfillmentTestMixin(object):
     def generate_open_order(self):
         """ Returns an open order, ready to be fulfilled. """
         user = self.create_user()
```
peter-fogg, line 14: It looks like this mixin needs to be used with `UserMixin` due to the `create_user` call. Could you add documentation of that? Or maybe it makes sense to leave it inheriting from `UserMixin`?

clintonb, line 14: All test cases should inherit from `ecommerce.tests.testcases.TestCase` which includes the `UserMixin`. I'll add a comment.

ecommerce/tests/mixins.py
```diff
@@ -243,16 +232,16 @@ def setUp(self):
         domain = 'testserver.fake'
         self.client = self.client_class(SERVER_NAME=domain)
 
-        # Create and configure the default site
-        Site.objects.all().delete()
-        self.site = Site.objects.create(id=1, domain=domain)
+        Site.objects.get_current().delete()
+        site_configuration = SiteConfigurationFactory(partner__name='edX',
```
peter-fogg: Formatting nitpick -- 

``` python
site_configuration = SiteConfigurationFactory(
    partner__name='edX',
    site__id=settings.SITE_ID,
    site__domain=domain,
)
```

peter-fogg:
Couple small things, otherwise :+1: (and the branch needs a rebase).

ecommerce/extensions/api/v2/tests/views/mixins.py
```diff
@@ -32,9 +30,7 @@ def serialize_catalog(self, catalog):
             'id': catalog.id,
             'partner': catalog.partner.id,
             'name': catalog.name,
-            'products': 'http://testserver' + reverse(
-                'api:v2:catalog-product-list',
-                kwargs={'parent_lookup_stockrecords__catalogs': catalog.id},
-            )
+            'products': self.get_full_url(reverse('api:v2:catalog-product-list',
```
rlucioni, line 33: `CourseCatalogTestMixin` inherits from `object`; from where does it get `get_full_url()`?

ecommerce/tests/mixins.py
clintonb: Updated.

PR #423, opened 2015-11-02
Updated order number generator

clintonb:
Order number generation now takes the site/partner into account for the order number prefix.

ECOM-2618

### DIFF
--- a/ecommerce/extensions/order/tests/test_utils.py
+++ b/ecommerce/extensions/order/tests/test_utils.py
@@ -5,7 +5,7 @@ from oscar.test.factories import create_basket as oscar_create_basket
 from oscar.test.newfactories import BasketFactory
 
 from ecommerce.extensions.fulfillment.status import ORDER
-from ecommerce.tests.factories import SiteConfigurationFactory
+from ecommerce.tests.factories import SiteConfigurationFactory, PartnerFactory
 from ecommerce.tests.testcases import TestCase
 
 NoShippingRequired = get_class('shipping.methods', 'NoShippingRequired')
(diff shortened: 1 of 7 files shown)


clintonb:
@rlucioni @peter-fogg @bderusha please review.

ecommerce/extensions/order/utils.py
```diff
@@ -26,7 +40,7 @@ def order_number_from_basket_id(self, basket_id):
             string: Order number.
         """
         order_id = int(basket_id) + self.OFFSET
-        return u'{prefix}-{order_id}'.format(prefix=settings.ORDER_NUMBER_PREFIX, order_id=order_id)
```
rlucioni, line 29: Since it looks like it's no longer used, `ORDER_NUMBER_PREFIX` can be removed from ecommerce's Django settings. It should also be removed from the defaults listed in the configuration repo.

clintonb, line 29: Done.

ecommerce/extensions/order/tests/test_utils.py
```diff
@@ -14,32 +14,41 @@
...
 
 
-class UtilsTest(TestCase):
-    """Unit tests for the order utility functions and classes. """
+class OrderNumberGeneratorTests(TestCase):
+    generator = OrderNumberGenerator()
 
-    ORDER_NUMBER_PREFIX = 'FOO'
+    def assert_order_number_matches_basket(self, basket, partner):
+        expected = '{}-{}'.format(partner.short_code.upper(), 100000 + basket.id)
```
rlucioni, line 21: I'd use `generator.OFFSET` instead of harcoding `100000`.

ecommerce/extensions/order/tests/test_utils.py
```diff
@@ -14,32 +14,41 @@
...
-        """
-        first_basket = BasketFactory()
-        second_basket = BasketFactory()
+    def test_order_number(self):
+        """ Verify the method returns an order number determined using the basket's site/partner and ID. """
+        basket = BasketFactory(site=self.site)
+        self.assert_order_number_matches_basket(basket, self.partner)
+
+    def test_order_number_for_basket_without_site(self):
+        """ Verify the order number is linked to the default site, if the basket has not associated site. """
```
rlucioni: s/has not/has no

ecommerce/extensions/order/tests/test_utils.py
clintonb, line 21: Leaving it as is. If anyone changes the offset (which they shouldn't), this is a reminder to check the tests.

ecommerce/extensions/order/tests/test_utils.py
clintonb: Updated.

rlucioni:
:+1:

peter-fogg:
:+1:

bderusha:
:+1:

PR #426, opened 2015-11-04
Fixed for duplicate payment ID received from Paypal.

tasawernawaz:
Paypal sometime sends duplicate payment ID, which cause error while getting basket or order. An exception will be logged when duplicate ID returned and user will be redirected to checkout error page served at `/commerce/checkout/error/`.

This PR depends upon this [PR](https://github.com/edx/edx-platform/pull/10502).

Please review this @awais786 @afeef @ahsan-ul-haq @zubair-arbi

### DIFF
--- a/ecommerce/extensions/payment/tests/test_views.py
+++ b/ecommerce/extensions/payment/tests/test_views.py
@@ -459,6 +459,81 @@ class PaypalPaymentExecutionViewTests(PaypalMixin, PaymentEventsMixin, TestCase)
             self._assert_order_placement_failure(error_message)
             self.assertTrue(fake_handle_order_placement.called)
 
+    @httpretty.activate
+    def test_payment_error_with_duplicate_payment_id(self):
+        """
+        Verify that we fail gracefully when PayPal sends us the wrong payment ID,
+        logging the exception and redirecting the user to an LMS checkout error page.
+        """
+        logger_name = 'ecommerce.extensions.payment.views'
+        with LogCapture(logger_name) as l:
+            self.mock_oauth2_response()
+
+            # Create payment records with different baskets which will have same payment ID
+            self.mock_payment_creation_response(self.basket)
+            self.processor.get_transaction_parameters(self.basket, request=self.request)
+
+            dummy_basket = factories.create_basket()
+            self.mock_payment_creation_response(dummy_basket)
+            self.processor.get_transaction_parameters(dummy_basket, request=self.request)
+
+            self._assert_error_page_redirect()
+            l.check(
+                (
+                    logger_name,
+                    'INFO',
+                    'Payment [{payment_id}] approved by payer [{payer_id}]'.format(
+                        payment_id=self.PAYMENT_ID,
+                        payer_id=self.PAYER_ID
+                    )
+                ),
+                (
+                    logger_name,
+                    'ERROR',
+                    'Duplicate payment ID [{payment_id}] received from PayPal.'.format(payment_id=self.PAYMENT_ID),
+                ),
+            )
+
+    @httpretty.activate
+    def test_payment_error_with_no_basket(self):
+        """
+        Verify that we fail gracefully when any Exception occurred in _get_basket() method,
+        logging the exception and redirecting the user to an LMS checkout error page.
+        """
+        with mock.patch.object(PaymentProcessorResponse.objects, 'get', side_effect=Exception):
+            logger_name = 'ecommerce.extensions.payment.views'
+            with LogCapture(logger_name) as l:
+                self.mock_oauth2_response()
+                self.mock_payment_creation_response(self.basket)
+                self.processor.get_transaction_parameters(self.basket, request=self.request)
+                self._assert_error_page_redirect()
+
+                l.check(
+                    (
+                        logger_name,
+                        'INFO',
+                        'Payment [{payment_id}] approved by payer [{payer_id}]'.format(
+                            payment_id=self.PAYMENT_ID,
+                            payer_id=self.PAYER_ID
+                        )
+                    ),
+                    (
+                        logger_name,
+                        'ERROR',
+                        'Unexpected error during basket retrieval while executing PayPal payment.'
+                    ),
+                )
+
+    def _assert_error_page_redirect(self):
+        """Verify redirection to the configured checkout error page after attempted failed payment execution."""
+        response = self.client.get(reverse('paypal_execute'), self.RETURN_DATA)
+
+        self.assertRedirects(
+            response,
+            self.processor.error_url,
+            fetch_redirect_response=False
+        )
+
 
 @mock.patch('ecommerce.extensions.payment.views.call_command')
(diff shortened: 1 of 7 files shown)


ecommerce/extensions/payment/views.py
```diff
@@ -159,15 +159,18 @@ class PaypalPaymentExecutionView(EdxOrderPlacementMixin, View):
...
+        try:
+            basket = PaymentProcessorResponse.objects.get(
+                processor_name=self.payment_processor.NAME,
+                transaction_id=payment_id
+            ).basket
+            basket.strategy = strategy.Default()
+            return basket
+        except MultipleObjectsReturned:
+            logging.exception(u"Duplicate payment ID [%s] is received from Paypal.", payment_id)
+            return None
```
zubair-arbi, line 194: No need to explicitly return `None`

ecommerce/extensions/payment/views.py
```diff
@@ -159,15 +159,18 @@ class PaypalPaymentExecutionView(EdxOrderPlacementMixin, View):
...
+        it will log exception and return."""
+        try:
+            basket = PaymentProcessorResponse.objects.get(
+                processor_name=self.payment_processor.NAME,
+                transaction_id=payment_id
+            ).basket
+            basket.strategy = strategy.Default()
+            return basket
+        except MultipleObjectsReturned:
+            logging.exception(u"Duplicate payment ID [%s] is received from Paypal.", payment_id)
```
zubair-arbi: nit: `Duplicate payment ID [%s] received from Paypal.`

ecommerce/extensions/payment/views.py
clintonb, line 194: I prefer explicit returns, even if the value is `None`. The explicit return removes any ambiguity future developer might have about the code.

clintonb:
Where are the tests?

zubair-arbi:
lgtm :+1:

tasawernawaz:
@rlucioni Please review this.

ecommerce/extensions/payment/views.py
```diff
@@ -159,15 +159,27 @@ class PaypalPaymentExecutionView(EdxOrderPlacementMixin, View):
...
+        try:
+            basket = PaymentProcessorResponse.objects.get(
+                processor_name=self.payment_processor.NAME,
+                transaction_id=payment_id
+            ).basket
+            basket.strategy = strategy.Default()
+            return basket
+        except MultipleObjectsReturned:
+            logging.exception(u"Duplicate payment ID [%s] received from Paypal.", payment_id)
+            return None
```
rlucioni, line 194: @tasawernawaz this looks good, but what if a different, unexpected exception type is raised? The user experience would be improved if we were more defensive here by adding a second `except` clause to catch and log other, unexpected errors which might occur during basket retrieval:

``` python
except:
    logging.exception(u"Unexpected error during basket retrieval while executing PayPal payment.")
    return None
```

ecommerce/extensions/payment/views.py
```diff
@@ -159,15 +159,27 @@ class PaypalPaymentExecutionView(EdxOrderPlacementMixin, View):
...
+        """
+        try:
+            basket = PaymentProcessorResponse.objects.get(
+                processor_name=self.payment_processor.NAME,
+                transaction_id=payment_id
+            ).basket
+            basket.strategy = strategy.Default()
+            return basket
+        except MultipleObjectsReturned:
+            logging.exception(u"Duplicate payment ID [%s] received from Paypal.", payment_id)
```
rlucioni: Nitpick: s/Paypal/PayPal

ecommerce/extensions/payment/views.py
```diff
@@ -159,15 +159,27 @@ class PaypalPaymentExecutionView(EdxOrderPlacementMixin, View):
...
+        Retrieve a basket using a payment ID.
 
-        basket.strategy = strategy.Default()
+        Arguments:
+            payment_id: payment_id received from Paypal.
 
-        return basket
+        Returns:
+            It will return related basket or log exception and return None if
+            duplicate payment_id received.
```
rlucioni: Thank you for adding this docstring!

ecommerce/extensions/payment/views.py
```diff
@@ -159,15 +159,30 @@ class PaypalPaymentExecutionView(EdxOrderPlacementMixin, View):
...
+                processor_name=self.payment_processor.NAME,
+                transaction_id=payment_id
+            ).basket
+            basket.strategy = strategy.Default()
+            return basket
+        except MultipleObjectsReturned:
+            logging.exception(u"Duplicate payment ID [%s] received from PayPal.", payment_id)
+            return None
+        except Exception:  # pylint: disable=broad-except
+            logging.exception(u"Unexpected error during basket retrieval while executing PayPal payment.")
```
tasawernawaz: @rlucioni How we can test this generic exception, because when I tried this
`with mock.patch.object(PaypalPaymentExecutionView, 'handle_payment', side_effect=Exception) as fake_handle_payment:`
It takes it as a generic exception in [PaypalPaymentExecutionView](https://github.com/edx/ecommerce/pull/426/files#diff-3a39a1e895cafeeed9c6a0c0dffd2dceL189) itself rather than as an exception in `_get_basket()` method.

ecommerce/extensions/payment/views.py
```diff
@@ -159,15 +159,30 @@ class PaypalPaymentExecutionView(EdxOrderPlacementMixin, View):
...
+        """
+        try:
+            basket = PaymentProcessorResponse.objects.get(
+                processor_name=self.payment_processor.NAME,
+                transaction_id=payment_id
+            ).basket
+            basket.strategy = strategy.Default()
+            return basket
+        except MultipleObjectsReturned:
+            logging.exception(u"Duplicate payment ID [%s] received from PayPal.", payment_id)
```
rlucioni: This and the below should be `logger.exception`.

ecommerce/extensions/payment/views.py
rlucioni: @tasawernawaz that would be because `handle_payment` isn't called until later in the `get` handler. Have you tried mocking the `get` method on `PaymentProcessorResponse.objects`?

tasawernawaz:
@rlucioni Code updated and test added for generic exception. Please review this.

ecommerce/extensions/payment/tests/test_views.py
```diff
@@ -459,6 +459,83 @@ def test_unanticipated_error_during_order_placement(self):
...
+                        )
+                    ),
+                    (
+                        logger_name,
+                        'ERROR',
+                        'Unexpected error during basket retrieval while executing PayPal payment.'
+                    ),
+                )
+
+    def _assert_execution_redirect_on_error_page(self):
```
rlucioni: I'd make this name a little clearer, `_assert_error_page_redirect`.

rlucioni:
Just one small suggestion, otherwise :+1: .

PR #427, opened 2015-11-04
Create new product category to support enrollment codes

vkaracic:
Created a new Client model to use for the enrollment codes. Added a new product class type via a migration. Added a `courses` field to the catalog model because a catalog will have one or more courses and a course will have one or more catalogs. Accompanying tests are included. 

https://openedx.atlassian.net/browse/SOL-1326

@mjfrey @clintonb please take a look.

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -59,3 +59,8 @@ class User(AbstractUser):
 
     def get_full_name(self):
         return self.full_name or super(User, self).get_full_name()
+
+
+class Client(User):
+    """The model for the business client."""
+    pass
--- a/ecommerce/extensions/catalogue/migrations/0002_auto_20150223_1052.py
+++ b/ecommerce/extensions/catalogue/migrations/0002_auto_20150223_1052.py
@@ -60,6 +60,15 @@ def create_catalog(apps, schema_editor):
     )
 
 
+def remove_catalog(apps, schema_editor):
+    """ Reverse function. """
+    Category = apps.get_model("catalogue", "Category")
+    ProductClass = apps.get_model("catalogue", "ProductClass")
+
+    Category.objects.filter(slug='seats').delete()
+    ProductClass.objects.filter(slug='seat').delete()
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
@@ -67,5 +76,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(create_catalog),
+        migrations.RunPython(create_catalog, remove_catalog),
     ]
--- a/ecommerce/extensions/catalogue/migrations/0012_enrollment_code_product_class.py
+++ b/ecommerce/extensions/catalogue/migrations/0012_enrollment_code_product_class.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+import oscar
+from django.db import models, migrations
+
+
+def create_product_class(apps, schema_editor):
+    """
+    Create a default EnrollmentCode class with added attributes:
+        - catalog (of courses)
+        - start date (of validity)
+        - end date
+        - type
+    """
+    AttributeOptionGroup = apps.get_model("catalogue", "AttributeOptionGroup")
+    AttributeOption = apps.get_model("catalogue", "AttributeOption")
+    Category = apps.get_model("catalogue", "Category")
+    ProductAttribute = apps.get_model("catalogue", "ProductAttribute")
+    ProductClass = apps.get_model("catalogue", "ProductClass")
+
+    # Create a new product class for enrollment codes
+    enrollment_code = ProductClass.objects.create(
+        track_stock=False,
+        requires_shipping=False,
+        name='Enrollment code',
+        slug='enrollment_code',
+    )
+
+    # Create product attributes for enrollment code products
+    ProductAttribute.objects.create(
+        product_class=enrollment_code,
+        name="Catalog",
+        code="catalog",
+        type="entity",
+        required=True
+    )
+    ProductAttribute.objects.create(
+        product_class=enrollment_code,
+        name="Start date",
+        code="start_date",
+        type="date",
+        required=True
+    )
+    ProductAttribute.objects.create(
+        product_class=enrollment_code,
+        name="End date",
+        code="end_date",
+        type="date",
+        required=True
+    )
+
+    # Option group for the type of enrollment code.
+    # single_use - Can be used once by one customer
+    # multi_use - Can be used multiple times by multiple customers
+    # once_per_customer - Can only be used once per customer
+    group = AttributeOptionGroup.objects.create(
+        name="Type"
+    )
+    AttributeOption.objects.create(
+        group=group,
+        option='single_use'
+    )
+    AttributeOption.objects.create(
+        group=group,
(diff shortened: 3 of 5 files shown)


ecommerce/extensions/catalogue/models.py
```diff
@@ -20,6 +20,7 @@ class Catalog(models.Model):
     name = models.CharField(max_length=255)
     partner = models.ForeignKey('partner.Partner', related_name='catalogs')
     stock_records = models.ManyToManyField('partner.StockRecord', blank=True, related_name='catalogs')
+    courses = models.ManyToManyField('courses.Course', blank=True)
```
clintonb: Why is this necessary? You can get to courses through the stock records and related products. A catalog is the general concept of a catalog. It includes more than just courses, so should share no direct links to courses.

ecommerce/extensions/catalogue/migrations/0012_enrollment_code_product_class.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0001_initial'),
+        ('catalogue', '0011_auto_20151019_0639')
+    ]
+    operations = [
+        migrations.RunPython(create_product_class)
```
clintonb: Include a reverse operation.

ecommerce/core/models.py
```diff
@@ -59,3 +59,8 @@ class Meta(object):
 
     def get_full_name(self):
         return self.full_name or super(User, self).get_full_name()
+
+
+class Client(User):
+    """Custom user model for clients."""
```
clintonb: This could use a better docstring. Explain why the model exists.

ecommerce/extensions/catalogue/migrations/0012_enrollment_code_product_class.py
mjfrey: Do we need reverse for all migrations?  I do not see one for the Seat migration.

clintonb: Yes, all migrations should have reverse migrations. Feel free to add one for the seat migration. That was created before I was on the team.

vkaracic:
@clintonb I have reverted the commit that added a new field to Catalog, addressed your other comments and changed the tests accordingly. Can you please take a look now?

mjfrey:
LGTM now.
:+1:

.travis.yml
```diff
@@ -36,6 +36,6 @@ script:
     - make validate_js
 branches:
     only:
-        - master
+        - ecodes/master
```
clintonb: Remove.

ecommerce/core/models.py
```diff
@@ -59,3 +59,8 @@ class Meta(object):
 
     def get_full_name(self):
         return self.full_name or super(User, self).get_full_name()
+
+
+class Client(User):
+    """The model for the customer that purchases enrollment codes."""
```
clintonb: `Business clients/purchasers.`

This model doesn't care about enrollment codes, so there should be no mention of enrollment codes in its comment.

ecommerce/extensions/catalogue/migrations/0002_auto_20150223_1052.py
```diff
@@ -60,6 +60,19 @@ def create_catalog(apps, schema_editor):
...
 
 
+def remove_catalog(apps, schema_editor):
+    """ Reverse function. """
+    Category = apps.get_model("catalogue", "Category")
+    ProductAttribute = apps.get_model("catalogue", "ProductAttribute")
+    ProductClass = apps.get_model("catalogue", "ProductClass")
+
+    Category.objects.filter(slug='seats').delete()
+    ProductAttribute.objects.filter(code='certificate_type').delete()
```
clintonb: Delete the `ProductClass` and the attributes will be deleted automatically.

ecommerce/extensions/catalogue/migrations/0012_enrollment_code_product_class.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def remove_product_class(apps, schema_editor):
+    """ Reverse function. """
+    Category = apps.get_model("catalogue", "Category")
+    ProductAttribute = apps.get_model("catalogue", "ProductAttribute")
+    ProductClass = apps.get_model("catalogue", "ProductClass")
+
+    Category.objects.filter(slug='enrollment_code').delete()
+    ProductAttribute.objects.filter(code='end_date').delete()
```
clintonb: Same as above. You only need to delete the `ProductClass`.

vkaracic:
@clintonb updated!

ecommerce/extensions/catalogue/tests/test_enrollment_codes.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_enrollment_code_product(self):
+        """
+        Test if an enrollment code is properly created
+        and has a course associated with it.
+        """
+        catalog = Catalog.objects.create(
+            partner_id=self.partner.id
+        )
+        course = Course.objects.create(
```
clintonb: Consider using `CourseFactory`.

ecommerce/extensions/catalogue/tests/test_enrollment_codes.py
```diff
@@ -0,0 +1,39 @@
...
+            id='org/number/run',
+            name='Test course'
+        )
+
+        ec_product_class = ProductClass.objects.get(slug='enrollment_code')
+        enrollment_code_product = Product.objects.create(
```
clintonb: Consider using `ProductFactory`.

ecommerce/extensions/catalogue/tests/test_enrollment_codes.py
```diff
@@ -0,0 +1,39 @@
+import datetime
+from oscar.core.loading import get_model
```
clintonb, line 3: Imports should be grouped: main lib, third-party, local. Each group should be alphabetized.

clintonb:
:+1:

ecommerce/extensions/catalogue/migrations/0012_enrollment_code_product_class.py
```diff
@@ -0,0 +1,78 @@
...
+    Category.objects.create(
+        description="All Enrollment Codes",
+        numchild=2,
+        slug="enrollment_codes",
+        depth=1,
+        path="0002",
+        image="",
+        name="Enrollment Codes"
+    )
+
```
mjfrey, line 91: I am just thinking we missed an attribute we need at fulfillment time.  The type of Voucher to create. Multi-use, Single-use etc...
@clintonb what do you think?
I suggest adding an attribute called "Type" and having it match the ENUM that Abstract Voucher uses.

clintonb, line 91: Agreed.

## 

Clinton Blackburn

vkaracic:
@mjfrey @clintonb I have assigned a new product attribute of type option to the enrollment code product class. Please take a look at the last commit. If you are happy with the code I believe I can merge this.

clintonb:
:+1:

mjfrey:
Great.

:+1:

PR #428, opened 2015-11-04
Catch valid use case exceptions and log info message

bderusha:
## [ECOM-2362](https://openedx.atlassian.net/browse/ECOM-2362)

@jimabramson @rlucioni Please review.

FYI @edx/ecommerce

### DIFF
--- a/ecommerce/extensions/payment/views.py
+++ b/ecommerce/extensions/payment/views.py
@@ -12,7 +12,7 @@ from django.utils.decorators import method_decorator
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import View
 from oscar.apps.partner import strategy
-from oscar.apps.payment.exceptions import PaymentError
+from oscar.apps.payment.exceptions import PaymentError, UserCancelled, TransactionDeclined
 from oscar.core.loading import get_class, get_model
 
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
@@ -112,6 +112,15 @@ class CybersourceNotifyView(EdxOrderPlacementMixin, View):
                         ppr.id
                     )
                     return HttpResponse(status=400)
+                except (UserCancelled, TransactionDeclined) as exception:
+                    logger.info(
+                        'CyberSource payment did not complete for basket [%d] because [%s]. '
+                        'The payment response was recorded in entry [%d].',
+                        basket.id,
+                        exception.__class__.__name__,
+                        ppr.id
+                    )
+                    return HttpResponse()
                 except PaymentError:
                     logger.exception(
                         'CyberSource payment failed for basket [%d]. The payment response was recorded in entry [%d].',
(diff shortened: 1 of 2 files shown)


jimabramson:
@bderusha couple things about this PR - 
- ECOM-2362 is actually for reducing log severity from ERROR to INFO in lms/djangoapps/shoppingcart.  (see https://github.com/edx/edx-platform/blob/master/lms/djangoapps/shoppingcart/processors/CyberSource2.py#L144-L153; check the splunk query in the ticket for example tracebacks).
- It's good that you've surfaced that the same adjustment to handling needs to happen here! - it's still worth doing - adding some comments inline about the approach.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -38,20 +41,27 @@ def handle_payment(self, response, basket):
...
+                processor_name=payment_event.processor_name,
+                reference=payment_event.reference,
+                user_id=basket.owner.id
+            )
+        except(UserCancelled, TransactionDeclined) as exception:
+            audit_log(
+                exception.__class__.__name__,
+                processor_name=self.payment_processor.NAME,
+                message=exception.message
+            )
```
jimabramson: we are already handling these exceptions one level up, in https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/payment/views.py#L115-L121 (`UserCancelled` and `TransactionDeclined` both are subclasses of `PaymentError`).  I think a simpler approach would be to check for those specific exceptions there, before the more general `PaymentError`, log at `info` instead of `exception` and then return an empty response to end processing without errors.

The audit logging feature in Otto is used specifically for splunk queries that drive metrics dashboards - I recommend just using raw logging to emit a one-liner that the user or card provider declined, as indicated by the exception type.

bderusha:
@jimabramson updates made based on your suggestions

ecommerce/extensions/payment/views.py
```diff
@@ -112,6 +112,15 @@ def post(self, request):
                         ppr.id
                     )
                     return HttpResponse(status=400)
+                except(UserCancelled, TransactionDeclined) as exception:
```
rlucioni: Nitpick: missing space between `except` and the opening parentheses.

ecommerce/extensions/payment/views.py
```diff
@@ -112,6 +112,15 @@ def post(self, request):
...
+                    logger.info(
+                        'CyberSource payment did not complete for basket [%d] because [%s]. \
+The payment response was recorded in entry [%d].',
```
rlucioni: Since you're within parentheses, you don't actually need to use the backslash. I'm surprised pylint didn't catch this. You can do:

``` python
logger.info(
    'CyberSource payment did not complete for basket [%d] because [%s]. '
    'The payment response was recorded in entry [%d].',
    basket.id,
    ...
```

rlucioni:
Just a few nitpicks, but :+1: otherwise.

jimabramson:
@bderusha nice :+1:

jimabramson:
thanks for making the changes.

PR #434, opened 2015-11-12
Updated Datatables dependency

clintonb:
This fixes a security vulnerability and fixes build failures resulting from dataTables.boostrap.js referencing "datatables.net" instead of "datatables".

ECOM-2314


clintonb:
@rlucioni @jimabramson 

FYI @feanil @maxrothman

bower.json
```diff
@@ -17,7 +17,7 @@
     "requirejs": "~2.1.15",
     "underscore": "~1.8.2",
     "bootstrapaccessibilityplugin": "~1.0.4",
-    "datatables": "~1.10.7",
+    "datatables": "~1.10.10",
```
rlucioni: Any chance we can pin this to 1.10.10 so the version of this package doesn't change under us again?

clintonb: Updated.

rlucioni:
:+1:

jimabramson:
:+1:  thx for fixing this @clintonb

PR #435, opened 2015-11-12
Add a wait to PayPal acceptance test to allow checkout form to load

rlucioni:
This test sometimes tries to submit the checkout form before it has been able to load, causing the test to fail. This change ensures that the page has loaded before the test tries to submit payment.

@clintonb @ronaldmulero @edx/ecommerce


clintonb:
:+1:

jimabramson:
:+1:

PR #439, opened 2015-11-19
Displaying list of credit providers in CAT

clintonb:
When creating/editing a credit seat, users will now see a list of available credit providers instead of an empty text field.

ECOM-2563


clintonb:
Course Detail View
screen shot 2015-11-19 at 3 43 46 pm

Course Edit View
screen shot 2015-11-19 at 3 44 03 pm

clintonb:
@AlasdairSwan @peter-fogg please review
FYI @edx/ecommerce

ecommerce/static/js/test/specs/views/course_edit_view_spec.js
```diff
@@ -185,7 +190,7 @@ define([
 
                 expect($creditElement.length).toBe(1);
                 _.each($creditSeats, function (seat) {
-                    expect($(seat).find('[name=credit_provider]').val()).toBe('Harvard');
+                    expect($(seat).find('[name=credit_provider]').val()).toBe('harvard');
```
AlasdairSwan, line 193: Should you be adding `.toLowerCase` to the value from the DOM so we don't care about case? Maybe we shouldn't, but maybe we should.

ecommerce/static/js/views/course_detail_view.js
```diff
@@ -1,3 +1,5 @@
+var ecommerce = ecommerce || {};
```
AlasdairSwan: Why is this being declared outside of RequireJS? This should be added via require, even if it's just with a shim file that returns `ecommerce || {}`

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
```diff
@@ -18,8 +18,13 @@ define([
             },
 
             bindings: {
-                'input[name=credit_provider]': {
+                'select[name=credit_provider]': {
```
AlasdairSwan, line 24: I would feel more comfortable using a class selector instead of element (so you wouldn't have had to make a change at all when swapping an input for a select as long as the select had the appropriate class)

ecommerce/static/js/test/specs/views/course_edit_view_spec.js
clintonb, line 193: Harvard isn't the best example. The value should be set to the credit provider's ID instead of it's display name (e.g. asu instead of Arizona State University).

AlasdairSwan, line 193: Cool.

ecommerce/static/js/views/course_seat_form_fields/credit_course_seat_form_field_row_view.js
clintonb, line 24: I see your point; but the likelihood of this changing again is quite low. If the input type does change, future developers will need a reason to come here and get rid of `selectOptions`.

ecommerce/courses/tests/test_views.py
```diff
@@ -39,20 +47,105 @@ def test_course_ids_required(self):
...
+            {
+                'id': 'shk',
+                'display_name': 'School of Hard Knocks'
+            },
+            {
+                'id': 'acme',
+                'display_name': 'Acme University'
+            }
+        ]
+        providers.sort(key=lambda provider: provider['display_name'])
```
peter-fogg, line 68: Is there a reason you don't just declare `providers` in sorted order?

ecommerce/static/js/apps/course_admin_app.js
```diff
@@ -1,9 +1,13 @@
+var ecommerce = ecommerce || {};
```
peter-fogg: Is there no nicer way to do this with RequireJS? I really don't like mixing old-style JS namespacing with proper require calls.

ecommerce/courses/tests/test_views.py
clintonb, line 68: Yes. I am implicitly verifying providers are sorted by the view.

ecommerce/static/js/test/specs/views/course_detail_view_spec.js
```diff
@@ -154,6 +158,7 @@ define([
                 };
 
             beforeEach(function () {
+                ecommerce.credit.providers = new CreditProviderCollection();
```
peter-fogg, line 161: If `ecommerce` is undefined on line 1, you'll need to give it the `credit` property before this line.

ecommerce/static/js/views/course_detail_view.js
peter-fogg: Ah, I see Alasdair has beat me to the punch...

AlasdairSwan: This it the way we're doing it in the programs repo https://github.com/edx/programs/blob/master/programs/static/js/shims/gettext.js

ecommerce/courses/tests/test_views.py
```diff
@@ -39,20 +47,105 @@ def test_course_ids_required(self):
 class CourseAppViewTests(TestCase):
     path = reverse('courses:app', args=[''])
 
+    def mock_credit_api_providers(self):
+        """
+        Mock GET requests to the Credit API's provider endpoint.
+
+        /api/credit/v1/providers
+        """
+        self.assertTrue(httpretty.is_enabled)
```
peter-fogg: #441

clintonb: Fixed.

rlucioni: I didn't realize #441 would be so timely. Glad we caught this and stopped it from propagating.

ecommerce/static/js/test/specs/views/course_detail_view_spec.js
clintonb, line 161: `ecommerce` is defined via require dependency.

peter-fogg, line 161: Right -- this was before the shim was added.

peter-fogg:
:+1:

ecommerce/static/js/views/course_form_view.js
```diff
@@ -372,6 +372,7 @@ define([
 
                 // Validate the input and display a message, if necessary.
                 if (!this.model.isValid(true)) {
+                    console.log(this.model.attributes);
```
AlasdairSwan: We shouldn't be committing console logs should we?

clintonb:
@AlasdairSwan `ecommerce.js` is only being called once, so we should be in the clear to merge.

AlasdairSwan:
Cool, then if you want that console.log to be there we are good to go :+1:

ecommerce/static/js/views/course_form_view.js
clintonb: Removed.

PR #440, opened 2015-11-19
Document use of payment processor switches

rlucioni:
@peter-fogg, we should have done this yesterday, but forgot.

FYI @edx/ecommerce


peter-fogg:
@rlucioni :+1:

It's debatable if this belongs in the doc right now, but these payment switches shouldn't all be disabled unless absolutely necessary until the corresponding platform changes are live. Otherwise users would see a payment page with no way to pay, and no message explaining what the issue is.

rlucioni:
@peter-fogg that's fair. Since it's very unlikely that all payment processors will be disabled in the time it'll take to get the platform changes deployed, and since the changes are already merged to master, I'm just going to go ahead and merge this.

PR #441, opened 2015-11-19
Actually check that httpretty is enabled when attempting to register URIs

rlucioni:
Drive-by bug fix. I stumbled across this while looking for something else.

`httpretty.is_enabled` returns a bound method which will always evaluate as truthy. Actually calling that method allows the assertion to work as expected and exposed a bug in the tests, too.

@clintonb FYI @edx/ecommerce


clintonb:
:+1:

PR #445, opened 2015-11-30
[wip] Add verified-with-audit type to CAT.

peter-fogg:


ecommerce/courses/models.py
```diff
@@ -86,6 +86,8 @@ def type(self):
         elif 'professional' in seat_types or 'no-id-professional' in seat_types:
             return 'professional'
         elif 'verified' in seat_types:
+            if 'honor' not in seat_types:
```
clintonb, line 89: You don't necessarily need a new course type. The current verified type still applies. Instead, modify `creatableSeatTypes` in `ecommerce/static/js/models/course_model.js` so that honor seats can no longer be created.

ecommerce/static/js/models/course_model.js
```diff
@@ -121,7 +127,7 @@ define([
              *
              * Note that audit seats cannot be created, only edited.
              */
-            creatableSeatTypes: ['honor', 'verified', 'professional', 'credit'],
+            creatableSeatTypes: ['audit', 'honor', 'verified', 'professional', 'credit'],
```
clintonb, line 130: Remove `'honor'` and update the comment above.

clintonb:
If you absolutely need an honor seat for some courses, perhaps this should be triggered in the UI (similar to how we toggle ID verification for professional education). Adding a new course type seems a bit strange to me.

ecommerce/courses/models.py
bderusha, line 89: @clintonb we still need to be able to create honor seats in case people need the old options for some reason.

PR #446, opened 2015-11-30
Update Pylint version.

peter-fogg:
@edx/ecommerce

### DIFF
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -2,6 +2,7 @@ import logging
 
 import analytics
 from django.dispatch import receiver
+from oscar.core.loading import get_class
 import waffle
 
 from ecommerce.courses.utils import mode_for_seat
@@ -9,7 +10,6 @@ from ecommerce.extensions.analytics.utils import is_segment_configured, parse_tr
 from ecommerce.extensions.checkout.utils import get_provider_data
 from ecommerce.notifications.notifications import send_notification
 from ecommerce.settings import get_lms_url
-from oscar.core.loading import get_class
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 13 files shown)


rlucioni:
:+1:

rlucioni:
FYI @AlasdairSwan, making a change like this one in https://github.com/edx/programs will fix the failures you saw earlier today.

peter-fogg:
FYI @renzo you may want to look again; upgrading seems to have introduced a few new violations to worry about which I've fixed.

ecommerce/extensions/checkout/signals.py
```diff
@@ -9,7 +10,6 @@
 from ecommerce.extensions.checkout.utils import get_provider_data
 from ecommerce.notifications.notifications import send_notification
 from ecommerce.settings import get_lms_url
-from oscar.core.loading import get_class
```
rlucioni, line 12: Did pylint catch this?

pylintrc
```diff
@@ -7,9 +7,6 @@
 # pygtk.require().
 #init-hook=''
 
-# Profiled execution.
-profile=no
```
rlucioni, line 11: Why are lines in this file being removed?

ecommerce/extensions/checkout/signals.py
peter-fogg, line 12: Yes.

pylintrc
peter-fogg, line 11: The removed lines are deprecated options which will be removed in Pylint 1.6.

clintonb:
:+1:

ecommerce/extensions/checkout/signals.py
rlucioni, line 12: That's awesome! We finally don't need to call these out on PRs. I'm so happy.

rlucioni:
:+1:

PR #447, opened 2015-12-02
Sending line item details to Cybersource

clintonb:
In addition to adding more info to each transaction, this may decrease our fees.

ECOM-3051

### DIFF
--- a/ecommerce/extensions/payment/migrations/0007_add_cybersource_level23_sample.py
+++ b/ecommerce/extensions/payment/migrations/0007_add_cybersource_level23_sample.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+SAMPLE_NAME = 'send_level_2_3_details_to_cybersource'
+
+
+def create_sample(apps, schema_editor):
+    """
+    Enable both existing payment processors.
+    """
+    Sample = apps.get_model('waffle', 'Sample')
+
+    Sample.objects.create(
+        name=SAMPLE_NAME,
+        percent=0,
+        note='When this sample is active, Level 2/3 transaction data will be sent to CyberSource.')
+
+
+def delete_sample(apps, schema_editor):
+    Sample = apps.get_model('waffle', 'Sample')
+    Sample.objects.get(name=SAMPLE_NAME).delete()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('payment', '0006_enable_payment_processors'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_sample, delete_sample)
+    ]
--- a/ecommerce/extensions/payment/processors/cybersource.py
+++ b/ecommerce/extensions/payment/processors/cybersource.py
@@ -1,9 +1,12 @@
 """ CyberSource payment processing. """
+from __future__ import unicode_literals
+
 import datetime
-from decimal import Decimal
 import logging
 import uuid
+from decimal import Decimal
 
+import waffle
 from django.conf import settings
 from oscar.apps.payment.exceptions import UserCancelled, GatewayError, TransactionDeclined
 from oscar.core.loading import get_model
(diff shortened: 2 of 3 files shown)


clintonb:
@rlucioni @mjfrey please take a look when you have time.

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -97,15 +99,38 @@ def get_transaction_parameters(self, basket, request=None):
...
+            for index, line in enumerate(basket.lines.all()):
+                parameters['item_{}_code'.format(index)] = line.product.get_product_class().slug
+                # Note (CCB): This indicates that the total_amount field below includes tax.
+                parameters['item_{}_gross_net_indicator'.format(index)] = 'Y'
+                parameters['item_{}_name'.format(index)] = line.product.title
+                parameters['item_{}_quantity'.format(index)] = line.quantity
+                parameters['item_{}_sku'.format(index)] = line.stockrecord.partner_sku
+                parameters['item_{}_tax_amount'.format(index)] = str(line.line_tax)
+                parameters['item_{}_tax_rate'.format(index)] = '0'
+                parameters['item_{}_total_amount '.format(index)] = str(line.line_price_incl_tax)
```
mjfrey: In doing the discount work I needed to calculate the item price less the discount other wise PayPal complained about the total not adding up from the line items.
line.price_incl_tax - line.discount_value
Maybe we want to do the same.

ecommerce/extensions/payment/tests/test_processors.py
```diff
@@ -96,51 +99,88 @@ def test_issue_credit_error(self):
...
+
+            for index, line in enumerate(self.basket.lines.all()):
+                expected['item_{}_code'.format(index)] = line.product.get_product_class().slug
+                expected['item_{}_gross_net_indicator'.format(index)] = 'Y'
+                expected['item_{}_name'.format(index)] = line.product.title
+                expected['item_{}_quantity'.format(index)] = line.quantity
+                expected['item_{}_sku'.format(index)] = line.stockrecord.partner_sku
+                expected['item_{}_tax_amount'.format(index)] = str(line.line_tax)
+                expected['item_{}_tax_rate'.format(index)] = '0'
+                expected['item_{}_total_amount '.format(index)] = str(line.line_price_incl_tax)
```
mjfrey: same here

ecommerce/extensions/payment/processors/cybersource.py
clintonb: Updated to use `line_price_incl_tax_incl_discounts`. Also, added `item_{}_discount_amount` key.

mjfrey:
:+1:

ecommerce/extensions/payment/migrations/0007_add_cybersource_level23_sample.py
```diff
@@ -0,0 +1,33 @@
...
+SAMPLE_NAME = 'send_level_2_3_details_to_cybersource'
+
+
+def create_sample(apps, schema_editor):
+    """
+    Enable both existing payment processors.
+    """
+    Sample = apps.get_model('waffle', 'Sample')
+
+    Sample.objects.create(
```
rlucioni, line 15: @clintonb why use a sample instead of a switch here? Would we ever want to use a percentage other than 100% or 0%?

clintonb, line 15: This is primarily for the initial rollout. We want to ramp it up to make sure we don't break for everyone.

clintonb:
An example payment is available at https://ebctest.cybersource.com/ebctest/transactionsearch/TransactionSearchDetailsLoad.do?requestSearch=true&requestId=4490125340885000001521. Note that it does not include the discount fields (which were added yesterday).

ecommerce/extensions/payment/migrations/0007_add_cybersource_level23_sample.py
rlucioni, line 15: Can't argue with that, it worked great for async.

clintonb, line 15: That's exactly why I added this a couple days ago after the async rollout success!

rlucioni:
:+1:. In the future, when switching a module to use `unicode_literals` (see what I did there?), it would be great if you could use a separate commit.

PR #452, opened 2015-12-03
Update getting_started.rst

jimabramson:
@rlucioni @srpearce not urgent, but mind taking a quick look?  if it makes sense to you I'll be making the exact same change to a few other repos with the same 'getting started' doc structure.


rlucioni:
:+1:

clintonb:
:+1:

PR #450, opened 2015-12-02
Voucher creation and generation of the CSV voucher report

marjev:
@clintonb, please review this PR. This is related to [SOL-1327](https://openedx.atlassian.net/browse/SOL-1327).

Here's the list of changes this PR introduces:
- Voucher creation (Range, Benefit, Condition and ConditionOffer are created under the hood)
- Tests that Voucher creation works as expected
- Test that Voucher usage works as expected (I wanted to make sure that applying a voucher to a basket containing a seat product works -- applies the benefit linked to a voucher)
- Voucher report generation in *.csv format
- Test that voucher report is formatted as expected

This work is continuation of work done by @vkaracic and @mjfrey. You'll see some changes to the lines merged by the first PR. Those are mostly replacements of `enrollment code` with `coupon`, since that's the new term we are using. Please notice the migration `0013_coupon_product_class.py`. It contains the reversal method that will should revert the changes introduced in `0012_enrollment_code_product_class`.

If you have any questions, please let me know.
Thanks!

PR #453, opened 2015-12-04
Created API endpoint for creating Coupon orders.

vkaracic:
This PR contains the API endpoint for coupon products, creating coupons, adding them to the basket, handling the order, fulfilling the order, with accompanying tests. (the test for the fulfillment module is not yet finished).

@clintonb can you please take a look at the code? Also note that #450 is a prerequisite for this so there is code that calls some models and methods from that PR, Travis will probably fail because of that too. But for now let's just fix any code issues that there might be. Thanks!

@mjfrey @marjev @dsego

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -26,6 +26,7 @@ ProductAttributeValue = get_model('catalogue', 'ProductAttributeValue')
 Refund = get_model('refund', 'Refund')
 Selector = get_class('partner.strategy', 'Selector')
 StockRecord = get_model('partner', 'StockRecord')
+Voucher = get_model('voucher', 'Voucher')
 
 COURSE_DETAIL_VIEW = 'api:v2:course-detail'
 PRODUCT_DETAIL_VIEW = 'api:v2:product-detail'
@@ -49,6 +50,11 @@ class ProductAttributeValueSerializer(serializers.ModelSerializer):
         return instance.attribute.name
 
     def get_value(self, obj):
+        if obj.attribute.name == 'Coupon vouchers':
+            request = self.context.get('request')
+            vouchers = obj.value.vouchers.all()
+            serializer = VoucherSerializer(vouchers, many=True, context={'request': request})
+            return serializer.data
         return obj.value
 
     class Meta(object):
@@ -77,11 +83,22 @@ class PartialStockRecordSerializerForUpdate(StockRecordSerializer):
 
 class ProductSerializer(serializers.HyperlinkedModelSerializer):
     """ Serializer for Products. """
-    attribute_values = ProductAttributeValueSerializer(many=True, read_only=True)
+    attribute_values = serializers.SerializerMethodField()
     product_class = serializers.SerializerMethodField()
     price = serializers.SerializerMethodField()
     is_available_to_buy = serializers.SerializerMethodField()
     stockrecords = StockRecordSerializer(many=True, read_only=True)
+
+    def get_attribute_values(self, product):
+        request = self.context.get('request')
+        attributes = product.attr
+        serializer = ProductAttributeValueSerializer(
+            attributes,
+            many=True,
+            read_only=True,
+            context={'request': request}
+        )
+        return serializer.data
 
     def get_product_class(self, product):
         return product.get_product_class().name
@@ -331,4 +348,24 @@ class CatalogSerializer(serializers.ModelSerializer):
             'api:v2:catalog-product-list',
             kwargs={'parent_lookup_stockrecords__catalogs': obj.id},
             request=self.context['request']
+        )
+
+
+class VoucherSerializer(serializers.ModelSerializer):
+    is_available_to_user = serializers.SerializerMethodField()
+    benefit = serializers.SerializerMethodField()
+
+    def get_is_available_to_user(self, obj):
+        request = self.context.get('request')
+        return obj.is_available_to_user(user=request.user)
+
+    def get_benefit(self, obj):
+        return (obj.offers.first().benefit.type, obj.offers.first().benefit.value)
+
+    class Meta(object):
+        model = Voucher
+        fields = (
+            'id', 'name', 'code', 'usage', 'start_datetime', 'end_datetime',
+            'num_basket_additions', 'num_orders', 'total_discount',
+            'date_created', 'offers', 'is_available_to_user', 'benefit'
         )
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -1,0 +1,261 @@
+from __future__ import unicode_literals
+
+import datetime
+import json
+
+from django.core.urlresolvers import reverse
+from django.db.utils import IntegrityError
+from django.test import RequestFactory
+from oscar.core.loading import get_model
+
+from ecommerce.core.models import Client
+from ecommerce.extensions.api.constants import APIConstants as AC
+from ecommerce.extensions.api.v2.views.coupons import CouponOrderCreateView
+from ecommerce.extensions.test.factories import create_coupon
+from ecommerce.tests.factories import SiteFactory, SiteConfigurationFactory
+from ecommerce.tests.testcases import TestCase
+
+Basket = get_model('basket', 'Basket')
+Benefit = get_model('offer', 'Benefit')
+Catalog = get_model('catalogue', 'Catalog')
+Course = get_model('courses', 'Course')
+Order = get_model('order', 'Order')
+Product = get_model('catalogue', 'Product')
+ProductClass = get_model('catalogue', 'ProductClass')
+StockRecord = get_model('partner', 'StockRecord')
+Voucher = get_model('voucher', 'Voucher')
+
+COUPONS_LINK = reverse('api:v2:coupons:create')
+
+
+class CouponOrderCreateViewTest(TestCase):
+    """Unit tests for creating coupon order."""
+
+    def setUp(self):
+        super(CouponOrderCreateViewTest, self).setUp()
+        self.user = self.create_user(is_staff=True)
+        self.client.login(username=self.user.username, password=self.password)
+
+        course_id = 'edx/Demo_Course/DemoX'
+        course = Course.objects.create(id=course_id)
+        course.create_or_update_seat('verified', True, 50, self.partner)
+
+        self.catalog = Catalog.objects.create(partner=self.partner)
+        self.product_class, __ = ProductClass.objects.get_or_create(name='Coupon')
+
+    def test_create(self):
+        """Test the create method."""
+        site_configuration = SiteConfigurationFactory(partner__name='TestX')
+        site = SiteFactory()
+        site.siteconfiguration = site_configuration
+        data = {
+            'title': 'Test coupon',
+            'client_username': 'Client',
+            'stock_record_ids': [1],
+            'start_date': '2015-1-1',
+            'end_date': '2020-1-1',
+            'code': '',
+            'benefit_type': Benefit.PERCENTAGE,
+            'benefit_value': 100,
+            'voucher_type': Voucher.SINGLE_USE,
+            'quantity': 1,
+            'price': 100
+        }
+        request = RequestFactory()
+        request.data = data
+        request.site = site
+
+        response = CouponOrderCreateView().create(request)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertDictEqual(
+            response.data,
+            {'payment_data': {'payment_processor_name': 'Invoice'}, 'id': 1, 'order': 1, 'coupon_id': 3}
+        )
+
+    def test_create_coupon_product(self):
+        """Test the created coupon data."""
+        coupon = create_coupon()
+        self.assertEqual(Product.objects.filter(product_class=self.product_class).count(), 1)
+        self.assertIsInstance(coupon, Product)
+        self.assertEqual(coupon.title, 'Test coupon')
+
+        self.assertEqual(StockRecord.objects.filter(product=coupon).count(), 1)
+        stock_record = StockRecord.objects.get(product=coupon)
+        self.assertEqual(stock_record.price_currency, 'USD')
+        self.assertEqual(stock_record.price_excl_tax, 100)
+
+        self.assertEqual(coupon.attr.coupon_vouchers.vouchers.count(), 5)
+
+    def test_append_to_existing_coupon(self):
+        """Test adding additional vouchers to an existing coupon."""
+        create_coupon(partner=self.partner, catalog=self.catalog)
+        data = {
+            'partner': self.partner,
+            'benefit_type': Benefit.PERCENTAGE,
+            'benefit_value': 100,
+            'catalog': self.catalog,
+            'end_date': datetime.date(2020, 1, 1),
+            'code': '',
+            'quantity': 2,
+            'start_date': datetime.date(2015, 1, 1),
+            'voucher_type': Voucher.MULTI_USE
+        }
+        coupon_append = CouponOrderCreateView().create_coupon_product(
+            title='Test coupon',
+            price=100,
+            data=data
+        )
+
+        self.assertEqual(Product.objects.filter(product_class=self.product_class).count(), 1)
+        self.assertEqual(StockRecord.objects.filter(product=coupon_append).count(), 1)
+        self.assertEqual(coupon_append.attr.coupon_vouchers.vouchers.count(), 7)
+        self.assertEqual(coupon_append.attr.coupon_vouchers.vouchers.filter(usage=Voucher.MULTI_USE).count(), 2)
+
+    def test_custom_code_string(self):
+        """Test creating a coupon with custom voucher code."""
+        data = {
+            'partner': self.partner,
+            'benefit_type': Benefit.PERCENTAGE,
+            'benefit_value': 100,
+            'catalog': self.catalog,
+            'end_date': datetime.date(2020, 1, 1),
+            'code': 'CUSTOMCODE',
+            'quantity': 1,
+            'start_date': datetime.date(2015, 1, 1),
+            'voucher_type': Voucher.ONCE_PER_CUSTOMER
+        }
+        custom_coupon = CouponOrderCreateView().create_coupon_product(
+            title='Custom coupon',
+            price=100,
+            data=data
+        )
+        self.assertEqual(custom_coupon.attr.coupon_vouchers.vouchers.count(), 1)
+        self.assertEqual(custom_coupon.attr.coupon_vouchers.vouchers.first().code, 'CUSTOMCODE')
+
+    def test_custom_code_integrity_error(self):
+        """Test custom coupon code duplication."""
+        data = {
+            'partner': self.partner,
+            'benefit_type': Benefit.PERCENTAGE,
+            'benefit_value': 100,
+            'catalog': self.catalog,
+            'end_date': datetime.date(2020, 1, 1),
+            'code': 'CUSTOMCODE',
+            'quantity': 1,
+            'start_date': datetime.date(2015, 1, 1),
+            'voucher_type': Voucher.SINGLE_USE
+        }
+        CouponOrderCreateView().create_coupon_product(
+            title='Custom coupon',
+            price=100,
+            data=data
+        )
+
+        with self.assertRaises(IntegrityError):
+            CouponOrderCreateView().create_coupon_product(
+                title='Coupon with integrity issue',
+                price=100,
+                data=data
+            )
+
+    def test_add_product_to_basket(self):
+        """Test adding a coupon product to a basket."""
+        coupon = create_coupon(partner=self.partner)
+        coupon_client = Client.objects.create(username='TestX')
+        basket = CouponOrderCreateView().add_product_to_basket(
+            product=coupon,
+            client=coupon_client,
+            site=self.site,
+            partner=self.partner
+        )
+
+        self.assertIsInstance(basket, Basket)
+        self.assertEqual(Basket.objects.count(), 1)
+        self.assertEqual(basket.lines.count(), 1)
+        self.assertEqual(basket.lines.first().price_excl_tax, 100)
+
+    def test_create_order(self):
+        """Test the order creation."""
+        coupon = create_coupon(partner=self.partner)
+        coupon_client = Client.objects.create(username='TestX')
+        basket = CouponOrderCreateView().add_product_to_basket(
+            product=coupon,
+            client=coupon_client,
+            site=self.site,
+            partner=self.partner
+        )
+        response_data = CouponOrderCreateView().create_order_for_invoice(basket, coupon_id=coupon.id)
+        self.assertEqual(response_data[AC.KEYS.BASKET_ID], 1)
+        self.assertEqual(response_data[AC.KEYS.ORDER], 1)
+        self.assertEqual(response_data[AC.KEYS.PAYMENT_DATA][AC.KEYS.PAYMENT_PROCESSOR_NAME], 'Invoice')
+
+        self.assertEqual(Order.objects.count(), 1)
+        self.assertEqual(Order.objects.first().status, 'Complete')
+        self.assertEqual(Order.objects.first().total_incl_tax, 100)
+        self.assertEqual(Basket.objects.first().status, 'Submitted')
+
+
+class CouponOrderCreateViewFunctionalTest(TestCase):
+    """Test the coupon order creation functionality."""
+
+    def setUp(self):
+        super(CouponOrderCreateViewFunctionalTest, self).setUp()
+        self.user = self.create_user(is_staff=True)
+        self.client.login(username=self.user.username, password=self.password)
+
+        course_id = 'edx/Demo_Course/DemoX'
+        course = Course.objects.create(id=course_id)
+        course.create_or_update_seat('verified', True, 50, self.partner)
+
+        course_id = 'edx/Demo_Course2/DemoX'
+        course = Course.objects.create(id=course_id)
+        course.create_or_update_seat('verified', True, 100, self.partner)
+
+        self.data = {
+            'title': 'Test coupon',
+            'client_username': 'TestX',
+            'stock_record_ids': [1, 2],
+            'start_date': '2015-01-01',
+            'end_date': '2020-01-01',
+            'code': '',
+            'benefit_type': Benefit.PERCENTAGE,
+            'benefit_value': 100,
+            'voucher_type': Voucher.SINGLE_USE,
+            'quantity': 2,
+            'price': 100
+        }
+        self.response = self.client.post(COUPONS_LINK, data=self.data, format='json')
+
+    def test_response(self):
+        """Test the response data given after the order was created."""
+        self.assertEqual(self.response.status_code, 200)
+        response_data = json.loads(self.response.content)
+        self.assertEqual(response_data[AC.KEYS.COUPON_ID], 5)
+        self.assertEqual(response_data[AC.KEYS.BASKET_ID], 1)
+        self.assertEqual(response_data[AC.KEYS.ORDER], 1)
+        self.assertEqual(response_data[AC.KEYS.PAYMENT_DATA][AC.KEYS.PAYMENT_PROCESSOR_NAME], 'Invoice')
+
+    def test_order(self):
+        """Test the order data after order creation."""
+        self.assertEqual(Order.objects.count(), 1)
+        self.assertEqual(Order.objects.first().status, 'Complete')
+        self.assertEqual(Order.objects.first().lines.count(), 1)
+        self.assertEqual(Order.objects.first().lines.first().product.title, 'Test coupon')
+
+    def test_authentication_required(self):
+        """Test that a guest cannot access the view."""
+        response = self.client.post(COUPONS_LINK, data=self.data)
+        self.assertEqual(response.status_code, 200)
+
+        self.client.logout()
+        response = self.client.post(COUPONS_LINK, data=self.data)
+        self.assertEqual(response.status_code, 401)
+
+    def test_authorization_required(self):
+        """Test that a non-staff user cannot access the view."""
+        user = self.create_user(is_staff=False)
+        self.client.login(username=user.username, password=self.password)
+
+        response = self.client.post(COUPONS_LINK, data=self.data)
+        self.assertEqual(response.status_code, 403)
--- a/ecommerce/extensions/api/v2/tests/views/test_products.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_products.py
@@ -9,10 +9,14 @@ import pytz
 from ecommerce.courses.models import Course
 from ecommerce.extensions.api.v2.tests.views import JSON_CONTENT_TYPE, ProductSerializerMixin
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.extensions.test.factories import create_coupon
 from ecommerce.tests.testcases import TestCase
 
+Benefit = get_model('offer', 'Benefit')
+Catalog = get_model('catalogue', 'Catalog')
 Product = get_model('catalogue', 'Product')
 ProductClass = get_model('catalogue', 'ProductClass')
+Voucher = get_model('voucher', 'Voucher')
 
 
 class ProductViewSetTests(ProductSerializerMixin, CourseCatalogTestMixin, TestCase):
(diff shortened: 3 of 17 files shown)


clintonb:
@rlucioni @jimabramson can one of you take a look at this as well, please?

ecommerce/extensions/api/serializers.py
```diff
@@ -332,3 +333,27 @@ def get_products(self, obj):
...
+    class Meta(object):
+        model = Voucher
+
+
+class CouponSerializer(serializers.ModelSerializer):
+    price = serializers.SerializerMethodField()
+    vouchers = serializers.SerializerMethodField()
+
+    def get_price(self, obj):
+        stock_record = StockRecord.objects.get(product=obj)
```
clintonb: See `ProductSerializer.get_price()`. It might be best to extract that into a separate function/mixin method.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -0,0 +1,236 @@
...
+
+        self.coupon = CouponOrderCreateView().create_coupon_product(
+            title='Test coupon',
+            price=100,
+            data=self.data
+        )
+
+    def test_list_coupons(self):
+        """Test coupon API endpoint list."""
+        response = self.client.get('/api/v2/coupons/')
```
clintonb: Use `reverse`.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -0,0 +1,236 @@
...
+        self.assertEqual(response.status_code, 200)
+        result = json.loads(response.content)['results']
+        self.assertEqual(len(result), 1)
+        self.assertEqual(result[0]['price'], 100)
+        self.assertEqual(len(result[0]['vouchers']), 5)
+
+        # If there are no products it returns an empty response.
+        Product.objects.all().delete()
+        response = self.client.get('/api/v2/coupons/')
```
clintonb: Use `reverse`.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -0,0 +1,236 @@
...
+
+        course_id = 'edx/Demo_Course/DemoX'
+        course = Course.objects.create(id=course_id)
+        course.create_or_update_seat('verified', True, 50, self.partner)
+
+        self.coupon_client = Client.objects.create(username='TestX')
+        self.catalog = Catalog.objects.create(partner=self.partner)
+        self.product_class, __ = ProductClass.objects.get_or_create(name='Coupon')
+
+        self.data = {
```
clintonb: This does not need to be an instance variable.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -0,0 +1,236 @@
...
+            'benefit_value': 100,
+            'catalog': self.catalog,
+            'end_date': datetime.date(2020, 1, 1),
+            'code': '',
+            'quantity': 5,
+            'start_date': datetime.date(2015, 1, 1),
+            'voucher_type': Voucher.SINGLE_USE
+        }
+
+        self.coupon = CouponOrderCreateView().create_coupon_product(
```
clintonb: This isn't needed by all tests. Move to a method, and call only when needed.

clintonb: Also, use a factory.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
clintonb: API tests belong in the api app.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -0,0 +1,263 @@
...
+Basket = get_model('basket', 'Basket')
+Catalog = get_model('catalogue', 'Catalog')
+logger = logging.getLogger(__name__)
+Order = get_model('order', 'Order')
+Product = get_model('catalogue', 'Product')
+ProductClass = get_model('catalogue', 'ProductClass')
+StockRecord = get_model('partner', 'StockRecord')
+
+
+class CouponOrderCreateView(generics.CreateAPIView, EdxOrderPlacementMixin, NonDestroyableModelViewSet):
```
clintonb: Combining `NonDestroyableModelViewSet` and `CreateAPIView` is odd. Why are you doing this?

ecommerce/extensions/api/v2/urls.py
```diff
@@ -21,6 +23,14 @@
     ),
 ]
 
+COUPON_URLS = [
+    url(r'^$', coupons_views.CouponOrderCreateView.as_view({'get': 'list'}), name='create'),
+]
+
+VOUCHER_URLS = [
+    url(r'^$', vouchers_views.VoucherViewSet.as_view({'get': 'list'}), name='list'),
```
clintonb: If you only need one access method, why bother with a viewset?

ecommerce/extensions/api/v2/urls.py
```diff
@@ -21,6 +23,14 @@
     ),
 ]
 
+COUPON_URLS = [
+    url(r'^$', coupons_views.CouponOrderCreateView.as_view({'get': 'list'}), name='create'),
```
clintonb: If you only need one access method, why bother with a viewset?

ecommerce/extensions/catalogue/utils.py
```diff
@@ -1,22 +1,73 @@
...
+            if set(catalog.stock_records.all()) == set(stock_records):
+                return catalog
+
+    catalog = Catalog.objects.create(name=name, partner=partner)
+    for stock_record in stock_records:
+        catalog.stock_records.add(stock_record)
+    return catalog
+
+
+def generate_upc(partner, title, catalog):
```
clintonb: We don't actually use UPCs. Also, UPCs follow a standard: https://en.wikipedia.org/wiki/Universal_Product_Code. This should be removed.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -0,0 +1,263 @@
...
+            A coupon product object.
+        """
+
+        upc = generate_upc(
+            title=title,
+            catalog=data['catalog'],
+            partner=data['partner']
+        )
+        product_class = ProductClass.objects.get(slug='coupon')
+        coupon_product, created = Product.objects.get_or_create(
```
clintonb: Use some combination of title, catalog, and partner to differentiate the products you are creating.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if created:
+            coupon_product.title = title
+
+            stock_record = StockRecord.objects.create(
+                product=coupon_product,
+                partner=data['partner'],
+                partner_sku=sku
+            )
+            stock_record.price_currency = 'USD'
```
clintonb: Move these to `create` and avoid the extra DB call.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -0,0 +1,263 @@
...
+            upc=upc
+        )
+
+        sku = generate_sku(
+            product=coupon_product,
+            partner=data['partner'],
+            catalog=data['catalog'],
+        )
+
+        if created:
```
clintonb: These two cases can be combined:

``` python
stock_record, _created = StockRecord.objects.get_or_create(product=coupon_product, partner=data['partner'], partner_sku=sku)
stock_record.price_currency = 'USD'
stock_record.price_excl_tax = price
stock_record.save()
```

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -0,0 +1,263 @@
...
+            stock_record = StockRecord.objects.get(
+                partner=data['partner'],
+                partner_sku=sku
+            )
+            stock_record.price_excl_tax = price
+            stock_record.save()
+
+        # Product validation.
+        try:
+            coupon_product.clean()
```
clintonb: Shouldn't this be validated before we go through the effort of creating stock records?

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -0,0 +1,263 @@
...
+
+        return coupon_product
+
+    def add_product_to_basket(self, product, client, site):
+        """Adds the coupon product to the user's basket."""
+        basket = Basket.get_basket(client, site)
+        basket.add_product(product)
+        logger.info(
+            'Added product with SKU [%s] to basket [%d]',
+            product.stockrecords.first().partner_sku, basket.id
```
clintonb: That's now how you get the SKU. Use `product.stockrecords.filter(partner=get_partner_for_site(site)).partner_sku`.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -0,0 +1,263 @@
...
+        response_data[AC.KEYS.ORDER] = order.id
+        logger.info(
+            'Created new order number [%s] from basket [%d]',
+            order_metadata[AC.KEYS.ORDER_NUMBER],
+            basket.id
+        )
+
+        # Invoice payment processor invocation.
+        payment_processor = InvoicePayment
+        payment_processor().handle_processor_response(response={}, basket=basket)
```
clintonb: Payment processing should occur before order placement.

ecommerce/extensions/catalogue/utils.py
vkaracic: Unfortunately there is no other way to differentiate a coupon (except for creating and deleting objects in the same method which wouldn't be ideal). To be precise, in the time of creating a coupon the only elements for creating it are `title` (which is not unique) and `product_class` (a constant) - none of those can be used as parameters for a `get_or_create` function.

mjfrey: So everyone is in agreement.  The goal here is to create a unique product and re-use / update an existing one if the same parameters are used in the Coupon creation form.

ecommerce/extensions/api/v2/views/coupons.py
vkaracic: `get_partner_for_site` expects the `request` as an argument. Can I just send the `partner` and:
`product.stockrecords.filter(partner=partner).first().partner_sku`, instead of the whole request?

ecommerce/extensions/catalogue/utils.py
clintonb: If that's the use case, this works. Just don't call it UPC.

vkaracic: It has to be UPC, because that's the only unique attribute that we can use from Oscar's Product model: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/catalogue/abstract_models.py#L267

I guess I could change the algorithm of creating it a bit. Or we'll need to adjust the Product model which we're trying hard to avoid.

clintonb: UPC is not the only unique attribute. We use a code/slug field for seats.
See Course.create_or_update_seat().

## 

Clinton Blackburn

ecommerce/extensions/api/v2/views/coupons.py
clintonb: That works.

ecommerce/extensions/catalogue/utils.py
vkaracic: Okay, I've overlooked the `slug` attribute in the Product model, sorry.

vkaracic:
I don't quite understand the test error:
``OperationalError: no such table: catalogue_productclass``
ProductClass is Oscar's model.

vkaracic:
@clintonb I believe I have addressed all your comments, I'll work on fixing the coverage, in the meantime can you take a second look?

ecommerce/extensions/api/serializers.py
```diff
@@ -332,3 +339,34 @@ def get_products(self, obj):
...
+
+    class Meta(object):
+        model = Voucher
+
+
+class CouponSerializer(serializers.ModelSerializer):
+    price = serializers.SerializerMethodField()
+    vouchers = serializers.SerializerMethodField()
+
+    def get_price(self, product):
```
clintonb: Move this, `get_info`, and the `price` field to a mixin shared with `ProductSerializer`.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -0,0 +1,15 @@
...
+from ecommerce.extensions.api import serializers
+
+
+Voucher = get_model('voucher', 'Voucher')
+
+
+class VoucherViewSet(viewsets.ReadOnlyModelViewSet):
+    queryset = Voucher.objects.all()
+    serializer_class = serializers.VoucherSerializer
+    permission_classes = (IsAuthenticated, IsAdminUser,)
```
clintonb: What's the use case of this API?

vkaracic: At some point we used it, but we've moved the listing to the coupons API. It's dead code, I'll remove it.

ecommerce/extensions/api/serializers.py
vkaracic: I've created a new mixin.

vkaracic:
@clintonb quality satisfied, and your last comments addressed, can you please take another look?

ecommerce/extensions/api/mixins.py
```diff
@@ -0,0 +1,19 @@
+from oscar.core.loading import get_class
+from rest_framework import serializers
+
+Selector = get_class('partner.strategy', 'Selector')
+
+
+class ProductInfoMixin(serializers.ModelSerializer):
```
clintonb: 1. This can/should reside in `serializers.py`.
2. Make it clear that this is intended for mixing with serializers, and avoid weird inheritance issues: `class ProductInfoSerializerMixin(object):`.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -0,0 +1,284 @@
...
+            'code': '',
+            'benefit_type': Benefit.PERCENTAGE,
+            'benefit_value': 100,
+            'voucher_type': Voucher.SINGLE_USE,
+            'quantity': 1,
+            'price': 100
+        }
+        request = RequestFactory()
+        request.data = data
+        request.site = site
```
clintonb: You can use `self.site`.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -0,0 +1,284 @@
...
+
+        response = CouponOrderCreateView().create(request)
```
clintonb, line 68: Call the API via HTTP: `self.client.get(...)`.

clintonb, line 68: I see what you're doing. You're actually separating unit/functional tests. That's fine. We generally start at the functional level for views since HTTP is the only supported interface.

vkaracic, line 68: I'm not trying to call the API, but to test the `create` method (which expects a request).

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -0,0 +1,240 @@
...
+Basket = get_model('basket', 'Basket')
+Catalog = get_model('catalogue', 'Catalog')
+logger = logging.getLogger(__name__)
+Order = get_model('order', 'Order')
+Product = get_model('catalogue', 'Product')
+ProductClass = get_model('catalogue', 'ProductClass')
+StockRecord = get_model('partner', 'StockRecord')
+
+
+class CouponOrderCreateView(EdxOrderPlacementMixin, NonDestroyableModelViewSet):
```
clintonb: If this is just a creation view, you don't need a `ViewSet`.

ecommerce/extensions/catalogue/migrations/0013_coupon_product_class.py
```diff
@@ -24,7 +24,7 @@ def create_product_class(apps, schema_editor):
         name='Coupon vouchers',
         code='coupon_vouchers',
         type='entity',
-        required=False
+        required=True
```
clintonb: You cannot modify migrations this way. This migration has already been run in production. Make a new migration with your changes.

ecommerce/extensions/catalogue/tests/test_utils.py
```diff
@@ -14,9 +33,79 @@ def test_generate_sku_for_course_seat(self):
         certificate_type = 'honor'
         product = course.create_or_update_seat(certificate_type, False, 0, self.partner)
 
-        _hash = u'{} {} {} {} {}'.format(certificate_type, course_id, 'False', '', self.partner.id)
```
clintonb, line 17: Why the change?

vkaracic, line 17: When there is `from __future__ import unicode_literals` is it needed anymore to explicitly add `u` in front of strings?

clintonb, line 17: Okay.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -281,3 +281,52 @@ def revoke_line(self, line):
...
+        logger.info("Attempting to fulfill 'Coupon' product types for order [%s]", order.number)
+
+        for line in lines:
+            line.set_status(LINE.COMPLETE)
+
+        logger.info("Finished fulfilling 'Coupon' product types for order [%s]", order.number)
+        return order, lines
+
+    def revoke_line(self, line):  # pragma: no cover
+        raise NotImplementedError("Revoke method not implemented!")
```
clintonb, line 340: How will this be implemented?

vkaracic: @mfrey

On Fri, Dec 11, 2015 at 2:33 PM, Clinton Blackburn <notifications@github.com
> wrote:

> In ecommerce/extensions/fulfillment/modules.py
> <https://github.com/edx/ecommerce/pull/453#discussion_r47354769>:
>
> > +            order (Order): The Order associated with the lines to be fulfilled.
> > +            lines (List of Lines): Order Lines, associated with purchased products in an Order. These should only
> > +                be 'Coupon' products.
> > +        Returns:
> > +            The original set of lines, with new statuses set based on the success or failure of fulfillment.
> > +        """
> > +        logger.info("Attempting to fulfill 'Coupon' product types for order [%s]", order.number)
> > +
> > +        for line in lines:
> > +            line.set_status(LINE.COMPLETE)
> > +
> > +        logger.info("Finished fulfilling 'Coupon' product types for order [%s]", order.number)
> > +        return order, lines
> > +
> > +    def revoke_line(self, line):  # pragma: no cover
> > +        raise NotImplementedError("Revoke method not implemented!")
>
> How will this be implemented?
>
> —
> <https://github.com/edx/ecommerce/pull/453/files#r47354769>.
>



-- 
Vedran Karačić

Mail:     vedran@edx.org
             vkaracic@extensionengine.com
Github: vkaracic <https://github.com/vkaracic>
Skype:  karacicvedran

clintonb:
@vkaracic minor issues identified. Let me know when they are addressed, and I'll review a final time.

ecommerce/extensions/fulfillment/modules.py
mjfrey, line 340: I don't think this has been decided yet.  I have had a discussion with product regarding refunds of Coupons, but no decision has been made.  For example what if 5 of 10 Vouchers have been redeemed and then we want to refund. Not sure how we do that.

mfrey, line 340: @vkaracic could you please remove me from your comment? i'm obviously not @mjfrey and i don't want to receive notifications about this pull request. thanks!

EDIT: well, the unsubscribe button is quite hidden (if you open the PR by mail there is no unsubscribe button). anyway, i've unsubscribed myself.

clintonb, line 340: Correction: @mfrey you can unsubscribe from the PR.

vkaracic, line 340: @mfrey sorry, was a typo :)

dsego, line 340: :joy:

vkaracic:
@clintonb ready for final review. Note that I have removed the coupons serializer because the coupons api endpoint won't list coupons, I've moved that to the products api endpoint where it belongs, therefor there won't be a need for the mixin class.

ecommerce/extensions/catalogue/migrations/0014_alter_couponvouchers_attribute.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0001_initial'),
+        ('catalogue', '0013_coupon_product_class')
+    ]
+    operations = [
+        migrations.RunPython(alter_couponvouchers_attribute)
```
clintonb: Reverse migration needed.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -0,0 +1,237 @@
...
+                code=data['code'] or None,
+                quantity=int(data['quantity']),
+                start_datetime=data['start_date'],
+                voucher_type=data['voucher_type']
+            )
+        except IntegrityError as ex:
+            logger.exception('Failed to create vouchers for [%s] coupon.', coupon_product.title)
+            raise IntegrityError(ex)  # pylint: disable=nonstandard-exception
+
+        coupon_vouchers = CouponVouchers.objects.get(coupon=coupon_product)
```
clintonb, line 164: @mjfrey remind me why we are creating vouchers here instead of during fulfillment.

mjfrey, line 164: @clintonb  We discussed this last week.  The first implementation was during fulfillment, but you suggested moving it to before.  I can not actually remember the reason why.  I think we discussed having a real invoice processor so we wanted the vouchers created before and not in fulfillment.

mjfrey, line 164: I think we also discussed the fact that we wanted the vouchers to part of the order as lines so we could reference them.  If we did them in fulfillment they would not be part of the line(s)

ecommerce/extensions/fulfillment/modules.py
marjev, line 340: I think you don't need to re-implement this method. It should be inherited from `BaseFulfillmentModule`.

vkaracic, line 340: I thought so too, but once I remove it I get this error:

```
2015-12-14 03:43:38,305 ERROR 6889 [ecommerce.extensions.fulfillment.api] api.py:67 - An unexpected error occurred while fulfilling order [EDX-100001].
  File "/home/vedran/Development/edX/otto/ecommerce/ecommerce/extensions/fulfillment/api.py", line 54, in fulfill_order
    module = module_class()
TypeError: Can't instantiate abstract class CouponFulfillmentModule with abstract methods revoke_line
```

mjfrey:
@clintonb can we get a last look

clintonb:
Rebase on master to fix the migration bug.

openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=453

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -0,0 +1,278 @@
...
+        self.assertEqual(response_data[AC.KEYS.ORDER], 1)
+        self.assertEqual(response_data[AC.KEYS.PAYMENT_DATA][AC.KEYS.PAYMENT_PROCESSOR_NAME], 'Invoice')
+
+        self.assertEqual(Order.objects.count(), 1)
+        self.assertEqual(Order.objects.first().status, 'Complete')
+        self.assertEqual(Order.objects.first().total_incl_tax, 100)
+        self.assertEqual(Basket.objects.first().status, 'Submitted')
+
+
+class CouponOrderCreateViewFunctionalTest(TestCase):
```
clintonb: You need to test authentication/authorization.

ecommerce/extensions/api/v2/views/coupons.py
clintonb, line 164: Okay.

ecommerce/extensions/catalogue/migrations/0014_alter_couponvouchers_attribute.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from oscar.core.loading import get_model
+
+ProductAttribute = get_model("catalogue", "ProductAttribute")
+
+
+def alter_couponvouchers_attribute(apps, shema_editor):
```
clintonb: schema_editor

ecommerce/extensions/catalogue/migrations/0014_alter_couponvouchers_attribute.py
```diff
@@ -0,0 +1,32 @@
...
+    """Change the coupon_vouchers product attribute to be required."""
+    coupon_vouchers = ProductAttribute.objects.get(code='coupon_vouchers')
+    coupon_vouchers.required = True
+    coupon_vouchers.save()
+
+
+def reverse_migration(apps, shema_editor):
```
clintonb: schema_editor

ecommerce/extensions/catalogue/utils.py
```diff
@@ -1,22 +1,74 @@
...
-        product.attr.course_key,
-        unicode(product.attr.id_verification_required),
-        getattr(product.attr, 'credit_provider', ''),
+    # Note: This currently supports coupons and seats.
+    # A new product type can be added via a new else if block.
+    if product.product_class and product.product_class.name == 'Coupon':
+        catalog = kwargs.get('catalog', '')
+        _hash = ' '.join((
+            unicode(product.id),
+            unicode(catalog.name),
```
clintonb: The name can change. Use `catalog.id`.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -281,3 +281,52 @@ def revoke_line(self, line):
             logger.exception('Failed to revoke fulfillment of Line [%d].', line.id)
 
         return False
+
+
+class CouponFulfillmentModule(BaseFulfillmentModule):
```
clintonb, line 286: This needs to be tested.

ecommerce/extensions/api/mixins.py
vkaracic: Actually when I inherit from `object` I get `ImproperlyConfigured` error:

```
Field name `price` is not valid for model `Product`.
```

vkaracic:
@clintonb I have updated the PR and addressed your latests comments. A couple of things to note, I have added the viewset again because we will be listing coupon products in the /coupons API endpoint (https://github.com/edx/ecommerce/pull/460#discussion_r47595671). The other thing is that I removed this dispatch method: https://github.com/vkaracic/ecommerce/commit/d31ef2359b99bae7fca79617a982e44861b1eb78#diff-c9a451f0afeebe0481e4cee1c16b7c94L44 because it gave out 
`TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.`
errors in the authentication/authorization tests. Without it all the tests pass, so I don't think that we need it, but I would like to hear your opinion about it.

ecommerce/extensions/api/serializers.py
```diff
@@ -26,11 +26,28 @@
 Refund = get_model('refund', 'Refund')
 Selector = get_class('partner.strategy', 'Selector')
 StockRecord = get_model('partner', 'StockRecord')
+Voucher = get_model('voucher', 'Voucher')
 
 COURSE_DETAIL_VIEW = 'api:v2:course-detail'
 PRODUCT_DETAIL_VIEW = 'api:v2:product-detail'
 
 
+class ProductInfoSerializerMixin(serializers.ModelSerializer):
```
clintonb: This should inherit for `object`, not `ModelSerializer`. I'm surprised there haven't been MRO errors.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -0,0 +1,295 @@
...
+        self.assertEqual(response_data[AC.KEYS.PAYMENT_DATA][AC.KEYS.PAYMENT_PROCESSOR_NAME], 'Invoice')
+
+    def test_order(self):
+        """Test the order data after order creation."""
+        self.assertEqual(Order.objects.count(), 1)
+        self.assertEqual(Order.objects.first().status, 'Complete')
+        self.assertEqual(Order.objects.first().lines.count(), 1)
+        self.assertEqual(Order.objects.first().lines.first().product.title, 'Test coupon')
+
+    def test_authorization_required(self):
```
clintonb: `test_authentication_required`

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -0,0 +1,295 @@
...
+        """Test that a guest cannot access the view."""
+        response = self.client.get(COUPONS_LINK)
+        self.assertEqual(response.status_code, 200)
+
+        self.client.logout()
+
+        response = self.client.get(COUPONS_LINK)
+        self.assertEqual(response.status_code, 401)
+
+    def test_authentication_required(self):
```
clintonb: `test_authorization_required`

ecommerce/extensions/api/serializers.py
vkaracic: I get MRO errors when inheriting from `serializers.Serializer`, when inheriting from object: https://github.com/edx/ecommerce/pull/453#discussion_r47650815

clintonb:
@vkaracic 
1. Rather than add another endpoint exposing products, add filtering to the existing one.
2. Did you investigate the error that occurred in the atomic block before removing it? The atomicity code has worked in production for few months. Please investigate the error to make sure removing that code is the best solution.

clintonb:
I'm seeing `catalog.name` used in many places. Names can change. IDs cannot. Consider using `id` instead of `name`, especially in places where we are generating codes/slugs that will be used for uniqueness or matching.

mjfrey:
@clintonb we did have this under the /products endpoint but you suggested moving it to its own since there are a lot of products.  So lets pick one way and stick to it.

clintonb:
Did you have general filtering, or a separate /api/v2/products/coupons/
endpoint? I am proposing generic filtering by product class (e.g.
/api/v2/products/?product_class=coupon). See
http://www.django-rest-framework.org/api-guide/filtering/#generic-filtering.

ecommerce/extensions/api/v2/views/coupons.py
clintonb, line 164: Please add a comment explaining why we are creating the order immediately.

ecommerce/extensions/api/serializers.py
clintonb: Leave as-is. I don't have a quick fix and, ultimately the price field needs to go away in lieu of stock records.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -281,3 +281,52 @@ def revoke_line(self, line):
...
+        """
+        logger.info("Attempting to fulfill 'Coupon' product types for order [%s]", order.number)
+
+        for line in lines:
+            line.set_status(LINE.COMPLETE)
+
+        logger.info("Finished fulfilling 'Coupon' product types for order [%s]", order.number)
+        return order, lines
+
+    def revoke_line(self, line):  # pragma: no cover
```
clintonb: Add a test, making sure this always raises an error. If someone comes along and changes this method, we want to make sure the changes are tested.

ecommerce/extensions/api/serializers.py
```diff
@@ -49,6 +50,11 @@ def get_name(self, instance):
         return instance.attribute.name
 
     def get_value(self, obj):
+        if obj.attribute.name == 'Coupon vouchers':
```
clintonb, line 53: This needs to be tested.

ecommerce/extensions/api/serializers.py
```diff
@@ -332,3 +349,23 @@ def get_products(self, obj):
             kwargs={'parent_lookup_stockrecords__catalogs': obj.id},
             request=self.context['request']
         )
+
+
+class VoucherSerializer(serializers.ModelSerializer):
```
clintonb, line 354: This needs to be tested.

clintonb:
:+1: pending...
- [x] test updates
- [x] voucher comment
- [x] filtering
- [x] squashed commits

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -28,6 +32,28 @@ def setUp(self):
         expires = datetime.datetime(2100, 1, 1, tzinfo=pytz.UTC)
         self.seat = self.course.create_or_update_seat('honor', False, 0, self.partner, expires=expires)
 
+    def create_coupon(self):
+        """Helper method for creating a coupon."""
+        catalog = Catalog.objects.create(partner=self.partner)
```
clintonb: Create a factory for this.

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -28,6 +32,28 @@ def setUp(self):
...
+            'benefit_value': 100,
+            'catalog': catalog,
+            'end_date': datetime.date(2020, 1, 1),
+            'code': '',
+            'quantity': 5,
+            'start_date': datetime.date(2015, 1, 1),
+            'voucher_type': Voucher.SINGLE_USE
+        }
+
+        coupon = CouponOrderCreateView().create_coupon_product(
```
clintonb: Create a factory.

ecommerce/extensions/api/v2/views/products.py
```diff
@@ -13,4 +14,9 @@
 class ProductViewSet(NestedViewSetMixin, NonDestroyableModelViewSet):
     queryset = Product.objects.all()
     serializer_class = serializers.ProductSerializer
+    # URL with filter example:
```
clintonb: This belongs in a docstring on the class.

ecommerce/extensions/api/v2/views/products.py
```diff
@@ -13,4 +14,9 @@
...
+    # /api/v2/products/?product_class__name=Coupon
+    # Note, it's case-sensitive
```
clintonb: Use a FilterSet to make this case-insensitive:
- http://www.django-rest-framework.org/api-guide/filtering/#specifying-a-filterset
- https://django-filter.readthedocs.org/en/latest/usage.html

vkaracic:
@clintonb I believe I have addressed all your comments, and I squashed the commits.

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -120,3 +124,47 @@ def test_no_partner_product(self):
...
+        self.assertEqual(response.status_code, 200)
+
+        response_data = json.loads(response.content)
+        self.assertEqual(response_data['id'], 3)
+        self.assertEqual(response_data['title'], 'Test coupon')
+        self.assertEqual(response_data['price'], '100.00')
+        self.assertEqual(response_data['attribute_values'][0]['name'], 'Coupon vouchers')
+        self.assertEqual(len(response_data['attribute_values'][0]['value']), 5)
+
+    def test_coupon_voucher_serializer(self):
```
clintonb, line 142: Ideally serializers should be tested separately of the view. Besides giving us definite test coverage of the serializers, this testing also allows us to confidently use the serializer to generate expected output.

Here is an example in the edx/course-discovery:
- Serializer tests: https://github.com/edx/course-discovery/blob/clintonb/courses/course_discovery/apps/api/tests/test_serializers.py
- Tests utilizing serializers: https://github.com/edx/course-discovery/blob/clintonb/courses/course_discovery/apps/api/v1/tests/test_views.py#L97-L104

clintonb:
@vkaracic looks good. I'd prefer to see explicit serializer tests. Your choice.

vkaracic:
@clintonb I couldn't find any existing serializer tests so I though you guys test it that way. I could do it as a separate task, so feel free to create a task/story on JIRA and assign me. But for now we need to get this merged since there's a lot of work pending this merge.

PR #454, opened 2015-12-04
Updated New Relic

clintonb:
ECOM-3099


clintonb:
FYI @feanil updating even if we don't think it's the cause.

PR #455, opened 2015-12-07
Add honor mode as additional option to CAT

bderusha:
@peter-fogg @clintonb Please review

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -88,7 +88,7 @@ class Course(models.Model):
         elif 'verified' in seat_types:
             return 'verified'
         else:
-            return 'honor'
+            return 'audit'
 
     @property
     def parent_seat_product(self):
--- a/ecommerce/courses/tests/test_models.py
+++ b/ecommerce/courses/tests/test_models.py
@@ -179,10 +179,10 @@ class CourseTests(CourseCatalogTestMixin, TestCase):
     def test_type(self):
         """ Verify the property returns a type value corresponding to the available products. """
         course = Course.objects.create(id='a/b/c', name='Test Course')
-        self.assertEqual(course.type, 'honor')
+        self.assertEqual(course.type, 'audit')
 
-        course.create_or_update_seat('honor', False, 0, self.partner)
-        self.assertEqual(course.type, 'honor')
+        course.create_or_update_seat('audit', False, 0, self.partner)
+        self.assertEqual(course.type, 'audit')
 
         course.create_or_update_seat('verified', True, 10, self.partner)
         self.assertEqual(course.type, 'verified')
--- a/ecommerce/static/js/models/course_model.js
+++ b/ecommerce/static/js/models/course_model.js
@@ -45,7 +45,8 @@ define([
                 id: null,
                 name: null,
                 type: null,
-                verification_deadline: null
+                verification_deadline: null,
+                honor_mode: null
             },
 
             validation: {
@@ -59,6 +60,10 @@ define([
                 type: {
                     required: true,
                     msg: gettext('You must select a course type.')
+                },
+                honor_mode: {
+                    required: true,
+                    msg: gettext('You must choose if an honor seat should be created.')
                 },
                 verification_deadline: function (value) {
                     var invalid;
@@ -91,7 +96,8 @@ define([
                 id: gettext('Course ID'),
                 name: gettext('Course Name'),
                 type: gettext('Course Type'),
-                verification_deadline: gettext('Verification Deadline')
+                verification_deadline: gettext('Verification Deadline'),
+                honor_mode: gettext('Include Honor Seat')
             },
 
             relations: [{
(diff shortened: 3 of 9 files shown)


peter-fogg:
:+1: Code looks good to me, but the tests look like they've broken due to the removal of `honor_course_seat_form_field_view.js`.

bderusha:
I wrestled with this a bit last night. Might be easier to add the file back for now and do the clean up as a fast follow

peter-fogg:
I think the problem is here: https://github.com/edx/ecommerce/blob/66049ce9c1b6d946732434cfb5abe03acbc598e0/ecommerce/static/js/test/specs/views/course_edit_view_spec.js#L5

You're still requiring the deleting file. Something is still failing for me locally after removing that, but at least some of the tests are passing.

ecommerce/courses/tests/test_models.py
```diff
@@ -179,10 +179,10 @@ def test_collision_avoidance(self):
     def test_type(self):
         """ Verify the property returns a type value corresponding to the available products. """
         course = Course.objects.create(id='a/b/c', name='Test Course')
-        self.assertEqual(course.type, 'honor')
+        self.assertEqual(course.type, 'audit')
 
-        course.create_or_update_seat('honor', False, 0, self.partner)
```
clintonb, line 184: You still need this to test the return of the `honor` type.

ecommerce/static/js/models/course_model.js
```diff
@@ -251,12 +290,27 @@ define([
...
                         name: this.get('name'),
                         verification_deadline: null
                     };
 
+                if (this.includeHonorMode()) {
+                    honorMode = this.get('honor_mode');
+
+                    if (honorMode) {
+                        honorSeatClass = CourseUtils.getCourseSeatModel('honor');
+                        /*jshint newcap: false */
```
clintonb: Why the disabling here? Your first letter of your class name should be capitalized.

ecommerce/static/js/views/course_form_view.js
```diff
@@ -253,12 +263,23 @@ define([
             },
 
             /**
+             * Displays, or hides, the verification deadline based on the course type.
```
clintonb: This comment is not accurate.

clintonb:
Tests?

bderusha:
@clintonb @peter-fogg ready for re-review

ecommerce/courses/models.py
```diff
@@ -88,7 +88,7 @@ def type(self):
         elif 'verified' in seat_types:
             return 'verified'
         else:
-            return 'honor'
```
clintonb, line 91: What about honor courses?

ecommerce/static/js/models/course_model.js
```diff
@@ -220,12 +252,30 @@ define([
             },
 
             /**
-             * Returns an array of seat types relevant to this Course, based on its type.
+             * Returns boolean indicating if this Course should include an honor seat.
+             *
+             * @returns {boolean}
+             */
+            includeHonorMode: function () {
+                return this.get('type') !== 'professional';
```
clintonb: Account for null values (e.g. new courses): `this.get('type') && this.get('type') !== 'professional'`.

peter-fogg: Isn't that case handled already? `undefined` is falsy.

bderusha: if `this.get('type')` is Falsey for some reason it will still work.  `null !== 'string'` is the same result as `'other-string' !== 'string'`

ecommerce/courses/models.py
bderusha, line 91: Since we are removing `honor` as a standalone type from the front end I also removed it here.  We can always leave this so that these show as `honor` courses on the display page, but we don't have a corresponding type for the edit/create page.  If that's okay I can put this back.

ecommerce/static/js/test/specs/views/course_form_view_spec.js
```diff
@@ -0,0 +1,44 @@
...
+        describe('course form view', function () {
+            describe('cleanHonorCode', function () {
+                it('should always return a boolean', function () {
+                    expect(view.cleanHonorMode('false')).toEqual(false);
+                    expect(view.cleanHonorMode('true')).toEqual(true);
+                });
+            });
+
+            describe('getActiveCourseTypes', function () {
+                it('should return expected course types', function () {
```
peter-fogg, line 25: It might be nice to steal something like [`withData` from edx-platform](https://github.com/edx/edx-platform/blob/master/common/static/common/js/spec_helpers/spec_helpers.js) to make these tests a little less verbose, but we can think about that for the future.

ecommerce/static/js/models/course_model.js
bderusha: :100: fixed

ecommerce/courses/models.py
bderusha, line 91: I will ask @mikigoyal what he would like to do about this and make sure the implementation matches.

ecommerce/static/templates/course_form.html
```diff
@@ -57,6 +57,25 @@
     </div>
 </div>
 
+<div class="fields">
+    <div class="form-group honor-mode hidden">
+        <label class="hd-4" for="honorMode"><%= gettext('Include Honor Seat') %></label>
```
peter-fogg: This should use `<%- %>` to HTML-escape the text.

ecommerce/static/templates/course_form.html
```diff
@@ -57,6 +57,25 @@
...
+            <label class="radio-inline">
+                <input type="radio" name="honor_mode" value="true"> Yes
+            </label>
+            <label class="radio-inline">
+                <input type="radio" name="honor_mode" value="false"> No
+            </label>
+        <!-- NOTE: This help-block is here for validation messages. -->
+        <span class="help-block"></span>
+        <span id="honorModeHelpBlock" class="help-block">
+            <%= gettext('Include an Honor Seat with this course') %>
```
peter-fogg: Same here.

ecommerce/courses/models.py
bderusha, line 91: Miki says to leave it.  I've added a helper note on the display page to make sure the change is explicitly communicated.

bderusha:
@peter-fogg @clintonb changes implemented

ecommerce/static/js/test/specs/views/course_form_view_spec.js
bderusha, line 25: agreed

peter-fogg:
:+1:

ecommerce/static/js/models/course_model.js
```diff
@@ -60,6 +61,9 @@ define([
                     required: true,
                     msg: gettext('You must select a course type.')
                 },
+                honor_mode: {
+                    required: true
```
clintonb: Add a `msg` field similar to above. The default message, "Include Honor Seat is required," is not ideal.

"You must choose if an honor seat should be created."

bderusha: fixed

clintonb:
"Select a course type." should go away once a course type has been selected. :+1: once this is fixed.
create

bderusha:
@clintonb bug filed for the above issue since it is in production already.
https://openedx.atlassian.net/browse/ECOM-3150

PR #457, opened 2015-12-08
Added management command to convert honor seats to audit

clintonb:


bderusha:
Should we have a check in here to ensure that a seat we are going to delete does not have any enrollments first?

bderusha:
Other than that question the code looks good.  :+1:

clintonb:
@bderusha if this command were to be long-lived, I would say yes. However, this command will die as soon as we get the CAT updated, and we know these courses are not live. Given that, we can forego the enrollment check.

PR #459, opened 2015-12-09
Update AUTHORS

marjev:


marjev:
@clintonb

clintonb:
Looks like there is a conflict. :+1:

PR #462, opened 2015-12-09
Update deadline copy to reflect defaults

bderusha:
# [ECOM-3044](https://openedx.atlassian.net/browse/ECOM-3044)
### Reviewers:
- @clintonb 
- @peter-fogg

### DIFF
--- a/ecommerce/static/templates/verified_course_seat_form_field.html
+++ b/ecommerce/static/templates/verified_course_seat_form_field.html
@@ -33,6 +33,10 @@
         <span id="expiresHelpBlock" class="help-block">
             <%= gettext('After this date/time, students can no longer enroll in this track.') %>
         </span>
+
+        <span class="help-block">
+            <%- gettext('Change the upgrade deadline for this course. The default deadline is 10 days before the course end date and time.') %>
+        </span>
         </div>
     </div>
 </div>
--- a/ecommerce/static/templates/course_form.html
+++ b/ecommerce/static/templates/course_form.html
@@ -52,7 +52,7 @@
             <%= gettext('After this date/time, students can no longer submit photos for ID verification.') %>
         </span>
         <span class="help-block">
-            <em>(<%= gettext('This is field is optional for course seats that require ID verification.') %>)</em>
+            <em>(<%- gettext('Change the verification deadline for this course. The default deadline is the course end date and time.') %>)</em>
         </span>
     </div>
 </div>
--- a/ecommerce/static/templates/professional_course_seat_form_field.html
+++ b/ecommerce/static/templates/professional_course_seat_form_field.html
@@ -33,6 +33,9 @@
         <span id="expiresHelpBlock" class="help-block">
             <%= gettext('After this date/time, students can no longer enroll in this track.') %>
         </span>
+        <span class="help-block">
+            <%- gettext('Change the upgrade deadline for this course. The default deadline is 10 days before the course end date and time.') %>
+        </span>
         </div>
     </div>
 </div>


clintonb:
Is there a reason we are HTML-escaping every string, including static strings?

ecommerce/static/templates/verified_course_seat_form_field.html
```diff
@@ -31,12 +31,16 @@
             <!-- NOTE: This help-block is here for validation messages. -->
             <span class="help-block"></span>
         <span id="expiresHelpBlock" class="help-block">
-            <%= gettext('After this date/time, students can no longer enroll in this track.') %>
+            <%- gettext('After this date/time, students can no longer enroll in this track.') %>
+        </span>
+
+        <span class="help-block">
+            <%- gettext('Change the upgrade deadline for this course. The default deadline is 10 days before the course end date and time.') %>
```
clintonb, line 38: Where is the deadline actually set? I don't see a default anywhere.

bderusha, line 38: Default deadline is set on the LMS side.  This is used as an override.

bderusha:
The HTML escaping is just future proofing.  Agreed that it's probably overkill, but since I was escaping something on the page I thought I'd make the whole page consistent.

ecommerce/static/templates/verified_course_seat_form_field.html
clintonb, line 38: When did we start setting a default upgrade deadline? How is that possible given that the course modes are created by CAT?

bderusha, line 38: We havent yet.  The code that is in master does it.  LMS listens to the course publish signal and sets a default if none exists.  Once set by CAT it will honor the CAT setting

clintonb, line 38: This means LMS and Otto are out of sync if users do not set a date here. That will lead to issues. If we want a default value, it should be set in CAT and propagated from there, not LMS.

bderusha, line 38: This code is already in master and has been approved across many levels.  If this is really an issue let's talk about it ASAP.  But I don't think it will cause issues.  If you think of the field as an override only I don't see the issue.

FYI @cpennington @mikigoyal @peter-fogg @rock345

clintonb, line 38: Otto is the system of record for course modes, upgrade deadlines, and verification deadlines. Any defaults should originate there.

bderusha, line 38: I guess the question is it okay to change the SoR for those value?  That's definitely true today.  Does it need to be true forever?  If yes then we need to change some things.

clintonb, line 38: Yes, Otto needs to be the system of record. We cannot rely on LMS to tell us if a product can be purchased.

cpennington, line 38: I agree with Clinton's concern that Otto shouldn't have to check with the LMS to know if we're past a date cutoff. Since we're basing the default dates on the end of the course, we'll have to push that data over to Otto when the course is published, I guess. @bderusha, how doable is that, based on what you've already got on master?

bderusha, line 38: To patch it in as a callback to Otto shouldn't be too hard.  But I think the real work is that if Otto is the source of truth, should the whole thing be rewired such that the default settings live in Otto too. That would be more work.

/CC @cpennington @clintonb @peter-fogg  @mikigoyal @marcotuts @rock345

bderusha:
@clintonb there is another ticket to deal with the issue of syncing lms and otto deadline dates https://openedx.atlassian.net/browse/ECOM-3174

Could you and @peter-fogg give this another quick look.

peter-fogg:
@bderusha :+1: 

@clintonb We should always be HTML-escaping translated strings, since translators have no need to put markup in their translations.

clintonb:
@peter-fogg that's not always true. There are some cases where including HTML in strings marked for translation is unavoidable.

peter-fogg:
@clintonb But the HTMl is not added by the translators themselves, thus the escaping.

PR #463, opened 2015-12-11
Add indexes to all TimeStampedModels to match their default ordering

cpennington:
TimeStampedModels have a default ordering of ('-modified', '-created').
In the django admin, this causes a full table scan. Indexing these
columns together will speed up those views.

Reference:
  https://github.com/django-extensions/django-extensions/blob/master/django_extensions/db/models.py#L29

### DIFF
--- a/ecommerce/extensions/refund/migrations/0002_auto_20151211_1417.py
+++ b/ecommerce/extensions/refund/migrations/0002_auto_20151211_1417.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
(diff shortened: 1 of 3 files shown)


clintonb:
Why not disable table scans?

cpennington:
I would happily error out on table scans in ecommerce, so that we get bug reports rather than bringing down the site.

cpennington:
Is that something we can do, though?

clintonb:
I think I'm missing context on this issue that is in HipChat. Carry on without my review. I must focus on the patch.

cpennington:
Note to self: This needs the migration generated.

clintonb:
:+1: (once migrations generated)

PR #464, opened 2015-12-11
Fix credit checkout page for courses with an audit mode.

peter-fogg:
# ECOM-3166

Fixes the credit checkout page for courses which have an audit seat.


clintonb:
:+1:

peter-fogg:
@clintonb I had to make some formatting changes for Pylint, and also changed the test to actually fail with the old code.

clintonb:
Not seeing a forced failure. Regardless, the fix remains accurate. :+1:

PR #465, opened 2015-12-11
Make Refund.status and RefundLine.status into choice fields

cpennington:
@rlucioni: As promised, choice fields!


clintonb:
:+1:

rlucioni:
:+1:

PR #466, opened 2015-12-11
Make the credit checkout page respect payment processor switches.

peter-fogg:
@edx/ecommerce I noticed that we weren't respecting the payment processor switches on the credit checkout page.

### DIFF
--- a/ecommerce/credit/tests/test_views.py
+++ b/ecommerce/credit/tests/test_views.py
@@ -4,9 +4,11 @@ Tests for the checkout page.
 from __future__ import unicode_literals
 import json
 
+import ddt
 from django.conf import settings
 from django.core.urlresolvers import reverse
 import httpretty
+from waffle.models import Switch
 
 from ecommerce.core.tests import toggle_switch
 from ecommerce.courses.models import Course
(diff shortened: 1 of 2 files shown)


clintonb:
:+1:

ecommerce/credit/tests/test_views.py
```diff
@@ -235,3 +236,29 @@ def test_get_without_credit_seat(self):
 
         response = self.client.get(self.path)
         self.assertEqual(response.context['error'], 'No credit seat is available for this course.')
+
+    @httpretty.activate
+    def test_no_providers_available(self):
```
jimabramson: this doesn't seem to test the behavior when a subset of providers are disabled, only when they all are.  would it be worth testing the former case as well?

peter-fogg: That's a good point, I'll add a similar test for a subset.

bderusha:
:+1:  once tests pass

peter-fogg:
@jimabramson I've DDTed the test to look at more possible combinations of switch settings.

jimabramson:
@peter-fogg nice :+1:

PR #467, opened 2015-12-11
Converted order API views to viewset

clintonb:
- Using a viewset to clean code
- Allowing staff users to retrieve orders
- Deprecated OrderByBasketRetrieveView

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_orders.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_orders.py
@@ -1,11 +1,11 @@
 import json
 
 import ddt
+import httpretty
+import mock
 from django.contrib.auth.models import Permission
 from django.core.urlresolvers import reverse
 from django.test import override_settings
-import httpretty
-import mock
 from oscar.core.loading import get_model
 from oscar.test import factories
 
@@ -20,10 +20,11 @@ Order = get_model('order', 'Order')
 ShippingEventType = get_model('order', 'ShippingEventType')
 
 
+@ddt.ddt
 class OrderListViewTests(AccessTokenMixin, ThrottlingMixin, TestCase):
     def setUp(self):
         super(OrderListViewTests, self).setUp()
-        self.path = reverse('api:v2:orders:list')
+        self.path = reverse('api:v2:order-list')
         self.user = self.create_user()
         self.token = self.generate_jwt_token_header(self.user)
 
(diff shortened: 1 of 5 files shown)


clintonb:
@rlucioni @peter-fogg

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -85,7 +85,7 @@ def test_with_other_users_orders(self):
 
     def test_super_user(self):
         """ The view should return all orders for when authenticating as a superuser. """
-        superuser = self.create_user(is_superuser=True)
+        superuser = self.create_user(is_superuser=True, is_staff=True)
```
peter-fogg: Does it make sense to test when `is_superuser=False`and `is_staff=True`?

clintonb: Yes. Fixing.

ecommerce/extensions/api/v2/views/orders.py
```diff
@@ -15,79 +17,26 @@
...
-class OrderFulfillView(generics.UpdateAPIView):
-    permission_classes = (IsAuthenticated, DjangoModelPermissions,)
-    throttle_classes = (ServiceUserThrottle,)
-    lookup_field = 'number'
-    queryset = Order.objects.all()
-    serializer_class = serializers.OrderSerializer
+        return queryset
 
-    def update(self, request, *args, **kwargs):
+    @detail_route(methods=['put', 'patch'])
```
rlucioni, line 37: I think it would make sense to allow POSTing to this endpoint. According to [the W3C](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html), PUT requests are supposed to be idempotent. However, PUT requests made to this endpoint will not be idempotent.

This endpoint does result in updating an existing resource, but that update is performed as a side effect of fulfillment (i.e., the order status is not directly updated).

clintonb, line 37: The original implementation used PATCH and PUT. If we want to change that,
let's do so in a future PR. At present we only use PUT.

rlucioni, line 37: That's fair, I just realized we were originally supporting PUT.

rlucioni:
:+1: once tests pass. Looks like conflicting migrations are breaking your build.

peter-fogg:
:-1:

peter-fogg:
Ack! That happens every damn time. :+1: :ship: :trophy:, etc.

clintonb:
LOL

PR #471, opened 2015-12-16
Only validate honor mode for non-prof-ed courses.

peter-fogg:
Quick fix for [ECOM-3170](https://openedx.atlassian.net/browse/ECOM-3170).

@bderusha @rlucioni

### DIFF
--- a/ecommerce/static/js/models/course_model.js
+++ b/ecommerce/static/js/models/course_model.js
@@ -62,7 +62,9 @@ define([
                     msg: gettext('You must select a course type.')
                 },
                 honor_mode: {
-                    required: true,
+                    required: function () {
+                        return this.includeHonorMode();
+                    },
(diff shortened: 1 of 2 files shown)


ecommerce/static/js/models/course_model.js
```diff
@@ -62,7 +62,9 @@ define([
                     msg: gettext('You must select a course type.')
                 },
                 honor_mode: {
-                    required: true,
+                    required: function () {
+                        return this.get('type') !== 'professional';
```
bderusha: Can we use the `includeHonorMode` method here to reduce duplicate code?

rlucioni:
:+1:

ecommerce/static/js/models/course_model.js
peter-fogg: Good idea, that's done.

bderusha:
:+1:

PR #472, opened 2015-12-17
Revert "Add indexes to all TimeStampedModels to match their default ordering"

clintonb:
Reverts edx/ecommerce#463

We do not have capacity to load test these changes.


clintonb:
FYI @edx/ecommerce

bderusha:
:+1:

clintonb:
FYI @edx/ecommerce before pulling this change, make sure you reverse the refund migrations before applying this change:

``` bash
./manage.py migrate refund 0001
```

PR #473, opened 2015-12-20
Add robots.txt

rlucioni:
Prevents well-behaved robots from visiting any pages on the site.

@clintonb @e0d please review.


clintonb:
:+1:

PR #475, opened 2015-12-22
[WIP] Add category to coupon model

vkaracic:
https://openedx.atlassian.net/browse/SOL-1539

Coupons should have their categories, "support", "student aid" etc.

In this PR:
- new `CategoryViewSet`. A read-only view set that lists out all the categories and supports filtering with `depth` to distinguish categories from sub-categories and `path` to make retrieving sub-categories of a specific category easier
- new `CategorySerializer`. Currently the depth=1 categories are displayed on the first level and all the sub-categories are nested inside.

```
- category1
    child: [
        sub-category1
            child: []
        sub-category2
            child: []
    ]
- category2
    child: []
```
- new migration that inserts all the default categories for coupons
- update to `CouponsViewSet` to handle categories and sub-categories
- new tests and updates to existing ones to test the new functionalities

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -6,6 +6,7 @@ import dateutil.parser
 
 from django.db import transaction
 from django.db.utils import IntegrityError
+from oscar.apps.catalogue.categories import create_from_breadcrumbs
 from oscar.core.loading import get_model
 from rest_framework import status
 from rest_framework.permissions import IsAuthenticated, IsAdminUser
@@ -24,6 +25,8 @@ from ecommerce.extensions.voucher.utils import create_vouchers
 
 Basket = get_model('basket', 'Basket')
 Catalog = get_model('catalogue', 'Catalog')
+Category = get_model('catalogue', 'Category')
+ProductCategory = get_model('catalogue', 'ProductCategory')
 logger = logging.getLogger(__name__)
 Order = get_model('order', 'Order')
 Product = get_model('catalogue', 'Product')
@@ -73,6 +76,7 @@ class CouponViewSet(EdxOrderPlacementMixin, NonDestroyableModelViewSet):
             quantity = request.data[AC.KEYS.QUANTITY]
             price = request.data[AC.KEYS.PRICE]
             partner = request.site.siteconfiguration.partner
+            category = request.data[AC.KEYS.CATEGORY] or 'Other'
 
             client, __ = Client.objects.get_or_create(username=client_username)
 
@@ -94,7 +98,8 @@ class CouponViewSet(EdxOrderPlacementMixin, NonDestroyableModelViewSet):
                 'code': code,
                 'quantity': quantity,
                 'start_date': start_date,
-                'voucher_type': voucher_type
+                'voucher_type': voucher_type,
+                'category': category
             }
 
             coupon_product = self.create_coupon_product(title, price, data)
@@ -127,6 +132,7 @@ class CouponViewSet(EdxOrderPlacementMixin, NonDestroyableModelViewSet):
                 - quantity (int)
                 - start_date (Datetime)
                 - voucher_type (str)
+                - category (str)
 
         Returns:
             A coupon product object.
@@ -168,6 +174,9 @@ class CouponViewSet(EdxOrderPlacementMixin, NonDestroyableModelViewSet):
         coupon_vouchers = CouponVouchers.objects.get(coupon=coupon_product)
 
         coupon_product.attr.coupon_vouchers = coupon_vouchers
+        category = create_from_breadcrumbs(data['category'])
+        ProductCategory.objects.get_or_create(category=category, product=coupon_product)
+
         coupon_product.save()
 
         sku = generate_sku(
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -76,7 +76,8 @@ class CouponViewSet(EdxOrderPlacementMixin, NonDestroyableModelViewSet):
             quantity = request.data[AC.KEYS.QUANTITY]
             price = request.data[AC.KEYS.PRICE]
             partner = request.site.siteconfiguration.partner
-            category = request.data[AC.KEYS.CATEGORY] or 'Other'
+            category = request.data[AC.KEYS.CATEGORY]
+            sub_category = request.data['sub_category']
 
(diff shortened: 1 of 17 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=475

vkaracic:
@marjev @dsego @mjfrey please take a look!

ecommerce/extensions/catalogue/tests/test_coupons.py
```diff
@@ -22,6 +22,8 @@ def test_coupon_product(self):
         voucherList = CouponVouchers.objects.create(coupon=coupon_product)
         voucherList.vouchers.add(voucher)
         coupon_product.attr.coupon_voucher = voucherList
+        coupon_product.attr.coupon_category = 'coupon-category'
+        coupon_product.save()
```
marjev: I think you don't need this to call `save` here.

marjev:
lgtm, just one small comment :+1:

dsego:
lgtm :+1:

vkaracic:
@clintonb can you please review this?

clintonb:
Oscar already has a notion of product categories. Why was this implementation chosen over that one?

Info
- http://django-oscar.readthedocs.org/en/releases-1.1/howto/how_to_create_categories.html
- https://github.com/edx/ecommerce/blob/87b9d1667d6971d11cc6ed08923c52438c46d10b/ecommerce/extensions/catalogue/migrations/0013_coupon_product_class.py#L29-L37
- https://github.com/edx/ecommerce/blob/5878a932c1a441e9e871cd417f36a432f17f5a90/ecommerce/courses/models.py#L43

vkaracic:
Because of this: http://django-oscar.readthedocs.org/en/releases-1.1/ref/apps/catalogue.html#oscar.apps.catalogue.abstract_models.AbstractCategory

but I guess this category is only going to be used for navigational purposes as well, so I'll change that.

vkaracic:
Code updated, please take another look.
@mjfrey @dsego @clintonb

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -69,6 +72,7 @@ def create(self, request, *args, **kwargs):
             quantity = request.data[AC.KEYS.QUANTITY]
             price = request.data[AC.KEYS.PRICE]
             partner = request.site.siteconfiguration.partner
+            category = request.data[AC.KEYS.CATEGORY] or 'Other'
```
clintonb: This needs to be limited to known group of categories. We don't want users making a bunch of random categories.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -164,6 +170,9 @@ def create_coupon_product(self, title, price, data):
         coupon_vouchers = CouponVouchers.objects.get(coupon=coupon_product)
 
         coupon_product.attr.coupon_vouchers = coupon_vouchers
+        category = create_from_breadcrumbs(data['category'])
```
clintonb: Make sure the specified category is nested below the "All Coupons" category.

ecommerce/extensions/api/v2/views/coupons.py
vkaracic: Can we limit it on the front end side? In the form of a dropdown of values?

On Tue, Jan 5, 2016 at 2:54 PM, Clinton Blackburn notifications@github.com
wrote:

> In ecommerce/extensions/api/v2/views/coupons.py
> https://github.com/edx/ecommerce/pull/475#discussion_r48846222:
> 
> > @@ -69,6 +72,7 @@ def create(self, request, _args, *_kwargs):
> >              quantity = request.data[AC.KEYS.QUANTITY]
> >              price = request.data[AC.KEYS.PRICE]
> >              partner = request.site.siteconfiguration.partner
> > -            category = request.data[AC.KEYS.CATEGORY] or 'Other'
> 
> This needs to be limited to known group of categories. We don't want users
> making a bunch of random categories.
> 
> —
> https://github.com/edx/ecommerce/pull/475/files#r48846222.

## 

Vedran Karačić

Mail:     vedran@edx.org
             vkaracic@extensionengine.com
Github: vkaracic https://github.com/vkaracic
Skype:  karacicvedran

clintonb: The limits need to be at both levels.

mjfrey: What list should we use?  A pre-defined set?

clintonb: The list needs to come from whomever proposed this change. If there is no known list of potential values, I recommend holding off on this addition until we have a better understanding of how the field will/should be used.

mjfrey: So we will create a list the user can choose from and add the ability to create new categories later.

mjfrey:
@clintonb 
Can you take a look.  For reference, this will just enable categories for coupons.  The UI will disable the field for now and we will just send CustomerSupport as the category. 
A future task is to enable full custom categories.

clintonb:
I see no use of "Customer Support" here.

mjfrey:
The UI will send that as default

clintonb:
What UI? The word "support" is nowhere to be found in this batch of commits.

mjfrey:
We have a PR for the coupon creation UI that will send "CustomerSupport" as the category and make the field read-only.

mjfrey:
@clintonb would it be better to just hard code "CustomerSupport" here?  Then use what is sent after we implement the logic to support custom categories.

clintonb:
1. I'd use "Customer Support" (two words).
2. I wouldn't add any category beyond "Catalogs". As soon as this is released, someone other than Student Support will use it, and your data will be wrong. I'd rather the data have no category than the wrong category.

Once again I question why we are bothering with minor UI changes when the major portions of this functionality have not been reviewed. This is a [bikeshedding](https://en.wikipedia.org/wiki/Parkinson%27s_law_of_triviality) moment.

mjfrey:
1. Product wanted "CustomerSupport" one word.
2. I agree with you.  I am just trying to fulfill Product's requirement of tagging these first coupons with the Support category.  But lets put this on hold and land the remaining bigger items and get back to this.

vkaracic:
@mjfrey @marjev please take a look.

vkaracic:
Closing this in favor of: https://github.com/edx/ecommerce/pull/552

PR #476, opened 2015-12-22
Coupon GET method support

vkaracic:
https://openedx.atlassian.net/browse/SOL-1545

In order to display necessary details about a coupon a new endpoint has to be created that combines some information from the coupon and order/basket.

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -3,6 +3,7 @@ from decimal import Decimal
 import logging
 
 from dateutil.parser import parse
+from django.conf import settings
 from django.db import transaction
 from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_model, get_class
@@ -16,6 +17,8 @@ from ecommerce.courses.models import Course
 
 logger = logging.getLogger(__name__)
 
+Basket = get_model('basket', 'Basket')
+Benefit = get_model('offer', 'Benefit')
 BillingAddress = get_model('order', 'BillingAddress')
 Catalog = get_model('catalogue', 'Catalog')
 Line = get_model('order', 'Line')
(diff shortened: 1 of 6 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=476

ecommerce/extensions/api/serializers.py
```diff
@@ -362,10 +370,57 @@ def get_is_available_to_user(self, obj):
...
+
+    def get_seats(self, obj):
+        voucher = obj.attr.coupon_vouchers.vouchers.first()
+        stockrecords = voucher.offers.first().condition.range.catalog.stock_records.all()
+        seats = Product.objects.filter(id__in=[sr.product.id for sr in stockrecords])
+        serializer = ProductSerializer(seats, many=True, context={'request': self.context['request']})
+        return serializer.data
+
+    def get_client(self, obj):
+        return Basket.objects.filter(lines__product_id=obj.id).first().owner.username
```
marjev: This will be a problem later on, in case the coupon is associated with multiple baskets (when the only thing that differs is the client username). In that case, only a username of the first 'client' will be returned, even thought multiple clients ordered the coupon.

FYI @mjfrey

mjfrey:
LGTM

@clintonb 
This is to add a Coupon endpoint that allows us to get the details needed for a given Coupon.  The info is a combination of both order and product.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -29,12 +31,14 @@
 StockRecord = get_model('partner', 'StockRecord')
 
 
-class CouponOrderCreateView(EdxOrderPlacementMixin, generics.CreateAPIView):
+class CouponOrderCreateViewSet(EdxOrderPlacementMixin, NonDestroyableModelViewSet):
```
clintonb: CouponViewSet. Also, update the docstring.

clintonb:
:+1: once names corrected for views and tests.

P.S. @mjfrey, if you don't already, please ensure there are stories to add documentation for all this coupons work. It's no good if only three developers know how it works.

PR #477, opened 2015-12-22
Disable course configuration in CAT after enrollment start.

peter-fogg:
# [ECOM-3254](https://openedx.atlassian.net/browse/ECOM-3254)

In order to avoid inconsistencies between LMS and CAT, this queries the enrollment API before publishing course modes and saving the course to Otto.

Note that this will require the changes in https://github.com/edx/edx-platform/pull/11053 to be deployed, since our Slumber-based API client adds a trailing slash to URLs.

@rlucioni @bderusha, can you review?

### DIFF
--- a/ecommerce/courses/publishers.py
+++ b/ecommerce/courses/publishers.py
@@ -1,11 +1,14 @@
 from __future__ import unicode_literals
+from datetime import datetime
 import json
 import logging
 
+import dateutil
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 from edx_rest_api_client.client import EdxRestApiClient
 from edx_rest_api_client.exceptions import SlumberHttpBaseException
+import pytz
 import requests
 
 from ecommerce.courses.utils import mode_for_seat
@@ -53,7 +56,7 @@ class LMSPublisher(object):
 
         api.courses(course_id).put(data)
 
-    def publish(self, course, access_token=None):
+    def publish(self, course, access_token=None, check_enrollment_start=False):
         """ Publish course commerce data to LMS.
 
         Uses the Commerce API to publish course modes, prices, and SKUs to LMS. Uses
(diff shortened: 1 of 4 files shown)


ecommerce/courses/publishers.py
```diff
@@ -78,6 +81,25 @@ def publish(self, course, access_token=None):
             logger.error('COMMERCE_API_URL is not set. Commerce data will not be published!')
             return error_message
 
+        try:
```
bderusha: This logic seams a little too overreaching to me.

This will block users from publishing **any** changes after the course has started.  It is a common use case (today) to push back the verification deadline on a course.  I'd imagine we would also want to allow PMs to "upgrade" modes as well.  EG audit to verified/audit. 

@marcotuts @mikigoyal @rock345 can you help clarify?

bderusha:
@peter-fogg I added a comment in the ticket.  I don't know if this is really what we want to be doing here.  If product says it is then :+1: .  Otherwise I'd image the logic around whether or not we can publish gets more complex and I think I'd prefer to see it pulled out into a helper method or something.

peter-fogg:
@rlucioni Could you also take a look at this, including the conversation in the JIRA comments?

ecommerce/courses/publishers.py
```diff
@@ -2,10 +2,13 @@
 import json
 import logging
 
+from datetime import datetime
```
rlucioni: Minor: datetime is part of the standard library and should be in the block of imports above. Pylint now checks for this; we should look into upgrading ecommerce.

rlucioni:
@peter-fogg sorry for the delay. I commented on the JIRA issue.

It's not clear to me what problem we're trying to solve with this change. What inconsistencies between the LMS and CAT have we observed? The CAT is the "source of truth" when it comes to course modes, so we shouldn't see differences between Otto's products and the course modes table. Being able to change upgrade and verification deadlines using the CAT seems important. Are we trying to prevent people from changing the seat types associated with a course once the course starts?

ecommerce/courses/publishers.py
peter-fogg: Fixed.

peter-fogg:
@rlucioni I've responded in JIRA.

peter-fogg:
Closing this, see the JIRA conversation.

PR #481, opened 2015-12-31
Disable fields for milestone 1

marjev:
@clintonb please review this.

JIRA ticket: https://openedx.atlassian.net/browse/SOL-1565

For milestone 1, product team would like some fields to be disabled and default values to be sent upon form submission.


clintonb:
:+1:

PR #482, opened 2015-12-31
Suggested UI text edits (SOL-1356 comment section)

dsego:
### Coupon Codes page

screen shot 2015-12-31 at 2 56 24 pm
### Create New Coupon page

screen shot 2015-12-31 at 2 56 48 pm
### Error messages

screen shot 2015-12-31 at 2 56 40 pm

screen shot 2015-12-31 at 2 57 08 pm

### DIFF
--- a/ecommerce/static/js/models/coupon_model.js
+++ b/ecommerce/static/js/models/coupon_model.js
@@ -18,6 +18,11 @@ define([
             ) {
         'use strict';
 
+        _.extend(Backbone.Validation.messages, {
+            required: gettext('This field is required'),
+            number: gettext('This value must be a number'),
+            date: gettext('This value must be a date')
+        });
         _.extend(Backbone.Model.prototype, Backbone.Validation.mixin);
 
         return Backbone.Model.extend({
@@ -31,27 +36,20 @@ define([
 
             validation: {
                 course_id: {
-                    pattern: 'courseId'
+                    pattern: 'courseId',
+                    msg: gettext('A valid course ID is required')
                 },
-                title: {
-                    required: true
-                },
-                client_username: {
-                    required: true
-                },
+                title: { required: true },
+                client_username: { required: true },
                 // seat_type is for validation only, stock_record_ids holds the values
-                seat_type: {
-                    required: true
-                },
                 price: {
                     pattern: 'number',
                     required: function () {
                         return this.isEnrollmentCode();
                     }
                 },
-                quantity: {
-                    pattern: 'digits'
-                },
+                seat_type: { required: true },
+                quantity: { pattern: 'number' },
                 benefit_value: {
                     pattern: 'number',
                     required: function () {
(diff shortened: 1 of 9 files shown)


ecommerce/static/js/models/coupon_model.js
```diff
@@ -18,6 +18,12 @@ define([
             ) {
         'use strict';
 
+        _.extend(Backbone.Validation.messages, {
+            required: gettext('This field is required'),
+            number: gettext('This value must be a number'),
+            digits: gettext('This value must only contain digits'),
```
clintonb: This doesn't seem necessary, given that you have the number message.

ecommerce/static/js/views/alert_view.js
```diff
@@ -17,7 +15,11 @@ define([
             },
 
             render: function () {
-                var body = this.template({title: this.title, message: this.message});
```
clintonb, line 20: Why has this been changed?

ecommerce/static/js/views/coupon_list_view.js
```diff
@@ -69,12 +69,12 @@ define([
...
                             },
                             {
-                                title: gettext('Voucher Report'),
+                                title: gettext('Coupon Report'),
                                 data: 'id',
                                 fnCreatedCell: function (nTd, sData, oData) {
                                     $(nTd).html(_s.sprintf('<a href="" ' +
                                         'class="btn btn-secondary btn-small voucher-report-button" ' +
-                                        'data-coupon-id="%s"> Download Voucher Report</a>', oData.id));
+                                        'data-coupon-id="%s"> Download Coupon Report</a>', oData.id));
```
clintonb: Mark for translation, and use `_.template()`.

ecommerce/static/js/views/form_view.js
```diff
@@ -49,10 +49,11 @@ define([
              * Renders alerts that will appear at the top of the page.
              *
              * @param {String} level - Severity of the alert. This should be one of success, info, warning, or danger.
-             * @param {Sring} message - Message to display to the user.
+             * @param {String} title - Title is displayed in bold.
+             * @param {String} message - Message to display to the user.
              */
-            renderAlert: function (level, message) {
-                var view = new AlertView({level: level, title: gettext('Error!'), message: message});
```
clintonb, line 55: Again, why the change for the _entire_ application, not just coupons? These are, in fact, errors on the user's part.

clintonb:
I'm okay with changing field validation messages, but not the error alerts.

dsego:
@clintonb  I need to change the alert message
“Error! Please complete all required fields.” to “You must complete all required fields.” 
How do you suggest I do it only for coupons?

clintonb:
You can change the message, but why remove the "Error!" prefix? Let's just
change the message and move on. The prefix is quite minor and doesn't need
to be touched.

dsego:
I agree. Actually all of this seems pretty minor, especially since it is an admin interface.

mjfrey:
@dsego Leave the Error prefix and lets move on.

mjfrey:
:+1:

PR #483, opened 2015-12-31
Fixes SOL-1554 and SOL-1555

dsego:
Issues in adding a course ID (SOL-1554) 
Total Paid field is empty (SOL-1555)

Screen recording: https://vid.me/e/ga8T

### DIFF
--- a/ecommerce/static/js/models/coupon_model.js
+++ b/ecommerce/static/js/models/coupon_model.js
@@ -31,7 +31,8 @@ define([
             defaults: {
                 quantity: 1,
                 stock_record_ids: [],
-                code: ''
+                code: '',
+                price: '0'
             },
 
             validation: {
@@ -42,18 +43,13 @@ define([
                 title: { required: true },
                 client_username: { required: true },
                 // seat_type is for validation only, stock_record_ids holds the values
-                price: {
-                    pattern: 'number',
-                    required: function () {
-                        return this.isEnrollmentCode();
-                    }
-                },
                 seat_type: { required: true },
                 quantity: { pattern: 'number' },
+                price: { pattern: 'number' },
                 benefit_value: {
                     pattern: 'number',
                     required: function () {
-                        return this.isDiscountCode();
+                        return this.get('code_type') === 'discount';
                     }
                 },
                 start_date: function (val) {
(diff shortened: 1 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @dsego! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=483

dsego:
@vkaracic @marjev Please take a look.

ecommerce/static/templates/coupon_form.html
```diff
@@ -54,7 +54,7 @@
             <label for="benefit_value"><%= gettext('Discount Value') %></label>
             <div class="input-group">
                 <div class="benefit-addon input-group-addon"></div>
-                <input type="text" class="form-control" name="benefit_value">
+                <input type="number" step="any" class="form-control" name="benefit_value">
```
marjev: why set step to `any`?

ecommerce/static/templates/coupon_form.html
```diff
@@ -81,7 +81,7 @@
             <label for="price"><%= gettext('Total Paid') %></label>
             <div class="input-group">
                 <div class="input-group-addon">$</div>
-                <input type="text" class="form-control" name="price">
+                <input type="number" step="any" class="form-control" name="price">
```
marjev: same question

marjev:
other than two small questions, :+1:

ecommerce/static/js/models/coupon_model.js
```diff
@@ -42,18 +43,12 @@ define([
                 title: { required: true },
                 client_username: { required: true },
                 // seat_type is for validation only, stock_record_ids holds the values
-                price: {
```
vkaracic, line 45: Why did you remove this validation?

dsego, line 45: Per sol-1555 the field can be left blank. Since you need the field to be non-empty for the backend, I will bring back the validation and add a default value.

mjfrey:
Looks good.

:+1:  Rebase and squash

PR #488, opened 2016-01-05
Prevent Celery from removing handlers on the root logger

rlucioni:
Allows use of custom logging handlers. ECOM-2249.

@jimabramson please review.


jimabramson:
:+1:

PR #489, opened 2016-01-06
Fix for ECOM-3321 flaky test

mjfrey:
Attempt to fix flaky test for ECOM-3321

### DIFF
--- a/ecommerce/static/js/test/specs/collections/coupon_collection_spec.js
+++ b/ecommerce/static/js/test/specs/collections/coupon_collection_spec.js
@@ -62,7 +62,8 @@ define([
 
             });
 
-            describe('fetch', function () {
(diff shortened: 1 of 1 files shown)


mjfrey:
@rlucioni This is to skip the test for ECOM-3321.

rlucioni:
FYI @peter-fogg

rlucioni:
:+1: after commits are squashed.

clintonb:
If this test is not going to be run, we should just remove it. There's is no point to keeping dead code around.

peter-fogg:
@clintonb I'd prefer keeping the test with `xit` for now, and prioritizing an actual fix.

mjfrey:
We are going to be changing this a bit so the test will be used in the future.

clintonb:
Okay.

PR #492, opened 2016-01-06
Do not use backbone-super

mjfrey:
Remove the use of backbone-super to see if it fixes the test in travis.

### DIFF
--- a/ecommerce/static/js/collections/coupon_collection.js
+++ b/ecommerce/static/js/collections/coupon_collection.js
@@ -1,12 +1,10 @@
 define([
         'backbone',
-        'backbone.super',
         'underscore',
         'collections/paginated_collection',
         'models/coupon_model'
     ],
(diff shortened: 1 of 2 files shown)


mjfrey:
@rlucioni @clintonb So this actually fixes the test.

ecommerce/static/js/collections/coupon_collection.js
```diff
@@ -20,7 +20,8 @@ define([
                 _.defaults(options.data || (options.data = {}), {
                     product_class__name: 'Coupon'
                 });
-                return this._super(options);
+                var tmp = PaginatedCollection.prototype.fetch.call(this, options);
+                return tmp;
```
rlucioni: Can you clean this up by removing `tmp`?

clintonb:
Remove the loading of `backbone.super`, then :+1:

PR #493, opened 2016-01-07
Disable atomicity for payment views, for real

rlucioni:
You can't apply a function decorator to a method. It must first be transformed into a method decorator. ECOM-2249.

@clintonb, please review.


clintonb:
Can we use a mixin to avoid duplication?

clintonb:
:+1:

rlucioni:
I gave it a shot, but it doesn't look like the method decorator gets inherited by subclasses using the mixin. Since use of a mixin won't save us an awful lot here and using the decorator is pretty clear to begin with, I'm going to press on without a mixin.

PR #495, opened 2016-01-07
Modify ecommerce worker service user privileges

rlucioni:
The service user should have permissions to fulfill orders, but should not be a superuser. ECOM-2249.

@clintonb pursuant to our conversation earlier today. Please review if you have a moment.


clintonb:
:+1:

PR #496, opened 2016-01-07
Only create honor or audit seats, not both.

bderusha:
Audit is the default seat, however if a PM adds the honor seat back in we need to remove the audit seat.

This is a short term fix so that we are not creating both seats.  There will be a follow on fix that includes visual changes to help PMs configure courses in both the new (audit) and old (honor) way. 

@peter-fogg @rlucioni please review

### DIFF
--- a/ecommerce/static/js/models/course_model.js
+++ b/ecommerce/static/js/models/course_model.js
@@ -300,6 +300,8 @@ define([
                     honorMode,
                     honorSeatClass,
                     honorSeat,
+                    products,
+                    auditSeat,
                     data = {
                         id: this.get('id'),
                         name: this.get('name'),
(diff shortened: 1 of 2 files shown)


ecommerce/static/js/models/course_model.js
```diff
@@ -314,6 +315,15 @@ define([
                         /*jshint newcap: false */
                         honorSeat = new honorSeatClass({course: this});
                         /*jshint newcap: true */
+
+
+                        // remove the audit seat and replace it with honor
+                        seats_no_audit = this.get('products').filter(function (product) {
+                            // Filter out audit seat
+                            return !!product.attributes.certificate_type
```
rlucioni: It'd be clearer to use Underscore's [isNull](http://underscorejs.org/#isNull).

rlucioni:
:+1: once tests pass.

peter-fogg:
LGTM but there's a failing test -- I think the test data in `course_model_spec.js` needs to be updated.

PR #500, opened 2016-01-13
New logic to generate coupon code

mjfrey:
Increase code length to 12
Base on UUID and hash


mjfrey:
@cpennington Please take a look.

ecommerce/extensions/voucher/utils.py
```diff
@@ -107,14 +110,11 @@ def _generate_code_string(length):
...
-        char for char in string.ascii_uppercase + string.digits
-        if char not in 'AEIOU1'
-    ]
-    voucher_code = string.join((random.choice(chars) for i in range(length)), '')
-    if Voucher.objects.filter(code__iexact=voucher_code).exists():
-        return _generate_code_string(length)
-    return voucher_code
+    tmp = uuid.uuid4().get_bytes()
+    h = hashlib.sha256()
+    h.update(tmp)
```
cpennington: Nit: you can pass `tmp` directly to the `sha256()` call.

ecommerce/settings/_oscar.py
```diff
@@ -251,6 +251,6 @@
 ENROLLMENT_FULFILLMENT_TIMEOUT = 7
 
 # Coupon code length
-VOUCHER_CODE_LENGTH = 8
+VOUCHER_CODE_LENGTH = 12
```
cpennington: Can we make this 14 or 16?

mjfrey: @cpennington  I went with 12 since we can nicely present this XXXX-XXXX-XXXX.  But if we feel strongly about increasing the length I am ok with it.

cpennington: It's probably already a large enough space that it doesn't matter, but I'd feel better taking amazon's 14 (or going larger). 16 makes a nice `XXXX-XXXX-XXXX-XXXX` pattern.

mjfrey:
@cpennington ok, length is pushed to 16.

cpennington:
:+1:

ecommerce/extensions/voucher/utils.py
```diff
@@ -107,13 +110,12 @@ def _generate_code_string(length):
...
         raise ValueError("Voucher code length must be a positive number.")
 
-    chars = [
-        char for char in string.ascii_uppercase + string.digits
-        if char not in 'AEIOU1'
-    ]
-    voucher_code = string.join((random.choice(chars) for i in range(length)), '')
+    h = hashlib.sha256()
+    h.update(uuid.uuid4().get_bytes())
+    voucher_code = str(base64.b32encode(h.digest()))[0:length]
```
vkaracic: I don't think you need the `str` conversion:

```
>>> str(base64.b32encode(h.digest()))[0:16] == base64.b32encode(h.digest())[0:16]
True
```

vkaracic:
What is the benefit of hashing over doing something simple as:

```
>>> uuid.uuid4().hex.upper()[0:16]
'1DC355963FE04C9C'
>>> uuid.uuid4().hex.upper()[0:16]
'A0DF648A052B4A97'
>>> uuid.uuid4().hex.upper()[0:16]
'9AD69809B625470F'
>>> uuid.uuid4().hex.upper()[0:16]
'6286E40FE206411E'
```

?

cpennington:
@vkaracic: base32 encoding gives you a much larger key space, and thus better protection against brute-force attacks. 32^16 >> 16^16.

vkaracic:
@cpennington okay, thanks for explaining!

PR #501, opened 2016-01-14
Use a Waffle Sample to control async order fulfillment

rlucioni:
Allows more controlled, careful release and use of this feature.

@clintonb, as we discussed. If you have a moment to review, I'd appreciate it. If not, let me know and I'll ping someone else.


clintonb:
:+1:

PR #505, opened 2016-01-15
ECOM-3464: Added the command to publish the courses to LMS

aamir-khan:
Added the command to take the course ids as input in text files and publish these course to LMS

Related Ticket: [ECOM-3464](https://openedx.atlassian.net/browse/ECOM-3464)

@adampalay Can you please review this PR.

### DIFF
--- a/ecommerce/courses/management/commands/publish_to_lms.py
+++ b/ecommerce/courses/management/commands/publish_to_lms.py
@@ -1,0 +1,63 @@
+""" This command publish the courses to LMS."""
+from __future__ import unicode_literals
+import logging
+from optparse import make_option
+import os
+
+from django.core.management import BaseCommand, CommandError
+
+from ecommerce.courses.models import Course
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """Publish the courses to LMS."""
+
+    help = 'Publish the courses to LMS'
+    option_list = BaseCommand.option_list + (
+        make_option(
+            '--course_ids_file',
+            action='store',
+            dest='course_ids_file',
+            default=None,
+            help='Path to file to read courses from.'
+        ),
+    )
+
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.DEBUG)
+    logger.addHandler(ch)
+
+    def handle(self, *args, **options):
+        failed = 0
+        course_ids_file = options['course_ids_file']
+        if not course_ids_file or not os.path.exists(course_ids_file):
+            raise CommandError("Pass the correct absolute path to course ids file as --course_ids_file argument.")
+
+        with open(course_ids_file, 'r') as file_handler:
+            course_ids = file_handler.readlines()
+            total_courses = len(course_ids)
+            logger.info("Publishing %d courses.", total_courses)
+            for index, course_id in enumerate(course_ids, start=1):
+                try:
+                    course_id = course_id.strip()
+                    course = Course.objects.get(id=course_id)
+                    publishing_error = course.publish_to_lms()
+                    if publishing_error:
+                        failed += 1
+                        logger.error(
+                            u"(%d/%d) Failed to publish %s: %s", index, total_courses, course_id, publishing_error
+                        )
+                    else:
+                        logger.info(u"(%d/%d) Successfully published %s.", index, total_courses, course_id)
+                except Course.DoesNotExist:
+                    failed += 1
+                    logger.error(
+                        u"(%d/%d) Failed to publish %s: Course does not exist.", index, total_courses, course_id
+                    )
+        if failed:
+            logger.error("Completed publishing courses. %d of %d failed.", failed, total_courses)
+        else:
+            logger.info("All %d courses successfully published.", total_courses)
--- a/ecommerce/courses/tests/test_publish_to_lms_command.py
+++ b/ecommerce/courses/tests/test_publish_to_lms_command.py
@@ -1,0 +1,170 @@
+# encoding: utf-8
+"""Contains the tests for publish to lms command."""
+
+from __future__ import unicode_literals
+import logging
+import os
+import tempfile
+
+import ddt
+from django.core.management import call_command, CommandError
+import mock
+from mock import call
+from testfixtures import LogCapture
+
+from ecommerce.courses.models import Course
+from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.tests.testcases import TransactionTestCase
+
+
+logger = logging.getLogger(__name__)
+LOGGER_NAME = 'ecommerce.courses.management.commands.publish_to_lms'
+
+
+@ddt.ddt
+class PublishCoursesToLMSTests(CourseCatalogTestMixin, TransactionTestCase):
+    """Tests the course publish command."""
+
+    tmp_file_path = os.path.join(tempfile.gettempdir(), "tmp-testfile.txt")
+
+    def setUp(self):
+        super(PublishCoursesToLMSTests, self).setUp()
+        self.course = CourseFactory()
+        self.create_course_ids_file(self.tmp_file_path, [self.course.id])
+
+    @classmethod
+    def tearDownClass(cls):
+        if os.path.exists(cls.tmp_file_path):
+            os.remove(cls.tmp_file_path)
+
+    def create_course_ids_file(self, file_path, course_ids):
+        """Write the course_ids list to the temp file."""
+
+        with open(file_path, 'w') as temp_file:
+            temp_file.write("\n".join(course_ids))
+
+    @ddt.data("", "fake/path")
+    def test_invalid_file_path(self, course_ids_file):
+        """ Verify command raises the CommandError for invalid file path. """
+
+        with self.assertRaises(CommandError):
+            call_command('publish_to_lms', course_ids_file=course_ids_file)
+
+    def test_invalid_course_id(self):
+        """ Verify invalid course_id fails. """
+
+        fake_course_id = "fake_course_id"
+        self.create_course_ids_file(self.tmp_file_path, [fake_course_id])
+        expected = (
+            (
+                LOGGER_NAME,
+                "INFO",
+                "Publishing 1 courses."
+            ),
+            (
+                LOGGER_NAME,
+                "ERROR",
+                u"(1/1) Failed to publish {}: Course does not exist.".format(fake_course_id)
+            ),
+            (
+                LOGGER_NAME,
+                "ERROR",
+                "Completed publishing courses. 1 of 1 failed."
+            )
+        )
+        with LogCapture(LOGGER_NAME) as lc:
+            call_command('publish_to_lms', course_ids_file=self.tmp_file_path)
+            lc.check(*expected)
+
+    def test_course_publish_successfully(self):
+        """ Verify all courses are successfully published."""
+
+        second_course = CourseFactory.create()
+        self.create_course_ids_file(self.tmp_file_path, [self.course.id, second_course.id])
+        expected = (
+            (
+                LOGGER_NAME,
+                "INFO",
+                "Publishing 2 courses."
+            ),
+            (
+                LOGGER_NAME,
+                "INFO",
+                u"(1/2) Successfully published {}.".format(self.course.id)
+            ),
+            (
+                LOGGER_NAME,
+                "INFO",
+                u"(2/2) Successfully published {}.".format(second_course.id)),
+            (
+                LOGGER_NAME,
+                "INFO",
+                "All 2 courses successfully published."
+            )
+        )
+        with mock.patch.object(Course, 'publish_to_lms', autospec=True) as mock_publish:
+            mock_publish.return_value = None
+            with LogCapture(LOGGER_NAME) as lc:
+                call_command('publish_to_lms', course_ids_file=self.tmp_file_path)
+                lc.check(*expected)
+        # Check that the mocked function was called twice.
+        self.assertListEqual(mock_publish.call_args_list, [call(self.course), call(second_course)])
+
+    def test_course_publish_failed(self):
+        """ Verify failed courses are logged."""
+
+        error_msg = "The failure message."
+        expected = (
+            (
+                LOGGER_NAME,
+                "INFO",
+                "Publishing 1 courses."
+            ),
+            (
+                LOGGER_NAME,
+                "ERROR",
+                u"(1/1) Failed to publish {}: {}".format(self.course.id, error_msg)
+            ),
+            (
+                LOGGER_NAME,
+                "ERROR",
+                "Completed publishing courses. 1 of 1 failed."
+            )
+        )
+        with mock.patch.object(Course, 'publish_to_lms') as mock_publish:
+            mock_publish.return_value = error_msg
+            with LogCapture(LOGGER_NAME) as lc:
+                call_command('publish_to_lms', course_ids_file=self.tmp_file_path)
+                lc.check(*expected)
+            mock_publish.assert_called_once_with()
+
+    def test_unicode_file_name(self):
+        """ Verify the unicode files name are read correctly."""
+        unicode_file = os.path.join(tempfile.gettempdir(), u"اول.txt")
+        self.create_course_ids_file(unicode_file, [self.course.id])
+        expected = (
+            (
+                LOGGER_NAME,
+                "INFO",
+                "Publishing 1 courses."
+            ),
+            (
+                LOGGER_NAME,
+                "INFO",
+                u"(1/1) Successfully published {}.".format(self.course.id)
+            ),
+            (
+                LOGGER_NAME,
+                "INFO",
+                "All 1 courses successfully published."
+            )
+        )
+        with mock.patch.object(Course, 'publish_to_lms') as mock_publish:
+            mock_publish.return_value = None
+            with LogCapture(LOGGER_NAME) as lc:
+                call_command('publish_to_lms', course_ids_file=unicode_file)
+                lc.check(*expected)
+
+        mock_publish.assert_called_once_with()
+        os.remove(unicode_file)


aamir-khan:
@edx/ecommerce @rlucioni @clintonb @bderusha @peter-fogg Can you please review this PR.

ecommerce/courses/management/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'arbisoft'
```
adampalay: this line should be removed

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    def handle(self, *args, **options):
+        errors_dict = {}
+        course_ids_file = options['course_ids_file']
+        if not course_ids_file or not os.path.exists(course_ids_file):
+            raise CommandError("Pass the correct absolute path to course ids file as --course_ids_file argument.")
+
+        with open(course_ids_file, 'r') as course_ids:
+            self.stderr.write("Publishing courses to LMS")
```
adampalay: why write to stderr? Why not just print?

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,57 @@
...
+            for course_id in course_ids:
+                try:
+                    course_id = course_id.strip()
+                    course = Course.objects.get(id=course_id.strip())
```
adampalay: `course_id` was already stripped in the line above

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,80 @@
...
+from ecommerce.courses.models import Course
+from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.tests.testcases import TestCase
+
+
+@ddt.ddt
+class PublishCoursesToLMSTests(TestCase):
+    """Tests the course publish command."""
+
+    tmp_file_path = os.path.join(tempfile.gettempdir(), "tmp-testfile.txt")
```
adampalay, line 29: would this tempfile be cleared after tests are run? It isn't clear to me that it would

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,80 @@
...
+        self.course = CourseFactory()
+        self.create_course_ids_file(self.course.id)
+
+    def create_course_ids_file(self, course_id):
+        with open(self.tmp_file_path, 'w') as f:
+            f.write(course_id)
+
+    @ddt.data("", "fake/path")
+    def test_invalid_file_path(self, course_ids_file):
+        """ The command should raise CommandError if invalid path is passed. """
```
adampalay: no need for doc strings in these tests (the function names are descriptive enough)

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,57 @@
...
+                    publishing_error = course.publish_to_lms()
+                    if publishing_error:
+                        errors_dict[course_id] = publishing_error
+                except Course.DoesNotExist:
+                    errors_dict[course_id] = "The course {course_id} does not exists.".format(course_id=course_id)
+        if errors_dict:
+            error_str = [
+                "Course_id={} Error={}".format(course_id, error_msg) for course_id, error_msg in errors_dict.items()
+                ]
```
adampalay: de-indent this bracket

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,57 @@
...
+            error_str = "\n".join(error_str)
+            error_str = "\n".join(["Following Courses failed while publishing through management command:", error_str])
+            self.stderr.write(error_str)
+            logger.error(error_str)
+        else:
+            self.stderr.write("All Courses published successfully.")
```
adampalay: @peter-fogg , @bderusha , strong opinions about whether or not this should be logged or printed?

adampalay:
@aamir-khan , first pass complete

ecommerce/courses/management/commands/publish_to_lms.py
clintonb: Use standard Python logging.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,57 @@
+""" This command publish the courses to LMS."""
+
+from __future__ import with_statement
```
clintonb: Why is this necessary?

ecommerce/courses/tests/test_publish_to_lms_command.py
clintonb: I disagree. I prefer tests to be explicit up front about what they are testing.

clintonb:
Why was it decided to use file input rather than re-publish all courses?

adampalay:
@clintonb , to give us flexibility if the audit script didn't give back a clean bill of health

ecommerce/courses/tests/test_publish_to_lms_command.py
aamir-khan, line 29: Added the tearDownClass method to clear the tempfile on test suit completion.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def setUp(self):
+        super(PublishCoursesToLMSTests, self).setUp()
+        self.course = CourseFactory()
+        self.create_course_ids_file(self.course.id)
+
+    @classmethod
+    def tearDownClass(cls):
+        if os.path.exists(cls.tmp_file_path):
+            try:
```
adampalay: why a try/except here? If this wasn't working, I think we'd want the test to fail so we'd know about it

ecommerce/courses/management/commands/publish_to_lms.py
adampalay: @aamir-khan ^^

cpennington: Actually, for management commands, using `self.stderr` and `self.stdout` to display messages to the user is the correct usage.

clintonb: We want this information to go to Splunk, which is setup using the Python logging.

cpennington: This message in particular seems like a bad one to go to splunk, since it has no interesting data in it. (Except maybe the timestamp?)

clintonb: Splunk filters to INFO severity or higher. If this message is not important, use `log.debug`.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,54 @@
...
+            self.stderr.write("Publishing courses to LMS")
+            for course_id in course_ids:
+                try:
+                    course_id = course_id.strip()
+                    course = Course.objects.get(id=course_id)
+                    publishing_error = course.publish_to_lms()
+                    if publishing_error:
+                        errors_dict[course_id] = publishing_error
+                except Course.DoesNotExist:
+                    errors_dict[course_id] = "The course {course_id} does not exists.".format(course_id=course_id)
```
peter-fogg: Wording: "does not exist.".

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,54 @@
...
+        if errors_dict:
+            error_str = ["Course_id={} Error={}".format(course_id, error_msg) for course_id, error_msg
+                         in errors_dict.items()]
+            error_str = "\n".join(["Following Courses failed while publishing through management command:"] + error_str)
+            self.stderr.write(error_str)
+            logger.error(error_str)
```
peter-fogg: I don't see any reason to both log and write to `stderr`.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,54 @@
...
+        else:
+            self.stderr.write("All Courses published successfully.")
+            logger.info("All Courses published successfully.")
+        self.stderr.write("Done.")
+        logger.info("Done.")
```
peter-fogg: Same here.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,92 @@
...
+                os.remove(cls.tmp_file_path)
+            except:  # pylint: disable=bare-except
+                logger.error("Error on deleting the temp file %s", cls.tmp_file_path)
```
peter-fogg: I agree with @adampalay above. Though, if we are going to catch the exception here, we should at least log its message.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def create_course_ids_file(self, course_id):
+        """Write the course_id to the temp file."""
+
+        with open(self.tmp_file_path, 'w') as f:
+            f.write(course_id)
```
peter-fogg: Might be worth testing the command's file parsing by using multiple course IDs.

ecommerce/courses/tests/test_publish_to_lms_command.py
clintonb: `logger.exception()`

ecommerce/courses/management/commands/publish_to_lms.py
peter-fogg: For this message, I don't see any reason I'd ever want to find it in Splunk. As such I lean towards printing.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,54 @@
...
+        if not course_ids_file or not os.path.exists(course_ids_file):
+            raise CommandError("Pass the correct absolute path to course ids file as --course_ids_file argument.")
+
+        with open(course_ids_file, 'r') as course_ids:
+            self.stderr.write("Publishing courses to LMS")
+            for course_id in course_ids:
+                try:
+                    course_id = course_id.strip()
+                    course = Course.objects.get(id=course_id)
+                    publishing_error = course.publish_to_lms()
```
peter-fogg, line 47: On the subject of logging, this would be a good place to note that we're publishing this course through the management command (and not through the normal publication endpoint).

ecommerce/courses/management/commands/publish_to_lms.py
aamir-khan: @peter-fogg My assumption for using both log and stderr is that stderr is for user running the command to see what happened to command he ran and the command progress on terminal instead of searching for logs on splunk. The log is used for splunk logging to check later if anything need to be verified. If it is not correct I can remove one of these logging. @adampalay

ecommerce/courses/tests/test_publish_to_lms_command.py
aamir-khan: Agreed removing the try/except.

ecommerce/courses/management/commands/publish_to_lms.py
clintonb: You can add a handler to send log output to the console. See https://docs.python.org/3.1/library/logging.html#configuring-logging.

aamir-khan:
@edx/ecommerce @rlucioni @clintonb @bderusha @peter-fogg @adampalay All comments have been addressed. Can you give this another review please?

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,53 @@
...
+                    if publishing_error:
+                        errors_dict[course_id] = publishing_error
+                except Course.DoesNotExist:
+                    errors_dict[course_id] = "The course {course_id} does not exist.".format(course_id=course_id)
+        if errors_dict:
+            error_str = ["Course_id={} Error={}".format(course_id, error_msg) for course_id, error_msg
+                         in errors_dict.items()]
+            error_str = "\n".join(["Following Courses failed while publishing through management command:"] + error_str)
```
clintonb: The use of `str.join()` here is somewhat confusing. Try this:

``` python
error_msg = "Errors occurred for the following courses while publishing:\n"
for course_id, error in errors_dict.items():
  error_msg += "{course_id}\t{error}\n".format(course_id=course_id, error=error)
```

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,109 @@
+"""Contains the tests for publish to lms command."""
+
+import logging
+import os
+import tempfile
+
+import ddt
+from django.core.management import call_command
+from django.core.management import CommandError
```
clintonb: Combine into a single import.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,53 @@
+""" This command publish the courses to LMS."""
+
+import logging
+from optparse import make_option
+import os
+
+from django.core.management import BaseCommand, CommandError
```
clintonb: Add a newline after this import.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,109 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        if os.path.exists(cls.tmp_file_path):
+            os.remove(cls.tmp_file_path)
+
+    def create_course_ids_file(self, course_ids):
+        """Write the course_ids list to the temp file."""
+
+        with open(self.tmp_file_path, 'w') as f:
+            # for course_id in course_ids:
```
clintonb: Remove dead code.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,109 @@
...
+    def test_invalid_file_path(self, course_ids_file):
+        """ The command should raise CommandError if invalid path is passed. """
+
+        with self.assertRaises(CommandError):
+            call_command('publish_to_lms', course_ids_file=course_ids_file)
+
+    def test_invalid_course_id(self):
+        """ The course failed due to invalid course id. """
+
+        # Mock the LMS call
```
clintonb: Remove.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @ddt.data("", "fake/path")
+    def test_invalid_file_path(self, course_ids_file):
+        """ The command should raise CommandError if invalid path is passed. """
+
+        with self.assertRaises(CommandError):
+            call_command('publish_to_lms', course_ids_file=course_ids_file)
+
+    def test_invalid_course_id(self):
+        """ The course failed due to invalid course id. """
```
clintonb: This is not a valid docstring for a test case. You should describe what the test validates.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,109 @@
...
+                        'Error=The course fake_course_id does not exist.'.format(fake_course_id)
+                    )
+                )
+
+    def test_course_publish_successfully(self):
+        """ All courses should be published successfully."""
+
+        second_course = CourseFactory.create()
+        self.create_course_ids_file([self.course.id, second_course.id])
+        # Mock the LMS call
```
clintonb: Remove.

ecommerce/courses/tests/test_publish_to_lms_command.py
clintonb: Remove. This comment isn't necessary as it is somewhat obvious what the mock is mocking.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,109 @@
...
+                    (
+                        LOGGER_NAME,
+                        'ERROR',
+                        'Following Courses failed while publishing through management command:\nCourse_id={} '
+                        'Error=The course fake_course_id does not exist.'.format(fake_course_id)
+                    )
+                )
+
+    def test_course_publish_successfully(self):
+        """ All courses should be published successfully."""
```
clintonb: Verify all courses are successfully published.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,109 @@
...
+                        'All Courses published successfully.'
+                    )
+                )
+        # Check that the mocked function was called twice.
+        self.assertListEqual(mock_publish.call_args_list, [call(self.course), call(second_course)])
+
+    def test_course_publish_failed(self):
+        """ The course failed to be published."""
+
+        # Mock the LMS call
```
clintonb: Remove.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,53 @@
...
+                        errors_dict[course_id] = publishing_error
+                except Course.DoesNotExist:
+                    errors_dict[course_id] = "The course {course_id} does not exist.".format(course_id=course_id)
+        if errors_dict:
+            error_str = ["Course_id={} Error={}".format(course_id, error_msg) for course_id, error_msg
+                         in errors_dict.items()]
+            error_str = "\n".join(["Following Courses failed while publishing through management command:"] + error_str)
+            logger.error(error_str)
+        else:
+            logger.debug("All Courses published successfully.")
```
clintonb: `logger.info`

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,109 @@
...
+        with mock.patch.object(Course, 'publish_to_lms') as mock_publish:
+            error_msg = "The failure message."
+            mock_publish.return_value = error_msg
+            with LogCapture(LOGGER_NAME) as l:
```
adampalay: nit: avoid single-character variable names. Can you do `log_capture` or something more descriptive?

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def test_course_publish_successfully(self):
+        """ All courses should be published successfully."""
+
+        second_course = CourseFactory.create()
+        self.create_course_ids_file([self.course.id, second_course.id])
+        # Mock the LMS call
+        with mock.patch.object(Course, 'publish_to_lms', autospec=True) as mock_publish:
+            mock_publish.return_value = None
+            with LogCapture(LOGGER_NAME) as l:
```
adampalay: see below about one-character variables

ecommerce/courses/management/commands/publish_to_lms.py
adampalay, line 47: @peter-fogg , good point. Should we also log each time a course is published?

ecommerce/courses/management/commands/publish_to_lms.py
clintonb: On second thought, I don't like the fact that we are waiting until the end of the process to log errors, or anything for that matter. If I am using this, and it takes 5 minutes to run, I see no output for five minutes. The command should log the state of every publication attempt immediately after said attempt. That way, if the first few attempts fail, we can stop the command and fix the issue.

ecommerce/courses/tests/test_publish_to_lms_command.py
clintonb: We normally use `lc`.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,53 @@
...
+        with open(course_ids_file, 'r') as course_ids:
+            for course_id in course_ids:
+                try:
+                    course_id = course_id.strip()
+                    course = Course.objects.get(id=course_id)
+                    publishing_error = course.publish_to_lms()
+                    if publishing_error:
+                        errors_dict[course_id] = publishing_error
+                except Course.DoesNotExist:
+                    errors_dict[course_id] = "The course {course_id} does not exist.".format(course_id=course_id)
```
peter-fogg: This string should be unicode (`u"The course ..."`), since course IDs can have non-ASCII characters. Same with all other strings in this file -- using `from __future__ import unicode_literals` might be a good idea.

ecommerce/courses/management/commands/publish_to_lms.py
peter-fogg, line 47: Yeah, that's my point -- I think `logger.info` would be the right logging level.

ecommerce/courses/management/commands/publish_to_lms.py
aamir-khan: @clintonb I agree with your suggestion for logging the state of every publication. I think, while iterating the courses we should log the state of every publication(success/failure) with DEBUG level that will only be displayed on terminal and will not go to splunk. All the failed publications should be logged at the end with ERROR level so that it will be easy to check which courses failed after the command completed.

clintonb: Why is everyone so afraid to log to Splunk. We _want_ to keep track of these logs, not let them die on the respective servers. Please log successes as INFO. Knowing which courses succeeded as just as important as knowing which failed.

ecommerce/courses/tests/test_publish_to_lms_command.py
```diff
@@ -0,0 +1,97 @@
...
+            os.remove(cls.tmp_file_path)
+
+    def create_course_ids_file(self, course_ids):
+        """Write the course_ids list to the temp file."""
+
+        with open(self.tmp_file_path, 'w') as temp_file:
+            temp_file.write("\n".join(course_ids))
+
+    @ddt.data("", "fake/path")
+    def test_invalid_file_path(self, course_ids_file):
```
awaisdar001, line 48: Can we also verify file nam with special chars and unicode chars?

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,60 @@
...
+                total_courses += 1
+                try:
+                    course_id = course_id.strip()
+                    course = Course.objects.get(id=course_id)
+                    publishing_error = course.publish_to_lms()
+                    if publishing_error:
+                        failed += 1
+                        logger.error(u"%s\t%s\n", course_id, publishing_error)
+                    else:
+                        msg = "The course published successfully."
```
clintonb: This variable is not necessary.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,60 @@
...
+                        logger.info(u"%s\t%s\n", course_id, msg)
+                except Course.DoesNotExist:
+                    failed += 1
+                    msg = u"The course {} does not exist.".format(course_id)
```
clintonb: This variable is not necessary.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def handle(self, *args, **options):
+        total_courses = 0
+        failed = 0
+        course_ids_file = options['course_ids_file']
+        if not course_ids_file or not os.path.exists(course_ids_file):
+            raise CommandError("Pass the correct absolute path to course ids file as --course_ids_file argument.")
+
+        with open(course_ids_file, 'r') as course_ids:
+            for course_id in course_ids:
```
clintonb: Log the total count early, so those running the command have an instant sanity check regarding how many courses are being loaded. In general, long-running commands should be somewhat verbose to avoid people thinking the command has somehow stalled or frozen. As @rock345 likes to say, over-communicate rather than under-communicate.

clintonb: Log the total count early, so those running the command have an instant sanity check regarding how many courses are being loaded. In general, long-running commands should be somewhat verbose to avoid people thinking the command has somehow stalled or frozen. As @rock345 likes to say, over-communicate rather than under-communicate.

ecommerce/courses/tests/test_publish_to_lms_command.py
aamir-khan, line 48: @awaisdar001 I added a test for this too.

ecommerce/courses/management/commands/publish_to_lms.py
adampalay: A note on the logging message — can you provide more context? If we're looking this up in splunk, this log is just not going to be useful. Which course published successfully? We should also know that it failed while running a management command, as opposed to failing during a regular publishing event.

adampalay:
@aamir-khan , I suggested improvements to the logging. We want to know more context:
- these logs are being thrown by this management command
- which course is being discussed

If you had to investigate a bug, what is the format of a bug that would be useful to you?

ecommerce/courses/management/commands/publish_to_lms.py
aamir-khan: @clintonb to get the total count we have to call readlines() function that will load all the data in memory instead of using the generator and may cause the problem if the file size is large.

clintonb: That shouldn't be more than 2000 strings today. It'll be fine.

aamir-khan:
@adampalay I have updated the logging.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,64 @@
...
+                            course_id, publishing_error)
+                    else:
+                        logger.info(u"The course %s published successfully through management command.\n", course_id)
+                except Course.DoesNotExist:
+                    failed += 1
+                    logger.error(
+                        u"Error in publishing course %s through management command.\tError detail: The course %s does "
+                        u"not exist.\n", course_id, course_id)
+        if failed:
+            logger.error("Management Command: %s course(s) failed out of %s.", failed, total_courses)
```
clintonb: It will be obvious from the logs that this message originated from a management command. The logger name will be `ecommerce.courses.management.commands.publish_to_lms`. There is no need to point out that these messages are coming from a management command anywhere.

peter-fogg: I realize this may be due to confusion over my comment above -- to clarify, Clinton's right that we don't need to point out that logging from this file is coming from a management command. I meant that we should be doing some logging here, and not just relying on whatever logging happens in the LMS.

adampalay: Ok, sounds good to me

adampalay:
@aamir-khan , sorry to turn this around, but @clintonb pointed out in a previous comment that we'd know what file this code was run in, so it wasn't as important to add information that this was run as a management command to the logs. That being said, still :+1: from me.

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,64 @@
...
+    def handle(self, *args, **options):
+        failed = 0
+        course_ids_file = options['course_ids_file']
+        if not course_ids_file or not os.path.exists(course_ids_file):
+            raise CommandError("Pass the correct absolute path to course ids file as --course_ids_file argument.")
+
+        with open(course_ids_file, 'r') as file_handler:
+            course_ids = file_handler.readlines()
+            total_courses = len(course_ids)
+            logger.info("Publishing %s course(s) through management command.", total_courses)
```
clintonb: There are much simpler ways to publish a single course.

`Publishing %s courses...`

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,64 @@
...
+            for index, course_id in enumerate(course_ids, start=1):
+                try:
+                    course_id = course_id.strip()
+                    course = Course.objects.get(id=course_id)
+                    publishing_error = course.publish_to_lms()
+                    if publishing_error:
+                        failed += 1
+                        logger.error(
+                            u"(%s/%s) Error in publishing course %s through management command.\tError detail: %s\n",
```
clintonb: `(%d/%d) Failed to publish %s: %s`

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,64 @@
...
+                            index, total_courses, course_id, publishing_error)
+                    else:
+                        logger.info(u"(%s/%s) The course %s published successfully through management command.\n",
```
clintonb: `(%d/%d) Successfully published %s.`

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,64 @@
...
+                                    index, total_courses, course_id)
+                except Course.DoesNotExist:
+                    failed += 1
+                    logger.error(
+                        u"(%s/%s) Error in publishing course %s through management command.\tError detail: "
```
clintonb: `(%d/%d) Failed to publish %s: Course does not exist.`

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,64 @@
...
+                        u"The course %s does not exist.\n", index, total_courses, course_id, course_id)
+        if failed:
+            logger.error("%s course(s) failed out of %s.", failed, total_courses)
```
clintonb: `Completed publishing courses. %d of %d failed.`

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -0,0 +1,64 @@
...
+        else:
+            logger.info("All %s course(s) published successfully.", total_courses)
```
clintonb: `All %d courses successfully published.`

clintonb:
:+1: (once log statements updated)

PR #509, opened 2016-01-20
Building all branches on Travis

clintonb:
There is no good reason to disable builds on non-master branches. Also, doing so allows developers to get build feedback without opening a PR.


clintonb:
FYI @edx/ecommerce

PR #514, opened 2016-01-22
Use the correct format when suggesting a course ID.

agaylard:


clintonb:
New format. The old one is still correct.

:+1:

PR #516, opened 2016-01-26
Use heartbeats and disable connection pooling to prevent broker connection loss

rlucioni:
When the broker is behind a load balancer, the load balancer may timeout Celery's connection to the broker, causing messages to be lost. Since we currently run with RabbitMQ behind a load balancer in production, these changes are necessary to prevent messages from being dropped.

@jimabramson, please review. I've verified with @maxrothman's help that these changes work as expected, preventing messages from being dropped on the floor when Rabbit is behind an ELB.


clintonb:
:+1:  Please add a ticket to the backlog to document this functionality, if one does not already exist.

jimabramson:
:+1:

PR #519, opened 2016-01-28
sleep between requests in management command to publish courses

adampalay:
@clintonb 
@bderusha 

May you please give this a review? I thought it would be prudent to sleep between requests so that we don't bombard the LMS.

I also de-indented some code so that if there's an exception while running the management command, we won't leave the course id filehandler open.

### DIFF
--- a/ecommerce/courses/management/commands/publish_to_lms.py
+++ b/ecommerce/courses/management/commands/publish_to_lms.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import logging
 from optparse import make_option
 import os
+import time
 
 from django.core.management import BaseCommand, CommandError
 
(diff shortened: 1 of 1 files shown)


adampalay:
It may be more helpful to review this without whitespace: https://github.com/edx/ecommerce/pull/519/files?w=1

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -38,25 +39,27 @@ def handle(self, *args, **options):
...
-                except Course.DoesNotExist:
+        total_courses = len(course_ids)
+        logger.info("Publishing %d courses.", total_courses)
+        for index, course_id in enumerate(course_ids, start=1):
+            try:
+                course_id = course_id.strip()
+                course = Course.objects.get(id=course_id)
+                publishing_error = course.publish_to_lms()
+                # sleep, so that we don't bombard the LMS
+                time.sleep(0.1)
```
clintonb, line 50: Instead of sleeping for every request, how about sleeping after every batch of requests? You should also check with DevOps to confirm we actually have throttling enabled. I see no settings for throttling in the repository or on my sandbox; however, the functionality may be configured in the secure settings.

adampalay, line 50: Yeah, we could sleep after a batch. I'm just concerned because I have no idea how slow the LMS endpoint is

clintonb, line 50: The endpoint is pretty quick. Re-publishing all courses should take around 4-5 minutes.

adampalay, line 50: OK, so maybe we don't have to do the sleep then. I'll batch them just to be safe

bderusha, line 50: Batching seems like the way to go. Could we manually batch just by running this command and breaking up the input file so we don't need to do a release first?

clintonb:
Is this still needed? If not, please close this pull request and delete the branch.

PR #523, opened 2016-01-29
Added support for API documentation via Swagger

clintonb:


clintonb:
@peter-fogg @adampalay please review

adampalay:
@clintonb , when I run this locally, I'm getting a bunch of 429s

screenshot 2016-01-29 09 22 51

clintonb:
Yep. We need to update throttling. For now, you can disable it by
commenting out that section in the DRF settings in base.py.

adampalay:
@clintonb , sorry, I don't understand if there's a blocker or not. Other than that, it looks good!

clintonb:
No blockers. I am opting not to increase throttling for the moment. Unless you are continually refreshing the docs, there should not be an issue with throttling.

peter-fogg:
:+1: Nice. Perhaps even swag.

PR #524, opened 2016-02-01
Reenable creation of discount codes

marjev:


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=524

marjev:
@mjfrey Here's the PR for reenabling creation and editing of discount codes. Please review. Thanks!

mjfrey:
looks good.

:+1:

PR #526, opened 2016-02-02
Removed sample/feature flag around CyberSource line item feature

clintonb:
This functionality has been active in production with no issues. It is safe to leave active permanently.

ECOM-3051

### DIFF
--- a/ecommerce/extensions/payment/migrations/0008_remove_cybersource_level23_sample.py
+++ b/ecommerce/extensions/payment/migrations/0008_remove_cybersource_level23_sample.py
@@ -17,14 +17,18 @@ def create_sample(apps, schema_editor):
 
 def delete_sample(apps, schema_editor):
     Sample = apps.get_model('waffle', 'Sample')
-    Sample.objects.get(name=SAMPLE_NAME).delete()
+
+    try:
+        Sample.objects.get(name=SAMPLE_NAME).delete()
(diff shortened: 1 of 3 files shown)


clintonb:
@rlucioni please review

ecommerce/extensions/payment/migrations/0008_remove_cybersource_level23_sample.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+SAMPLE_NAME = 'send_level_2_3_details_to_cybersource'
+
+
+def create_sample(apps, schema_editor):
+    """
+    Enable both existing payment processors.
```
rlucioni: This docstring is inaccurate. The same is true of the docstring in [0007](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/payment/migrations/0007_add_cybersource_level23_sample.py), but I missed it there. Since they don't add any real value, I would remove both docstrings.

rlucioni:
One minor comment about docstrings, :+1: otherwise.

ecommerce/extensions/payment/migrations/0008_remove_cybersource_level23_sample.py
clintonb: Updated.

PR #527, opened 2016-02-02
Updated to Django 1.8.9

clintonb:


clintonb:
FYI @edx/ecommerce

rlucioni:
:+1:

PR #528, opened 2016-02-02
Change IRC to Slack on README

vkaracic:
Open source communication has been moved from IRC to Slack.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,7 +36,7 @@ Reporting Security Issues
 
 Please do not report security issues in public. Please email security@edx.org.
 
-Mailing List and IRC Channel
-----------------------------
+Mailing List and Slack
+----------------------
 
-You can discuss this code on the `edx-code Google Group <https://groups.google.com/forum/#!forum/edx-code>`_ or in the ``#edx-code`` IRC channel on Freenode.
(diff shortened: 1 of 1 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=528

clintonb:
:+1:

README.rst
```diff
@@ -39,4 +39,4 @@ Please do not report security issues in public. Please email security@edx.org.
 Mailing List and IRC Channel
```
nedbat: Maybe also change the title of the section?

vkaracic: Is it correct to change it to `Slack Channel` ? What's the equivalent on Slack for an IRC channel?

nedbat: The term is "Slack Team", but a heading of "Mailing List and Slack" would be fine :)

README.rst
```diff
@@ -36,7 +36,7 @@ Reporting Security Issues
 
 Please do not report security issues in public. Please email security@edx.org.
 
-Mailing List and IRC Channel
+Mailing List and Slack
 ----------------------------
```
clintonb: This should be the same length of the text above.

PR #530, opened 2016-02-03
Refactor prepare_basket function

vkaracic:
New basket utility functions needed for the new basket work. 
Added new `StockRecordFactory` that contains `partner` and `partner_sku`.

### DIFF
--- a/ecommerce/extensions/basket/models.py
+++ b/ecommerce/extensions/basket/models.py
@@ -16,34 +16,21 @@ class Basket(AbstractBasket):
         return OrderNumberGenerator().order_number(self)
 
     @classmethod
-    def get_basket(cls, user, site, new_basket=False):
+    def get_basket(cls, user, site):
         """Retrieve the basket belonging to the indicated user.
 
         If no such basket exists, create a new one. If multiple such baskets exist,
         merge them into one.
-
-        If a new basket is requested find all open baskets associated with that user and delete them.
-        This is needed for the single-item basket flow.  We want to ensure an empty basket so no
-        other products / vouchers are already included.  This flow is for redemption of a single product
-        as well as used for redeeming a product with a dicount code.
         """
-
-        if new_basket:
-            # If new_baket is requested -- delete all open ones for a user and create new one.
-            all_user_baskets = cls.objects.filter(owner=user, status__in=Basket.editable_statuses)
-            for user_basket in all_user_baskets:
-                user_basket.delete()
+        editable_baskets = cls.objects.filter(site=site, owner=user, status__in=Basket.editable_statuses)
+        if len(editable_baskets) == 0:
             basket = cls.objects.create(site=site, owner=user)
         else:
-            editable_baskets = cls.objects.filter(site=site, owner=user, status__in=Basket.editable_statuses)
-            if len(editable_baskets) == 0:
-                basket = cls.objects.create(site=site, owner=user)
-            else:
-                stale_baskets = list(editable_baskets)
-                basket = stale_baskets.pop(0)
-                for stale_basket in stale_baskets:
-                    # Don't add line quantities when merging baskets
-                    basket.merge(stale_basket, add_quantities=False)
+            stale_baskets = list(editable_baskets)
+            basket = stale_baskets.pop(0)
+            for stale_basket in stale_baskets:
+                # Don't add line quantities when merging baskets
+                basket.merge(stale_basket, add_quantities=False)
 
         # Assign the appropriate strategy class to the basket
         basket.strategy = Selector().strategy(user=user)
--- a/ecommerce/extensions/basket/models.py
+++ b/ecommerce/extensions/basket/models.py
@@ -16,21 +16,34 @@ class Basket(AbstractBasket):
         return OrderNumberGenerator().order_number(self)
 
     @classmethod
-    def get_basket(cls, user, site):
+    def get_basket(cls, user, site, new_basket=False):
         """Retrieve the basket belonging to the indicated user.
 
         If no such basket exists, create a new one. If multiple such baskets exist,
         merge them into one.
+
+        If a new basket is requested find all open baskets associated with that user and delete them.
+        This is need for the single-item basket flow.  We want to ensure an empty basket so no
+        other products / vouchers are already included.  This flow is for redemption of a single product
+        as well as used for redeeming a product with a dicount code.
         """
-        editable_baskets = cls.objects.filter(site=site, owner=user, status__in=Basket.editable_statuses)
-        if len(editable_baskets) == 0:
+
+        if new_basket:
+            # If new_baket is requested -- delete all open ones for a user and create new one.
+            all_user_baskets = cls.objects.filter(owner=user, status__in=Basket.editable_statuses)
+            for user_basket in all_user_baskets:
+                user_basket.delete()
             basket = cls.objects.create(site=site, owner=user)
         else:
-            stale_baskets = list(editable_baskets)
-            basket = stale_baskets.pop(0)
-            for stale_basket in stale_baskets:
-                # Don't add line quantities when merging baskets
-                basket.merge(stale_basket, add_quantities=False)
+            editable_baskets = cls.objects.filter(site=site, owner=user, status__in=Basket.editable_statuses)
+            if len(editable_baskets) == 0:
+                basket = cls.objects.create(site=site, owner=user)
+            else:
+                stale_baskets = list(editable_baskets)
+                basket = stale_baskets.pop(0)
+                for stale_basket in stale_baskets:
+                    # Don't add line quantities when merging baskets
+                    basket.merge(stale_basket, add_quantities=False)
 
         # Assign the appropriate strategy class to the basket
         basket.strategy = Selector().strategy(user=user)
--- a/ecommerce/extensions/basket/models.py
+++ b/ecommerce/extensions/basket/models.py
@@ -23,7 +23,7 @@ class Basket(AbstractBasket):
         merge them into one.
 
         If a new basket is requested find all open baskets associated with that user and delete them.
-        This is need for the single-item basket flow.  We want to ensure an empty basket so no
+        This is needed for the single-item basket flow.  We want to ensure an empty basket so no
         other products / vouchers are already included.  This flow is for redemption of a single product
         as well as used for redeeming a product with a dicount code.
         """
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -1,0 +1,58 @@
+import logging
+
+from oscar.core.loading import get_class, get_model
+
+Applicator = get_class('offer.utils', 'Applicator')
+Basket = get_model('basket', 'Basket')
+logger = logging.getLogger(__name__)
+Selector = get_class('partner.strategy', 'Selector')
+StockRecord = get_model('partner', 'StockRecord')
+Voucher = get_model('voucher', 'Voucher')
+
+
+def prepare_basket(request, user, product, voucher=None):
+    """
+    Prepare the basket, add the product, and apply a voucher.
+
+    Existing baskets are merged and flushed. The specified product will
+    be added to the remaining open basket, and the basket will be frozen.
+    The Voucher is applied to the basket and checked for discounts.
+
+    Arguments:
+    site (Site): The site from which the request came.
+    user (User): User who made the request.
+    product (Product): Product to be added to the basket.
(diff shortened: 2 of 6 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=530

ecommerce/tests/factories.py
```diff
@@ -24,3 +27,15 @@ class Meta(object):
 
     site = factory.SubFactory(SiteFactory)
     partner = factory.SubFactory(PartnerFactory)
+
+
+class StockRecordFactory(factory.DjangoModelFactory):
+    product = factory.SubFactory(ProductFactory)
+    partner = factory.SubFactory(PartnerFactory)
+    partner_sku = factory.Sequence(lambda n: 'unit%d' % n)
+    price_currency = "GBP"
```
clintonb: USD

mjfrey:
@bderusha Can you take a look?  I added a param to get_basket to allow for a completely new basket even if one already exists for the user.  We need these utils for the upcoming otto checkout work.

@vkaracic as well.

ecommerce/extensions/basket/models.py
```diff
@@ -16,21 +16,31 @@ def order_number(self):
...
 
     @classmethod
-    def get_basket(cls, user, site):
+    def get_basket(cls, user, site, new_basket=False):
         """Retrieve the basket belonging to the indicated user.
 
         If no such basket exists, create a new one. If multiple such baskets exist,
         merge them into one.
+
+        If a new basket is requested find all open baskets associated with that user and delete them.
```
clintonb: Please clarify the use case(s) for this kwarg in the docstring.

mjfrey:
or @clintonb as you are already looking

ecommerce/extensions/basket/models.py
```diff
@@ -16,21 +16,34 @@ def order_number(self):
         return OrderNumberGenerator().order_number(self)
 
     @classmethod
-    def get_basket(cls, user, site):
+    def get_basket(cls, user, site, new_basket=False):
```
clintonb: This needs to be tested.

mjfrey: @clintonb added a test to the model tests.

ecommerce/extensions/basket/models.py
```diff
@@ -16,21 +16,34 @@ def order_number(self):
...
+        This is needed for the single-item basket flow.  We want to ensure an empty basket so no
+        other products / vouchers are already included.  This flow is for redemption of a single product
+        as well as used for redeeming a product with a dicount code.
         """
-        editable_baskets = cls.objects.filter(site=site, owner=user, status__in=Basket.editable_statuses)
-        if len(editable_baskets) == 0:
+
+        if new_basket:
+            # If new_baket is requested -- delete all open ones for a user and create new one.
+            all_user_baskets = cls.objects.filter(owner=user, status__in=Basket.editable_statuses)
```
clintonb: The actions taken on one site should not affect the other site. This should only touch baskets for the passed in site.

mjfrey: This is actually the problem that this solves.  The current code will find and merge a basket that is not part of the current site. So the new basket option finds all baskets for a given user and deletes them regardless of site.

clintonb: Everything I'm seeing in the method prior to your changes filters by site. Where are you seeing baskets being merged that are not linked to the current site?

mjfrey: In my testing there have been baskets that where not associated with a site and oscar was merging them.  I never could figure out why that was.  That is the reason I opted for the new basket option which would ensure a fresh start.

clintonb: We can't delete baskets. That will lead to a poor user experience. If you need to figure out what bit of code is creating site-less baskets, make the Basket.site column non-nullable and run your tests until failure. The resulting stack trace will give you the culprit.

mjfrey: ok, let me back out the basket change and I will do more debugging.  So we can move forward.

mjfrey:
@clintonb ok I reverted the deletion of baskets.

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    Looks up a product stock record for a given partner/sku
+
+    Arguments:
+    partner: The partner that the product belongs.
+    sku: The sku for the product to lookup.
+
+    Returns:
+    Product: The product that matches the sku.
+    Msg: An error message if the product is not found.
```
clintonb: Raise an exception.

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    basket.add_product(product, 1)
+    if voucher:
+        basket.vouchers.add(voucher)
+        Applicator().apply(basket, user, request)
+        logger.info('Applied Voucher [%s] to basket.', voucher.code)
+
+    return basket
+
+
+def get_product_from_sku(partner, sku):
```
clintonb: What's the use case for this?

ecommerce/extensions/basket/utils.py
mjfrey: why not a message?

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    Looks up a product stock record for a given partner/sku
+
+    Arguments:
+    partner: The partner that the product belongs.
+    sku: The sku for the product to lookup.
+
+    Returns:
+    Product: The product that matches the sku.
```
clintonb: Indent.

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+        basket.vouchers.add(voucher)
+        Applicator().apply(basket, user, request)
+        logger.info('Applied Voucher [%s] to basket.', voucher.code)
+
+    return basket
+
+
+def get_product_from_sku(partner, sku):
+    """
+    Looks up a product stock record for a given partner/sku
```
clintonb: SKU

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Arguments:
+    partner: The partner that the product belongs.
```
clintonb: Indent, and provide the data type

ecommerce/extensions/basket/utils.py
mjfrey: We need this when looking up the product to add when passed a sku.  This is setup for the upcoming checkout page.

ecommerce/extensions/basket/utils.py
clintonb: Because the message is essentially doing the work of exceptions. If you raise an exception, it is clear exactly what went wrong and where (stack trace).

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+def prepare_basket(request, user, product, voucher=None):
+    """
+    Prepare the basket, add the product, and apply a voucher.
+
+    Existing baskets are merged and flushed. The specified product will
+    be added to the remaining open basket, and the basket will be frozen.
+    The Voucher is applied to the basket and checked for discounts.
+
+    Arguments:
+    site (Site): The site from which the request came.
```
clintonb: Indent

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    user (User): User who made the request.
+    product (Product): Product to be added to the basket.
+    voucher (Voucher): Voucher to apply to the basket.
+
+    Returns:
+    basket (Basket): Contains the product to be redeemed and the Voucher applied.
```
clintonb: Indent

ecommerce/extensions/basket/utils.py
clintonb: How often will it be used? Can you link to that code? Without the context, reviewing this is difficult.

ecommerce/extensions/basket/utils.py
clintonb: Also, there is no site variable being passed.

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+Basket = get_model('basket', 'Basket')
+logger = logging.getLogger(__name__)
+Selector = get_class('partner.strategy', 'Selector')
+StockRecord = get_model('partner', 'StockRecord')
+Voucher = get_model('voucher', 'Voucher')
+
+
+def prepare_basket(request, user, product, voucher=None):
+    """
+    Prepare the basket, add the product, and apply a voucher.
```
clintonb: Prepare the basket for what?

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Existing baskets are merged and flushed. The specified product will
+    be added to the remaining open basket, and the basket will be frozen.
+    The Voucher is applied to the basket and checked for discounts.
+
+    Arguments:
+    site (Site): The site from which the request came.
+    user (User): User who made the request.
```
clintonb: User can be pulled from the request.

ecommerce/extensions/basket/utils.py
mjfrey: https://github.com/edx/ecommerce/compare/coupons/new-basket#diff-398e80deb2588fea52e2551897aac3e2R43

The whole point of this was to make the checkout page PR smaller by breaking out some smaller utilities.

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    # remove all existing vouchers from the basket
+    for v in basket.vouchers.all():
+        basket.vouchers.remove(v)
+    basket.reset_offer_applications()
+    basket.flush()
+    basket.add_product(product, 1)
+    if voucher:
+        basket.vouchers.add(voucher)
+        Applicator().apply(basket, user, request)
+        logger.info('Applied Voucher [%s] to basket.', voucher.code)
```
clintonb: Which basket?

ecommerce/extensions/basket/utils.py
clintonb: I get that, and I'm not trying to be difficult. This might be too small of a chunk, however.

clintonb: This doesn't need to be a public utility method given that it is used in only one location. I'd make it a private method on `BasketSingleItemView`. You still want to let the `DoesNotExist` exception bubble up so that you can handle it in `BasketSingleItemView.get()`.

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+Applicator = get_class('offer.utils', 'Applicator')
+Basket = get_model('basket', 'Basket')
+logger = logging.getLogger(__name__)
+Selector = get_class('partner.strategy', 'Selector')
+StockRecord = get_model('partner', 'StockRecord')
+Voucher = get_model('voucher', 'Voucher')
+
+
+def prepare_basket(request, user, product, voucher=None):
```
clintonb: Same as below. This doesn't need to be a utility function. It can be a private method.

mjfrey: I will just modify this stuff in the main branch.  Leave it in the main basket PR.

mjfrey: The reason for this being a utility is it is also used in the enrollment code redemption view.  Not just single-item view.  So a private method would not work.

vkaracic:
@clintonb I've additionally removed some of the code so that only the `prepare_basket` function is left. The reason for placing that method in the basket/utils.py is because we're going to use it in here: https://github.com/edx/ecommerce/blob/coupons/new-basket/ecommerce/extensions/basket/views.py#L52 _(when/if this get's merged the function there will be refactored to send the updated parameter list)_

Can you please take another look?

clintonb:
The goal of the team's request for smaller PRs is to make the review process easier/faster for the reviewer. If we're down to reviewing function-by-function, and you're telling me the function might change in another review, that's not making things easier.

I took a look at your branch, and propose the following breakdown/ordering of PRs:
1. Acceptance tets
2. Basket views
3. Checkout views
4. UI

vkaracic:
Sounds like a plan!

PR #531, opened 2016-02-03
Remove devserve

vkaracic:
Removing the command and documentation for `make devserve` added here: https://github.com/edx/ecommerce/commit/d8e258948cb6e8dc55e740efd2d27261edfc042c
Ecommerce on devstack uses the devstack settings by default now: https://github.com/edx/configuration/pull/2562

@adampalay @clintonb


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=531&repo=edx%2Fecommerce

clintonb:
:+1:

adampalay:
:-1: 
When I run `make serve` on devstack, it runs with production settings:
screenshot 2016-02-04 10 16 21

Maybe there's a bug with configuration? Or maybe I installed my devstack before your config change?

vkaracic:
I believe it's the latter. This is what I get:
screenshot from 2016-02-04 16-20-47

adampalay:
Ok, can we leave this in for backwards compatibility's sake (for example, I don't want to rebuild my devstack just for this?) You can add a note to `make devserve` that it's only in for backwards compatibility purposes.

vkaracic:
Fine by me :)
@clintonb ?

clintonb:
I'd rather not leave unnecessary code in the code base. @adampalay you don't need to rebuild devstack. Modify /edx/app/ecommerce/ecommerce_env to set `DJANGO_SETTINGS_MODULE` to `ecommerce.settings.devstack`.

adampalay:
@clintonb , cool, but we should add that to documentation

PR #532, opened 2016-02-04
Checkout functionality

vkaracic:
@mjfrey

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_checkout.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_checkout.py
@@ -1,0 +1,82 @@
+from __future__ import unicode_literals
+
+import json
+
+from django.conf import settings
+from django.core.urlresolvers import reverse
+from django.test import override_settings
+from oscar.core.loading import get_model
+from waffle.models import Switch
+
+from ecommerce.extensions.payment.tests.processors import DummyProcessor
+from ecommerce.tests.testcases import TestCase
+
+Basket = get_model('basket', 'Basket')
+
+
+class DummyProcessorWithUrl(DummyProcessor):
+    """ Dummy payment processor class that has a test payment page url. """
+    NAME = 'dummy_with_url'
+
+    def get_transaction_parameters(self, basket, request=None):
+        dummy_values = {
+            'payment_page_url': 'test_processor.edx',
+            'transaction_param': 'test_trans_param'
+        }
+        return dummy_values
+
+
+class CheckoutViewTests(TestCase):
+    """ Tests for CheckoutView API view. """
+    path = reverse('api:v2:checkout')
+
+    def setUp(self):
+        super(CheckoutViewTests, self).setUp()
+        self.user = self.create_user()
+        self.client.login(username=self.user.username, password=self.password)
+        Basket.objects.create(owner=self.user)
+        self.data = {
+            'basket_id': 1,
+            'payment_processor': DummyProcessorWithUrl.NAME
+        }
+
+    def toggle_payment_processor(self):
+        """ Helper function to activate the dummy payment processor. """
+        switch, __ = Switch.objects.get_or_create(
+            name=settings.PAYMENT_PROCESSOR_SWITCH_PREFIX + DummyProcessorWithUrl.NAME
+        )
+        switch.active = True
+        switch.save()
+
+    @override_settings(
+        PAYMENT_PROCESSORS=['ecommerce.extensions.api.v2.tests.views.test_checkout.DummyProcessorWithUrl']
+    )
+    def test_authentication_required(self):
+        """ Verify that a guest cannot access the view. """
+        self.toggle_payment_processor()
+        response = self.client.post(self.path, data=self.data)
+        self.assertEqual(response.status_code, 200)
+
+        self.client.logout()
+        response = self.client.post(self.path, data=self.data)
+        self.assertEqual(response.status_code, 401)
+
+    def test_no_basket(self):
+        """ Verify an error happens when a user has no baskets associated. """
+        self.user.baskets.all().delete()
+        response = self.client.post(self.path, data=self.data)
+        self.assertEqual(response.status_code, 400)
+
+    @override_settings(
+        PAYMENT_PROCESSORS=['ecommerce.extensions.api.v2.tests.views.test_checkout.DummyProcessorWithUrl']
+    )
+    def test_view_response(self):
+        """ Verify proper response happens. """
+        self.toggle_payment_processor()
+        response = self.client.post(self.path, data=self.data)
+        self.assertEqual(response.status_code, 200)
+
+        response_data = json.loads(response.content)
+        self.assertEqual(response_data['payment_form_data']['transaction_param'], 'test_trans_param')
+        self.assertEqual(response_data['payment_page_url'], 'test_processor.edx')
+        self.assertEqual(response_data['payment_processor'], 'dummy_with_url')
--- a/ecommerce/extensions/api/v2/tests/views/test_checkout.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_checkout.py
@@ -26,12 +26,16 @@ class DummyProcessorWithUrl(DummyProcessor):
         return dummy_values
 
 
+@override_settings(
+    PAYMENT_PROCESSORS=['ecommerce.extensions.api.v2.tests.views.test_checkout.DummyProcessorWithUrl']
+)
 class CheckoutViewTests(TestCase):
     """ Tests for CheckoutView API view. """
     path = reverse('api:v2:checkout')
 
     def setUp(self):
         super(CheckoutViewTests, self).setUp()
+        self.activate_payment_processor()
         self.user = self.create_user()
         self.client.login(username=self.user.username, password=self.password)
         Basket.objects.create(owner=self.user)
@@ -40,7 +44,7 @@ class CheckoutViewTests(TestCase):
             'payment_processor': DummyProcessorWithUrl.NAME
         }
 
-    def toggle_payment_processor(self):
+    def activate_payment_processor(self):
         """ Helper function to activate the dummy payment processor. """
         switch, __ = Switch.objects.get_or_create(
             name=settings.PAYMENT_PROCESSOR_SWITCH_PREFIX + DummyProcessorWithUrl.NAME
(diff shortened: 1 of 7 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=532&repo=edx%2Fecommerce

mjfrey:
This PR implements a Checkout API endpoint. It takes a basket and a payment processor and prepares the request that can be sent.

This is the first of a few PR's that will implement full checkout for LMS in Otto.

@clintonb @bderusha Can you please look.

ecommerce/extensions/api/v2/tests/views/test_checkout.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def toggle_payment_processor(self):
+        """ Helper function to activate the dummy payment processor. """
+        switch, __ = Switch.objects.get_or_create(
+            name=settings.PAYMENT_PROCESSOR_SWITCH_PREFIX + DummyProcessorWithUrl.NAME
+        )
+        switch.active = True
+        switch.save()
+
+    @override_settings(
```
clintonb: You can avoid duplication by setting this on the class rather than the individual test methods.

ecommerce/extensions/api/v2/tests/views/test_checkout.py
```diff
@@ -0,0 +1,82 @@
...
+        super(CheckoutViewTests, self).setUp()
+        self.user = self.create_user()
+        self.client.login(username=self.user.username, password=self.password)
+        Basket.objects.create(owner=self.user)
+        self.data = {
+            'basket_id': 1,
+            'payment_processor': DummyProcessorWithUrl.NAME
+        }
+
+    def toggle_payment_processor(self):
```
clintonb: For the sake of simplicity, call this method from `setUp`. Also, `activate_payment_processor` is a more appropriate name.

ecommerce/extensions/api/v2/tests/views/test_checkout.py
```diff
@@ -0,0 +1,82 @@
...
+            name=settings.PAYMENT_PROCESSOR_SWITCH_PREFIX + DummyProcessorWithUrl.NAME
+        )
+        switch.active = True
+        switch.save()
+
+    @override_settings(
+        PAYMENT_PROCESSORS=['ecommerce.extensions.api.v2.tests.views.test_checkout.DummyProcessorWithUrl']
+    )
+    def test_authentication_required(self):
+        """ Verify that a guest cannot access the view. """
```
clintonb: Verify the endpoint requires authentication.

ecommerce/extensions/api/v2/tests/views/test_checkout.py
```diff
@@ -0,0 +1,82 @@
...
+        self.toggle_payment_processor()
+        response = self.client.post(self.path, data=self.data)
+        self.assertEqual(response.status_code, 200)
```
clintonb: This block isn't necessary given that you run the same code in other tests.

ecommerce/extensions/api/v2/tests/views/test_checkout.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self.client.logout()
+        response = self.client.post(self.path, data=self.data)
+        self.assertEqual(response.status_code, 401)
+
+    def test_no_basket(self):
+        """ Verify an error happens when a user has no baskets associated. """
```
clintonb: Verify the endpoint returns HTTP 400 if the user has no associated baskets.

ecommerce/extensions/api/v2/tests/views/test_checkout.py
```diff
@@ -0,0 +1,82 @@
...
+        self.user.baskets.all().delete()
+        response = self.client.post(self.path, data=self.data)
+        self.assertEqual(response.status_code, 400)
+
+    @override_settings(
+        PAYMENT_PROCESSORS=['ecommerce.extensions.api.v2.tests.views.test_checkout.DummyProcessorWithUrl']
+    )
+    def test_view_response(self):
+        """ Verify proper response happens. """
```
clintonb: Verify the endpoint returns a successful response when the user is able to checkout.

ecommerce/extensions/api/v2/views/checkout.py
```diff
@@ -0,0 +1,50 @@
...
+
+from ecommerce.extensions.api.serializers import CheckoutSerializer
+from ecommerce.extensions.payment.helpers import get_processor_class_by_name
+
+Applicator = get_class('offer.utils', 'Applicator')
+
+
+class CheckoutView(APIView):
+    """
+    Prepares a basket for checkout and returns information necessary for the browser to redirect the client
```
clintonb: Freezes a basket, and returns the information necessary to start the payment process.

ecommerce/extensions/api/v2/views/checkout.py
```diff
@@ -0,0 +1,50 @@
...
+        # Get the basket, and make sure it belongs to the current user.
+        try:
+            basket = request.user.baskets.get(id=basket_id)
+        except ObjectDoesNotExist:
+            return HttpResponseBadRequest('Basket [{}] not found'.format(basket_id))
+
+        # Freeze the basket so that it cannot be modified
+        basket.strategy = request.strategy
+        Applicator().apply(basket, request.user, request)
+        basket.save()
```
clintonb: No need for this call. `basket.freeze()` will call `save()`.

ecommerce/extensions/api/v2/views/checkout.py
```diff
@@ -0,0 +1,50 @@
...
+
+        basket.freeze()
+
+        # Return the payment info
+        payment_processor = get_processor_class_by_name(payment_processor)()
```
clintonb: This will raise `ProcessorNotFoundError` if the given payment processor does not exist. You should catch that error, log it using `logging.exception()`, and return the appropriate 4xx response. Also, add tests.

ecommerce/static/js/pages/checkout_payment.js
```diff
@@ -0,0 +1,59 @@
+/**
```
clintonb, line 1: This needs tests.

clintonb:
Preliminary review looks good. The JS component needs tests.

ecommerce/extensions/api/v2/tests/views/test_checkout.py
vkaracic: That means that `test_no_basket` test would unnecessarily execute this. In my opinion moving the decorator from other tests to this one and calling it on the tests that need it is simple enough.
But if you insist we can move it.

ecommerce/static/js/pages/checkout_payment.js
```diff
@@ -0,0 +1,59 @@
...
+
+            $paymentButtons.find('.payment-button').click(function (e) {
+                var $btn = $(e.target),
+                    paymentProcessor = $btn.val(),
+                    data = {
+                        basket_id: basketId,
+                        payment_processor: paymentProcessor
+                    };
+
+                $.ajax({
```
bderusha, line 50: Can we add a safety mechanism to this by disabling the button after it's clicked.

Adam recently sent an email to this effect linking to this code snippet https://github.com/edx/edx-platform/blob/master/common/static/common/js/components/utils/view_utils.js#L118-L122

Also what should this do if 400 is returned? Do we still want to redirect to payment provider?

ecommerce/extensions/api/v2/urls.py
```diff
@@ -48,6 +48,7 @@
 
 urlpatterns = [
     url(r'^baskets/', include(BASKET_URLS, namespace='baskets')),
+    url(r'^checkout/$', checkout_views.CheckoutView.as_view(), name='checkout'),
```
bderusha: NIT: Might be nice to follow the existing pattern here. `include(CHECKOUT_URLS, namespace='checkout')`

ecommerce/static/js/pages/checkout_payment.js
marjev, line 1: This is NOT testable.
All the code is hidden in the $(document).ready() closure.
The event handler is anonymous function passed directly into the on function.

ecommerce/static/js/pages/checkout_payment.js
```diff
@@ -0,0 +1,63 @@
...
+                    dataType: 'json',
+                    headers: {
+                        'X-CSRFToken': $.cookie('ecommerce_csrftoken')
+                    },
+                    data: JSON.stringify(data)
+                }).done(function (data) {
+                    redirectToPaymentProvider(data);
+                }).fail(function(jqXHR, errorThrown) {
+                    $('#messages').empty().append(
+                        '<div class="error">' + jqXHR.status + errorThrown + ' occurred during checkout.</div>'
```
bderusha: Is this the messaging we want to expose to the user?  Regardless it will need to go through i18n, but I don't think we typically expose the error message returned by the server directly.

CC/ @srpearce can you review?

ecommerce/static/js/pages/checkout_payment.js
bderusha, line 50: Thank you for adding a failure case, but the button disabling is still not addressed.  Without button disabling I can click the button multiple times; each time sending a checkout request to the the server.

screen shot 2016-02-05 at 10 00 46 am

marjev, line 50: I'm working on that right now. I have to write a test before I can commit the change, otherwise coverage will decrease.

ecommerce/static/js/pages/checkout_payment.js
clintonb, line 1: Of course it's testable. Pretty much everything is testable. You need only make functions for your event handlers:

``` javascript
var onDone = function(data){};
var onFail = function(jqXHR, errorThrown){};
var onReady = function(){
  $.ajax({
    ...
  })
  .done(onDone)
  .fail(onFail);
};

$(document).ready(onReady);
```

ecommerce/static/js/pages/checkout_payment.js
bderusha, line 50: Ah cool!  Sorry I jumped the gun

ecommerce/static/js/pages/checkout_payment.js
marjev: Not sure what kind of messaging we want here since it wasn't specified in the requirements.
Leaving it without message seemed pointless.

clintonb: Please get clarification on the requirements from @mjfrey and @pbaruah.

mjfrey: @marjev "Problem occurred during checkout. Please contact support"

mjfrey:
@clintonb @bderusha I think we have addressed everything.  Can we get a pass for the JS test and add some at a later date?

bderusha:
My comments are addressed. :+1: from me.  I'll defer to clinton about waiting on JS tests.

bderusha:
Also, "at a later date" generally reads as "probably never" can we make sure there is a ticket in the next sprint to address it?

mjfrey:
Yes I will create a ticket for the test.

mjfrey:
Link to ticket:
https://openedx.atlassian.net/browse/SOL-1624

clintonb:
We cannot give a pass on testing this. This is user-facing code that collects payment from users. If it fails, we don't get paid. It must be tested.

mjfrey:
ok, will work on JS test now. 
This PR is now at
https://github.com/edx/ecommerce/pull/536

PR #534, opened 2016-02-05
Add new Basket summary view that can handle Coupons.

mjfrey:
Add a new basket view in otto that displays the product as well as any offers applied.  

The intent of this is to replace the current LMS payment page.

The PR implements the following:
Basket summary view 
Single item api endpoint (adds a single item to a basket, and optionally accepts a coupon code)
coupon tests that use the new api

### DIFF
--- a/ecommerce/core/tests/test_context_processors.py
+++ b/ecommerce/core/tests/test_context_processors.py
@@ -4,10 +4,11 @@ from ecommerce.core.context_processors import core
 from ecommerce.tests.testcases import TestCase
 
 PLATFORM_NAME = 'Test Platform'
+SUPPORT_URL = 'example.com'
 
 
 class CoreContextProcessorTests(TestCase):
-    @override_settings(PLATFORM_NAME=PLATFORM_NAME)
+    @override_settings(PLATFORM_NAME=PLATFORM_NAME, SUPPORT_URL=SUPPORT_URL)
     def test_core(self):
         request = RequestFactory().get('/')
-        self.assertDictEqual(core(request), {'platform_name': PLATFORM_NAME})
+        self.assertDictEqual(core(request), {'platform_name': PLATFORM_NAME, 'support_url': SUPPORT_URL})
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -1,24 +1,21 @@
 import datetime
-from decimal import Decimal
-import json
 
 import ddt
 from django.conf import settings
 from django.core.urlresolvers import reverse
-from django.http import HttpResponseRedirect
 from django.utils.timezone import now
 from django.utils.translation import ugettext_lazy as _
 import httpretty
 from oscar.core.loading import get_class, get_model
-from oscar.test.factories import (OrderFactory, ConditionalOfferFactory, VoucherFactory,
-                                  RangeFactory, BenefitFactory, ProductFactory)
+from oscar.test.factories import OrderFactory, ConditionalOfferFactory, VoucherFactory, RangeFactory
 from oscar.test.utils import RequestFactory
 import pytz
 
-from ecommerce.coupons.views import get_voucher, voucher_is_valid
+from ecommerce.coupons.views import get_voucher_from_code, voucher_is_valid
 from ecommerce.courses.tests.factories import CourseFactory
-from ecommerce.extensions.test.factories import create_coupon
+from ecommerce.extensions.test.factories import create_coupon, prepare_voucher
 from ecommerce.settings import get_lms_url
+from ecommerce.tests.mixins import LmsApiMockMixin
 from ecommerce.tests.testcases import TestCase
 
 Applicator = get_class('offer.utils', 'Applicator')
@@ -29,6 +26,9 @@ Course = get_model('courses', 'Course')
 StockRecord = get_model('partner', 'StockRecord')
 Voucher = get_model('voucher', 'Voucher')
 VoucherApplication = get_model('voucher', 'VoucherApplication')
+
+CONTENT_TYPE = 'application/json'
+COUPON_CODE = 'COUPONTEST'
 
 
 class CouponAppViewTests(TestCase):
@@ -57,80 +57,39 @@ class CouponAppViewTests(TestCase):
         self.assertEqual(response.status_code, 200)
 
 
-@ddt.ddt
-class CouponOfferViewTests(TestCase):
-    path = reverse('coupons:offer')
-    path_with_code = '{path}?code={code}'.format(path=path, code='COUPONTEST')
+class GetVoucherTests(TestCase):
 
-    def setUp(self):
-        super(CouponOfferViewTests, self).setUp()
+    def test_get_voucher_from_code(self):
+        """ Verify that get_voucher_from_code() returns product and voucher. """
+        original_voucher, original_product = prepare_voucher(code=COUPON_CODE)
+        voucher, product = get_voucher_from_code(code=COUPON_CODE)
 
-    def prepare_voucher(self, range_=None, start_datetime=None, benefit_value=100, benefit_type='Percentage'):
-        """ Create a voucher and add an offer to it that contains a created product. """
-        if range_ is None:
-            product = ProductFactory(title='Test product')
-            range_ = RangeFactory(products=[product, ])
-        else:
-            product = range_.all_products()[0]
-
-        if start_datetime is None:
-            start_datetime = now() - datetime.timedelta(days=1)
-
-        voucher = VoucherFactory(code='COUPONTEST', start_datetime=start_datetime, usage=Voucher.SINGLE_USE)
-        benefit = BenefitFactory(type=benefit_type, range=range_, value=benefit_value)
-        offer = ConditionalOfferFactory(benefit=benefit)
-        voucher.offers.add(offer)
-        return voucher, product
-
-    def prepare_course_information(self):
-        """ Helper function to prepare an API endpoint that provides course information. """
-        course = CourseFactory()
-        seat = course.create_or_update_seat('verified', True, 50, self.partner)
-        sr = StockRecord.objects.get(product=seat)
-        catalog = Catalog.objects.create(name='Test catalog', partner=self.partner)
-        catalog.stock_records.add(sr)
-        range_ = RangeFactory(catalog=catalog)
-        course_info = {
-            "media": {
-                "course_image": {
-                    "uri": "/asset-v1:edX+DemoX+Demo_Course+type@asset+block@images_course_image.jpg"
-                }
-            },
-            "name": "edX Demonstration Course",
-        }
-        course_info_json = json.dumps(course_info)
-        course_url = get_lms_url('api/courses/v1/courses/{}/'.format(course.id))
-        httpretty.register_uri(httpretty.GET, course_url, body=course_info_json, content_type='application/json')
-        return range_
-
-    def test_get_voucher(self):
-        """ Verify that get_voucher() returns product and voucher. """
-        self.prepare_voucher()
-        voucher, product = get_voucher(code='COUPONTEST')
         self.assertIsNotNone(voucher)
-        self.assertEqual(voucher.code, 'COUPONTEST')
+        self.assertEqual(voucher, original_voucher)
+        self.assertEqual(voucher.code, COUPON_CODE)
         self.assertIsNotNone(product)
-        self.assertEqual(product.title, 'Test product')
+        self.assertEqual(product, original_product)
 
     def test_no_product(self):
         """ Verify that None is returned if there is no product. """
         voucher = VoucherFactory(code='NOPRODUCT')
         offer = ConditionalOfferFactory()
         voucher.offers.add(offer)
-        voucher, product = get_voucher(code='NOPRODUCT')
+        voucher, product = get_voucher_from_code(code='NOPRODUCT')
+
         self.assertIsNotNone(voucher)
         self.assertEqual(voucher.code, 'NOPRODUCT')
         self.assertIsNone(product)
 
     def test_get_non_existing_voucher(self):
-        """ Verify that get_voucher() returns None for non-existing voucher. """
-        voucher, product = get_voucher(code='INVALID')
+        """ Verify that get_voucher_from_code() returns None for a non-existing voucher. """
+        voucher, product = get_voucher_from_code(code='INVALID')
         self.assertIsNone(voucher)
         self.assertIsNone(product)
 
     def test_valid_voucher(self):
         """ Verify voucher_is_valid() assess that the voucher is valid. """
-        voucher, product = self.prepare_voucher()
+        voucher, product = prepare_voucher(code=COUPON_CODE)
         request = RequestFactory().request()
         valid, msg = voucher_is_valid(voucher=voucher, product=product, request=request)
         self.assertTrue(valid)
@@ -145,7 +104,7 @@ class CouponOfferViewTests(TestCase):
     def test_expired_voucher(self):
         """ Verify voucher_is_valid() assess that the voucher has expired. """
         future_datetime = now() + datetime.timedelta(days=10)
-        voucher, product = self.prepare_voucher(start_datetime=future_datetime)
+        voucher, product = prepare_voucher(code=COUPON_CODE, start_datetime=future_datetime)
         valid, msg = voucher_is_valid(voucher=voucher, product=product, request=None)
         self.assertFalse(valid)
         self.assertEqual(msg, _('Coupon expired'))
@@ -153,14 +112,14 @@ class CouponOfferViewTests(TestCase):
     def test_voucher_unavailable_to_buy(self):
         """ Verify that False is returned for unavialable products. """
         request = RequestFactory().request()
-        voucher, product = self.prepare_voucher()
+        voucher, product = prepare_voucher(code=COUPON_CODE)
         product.expires = pytz.utc.localize(datetime.datetime.min)
         valid, __ = voucher_is_valid(voucher=voucher, product=product, request=request)
         self.assertFalse(valid)
 
     def test_used_voucher(self):
         """ Verify voucher_is_valid() assess that the voucher is unavailable. """
-        voucher, product = self.prepare_voucher()
+        voucher, product = prepare_voucher(code=COUPON_CODE)
         order = OrderFactory()
         user = self.create_user()
         VoucherApplication.objects.create(voucher=voucher, user=user, order=order)
@@ -168,6 +127,28 @@ class CouponOfferViewTests(TestCase):
         valid, msg = voucher_is_valid(voucher=voucher, product=product, request=request)
         self.assertFalse(valid)
         self.assertEqual(msg, _('This coupon has already been used'))
+
+
+@httpretty.activate
+@ddt.ddt
+class CouponOfferViewTests(LmsApiMockMixin, TestCase):
+    path = reverse('coupons:offer')
+    path_with_code = '{path}?code={code}'.format(path=path, code=COUPON_CODE)
+
+    def setUp(self):
+        super(CouponOfferViewTests, self).setUp()
+        self.mock_footer_api_response()
+
+    def prepare_course_information(self):
+        """ Helper function to prepare an API endpoint that provides course information. """
+        course = CourseFactory(name='Test course')
+        seat = course.create_or_update_seat('verified', True, 50, self.partner)
+        stock_record = StockRecord.objects.get(product=seat)
+        catalog = Catalog.objects.create(name='Test catalog', partner=self.partner)
+        catalog.stock_records.add(stock_record)
+        _range = RangeFactory(catalog=catalog)
+        self.mock_course_api_response(course=course)
+        return _range
 
     def test_no_code(self):
         """ Verify a proper response is returned when no code is supplied. """
@@ -180,16 +161,15 @@ class CouponOfferViewTests(TestCase):
         response = self.client.get(url)
         self.assertEqual(response.context['error'], _('Coupon does not exist'))
 
-    @httpretty.activate
     def test_course_information_error(self):
         """ Verify a response is returned when course information is not accessable. """
         course = CourseFactory()
         seat = course.create_or_update_seat('verified', True, 50, self.partner)
-        range_ = RangeFactory(products=[seat, ])
-        self.prepare_voucher(range_=range_)
+        _range = RangeFactory(products=[seat, ])
+        prepare_voucher(code=COUPON_CODE, _range=_range)
 
         course_url = get_lms_url('api/courses/v1/courses/{}/'.format(course.id))
-        httpretty.register_uri(httpretty.GET, course_url, status=404, content_type='application/json')
+        httpretty.register_uri(httpretty.GET, course_url, status=404, content_type=CONTENT_TYPE)
 
         response = self.client.get(self.path_with_code)
         response_text = (
@@ -198,16 +178,14 @@ class CouponOfferViewTests(TestCase):
         ).format(course.id)
         self.assertEqual(response.context['error'], _(response_text))
 
-    @httpretty.activate
     def test_proper_code(self):
         """ Verify that proper information is returned when a valid code is provided. """
-        range_ = self.prepare_course_information()
-        self.prepare_voucher(range_=range_)
+        _range = self.prepare_course_information()
+        prepare_voucher(code=COUPON_CODE, _range=_range)
         response = self.client.get(self.path_with_code)
-        self.assertEqual(response.context['course']['name'], _('edX Demonstration Course'))
-        self.assertEqual(response.context['code'], _('COUPONTEST'))
+        self.assertEqual(response.context['course']['name'], _('Test course'))
+        self.assertEqual(response.context['code'], COUPON_CODE)
 
-    @httpretty.activate
     @ddt.data(
         (5, '45.00'),
         (100000, '0.00'),
@@ -215,8 +193,8 @@ class CouponOfferViewTests(TestCase):
     @ddt.unpack
     def test_fixed_amount(self, benefit_value, new_price):
         """ Verify a new price is calculated properly with fixed price type benefit. """
-        range_ = self.prepare_course_information()
-        self.prepare_voucher(range_=range_, benefit_value=benefit_value, benefit_type=Benefit.FIXED_PRICE)
+        _range = self.prepare_course_information()
+        prepare_voucher(code=COUPON_CODE, _range=_range, benefit_value=benefit_value, benefit_type=Benefit.FIXED_PRICE)
         response = self.client.get(self.path_with_code)
         self.assertEqual(response.context['new_price'], new_price)
 
@@ -233,6 +211,17 @@ class CouponRedeemViewTests(TestCase):
 
         self.catalog = Catalog.objects.create(partner=self.partner)
         self.catalog.stock_records.add(StockRecord.objects.get(product=self.seat))
+
+    def create_and_test_coupon(self):
+        """ Creates enrollment code coupon. """
+        create_coupon(catalog=self.catalog, code=COUPON_CODE)
+        self.assertEqual(Voucher.objects.filter(code=COUPON_CODE).count(), 1)
+
+    def assert_redemption_page_redirects(self, expected_url, target=200):
+        """ Verify redirect from redeem page to expected page. """
+        url = self.redeem_url + '?code={}'.format(COUPON_CODE)
+        response = self.client.get(url)
+        self.assertRedirects(response, expected_url, status_code=302, target_status_code=target)
 
     def test_login_required(self):
         """ Users are required to login before accessing the view. """
@@ -252,43 +241,33 @@ class CouponRedeemViewTests(TestCase):
         response = self.client.get(url)
         self.assertEqual(response.context['error'], _('Coupon does not exist'))
 
-    def test_basket_not_free(self):
-        """ Verify a response message is returned when the basket is not free. """
-        self.assertEqual(StockRecord.objects.get(product=self.seat).price_excl_tax, Decimal('50.00'))
-        create_coupon(catalog=self.catalog, code='COUPONTEST', benefit_value=0)
-        url = self.redeem_url + '?code={}'.format('COUPONTEST')
-        response = self.client.get(url)
-        self.assertEqual(str(response.context['error']), _('Basket total not $0, current value = $50.00'))
-
     def test_order_not_completed(self):
         """ Verify a response message is returned when an order is not completed. """
-        create_coupon(catalog=self.catalog, code='COUPONTEST')
-        self.assertEqual(Voucher.objects.filter(code='COUPONTEST').count(), 1)
-
-        url = self.redeem_url + '?code={}'.format('COUPONTEST')
+        self.create_and_test_coupon()
+        url = self.redeem_url + '?code={}'.format(COUPON_CODE)
         response = self.client.get(url)
         self.assertEqual(response.context['error'], _('Error when trying to redeem code'))
 
-    @httpretty.activate
-    def test_redirect(self):
-        """ Verify a redirect happens when valid info is provided. """
-        create_coupon(catalog=self.catalog, code='COUPONTEST')
-        self.assertEqual(Voucher.objects.filter(code='COUPONTEST').count(), 1)
+    def test_basket_redirect_discount_code(self):
+        """ Verify the view redirects to the basket single-item view when a discount code is provided. """
+        create_coupon(catalog=self.catalog, code=COUPON_CODE, benefit_value=5)
+        sku = StockRecord.objects.get(product=self.seat).partner_sku
+        test_server_url = self.get_full_url(path=reverse('basket:single-item'))
+        expected_url = '{url}?sku={sku}&code={code}'.format(url=test_server_url, sku=sku, code=COUPON_CODE)
+        self.assert_redemption_page_redirects(expected_url, 303)
 
+    @httpretty.activate
+    def test_basket_redirect_enrollment_code(self):
+        """ Verify the view redirects to LMS when an enrollment code is provided. """
+        self.create_and_test_coupon()
         httpretty.register_uri(httpretty.POST, settings.ENROLLMENT_API_URL, status=200)
-        url = self.redeem_url + '?code={}'.format('COUPONTEST')
-        response = self.client.get(url)
-        self.assertIsInstance(response, HttpResponseRedirect)
+        self.assert_redemption_page_redirects(settings.LMS_URL_ROOT)
 
     @httpretty.activate
     def test_multiple_vouchers(self):
-        """ Verify a redirect happens when a basket with already existing vouchers is used. """
-        create_coupon(catalog=self.catalog, code='COUPONTEST')
-        self.assertEqual(Voucher.objects.filter(code='COUPONTEST').count(), 1)
+        """ Verify a redirect to LMS happens when a basket with already existing vouchers is used. """
+        self.create_and_test_coupon()
         basket = Basket.get_basket(self.user, self.site)
-        basket.vouchers.add(Voucher.objects.get(code='COUPONTEST'))
+        basket.vouchers.add(Voucher.objects.get(code=COUPON_CODE))
         httpretty.register_uri(httpretty.POST, settings.ENROLLMENT_API_URL, status=200)
-
-        url = self.redeem_url + '?code={}'.format('COUPONTEST')
-        response = self.client.get(url)
-        self.assertIsInstance(response, HttpResponseRedirect)
+        self.assert_redemption_page_redirects(settings.LMS_URL_ROOT)
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 import logging
 
 from django.contrib.auth.decorators import login_required
+from django.core.urlresolvers import reverse
 from django.http import HttpResponseRedirect
 from django.utils.decorators import method_decorator
 from django.utils.translation import ugettext_lazy as _
@@ -16,6 +17,7 @@ from ecommerce.core.views import StaffOnlyMixin
 from ecommerce.extensions.api import data as data_api
 from ecommerce.extensions.api.constants import APIConstants as AC
 from ecommerce.extensions.api.data import get_lms_footer
+from ecommerce.extensions.basket.utils import prepare_basket
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
 from ecommerce.extensions.fulfillment.status import ORDER
 from ecommerce.settings import get_lms_url
@@ -29,9 +31,9 @@ StockRecord = get_model('partner', 'StockRecord')
 Voucher = get_model('voucher', 'Voucher')
 
 
-def get_voucher(code):
+def get_voucher_from_code(code):
     """
-    Returns a voucher and prodcut for a given code.
+    Returns a voucher and product for a given code.
 
     Arguments:
         code (str): The code of a coupon voucher.
@@ -100,7 +102,7 @@ class CouponOfferView(TemplateView):
         footer = get_lms_footer()
         code = self.request.GET.get('code', None)
         if code is not None:
-            voucher, product = get_voucher(code=code)
+            voucher, product = get_voucher_from_code(code=code)
             valid_voucher, msg = voucher_is_valid(voucher, product, self.request)
             if valid_voucher:
                 api = EdxRestApiClient(
@@ -165,12 +167,12 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
         if not code:
             return render(request, template_name, {'error': _('Code not provided')})
 
-        voucher, product = get_voucher(code=code)
+        voucher, product = get_voucher_from_code(code=code)
         valid_voucher, msg = voucher_is_valid(voucher, product, request)
         if not valid_voucher:
             return render(request, template_name, {'error': msg})
 
-        basket = self._prepare_basket(request.site, request.user, product, voucher)
+        basket = prepare_basket(request, product, voucher)
         if basket.total_excl_tax == AC.FREE:
             basket.freeze()
             order_metadata = data_api.get_order_metadata(basket)
@@ -194,54 +196,17 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
                 order_total=order_metadata[AC.KEYS.ORDER_TOTAL],
             )
         else:
-            return render(
-                request,
-                template_name,
-                {'error': _('Basket total not $0, current value = ${basket_price}'.format(
-                    basket_price=basket.total_excl_tax
-                ))}
+            partner = request.site.siteconfiguration.partner
+            sku = StockRecord.objects.get(product=product, partner=partner).partner_sku
+            url = '{url}?sku={sku}&code={code}'.format(
+                url=reverse('basket:single-item'),
+                sku=sku,
+                code=code
             )
+            return HttpResponseRedirect(url)
 
         if order.status is ORDER.COMPLETE:
             return HttpResponseRedirect(get_lms_url(''))
         else:
             logger.error('Order was not completed [%s]', order.id)
             return render(request, template_name, {'error': _('Error when trying to redeem code')})
-
-    def _prepare_basket(self, site, user, product, voucher):
-        """
-        Prepare the basket, add the product, and apply a voucher.
-
-        Existing baskets are merged and flushed. The specified product will
-        be added to the remaining open basket, and the basket will be frozen.
-        The Voucher is applied to the basket and checked for discounts.
-
-        Arguments:
-            site (Site): The site from which the request came.
-            user (User): User who made the request.
-            product (Product): Product to be added to the basket.
-            voucher (Voucher): Voucher to apply to the basket.
-
-        Returns:
-            basket (Basket): Contains the product to be redeemed and the Voucher applied.
-        """
-        basket = Basket.get_basket(user, site)
-        basket.thaw()
-        # remove all existing vouchers from the basket
-        for v in basket.vouchers.all():
-            basket.vouchers.remove(v)
-        basket.reset_offer_applications()
-        basket.flush()
-        basket.add_product(product, 1)
-        basket.vouchers.add(voucher)
-        Applicator().apply(basket, user, self.request)
-        discounts = basket.offer_applications
-        # Look for discounts from this new voucher
-        for discount in discounts:
-            if discount['voucher'] and discount['voucher'] == voucher:
-                logger.info('Applied Voucher [%s] to basket.', voucher.code)
-                break
-            else:
-                logger.info('Voucher [%s] does not offer a discount.', voucher.code)
-                basket.vouchers.remove(voucher)
-        return basket
--- a/ecommerce/extensions/basket/app.py
+++ b/ecommerce/extensions/basket/app.py
@@ -1,0 +1,23 @@
+from django.conf.urls import url
+from django.contrib.auth.decorators import login_required
+from oscar.apps.basket import app
+from oscar.core.loading import get_class
+
+
+class BasketApplication(app.BasketApplication):
+    single_item_view = get_class('basket.views', 'BasketSingleItemView')
+    summary_view = get_class('basket.views', 'BasketSummaryView')
+
+    def get_urls(self):
+        urls = [
+            url(r'^$', self.summary_view.as_view(), name='summary'),
+            url(r'^add/(?P<pk>\d+)/$', self.add_view.as_view(), name='add'),
+            url(r'^vouchers/add/$', self.add_voucher_view.as_view(), name='vouchers-add'),
+            url(r'^vouchers/(?P<pk>\d+)/remove/$', self.remove_voucher_view.as_view(), name='vouchers-remove'),
+            url(r'^saved/$', login_required(self.saved_view.as_view()), name='saved'),
+            url(r'^single-item/$', login_required(self.single_item_view.as_view()), name='single-item'),
+        ]
+        return self.post_process_urls(urls)
+
+
+application = BasketApplication()
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -1,0 +1,72 @@
+import ddt
+
+from django.test import RequestFactory
+from oscar.core.loading import get_model
+from oscar.test.factories import ProductFactory, RangeFactory
+
+from ecommerce.extensions.basket.utils import get_certificate_type_display_value, prepare_basket
+from ecommerce.extensions.partner.models import StockRecord
+from ecommerce.extensions.test.factories import prepare_voucher
+from ecommerce.tests.factories import SiteConfigurationFactory
+from ecommerce.tests.testcases import TestCase
+
+Benefit = get_model('offer', 'Benefit')
+Basket = get_model('basket', 'Basket')
+
+
+@ddt.ddt
+class BasketUtilsTests(TestCase):
+    """ Tests for basket utility functions. """
+
+    def setUp(self):
+        super(BasketUtilsTests, self).setUp()
+        self.request = RequestFactory()
+        self.request.user = self.create_user()
+        site_configuration = SiteConfigurationFactory(partner__name='Tester')
+        self.request.site = site_configuration.site
+
+    def test_prepare_basket_with_voucher(self):
+        """ Verify a basket is returned and contains a voucher and the voucher is applied. """
+        # Prepare a product with price of 100 and a voucher with 10% discount for that product.
+        product = ProductFactory(stockrecords__price_excl_tax=100)
+        new_range = RangeFactory(products=[product, ])
+        voucher, product = prepare_voucher(_range=new_range, benefit_value=10)
+
+        stock_record = StockRecord.objects.get(product=product)
+        self.assertEqual(stock_record.price_excl_tax, 100.00)
+
+        basket = prepare_basket(self.request, product, voucher)
+        self.assertIsNotNone(basket)
+        self.assertEqual(basket.status, Basket.FROZEN)
+        self.assertEqual(basket.lines.count(), 1)
+        self.assertEqual(basket.lines.first().product, product)
+        self.assertEqual(basket.vouchers.count(), 1)
+        self.assertIsNotNone(basket.applied_offers())
+        self.assertEqual(basket.total_discount, 10.00)
+        self.assertEqual(basket.total_excl_tax, 90.00)
+
+    def test_prepare_basket_without_voucher(self):
+        """ Verify a basket is returned and does not contain a voucher. """
+        product = ProductFactory()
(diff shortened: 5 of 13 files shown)


mjfrey:
@clintonb @bderusha Can you take a look?

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+            return HttpResponseServerError('No Partner is associated with this site.')
+
+        sku = request.GET.get('sku', None)
+        code = request.GET.get('code', None)
+
+        if not sku:
+            return HttpResponseBadRequest('No SKU provided.')
+
+        if code:
+            voucher, _ = get_voucher_from_code(code=code)
```
clintonb: Use two underscores, `__`.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+logger = logging.getLogger(__name__)
+StockRecord = get_model('partner', 'StockRecord')
+
+
+class BasketSingleItemView(View):
+    def get(self, request):
+        partner = get_partner_for_site(request)
+        if not partner:
+            return HttpResponseServerError('No Partner is associated with this site.')
```
clintonb: Mark for translation.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+        sku = request.GET.get('sku', None)
+        code = request.GET.get('code', None)
+
+        if not sku:
+            return HttpResponseBadRequest('No SKU provided.')
```
clintonb: Mark for translation.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+        if code:
+            voucher, _ = get_voucher_from_code(code=code)
+        else:
+            voucher = None
+
+        try:
+            product = StockRecord.objects.get(partner=partner, partner_sku=sku).product
+        except StockRecord.DoesNotExist:
+            return HttpResponseBadRequest('SKU [{sku}] does not exist.'.format(sku=sku))
```
clintonb: Mark for translation.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+        purchase_info = request.strategy.fetch_for_product(product)
+        if not purchase_info.availability.is_available_to_buy:
+            return HttpResponseBadRequest('Product [{product}] not available to buy.'.format(product=product.title))
```
clintonb: Mark for translation.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+        prepare_basket(request, product, voucher)
+        return HttpResponseRedirect(reverse('basket:summary'), status=303)
+
+
+class BasketSummaryView(BasketView):
+    def get_context_data(self, **kwargs):
+        context = super(BasketSummaryView, self).get_context_data(**kwargs)
+        lines = self.request.basket.lines.all()
+        api = EdxRestApiClient(
```
clintonb: Make this one line.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+from ecommerce.extensions.basket.utils import prepare_basket
+from ecommerce.extensions.payment.helpers import get_processor_class
+from ecommerce.extensions.partner.shortcuts import get_partner_for_site
+from ecommerce.settings import get_lms_url
+
+logger = logging.getLogger(__name__)
+StockRecord = get_model('partner', 'StockRecord')
+
+
+class BasketSingleItemView(View):
```
clintonb: Add a docstring explaining the purpose of this view.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+        purchase_info = request.strategy.fetch_for_product(product)
+        if not purchase_info.availability.is_available_to_buy:
+            return HttpResponseBadRequest('Product [{product}] not available to buy.'.format(product=product.title))
+
+        prepare_basket(request, product, voucher)
+        return HttpResponseRedirect(reverse('basket:summary'), status=303)
+
+
+class BasketSummaryView(BasketView):
```
clintonb: Add a docstring explaining the purpose of this view.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+        context = super(BasketSummaryView, self).get_context_data(**kwargs)
+        lines = self.request.basket.lines.all()
+        api = EdxRestApiClient(
+            get_lms_url('api/courses/v1/'),
+        )
+        for line in lines:
+            course_id = line.product.course_id
+
+            try:
+                line.certificate_type = line.product.attr.certificate_type.capitalize()
```
clintonb: What are you doing here?

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+            except AttributeError as e:
+                line.certificate_type = None
+                logger.exception('Product has no certificate type. [%s]', e)
```
clintonb: This message will not be useful for debugging. It's not clear which product has no certificate type, or why you're even checking for certificate type. Also, there is no need include the exception in the log message. `logger.exception()` automatically does this.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+            try:
+                course = api.courses(course_id).get()
+                course['image_url'] = get_lms_url(course['media']['course_image']['uri'])
+                line.course = course
+            except SlumberHttpBaseException as e:
+                logger.exception('Could not get course information. [%s]', e)
```
clintonb: Same feedback as above. At the very least include the course ID.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+            'payment_processors': self.get_payment_processors(),
+            'homepage_url': get_lms_url(''),
+            'footer': get_lms_footer(),
+            'lines': lines,
+        })
+        return context
+
+    def get_payment_processors(self):
+        """ Retrieve the list of active payment processors. """
+        # TODO Retrieve this information from SiteConfiguration
```
clintonb, line 112: Is this going to be done?

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+            try:
+                course = api.courses(course_id).get()
+                course['image_url'] = get_lms_url(course['media']['course_image']['uri'])
+                line.course = course
+            except SlumberHttpBaseException as e:
+                logger.exception('Could not get course information. [%s]', e)
+
+        applied_offers = self.request.basket.applied_offers()
+        benefit = applied_offers.values()[0].benefit if applied_offers else ''
```
clintonb: What's the significance of empty string? Can `None` be used instead?

ecommerce/tests/factories.py
```diff
@@ -24,3 +27,15 @@ class Meta(object):
 
     site = factory.SubFactory(SiteFactory)
     partner = factory.SubFactory(PartnerFactory)
+
+
+class StockRecordFactory(factory.DjangoModelFactory):
```
clintonb: Oscar already has a `StockRecordFactory`. Why not inherit it and override the relevant fields?

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,92 @@
...
+            course_id = line.product.course_id
+
+            try:
+                line.certificate_type = line.product.attr.certificate_type.capitalize()
+            except AttributeError as e:
+                line.certificate_type = None
+                logger.exception('Product has no certificate type. [%s]', e)
+
+            try:
+                course = api.courses(course_id).get()
```
clintonb: This should be cached.

ecommerce/extensions/basket/views.py
clintonb: Is the second return value of `get_voucher_from_code()` ever used? If not, it should be removed.

ecommerce/coupons/views.py
```diff
@@ -29,7 +31,7 @@
 Voucher = get_model('voucher', 'Voucher')
 
 
-def get_voucher(code):
+def get_voucher_from_code(code):
     """
     Returns a voucher and prodcut for a given code.
```
clintonb: Product is misspelled.

ecommerce/extensions/basket/views.py
clintonb: Actually, looking at `get_voucher_from_code()`, the second return value should definitely be removed as it makes the assumption that vouchers are linked to one product, which is not true.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -0,0 +1,57 @@
+from django.test import RequestFactory
+from oscar.core.loading import get_model
+from oscar.test.factories import (ConditionFactory, ConditionalOfferFactory,
+                                  BenefitFactory, ProductFactory, RangeFactory, VoucherFactory)
+
+from ecommerce.extensions.basket.utils import prepare_basket
+from ecommerce.tests.factories import SiteConfigurationFactory
+from ecommerce.tests.testcases import TestCase
+from ecommerce.extensions.partner.models import StockRecord
```
clintonb: Alphabetize.

ecommerce/coupons/tests/test_views.py
```diff
@@ -30,6 +28,26 @@
 Voucher = get_model('voucher', 'Voucher')
 VoucherApplication = get_model('voucher', 'VoucherApplication')
 
+COUPON_CODE = 'COUPONTEST'
+
+
+def prepare_voucher(range_=None, start_datetime=None, benefit_value=100, benefit_type='Percentage'):
```
clintonb: Rename this to something that doesn't conflict with the other `prepare_voucher` function (e.g. `create_voucher`).

clintonb: If this needs to be shared by multiple tests, move it to a mixin.

ecommerce/coupons/tests/test_views.py
```diff
@@ -57,58 +75,17 @@ def test_staff_user_required(self):
...
-            "name": "edX Demonstration Course",
-        }
-        course_info_json = json.dumps(course_info)
-        course_url = get_lms_url('api/courses/v1/courses/{}/'.format(course.id))
-        httpretty.register_uri(httpretty.GET, course_url, body=course_info_json, content_type='application/json')
-        return range_
+    def test_get_voucher_from_code(self):
+        """ Verify that get_voucher_from_code() returns product and voucher. """
+        prepare_voucher()
+        voucher, product = get_voucher_from_code(code=COUPON_CODE)
```
clintonb, line 65: There are no assertions in this test. There should be at least one assertion.

ecommerce/coupons/tests/test_views.py
```diff
@@ -169,6 +147,43 @@ def test_used_voucher(self):
...
+    def setUp(self):
+        super(CouponOfferViewTests, self).setUp()
+        footer_url = get_lms_url('api/branding/v1/footer')
+        footer_content = {
+            'footer': 'edX Footer'
+        }
+        content_json = json.dumps(footer_content)
+        httpretty.register_uri(httpretty.GET, footer_url, body=content_json, content_type='application/json')
+
+    def prepare_course_information(self):
```
clintonb, line 142: This method does too much. Ideally functions/methods should do one thing well. In this case you are generating data and mocking an API endpoint. Break this into two methods.

ecommerce/coupons/tests/test_views.py
```diff
@@ -169,6 +147,43 @@ def test_used_voucher(self):
...
+        """ Helper function to prepare an API endpoint that provides course information. """
+        course = CourseFactory()
+        seat = course.create_or_update_seat('verified', True, 50, self.partner)
+        sr = StockRecord.objects.get(product=seat)
```
clintonb: Avoid unnecessary abbreviation so that it is immediately clear what the variable refers to. `stock_record`.

ecommerce/coupons/tests/test_views.py
```diff
@@ -169,6 +147,43 @@ def test_used_voucher(self):
...
+        catalog = Catalog.objects.create(name='Test catalog', partner=self.partner)
+        catalog.stock_records.add(sr)
+        range_ = RangeFactory(catalog=catalog)
```
clintonb: What's with the trailing underscore?

ecommerce/coupons/tests/test_views.py
```diff
@@ -169,6 +147,43 @@ def test_used_voucher(self):
...
+    path_with_code = '{path}?code={code}'.format(path=path, code=COUPON_CODE)
+
+    def setUp(self):
+        super(CouponOfferViewTests, self).setUp()
+        footer_url = get_lms_url('api/branding/v1/footer')
+        footer_content = {
+            'footer': 'edX Footer'
+        }
+        content_json = json.dumps(footer_content)
+        httpretty.register_uri(httpretty.GET, footer_url, body=content_json, content_type='application/json')
```
clintonb: Move `'application/json'` to a constant shared across the file.

ecommerce/extensions/basket/views.py
mjfrey: We use this to get the product based on a voucher code.  This is how we build the offer landing page with the matching product.  It will eventually be a list.

ecommerce/coupons/tests/test_views.py
```diff
@@ -198,16 +212,14 @@ def test_course_information_error(self):
...
-    @httpretty.activate
     def test_proper_code(self):
         """ Verify that proper information is returned when a valid code is provided. """
         range_ = self.prepare_course_information()
-        self.prepare_voucher(range_=range_)
+        prepare_voucher(range_=range_)
         response = self.client.get(self.path_with_code)
         self.assertEqual(response.context['course']['name'], _('edX Demonstration Course'))
-        self.assertEqual(response.context['code'], _('COUPONTEST'))
+        self.assertEqual(response.context['code'], _(COUPON_CODE))
```
clintonb: Coupon codes should not be marked for translation. Unless you are explicitly testing translations, you shouldn't need `ugettext` at all in these tests.

ecommerce/extensions/basket/views.py
clintonb: It would be ideal if it returned a list now. Do the limiting in the view/template. This can be done in a separate PR.

ecommerce/coupons/tests/test_views.py
```diff
@@ -234,6 +246,17 @@ def setUp(self):
         self.catalog = Catalog.objects.create(partner=self.partner)
         self.catalog.stock_records.add(StockRecord.objects.get(product=self.seat))
 
+    def create_and_test_coupon(self):
+        """ Creates enrollment code coupon. """
+        create_coupon(catalog=self.catalog, code=COUPON_CODE)
+        self.assertEqual(Voucher.objects.filter(code=COUPON_CODE).count(), 1)
+
+    def assert_redirect(self, expected_url, target=200):
```
clintonb: This name is confusing alongside `assertRedirects`. Try something that more accurately reflects what the method does (e.g. `assert_redemption_page_redirects`).

ecommerce/coupons/views.py
```diff
@@ -194,54 +196,16 @@ def get(self, request):
                 order_total=order_metadata[AC.KEYS.ORDER_TOTAL],
             )
         else:
-            return render(
-                request,
-                template_name,
-                {'error': _('Basket total not $0, current value = ${basket_price}'.format(
-                    basket_price=basket.total_excl_tax
-                ))}
+            sku = StockRecord.objects.get(product=product).partner_sku
```
clintonb: Don't assume products have a single stock record. Include the partner as well.

clintonb: To clarify, if we ever move to offering multi-locale pricing, a product will have multiple stock records.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -0,0 +1,57 @@
...
+    """ Tests for basket utility functions. """
+
+    def setUp(self):
+        super(BasketUtilsTests, self).setUp()
+        self.request = RequestFactory()
+        self.request.user = self.create_user()
+        site_configuration = SiteConfigurationFactory(partner__name='Tester')
+        self.request.site = site_configuration.site
+
+    def _prepare_basket(self, product, voucher=None):
```
clintonb: Is this a helper function or a test? It seems to be both, which is odd. If you want to test the output of `prepare_basket`, write a test that does so. There is no need to couple the helper method with the test.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -0,0 +1,57 @@
...
+        self.assertEqual(basket.lines.count(), 1)
+        self.assertEqual(basket.lines.first().product, product)
+        return basket
+
+    def test_prepare_basket_with_voucher(self):
+        """ Verify a basket is returned and contains a voucher and the voucher is applied. """
+        # Prepare a product with price of 100 and a voucher with 10% discount for that product.
+        product = ProductFactory(stockrecords__price_excl_tax=100)
+        voucher = VoucherFactory()
+        range_ = RangeFactory(products=[product, ])
```
clintonb: Why the trailing underscore?

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -0,0 +1,57 @@
...
+        benefit = BenefitFactory(type=Benefit.PERCENTAGE, range=range_, value=10)
+        condition = ConditionFactory(value=1, range=range_)
+        offer = ConditionalOfferFactory(benefit=benefit, condition=condition)
+        voucher.offers.add(offer)
+
+        stock_record = StockRecord.objects.get(product=product)
+        self.assertEqual(str(stock_record.price_excl_tax), '100.00')
```
clintonb: Why the string conversion? Just compare `Decimal` values.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+from ecommerce.extensions.payment.tests.processors import DummyProcessor
+from ecommerce.settings import get_lms_url
+from ecommerce.tests.factories import StockRecordFactory
+from ecommerce.tests.testcases import TestCase
+
+Basket = get_model('basket', 'Basket')
+Selector = get_class('partner.strategy', 'Selector')
+
+
+def prepare_course_api_response(course=None):
```
clintonb: `mock_course_api`

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+    """ Helper function to register an API endpoint for the course information. """
+    course_info = {
+        "media": {
+            "course_image": {
+                "uri": "/asset-v1:edX+DemoX+Demo_Course+type@asset+block@images_course_image.jpg"
+            }
+        }
+    }
+    course_info_json = json.dumps(course_info)
+    if not course:
```
clintonb: This seems wrong. The list endpoint does not return the same data as the detail endpoint.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+        course_url = get_lms_url('api/courses/v1/courses/')
+    else:
+        course_url = get_lms_url('api/courses/v1/courses/{}/'.format(course.id))
+    httpretty.register_uri(httpretty.GET, course_url, body=course_info_json, content_type='application/json')
+
+
+def prepare_footer_api_response():
```
clintonb: `mock_footer_api`

clintonb: Move to a mixin.

ecommerce/extensions/basket/tests/test_views.py
clintonb: Move to a mixin.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+        self.assertRedirects(response, expected_url, target_status_code=302)
+
+    def test_missing_sku(self):
+        """ The view should return HTTP 400 if no SKU is provided. """
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.content, 'No SKU provided.')
+
+    def test_missing_product(self):
+        """ The view should return HTTP 400 if SKU with no product is provided. """
```
clintonb: ...if the SKU has no associated product.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+    def setUp(self):
+        super(BasketSingleItemViewTests, self).setUp()
+        self.user = self.create_user()
+        self.client.login(username=self.user.username, password=self.password)
+
+        product = factories.ProductFactory()
+        self.stock_record = StockRecordFactory(product=product, partner=self.partner)
+
+    def test_login_required(self):
+        """ The view should redirect to login page if user is not logged in. """
```
clintonb: ...to the login page if the user is not logged in.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+        """ Verify the view redirects to the checkout payment page. """
+        prepare_footer_api_response()
+        prepare_course_api_response()
+        url = '{path}?sku={sku}'.format(path=self.path, sku=self.stock_record.partner_sku)
+        response = self.client.get(url)
+        expected_url = self.get_full_url(reverse('basket:summary'))
+        self.assertRedirects(response, expected_url, status_code=303)
+
+    @httpretty.activate
+    def test_redirect(self):
```
clintonb: This test is redundant with the one below. Remove it, and update the name/docstring of `test_basket` to better reflect what it does.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+        """ The view should redirect the user to the payment page. """
+        self.assert_view_redirects_to_checkout_payment()
+
+    @httpretty.activate
+    def test_basket(self):
+        """ The user's latest Basket should contain one instance of the specified product and be frozen. """
```
clintonb: Verify the view redirects to the checkout page, and the user's basket includes only the requested product.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+        self.assert_view_redirects_to_checkout_payment()
+
+        basket = Basket.get_basket(self.user, self.site)
+        self.assertEqual(basket.status, Basket.OPEN)
```
clintonb: Should this be frozen?

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+    """ BasketSummaryView basket view tests. """
+    path = reverse('basket:summary')
+
+    def setUp(self):
+        super(BasketSummaryViewTests, self).setUp()
+        self.user = self.create_user()
+        self.client.login(username=self.user.username, password=self.password)
+        self.course = CourseFactory()
+
+    def _prepare_basket(self, product):
```
clintonb: As above, this needs to either be a helper method or a test method, not both.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+        """ Helper function for creating and adding a product to a basket. """
+        basket = factories.BasketFactory(owner=self.user)
+        basket.add_product(product, 1)
+        self.assertEqual(basket.lines.count(), 1)
+        return basket
+
+    @httpretty.activate
+    def test_connection_to_course_error(self):
+        """ Verify a connection error is logged when a connection error happens. """
```
clintonb: What about the response seen by the user? You need to check the response status and content.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+    @httpretty.activate
+    @override_settings(PAYMENT_PROCESSORS=['ecommerce.extensions.payment.tests.processors.DummyProcessor'])
+    def test_response_success(self):
+        """ Verify a successful response is returned. """
+        seat = self.course.create_or_update_seat('verified', True, 50, self.partner)
+        self._prepare_basket(seat)
+        prepare_course_api_response(self.course)
+        prepare_footer_api_response()
+
+        switch, __ = Switch.objects.get_or_create(name=settings.PAYMENT_PROCESSOR_SWITCH_PREFIX + DummyProcessor.NAME)
```
clintonb: Use `ecommerce.core.tests.toggle_switch()`.

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,32 @@
+import logging
+
+from oscar.core.loading import get_class, get_model
+
+Applicator = get_class('offer.utils', 'Applicator')
+Basket = get_model('basket', 'Basket')
+logger = logging.getLogger(__name__)
+
+
+def prepare_basket(request, product, voucher=None):
```
clintonb, line 11: Testing of this would be much simpler if you split this in half. Retrieve the basket and add the product separately from the voucher application.

clintonb:
I've completed an initial review. Here's what's necessary to get this merged:
- [x] Address comments from initial review
- [x] Review from @bderusha 
- [x] Another round of review from @clintonb and @bderusha, if necessary
- [x] Commits squashed into 1-2 logical commits

ecommerce/extensions/basket/views.py
mjfrey: This tells the UI what type of course.  Verified or professional so we can put the appropriate text in the checkout page.

clintonb: Please add a comment stating that fact. Also, there is no need to capitalize. All certificate types are lower-cased.

ecommerce/extensions/basket/views.py
clintonb: Based on the comments above, this log statement should either be downgraded to debug or removed. The odds of a course seat properly setup via CAT not having an associated certificate type are pretty close to zero.

ecommerce/coupons/tests/test_views.py
vkaracic: It's to differentiate between the range object and Python's range keyword.

ecommerce/coupons/tests/test_views.py
vkaracic, line 65: I don't understand what you mean here. Here's what I'm seeing:
https://github.com/edx/ecommerce/blob/78d913bfef262ca2f97e8aa7f5a7c62982f370bf/ecommerce/coupons/tests/test_views.py#L88-L91

ecommerce/extensions/basket/tests/test_utils.py
vkaracic: This way `Decimal` doesn't have to be imported.

ecommerce/coupons/tests/test_views.py
```diff
@@ -169,6 +129,28 @@ def test_used_voucher(self):
...
+        self.mock_footer_api_response()
+
+    def prepare_course_information(self):
+        """ Helper function to prepare an API endpoint that provides course information. """
+        course = CourseFactory()
+        seat = course.create_or_update_seat('verified', True, 50, self.partner)
+        stock_record = StockRecord.objects.get(product=seat)
+        catalog = Catalog.objects.create(name='Test catalog', partner=self.partner)
+        catalog.stock_records.add(stock_record)
+        range_ = RangeFactory(catalog=catalog)
```
clintonb: Please use "range". The trailing underscore is confusing. We have not had issues in the past differentiating between objects and variables.

clintonb:
@vkaracic it seems your pushes have confused GitHub. Please ensure that all previous comments are addressed. If the changes have been made, reply to that affect.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,143 @@
+import json
+
+import datetime
+import httpretty
```
bderusha: NIT: this should be with the 3rd party packages below and `datetime` should be with `json`

ecommerce/coupons/tests/test_views.py
mjfrey: I already changed the var name.  Range is a python reserved word not a method name.

vkaracic: That's alright. While I'm not _too_ comfortable using it, Oscar uses it as well it seems so I guess it's OK.

clintonb: Thanks for further clarifying @mjfrey. I forgot about the `range` keyword. We generally tend to use prefixed underscores (e.g. `_range`) to avoid these issues.

vkaracic: Now I'm confused. @clintonb I thought this was your answer to: https://github.com/edx/ecommerce/pull/534#discussion_r52283133
What's the final decision then, `_range` or `range` ?

clintonb: `_range`

ecommerce/extensions/basket/app.py
```diff
@@ -0,0 +1,23 @@
...
+            url(r'^add/(?P<pk>\d+)/$', self.add_view.as_view(), name='add'),
+            url(r'^vouchers/add/$', self.add_voucher_view.as_view(), name='vouchers-add'),
+            url(r'^vouchers/(?P<pk>\d+)/remove/$', self.remove_voucher_view.as_view(), name='vouchers-remove'),
+            url(r'^saved/$', login_required(self.saved_view.as_view()), name='saved'),
+            url(r'^single-item/$', login_required(self.single_item_view.as_view()), name='single-item'),
+        ]
+        return self.post_process_urls(urls)
+
+
+application = BasketApplication()
```
bderusha, line 23: Question about these views.  Are these urls our design or are we just piggybacking on oscar here? I feel like there could be more continuity between the names of these endpoints.  If it's consistent with what Oscar does then we should use it, if this is of our own creation I think we can iterate on this a little and come up with a more consistent endpoint name pattern

clintonb, line 23: single-item is the only new URL. The rest come from Oscar.

bderusha, line 23: Answered my own question
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/basket/app.py

This is fine as is.

::EDIT::
Ah thanks, @clintonb.  I missed that `single-item` was different.  Commenting in-line.

ecommerce/extensions/basket/app.py
```diff
@@ -0,0 +1,23 @@
...
+    summary_view = get_class('basket.views', 'BasketSummaryView')
+
+    def get_urls(self):
+        urls = [
+            url(r'^$', self.summary_view.as_view(), name='summary'),
+            url(r'^add/(?P<pk>\d+)/$', self.add_view.as_view(), name='add'),
+            url(r'^vouchers/add/$', self.add_voucher_view.as_view(), name='vouchers-add'),
+            url(r'^vouchers/(?P<pk>\d+)/remove/$', self.remove_voucher_view.as_view(), name='vouchers-remove'),
+            url(r'^saved/$', login_required(self.saved_view.as_view()), name='saved'),
+            url(r'^single-item/$', login_required(self.single_item_view.as_view()), name='single-item'),
```
bderusha, line 18: Given that this view doesn't actually display a single basket item, can give it a more appropriate name.  It seems to me that this view takes a sku and optional voucher and applies them to a basket before forwarding you to the summary page.  

IMO this should be a `POST` request and not a `GET` request. I don't know enough about the guts of how this is being called to know how drastic of a change that is.

ecommerce/extensions/basket/app.py
```diff
@@ -0,0 +1,23 @@
+from django.conf.urls import url
+from django.contrib.auth.decorators import login_required
+from oscar.apps.basket import app
+from oscar.core.loading import get_class
+
+
+class BasketApplication(app.BasketApplication):
+    single_item_view = get_class('basket.views', 'BasketSingleItemView')
```
bderusha, line 8: Per my comment below we may want to update the name of this variable and class as well

bderusha:
Done with my review.  Looks good except for my question about that `single-item` endpoint.

ecommerce/extensions/basket/app.py
mjfrey, line 18: @bderusha  We certainly could change this to a POST.  Right now from the LMS side we build href links to this endpoint within the course_mode view based on the mode.  We would need to change those to trigger a form submit.  

How strongly do you feel about making this change?

clintonb, line 18: Changing this from `GET` to `POST` isn't necessary to merge this.

mjfrey:
@clintonb I believe all comments have been addressed.

ecommerce/extensions/basket/views.py
vkaracic, line 112: This TODO comment was left because we are working on one of your commits: https://github.com/edx/ecommerce/commit/7fa942922a25f709b6f2fc8662ea89f1035d3657#diff-c5add78f78414380f98a613743711c07R36
That being said, I'm not sure how exactly do you plan on implementing this since payment processors from `SiteConfiguration` class is just a CharField where you can put just about anything.
I've implemented a rudimentary mechanics but I don't feel this is optimal:
https://github.com/edx/ecommerce/commit/ebd623c8531fd7891b393a34ac52363659671389#diff-398e80deb2588fea52e2551897aac3e2R98

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,111 @@
...
+            'payment_processors': self.get_payment_processors(self.request),
+            'homepage_url': get_lms_url(''),
+            'footer': get_lms_footer(),
+            'lines': lines,
+        })
+        return context
+
+    def get_payment_processors(self, request):
+        """ Retrieve the list of active payment processors. """
+        # TODO Retrieve this information from SiteConfiguration
```
clintonb, line 112: If this isn't going to be done, please remove the comment.

vkaracic, line 112: https://github.com/edx/ecommerce/pull/534#discussion_r52434544

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,111 @@
+from __future__ import unicode_literals
+import hashlib
+import logging
+
+from django.conf import settings
+from django.http import HttpResponseRedirect, HttpResponseBadRequest
```
clintonb: Alphabetize.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,111 @@
...
+        if not purchase_info.availability.is_available_to_buy:
+            return HttpResponseBadRequest(_('Product [{product}] not available to buy.'.format(product=product.title)))
+
+        prepare_basket(request, product, voucher)
+        return HttpResponseRedirect(reverse('basket:summary'), status=303)
+
+
+class BasketSummaryView(BasketView):
+    """
+    View to create a context that shows the user's basket contents
```
clintonb: Display basket contents and checkout/payment options.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,111 @@
...
+        for line in lines:
+            course_id = line.product.course_id
+
+            # Get each course type so we can display to the user at checkout.
+            try:
+                line.certificate_type = line.product.attr.certificate_type
+            except AttributeError:
+                line.certificate_type = None
+
+            course_hash = hashlib.md5(course_id).hexdigest()
```
clintonb: Your hash key needs to be more specific to avoid collisions: 

``` python
cache_key = 'courses_api_detail_{}'.format(course_id)
cache_key = hashlib.md5(cache_key).hexdigest()
```

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,111 @@
...
+            try:
+                course = cache.get(course_hash)
+                if not course:
+                    course = api.courses(course_id).get()
+                    course['image_url'] = get_lms_url(course['media']['course_image']['uri'])
+                    # cache the course for an hour
+                    cache.set(course_hash, course, 3600)
```
clintonb: The timeout value should be a setting, `COURSES_API_CACHE_TIMEOUT`. The value can default to 3600 in `settings/base.py`.

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,111 @@
...
+                line.course = course
+            except SlumberHttpBaseException:
+                logger.exception('Could not get course information. [%s]', course_id)
```
clintonb: Failed to retrieve data from Course API for course [%s].

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,111 @@
...
+
+        applied_offers = self.request.basket.applied_offers()
+        benefit = applied_offers.values()[0].benefit if applied_offers else None
```
clintonb: Why is only the first selected? What if there are multiple offers?

ecommerce/tests/factories.py
```diff
@@ -24,3 +25,8 @@ class Meta(object):
 
     site = factory.SubFactory(SiteFactory)
     partner = factory.SubFactory(PartnerFactory)
+
+
+class OttoStockRecordFactory(StockRecordFactory):
```
clintonb: Use `StockRecordFactory` to maintain consistency. Import the other one as `OscarStockRecordFactory`.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+        self.mock_footer_api_response()
+        seat = self.course.create_or_update_seat('verified', True, 50, self.partner)
+        basket = factories.BasketFactory(owner=self.user)
+        basket.add_product(seat, 1)
+        self.assertEqual(basket.lines.count(), 1)
+        course_url = get_lms_url('api/courses/v1/courses/')
+        logger_name = 'ecommerce.extensions.basket.views'
+        with self.assertRaises(ConnectionError):
+            with LogCapture(logger_name) as l:
+                self.client.get(self.path)
```
clintonb: I commented on this earlier, but the comment was not acknowledged. What does the user see when this occurs? The response should be tested to ensure the correct user experience.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+    @override_settings(PAYMENT_PROCESSORS=['ecommerce.extensions.payment.tests.processors.DummyProcessor'])
+    def test_response_success(self):
+        """ Verify a successful response is returned. """
+        seat = self.course.create_or_update_seat('verified', True, 50, self.partner)
+        basket = factories.BasketFactory(owner=self.user)
+        basket.add_product(seat, 1)
+        self.assertEqual(basket.lines.count(), 1)
+        self.mock_course_api_response(self.course)
+        self.mock_footer_api_response()
+        self.site.siteconfiguration.payment_processors = DummyProcessor.NAME
```
clintonb: This is not necessary given that `SiteConfiguration.payment_processors` is not used.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.context['benefit'], None)
+        self.assertEqual(response.context['payment_processors'][0].NAME, DummyProcessor.NAME)
+        self.assertEqual(json.loads(response.context['footer']), {'footer': 'edX Footer'})
+
+    @httpretty.activate
+    @override_settings(PAYMENT_PROCESSORS=['ecommerce.extensions.payment.tests.processors.DummyProcessor'])
+    def test_cached_course(self):
+        """ Verify a successful response is returned. """
```
clintonb: Either this docstring or the test name is wrong. Either way, they should match.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+        seat = self.course.create_or_update_seat('verified', True, 50, self.partner)
+        basket = factories.BasketFactory(owner=self.user)
+        basket.add_product(seat, 1)
+        self.assertEqual(basket.lines.count(), 1)
+        self.mock_course_api_response(self.course)
+        self.mock_footer_api_response()
+        self.site.siteconfiguration.payment_processors = DummyProcessor.NAME
```
clintonb: Remove.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+                    (
+                        logger_name, 'ERROR',
+                        u'Could not get course information. [Client Error 404: {course_url}{course_id}/]'.format(
+                            course_url=course_url, course_id=self.course.id
+                        )
+                    )
+                )
+
+    @httpretty.activate
+    @override_settings(PAYMENT_PROCESSORS=['ecommerce.extensions.payment.tests.processors.DummyProcessor'])
```
clintonb: You can avoid duplication by moving this to decorate the class. Move the `toggle_switch` call to `setUp()`.

ecommerce/extensions/basket/tests/test_views.py
clintonb: Actually, now that I'm seeing a couple lines above, the user experience is quite poor. You are essentially preventing users from checking out because we failed to connect to an API. The view/template should be setup to gracefully degrade (e.g. don't show an image, various course details) when we fail to get API data. Raising an error, and showing the user an error page, is a good way to lose users.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+    path = reverse('basket:summary')
+
+    def setUp(self):
+        super(BasketSummaryViewTests, self).setUp()
+        self.user = self.create_user()
+        self.client.login(username=self.user.username, password=self.password)
+        self.course = CourseFactory()
+
+    @httpretty.activate
+    def test_connection_to_course_error(self):
```
clintonb: We rarely see `ConnectionError` in production. Rename this to `test_course_api_failure`, and use `ddt` to test for `SlumberBaseException` and `Timeout` in addition to `ConnectionError`.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+        expected_content = 'SKU [{}] does not exist.'.format(sku)
+        url = '{path}?sku={sku}'.format(path=self.path, sku=sku)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.content, expected_content)
+
+    def test_unavailable_product(self):
+        """ The view should return HTTP 400 if the product is not available for purchase. """
+        product = self.stock_record.product
+        product.expires = pytz.utc.localize(datetime.datetime.min)
```
clintonb, line 77: You don't need need to localize the beginning of time.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+        self.assertFalse(Selector().strategy().fetch_for_product(product).availability.is_available_to_buy)
+
+        expected_content = 'Product [{}] not available to buy.'.format(product.title)
+        url = '{path}?sku={sku}'.format(path=self.path, sku=self.stock_record.partner_sku)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.content, expected_content)
+
+    @httpretty.activate
+    def test_basket(self):
```
clintonb: `test_redirect_to_basket_summary`

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Verify redirect to basket summary and that the Basket contains
```
clintonb: Verify the view redirects to the basket summary page, and that the user's basket is prepared for checkout.

ecommerce/extensions/basket/tests/test_views.py
clintonb: Either update this test, or add a new one, to test the addition of a voucher to the basket.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,161 @@
+import datetime
+import json
+
+from django.conf import settings
+from django.core.urlresolvers import reverse
+from django.test import override_settings
+
```
clintonb: Remove this newline.

ecommerce/extensions/test/factories.py
```diff
@@ -73,3 +77,22 @@ def create_coupon(title='Test coupon', price=100, partner=None, catalog=None, co
         data=data
     )
     return coupon
+
+
+def prepare_voucher(code='COUPONTEST', _range=None, start_datetime=None, benefit_value=100, benefit_type='Percentage'):
```
clintonb: Replace the hardcoded `'Percentage'` with `Benefit.PERCENTAGE`. We should generally avoid hardcoding values.

ecommerce/coupons/tests/test_views.py
```diff
@@ -57,58 +57,17 @@ def test_staff_user_required(self):
...
-        super(CouponOfferViewTests, self).setUp()
-
-    def prepare_voucher(self, range_=None, start_datetime=None, benefit_value=100, benefit_type='Percentage'):
-        """ Create a voucher and add an offer to it that contains a created product. """
-        if range_ is None:
-            product = ProductFactory(title='Test product')
-            range_ = RangeFactory(products=[product, ])
-        else:
-            product = range_.all_products()[0]
+        super(GetVoucherTests, self).setUp()
```
clintonb: There is no need for this override.

ecommerce/coupons/tests/test_views.py
```diff
@@ -57,58 +57,17 @@ def test_staff_user_required(self):
...
 
-    def test_get_voucher(self):
-        """ Verify that get_voucher() returns product and voucher. """
-        self.prepare_voucher()
-        voucher, product = get_voucher(code='COUPONTEST')
         self.assertIsNotNone(voucher)
-        self.assertEqual(voucher.code, 'COUPONTEST')
+        self.assertEqual(voucher.code, COUPON_CODE)
         self.assertIsNotNone(product)
         self.assertEqual(product.title, 'Test product')
```
clintonb: Test that the product is the actual object you expect, rather than checking attributes.

ecommerce/coupons/views.py
```diff
@@ -29,9 +31,9 @@
 Voucher = get_model('voucher', 'Voucher')
 
 
-def get_voucher(code):
+def get_voucher_from_code(code):
```
clintonb, line 34: Please create a ticket to refactor this function. We should raise an exception in exceptional cases (e.g. the voucher does no exist, or the voucher has no associated products). Using special return values (such as `None`) is prone to errors, especially when those special values are not documented (which they aren't in this case).

ecommerce/extensions/basket/views.py
mjfrey: We currently only support a single offer per basket.

clintonb:
I've added more comments. Here's what's left:
- [x] Address comments
- [x] Create ticket to refactor `get_voucher_from_code`
- [x] Squash commits

ecommerce/extensions/basket/tests/test_views.py
mjfrey: We do check in the response that processor is set correctly.

ecommerce/extensions/basket/views.py
clintonb, line 112: If you're going to move forward with this, please do the following:
1. Remove the TODO statement, because there is nothing left to do.
2. File a ticket to (a) update documentation, and (b) ensure these values are setup properly for existing sites. (Any staff user can view/modify the configuration via Django admin.)

ecommerce/extensions/basket/tests/test_views.py
mjfrey, line 77: Need this or tests fail.
OverflowError: date value out of range

ecommerce/extensions/basket/views.py
mjfrey, line 112: I think we will stick to getting the payment processors via settings for now.

ecommerce/coupons/views.py
mjfrey, line 34: Created ticket to re-factor:
https://openedx.atlassian.net/browse/SOL-1636

ecommerce/extensions/basket/tests/test_views.py
mjfrey: We do not show the user an error page and the course information is just blank.  They can still purchase the product.

clintonb: That's not what your tests say. `with self.assertRaises(ConnectionError)` implies that you are expected `self.client.get(self.path)` to raise an error. If such an error is raised, the user will see an error page.

mjfrey: Well the test is wrong.  We do not raise an exception in the summary view.  The context for the course info will just be blank.
@vkaracic Please update this test. Or I wonder if this test is needed at all

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -0,0 +1,166 @@
...
+        super(BasketSummaryViewTests, self).setUp()
+        self.user = self.create_user()
+        self.client.login(username=self.user.username, password=self.password)
+        self.course = CourseFactory()
+        self.site.siteconfiguration.payment_processors = DummyProcessor.NAME
+        self.site.siteconfiguration.save()
+        toggle_switch(settings.PAYMENT_PROCESSOR_SWITCH_PREFIX + DummyProcessor.NAME, True)
+
+    @httpretty.activate
+    def test_connection_to_course_error(self):
```
clintonb: As previously commented: We rarely see ConnectionError in production. Rename this to test_course_api_failure, and use `ddt` to test for SlumberBaseException and Timeout in addition to ConnectionError.

ecommerce/extensions/basket/tests/test_views.py
clintonb: This test is very much needed. Given that it is passing, your view is not properly handling these exceptions. At present you only account for `SlumberBaseException`. You also need to account for `Timeout`.

mjfrey:
@clintonb ok, think we are done.

ecommerce/core/tests/test_context_processors.py
```diff
@@ -4,10 +4,12 @@
 from ecommerce.tests.testcases import TestCase
 
 PLATFORM_NAME = 'Test Platform'
+SUPPORT_URL = 'example.com'
 
 
 class CoreContextProcessorTests(TestCase):
     @override_settings(PLATFORM_NAME=PLATFORM_NAME)
+    @override_settings(SUPPORT_URL=SUPPORT_URL)
```
clintonb: Combine with the line above: `@override_settings(PLATFORM_NAME=PLATFORM_NAME, SUPPORT_URL=SUPPORT_URL)`

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,117 @@
...
+        context = super(BasketSummaryView, self).get_context_data(**kwargs)
+        lines = self.request.basket.lines.all()
+        api = EdxRestApiClient(get_lms_url('api/courses/v1/'))
+        for line in lines:
+            course_id = line.product.course_id
+
+            # Get each course type so we can display to the user at checkout.
+            try:
+                line.certificate_type = _(
+                    '{certificate_type}'.format(certificate_type=line.product.attr.certificate_type.capitalize())
```
clintonb: Why are you calling `.format()` here? Also, note that this will not be translated as expected. When we pull strings for translation, there is no context (e.g. basket, line items). In the interest of time, you should file a ticket and immediately fix this before going to production.

You'll most likely want something like this:

``` python
# ecommerce/extensions/basket/views.py
line.certificate_type = get_certificate_type_display_value(line.product.attr.certificate_type)

# someplace else
def get_certificate_type_display_value(certificate_type):
  display_values = {
    'audit': _('Audit'),
    'verified': _('Verified'),
    ...
  }

  # TODO Account for cases with an unknown certificate type. Should never occur, but defensive programming FTW!

  return display_values[certificate_type]
```

ecommerce/extensions/basket/views.py
```diff
@@ -0,0 +1,117 @@
...
+                line.certificate_type = None
+
+            cache_key = 'courses_api_detail_{}'.format(course_id)
+            cache_hash = hashlib.md5(cache_key).hexdigest()
+            try:
+                course = cache.get(cache_hash)
+                if not course:
+                    course = api.courses(course_id).get()
+                    course['image_url'] = get_lms_url(course['media']['course_image']['uri'])
+                    # cache the course for COURSES_API_CACHE_TIMEOUT time
```
clintonb: Unnecessary comment is unnecessary.

ecommerce/extensions/basket/views.py
clintonb, line 112: If you're going to add a `TODO`, it needs to have a ticket associated with it.

ecommerce/tests/factories.py
```diff
@@ -24,3 +25,8 @@ class Meta(object):
 
     site = factory.SubFactory(SiteFactory)
     partner = factory.SubFactory(PartnerFactory)
+
+
+class StockRecordFactory(OscarStockRecordFactory):
+    product = factory.SubFactory(ProductFactory)
+    price_currency = "USD"
```
clintonb: Nit: Maintain consistency of quotes in the file. Use single quotes.

clintonb:
:+1: pending:
- [x] Remaining comments addressed
- [x] Commits squashed

ecommerce/extensions/basket/views.py
mjfrey: @clintonb please take a look.  I did this now.

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,47 @@
...
+    basket.freeze()
+
+    return basket
+
+
+def get_certificate_type_display_value(certificate_type):
+    display_values = {
+        'audit': _('Audit'),
+        'verified': _('Verified'),
+        'professional': _('Professional')
```
clintonb: The honor type is missing.

ecommerce/extensions/basket/utils.py
```diff
@@ -0,0 +1,47 @@
...
+    if voucher:
+        basket.vouchers.add(voucher)
+        Applicator().apply(basket, request.user, request)
+        logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
+    basket.freeze()
+
+    return basket
+
+
+def get_certificate_type_display_value(certificate_type):
```
clintonb, line 37: This needs tests.

mjfrey, line 37: ok tests added.

PR #536, opened 2016-02-05
CheckoutView API endpoint

mjfrey:
Checkout API endpoint

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -422,3 +422,12 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     class Meta(object):
         model = Product
         fields = ('id', 'title', 'coupon_type', 'last_edited', 'seats', 'client', 'price', 'vouchers',)
+
+
+class CheckoutSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    payment_form_data = serializers.SerializerMethodField()
+    payment_page_url = serializers.URLField()
+    payment_processor = serializers.CharField()
+
+    def get_payment_form_data(self, obj):
+        return obj['payment_form_data']
--- a/ecommerce/extensions/api/v2/tests/views/test_checkout.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_checkout.py
@@ -1,0 +1,80 @@
+from __future__ import unicode_literals
+
+import json
+
+from django.conf import settings
+from django.core.urlresolvers import reverse
+from django.test import override_settings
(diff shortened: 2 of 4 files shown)


mjfrey:
@clintonb @bderusha This is the new PR.  I messed up the old one.  Working on the JS test now.

clintonb:
Given that the JavaScript file is not used, I would be fine if you merged without those files. They still need to be tested; but, once removed, they are no longer blocking this from being merged.

mjfrey:
@clintonb added JS tests.

ecommerce/static/js/test/specs/pages/checkout_payment_spec.js
```diff
@@ -0,0 +1,51 @@
...
+                it('should fill form inputs for each data key/value pair', function () {
+                    var data = {
+                        payment_page_url: 'http://www.dummy-url.com/',
+                        type: 'Dummy Type',
+                        model: '500',
+                        color: 'white',
+                    };
+                    spyOn(_, 'each');
+                    CheckoutPayment.onDone(data);
+                    expect(_.each.calls.count()).toEqual(1);
```
clintonb: The behavior we care most about for this function is the form submission. Counting the calls to `_.each()` doesn't validate the functionality we care most about.

ecommerce/static/js/test/specs/pages/checkout_payment_spec.js
```diff
@@ -0,0 +1,51 @@
...
+            });
+
+            describe('onReady', function () {
+                it('should disable payment button before making ajax call', function () {
+                    $('<div class="payment-buttons"><button class="payment-button">Pay</button></div>')
+                        .appendTo('body');
+                    spyOn(Utils, 'disableElementWhileRunning');
+                    CheckoutPayment.onReady();
+                    $('button.payment-button').trigger('click');
+                    expect(Utils.disableElementWhileRunning).toHaveBeenCalled();
```
clintonb: As above, disabling the button is not what we care most about. We want to make sure the function issues an AJAX request.

clintonb:
These JavaScript tests are not sufficient as they don't actually test checkout initiation. As I mentioned earlier, given that this functionality is not needed by this PR, I recommend removing it. We can review the code again in a fronted PR that actually makes use of the JavaScript.

mjfrey:
Moved JS out of this PR and into the basket PR.

clintonb:
:+1: pending:
- [x] All JS files removed (If the code isn't being used, it doesn't need to be in the PR/repository.)
- [x] Commits squashed into single commit

PR #537, opened 2016-02-08
Basket UI

vkaracic:
UI That implements a Basket Summary and adds the ability to enter a coupon code to apply an offer to the basket.  

The basket makes use of the basket endpoint and also uses the checkout endpoint to initiate payment.

screen shot 2016-02-05 at 12 26 33 pm


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=537

marjev:
@clintonb, @bderusha could you please review the code changes? This PR includes all of the basket UI changes.

Thanks!

bderusha:
LGTM :+1: 

@clintonb ?

ecommerce/static/js/test/specs/pages/basket_spec.js
```diff
@@ -0,0 +1,55 @@
...
+              Utils,
+              Basket) {
+        'use strict';
+
+        describe('Basket Page', function () {
+            beforeEach(function () {
+                $('<div id="voucher_form_container"><input id="id_code">' +
+                    '<a id="voucher_form_cancel"></a></button></div>'
+                ).appendTo('body');
+                $('<div id="voucher_form_link"><a href=""></a></div>').appendTo('body');
```
clintonb: Any reason not to combine this with the call above?

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -0,0 +1,46 @@
+{% extends 'edx/base.html' %}
+{% load thumbnail %}
```
clintonb: This doesn't appear to be used anywhere in this file.

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -0,0 +1,46 @@
...
+{% load i18n %}
+{% load humanize %}
```
clintonb: This doesn't appear to be used anywhere in this file.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,144 @@
+{% load i18n %}
+{% load thumbnail %}
```
clintonb: This doesn't appear to be used anywhere in this file. Only load what is needed.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,144 @@
...
+                                <div class="col-sm-3 discount">
+                                    <div class="benefit">
+                                        {% if benefit.type == 'Percentage' and benefit.value < 100.00 %}
+                                            {% blocktrans with percentage=benefit.value|stringformat:'d' %}
+                                                {{percentage}}% off
+                                            {% endblocktrans %}
+                                        {% endif %}
+                                    </div>
+                                    <div class="old-price">
+                                        {% if line.is_tax_known %}
```
clintonb: All prices should be displayed with tax. We may not collect tax today, but that may change in the future. When it does, we want to avoid having to change every single place where we display prices.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,144 @@
...
+                                    <img class="thumbnail" src="{{ line.course.image_url }}" alt="{{ line.course.name }}"/>
+                            </div>
+                            <div class="col-sm-5">
+                                <p class="course_name">{{ line.course.name }} - {{ line.course.org }} ({{ line.course.number }})</p>
+                                <p class="course_description">{{ line.course.short_description }}</p>
+                            </div>
+                            {% if line.has_discount %}
+                                <div class="col-sm-3 discount">
+                                    <div class="benefit">
+                                        {% if benefit.type == 'Percentage' and benefit.value < 100.00 %}
```
clintonb: Don't hardcode the benefit type and values in the template. As I pointed out on #534, the assumption that there is a single offer is not a good one. If you insist on doing this, at least move this logic to the view instead of the template.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,144 @@
...
+                {% csrf_token %}
+                {{ formset.management_form }}
+
+                <div class="basket-items">
+                    {% for line in lines %}
+                        {% purchase_info_for_line request line as session %}
+                        {% if line.certificate_type %}
+                        <p class="certificate_type">
+                            {% blocktrans with certificate_type=line.certificate_type %}
+                                Enroll and Pursue a {{certificate_type}} Certificate in
```
clintonb, line 21: This will be a poor user experience. `certificate_type` is lower-cased and in English.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,144 @@
...
+                                    <div class="old-price">
+                                        {% if line.is_tax_known %}
+                                            {{ line.line_price_incl_tax|currency:line.price_currency }}
+                                        {% else %}
+                                            {{ line.line_price_excl_tax|currency:line.price_currency }}
+                                        {% endif %}
+                                    </div>
+                                </div>
+                            {% endif %}
+                            <div class="price {% if line.has_discount %}green{% endif %}">
```
clintonb: "green" is not a good class name as you tie the style name to the display color. "discounted" more accurately describes the intent.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,144 @@
...
+                            {% trans "Checkout with PayPal" %}
+                        {% endif %}
+                    </button>
+                {% endfor %}
+            </div>
+        </div>
+
+        <div class="row verification-note">
+            <span><i class="fa fa-exclamation-circle"></i></span>
+            <b>{% trans "Note:" %}</b>{% trans " You will need a " %}
```
clintonb: You cannot breakup translations in this manner. The translators will have no context. In this case it is okay to include the markup in the text to be translated. You could also simply remove the markup.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,144 @@
...
+            <b>{% trans "webcam" %}</b>{% trans " and a " %}
+            <b>{% trans "photo ID" %}</b>{% trans " to verify your identity before the course starts. " %}
+        </div>
+        <div class="row help">
+            <p><b>{% trans "Have questions?" %}</b></p>
+            {% blocktrans with link_start='<a href="https://www.edx.org/gfa" target="_blank">' link_end='</a>' %}
```
clintonb: 1. That's the wrong link.
2. Don't hardcode links. Use a setting (e.g. `SUPPORT_URL`).

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,144 @@
...
+                Please read {{ link_start }}our FAQs to view common questions about our certificates.{{ link_end }}
+            {% endblocktrans %}
+        </div>
+    </div>
+{% else %}
+    {% block emptybasket %}
+        <p>
+            {% trans "Your basket is empty." %}
+            <a href="{{ homepage_url }}">{% trans "Continue shopping" %}</a>
```
clintonb: Discover new courses.

ecommerce/templates/oscar/basket/partials/basket_totals.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}
+{% load currency_filters %}
+{% load shipping_tags %}
+
+<div id="basket_totals" class="col-sm-2">
+    {% block order_total %}
+        {% trans "Total:" %}
+        {% if order_total.is_tax_known %}
```
clintonb: Always show the value with tax. That is what we will always use for payment.

ecommerce/templates/oscar/basket/partials/basket_totals.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}
+{% load currency_filters %}
+{% load shipping_tags %}
```
clintonb: This is not used.

ecommerce/templates/oscar/basket/partials/basket_content.html
marjev: What seems to be the issue here? If the tax is known, the price will be displayed with tax included and if it's not, the price will be displayed with tax excluded.

Why would we want to change anything here, if we decide to collect the tax? This is, after all, the way Oscar handles price display.

ecommerce/templates/oscar/basket/partials/basket_content.html
marjev, line 21: I'm not sure I understand your comment here.
Here on [Invision design prototype],(https://projects.invisionapp.com/share/MN51RYKU8#/screens/117681518) `certificate_type` is upper-cased and, as I recall, LMS translates certificate_type.

Could you please explain the issue?

clintonb, line 21: Looking at the underlying view, this value comes directly from `line.product.attr.certificate_type`. That value is lower-cased and untranslated. You should work with @mjfrey and @vkaracic to fix the view before it is merged.

ecommerce/templates/oscar/basket/partials/basket_content.html
clintonb: Use `{{ line.line_price_incl_tax|currency:line.price_currency }}`. That value will always be accurate. If we don't charge tax, the tax amount will be 0. The less branching you have in your code, the fewer things you have that can go wrong and/or need to be tested.

ecommerce/templates/oscar/basket/partials/basket_content.html
marjev, line 21: Ok, it will be resolved in #534.

marjev:
@clintonb, all the comments are addressed. Could you take another look?

ecommerce/templates/oscar/basket/partials/basket_content.html
vkaracic: I couldn't find any support URL in the ecommerce settings. Would an implementation like this suffice: https://github.com/edx/ecommerce/commit/f01d151e523a5cc9b553d6396925503cb714d056 ?

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -0,0 +1,44 @@
+{% extends 'edx/base.html' %}
+{% load i18n %}
+{% load staticfiles %}
+
+
+{# Hide mini-basket so we don't have two baskets to keep in sync #}
+{% block mini_basket %}{% endblock %}
```
clintonb: This block doesn't exist in `edx/base.html`. Was this originally based on Oscar's checkout page?

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -0,0 +1,44 @@
...
+
+{% block title %}
+{% trans "Basket" %}
+{% endblock title %}
+
+
+{% block javascript %}
+    <script src="{% static 'js/pages/basket.js' %}"></script>
```
clintonb: Is there a reason these scripts are separated?

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -0,0 +1,44 @@
...
+{% block content %}
+    <div class="basket">
+        <div class="container">
+            {# Use a partial template so that AJAX can be used to re-render basket #}
+            {% include 'basket/partials/basket_content.html' %}
+        </div>
+    </div>
+{% endblock content %}
+
+{% block onbodyload %}
```
clintonb: I don't think this is ever loaded. This block doesn't exist in `edx/base.html`.

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -0,0 +1,44 @@
...
+    {{ block.super }}
+    {# We pass in the URL to send AJAX requests to #}
+    var options = {
+        'basketURL': '{% url 'basket:summary' %}'
+    };
+    oscar.basket.init(options);
+{% endblock %}
+
+
+{% block footer %}
```
clintonb, line 23: Why not move this to the base template?

ecommerce/templates/oscar/basket/partials/basket_content.html
clintonb: Yes.

ecommerce/templates/oscar/basket/basket.html
marjev: This is c/p'd Oscar basket. I don't know what it was based on. I just got a template page and had to make it look like the Invision design.

I'll remove that since we don't use it.

ecommerce/templates/oscar/basket/basket.html
marjev: basket.js contains JS code specifically written for the basket page. I don't know the purpose behind the other script since I didn't write it.

ecommerce/templates/oscar/basket/basket.html
marjev, line 23: As far as I know, the requirement is that only basket page has the branding footer so that student doesn't feel like he's been sent to another 'page'.

Other Otto pages should override the footer in their own fashion.

marjev:
@clintonb I addressed you latest comment and left a comment on those that I thought are still open for discussion or already resolved.

Could you take a final look and perhaps make a list of things that need to be done before merging? Thanks!

ecommerce/templates/oscar/basket/basket.html
clintonb: See http://open-edx-ecommerce-guide.readthedocs.org/en/latest/asset_pipeline.html. This page needs to be added to `build.js`.

ecommerce/templates/oscar/basket/basket.html
clintonb, line 23: Why is it that these comments get hidden? I have never seen that on other PRs.

ecommerce/static/js/utils/basket_utils.js
```diff
@@ -0,0 +1,99 @@
+/**
```
clintonb: There is no reason for this to be split from `pages/basket.js`. Why was it?

ecommerce/static/js/utils/basket_utils.js
```diff
@@ -0,0 +1,99 @@
...
+            }).appendTo(form);
+        },
+        hideVoucherForm = function() {
+            $('#voucher_form_container').hide();
+            $('#voucher_form_link').show();
+        },
+        onFail = function(){
+            var message = gettext('Problem occurred during checkout. Please contact support');
+            $('#messages').empty().append(
+                '<div class="error">' + message + '</div>'
```
clintonb: You can use [`_s.sprintf`](http://gabceb.github.io/underscore.string.site/#sprintf).

ecommerce/static/js/utils/basket_utils.js
```diff
@@ -0,0 +1,99 @@
...
+                    promise = deferred.promise(),
+                    paymentProcessor = $btn.val(),
+                    data = {
+                        basket_id: basketId,
+                        payment_processor: paymentProcessor
+                    };
+
+                Utils.disableElementWhileRunning($btn, function() { return promise; });
+
+                $.ajax({
```
clintonb: There needs to be a functional test here. See https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/test/specs/models/course_model_spec.js#L242-L282 for an example.

ecommerce/templates/oscar/basket/basket.html
clintonb: Was this addressed?

clintonb:
:+1: pending:
- [x] All previous comments are addressed. This can be as simple as replying with the word "Done". (Normally, I do this before pushing the change/fix to avoid having to hunt through hidden comments.)
- [x] All identified JS issues addressed.

ecommerce/static/js/utils/basket_utils.js
marjev: To be able to write the test for basket_utils functions and be able to reference those in basket HTML file.

clintonb: The functions can still be functions; but, these are not utility functions. They are all specific to the basket page. As such they belong in the page JS file. I don't understand what you mean by `reference those in basket HTML file`. If your HTML has anything other than a reference to load the JS file, something is wrong.

marjev: It's not about the HTML, it's about the JS file. Currently, it's not written like a module. If you reference a file in a test, you can't call the functions within, (at least I didn't find a way to do that).

clintonb: Why not write it as a module or Backbone view?

marjev: Here's the answer to your question: [Checkout payment script written in non-modular fashion and without using BackboneJS](https://github.com/edx/ecommerce/blob/7b0d5a040fb601b1dd4485b3e0d09d2ba06cdca2/ecommerce/static/js/pages/checkout_payment.js)

You should talk to the author of the code.

clintonb: I spoke to the author. He says if you are going to integrate his prototype, you should make it production-worthy.

clintonb:
P.S. We can compromise on the functional test. File a ticket to update the JS and add the test. However, your acceptance tests **MUST** cover complete payment flow without exception.

ecommerce/templates/oscar/basket/basket.html
marjev: Done

ecommerce/static/js/utils/basket_utils.js
marjev: Done

ecommerce/static/js/utils/basket_utils.js
marjev: Done

ecommerce/static/js/utils/basket_utils.js
marjev: Done

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,132 @@
...
+                        {# Hide the entire section if a custom BasketView doesn't pass in a voucher form #}
+                        {% if voucher_form %}
+                            <div class="use-voucher col-sm-6">
+                                <p id="voucher_form_link"><a href="#voucher">{% trans "Apply a promotion code" %}</a></p>
+
+                                <div id="voucher_form_container">
+                                    <form id="voucher_form" action="{% url 'basket:vouchers-add' %}" method="post">
+                                        {% csrf_token %}
+                                        <div class="code">
+                                            <input id="id_code" maxlength="128" name="code" type="text" placeholder="{% trans 'enter promo code' %}">
```
clintonb: `maxlength` isn't necessary.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -0,0 +1,132 @@
...
+                                            <button type="submit" class="btn btn-default" data-loading-text="{% trans 'Adding...' %}">{% trans "Apply" %}</button>
```
clintonb: s/Adding/Applying

ecommerce/templates/oscar/basket/partials/basket_content.html
marjev: Done

ecommerce/templates/oscar/basket/partials/basket_content.html
marjev: Done

marjev:
@clintonb I addressed all the comments and squash commits.

clintonb:
Merge when ready.

PR #541, opened 2016-02-10
[SOL-1631] Add upper limit to percentage benefit value

marjev:
@mjfrey Please take a look. This should solve the https://openedx.atlassian.net/browse/SOL-1631 issue.


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=541&repo=edx%2Fecommerce

mjfrey:
:+1:

PR #542, opened 2016-02-10
Only call fulfill_product if there are supported lines

mjfrey:
Check to see if there are actually supported lines before calling fulfill_product on the fulfillment module.


mjfrey:
@clintonb Fix for ECOM-3680

clintonb:
:+1:

PR #543, opened 2016-02-10
WL-286 Add support for multitenant OAuth

douglashall:
- Pull URL references from SiteConfiguration instead of settings.
- Upgrade edx-auth-backends to 0.2.0 https://github.com/edx/auth-backends/pull/9

Sandbox URLs:
https://ecommerce-multitenant.sandbox.edx.org/
https://ecommerce-hms-multitenant.sandbox.edx.org/
https://ecommerce-mit-multitenant.sandbox.edx.org/

### DIFF
--- a/ecommerce/core/context_processors.py
+++ b/ecommerce/core/context_processors.py
@@ -1,8 +1,11 @@
 from django.conf import settings
 
+from ecommerce.core.url_utils import get_lms_dashboard_url
+
 
 def core(_request):
     return {
+        'lms_dashboard_url': get_lms_dashboard_url(),
+        'platform_name': settings.PLATFORM_NAME,
         'support_url': settings.SUPPORT_URL,
-        'platform_name': settings.PLATFORM_NAME
     }
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -34,6 +34,13 @@ class SiteConfiguration(models.Model):
         null=False,
         blank=False
     )
+    oauth_settings = JSONField(
+        verbose_name=_('OAuth settings'),
+        help_text=_("JSON string containing OAuth backend settings."),
+        null=False,
+        blank=False,
+        default='{}'
+    )
 
     class Meta(object):
         unique_together = ('site', 'partner')
--- a/ecommerce/core/tests/test_views.py
+++ b/ecommerce/core/tests/test_views.py
@@ -24,6 +24,7 @@ class HealthTests(TestCase):
     """Tests of the health endpoint."""
 
     def setUp(self):
+        super(HealthTests, self).setUp()
         self.fake_lms_response = Response()
 
     def test_all_services_available(self, mock_lms_request):
@@ -33,6 +34,8 @@ class HealthTests(TestCase):
 
         self._assert_health(status.HTTP_200_OK, Status.OK, Status.OK, Status.OK)
 
+    @mock.patch('ecommerce.core.views.get_lms_heartbeat_url', mock.Mock(return_value=''))
+    @mock.patch('django.contrib.sites.middleware.get_current_site', mock.Mock(return_value=None))
     @mock.patch('django.db.backends.base.base.BaseDatabaseWrapper.cursor', mock.Mock(side_effect=DatabaseError))
     def test_database_outage(self, mock_lms_request):
         """Test that the endpoint reports when the database is unavailable."""
--- a/ecommerce/core/url_utils.py
+++ b/ecommerce/core/url_utils.py
@@ -1,0 +1,52 @@
+from urlparse import urljoin
+
+from threadlocals.threadlocals import get_current_request
+
+from ecommerce.core.exceptions import MissingRequestError
+
+
+def get_ecommerce_url(path=''):
+    """
+    Returns path joined with the appropriate ecommerce URL root for the current site
+
+    Raises:
+        MissingRequestError: If the current ecommerce site is not in threadlocal storage
+    """
+    request = get_current_request()
+    if request:
+        ecommerce_url_root = "{}://{}".format(request.scheme, request.site.domain)
+        return urljoin(ecommerce_url_root, path)
+    raise MissingRequestError
+
+
+def get_lms_commerce_api_url():
+    return get_lms_url('/api/commerce/v1/')
+
+
+def get_lms_dashboard_url():
+    return get_lms_url('/dashboard')
+
+
+def get_lms_enrollment_api_url():
+    return get_lms_url('/api/enrollment/v1/enrollment')
+
+
+def get_lms_heartbeat_url():
+    return get_lms_url('/heartbeat')
+
+
+def get_lms_url(path=''):
+    """
+    Returns path joined with the appropriate LMS URL root for the current site
+
+    Raises:
+        MissingRequestError: If the current ecommerce site is not in threadlocal storage
+    """
+    request = get_current_request()
+    if request:
+        return urljoin(request.site.siteconfiguration.lms_url_root, path)
+    raise MissingRequestError
+
+
+def get_oauth2_provider_url():
+    return get_lms_url('/oauth2')
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -11,11 +11,11 @@ from oscar.test.factories import OrderFactory, ConditionalOfferFactory, VoucherF
 from oscar.test.utils import RequestFactory
 import pytz
 
+from ecommerce.core.url_utils import get_lms_url, get_lms_enrollment_api_url
 from ecommerce.coupons.views import get_voucher_from_code, voucher_is_valid
 from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.extensions.test.factories import prepare_voucher
-from ecommerce.settings import get_lms_url
 from ecommerce.tests.mixins import CouponMixin, LmsApiMockMixin
 from ecommerce.tests.testcases import TestCase
 
@@ -187,7 +187,7 @@ class CouponOfferViewTests(CourseCatalogTestMixin, LmsApiMockMixin, TestCase):
         response = self.client.get(self.path_with_code)
         response_text = (
             'Could not get course information. '
-            '[Client Error 404: http://127.0.0.1:8000/api/courses/v1/courses/{}/]'
+            '[Client Error 404: http://lms.testserver.fake/api/courses/v1/courses/{}/]'
         ).format(course.id)
         self.assertEqual(response.context['error'], _(response_text))
 
@@ -264,8 +264,8 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, TestCase):
     def test_basket_redirect_enrollment_code(self):
         """ Verify the view redirects to LMS when an enrollment code is provided. """
         self.create_and_test_coupon()
-        httpretty.register_uri(httpretty.POST, settings.ENROLLMENT_API_URL, status=200)
-        self.assert_redemption_page_redirects(settings.LMS_URL_ROOT)
+        httpretty.register_uri(httpretty.POST, get_lms_enrollment_api_url(), status=200)
+        self.assert_redemption_page_redirects(get_lms_url())
 
     @httpretty.activate
     def test_multiple_vouchers(self):
@@ -273,5 +273,5 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, TestCase):
         self.create_and_test_coupon()
         basket = Basket.get_basket(self.user, self.site)
         basket.vouchers.add(Voucher.objects.get(code=COUPON_CODE))
-        httpretty.register_uri(httpretty.POST, settings.ENROLLMENT_API_URL, status=200)
-        self.assert_redemption_page_redirects(settings.LMS_URL_ROOT)
+        httpretty.register_uri(httpretty.POST, get_lms_enrollment_api_url(), status=200)
+        self.assert_redemption_page_redirects(get_lms_url())
--- a/ecommerce/courses/publishers.py
+++ b/ecommerce/courses/publishers.py
@@ -8,8 +8,8 @@ from edx_rest_api_client.client import EdxRestApiClient
 from edx_rest_api_client.exceptions import SlumberHttpBaseException
 import requests
 
+from ecommerce.core.url_utils import get_lms_url, get_lms_commerce_api_url
 from ecommerce.courses.utils import mode_for_seat
-from ecommerce.settings import get_lms_url
 
 logger = logging.getLogger(__name__)
 
@@ -74,8 +74,9 @@ class LMSPublisher(object):
             course_id=course_id
         )
 
-        if not settings.COMMERCE_API_URL:
-            logger.error('COMMERCE_API_URL is not set. Commerce data will not be published!')
+        commerce_api_url = get_lms_commerce_api_url()
+        if not commerce_api_url:
+            logger.error('Commerce API URL is not set. Commerce data will not be published!')
             return error_message
 
         name = course.name
@@ -109,7 +110,7 @@ class LMSPublisher(object):
             'modes': modes,
         }
 
-        url = '{}/courses/{}/'.format(settings.COMMERCE_API_URL.rstrip('/'), course_id)
+        url = '{}/courses/{}/'.format(commerce_api_url.rstrip('/'), course_id)
 
         headers = {
             'Content-Type': 'application/json',
--- a/ecommerce/extensions/api/data.py
+++ b/ecommerce/extensions/api/data.py
@@ -2,12 +2,11 @@
 import logging
 import requests
 
-from django.conf import settings
 from oscar.core.loading import get_model, get_class
 
+from ecommerce.core.url_utils import get_lms_url
 from ecommerce.extensions.api import exceptions
 from ecommerce.extensions.api.constants import APIConstants as AC
-from ecommerce.settings import get_lms_url
 
 NoShippingRequired = get_class('shipping.methods', 'NoShippingRequired')
 OrderTotalCalculator = get_class('checkout.calculators', 'OrderTotalCalculator')
@@ -59,8 +58,9 @@ def get_lms_footer():
         str: HTML representation of the footer.
     """
     try:
+        footer_api_url = get_lms_url('api/branding/v1/footer')
         response = requests.get(
-            get_lms_url('api/branding/v1/footer'),
+            footer_api_url,
             data={'language': 'en'}
         )
         if response.status_code == 200:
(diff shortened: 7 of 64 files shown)


douglashall:
@clintonb This is ready for a final review. The tests will not run on Travis until be get edx-auth-backends==0.2.0 pushed up to pypi. They do run successfully locally with auth-backends:douglashall/WL-286/support_multi_tenancy source installed.

ecommerce/social_auth/strategies.py
```diff
@@ -0,0 +1,18 @@
...
+class CurrentSiteDjangoStrategy(DjangoStrategy):
+    """
+    Python Social Auth strategy which accounts for the current
+    Site when enabling third party authentication.
+    """
+
+    def get_setting(self, name):
+        # First check the current SiteConfiguration for the setting
+        setting = self.request.site.siteconfiguration.social_auth_settings.get(name)
+        if not setting:
```
clintonb, line 15: This doesn't seem right. We should rely solely on site configuration. Having two sources of data has the potential to lead to issues with debugging. Also, it will be nice to get rid of these unnecessary settings.

douglashall, line 15: We are going to need this to roll this change out to production. We can go back and fully rely on SiteConfiguration later on.

clintonb, line 15: We can deploy to a single cluster outside of the ELB. Once the settings are in place, we can cut over.

douglashall, line 15: If you are signing up to manage that deployment, then I can come onboard. Otherwise, I would prefer to stick with what we have here for now.

clintonb, line 15: This is the cleanest deployment strategy. Say we keep this as written. How do you know your changes from `SiteConfiguration` are being used instead of the standard Django settings? As is, there is no assurance the new path/settings are being used. The full switchover will require some manual intervention, but we will know it works.

clintonb, line 15: Hmm...even if we go with the full cutover, this will need to be deployed in two phases: (1) add column to configuration, (2) switch to using new strategy. If we do the full cutover, there will be no way to log into the server. However, we could define a password for someone's staff account and bypass OIDC login altogether. Either way, you have options. My preference is that we make a full cutover rather than take a hybrid approach.

douglashall, line 15: It will work well as written. Not sure what you mean by "no assurance the new path/settings are being used". The new sites that we add will not work at all if we are not pulling settings from SiteConfiguration. The existing site will work because it will be using this fallback to grab settings from django.

douglashall, line 15: I am happy to create a follow up Jira task to remove the fallback and pull it into our next sprint if the concern is that we will never remove the fallback.

clintonb, line 15: Cool.

douglashall, line 15: https://openedx.atlassian.net/browse/WL-333

douglashall:
@mattdrayer Could you give this a second review?

ecommerce/core/admin.py
```diff
@@ -6,7 +6,7 @@
 
 
 class SiteConfigurationAdmin(admin.ModelAdmin):
-    list_display = ('site', 'partner', 'lms_url_root', 'theme_scss_path', 'payment_processors')
+    list_display = ('site', 'partner', 'lms_url_root', 'theme_scss_path', 'payment_processors', 'social_auth_settings')
```
clintonb: These settings don't need to be in the list display.

ecommerce/core/context_processors.py
```diff
@@ -1,8 +1,11 @@
 from django.conf import settings
 
+from ecommerce.core import get_lms_dashboard_url
+
 
 def core(_request):
     return {
+        'lms_dashboard_url': get_lms_dashboard_url(),
+        'platform_name': settings.PLATFORM_NAME,
```
clintonb, line 9: This will eventually need to be changed to either the site or partner name.

ecommerce/core/models.py
```diff
@@ -34,6 +34,13 @@ class SiteConfiguration(models.Model):
         null=False,
         blank=False
     )
+    social_auth_settings = JSONField(
+        verbose_name=_('Python Social Auth settings'),
+        help_text=_("JSON string containing Python Social Auth backend settings."),
+        max_length=255,
```
clintonb: There shouldn't be a maximum length here.

ecommerce/core/models.py
```diff
@@ -34,6 +34,13 @@ class SiteConfiguration(models.Model):
...
+        null=False,
```
clintonb, line 40: Unless you specify a default, this won't work.

ecommerce/core/models.py
```diff
@@ -34,6 +34,13 @@ class SiteConfiguration(models.Model):
         null=False,
         blank=False
     )
+    social_auth_settings = JSONField(
```
clintonb: We shouldn't tie our field names to third-party libraries. What if we decide to replace PSA? `oauth_settings` is more appropriate. The `verbose_name` and `help_text` values should be modified as well.

ecommerce/core/tests/test_views.py
```diff
@@ -33,6 +34,8 @@ def test_all_services_available(self, mock_lms_request):
 
         self._assert_health(status.HTTP_200_OK, Status.OK, Status.OK, Status.OK)
 
+    @mock.patch('ecommerce.core.views.get_lms_heartbeat_url', mock.Mock(return_value=''))
```
clintonb, line 37: Why are these mocks only necessary for the database test? If they must be mocked, more realistic values should be used.

ecommerce/coupons/tests/test_views.py
```diff
@@ -129,7 +129,6 @@ def test_used_voucher(self):
         self.assertEqual(msg, _('This coupon has already been used'))
 
 
-@httpretty.activate
```
clintonb, line 143: Why was this removed? `httpretty` is used by almost every test in this class.

ecommerce/coupons/tests/test_views.py
```diff
@@ -11,10 +11,10 @@
 from oscar.test.utils import RequestFactory
```
clintonb, line 11: What's going on in this file? Aside form changing `get_lms_url`, why are any of the other changes necessary?

ecommerce/extensions/api/data.py
```diff
@@ -59,8 +59,9 @@ def get_lms_footer():
         str: HTML representation of the footer.
     """
     try:
+        lms_url = get_lms_url('api/branding/v1/footer')
```
clintonb: `footer_api_url`

ecommerce/extensions/api/serializers.py
```diff
@@ -373,8 +373,7 @@ def get_benefit(self, obj):
         return (obj.offers.first().benefit.type, obj.offers.first().benefit.value)
 
     def get_redeem_url(self, obj):
-        domain = settings.ECOMMERCE_URL_ROOT
-        return "{}/coupons/offer/?code={}".format(domain, obj.code)
+        return get_ecommerce_url("/coupons/offer/?code={}".format(obj.code))
```
clintonb: Append the query string after you get the URL. `get_ecommerce_url` expects to receive a path, so that's all it should receive.

``` python
def get_redeem_url(self, obj):
  url = get_ecommerce_url('/coupons/offer/')
  return '{url}?code={code}'.format(url=url, code=obj.code)
```

ecommerce/extensions/dashboard/users/views.py
```diff
@@ -1,11 +1,14 @@
 import logging
+import requests
```
clintonb: Put these back where they were. Neither is a default Python library.

ecommerce/extensions/voucher/utils.py
```diff
@@ -77,7 +79,7 @@ def generate_coupon_report(coupon_vouchers):
                 discount = _("{percentage} %").format(percentage=benefit_value)
             else:
                 discount = currency(benefit_value)
-            URL = '{}/coupons/redeem/?code={}'.format(settings.ECOMMERCE_URL_ROOT, voucher.code)
+            URL = get_ecommerce_url('/coupons/redeem/?code={}'.format(voucher.code))
```
clintonb: Same as above. Pass only the path to `get_ecommerce_url`.

ecommerce/settings/base.py
```diff
@@ -209,12 +211,6 @@
 # Absolute URL used to construct LMS URLs.
 LMS_URL_ROOT = None
```
clintonb: Is this still needed?

ecommerce/settings/test.py
```diff
@@ -45,15 +47,9 @@ def __getitem__(self, item):
 
 
 # URL CONFIGURATION
-ECOMMERCE_URL_ROOT = 'http://localhost:8002'
-
-LMS_URL_ROOT = 'http://127.0.0.1:8000'
+ECOMMERCE_URL_ROOT = 'http://testserver.fake'
```
clintonb: This should not be the same as `LMS_URL_ROOT`. Make them different to avoid confusion when debugging test failures.

ecommerce/settings/test.py
```diff
@@ -70,6 +66,9 @@ def __getitem__(self, item):
 PASSWORD_HASHERS = (
     'django.contrib.auth.hashers.MD5PasswordHasher',
 )
+
+SOCIAL_AUTH_EDX_OIDC_MULTISITE_KEY = 'fake-key'
```
clintonb: These aren't needed.

ecommerce/social_auth/tests/test_strategies.py
```diff
@@ -0,0 +1,38 @@
...
+from ecommerce.tests.testcases import TestCase
+
+
+class CurrentSiteDjangoStrategyTests(TestCase):
+    """Tests of the CurrentSiteDjangoStrategy."""
+
+    def setUp(self):
+        super(CurrentSiteDjangoStrategyTests, self).setUp()
+        self.site.siteconfiguration.social_auth_settings = {
+            'SOCIAL_AUTH_EDX_OIDC_MULTISITE_KEY': 'test-key'
```
clintonb: MULTISITE is no longer valid.

ecommerce/tests/mixins.py
```diff
@@ -247,6 +249,7 @@ def setUp(self):
 
         Site.objects.get_current().delete()
         site_configuration = SiteConfigurationFactory(
+            lms_url_root="http://%s" % domain,
```
clintonb: Move this to the factory. Also, use 'https' and `string.format()`. See the decorator pattern for lazy attributes at https://factoryboy.readthedocs.org/en/latest/introduction.html#lazyattribute, and pull the value from `obj.site.domain`.

ecommerce/core/context_processors.py
douglashall, line 9: Yes, let's make that change separately.

ecommerce/core/tests/test_views.py
douglashall, line 37: Because by mocking the database cursor, we cause the Django CurrentSiteMiddleware to fail when it tries to look up the current site. Not sure what you mean by more realistic value?

clintonb, line 37: "more realistic value" -> an actual URL

douglashall, line 37: The URL returned by the mock is inconsequential considering the purpose of this test.

ecommerce/coupons/tests/test_views.py
douglashall, line 11: For some reason we were requiring an LMS instance to be up and running in order for most of these tests to run. I have fixed this by mocking the calls to the LMS footer endpoint.

ecommerce/coupons/tests/test_views.py
douglashall, line 143: Decorating the individual test methods seems more clear when every test does not have a request being mocked.

ecommerce/coupons/tests/test_views.py
clintonb, line 11: Something's wrong here. These tests did not require an LMS instance. If they had they never would have been merged, because Travis doesn't run LMS.

clintonb, line 11: Taking a closer look, `CouponOfferViewTests.setUp()` mocks the call to the LMS Footer API.

douglashall, line 11: Right. I will revert these changes.

ecommerce/settings/base.py
douglashall: Yes. There are tests that reference this setting.

clintonb: Which tests? Why not update them and remove this setting, especially since there's potential for brand leakage?

douglashall: I have updated the tests to use `ecommerce.core.get_lms_url()` instead of the setting. There is still a management command that wants the setting. Any objections to making the LMS URL an option on the command?

https://github.com/edx/ecommerce/blob/douglashall/WL-286/support_multitenant_oauth/ecommerce/extensions/catalogue/management/commands/migrate_course.py#L42

clintonb: `migrate_course`? It takes a `Partner` identifier. Use that to get the `SiteConfiguration` and the correct LMS URL.

douglashall: `Partner` can have more than one `SiteConfiguration` though, no?

clintonb: Yeah...replace `Partner` with `Site`.

douglashall:
@clintonb This is ready for another look. I have removed all references to URLs in Django settings.

ecommerce/core/__init__.py
```diff
@@ -1 +1,47 @@
...
+
 default_app_config = 'ecommerce.core.config.CoreAppConfig'  # pragma: no cover
+
+
+def get_ecommerce_url(path=''):
+    request = get_current_request()
+    if request:
+        ecommerce_url_root = "{}://{}".format(request.scheme, request.site.domain)
+        return urljoin(ecommerce_url_root, path)
+    return None
```
clintonb: An exception (e.g. `MissingRequestError`) should be raised if there is no `request` value. Otherwise, you run the risk of consuming code not checking the return value, and displaying an invalid URL. Also, add a docstring.

ecommerce/core/__init__.py
```diff
@@ -1 +1,47 @@
...
+    if request:
+        return urljoin(request.site.siteconfiguration.lms_url_root, path)
+    return None
+
+
+def get_oauth2_provider_url():
+    return get_lms_url('/oauth2')
+
+
+def rewrite_lms_url_for_current_site(url):
```
clintonb: Update the payment providers settings to accept a receipt/error page path instead of a URL.

ecommerce/core/tests/test_init.py
```diff
@@ -0,0 +1,10 @@
+import mock
```
clintonb: Consider moving the code being tested out of `__init__.py` and into `url_utils.py`. Rename this file to `test_url_utils.py`.

ecommerce/core/tests/test_init.py
```diff
@@ -0,0 +1,10 @@
...
+
+from ecommerce.core import get_lms_url
+from ecommerce.tests.testcases import TestCase
+
+
+class UrlUtilityTests(TestCase):
+    @mock.patch('ecommerce.core.get_current_request', mock.Mock(return_value=None))
+    def test_get_lms_url_with_no_current_request(self):
```
clintonb: What if there _is_ a request? The positive case needs to be tested.

douglashall: The positive case is already covered by numerous other tests.

ecommerce/courses/publishers.py
```diff
@@ -74,8 +74,9 @@ def publish(self, course, access_token=None):
             course_id=course_id
         )
 
-        if not settings.COMMERCE_API_URL:
-            logger.error('COMMERCE_API_URL is not set. Commerce data will not be published!')
+        commerce_api_url = get_lms_commerce_api_url()
+        if not commerce_api_url:
+            logger.error('Could not find the commerce api url. Commerce data will not be published!')
```
clintonb: Commerce API URL si not set. Commerce data will not be published.

ecommerce/core/tests/test_init.py
douglashall: I would like to leave these in `__init__.py`

ecommerce/core/__init__.py
douglashall: The idea here was to make for a smooth rollout. Do we store the payment processor settings in https://github.com/edx/configuration somewhere? Can you point me to the right place?

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -171,30 +165,30 @@ class Command(BaseCommand):
...
                     help='Save the migrated data to the database. If this is not set, '
                          'migrated data will NOT be saved to the database.'),
-        make_option('--partner',
+        make_option('--site',
                     action='store',
-                    dest='partner_short_code',
+                    dest='site_domain',
                     default=None,
-                    help='Short code for the partner providing the course.'),
+                    help='Domain for the site providing the course.'),
```
clintonb: ecommerce site

ecommerce/settings/production.py
```diff
@@ -11,6 +11,10 @@
 from ecommerce.settings.logger import get_logger_config
 
 
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#site-id
+# This needs to be set to None in order to support multitenancy
+SITE_ID = None
```
clintonb, line 16: The value is not defined by default. Is this really necessary?

douglashall, line 16: Correct. I'm erring on the side of over documentation here.

ecommerce/core/__init__.py
clintonb: https://github.com/edx/configuration/blob/master/playbooks/roles/ecommerce/defaults/main.yml#L62-L80

ecommerce/core/tests/test_init.py
clintonb: Why?

douglashall: Makes for shorter imports. Is there a reason we would want to move them?

ecommerce/core/__init__.py
douglashall: Do we have a module that houses custom errors somewhere? I can't seem to find one.

ecommerce/core/tests/test_init.py
clintonb: Organization. All of the functions are URL utilities. Grouping them into a package seems logical. Shorter imports don't really matter.

ecommerce/core/__init__.py
clintonb: Custom exceptions generally go into an exceptions package. In this case, you'll want to create `ecommerce/core/exceptions.py`.

ecommerce/core/tests/test_init.py
douglashall: Sounds good. This is something we would want to raise earlier in the review process.

douglashall:
@clintonb All comments addressed. Thanks again for all your help.

ecommerce/core/url_utils.py
```diff
@@ -0,0 +1,46 @@
+from urlparse import urljoin
+
+from threadlocals.threadlocals import get_current_request
+
+from ecommerce.core.exceptions import MissingRequestError
+
+
+def get_ecommerce_url(path=''):
+    """
+    Returns path joined with the appropriate ecommerce URL root for the current site
```
clintonb, line 10: Document the exception being raised.

clintonb:
:+1:

PR #547, opened 2016-02-12
Don't freeze a basket in the preparation step

vkaracic:
The basket shouldn't be frozen while it's still being prepared, but rather on the checkout step:
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/v2/views/checkout.py#L37


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=547&repo=edx%2Fecommerce

clintonb:
:+1:

PR #549, opened 2016-02-14
Log seconds since placement after trying to fulfill an order

rlucioni:
Allows easier monitoring of the time orders spend in an unfulfilled state.

@bderusha or @cpennington, please review.


bderusha:
:+1: lgtm

PR #550, opened 2016-02-15
[SOL-1640] Change 'Add to basket' link from basket:single-item to basket:summary

vkaracic:
`prepare_basket` was called twice, once in `CouponRedeemView` once in `BasketSingleItemView` and caused the product to be added twice. Since the `BasketSingleItemView` view does the same checks that are already done in `CouponRedeemView` there is no need to route through that view and therefore it's been rerouted to the summary page.

Maybe even change the redeem links to point to the basket single-item view in the first place?

@marjev @mjfrey


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=550

marjev:
LGTM :+1:

mjfrey:
:+1:

PR #551, opened 2016-02-16
[SOL-1626] Modifications to landing page - Discount code flow

marjev:
screen shot 2016-02-18 at 14 36 38


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=551

marjev:
@mjfrey, @vkaracic: Could you take a look?

JIRA ticket: https://openedx.atlassian.net/browse/SOL-1626

@srpearce, could you please review the text in the template? Check the JIRA ticket (link above) to see the design prototype.

ecommerce/coupons/views.py
```diff
@@ -118,16 +120,18 @@ def get_context_data(self, **kwargs):
                         'footer': footer
                     }
                 course['image_url'] = get_lms_url(course['media']['course_image']['uri'])
-                stock_records = voucher.offers.first().benefit.range.catalog.stock_records.first()
+                benefit = voucher.offers.first().benefit
+                stock_record = benefit.range.catalog.stock_records.first()
                 benefit_type = voucher.offers.first().benefit.type
                 benefit_value = voucher.offers.first().benefit.value
```
vkaracic: Since you're setting the benefit to `benefit` variable, can you refactor these two lines to be `benefit.type` and `benefit.value`, please.

vkaracic: Actually, you don't need these two lines, just use `benefit.type` or `benefit.value` instead of `benefit_type` and `benefit_value`.

marjev: Done.

ecommerce/extensions/basket/views.py
```diff
@@ -87,17 +88,9 @@ def get_context_data(self, **kwargs):
...
-                    line.is_discounted = True if benefit.value < 100 else False
-                else:
-                    line.is_discounted = True
-                    discount_percentage = float(benefit.value / line.unit_price_excl_tax) * 100.0
-                    line.discount_percentage = 100.0 if discount_percentage > 100 else discount_percentage
-            else:
-                line.is_discounted = False
-                line.discount_percentage = 0
+            discount_info = get_voucher_discount_info(benefit, line.unit_price_excl_tax)
+            line.discount_percentage = discount_info['discount_percentage']
```
vkaracic, line 92: You'll need to format `discount_percentage` to two decimal places.
screenshot from 2016-02-19 12-19-41

vkaracic:
Other than the couple comments :+1:

marjev:
UI nitpicks commit includes:
- [x] Integer percentage shouldn't be displayed as float values.
- [x] Fonts on the redeem page should match those on the basket page.

mjfrey:
:+1:

PR #539, opened 2016-02-09
[WIP][SOL-1616] Coupons category and subcategory implementation

swpdsgn:
@mjfrey

PR #552, opened 2016-02-16
Coupon categories

vkaracic:
This is a combination of:
- backend work: https://openedx.atlassian.net/browse/SOL-1539 PR: https://github.com/edx/ecommerce/pull/475
- frontend work: https://openedx.atlassian.net/browse/SOL-1616 PR: https://github.com/edx/ecommerce/pull/539

@swpdsgn
@mjfrey take a look. 

In this PR we’re introducing categories for coupons.

_FRONTEND:_
The additional fields in the UI look like this:
https://drive.google.com/a/edx.org/file/d/0ByBQUYJ4LMyAa2l1dUFhbXZyaXM/view?usp=sharing
https://drive.google.com/a/edx.org/file/d/0ByBQUYJ4LMyAT2hCWFY4VGs2OG8/view?usp=sharing

The category dropdown get’s it’s content from the coupons categories that already exist. Deleting and adding new coupon categories is done via the Django admin panel.
The sub-category field is a typeahead where a user can select an already existing sub-category or add a new one - only one sub-category at a time can be added.

_BACKEND:_
The backend works so that it:
1. retrieves the category (and sub-category) values from the request
2. checks if the category for the coupon exists, if not report an error
3. if a sub-category value is provided it is added to the category object (category > subcategory)

After all that the category is via `ProductCategory` associated with that particular coupon.
- a new `CategoryViewSet` is introduced. A read-only view set that lists out all the categories and supports filtering with `depth` to distinguish categories from sub-categories and `path` to make retrieving sub-categories of a specific category easier
  For example to filter for all coupon categories (which has a `depth=1` and `path=0002`):

```
api/v2/categories?depth=1&path=0002
```
- coupons API now displays the category for the coupons and the sub-category if it exists, if not it’s `null`. `CategorySerializer` displays all the category information along with child sub-categories.
- new migration that inserts all the default categories for coupons, and assigns exactly one to the existing coupons to avoid database errors


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=552

ecommerce/extensions/catalogue/migrations/0015_default_categories.py
```diff
@@ -0,0 +1,75 @@
...
+
+Category = get_model("catalogue", "Category")
+Product = get_model("catalogue", "Product")
+ProductCategory = get_model("catalogue", "ProductCategory")
+
+
+def create_default_categories(apps, schema_editor):
+    """Create default coupon categories."""
+    categories = (
+        'Coupons > AffiliatePromo',
```
clintonb: Use full names instead of abbreviations (e.g. Affiliate Promotion) here and for the other categories.

pbaruah:
@vkaracic or Duje

Course Promotion
Bulk Enrollment
Customer Service
Marketing-Other
Paid Cohort
ConnectEd
Marketing Partner Promotion
Services-Other
Financial Assistance
Retention Promotion
Support-Other
Affiliate Promotion
Upsell Promotion
Geography Promotion

mjfrey:
LGTM :+1:

mjfrey:
@vkaracic squash commits

mjfrey:
@clintonb can take a look?

ecommerce/extensions/api/serializers.py
```diff
@@ -385,6 +387,19 @@ class Meta(object):
         )
 
 
+class CategorySerializer(serializers.ModelSerializer):
+    child = serializers.SerializerMethodField()
```
clintonb, line 391: children

ecommerce/extensions/api/serializers.py
```diff
@@ -385,6 +387,19 @@ class Meta(object):
...
+
+    def get_child(self, obj):
+        child = obj.get_children()
+        serializer = CategorySerializer(child, many=True)
+        return serializer.data
+
+    class Meta(object):
+        model = Category
+        fields = ('id', 'name', 'slug', 'description', 'path', 'depth', 'numchild', 'image', 'child')
```
clintonb, line 400: Is there an actual use case for all of these fields? Stick to the bare minimum. It's easier to add fields than it is to modify or remove them.

ecommerce/extensions/api/serializers.py
clintonb, line 391: Use [`source`](http://www.django-rest-framework.org/api-guide/fields/#source).

``` python
children = CategorySerializer(source='get_children', many=True)
```

clintonb:
The breaking out of category and sub-category unnecessarily complicates the code and UI. Oscar's standard product creation UI uses a much simpler selector that even supports search. A similar approach should be taken here.

screen shot 2016-02-18 at 12 39 13 pm

vkaracic:
Closing this PR due to an update in product specifications.

PR #553, opened 2016-02-17
Fix coupons date update

vkaracic:
Coupons update expected `start_datetime` and `start_endtime` keywords instead of `start_date` and `end_date`. This PR fixes that.

@marjev 
https://openedx.atlassian.net/browse/SOL-1630


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=553

marjev:
lgtm :+1:

PR #554, opened 2016-02-18
Include 'jquery-cookie' to fix JS errors in basket page.

mjfrey:
Need to include query-cookie to fit JS error on the basket page.


mjfrey:
@marjev

marjev:
LGTM :+1:  I'll merge this. Thanks for fixing this.

PR #555, opened 2016-02-18
Change benefit type to Absolute

vkaracic:
The value of the `Benefit.FIXED` keyword is `Absolute` and not `Fixed`.
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/abstract_models.py#L390-L396

@marjev @mjfrey


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=555

marjev:
LGTM :+1:

mjfrey:
:+1:

PR #557, opened 2016-02-18
Only add a single quantity of a product

mjfrey:
Only allow a single quantity of a product when adding to a basket.


mjfrey:
@vkaracic

vkaracic:
:+1:

PR #561, opened 2016-02-19
Fix for firefox form submit

mjfrey:
Append form to body so Firefox submits it.


clintonb:
The form needs to be hidden. See https://github.com/edx/edx-platform/blob/master/lms/static/js/commerce/credit.js#L28. This is the second bug fix for Firefox. Has someone actually manually walked through checkout for both payment processors and made sure the functionality works?

mjfrey:
@marjev please fix the tests for payment JS. Here are examples. 
https://github.com/edx/edx-platform/blob/master/lms/static/js/spec/verify_student/make_payment_step_view_spec.js

PR #562, opened 2016-02-19
Flush basket as we only support single item for now.

mjfrey:
Remove all existing lines in a basket as our basket page does not support managing of multiple items.

### DIFF
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -56,16 +56,17 @@ class BasketUtilsTests(TestCase):
         self.assertFalse(basket.vouchers.all())
         self.assertFalse(basket.applied_offers())
 
-    def test_prepare_basket_with_multiple_quantity(self):
-        """ Verify a basket is returned and only contains a single quantity of product. """
-        product = ProductFactory()
-        basket = prepare_basket(self.request, product)
-        basket = prepare_basket(self.request, product)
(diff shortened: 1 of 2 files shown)


mjfrey:
@vkaracic

vkaracic:
Do we need a test for this? In which you add two different products to a basket and assert that only the second one survived.

mjfrey:
Yes we do.  Let me add one.

mjfrey:
@vkaracic add test.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -56,16 +56,17 @@ def test_prepare_basket_without_voucher(self):
...
         self.assertFalse(basket.applied_offers())
 
-    def test_prepare_basket_with_multiple_quantity(self):
-        """ Verify a basket is returned and only contains a single quantity of product. """
-        product = ProductFactory()
-        basket = prepare_basket(self.request, product)
-        basket = prepare_basket(self.request, product)
+    def test_prepare_basket_with_multiple_products(self):
+        """ Verify a basket is returned and only contains a single product. """
+        product1 = ProductFactory(title='Product One', stockrecords__partner__short_code='test1')
```
vkaracic: Since you're not using it anywhere in the test you can remove the title argument.

vkaracic:
One nit, otherwise :+1:

PR #563, opened 2016-02-19
[WIP] Coupon categories and notes

vkaracic:
Partial continuation of https://github.com/edx/ecommerce/pull/552

In this PR we’re introducing categories for coupons and the notes coupon product attribute.

_FRONTEND:_
- two new fields are added to the coupon creation form:
  - `category` dropdown that is populated by all the coupon sub-categories [*required field]
  - `note` text field
- `category` and `note` fields added to the coupon details page, coupon creation page and coupon editing page
- `coupon categories` backbone collection added 

Screenshots: https://drive.google.com/a/edx.org/folderview?id=0ByBQUYJ4LMyAdmdkRDVRNHFJWG8&usp=sharing

_BACKEND:_
- a new `CategoryViewSet` is introduced. A read-only view set that lists out all the categories and supports filtering with `depth` to distinguish categories from sub-categories (currently we have `Seat` and `Coupon` as categories on the top level) and `path` to make retrieving sub-categories of a specific category easier
  For example to filter for all coupon categories (which has a `depth=2` and `path=0002`):

```
api/v2/categories?depth=1&path=0002
```
- coupons API now displays the note information and category for the coupons. `CategorySerializer` displays all the category information together with the category's children categories.
- new migration that inserts all the default categories for coupons, and assigns exactly one to the existing coupons to avoid database errors
- new migration that creates the optional `Note` product attribute and assigns it to the coupon product class


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=563

ecommerce/extensions/api/serializers.py
```diff
@@ -419,9 +428,23 @@ def get_vouchers(self, obj):
...
+    def get_category(self, obj):
+        category = ProductCategory.objects.get(product=obj).category
+        serializer = CategorySerializer(category)
+        return serializer.data
+
+    def get_note(self, obj):
+        try:
+            return obj.attr.note
+        except AttributeError:
+            return ''
```
clintonb: Use `None`. An empty string implies an empty value has been set.

ecommerce/extensions/api/serializers.py
```diff
@@ -392,6 +399,8 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     seats = serializers.SerializerMethodField()
     client = serializers.SerializerMethodField()
     vouchers = serializers.SerializerMethodField()
+    category = serializers.SerializerMethodField()
```
clintonb, line 409: Try: `categories = CategorySerializer(many=true)`

ecommerce/extensions/api/serializers.py
clintonb: Looking forward, if other attributes are added, it may be simpler to use `ProductAttributeValueSerializer` as we do with `ProductSerializer`.

swpdsgn:
Pinging every1 @vkaracic  @vkaracic @mjfrey 
The changes have been made and pushed. Michael, a screenshot is not neccesarry cause nothing changed from the UI perspective, the changes that were made were mostly in the view models (apart from the Subcategory to Note label change)

vkaracic:
I've added the screenshots as we are adding new fields to the UI. @swpdsgn please check the frontend documentation to see if I missed something.

ecommerce/extensions/api/serializers.py
```diff
@@ -385,13 +387,37 @@ class Meta(object):
         )
 
 
+class CategorySerializer(serializers.ModelSerializer):
+    children = serializers.SerializerMethodField()
```
clintonb, line 391: Does this not work?

``` python
children = CategorySerializer(many=True, source='get_children')
```

vkaracic, line 391: No. The `Coupons` category shows the number of children (15 or 16) but the children attribute is emtpy `[]`.

clintonb, line 391: Are you certain your data is correct locally? I created a child category via Oscar's dashboard, and `get_children()` returns the expected child.

ecommerce/extensions/api/serializers.py
```diff
@@ -385,13 +387,37 @@ class Meta(object):
...
+        serializer = CategorySerializer(children, many=True)
+        return serializer.data
+
+    class Meta(object):
+        model = Category
+
+
 class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     """ Serializer for Coupons. """
+    attribute_values = serializers.SerializerMethodField()
```
clintonb, line 404: Move this, and the associated method, a mixin (e.g. `ProductAttributeValuesMixin`) shared with `ProductSerializer`.

ecommerce/extensions/api/serializers.py
```diff
@@ -385,13 +387,37 @@ class Meta(object):
...
     coupon_type = serializers.SerializerMethodField()
     last_edited = serializers.SerializerMethodField()
     seats = serializers.SerializerMethodField()
     client = serializers.SerializerMethodField()
-    vouchers = serializers.SerializerMethodField()
+    category = serializers.SerializerMethodField()
+
+    def get_attribute_values(self, product):
+        request = self.context.get('request')
+        attributes = product.attr
```
clintonb, line 413: Variable not necessary.

ecommerce/extensions/api/serializers.py
```diff
@@ -385,13 +387,37 @@ class Meta(object):
...
+
 class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     """ Serializer for Coupons. """
+    attribute_values = serializers.SerializerMethodField()
     coupon_type = serializers.SerializerMethodField()
     last_edited = serializers.SerializerMethodField()
     seats = serializers.SerializerMethodField()
     client = serializers.SerializerMethodField()
-    vouchers = serializers.SerializerMethodField()
+    category = serializers.SerializerMethodField()
```
clintonb, line 409: `categories = CategorySerializer(many=True)`

clintonb, line 409: `CategorySerializer` should be modified to accept a depth parameter. I doubt you want to retrieve the entire category subtree when getting data for one coupon.

ecommerce/extensions/api/v2/views/categories.py
```diff
@@ -0,0 +1,28 @@
...
+    depth = django_filters.NumberFilter(name='depth', lookup_type='exact')
+    path = django_filters.CharFilter(name='path', lookup_type='startswith')
+
+    class Meta(object):
+        model = Category
+        fields = ('depth', 'path')
+
+
+class CategoryViewSet(viewsets.ReadOnlyModelViewSet):
+    queryset = Category.objects.filter(depth=1)
```
clintonb, line 24: Why is this filter necessary? It seems to unnecessarily limit the scope of the API solely to top-level categories. I tend to avoid filtering so that the API can be useful beyond the initial use case. This is especially helpful for debugging.

ecommerce/extensions/api/v2/views/categories.py
```diff
@@ -0,0 +1,28 @@
...
+    serializer_class = serializers.CategorySerializer
+    filter_backends = (filters.DjangoFilterBackend,)
+    filter_class = CategoryFilter
+    permission_classes = (IsAuthenticated, IsAdminUser,)
```
clintonb, line 28: `IsAuthenticated` is sufficient. These aren't exactly state secrets.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -75,6 +77,8 @@ def create(self, request, *args, **kwargs):
             quantity = request.data[AC.KEYS.QUANTITY]
             price = request.data[AC.KEYS.PRICE]
             partner = request.site.siteconfiguration.partner
+            category = request.data[AC.KEYS.CATEGORY]
```
clintonb, line 80: This block is a bit too large. Use a `Serializer` or `Form` to contain it. You also get the benefit of actually validating your data as well.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -168,10 +175,25 @@ def create_coupon_product(self, title, price, data):
...
 
+        category = Category.objects.get(id=data['category'])
+
+        # The reason why we are not using:
+        #   ProductCategory.objects.get_or_create(product=product, category=category)
+        # here is that in case an existing coupon's category has been changed this will
+        # create a new product category which we don't want. We want one coupon to have
+        # one product category. Therefor if a coupon's category is edited the category of the
+        # existing product category will be edited.
+        try:
```
clintonb, line 190: How about this?

``` python
coupon_product.categories.all().delete()
coupon_product.categories.create(category=category)
```

ecommerce/extensions/catalogue/migrations/0015_default_categories.py
```diff
@@ -0,0 +1,76 @@
...
+        'Coupons > Upsell Promotion',
+        'Coupons > Other'
+    )
+    for breadcrumbs in categories:
+        create_from_breadcrumbs(breadcrumbs)
+
+
+def remove_default_categories(apps, schema_editor):
+    """Remove default coupon categories."""
+    default_categories = [
```
clintonb, line 38: Move this to a global variable, and use it in `create_default_categories`.

ecommerce/extensions/catalogue/migrations/0015_default_categories.py
```diff
@@ -0,0 +1,76 @@
...
+        'Affiliate Promotion', 'Bulk Enrollment', 'ConnectEd', 'Course Promotion',
+        'Customer Service', 'Financial Assistance', 'Geography Promotion',
+        'Marketing Partner Promotion', 'Marketing-Other', 'Paid Cohort', 'Other',
+        'Retention Promotion', 'Services-Other', 'Support-Other', 'Upsell Promotion',
+    ]
+    for category in default_categories:
+        Category.objects.get(name=category).delete()
+
+
+def assign_categories_to_coupons(self, schema_editor):
```
clintonb, line 48: No. Don't make a fake category. Manually go in and fix the data.

clintonb:
This PR is _way_ too big for the functionality being added. I took some time to implement scale it back. Please make use of the code in https://github.com/edx/ecommerce/commits/clintonb/coupon-category.

FYI @mjfrey

PR #564, opened 2016-02-19
[SOL-1633] Coupon checkout page title in browser

swpdsgn:
@mjfrey @vkaracic

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -64,8 +64,15 @@ class BasketSummaryView(BasketView):
         context = super(BasketSummaryView, self).get_context_data(**kwargs)
         lines = self.request.basket.lines.all()
         api = EdxRestApiClient(get_lms_url('api/courses/v1/'))
-        page_title = 'Upgrade your enrollment'
-        for line in lines:
+        if 'HTTP_REFERER' in self.request.META:
+            referer = self.request.META['HTTP_REFERER']
+        else:
+            referer = ''
+        if referer.find('dashboard') == -1:
+            page_title = 'Enroll in '
+        else:
+            page_title = 'Upgrade your enrollment for '
+        for index, line in enumerate(lines):
             course_id = line.product.course_id
             # Get each course type so we can display to the user at checkout.
             try:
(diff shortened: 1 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @swpdsgn! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=564

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -3,7 +3,7 @@
 {% load staticfiles %}
 
 {% block title %}
-{% trans "Basket" %}
+{% blocktrans %} {{ page_title }} {% endblocktrans %}
```
clintonb, line 6: This will not work as expected. We do not have dynamic translations.

ecommerce/extensions/basket/views.py
```diff
@@ -64,9 +64,16 @@ def get_context_data(self, **kwargs):
         context = super(BasketSummaryView, self).get_context_data(**kwargs)
         lines = self.request.basket.lines.all()
         api = EdxRestApiClient(get_lms_url('api/courses/v1/'))
-        for line in lines:
+        if 'HTTP_REFERER' in self.request.META:
+            referer = self.request.META['HTTP_REFERER']
+        else:
+            referer = ''
+        if referer.find('dashboard') == -1:
```
clintonb, line 71: This is brittle. I recommend keeping this simple, and sticking with "Basket" or something similar. It's a common e-commerce term, and we don't assume our products are always course seats.

ecommerce/templates/oscar/basket/basket.html
swpdsgn, line 6: Whats your recommendation on this one? Just remove the trans block and simply echo the page title?

ecommerce/extensions/basket/views.py
swpdsgn, line 71: Please elaborate on this. This seemed to me the best way to check the origin of the request without messing and adding new stuff to the models. Why would you say its brittle when it'll allways fall back to "Enroll in [TITLE_COURSE]" which is basically the default.

clintonb:
My recommendation, as it was the last time a similar PR came up, is that we do not do this. The value delivered by this change is minuscule at best, and they are too easy to get wrong. If you look at pretty much every major e-commerce site, the title simply reads "Basket" or "Shopping Cart".

FYI @pbaruah @mjfrey @rock345

swpdsgn:
OK then, I'll wait for you guys to reach an agreement on this.

PR #565, opened 2016-02-22
	Updated basket creation endpoint

clintonb:
Instead of merging baskets, we create a new basket every time. This ensures we don't have issues with race conditions.

ECOM-3663

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_baskets.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_baskets.py
@@ -2,7 +2,6 @@
 from collections import namedtuple
 from decimal import Decimal
 import json
-import logging
 
 import ddt
 from django.contrib.auth import get_user_model
@@ -12,7 +11,6 @@ import mock
 from oscar.core.loading import get_model
 from oscar.test import factories
 from rest_framework.throttling import UserRateThrottle
-from testfixtures import LogCapture
 
 from ecommerce.extensions.api import exceptions as api_exceptions
 from ecommerce.extensions.api.constants import APIConstants as AC
@@ -91,18 +89,18 @@ class BasketCreateViewTests(BasketCreationMixin, ThrottlingMixin, TransactionTes
         self.assert_successful_basket_creation(skus, checkout, payment_processor_name, requires_payment)
 
     def test_multiple_baskets(self):
-        """Test that basket operations succeed if multiple editable baskets exist for the user."""
-        user = User.objects.create_user(
-            username=self.USER_DATA['username'],
-        )
-
+        """ Test that basket operations succeed if the user has editable baskets. The endpoint should
+        ALWAYS create a new basket. """
         # Create two editable baskets for the user
-        for _ in xrange(2):
-            basket = Basket(owner=user, status='Open')
+        basket_count = 2
+        for _ in xrange(basket_count):
+            basket = Basket(owner=self.user, status='Open')
             basket.save()
 
+        self.assertEqual(self.user.baskets.count(), basket_count)
         response = self.create_basket(skus=[self.PAID_SKU], checkout=True)
         self.assertEqual(response.status_code, 200)
+        self.assertEqual(self.user.baskets.count(), basket_count + 1)
 
     @mock.patch('oscar.apps.partner.strategy.Structured.fetch_for_product')
     def test_order_unavailable_product(self, mock_fetch_for_product):
@@ -145,7 +143,7 @@ class BasketCreateViewTests(BasketCreationMixin, ThrottlingMixin, TransactionTes
             self.PATH,
             data=json.dumps(request_data),
             content_type=JSON_CONTENT_TYPE,
-            HTTP_AUTHORIZATION='JWT ' + self.generate_token(self.USER_DATA)
+            HTTP_AUTHORIZATION=self.generate_jwt_token_header(self.user)
         )
         self.assertEqual(response.status_code, 400)
         self.assertEqual(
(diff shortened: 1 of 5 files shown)


clintonb:
@rlucioni @bderusha please review

FYI @edx/ecommerce

ecommerce/tests/mixins.py
```diff
@@ -127,12 +124,11 @@ def create_basket(self, skus=None, checkout=None, payment_processor_name=None, a
             request_data[AC.KEYS.PAYMENT_PROCESSOR_NAME] = payment_processor_name
 
         if auth:
-            token = token or self.generate_token(self.USER_DATA)
             response = self.client.post(
                 self.PATH,
                 data=json.dumps(request_data),
                 content_type='application/json',
-                HTTP_AUTHORIZATION='JWT ' + token
+                HTTP_AUTHORIZATION='JWT {}'.format(token) if token else self.generate_jwt_token_header(self.user)
```
bderusha, line 131: NIT: prefer named variables in interpolated strings.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -186,8 +174,8 @@ def create(self, request, *args, **kwargs):
             try:
                 response_data = self._checkout(basket, payment_processor())
             except Exception as ex:  # pylint: disable=broad-except
-                basket.thaw()
-                logger.exception('Failed to initiate checkout for Basket [%d]. Basket has been thawed.', basket.id)
+                basket.delete()
```
bderusha, line 192: For my own edification, are we doing this here because right now we end up needing to go in to the system and delete old thawed baskets?  Mostly wondering why we were thawing before and what has changed to make us want to delete instead.

clintonb, line 192: Prior to this change we were recycling and merging existing baskets. That's the primary source of the production issue. Given that we are no longer reusing baskets, it makes no sense to keep it. We have enough info in the logs (esp. the items added to the basket) to recreate the scenario, if necessary.

rlucioni:
:+1:

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -226,16 +225,14 @@ def test_checkout_exception(self):
         """ If an exception is raised when initiating the checkout process, a PaymentProcessorResponse should be
         recorded, and the view should return HTTP status 500. """
 
+        self.user.baskets.all().delete()
```
bderusha, line 231: Wondering if it makes sense to put this in the `setUp` so that we know that each test is being run against the set of baskets created specifically for that test.

ecommerce/extensions/api/v2/views/baskets.py
bderusha, line 192: Ah makes sense. Thanks!

bderusha:
:+1:

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -31,14 +31,6 @@ class BasketCreateView(EdxOrderPlacementMixin, generics.CreateAPIView):
     """
     permission_classes = (IsAuthenticated,)
 
-    # Disable atomicity for the view. Otherwise, we'd be unable to commit to the database
-    # until the request had concluded; Django will refuse to commit when an atomic() block
-    # is active, since that would break atomicity. Without an order present in the database
-    # at the time fulfillment is attempted, asynchronous order fulfillment tasks will fail.
-    @method_decorator(transaction.non_atomic_requests)
```
rlucioni, line 38: @clintonb I'm retracting my thumb (sorry!) until this and the manual transaction management is restored. When used in combination with async order fulfillment, it's possible for an ecommerce worker to attempt fulfillment of an order before the order has been committed to the database. If an ecommerce worker hits the [order fulfillment endpoint](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/v2/views/orders.py#L40) before the order has been committed to the database, the endpoint will return a 404. When the worker gets a 404 back, it thinks that it's been asked to fulfill a non-existent order and [gives up on fulfillment](https://github.com/edx/ecommerce-worker/blob/master/ecommerce_worker/fulfillment/v1/tasks.py#L33-L42).

rlucioni, line 38: For more on this race condition, see [this comment](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/checkout/mixins.py#L102-L106).

clintonb, line 38: Rather than code around this, we should use [Django 1.9's `on_commit`](https://docs.djangoproject.com/en/1.9/topics/db/transactions/#performing-actions-after-commit).

rlucioni, line 38: Sure, that seems well-suited to our needs.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
cpennington, line 231: :+1: to that.

cpennington, line 231: Although, testing against users that already have more than 1 basket is crucial as well (since that's a situation that exists now [and might exist again, if a user is buying multiple items in parallel])

ecommerce/extensions/api/v2/views/baskets.py
cpennington, line 38: Does that mean we're going to block on merging this until we upgrade to django 1.9?

clintonb, line 38: No. Haystack, used by Oscar, does not yet support Django 1.9. Also, we have a viable alternative (e.g. stick with the existing transaction logic).

ecommerce/extensions/api/v2/tests/views/test_baskets.py
bderusha, line 231: That sounds like a great test case to explicitly set up ;)

ecommerce/tests/mixins.py
clintonb, line 131: If we had multiple substitutions, yes. Since there's only one, the extra identifier isn't necessary.

ecommerce/extensions/api/v2/views/baskets.py
clintonb, line 38: The transaction bits have been restored.

rlucioni:
:+1:

PR #566, opened 2016-02-22
Added basket deletion endpoint

clintonb:
This endpoint will allow any superuser to delete a basket. It is primarily intended to be used to delete baskets with multiple items.

ECOM-3663

### DIFF
--- a/ecommerce/extensions/api/permissions.py
+++ b/ecommerce/extensions/api/permissions.py
@@ -25,3 +25,10 @@ class IsStaffOrOwner(permissions.BasePermission):
 
     def has_object_permission(self, request, view, obj):
         return request.user and (request.user.is_staff or obj.user == request.user)
+
+
+class IsSuperUser(permissions.BasePermission):
+    """ Allows access only to super users. """
+
+    def has_permission(self, request, view):
+        return request.user and request.user.is_superuser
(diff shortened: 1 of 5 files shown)


clintonb:
@bderusha please review.
@rlucioni review, if you care/have time.

bderusha:
LGTM :+1:

rlucioni:
@clintonb is the Django admin not sufficient for this because we'd like to delete baskets programmatically?

clintonb:
@rlucioni we need to delete 200+ baskets. Django admin is not a good alternative.

ecommerce/extensions/api/permissions.py
```diff
@@ -25,3 +25,10 @@ class IsStaffOrOwner(permissions.BasePermission):
 
     def has_object_permission(self, request, view, obj):
         return request.user and (request.user.is_staff or obj.user == request.user)
+
+
+class IsSuperUser(permissions.BasePermission):
```
rlucioni, line 30: Is DRF's `IsAdminUser` not sufficient? `IsAdminUser` only grants permission to staff users. Is there a real distinction to be made here between staff users and superusers?

clintonb, line 30: In an ideal users all staff users would _not_ be superusers. We want to be explicit about that distinction, even if we don't yet live in that ideal world, since we are deleting data.

rlucioni:
:+1:

PR #567, opened 2016-02-23
Updated clean_static make target

clintonb:
The contents of the static directories are deleted, but the parent directories are retained. This avoids issues with Django being unable to find ecommerce/static/build after running `make clean_static`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ clean:
 	rm -rf coverage htmlcov
 
 clean_static:
-	rm -rf assets/ ecommerce/static/build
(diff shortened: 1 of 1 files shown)


clintonb:
FYI @edx/ecommerce

rlucioni:
I'm surprised you needed to do this. What behavior were you trying to remedy?

clintonb:
See the commit message.

rlucioni:
I see, thanks.

cpennington:
Another option would be to add those directories as make targets, if the `make static` target depends on them existing.

PR #568, opened 2016-02-23
Fix for checkout page JS submit

mjfrey:
Fix for submitting hidden form.


mjfrey:
@clintonb can you take a look.  Can I also get the merge enabled as this functionality will be tested as part of the acceptance tests.

clintonb:
I have validated this locally.

PR #570, opened 2016-02-23
Added email search to order and refund dashboards

clintonb:
ECOM-2370

### DIFF
--- a/ecommerce/extensions/dashboard/orders/views.py
+++ b/ecommerce/extensions/dashboard/orders/views.py
@@ -7,6 +7,8 @@ from oscar.apps.dashboard.orders.views import (
 )
 from oscar.core.loading import get_model
 
+from ecommerce.extensions.dashboard.views import FilterFieldsMixin
+
 Order = get_model('order', 'Order')
 Partner = get_model('partner', 'Partner')
 Refund = get_model('refund', 'Refund')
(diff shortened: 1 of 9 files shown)


clintonb:
This is manually tested.

@bderusha please review.

FYI @edx/ecommerce

clintonb:
Yes, there is duplication; however, I'm not certain it is worth extracting to a mixin.

rlucioni:
We should notify the support team when this is deployed.

ecommerce/extensions/dashboard/refunds/views.py
```diff
@@ -24,12 +26,15 @@ def get_queryset(self):
...
-                    queryset = queryset.filter(status__in=value)
-                elif field == 'username' and value:
-                    queryset = queryset.filter(user__username__istartswith=value)
-                elif value:
-                    queryset = queryset.filter(**{field: value})
+                if value:
+                    _filter = self.exposed_fields.get(field)
+
+                    if field == 'status':
+                        queryset = queryset.filter(status__in=value)
```
bderusha: Is there a way in which we can have views with other filters declare them at the top to reduce a list of `if field == 'field1':..  if field == 'field2':... etc`  and instead have the top of the class define `exposed_fields['status'] = 'status__in'` or something and let the queryset filter builder just work?  Thoughts?

ecommerce/extensions/dashboard/orders/views.py
```diff
@@ -39,11 +43,16 @@ def dispatch(self, request, *args, **kwargs):
...
+        # Note (CCB): We set self.form here because the super method does not always pass request.GET
+        # to the form constructor. This results in the form not being populated when re-rendered.
+        self.form = self.form_class(self.request.GET)
+        if self.form.is_valid():
+            for field, value in self.form.cleaned_data.iteritems():
+                if value:
+                    _filter = self.exposed_fields.get(field)
+
+                    if _filter:
+                        queryset = queryset.filter(**{_filter: value})
```
bderusha: Why does this line not have this line as a catchall like the file below?

``` python
else:
    queryset = queryset.filter(**{field: value})
```

bderusha:
A couple of minor questions.  :+1: pending answers.  Also don't forget to :eggplant: commits

ecommerce/extensions/dashboard/orders/views.py
clintonb: The `RefundListView` is unique because it is completely custom. The `OrderListView` extends the existing Oscar dashboard view, which performs the base filtering.

PR #571, opened 2016-02-23
Add discount to line items

mjfrey:
Use line.line_price_incl_tax_incl_discounts so PayPal knows the proper price.


mjfrey:
@clintonb please look.  I had made this change but got lost somehow.

clintonb:
:+1:

PR #572, opened 2016-02-23
Re-apply offers in the basket so the correct total and discounts are …

mjfrey:
…recorded

Need to recalculate the basket before we record the order.

### DIFF
--- a/ecommerce/extensions/payment/views.py
+++ b/ecommerce/extensions/payment/views.py
@@ -23,6 +23,7 @@ from ecommerce.extensions.payment.processors.paypal import Paypal
 
 logger = logging.getLogger(__name__)
 
+Applicator = get_class('offer.utils', 'Applicator')
 Basket = get_model('basket', 'Basket')
 BillingAddress = get_model('order', 'BillingAddress')
 Country = get_model('address', 'Country')
@@ -70,6 +71,7 @@ class CybersourceNotifyView(EdxOrderPlacementMixin, View):
             basket_id = int(basket_id)
             basket = Basket.objects.get(id=basket_id)
             basket.strategy = strategy.Default()
+            Applicator().apply(basket, basket.owner, self.request)
             return basket
         except (ValueError, ObjectDoesNotExist):
             return None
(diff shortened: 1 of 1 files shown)


mjfrey:
@clintonb please take a look.  This also got lost somehow.  This is needed so that we record the actual discount on the order.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -38,6 +38,7 @@ def handle_payment(self, response, basket):
         events (using add_payment_event) so they can be
         linked to the order when it is saved later on.
         """
+        Applicator().apply(basket, basket.owner, self.request)
```
clintonb: Why is this done here as opposed to earlier in the request lifecycle?

mjfrey: It is done before payment and must be done here as well once payment is received by the processor. Offers and discounts are not persistent in the basket model. If we don't do this here the discount will not be reflected in the recorded order.

clintonb: This belongs in the `_get_basket()` methods of the payment views ([here](https://github.com/edx/ecommerce/blob/c8f4fd99b8ba3e989e7ecd9ac0e9ebe89cb5a3c0/ecommerce/extensions/payment/views.py#L65-L75) and [here](https://github.com/edx/ecommerce/blob/c8f4fd99b8ba3e989e7ecd9ac0e9ebe89cb5a3c0/ecommerce/extensions/payment/views.py#L182-L206)). The offers should be applied as early in the request lifecycle as possible to avoid other parts of the code getting invalid baskets.

mjfrey:
@clintonb ok thanks.  Moved earlier to the processor _get_basket methods.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -9,7 +9,6 @@
 
 from ecommerce.extensions.analytics.utils import audit_log
 
-
```
clintonb, line 12: Revert.

clintonb:
:+1:

PR #573, opened 2016-02-24
Coupon test mixin

vkaracic:
New `CouponMixin` test mixin for containing all helper functions for coupon tests. Currently only contains `create_coupon` helper method for creating a coupon. In the future it will have a `setUp` method which creates a default coupon category which will be needed for creating coupons.

@mjfrey @clintonb


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=573&repo=edx%2Fecommerce

mjfrey:
@rlucioni @bderusha Can you take a look?

clintonb:
:+1:

bderusha:
:+1:

PR #575, opened 2016-02-24
Reorganized acceptance test page classes

clintonb:
The page classes are now grouped in sub-packages by system/IDA.


clintonb:
FYI @edx/ecommerce @mjfrey @marjev @vkaracic

PR #578, opened 2016-02-25
Do not check order status as it may take longer.

mjfrey:
Just take the user to the dashboard after enrollment.  Order fulfillment is async and could take longer.


mjfrey:
@vkaracic

vkaracic:
What happens in case an order fails?

vkaracic:
:+1:

PR #579, opened 2016-02-25
Coupon category

mjfrey:
Adds Category attribute to Coupons.
Also implements the front end UI changes needed to support Categories.

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -14,13 +14,13 @@ import waffle
 from ecommerce.core.constants import ISO_8601_FORMAT, COURSE_ID_REGEX
 from ecommerce.courses.models import Course
 
-
 logger = logging.getLogger(__name__)
 
 Basket = get_model('basket', 'Basket')
 Benefit = get_model('offer', 'Benefit')
 BillingAddress = get_model('order', 'BillingAddress')
 Catalog = get_model('catalogue', 'Catalog')
+Category = get_model('catalogue', 'Category')
 Line = get_model('order', 'Line')
 Order = get_model('order', 'Order')
 Product = get_model('catalogue', 'Product')
@@ -109,9 +109,8 @@ class ProductSerializer(ProductPaymentInfoMixin, serializers.HyperlinkedModelSer
 
     def get_attribute_values(self, product):
         request = self.context.get('request')
-        attributes = product.attr
         serializer = ProductAttributeValueSerializer(
-            attributes,
+            product.attr,
             many=True,
             read_only=True,
             context={'request': request}
@@ -379,10 +378,17 @@ class VoucherSerializer(serializers.ModelSerializer):
     class Meta(object):
         model = Voucher
         fields = (
-            'id', 'name', 'code', 'redeem_url', 'usage', 'start_datetime', 'end_datetime',
-            'num_basket_additions', 'num_orders', 'total_discount',
-            'date_created', 'offers', 'is_available_to_user', 'benefit'
+            'id', 'name', 'code', 'redeem_url', 'usage', 'start_datetime', 'end_datetime', 'num_basket_additions',
+            'num_orders', 'total_discount', 'date_created', 'offers', 'is_available_to_user', 'benefit'
         )
+
+
+class CategorySerializer(serializers.ModelSerializer):
+    # NOTE (CCB): We are explicitly ignoring child categories. They are not relevant to our current needs. Support
+    # should be added later, if needed.
+
+    class Meta(object):
+        model = Category
 
 
 class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
@@ -392,17 +398,18 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     seats = serializers.SerializerMethodField()
     client = serializers.SerializerMethodField()
     vouchers = serializers.SerializerMethodField()
+    categories = CategorySerializer(many=True, read_only=True)
 
     def get_coupon_type(self, obj):
         voucher = obj.attr.coupon_vouchers.vouchers.first()
         benefit = voucher.offers.first().benefit
         if benefit.type == Benefit.PERCENTAGE and benefit.value == 100:
-            return "Enrollment code"
-        return "Discount code"
+            return 'Enrollment code'
+        return 'Discount code'
 
     def get_last_edited(self, obj):
         history = obj.history.latest()
-        return (history.history_user.username, history.history_date)
+        return history.history_user.username, history.history_date
 
     def get_seats(self, obj):
         voucher = obj.attr.coupon_vouchers.vouchers.first()
@@ -421,7 +428,7 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
 
     class Meta(object):
         model = Product
-        fields = ('id', 'title', 'coupon_type', 'last_edited', 'seats', 'client', 'price', 'vouchers',)
+        fields = ('id', 'title', 'coupon_type', 'last_edited', 'seats', 'client', 'price', 'vouchers', 'categories',)
 
 
 class CheckoutSerializer(serializers.Serializer):  # pylint: disable=abstract-method
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -5,21 +5,24 @@ import json
 from django.core.urlresolvers import reverse
 from django.db.utils import IntegrityError
 from django.test import RequestFactory
+from oscar.apps.catalogue.categories import create_from_breadcrumbs
 from oscar.core.loading import get_model
 
 from ecommerce.core.models import Client
 from ecommerce.extensions.api.constants import APIConstants as AC
 from ecommerce.extensions.api.v2.views.coupons import CouponViewSet
 from ecommerce.tests.factories import SiteConfigurationFactory, SiteFactory
-from ecommerce.tests.mixins import CouponMixin
+from ecommerce.tests.mixins import CouponMixin, ThrottlingMixin
 from ecommerce.tests.testcases import TestCase
 
 Basket = get_model('basket', 'Basket')
 Benefit = get_model('offer', 'Benefit')
 Catalog = get_model('catalogue', 'Catalog')
+Category = get_model('catalogue', 'Category')
 Course = get_model('courses', 'Course')
 Order = get_model('order', 'Order')
 Product = get_model('catalogue', 'Product')
+ProductCategory = get_model('catalogue', 'ProductCategory')
 ProductClass = get_model('catalogue', 'ProductClass')
 StockRecord = get_model('partner', 'StockRecord')
 Voucher = get_model('voucher', 'Voucher')
@@ -51,6 +54,7 @@ class CouponViewSetTest(CouponMixin, TestCase):
             'quantity': 2,
             'start_date': '2015-1-1',
             'voucher_type': Voucher.MULTI_USE,
+            'categories': [self.category]
         }
 
     def test_create(self):
@@ -64,6 +68,7 @@ class CouponViewSetTest(CouponMixin, TestCase):
             'stock_record_ids': [1],
             'voucher_type': Voucher.SINGLE_USE,
             'price': 100,
+            'category_ids': [self.category.id]
         })
         request = RequestFactory()
         request.data = self.coupon_data
@@ -90,6 +95,8 @@ class CouponViewSetTest(CouponMixin, TestCase):
         self.assertEqual(stock_record.price_excl_tax, 100)
 
         self.assertEqual(coupon.attr.coupon_vouchers.vouchers.count(), 5)
+        category = ProductCategory.objects.get(product=coupon).category
+        self.assertEqual(category, self.category)
 
     def test_append_to_existing_coupon(self):
         """Test adding additional vouchers to an existing coupon."""
@@ -176,7 +183,7 @@ class CouponViewSetTest(CouponMixin, TestCase):
         self.assertEqual(Basket.objects.first().status, 'Submitted')
 
 
-class CouponViewSetFunctionalTest(CouponMixin, TestCase):
+class CouponViewSetFunctionalTest(CouponMixin, ThrottlingMixin, TestCase):
     """Test the coupon order creation functionality."""
 
     def setUp(self):
@@ -203,6 +210,7 @@ class CouponViewSetFunctionalTest(CouponMixin, TestCase):
             'voucher_type': Voucher.SINGLE_USE,
             'quantity': 2,
             'price': 100,
+            'category_ids': [self.category.id]
         }
         self.response = self.client.post(COUPONS_LINK, data=self.data, format='json')
 
@@ -243,8 +251,7 @@ class CouponViewSetFunctionalTest(CouponMixin, TestCase):
         """Test that the endpoint returns information needed for the details page."""
         response = self.client.get(COUPONS_LINK)
         self.assertEqual(response.status_code, 200)
-        response_data = json.loads(response.content)
-        coupon_data = response_data['results'][0]
+        coupon_data = json.loads(response.content)['results'][0]
         self.assertEqual(coupon_data['title'], 'Test coupon')
         self.assertEqual(coupon_data['coupon_type'], 'Enrollment code')
         self.assertIsNotNone(coupon_data['last_edited'][0])
@@ -301,3 +308,22 @@ class CouponViewSetFunctionalTest(CouponMixin, TestCase):
         self.assertEqual(new_coupon.attr.coupon_vouchers.vouchers.last().start_datetime.year, 2030)
         self.assertEqual(new_coupon.attr.coupon_vouchers.vouchers.first().end_datetime.year, 2035)
         self.assertEqual(new_coupon.attr.coupon_vouchers.vouchers.last().end_datetime.year, 2035)
+
+
+class CouponCategoriesListViewTests(TestCase):
+    """ Tests for the coupon category list view. """
+    path = reverse('api:v2:coupons:coupons_categories')
+
+    def setUp(self):
+        super(CouponCategoriesListViewTests, self).setUp()
+        self.user = self.create_user()
+        self.client.login(username=self.user.username, password=self.password)
+        Category.objects.all().delete()
+        create_from_breadcrumbs('Coupons > Coupon test category')
+
+    def test_category_list(self):
+        """ Verify the endpoint returns successfully. """
+        response = self.client.get(self.path)
+        response_data = json.loads(response.content)
+        self.assertEqual(response_data['count'], 1)
+        self.assertEqual(response_data['results'][0]['name'], 'Coupon test category')
--- a/ecommerce/extensions/api/v2/urls.py
+++ b/ecommerce/extensions/api/v2/urls.py
@@ -40,6 +40,7 @@ REFUND_URLS = [
 
 COUPON_URLS = [
     url(r'^coupon_reports/(?P<coupon_id>[\d]+)/$', CouponReportCSVView.as_view(), name='coupon_reports'),
+    url(r'^categories/$', coupon_views.CouponCategoriesListView.as_view(), name='coupons_categories'),
 ]
 
 CHECKOUT_URLS = [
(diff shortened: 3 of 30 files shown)


mjfrey:
@bderusha Can you please take a look.

ecommerce/extensions/api/serializers.py
```diff
@@ -1,7 +1,8 @@
 """Serializers for data manipulated by ecommerce API endpoints."""
-from decimal import Decimal
 import logging
+from decimal import Decimal
```
bderusha, line 2: Why did this move?  Seems like it was in the correct place alphabetically

ecommerce/extensions/api/serializers.py
```diff
@@ -1,7 +1,8 @@
...
 
+import waffle
```
bderusha: same here

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -51,6 +54,7 @@ def setUp(self):
             'quantity': 2,
             'start_date': '2015-1-1',
             'voucher_type': Voucher.MULTI_USE,
+            'category': self.category,
```
bderusha: Why is this `category` but below `category.id`. Can we pick one and just use the default?

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -90,6 +95,8 @@ def test_create_coupon_product(self):
         self.assertEqual(stock_record.price_excl_tax, 100)
 
         self.assertEqual(coupon.attr.coupon_vouchers.vouchers.count(), 5)
+        category = ProductCategory.objects.get(product=coupon).category
+        self.assertEqual(category.name, self.category.name)
```
bderusha: should these objects match?  Can we `assertEqual(category, self.category)`?

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -271,6 +279,22 @@ def test_list_discount_coupons(self):
         self.assertEqual(response_data['results'][0]['coupon_type'], 'Discount code')
         self.assertEqual(response_data['results'][0]['vouchers'][0]['benefit'][1], 20.0)
 
+    def test_existing_coupon_category_change(self):
+        """ Test that the existing product category has been edited on change of category. """
+        response_data = json.loads(self.response.content)
+        coupon = Product.objects.get(id=response_data['coupon_id'])
+        self.assertEqual(ProductCategory.objects.filter(product=coupon).count(), 1)
+        product_category = ProductCategory.objects.get(product=coupon)
+        self.assertEqual(product_category.category.name, self.category.name)
```
bderusha: same question about asserting if objects are equal

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -271,6 +279,22 @@ def test_list_discount_coupons(self):
         self.assertEqual(response_data['results'][0]['coupon_type'], 'Discount code')
         self.assertEqual(response_data['results'][0]['vouchers'][0]['benefit'][1], 20.0)
 
+    def test_existing_coupon_category_change(self):
```
bderusha: Can we add a few inline comments here about what these two sections of code are testing.

ecommerce/extensions/api/v2/urls.py
```diff
@@ -62,6 +62,8 @@
     url(r'^payment/', include(PAYMENT_URLS, namespace='payment')),
     url(r'^refunds/', include(REFUND_URLS, namespace='refunds')),
     url(r'^publication/', include(ATOMIC_PUBLICATION_URLS, namespace='publication')),
+    url(r'^coupons/categories/$', coupon_views.CouponCategoriesListView.as_view(), name='coupons-categories'),
```
bderusha: please use the `include` pattern to be consistent with the rest of this file

ecommerce/extensions/api/v2/urls.py
```diff
@@ -62,6 +62,8 @@
...
+
```
bderusha: remove bonus space

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -7,18 +7,16 @@
...
-from rest_framework import status
+from rest_framework import generics, status
 from rest_framework.permissions import IsAdminUser, IsAuthenticated
 from rest_framework.response import Response
 
 from ecommerce.core.models import Client
 from ecommerce.extensions.api import data as data_api
 from ecommerce.extensions.api.constants import APIConstants as AC
-from ecommerce.extensions.api.serializers import CouponSerializer
+from ecommerce.extensions.api.serializers import CouponSerializer, CategorySerializer
```
bderusha: NIT: Alphabetize serializers

ecommerce/static/js/apps/coupon_admin_app.js
```diff
@@ -1,17 +1,26 @@
 require([
         'backbone',
         'routers/coupon_router',
-        'utils/navigate'
+        'utils/navigate',
+        'ecommerce',
+        'collections/category_collection'
```
bderusha: alphabetize

ecommerce/static/js/models/coupon_model.js
```diff
@@ -1,21 +1,28 @@
 define([
         'backbone',
+ //       'backbone.relational',
```
bderusha: if this is not required anymore let's remove it

ecommerce/static/js/models/coupon_model.js
```diff
@@ -1,21 +1,28 @@
...
+//              BackboneRelational,
               BackboneSuper,
               BackboneValidation,
               $,
               $cookie,
               _,
               moment
-            ) {
+              //   ,
+              // CategoryCollection,
```
bderusha: Same with these

ecommerce/static/js/models/coupon_model.js
```diff
@@ -84,6 +92,15 @@ define([
...
             },
 
+            // relations: [{
+            //     collectionType: CategoryCollection,
+            //     type: Backbone.HasMany,
+            //     key: 'category',
+            //     relatedModel: Category,
+            //     includeInJSON: false,
+            //     parse: true
+            // }],
```
bderusha: Same here

ecommerce/static/js/test/specs/models/coupon_model_spec.js
```diff
@@ -49,7 +49,8 @@ define([
                 course_id: 'a/b/c',
                 seat_type: 'verified',
                 course: course,
-                price: 100
+                price: 100,
+                category: 4
```
bderusha, line 53: Again, I'm having some confusion about whether this should be `category` versus `category_id`.  What should this be?

ecommerce/static/js/test/specs/utils/utils_spec.js
```diff
@@ -10,6 +10,169 @@ define([
               CouponFormView) {
         'use strict';
 
+        var categories = [
```
bderusha: can we pull this data out and share it across the tests?

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -12,6 +12,7 @@ define([
         'text!templates/coupon_form.html',
         'models/course_model',
         'views/form_view',
+        'ecommerce'
```
bderusha: alphabetize

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -23,9 +24,12 @@ define([
               Utils,
               CouponFormTemplate,
               Course,
-              FormView) {
+              FormView,
+              ecommerce) {
         'use strict';
 
+        window.ecommerce = ecommerce;
```
bderusha: Do we _have_ to attach this to the global namespace?

ecommerce/tests/mixins.py
```diff
@@ -335,6 +340,7 @@ def create_coupon(
             'quantity': quantity,
             'start_date': datetime.date(2015, 1, 1),
             'voucher_type': Voucher.SINGLE_USE,
+            'category': self.category
```
bderusha: and here: category VS category_id

bderusha:
@mjfrey @vkaracic done with initial review.  Ping me when these are addressed/responded to and I will take a final look.

ecommerce/static/js/test/specs/models/coupon_model_spec.js
mjfrey, line 53: The front end UI deals in id the backend is category.

ecommerce/tests/mixins.py
mjfrey: yes this is correct.  We use the category not the id in the backend.

ecommerce/extensions/api/v2/urls.py
```diff
@@ -40,6 +40,7 @@
 
 COUPON_URLS = [
     url(r'^coupon_reports/(?P<coupon_id>[\d]+)/$', CouponReportCSVView.as_view(), name='coupon_reports'),
+    url(r'^categories/$', coupon_views.CouponCategoriesListView.as_view(), name='coupons-categories'),
```
bderusha: NIT: underscore between multi-word names `coupons_categories`

ecommerce/static/js/apps/coupon_admin_app.js
```diff
@@ -1,17 +1,26 @@
 require([
         'backbone',
+        'collections/category_collection',
```
bderusha, line 3: When reordering the dependencies you must also reorder them in the function definition below.

ecommerce/static/js/apps/coupon_admin_app.js
```diff
@@ -1,17 +1,26 @@
...
+        'ecommerce',
         'routers/coupon_router',
-        'utils/navigate'
+        'utils/navigate',
     ],
     function (Backbone,
               CouponRouter,
```
bderusha: here ^^

mjfrey: nice catch.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -23,9 +24,12 @@ define([
               Utils,
               CouponFormTemplate,
               Course,
-              FormView) {
+              FormView,
+              ecommerce) {
```
bderusha: Same here.  Reordering dependency, we also need to reorder it here.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
vkaracic: The `coupon_data` in this particular test class uses the actual `category` object because here we unit test the methods within the `CouponViewSet`. The `test_create` one below tests the `create` method which receives the request data in which a category ID is sent and the category object is immediately retrieved: 
https://github.com/edx/ecommerce/blob/clintonb/coupon-category/ecommerce/extensions/api/v2/views/coupons.py#L78
and sent to the `create_coupon_product` method, which is tested in most of the rest unit tests in this class.

I do agree that we might need to change the keyword from `category` to `category_id` in the request data, but everything else is as it should be.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
vkaracic: I'm sorry I have a nasty habit of comparing object attributes. I'll stop. :sweat_smile:

vkaracic:
Hey @bderusha thanks for reviewing. I addressed a couple of comments and left an explanation for the category objects/id. Please take a look when you get a chance.

ecommerce/static/js/models/coupon_model.js
bderusha: @vkaracic bump ^

ecommerce/static/js/test/specs/utils/utils_spec.js
bderusha: @vkaracic bump ^

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -26,6 +28,8 @@ define([
               FormView) {
         'use strict';
 
+        window.ecommerce = ecommerce;
```
bderusha: @vkaracic is there a way to implement this without adding `ecommerce` to the global namespace?  It seems like the only reason we are doing it is to support tests. Is there another way to mock the `selectOptions` `collection` method rather than the `ecommerce` object

ecommerce/extensions/api/v2/tests/views/test_coupons.py
bderusha: @vkaracic we need to update the request data to reflect that it's a `category_id` where ever that's what the data actually is.

bderusha:
:+1:  Pending:
- [x] update the request data to use `category_id` where appropriate
- [x] remove commented out dependency
- [x] DRY up categories data in tests (so we don't need to make changes in 3 places)
- [x] get `ecommerce` out of the global name space if possible

ecommerce/static/js/test/specs/utils/utils_spec.js
swpdsgn: You do realize that mock data like this is not DRYed up thruout the codebase? At least for the JS part. Just go thru other specs and you'll see.

ecommerce/static/js/views/coupon_form_view.js
swpdsgn: You are correct its cuase of the tests. Last time i did the category work there was no need for this, however Clinton made a new branch and changed some code, and this was the fastest and best way to get it working. In order to remove commerce from the global namespace i would need to rewrite most of the stuff he did.

clintonb: Why not mock `ecommerce` in your tests?

clintonb: See `ecommerce.credit.providers` in #439.

ecommerce/extensions/api/serializers.py
```diff
@@ -392,17 +398,19 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     seats = serializers.SerializerMethodField()
     client = serializers.SerializerMethodField()
     vouchers = serializers.SerializerMethodField()
+    category = serializers.SerializerMethodField()
+    CategorySerializer(many=True, read_only=True)
```
clintonb: This line does nothing.

ecommerce/extensions/api/serializers.py
```diff
@@ -392,17 +398,19 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     seats = serializers.SerializerMethodField()
     client = serializers.SerializerMethodField()
     vouchers = serializers.SerializerMethodField()
+    category = serializers.SerializerMethodField()
```
clintonb: I highly recommend sticking with multiple categories, as I originally wrote. If you only want to _display_ a single category, do so; but, don't limit this functionality at the API level. It is far easier to change the UI than the API (which would require a new version to be release, and other code to be deprecated). We must be cognizant of the future ramifications of changes like this.

clintonb: Also, the closer we stick to the original Oscar data models/API, the easier it is for open source users to extend this functionality.

vkaracic: I've been told that we're going to have only one category, I can change that to accept multiple categories but I'd like for you guys to give me a definite decision so that I don't go back and forth.

@mjfrey @pbaruah

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -271,6 +279,27 @@ def test_list_discount_coupons(self):
...
+        response_data = json.loads(self.response.content)
+        coupon = Product.objects.get(id=response_data['coupon_id'])
+        self.assertEqual(ProductCategory.objects.filter(product=coupon).count(), 1)
+        product_category = ProductCategory.objects.get(product=coupon)
+        self.assertEqual(product_category.category, self.category)
+
+        # Create a new category and post the same data but with new category.
+        # Verify that the coupon still has only one category associated with it,
+        # and the category is the new one.
+        new_category = Category.objects.create(name="A new category", depth=2)
```
clintonb: Use `CategoryFactory`.

ecommerce/extensions/api/serializers.py
clintonb: UI design should not necessarily dictate the API. Given the data supports multiple categories, the API should offer such functionality. Again, we may only expose one category _today_, but that may not be the case tomorrow.

ecommerce/extensions/catalogue/migrations/0015_default_categories.py
```diff
@@ -0,0 +1,38 @@
...
+    'Customer Service', 'Financial Assistance', 'Geography Promotion',
+    'Marketing Partner Promotion', 'Marketing-Other', 'Paid Cohort', 'Other',
+    'Retention Promotion', 'Services-Other', 'Support-Other', 'Upsell Promotion',
+]
+
+
+def create_default_categories(apps, schema_editor):
+    """Create default coupon categories."""
+    for category in DEFAULT_CATEGORIES:
+        create_from_breadcrumbs('Coupons > %s' % category)
```
clintonb: Use `str.format()`

ecommerce/extensions/catalogue/migrations/0015_default_categories.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def remove_default_categories(apps, schema_editor):
+    """Remove default coupon categories."""
+    for category in DEFAULT_CATEGORIES:
```
clintonb: `Category.objects.get(name='Coupons').get_children().filter(name__in= DEFAULT_CATEGORIES, depth=).delete()`

You will want to move `'Coupons'` to a constant shared across both functions.

ecommerce/static/js/test/specs/mock_data/categories.js
```diff
@@ -0,0 +1,168 @@
+define([], function(){
```
clintonb: Only specs should be in the specs directory. Move this up one level.

ecommerce/static/js/test/specs/utils/utils_spec.js
```diff
@@ -2,12 +2,14 @@ define([
         'backbone',
         'test/spec-utils',
         'utils/utils',
-        'views/coupon_form_view'
+        'views/coupon_form_view',
+        '../mock_data/categories.js'
```
clintonb: `'test/mock_data/categories'`

ecommerce/static/js/test/specs/mock_data/categories.js
swpdsgn: will do

ecommerce/static/js/views/coupon_form_view.js
swpdsgn: Did that finally

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -75,7 +75,7 @@ def create(self, request, *args, **kwargs):
             quantity = request.data[AC.KEYS.QUANTITY]
             price = request.data[AC.KEYS.PRICE]
             partner = request.site.siteconfiguration.partner
-
+            categories = Category.objects.filter(pk__in=request.data['category_ids'])
```
bderusha: `category_ids` plural here seems semantically correct, but we are still referencing `category_id` everywhere else.

ecommerce/static/js/models/coupon_model.js
```diff
@@ -138,18 +141,19 @@ define([
                 data.start_date = moment.utc(this.get('start_date'));
                 data.end_date = moment.utc(this.get('end_date'));
                 data.price = this.get('price');
+                data.category_id = this.get('category');
```
bderusha: If this is actually an ID then we should be calling `this.get('category_id');` since that should also be an ID

ecommerce/extensions/api/v2/views/coupons.py
vkaracic: We're still working on that implementation :)

ecommerce/static/js/test/specs/models/coupon_model_spec.js
bderusha, line 53: This is still confusing, probably because of the inconsistent use of `category` and `category_id`.  But if this is a category object can we please update the mock to reflect that as well.  Something like `category: { id: 4 }` would suffice.

ecommerce/extensions/api/v2/views/coupons.py
bderusha: Ah sorry I jumped the gun!

bderusha:
One small open question about updating the `discount.category` attribute in `coupon_model_spec.js`.

The rest looks good :+1: once tests are passing

bderusha:
and of course please squash commits :eggplant:

PR #580, opened 2016-02-26
Aamir khan/ecom 3661 update order script

aamir-khan:
Added the script to update the order flagged in Feb audit ECOM-3661.


clintonb:
This should not be committed to the repo. One-off scripts should remain private (e.g. private gist).

aamir-khan:
Closing and moving the script to private gist.

PR #584, opened 2016-02-29
Fix quotes

mjfrey:
Remove smart quotes.


mjfrey:
@clintonb

mjfrey:
@pbaruah

pbaruah:
:+1:

PR #585, opened 2016-03-01
Fixed credit provider bug

clintonb:
The credit seat details view now uses a specific attribute for the credit provider display name, rather than overriding the credit_provider attribute (and breaking editing).

ECOM-3659

### DIFF
--- a/ecommerce/static/js/models/course_seats/credit_seat.js
+++ b/ecommerce/static/js/models/course_seats/credit_seat.js
@@ -1,7 +1,9 @@
 define([
-        'models/course_seats/course_seat'
+        'models/course_seats/course_seat',
+        'ecommerce'
     ],
-    function (CourseSeat) {
+    function (CourseSeat,
+              ecommerce) {
         'use strict';
 
         return CourseSeat.extend({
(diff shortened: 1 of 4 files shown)


clintonb:
@Ayub-Khan @bderusha @awaisdar001  please review.

awaisdar001:
@bderusha  Can you have a look at this first, I will review after you. 
Thanks.

ecommerce/static/js/models/course_seats/credit_seat.js
```diff
@@ -17,9 +19,14 @@ define([
 
             validation: _.extend({}, CourseSeat.prototype.validation,
                 {
-                    credit_provider: {
-                        required: true,
-                        msg: gettext('All credit seats must have a credit provider.')
+                    credit_provider: function (value) {
```
awaisdar001, line 22: Is this called while saving the model?

awaisdar001:
Just a question, but this looks good to me.
:+1:

ecommerce/static/js/models/course_seats/credit_seat.js
clintonb, line 22: It can be; however, we actually validate every time a field is changed.

PR #587, opened 2016-03-01
Lines should be empty for non-existing baskets.

vkaracic:
We got this!

In case a user goes to `/basket` without adding anything to the basket `line_list` would not exist in the context. This change adds a default empty list in that case. 
https://openedx.atlassian.net/browse/SOL-1675

### DIFF
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -181,8 +181,15 @@ class BasketSummaryViewTests(LmsApiMockMixin, TestCase):
 
         response = self.client.get(self.path)
         self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.context['lines']), 1)
         self.assertEqual(response.context['payment_processors'][0].NAME, DummyProcessor.NAME)
         self.assertEqual(json.loads(response.context['footer']), {'footer': 'edX Footer'})
+
+    def test_no_basket_response(self):
+        """ Verify there are no lines in the context for a non-existing basket. """
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.context['lines'], [])
 
     def test_line_item_discount_data(self):
         """ Verify that line item has correct discount data. """
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -63,7 +63,7 @@ class BasketSummaryView(BasketView):
     """
     def get_context_data(self, **kwargs):
         context = super(BasketSummaryView, self).get_context_data(**kwargs)
-        lines = context['line_list']
+        lines = context.get('line_list', [])
         api = EdxRestApiClient(get_lms_url('api/courses/v1/'))
         for line in lines:
             course_id = line.product.course_id


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=587&repo=edx%2Fecommerce

vkaracic:
@mjfrey @marjev

clintonb:
The empty basket page needs better styling.
screen shot 2016-03-01 at 9 05 10 am

If the cause of this issue is users clicking the back button, how will you account for that problem?

mjfrey:
@clintonb when the user clicks checkout the basket is frozen.  Is it not correct to have an empty basket at that point?

clintonb:
Yes, but if the user cancels/clicks the back button she expects to see the basket she was just viewing. The user knows nothing about freezing/thawing baskets. She simply knows she wants to go back to view the basket.

This was not an issue until the switch, because the back button took the user to a page on LMS that didn't care about baskets. Now that we are displaying the basket, we need to account for this user behavior. See [this discussion thread](https://groups.google.com/forum/#!msg/django-oscar/B_VRAMmGWnA/1EOj31n9bgAJ) for a potential solution.

pbaruah:
@vkaracic - Let's change it to "Your basket is empty. Please select a course and try again". Select a course can be linked to the course search page.

If it is a small effort to address the issue properly as suggested by @clintonb, let's do that. Else we can do this fix and then address it as a high priority fix immediately.

We can use the styling for error messages with the red border on top and proper header and footer. This needs to be addressed for all error messages.

mjfrey:
@clintonb If we did in fact check for a submitted basket in the summary view, how would we know if it had not actually been a complete purchase?

clintonb:
Baskets are only frozen at payment time. They are not submitted until payment has been received, and an order placed.

You need to account for two scenarios for frozen baskets:
1. User clicks the cancel button on the payment page. We control where the user goes after clicking the cancel button. When the user clicks cancel she should be sent to a page that thaws the basket before redirecting back to the basket summary view. This is described in the [discussion thread](https://groups.google.com/forum/#!msg/django-oscar/B_VRAMmGWnA/1EOj31n9bgAJ).
2. User clicks back button. There is not much we can do about this. Adding a message below the checkout buttons asking the user to _not_ use the back button will help. You can also add a note on the empty basket page explaining that the basket may be empty because the user clicked the back button, instead of cancel.

vkaracic:
AFAIK, only baskets which are paid for are of status `Submitted`, correct me if I'm wrong. We can account for when a user clicks the cancel button the payment provider by setting a cancel URL in the processor's configuration. Not sure about the back button though

mjfrey:
Currently if the user clicks cancel we send them to a page saying such.  Is this not sufficient?

I think we should just change the text on the checkout back regarding the back button.

clintonb:
Say you have a basket at Amazon. You go through the process of picking your items, get to payment and realize you forgot something. You click the cancel button, and all of your basket items have disappeared. You now have to go back and recreate your basket. You probably wouldn't shop at Amazon very often.

Although we only sell one item per basket now, we want to make every part of the checkout process as efficient as possible. If a user mistakenly clicks PayPal, we don't want to punish her by forcing her to recreate the basket. We want to meet the user's expectation (that the basket is still present) and allow her to select another payment provider, review the basket, or do whatever prompted the user to cancel.

pbaruah:
I agree, clicking on cancel or back to take the user back to the checkout page.

vkaracic:
I can look through the available code for the payment processors that we use and see if there is a way that they account for the user going back. 
Worst case scenario, and from the top of my head, we can add an additional basket status, something like pre-submitted, which we can use for situations like this.
That being said, this PR addresses the bug that was linked up there. If we need/want additional changes let's log tickets for them with details, I don't want to add changes that are out of scope of the ticket this PR addresses.

clintonb:
Again, please see the django-oscar-paypal code in the linked discussion thread. This problem has already been solved. There is no need for additional basket states.

vkaracic:
If you're referring to this, which I gather is the class mentioned in that discussion: https://github.com/django-oscar/django-oscar-paypal/blob/master/paypal/express/views.py#L151

I believe this is used for when a user cancels a payment, not for when the back button was hit. I was talking about that use-case when I mentioned additional basket state.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -181,9 +181,16 @@ def test_response_success(self, benefit_type, benefit_value):
...
         self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.context['lines']), 1)
         self.assertEqual(response.context['payment_processors'][0].NAME, DummyProcessor.NAME)
         self.assertEqual(json.loads(response.context['footer']), {'footer': 'edX Footer'})
 
+    def test_no_basket_response(self):
+        """ Verify there are no lines in the context for a non-existing basket. """
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.context['lines']), 0)
```
clintonb: `self.assertListEqual(response.context['lines'], [])`

clintonb:
:+1:

mjfrey:
Created ticket SOL-1679 to address the cancel use case.

PR #588, opened 2016-03-01
[SOL-1658] Display benefit value correctly

marjev:
@mjfrey and @vkaracic, please review the code.

[JIRA ticket](https://openedx.atlassian.net/browse/SOL-1658)

### DIFF
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -17,6 +17,7 @@ from testfixtures import LogCapture
 
 from ecommerce.core.tests import toggle_switch
 from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.extensions.offer.utils import format_benefit_value
 from ecommerce.extensions.payment.tests.processors import DummyProcessor
 from ecommerce.extensions.test.factories import prepare_voucher
 from ecommerce.settings import get_lms_url
@@ -179,9 +180,13 @@ class BasketSummaryViewTests(LmsApiMockMixin, TestCase):
         self.mock_course_api_response(self.course)
         self.mock_footer_api_response()
 
+        benefit, __ = Benefit.objects.get_or_create(type=benefit_type, value=benefit_value)
+
         response = self.client.get(self.path)
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.context['lines']), 1)
+        self.assertEqual(response.context['lines'][0].benefit_value, format_benefit_value(benefit))
+        self.assertEqual(response.context['lines'][0].has_discount, True)
         self.assertEqual(response.context['payment_processors'][0].NAME, DummyProcessor.NAME)
         self.assertEqual(json.loads(response.context['footer']), {'footer': 'edX Footer'})
 
@@ -208,10 +213,10 @@ class BasketSummaryViewTests(LmsApiMockMixin, TestCase):
         basket.add_product(seat_without_benefit, 1)
 
         response = self.client.get(self.path)
-        self.assertEqual(response.context['lines'][0].discount_percentage, 50)
+        self.assertEqual(response.context['lines'][0].benefit_value, '50%')
         self.assertEqual(response.context['lines'][0].has_discount, True)
 
-        self.assertEqual(response.context['lines'][1].discount_percentage, 0)
+        self.assertEqual(response.context['lines'][1].benefit_value, None)
         self.assertEqual(response.context['lines'][1].has_discount, False)
 
     def test_cached_course(self):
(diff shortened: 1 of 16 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=588&repo=edx%2Fecommerce

ecommerce/extensions/offer/tests/test_utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def test_format_benefit_value(self):
+        """ format_benefit_value(benefit) should format benefit value based on benefit type """
+        benefit_value = format_benefit_value(self.percentage_benefit)
+        self.assertEqual(benefit_value, '35%')
+
+        benefit_value = format_benefit_value(self.value_benefit)
+        self.assertEqual(benefit_value, currency(self.seat_price / 3))
+
+    def test_remove_exponent_and_trailing_zeros(self):
```
vkaracic: You could use `ddt` for these tests to not repeat yourself.

ecommerce/extensions/offer/utils.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Remove exponent and trailing zeros.
+
+    Arguments:
+        decimal (Decimal): Decimal number that needs to be modified
+
+    Returns:
+        decimal (Decimal): Modified decimal number without exponent and trailing zeros.
+    """
+    return decimal.quantize(Decimal(1)) if decimal == decimal.to_integral() else decimal.normalize()
```
vkaracic, line 21: In my opinion we should limit this to two decimal places. This code would, for a highly unlikely number of `5.0000100` return `5.0001`, I these cases I suggest using only 2 decimal places and convert it just `5`. Something like this:

```
>>> "%.3g" % D('1.111111')
'1.11'
>>> "%.3g" % D('1.01')
'1.01'
>>> "%.3g" % D('1.001')
'1'
>>> "%.3g" % D('1.00')
'1'
>>> "%.3g" % D('1.34')
'1.34' 
```

@pbaruah from a product standpoint would it be OK that we limit ourselves to only two decimal places?
EDIT: never mind the example, it's a wrong usage.

marjev, line 21: Benefit value is already limited to two decimal places. This utility method simply removes exponent and trailing zeros because, for some reason, that's not the default behaviour of Decimal values.

https://github.com/django-oscar/django-oscar/blob/9058df1a8f6086f38bee758296433d67993dadd3/src/oscar/apps/offer/abstract_models.py#L413

vkaracic, line 21: Ahh, correct.

vkaracic:
:+1:

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -208,10 +208,10 @@ def test_line_item_discount_data(self):
         basket.add_product(seat_without_benefit, 1)
 
         response = self.client.get(self.path)
-        self.assertEqual(response.context['lines'][0].discount_percentage, 50)
+        self.assertEqual(response.context['lines'][0].benefit_value, '50%')
```
mjfrey, line 216: Should we be also testing a fixed dollar discount?

ecommerce/extensions/offer/tests/test_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    def setUp(self):
+        super(UtilTests, self).setUp()
+        self.course = CourseFactory()
+        self.verified_seat = self.course.create_or_update_seat('verified', False, 100, self.partner)
+        self.stock_record = StockRecord.objects.filter(product=self.verified_seat).first()
+        self.seat_price = self.stock_record.price_excl_tax
+        self._range = RangeFactory(products=[self.verified_seat, ])
+
+        self.percentage_benefit = BenefitFactory(type=Benefit.PERCENTAGE, range=self._range, value=35.00)
+        self.value_benefit = BenefitFactory(type=Benefit.FIXED, range=self._range, value=self.seat_price / 3)
```
mjfrey: Fixed benefit should be a dollar amount not a division.  Example: seat_price - 10

ecommerce/extensions/offer/tests/test_utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def test_format_benefit_value(self):
+        """ format_benefit_value(benefit) should format benefit value based on benefit type """
+        benefit_value = format_benefit_value(self.percentage_benefit)
+        self.assertEqual(benefit_value, '35%')
+
+        benefit_value = format_benefit_value(self.value_benefit)
+        self.assertEqual(benefit_value, currency(self.seat_price / 3))
```
mjfrey: Same as above. Make it a fixed dollar amount

marjev:
@mjfrey Comments addressed.

mjfrey:
:+1:

PR #590, opened 2016-03-02
Check if there is frozen basket for the user, if so thaw and use.

mjfrey:
Check if a FROZEN basket exists.  If so, that and let it get used in get_basket.  This will ensure we present the user with his current basket if the summary view is called.  It will also get merged in the single-item view if something new is added to the user's basket.

### DIFF
--- a/ecommerce/extensions/basket/middleware.py
+++ b/ecommerce/extensions/basket/middleware.py
@@ -36,6 +36,9 @@ class BasketMiddleware(OscarBasketMiddleware):
             if basket:
                 basket = basket[0]
                 basket.thaw()
+                # Delete any remaining stale FROZEN baskets.
+                for b in basket[1:]:
+                    b.delete()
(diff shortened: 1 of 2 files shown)


mjfrey:
@clintonb Take a look.  This solves the page refresh and back button issue during checkout.

clintonb:
This won't work as expected. Say I have two devices. I get through to checkout on Device 1, but pause on the payment page. At this point Basket 1 is frozen. I switch to Device 2 and start checkout. Basket 1 has now been thawed, and I begin modifying it. Now I go back to Device 1, and resume checkout. An order is created, but it doesn't match up with the basket anymore.

Additionally, unless you can be absolutely certain a user is not viewing a frozen basket at checkout, you should not delete a frozen basket. The risk of someone purchasing a deleted basket is too great. I'd much rather deal with user unhappiness over an empty basket page than paying for a deleted basket.

PR #593, opened 2016-03-02
Updated PayPal and requests dependencies

clintonb:
These updates ensure we do not encounter issues with the forthcoming PayPal TLS switchover.

ECOM-3835

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -18,6 +18,7 @@ edx-ecommerce-worker==0.3.0
 edx-rest-api-client==1.3.0
 jsonfield==1.0.3
 libsass==0.9.2
+ndg-httpsclient==0.4.0
 paypalrestsdk==1.11.5
 premailer==2.9.2
 pycountry==1.18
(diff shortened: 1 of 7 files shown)


clintonb:
FYI @edx/ecommerce

requirements/base.txt
```diff
@@ -18,10 +18,11 @@ edx-ecommerce-worker==0.3.0
 edx-rest-api-client==1.3.0
 jsonfield==1.0.3
 libsass==0.9.2
-paypalrestsdk==1.11.4
+ndg-httpsclient==0.4.0
```
rlucioni, line 21: Can you help me understand why we need this? Why isn't requests sufficient?

clintonb, line 21: This is needed to support TLS 1.2.

rlucioni:
:+1:

jimabramson:
:+1: , but does PyOpenSSL need to be in place for this to work?

clintonb:
PyOpenSSL is a dependency of one of the libraries. It is already installed.

clintonb:
Tests break with ndg-httpsclient due to https://github.com/gabrielfalcao/HTTPretty/issues/242.

PR #594, opened 2016-03-02
Fix for SOL-1681

mjfrey:
New messaging for empty basket page.

### DIFF
--- a/ecommerce/templates/oscar/basket/partials/basket_content.html
+++ b/ecommerce/templates/oscar/basket/partials/basket_content.html
@@ -1,4 +1,5 @@
 {% load i18n %}
+{% load core_extras %}
 {% load currency_filters %}
 {% load purchase_info_tags %}
 {% load widget_tweaks %}
@@ -96,7 +97,7 @@
         </div>
 
         <div class="row">
-            <div class="payment-buttons col-sm-3" data-basket-id="{{ basket.id }}">
+            <div class="payment-buttons col-sm-4" data-basket-id="{{ basket.id }}">
                 {% for processor in payment_processors %}
                     <button class="btn btn-success payment-button" value="{{ processor.NAME|lower }}" id="{{ processor.NAME|lower }}">
                         {% if processor.NAME == 'cybersource' %}
@@ -128,9 +129,21 @@
     </div>
 {% else %}
     {% block emptybasket %}
-        <p>
-            {% trans "Your basket is empty." %}
-            <a href="{{ homepage_url }}">{% trans "Discover new courses." %}</a>
-        </p>
+    <div class="depth depth-2 message-error">
+        <h3>{% trans "Your purchase could not be completed" %}</h3>
+        {% captureas dashboard_link %}
+            <a href="{{ homepage_url }}">{% trans "Dashboard" %}</a>
+        {% endcaptureas %}
+        {% captureas support_link %}
+            <a href="{{ support_url }}">
+                {% blocktrans %}
+                    contact {{ platform_name }} Support
+                {% endblocktrans %}
+            </a>
+        {% endcaptureas %}
+        {% blocktrans %}
+            You have not been charged. Return to your {{dashboard_link}} to try again, or {{support_link}}.
+        {% endblocktrans %}
+    </div>
     {% endblock %}
 {% endif %}
--- a/ecommerce/static/sass/base/_base.scss
+++ b/ecommerce/static/sass/base/_base.scss
@@ -33,3 +33,39 @@ a {
     }
   }
 }
+
+.message-error {
+  border-top: 4px solid #B20610;
+  h3 {
+    margin-bottom: 10px;
+    font-weight: bold;
+  }
+}
+
+.depth {
+  padding: 1.25rem 1.25rem;
+}
+
+.depth-2 {
+  background: #fcfcfc;
+  box-shadow: 0 1px 2px 1px rgba(167, 164, 164, 0.25);
+}
+
+.depth--3 {
+  background: #34383a;
+  text-align: center;
+  .copy {
+    color: #e7e6e6;
+    @include container-fixed;
+
+    @media (min-width: $screen-sm-min) {
+      width: $container-sm;
+    }
+    @media (min-width: $screen-md-min) {
+      width: $container-md;
+    }
+    @media (min-width: $screen-lg-min) {
+      width: $container-lg;
+    }
+  }
+}
--- a/ecommerce/static/sass/views/_basket.scss
+++ b/ecommerce/static/sass/views/_basket.scss
@@ -1,6 +1,6 @@
 .basket {
     .container {
-        padding: 60px 0 30px 0;
+        padding: 60px 20px 30px 20px;
 
         .basket-items,
         .total,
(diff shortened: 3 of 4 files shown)


ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -129,8 +129,8 @@
 {% else %}
     {% block emptybasket %}
         <p>
-            {% trans "Your basket is empty." %}
-            <a href="{{ homepage_url }}">{% trans "Discover new courses." %}</a>
+            {% trans "Your purchase could not be completed" %} </br>
```
clintonb: This assumes everyone arriving at the page has clicked the back button. If I arrive at the page by simply navigating to /basket/, this message is not valid. Try something along the following lines:

```
Your basket is empty. Discover new courses.

If you arrived at this page by clicking your browser's "Back" button on the payment page, please go forward, and click the "Cancel" button on the payment page.
```

@srpearce may be able to provide better language.

mjfrey: This is what @srpearce and @pbaruah came up with.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -129,8 +129,8 @@
...
+            {% trans "You have not been charged. Return to your " %}<a href="{{ homepage_url }}">{% trans "Dashboard" %}</a> {% trans "to try again, or " %} <a href="{{ support_url }}">{% trans "contact edX support." %}</a>
```
clintonb: This won't bet properly translated. Use `captureas`. See https://github.com/edx/ecommerce/blob/master/ecommerce/templates/edx/credit/checkout.html#L23-L28.

ecommerce/templates/oscar/basket/partials/basket_content.html
clintonb: This messaging is not good for the buyer. If I arrive at this page by accident, I am confused because I have no clue why the page is talking about payment. (This is most likely an edge case.) If I get to this page by clicking the back button, I now have to start the entire process over...from the dashboard...which has almost nothing to do with payment. At a minimum we should take steps to inform the user _why_ an issue occurred so s/he does not repeat the mistake. It's even better if we provide options to resolve the issue. Leaving the user with only the dashboard option will lead to drop-off.

pbaruah:
I think this should take care of the majority of the users who might be in this state. We can improve in future or have a better solution for the back button issue.

:+1:

rock345:
I am a thumbs up given we a dress the Clinton translation concerns.  Partha can you create the follow up stories?

rock345:
I am a thumbs up given we a dress the Clinton translation concerns.  Partha can you create the follow up stories?

marjev:
@mjfrey, here's the screenshot that shows the Empty basket message style:

screen shot 2016-03-03 at 12 23 14

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -128,9 +128,9 @@
     </div>
 {% else %}
     {% block emptybasket %}
-        <p>
-            {% trans "Your basket is empty." %}
-            <a href="{{ homepage_url }}">{% trans "Discover new courses." %}</a>
-        </p>
+    <div class="depth depth-2 message-error">
+        <h3>{% trans "Your purchase could not be completed" %}</h3>
+        {% trans "You have not been charged. Return to your " %}<a href="{{ homepage_url }}">{% trans "Dashboard" %}</a> {% trans "to try again, or " %} <a href="{{ support_url }}">{% trans "contact edX support." %}</a>
```
clintonb: This won't bet properly translated. Use `captureas`. See https://github.com/edx/ecommerce/blob/master/ecommerce/templates/edx/credit/checkout.html#L23-L28.

clintonb:
:+1: (once translation string fixed)

ecommerce/templates/oscar/basket/partials/basket_content.html
marjev: I don't see the need for a custom tag. Simply using blocktrans should suffice. It's simpler.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -128,9 +128,11 @@
...
 {% else %}
     {% block emptybasket %}
-        <p>
-            {% trans "Your basket is empty." %}
-            <a href="{{ homepage_url }}">{% trans "Discover new courses." %}</a>
-        </p>
+    <div class="depth depth-2 message-error">
+        <h3>{% trans "Your purchase could not be completed" %}</h3>
+        {% blocktrans %}
+            You have not been charged. Return to your <a href="{{ homepage_url }}">Dashboard</a> to try again, or <a href="{{ support_url }}">contact edX support</a>.
```
clintonb: You assume translators will properly copy over the HTML. This is a bad assumption. Experience has shown me that not all translators are perfect. They _will_ mess this up. If they mess up the HTML, you won't notice it unless you manually view the site in every language. If they mess up a placeholder (usually by forgetting the curly braces), the translations will fail to compile.

clintonb: "edX" should come from the site/site configuration. Support should be capitalized.

marjev: can you be more specific about where "edX" should come from and let me know how I can verify that locally? I don't know what you mean by `site/site configuration`.

clintonb: Every request is tied to a `Site`, which is tied to a `SiteConfiguration` instance, which is tied to a `Partner` instance. We must remember that what we produce must work for all tenants in addition to open source users. "edX" should not be hard coded anywhere. That name should come from the partner's name.

For now, you should be safe using `{{ platform_name }}` which is set in `ecommerce/core/context.py`.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -128,9 +129,21 @@
...
+        {% endcaptureas %}
+        {% captureas support_link %}
+            <a href="{{ support_url }}">
+                {% blocktrans %}
+                    contact {{ platform_name }} Support
+                {% endblocktrans %}
+            </a>
+        {% endcaptureas %}
+        {% blocktrans %}
+            You have not been charged. Return to your {{dashboard_link}} to try again, or {{support_link}}.
```
clintonb, line 145: This is wrong. You want to put the entire sentence/paragraph in the translation block. There is not substitution when we extract translation strings. The correct implementation is below.

``` jinja
{% captureas dashboard_link_start %}
  <a href="{{ homepage_url }}">
{% endcaptureas %}

{% captureas support_link_start %}
  <a href="{{ support_url }}">
{% endcaptureas %}


{% blocktrans with link_end="</a>" %}
  You have not been charged. Return to your {{ dashboard_link_start }}dashboard{{ link_end }} to try again, or {{ support_link_start }}contact {{ platform_name }} Support{{ link_end }}.
{% endblocktrans %}
```

PR #600, opened 2016-03-04
Add better errror messaging

mjfrey:
Add better messaging when a voucher is no longer active.

### DIFF
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -8,6 +8,7 @@ from django.http import HttpResponseRedirect
 from django.utils.decorators import method_decorator
 from django.utils.translation import ugettext_lazy as _
 from django.shortcuts import render
+from django.utils import timezone
 from django.views.generic import TemplateView, View
 from oscar.core.loading import get_class, get_model
 
@@ -78,7 +79,11 @@ def voucher_is_valid(voucher, product, request):
         return False, _('Coupon does not exist')
 
     if not voucher.is_active():
-        return False, _('Coupon expired')
+        now = timezone.now()
+        if voucher.start_datetime > now:
+            return False, _('This coupon code is not yet valid.')
+        elif voucher.end_datetime < now:  # pragma: no cover
+            return False, _('This coupon code has expired.')
 
     avail, msg = voucher.is_available_to_user(request.user)
     if not avail:
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -133,8 +133,8 @@ class BasketSummaryViewTests(LmsApiMockMixin, TestCase):
         basket.add_product(product, 1)
         return basket
 
-    def create_seat(self, course, seat_price=100):
-        return course.create_or_update_seat('verified', True, seat_price, self.partner)
+    def create_seat(self, course, seat_price=100, cert_type='verified'):
+        return course.create_or_update_seat(cert_type, True, seat_price, self.partner)
 
     def create_and_apply_benefit_to_basket(self, basket, product, benefit_type, benefit_value):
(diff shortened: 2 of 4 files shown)


mjfrey:
@vkaracic

mjfrey:
@clintonb I can not get codecov to not mark a line of code as partial.  Can you take a look?

ecommerce/coupons/views.py
```diff
@@ -78,7 +79,11 @@ def voucher_is_valid(voucher, product, request):
         return False, _('Coupon does not exist')
 
     if not voucher.is_active():
-        return False, _('Coupon expired')
+        now = timezone.now()
+        if voucher.start_datetime > now:
+            return False, _('This coupon code is not yet valid.')
+        elif voucher.end_datetime < now:
```
clintonb: [Exclude this clause from coverage](http://coverage.readthedocs.org/en/coverage-4.0.3/excluding.html). @nedbat, when you get a chance, can you take a look at https://codecov.io/github/edx/ecommerce/compare/55e1717...fdee5c7 and confirm this is the correct fix? @mjfrey no need to wait on this confirmation.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -133,8 +133,8 @@ def create_basket_and_add_product(self, product):
         basket.add_product(product, 1)
         return basket
 
-    def create_seat(self, course, seat_price=100):
-        return course.create_or_update_seat('verified', True, seat_price, self.partner)
+    def create_seat(self, course, seat_price=100, cert_type='verified'):
+        return course.create_or_update_seat(cert_type, True, seat_price, self.partner)
```
vkaracic, line 137: I'm not sure if we really need an extracted method for a one-liner.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -191,6 +191,12 @@ def test_no_basket_response(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.context['lines'], [])
 
+    def test_line_no_cert_type(self):
+        seat = self.create_seat(course=self.course, cert_type='TEST', seat_price=100)
+        self.create_basket_and_add_product(seat)
+        response = self.client.get(self.path)
+        self.assertEqual(response.context['lines'][0].certificate_type, None)
```
vkaracic: When you assert if an object is `None` you can always use `assertIsNone`.

ecommerce/coupons/views.py
```diff
@@ -78,7 +79,11 @@ def voucher_is_valid(voucher, product, request):
         return False, _('Coupon does not exist')
 
     if not voucher.is_active():
-        return False, _('Coupon expired')
+        now = timezone.now()
+        if voucher.start_datetime > now:
+            return False, _('This coupon code is not yet valid.')
+        elif voucher.end_datetime < now:  # pragma: no cover
```
vkaracic, line 85: I think you have this one covered with the `test_expired_voucher` test. 
(I'm having issues running coverage locally so I can't confirm it.)

vkaracic:
You have a typo in the commit message, once that is fixed :+1:
The rest of the comments are nits and not _required_ to fix.

ecommerce/coupons/views.py
nedbat, line 85: Two points about this pragma:
- the line of code is covered, but only partially, so a better pragma is `# pragma: no branch`.
- but I'm concerned about the code itself.  The reason this is marked as partially covered is because when line 85 is reached, the condition is always true.  The "invisible else" was never taken.  Is there a scenario that could execute that else?  If so, let's add a test for it.  If not, why are we checking this always-true condition?

nedbat, line 85: @clintonb @mjfrey ^^

PR #601, opened 2016-03-08
Stripe integration 

jbzdak:
This implements stripe processor for ecommerce. 
## Outstanding issues:
- [x] Stripe processor is intended to work using "new" flow where payment process is done entirely in otto ([I mean this PR](https://github.com/edx/ecommerce/pull/537)). Implementing it in LMS would be either non trivial, or non-optimal from user standpoint. Check if this is OK. 
- [x] Some automatic tests fail 
- [x] Figure out what to do with Braintree procssor (probably just remove it). 
- [ ] Squash commits
- [ ] This will require changes to the edx/configuration repository, [due to this problem](https://github.com/edx/ecommerce/pull/601#discussion_r56267683) if we merge this now it will break all e-commerce sandboxes.

Testing instructions are quite long, so they are in a [separete document](https://docs.google.com/document/d/1yDDuRaIzbn3nMKDFRKV-cBd7JSPgxVktw6iBWbg9md4/edit#heading=h.3nnyxcsymae1).

Sandbox url: http://pr11788.sandbox.opencraft.com/. 
## Current Stripe Payment flow:

Student chooses to pursue verified certificate: 

He gets redirected to Otto Basket Page: 

He chooses Stripe Payment: 

He enters CC data

Payment gets processed:


openedx-webhooks:
Thanks for the pull request, @jbzdak! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=601

Kelketek:
@jbzdak As the testing instructions are very long, it would be best to create a sandbox that will handle this. You can create a pull request with an empty commit against edx-platform to generate one, and configure it from there.

ecommerce/extensions/payment/processors/braintree.py
```diff
@@ -0,0 +1,190 @@
...
+            AttributeError: If ECOMMERCE_URL_ROOT setting is not set.
+        """
+        configuration = self.configuration
+        self.merchant_id = configuration['merchant_id']
+        self.public_key = configuration['public_key']
+        self.private_key = configuration['private_key']
+        self.receipt_page_url = configuration['receipt_page_url']
+        self.ecommerce_url_root = settings.ECOMMERCE_URL_ROOT
+
+        # TODO Configure sandbox-production via settings
```
Kelketek: Is this something you intend to do before merge, or do we have a separate ticket for it?

ecommerce/extensions/payment/processors/braintree.py
```diff
@@ -0,0 +1,190 @@
...
+
+        transaction = result.transaction
+        if result.is_success:
+            # Create Source to track all transactions related to this processor and order
+            source_type, __ = SourceType.objects.get_or_create(name=self.NAME)
+            currency = basket.currency
+            total = basket.total_incl_tax
+            transaction_id = transaction.id
+
+            # TODO Check for PayPal data
```
Kelketek: Same question for this TODO. Is this to be resolved in this PR or do we have a ticket for it?

ecommerce/extensions/payment/processors/braintree.py
```diff
@@ -0,0 +1,190 @@
...
+                label = transaction.credit_card_details.masked_number if transaction.credit_card['bin'] else None
+                card_type = transaction.credit_card_details.card_type
+            elif hasattr(transaction, 'paypal_details'):
+                email = transaction.paypal_details.payer_email
+                label = 'PayPal ({})'.format(email) if email else 'PayPal Account'
+            elif hasattr(transaction, 'coinbase_details'):
+                email = transaction.coinbase_details.user_email
+                label = 'Coinbase ({})'.format(email) if email else 'Coinbase Account'
+
+            # TODO Serialize transaction to JSON
```
Kelketek: Another TODO, same question

ecommerce/extensions/payment/processors/braintree.py
```diff
@@ -0,0 +1,190 @@
...
+            # self.record_processor_response(self.braintree_object_tree_to_dict(transaction),
```
Kelketek: What did this commented out code do, and do we still need it?

ecommerce/extensions/payment/processors/braintree.py
```diff
@@ -0,0 +1,190 @@
...
+
+            return source, event
+        elif transaction:
+            logger.error('Transaction failed: [%s] - [%s]',
+                         transaction.processor_response_code,
+                         transaction.processor_response_text)
+            raise GatewayError
+
+        else:
+            # TODO Log the error!
```
Kelketek: You should be able to resolve this particular TODO before merge.

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,162 @@
...
+        return {
+            'stripe_publishable_key': self.publishable_key,
+            'stripe_process_payment_url': reverse('stripe_checkout', kwargs={
+                'basket': basket.pk
+            }),
+            'stripe_amount_cents': self.get_total(basket),
+            'stripe_image_url': self.image_url,
+            'stripe_currency': basket.currency,
+            'stripe_user_email': user.email,
+            # TODO: Description could describe payment more.
```
Kelketek: Sounds like something you intend to resolve before merge?

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -36,6 +37,11 @@ class Paypal(BasePaymentProcessor):
     NAME = u'paypal'
     DEFAULT_PROFILE_NAME = 'default'
 
+    @property
+    def payment_label(self):
+        # Translators: Do NOT translate the name PayPal.
+        return _("Checkout with PayPal")
```
Kelketek, line 43: When used as a verb, it's 'Check out'. 'Checkout' is used for describing a place where a person checks out.

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,162 @@
...
+            'stripe_payment_description': self.get_description(basket),
+            'button_label': self.payment_label,
+            "basket": basket
+        }
+
+    @property
+    def payment_label(self):
+        return _("Checkout using Credit-Card")
```
Kelketek: "Check out"

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,162 @@
...
+                }
+            )
+            logger.info(charge)
+
+            source_type, __ = SourceType.objects.get_or_create(name=self.NAME)
+            currency = basket.currency
+            total = basket.total_incl_tax
+            transaction_id = charge.id
+
+            # TODO: stripe guarantees that transaction_id is shorter than 255
```
Kelketek: Not sure what the requirement for fulfilling this TODO is? Planning to add an assertion? Same question here as for the other TODOs

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,162 @@
...
+            # and reference is 128 long
+            source = Source(source_type=source_type,
+                            currency=currency,
+                            amount_allocated=total,
+                            amount_debited=total,
+                            reference=transaction_id,
+                            label='Stripe',
+                            card_type=None)
```
Kelketek: Paren should be on next line, and the args should be started on the next line, with only one level of indentation, per edx style.

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,162 @@
...
+
+            # Create PaymentEvent to track
+            event_type, __ = PaymentEventType.objects.get_or_create(name=PaymentEventTypeName.PAID)
+            event = PaymentEvent(event_type=event_type, amount=total, reference=transaction_id,
```
Kelketek: edx indentation style here, too.

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,162 @@
...
+        transaction_id = source.reference
+        order = source.order
+
+        source.refund(amount, reference=transaction_id)
+
+        event_type, __ = PaymentEventType.objects.get_or_create(name=PaymentEventTypeName.REFUNDED)
+        PaymentEvent.objects.create(event_type=event_type, order=order, amount=amount, reference=transaction_id,
+                                    processor_name=self.NAME)
+
+    # TODO: Remove this note: issues refund for a given transaction
```
Kelketek: Remove which note? What needs to be done here?

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,162 @@
...
+    def issue_credit(self, source, amount, currency):
+
+        # TODO: Technically stripe guarantees all fields to be shorter than 255
```
Kelketek: Similar question to other TODOs

ecommerce/extensions/payment/views.py
```diff
@@ -313,3 +332,98 @@ def get(self, request, *_args, **_kwargs):
...
+    def get_payment_data(self, request):
+        raise NotImplementedError
+
+    def locate_basket(self):
+        basket_id = self.request.POST['basket_id']
+        basket = self.get_basket(self.request, basket_id)
+
+        if not basket:
+            logger.error('Payment requested for non-existent basket [%s].', basket_id)
+            # TODO Handle this better (perhaps redirect to an error page).
```
Kelketek: Planning to do this as part of this PR, or a should a separate ticket be created?

ecommerce/extensions/payment/views.py
```diff
@@ -313,3 +332,98 @@ def get(self, request, *_args, **_kwargs):
...
+
+        # Retrieve the basket, or bail out, if it cannot be found.
+        basket = self.locate_basket()
+
+        try:
+            with transaction.atomic():
+                try:
+                    self.handle_payment(payment_data, basket)
+                except PaymentError:
+                    # TODO Handle this better (perhaps redirect to an error page).
```
Kelketek: Same standard question with this TODO.

ecommerce/extensions/payment/views.py
```diff
@@ -313,3 +332,98 @@ def get(self, request, *_args, **_kwargs):
...
+                    return HttpResponseBadRequest()
+        except:  # pylint: disable=bare-except
+            logger.exception('Attempts to handle payment for basket [%d] failed.', basket.id)
+            # TODO Handle this better (perhaps redirect to an error page).
```
Kelketek: Same standard question with this TODO.

Kelketek:
@jbzdak Seems mostly sane so far, just some questions about tech debt introduced via TODOs, a few style issues, and a grammar thing or two. Let me know when the sandbox is working.

jbzdak:
@Kelketek Thanks for pointing the `#TODO`s out. I 'inherited' most of them from branch I was basing on and I stopped noticing them (I should notice them).  I think that for some of them I'll need to explicitly ask E-commerce team, but I'll look into them tomorrow.

ecommerce/extensions/basket/views.py
```diff
@@ -95,15 +96,37 @@ def get_context_data(self, **kwargs):
...
+        basket = self.request.basket
+        user = self.request.user
+        processors = (
+            get_processor_class(path)
+            for path in settings.PAYMENT_PROCESSORS
+        )
+        enabled_processors = [
+            processor() for processor in processors if processor.is_enabled()
+        ]
+        filter = lambda sequence: [item for item in sequence if item]
```
bradenmacdonald: `filter` is a built-in method - you don't need to define it and should not be overwriting it.

ecommerce/extensions/basket/views.py
```diff
@@ -95,15 +96,37 @@ def get_context_data(self, **kwargs):
...
+            'faq_url': get_lms_url('') + '/verified-certificate',
         })
+        context.update(self.get_payment_processors())
         return context
 
     def get_payment_processors(self):
         """ Retrieve the list of active payment processors. """
         # TODO Retrieve this information from SiteConfiguration
-        processors = (get_processor_class(path) for path in settings.PAYMENT_PROCESSORS)
-        return [processor for processor in processors if processor.is_enabled()]
```
bradenmacdonald, line 109: The code that was here before was cleaner IMHO.

I'd rather you just keep it as it was and change the usage above to:

``` python
          processors = self.get_payment_processors()
          context.update({
              'free_basket': context['order_total'].incl_tax == 0,
              'payment_processors': processors,
              'extra_scripts': filter(None, (processor.get_payment_page_script(basket, user) for processor in processors)),
              ...
          })
```

(Here, `filter(None, ...)` is the built-in method, not a custom method)

jbzdak, line 109: I wanted to avoid this `filter(None,`, but you are right I should use a builtin.

ecommerce/extensions/basket/views.py
```diff
@@ -95,15 +96,37 @@ def get_context_data(self, **kwargs):
...
+        filter = lambda sequence: [item for item in sequence if item]
+        return {
+            "payment_processors": [
+                processor.render_payment_button(basket, user)
+                for processor in enabled_processors
+            ],
+            "payment_processors_scripts": filter(
+                processor.get_payment_page_script(basket, user)
+                for processor in enabled_processors
+            )
```
bradenmacdonald: Per edX Python conventions, you should never return a `dict` with a fixed form - use a `namedtuple` instead. But I think in this case it's best to return the Processor objects themselves, as the code was originally doing.

ecommerce/extensions/payment/processors/__init__.py
```diff
@@ -33,18 +41,67 @@ def get_transaction_parameters(self, basket, request=None):
         """
         raise NotImplementedError
 
+    def render_payment_button(self, basket, user):
```
bradenmacdonald: `basket` and `user` are unused here. A linter should have caught this?

ecommerce/extensions/payment/processors/__init__.py
```diff
@@ -14,6 +16,12 @@ class BasePaymentProcessor(object):  # pragma: no cover
 
     NAME = None
 
+    STANDARD_BUTTON_TEMPLATE = u"""
+    <button class="{button_class}" value="{processor_name}">
+        {button_label}
+    </button>
+    """.strip()
```
bradenmacdonald: HTML belongs in templates - I think this should be left in `basket_content.html`

ecommerce/extensions/basket/views.py
bradenmacdonald, line 109: ^ Updated a bit.

ecommerce/extensions/payment/processors/__init__.py
jbzdak: They are used in the subclass (I think). This will go away as stated in earlier comment.

ecommerce/extensions/payment/processors/__init__.py
```diff
@@ -33,18 +41,67 @@ def get_transaction_parameters(self, basket, request=None):
...
+            processor_name=self.NAME
+        )
+
+    @property
+    def payment_label(self):
+        """
+        Returns: Label for payment button
+        """
+        return _(u"Checkout with {processor_name}").format(
+            processor_name=self.NAME.capitalize()
```
bradenmacdonald, line 44: `NAME.capitalize()` doesn't work in some cases such as "paypal" which should be "PayPal". Instead, you should have a separate field for "name" (PayPal) and "id" (paypal) (or whatever you want to call these two fields).

ecommerce/extensions/payment/processors/__init__.py
```diff
@@ -33,18 +41,67 @@ def get_transaction_parameters(self, basket, request=None):
...
+    @property
+    def payment_button_classes(self):
+        """
+        Returns: CSS classes for payment button
+        """
+        return u"btn btn-success payment-button builtin-handling"
+
+    def get_payment_page_script(self, basket, user):
+        """
+
```
bradenmacdonald, line 63: Nit: blank line here. Docstrings should begin with a one-line summary and then have details on the lines below.

ecommerce/extensions/payment/processors/__init__.py
jbzdak, line 44: This is the default, it is overriden for PayPal (and most other processors), just trying to provide sane default here.

ecommerce/extensions/payment/processors/__init__.py
```diff
@@ -33,18 +41,67 @@ def get_transaction_parameters(self, basket, request=None):
...
+        return _(u"Checkout with {processor_name}").format(
+            processor_name=self.NAME.capitalize()
+        )
+
+    @property
+    def payment_button_classes(self):
+        """
+        Returns: CSS classes for payment button
+        """
+        return u"btn btn-success payment-button builtin-handling"
```
bradenmacdonald: These classes should be kept in the template if possible. Can you not just put `class="btn btn-success payment-button payment-button-{{processor.NAME}}"` in the template?

ecommerce/extensions/payment/processors/__init__.py
```diff
@@ -59,7 +116,10 @@ def configuration(self):
         Raises:
             KeyError: If no settings found for this payment processor
```
bradenmacdonald: This docstring is now out of date, since you've changed the behavior.

jbzdak: Good point.

ecommerce/extensions/payment/processors/__init__.py
bradenmacdonald, line 44: Ah, I see. Ok.

ecommerce/extensions/payment/processors/__init__.py
jbzdak: The point is the javascript on checkout page automatically handles all payment buttons, these classes allow me to inhibit this behaviour. I could leave classes alone and then add some `data-inhibit-default-handler=True`, which would solve the issue.

ecommerce/extensions/payment/processors/braintree.py
```diff
@@ -0,0 +1,190 @@
+""" Braintree payment processing. """
```
bradenmacdonald: Does the Braintree processor work, or is there a good reason to include those WIP commits from the earlier branch? If not, I'd rather you leave it out of this PR for now.

ecommerce/extensions/payment/processors/__init__.py
bradenmacdonald: Yeah, even if you just do `data-has-custom-handler-script="{{ processor.has_script }}"` (I'm being verbose but you get the idea). CSS classes and HTML attributes should stay in the template.

ecommerce/extensions/payment/processors/braintree.py
jbzdak: This is an explicit point I was unsure about (see PR cover letter). Then I'll delete it.

ecommerce/extensions/payment/processors/__init__.py
jbzdak: OK this makes sense, and is cleaner than proposed solution. Thank you, I'll move the button template to the templates.

ecommerce/extensions/payment/processors/braintree.py
bradenmacdonald: Ah, sorry I missed that. Please delete it. (But if you've made changes to adapt it to the current API, then make sure to keep those commits available somewhere so someone else can pick it up again in the future - don't squash that work into oblivion.)

Also AFAIK there's no reason to worry about the old checkout flow in the LMS. Focus on the new stuff.

clintonb:
Before you go too much further on this, we should take some time to discuss your integration. The prototype code you are adapting for Braintree/Stripe doesn't necessarily fit with where we want to take payment integration. We ended up with the current design because our checkout page was hosted by LMS. Now that the checkout page is hosted here in Otto, we want to bring things more in line with Oscar, similar to [django-oscar-paypal](https://github.com/django-oscar/django-oscar-paypal).

I realize I am being a bit vague, because we haven't yet thought through the future architecture. However, I want to make sure we have this discussion now rather than after you've done all the work and are awaiting a final review.

Questions
1. Do you want to use Stripe Checkout or Stripe.js?
2. What will the UI look like?

jbzdak:
@clintonb Just to have a baseline for discussion, I see two possible approaches: 
1. Use `checkout.js` appraoch; It has the downside that `PaymentProcessors` need to be able to attach custom `.js` to the payment page, and has the upside that payment is ultimately handled by Stripe. 
2. Create a landing page for payment on E-Commerce; which integrates better with the current flow, but we need to host a payment form on our page. 

I went for `checkout.js` approach, so when user clicks `Pay with Credit Card` he gets an Ajax popup, then, after submitting payment details payment is made using: `CheckoutViewMixin`. The flow feels natural (at least works like on couple of pages that I used stripe on), and works nicely. 

I noticed that eventually we'll be moving towards using `oscar` plugins, but there is no working `django-oscar-stripe` plugin yet.

jbzdak:
Sorry for the mail noise, I clicked the wrong button when responding to Clinton.

ecommerce/extensions/payment/processors/braintree.py
jbzdak: This is another procesor (braintree) it will most probably be removed from this PR, it is leftover work from the WIP PR I based on.

ecommerce/extensions/payment/processors/stripe.py
jbzdak: Precisely, but obtaining a concise description from a basket that can contain many items either requires some heuristics or I missed something here. Anyways upstream I'm counting on upstream on that.

ecommerce/extensions/payment/processors/paypal.py
jbzdak, line 43: This one is actually an [existing label in ecommerce](https://github.com/edx/ecommerce/pull/601/files#diff-b9b26b06e74e62ca7c4ba347c28eb249L108), it also seems translated and I'd rather not change it without a good reason. I just moved the label from ifelse in template to a class.

ecommerce/extensions/payment/processors/stripe.py
jbzdak: This something I probably will have to fix on way or another --- the problem is that `Source.reference` is `VARCHAR(128)`, and stripe uses ids that are shorter than that, but they **guarantee** only that they will be shorted than 255. There are couple of possible solutions, but I'll need to choose something.

ecommerce/extensions/payment/processors/stripe.py
jbzdak: The same as [here](https://github.com/edx/ecommerce/pull/601#discussion_r55835140)

ecommerce/extensions/basket/views.py
clintonb, line 109: I'd prefer that this method remain in its simplified form as well. If you need scripts, create a new method, `get_payment_processor_scripts`.

jbzdak, line 109: If you wait about half an hour I'll post new version PR with most of cleanups suggested by @bradenmacdonald.

ecommerce/extensions/payment/processors/stripe.py
jbzdak: Here I'd stay consistent with "Checkout with PayPal" :)

bradenmacdonald: It should just say "Checkout with credit card" or "Check out with credit card" - there's no hyphen and credit card is not a proper noun.

clintonb: "Credit Card" (without the hyphen) makes more sense given that is the actual payment mechanism. I doubt most people know what Stripe is since they don't regularly interact with Stripe as they would PayPal.

ecommerce/templates/edx/base.html
```diff
@@ -54,5 +54,8 @@
     {% block javascript %}
     {% endblock javascript %}
 {% endcompress %}
+
+{% block uncompressedjs %}
```
jbzdak, line 58: Stripe JS can't be compressed during collectstatic as processors are meant to be enabled and disabled interactively during website operation by waffle switches, so AFAIK it's just better to attach them dynamically. This block must be after other js blocks to see `requirejs`.

ecommerce/extensions/payment/processors/stripe.py
jbzdak: After some thought current description is OK, does the job and is brief, I'm not sure if there are viable alternatives.

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,168 @@
...
+        try:
+            configuration = self.configuration
+            self.publishable_key = configuration['publishable_key']
+            self.secret_key = configuration['secret_key']
+            self.receipt_page_url = configuration['receipt_url']
+            self.error_page_url = configuration['error_url']
+            self.image_url = configuration['image_url']
+            self.ecommerce_url_root = settings.ECOMMERCE_URL_ROOT
+        except KeyError as e:
+            raise ImproperlyConfigured(
```
jbzdak, line 52: PaymentProcessors are instantiated during `views.py` module creation, and if module (when it's being created) raises `KeyError` then weird things happen. `ImproperlyConfigured` is way safer in this regard, as nothing catches it explicitly.

ecommerce/extensions/payment/views.py
```diff
@@ -313,3 +331,91 @@ def get(self, request, *_args, **_kwargs):
...
+            # Raise the error to give clear 500, in future (maybe?) introduce better way of notifying
+            # admins with clear message.
+            raise
+
+        receipt_url = u'{}?basket_id={}'.format(self.payment_processor.receipt_page_url, basket.id)
+        return redirect(receipt_url)
+
+
+class StripeCheckoutView(CheckoutViewMixin, View):
+    payment_processor = StripeProcessor()
```
jbzdak, line 407: `payment_processor` attribute is required by `EdxOrderPlacementMixin`, and needs to be instantiated instance of PaymentProcessor. PaymentProcessors explode with either a `KeyError` or `InproperlyConfigured` error when instantiated, and missing settings (even those processors that are not enabled). This is some deeper problem in ECommerce, not really solvable in this PR, but worth pointing out. 

This has the side effect that every processor needs to have configuration, with all the required keys --- even if values make no sense.

acceptance_tests/test_coupon_checkout.py
```diff
@@ -59,6 +61,7 @@ def test_enrollment_code_redemption(self):
         self.assert_order_created_and_completed()
         self.assert_user_enrolled(self.username, self.course_id, mode='verified')
 
+    @skipUnless(ENABLE_PAYPAL_TESTS, "PayPal tests are not enabled")
```
jbzdak, line 64: Due to https://openedx.atlassian.net/browse/ECOM-3904 I had to disable PayPal tests locally, it could make sense to disable each payment processor, as now we have three, and both Stripe and PayPal are easy to set up on sandbox

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,200 @@
...
+            logger.info(charge)
+
+            self.record_processor_response(charge, basket=basket, transaction_id=charge.id)
+
+            source_type, __ = SourceType.objects.get_or_create(name=self.NAME)
+            currency = basket.currency
+            total = basket.total_incl_tax
+            transaction_id = charge.id
+
+            # TODO: Stripe guarantees that transaction_id is shorter than 255
```
jbzdak, line 123: This todo is problematic. Ideas appreciated. [This page explains what a compatible change means to Stripe](https://stripe.com/docs/upgrades#what-changes-does-stripe-consider-to-be-backwards-compatible).

ecommerce/extensions/payment/processors/stripe.py
```diff
@@ -0,0 +1,198 @@
...
+                reference=transaction_id,
+                label='Stripe',
+                card_type=None
+            )
+
+            # Create PaymentEvent to track
+            event_type, __ = PaymentEventType.objects.get_or_create(
+                name=PaymentEventTypeName.PAID
+            )
+            event = PaymentEvent(event_type=event_type, amount=total, reference=transaction_id,
```
Kelketek: Don't use visual hanging indents-- when using large sets of arguments, start them on the next line with indentation, and close them with the closing paren on a separate line.

jbzdak: OK. I'll check rest of the code for this.

ecommerce/extensions/payment/processors/stripe.py
bradenmacdonald, line 123: I think you should open a quick fix PR against django-oscar.

But for now, we'll need to either (A) ignore this, which should be fine, or (B) store the transaction_id in a new separate field, other than `reference`, on the Otto `Source` model which we can more easily modify then the Oscar `AbstractSource` model.

My feeling is the cleanest way to handle this is just to open the Otto PR, and hope it gets merged soon, and not use ugly workarounds.

Kelketek:
:+1:

ecommerce/extensions/payment/processors/stripe.py
jbzdak, line 123: @bradenmacdonald I think that oscar has some sort of plugin mechanism for models, it won't classify  as an ugly workaround, but if django-oscar would accept this PR this would make things much easier.

jbzdak, line 123: While I'm not suggesting that we should leave this matter open, I just got confirmation that they are not planning to make ids longer in the "near future".

bradenmacdonald, line 123: @jbzdak Great! I still think we should open a Django-Oscar PR to bump up that field length, just to future-proof them and since it's a minor change that should have no ill effects. (Actually, if it's an indexed column and people use utf8mb4 on older versions of MySQL they could hit index size limits :-/ but hopefully that's not the case.)

pomegranited:
@clintonb We're looking at picking this up again for the Ficus upgrade.  Do you have any guidance or in-progress PRs that we should look at to get Stripe integration working with the latest codebase?

clintonb:
@pomegranited I was hoping to have broken out the CyberSource code by now so that the JS that gets loaded is dependent on the processor not hard-coded for CyberSource. Looking at that code, is a good place to start. If you can help split that code, I can provide more details.

Relevant files
* https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/payment/views/cybersource.py
* https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/payment_processors/cybersource.js
* https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/payment/processors/cybersource.py

pomegranited:
Thanks for the info, @clintonb !  Will let you know when/if we restart this work.

clintonb:
I *finally* found time to do some refactoring of the CyberSource processor, and create a Stripe implementation. Now I just need to find some time to finish the unit and acceptance tests. See https://github.com/edx/ecommerce/pull/1147.

PR #604, opened 2016-03-08
[SOL-1684] Remove coupon tests dependence on the migrations

marjev:
JIRA ticket: [SOL-1684](https://openedx.atlassian.net/browse/SOL-1684)

1st review: tagging both @mjfrey and @vkaracic. Please take a look.

2nd review: @rlucioni, I would like you to take a look as well since you reported the issue.

### DIFF
--- a/ecommerce/tests/mixins.py
+++ b/ecommerce/tests/mixins.py
@@ -12,7 +12,6 @@ from django.core.urlresolvers import reverse
 import httpretty
 import jwt
 from mock import patch
-from oscar.apps.catalogue.categories import create_from_breadcrumbs
 from oscar.core.loading import get_class, get_model
 from oscar.test import factories
 from social.apps.django_app.default.models import UserSocialAuth
@@ -31,6 +30,8 @@ Selector = get_class('partner.strategy', 'Selector')
 ShippingEventType = get_model('order', 'ShippingEventType')
 Order = get_model('order', 'Order')
 Partner = get_model('partner', 'Partner')
+ProductAttribute = get_model('catalogue', 'ProductAttribute')
+ProductClass = get_model('catalogue', 'ProductClass')
 User = get_user_model()
 Voucher = get_model('voucher', 'Voucher')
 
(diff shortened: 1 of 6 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=604

vkaracic:
The text in the ticket is a bit vague, is this PR regarding the coupons and basket tests or all tests? Because when I run `make validate` with `DISABLE_MIGRATIONS=1` there are a few tests failing, mostly regarding payment processor tests.

ecommerce/courses/tests/factories.py
```diff
@@ -12,3 +18,13 @@ class Meta(object):
 
     id = FuzzyText(prefix='course-id-')
     name = FuzzyText(prefix='course-name-')
+
+    @classmethod
+    def create(cls, **kwargs):
+        if os.getenv('DISABLE_MIGRATIONS'):
```
clintonb: This is not necessary. [CourseCatalogTestMixin](https://github.com/edx/ecommerce/blob/ff2c224ece949a4d85c41f34235b97ec0c068946/ecommerce/extensions/catalogue/tests/mixins.py#L14) does this for you.

ecommerce/tests/mixins.py
```diff
@@ -298,7 +302,13 @@ class CouponMixin(object):
     """ Mixin for preparing data for coupons and creating coupons. """
     def setUp(self):
         super(CouponMixin, self).setUp()
-        self.category = create_from_breadcrumbs('Coupons > Test category')
+        self.category = factories.CategoryFactory()
+
+        if os.getenv('DISABLE_MIGRATIONS'):
```
clintonb: Remove. This mixin has nothing to do with course seats, and should focus only on coupons. Tests that need to test course seats should inherit [CourseCatalogTestMixin](https://github.com/edx/ecommerce/blob/ff2c224ece949a4d85c41f34235b97ec0c068946/ecommerce/extensions/catalogue/tests/mixins.py#L14).

ecommerce/tests/mixins.py
```diff
@@ -298,7 +302,13 @@ class CouponMixin(object):
...
+            get_or_create_seat_product_class()
+            get_or_create_seat_category()
+            coupon = get_or_create_coupon_product_class()
```
clintonb: Follow the pattern set by [CourseCatalogTestMixin](https://github.com/edx/ecommerce/blob/ff2c224ece949a4d85c41f34235b97ec0c068946/ecommerce/extensions/catalogue/tests/mixins.py#L14). Stop using `if os.getenv('DISABLE_MIGRATIONS')`. Your migrations should be smart enough to detect the existing data. If they aren't, update them.

marjev:
@vkaracic I fixed the coupon tests since I'm familiar only with that part of the code.

ecommerce/courses/tests/factories.py
marjev: Done.

ecommerce/tests/mixins.py
marjev: Done.

ecommerce/tests/mixins.py
marjev: Done.

rlucioni:
Thanks @marjev. This resolves the issues I was seeing with the coupon tests. :+1:

vkaracic:
:+1:

PR #606, opened 2016-03-08
Aamir khan/ecom 3661 update order script

aamir-khan:
Added the management command to updated the order affected in ECOM-3661

@awaisdar001 @Shrhawk @mushtaqak @Qubad786 @Ayub-Khan Please review this PR

### DIFF
--- a/ecommerce/extensions/order/management/commands/update_order.py
+++ b/ecommerce/extensions/order/management/commands/update_order.py
@@ -1,0 +1,56 @@
+"""This scripts remove the data inconsistency of ECOM-3661."""
+
+import logging
+
+from django.core.management import BaseCommand, CommandError
+from django.db import transaction
+from oscar.core.loading import get_model
+
+BasketLine = get_model('basket', 'Line')
+Order = get_model('order', 'Order')
+LinePrice = get_model('order', 'LinePrice')
+PaymentEventQuantity = get_model('order', 'PaymentEventQuantity')
+OrderNote = get_model('order', 'OrderNote')
+logger = logging.getLogger(__name__)
+
+ORDER_NUMBER = "EDX-5266181"
+AMOUNT = 49.0
+
+
+class Command(BaseCommand):
+    help = 'Update the order flagged in ECOM-3661.'
+
+    def handle(self, *args, **options):
+        logger.info("Updating the order [%s] with new amount [%s].", ORDER_NUMBER, AMOUNT)
+        try:
+            order = Order.objects.get(number=ORDER_NUMBER)
+        except Order.DoesNotExist:
+            logger.error("The order [%s] does not exist.", ORDER_NUMBER)
+            raise CommandError("The order {} does not exist.".format(ORDER_NUMBER))
+        with transaction.atomic():
+            order.total_incl_tax = AMOUNT
+            order.total_excl_tax = AMOUNT
+            order.save()
+            order.lines.update(
+                quantity=1,
+                line_price_incl_tax=AMOUNT,
+                line_price_excl_tax=AMOUNT,
+                line_price_before_discounts_incl_tax=AMOUNT,
+                line_price_before_discounts_excl_tax=AMOUNT
+            )
+
+            LinePrice.objects.filter(order=order).update(quantity=1)
+
+            PaymentEventQuantity.objects.filter(line__in=order.lines.all()).update(quantity=1)
+
+            # Make sure it does not update the lines whom baskets were deleted.
+            if order.basket:
+                BasketLine.objects.filter(basket=order.basket).update(quantity=1)
+
+            OrderNote.objects.create(
+                order=order, message="Flagged in Feb audit. The user was charged correctly for one seat on cybersource "
+                                     "but the order was created for two seats. This was due to some error in basket "
+                                     "creation. Updating the order amount to sync with the payments. "
+                                     "Issue detail in ECOM-3661."
+            )
+        logger.info("The order %s updated successfully.", ORDER_NUMBER)
--- a/ecommerce/extensions/order/tests/test_update_order_command.py
+++ b/ecommerce/extensions/order/tests/test_update_order_command.py
@@ -1,0 +1,88 @@
+from django.core.management import call_command, CommandError
+from oscar.core.loading import get_model
+from oscar.test import factories
+
+from ecommerce.tests.testcases import TestCase
+
+BasketLine = get_model('basket', 'Line')
+Order = get_model('order', 'Order')
+LinePrice = get_model('order', 'LinePrice')
+PaymentEventQuantity = get_model('order', 'PaymentEventQuantity')
+OrderNote = get_model('order', 'OrderNote')
+
+LOGGER_NAME = 'ecommerce.extensions.order.management.commands.update_order'
+
+
+class UpdateOrderCommandTests(TestCase):
+    ORDER_NUMBER = "EDX-5266181"
+    AMOUNT = 49.0
+
+    def setUp(self):
+        super(UpdateOrderCommandTests, self).setUp()
+        self.correct_order = factories.create_order()
+        self.affected_order = factories.create_order(number=self.ORDER_NUMBER)
+        self.affected_order.total_incl_tax = 98.0
+        self.affected_order.total_excl_tax = 98.0
+        self.affected_order.save()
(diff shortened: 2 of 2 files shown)


aamir-khan:
@adampalay Please review this PR. I just moved the code to this PR.

ecommerce/extensions/order/management/commands/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'arbisoft'
```
adampalay: should we remove these?

ecommerce/extensions/order/management/commands/update_order.py
```diff
@@ -0,0 +1,65 @@
...
+PaymentEventQuantity = get_model('order', 'PaymentEventQuantity')
+OrderNote = get_model('order', 'OrderNote')
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Update the order flagged in ECOM-3661.'
+
+    option_list = BaseCommand.option_list + (
+        make_option('--order_number',
```
adampalay: guidelines for braking long lines: https://github.com/edx/edx-platform/wiki/Python-Guidelines#breaking-long-lines

ecommerce/extensions/order/management/commands/update_order.py
```diff
@@ -0,0 +1,65 @@
...
+        if not order_number:
+            raise CommandError("Order number is required.")
+        logger.info("Updating the order [%s].", order_number)
+        try:
+            order = Order.objects.get(number=order_number)
+        except Order.DoesNotExist:
+            logger.error("The order [%s] does not exist.", order_number)
+            return
+        with transaction.atomic():
+            order.total_incl_tax = 49
```
adampalay: should this number be configured?

adampalay:
Lines that need coverage:
https://codecov.io/github/edx/ecommerce/ecommerce/extensions/order/management/commands/update_order.py?ref=11ae6ec4f23b753512977d5df2f6eb0258fa6abd

ecommerce/extensions/order/tests/test_update_order_command.py
```diff
@@ -0,0 +1,67 @@
...
+        PaymentEventQuantity.objects.filter(line__in=self.affected_order.lines.all()).update(quantity=2)
+        BasketLine.objects.filter(basket=self.affected_order.basket).update(quantity=2)
+
+    def test_handle(self):
+        """ Verify the management command update the affected order."""
+
+        self.assertEqual(OrderNote.objects.filter(order=self.affected_order).count(), 0)
+
+        # Verify that the command raises exception if order_number is not provided.
+        with self.assertRaises(CommandError):
```
adampalay: L32-33 should be split off into a separate test case

There should also be a test case for an order number that doesn't exist

ecommerce/extensions/order/tests/test_update_order_command.py
```diff
@@ -0,0 +1,67 @@
...
+        self.affected_order.total_excl_tax = 98.0
+        self.affected_order.save()
+        LinePrice.objects.filter(order=self.affected_order).update(quantity=2)
+        PaymentEventQuantity.objects.filter(line__in=self.affected_order.lines.all()).update(quantity=2)
+        BasketLine.objects.filter(basket=self.affected_order.basket).update(quantity=2)
+
+    def test_handle(self):
+        """ Verify the management command update the affected order."""
+
+        self.assertEqual(OrderNote.objects.filter(order=self.affected_order).count(), 0)
```
adampalay: You can use [`.exists()`](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#django.db.models.query.QuerySet.exists) instead of `.count()`

ecommerce/extensions/order/tests/test_update_order_command.py
```diff
@@ -0,0 +1,67 @@
...
+        self.assertEqual(correct_order.total_excl_tax, self.correct_order.total_excl_tax)
+
+        # Verify that the command updated the order lines.
+        for line in affected_order.lines.all():
+            self.assertEqual(line.quantity, 1)
+            self.assertEqual(line.line_price_incl_tax, 49.0)
+            self.assertEqual(line.line_price_excl_tax, 49.0)
+            self.assertEqual(line.line_price_before_discounts_excl_tax, 49.0)
+        correct_order_line = correct_order.lines.all()[0]
+        correct_line_b4_update = self.correct_order.lines.all()[0]
```
adampalay: type out `before` instead of `b4`

adampalay:
@aamir-khan , pass completed!

aamir-khan:
@adampalay Comments are addressed please review.

ecommerce/extensions/order/management/commands/update_order.py
```diff
@@ -0,0 +1,73 @@
...
+        order_number = options.get('order_number')
+        amount = options.get("amount")
+        if not order_number or not amount:
+            raise CommandError("Order number and amount is required.")
+        logger.info("Updating the order [%s] with new amount [%s].", order_number, amount)
+        try:
+            order = Order.objects.get(number=order_number)
+        except Order.DoesNotExist:
+            logger.error("The order [%s] does not exist.", order_number)
+            return
```
adampalay: why not raise a `CommandError` here as well?

ecommerce/extensions/order/management/commands/update_order.py
```diff
@@ -0,0 +1,73 @@
...
+class Command(BaseCommand):
+    help = 'Update the order flagged in ECOM-3661.'
+
+    option_list = BaseCommand.option_list + (
+        make_option(
+            '--order_number',
+            action='store',
+            dest='order_number',
+            default=None,
+            help='Order number to be updated.'),
```
adampalay: nit: breaking long lines

adampalay:
@aamir-khan , pass completed

ecommerce/extensions/order/management/commands/update_order.py
```diff
@@ -0,0 +1,75 @@
...
+    def handle(self, *args, **options):
+        order_number = options.get('order_number')
+        amount = options.get("amount")
+        if not order_number or not amount:
+            raise CommandError("Order number and amount is required.")
+        logger.info("Updating the order [%s] with new amount [%s].", order_number, amount)
+        try:
+            order = Order.objects.get(number=order_number)
+        except Order.DoesNotExist:
+            logger.error("The order [%s] does not exist.", order_number)
```
adampalay: I don't think you'll need the logger and the traceback — the traceback should get logged

adampalay:
one small comment. This also needs a second review

aamir-khan:
@clintonb Can you please review this PR. This script was review by you but now the script is moved to management command.

ecommerce/extensions/order/management/commands/update_order.py
```diff
@@ -0,0 +1,75 @@
...
+LinePrice = get_model('order', 'LinePrice')
+PaymentEventQuantity = get_model('order', 'PaymentEventQuantity')
+OrderNote = get_model('order', 'OrderNote')
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Update the order flagged in ECOM-3661.'
+
+    option_list = BaseCommand.option_list + (
```
clintonb: Why are there options? There is literally one order that needs to be updated.

clintonb:
:+1: pending
- [x] Command line options removed, and values hardcoded.
- [x] Ticket filed to revert this from the repository immediately after it's been run.

ecommerce/extensions/order/management/commands/update_order.py
adampalay: @clintonb , the reason I thought it would be better not to hardcode this stuff was to make sure no sensitive information was disclosed in the open

clintonb: Order numbers aren't sensitive. This should be hardcoded to avoid affecting other orders.

adampalay: Ok, fair enough — is there a list somewhere of info that we should be discreet about and info that we don't have to be discreet about?

clintonb: No such list exists.

aamir-khan:
Created https://openedx.atlassian.net/browse/ECOM-3913 for removing this code once it's been run.

PR #608, opened 2016-03-08
Make credit tests independent of migrations

rlucioni:
These tests formerly relied on data migrations to enable payment processors. The tests now enable payment processors on their own.

@clintonb @marjev


clintonb:
:+1: 
FYI @douglashall

PR #610, opened 2016-03-10
[SOL-1689 & SOL-1688] Basket and coupon offer page analytics

vkaracic:
This PR is a combination of these tickets:
https://openedx.atlassian.net/browse/SOL-1688
https://openedx.atlassian.net/browse/SOL-1689

Contains added analytics for basket checkout page and coupon offer page. We're using the existing analytics logic that the credit app is using with some DRY'ed methods.

### DIFF
--- a/ecommerce/extensions/analytics/tests/test_utils.py
+++ b/ecommerce/extensions/analytics/tests/test_utils.py
@@ -1,0 +1,37 @@
+import json
+
+from django.contrib.auth.models import AnonymousUser
+from django.test import override_settings
+
+from ecommerce.extensions.analytics.utils import prepare_analytics_data
+from ecommerce.tests.testcases import TestCase
+
+
+@override_settings(SEGMENT_KEY='TEST123')
+class UtilsTest(TestCase):
+    """ Tests for the analytics utils. """
+
+    def test_prepare_analytics_data(self):
+        """ Verify the function returns correct analytics data for a logged in user."""
+        user = self.create_user(
+            username='Tester',
+            first_name='John',
+            last_name='Doe',
+            email='test@example.com'
+        )
+        data = prepare_analytics_data('a/b/c', user)
+        self.assertDictEqual(json.loads(data), {
+            'course': {'courseId': 'a/b/c'},
+            'tracking': {'segmentApplicationId': 'TEST123'},
+            'user': {'username': 'Tester', 'name': 'John Doe', 'email': 'test@example.com'}
+        })
+
+    def test_anon_prepare_analytics_data(self):
+        """ Verify the function returns correct analytics data for an anonymous user."""
+        user = AnonymousUser()
+        data = prepare_analytics_data('a/b/c', user)
+        self.assertDictEqual(json.loads(data), {
+            'course': {'courseId': 'a/b/c'},
+            'tracking': {'segmentApplicationId': 'TEST123'},
+            'user': 'AnonymousUser'
+        })
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -1,8 +1,8 @@
 from functools import wraps
+import json
 import logging
 
 from django.conf import settings
-
 
 logger = logging.getLogger(__name__)
 
@@ -85,3 +85,38 @@ def audit_log(name, **kwargs):
     message = u'{name}: {payload}'.format(name=name, payload=payload)
 
     logger.info(message)
+
+
+def prepare_analytics_data(course_id, user):
+    """ Helper function for preparing necessary data for analytics.
+
+    Arguments:
+        course_id (str): The course ID.
+        user (User): The user making the request.
+
+    Returns:
+        JSON object with the data for analytics.
+    """
+    data = {
+        'course': {
+            'courseId': course_id
+        },
+        'tracking': {
+            'segmentApplicationId': settings.SEGMENT_KEY
+        }
+    }
+
+    if user.is_authenticated():
+        user_data = {
+            'user': {
+                'username': user.get_username(),
+                'name': user.get_full_name(),
+                'email': user.email
+            }
+        }
+    else:
+        user_data = {
+            'user': 'AnonymousUser'
+        }
+    data.update(user_data)
+    return json.dumps(data)
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -14,6 +14,7 @@ from edx_rest_api_client.client import EdxRestApiClient
 from slumber.exceptions import SlumberBaseException
 
 from ecommerce.coupons.views import get_voucher_from_code
+from ecommerce.extensions.analytics.utils import prepare_analytics_data
 from ecommerce.extensions.api.data import get_lms_footer
 from ecommerce.extensions.basket.utils import get_certificate_type_display_value, prepare_basket
 from ecommerce.extensions.offer.utils import format_benefit_value
@@ -92,6 +93,10 @@ class BasketSummaryView(BasketView):
                 line.benefit_value = format_benefit_value(benefit)
             else:
                 line.benefit_value = None
+
+            context.update({
+                'analytics_data': prepare_analytics_data(course_id, self.request.user)
+            })
 
         context.update({
             'free_basket': context['order_total'].incl_tax == 0,
--- a/ecommerce/static/js/test/specs/pages/basket_page_spec.js
+++ b/ecommerce/static/js/test/specs/pages/basket_page_spec.js
@@ -3,11 +3,20 @@ define([
         'underscore',
         'pages/basket_page',
         'utils/utils',
+        'utils/analytics_utils',
+        'models/tracking_model',
+        'models/user_model',
+        'views/analytics_view',
     ],
     function ($,
               _,
               BasketPage,
-              Utils) {
+              Utils,
+              AnalyticsUtils,
+              TrackingModel,
+              UserModel,
+              AnalyticsView
+              ) {
         'use strict';
 
         describe('Basket Page', function () {
@@ -18,9 +27,21 @@ define([
                 $('<div id="voucher_form_container"><input id="id_code">' +
                     '<a id="voucher_form_cancel"></a></button></div>' +
                     '<div id="voucher_form_link"><a href=""></a></div>' +
-                    '<div class="payment-buttons"><button class="">' +
-                    '</div>'
+                    '<button type="submit" class="apply_voucher"' +
+                    'data-track-type="click" data-track-event="edx.bi.ecommerce.basket.voucher_applied"' +
+                    'data-voucher-code="ABCDEF"></button></div>' +
+                    '<div class="payment-buttons">' +
+                    '<button data-track-type="click"' +
+                    'data-track-event="edx.bi.ecommerce.basket.payment_selected"' +
+                    'data-track-category="cybersource"' +
+                    'data-processor-name="cybersource"' +
+                    'class="btn btn-success payment-button"' +
+                    'value="cybersource"' +
+                    'id="cybersource"></button></div>'
                 ).appendTo('body');
+
+
+                $('<script type="text/javascript">var initModelData = {};</script>').appendTo('body');
 
                 data = {
                     basket_id: 1,
@@ -75,6 +96,14 @@ define([
                     expect($('#voucher_form_container').is(':visible')).toBeFalsy();
                     expect($('#voucher_form_link').is(':visible')).toBeTruthy();
                 });
+
+                it('should disable payment button before making ajax call', function () {
+                    spyOn(Utils, 'disableElementWhileRunning').and.callThrough();
+                    BasketPage.onReady();
+                    $('button.payment-button').trigger('click');
+                    expect(Utils.disableElementWhileRunning).toHaveBeenCalled();
+                    expect($('button#cybersource').hasClass('is-disabled')).toBeTruthy();
+                });
             });
 
             describe('appendToForm', function () {
@@ -94,17 +123,6 @@ define([
                     expect(error_messages_div.text()).toEqual(
                         'Problem occurred during checkout. Please contact support'
                     );
-                });
-            });
-
-            describe('onReady', function () {
-                it('should disable payment button before making ajax call', function () {
-                    $('<div class="payment-buttons"><button class="payment-button">Pay</button></div>')
-                        .appendTo('body');
-                    spyOn(Utils, 'disableElementWhileRunning');
-                    BasketPage.onReady();
-                    $('button.payment-button').trigger('click');
-                    expect(Utils.disableElementWhileRunning).toHaveBeenCalled();
                 });
             });
 
@@ -128,6 +146,38 @@ define([
                     expect(args.contentType).toEqual('application/json; charset=utf-8');
                     expect(args.headers).toEqual({'X-CSRFToken': cookie});
                     expect(JSON.parse(args.data)).toEqual(data);
+                });
+            });
+
+            describe('Analytics', function() {
+                beforeEach(function () {
+                    spyOn(TrackingModel.prototype, 'isTracking').and.callFake(function() {
+                        return true;
+                    });
+                    spyOn(AnalyticsView.prototype, 'track');
+                    BasketPage.onReady();
+                    spyOn(window.analytics, 'page');
+                });
+
+                it('should trigger voucher applied analytics event', function() {
+                    $('button.apply_voucher').trigger('click');
+                    expect(AnalyticsView.prototype.track).toHaveBeenCalledWith(
+                        'edx.bi.ecommerce.basket.voucher_applied',
+                        { type: 'click' }
+                    );
+                });
+
+                it('should trigger checkout analytics event', function() {
+                    $('button.payment-button').trigger('click');
+                    expect(AnalyticsView.prototype.track).toHaveBeenCalledWith(
+                        'edx.bi.ecommerce.basket.payment_selected',
+                        { category: 'cybersource', type: 'click' }
+                    );
+                });
+
+                it('should trigger page load analytics event', function() {
+                    AnalyticsUtils.analyticsSetUp();
+                    expect(window.analytics.page).toHaveBeenCalled();
                 });
             });
         });
--- a/ecommerce/static/js/test/specs/pages/coupon_offer_spec.js
+++ b/ecommerce/static/js/test/specs/pages/coupon_offer_spec.js
@@ -1,0 +1,49 @@
+define([
+        'jquery',
+        'pages/coupon_offer_page',
+        'models/tracking_model',
+        'models/user_model',
+        'views/analytics_view',
+    ],
+    function($,
+             CouponOfferPage,
+             TrackingModel,
+             UserModel,
+             AnalyticsView
+             ) {
+        'use strict';
+
+        describe('Coupon offer page', function() {
+            beforeEach(function() {
+                $('<a href=""'+
+                'id="PurchaseCertificate"'+
+                'class="btn btn-success btn-purchase"'+
+                'data-track-type="click"'+
+                'data-track-event="edx.bi.ecommerce.coupons.accept_offer"'+
+                'data-track-category="Coupons accepted offer"'+
+                'data-course-id="{{ course.id }}">'+
+                'Purchase Certificate'+
+                '</a>').appendTo('body');
+                $('<script type="text/javascript">var initModelData = {};</script>').appendTo('body');
+            });
+
+            afterEach(function () {
+                $('body').empty();
+            });
+
+            describe('Analytics', function() {
+                it('should trigger purchase certificate event', function() {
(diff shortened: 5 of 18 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=610&repo=edx%2Fecommerce

ecommerce/extensions/basket/views.py
```diff
@@ -93,6 +94,22 @@ def get_context_data(self, **kwargs):
             else:
                 line.benefit_value = None
 
+            context.update({
```
mjfrey, line 97: should we use 
'analytics_data': prepare_analytics_data(course.id, self.request.user)

ecommerce/extensions/analytics/utils.py
```diff
@@ -85,3 +86,30 @@ def audit_log(name, **kwargs):
...
+        JSON object with the data for analytics.
+    """
+    data = {
+        'course': {
+            'courseId': course_id
+        },
+        'tracking': {
+            'segmentApplicationId': settings.SEGMENT_KEY
+        },
+        'user': {
```
mjfrey: Need to account for AnonymousUser

vkaracic:
@mjfrey comments addressed! :)

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -13,6 +13,7 @@
 
 {% block javascript %}
     <script src="{% static 'js/apps/basket_app.js' %}"></script>
+    <script src="{% static 'js/apps/credit_checkout.js' %}"></script>
```
clintonb: This should not be here.

ecommerce/static/js/views/leaving_view.js
```diff
@@ -0,0 +1,29 @@
+define(['underscore', 'backbone'],
```
clintonb: 1. This is more of a utility than a view.
2. Each package and function argument in the initial definition should be on its own line.

ecommerce/templates/coupons/_offer_discount.html
```diff
@@ -30,7 +30,15 @@
             <div class="data-row total-row">{% trans "Total:" %}
                 <div class="text-right">${{new_price}}</div>
             </div>
-            <a href="{% url 'coupons:redeem' %}?code={{ code }}" id="PurchaseCertificate" class="btn btn-success btn-purchase">{% trans "Purchase Certificate" %}</a>
+            <a href="{% url 'coupons:redeem' %}?code={{ code }}"
+                id="PurchaseCertificate"
+                class="btn btn-success btn-purchase"
+                data-track-type="click"
+                data-track-event="edx.bi.ecommerce.coupon_offer.purchased_certificate"
```
clintonb: You cannot purchase a certificate. I'll let you deal with Product regarding the text displayed to the user; but, the event should be named `"edx.bi.ecommerce.coupons.accept_offer"`.

pbaruah:
The button text has been reviewed by product, UX and doc. We are working on making some general improvements to UX and text on that page. It is good for this release.
:+1:

vkaracic:
@rlucioni here's the PR. Please take a look when/if you have time.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -130,6 +145,53 @@ define([
...
                 });
             });
+
+            describe('Analytics', function() {
+                beforeEach(function () {
+                    /* jshint ignore:start */
+                    // jscs:disable
+                    window.analytics = window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","group","track","ready","alias","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t){var e=document.createElement("script");e.type="text/javascript";e.async=!0;e.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(e,n)};analytics.SNIPPET_VERSION="3.0.1";}
```
rlucioni: Your unit tests shouldn't require a real version of Segment's library. Can't you stub it out?

ecommerce/static/js/test/specs/pages/coupon_offer_spec.js
```diff
@@ -0,0 +1,56 @@
...
+                    new CouponOfferPage();
+                    $('a#PurchaseCertificate').trigger('click');
+                    expect(AnalyticsView.prototype.track).toHaveBeenCalledWith(
+                        'edx.bi.ecommerce.coupon_offer.purchased_certificate',
+                        { category: 'coupon_offer_purchase', type: 'click' }
+                    );
+                });
+            });
+    });
+});
```
rlucioni: Nit: missing newline.

ecommerce/static/js/utils/analytics_utils.js
```diff
@@ -0,0 +1,65 @@
...
+                /* jshint ignore:start */
+                // initModelData is set by the Django template at render time.
+                trackingModel.set(initModelData.tracking);
+                userModel.set(initModelData.user);
+                courseModel.set(initModelData.course);
+                /* jshint ignore:end */
+
+                /*
+                 Triggering the analytics events on clicking the payment buttons on checkoutpage.
+                 Buttons has the data-track- attributes type , event and category.
```
rlucioni: Is this comment necessary?

ecommerce/static/js/utils/analytics_utils.js
```diff
@@ -0,0 +1,65 @@
...
+                });
+
+                return {
+                    userModel: userModel,
+                    trackingModel: trackingModel
+                };
+            }
+        };
+    }
+);
```
rlucioni: Nit: missing newline.

ecommerce/static/js/utils/utils.js
```diff
@@ -138,6 +138,25 @@ define([
...
+                /* jshint ignore:start */
+                // jscs:disable
+                window.analytics = window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","group","track","ready","alias","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t){var e=document.createElement("script");e.type="text/javascript";e.async=!0;e.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(e,n)};analytics.SNIPPET_VERSION="3.0.1";}
+                // jscs:enable
+                /* jshint ignore:end */
+
+                window.onbeforeunload = function() {
```
rlucioni: Is this going to end up being logged for all users, regardless of how they leave the checkout page?

ecommerce/static/js/utils/utils.js
```diff
@@ -138,6 +138,25 @@ define([
...
+                    window.analytics.page('Leaving', {
```
rlucioni: The `page` method is meant to track pageviews. See https://segment.com/docs/libraries/analytics.js/#page. The argument should be the name of the page being viewed, and it should be logged when the user visits the page, not when the user leaves.

rlucioni: FYI, `page` should be called by the Segment snippet being loaded onto the page.

ecommerce/templates/coupons/offer.html
```diff
@@ -20,6 +20,12 @@
             {% endif %}
         {% endif %}
     </div>
+
+    {% if analytics_data %}
+        <script type="text/javascript">
+            var initModelData = {{ analytics_data|safe }};
```
rlucioni: Why is the `safe` filter required here? I don't think it's true that this data requires no additional escaping. User-provided data can end up here (username, full name).

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -101,10 +108,25 @@
         <div class="row">
             <div class="payment-buttons col-sm-4" data-basket-id="{{ basket.id }}">
                 {% if free_basket %}
-                <a href="{% url 'checkout:free-checkout' %}" class="btn btn-success checkout-button">{% trans "Place Order" %}</a>
+                <a  href="{% url 'checkout:free-checkout' %}"
+                    data-track-type="click"
+                    data-track-event="edx.bi.ecommerce.basket.free_checkout"
+                    data-track-category="{{ processor.NAME }}"
+                    data-processor-name="{{ processor.NAME }}"
```
rlucioni: Why record a payment processor name for a free checkout?

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -75,6 +94,13 @@ define([
                     expect($('#voucher_form_container').is(':visible')).toBeFalsy();
                     expect($('#voucher_form_link').is(':visible')).toBeTruthy();
                 });
+
+                it('should disable payment button before making ajax call', function () {
+                    spyOn(Utils, 'disableElementWhileRunning');
+                    BasketPage.onReady();
+                    $('button.payment-button').trigger('click');
+                    expect(Utils.disableElementWhileRunning).toHaveBeenCalled();
```
rlucioni, line 104: You should verify that the utility method was called for the right element.

ecommerce/static/js/utils/utils.js
vkaracic: The acceptance criteria is: `Verify that an event is triggered when the learner goes away from the page`.
@mjfrey @pbaruah should we update this criteria to be more precise about which kind of "going away" we want to capture?

ecommerce/templates/coupons/offer.html
vkaracic: It's because credit checkout page uses the same code:
https://github.com/edx/ecommerce/blob/master/ecommerce/templates/edx/credit/checkout.html#L121

Should I remove it there too?

ecommerce/static/js/utils/analytics_utils.js
vkaracic: It's an existing comment: https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/pages/credit_checkout.js#L57-L58
When changing existing code like this I don't necessarily want to remove existing documentation for it. I can remove it if you deem it unnecessary, but for now I'll just fix the grammar.

pbaruah:
Leaving the page - exit the page without clicking either of the checkout buttons. 
Since we have an event to measure the successful outcome of "Clicking on checkout or checkout through Paypal", leaving the page is expected to cover the negative scenario.

vkaracic:
@rlucioni thanks for doing the review. I believe all comments have been addressed (thumbs up reaction on the comment) or answered. Can you take a second look, please?

ecommerce/static/js/utils/utils.js
```diff
@@ -138,6 +138,31 @@ define([
...
+             * @param userModel: The user model from which the username 
+             * is retrieved.
+             */
+            logLeaving: function(userModel) {
+                /* jshint ignore:start */
+                // jscs:disable
+                window.analytics = window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","group","track","ready","alias","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t){var e=document.createElement("script");e.type="text/javascript";e.async=!0;e.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(e,n)};analytics.SNIPPET_VERSION="3.0.1";}
```
clintonb: This is copied from `analytics_view.js`. Avoid duplicating code. Move shared code to a common location.

ecommerce/static/js/utils/utils.js
rlucioni: @vkaracic @pbaruah it doesn't make sense to log a pageview when a user _leaves_ the page. You're tracking pageviews and button clicks. Those two pieces of information are enough to tell you who is accessing the page without clicking a checkout button. This `logLeaving` utility is not necessary.

ecommerce/templates/coupons/offer.html
rlucioni: @vkaracic it was introduced as part of https://github.com/edx/ecommerce/pull/198. Unless there's a clear need for it, I'd remove it.

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -22,8 +22,15 @@
             {% include 'basket/partials/basket_content.html' %}
         </div>
     </div>
+
+    {% if analytics_data %}
+        <script type="text/javascript">
+            var initModelData = {{ analytics_data|safe }};
```
rlucioni: @vkaracic same here. Unless there's a clear need for the `safe` filter, remove it.

ecommerce/extensions/analytics/tests/test_utils.py
```diff
@@ -0,0 +1,37 @@
...
+            "tracking": {"segmentApplicationId": "TEST123"},
+            "user": {"username": "Tester", "name": "John Doe", "email": "test@example.com"}
+        })
+
+    def test_anon_prepare_analytics_data(self):
+        """ Verify the function returns correct analytics data for an anonymous user."""
+        user = AnonymousUser()
+        data = prepare_analytics_data('a/b/c', user)
+        self.assertDictEqual(json.loads(data), {
+            "course": {"courseId": "a/b/c"},
```
rlucioni: Nitpick: please keep your quote levels consistent. Stick to single-quotes unless there's a clear need for doubles.

pbaruah:
Since we are able to calculate the number of people who enters the page and clicks on checkout, we can get the number of people who leaves without clicking checkout. It makes sense to remove the logLeaving logic to trigger an event when a user leaves the page.
:+1:

vkaracic:
@rlucioni @clintonb last comments addressed!

ecommerce/extensions/basket/views.py
clintonb, line 97: Merge this with the call below. There is no need to call `context.update()` twice.

ecommerce/static/js/utils/analytics_utils.js
```diff
@@ -0,0 +1,60 @@
...
+
+                /* jshint ignore:start */
+                // initModelData is set by the Django template at render time.
+                trackingModel.set(initModelData.tracking);
+                userModel.set(initModelData.user);
+                courseModel.set(initModelData.course);
+                /* jshint ignore:end */
+
+                /*
+                 Triggering the analytics events on clicking the payment buttons on checkoutpage.
```
clintonb: Remove this dangling comment.

ecommerce/extensions/basket/views.py
vkaracic, line 97: This is in the for loop, if the basket has no lines (user opens `/basket` or some other reason) it will report an error because `course_id` is defined in the loop.

clintonb, line 97: That's fine, for now. Ideally this should not be in a loop. If we did move a notion of multiple course checkout, only the last course would fire an event.

clintonb:
:+1: once comments addressed, and commits squashed.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -101,10 +107,24 @@
...
+                    data-course-id="{{ course.id }}"
+                    class="btn btn-success checkout-button">
+                    {% trans "Place Order" %}
+                </a>
                 {% else %}
                 {% for processor in payment_processors %}
-                    <button class="btn btn-success payment-button" value="{{ processor.NAME|lower }}" id="{{ processor.NAME|lower }}">
+                    <button data-track-type="click"
+                            data-track-event="edx.bi.ecommerce.basket.payment_selected"
+                            data-track-category="{{ processor.NAME }}"
```
rlucioni: @vkaracic in order for these events to be grouped when viewing them in GA, they should all have the same category value (e.g., `checkout`).

clintonb:
Speaking of squashing, see https://robots.thoughtbot.com/autosquashing-git-commits, if you've never used `fixup`.

rlucioni:
:+1: once commits squashed and final comment addressed.

vkaracic:
@clintonb I've used `fixup` only when interactively rebasing but I didn't know about `--fixup` commiting argument. Thanks for the info!

All comments addressed, thanks everyone for your time and effort!

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -101,10 +107,24 @@
         <div class="row">
             <div class="payment-buttons col-sm-4" data-basket-id="{{ basket.id }}">
                 {% if free_basket %}
-                <a href="{% url 'checkout:free-checkout' %}" class="btn btn-success checkout-button">{% trans "Place Order" %}</a>
+                <a  href="{% url 'checkout:free-checkout' %}"
+                    data-track-type="click"
+                    data-track-event="edx.bi.ecommerce.basket.free_checkout"
+                    data-track-category="Free checkout"
```
rlucioni: @vkaracic you need to update this data attribute, too.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -77,13 +77,19 @@
...
                                 <div id="voucher_form_container">
                                     <form id="voucher_form" action="{% url 'basket:vouchers-add' %}" method="post">
                                         {% csrf_token %}
                                         <div class="code">
                                             <input id="id_code" name="code" type="text" placeholder="{% trans 'Enter a coupon code' %}">
-                                            <button type="submit" class="btn btn-default" data-loading-text="{% trans 'Applying...' %}">{% trans "Apply" %}</button>
+                                            <button type="submit"
+                                                    class="btn btn-default"
+                                                    data-loading-text="{% trans 'Applying...' %}"
+                                                    data-track-type="click"
```
rlucioni, line 88: This event has no category?

ecommerce/templates/coupons/_offer_discount.html
```diff
@@ -30,7 +30,15 @@
             <div class="data-row total-row">{% trans "Total:" %}
                 <div class="text-right">${{new_price}}</div>
             </div>
-            <a href="{% url 'coupons:redeem' %}?code={{ code }}" id="PurchaseCertificate" class="btn btn-success btn-purchase">{% trans "Purchase Certificate" %}</a>
+            <a href="{% url 'coupons:redeem' %}?code={{ code }}"
+                id="PurchaseCertificate"
+                class="btn btn-success btn-purchase"
+                data-track-type="click"
+                data-track-event="edx.bi.ecommerce.coupons.accept_offer"
+                data-track-category="Coupons accepted offer"
```
rlucioni: This category also needs to be updated.

ecommerce/templates/oscar/basket/partials/basket_content.html
vkaracic: Hmm, I was thinking on logging the free checkout as it's own category. I'll change it to checkout.

ecommerce/templates/coupons/_offer_discount.html
vkaracic: To `Accepted coupon offer`? The problem here is that the user doesn't really checkout when the offer is not a free enrollment but gets redirected to the basket page.

rlucioni: @vkaracic you want to be grouping these events using broad, shared terms so that they're easy to find when viewing them in GA. The events you're adding all seem to be related to checkout, hence my suggestion to use the category `checkout`. If that's not the case, consider using a category like `coupons` here.

vkaracic: We decided on `coupon-codes`.

PR #612, opened 2016-03-11
(SOL-1701) Pre-a11y review changes

marjev:
@mjfrey Please take a look at these changes. I did a first pass review of our pages and I saw that some improvements could be made.


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=612&repo=edx%2Fecommerce

mjfrey:
:+1:

PR #614, opened 2016-03-14
[WIP] mattdrayer/WL-330: Multi-site Theming Support

mattdrayer:
@malikshahzad228 @mjfrey FYI


mattdrayer:
@saleem-latif FYI, I have closed this PR because "site" should now be available in the basket thanks to the use of the newly-introduced [django_sites_extensions.middleware.CurrentSiteWithDefaultMiddleware](https://github.com/edx/edx-django-sites-extensions/blob/master/django_sites_extensions/middleware.py#L10) in master, here:  https://github.com/edx/ecommerce/blob/master/ecommerce/settings/base.py#L195:

PR #618, opened 2016-03-15
Management command to update course seat expires

awaisdar001:
Create a management command that updates the expires date for all course seat products. This will ensure that, if there are other pages causing this issue, seats for older courses cannot be added to the basket/purchased.

[ECOM-3829](https://openedx.atlassian.net/browse/ECOM-3829)
Related PR: https://github.com/edx/edx-platform/pull/11809

### DIFF
--- a/ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
+++ b/ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
@@ -1,0 +1,119 @@
+from __future__ import unicode_literals
+import logging
+import time
+from optparse import make_option
+
+from dateutil import parser
+from django.core.management import BaseCommand, CommandError
+from edx_rest_api_client.client import EdxRestApiClient
+from slumber.exceptions import HttpClientError
+
+from ecommerce.core.url_utils import get_lms_url
+from ecommerce.courses.models import Course
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """Update seat expire dates."""
+
+    help = 'Update seat expire with the course enrollment end date.'
+    option_list = BaseCommand.option_list + (
+        make_option('--commit',
+                    action='store_true',
+                    dest='commit',
+                    default=False,
+                    help='Save the data to the database. If this is not set, '
+                         'expires date will not be updated'),
+    )
+
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.DEBUG)
+    logger.addHandler(ch)
+    enrollment_date_not_found = set()
+    pause_time = 5
+    max_tries = 5
+
+    def handle(self, *args, **options):
+        seats_to_update = ['honor', 'audit', 'no-id-professional', 'professional']
+        save_to_db = options.get('commit', False)
+        courses_enrollment_info = self._get_courses_enrollment_info()
+
+        if not courses_enrollment_info:
+            msg = 'No course enrollment information found.'
+            logger.error(msg)
+            raise CommandError(msg)
+
+        courses = Course.objects.all().order_by('id')
+        logger.info('[%d] courses found for update.', courses.count())
+
+        for course in courses:
+            enrollment_end_date = courses_enrollment_info.get(course.id)
+
+            # Only proceed if course enrollment information is present
+            if not enrollment_end_date:
+                logger.error('Enrollment missing for course [%s]', course.id)
+                continue
+
+            if save_to_db:
+                course_seats = course.seat_products.filter(
+                    attributes__name='certificate_type',
+                    attribute_values__value_text__in=seats_to_update
+                )
+                expires = parser.parse(enrollment_end_date)
+                course_seats.update(expires=expires)
+                logger.info(
+                    'Updated expiration date for [%s] seats: [%s]',
+                    course.id,
+                    ', '.join([str(seat.id) for seat in course_seats]),
+                )
+
+    def _get_courses_enrollment_info(self):
+        """
+        Retrieve the enrollment information for all the courses.
+
+        Returns:
+            Dictionary representing the key-value pair (course_key, enrollment_end) of course.
+        """
+        def _parse_response(api_response):
+            response_data = api_response.get('results', [])
+
+            # Map course_id with enrollment end date.
+            courses_enrollment = dict(
+                (course_info['course_id'], course_info['enrollment_end'])
+                for course_info in response_data
+            )
+            return courses_enrollment, api_response['pagination'].get('next', None)
+
+        querystring = {'page_size': 50}
+        api = EdxRestApiClient(get_lms_url('api/courses/v1/'))
+        course_enrollments = {}
+
+        page = 0
+        throttling_attempts = 0
+        next_page = True
+        while next_page:
+            page += 1
+            querystring['page'] = page
+            try:
+                response = api.courses().get(**querystring)
+                throttling_attempts = 0
+            except HttpClientError as exc:
+                # this is a known limitation; If we get HTTP429, we need to pause execution for a few seconds
+                # before re-requesting the data. raise any other errors
+                if exc.response.status_code == 429 and throttling_attempts < self.max_tries:
+                    logger.warning(
+                        'API calls are being rate-limited. Waiting for [%d] seconds before retrying...',
+                        self.pause_time
+                    )
+                    time.sleep(self.pause_time)
+                    page -= 1
+                    throttling_attempts += 1
+                    logger.info('Retrying [%d]...', throttling_attempts)
+                    continue
+                else:
+                    raise
+            enrollment_info, next_page = _parse_response(response)
+            course_enrollments.update(enrollment_info)
+        return course_enrollments
--- a/ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
+++ b/ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
@@ -1,0 +1,217 @@
+# encoding: utf-8
+from __future__ import unicode_literals
+import datetime
+import json
+import logging
+import mock
+from pytz import UTC
+
+import ddt
+from django.core.management import call_command, CommandError
+import httpretty
+from slumber.exceptions import HttpClientError
+from testfixtures import LogCapture
+
+from ecommerce.core.url_utils import get_lms_url
+from ecommerce.courses.models import Product
+from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.tests.testcases import TestCase
+
+logger = logging.getLogger(__name__)
+LOGGER_NAME = 'ecommerce.extensions.catalogue.management.commands.update_course_seat_expire'
+JSON = 'application/json'
+
+
+@ddt.ddt
+class UpdateSeatExpireDateTests(CourseCatalogTestMixin, TestCase):
+    """Tests the update course seat expire command ."""
+
+    def setUp(self):
+        """ Setup course and seats required to run tests """
+        super(UpdateSeatExpireDateTests, self).setUp()
+        now = datetime.datetime.now(tz=UTC)
+        self.expire_date = now - datetime.timedelta(days=7)
+        self.verified_expire_date = now - datetime.timedelta(days=8)
+        self.seats_to_update = ['honor', 'audit', 'no-id-professional', 'professional']
+        self.course = CourseFactory()
+        self.course_info = {
+            'pagination': {},
+            'results': [
+                {
+                    'enrollment_end': unicode(self.expire_date),
+                    'course_id': self.course.id
+                },
+            ],
+        }
+
+        self.honor_seat = self.course.create_or_update_seat('honor', False, 0, self.partner)
+        self.verified_seat = self.course.create_or_update_seat(
+            'verified', False, 500, self.partner, expires=self.verified_expire_date
+        )
+        self.professional_seat = self.course.create_or_update_seat('professional', False, 0, self.partner)
+
+    def mock_courses_api(self, status, body=None):
+        """ Mock Courses API with specific status and body. """
+        self.assertTrue(httpretty.is_enabled(), 'httpretty must be enabled to mock Course API calls.')
+
+        body = body or {}
+        url = get_lms_url('/api/courses/v1/courses/?page_size=1')
+        httpretty.register_uri(
+            httpretty.GET,
+            url,
+            status=status,
+            body=json.dumps(body),
+            content_type=JSON
+        )
+
+    @httpretty.activate
+    def test_update_course_with_commit(self):
+        """ Verify all course seats are updated successfully, when commit option is provided. """
+        seats_expected_to_update = self.course.seat_products.filter(
+            attributes__name='certificate_type',
+            attribute_values__value_text__in=self.seats_to_update
+        )
+        self.mock_courses_api(status=200, body=self.course_info)
+
+        expected = [
+            (
+                LOGGER_NAME,
+                'INFO',
+                '[1] courses found for update.'
+            ),
+            (
+                LOGGER_NAME,
+                'INFO',
+                'Updated expiration date for [{}] seats: [{}]'.format(
+                    self.course.id,
+                    ', '.join([str(seat.id) for seat in seats_expected_to_update])
+                )
+            ),
+        ]
+
+        with LogCapture(LOGGER_NAME) as lc:
+            call_command('update_course_seat_expire', commit=True)
+            lc.check(*expected)
+
+        # Verify course seats have been updated
+        for seat in seats_expected_to_update:
+            fetched_seat = Product.objects.get(id=seat.id)
+            self.assertIsNotNone(fetched_seat.expires)
+            self.assertEqual(fetched_seat.expires, self.expire_date)
+
+        # Verify that 'verified' seat has not been updated.
+        verified_seat = Product.objects.get(id=self.verified_seat.id)
+        self.assertEqual(verified_seat.expires, self.verified_expire_date)
+
+    @httpretty.activate
+    def test_update_course_without_commit(self):
+        """ Verify all course seats are not updated with commit option is not provided. """
+        seats_expected_to_update = self.course.seat_products.filter(
+            attributes__name='certificate_type',
+            attribute_values__value_text__in=self.seats_to_update
+        )
+        self.mock_courses_api(status=200, body=self.course_info)
+
+        expected = [
+            (
+                LOGGER_NAME,
+                'INFO',
+                '[1] courses found for update.'
+            ),
+        ]
+
+        with LogCapture(LOGGER_NAME) as lc:
+            call_command('update_course_seat_expire', commit=False)
+            lc.check(*expected)
+
+        # Verify course seats have not been updated
+        for seat in seats_expected_to_update:
+            fetched_seat = Product.objects.get(id=seat.id)
+            self.assertIsNone(fetched_seat.expires)
+
+        # Verify that 'verified' seat has not been updated.
+        verified_seat = Product.objects.get(id=self.verified_seat.id)
+        self.assertEqual(verified_seat.expires, self.verified_expire_date)
+
+    @httpretty.activate
+    def test_update_course_with_no_data(self):
+        """ Verify all course seats are updated successfully, when commit option is provided"""
+        self.course_info['results'] = []
+        self.mock_courses_api(status=200, body=self.course_info)
+
+        expected = [
+            (
+                LOGGER_NAME,
+                'ERROR',
+                'No course enrollment information found.'
+            )
+        ]
+
+        with LogCapture(LOGGER_NAME) as lc:
+            with self.assertRaises(CommandError):
+                call_command('update_course_seat_expire')
+                lc.check(*expected)
+
+    @httpretty.activate
+    def test_update_course_with_missing_enrolment(self):
+        """
+        Verify that management command logs `enrollment missing` log for all courses
+        which are missing `enrollment_end` date.
+        """
+        self.course_info['results'][0]['enrollment_end'] = None
+        self.mock_courses_api(status=200, body=self.course_info)
+
+        expected = [
+            (
+                LOGGER_NAME,
+                'INFO',
+                '[1] courses found for update.'
+            ),
+            (
+                LOGGER_NAME,
+                'ERROR',
+                'Enrollment missing for course [{}]'.format(self.course.id)
+            )
+        ]
+
+        with LogCapture(LOGGER_NAME) as lc:
+            call_command('update_course_seat_expire')
+            lc.check(*expected)
+
+    @httpretty.activate
+    @mock.patch(
+        'ecommerce.extensions.catalogue.management.commands.update_course_seat_expire.Command.max_tries',
+        new_callable=mock.PropertyMock,
+        return_value=1
+    )
+    @mock.patch(
+        'ecommerce.extensions.catalogue.management.commands.update_course_seat_expire.Command.pause_time',
+        new_callable=mock.PropertyMock,
+        return_value=1
+    )
+    def test_update_course_with_exception(self, mock_max_tries, mock_pause_time):
+        """
+        Verify that management command logs throttling errors when rate-limit to API
+        exceeds.
+        """
+        self.mock_courses_api(status=429, body=self.course_info)
+        expected = [
+            (
+                LOGGER_NAME,
+                'WARNING',
+                'API calls are being rate-limited. Waiting for [1] seconds before retrying...'
+            ),
+            (
+                LOGGER_NAME,
+                'INFO',
+                'Retrying [1]'
+            ),
+        ]
+        with self.assertRaises(HttpClientError):
+            with LogCapture(LOGGER_NAME) as lc:
+                call_command('update_course_seat_expire')
+                lc.check(*expected)
+
+        self.assertEqual(mock_max_tries.call_count, 2)
+        self.assertEqual(mock_pause_time.call_count, 2)


ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,110 @@
...
+        Retrieve the enrollment information for all the courses.
+
+        Arguments:
+            exclude_empty: Boolean value if you want to exclude courses with
+            empty enrollment_end dates
+
+        Returns a key, value pair of (course_key, enrollment_end) course and enrollment info.
+        """
+        try:
+            url = get_lms_url('api/courseoverview/enrollment_dates/?exclude_empty={exclude_empty}').format(
```
clintonb: This is not a real endpoint.

awaisdar001: @clintonb I created a simple API in courseoverview to get the course enrollment end dates.
https://github.com/edx/edx-platform/pull/11809

clintonb: There are numerous issues with this approach. First, your API is neither versioned nor tested. Second, I'm fairly certain we recently got rid of the `course_overviews` API. Finally, you are returning a lot of information in a single API response. Although, you make a lot of calls, this would be much simpler/cleaner if you used the existing Courses API (https://courses.stage.edx.org/api/courses/v1/courses/course-v1:ASUx+AST111+3T2015/).

awaisdar001: @clintonb We would need to make requests for each course right? The main idea for returning all data was to make a single request and fetch only required data. 

So seems like we need to call Courses API for each course to fetch information. Right?

clintonb: Yes. Aside from throttling issues, which we can fix for admin users, there is nothing wrong with making multiple calls.

awaisdar001: Ok, seems like i should switch to courses api.

clintonb: Please do. In general, we should look to use our existing APIs, rather than create new ones. We need to consolidate.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,122 @@
...
+
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.DEBUG)
+    logger.addHandler(ch)
+    enrollment_date_not_found = set()
+    no_data = []
+
+    def handle(self, *args, **options):
+        seats_updated = 0
+        seats_to_update = ['honor', 'audit']
```
clintonb: What about professional education courses?

awaisdar001: @clintonb Looking at this.

awaisdar001:
@clintonb @adampalay  Please review.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
+from __future__ import unicode_literals
+
+from optparse import make_option
+import logging
+from dateutil import parser
+
+from django.core.management import BaseCommand
+
+from edx_rest_api_client.client import EdxRestApiClient
```
clintonb: This import ordering is incorrect. Imports should be grouped into the following categores, and each group should be alphabetized:
- Python
- Third-party
- Project-specific

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+        seats_updated = 0
+        seats_to_update = ['honor', 'audit', 'no-id-professional', 'professional']
+        save_to_db = options.get('commit', False)
+        courses_enrollment_info = self._get_courses_information(page_size=50)
+
+        # If there is some exception fetching the course information, do not proceed.
+        if not courses_enrollment_info:
+            return
+
+        child_products = Product.objects.filter(structure=Product.CHILD)
```
clintonb: Limit to courses product class.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+        logger.info("[%d] Child products found for update.", child_products.count())
```
clintonb: Lowercase "child".

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
clintonb: Iterating over products is incorrect. You will end up making multiple API calls per course. Iterate over `Course.objects.all()` to cap the number of API calls.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+        self.log_enrollment_missing()
+
+    def log_enrollment_missing(self):
+        """ log courses which are missing enrollment end date """
+        logger.info("Enrollment missing for: [%d] courses", len(self.enrollment_date_not_found))
+
+        for course_key in self.enrollment_date_not_found:
+            logger.info("%s", course_key)
+
+    def _get_courses_information(self, **querystring):
```
clintonb: You are always calling this with `page_size=50`. Hardcode that in the call below, and avoid the hassle of the kwarg.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+                for course_info in response_data
+            )
+            return courses_enrollment, api_response['pagination'].get('next', None)
+
+        querystring = {} if not querystring else querystring
+        api = EdxRestApiClient(get_lms_url('api/courses/v1/'))
+        course_enrollments = {}
+        try:
+            page = 1
+            response = api.courses().get(**querystring)
```
clintonb: Remove this duplicate code. Set `next_page` to some non-false value.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+            course_enrollments.update(enrollment_info)
+            while next_page:
+                page += 1
+                querystring['page'] = page
+                response = api.courses().get(**querystring)
+                enrollment_info, next_page = _parse_response(response)
+                course_enrollments.update(enrollment_info)
+            return course_enrollments
+
+        except Exception:  # pylint: disable=broad-except
```
clintonb: Let the exception bubble up. We cannot recover from it.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+        course_enrollments = {}
+        try:
+            page = 1
+            response = api.courses().get(**querystring)
+            enrollment_info, next_page = _parse_response(response)
+            course_enrollments.update(enrollment_info)
+            while next_page:
+                page += 1
+                querystring['page'] = page
+                response = api.courses().get(**querystring)
```
clintonb: You need to account for throttling errors.

ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
```diff
@@ -0,0 +1,145 @@
...
+from __future__ import unicode_literals
+import datetime
+import json
+import logging
+
+import ddt
+import httpretty
+import mock
+from pytz import UTC
+
```
clintonb, line 8: Remove, and group all third-party imports.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+
+        logger.info("[%d] seats updated", seats_updated)
+        self.log_enrollment_missing()
+
+    def log_enrollment_missing(self):
+        """ log courses which are missing enrollment end date """
+        logger.info("Enrollment missing for: [%d] courses", len(self.enrollment_date_not_found))
+
+        for course_key in self.enrollment_date_not_found:
+            logger.info("%s", course_key)
```
clintonb: No need for string interpolation.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+            # Only proceed if course enrollment information is present
+            if not enrollment_end_date:
+                self.enrollment_date_not_found.add(product.course_id)
+                continue
+
+            if not product.expires and seat_mode in seats_to_update:
+                product.expires = parser.parse(enrollment_end_date)
+
+                if save_to_db:
+                    logger.info(
```
clintonb: This is noisy. Do it once per course, not seat.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+        for product in child_products:
+            seat_mode = mode_for_seat(product)
+            enrollment_end_date = courses_enrollment_info.get(product.course.id)
+
+            # Only proceed if course enrollment information is present
+            if not enrollment_end_date:
+                self.enrollment_date_not_found.add(product.course_id)
+                continue
+
+            if not product.expires and seat_mode in seats_to_update:
```
clintonb: The existence of `product.expires` doesn't matter. We want to sync the latest data.

clintonb:
I will re-review once the management command has been updated.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
adampalay: Hey Clinton, is there an ecommerce style guide we should be reviewing before submitting PRs?

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def handle(self, *args, **options):
+        seats_updated = 0
+        seats_to_update = ['honor', 'audit', 'no-id-professional', 'professional']
+        save_to_db = options.get('commit', False)
+        courses_enrollment_info = self._get_courses_information(page_size=50)
+
+        # If there is some exception fetching the course information, do not proceed.
+        if not courses_enrollment_info:
+            return
```
adampalay: throw a `CommandError` instead, to be more descriptive?

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
clintonb: - https://google.github.io/styleguide/pyguide.html
- https://www.python.org/dev/peps/pep-0008/

ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
```diff
@@ -0,0 +1,145 @@
...
+        with LogCapture(LOGGER_NAME) as lc:
+            call_command('update_course_seat_expire', commit=commit)
+            lc.check(*expected)
+
+        # Verify course seats
+        verified_seat = Product.objects.get(id=verified_seat.id)
+        for seat in seats_expected_to_update:
+            fetched_seat = Product.objects.get(id=seat.id)
+            if commit:
```
adampalay: I'm not a huge fan of if/else statements in tests — there's not a good way to confirm that every line is being hit in a test otherwise. May make sense just to split out into two tests?

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
awaisdar001: @clintonb `self._get_courses_information()` method fetches all the courses information and returns key, value structure `courses_enrollment_info` as `course_id` and `enrollment_end`. 

We are not making additional API calls while executing `courses_enrollment_info.get(seat.course.id)`

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
awaisdar001: Can you please elaborate how can I account the throttling errors?

clintonb: If you get HTTP 429, you will need to pause execution for a few seconds before re-requesting the data.

http://www.django-rest-framework.org/api-guide/throttling/

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
clintonb: Okay. Nevertheless, we have more than just course products. You must filter by product class or iterate only over the `Course` objects.

awaisdar001: I am on my way of updating it to iterate over courses. 
`courses = Course.objects.all()` and then only fetching the seats we are required to update using filtering. and updating them using `.update()` of `Queryset`

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
awaisdar001: Ok, Couple of questions.
1. How can I know if an API view is handling Throttling? I looked at [CourseListView](https://github.com/edx/edx-platform/blob/5df15fcf0339e8378df3afe0daad60aa408121a0/lms/djangoapps/course_api/views.py#L122) but coudln't get a clue.
2. So seems like upon Throttling, the `response = api.courses().get(**querystring)` will raise an `Exception`? and we need to pause the execution for that time?

clintonb: 1. See the link I sent.
2. Catch [`HttpClientError`](https://github.com/samgiles/slumber/blob/master/slumber/exceptions.py#L21) and check the `status_code`.

awaisdar001:
@clintonb Updated management command.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
+from __future__ import unicode_literals
+from dateutil import parser
+import time
+import logging
```
clintonb, line 2: Alphabetize.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
...
+
+        for course in courses:
+            enrollment_end_date = courses_enrollment_info.get(course.id)
+
+            # Only proceed if course enrollment information is present
+            if not enrollment_end_date:
+                self.enrollment_date_not_found.add(course.id)
+                continue
+
+            course_seats = course.seat_products.filter(
```
clintonb: Move this inside the `if` statement below.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
...
+        seats_to_update = ['honor', 'audit', 'no-id-professional', 'professional']
+        save_to_db = options.get('commit', False)
+        courses_enrollment_info = self._get_courses_information()
+
+        # If not data is fetched from LMS API, do not proceed.
+        if not courses_enrollment_info:
+            logger.error("No course enrollment information found.")
+            return
+
+        courses = Course.objects.all()
```
clintonb: `order_by('id')`

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
...
+            if not enrollment_end_date:
+                self.enrollment_date_not_found.add(course.id)
+                continue
+
+            course_seats = course.seat_products.filter(
+                attributes__name='certificate_type',
+                attribute_values__value_text__in=seats_to_update
+            )
+            if save_to_db:
+                logger.info(
```
clintonb: Move this after the update operation. The message should be: `Updated expiration date for [{course_id}] seats: [{comma_separated_product_id_list}]`.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
...
+                    "Updating (expires): [%s], for seats: [%s] to: [%s]",
+                    unicode(course),
+                    unicode(course_seats),
+                    enrollment_end_date
+                )
+                expires = parser.parse(enrollment_end_date)
+                course_seats.update(expires=expires)
+                seats_updated += len(course_seats)
+
+        logger.info("[%d] seats updated", seats_updated)
```
clintonb: We don't really need to track to this level of detail. This can go away.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
...
+
+        courses = Course.objects.all()
+        logger.info("[%d] courses found for update.", courses.count())
+
+        for course in courses:
+            enrollment_end_date = courses_enrollment_info.get(course.id)
+
+            # Only proceed if course enrollment information is present
+            if not enrollment_end_date:
+                self.enrollment_date_not_found.add(course.id)
```
clintonb: Log the error here. Don't wait until after we have iterated over every course.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
...
+                    unicode(course),
+                    unicode(course_seats),
+                    enrollment_end_date
+                )
+                expires = parser.parse(enrollment_end_date)
+                course_seats.update(expires=expires)
+                seats_updated += len(course_seats)
+
+        logger.info("[%d] seats updated", seats_updated)
+        self.log_enrollment_missing()
```
clintonb: Remove this. Log the issue earlier.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
...
+        logger.info("Enrollment missing for: [%d] courses", len(self.enrollment_date_not_found))
+
+        for course_key in self.enrollment_date_not_found:
+            logger.info(course_key)
+
+    def _get_courses_information(self):
+        """
+        Retrieve the enrollment information for all the courses.
+
+        Arguments:
```
clintonb: Remove.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
...
+        course_enrollments = {}
+
+        page = 0
+        next_page = True
+        while next_page:
+            page += 1
+            querystring['page'] = page
+            try:
+                response = api.courses().get(**querystring)
+            except HttpClientError, exc:
```
clintonb: `except HttpClientError as exc:`

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,125 @@
...
+                # this is a known limitation; If we get HTTP429, we need to pause execution for a few seconds
+                # before re-requesting the data.
+                status_code = exc.response.status_code
+                logger.warning("HttpClientError with status_code [%d].", status_code)
+                if status_code == 429:
```
clintonb: This is the only case we care about. If you get other types of errors, you should raise them, not continue.

``` python
if exc.response.status_code != 429:
  raise
else:
  logger.warning('API calls are being rate-limited. Waiting for [%d] seconds before retrying...', self.pause_time)
  time.sleep(self.pause_time)
  page -= 1
  logger.info('Retrying.')
  continue
```

awaisdar001:
@clintonb Comments addressed. Please review.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def handle(self, *args, **options):
+        seats_to_update = ['honor', 'audit', 'no-id-professional', 'professional']
+        save_to_db = options.get('commit', False)
+        courses_enrollment_info = self._get_courses_information()
+
+        # If not data is fetched from LMS API, do not proceed.
+        if not courses_enrollment_info:
+            logger.error("No course enrollment information found.")
+            return
```
clintonb:
``` python
msg = 'No course enrollment information found.'
logger.error(msg)
raise CommandError(msg)
```

Raise an error to ensure the correct exit code is returned. Simply returning, with no exit code, implies the command executed successfully.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,116 @@
...
+
+        courses = Course.objects.all().order_by('id')
+        logger.info("[%d] courses found for update.", courses.count())
+
+        for course in courses:
+            enrollment_end_date = courses_enrollment_info.get(course.id)
+
+            # Only proceed if course enrollment information is present
+            if not enrollment_end_date:
+                logger.error("Enrollment missing for course: [%s]", course.id)
```
clintonb: No need for the colon.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,116 @@
...
+            if save_to_db:
+                course_seats = course.seat_products.filter(
+                    attributes__name='certificate_type',
+                    attribute_values__value_text__in=seats_to_update
+                )
+                expires = parser.parse(enrollment_end_date)
+                course_seats.update(expires=expires)
+                logger.info(  # todo
+                    "Updated expiration date for [%s] seats: [%s]",
+                    unicode(course),
```
clintonb: `course.id`

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,116 @@
...
+                continue
+
+            if save_to_db:
+                course_seats = course.seat_products.filter(
+                    attributes__name='certificate_type',
+                    attribute_values__value_text__in=seats_to_update
+                )
+                expires = parser.parse(enrollment_end_date)
+                course_seats.update(expires=expires)
+                logger.info(  # todo
```
clintonb: Remove the `# todo`.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,116 @@
...
+        course_enrollments = {}
+
+        page = 0
+        next_page = True
+        while next_page:
+            page += 1
+            querystring['page'] = page
+            try:
+                response = api.courses().get(**querystring)
+            except HttpClientError as exc:
```
clintonb, line 102: This needs to be tested.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,116 @@
...
+    pause_time = 5
+
+    def handle(self, *args, **options):
+        seats_to_update = ['honor', 'audit', 'no-id-professional', 'professional']
+        save_to_db = options.get('commit', False)
+        courses_enrollment_info = self._get_courses_information()
+
+        # If not data is fetched from LMS API, do not proceed.
+        if not courses_enrollment_info:
+            logger.error("No course enrollment information found.")
```
clintonb: Single quotes.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,116 @@
...
+            return
+
+        courses = Course.objects.all().order_by('id')
+        logger.info("[%d] courses found for update.", courses.count())
```
clintonb: Single quotes.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
clintonb: Single quotes.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,116 @@
...
+
+            if save_to_db:
+                course_seats = course.seat_products.filter(
+                    attributes__name='certificate_type',
+                    attribute_values__value_text__in=seats_to_update
+                )
+                expires = parser.parse(enrollment_end_date)
+                course_seats.update(expires=expires)
+                logger.info(  # todo
+                    "Updated expiration date for [%s] seats: [%s]",
```
clintonb: Single quotes.

ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
```diff
@@ -0,0 +1,181 @@
...
+            ],
+        }
+
+        self.honor_seat = self.course.create_or_update_seat('honor', False, 0, self.partner)
+        self.verified_seat = self.course.create_or_update_seat(
+            'verified', False, 500, self.partner, expires=self.expire_date
+        )
+        self.professional_seat = self.course.create_or_update_seat('professional', False, 0, self.partner)
+
+    def tearDown(self):
```
clintonb: This is not necessary.

ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
```diff
@@ -0,0 +1,181 @@
...
+
+logger = logging.getLogger(__name__)
+LOGGER_NAME = 'ecommerce.extensions.catalogue.management.commands.update_course_seat_expire'
+JSON = 'application/json'
+Product = get_model('catalogue', 'Product')
+Course = get_model('courses', 'Course')
+
+
+@ddt.ddt
+class UpdateSeatExpireDateTests(CourseCatalogTestMixin, TransactionTestCase):
```
clintonb: Why `TransactionTestCase` instead of `TestCase`?

ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
```diff
@@ -0,0 +1,181 @@
...
+        self.honor_seat = self.course.create_or_update_seat('honor', False, 0, self.partner)
+        self.verified_seat = self.course.create_or_update_seat(
+            'verified', False, 500, self.partner, expires=self.expire_date
+        )
+        self.professional_seat = self.course.create_or_update_seat('professional', False, 0, self.partner)
+
+    def tearDown(self):
+        super(UpdateSeatExpireDateTests, self).tearDown()
+
+    def _mock_lms_api(self, status, body=None):
```
clintonb: `mock_course_api`

ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
```diff
@@ -0,0 +1,181 @@
...
+        """ Mock LMS API with specific status and body. """
```
clintonb: s/LMS/Course

ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
```diff
@@ -0,0 +1,181 @@
...
+        self.assertTrue(httpretty.is_enabled(), 'httpretty must be enabled to mock Commerce API calls.')
```
clintonb: s/Commerce/Course

ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # Verify course seats have been updated
+        for seat in seats_expected_to_update:
+            fetched_seat = Product.objects.get(id=seat.id)
+            self.assertIsNotNone(fetched_seat.expires)
+            self.assertEqual(fetched_seat.expires, self.expire_date)
+
+        # Verify that 'verified' seat has not been updated.
+        verified_seat = Product.objects.get(id=self.verified_seat.id)
+        self.assertEqual(verified_seat.expires, self.expire_date)
```
clintonb: The verified seat should have a different expiration date; otherwise, you don't truly know if the date was untouched.

awaisdar001:
@clintonb I have changed a little in the management command. 
I am using a new variable `max_tries` and keep track to make finite tries when throttling occurs.
I also added another test `test_update_course_with_exception` as you requested. 

I have addressed the comments also, Please have a look.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,120 @@
...
+from dateutil import parser
+from django.core.management import BaseCommand, CommandError
+from django.db.models.loading import get_model
+from edx_rest_api_client.client import EdxRestApiClient
+from slumber.exceptions import HttpClientError
+
+from ecommerce.settings import get_lms_url
+
+logger = logging.getLogger(__name__)
+Course = get_model('courses', 'Course')
```
clintonb: Load the `Course` model like any other. It's not an Oscar model.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,120 @@
...
+    enrollment_date_not_found = set()
+    pause_time = 5
+    max_tries = 5
+
+    def handle(self, *args, **options):
+        seats_to_update = ['honor', 'audit', 'no-id-professional', 'professional']
+        save_to_db = options.get('commit', False)
+        courses_enrollment_info = self._get_courses_information()
+
+        # If not data is fetched from LMS API, do not proceed.
```
clintonb: This comment is not necessary.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,120 @@
...
+                    course.id,
+                    ', '.join([str(seat.id) for seat in course_seats]),
+                )
+
+    def _get_courses_information(self):
+        """
+        Retrieve the enrollment information for all the courses.
+
+        Returns:
+            key, value pair of (course_key, enrollment_end) course and enrollment info.
```
clintonb: Dictionary

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,120 @@
...
+        next_page = True
+        while next_page:
+            page += 1
+            querystring['page'] = page
+            try:
+                response = api.courses().get(**querystring)
+            except HttpClientError as exc:
+                # this is a known limitation; If we get HTTP429, we need to pause execution for a few seconds
+                # before re-requesting the data. raise any other errors
+                status_code = exc.response.status_code
```
clintonb: No need for this variable.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,120 @@
...
+                for course_info in response_data
+            )
+            return courses_enrollment, api_response['pagination'].get('next', None)
+
+        querystring = {'page_size': 50}
+        api = EdxRestApiClient(get_lms_url('api/courses/v1/'))
+        course_enrollments = {}
+
+        page = 0
+        throttling_tries = 0
```
clintonb: `throttling_attempts`

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,120 @@
...
+        next_page = True
+        while next_page:
+            page += 1
+            querystring['page'] = page
+            try:
+                response = api.courses().get(**querystring)
```
clintonb, line 100: Set `throttling_attempts` to zero after this call.

clintonb:
:+1: once comments addressed, and tests pass.

awaisdar001:
@clintonb Comments addressed. 
@adampalay, @aamir-khan  Please review.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,118 @@
...
+    ch.setLevel(logging.DEBUG)
+    logger.addHandler(ch)
+    enrollment_date_not_found = set()
+    pause_time = 5
+    max_tries = 5
+
+    def handle(self, *args, **options):
+        seats_to_update = ['honor', 'audit', 'no-id-professional', 'professional']
+        save_to_db = options.get('commit', False)
+        courses_enrollment_info = self._get_courses_information()
```
aamir-khan: This is not the course's enrollment related information. It is confusing can you rename it to just courses_info

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,118 @@
...
+
+        if not courses_enrollment_info:
+            msg = 'No course enrollment information found.'
+            logger.error(msg)
+            raise CommandError(msg)
+
+        courses = Course.objects.all().order_by('id')
+        logger.info('[%d] courses found for update.', courses.count())
```
aamir-khan, line 49: I think you should use `len(courses)` instead of count() because count makes a db call.

ecommerce/extensions/catalogue/tests/test_update_course_seat_expire.py
```diff
@@ -0,0 +1,217 @@
...
+import logging
+import mock
+from pytz import UTC
+
+import ddt
+from django.core.management import call_command, CommandError
+import httpretty
+from slumber.exceptions import HttpClientError
+from testfixtures import LogCapture
+from ecommerce.courses.models import Product
```
aamir-khan: Align this to below group

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
clintonb, line 49: Both make a database call. If you only need the count, use count().

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
awaisdar001: Good point Aamir. Actually `self._get_courses_information()` is returning enrollment information, so I think I should name it to something `self._get_courses_enrollment_info()`

awaisdar001:
@aamir-khan Comments addressed

awaisdar001:
@adampalay @aamir-khan PR updated.

awaisdar001:
@mushtaqak Aamir is out, Can you please take a look.

ecommerce/extensions/catalogue/management/commands/update_course_seat_expire.py
```diff
@@ -0,0 +1,119 @@
...
+            if not enrollment_end_date:
+                logger.error('Enrollment missing for course [%s]', course.id)
+                continue
+
+            if save_to_db:
+                course_seats = course.seat_products.filter(
+                    attributes__name='certificate_type',
+                    attribute_values__value_text__in=seats_to_update
+                )
+                expires = parser.parse(enrollment_end_date)
```
mushtaqak, line 64: I think `if save_to_db:` should be below this line as this is the only thing that is updating the course seat. If update, we should log, seat updated. 

If I do not give commit option, that means I am just dry-running the command. So, I would still want to get the the courses which **could be affected** by this command. The log should always be printed.

mushtaqak:
One suggestion, it is upto you if you want to do it :) 

Nice work :+1:  from me too

PR #619, opened 2016-03-15
Updated checkout page, now each provider can have their own fulfillment instructions.

waheedahmed:
[ECOM-2934](https://openedx.atlassian.net/browse/ECOM-2934)

Now each credit provider can have their own fulfillment instructions (which now includes a paragraph that was previously not parameterized)
## Screenshots:
- Collapsed
  screen shot 2016-03-17 at 3 21 24 pm
- Expanded
  screen shot 2016-03-17 at 3 19 37 pm

@zubair-arbi @tasawernawaz @awais786 Please review this.

### DIFF
--- a/ecommerce/templates/edx/credit/_provider_detail.html
+++ b/ecommerce/templates/edx/credit/_provider_detail.html
@@ -29,39 +29,7 @@
             </button>
 
             <div class="collapse" id="fulfillment-instructions-{{ provider.id|lower }}">
-                <p>
-                    {% blocktrans with display_name=provider.display_name %}
-                        You can purchase credit through {{ display_name }} for this course for up to one year
-                        after the course end date. You can use the credit at {{ display_name }} or transfer the
-                        credit to another institution that accepts the credit.
-                    {% endblocktrans %}
-                </p>
-
-                <p>{% trans "To purchase course credit:" %}</p>
-
-                <ol>
-                    <li>
-                        {% trans "On this page, select a checkout option." %}
-                    </li>
-                    <li>{% trans "On the payment page, enter your payment information." %}</li>
-                    <li>
-                        {% blocktrans with display_name=provider.display_name %}
-                            When your payment is complete, follow the link to the {{ display_name }} website.
-                        {% endblocktrans %}
-                    </li>
-                    <li>
-                        {% blocktrans with display_name=provider.display_name %}
-                            On the {{ display_name }} website, enter all required information to request your course
-                            credit.
-                        {% endblocktrans %}
-                    </li>
-                </ol>
-                <p>
-                    {% blocktrans with display_name=provider.display_name %}
-                        After {{ display_name }} approves your credit, you can request an official transcript from
-                        {{ display_name }}.
-                    {% endblocktrans %}
-                </p>
+                {{ provider.fulfillment_instructions|safe }}
             </div>
         </div>
         <div class="col-sm-2 col-xs-2 hidden-xs">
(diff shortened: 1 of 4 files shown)


tasawernawaz:
lgtm :+1:

waheedahmed:
@rlucioni Please review this.

ecommerce/templates/edx/credit/_provider_detail.html
```diff
@@ -29,39 +29,7 @@ <h3 class="title">{{ provider.display_name }}</h3>
...
-                        {% endblocktrans %}
-                    </li>
-                </ol>
-                <p>
-                    {% blocktrans with display_name=provider.display_name %}
-                        After {{ display_name }} approves your credit, you can request an official transcript from
-                        {{ display_name }}.
-                    {% endblocktrans %}
-                </p>
+                {{ provider.fulfillment_instructions|safe }}
```
rlucioni, line 32: @waheedahmed I'm a little fuzzy on the details here, but does it really make sense to include a big HTML string in `fulfillment_instructions`? I took a look at the description of ECOM-2934, and it sounds like the goal is to allow parametrization of one paragraph in the fulfillment instructions. Instead of replacing the contents of this div, why not add a paragraph with the additional content?

Relatedly, where are `fulfillment_instructions` configured/specified for each provider?

waheedahmed, line 32: @rlucioni I don't think so, because the provided fulfillment instruction's (check the comment on ticket for provided parameters) have everything not only a single paragraph. And the goal of this story is that each provider can have their own fulfillment instructions not only a single paragraph.

rlucioni, line 32: @waheedahmed where would the HTML for these instructions be entered? I'm guessing it's the Django admin, in which case I'd encourage you to think about a more structured way of doing this. I think entering raw HTML in the admin is an antipattern.

waheedahmed, line 32: @rlucioni I'm confused here, I did exactly according to the ticket and we are also using the same approach in web certificates thats why I did this.
I don't know how we are going to get whole text from db and put different HTML tags around different lines of text and then render. Can you please suggest a better way to implement this?

FYI @zubair-arbi @marcotuts

marcotuts, line 32: @rlucioni That (rat HTML in the admin) may be something we don't want to be doing, but considerations for that aren't necessarily in scope for this story.

rlucioni, line 32: @waheedahmed what you have here will work, but in the future we should try to avoid this pattern. It should be possible to break fulfillment instructions down into a variety of smaller fields which we can then inject into a single template tracked in the repo.

rlucioni:
:+1:

ecommerce/templates/edx/credit/_provider_detail.html
waheedahmed, line 32: @rlucioni Sure thing, thanks!

waheedahmed:
@rlucioni Can you please help me finding the reason why the build is failing after rebasing with master?

vkaracic:
@waheedahmed pylint started failing for a couple of unrelated files:
https://travis-ci.org/edx/ecommerce/builds/117647673#L1050

This was happening to me too, so I added https://github.com/edx/ecommerce/pull/627/commits/70c76e156c5594ace719286c821700493f87e2ba until we get to the bottom of the sudden failures.

waheedahmed:
@vkaracic Should I cherry-pick this? If yes, can I merge my PR after this?

vkaracic:
Pylint's `bad-super-call` exception is this case is a false-positive and I'm not quite sure why it's happening now. You can cherry-pick+merge this. I can remove the pylint disable's once we figure out why it's doing this.

@rlucioni @clintonb thoughts?

waheedahmed:
@vkaracic OK, thanks!

PR #621, opened 2016-03-16
[ON HOLD] [SOL-1679] Cancel payment view

vkaracic:
https://openedx.atlassian.net/browse/SOL-1679

When a user cancels his purchase in one of the payment processors' payment pages they get redirected to the `cancel_path` in each of the processors' configuration. It used to be that the user gets redirect to a page on LMS saying that his order has been cancelled.
We want that, for whatever reason, the user cancels on the payment processor's page, she gets redirect to the basket page and her basket to be thawed for further use.
- [ ] change cancel URL in the settings
- [ ] change `cancel_page_url` properties in payment processors

---
- [ ] confirm the fix on the sandbox

### DIFF
--- a/ecommerce/extensions/checkout/app.py
+++ b/ecommerce/extensions/checkout/app.py
@@ -5,10 +5,12 @@ from oscar.core.loading import get_class
 
 class CheckoutApplication(app.CheckoutApplication):
     free_checkout = get_class('checkout.views', 'FreeCheckoutView')
+    cancel_response = get_class('checkout.views', 'CancelResponseView')
 
     def get_urls(self):
         urls = [
             url(r'^free-checkout/$', self.free_checkout.as_view(), name='free-checkout'),
+            url(r'^cancel/(?P<basket_id>\d+)/$', self.cancel_response.as_view(), name='cancel'),
 
             url(r'^$', self.index_view.as_view(), name='index'),
 
--- a/ecommerce/extensions/checkout/views.py
+++ b/ecommerce/extensions/checkout/views.py
@@ -7,6 +7,7 @@ import logging
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
 from django.core.urlresolvers import reverse
+from django.shortcuts import get_object_or_404
 from django.utils.decorators import method_decorator
 from django.views.generic import RedirectView
 from oscar.apps.checkout.views import *  # pylint: disable=wildcard-import, unused-wildcard-import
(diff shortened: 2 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=621

ecommerce/extensions/checkout/app.py
```diff
@@ -5,10 +5,12 @@
 
 class CheckoutApplication(app.CheckoutApplication):
     free_checkout = get_class('checkout.views', 'FreeCheckoutView')
+    cancel_response = get_class('checkout.views', 'CancelResponseView')
 
     def get_urls(self):
         urls = [
             url(r'^free-checkout/$', self.free_checkout.as_view(), name='free-checkout'),
+            url(r'^cancel-payment/(?P<basket_id>\d+)/$', self.cancel_response.as_view(), name='cancel_payment'),
```
vkaracic: Perhaps change this to just `cancel`.

ecommerce/extensions/checkout/views.py
```diff
@@ -52,3 +53,21 @@ def get_redirect_url(self, *args, **kwargs):
...
+    """ Handles behaviour for the 'cancel' redirect. """
+    permanent = False
+
+    def get(self, request, *args, **kwargs):
+        """ Retrieves the previously frozen basket from the kwargs and thaws it. """
+        basket = get_object_or_404(Basket, id=kwargs['basket_id'],
+                                   status=Basket.FROZEN)
+        basket.thaw()
+        logger.info('Payment cancelled (token %s) - basket #%s thawed',
+                    request.GET.get('token', '<no token>'), basket.id)
```
clintonb, line 68: What is this token?

vkaracic, line 68: My guess is a token identifier returned from the payment processor, I'll check it once I get a chance to test it on a sandbox.

ecommerce/extensions/checkout/app.py
clintonb: Yes.

ecommerce/extensions/checkout/views.py
vkaracic, line 68: token identifier identifying the transaction*

clintonb, line 68: 1. Try: Payment for basket [%d] was cancelled. Transaction ID is [%s]. 
2. Let me know if you need help running this locally. No need to wait for sandboxes.

clintonb:
Looks good. You just need to update the payment processors/settings.

vkaracic:
About the settings part, the only place I found I should change is this line:
https://github.com/edx/ecommerce/blob/master/ecommerce/settings/_oscar.py#L106

Is there some setting that I'm missing perhaps?

clintonb:
@vkaracic You want to remove all existing settings for the cancel page since they all point to LMS. You will also want to update the values in `PAYMENT_PROCESSOR_CONFIG`, removing them.

The base payment processor should be updated to generate a URL using only a given basket and the new view (e.g. `get_cancel_url(self, basket)`).

edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=621

vkaracic:
@JoyceZhu there is some code here that might be of use to you. If there is no, can you please let me know, or delete this branch?

JoyceZhu:
@vkaracic , I can definitely use this in my work -- thanks for letting me know!

PR #623, opened 2016-03-16
Revert "WL-286 Add support for multitenant OAuth"

douglashall:
Reverts edx/ecommerce#543

There is an issue we need to resolve with how we perform health checks against the application in stage/prod.


douglashall:
@mjfrey @clintonb

PR #625, opened 2016-03-17
Fix issue when try to add new course using a course id which is alrea…

attiyaIshaque:
### Description

In the `CAT server`, When try to add a new course using a course id which is already associated with a course, It will give an  error `This course id is associated with another course seat` and do no let user save the existing course seat.

[ECOM-3859](https://openedx.atlassian.net/browse/ECOM-3859)

### DIFF
--- a/ecommerce/static/js/test/specs/views/form_view_spec.js
+++ b/ecommerce/static/js/test/specs/views/form_view_spec.js
@@ -11,7 +11,9 @@ define([
         describe('form view', function () {
             var view,
                 TestModel,
-                model;
+                model,
+                errorHTML,
+                courseId;
 
             beforeEach(function () {
                 TestModel = Backbone.Model.extend({ url: 'fake/url' });
@@ -19,10 +21,11 @@ define([
                 view = new FormView({ model: model });
                 view.$el.prepend('<div class="alerts" tabindex="-1" aria-live="polite"></div>');
                 view.render();
+                courseId = 'course-v1:edX+DemoX+Demo_Course';
             });
 
             it('should throw an error if submitted with blank fields', function () {
-                var errorHTML = '<strong>Error!</strong> You must complete all required fields.';
+                errorHTML = '<strong>Error!</strong> You must complete all required fields.';
                 spyOn(model, 'isValid').and.returnValue(false);
                 view.submit($.Event('click'));
                 expect(view.$el.find('.alert').length).toBe(1);
@@ -41,6 +44,68 @@ define([
                 expect(Backbone.history.navigate).toHaveBeenCalled();
             });
 
+            it('should validate if course id is duplicate', function () {
+                 var courseDataSuccess =  {
+                        id: courseId,
+                        name: 'Demo Course',
+                        type: 'audit'
+                 },
+                     redirected_url = '/courses/' + courseId,
+                     html = '<a href="'+redirected_url+'"> Click here to view the existing course</a>';
+                errorHTML = '<strong>Error!</strong> A course with the specified ID already exists.' + html;
+
+                 //check the error msg is empty
+                 expect(view.$el.find('.alert').length).toBe(0);
+
+                 spyOn($, 'ajax').and.callFake( function (callback) {
+                     //Check the success method of ajax
+                     callback.success(courseDataSuccess);
+                    });
+                 view.checkCourseAlreadyExist(courseId);
+
+                 expect(view.$el.find('.alert').length).toBe(1);
+                 expect(view.$el.find('.alert').html()).toBe(errorHTML);
+             });
+
+            it('should not validate if course id is not duplicate', function () {
+                 var errorResponse = {
+                     status : 404
+                 };
+
+                 //check the error msg is empty
+                 expect(view.$el.find('.alert').length).toBe(0);
+
+                 //Check the failure method of ajax
+                 spyOn($, 'ajax').and.callFake( function (callback) {
+                     callback.error(errorResponse);
+                    });
+                 view.checkCourseAlreadyExist(courseId);
+                 expect(view.$el.find('.alert').length).toBe(0);
+             });
+
+            it('should check ajax method has been called', function () {
+                 var args;
+                 spyOn($, 'ajax');
+                 view.checkCourseAlreadyExist(courseId);
+
+                 // $.ajax should have been called
+                 expect($.ajax).toHaveBeenCalled();
+                 args = $.ajax.calls.argsFor(0)[0];
+                 expect(args.method).toEqual('get');
+                 expect(args.url).toEqual('/api/v2/courses/'+courseId);
+                 expect(args.contentType).toEqual('application/json');
+             });
+
+            it('should check validateCourseID method has been called' , function(){
+                view.$el.prepend('<input type="text" class="form-control" name="id" ' +
+                     'value = "course-v1:edX+DemoX+Demo_Course" >');
+                var courseIdInput = $(view.el).find('input[name=id]');
+                spyOn(view, 'checkCourseAlreadyExist');
+                view.validateCourseID();
+                courseIdInput.trigger('focusout');
+                expect(view.checkCourseAlreadyExist).toHaveBeenCalled();
+            });
+
             it('should call saveSuccess when model is successfully saved', function () {
                 spyOn(view, 'saveSuccess');
                 spyOn(model, 'isValid').and.returnValue(true);
@@ -52,8 +117,8 @@ define([
             });
 
             it('should throw an error if saving the data fails', function () {
-                var errorHTML = '<strong>Error!</strong> An error occurred while saving the data.',
-                    errorObj;
+                var errorObj;
+                errorHTML = '<strong>Error!</strong> An error occurred while saving the data.';
 
                 function testErrorResponse() {
                     view.submit($.Event('click'));
--- a/ecommerce/static/js/views/form_view.js
+++ b/ecommerce/static/js/views/form_view.js
@@ -67,6 +67,44 @@ define([
                 return this;
             },
 
+            validateCourseID : function(){
+                var self = this;
+                var courseIdInput = self.$('input[name=id]');
+                courseIdInput.focusout(function() {
+                    var courseId = courseIdInput.val();
+                    self.checkCourseAlreadyExist(courseId);
+                });
+            },
+            /**
+             * Validate the courseId input if it already exists
+             */
+            checkCourseAlreadyExist : function(courseId){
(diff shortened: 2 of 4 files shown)


ecommerce/static/js/models/course_model.js
```diff
@@ -45,7 +45,34 @@ define([
             validation: {
                 id: {
                     required: true,
-                    pattern: 'courseId'
+                    pattern: 'courseId',
+                    // If the course already exists,the Showed a message with a link to the existing course.
+                    fn: function (courseId) {
```
clintonb: This approach is incorrect. If you check for the existence of the course on the course model's validation, you will never be able to update a course.

You only need to check if a course exists on the course creation page. When a user inputs a course ID, do the following:
1. Poll the API to determine if the course exists.
2. If the course does not exist, do nothing.
3. If the course exists, [display an alert](https://github.com/edx/ecommerce/blob/ac642df791f6eea8c262960f21eb22de6ced0c46/ecommerce/static/js/views/form_view.js#L55).

attiyaIshaque:
@clintonb  and now I am using {{ajax}} to check if course already exists. Now, [JS test](https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/test/specs/models/course_model_spec.js#L262) is failing here because of the change. 

In the test, when the model is not validated.
`expect(model.validate()).toBeFalsy();`
Why we are expacting the an ajax call to the `publication` api?

Second we used stickit for validation and it validates everytime user inputs value. If user clicks anywhere on the page it again validates the model. This specifically happens when user clicks on the `link` to go to an existing course which calls the validation and anchor link is not followed.

ecommerce/static/js/models/course_model.js
attiyaIshaque: @clintonb Okay! I will trigger an event on focus out the element and make the API call there.
Is this correct?

attiyaIshaque: @clintonb I triggered `focusout` event  for checking the validation if the course exist, its dispplay an error.
But  on submit event [here](https://github.com/edx/ecommerce/blob/ac642df791f6eea8c262960f21eb22de6ced0c46/ecommerce/static/js/views/form_view.js#L118) its validate the input and pass it.

attiyaIshaque: @clintonb ??

clintonb: I don't understand your question. Why are you worried about the submit event?

awaisdar001: @clintonb  Ok, so it seems, what she is trying to say is, a `focusout` event trigers on the Course ID input field. Which displays an error for an existing course ID. She thinks displaying an error message is not sufficient but also to restrict user from submitting the from. 

Like, If one enters an existing `course_id`, an error message is displayed, but submitting the course create form with the `course_id` does not restrict the user. She is asking what approach/way should be used that also restrict user from submitting the course create form in CAT.

@attiyaIshaque Please feel free to add what else you think.

clintonb: Go for it. I agree, the user should not be able to re-create an existing course.

attiyaIshaque: @awaisdar001 Thanks for explaining.
@clintonb Okay.

attiyaIshaque:
@aamir-khan @mushtaqak Please review.

ecommerce/static/js/test/specs/views/form_view_spec.js
```diff
@@ -52,8 +114,8 @@ define([
             });
 
             it('should throw an error if saving the data fails', function () {
-                var errorHTML = '<strong>Error!</strong> An error occurred while saving the data.',
-                    errorObj;
+                errorHTML = '<strong>Error!</strong> An error occurred while saving the data.';
+                var errorObj;
```
aamir-khan, line 120: Define the var errorObj before the errorHTML. The variable definition should be in the start of function.

ecommerce/static/js/views/form_view.js
```diff
@@ -120,6 +158,10 @@ define([
                     this.renderAlert('danger', gettext('You must complete all required fields.'));
                     return;
                 }
+                // Check the courseID is already exist.
```
mushtaqak: nit : You may modify comment like this
// Check if the courseID already exists.

ecommerce/static/js/views/form_view.js
```diff
@@ -67,6 +67,43 @@ define([
...
+                $.ajax({
+                    url: url,
+                    method: 'get',
+                    contentType: 'application/json',
+                    async: false,
+                    success: function (data) {
+                        if(data.id === courseId) {
+                            _this.clearAlerts();
+                            _this.renderAlert('danger',
+                                gettext('This course id is associated with another Course seat'));
```
mushtaqak: Can this line be adjusted above ?

ecommerce/static/js/views/form_view.js
```diff
@@ -67,6 +67,43 @@ define([
...
+                    self.checkCourseAlreadyExist(courseId);
+                });
+            },
+            /**
+             * Validate the courseId input if it is already exist
+             */
+            checkCourseAlreadyExist : function(courseId){
+                var courseIDFound = false;
+                var _this = this;
+                var url = '/api/v2/courses/' + courseId;
```
mushtaqak: Just like you have done here for courseId

ecommerce/static/js/views/form_view.js
```diff
@@ -67,6 +67,43 @@ define([
...
+            validateCourseID : function(){
+                var self = this;
+                var courseIdInput = self.$('input[name=id]');
+                courseIdInput.focusout(function() {
+                    var courseId = courseIdInput.val();
+                    self.checkCourseAlreadyExist(courseId);
+                });
+            },
+            /**
+             * Validate the courseId input if it is already exist
```
mushtaqak: nit : Validate the courseId input if it already exists

ecommerce/static/js/test/specs/views/form_view_spec.js
```diff
@@ -41,6 +44,65 @@ define([
...
+            it('should check ajax method has been called', function () {
+                 var args;
+                 spyOn($, 'ajax');
+                 view.checkCourseAlreadyExist(courseId);
+
+                 // $.ajax should have been called
+                 expect($.ajax).toHaveBeenCalled();
+                 args = $.ajax.calls.argsFor(0)[0];
+                 expect(args.method).toEqual('get');
+                 expect(args.url).toEqual('/api/v2/courses/course-v1:edX+DemoX+Demo_Course');
```
mushtaqak: We can add courseId here too

ecommerce/static/js/test/specs/views/form_view_spec.js
```diff
@@ -41,6 +44,65 @@ define([
...
+                        name: 'Demo Course',
+                        type: 'audit'
+                 };
+                 errorHTML = '<strong>Error!</strong> This course id is associated with another Course seat';
+
+                 //check the error msg is empty
+                 expect(view.$el.find('.alert').length).toBe(0);
+
+                 spyOn($, 'ajax').and.callFake( function (callback) {
+                     callback.success(courseDataSuccess); //Check the success method of ajax
```
mushtaqak: move comment to above line

ecommerce/static/js/test/specs/views/form_view_spec.js
```diff
@@ -41,6 +44,65 @@ define([
                 expect(Backbone.history.navigate).toHaveBeenCalled();
             });
 
+            it('should validate if course id is duplicate', function () {
+                 var courseDataSuccess =  {
+                        id: 'course-v1:edX+DemoX+Demo_Course',
```
mushtaqak: can we set id to `courseId` that we had previously set in `beforeEach`

attiyaIshaque:
@aamir-khan and @mushtaqak  I have addressed all comments.

aamir-khan:
:+1:

mushtaqak:
LGTM :+1:

attiyaIshaque:
@clintonb Please review.

clintonb:
Please share a sandbox link.

attiyaIshaque:
@clintonb okay.

attiyaIshaque:
@clintonb I have made a sandbox through master and then checkout my branch. when I update the static files, having error 

```
python manage.py collectstatic --noinput -v0
 File "manage.py", line 8, in <module>
    from django.core.management import execute_from_command_line
ImportError: No module named django.core.management
make: *** [static] Error 1
```

 What would i do?

clintonb:
Use virtulenv. 
source /edx/app/ecommerce/ecommerce_env

attiyaIshaque:
@clintonb I have already tried this. but getting the same error. I can't update the static files.

clintonb:
I'm going to challenge you to figure this one out on your own so that you can learn more about production-style deployments, and potentially help others. Here are the items I check when debugging sandbox issues:
1. virtualenv (including installed packages)
2. settings
3. git branch

I'll further challenge you to take a moment to understand how nginx interacts with supervisor, which interacts with gunicorn, which interacts with Django. The more you understand this interaction, the better able you'll be able to resolve these issues on your own.

attiyaIshaque:
@clintonb thanks.

attiyaIshaque:
@clintonb 
Sandbox here: https://ecom.sandbox.edx.org/

ecommerce/static/js/test/specs/views/form_view_spec.js
```diff
@@ -41,6 +44,66 @@ define([
                 expect(Backbone.history.navigate).toHaveBeenCalled();
             });
 
+            it('should validate if course id is duplicate', function () {
+                 var courseDataSuccess =  {
+                        id: courseId,
+                        name: 'Demo Course',
+                        type: 'audit'
+                 };
+                 errorHTML = '<strong>Error!</strong> This course id is associated with another Course seat';
```
clintonb: A course with the specified ID already exists.

clintonb:
👍  (after rebasing and updating the error message)

ecommerce/static/js/test/specs/views/form_view_spec.js
clintonb: It would also be nice if you added, "Click here to view the existing course." and linked to the course details page.

attiyaIshaque:
@clintonb I have addressed the comments.

attiyaIshaque:
Merged. 
FYI : @clintonb @mushtaqak @aamir-khan

PR #627, opened 2016-03-18
[SOL-1664] Add Notes text field while creating codes

vkaracic:
This PR adds a `note` field to the coupon model.

### DIFF
--- a/ecommerce/extensions/catalogue/migrations/0016_coupon_note_attribute.py
+++ b/ecommerce/extensions/catalogue/migrations/0016_coupon_note_attribute.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from oscar.core.loading import get_model
+
+ProductAttribute = get_model("catalogue", "ProductAttribute")
+ProductClass = get_model("catalogue", "ProductClass")
+
+
+def create_note_attribute(apps, schema_editor):
+    """Create coupon note attribute."""
+    coupon = ProductClass.objects.get(name='Coupon')
+    ProductAttribute.objects.create(
+        product_class=coupon,
+        name='Note',
+        code='note',
+        type='text',
+        required=False
+    )
+
+
+def remove_note_attribute(apps, schema_editor):
+    """Remove coupon note attribute."""
+    coupon = ProductClass.objects.get(name='Coupon')
+    ProductAttribute.objects.get(product_class=coupon, name='Note').delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0001_initial'),
+        ('catalogue', '0015_default_categories')
+    ]
+    operations = [
+        migrations.RunPython(create_note_attribute, remove_note_attribute)
+    ]
--- a/ecommerce/static/templates/coupon_form.html
+++ b/ecommerce/static/templates/coupon_form.html
@@ -97,6 +97,11 @@
             <select class="form-control" name="seat_type"></select>
             <p class="help-block"></p>
         </div>
+        <div class="form-group">
+            <label for="note"><%= gettext('Note') %></label>
+            <input type="text" class="form-control" name="note" maxlength="100">
+            <p class="help-block"></p>
+        </div>
     </div>
 </div>
 
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -399,6 +399,7 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     client = serializers.SerializerMethodField()
     vouchers = serializers.SerializerMethodField()
     categories = CategorySerializer(many=True, read_only=True)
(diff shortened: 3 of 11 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=627&repo=edx%2Fecommerce

ecommerce/extensions/catalogue/migrations/0016_coupon_note_attribute.py
```diff
@@ -0,0 +1,36 @@
...
+        name='Note',
+        code='note',
+        type='text',
+        required=False
+    )
+
+
+def remove_note_attribute(apps, schema_editor):
+    """Remove coupon note attribute."""
+    ProductAttribute.objects.get(name='Note').delete()
```
clintonb: Include the product class.

clintonb:
1. I see you tried TDD, or at least writing tests first. What do you think?
2. Make sure we can run these tests without migrations. Update the coupon test mixin to create this new attribute. 
3. :+1:

vkaracic:
I'm still getting used to TDD, it's good to for initial structuring of the changes but, as you can see, I fail to cover everything at first. But I'll keep at it. 

I've added to the mixin: https://github.com/edx/ecommerce/pull/627/files#diff-9cb070bbae2784c76985cdd9fff7f702
Locally it passes for me when I disable migrations.

marjev:
You'll have to disable the input element for notes when editing form is open since we currently don't support editing of the notes. Also, styling seems to be a little off when the form is in editing mode. Perhaps disabling the input element will solve the issue. If not, please fix that as well.

Otherwise:+1:

clintonb:
Please post screenshots of the creation, editing, and detail views.

marjev:
Create new coupon:

## screen shot 2016-03-21 at 14 27 55

Coupon details:

## screen shot 2016-03-21 at 14 29 23

Edit coupon:
screen shot 2016-03-21 at 14 29 34

ecommerce/static/templates/coupon_form.html
```diff
@@ -97,6 +97,11 @@
             <select class="form-control" name="seat_type"></select>
             <p class="help-block"></p>
         </div>
+        <div class="form-group">
+            <label for="note"><%= gettext('Note') %></label>
+            <input type="text" class="form-control" name="note">
```
clintonb: Why not `<textarea>`?

vkaracic: The ticket mentioned a max length of 30:
https://openedx.atlassian.net/browse/SOL-1664
Maybe set `maxlength` in the element?

clintonb: Okay, why a max length of 30? I'm not seeing any discussion on the ticket? 30 characters seems rather arbitrary.

vkaracic: The notes are a way of further describing the coupon, it's a better replacement for sub-categories that we had going some time ago. I believe two-three words will ever be used for that, but @pbaruah suggest 150 character length and I'll change it to that.

clintonb: Before we go and make another change, let's pause. What will this field _actually_ be used for? Let's get some concrete use cases, please. The sub-categories idea was not well-thought because there were no use cases. Let's avoid a repeat of that, please.

pbaruah: @vkaracic @clintonb - 

Here is a background on the original sub-category thought/use cases -
1. If an edX user wants to create a coupon for a one-off campaign and there is no matching category, the user can select "other" as major category and create a sub-category to provide some clarification why the coupon was created. 
2. For some campaigns where there is a category (e.g. Geography Promo), the coupon creator can create a sub-category called "Latin America" which further describes the major category.
3. For bulk enrollment for corporate clients, one can use "Bulk Enrollment" as the major category and create a "Boeing" as a sub-category. When reporting, we can search all the coupons created under bulk enrollment and see which clients it was created for. We can also narrow it down further and bring up only "Boeing" coupons created for "Bulk Enrollment"

The idea was to create these sub-categories like Jira labels so that if it does not exist, it creates a new sub-category but if it exists, one can select it. In order to reduce implementation time, it was later decided to use a text field instead of sub-categories and it was decided to change the name to "Notes". 

The 30 character limit was a left over from the sub-category idea. Since it is a text field now, we can increase the length to 100 characters (not 150 upon further thought as we just want a precise record of what it was created for and not want long essays). 

Hope this clears your doubts and the PR can be accepted and merged.
I

vkaracic:
I'm a bit unclear about whether or not I can proceed on merging this?
@mjfrey @pbaruah @clintonb

pbaruah:
@vkaracic - I saw a thumbs up from @clintonb on Monday. Are we still waiting for any clarifications from anyone? Please let me know.

vkaracic:
I'll consider the thumbs up on this comment https://github.com/edx/ecommerce/pull/627#discussion_r56839865 to resolve the notes issues and go ahead with the merge.

PR #628, opened 2016-03-18
[SOL-1713] Solve Integrity Error on Create Coupon request

marjev:
@mjfrey please review this.

I guess the reason for IntegrityError is explained in Django documentation.
https://docs.djangoproject.com/en/1.9/ref/models/querysets/#django.db.models.query.QuerySet.get_or_create

I don't see another reason for it. There are many unofficial ways to solve this problem like creating our own get_or_create method, but that's not global solution like this is. Let me know what you think.

### DIFF
--- a/ecommerce/settings/production.py
+++ b/ecommerce/settings/production.py
@@ -1,5 +1,6 @@
 """Production settings and globals."""
 from os import environ
+import psycopg2.extensions
 
 # Normally you should not import ANYTHING from Django directly
 # into your settings, but ImproperlyConfigured is an exception.
@@ -58,3 +59,5 @@ DB_OVERRIDES = dict(
 
 for override, value in DB_OVERRIDES.iteritems():
     DATABASES['default'][override] = value
+
+DATABASES['OPTIONS']['isolation_level'] = psycopg2.extensions.ISOLATION_LEVEL_READ_COMMITTED
--- a/ecommerce/settings/production.py
+++ b/ecommerce/settings/production.py
@@ -1,6 +1,5 @@
 """Production settings and globals."""
 from os import environ
-import psycopg2.extensions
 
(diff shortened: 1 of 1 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=628&repo=edx%2Fecommerce

ecommerce/settings/production.py
```diff
@@ -1,5 +1,6 @@
 """Production settings and globals."""
 from os import environ
+import psycopg2.extensions
```
clintonb: We use MySQL.

marjev: I fixed this. Could you please take a look at this PR?
This [JIRA ticket](https://openedx.atlassian.net/browse/SOL-1713) has a traceback that we got from Splunk. My first comment contains a probable cause of the issue (see the link to Django documentation).

Could you propose a way of testing this so that we can verify that this would fix the issue mentioned in the bug report ticket?

clintonb: The error described in the ticket has nothing to do with database settings, and everything to do with duplicate usernames. This is a serious concern @mjfrey and @pbaruah. We are treating clients as users. If you are making new users on ecommerce, and they aren't being synced back to LMS, there will be a major issue if a real user comes along with the same username.

mjfrey: @clintonb That was always the design and talked about in detail.  Creating a user to represent the client, as that would also be who owned the basket.  Do you have a suggestion on how to modify this?

clintonb: The user needs to be created on LMS first.

mjfrey:
Going to re-work the way we record coupon orders / invoices.  Closing this PR.

PR #629, opened 2016-03-18
WL-286 Added multitenancy support using Django sites framework

douglashall:
Reintroducing this work after having to revert previously due to [this](https://groups.google.com/forum/#!topic/edx-code/5q6riHK2_P8) issue described on edx-code.

Original PR: https://github.com/edx/ecommerce/pull/543

Diff with original PR: https://github.com/edx/ecommerce/compare/48531e2d8b2a3201e12d7b95887d4eeab4587844...douglashall/WL-286/multitenancy_support

PR #630, opened 2016-03-21
Multitenant payment processor config

e-kolpakov:
**Description:** Adds multitenancy support for payment processor configuration, using Django sites framework.
**Dependencies:** #629 
**Merge deadline:** 1 Apr
**Related:** [Discovery document](https://docs.google.com/document/d/17BRYvkO4ETvBV9TFodElpbJaK4oze7e0mNqf9ztnGnA/edit#heading=h.70eeasm5mr1q)
**Sandbox:** 
- edX site: [LMS](http://pr11898.sandbox.opencraft.com), [Studio](http://studio.pr11898.sandbox.opencraft.com), [eCommerce](https://ecommerce.pr11898.sandbox.opencraft.com)
- WL site: [LMS](http://wlsite.pr11898.sandbox.opencraft.com), [eCommerce](https://ecommerce.wlsite.pr11898.sandbox.opencraft.com)

Note both HTTP and HTTPS are available for LMS and Otto.

**JIRA:** [WL-362](https://openedx.atlassian.net/browse/WL-362)

**Preparing environment**
1. Switch to current branch
2. Modify settings - move current contents of PAYMENT_PROCESSOR_CONFIG into a new nested dict with  `default` key, i.e. 
    `PAYMENT_PROCESSOR_CONFIG = { 'default': OLD_PAYMENT_PROCESSOR_CONFIG_VALUE }`
3. Migrate DB (adds "OAuth settings" field to `SiteConfiguration`) model.

**LMS and Otto configuration**

The idea is to set up two Sites in Otto and configure Otto and LMS so that both sites are operational.
We will be using two "sites": "edx" and "wl"; so [sitename] is one of (`edx`,  `wl`)

**Configuring LMS:**

The goal here is to configure two (or more) OAuth clients.
1. Go to http://localhost:8000/admin/oauth2/client/ -> Add Client
2. Set values: 
   1. _Url:_ http://[sitename].lms.local:8002
   2. _Redirect Uri_: http://[sitename].lms.local:8002/complete/edx-oidc/
   3. _Client Id:_ ecommerce-[sitename]-key
   4. _Client Secret:_ ecommerce-[sitename]-secret
   5. _Client type:_ Confidential
3. Go to http://localhost:8000/admin/oauth2_provider/trustedclient/ and designate both clients added in (2) as trusted (add one record per client)
4. In http://localhost:8000/admin/commerce/commerceconfiguration/ add/edit the configuration and set check "Checkout on ecommerce service".
5. Make sure FEATURES['ENABLE_OAUTH2_PROVIDER'] is set to True in lms.env.json (required to generate credit course records in course catalog - can be skipped if no credit-mode courses are used)

**Configuring Otto**

Otto configuration: Configure two Sites and SiteConfiguration records so that Otto can connect to LMS
1. Add two site records at http://localhost:8002/admin/sites/site/ (i.e. edx.lms.local:8002 and wl.lms.local:8002)
2. Add two partner records (http://localhost:8002/admin/partner/partner/) - choose unique name and short_code, don't select users.
3. Add corresponding SiteConfiguration records (http://mitpe.lms.analytics.local:8002/admin/core/siteconfiguration/):
   1. _Site:_ choose one of sites selected earlier
   2. _Parner:_ choose unique partner (or create new)
   3. _Path to custom site theme:_ anything, but empty string (required, but works fine if file does not exist)
   4. *Payment processors:' comma separated list of supported processors (supported so far: paypal, cybersource), i.e. `paypal,cybersource`
   5. OAuth settings:

```
    {
      "SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY":"ecommerce-[sitename]-secret",
      "SOCIAL_AUTH_EDX_OIDC_KEY":"ecommerce-[sitename]-key",
      "SOCIAL_AUTH_EDX_OIDC_SECRET":"ecommerce-[sitename]-secret",
      "SOCIAL_AUTH_EDX_OIDC_URL_ROOT": "http://[sitename].lms.local:8000/oauth2",
      "OAUTH_PROVIDER_URL":"http://[sitename].lms.local:8000/"
    }
```

**Set up payment processor configuration in Otto**

Add payment processor configuration for each site - add new top level records to PAYMENT_PROCESSOR_CONFIG setting:

```
PAYMENT_PROCESSOR_CONFIG = {
    'default': { ... },
    '[sitename]': { 
        'cybersource': { $cybersource_config$ },
        'paypal': { $paypal_config$ },
}
```

$cybersource_config$ simplest way to do everything right is to copy PAYMENT_PROCESSOR_CONFIG['default'] value and than replace relevant settings. For cybersource 'merchant_id', 'transaction_key':  'profile_id', 'access_key',  'secret_key' should be replaced.

**Configure WL microsite in LMS**

TBD - this part is reworked by @mattdrayer (microsite -> comprehensive theming). Meanwhile, just continue using microsite framework - add the following to lms.env.json:

```
"MICROSITE_CONFIGURATION": {
        "wl": {
                        "cybersource_config_key": "wl",
                        "domain_prefix": "wl",
                        "university": "wl",
                        "platform_name": "White Label Site",
                        "css_overrides_file": "css/main.css",
                        "SITE_NAME": "wl.exampl.ecom",
                        "course_org_filter": "wlX",
                        "show_partners": false,
                        "ECOMMERCE_PUBLIC_URL_ROOT":  "http://[sitename].lms.local:8002/"
                }
    },
```

LMS does not support checkout with Otto for themed sites/microsites in master branch, so you'd need to switch to [alternative branch](https://github.com/open-craft/edx-platform/tree/ekolpakov/theme-ecommerce-support) until it is merged to master.

**Generate course catalog entries**
1. Create courses in Studio (documented elsewhere)
2. Verify both main site and WL site have some courses - go to LMS (at edx.lms.local and wl.lms.local) and make sure there are some courses displayed in "Find Courses".
3. Go to http://localhost:8002/courses/ and add new course records to course catalog (documented elsewhere)

**Update host machine hosts file**
Add edx.lms.local and wl.lms.local to `/etc/hosts` (or equivalent) as aliases of 127.0.0.1

**Testing instructions:**
For both sites (wl and edx):
1. Log out from both Otto (localhost:8002) and LMS (localhost:8000 / edx.lms.local / wl.lms.local)
2. Log in to LMS as honor (or create new user)
3. Go to course catalogue
4. Enroll in course (if Verified track is avalilable) and upgrade to "Verified" track or add course to a cart (if "professional" track is available).
5. You'll arrive at checkout page. This page contains one "Checkout" button per available payment processor (Cybersource/Paypal). _Expected result:_ buttons are shown only for processors enabled in SiteConfiguration model for that site.
6. Click "Checkout" - you'll be taken to the payment processor site for payment. _Expected result:_ correct PAYMENT_PROCESSOR_CONFIG entry is used. The simplest would be to design visually different payment pages for different configs (i.e. include different subset of optional fields).
7. Complete payment
8. You should be redirected back to appropriate microsite/main site (all payment processors), and have the course upgraded to Verified track (only Paypal - Cybersource can not connect to VM boxes, unless they are exposed to the world - sandbox to test that is currently provisioning)


openedx-webhooks:
Thanks for the pull request, @e-kolpakov! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=630&repo=edx%2Fecommerce

ecommerce/core/models.py
```diff
@@ -34,10 +34,25 @@ class SiteConfiguration(models.Model):
...
+        blank=False,
+        default='{}'
+    )
 
     class Meta(object):
         unique_together = ('site', 'partner')
 
+    @property
+    def allowed_payment_processors(self):
+        payment_processors = (
```
jbzdak: Possibly (for backward compatibility) it would be good to have empty `payment_processors` mean 'every processor', as this field my be unfilled on various production instances.

ecommerce/core/models.py
```diff
@@ -34,10 +34,25 @@ class SiteConfiguration(models.Model):
...
+            raw_processor_value.strip()
+            for raw_processor_value in self.payment_processors.split(',')
```
jbzdak, line 67: If we are considering maintainability of this code, it would be good to move this to `EdxOrderPlacementMixin`, as this is the type that requires `self.payment_processor` to be a defined instance, and define payment processor type as a class constant.

jbzdak:
@e-kolpakov It works as advertised, ping me when you finish --- add tests and so on --- and I'll re-check :)

ecommerce/extensions/payment/config.py
```diff
@@ -7,7 +7,7 @@ class PaymentConfig(config.PaymentConfig):
     name = 'ecommerce.extensions.payment'
 
     def ready(self):
-        paypal_configuration = settings.PAYMENT_PROCESSOR_CONFIG.get('paypal')
+        paypal_configuration = settings.PAYMENT_PROCESSOR_CONFIG['default'].get('paypal')
```
jbzdak: Please double-check this actually makes sense. `paypalrestsdk` could have different config than currently used microsite, which could lead to trouble.

ecommerce/settings/devstack.py
```diff
@@ -23,25 +21,27 @@
...
-    },
-    'paypal': {
-        'mode': 'sandbox',
-        'client_id': 'fake-client-id',
-        'client_secret': 'fake-client-secret',
-        'receipt_url': get_lms_url(settings.RECEIPT_PAGE_PATH),
-        'cancel_url': get_lms_url('/commerce/checkout/cancel/'),
-        'error_url': get_lms_url('/commerce/checkout/error/'),
-    },
+    'default': {
```
jbzdak: Default is a misleading name here. I read default as: "If I don't override settings for a my microsite, it will use default settings". Moreover `'default'` key is used in tests and settings, only occourence in code [here](https://github.com/edx/ecommerce/pull/630/files#r56902761), so possibly it should just go away.

jbzdak:
@e-kolpakov Thank you for very detailed setup instructions :)

ecommerce/core/models.py
e-kolpakov: @jbzdak maybe, but if there are production instance(s) with empty `payment_processor` fields, it would be _very_ easy to spot (no "checkout" buttons) and easy to fix. Also, it is a direct violation of setup instructions, so I don't think there are any.

So, in my opinion, we shouldn't add this little workaround/hack to support backward compatibility. Let's see if upstream reviewer (@douglashall ?) have something to say about it.

ecommerce/core/models.py
e-kolpakov, line 67: @jbzdak are you sure this is the right place for this comment? If it is, than no, this setting should stay here, as it just transforms comma-separated `payment_processors` field into a set of allowed payment processor names.

Also, only two children of `EdxOrderPlacementMixin` uses `payment_processor` property, so it does not look like `payment_processor` belongs to it.

ecommerce/extensions/payment/views.py
```diff
@@ -35,7 +35,9 @@
 
 class CybersourceNotifyView(EdxOrderPlacementMixin, View):
     """ Validates a response from CyberSource and processes the associated basket/order appropriately. """
-    payment_processor = Cybersource()
+    @property
+    def payment_processor(self):
+        return Cybersource()
```
jbzdak: If we are considering maintainability of this code, it would be good to move this to EdxOrderPlacementMixin, as this is the type that requires self.payment_processor to be a defined instance, and define payment processor type as a class constant.

ecommerce/core/models.py
jbzdak, line 67: OK you are right --- i pasted it in the wrong place (I was moving comments from #632 to here), and you are right on that that there are more subclasses `EdxOrderPlacementMixin` than I was aware of.

ecommerce/core/tests/test_models.py
```diff
@@ -38,3 +40,21 @@ def test_get_full_name(self):
 
         user = self.create_user(full_name=full_name, first_name=first_name, last_name=last_name)
         self.assertEquals(user.get_full_name(), full_name)
+
+
+@ddt.ddt
+class SiteConfigurationTests(TestCase):
+    @ddt.data(
+        ("", set()),
+        (",,", set()),
```
jbzdak: I like this case :)

ecommerce/core/tests/test_models.py
```diff
@@ -38,3 +40,21 @@ def test_get_full_name(self):
...
+        ("paypal", {"paypal"}),
+        ("paypal,", {"paypal"}),
+        ("paypal,cybersource", {"paypal", "cybersource"}),
+        ("paypal,,cybersource", {"paypal", "cybersource"}),
+        (",paypal,,cybersource", {"paypal", "cybersource"}),
```
jbzdak: Maybe add something like `, paypal   , ,  cybersource`, to check for regressions when spaces arn't properly stripped.

ecommerce/extensions/payment/helpers.py
```diff
@@ -87,3 +89,18 @@ def sign(message, secret):
...
+
+
+def get_payment_processors():
+    request = get_current_request()
+    if request:
+        all_processors = [get_processor_class(path) for path in settings.PAYMENT_PROCESSORS]
+        available_processors = [
+            processor for processor in all_processors
+            if (
+                processor.NAME in request.site.siteconfiguration.allowed_payment_processors and
```
jbzdak: NIT: Create a local variable for `request.site.siteconfiguration.allowed_payment_processors`, and then whole if would fit in a single line :)

ecommerce/extensions/payment/tests/test_helpers.py
```diff
@@ -40,3 +44,58 @@ def test_sign(self):
...
         expected = "qU4fRskS/R9yZx/yPq62sFGOUzX0GSUtmeI6bPVsqao="
         self.assertEqual(helpers.sign(message, secret), expected)
+
+    @ddt.data(
+        ([], []),
+        ([DummyProcessor], [DummyProcessor]),
+        ([DummyProcessor, AnotherDummyProcessor], [DummyProcessor, AnotherDummyProcessor]),
+    )
+    @ddt.unpack
+    def test_get_payment_processors(self, processors, expected_result):
```
jbzdak: It looks like in every case `processors==expected_result` so maybe either change them to vary or merge them to a single variable.

e-kolpakov: @jbzdak that's just a coincidence - expected result have the same format and value as test input, so I would rather keep it separated (for flexibility) and in line with other tests.

jbzdak: OK. Agreed.

jbzdak:
Solid code +1, I have verified using provided sandbox. Before the merge we need to verify that there is proper redirection (wlsite.lms redirects to wlsite.otto and plain lms redirects to plain otto).

bradenmacdonald:
@e-kolpakov Can you please rebase this too, now that Doug's PR has merged?

e-kolpakov:
@clintonb @douglashall please review.

ecommerce/core/models.py
douglashall: I do not think that an empty `payment_processors` field should convert to all payment processors.

ecommerce/core/models.py
```diff
@@ -45,6 +45,15 @@ class SiteConfiguration(models.Model):
     class Meta(object):
         unique_together = ('site', 'partner')
 
+    @property
+    def allowed_payment_processors(self):
```
douglashall: Can we rename this to `payment_processors_set`?

ecommerce/core/models.py
```diff
@@ -45,6 +45,15 @@ class SiteConfiguration(models.Model):
...
+        payment_processors = (
+            raw_processor_value.strip()
+            for raw_processor_value in self.payment_processors.split(',')
+            if raw_processor_value and not raw_processor_value.isspace()
+        )
+        return set(payment_processors)
```
douglashall: You could use a set comprehension here instead.

```
return {
    raw_processor_value.strip()
    for raw_processor_value in self.payment_processors.split(',')
    if raw_processor_value and not raw_processor_value.isspace()
}
```

ecommerce/extensions/payment/helpers.py
```diff
@@ -87,3 +89,16 @@ def sign(message, secret):
...
+def get_payment_processors():
+    request = get_current_request()
+    if request:
+        all_processors = [get_processor_class(path) for path in settings.PAYMENT_PROCESSORS]
+        site_config = request.site.siteconfiguration
+        available_processors = [
+            processor for processor in all_processors
+            if processor.NAME in site_config.allowed_payment_processors and processor.is_enabled()
+        ]
+        return available_processors
```
douglashall: Prefer:

```
request = get_current_request()
if request:
    allowed_payment_processors = request.site.siteconfiguration.allowed_payment_processors
    return [
        processor for processor in all_processors
        if processor.is_enabled() and processor.NAME in allowed_payment_processors
    ]
```

ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -26,7 +28,7 @@ class Command(BaseCommand):
 
     The 'enable' and 'disable' actions are idempotent so it is safe to run them repeatedly in the same environment.
     """
-    args = "action [id] [json]"
+    args = "action config [id] [json]"
```
douglashall: Rename `config` to `partner`.

douglashall:
:+1: once comments addressed. Thanks!

douglashall:
@e-kolpakov We will also need a corresponding PR in the configuration [repo](https://github.com/edx/configuration/blob/master/playbooks/roles/ecommerce/defaults/main.yml#L82) to update the payment processor configuration.

e-kolpakov:
@douglashall Config repo PR: https://github.com/edx/configuration/pull/2886

e-kolpakov:
@clintonb / @mjfrey do you want to review this, or is @douglashall approval enough for the merge?

ecommerce/core/models.py
```diff
@@ -45,6 +45,14 @@ class SiteConfiguration(models.Model):
     class Meta(object):
         unique_together = ('site', 'partner')
 
+    @property
+    def payment_processors_set(self):
+        return {
+            raw_processor_value.strip()
+            for raw_processor_value in self.payment_processors.split(',')
+            if raw_processor_value and not raw_processor_value.isspace()
```
clintonb: Validate the data before it's stored, and you don't have to worry about cases like these.

ecommerce/core/tests/test_models.py
clintonb: Validate the data before it's stored, and you don't have to worry about cases like these.

ecommerce/extensions/payment/helpers.py
```diff
@@ -87,3 +89,15 @@ def sign(message, secret):
     signature = base64.b64encode(digest).decode()
 
     return signature
+
+
+def get_payment_processors():
+    request = get_current_request()
```
clintonb: Don't bother with raising an error. Pass the request as an argument and let the caller worry about where that request comes from. Better yet, make this an instance method on `SiteConfiguration`.

ecommerce/settings/_oscar.py
```diff
@@ -107,23 +107,25 @@
...
-    },
-    'paypal': {
-        # 'mode' can be either 'sandbox' or 'live'
-        'mode': None,
-        'client_id': None,
-        'client_secret': None,
-        'receipt_path': PAYMENT_PROCESSOR_RECEIPT_PATH,
-        'cancel_path': PAYMENT_PROCESSOR_CANCEL_PATH,
-        'error_path': PAYMENT_PROCESSOR_ERROR_PATH,
+    'edX': {
```
clintonb: Always compare the lowercase partner code to avoid issues with case sensitivity.

ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -47,9 +50,33 @@ def handle(self, *args, **options):
...
-        action = None
+
+        if len(args) < 2:
+            raise CommandError("Required arguments 'config' and 'action' are missing")
+
+        partner = args.pop(0)
+        action = args.pop(0)
+
+        if partner not in settings.PAYMENT_PROCESSOR_CONFIG:
+            raise CommandError("Payment Processor configuration for partner {0} not found".format(partner))
```
clintonb: Single quotes or double? Pick one is be consistent throughout the file.

ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -47,9 +50,33 @@ def handle(self, *args, **options):
         Main dispatch.
         """
         args = list(args)
-        action = None
+
+        if len(args) < 2:
+            raise CommandError("Required arguments 'config' and 'action' are missing")
+
+        partner = args.pop(0)
```
clintonb, line 57: Lowercase this value.

ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -47,9 +50,33 @@ def handle(self, *args, **options):
...
+        action = args.pop(0)
+
+        if partner not in settings.PAYMENT_PROCESSOR_CONFIG:
```
clintonb: Combine these two checks: 

```
paypal_configuration = settings.PAYMENT_PROCESSOR_CONFIG.get(partner, {}).get(self.PAYPAL_CONFIG_KEY)
if not paypal_configuration:
  ...
```

ecommerce/core/models.py
e-kolpakov: @clintonb I can't guarantee there are no "poisonous" data already in staging/production DBs, so at least for initial phase this code should be tolerant to such cases

clintonb: This is the first time we are actually using the fields. Anyone configuring a Site will need to update these values. We should be as strict as possible here. If your site breaks because the values are wrong, that's a good thing. You need to fix your site.

ecommerce/extensions/payment/processors/__init__.py
```diff
@@ -59,7 +62,11 @@ def configuration(self):
         Raises:
             KeyError: If no settings found for this payment processor
         """
-        return settings.PAYMENT_PROCESSOR_CONFIG[self.NAME]
+        request = get_current_request()
```
clintonb, line 65: I really don't like this pattern of looking for a request. I would much rather the request be a parameter on the methods/constructors.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -47,6 +47,14 @@ def __init__(self):
         self.retry_attempts = self.configuration.get('retry_attempts', 1)
 
     @property
```
clintonb: Use `@cached_property` to avoid instantiating this multiple times.

ecommerce/extensions/payment/tests/test_processors.py
```diff
@@ -61,6 +62,18 @@ class PaymentProcessorTestCaseMixin(RefundTestMixin, CourseCatalogTestMixin, Pay
 
     CERTIFICATE_TYPE = 'test-certificate-type'
 
+    @classmethod
+    def setUpClass(cls):
+        super(PaymentProcessorTestCaseMixin, cls).setUpClass()
+        paypal_configuration = settings.PAYMENT_PROCESSOR_CONFIG['edX'].get('paypal')
+        if paypal_configuration:
```
clintonb: Why is this necessary given that the payment processor does this already?

ecommerce/extensions/payment/tests/test_processors.py
```diff
@@ -50,6 +50,7 @@
 log = logging.getLogger(__name__)
 
 
+@ddt.ddt
```
clintonb: This file has grown a bit too large. My cutoff is around 400 lines. Please break this out, preferably in a separate commit, into multiple files, one per processor.

ecommerce/extensions/payment/processors/__init__.py
douglashall, line 65: Maybe move the pattern into a custom `get_current_request()` function? The issue with requiring a a request parameter is that you are just pushing the pattern upstream.

clintonb, line 65: I'm okay with pushing the pattern upstream, especially given the lack of documentation of the request requirement and exception. I would much rather know that I need a request to instantiate the class than learn about it via an unhandled exception later. I see no place where we are catching `MissingRequestError` outside of tests. This code would be much safer if it were explicit about its need for a request.

clintonb:
Looks good, but I would like to see the outstanding comments addressed before this is merged.

ecommerce/extensions/payment/processors/__init__.py
e-kolpakov, line 65: @clintonb @douglashall Initial implementation passed configuration to payment processors constructor/initializer (i.e. `__init__` - some are just pedantic about it not being an actual constructor :) ). However, there was a problem somewhere, so I had to revert to getting settings in base class property from using threadlocal request. I can't recall what was the exact problem (likely it was that payment processor was created in some place _very_ far from where `request` is available) - I'll try again, but I trust my past self on it, so likely it'll have to stay as it is.

ecommerce/extensions/payment/tests/test_processors.py
e-kolpakov: @clintonb the test itself creates a lot of paypal SDK objects, and they require the api, either statically configured (like this), or passed as constructor parameter (like in `Paypal` class). Without this, the test fails with

```
  File "/work/OpenCraft/edX/analytics-devstack/ecommerce/ecommerce/extensions/payment/tests/test_processors.py", line 631, in test_issue_credit_error
    paypal_refund = paypalrestsdk.Refund({'error': expected_response})
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/paypalrestsdk/resource.py", line 14, in __init__
    self.__dict__['api'] = api or default_api()
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/paypalrestsdk/api.py", line 311, in default
    Refer https://github.com/paypal/rest-api-sdk-python#configuration")
MissingConfig: Required PAYPAL_CLIENT_ID and PAYPAL_CLIENT_SECRET.  
```

clintonb: Please add this reasoning as a comment.

Also, what happens if `paypal_configuration` is empty? If this is required, the line above should use `['paypal']` instead of `get()`.

ecommerce/core/models.py
e-kolpakov: @clintonb ok.

ecommerce/extensions/payment/processors/__init__.py
e-kolpakov, line 65: @clintonb @douglashall yes, the problematic place is `Refund._issue_credit:

```
 processor = get_processor_class_by_name(source.source_type.name)()
```

The longest call chain is as follows:
- RefundCreateView.create - request available as parameter
- refund.api.create_refunds
- Refund.create_with_lines
- Refund.approve
- Refund._issue_credit

So, to me using threadlocal request is a lesser evil than passing the request / payment processor configuration through that four calls.

e-kolpakov:
@clintonb I've addressed your comments (except one, which I would like to discuss first) - please take another look.

e-kolpakov:
@clintonb still need your approval on this.

@douglashall the contents have changed since your approval (not much, but still) - does your +1 still holds?

ecommerce/core/models.py
```diff
@@ -45,6 +50,35 @@ class SiteConfiguration(models.Model):
     class Meta(object):
         unique_together = ('site', 'partner')
 
+    @property
+    def payment_processors_set(self):
+        return {
```
clintonb, line 65: This should return a `list`. There is no need to return a `dict`, given that you only ever look at the keys.

ecommerce/core/models.py
```diff
@@ -45,6 +50,35 @@ class SiteConfiguration(models.Model):
...
+            raw_processor_value.strip()
+            for raw_processor_value in self.payment_processors.split(',')
+        }
+
+    def _clean_payment_processors(self):
+        if self.payment_processors.isspace():
+            raise ValidationError("Invalid payment processors field: must not only contain whitespace characters")
```
clintonb: Single quotes.

ecommerce/core/models.py
```diff
@@ -45,6 +50,35 @@ class SiteConfiguration(models.Model):
...
 
+    @property
+    def payment_processors_set(self):
+        return {
+            raw_processor_value.strip()
+            for raw_processor_value in self.payment_processors.split(',')
+        }
+
+    def _clean_payment_processors(self):
+        if self.payment_processors.isspace():
```
clintonb: You're really like this `isspace()` case. This can be improved a tad:

``` python
def _clean_payment_processors(self):
  value = self.payment_processors.strip()
  if not value:
    raise ValidationError('The payment processors field must contain a non-empty value.')

processor_names = value.split(',')
for name in processor_names:
  try:
    get_processor_class_by_name(name.strip())
  except ProcessorNotFoundError:
    raise ValidationError('[{name}] is not a valid payment processor.' .format(name=name))
```

ecommerce/core/tests/test_models.py
```diff
@@ -45,3 +54,110 @@ class BusinessClientTests(TestCase):
...
+@ddt.ddt
+class SiteConfigurationTests(TestCase):
+    @ddt.data(
+        ("paypal", {"paypal"}),
+        ("paypal", {"paypal"}),
+        ("paypal,cybersource", {"paypal", "cybersource"}),
+        ("paypal,cybersource,something_else", {"paypal", "cybersource", "something_else"}),
+    )
+    @ddt.unpack
+    def test_allowed_payment_processors(self, payment_processors, expected_result):
```
clintonb: 1. Give me a docstring for every test explaining what it's testing.
2. This test is more appropriately named `test_payment_processor_field_parsing`.

ecommerce/core/tests/test_models.py
```diff
@@ -45,3 +54,110 @@ class BusinessClientTests(TestCase):
...
+        site_config = SiteConfiguration(payment_processors=payment_processors)
+        self.assertEqual(site_config.payment_processors_set, expected_result)
+
+    @mock.patch("ecommerce.extensions.payment.helpers.get_processor_class_by_name")
```
clintonb: Nit: You can move this inside the test method.

ecommerce/core/tests/test_models.py
```diff
@@ -45,3 +54,110 @@ class BusinessClientTests(TestCase):
...
+    @ddt.data(" ", "  \t ", "\t\n\r")
+    def test_clean_fields_whitespace_payment_processor_fail_validation(self, payment_processors):
+        site_config = SiteConfiguration(payment_processors=payment_processors)
+        with self.assertRaises(ValidationError) as err:
+            site_config.clean_fields()
+            self.assertEqual(
+                err.message, "Invalid payment processors field: must not only contain whitespace characters"
+            )
+
+    @mock.patch("ecommerce.extensions.payment.helpers.get_processor_class_by_name")
```
clintonb: You don't need to patch this.

ecommerce/core/tests/test_models.py
```diff
@@ -45,3 +54,110 @@ class BusinessClientTests(TestCase):
...
+    def test_clean_fields_payment_processor_excluded_always_pass(self, patched_proc_by_name):
+        patched_proc_by_name.side_effect = exceptions.ProcessorNotFoundError
+        try:
+            site_config = SiteConfiguration(payment_processors="")
+            site_config.clean_fields(exclude={"payment_processors"})
+
+            site_config.payment_processors = "irrelevant-get_processor_by_name-is-patched"
+            site_config.clean_fields(exclude={"payment_processors"})
+        except ValidationError as exc:
```
clintonb: Don't bother catching the exception. Raise exceptions will fail the test.

ecommerce/core/tests/test_models.py
```diff
@@ -45,3 +54,110 @@ class BusinessClientTests(TestCase):
...
+
+    @mock.patch("ecommerce.extensions.payment.helpers.get_processor_class_by_name")
+    def test_clean_fields_unknown_payment_processor_fail_validation(self, patched_proc_by_name):
+        site_config = SiteConfiguration(payment_processors="irrelevant-get_processor_by_name-is-patched")
+        patched_proc_by_name.side_effect = exceptions.ProcessorNotFoundError
+
+        with self.assertRaises(ValidationError):
+            site_config.clean_fields()
+
+    @mock.patch("ecommerce.extensions.payment.helpers.get_processor_class_by_name")
```
clintonb: No need to patch.

ecommerce/core/tests/test_models.py
```diff
@@ -45,3 +54,110 @@ class BusinessClientTests(TestCase):
...
+    def _enable_processor_switches(processors):
+        for processor in processors:
+            toggle_switch(settings.PAYMENT_PROCESSOR_SWITCH_PREFIX + processor.NAME, True)
+
+    @override_settings(PAYMENT_PROCESSORS=[
+        'ecommerce.extensions.payment.tests.processors.DummyProcessor',
+        'ecommerce.extensions.payment.tests.processors.AnotherDummyProcessor',
+    ])
+    @ddt.data(
+        ([], []),
```
clintonb, line 141: This is not a valid case.

ecommerce/core/tests/test_models.py
```diff
@@ -45,3 +54,110 @@ class BusinessClientTests(TestCase):
...
+        result = site_config.get_payment_processors()
+        self.assertEqual(result, expected_result)
+
+    @override_settings(PAYMENT_PROCESSORS=[
+        'ecommerce.extensions.payment.tests.processors.DummyProcessor',
+    ])
+    @ddt.data(
+        ([], []),
+        ([DummyProcessor], [DummyProcessor]),
+        ([DummyProcessor, AnotherDummyProcessor], [DummyProcessor]),
```
clintonb: This seems wrong. If only a subset of processors exists, the others should not be in the list, and an error should be raised.

ecommerce/extensions/api/v2/tests/views/test_payments.py
```diff
@@ -16,8 +17,13 @@ class PaymentProcessorListViewTests(TestCase):
     def setUp(self):
         super(PaymentProcessorListViewTests, self).setUp()
         self.token = self.generate_jwt_token_header(self.create_user())
-        self.toggle_payment_processor('dummy', True)
-        self.toggle_payment_processor('another-dummy', True)
+        self.toggle_payment_processor(DummyProcessor.NAME, True)
+        self.toggle_payment_processor(AnotherDummyProcessor.NAME, True)
+
+        site_config, __ = SiteConfiguration.objects.get_or_create(site__id=1)
+        site_config.payment_processors = ",".join([DummyProcessor.NAME, AnotherDummyProcessor.NAME])
```
clintonb, line 26: Use `update_or_create` and pass this value in the defaults dict.

clintonb, line 26: Alternatively, just use `self.site.site_configuration`.

ecommerce/core/models.py
douglashall, line 65: This is a set comprehension. A set will be returned.

clintonb, line 65: Ah. Add a docstring, please.

ecommerce/core/models.py
```diff
@@ -45,6 +50,35 @@ class SiteConfiguration(models.Model):
...
+            raise ValidationError("Invalid payment processors field: must not only contain whitespace characters")
+
+        split_payment_processors = self.payment_processors.split(',')
+        for proc_name in split_payment_processors:
+            try:
+                get_processor_class_by_name(proc_name.strip())
+            except ProcessorNotFoundError as exc:
+                raise ValidationError(exc.message)
+
+    def get_payment_processors(self):
```
clintonb, line 93: Add a docstring explaining what this returns.

ecommerce/core/models.py
```diff
@@ -45,6 +50,35 @@ class SiteConfiguration(models.Model):
...
+        all_processors = [get_processor_class(path) for path in settings.PAYMENT_PROCESSORS]
+        return [
+            processor for processor in all_processors
+            if processor.NAME in self.payment_processors_set and processor.is_enabled()
+        ]
+
+    def clean_fields(self, exclude=None):
```
clintonb, line 115: Docstring.

ecommerce/core/tests/test_models.py
e-kolpakov: @clintonb this way is slightly semantically better as:
1. It explicitly expresses that test checks for raised/not raised exception (of particular type); without this a comment like "if it raises - test failed" is needed.
2. It expects a particular exception to be not raised, so raising that exception is technically a failed test, not an error.

clintonb: Regardless of the exception raised, an exception is a failure. All you're doing by catching the exception is hiding the stack trace, which would actually help debug the broken test.

ecommerce/core/tests/test_models.py
e-kolpakov: @clintonb there are at least one somewhat valid use case for this:
- Initially both Dummy and AnotherDummy processors was enabled in settings
- SiteConfiguration with both Dummy and AnotherDummy processor was created
- AnotherDummy was later removed from settings.

In two words, list of "available" processors can change after a model is created, and there's no code so far that could protect from such change.

Also, this test tests the method, not a model itself, so it can't rely on the fact that validation will fail a record with unknown processor.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -122,6 +123,10 @@ def setUp(self):
         self.user = self.create_user()
         self.client.login(username=self.user.username, password=self.password)
         self.course = CourseFactory(name='BasketSummaryTest')
+        self.site_configuration = SiteConfiguration.objects.get(site__id=1)
```
clintonb: `self.site_configuration = self.site.site_configuration`

ecommerce/core/tests/test_models.py
clintonb: If there is concern about settings changing, and breaking the model, we should have startup validation. The app doesn't start unless every `Site` has access to its listed payment processors. At the very least, if you don't want to go the validation route, you should log a warning whenever the `Site` expects a payment processor that is not configured in settings.

I would much rather have startup validation given that we have different settings across environments. Things might be fine in stage, where we run acceptance tests. However, we don't test production after a release. We would have to wait for error alerts or support calls. If something is going to break, it should break as early in the process as possible.

ecommerce/core/models.py
e-kolpakov: @clintonb ok, but I would like to point out that most of the user-facing strings use double quotes. And "internal" strings (dictionary keys, internal code, etc.) use single quote. And in some places these rules are violated, including this file (i.e. `theme_scss_path` and `payment_processors` help_texts).

ecommerce/extensions/payment/tests/processors/base.py
```diff
@@ -0,0 +1,71 @@
+# -*- coding: utf-8 -*-
```
clintonb: Rename to mixins.py.

ecommerce/core/models.py
clintonb: You have the dubious honor of being the first, and probably, only person's code I review today. That means you get the well-rested reviewer, hence more strict about using one quote style throughout a file/module/codebase.

`lms_url_root` and `payment_processors` are exceptions because they include single quotes in the help text. If you see violations, please fix them. Others might not be as strict as I am, hence the mix of styles.

ecommerce/extensions/payment/processors/__init__.py
clintonb, line 65: Leave it, but document all errors being raised.

clintonb:
Here's what needs to be done to earn my approval:
- [x] Remaining comments addressed
- [x] Some form of validation added to ensure `Site`s are not configured to use payment processors that are not in settings. The last thing we want is to display a checkout form to the user with no ability to pay.

ecommerce/core/tests/test_models.py
e-kolpakov, line 141: @clintonb it is valid for the method under test; `payment_processor` field "non-emptiness" is enforced by validation code and `blank=False`, `null=False` in field declaration, which will make this test depend on validation/field configuration, i.e. make it more brittle.

ecommerce/core/tests/test_models.py
e-kolpakov: @clintonb ok, I'll see if I could add validating site configurations at startup.

ecommerce/core/models.py
e-kolpakov, line 115: @clintonb self-explanatory, but ok, I'll add it.

ecommerce/core/models.py
e-kolpakov: @clintonb sure, but could I have a list of rules to follow on this? It looks like it is as foolows:
1. If the string contains `'` or `"` in body, use other quotes to avoid escaping.
2. If it is a docstring use three double quotes
3. If it is an internal thing (i.e. dict key, `getattr` argument, etc.) - us single quote.
4. Otherwise use double quote.

Is that list correct and complete?

ecommerce/core/tests/test_models.py
e-kolpakov: @clintonb technically, non-catched exception is an error, not a failure; little attention is paid to the difference (as both "fail" the entire suite/build/etc.), but to me "failure" is an expected thing - we expect this to do that and it was not the case, while "error" is - we expect this to do that, but something totally unrelated prevented it from doing so.

Anyway, in order to not engage into a prolonged debate for three lines of code, I'll remove the `try..except` clause.

ecommerce/extensions/basket/tests/test_views.py
e-kolpakov: @clintonb does not work like that:

```
  File "/work/OpenCraft/edX/analytics-devstack/ecommerce/ecommerce/extensions/basket/tests/test_views.py", line 127, in setUp
    site_configuration = self.site.site_configuration
AttributeError: 'Site' object has no attribute 'site_configuration'
```

ecommerce/extensions/api/v2/tests/views/test_payments.py
e-kolpakov, line 26: @clintonb There's a problem with `update_or_create`, sine it requires passing partner_id, site_id and so on, and my best guesses was not right (site_id=1 and partner_id=1 caused DB integrity violation), so it looks like `get_or_create` is more robust.

`self.site.site_configuration` raises `AttributeError`

douglashall:
@e-kolpakov This still looks good to me. Thanks for all your work on this.

ecommerce/core/models.py
clintonb: The only time you need double quotes are for docstrings and strings containing a single quote that you don't want to escape. That's the de-facto rule that many of us have been following.

ecommerce/extensions/basket/tests/test_views.py
clintonb: My mistake. This is not a foreign key relationship. In that case, minor improvement:

``` python
self.site_configuration, __ = SiteConfiguration.objects.update_or_create(site__id=self.site.id, defaults={'payment_processors': DummyProcessor.NAME})
```

ecommerce/extensions/api/v2/tests/views/test_payments.py
clintonb, line 26: `self.site.siteconfiguration`

ecommerce/extensions/basket/tests/test_views.py
clintonb: `self.site.siteconfiguration`

ecommerce/wsgi.py
```diff
@@ -32,3 +31,11 @@
 # Apply WSGI middleware here.
 # from helloworld.wsgi import HelloWorldApplication
 # application = HelloWorldApplication(application)
+
+
+def application_ready():
```
e-kolpakov: @clintonb here's the motivation for it:
1. In order to validate models field values we need to fetch them from DB.
2. We can't do that in AppConfig, as [Django docs](https://docs.djangoproject.com/en/1.9/ref/applications/#django.apps.AppConfig.ready) explicitly tells it's a bad idea (and I've tried that anyway and it couldn't even run the tests)
3. We can't do that in `models.py`, as SiteConfiguration references model from other app and it is not ready (and likely that's just the first problem on the way; most likely it won't work anyway, since `syncdb` needs models to be ready to create DB, and we're trying to access DB before models module is processed)
4. I've considered using middleware for this, but it would be executed on every request, which means +N SQL queries for evry request (N = SiteConfiguration). It could be solved by some global flag, but it felt more and more hackish, so I have decided to not follow that path.

And that's how I ended with this - it is executed once, fails startup (raises ValidateException, does not respond to HTTP requests at all) and covered by tests to the maximum possible extent.

If you've got better ideas for how to achieve this - let me know.

Makefile
```diff
@@ -78,6 +79,12 @@ static:
 html_coverage:
 	coverage html && open htmlcov/index.html
 
+diff_coverage: validate fast_diff_coverage
+
+fast_diff_coverage:
+	coverage xml
+	diff-cover coverage.xml --compare-branch=$(DIFF_COVER_BASE_BRANCH)
```
e-kolpakov, line 86: @clintonb allows checking diff coverage locally and not wait for push, CI build and codecov processing.

e-kolpakov:
@clintonb addressed remaining comments (including validating models on startup), please take another look.

P.S. sorry if I'm getting bothersome - merge deadline is close (I've just added it to description, sorry for not communicating it earlier), so I would like to speed up the process (in terms of results, not the effort put)

Makefile
clintonb, line 86: Diff coverage isn't necessary if you ensure your code has 100% coverage.

ecommerce/core/models.py
```diff
@@ -45,6 +55,63 @@ class SiteConfiguration(models.Model):
...
+        Returns:
+            list[BasePaymentProcessor]: Returns payment processor classes enabled for the corresponding Site
+        """
+        all_processors = [get_processor_class(path) for path in settings.PAYMENT_PROCESSORS]
+        all_processor_names = {processor.NAME for processor in all_processors}
+
+        missing_processor_configurations = self.payment_processors_set - all_processor_names
+        if missing_processor_configurations:
+            processor_config_repr = ", ".join(missing_processor_configurations)
+            log.warning(
```
clintonb, line 106: If you're using this for startup validation, logging a warning won't be good enough. You need to raise an error to actually stop execution.

ecommerce/wsgi.py
clintonb: I would prefer to see this in `AppConfig.ready()`, despite Django's warning. We are definitely not modifying any data, and the handful of instances we'd be iterating over are not enough to drastically slow down execution of management commands.

You'll most likely have to either duplicate code, or move it to a utility function/module, to avoid importing `models.py` at startup.

Makefile
e-kolpakov, line 86: @clintonb this is a tool to help ensure my code have 100% coverage. Also, codecov checks for 90% diff coverage, and does not allow overall coverage to drop. The only reasonable way to satisfy both conditions is to check that diff have 100% coverage. So this tool helps to ensure that.

Also, `make validate` coverage report rounds the coverage to integer, so it is not possible to tell if "coverage decrease" check is passed or not.

clintonb, line 86: Okay.

ecommerce/core/models.py
e-kolpakov, line 106: @clintonb no, validation is performed in:
- wsgi.py -> application_ready
- core.models -> validate_configuration
- core.models.SiteConfiguration -> clean_fields

clintonb, line 106: Ah. Yes. My mistake.

ecommerce/core/models.py
```diff
@@ -45,6 +55,63 @@ class SiteConfiguration(models.Model):
...
+        value = self.payment_processors.strip()
+        if not value:
+            raise ValidationError('Invalid payment processors field: must not only contain whitespace characters')
+
+        processor_names = value.split(',')
+        for name in processor_names:
+            try:
+                get_processor_class_by_name(name.strip())
+            except ProcessorNotFoundError as exc:
+                raise ValidationError(exc.message)
```
clintonb, line 91: Add a call to `log.exception()` as well, so we don't lose the stack trace.

ecommerce/wsgi.py
e-kolpakov: @clintonb I've tried putting it into `AppConfig.ready` and it didn't work: AppConfig.ready is executed before an app can sync models to DB, so even reading form DB is not possible. In particular:

```
# core.config.py
from django.apps import AppConfig

class CoreAppConfig(AppConfig):
    name = 'ecommerce.core'
    verbose_name = 'Core'

    def ready(self):
        super(CoreAppConfig, self).ready()

        # Ensures that the initialized Celery app is loaded when Django starts.
        # Allows Celery tasks to bind themselves to an initialized instance of the Celery library.
        from ecommerce import celery_app  # pylint: disable=unused-variable

        from .models import validate_configuration
        validate_configuration()
```

Crashed with

```
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 328, in execute
    django.setup()
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/__init__.py", line 18, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/apps/registry.py", line 115, in populate
    app_config.ready()
  File "/work/OpenCraft/edX/analytics-devstack/ecommerce/ecommerce/core/config.py", line 16, in ready
    validate_configuration()
  File "/work/OpenCraft/edX/analytics-devstack/ecommerce/ecommerce/core/models.py", line 152, in validate_configuration
    for config in SiteConfiguration.objects.all():
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 162, in __iter__
    self._fetch_all()
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 238, in iterator
    results = compiler.execute_sql()
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 840, in execute_sql
    cursor.execute(sql, params)
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/db/utils.py", line 98, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/home/john/.virtualenvs/ecommerce/local/lib/python2.7/site-packages/django/db/backends/sqlite3/base.py", line 318, in execute
    return Database.Cursor.execute(self, query, params)
OperationalError: no such table: core_siteconfiguration
```

mattdrayer:
@saleem-latif, just FYI on the work being done in this branch

ecommerce/wsgi.py
clintonb: Catch, and log, the `OperationalError`. Add a comment stating why it's safe to ignore for tests.

e-kolpakov:
@clintonb anything else needs to be done here?

ecommerce/core/config.py
```diff
@@ -1,13 +1,32 @@
...
         # Allows Celery tasks to bind themselves to an initialized instance of the Celery library.
         from ecommerce import celery_app  # pylint: disable=unused-variable
+
+        from ecommerce.core.models import validate_configuration
+        # Operational error means database did not contain SiteConfiguration table - ok to skip since it means there
+        # are no SiteConfiguration models to validate. Also, this exception was only observed in tests and test run
+        # just fine with this suppression - most likely test code hit this line before the DB is populated
+        try:
+            # Here we're are consciously violating Django's warning about not interacting with DB in AppConfig.ready
+            # We now what we're doing, have considered a couple of other approaches and discussed it in great length:
```
clintonb: s/now/know

clintonb:
:+1: pending last comment addressed, commits squashed

ecommerce/wsgi.py
smarnach: When running `./manage.py migrate` with a completely empty DB, I got a `ProgrammingError` here rather than an `OperationalError`:

```
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 328, in execute
    django.setup()
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/__init__.py", line 18, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/apps/registry.py", line 115, in populate
    app_config.ready()
  File "/edx/app/ecommerce/ecommerce/ecommerce/core/config.py", line 30, in ready
    validate_configuration()
  File "/edx/app/ecommerce/ecommerce/ecommerce/core/models.py", line 158, in validate_configuration
    for config in SiteConfiguration.objects.all():
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 162, in __iter__
    self._fetch_all()
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 238, in iterator
    results = compiler.execute_sql()
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 840, in execute_sql
    cursor.execute(sql, params)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/utils.py", line 98, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/backends/mysql/base.py", line 124, in execute
    return self.cursor.execute(query, args)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/MySQLdb/cursors.py", line 205, in execute
    self.errorhandler(self, exc, value)
    raise errorclass, errorvalue
django.db.utils.ProgrammingError: (1146, "Table 'ecommerce.core_siteconfiguration' doesn't exist")
```

I got there by dropping and recreating the `ecommerce` database manually, but I suspect the same thing would happen on a fresh install.  I've got no idea why a missing table results in different exceptions being raised, but simply adding `ProgrammingError` to the list of ignored exceptions fixed the problem for me.

bradenmacdonald: @smarnach Yep, the ECOM team discovered that and we have https://openedx.atlassian.net/browse/WL-390 planned to fix it.

openedx-webhooks:
Thanks for the pull request, @e-kolpakov! I've created [OSPR-1694](https://openedx.atlassian.net/browse/OSPR-1694) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

If you like, you can add yourself to the [AUTHORS](https://github.com/edx/ecommerce/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

e-kolpakov:
@edx-webhook That's so kind of you, thank you! Please note this PR was merged like a year ago though. :)

gsong:
@e-kolpakov Yeah, we're investigating some bugs with the bot.

PR #632, opened 2016-03-21
Multitenant payment processor configuration

e-kolpakov:
**Description:** Same as #630, but against the base branch to facilitate code review (as it only includes changes related to the task).

See #630 for description, explanation, JIRA ticket, testing instructions, etc.

### DIFF
--- a/ecommerce/extensions/payment/config.py
+++ b/ecommerce/extensions/payment/config.py
@@ -1,21 +1,5 @@
-from django.conf import settings
 from oscar.apps.payment import config
-import paypalrestsdk
 
 
 class PaymentConfig(config.PaymentConfig):
     name = 'ecommerce.extensions.payment'
-
-    def ready(self):
-        paypal_configuration = settings.PAYMENT_PROCESSOR_CONFIG.get('paypal')
-        if paypal_configuration:
-            # Initialize the PayPal REST SDK
-            paypalrestsdk.configure({
-                'mode': paypal_configuration['mode'],
-                'client_id': paypal_configuration['client_id'],
-                'client_secret': paypal_configuration['client_secret']
-            })
-
-        # Register signal handlers
-        # noinspection PyUnresolvedReferences
-        import ecommerce.extensions.payment.signals  # pylint: disable=unused-variable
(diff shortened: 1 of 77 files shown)


openedx-webhooks:
Thanks for the pull request, @e-kolpakov! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=632&repo=edx%2Fecommerce

ecommerce/extensions/payment/views.py
```diff
@@ -35,7 +35,9 @@
 
 class CybersourceNotifyView(EdxOrderPlacementMixin, View):
     """ Validates a response from CyberSource and processes the associated basket/order appropriately. """
-    payment_processor = Cybersource()
+    @property
```
jbzdak: If we are considering maintainability of this code, it would be good to move this to ``EdxOrderPlacementMixin``, as this is the type that requires ``self.payment_processor`` to be a defined instance, and define payment processor type as a class constant.

mattdrayer:
@douglashall, just FYI on this PR

ecommerce/extensions/payment/config.py
```diff
@@ -7,7 +7,7 @@ class PaymentConfig(config.PaymentConfig):
     name = 'ecommerce.extensions.payment'
 
     def ready(self):
-        paypal_configuration = settings.PAYMENT_PROCESSOR_CONFIG.get('paypal')
+        paypal_configuration = settings.PAYMENT_PROCESSOR_CONFIG['default'].get('paypal')
         if paypal_configuration:
             # Initialize the PayPal REST SDK
             paypalrestsdk.configure({
```
jbzdak: Please double-check this actually makes sense. ``paypalrestsdk`` could have different config than currently used microsite, which could lead to trouble.

PR #633, opened 2016-03-21
Correct messages logged during LMS footer retrieval

rlucioni:
Having correct log messages helps a lot when trying to understand errors in production. [Change-detector tests](http://googletesting.blogspot.com/2015/01/testing-on-toilet-change-detector-tests.html) are also hard to work with. FYI @vkaracic.


clintonb:
:+1:

vkaracic:
@rlucioni thanks for fixing this :+1:

PR #634, opened 2016-03-22
Add separate requirements file for acceptance testing

rlucioni:
Jenkins nodes don't currently have libffi-dev installed. The solution presented here, a separate requirements file for acceptance tests, isn't ideal because it forces the ecommerce project to duplicate some dependencies listed in other (modular) requirements files. That said, having Jenkins install a bunch of superfluous dependencies isn't ideal either.

Here's the challenge. Say we include the new acceptance.txt as part of test.txt. We then list `requests` in acceptance.txt. Developers can no longer install dependencies using local.txt, since `requests` is now listed twice; it's already listed in base.txt. We don't want to move `requests` out of base.txt because anyone running production.txt won't have it installed, and we don't want to install acceptance testing dependencies in production.

@benpatterson @clintonb @schenedx @jzoldak thoughts?

### DIFF
--- a/acceptance_tests/pages/coupons.py
+++ b/acceptance_tests/pages/coupons.py
@@ -1,7 +1,8 @@
+import random
+# The string module is not deprecated. This is a bug in Pylint: https://www.logilab.org/ticket/2481.
+import string  # pylint: disable=deprecated-module
+
 from bok_choy.javascript import wait_for_js
-
-from factory.fuzzy import FuzzyText
-
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
@@ -14,7 +15,9 @@ from acceptance_tests.pages.ecommerce import EcommerceAppPage
 def _get_coupon_name(is_discount):
     """ Returns an appropriate coupon name. """
     prefix = 'test-discount-code-' if is_discount else 'test-enrollment-code-'
-    return FuzzyText(length=3, prefix=prefix).fuzz()
+    postfix = ''.join(random.choice(string.ascii_letters) for _ in range(3))
+
+    return prefix + postfix
 
 
 class BasketPage(EcommerceAppPage):
--- a/acceptance_tests/requirements.txt
+++ b/acceptance_tests/requirements.txt
@@ -1,0 +1,8 @@
+# Packages required to run acceptance tests
+bok-choy==0.5.0
+ddt==1.0.1
+django-nose==1.4.2
+edx-rest-api-client==1.5.0
+nose-ignore-docstring==0.2
+requests==2.9.1
+selenium>=2.53.1
--- a/requirements/monitoring/requirements.txt
+++ b/requirements/monitoring/requirements.txt
@@ -10,3 +10,4 @@
 -r ../optional.txt
 -r ../production.txt
 -r ../test.txt
+-r ../../acceptance_tests/requirements.txt


clintonb:
1. What's the problem you are trying to solve?
2. Where does `libffi-dev` come into play?
3. Why can't `libffi-dev` be installed on Jenkins?

rlucioni:
> 1. What's the problem you are trying to solve?

The goal is to run acceptance tests using Jenkins jobs. In order to do so, these jobs need to install dependencies required to run the tests.

> 2. Where does `libffi-dev` come into play?

At the moment, `libssl-dev` and `libffi-dev` are prerequisites to installing `paypalrestsdk`, listed in ecommerce's base requirements. (Ecommerce's base requirements are installed when installing test requirements via test.txt.) I assume `libssl-dev` is already installed on Jenkins worker nodes.

> 3. Why can't `libffi-dev` be installed on Jenkins?

It probably can be, but I don't know what installing it involves. @benpatterson?

benpatterson:
My position is if we are going to do remote testing on a node, then that node should only have what it needs to do remote testing. The way requirements are organized now is around localhost.

I don't want to create a separate node for every kind of IDA we might be running end-to-end tests for, especially when doing that task really requires the same kinds of dependencies for all IDAs (e.g., bok-choy, nose, etc). I understand the notion of keeping requirements DRY, but duplication in requirements files is far cheaper than producing several different nodes depending on what's being tested.

I do think we can set up a model where everyone wins if we rethink how the requirements are laid out. What we really want is some sort of acceptance-test.txt that does _not_ call base.txt, but maybe it calls a common test.txt file. 

Let me know if you need help with this; I can put some extra thought into it.

schenedx:
@rlucioni How did you test the set of dependencies within this document is the minimum set we need for acceptance testing?
I think with what we are trying to do, we have to duplicate the dependency somehow like @rlucioni mentioned. I hate to bring this up, but it seems to me that acceptance tests are more and more detached from the actual code within the code base. There is nothing stops us from combining all acceptance tests for edx-platform, programs, ecommerce and even credentials move into a single repo called "edx-acceptance-tests". @benpatterson Is this something we considered and decide to not go to that route?

clintonb:
:+1: pending
- [x] Move requirements to `acceptance_tests/requirements.txt`
- [x] Add reference to new file in `requirements/monitoring/requirements.txt`.

benpatterson:
@schenedx in reference to your earlier question:
- It is ok to ask...
- We have considered it, and coming up with an answer is basically next on our list once we land on the right integration test (i.e., API contract validation) framework. At this point we are leaning towards keeping tests in their respective repos, and having a way of collecting them all for a unified test run. But...I'm talking more about _api contracts_ testing, which is easier to isolate. When it comes to _end-to-end_ testing (like what this PR is attempting to address), I'd like to keep things generally closer to the pertinent repos, but also recognize there are going to be test cases that really cover multiple repos (which is the whole idea of end-to-end tests). We have a nascent repo for that based on some work the solutions team is doing, but we haven't formalized anything. This, too, is basically something we'll think about once we land on an API contracts testing pattern/framework. Happy to discuss more in chat or in person. For this PR though, I'd call a solution to that problem as out of scope.

rlucioni:
Thanks for the input everyone. @clintonb I've addressed your comments. @schenedx, I determined this to be the minimum set of dependencies required by looking at the packages used by the acceptance tests, installing them in a fresh virtualenv, then running the tests.

schenedx:
:+1:

rlucioni:
@clintonb updated to remove the factory-boy dependency, per our offline conversation.

PR #635, opened 2016-03-22
Rework coupon order and invoice record.

mjfrey:
This PR re-works the way we record orders for Coupons and Invoices.

Changes include:
- Create new InvoiceClient model
- Added InvoiceClient to the Django admin
- Add following FKs to Invoice model
  - Order
  - InvoiceClient
- Added Invoices the the Django admin
- Updated Coupon API
  - request.user for order
  - Create Invoice using Order and InvoiceClient
- Update Invoice Payment Processor to user Order and InvoiceClient
- Updated tests

### DIFF
--- a/ecommerce/core/admin.py
+++ b/ecommerce/core/admin.py
@@ -2,7 +2,7 @@ from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
 from django.utils.translation import ugettext_lazy as _
 
-from ecommerce.core.models import SiteConfiguration, User
+from ecommerce.core.models import BusinessClient, SiteConfiguration, User
 
 
 class SiteConfigurationAdmin(admin.ModelAdmin):
@@ -20,6 +20,10 @@ class EcommerceUserAdmin(UserAdmin):
         (_('Important dates'), {'fields': ('last_login', 'date_joined')}),
     )
 
+
+@admin.register(BusinessClient)
+class BusinessClientAdmin(admin.ModelAdmin):
+    pass
 
 admin.site.register(SiteConfiguration, SiteConfigurationAdmin)
 admin.site.register(User, EcommerceUserAdmin)
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -69,5 +69,13 @@ class User(AbstractUser):
 
 
 class Client(User):
-    """The model for the business client."""
     pass
+
+
+class BusinessClient(models.Model):
+    """The model for the business client."""
+
+    name = models.CharField(_('Name'), unique=True, max_length=255, blank=False, null=False)
+
+    def __str__(self):
+        return self.name
--- a/ecommerce/core/tests/test_models.py
+++ b/ecommerce/core/tests/test_models.py
@@ -1,4 +1,4 @@
-from ecommerce.core.models import User
+from ecommerce.core.models import BusinessClient, User
 from ecommerce.tests.testcases import TestCase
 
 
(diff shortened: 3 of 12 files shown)


mjfrey:
@clintonb @vkaracic can you please take a look.

ecommerce/invoice/models.py
```diff
@@ -10,20 +10,22 @@ class Invoice(TimeStampedModel):
...
+    basket = models.ForeignKey('basket.Basket', null=True, blank=False)
+    order = models.ForeignKey('order.Order', null=True, blank=False)
+    invoice_client = models.ForeignKey('core.InvoiceClient', null=True, blank=False)
     state = models.CharField(max_length=255, default=NOT_PAID, choices=state_choices)
 
     history = HistoricalRecords()
 
     def __str__(self):
-        return 'Invoice {id} for order number {order}'.format(id=self.id, order=self.basket.order.number)
+        return 'Invoice {id} for order number {order}'.format(id=self.id, order=self.order)
```
vkaracic: Should the order part be from `self.order.number` ?

vkaracic:
I presume the idea is that the user who creates the coupon is bound to the basket and the client is to the order now?

mjfrey:
@vkaracic The order is now recorded as who creates the Coupon.  The Invoice contains the client and a link to the order.

ecommerce/core/models.py
```diff
@@ -62,5 +62,14 @@ def get_full_name(self):
 
 
 class Client(User):
-    """The model for the business client."""
     pass
+
+
+class InvoiceClient(models.Model):
```
clintonb: `BusinessClient`, or anything not related to invoices. This is more than just invoices.

ecommerce/core/models.py
```diff
@@ -62,5 +62,14 @@ def get_full_name(self):
...
+    """The model for the business client."""
+
+    full_name = models.CharField(_('Full Name'), unique=True, default="",
```
clintonb: This field should just be `name`. Remove the default.

ecommerce/core/tests/test_models.py
```diff
@@ -38,3 +38,10 @@ def test_get_full_name(self):
 
         user = self.create_user(full_name=full_name, first_name=first_name, last_name=last_name)
         self.assertEquals(user.get_full_name(), full_name)
+
+
+class InvoiceClientTests(TestCase):
+
+    def test_full_name(self):
```
clintonb: No need to test basic model fields. Test the `str` method instead.

ecommerce/core/models.py
```diff
@@ -62,5 +62,14 @@ def get_full_name(self):
 
 
 class Client(User):
```
vkaracic, line 71: Can we remove this class?

ecommerce/core/tests/test_models.py
vkaracic: Isn't that what's being tested here?

clintonb: It is, actually. The test is mis-named.

ecommerce/core/models.py
mjfrey, line 71: Not yet.  Need to migrate all data from it first.  Then drop.

mjfrey:
@clintonb @vkaracic Comments addressed.  Can you give another look?

ecommerce/core/admin.py
```diff
@@ -21,5 +21,9 @@ class EcommerceUserAdmin(UserAdmin):
     )
 
 
+@admin.register(BusinessClient)
+class InvoiceClientAdmin(admin.ModelAdmin):
```
clintonb: Correct the name.

ecommerce/invoice/models.py
```diff
@@ -10,20 +10,22 @@ class Invoice(TimeStampedModel):
         (NOT_PAID, _('Not Paid')),
         (PAID, _('Paid')),
     )
-    basket = models.ForeignKey('basket.Basket', null=False, blank=False)
+    basket = models.ForeignKey('basket.Basket', null=True, blank=False)
```
clintonb: `blank=True`

clintonb:
:+1: once comments addressed.

vkaracic:
Sorry, I forgot to leave the emoticon
:+1:

PR #637, opened 2016-03-22
Added gate around CyberSource Level II/III fields

clintonb:
Not all CyberSource payment processors support this data. This flag ensures open source users have the option of disabling this functionality.

ECOM-3960

### DIFF
--- a/ecommerce/extensions/payment/tests/test_processors.py
+++ b/ecommerce/extensions/payment/tests/test_processors.py
@@ -2,6 +2,7 @@
 """Unit tests of payment processor implementations."""
 from __future__ import unicode_literals
 
+import copy
 import datetime
 import json
 import logging
@@ -11,6 +12,7 @@ from uuid import UUID
 import ddt
 from django.conf import settings
 from django.core.urlresolvers import reverse
+from django.test import override_settings
 from django.test.client import RequestFactory
 import mock
 from oscar.apps.payment.exceptions import TransactionDeclined, UserCancelled, GatewayError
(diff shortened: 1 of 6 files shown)


clintonb:
@rlucioni @peter-fogg please review

ecommerce/extensions/payment/tests/test_processors.py
```diff
@@ -153,18 +159,37 @@ def get_expected_transaction_parameters(self, transaction_uuid):
...
 
+    def test_get_transaction_parameters(self):
+        """ Verify the processor returns the appropriate parameters required to complete a transaction. """
+        # NOTE (CCB): Make a deepcopy of the settings so that we can modify them without affecting the real settings.
+        # This is a bit simpler than using modify_copy(), which would does not support nested dictionaries.
+        payment_processor_config = copy.deepcopy(settings.PAYMENT_PROCESSOR_CONFIG)
+        payment_processor_config[self.processor_name]['send_level_2_3_details'] = False
+
+        with override_settings(PAYMENT_PROCESSOR_CONFIG=payment_processor_config):
+            self.assert_correct_transaction_parameters(False)
```
rlucioni: For clarity, I might use the keyword argument you've added as a keyword argument:

``` python
self.assert_correct_transaction_parameters(include_level_2_3_details=False)
```

rlucioni:
:+1:

ecommerce/extensions/payment/tests/test_processors.py
clintonb: Updated.

PR #639, opened 2016-03-23
Avoid unnecessary lambda in acceptance tests

rlucioni:
`EmptyPromise` must be initialized with a callable, but that doesn't require use of a lambda. @clintonb @schenedx


clintonb:
:+1:

PR #641, opened 2016-03-25
Default value of oauth_settings should be an empty dict, not a str

douglashall:
This fixes a problem with rolling out the multitenancy work to stage/production, allowing us to fall back on configured Django settings until we have a change to update the oauth_settings field of the SiteConfiguration with the appropriate values.

@mjfrey @clintonb


mjfrey:
:+1:

PR #642, opened 2016-03-28
Default JWT_LEEWAY to 1 second

rlucioni:
Allows ecommerce to tolerate slight clock skew. Companion to https://github.com/edx/configuration/pull/2892. The base JWT_AUTH dictionary modified here is [overwritten](https://github.com/edx/ecommerce/blob/master/ecommerce/settings/production.py#L48) in production.py. This change makes the setting an application default.

@jimabramson


jimabramson:
:+1:

PR #643, opened 2016-03-29
Add coupon attribute to audit log

mjfrey:
Add an attribute to the audit log for determining if the order was for a Coupon.  This allows us to modify alerts and filter out such orders.

@vkaracic @rlucioni Please take a look.

### DIFF
--- a/ecommerce/extensions/checkout/mixins.py
+++ b/ecommerce/extensions/checkout/mixins.py
@@ -101,7 +101,8 @@ class EdxOrderPlacementMixin(OrderPlacementMixin):
             basket_id=order.basket.id,
             currency=order.currency,
             order_number=order.number,
-            user_id=order.user.id
+            user_id=order.user.id,
+            contains_coupon=order.contains_coupon
         )
 
         if waffle.sample_is_active('async_order_fulfillment'):
--- a/ecommerce/extensions/order/models.py
+++ b/ecommerce/extensions/order/models.py
@@ -15,6 +15,12 @@ class Order(AbstractOrder):
         """Returns a boolean indicating if order can be fulfilled."""
         return self.status in (ORDER.OPEN, ORDER.FULFILLMENT_ERROR)
 
+    @property
+    def contains_coupon(self):
+        """ Return a boolean if the order contains a Coupon. """
+        return any(line.product.get_product_class().name == 'Coupon' for line in
+                   self.basket.all_lines())
+
 
 class PaymentEvent(AbstractPaymentEvent):
     processor_name = models.CharField(_("Payment Processor"), max_length=32, blank=True, null=True)
--- a/ecommerce/extensions/order/tests/test_models.py
+++ b/ecommerce/extensions/order/tests/test_models.py
@@ -27,3 +27,14 @@ class OrderTests(TestCase):
         self.order.status = status
         self.order.save()
         self.assertFalse(self.order.is_fulfillable)
+
+    def test_contains_coupon(self):
+        self.assertFalse(self.order.contains_coupon)
+
+        product_class = u'Coupon'
+        product = factories.create_product(product_class=product_class)
+        basket = factories.create_basket(empty=True)
+        factories.create_stockrecord(product, num_in_stock=1)
+        basket.add_product(product)
+        order = factories.create_order(basket=basket)
+        self.assertTrue(order.contains_coupon)
--- a/ecommerce/extensions/checkout/tests/test_mixins.py
+++ b/ecommerce/extensions/checkout/tests/test_mixins.py
@@ -103,9 +103,11 @@ class EdxOrderPlacementMixinTests(BusinessIntelligenceMixin, RefundTestMixin, Te
                 (
                     LOGGER_NAME,
                     'INFO',
-                    'order_placed: amount="{}", basket_id="{}", currency="{}", order_number="{}", user_id="{}"'.format(
+                    'order_placed: amount="{}", basket_id="{}", contains_coupon="{}", currency="{}",'
+                    ' order_number="{}", user_id="{}"'.format(
                         self.order.total_excl_tax,
                         self.order.basket.id,
+                        self.order.contains_coupon,
                         self.order.currency,
                         self.order.number,
                         self.order.user.id


ecommerce/extensions/checkout/utils.py
```diff
@@ -9,6 +9,20 @@
 logger = logging.getLogger(__name__)
 
 
+def order_contains(class_name, order):
+    """ Check to see if order contains a particualr product class
```
vkaracic: Typo :)

ecommerce/extensions/checkout/tests/test_utils.py
```diff
@@ -41,3 +43,8 @@ def test_exceptions(self, exception):
         """ Verify the function returns None when a request exception is raised. """
         with mock.patch.object(requests, 'get', mock.Mock(side_effect=exception)):
             self.assertIsNone(get_provider_data('ABC'))
+
+    def test_order_contains(self):
+        """ Verify order_contains matches on the provided product class """
+        order = factories.create_order()
+        self.assertTrue(order_contains(u'Dùｍϻϒ item class', order))
```
vkaracic: It might be a bit over-the-top, but what do you think about adding a custom product class name instead of relying on Oscar's which could change in the future and we would need to change this. Something like:

```
        product_class = u'Dùｍϻϒ'
        product = factories.create_product(product_class=product_class)
        basket = factories.create_basket(empty=True)
        factories.create_stockrecord(product, num_in_stock=1)
        basket.add_product(product)
        order = factories.create_order(basket=basket)
        self.assertTrue(order_contains(product_class, order))
```

Ignore this comment if you think it's unnecessary.

vkaracic:
The code looks good to me, although I'm not familiar how the filtering is set up in Splunk.
:+1:

ecommerce/extensions/checkout/tests/test_utils.py
```diff
@@ -41,3 +42,14 @@ def test_exceptions(self, exception):
...
+    def test_order_contains(self):
+        """ Verify order_contains matches on the provided product class """
+        product_class = u'Seat'
+        product = factories.create_product(product_class=product_class)
+        basket = factories.create_basket(empty=True)
+        factories.create_stockrecord(product, num_in_stock=1)
+        basket.add_product(product)
+        order = factories.create_order(basket=basket)
+        self.assertTrue(order_contains(product_class, order))
+        self.assertTrue(order_contains(u'Seat', order))
```
vkaracic: You're doing the same assertions as the one above.

ecommerce/extensions/checkout/utils.py
```diff
@@ -9,6 +9,20 @@
 logger = logging.getLogger(__name__)
 
 
+def order_contains(class_name, order):
```
rlucioni: @mjfrey this would work nicely as a property on the Order model (e.g., `contains_coupon`). For an example, see [`is_fulfillable`](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/order/models.py#L13-L16).

ecommerce/extensions/checkout/tests/test_utils.py
```diff
@@ -41,3 +42,13 @@ def test_exceptions(self, exception):
         """ Verify the function returns None when a request exception is raised. """
         with mock.patch.object(requests, 'get', mock.Mock(side_effect=exception)):
             self.assertIsNone(get_provider_data('ABC'))
+
+    def test_order_contains(self):
```
rlucioni: You'll want to group this test with our other Order model tests.

ecommerce/extensions/checkout/tests/test_utils.py
```diff
@@ -41,3 +42,13 @@ def test_exceptions(self, exception):
...
+        """ Verify order_contains matches on the provided product class """
+        product_class = u'Seat'
+        product = factories.create_product(product_class=product_class)
+        basket = factories.create_basket(empty=True)
+        factories.create_stockrecord(product, num_in_stock=1)
+        basket.add_product(product)
+        order = factories.create_order(basket=basket)
+        self.assertTrue(order_contains(product_class, order))
```
rlucioni: Please also assert that the property returns False when the order does not contain a coupon product.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -101,7 +102,8 @@ def handle_successful_order(self, order):
             basket_id=order.basket.id,
             currency=order.currency,
             order_number=order.number,
-            user_id=order.user.id
+            user_id=order.user.id,
+            coupon_order=order_contains(class_name='Coupon', order=order)
```
rlucioni: For readability, let the field name match the property name (e.g., `contains_coupon`).

rlucioni:
@mjfrey I've completed my pass. I've made some comments to help you clean this up a bit.

mjfrey:
@rlucioni Comments addressed.

ecommerce/extensions/order/tests/test_models.py
```diff
@@ -27,3 +27,13 @@ def test_is_not_fulfillable(self, status):
...
+
+    def test_contains_coupon(self):
+        product_class = u'Coupon'
+        product = factories.create_product(product_class=product_class)
+        basket = factories.create_basket(empty=True)
+        factories.create_stockrecord(product, num_in_stock=1)
+        basket.add_product(product)
+        order = factories.create_order(basket=basket)
+        self.assertTrue(order.contains_coupon)
+        self.assertFalse(self.order.contains_coupon)
```
rlucioni: For readability, start with this assertion on its own line, then include the rest separated by a blank line. It's surprising to see it tacked on down here.

ecommerce/extensions/order/models.py
```diff
@@ -6,6 +6,8 @@
 
 from ecommerce.extensions.fulfillment.status import ORDER
 
+COUPON_CLASS_NAME = 'Coupon'
```
rlucioni: The 'Coupon' string is only used in a single place here, so I don't see the need for this constant. I'd use the string directly below.

rlucioni:
@mjfrey I made a few additional, minor comments. :+1: once those are addressed and commits are squashed.

PR #644, opened 2016-03-29
[SOL-1636] Refactor get_voucher_from_code

vkaracic:
When no voucher exists for given code and when no products are associated with the voucher, raise exceptions instead of returning None.

This is a fix for: https://openedx.atlassian.net/browse/SOL-1636

### DIFF
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -17,6 +17,7 @@ from edx_rest_api_client.exceptions import SlumberHttpBaseException
 
 from ecommerce.core.url_utils import get_lms_url
 from ecommerce.core.views import StaffOnlyMixin
+from ecommerce.extensions.api import exceptions
 from ecommerce.extensions.analytics.utils import prepare_analytics_data
 from ecommerce.extensions.api.constants import APIConstants as AC
 from ecommerce.extensions.api.data import get_lms_footer
@@ -44,20 +45,19 @@ def get_voucher_from_code(code):
     Returns:
         voucher (Voucher): The Voucher for the passed code.
         product (Product): The Product associated with the Voucher.
+
+    Raises:
+        Voucher.DoesNotExist: When no vouchers with provided code exist.
+        ProductNotFoundError: When no products are associated with the voucher.
     """
-    voucher = None
-    product = None
-    # Check to see if a voucher exists for the code.
-    try:
-        voucher = Voucher.objects.get(code=code)
-    except Voucher.DoesNotExist:
-        logger.exception('Voucher does not exist for code [%s].', code)
-        return voucher, product
+    voucher = Voucher.objects.get(code=code)
 
     # Just get the first product.
     products = voucher.offers.all()[0].benefit.range.all_products()
     if products:
         product = products[0]
+    else:
+        raise exceptions.ProductNotFoundError()
     return voucher, product
 
 
@@ -114,7 +114,18 @@ class CouponOfferView(TemplateView):
         footer = get_lms_footer()
         code = self.request.GET.get('code', None)
         if code is not None:
-            voucher, product = get_voucher_from_code(code=code)
+            try:
+                voucher, product = get_voucher_from_code(code=code)
+            except Voucher.DoesNotExist:
+                return {
+                    'error': _('Coupon does not exist'),
+                    'footer': footer,
+                }
+            except exceptions.ProductNotFoundError:
+                return {
+                    'error': _('The voucher is not applicable to your current basket.'),
+                    'footer': footer,
+                }
             valid_voucher, msg = voucher_is_valid(voucher, product, self.request)
             if valid_voucher:
                 api = EdxRestApiClient(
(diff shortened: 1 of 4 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=644&repo=edx%2Fecommerce

vkaracic:
@mjfrey please review.

vkaracic:
I've added try..except blocks to preserve the existing error messages.

ecommerce/coupons/views.py
```diff
@@ -114,7 +114,13 @@ def get_context_data(self, **kwargs):
         footer = get_lms_footer()
         code = self.request.GET.get('code', None)
         if code is not None:
-            voucher, product = get_voucher_from_code(code=code)
+            try:
+                voucher, product = get_voucher_from_code(code=code)
+            except Voucher.DoesNotExist:
```
mjfrey, line 119: should you add exceptions.ProductNotFoundError to this line?

ecommerce/coupons/views.py
```diff
@@ -182,8 +188,11 @@ def get(self, request):
 
         if not code:
             return render(request, template_name, {'error': _('Code not provided')})
+        try:
+            voucher, product = get_voucher_from_code(code=code)
+        except Voucher.DoesNotExist:
```
mjfrey, line 198: same here.

vkaracic:
@mjfrey I've added an exception catch for `ProductNotFoundError` and I've set it to return `No courses associated with coupon` as the error message.

mjfrey:
:+1:

vkaracic:
The project test coverage was failing even though the patch coverage is at 100% so I've added a couple of tests that are unrelated to boost that coverage. I don't want a negative project test coverage to affect other PR's.

mjfrey:
:+1: 
Please squash

vkaracic:
I'd like to keep the commits separate because they aren't related (the two last ones are just tests for unrelated code to keep the project coverage).

ecommerce/coupons/views.py
```diff
@@ -182,8 +193,13 @@ def get(self, request):
...
+            return render(request, template_name, {'error': _('Coupon does not exist')})
+        except exceptions.ProductNotFoundError:
+            return render(request, template_name, {'error': _('No courses associated with coupon')})
```
clintonb: No products are associated with this coupon.

ecommerce/coupons/views.py
```diff
@@ -182,8 +193,13 @@ def get(self, request):
 
         if not code:
             return render(request, template_name, {'error': _('Code not provided')})
+        try:
+            voucher, product = get_voucher_from_code(code=code)
+        except Voucher.DoesNotExist:
+            return render(request, template_name, {'error': _('Coupon does not exist')})
```
clintonb: The coupon is not valid.

ecommerce/extensions/fulfillment/tests/test_api.py
```diff
@@ -51,6 +52,14 @@ def test_fulfill_order_incorrect_module(self):
         self.assertEquals(ORDER.FULFILLMENT_ERROR, self.order.status)
         self.assertEquals(LINE.FULFILLMENT_CONFIGURATION_ERROR, self.order.lines.all()[0].status)
 
+    @override_settings(FULFILLMENT_MODULES=['ecommerce.extensions.fulfillment.tests.modules.InvalidModule', ])
```
clintonb: It is not clear what you're trying to test here. What is an "invalid" module?

ecommerce/extensions/test/factories.py
```diff
@@ -54,7 +54,10 @@ def prepare_voucher(code='COUPONTEST', _range=None, start_datetime=None, end_dat
         product = ProductFactory(title='Test product')
         _range = RangeFactory(products=[product, ])
     else:
-        product = _range.all_products()[0]
+        if _range.num_products() > 0:
```
clintonb: 1. Combine with the `else` above.
2. Why is this necessary?

ecommerce/extensions/fulfillment/tests/test_api.py
vkaracic: It's a empty class module:
https://github.com/edx/ecommerce/pull/644/files#diff-2bfb94efa2bc715ec47e21d16132c336R66
used to trigger an unexpected error that would in turn trigger an exception catch here:
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/fulfillment/api.py#L68
That's the line not covered by a test.

ecommerce/extensions/test/factories.py
vkaracic: It's for the `prepare_voucher` function to accept a range without a product:
https://github.com/edx/ecommerce/pull/644/files#diff-549419197ceb5146d6c1f4c27c33fee2R188

ecommerce/extensions/fulfillment/tests/test_api.py
clintonb: This would be a bit cleaner if you patched a method on an existing fulfillment module, even the dummy one, to raise an error.

ecommerce/extensions/test/factories.py
clintonb: Okay. Merge into an `elif`.

clintonb:
:+1: once comments addressed.

ecommerce/coupons/views.py
clintonb: Actually, use the current message: No voucher found with code '??'.

ecommerce/coupons/views.py
clintonb: Correction: The voucher is not applicable to your current basket.

vkaracic:
@clintonb I believe I have addressed all your comments, can you take another look?

clintonb:
:+1:

PR #646, opened 2016-03-30
Remove multi-use coupon usage limitation

vkaracic:
We currently are not supporting multi-use voucher types and this PR removes that option from the `Usage Limitations` dropdown.

@mjfrey

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -55,7 +55,7 @@ class CouponViewSetTest(CouponMixin, CourseCatalogTestMixin, TestCase):
             'code': '',
             'quantity': 2,
             'start_date': '2015-1-1',
-            'voucher_type': Voucher.MULTI_USE,
+            'voucher_type': Voucher.ONCE_PER_CUSTOMER,
             'categories': [self.category],
             'note': None,
         }
@@ -117,7 +117,7 @@ class CouponViewSetTest(CouponMixin, CourseCatalogTestMixin, TestCase):
         self.assertEqual(Product.objects.filter(product_class=self.product_class).count(), 1)
         self.assertEqual(StockRecord.objects.filter(product=coupon_append).count(), 1)
         self.assertEqual(coupon_append.attr.coupon_vouchers.vouchers.count(), 7)
-        self.assertEqual(coupon_append.attr.coupon_vouchers.vouchers.filter(usage=Voucher.MULTI_USE).count(), 2)
+        self.assertEqual(coupon_append.attr.coupon_vouchers.vouchers.filter(usage=Voucher.ONCE_PER_CUSTOMER).count(), 2)
 
     def test_custom_code_string(self):
         """Test creating a coupon with custom voucher code."""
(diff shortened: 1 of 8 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=646&repo=edx%2Fecommerce

mjfrey:
:+1:

vkaracic:
@marjev can you please review this?

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -55,7 +55,7 @@ def setUp(self):
             'code': '',
             'quantity': 2,
             'start_date': '2015-1-1',
-            'voucher_type': Voucher.MULTI_USE,
+            'voucher_type': Voucher.ONCE_PER_CUSTOMER,
```
marjev, line 58: What happens if API endpoint receives voucher type `Voucher.ONCE_PER_CUSTOMER`?

vkaracic, line 58: Creates new vouchers with `ONCE_PER_CUSTOMER` type.

marjev, line 58: Sorry, I meant to ask what happens if API endpoint receives voucher type `Voucher.MULTI_USE`?

I think that we want to test that use case as well.

vkaracic, line 58: Oh, well the whole purpose of this PR is to remove that voucher type from our flow. That option is removed in the dropdown on the coupon creation form and should never be sent out.

marjev, line 58: Yes, but the API is not aware of the form. Someone can send a request via API and our backend would process that, so we must handle that case. The Open-source community is not aware of our internal work and since this is not properly documented I believe that the best course of action here is to test that as well to be sure that API endpoint accepts that voucher type or that it raises an error on back-end if we don't support that on our backend. That's my only concern here. I'm not that familiar with the backend requirements here, but we must test that the result matches the requirements.

marjev:
Other than one concern, code looks great! :+1:

ecommerce/extensions/api/v2/tests/views/test_coupons.py
vkaracic, line 58: Good thinking! I think this commit cover that use-case:
https://github.com/edx/ecommerce/pull/646/commits/afa840fe499041f8298d78e873c1c6fc81c5d32c

Please take another look :)

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -85,6 +85,10 @@ def create(self, request, *args, **kwargs):
             client, __ = BusinessClient.objects.get_or_create(name=client_username)
             note = request.data.get('note', None)
 
+            # We currently do not support multi-usa voucher types.
```
marjev: I think you have a typo in the comment. Nitpick.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
marjev, line 58: Looks much better. Thanks!

PR #647, opened 2016-03-31
Make sure basket is empty before coupon order

mjfrey:
Flush the basket before we create a coupon order.

### DIFF
--- a/ecommerce/tests/mixins.py
+++ b/ecommerce/tests/mixins.py
@@ -23,6 +23,7 @@ from ecommerce.core.url_utils import get_lms_url
 from ecommerce.courses.utils import mode_for_seat
 from ecommerce.extensions.api.constants import APIConstants as AC
 from ecommerce.extensions.api.v2.views.coupons import CouponViewSet
+from ecommerce.extensions.basket.utils import prepare_basket
 from ecommerce.extensions.fulfillment.signals import SHIPPING_EVENT_NAME
 from ecommerce.tests.factories import PartnerFactory, SiteConfigurationFactory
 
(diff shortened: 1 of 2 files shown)


mjfrey:
@rlucioni

rlucioni:
@mjfrey please add a regression test. Also, is there still a place for [`ecommerce.extensions.baskets.utils.prepare_basket()`](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/utils.py#L11)? It performs similar operations, including flushing the basket.

mjfrey:
@rlucioni Took your suggestion as this is better.

ecommerce/tests/mixins.py
```diff
@@ -391,12 +392,11 @@ def create_coupon(
...
-            product=coupon,
-            client=factories.UserFactory(),
-            site=self.site,
-            partner=partner
-        )
+        request = RequestFactory()
+        request.site = self.site
+        request.user = factories.UserFactory()
+
+        self.basket = prepare_basket(request, coupon)
```
rlucioni, line 399: @mjfrey this doesn't actually test the observed behavior, does it? We want to verify that the basket is flushed.

mjfrey, line 399: There are already tests for prepare_basket that ensure a single item. They are in test_utils.py

rlucioni, line 399: @mjfrey never mind, prepare_basket's tests cover this case: https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/tests/test_utils.py#L73

rlucioni:
:+1:

mjfrey:
@rlucioni not sure on how to merge since I am .01 coverage short.
Can you merge this?

rlucioni:
@mjfrey I cannot, only @clintonb has that power :expressionless:. I'm looking around in our codecov settings to see if we can get the build to pass when coverage drops by small percentages like this. Coveralls allows it, I'm not sure about Codecov.

rlucioni:
@mjfrey I changed a setting which should pass the build as long as coverage is > 95%. Try restarting your build.

rlucioni:
Voila!

PR #648, opened 2016-04-01
[SOL-1721] Change Total Paid Coupon Field

marjev:


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=648&repo=edx%2Fecommerce

marjev:
@mjfrey Here's a PR for [SOL-1721](https://openedx.atlassian.net/browse/SOL-1721). Please review.

Here's the screenshot:
new-create-new-coupon-page

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -247,23 +251,30 @@ define([
...
                 }
             },
 
+            changePrice: function () {
+                this.updatePrice(this.getSeatData());
+            },
+
+            updatePrice: function (seatData) {
+                var quantity = this.$el.find('input[name=quantity]').val(),
+                    price = quantity * seatData.price;
```
mjfrey, line 273: Shouldn't we be calculating the price less any discount?  If they are discount codes we need to take that into account.  The total price would be (unit price \* discount) \* quantity

marjev, line 273: Please discuss with @pbaruah. I just implemented the calculation that I found in the story description, which is:

> Calculation: Seat price X number of codes

pbaruah:
@marjev , @mjfrey - this is the total value of the coupon before discount. This amount goes in the "Price" field in the downloadable CSV report. Could you please take out the word discount to read "number of coupon codes X unit price per seat"

marjev:
@mjfrey Is it ok to merge this PR?

mjfrey:
@marjev can you update the screen shot?  Did you change the text according to @pbaruah

marjev:
@mjfrey yes, I did the update. Here's the new screenshot:

screen shot 2016-04-06 at 14 49 53

mjfrey:
:+1:

PR #649, opened 2016-04-01
WL-330: Multi-Site Branding in Otto

saleem-latif:
Hi @mattdrayer 

I have completed changes related to WL-330, Kindly review these changes and share your thoughts on this.

**Description of WL-330:**
Changes included customization of
- Static Files Finders and Static Storage
- Sass compilation workflow
- Template Loaders

and helper functions for enabling comprehensive theming for Otto.


mattdrayer:
@saleem-latif -- just FYI, I rebased this branch onto the latest master -- please push your latest commits when you have a chance -- thanks!

clintonb:
This looks a lot like the app from edx-platform. Given that we have multiple IDAs that will need to consume this, this code should live in an independent app.

mattdrayer:
Thanks @clintonb -- I agree we should consider generalizing the theming implementation.  However I want to make sure we have a better understanding of what is truly reusable before jumping into the creation of another package.  So let's focus on getting this second effort working correctly and afterwards we can move forward with the generalization work -- it has to happen as part of the microsites cleanup anyway.  I've logged WL-1749 to keep track of this job.

ecommerce/settings/base.py
```diff
@@ -112,11 +112,14 @@
 
 # See: https://docs.djangoproject.com/en/dev/ref/contrib/staticfiles/#staticfiles-finders
 STATICFILES_FINDERS = (
+    'ecommerce.theming.finders.ThemeFilesFinder',
     'django.contrib.staticfiles.finders.FileSystemFinder',
     'django.contrib.staticfiles.finders.AppDirectoriesFinder',
     'compressor.finders.CompressorFinder',
 )
 
+STATICFILES_STORAGE = "ecommerce.theming.storage.ThemeStorage"
```
mattdrayer, line 125: Please add a comment here explaining the general idea behind ThemeStorage

ecommerce/settings/base.py
```diff
@@ -152,14 +155,17 @@
...
-        'APP_DIRS': True,
         'DIRS': (
             normpath(join(DJANGO_ROOT, 'templates')),
             # Templates which override default Oscar templates
             normpath(join(DJANGO_ROOT, 'templates/oscar')),
             OSCAR_MAIN_TEMPLATE_DIR,
         ),
         'OPTIONS': {
+            'loaders': [
+                'ecommerce.theming.template_loaders.ThemeTemplateLoader',
```
mattdrayer, line 175: If order is important for these loaders we should note it with a comment

ecommerce/settings/base.py
```diff
@@ -112,11 +112,14 @@
 
 # See: https://docs.djangoproject.com/en/dev/ref/contrib/staticfiles/#staticfiles-finders
```
mattdrayer, line 113: Please add some information here about the ThemeFilesFinder -- purpose, ordering requirements, etc.

ecommerce/startup.py
```diff
@@ -0,0 +1,10 @@
+"""
+Module for code that should run during LMS startup
+"""
+from django.conf import settings
+
+from ecommerce.theming.core import enable_theming
+
+
+def run():
+    enable_theming(settings.COMPREHENSIVE_THEME_DIR)
```
mattdrayer: Please add a docstring for this method

ecommerce/static/sass/base/main.scss
```diff
@@ -1,8 +1,8 @@
 // utilities
 // --------------------
-@import '../utilities/developer';    // developer-friendly file: add rough/WIP styling that needs UI triage
-@import '../utilities/variables';
-@import '../utilities/mixins';
+@import 'utilities/developer';    // developer-friendly file: add rough/WIP styling that needs UI triage
```
mattdrayer, line 3: We should ask someone skilled with Sass to take a look at these particular changes -- perhaps @AlasdairSwan or @andy-armstrong can give us a hand.

ecommerce/theming/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""
+This app contains theming and branding logic for ecommerce.
+"""
```
mattdrayer, line 41: We should include more information here on how to work with the theming app -- it does not need to substitute the official documentation (which will need to be written), but we can include information here specifically for developers.

ecommerce/theming/helpers.py
```diff
@@ -0,0 +1,261 @@
...
+    """
+    Returns boolean indicating whether comprehensive theming functionality is enabled or disabled.
+    Example:
+        >> is_comprehensive_theming_enabled()
+        True
+
+    Returns:
+         (bool): True if comprehensive theming is enabled else False
+    """
+    return True if settings.COMPREHENSIVE_THEME_DIR else False
```
mattdrayer: Several people have commented to me that it would be nicer to have an actual feature flag for comprehensive theming, versus keying-off this directory setting -- I tend to agree with this.

ecommerce/theming/helpers.py
```diff
@@ -0,0 +1,261 @@
...
+        return Path(themes_dir) / theme_dir
+
+
+def get_base_themes_dir():
+    """
+    Return base directory that contains all the themes.
+
+    Example:
+        >> get_base_themes_dir()
+        '/edx/app/edxapp/edx-platform/themes'
```
mattdrayer: Should change this to an Otto-specific example

mattdrayer:
@saleem-latif overall this changeset is looking good -- there are some nitpicky things to address in order to add clarification for future developers but in general I think we're on the right track.  As noted by @clintonb there's a lot of similarity between this change set and what has been implemented in the LMS.  It would be good to begin thinking about how we might design a package that could be used by multiple IDAs, which pieces could move there, which would need to stay -- but let's not view that as a requirement for the current effort.

We have some failing Travis checks to address and some conflicts to resolve, but otherwise I think we're ready to move into full-on review mode.  Let's squash this initial group of commits after the build turns green and then we can do a more formal review.

clintonb:
Having failed to make theming work on LMS devstack/sandboxes after 2 hours last night, I need to see documentation in the docs directory before this gets merged. At a minimum, the following topics need to be covered:
1. Overall theming architecture.
2. Settings/configuration necessary to get started
3. How to create a new theme
4. How to activate/deactivate a theme

mattdrayer:
I agree -- more people are beginning to use theming now and there needs to be clear instruction on how to configure the feature.

saleem-latif:
@clintonb yeah sure, I will start working on docs for theming otto, [this](https://openedx.atlassian.net/wiki/display/SOL/Theming) confluence article has a small checklist for troubleshooting theming on lms that might help in your case.

saleem-latif:
@clintonb I have added docs for theming, kindly take a look

ecommerce/static/sass/base/main.scss
AlasdairSwan, line 3: @saleem-latif can you please explain how the `../` can be removed here?

ecommerce/theming/management/commands/update_assets.py
```diff
@@ -0,0 +1,270 @@
...
+            default='nested',
+            help='Coding style for compiled sass (default="nested").',
+        )
+
+        parser.add_argument(
+            '--enable-source-comments',
+            dest='source_comments',
+            action='store_true',
+            default=False,
+            help="Add source comments in compiles sass.",
```
jimabramson: s/complies/compiled/

ecommerce/theming/management/commands/update_assets.py
```diff
@@ -0,0 +1,270 @@
...
+            *args: Positional arguments passed to the update_assets command
+            **options: optional arguments passed to the update_assets command
+        Returns:
+            A tuple containing parsed values for themes, system, source comments and output style.
+            1. themes (list): list of Theme objects
+            2. system (bool): True if system sass need to be compiled, False otherwise
+            3. source_comments (bool): True if source comments need to be included in output, False otherwise
+            4. output_style (str): Coding style for compiled css files.
+        """
+        system = options.get("system", 'yes').lower()
```
jimabramson: why not just make the option a boolean flag, like `--skip-system` to override the default?

ecommerce/theming/tests/test_theme_style_overrides.py
```diff
@@ -0,0 +1,95 @@
...
+
+        self.assertIn("background-color: #00fa00", main_css)
+
+
+def compile_sass():
+    """
+    Call update assets command to compile system and theme sass.
+    """
+    with override_settings(DEBUG=True):
+        # Compile system adn theme sass files
```
jimabramson: s/adn/and

ecommerce/theming/management/commands/update_assets.py
```diff
@@ -0,0 +1,270 @@
...
+
+    Returns:
+        A tuple containing sass source dir, css destination dir and duration of sass compilation process
+    """
+    output_style = kwargs.get('output_style', 'compressed')
+    source_comments = kwargs.get('source_comments', False)
+    start = datetime.datetime.now()
+
+    if not sass_source_dir.isdir():
+        logger.warning("Sass dir '%s' does not exists.", sass_source_dir)
```
jimabramson: s/exists/exist/

ecommerce/theming/management/commands/update_assets.py
```diff
@@ -0,0 +1,270 @@
...
+            # convert theme names to Theme objects
+            themes = [available_themes.get(theme) for theme in given_themes]
+
+        return themes, system, source_comments, output_style, collect
+
+    def handle(self, *args, **options):
+        """
+        Handle update_assets command.
+        """
+        logger.info("Sass compilation started.\n")
```
jimabramson: is it necessary to emit line breaks / blank lines to logs?

ecommerce/theming/management/commands/update_assets.py
```diff
@@ -0,0 +1,270 @@
...
+
+    return sass_source_dir, css_destination_dir, duration
+
+
+def collect_assets():
+    """
+    Collect static assets.
+    """
+    logger.info("\t\tStarted collecting static assets.")
+    call_command("collectstatic")
```
jimabramson: i suppose it is not confusing to django to call a management command from within another one?   (if it's working and not discouraged, i have no issue with it...)

ecommerce/theming/management/commands/update_assets.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+    output_style = kwargs.get('output_style', 'compressed')
+    source_comments = kwargs.get('source_comments', False)
+    start = datetime.datetime.now()
+
+    if not sass_source_dir.isdir():
+        logger.warning("Sass dir '%s' does not exists.", sass_source_dir)
+        raise ValueError("Sass dir '{dir}' must be a valid directory.".format(dir=sass_source_dir))
+    if not css_destination_dir.isdir():
+        # If css destination directory does not exists, then create one
```
jimabramson: ditto

ecommerce/theming/core.py
```diff
@@ -0,0 +1,27 @@
...
+    Args:
+        themes_dir (str): path to base theme directory
+    """
+    if is_comprehensive_theming_enabled():
+        # proceed only if comprehensive theming in enabled
+        if isinstance(themes_dir, basestring):
+            themes_dir = path(themes_dir)
+
+        for theme in get_themes(themes_dir):
+            locale_dir = os.path.join(themes_dir, theme.theme_dir, "conf", "locale")
```
jimabramson: does the path.py API, widely used elsewhere in this PR, not support this operation?

ecommerce/startup.py
```diff
@@ -0,0 +1,14 @@
+"""
+Module for code that should run during system startup
```
jimabramson: The standard Django way of doing this seems to use a different API, described here https://docs.djangoproject.com/en/dev/ref/applications/#django.apps.AppConfig.ready - is this startup / run API a custom alternative?

ecommerce/static/sass/base/main.scss
saleem-latif, line 3: @AlasdairSwan `ecommerce/static/sass/partials` is added to sass `lookup_paths` when calling [`sass.compile`](https://github.com/edx/ecommerce/pull/649/files#diff-fa75a60e6e7428d0625bbf232e768211R248). I can remove `../` since all of these imports are present in `ecommerce/static/sass/partials`.

The reason behind this change is that I want sass lookup to use absolute paths instead of relative paths for the reason that with relative paths we can not offer overridable sass functionality in themes. With relative paths `@import` will always pick sass from current directory and theme's sass overrides will never be picked.

ecommerce/theming/template_loaders.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class ThemeTemplateLoader(FilesystemLoader):
+    """
+    Filesystem Template loaders to pickup templates from theme directory based on the current site.
+    """
+    is_usable = True
+    _accepts_engine_in_init = True
+
+    def get_template_sources(self, template_name, template_dirs=None):
```
jimabramson, line 20: the superclass return type is `Origin`: https://github.com/django/django/blob/master/django/template/loaders/filesystem.py#L47-L51

Is it a deliberate decision to deviate from that?

ecommerce/theming/tests/test_core.py
```diff
@@ -0,0 +1,36 @@
...
+        expected_locale_paths = (TEST_THEME / "conf" / "locale", ) + settings.LOCALE_PATHS
+        enable_theming(TEST_THEME.dirname())
+
+        self.assertItemsEqual(expected_locale_paths, settings.LOCALE_PATHS)
+
+    def test_enable_theming_red_theme(self):
+        """
+        Tests that locale path is added only if it exists.
+        """
+        red_theme = Path(settings.COMPREHENSIVE_THEME_DIR + "/red-theme")
```
jimabramson: it may benefit future maintainers to add an inline comment to make explicit the assumption that the red-theme has no translations dir in it.

ecommerce/theming/helpers.py
```diff
@@ -0,0 +1,264 @@
...
+    if not is_comprehensive_theming_enabled():
+        return None
+
+    site_theme = site and site.themes.first()
+    theme_dir = getattr(site_theme, "theme_dir_name") if site_theme else None
+
+    if theme_dir:
+        themes_dir = get_base_themes_dir()
+
+        return Path(themes_dir) / theme_dir
```
jimabramson, line 162: Initially i misread these lines - I think an explicit `else: return None` might make sense

ecommerce/theming/apps.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    name = 'ecommerce.theming'
+    verbose_name = 'Theming'
+
+    def ready(self):
+        """
+        startup run method, this method is called after the application has successfully initialized.
+        Anything that needs to executed once (and only once) the theming app starts can be placed here.
+        """
+        enable_theming(settings.COMPREHENSIVE_THEME_DIR)
```
jimabramson: if `COMPREHENSIVE_THEME_DIR` is set to a non-existent / invalid directory, will any error surface upon startup?   Looking at `enable_theming` and the underlying helpers, it seems like it might just do nothing, and the problem might not be apparent until a request arrived, or theme overrides might simply not work.

Would it be better to validate the setting / raise here during startup, giving the admin a chance to fix the configuration?

ecommerce/theming/template_loaders.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        Returns the absolute paths to "template_name", when appended to each
+        directory in "template_dirs". Any paths that don't lie inside one of the
+        template dirs are excluded from the result set, for security reasons.
+        """
+        if not template_dirs:
+            template_dirs = self.engine.dirs
```
jimabramson, line 27: suggest using [`self.get_dirs()`](https://github.com/django/django/blob/master/django/template/loaders/filesystem.py#L19) here instead

ecommerce/theming/tests/test_commands.py
```diff
@@ -0,0 +1,117 @@
+"""
+Tests for Management commands of comprehensive theming.
+"""
+from path import Path
+from mock import patch, Mock
```
jimabramson, line 4: nit: 3p imports should be grouped with the django ones, below, and alphabetized

ecommerce/theming/tests/test_commands.py
```diff
@@ -0,0 +1,117 @@
...
+
+from django.conf import settings
+from django.test import TestCase
+from django.core.management import call_command, CommandError
+
+from ecommerce.theming.helpers import get_themes
+from ecommerce.theming.management.commands.update_assets import SYSTEM_SASS_PATHS
+from ecommerce.theming.management.commands.update_assets import get_sass_directories, compile_sass, Command
```
jimabramson: nit: condense multiple imports from same module

ecommerce/theming/tests/test_commands.py
```diff
@@ -0,0 +1,117 @@
...
+                "lookup_paths": SYSTEM_SASS_PATHS,
+            }
+        ]
+
+        returned_dirs = get_sass_directories(themes=[], system=True)
+        self.assertItemsEqual(expected_directories, returned_dirs)
+
+    def test_get_theme_sass_directories_with_no_themes(self):
+        """
+        Test that proper sass dirs are returned by get_theme_sass_directories
```
jimabramson: nit: docstring is identical to that of previous test, probably needs better description

ecommerce/theming/tests/test_commands.py
```diff
@@ -0,0 +1,117 @@
...
+        with self.assertRaises(ValueError):
+            compile_sass(
+                sass_source_dir=RED_THEME / "sass" / "non-existent",
+                css_destination_dir=RED_THEME / "static" / "css" / "base",
+                lookup_paths=[RED_THEME / "static" / "sass" / "partials"] + SYSTEM_SASS_PATHS
+            )
+
+    def test_collect_static(self):
+        """
+        Test ValueError is raised if sass directory provided to the compile_sass method does not exist.
```
jimabramson: docstring is incorrect

ecommerce/theming/tests/test_finders.py
```diff
@@ -0,0 +1,42 @@
...
+from django.test import TestCase
+
+from ecommerce.theming.finders import ThemeFilesFinder
+
+RED_THEME = Path(settings.COMPREHENSIVE_THEME_DIR + "/red-theme")
+
+
+class TestThemeFinders(TestCase):
+    """
+    Test comprehensive theming static files storage.
```
jimabramson: docstring is incorrect

ecommerce/theming/tests/test_core.py
```diff
@@ -0,0 +1,36 @@
+"""
+Comprehensive Theming tests for core functionality.
+"""
+from path import Path
+
+from django.conf import settings
```
jimabramson, line 4: nit: group / alphabetize 3p imports

ecommerce/theming/tests/test_finders.py
```diff
@@ -0,0 +1,42 @@
...
+
+from django.conf import settings
+from django.test import TestCase
+
+from ecommerce.theming.finders import ThemeFilesFinder
+
+RED_THEME = Path(settings.COMPREHENSIVE_THEME_DIR + "/red-theme")
+
+
+class TestThemeFinders(TestCase):
```
jimabramson, line 11: This test case seems to cover only a small portion of the non-trivial logic in `ThemeFilesFinder`.  Would you consider adding more coverage, or is that provided by other tests?

ecommerce/theming/tests/test_helpers.py
```diff
@@ -0,0 +1,153 @@
...
+"""
+from path import Path
+from mock import patch
+
+from django.test import override_settings
+from django.conf import settings, ImproperlyConfigured
+
+from ecommerce.tests.testcases import TestCase
+from ecommerce.theming.helpers import get_current_site_theme_dir, get_themes, Theme, get_theme_dir, \
+    get_current_theme_template_dirs, get_all_theme_template_dirs, get_base_themes_dir
```
jimabramson: nit: consider grouping using multiline, comma-separated names enclosed in parens

ecommerce/theming/tests/test_helpers.py
```diff
@@ -0,0 +1,153 @@
...
+        Tests get_all_theme_template_dirs returns correct template dirs for all the themes.
+        """
+        theme_dir = Path(settings.DJANGO_ROOT + "/themes/red-theme")
+
+        expected_theme_dirs = [
+            theme_dir / "templates",
+            theme_dir / "templates" / "oscar",
+        ]
+        actual_theme_dirs = get_all_theme_template_dirs()
+        self.assertItemsEqual(expected_theme_dirs, actual_theme_dirs)
```
jimabramson: it seems like this should at least be tested with more than one theme.

ecommerce/theming/tests/test_helpers.py
```diff
@@ -0,0 +1,153 @@
+"""
+Tests of comprehensive theming.
+"""
```
jimabramson, line 3: general comment on this test module: a lot of tests seem (mostly) duplicated across the base settings and explicitly overridden settings.  this seems redundant and, in the former case, somewhat brittle.  My suggestion would be to favor the tests that explicitly override settings.

I would also suggest to create a comprehensive theme dir specifically for tests and put >1 theme in it, so that logic dealing with multiple themes can be tested more realistically.  Then you can point `COMPREHENSIVE_THEME_DIR` at this dir in settings/test.py and avoid the need for a `TEST_THEME` setting.

ecommerce/theming/tests/test_storage.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class TestThemeStorage(TestCase):
+    """
+    Test comprehensive theming static files storage.
+    """
+
+    def setUp(self):
+        super(TestThemeStorage, self).setUp()
+        self.themes_dir = get_base_themes_dir()
```
jimabramson: since this function is not under test I would suggest using `settings.COMPREHENSIVE_THEME_DIR` directly here instead.

ecommerce/theming/tests/test_theme_style_overrides.py
```diff
@@ -0,0 +1,95 @@
...
+    def setUp(self):
+        """
+        Clear static file finders cache and register cleanup methods.
+        """
+        super(TestComprehensiveTheme, self).setUp()
+
+        # Clear the internal staticfiles caches, to get test isolation.
+        staticfiles.finders.get_finder.cache_clear()
+
+        # create a user and log him in
```
jimabramson: s/him// :-)

ecommerce/theming/tests/test_theme_style_overrides.py
```diff
@@ -0,0 +1,95 @@
...
+        self.client.login(username=self.user.username, password=self.password)
+
+    @classmethod
+    def setUpClass(cls):
+        """
+        Enable Comprehensive theme and compile sass files.
+        """
+        # Apply Comprehensive theme and compile sass assets for test theme.
+        with override_settings(COMPREHENSIVE_THEME_DIR=TEST_THEME.dirname()):
+            compile_sass()
```
jimabramson: is this going to add a bunch of time to python test runs?

ecommerce/theming/tests/test_theme_style_overrides.py
```diff
@@ -0,0 +1,95 @@
...
+def compile_sass():
+    """
+    Call update assets command to compile system and theme sass.
+    """
+    with override_settings(DEBUG=True):
+        # Compile system and theme sass files
+        call_command('update_assets')
+
+
+def compress_templates():
```
jimabramson: this does not appear to be used.

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
+Comprehensive Theming
+=====================
+
+Any application, including Otto, can be loosely divided into two parts:
+
+- the user interface ("how it looks"),
+- and the application logic ("how it works").
+
+For example, when considering a product in otto,
```
jimabramson: capitalize Otto

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+---------------
+Theme Structure
+---------------
+From a technical perspective, theming consists of overriding core
+
+* Templates,
+* Static Assets and
+* Sass
+
+with themed templates, themed assets and themed sass. Every theme must conform to a directory structure,
```
jimabramson: s/templates, themed assets and themed sass/versions of those resources/
s/structure, this/structure. This/

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+this structure is the replica of Otto directory structure and looks like this
```
jimabramson: s/is the replica of/mirrors the/

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+        └── templates
+                └── oscar
+                |     └── dashboard
+                |             └── index.html
+                └── 404.html
+
+-----------
+- Templates
+-----------
+Any template included in ``ecommerce/templates`` directory is themable. However, Make sure not to override
```
jimabramson: s/Make/make/

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+class names or id values of html elements inside a template as these are used by javascript and/or css and overriding
```
jimabramson: s/template/template,/

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+------
+- Sass
+------
+Sass overrides are a little different from static asset or template overrides.
+There are two types of styles included in ``ecommerce/static/sass``:
+
+- `base`
+- `partials`
+
+Styles present ``ecommerce/static/sass/base`` (, while they can be, ) should not be overridden as overriding these
```
jimabramson: i would suggest removing the part in parentheses, and perhaps prepending 'Warning:' to this entire sentence.

jimabramson: s/present/present in/

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+could result in an unexpected behavior.
+
+Any styles included in ``ecommerce/static/sass/partials`` can be overridden.
+Style included in this directory contain variable definitions that are used by main sass files and is the best place
```
jimabramson: s/Style/Styles/
s/files and/files. This is/

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+if you want to update things like header/footer background, fonts etc.
+
+----------------
+Enabling a Theme
+----------------
+To enable a theme, you must first install your theme onto the same server that is running Otto.
+If you are using devstack or fullstack to run Otto, you must be sure that the theme is present on the Vagrant virtual machine.
+It is up to you where to install the theme on the server, but a good default location is ``/edx/app/ecommerce/ecommerce/themes``.
```
jimabramson, line 100: I would mention here that all themes must reside in the same physical directory.

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Once the theme is installed on the server, you can activate it by setting the ``COMPREHENSIVE_THEME_DIR`` variable
+to the full path of your theme's directory. You can do that by adding the following value to your settings file:
```
jimabramson: In order for Otto to use the installed themes, you must specify the location of the theme directory in Django settings by setting `COMPREHENSIVE_THEME_DIR` in your settings file:

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+
+.. code-block:: yaml
+
+    COMPREHENSIVE_THEME_DIR: "/edx/app/ecommerce/ecommerce/themes"
+
+Where ``/edx/app/ecommerce/ecommerce/themes`` is the path to where you have installed the
+themes on your server.
+
+After installing a theme, it must be enabled by adding appropriate entries to the following tables
```
jimabramson: s/must be enabled/is associated with sites/

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+--------------------
+Compiling Theme Sass
+--------------------
+Management command ``update_assets`` can be used for compiling and collecting themed sass.
+
+.. code-block:: yaml
+
+    python manage.py update_assets
+
+``update_assets`` accespts following optional arguments
```
jimabramson: s/accespts following/accepts the following/

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+
+    :--settings: settings file to use, ``default: ecommerce.settings.devstack``
+
+        .. code-block:: yaml
+
+            python manage.py update_assets --settings=ecommerce.settings.production
+
+    :--themes: Space separated list of themes to compile sass for. 'all' for all themes,
+        'no' to skip sass compilation for themes,  ``default: 'all'``
```
jimabramson, line 151: what is an example situation where I would want to skip sass compilation for themes, or only specify individual ones (instead of all of them)?

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+            python manage.py update_assets --themes my-theme
+
+    :--output-style: Coding style for compiled css files. Possible options are ``nested``, ``expanded``,
+        ``compact`` and ``compressed``. ``default: 'nested'``
+
+        .. code-block:: yaml
+
+            python manage.py update_assets --output-style='compressed'
+
+    :--skip-system: This flag disables system sass compilation.
```
jimabramson, line 173: when might someone use this?

docs/theming.rst
```diff
@@ -0,0 +1,178 @@
...
+
+            python manage.py update_assets --skip-system
+
+    :--enable-source-comments: This flag enables source comments in generated css files
+
+        .. code-block:: yaml
+
+            python manage.py update_assets --enable-source-comments
+
+    :--skip-collect: This flag can be used to skip collectstatic call after sass compilation
```
jimabramson, line 186: when might someone use this?

ecommerce/theming/finders.py
```diff
@@ -0,0 +1,96 @@
...
+            if storage.exists(''):  # check if storage location exists
+                for path in utils.get_files(storage, ignore_patterns):
+                    yield path, storage
+
+    def find(self, path, all=False):  # pylint: disable=redefined-builtin
+        """
+        Looks for files in the theme directories.
+        """
+        matches = []
+        theme_dir = path.split("/", 1)[0]
```
jimabramson, line 72: perhaps an edge case, but if i have a subdirectory that has the same name as one of the known themes, this function will get confused, right?   If so, might be worth adding a warning to the docs.

ecommerce/settings/base.py
```diff
@@ -111,12 +111,32 @@
...
+# ThemeFilesFinder looks for static assets inside theme directories. It creates separate storage for each theme.
+# During collectstatic run it utilizes all of these storages to fetch static assets for each theme.
+# Since, storage saves each theme asset in its own sub directory (i.e. {STATIC_ROOT}/red-theme/images/default-logo.png)
+# and also creates asset url with theme name prefixed (i.e. /static/red-theme/images/default-logo.png),
+# during post-processing and development mode, finder will know which theme asset is
+# being accessed (judging from the prefix), and will use the corresponding storage to access the file.
+# If prefix is not a theme name (we know all theme names via get_themes helper),
+# ThemeFilesFinder will know that a system asset is being accessed.
```
jimabramson: should this documentation really go with the code itself (in the theming package?)

ecommerce/settings/base.py
```diff
@@ -111,12 +111,32 @@
...
+# ThemeStorage stores and retrieves files with theming in mind, during collectstatic run it will take
+# all themed assets and place theme in theme subdirectory (e.g. red-theme) inside static directory. All system assets
+# are saved normally. e.g. 'red-theme' assets will be stored in '{STATIC_ROOT}/red-theme/' and system assets
+# will be stored in '{STATIC_ROOT}/'.
+#
+# While serving static assets it prefixes asset url with theme name (e.g. 'red-theme'),
+# only if assets is overridden by theme. e.g. if red-theme provides 'images/default-logo.png' then corresponding url
+# will be "/static/red-theme/images/default-logo.png" and if red-theme does not provide 'images/default-logo.png'
+# then corresponding url will be "/static/images/default-logo.png"
+STATICFILES_STORAGE = "ecommerce.theming.storage.ThemeStorage"
```
jimabramson, line 125: ditto

jimabramson:
is it considered suitably documented how to run Otto without making any use of theming?  Presumably this is already something that just works?

ecommerce/theming/tests/test_core.py
jimabramson, line 4: the above nitpick applies to a number of the python modules in this PR; i'll stop calling them out individually

jimabramson:
just to confirm, can we assume this PR is compatible with compressor, and/or any other staticfiles implementations that use manifests/caching etc?

ecommerce/static/sass/base/main.scss
jimabramson, line 3: @andy-armstrong @AlasdairSwan are you comfortable with this approach?

ecommerce/tests/test-theme/static/sass/partials/utilities/_variables.scss
```diff
@@ -0,0 +1,38 @@
+// config: variables
+// ====================
+// NOTE: these variables override Bootstrap-centric variables.
```
jimabramson: assuming this needs to stay largely in sync with the base _variables, can a comment be added somewhere to make maintainers aware of this?

alternatively, if this is never rendered and only used in unit tests, perhaps the amount of content in here can be reduced to the minimum of what is needed in those tests?

ecommerce/tests/test-theme/templates/oscar/dashboard/index.html
```diff
@@ -0,0 +1,192 @@
+{% extends 'dashboard/layout.html' %}
```
jimabramson: same comment here as for the _variables module.

ecommerce/themes/red-theme/static/sass/partials/utilities/_variables.scss
```diff
@@ -0,0 +1,38 @@
+// config: variables
```
jimabramson: we should try and prevent this theme from drifting out of sync with the base - a comment in the base module might be best to call this to the attention of maintainers.

ecommerce/themes/red-theme/templates/bootstrap-demo.html
```diff
@@ -0,0 +1,657 @@
+{% extends "edx/base.html" %}
```
jimabramson, line 1: same comment here as above

ecommerce/themes/red-theme/templates/oscar/dashboard/index.html
```diff
@@ -0,0 +1,193 @@
+Testing Red Theme!
```
jimabramson: same

jimabramson:
@saleem-latif I am done with my first pass of the review.

ecommerce/theming/template_loaders.py
saleem-latif, line 20: @jimabramson return type `Origin`  is added in django 1.9. [Ref django 1.8](https://github.com/django/django/blob/stable/1.8.x/django/template/loaders/filesystem.py#L27)

ecommerce/theming/template_loaders.py
saleem-latif, line 27: @jimabramson `get_dirs` is also added in django 1.9 and is not present in [django 1.8](https://github.com/django/django/blob/stable/1.8.x/django/template/loaders/filesystem.py#L15-L47)

ecommerce/theming/helpers.py
saleem-latif, line 162: Done

docs/theming.rst
saleem-latif, line 100: @jimabramson I have added a Note saying this in this section.

docs/theming.rst
saleem-latif, line 151: @jimabramson I have added comments with examples to state some use cases. However, some of the use cases are related to development mode e.g. compiling sass for only one theme would be useful if a developer adds a new theme and wants to test it without changing other themes.
kindly take a look at and let me know if updates are required.

ecommerce/theming/finders.py
saleem-latif, line 72: @jimabramson  Yes, that is an edge case, I have added a Caution note in the docs stating this.

ecommerce/settings/base.py
saleem-latif: @jimabramson I have also added this info in theming app [here](https://github.com/edx/ecommerce/pull/649/files#diff-260c94995d0994e9cef0104ab947bac1R11). I added it here to explain the purpose of `ThemeFilesFinder`. 
If it does not seem good, I can remove details and add only a small comment related ThemeFinders?

saleem-latif:
@jimabramson 

> is it considered suitably documented how to run Otto without making any use of theming? Presumably this is already something that just works?

`ENABLE_COMPREHENSIVE_THEMING` setting can be set to `False` to disable theming, corresponding section of docs is [here](https://github.com/edx/ecommerce/pull/649/files#diff-d32c1b5d30306b925f775db5c5c231faR23).

> just to confirm, can we assume this PR is compatible with compressor, and/or any other staticfiles implementations that use manifests/caching etc?

I have tested overridden templates (compression), static assets (caching) and sass overrides locally for multiple themes simultaneously. But, I think it would be a good idea for someone, with more info on Otto internals, to  test it.

ecommerce/theming/template_loaders.py
jimabramson, line 20: ah, understood, thanks

ecommerce/theming/template_loaders.py
jimabramson, line 27: got it.

ecommerce/tests/themes/test-theme-2/static/sass/base/_base.scss
```diff
@@ -0,0 +1,3 @@
+@import "ecommerce/static/sass/base/base";
+
+// This file is here for testing only
```
jimabramson: nit: newline

jimabramson:
@saleem-latif thanks for taking my suggestion regarding /tests/themes/ and reducing the amount of content there.  Question about the latest changes to red-theme, though - is that still useful with the templates gutted?  My assumption (which may be wrong) was that red-theme is for user-facing demonstration of the theming capability, so I just want to make sure it can still serve that purpose (e.g. with Oscar dashboard template mostly empty now).

jimabramson:
@mattdrayer @saleem-latif I took a quick look at parts of the edx-platform implementation for multi-site theming, and noticed some stuff I don't see in this PR.  For example the cached files and theme pipeline mixins defined [here](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/theming/storage.py#L106).

Do we think those will not be needed in Otto?  The cached files stuff seems like it may be necessary to work with offline compression and asset versioning, a concern I mentioned earlier.

EDIT: after looking at this with @mattdrayer we agree it is important to ensure that this approach is compatible with [offline compression](https://django-compressor.readthedocs.org/en/latest/usage/#pre-compression), which is used to pre-compile js and css in production environments.

In addition, while Otto does not presently use a caching storage backend for static images, it may in the future, and we may have compatibility issues there too.  (For example In the credentials IDA we make use of the [ManifestStaticFilesStorage](https://github.com/edx/configuration/blob/master/playbooks/roles/credentials/defaults/main.yml#L124) implementation which is useful for adding hashes to filenames ensuring they can be safely cached forever on CDNs, but it's not clear whether we could use that in concert with this theming infrastructure.)

In terms of the needs of the present PR I'd say the offline compression compatibility is a blocker, while the image issues may or may not be; as long as the cache-busting query strings currently being appended by Django to static images continue to be applied to base / themed assets then we could leave that for a separate change.     cc: @clintonb

ecommerce/settings/base.py
jimabramson: My major concern is to ensure the same detailed documentation isn't duplicated in two places, because sooner or later they will get out of sync.  If the doc content is different enough that it makes sense to have in both places and maintain separately, then I have no objections.

ecommerce/theming/tests/test_storage.py
```diff
@@ -0,0 +1,88 @@
...
+    Test comprehensive theming static files storage.
+    """
+
+    def setUp(self):
+        super(TestThemeStorage, self).setUp()
+        self.themes_dir = Path(settings.COMPREHENSIVE_THEME_DIR)
+        self.enabled_theme = "test-theme"
+        self.storage = ThemeStorage(location=self.themes_dir / self.enabled_theme / 'static')
+
+    @override_settings(DEBUG=True)
```
jimabramson: I think you can put this directive at the top of the class definition, instead of above each test method individually, and it will do the same thing.

ecommerce/theming/helpers.py
```diff
@@ -0,0 +1,281 @@
...
+    """
+    Returns boolean indicating whether comprehensive theming functionality is enabled or disabled.
+    Example:
+        >> is_comprehensive_theming_enabled()
+        True
+
+    Returns:
+         (bool): True if comprehensive theming is enabled else False
+    """
+    return settings.ENABLE_COMPREHENSIVE_THEMING
```
jimabramson: @saleem-latif - @mattdrayer and I discussed some possibilities for a feature switch allowing for a dark launch of this code (such that theming would be ignored at runtime, but the mgmt commands would still work so that assets could be compiled and ready to go).

Not sure if it's the optimal approach, but this seems like one possible place to do this - you could check an additional boolean whose meaning was "disable theming at runtime", and if it is set _and_ `get_current_request` was not `None`, return `False` from this function.

For a way to set the toggle at runtime, [waffle](http://waffle.readthedocs.org/en/latest/index.html) is already installed and used elsewhere in this code - searching the repo "waffle" should point you to a number of examples of its usage.

jimabramson:
@saleem-latif it seems like what the last change does is to gracefully fail offline compression for themed assets rather than pull them along into the manifest - is that correct?  If so won't this have a significant performance impact when themes are in use, since they will continue to be compressed on the fly?

saleem-latif:
@jimabramson Yes, you are right, I tried multiple approaches for django-compressor work with themed templates, but I was not able to achieve that. This is the only solution I could find and I thought it could be a temporary solution until we figure out and resolve django-compressor problems. We might even move to some other compression technique while implementing stand-alone theming app.

saleem-latif:
@jimabramson After reviewing @clintonb 's suggestion [here](https://openedx.atlassian.net/browse/WL-330?focusedCommentId=183132&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-183132), I was able to implement a way to compress all themed templates. 
I have removed the commit of "gracefully fail offline compression" (so discard my [previous comment](https://github.com/edx/ecommerce/pull/649#issuecomment-211212837)).
I have squashed the commits and rebased the branch with master. [This](https://github.com/edx/ecommerce/blob/saleem-latif/WL-330/ecommerce/theming/compressor.py#L1-L26), [this](https://github.com/edx/ecommerce/blob/saleem-latif/WL-330/ecommerce/templates/edx/base.html#L15-L21) and [this](https://github.com/edx/ecommerce/blob/saleem-latif/WL-330/ecommerce/templates/rest_framework_swagger/index.html#L12-L24) is the change set to resolves compression problem. This approach also requires to use [`django-compressor=2.0`](https://github.com/edx/ecommerce/blob/saleem-latif/WL-330/requirements/base.txt#L4), please let me know if the version can be updated like this, or their would be other changes required in the system to make version update.

jimabramson:
@saleem-latif nice!  this solution looks much cleaner and more complete (thanks for the direct links too).

I'm 👍  with this, assuming it's been manually run & validated, and pending confirmation from @andy-armstrong or @AlasdairSwan on the outstanding question above about sass paths.

ecommerce/static/sass/base/main.scss
andy-armstrong, line 3: @saleem-latif @jimabramson This approach makes sense to me.

AlasdairSwan:
Looks good @saleem-latif! Once rebased with master 👍

jibsheet:
@mattdrayer this seems to have caused build failures on the CI sandbox.
Are there configuration changes or other theme related changes that should have been made?
http://jenkins.edx.org:8080/job/ansible-provision-continuous-integration/26518/consoleFull

PR #650, opened 2016-04-01
Turn off startup validation for core app because it is breaking migrations

douglashall:
Sandbox provisioning is broken due to the startup validation introduced in https://github.com/edx/ecommerce/pull/630.

http://jenkins.edx.org:8080/job/ansible-provision/6495/console

@e-kolpakov @clintonb @mjfrey @maxrothman

### DIFF
--- a/ecommerce/core/config.py
+++ b/ecommerce/core/config.py
@@ -27,6 +27,10 @@ class CoreAppConfig(AppConfig):
             # Here we're are consciously violating Django's warning about not interacting with DB in AppConfig.ready
             # We know what we're doing, have considered a couple of other approaches and discussed it in great length:
             # https://github.com/edx/ecommerce/pull/630#discussion-diff-58026881
-            validate_configuration()
+            # TODO: This was causing an issue with running migrations from scratch
+            # on a sandbox. We need to fix this before uncommenting.
+            # http://jenkins.edx.org:8080/job/ansible-provision/6495/console
+            # validate_configuration()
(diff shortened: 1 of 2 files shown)


douglashall:
@bradenmacdonald

ecommerce/core/config.py
```diff
@@ -27,6 +27,10 @@ def ready(self):
             # Here we're are consciously violating Django's warning about not interacting with DB in AppConfig.ready
             # We know what we're doing, have considered a couple of other approaches and discussed it in great length:
             # https://github.com/edx/ecommerce/pull/630#discussion-diff-58026881
-            validate_configuration()
+            # TODO: This was causing an issue with running migrations from scratch
+            # on a sandbox. We need to fix this before uncommenting.
+            # http://jenkins.edx.org:8080/job/ansible-provision/6495/console
+            # validate_configuration()
+            pass
         except OperationalError:
```
clintonb, line 35: Based on the logs, the solution here is to catch `ProgrammingError`.

douglashall, line 35: True. Still I get a ValidationError when running migrations on my existing devstack.

I think we need to further evaluate the side effects of this validation before reintroducing it.

douglashall:
@clintonb Look good?

clintonb:
I'd much rather us just delete the malfunctioning code than comment it out. We have the git history for when it's time to fix this.

PR #651, opened 2016-04-01
Revert "Multitenant payment processor config"

douglashall:
Reverts edx/ecommerce#630

The core app startup validation which this changeset introduced was causing migrations to fail on sandboxes.

http://jenkins.edx.org:8080/job/ansible-provision/6495/console

We need to deploy these changes to a sandbox and test before reintroducing.

@e-kolpakov @clintonb @mjfrey @maxrothman


douglashall:
@bradenmacdonald

mjfrey:
:+1:

clintonb:
This is the second time we will have reverted a fairly large change.
1. What are our options for fixing this rather than starting over completely?
2. How do we avoid this in the future?

douglashall:
@clintonb Just reverting the startup validation is the other option.

https://github.com/edx/ecommerce/pull/650

We should have asked for a sandbox with this changeset deployed to it the first time around. That said, I don't see it as the end of the world that we have a need to revert a changeset. Happy to chat about it further offline.

peter-fogg:
@douglashall I'm going to go ahead and merge this at @maxrothman's request.

e-kolpakov:
@douglashall @clintonb It was deployed to the sandboxes listed in #630. I have tried running migrate, syncdb, and a couple of other management commands, but with the existing DB.

I believe a better approach would be to split startup validation into a dedicated PR - this way the body of the work would stay, but that option didn't come into minds of anyone involved with #630, unfortunately.

e-kolpakov:
@douglashall I can't access http://jenkins.edx.org:8080/job/ansible-provision/6495/console - can I have the relevant logs copied somewhere (or emailed to me), so that I could understand the problem and address it?

maxrothman:
Looping @fredsmith in because he's now on-call. Is this issue resolved?

clintonb:
@peter-fogg @maxrothman @fredsmith In the future let's actually _discuss_ these issues before reverting. A lot of time was invested reviewing these changes, and we now have to review again. The ultimate fix for this was a one line change.

PR #656, opened 2016-04-04
[SOL-1671] Mark the Client field as required on Coupon creation form

marjev:
@mjfrey This is a PR for [SOL-1671](https://openedx.atlassian.net/browse/SOL-1671). Please take a look.

Here's the screenshot that shows the change:
screen shot 2016-04-04 at 14 24 11


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=656&repo=edx%2Fecommerce

mjfrey:
:+1:

PR #657, opened 2016-04-05
Expanded Product API filtering

clintonb:
- Filtering on product class properly supports examining the product classes of children's parents.
- Added support for filtering by structure.

ECOM-3981


clintonb:
FYI @edx/ecommerce

PR #658, opened 2016-04-05
Fixing default value of SiteConfiguration.oauth_settings to match migration

douglashall:
This is a followup PR to https://github.com/edx/ecommerce/pull/641 which fixed the default value of the SiteConfiguration.oauth_settings field in migrations, but not in the actual model. This PR remedies that condition.

@mjfrey @clintonb

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -49,7 +49,7 @@ class SiteConfiguration(models.Model):
         help_text=_('JSON string containing OAuth backend settings.'),
         null=False,
         blank=False,
-        default='{}'
+        default={}
     )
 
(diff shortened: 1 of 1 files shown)


douglashall:
@mjfrey @clintonb Bump on this one line patch. Please review when you get a chance.

clintonb:
This requires a migration.

clintonb:
:+1: once migration added.

douglashall:
@clintonb There is a migration which already covers this change. This PR corrects a previous miguided PR which modified a migration but did not modify the model. See https://github.com/edx/ecommerce/pull/641. Thanks.

PR #660, opened 2016-04-06
(WIP) mattdrayer/WL-381-coupon-app: Move coupon creation to djangoapp

mattdrayer:
@mjfrey -- let's use this branch instead -- it has only the coupon app work

This PR adds a new "api.py" Python integration module to the Coupons djangoapp, and moves some work being conducted in the Coupons API view that I would like to reuse in the Course model's creation workflow.  Rather than adding an upwards dependency from the model to the api, it seems better to move the logic out of the view to a neutral location.

I've also made some modifications to the coupons test harness.  I've avoided altering existing test assertions in order to preserve the semantics of the system, however I did change one assert statement in ecommerce/extensions/api/v2/tests/views/test_products.py that I felt was acceptable.

This change set also introduces the concept of including the creation of a 'bulk purchase' product during seat creation, although it is not currently wired-up to anything and still needs targeted test coverage.


mattdrayer:
@clintonb FYI on this PR as well

clintonb:
@mattdrayer Please file an ECOM ticket, if you haven't already. Our sprint planning is today.

mattdrayer:
Done -- I filed a ticket for WL-330 as well

ecommerce/courses/models.py
```diff
@@ -227,3 +227,13 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
             ).delete()
 
         return seat
+
```
mattdrayer: these changes can be removed from this PR

ecommerce/extensions/api/serializers.py
```diff
@@ -301,6 +301,10 @@ def save(self):
                         credit_hours=credit_hours,
                     )
 
+                    # In addition to the seat product, we also need to create an enrollment code product
+                    # for each new course.  This enables bulk purchasing of course seats via the LMS service.
+                    # course.create_or_update_enrollment_code()
```
mattdrayer: These changes can be removed from this PR

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -148,59 +145,7 @@ def create_coupon_product(self, title, price, data):
...
-            product=coupon_product,
-            partner=data['partner'],
-            partner_sku=sku
-        )
-        stock_record.price_currency = 'USD'
-        stock_record.price_excl_tax = price
-        stock_record.save()
-
-        return coupon_product
+        return coupons_api.create_coupon_product(title, price, data)
```
mjfrey: Why not just remove this method all together and call the above line in create.

mattdrayer:
Closed this in favor of #677 -- I ran into too many issues trying to get this moved code to work (especially with the tests) and we need to focus on shipping something usable.

ecommerce/coupons/api.py
```diff
@@ -0,0 +1,119 @@
...
+    Returns:
+        A coupon product object.
+
+    Raises:
+        IntegrityError: An error occured when create_vouchers method returns
+                        an IntegrityError exception
+    """
+    # Use the provided information to locate (or create) the catalog
+    stock_records_string = ' '
+    if stock_record_ids:
```
vkaracic, line 57: Will there every not be stock records?

ecommerce/coupons/tests/test_api.py
```diff
@@ -0,0 +1,46 @@
...
+class CouponApiTest(CouponMixin, TestCase):
+    """
+    Unit tests covering integration scenarios for the Coupons app
+    """
+
+    def setUp(self):
+        super(CouponApiTest, self).setUp()
+        self.product_class, __ = ProductClass.objects.get_or_create(name='Coupon')
+        self.coupon = self.create_coupon(
+            voucher_type=Voucher.MULTI_USE,
```
vkaracic, line 29: A small heads-up, `MULTI-USE` voucher type has been removed for the time being:
https://github.com/edx/ecommerce/pull/646

ecommerce/tests/mixins.py
```diff
@@ -340,65 +344,79 @@ def coupon_product_class(self):
...
+            course=None
     ):
         """Helper method for creating a coupon.
 
         Arguments:
             title(str): Title of the coupon
             price(int): Price of the coupon
             partner(Partner): Partner used for creating a catalog
-            catalog(Catalog): Catalog of courses for which the coupon applies
+            catalog(Catalog): Catalog of seat products for which the coupon applies
```
vkaracic: This is not true anymore. According to the parameter list above this is now the name of a catalog / string.

PR #654, opened 2016-04-04
Multitenant payment processor configuration

e-kolpakov:
**Description:** This PR reintroduces changes made in #630, but without startup validation that caused [some trouble](https://openedx.atlassian.net/browse/WL-390)
**Merge deadline:** 8 Apr
**Related:** [Discovery document](https://docs.google.com/document/d/17BRYvkO4ETvBV9TFodElpbJaK4oze7e0mNqf9ztnGnA/edit#heading=h.70eeasm5mr1q)
**Sandbox:** 
- edX site: [LMS](http://pr11898.sandbox.opencraft.com), [Studio](http://studio.pr11898.sandbox.opencraft.com), [eCommerce](https://ecommerce.pr11898.sandbox.opencraft.com)
- WL site: [LMS](http://wlsite.pr11898.sandbox.opencraft.com), [eCommerce](https://ecommerce.wlsite.pr11898.sandbox.opencraft.com)

**JIRA:** [WL-362](https://openedx.atlassian.net/browse/WL-362), [WL-390](https://openedx.atlassian.net/browse/WL-390)

**Testing instructions:** See #630

PR #663, opened 2016-04-06
Fix payment processor settings overrides

douglashall:
We missed this in https://github.com/edx/ecommerce/pull/654. Payment processor path settings need to be overridden in production settings.

@e-kolpakov @mjfrey @clintonb

### DIFF
--- a/ecommerce/settings/production.py
+++ b/ecommerce/settings/production.py
@@ -61,10 +61,11 @@ for override, value in DB_OVERRIDES.iteritems():
 
 
 # PAYMENT PROCESSOR OVERRIDES
-for __, config in PAYMENT_PROCESSOR_CONFIG.iteritems():
-    config.update({
-        'receipt_path': PAYMENT_PROCESSOR_RECEIPT_PATH,
-        'cancel_path': PAYMENT_PROCESSOR_CANCEL_PATH,
-        'error_path': PAYMENT_PROCESSOR_ERROR_PATH,
-    })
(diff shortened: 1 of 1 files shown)


douglashall:
FYI @bradenmacdonald

bradenmacdonald:
Thanks for fixing this @douglashall.

In retrospect we probably should have tried something more backwards-compatible, like make the new setting use a different name and add a fallback to the old setting.

e-kolpakov:
@douglashall :+1: I wonder why I haven't seen the error in sandbox; maybe that's because sandbox have `*_path` instead of `*_url` in `ecommerce.yml`?

douglashall:
@e-kolpakov No worries. I should have caught this during my review. As far as sandbox testing goes, @mjfrey raised this issue because he was seeing it in his newly deployed sandbox. Is it possible that you didn't perform a manual test of the checkout workflow on your sandbox?

e-kolpakov:
@douglashall if by "checkout workflow" you mean creating sandbox from scratch - than yes, I haven't done that, as configuring sandbox from fresh install takes some 2-4 hours (creating DB records, sharing keys with LMS, etc.). Except that initially I created a sandbox from earlier version of the #630.

clintonb:
@e-kolpakov @douglashall if creating a sandbox takes 2-4 hours, please create management commands and update configuration to make it a 1-click process.

mjfrey:
I just created a new sandbox from this branch and it fixes the issue.
:+1:

PR #665, opened 2016-04-11
Capture acceptance test results in an xunit.xml file.

benpatterson:
This will allow test results to be more easily consumed by:
- CI infrastructure
- A developer who wants to understand results
- Logging infrastructure we use wherein we can roll together test results across repos

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ fast_diff_coverage:
 	diff-cover coverage.xml --compare-branch=$(DIFF_COVER_BASE_BRANCH)
 
 accept:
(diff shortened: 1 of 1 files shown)


benpatterson:
@rlucioni @schenedx 

I'd like to capture test results from this repo in an xunit file, which we can then feed into Jenkins output for usability, trending, etc.

rlucioni:
👍

schenedx:
:+1: @benpatterson Are you going to make the same change for Programs IDA at https://github.com/edx/programs/blob/master/Makefile? Thanks

benpatterson:
@schenedx Yep...that's my plan!

PR #671, opened 2016-04-14
Pin python-social-auth

mjfrey:
Pin python-social-auth==0.2.14


mjfrey:
@clintonb @rlucioni

rlucioni:
Could you please add a comment linking to https://github.com/omab/python-social-auth/pull/897? It should be possible to remove the pin once the changes in that PR are released.

mjfrey:
@rlucioni comment added.

clintonb:
:+1:

rlucioni:
👍

PR #673, opened 2016-04-15
Added edx-drf-extensions

clintonb:
Replaced local version of IsSuperUser with version from package.


clintonb:
FYI @edx/ecommerce

rlucioni:
👍

PR #674, opened 2016-04-15
(WIP) WL-382: Quantity field added to basket

malikshahzad228:


malikshahzad228:
@mattdrayer FYI

vkaracic:
Have you looked how Oscar is handling quantities? You can checkout out their demo site: http://latest.oscarcommerce.com/ - try adding a couple of items to your basket and view the basket, a quantity field should appear. This is the relevant template code: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/basket/partials/basket_content.html#L62
If you could, please try and use Oscar's native implementation of basket item quantities.

PR #675, opened 2016-04-15
[SOL-1741 Fix for the Coupon Invoice Django Admin Page]

marjev:
@mjfrey Here's a PR that fixes the broken Coupon Invoice Django Admin page. Please review.

JIRA ticket: [SOL-1741](https://openedx.atlassian.net/browse/SOL-1741)

Here's the screenshot that shows how it looks now:
screencapture-localhost-8002-admin-invoice-invoice-1460723626570

### DIFF
--- a/ecommerce/invoice/admin.py
+++ b/ecommerce/invoice/admin.py
@@ -6,4 +6,6 @@ from ecommerce.invoice.models import Invoice
 
 @admin.register(Invoice)
 class InvoiceAdmin(SimpleHistoryAdmin):
-    pass
+    list_display = ('id', 'basket', 'order', 'business_client', 'state',)
+    list_filter = ('state',)
+    search_fields = ['order__number', 'business_client__name']
--- a/ecommerce/invoice/models.py
+++ b/ecommerce/invoice/models.py
@@ -17,9 +17,6 @@ class Invoice(TimeStampedModel):
 
     history = HistoricalRecords()
 
-    def __str__(self):
-        return 'Invoice {id} for order number {order}'.format(id=self.id, order=self.order.number)
-
     @property
     def total(self):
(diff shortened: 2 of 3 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=675

clintonb:
How about this?
1. Remove the order number from the string method. 
2. Create a proper admin with sortable/searchable list display: https://docs.djangoproject.com/en/1.8/ref/contrib/admin/.

As you develop new models, consider making the admin the third thing you do after creating the model and migrations. The admin will be the first place we look when debugging, so it should be fully functional.

marjev:
@clintonb, @mjfrey Below is the new look of the Invoice Django Admin Page. Please take a look at the screenshot below and the code changes. Thanks!

screencapture-localhost-8002-admin-invoice-invoice-1460729638276

ecommerce/invoice/admin.py
```diff
@@ -6,4 +6,5 @@
 
 @admin.register(Invoice)
 class InvoiceAdmin(SimpleHistoryAdmin):
-    pass
+    list_display = ('id', 'basket', 'order', 'business_client', 'state',)
+    search_fields = ['id', 'order__number', 'business_client__name', 'state']
```
clintonb: Don't search on state or id. If you have the ID, you can just change the URL, and include the ID. For state, you should use [`list_filter`](https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_filter).

marjev: Done.

mjfrey:
:+1:

PR #677, opened 2016-04-18
(WIP) mattdrayer/WL-381: Bulk seat purchase coupons

mattdrayer:
Still a WIP, but getting closer to an actual implementation here -- see WL-381 and #660 for some backstory/context.  

@mjfrey @vkaracic, FYI

### DIFF
--- a/ecommerce/coupons/api.py
+++ b/ecommerce/coupons/api.py
@@ -45,16 +45,15 @@ def create_coupon_product(title, price, data):
                         an IntegrityError exception
     """
     coupon_slug = generate_coupon_slug(title=title, catalog=data['catalog'], partner=data['partner'])
-    defaults = {'requires_shipping': False, 'track_stock': False, 'name': 'Coupon'}
-    product_class, __ = ProductClass.objects.get_or_create(slug='coupon', defaults=defaults)
+    pc_defaults = {'requires_shipping': False, 'track_stock': False, 'name': 'Coupon'}
+    product_class, __ = ProductClass.objects.get_or_create(slug='coupon', defaults=pc_defaults)
     coupon_product, __ = Product.objects.get_or_create(
         title=title,
         product_class=product_class,
         slug=coupon_slug
     )
 
-    for category in data['categories']:
-        ProductCategory.objects.get_or_create(product=coupon_product, category=category)
+    assign_categories_to_coupon(coupon=coupon_product, categories=data['categories'])
 
     # Vouchers are created during order and not fulfillment like usual
     # because we want vouchers to be part of the line in the order.
@@ -100,3 +99,14 @@ def create_coupon_product(title, price, data):
     stock_record.save()
 
     return coupon_product
+
+
+def assign_categories_to_coupon(coupon, categories):
+    """
+    Assign categories to a coupon. If a category is already assigned, it will be fetch instead.
+    Arguments:
+        coupon (Product): Coupon product
+        categories (List): List of categories to be assigned to a coupon
+    """
+    for category in categories:
+        ProductCategory.objects.get_or_create(product=coupon, category=category)
--- a/ecommerce/coupons/api.py
+++ b/ecommerce/coupons/api.py
@@ -1,0 +1,102 @@
+from decimal import Decimal
+import logging
+
+from django.db.utils import IntegrityError
+
+from oscar.core.loading import get_model
+
+from ecommerce.extensions.catalogue.utils import generate_coupon_slug, generate_sku
+from ecommerce.extensions.voucher.models import CouponVouchers
+from ecommerce.extensions.voucher.utils import create_vouchers as utils_create_vouchers
+
+logger = logging.getLogger(__name__)
+
+Product = get_model('catalogue', 'Product')
+ProductCategory = get_model('catalogue', 'ProductCategory')
+ProductClass = get_model('catalogue', 'ProductClass')
+StockRecord = get_model('partner', 'StockRecord')
+
+
+def create_coupon_product(title, price, data):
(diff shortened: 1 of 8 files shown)


ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -253,7 +257,9 @@ def setUp(self):
             'quantity': 2,
             'price': 100,
             'category_ids': [self.category.id],
+            'create_vouchers': True,
         }
+        print COUPONS_LINK
```
vkaracic: :)

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -180,28 +182,29 @@ def create_coupon_product(self, title, price, data):
...
-                voucher_type=data['voucher_type'],
-                max_uses=data['max_uses'],
-                coupon_id=coupon_product.id
-            )
-        except IntegrityError as ex:
-            logger.exception('Failed to create vouchers for [%s] coupon.', coupon_product.title)
-            raise IntegrityError(ex)  # pylint: disable=nonstandard-exception
-
-        coupon_vouchers = CouponVouchers.objects.get(coupon=coupon_product)
+        if data.get('create_vouchers', False):
```
vkaracic: Not that this is false, but I'm curious is `create_vouchers` already False regarding the line number 90 above.

mattdrayer: It would already be false if this method is called from 'create' above, yes, but it's possible that this method will be used by others in the future so I think we still need to check for the parameter's existence and provide a reasonable default value?

mattdrayer:
@vkaracic, FYI I have pushed some updates to the branch -- please take a look when you have a moment, thanks!

ecommerce/courses/models.py
```diff
@@ -225,5 +232,58 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
...
+            Product:  The "Bulk Seat" coupon product that has been created or updated.
+        """
+        stock_records_string = ' '.join(str(id) for id in stock_record_ids)
+        coupon_catalog, __ = get_or_create_catalog(
+            name='Catalog for stock records: {}'.format(stock_records_string),
+            partner=partner,
+            stock_record_ids=stock_record_ids
+        )
+
+        category, __ = Category.objects.get_or_create(name='Coupons', defaults={'depth': 1, 'path': 'coupons'})
```
vkaracic, line 257: Oscar has a function for getting or creating categories from strings: http://django-oscar.readthedocs.org/en/releases-1.1/howto/how_to_create_categories.html so you could just say `create_from_breadcrumbs('Coupons')`
Also is the parameter `category` still needed in the list of parameters for this method?

ecommerce/courses/tests/test_models.py
```diff
@@ -256,3 +256,31 @@ def test_type(self):
 
         course.create_or_update_seat('credit', True, 1000, self.partner, credit_provider='SMU')
         self.assertEqual(course.type, 'credit')
+
+    def test_create_or_update_bulk_seat(self):
+        """ Verify the method creates or updates a seat Product. """
+        course = Course.objects.create(
```
vkaracic, line 262: You could use [ `CourseFactory` ](https://github.com/edx/ecommerce/blob/master/ecommerce/courses/tests/factories.py#L9) and just send the bulk purchase argument.

ecommerce/courses/tests/test_models.py
```diff
@@ -256,3 +256,31 @@ def test_type(self):
...
+            id='a/b/c',
+            name='Test Course',
+            bulk_purchasing_enabled=True
+        )
+
+        # Test seat creation
+        certificate_type = 'verified'
+        id_verification_required = True
+        price = 5
+        course.create_or_update_seat(certificate_type, id_verification_required, price, self.partner)
```
vkaracic, line 272: If you don't necessarily need to assert that two seats are created, you can just save the returned value here in a `seat` variable and use that in the assertion.

ecommerce/extensions/api/serializers.py
```diff
@@ -304,6 +304,9 @@ def save(self):
                         credit_hours=credit_hours,
                     )
 
+                if course.bulk_purchasing_enabled:
+                    course.create_or_update_bulk_seat_product(seat_product)
```
vkaracic, line 308: Aren't `partner, stock_record_ids, category, price` also needed as arguments?

ecommerce/coupons/api.py
```diff
@@ -0,0 +1,112 @@
...
+        A coupon product object.
+
+    Raises:
+        IntegrityError: An error occured when create_vouchers method returns
+                        an IntegrityError exception
+    """
+    coupon_slug = generate_coupon_slug(title=title, catalog=data['catalog'], partner=data['partner'])
+    pc_defaults = {'requires_shipping': False, 'track_stock': False, 'name': 'Coupon'}
+    product_class, __ = ProductClass.objects.get_or_create(slug='coupon', defaults=pc_defaults)
+    coupon_product, __ = Product.objects.get_or_create(
```
vkaracic, line 50: Are you sure you need to `get_or_create` these two?

ecommerce/courses/models.py
clintonb, line 257: Pause. 
1. The Coupons category already exists.
2. You should almost never be creating categories on the fly in this manner. Categories should come from migrations.

ecommerce/coupons/api.py
vkaracic, line 50: Okay you need it for `coupon_product` but I doubt you need it for `product_class`.

PR #680, opened 2016-04-19
[SOL-1673] Cancel button on the Edit Coupons form should lead to Details Page

marjev:
@mjfrey Here's the fix for [SOL-1673](https://openedx.atlassian.net/browse/SOL-1673). Please review.


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=680

mjfrey:
👍

PR #682, opened 2016-04-20
add segment-key to siteconfiguration

zubair-arbi:
ECOM-4097
@awais786 @ahsan-ul-haq @tasawernawaz @waheedahmed 
- Add new optional field `segment_key` for multi-tenant analytics tracking with [segement](https://segment.com/).
- Add migration

### DIFF
--- a/ecommerce/core/management/commands/create_or_update_site.py
+++ b/ecommerce/core/management/commands/create_or_update_site.py
@@ -17,65 +17,71 @@ class Command(BaseCommand):
     help = 'Create or update Site, Partner, and SiteConfiguration'
 
     def add_arguments(self, parser):
-        parser.add_argument('-i', '--site-id',
+        parser.add_argument('--site-id',
                             action='store',
                             dest='site_id',
                             type=int,
                             help='ID of the Site to update.')
-        parser.add_argument('-d', '--site-domain',
+        parser.add_argument('--site-domain',
                             action='store',
                             dest='site_domain',
                             type=str,
                             required=True,
                             help='Domain of the Site to create or update.')
-        parser.add_argument('-n', '--site-name',
+        parser.add_argument('--site-name',
                             action='store',
                             dest='site_name',
                             type=str,
                             default='',
                             help='Name of the Site to create or update.')
-        parser.add_argument('-c', '--partner-code',
+        parser.add_argument('--partner-code',
                             action='store',
                             dest='partner_code',
                             type=str,
                             required=True,
                             help='Partner code to select/create and associate with site.')
-        parser.add_argument('-p', '--partner-name',
+        parser.add_argument('--partner-name',
                             action='store',
                             dest='partner_name',
                             type=str,
                             default='',
                             help='Partner name to select/create and associate with site.')
-        parser.add_argument('-l', '--lms-url-root',
+        parser.add_argument('--lms-url-root',
                             action='store',
                             dest='lms_url_root',
                             type=str,
                             required=True,
                             help='Root URL of LMS (e.g. https://localhost:8000)')
-        parser.add_argument('-t', '--theme-scss-path',
+        parser.add_argument('--theme-scss-path',
                             action='store',
                             dest='theme_scss_path',
                             type=str,
                             default='',
                             help='Path to theme SCSS relative to STATIC_DIR (e.g. sass/themes/edx.scss)')
-        parser.add_argument('-m', '--payment-processors',
+        parser.add_argument('--payment-processors',
                             action='store',
                             dest='payment_processors',
                             type=str,
                             default='',
                             help='Comma-delimited list of payment processors (e.g. cybersource,paypal)')
-        parser.add_argument('-k', '--client-id',
+        parser.add_argument('--client-id',
                             action='store',
                             dest='client_id',
                             type=str,
                             required=True,
                             help='client ID')
-        parser.add_argument('-s', '--client-secret',
+        parser.add_argument('--client-secret',
                             action='store',
                             dest='client_secret',
                             type=str,
                             required=True,
                             help='client secret')
+        parser.add_argument('--segment-key',
+                            action='store',
+                            dest='segment_key',
+                            type=str,
+                            required=False,
+                            help='segment key')
 
     def handle(self, *args, **options):
         site_id = options.get('site_id')
@@ -86,6 +92,7 @@ class Command(BaseCommand):
         lms_url_root = options.get('lms_url_root')
         client_id = options.get('client_id')
         client_secret = options.get('client_secret')
+        segment_key = options.get('segment_key')
 
         try:
             site = Site.objects.get(id=site_id)
@@ -113,6 +120,7 @@ class Command(BaseCommand):
                 'lms_url_root': lms_url_root,
                 'theme_scss_path': options['theme_scss_path'],
                 'payment_processors': options['payment_processors'],
+                'segment_key': segment_key,
                 'oauth_settings': {
                     'SOCIAL_AUTH_EDX_OIDC_URL_ROOT': '{lms_url_root}/oauth2'.format(lms_url_root=lms_url_root),
                     'SOCIAL_AUTH_EDX_OIDC_KEY': client_id,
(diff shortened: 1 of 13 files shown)


tasawernawaz:
@clintonb Please review this PR.

ecommerce/core/management/commands/create_or_update_site.py
```diff
@@ -76,6 +76,12 @@ def add_arguments(self, parser):
                             type=str,
                             required=True,
                             help='client secret')
+        parser.add_argument('-e', '--segment-key',
```
clintonb: Please get rid of all of the short flags (e.g. `-e`). They barely relate to the full name, and are somewhat confusing.

ecommerce/core/models.py
```diff
@@ -51,6 +51,13 @@ class SiteConfiguration(models.Model):
         blank=False,
         default={}
     )
+    segment_key = models.CharField(
+        verbose_name=_('Segment key'),
+        help_text=_('Write key for segment source.'),
```
clintonb: `Segment write/API key`

ecommerce/extensions/analytics/config.py
```diff
@@ -11,5 +12,8 @@ def ready(self):
             from oscar.apps.analytics import receivers  # noqa pylint: disable=unused-variable
 
         # Initialize Segment
-        analytics.write_key = settings.SEGMENT_KEY
+        request = get_current_request()
```
clintonb: This will almost certainly always be `None` when the project is starting. A better solution may be to create a `Client` and associate it with each `Site`: https://segment.com/docs/libraries/python/#options.

clintonb:
Looks fine, aside from the outstanding comments.

ecommerce/extensions/analytics/config.py
tasawernawaz: @clintonb Can you please help on this, do I need to create a client here ? and how we can get site ?

clintonb: Create a cached property on the Site object, and remove this code.

tasawernawaz:
@clintonb I have updated code according to your suggestions, Please review it.

ecommerce/core/models.py
```diff
@@ -117,6 +125,10 @@ def clean_fields(self, exclude=None):
         if not exclude or 'payment_processors' not in exclude:
             self._clean_payment_processors()
 
+    @cached_property
+    def segment_key_value(self):
```
clintonb: This should be the Segment `Client`. There is no need to cache the key. What we want is a simple location to retrieve the Segment `Client` for each `Site`

``` python
from analytics import Client

class SiteConfiguration(models.Model):
  @cached_property
  def segment_client(self):
    return Client(self.segment_key, debug=settings.DEBUG)
```

Once you have this setup, all _Python_ calls to `analytics.*` should be updated to `{site}.*` where `{site}` is `Site` instance for the current request.

ecommerce/extensions/analytics/config.py
```diff
@@ -1,5 +1,6 @@
 import analytics
```
clintonb: This file will no longer be necessary once you implement `Site.segment_client`.

ecommerce/core/models.py
tasawernawaz: @clintonb Thanks. I will update this. There is already class `Client` defined here. Do you have idea why we have it?

ref: https://github.com/edx/ecommerce/blob/zub/ecom-4097-add-multitenant-segement-key/ecommerce/core/models.py#L155

clintonb: Ignore that poorly-documented model. It is not related. Use `from analytics import Client as SegmentClient`.

ecommerce/extensions/refund/tests/test_signals.py
```diff
@@ -8,7 +7,6 @@
 from ecommerce.tests.testcases import TestCase
 
 
-@override_settings(SEGMENT_KEY='dummy-key')
 @patch('analytics.track')
```
tasawernawaz: @clintonb I am not able to mock after using `cached_property`.

I have tried to get current request and try to mock with segment_client but it gives error that no module name `request`.

clintonb: You want to mock `Site.segment_client`.

ecommerce/extensions/analytics/config.py
```diff
@@ -1,15 +0,0 @@
-import analytics
-from django.conf import settings
-from oscar.apps.analytics import config
-
-
-class AnalyticsConfig(config.AnalyticsConfig):
-    name = 'ecommerce.extensions.analytics'
-
-    def ready(self):
-        if settings.INSTALL_DEFAULT_ANALYTICS_RECEIVERS:
```
zubair-arbi, line 10: @rlucioni @clintonb  we have a failing test in `ecommerce.extensions.analytics.tests.test_analytics.AnalyticsTests#test_order_receiver_enabled` due to removing of this file as we have move the config to `SiteConfiguration`.

```
ERROR: test_order_receiver_enabled (ecommerce.extensions.analytics.tests.test_analytics.AnalyticsTests)

----------------------------------------------------------------------

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/test/utils.py", line 196, in inner
    return test_func(*args, **kwargs)
  File "/home/travis/build/edx/ecommerce/ecommerce/extensions/analytics/tests/test_analytics.py", line 37, in test_order_receiver_enabled
    product = ProductRecord.objects.get(product=self.free_product)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/db/models/query.py", line 334, in get
    self.model._meta.object_name

DoesNotExist: ProductRecord matching query does not exist.
```

Should we remove the settings variable `INSTALL_DEFAULT_ANALYTICS_RECEIVERS` and the tests `ecommerce/extensions/analytics/tests/test_analytics.py`.

clintonb, line 10: Restore this file, and remove the references to the `analytics` package.

ecommerce/core/models.py
```diff
@@ -1,9 +1,11 @@
+from analytics import Client as SegmentClient
```
clintonb: This should be grouped with the third-party imports.

clintonb:
Make sure you've replaced all Python-based calls to `analytics.track` and `analytics.page`. You missed one: https://github.com/edx/ecommerce/blob/163fb716ec5148ce39ccfeae03db7776974f7ea7/ecommerce/extensions/checkout/signals.py#L32.

zubair-arbi:
@clintonb please give another review.

ecommerce/extensions/checkout/signals.py
```diff
@@ -29,7 +29,8 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
 
     user_tracking_id, lms_client_id, lms_ip = parse_tracking_context(order.user)
 
-    analytics.track(
+    request = get_current_request()
+    request.site.siteconfiguration.segment_client.track(
```
clintonb: Account for `request` not being present.

clintonb: Actually, you can use `order.site`.

ecommerce/extensions/refund/signals.py
```diff
@@ -21,7 +22,8 @@ def track_completed_refund(sender, refund=None, **kwargs):  # pylint: disable=un
     # Ecommerce transaction reversal, performed by emitting an event which is the inverse of an
     # order completion event emitted previously.
     # See: https://support.google.com/analytics/answer/1037443?hl=en
-    analytics.track(
+    request = get_current_request()
+    request.site.siteconfiguration.segment_client.track(
```
clintonb: `refund.order.site`

clintonb:
👍  (once final comments addressed)

awais786:
👍

PR #683, opened 2016-04-20
call analytics.page on ecommerce pages

zubair-arbi:
ECOM-4100
@awais786 @ahsan-ul-haq @tasawernawaz @waheedahmed 
- Add `segment` script for `Otto` in `base.html` so that the event `analytics.page()` is called on all pages.

_**Note:**The field `segment_key` in the model `SiteConfiguration` is introduced in a separate PR: https://github.com/edx/ecommerce/pull/682/_

### DIFF
--- a/ecommerce/core/context_processors.py
+++ b/ecommerce/core/context_processors.py
@@ -3,9 +3,9 @@ from django.conf import settings
 from ecommerce.core.url_utils import get_lms_dashboard_url
 
 
-def core(_request):
+def core(request):
     return {
         'lms_dashboard_url': get_lms_dashboard_url(),
-        'platform_name': _request.site.name,
+        'platform_name': request.site.name,
         'support_url': settings.SUPPORT_URL,
     }
--- a/ecommerce/extensions/analytics/tests/test_context_processors.py
+++ b/ecommerce/extensions/analytics/tests/test_context_processors.py
@@ -1,0 +1,19 @@
+from django.test import RequestFactory
+from ecommerce.extensions.analytics.context_processors import analytics
+from ecommerce.extensions.analytics.utils import prepare_analytics_data
+from ecommerce.tests.testcases import TestCase
+
+
+class AnalyticsContextProcessorTests(TestCase):
+    def test_analytics(self):
+        request = RequestFactory().get('/')
+        request.user = self.create_user()
+        request.site = self.site
+        analytics_data = prepare_analytics_data(request.user, request.site.siteconfiguration.segment_key)
+
+        self.assertDictEqual(
+            analytics(request),
+            {
+                'analytics_data': analytics_data,
+            }
+        )
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -4,6 +4,7 @@ import logging
 
 from threadlocals.threadlocals import get_current_request
 
+
 logger = logging.getLogger(__name__)
 
 
@@ -87,22 +88,23 @@ def audit_log(name, **kwargs):
     logger.info(message)
 
 
-def prepare_analytics_data(course_id, user):
+def prepare_analytics_data(user, segment_key, course_id=None):
     """ Helper function for preparing necessary data for analytics.
 
     Arguments:
-        course_id (str): The course ID.
         user (User): The user making the request.
+        segment_key (str): Segment write/API key.
+        course_id (str): The course ID.
 
     Returns:
-        JSON object with the data for analytics.
+        str: JSON object with the data for analytics.
     """
     data = {
         'course': {
             'courseId': course_id
         },
         'tracking': {
-            'segmentApplicationId': get_current_request().site.siteconfiguration.segment_key
+            'segmentApplicationId': segment_key
         }
     }
 
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -2,14 +2,14 @@ from functools import wraps
 import json
 import logging
 
-from django.conf import settings
+from threadlocals.threadlocals import get_current_request
 
 logger = logging.getLogger(__name__)
 
 
 def is_segment_configured():
     """Returns a Boolean indicating if Segment has been configured for use."""
-    return bool(settings.SEGMENT_KEY)
+    return bool(get_current_request().site.siteconfiguration.segment_key)
 
 
 def parse_tracking_context(user):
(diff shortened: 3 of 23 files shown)


ecommerce/core/context_processors.py
```diff
@@ -1,11 +1,13 @@
 from django.conf import settings
 
 from ecommerce.core.url_utils import get_lms_dashboard_url
+from threadlocals.threadlocals import get_current_request
 
 
 def core(_request):
```
clintonb: `request`

ecommerce/core/context_processors.py
```diff
@@ -1,11 +1,13 @@
...
     return {
         'lms_dashboard_url': get_lms_dashboard_url(),
         'platform_name': _request.site.name,
         'support_url': settings.SUPPORT_URL,
+        'segment_key': get_current_request().site.siteconfiguration.segment_key,
```
clintonb: `request.site.siteconfiguration.segment_key`

ecommerce/templates/edx/base.html
```diff
@@ -25,6 +25,13 @@
         {% block stylesheets %}
         {% endblock %}
     {% endcompress %}
+
+    <script type="text/javascript">
```
clintonb: Find a way to incorporate `analytics_view.js`.

zubair-arbi:
@clintonb @mjfrey can you please check if the approach is correct?

ecommerce/core/context_processors.py
```diff
@@ -1,11 +1,15 @@
...
-def core(_request):
+def core(request):
+    analytics_data = prepare_initial_analytics_data(request.user, request.site.siteconfiguration.segment_key)
+
     return {
         'lms_dashboard_url': get_lms_dashboard_url(),
-        'platform_name': _request.site.name,
+        'platform_name': request.site.name,
         'support_url': settings.SUPPORT_URL,
+        'analytics_data': analytics_data,
```
clintonb: Move this to context processor in the `ecommerce.extensions.analytics`. `core` should not be dependent on any internal app. Otherwise, you run the risk of a circular dependency.

ecommerce/core/tests/test_context_processors.py
```diff
@@ -1,22 +1,30 @@
...
 from threadlocals.threadlocals import get_current_request
 
 from ecommerce.core.context_processors import core
 from ecommerce.core.url_utils import get_lms_dashboard_url
+from ecommerce.extensions.analytics.utils import prepare_initial_analytics_data
 from ecommerce.tests.testcases import TestCase
 
+
 SUPPORT_URL = 'example.com'
+User = get_user_model()
```
clintonb: This is not needed.

ecommerce/static/js/common.js
```diff
@@ -4,14 +4,18 @@ require([
         'backbone.validation',
         'bootstrap',
         'bootstrap_accessibility',
-        'underscore'
+        'underscore',
+        'utils/analytics_utils'
     ],
-    function () {
+    function ($, Backbone, BackboneValidation, Bootstrap, BootstrapAccessibility, _, AnalyticsUtils) {
```
clintonb: Each of these should be on its own line.

ecommerce/static/js/views/analytics_view.js
```diff
@@ -66,8 +66,12 @@ define([
                 // provide our application key for logging
                 analytics.load(applicationKey);
 
-                // this needs to be called once
-                analytics.page(this.buildCourseProperties());
+                if (this.options.courseModel.get('courseId')) {
```
clintonb, line 77: This can be cleaned up by moving the logic of compiling the data sent to Segment to a single function:

``` javascript
getSegmentPageData: function() {
  if (this.options.courseModel.get('courseId')) {
    return this.buildCourseProperties();
  }

  return {};
}

var pageData = this.getSegmentPageData()
analytics.page(pageData);
```

ecommerce/extensions/analytics/utils.py
```diff
@@ -120,3 +120,39 @@ def prepare_analytics_data(course_id, user):
...
+def prepare_initial_analytics_data(user, segment_key, course_id=None):
+    """ Helper function for preparing necessary data for analytics.
+
+    Arguments:
+        user (User): The user making the request.
+        segment_key (str): Segment write/API key.
+        course_id (str): The course ID.
+
+    Returns:
+        JSON object with the data for analytics.
```
clintonb: str: JSON object with the data for analytics.

ecommerce/extensions/analytics/utils.py
```diff
@@ -120,3 +120,39 @@ def prepare_analytics_data(course_id, user):
         }
     data.update(user_data)
     return json.dumps(data)
+
+
+def prepare_initial_analytics_data(user, segment_key, course_id=None):
```
clintonb: Wait...why is this copied from above? Modify the function above to take a `segment_key` argument.

zubair-arbi:
@clintonb @awais786 @ahsan-ul-haq please review

ecommerce/extensions/analytics/tests/test_context_processors.py
```diff
@@ -0,0 +1,19 @@
+from threadlocals.threadlocals import get_current_request
+
+from ecommerce.extensions.analytics.context_processors import analytics
+from ecommerce.extensions.analytics.utils import prepare_analytics_data
+from ecommerce.tests.testcases import TestCase
+
+
+class AnalyticsContextProcessorTests(TestCase):
+    def test_analytics(self):
+        request = get_current_request()
```
clintonb: This shouldn't be used for tests. Use [`RequestFactory`](https://docs.djangoproject.com/en/1.8/topics/testing/advanced/#the-request-factory).:

``` python
from django.test import RequestFactory

request = RequestFactory().get('/')
request.user = self.create_user()
request.site = self.site
```

zubair-arbi:
@clintonb I have addressed your comment. Please take another look.

ecommerce/extensions/analytics/tests/test_context_processors.py
```diff
@@ -1,14 +1,14 @@
-from threadlocals.threadlocals import get_current_request
-
+from django.test import RequestFactory
```
clintonb, line 1: Group imports properly.

clintonb:
👍

clintonb:
Rebase and merge this PR to close both PRs simultaneously.

PR #684, opened 2016-04-20
[WIP] Receipt page for new checkout flow

ahsan-ul-haq:
WIP for Moving receipt pages from lms to ecommerce for new checkout flow

### DIFF
--- a/ecommerce/receipts/urls.py
+++ b/ecommerce/receipts/urls.py
@@ -1,0 +1,8 @@
+from django.conf.urls import url
+
+from ecommerce.receipts import views
+
+urlpatterns = [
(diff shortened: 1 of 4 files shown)


clintonb:
@ahsan-ul-haq before you do too much work here:
1. Oscar already has an order confirmation/receipt view. Porting the receipt page should primarily be a matter of creating/updating templates. See https://ecommerce.stage.edx.org/checkout/thank-you/?order_number=EDX-103543 for an example.
2. ECOM-4077 is not currently pointed, in a sprint, or assigned to anyone. I appreciate the initiative, but we should properly add this work to a sprint before starting.

PR #685, opened 2016-04-21
[WIP] [WL-428] Bulk enrollment fulfillment module.

vkaracic:
@mattdrayer this is the fulfillment module WIP.

Some of the open questions:
- [ ] to what should the end date be set on the vouchers generated this way? Currently it's being set to now() + 10 years

This PR contains:
- [x] bulk enrollment model that ties the specific vouchers the user bought with the order
- [x] `Bulk enrollment` category migration that adds a new category to the coupon parent category distinguishing bulk enrollment coupons created with course seats
- [ ] bulk enrollment fulfillment module which fulfills orders that contain coupons with the `Bulk enrollment` category. Creates enrollment vouchers, associates them to the order and sends an email to the users with an attached CSV that contains all the voucher codes and redemption URLs
- [x] forked `customer` Oscar app, needed to extend the `Dispatcher` class to include email attachments
- [ ] email templates for bulk enrollment product orders (probably needs UX)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=685&repo=edx%2Fecommerce

ecommerce/extensions/voucher/models.py
```diff
@@ -6,5 +6,10 @@ class CouponVouchers(models.Model):
     coupon = models.ForeignKey('catalogue.Product', related_name='coupon_vouchers')
     vouchers = models.ManyToManyField('voucher.Voucher', blank=True, related_name='coupon_vouchers')
 
+
+class BulkEnrollmentCoupon(models.Model):
```
mattdrayer: The name of this model seems odd when it only contains order and voucher references -- I'd rather see it named "OrderVouchers" if this is the case.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -338,3 +348,106 @@ def revoke_line(self, line):
...
+        """ Return a list of lines containing bulk enrollment coupon products that can be fulfilled.
+
+        Args:
+            lines (List of Lines): Order Lines, associated with purchased products in an Order.
+        Returns:
+            A supported list of unmodified lines associated with a bulk enrollment coupon product.
+        """
+        return [line for line in lines if self.supports_line(line)]
+
+    def fulfill_product(self, order, lines):
```
mattdrayer, line 381: Does it make sense to rename this method to "fulfill_order"?

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -338,3 +348,106 @@ def revoke_line(self, line):
...
+        """
+        logger.info("Attempting to fulfill 'Bulk Enrollment Coupon' product types for order [%s]", order.number)
+
+        file_name = 'Bulk enrollment CSV order num {}.csv'.format(order.number)
+        f = open(file_name, 'w')
+        writer = csv.writer(f)
+        writer.writerow(['Order Number', order.number])
+        writer.writerow([])
+
+        for line in lines:
```
mattdrayer, line 398: I don't really understand why we are passing in separate "order" and "lines" arguments -- are the lines not part of the order object at this point?

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -338,3 +348,106 @@ def revoke_line(self, line):
...
+            lines (List of Lines): Order Lines, associated with purchased products in an Order.
+
+        Returns:
+            The original set of lines, with new statuses set based on the success or failure of fulfillment.
+        """
+        logger.info("Attempting to fulfill 'Bulk Enrollment Coupon' product types for order [%s]", order.number)
+
+        file_name = 'Bulk enrollment CSV order num {}.csv'.format(order.number)
+        f = open(file_name, 'w')
+        writer = csv.writer(f)
```
mattdrayer: I think we should separate the work of creating the CSV file from the work of creating the vouchers for the order.  We want to ensure there are no issues creating the vouchers before we begin writing the CSV.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -338,3 +348,106 @@ def revoke_line(self, line):
...
+                stock_record = StockRecord.objects.get(product=line.product)
+                catalog.stock_records.add(stock_record)
+
+            vouchers = create_vouchers(
+                name='Bulk enrollment voucher [{}]'.format(line.product.title),
+                benefit_type=Benefit.PERCENTAGE,
+                benefit_value=100,
+                catalog=catalog,
+                coupon=line.product,
+                end_datetime=datetime.datetime.now() + datetime.timedelta(days=3640),
```
mattdrayer: Are end dates arbitrary?  This seems like a good use case for a constant or setting.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -338,3 +348,106 @@ def revoke_line(self, line):
             True, if the product is revoked; otherwise, False.
         """
         raise NotImplementedError("Revoke method not implemented!")
+
+
+class BulkEnrollmentFulfillmentModule(BaseFulfillmentModule):
+
+    def supports_line(self, line):
```
mattdrayer, line 354: Might be better to name this something like "is_bulk_enrollment_product", or "is_bulk_enrollment_coupon_product"

ecommerce/extensions/voucher/utils.py
```diff
@@ -432,3 +432,26 @@ def update_voucher_offer(offer, benefit_value, benefit_type):
         benefit_value=benefit_value,
         benefit_type=benefit_type
     )
+
+
+def create_voucher_csv(vouchers):
```
mattdrayer: We're not really creating a csv here -- we should change the name of the method to something different -- also it looks like we're really only using this operation in one place (aside from the tests), so we should probably move it into the fulfillment module itself.

ecommerce/extensions/fulfillment/modules.py
vkaracic, line 381: It has to be `fulfill_product` because is extending an abstract method: https://github.com/edx/ecommerce/blob/vkaracic/WL-428/ecommerce/extensions/fulfillment/modules.py#L71

ecommerce/extensions/fulfillment/modules.py
vkaracic, line 398: They are, but again, that's the way the base class' method is implemented.

ecommerce/extensions/fulfillment/modules.py
vkaracic: Yes, I mentioned so in the PR's conversation. A setting sounds good, can we agree to set it to 10 years from now, or should I use a different end date?

ecommerce/extensions/voucher/utils.py
vkaracic: I agree that the name does not reflect the method. But instead of renaming I'd like to change the method to accept the order and create a CSV and returns it. That way we'd have a function that can be called to re-generated a CSV apart from the fulfillment module. Maybe even set it so that this function emails the user too?

ecommerce/extensions/fulfillment/modules.py
vkaracic, line 354: It's the base class' abstract method that we implement in modules.

ecommerce/extensions/voucher/utils.py
vkaracic: Okay, take a look: https://github.com/edx/ecommerce/pull/685/commits/7c7b082074090bbc9a1e9fe362152389bf0f35a0 (tests are missing for the time being)

PR #690, opened 2016-04-26
[WIP!] Basket quantity

vkaracic:
Quick and dirty hack. I cannot reuse Oscar's quantity field here because, upon taking a close look at our frontend, we currently have a different structure of the basket UI. Oscar is using a composition of forms for it's basket while we have a mere display of the content of lines in the basket.

If there will be a need for new functionality on the basket page that Oscar has (that might even include this one) we'll need to do some refactoring here.

@mattdrayer FYI.

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -22,7 +22,9 @@ from ecommerce.extensions.offer.utils import format_benefit_value
 from ecommerce.extensions.partner.shortcuts import get_partner_for_site
 
 Benefit = get_model('offer', 'Benefit')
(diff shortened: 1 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=690&repo=edx%2Fecommerce

mattdrayer:
@vkaracic, we also have the prototype built by @malikshahzad228 in #674 that we can use as a reference -- I assume we'll end up with some sort of hybrid of these two PRs plus some additional changes when complete.

PR #691, opened 2016-04-26
[SOL-1667] Enable date picker in all browsers (coupon and course admin)

iivic:
@mjfrey  Date pickers should appear in all browsers. I used jquery-ui to solve this issue.
screen shot 2016-04-26 at 16 14 18

### DIFF
--- a/ecommerce/static/js/config.js
+++ b/ecommerce/static/js/config.js
@@ -18,6 +18,7 @@ require.config({
         'models': 'js/models',
         'moment': 'bower_components/moment/moment',
         'pages': 'js/pages',
+        'pikaday': 'bower_components/pikaday/pikaday',
         'requirejs': 'bower_components/requirejs/require',
         'routers': 'js/routers',
         'templates': 'templates',
@@ -43,6 +44,9 @@ require.config({
         },
         'jquery-cookie': {
             deps: ['jquery']
+        },
+        'pikaday': {
+            deps: ['moment']
         }
     }
 });
--- a/ecommerce/static/js/views/coupon_form_view.js
+++ b/ecommerce/static/js/views/coupon_form_view.js
@@ -12,7 +12,7 @@ define([
         'utils/utils',
         'text!templates/coupon_form.html',
         'models/course_model',
-        'views/form_view',
+        'views/form_view'
     ],
     function ($,
               Backbone,
@@ -329,6 +329,9 @@ define([
                     this.$el.find('[name=benefit_value]').attr('max', 100);
                     this.$el.find('button[type=submit]').html(gettext('Create Coupon'));
                 }
+
+                // Add date picker
+                Utils.addDatePicker(this);
 
                 this._super();
                 return this;
--- a/ecommerce/static/js/views/course_form_view.js
+++ b/ecommerce/static/js/views/course_form_view.js
@@ -17,7 +17,8 @@ define([
         'views/course_seat_form_fields/professional_course_seat_form_field_view',
         'views/course_seat_form_fields/credit_course_seat_form_field_view',
         'views/form_view',
-        'utils/course_utils'
+        'utils/course_utils',
+        'utils/utils'
     ],
     function ($,
               Backbone,
@@ -35,7 +36,8 @@ define([
               ProfessionalCourseSeatFormFieldView,
               CreditCourseSeatFormFieldView,
               FormView,
-              CourseUtils) {
+              CourseUtils,
+              Utils) {
         'use strict';
 
         return FormView.extend({
@@ -311,6 +313,9 @@ define([
                 _.each(activeSeats, function (seat) {
                     $courseSeats.filter('.' + seat).show();
                 });
+
+                // Add date picker
+                Utils.addDatePicker(this);
 
                 return this;
             }
--- a/ecommerce/templates/coupons/coupon_app.html
+++ b/ecommerce/templates/coupons/coupon_app.html
@@ -4,6 +4,11 @@
 
 {% block title %}{% trans "Coupon Codes" %}{% endblock %}
 
+{% block stylesheets %}
+    <link rel="stylesheet" href="{% static 'bower_components/pikaday/css/pikaday.css' %}"
+          type="text/x-scss">
+{% endblock %}
+
 {% block navbar %}
     <nav class="navbar navbar-default" aria-label="Account">
         <div class="container">
--- a/ecommerce/templates/courses/course_app.html
+++ b/ecommerce/templates/courses/course_app.html
@@ -4,6 +4,11 @@
 
 {% block title %}{% trans "Courses" %}{% endblock %}
 
+{% block stylesheets %}
+    <link rel="stylesheet" href="{% static 'bower_components/pikaday/css/pikaday.css' %}"
+          type="text/x-scss">
+{% endblock %}
+
 {% block navbar %}
     <nav class="navbar navbar-default" aria-label="Account">
         <div class="container">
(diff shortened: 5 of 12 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=691&repo=edx%2Fecommerce

clintonb:
1. Why is this necessary? Every modern browser renders a date picker. 
2. No jquery-UI. It's overkill for this.

iivic:
1. It isn't supported by every modern browser http://caniuse.com/#feat=input-datetime. That's the reason this bug is reported.
2. If jquery-ui isn't needed can you suggest another solution?

clintonb:
@andy-armstrong @AlasdairSwan do we have a datepicker in the edX toolkit? If not, how can we get one, and/or can you recommend one?

andy-armstrong:
@clintonb @AlasdairSwan Unfortunately we don't have a date picker in the UI Toolkit. Currently edx-platform is using the JQuery UI date picker, but I'd also like to see us remove it as it is too heavyweight.

From some brief research, this looks like a good bet. It also plays well with moment.js which is our preferred library for client-side date rendering:

https://github.com/dbushell/Pikaday

I've added date pickers to the list of requested patterns for the Pattern Library:

https://openedx.atlassian.net/wiki/display/FEDX/Pattern+Status

iivic:
This solution doesn't support time (only date). But there are GitHub forks for Pikaday which enable time picking. Should i use one of them as package or another solution?
Both date and time are needed in those forms.

mjfrey:
actually I do not think we need time.  That was just there because the widget in chrome has it be default.

clintonb:
You do need time. The field records a date-time value, and we have course runs that open/close at times other than midnight.

Use anything except jQuery UI for this.

mjfrey:
Interesting.  Since safari and firefox currently do not support time in that field.  Both for Coupons and Courses.

iivic:
The new date picker uses Pikaday and supports time pick. Someone review and thumbs up.

ecommerce/static/js/config.js
```diff
@@ -17,6 +17,7 @@ require.config({
         'jquery-cookie': 'bower_components/jquery-cookie/jquery.cookie',
         'models': 'js/models',
         'moment': 'bower_components/moment/moment',
+        'pikaday': 'bower_components/pikaday/pikaday',
```
clintonb: Minor: alphabetize with the existing list.

ecommerce/templates/edx/base.html
```diff
@@ -18,6 +18,7 @@
         {% else %}
             <link rel="stylesheet" href="{% static 'css/base/main.css' %}" type="text/x-scss">
         {% endif %}
+            <link rel="stylesheet" href="{% static 'bower_components/pikaday/css/pikaday.css' %}">
```
clintonb: This should not be included on every page. Add it only to the pages where it is used.

iivic: Done.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -330,10 +332,37 @@ define([
...
 
+                // Add date picker for browsers which doesn't support input[type="date"] with pikaday
+                this.addDatePicker();
+
                 this._super();
                 return this;
             },
 
+            addDatePicker: function() {
+                // Check if browser supports input[type="date"]
```
clintonb: Don't bother with this. Use the same UI element for all browsers.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -330,10 +332,37 @@ define([
...
+
+                if (testElement.type === 'text' ) {
+                    _.each(this.$el.find('[type=datetime-local]'), function(el) {
+                        if (el.getAttribute('datepicker-initialized') !== 'true') {
+                            new Pikaday({
+                                field: el,
+                                format: 'MM/DD/YYYY hh:mm a',
+                                defaultDate: this.model.get(el.name),
+                                setDefaultDate: true,
+
```
clintonb: Remove the unnecessary space.

ecommerce/static/js/views/course_form_view.js
```diff
@@ -312,8 +317,35 @@ define([
...
                 });
 
+                // Expires input field doesn't exist when the page loads
+                this.addDatePicker();
+
                 return this;
-            }
+            },
+
+            addDatePicker: function() {
```
clintonb: Don't duplicate code. Move this to a shared utility.

ecommerce/templates/coupons/coupon_app.html
```diff
@@ -4,6 +4,10 @@
 
 {% block title %}{% trans "Coupon Codes" %}{% endblock %}
 
+{% block stylesheets %}
+    <link rel="stylesheet" href="{% static 'bower_components/pikaday/css/pikaday.css' %}">
```
clintonb: Mark for compression.

ecommerce/templates/courses/course_app.html
```diff
@@ -4,6 +4,10 @@
 
 {% block title %}{% trans "Courses" %}{% endblock %}
 
+{% block stylesheets %}
+    <link rel="stylesheet" href="{% static 'bower_components/pikaday/css/pikaday.css' %}">
```
clintonb: Mark for compression.

clintonb:
👍 (once comments addressed)

ecommerce/static/js/views/coupon_form_view.js
iivic: Without this check chrome will display both datepickers Pikaday and the default one.

clintonb: Remove the `type` attribute from the fields.

iivic:
@clintonb Thank you for the review it was helpful. But now tests are failing I guess it's because of the compression but I don't know how to fix this issue. 
compressor.exceptions.OfflineGenerationError: You have offline compression enabled but key "31e3c658995c10d3c1759ad329ff040d" is missing from offline manifest. You may need to run "python manage.py compress". 
This is the error message you can see more in the test section.

iivic:
The issue is fixed. I will wait with the merge just in case someone wants to look at the code. Merging at end of the day.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -330,6 +330,9 @@ define([
                     this.$el.find('button[type=submit]').html(gettext('Create Coupon'));
                 }
 
+                // Add date picker
+                Utils.addDatePicker(this, 'MM/DD/YYYY hh:mm a');
```
clintonb: Please use an ISO-friendly format (e.g. `YYYY-MM-DD hh:mm a`) to avoid confusing non-American users.

ecommerce/static/js/views/course_form_view.js
```diff
@@ -312,6 +314,9 @@ define([
                     $courseSeats.filter('.' + seat).show();
                 });
 
+                // Add date picker
+                Utils.addDatePicker(this, 'YYYY-MM-DDTHH:mm:ss');
```
clintonb: Use the same format for all datepickers. This should probably be a constant in `Utils`.

ecommerce/static/js/views/coupon_form_view.js
iivic: OK we can use format `YYYY-MM-DDTHH:mm:ss` because only this format passes form validation in course_form. In this way both formats will be the same.

iivic:
All issues solved if none occur by tomorrow I will merge the branch.

PR #692, opened 2016-04-27
Fix default Upgrade Deadline for courses

awaisdar001:
ECOM-3657

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -8,6 +8,7 @@ from oscar.core.loading import get_model
 from simple_history.models import HistoricalRecords
 
 from ecommerce.courses.publishers import LMSPublisher
+from ecommerce.courses.utils import get_default_seat_upgrade_deadline
 from ecommerce.extensions.catalogue.utils import generate_sku
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


awaisdar001:
@marcotuts This is the PR. 
FYI: @bderusha

clintonb:
@bderusha please confirm that there is an issue, and this resolves it. I was under the impression this was deadline is reinforced in LMS.

marcotuts:
Thanks for the note here @awaisdar001. Ultimately someone on the dev team will need to review, possibly @bderusha  (as noted by Clinton earlier.)

awaisdar001:
@bderusha  Ping.

bderusha:
@awaisdar001 Unfortunately this ticket and subsequent work is unnecessary.

The default dates are handled by the LMS and calculated based on the course end date.
This code would cause issues because the first time a course is saved we would set the date to the default in LMS, but then if the course end date changes, altering the upgrade deadline e-commerce and lms would be out of sync.

Secondly, I believe that this is a data bug, not a system bug.  All courses created after the default date code released work correctly.  All courses with deadlines set in ecommerce work correctly.  The gap is courses created before the default code launched which did not set a deadline in ecommerce.  Those now look like they should be getting a default date. but they aren't.

I will run some queries and attach the impacted courses to the ticket.

I think we can close this PR.  Sorry for not catching this earlier before you started work.

awaisdar001:
Closing  :(

PR #693, opened 2016-04-27
WL-330: Update edx.org sass to show edx.org logo

saleem-latif:
Hi @mattdrayer ,  @jimabramson  , @clintonb  

kindly review this change set, I have updated sass for edx.org to override default header logo.


mattdrayer:
👍 -- @clintonb @jimabramson this is a small change and we need it to fix the stage AMI, so I'm going to merge it now

jimabramson:
my sass chops are very limited, but the changes look(ed) reasonable to me.  👍

PR #699, opened 2016-04-28
WL-330: Add DEFAULT_SITE_THEME settings

saleem-latif:
Hi @mattdrayer 

This PR contains addition of DEFAULT_SITE_THEME setting that can be used to define a default theme that should be used when theming is enabled and there is no theme for the current site.

### DIFF
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -507,6 +507,10 @@ DISABLE_THEMING_ON_RUNTIME_SWITCH = "disable_theming_on_runtime"
 # Directory that contains all themes
 COMPREHENSIVE_THEME_DIR = DJANGO_ROOT + "/themes"
 
+# Theme to use when no site or site theme is defined,
+# set to None if you want to use openedx theme
+DEFAULT_SITE_THEME = None
+
 # Cache time out for theme templates and related assets
 
 THEME_CACHE_TIMEOUT = 30 * 60
--- a/ecommerce/theming/models.py
+++ b/ecommerce/theming/models.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.conf import settings
 from django.contrib.sites.models import Site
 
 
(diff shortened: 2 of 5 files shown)


ecommerce/settings/base.py
```diff
@@ -507,6 +507,10 @@
 # Directory that contains all themes
 COMPREHENSIVE_THEME_DIR = DJANGO_ROOT + "/themes"
 
+# Theme to use when no site or site theme is defined,
+# set to None if you want to use openedx theme
+DEFAULT_SITE_THEME = "edx.org"
```
clintonb: The default should be the OpenEdX, never edx.org.

ecommerce/theming/models.py
```diff
@@ -12,3 +13,25 @@ class SiteTheme(models.Model):
...
+        Args:
+            site (django.contrib.sites.models.Site): site object related to the current site.
+
+        Returns:
+            SiteTheme object for given site or a default site set by `DEFAULT_SITE_THEME`
+        """
+        if not site:
+            return None
+
+        theme = site.themes.first()
```
clintonb, line 32: Why would a site ever have multiple themes?

clintonb, line 32: Ah. The `ForeignKey` on `SiteTheme` should be a `OneToOneField`.

mattdrayer, line 32: @clintonb I'd prefer to leave this as a ForeignKey field because we have discussed a use case for "stacking" multiple themes onto a site.  If we change to a OneToOneField then we'd have to change back to ForeignKey in order to support this behavior, so it seems to make sense to keep it a ForeignKey?

ecommerce/settings/base.py
mattdrayer: I've updated this value -- we'll set the default to 'edx.org' in the deployment config -- @jibsheet, FYI

ecommerce/theming/models.py
clintonb, line 32: (mindblown)

clintonb:
👍 (not that you asked for it. One day I'll stop reviewing everything.)

mattdrayer:
I appreciate the review -- I was going to ask for it anyway 😉

PR #700, opened 2016-04-28
mattdrayer/default-theme-override: Add default theme setting environ override

mattdrayer:
@clintonb -- FYI, looks like I needed to plumb the setting through to the environment config.

### DIFF
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -509,7 +509,7 @@ COMPREHENSIVE_THEME_DIR = DJANGO_ROOT + "/themes"
 
 # Theme to use when no site or site theme is defined,
 # set to None if you want to use openedx theme
-DEFAULT_SITE_THEME = None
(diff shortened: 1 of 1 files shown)


ecommerce/settings/base.py
```diff
@@ -509,7 +509,7 @@
 
 # Theme to use when no site or site theme is defined,
 # set to None if you want to use openedx theme
-DEFAULT_SITE_THEME = None
+DEFAULT_SITE_THEME = os.environ.get('DEFAULT_SITE_THEME', None)
```
clintonb, line 512: `production.py` already does this. We read from the YAML file and just add everything to the global namespace. See https://github.com/edx/ecommerce/blob/master/ecommerce/settings/production.py#L45-L48.

clintonb, line 512: You may want to override this in configuration; but, since we should be configuring sites, I see no need.

mattdrayer:
Closing in favor of https://github.com/edx/configuration/pull/2991

PR #701, opened 2016-04-28
Remove enrollment code migration functionality

vkaracic:
@clintonb

### DIFF
--- a/ecommerce/extensions/catalogue/migrations/0012_enrollment_code_product_class.py
+++ b/ecommerce/extensions/catalogue/migrations/0012_enrollment_code_product_class.py
@@ -3,100 +3,7 @@ from __future__ import unicode_literals
 import oscar
 from django.db import models, migrations
 
-
-def create_product_class(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=701&repo=edx%2Fecommerce

ecommerce/extensions/catalogue/migrations/0012_enrollment_code_product_class.py
```diff
@@ -3,100 +3,15 @@
...
-        depth=1,
-        path="0002",
-        image="",
-        name="Enrollment Codes"
-    )
+def create_product_class(apps, schema_editor):
+    pass
 
 
 def remove_product_class(apps, schema_editor):
```
clintonb: If you make `operations = []` below, I believe you can get rid of all of this.

clintonb:
👍

PR #702, opened 2016-04-29
[WIP] [WL-381] Enrollment codes

vkaracic:
Needs to be finished/addressed:
- [x] CSV file creation (permission issue) - CSV won't be sent in the email, just a link to download it
- [x] test the email sending
- [ ] write tests
- [ ] add checkbox to enable `create_enrollment_code` mode in CAT
- [x] email template design and content

@mattdrayer


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=702&repo=edx%2Fecommerce

clintonb:
Break this up into multiple PRs, please.

vkaracic:
Will do, but for now we need it all in one branch so that we can test it on the sandbox.

ecommerce/extensions/catalogue/migrations/0017_enrollment_code_product_class.py
```diff
@@ -0,0 +1,61 @@
...
+    ProductAttribute.objects.create(
+        product_class=enrollment_code,
+        name='Seat type',
+        code='seat_type',
+        type='text',
+        required=True
+    )
+
+    seat_product_class = ProductClass.objects.get(slug='seat')
+    ProductAttribute.objects.create(
```
vkaracic: TODO: remove this ProductAttribute in the remove function.

mattdrayer:
We'll also need to include the bulk sku with each seat product in the publish step from Otto to LMS -- see https://github.com/edx/ecommerce/blob/master/ecommerce/courses/publishers.py#L36 and https://github.com/edx/edx-platform/pull/12315/files#diff-682d1471c2d3e107e886a0cca174aaecR36

ecommerce/coupons/views.py
```diff
@@ -224,3 +228,65 @@ def get(self, request):
             return HttpResponseRedirect(reverse('basket:summary'))
 
         return HttpResponseRedirect(get_lms_url(''))
+
+
+class EnrollmentCodeCSVView(View):
```
mattdrayer: EnrollmentCodeCSVView -> EnrollmentCodeCsvView(

ecommerce/coupons/views.py
```diff
@@ -224,3 +228,65 @@ def get(self, request):
...
+            order (Order): The bulk enrollment order.
+            csv_lines(list): List of the order lines together with the particular
+                vouchers for that line:
+                [{
+                    'title'(str): The title of the product in the line.
+                    'vouchers'(list): List of the vouchers associated to that line.
+                }]
+        """
+
+        order = Order.objects.get(number=number)
```
mattdrayer: Do we need to defend against an invalid order number / missing entry?

ecommerce/coupons/views.py
```diff
@@ -224,3 +228,65 @@ def get(self, request):
...
+        if request.user != order.user:
+            raise Exception('Not allowed!')
```
mattdrayer: Please add a more descriptive exception message

ecommerce/coupons/views.py
```diff
@@ -224,3 +228,65 @@ def get(self, request):
...
+
+        order_line_vouchers = OrderLineVouchers.objects.filter(line__order=order)
```
mattdrayer: Please move this closer to where it's being used on line 281

ecommerce/coupons/views.py
```diff
@@ -224,3 +228,65 @@ def get(self, request):
...
+
+        file_name = 'Enrollment code CSV order num {}'.format(order.number)
```
mattdrayer, line 274: Please use named variables for format() operations

ecommerce/coupons/views.py
```diff
@@ -224,3 +228,65 @@ def get(self, request):
...
+        file_name = '{}.csv'.format(slugify(file_name))
+
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename={}'.format(file_name)
+
+        redeem_url = get_ecommerce_url(reverse('coupons:offer'))
+        voucher_field_names = ('Code', 'Redemption URL')
+        writer = csv.writer(response)
+        voucher_writer = csv.DictWriter(response, fieldnames=voucher_field_names)
```
mattdrayer, line 282: Please move this closer to where it is being used

ecommerce/coupons/views.py
```diff
@@ -224,3 +228,65 @@ def get(self, request):
...
+        order_line_vouchers = OrderLineVouchers.objects.filter(line__order=order)
+
+        file_name = 'Enrollment code CSV order num {}'.format(order.number)
+        file_name = '{}.csv'.format(slugify(file_name))
+
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename={}'.format(file_name)
+
+        redeem_url = get_ecommerce_url(reverse('coupons:offer'))
+        voucher_field_names = ('Code', 'Redemption URL')
```
mattdrayer, line 281: Please move this closer to where it is being used

ecommerce/coupons/views.py
vkaracic: I know. Every user-facing message is just plain placeholder for the time being, because they change way too often in the process so I'm saving them for last.

ecommerce/coupons/views.py
```diff
@@ -224,3 +228,65 @@ def get(self, request):
...
+        writer.writerow(('Order Number:', order.number))
+        writer.writerow([])
+
+        for order_line_voucher in order_line_vouchers:
+            writer.writerow([order_line_voucher.line.product.title])
+            voucher_writer.writeheader()
+
+            for voucher in order_line_voucher.vouchers.all():
+                voucher_writer.writerow({
+                    'Code': voucher.code,
```
mattdrayer: Please use the field names defined in voucher_field_names for these two properties

ecommerce/courses/models.py
```diff
@@ -208,9 +214,7 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
             )
 
         stock_record.price_excl_tax = price
-
-        # TODO Expose via setting
-        stock_record.price_currency = 'USD'
+        stock_record.price_currency = settings.OSCAR_DEFAULT_CURRENCY
```
mattdrayer, line 217: For my own knowledge, is there ever only a single currency defined for an instance, or is this something that is configurable?

ecommerce/courses/models.py
```diff
@@ -227,3 +231,32 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
             ).delete()
 
         return seat
+
+    def _create_enrollment_code(self, course_id, certificate_type, partner, price):
+        """Create a enrollment code product and a corresponding stock record."""
+        enrollment_code_product_class = ProductClass.objects.get(slug='enrollment_code')
```
mattdrayer, line 251: Need to protect against a possible DoesNotExist here, yes?

ecommerce/courses/models.py
```diff
@@ -227,3 +231,32 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
...
+        title = '{} ({}) Enrollment Code'.format(self.name, certificate_type)
+        enrollment_code = Product.objects.create(
+            title=title,
+            product_class=enrollment_code_product_class
+        )
+
+        enrollment_code.attr.course_key = course_id
+        enrollment_code.attr.seat_type = certificate_type
+        enrollment_code.save()
```
mattdrayer: Any chance we could collide with an existing record here?

ecommerce/courses/models.py
```diff
@@ -227,3 +231,32 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
...
+        except StockRecord.DoesNotExist:
+            enrollment_code_sku = generate_sku(enrollment_code, partner)
+            stock_record = StockRecord.objects.create(
+                product=enrollment_code,
+                partner=partner,
+                partner_sku=enrollment_code_sku
+            )
+
+        stock_record.price_excl_tax = price
+        stock_record.price_currency = settings.OSCAR_DEFAULT_CURRENCY
```
mattdrayer, line 274: Nit: Set the currency further back and reuse here and on L217

ecommerce/courses/models.py
vkaracic, line 217: AFAIK, this is being used for the currency template tag:
https://github.com/django-oscar/django-oscar/blob/ef863b53b840b549e3fc65927530748e0fd63e68/src/oscar/templatetags/currency_filters.py#L24
I believe this is configurable.

ecommerce/courses/models.py
vkaracic, line 251: I'd like to leave this as is, because if this fails then the migration which creates it failed and that migration creates everything we need for the enrollment code functionality which means we'd have bigger problems. Apart from that I don't believe it's common practice on this project to create/delete `ProductClass` objects outside of migrations (and tests).

ecommerce/extensions/api/serializers.py
```diff
@@ -77,6 +77,10 @@ def get_value(self, obj):
             vouchers = obj.value.vouchers.all()
             serializer = VoucherSerializer(vouchers, many=True, context={'request': request})
             return serializer.data
+        if obj.attribute.name == 'Enrollment code':
```
mattdrayer: This seems brittle to me -- it would be better to move "Enrollment code" into a parameter, constant, or setting which can be referenced throughout the system.

ecommerce/extensions/basket/views.py
```diff
@@ -64,40 +65,55 @@ class BasketSummaryView(BasketView):
...
+
+            lines_data.append({
+                'seat_type': seat_type,
+                'course_name': course['name'],
+                'course_org': course_key.org,
+                'course_number': course_key.run,
+                'image_url': image_url,
+                'course_short_description': short_description,
+                'benefit_value': benefit_value,
+                'enrollment_code': line.product.get_product_class().name == 'Enrollment code',
```
mattdrayer: Here's another spot where it would be nice to have "Enrollment code" available as a setting/param/constant

ecommerce/extensions/catalogue/migrations/0017_enrollment_code_product_class.py
```diff
@@ -0,0 +1,62 @@
...
+        code='enrollment_code',
+        type='entity',
+        required=False
+    )
+
+
+def remove_enrollment_code_product_class(apps, schema_editor):
+    """Remove the Enrollment code product class."""
+    ProductClass.objects.filter(slug='enrollment_code').delete()
+    ProductAttribute.objects.get(code='enrollment_code').delete()
```
mattdrayer, line 50: I'm not entirely clear on the data model, but it seems we might need to remove the ProductAttribute before the ProductClass due to referential integrity here

ecommerce/extensions/catalogue/utils.py
```diff
@@ -23,6 +23,12 @@ def generate_sku(product, partner, **kwargs):
             unicode(catalog.id),
             str(partner.id)
         ))
+    elif product.product_class and product.product_class.name == 'Enrollment code':
```
mattdrayer: "Enrollment code" --> setting/constant/param

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -338,3 +347,93 @@ def revoke_line(self, line):
...
+        Check whether the product in line is an Enrollment code.
+
+        Args:
+            line (Line): Line to be considered.
+
+        Returns:
+            True if the line contains an Enrollment code.
+            False otherwise.
+        """
+        return line.product.get_product_class().name == 'Enrollment code'
```
mattdrayer: "Enrollment code" --> setting/constant/param

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -338,3 +347,93 @@ def revoke_line(self, line):
...
+                'vouchers': vouchers
+            }
+            csv_lines.append(csv_line)
+            line.set_status(LINE.COMPLETE)
+
+        send_notification(order.user, 'ORDER_WITH_CSV', context={
+            'csv_download_link': get_ecommerce_url(reverse('coupons:enrollment_code_csv', args=[order.number])),
+        })
+
+        logger.info("Finished fulfilling 'Enrollment code' product types for order [%s]", order.number)
```
mattdrayer, line 445: Several spots in this module where we could reference a setting/constant/param instead of "Enrollment code"

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -338,3 +347,93 @@ def revoke_line(self, line):
...
+    def revoke_line(self, line):
+        """ Revokes the specified line.
+
+        Args:
+            line (Line): Order Line to be revoked.
+
+        Returns:
+            True, if the product is revoked; otherwise, False.
+        """
+        raise NotImplementedError("Revoke method not implemented!")
```
mattdrayer, line 457: "The 'revoke_line' method is not implemented."

ecommerce/extensions/voucher/models.py
```diff
@@ -6,5 +6,10 @@ class CouponVouchers(models.Model):
     coupon = models.ForeignKey('catalogue.Product', related_name='coupon_vouchers')
     vouchers = models.ManyToManyField('voucher.Voucher', blank=True, related_name='coupon_vouchers')
 
+
+class OrderLineVouchers(models.Model):
+    line = models.ForeignKey('order.Line', related_name='order_vouchers')
```
mattdrayer: Should the related_name be 'order_line_vouchers'?

ecommerce/extensions/voucher/utils.py
```diff
@@ -320,7 +320,8 @@ def create_vouchers(
         voucher_type,
         coupon_id=None,
         code=None,
-        max_uses=None):
+        max_uses=None,
+        _range=None):
```
mattdrayer, line 324: For my own knowledge, what does it mean to have an argument prefixed with an underscore?

ecommerce/settings/base.py
```diff
@@ -516,3 +516,6 @@
 THEME_CACHE_TIMEOUT = 30 * 60
 
 # End Theme settings
+
+# Expiration date for bulk enrollment vouchers
+BULK_ENROLLMENT_VOUCHER_END_DATE = '2026-01-01'
```
mattdrayer: Does this have to be an exact value?  Could it instead be a relativedelta of +10 years from today?

ecommerce/extensions/voucher/utils.py
vkaracic, line 324: I believe it's bad practice to use languages' keywords as variable names, hence the distinction.

ecommerce/settings/base.py
mattdrayer: Also I believe that at some point (maybe not now) there will be an override available for this value on a per-product basis -- we might want to think about how this setting will be applied in that case.

ecommerce/templates/customer/emails/commtype_order_with_csv_body.html
```diff
@@ -0,0 +1,78 @@
+{% extends 'customer/email_base.html' %}
```
mattdrayer, line 1: I checked with @griffresch and for this template we can borrow the content from ecommerce/templates/customer/emails/commtype_course_purchased_body.html

ecommerce/templates/customer/emails/commtype_order_with_csv_body.txt
```diff
@@ -0,0 +1,6 @@
+{% load i18n %}
```
mattdrayer, line 1: I checked with @griffresch and for this template we can borrow the content from ecommerce/templates/customer/emails/commtype_course_purchased_body.txt

ecommerce/templates/customer/emails/commtype_order_with_csv_subject.txt
```diff
@@ -0,0 +1,2 @@
+{% load i18n %}
+{% trans "Enrollment Code Order Receipt" %}
```
mattdrayer: Is it possible to include the partner short name and order number in the subject line?  eg:

"{partner_short_name}: Enrollment Code Purchase Receipt [Order No. {order_number}] "

@griffresch

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -13,28 +13,39 @@
...
+                                <p class="course_name">{{ line_data.course_name }} - {{ line_data.course_org }} ({{ line.course.number }})</p>
+                                <p class="course_description">{{ line_data.course_short_description }}</p>
                             </div>
-                            <div class="col-sm-5 course_prices">
+                            {% if line_data.enrollment_code %}
+                            <div class="col-sm-2">
+                                <div class="checkout-quantity">
+                                    <div class="input-group  {% if form.errors %}error{% endif %}">
+                                        {% render_field form.quantity class+="quantity" %}
+                                    <button class="btn btn-default" type="submit" data-loading-text="{% trans 'Updating...' %}">{% trans "Update" %}</button>
```
mattdrayer, line 39: I think due to timing we're getting close to "nice to have" on this, but the original ask was to provide a quantity field that auto-updated on a change of value.  I think @malikshahzad228 was able to get a prototype of this up and running in his branch -- see https://github.com/edx/ecommerce/pull/674/files#diff-a77e90eb28521ce0645b1b9831b34192R82

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -48,13 +59,15 @@
...
                         </div>
+                        {% endwith %}
                     {% endfor %}
                 </div>
             </form>
         {% endblock %}
 
         <div class="total">
             <div class="row">
+            {% if line_data.enrollment_code == False %}
```
mattdrayer: Can this be written as "!line_data.enrollment_code" ?

ecommerce/courses/models.py
mattdrayer, line 251: Okay -- can we add to the docstring explaining the intended use case for this method, so future devs are aware?

vkaracic, line 251: Of course. This whole code is still very much a WIP so I didn't get to write proper documentation just yet.

ecommerce/templates/customer/emails/commtype_order_with_csv_body.html
mattdrayer, line 1: Also @saleem-latif, FYI on these email templates which will require theming once available.

ecommerce/courses/models.py
vkaracic, line 274: These are two different stock records.

ecommerce/templates/oscar/basket/partials/basket_content.html
vkaracic, line 39: Definitely a 'nice to have' at this time. Each line is an object in the database and therefor the quantity form needs to update the object's `quantity` field before a checkout happens.

ecommerce/coupons/views.py
vkaracic, line 282: It's being used in a for-loop, there is no need to initialize the same thing several times.

ecommerce/extensions/fulfillment/modules.py
vkaracic, line 457: This language has been used in all the unimplemented methods in this document, I'd rather not deviate.

ecommerce/settings/base.py
vkaracic: Overrides should happen in the fulfillment module that sets the voucher dates.

ecommerce/extensions/api/serializers.py
```diff
@@ -12,7 +12,7 @@
 from rest_framework.reverse import reverse
 import waffle
 
-from ecommerce.core.constants import ISO_8601_FORMAT, COURSE_ID_REGEX
+from ecommerce.core.constants import COURSE_ID_REGEX, ENROLLMNENT_CODE, ISO_8601_FORMAT
```
mattdrayer: ENROLLMNENT_CODE -> ENROLLMENT_CODE

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -432,11 +432,14 @@ def fulfill_product(self, order, lines):
             line.set_status(LINE.COMPLETE)
 
         send_notification(order.user, 'ORDER_WITH_CSV', context={
-            'enrollment_code_title': lines[0].product.title,
+            'contact_url': "{}/contact".format(site.siteconfiguration.lms_url_root)
+            'course_name': order.lines.first().product.course.name,
```
vkaracic, line 436: Is this right? This would imply that only one course enrollment codes can be purchased at a time.

mattdrayer:
closed in favor of #712, #718, and #719

PR #703, opened 2016-04-29
Fix acceptance/end-to-end tests

rlucioni:
Fixes include: interacting with the iframe on PayPal's new login page, updates to deal with the new edX checkout page, loading the edX receipt page several times to deal with slow payment processing, more robust interaction with the coupon admin app, and allowing coupon admin tests to be skipped (which they are, by default).

@schenedx @jimabramson @ronaldmulero @mjfrey FYI

### DIFF
--- a/acceptance_tests/api.py
+++ b/acceptance_tests/api.py
@@ -1,7 +1,7 @@
 import requests
 from requests.auth import AuthBase
 
-from acceptance_tests.config import ENROLLMENT_API_URL, ENROLLMENT_API_TOKEN
+from acceptance_tests.config import ACCESS_TOKEN, ENROLLMENT_API_URL
 
 
 class BearerAuth(AuthBase):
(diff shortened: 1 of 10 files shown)


clintonb:
>  ...and allowing coupon admin tests to be skipped (which they are, by default).

Why are these tests skipped?

rlucioni:
@clintonb they can be skipped because the same functionality is covered by the coupon checkout tests. I've left the admin tests in place for now because I'm not sure if there's benefit to being able to run them independently. If they need to be removed, I'd prefer that be done separately. There was talk of using an API to create coupons during the checkout tests, in which case the admin tests would become useful again.

clintonb:
I made the recommendation of correcting the checkout tests. They should definitely be using the API rather than the UI. Please file a SOL ticket so that this can be fixed.

rlucioni:
@clintonb see [SOL-1775](https://openedx.atlassian.net/browse/SOL-1775).

mjfrey:
For completeness I created  [ECOM-4319](https://openedx.atlassian.net/browse/ECOM-4319).  As there currently are no acceptance tests for the Course Admin Tool.

PR #704, opened 2016-04-29
Add Codecov YAML config

rlucioni:
@clintonb FYI. Corresponds to the config added to programs and the cookiecutter repo.


clintonb:
👍

PR #711, opened 2016-05-04
Adjust PayPal information for checking out baskets with >1 quantity lines

vkaracic:
The PayPal API requires for the [ `item_list` ](https://developer.paypal.com/docs/api/#item-object) to be populated with item data so that the sum of item prices, where every item price is `price * quantity`, equals to the `amount['total']` that you're sending in the `transactions` values.


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=711&repo=edx%2Fecommerce

clintonb:
👍

PR #717, opened 2016-05-05
Credit checkout update

clintonb:
- Checking to see if a seat is actually available for purchase, before displaying info to user.
- Displaying a more helpful error message if no seat is available.

ECOM-4378

### DIFF
--- a/ecommerce/credit/tests/test_views.py
+++ b/ecommerce/credit/tests/test_views.py
@@ -4,15 +4,17 @@ Tests for the checkout page.
 from __future__ import unicode_literals
 
 import json
+from datetime import timedelta
 
 import ddt
+import httpretty
 from django.conf import settings
 from django.core.urlresolvers import reverse
-import httpretty
+from django.utils import timezone
 from waffle.models import Switch
 
 from ecommerce.core.url_utils import get_lms_url
-from ecommerce.courses.models import Course
+from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.extensions.payment.helpers import get_processor_class
 from ecommerce.tests.mixins import JwtMixin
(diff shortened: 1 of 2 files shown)


clintonb:
@rlucioni please review

ecommerce/credit/tests/test_views.py
```diff
@@ -6,13 +6,14 @@
 import json
 
 import ddt
+from datetime import datetime, timedelta
```
rlucioni: Nitpick: `datetime` is part of the standard library. pylint>=1.5.4 will catch this.

ecommerce/credit/views.py
```diff
@@ -32,34 +34,48 @@ def get_context_data(self, **kwargs):
         context = super(Checkout, self).get_context_data(**kwargs)
 
         course = get_object_or_404(Course, id=kwargs.get('course_id'))
+        context['course'] = course
```
rlucioni, line 37: Left over from debugging? This was added but I don't see it being used.

ecommerce/credit/tests/test_views.py
```diff
@@ -285,3 +268,18 @@ def test_disabled_providers(self, disabled_processors, expected_context):
         response = self.client.get(self.path)
         self.assertEqual(response.status_code, 200)
         self.assertDictContainsSubset(expected_context, response.context)
+
+    def test_seat_unavailable(self):
```
rlucioni, line 274: Failures suggest you may need to use activate/use httpretty for this test.

rlucioni, line 274: Looks like you'll at least want to `_mock_eligibility_api`.

clintonb, line 274: Fixed.

ecommerce/credit/tests/test_views.py
clintonb: Fixed.

ecommerce/credit/views.py
clintonb, line 37: Nope. Previously, the error conditions did not add the course to the context. This resulted in the title of the page being incorrect. This fixes that.

rlucioni:
👍

clintonb:
screen shot 2016-05-05 at 5 58 28 pm

PR #721, opened 2016-05-10
Only check max-applications if the Voucher is not SINGLE_USE

mjfrey:
We should only check max-applications on non SINGLE-USE vouchers.


mjfrey:
@vkaracic

PR #722, opened 2016-05-10
This improves New Relic's error analytics

jibsheet:
@edx/ecommerce If you want to see this in action in NR, you can look at the lms or studio apps and check out a few enhancements to the response times page, but the nice win is that it enables their new Error Analytics charting.

### DIFF
--- a/requirements/optional.txt
+++ b/requirements/optional.txt
@@ -1,2 +1,2 @@
 # Optional packages
-newrelic==2.58.2.45
+newrelic==2.64.0.48


clintonb:
👍

rlucioni:
👍

jibsheet:
Ping me if I'm not on call during the next release and I'll verify in New Relic.  If needed, anyone else with NR access can confirm the app is still reporting and we'll be mostly good.

PR #723, opened 2016-05-10
Referral Tracking: Attribute a basket with the affiliate

bderusha:
Ready for review

### DIFF
--- a/ecommerce/extensions/basket/models.py
+++ b/ecommerce/extensions/basket/models.py
@@ -8,8 +8,9 @@ Selector = get_class('partner.strategy', 'Selector')
 
 
 class Basket(AbstractBasket):
-    site = models.ForeignKey('sites.Site', verbose_name=_("Site"), null=True, blank=True, default=None,
-                             on_delete=models.SET_NULL)
+    site = models.ForeignKey(
+        'sites.Site', verbose_name=_("Site"), null=True, blank=True, default=None, on_delete=models.SET_NULL
+    )
 
     @property
     def order_number(self):
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -7,6 +7,7 @@ from oscar.test.factories import ProductFactory, RangeFactory, VoucherFactory
 from ecommerce.extensions.basket.utils import get_certificate_type_display_value, prepare_basket
 from ecommerce.extensions.partner.models import StockRecord
 from ecommerce.extensions.test.factories import prepare_voucher
+from ecommerce.referrals.models import Referral
 from ecommerce.tests.factories import SiteConfigurationFactory
 from ecommerce.tests.testcases import TestCase
 
@@ -21,6 +22,7 @@ class BasketUtilsTests(TestCase):
     def setUp(self):
         super(BasketUtilsTests, self).setUp()
         self.request = RequestFactory()
+        self.request.COOKIES = {}
         self.request.user = self.create_user()
         site_configuration = SiteConfigurationFactory(partner__name='Tester')
         self.request.site = site_configuration.site
@@ -81,6 +83,34 @@ class BasketUtilsTests(TestCase):
         self.assertEqual(basket.lines.count(), 1)
         self.assertEqual(basket.lines.first().product, product2)
         self.assertEqual(basket.product_quantity(product2), 1)
+
+    def test_prepare_basket_affiliate_cookie_lifecycle(self):
+        """ Verify a basket is returned and referral captured. """
+        product = ProductFactory()
+        affiliate_id = 'test_affiliate'
+        self.request.COOKIES['affiliate_id'] = affiliate_id
+        basket = prepare_basket(self.request, product)
+
+        # test affiliate id from cookie saved in referral
+        referral = Referral.objects.get(basket_id=basket.id)
+        self.assertEqual(referral.affiliate_id, affiliate_id)
+
+        # update cookie
+        new_affiliate_id = 'new_affiliate'
+        self.request.COOKIES['affiliate_id'] = new_affiliate_id
+        basket = prepare_basket(self.request, product)
+
+        # test new affiliate id saved
+        referral = Referral.objects.get(basket_id=basket.id)
+        self.assertEqual(referral.affiliate_id, new_affiliate_id)
+
+        # expire cookie
+        del self.request.COOKIES['affiliate_id']
+        basket = prepare_basket(self.request, product)
+
+        # test referral record is deleted when no cookie set
+        with self.assertRaises(Referral.DoesNotExist):
+            Referral.objects.get(basket_id=basket.id)
 
     @ddt.data(
         ('honor', 'Honor'),
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -1,16 +1,20 @@
 import logging
 
+from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_class, get_model
 
+from ecommerce.referrals.models import Referral
+
 Applicator = get_class('offer.utils', 'Applicator')
 Basket = get_model('basket', 'Basket')
+
 logger = logging.getLogger(__name__)
 
 
 def prepare_basket(request, product, voucher=None):
     """
-    Create or get the basket, add the product, and apply a voucher.
+    Create or get the basket, add the product, apply a voucher, and record referral data.
 
     Existing baskets are merged. The specified product will
     be added to the remaining open basket. If a voucher is passed, all existing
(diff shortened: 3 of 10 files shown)


ecommerce/extensions/basket/middleware.py
```diff
@@ -19,6 +20,30 @@ def get_cookie_key(self, request):
         key = '{base}_{site_id}'.format(base=key, site_id=request.site.id)
         return key
 
+    def get_affiliate_cookie_key(self, request):
```
bderusha: No need for this method.  Does not `site_id`

ecommerce/extensions/basket/middleware.py
```diff
@@ -19,6 +20,30 @@ def get_cookie_key(self, request):
...
+            request (Request) -- current request being processed
+
+        Returns:
+            str - cookie name
+        """
+        return '{base}_{site_id}'.format(base=AFFILIATE_COOKIE_KEY, site_id=request.site.id)
+
+    def get_affiliate_id(self, cookie_key, request):
+        """
+        Return the affiliate cookie name for referral attribution.
```
clintonb: affiliate ID

ecommerce/extensions/basket/models.py
```diff
@@ -8,8 +8,10 @@
 
 
 class Basket(AbstractBasket):
-    site = models.ForeignKey('sites.Site', verbose_name=_("Site"), null=True, blank=True, default=None,
-                             on_delete=models.SET_NULL)
+    site = models.ForeignKey(
+        'sites.Site', verbose_name=_("Site"), null=True, blank=True, default=None, on_delete=models.SET_NULL
+    )
+    affiliate_id = models.CharField(_('Affiliate ID'), null=True, default=None, max_length=255)
```
clintonb: `blank=True`

ecommerce/extensions/basket/middleware.py
```diff
@@ -2,6 +2,7 @@
 from oscar.core.loading import get_model
 
 Basket = get_model('basket', 'basket')
+AFFILIATE_COOKIE_KEY = 'affiliate_partner'
```
peter-fogg: This should be `'affiliate_id'`.

ecommerce/extensions/basket/middleware.py
```diff
@@ -19,6 +20,30 @@ def get_cookie_key(self, request):
...
+        """
+        Return the affiliate cookie name for referral attribution.
+
+        Parameters:
+            request (Request) -- current request being processed
+
+        Returns:
+            str - cookie name
+        """
+        return request.COOKIES.get(cookie_key)
```
peter-fogg: This could also return None if there's no cookie, so I guess we either want to assign `request.affiliate_id = None` or not set it?

clintonb:
@bderusha be careful with adding a column to the `Basket` and `Order` models. Those are rather large tables, and modifying them will lead to downtime. We do have a management command to clear all ordered baskets; but, you should work with @mjfrey to ensure clearing these baskets doesn't break the `Invoice` model which has a non-nullable foreign key to the `Basket` model.

P.S. Otherwise, I like this approach.

bderusha:
@clintonb @peter-fogg please review

ecommerce/extensions/order/utils.py
```diff
@@ -99,4 +102,14 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
...
         order.save()
+
+        try:
+            referral = Referral.objects.get(basket=basket)
+            referral.order = order
+            referral.save()
+        except ObjectDoesNotExist:
+            logger.debug('Order [%d] has no referral associated with its basket.', order.id)
+        except Exception as exception:  # pylint: disable=broad-except
+            logger.error('Referral for Order [%d] failed to save: [%s] .', order.id, exception)
```
clintonb: Use `logger.exception` and you don't need to worry about the `exception` variable.

ecommerce/referrals/models.py
```diff
@@ -0,0 +1,14 @@
...
+
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+from django_extensions.db.models import TimeStampedModel
+
+logger = logging.getLogger(__name__)
+
+
+class Referral(TimeStampedModel):
+    affiliate_id = models.CharField(_('Affiliate ID'), null=True, blank=True, default=None, max_length=255)
```
clintonb: `null=False blank=False`

This field is null, thee row shouldn't exist.

ecommerce/extensions/order/utils.py
bderusha: Will that still grab the right exception instance and thus the appropriate message?

clintonb: Yes.

ecommerce/referrals/models.py
```diff
@@ -0,0 +1,14 @@
...
+    basket = models.ForeignKey('basket.Basket', related_name='basket', null=True, blank=True)
```
clintonb: [`OneToOneField`](https://docs.djangoproject.com/en/1.9/topics/db/examples/one_to_one/) seems more appropriate for these relationships.

ecommerce/extensions/basket/utils.py
```diff
@@ -35,6 +39,12 @@ def prepare_basket(request, product, voucher=None):
         Applicator().apply(basket, request.user, request)
         logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
 
+    affiliate_id = request.COOKIES.get(settings.AFFILIATE_COOKIE_KEY)
+    Referral.objects.update_or_create(
```
peter-fogg: In combination with Clinton's comment on the model definition, it seems like we should check to see if `affiliate_id is not None` here.

ecommerce/extensions/order/utils.py
```diff
@@ -99,4 +102,14 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
             order_data.update(extra_order_fields)
         order = Order(**order_data)
         order.save()
+
+        try:
+            referral = Referral.objects.get(basket=basket)
+            referral.order = order
+            referral.save()
+        except ObjectDoesNotExist:
```
peter-fogg: Is there a reason to catch `ObjectDoesNotExist` instead of the `Referral`-specific exception?

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -82,6 +84,34 @@ def test_prepare_basket_with_multiple_products(self):
...
+        new_affiliate_id = 'new_affiliate'
+        self.request.COOKIES['affiliate_id'] = new_affiliate_id
+        basket = prepare_basket(self.request, product)
+
+        # test new affiliate id saved
+        referral = Referral.objects.get(basket_id=basket.id)
+        self.assertEqual(referral.affiliate_id, new_affiliate_id)
+
+        # expire cookie
+        self.request.COOKIES['affiliate_id'] = None
```
peter-fogg: It might be good to explicitly delete the `affiliate_id` cookie instead of setting it to `None` (https://docs.djangoproject.com/en/1.9/ref/request-response/#django.http.HttpResponse.delete_cookie). That way we can ensure that no code assumes the cookie is there, ending up with a `KeyError` when the cookie isn't set.

bderusha: I don't know that that will work here since that is a response method and we are working with the request here.  I didn't see anything similar in the docs.

ecommerce/extensions/order/utils.py
bderusha: fixed

ecommerce/extensions/basket/utils.py
bderusha: addressed along with clinton's comment

bderusha:
@peter-fogg @clintonb comments addressed

ecommerce/extensions/basket/tests/test_utils.py
peter-fogg: Erk, you're totally right about that. Maybe `del self.request.COOKIES['affiliate_id']`?

peter-fogg:
👍 :shipit:

ecommerce/referrals/models.py
```diff
@@ -0,0 +1,14 @@
...
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+from django_extensions.db.models import TimeStampedModel
+
+logger = logging.getLogger(__name__)
+
+
+class Referral(TimeStampedModel):
+    affiliate_id = models.CharField(_('Affiliate ID'), null=False, blank=False, default=None, max_length=255)
+    basket = models.OneToOneField('basket.Basket', related_name='basket', null=True, blank=True)
```
clintonb: Remove `related_name`. The default, `referral` should be sufficient. As is, this will point `basket.basket` to a `Referral` instance.

clintonb:
👍 (once `related_name` fixed)

PR #725, opened 2016-05-11
Enforce proper max_uses value for coupons.

vkaracic:
The issue was that a `max_uses` value would be sent for `SINGLE_USE` voucher type coupons even though the `max_uses` input field on the UI was hidden. I missed that in the initial implementation.
This PR enforces  `max_uses` to be `None` (default) for voucher types other than the multi usage one (`ONCE_PER_CUSTOMER`).

https://openedx.atlassian.net/browse/SOL-1792
@mjfrey


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=725&repo=edx%2Fecommerce

mjfrey:
:+1

mjfrey:
:+1:

PR #727, opened 2016-05-11
Management command to set SINGLE_USE vouchers global_applications

mjfrey:
Some SINGLE_USE vouchers may have been created with the wrong max_applications set.  This command fixes them.

### DIFF
--- a/ecommerce/extensions/voucher/management/commands/fix_single_use.py
+++ b/ecommerce/extensions/voucher/management/commands/fix_single_use.py
@@ -1,0 +1,30 @@
+""" This command fixes SINGLE_USE vouchers."""
+import logging
+
+from django.core.management import BaseCommand
+
+from oscar.core.loading import get_model
+
+logger = logging.getLogger(__name__)
+
+Voucher = get_model('voucher', 'Voucher')
+
+
+class Command(BaseCommand):
+    """
+    Fix all SINGLE_USE vouchers.
+    Some SINGLE_USE vouchers may have been created with the wrong max_applications set.
+    This command sets their max_global_applications to the default value.
+    """
+
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.DEBUG)
+    logger.addHandler(ch)
+
+    def handle(self, *args, **options):
+        vouchers = Voucher.objects.filter(usage=Voucher.SINGLE_USE, offers__max_global_applications=1)
+        for v in vouchers:
+            offer = v.offers.first()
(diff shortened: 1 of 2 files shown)


mjfrey:
@vkaracic

vkaracic:
I've added some slight changes. I've changed the command to set `None` for the `max_global_applications` because that is the default. Although 10000 should be fine, I'd rather use the default value because I'm not sure if changing to 10000 will do some unexpected changes on other place(s).

mjfrey:
👍 
squash commits.

mjfrey:
@douglashall or @mattdrayer can you take a quick look?

ecommerce/extensions/voucher/management/commands/fix_single_use.py
```diff
@@ -0,0 +1,26 @@
...
+
+from oscar.core.loading import get_model
+
+logger = logging.getLogger(__name__)
+
+Voucher = get_model('voucher', 'Voucher')
+
+
+class Command(BaseCommand):
+    """Fix all SINGLE_USE vouchers."""
```
mattdrayer: Why do they need to be fixed?  Is this intended to be a one-time operation?  Would be good to include more context in the docstring.

ecommerce/extensions/voucher/management/commands/fix_single_use.py
```diff
@@ -0,0 +1,26 @@
...
+
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.DEBUG)
+    logger.addHandler(ch)
+
+    def handle(self, *args, **options):
+        vouchers = Voucher.objects.filter(usage=Voucher.SINGLE_USE, offers__max_global_applications=1)
+        for v in vouchers:
+            offer = v.offers.first()
```
mattdrayer, line 27: Do we need to accommodate for the case where there are no offers?

ecommerce/extensions/voucher/management/commands/fix_single_use.py
vkaracic: This is a one-time operation. The fix for the bug that was causing these issues has been deployed in yesterdays release.

ecommerce/extensions/voucher/management/commands/fix_single_use.py
vkaracic, line 27: Hmm, all the vouchers need to have one offer on them. If there are any without offers then that would be a different problem.

mattdrayer:
👍

PR #729, opened 2016-05-12
mattdrayer/from-email-default: Allow null field values

mattdrayer:
@clintonb @mjfrey -- including a default value for the new SiteConfiguration.from_email model field.

### DIFF
--- a/ecommerce/core/migrations/0015_siteconfiguration_from_email.py
+++ b/ecommerce/core/migrations/0015_siteconfiguration_from_email.py
@@ -14,7 +14,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='siteconfiguration',
             name='from_email',
-            field=models.CharField(default='oscar@example.com', help_text='Address from which emails are sent.', max_length=255, verbose_name='From email'),
-            preserve_default=False,
+            field=models.CharField(help_text='Address from which emails are sent.', max_length=255, null=True, verbose_name='From email', blank=True),
         ),
     ]
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -64,8 +64,8 @@ class SiteConfiguration(models.Model):
         verbose_name=_('From email'),
         help_text=_('Address from which emails are sent.'),
         max_length=255,
-        null=False,
-        blank=False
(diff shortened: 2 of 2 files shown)


ecommerce/core/migrations/0015_siteconfiguration_from_email.py
```diff
@@ -14,7 +14,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='siteconfiguration',
             name='from_email',
-            field=models.CharField(default='oscar@example.com', help_text='Address from which emails are sent.', max_length=255, verbose_name='From email'),
-            preserve_default=False,
+            field=models.CharField(default=b'oscar@example.com', help_text='Address from which emails are sent.', max_length=255, verbose_name='From email'),
```
mattdrayer: Hmm, this isn't doing what I would expect -- is there a way to make this a dynamic value?  Can I modify the actual migration file to pull settings.OSCAR_FROM_EMAIL?

ecommerce/core/models.py
```diff
@@ -65,7 +65,8 @@ class SiteConfiguration(models.Model):
         help_text=_('Address from which emails are sent.'),
         max_length=255,
         null=False,
```
clintonb: Skip the default and make the field nullable.

mattdrayer: Okay, I've updated the model to allow for null values and regenerated the migration file.

clintonb:
:+1:

ecommerce/core/migrations/0015_siteconfiguration_from_email.py
```diff
@@ -14,7 +14,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='siteconfiguration',
             name='from_email',
-            field=models.CharField(default='oscar@example.com', help_text='Address from which emails are sent.', max_length=255, verbose_name='From email'),
-            preserve_default=False,
+            field=models.CharField(help_text='Address from which emails are sent.', max_length=255, null=True, verbose_name='From email', blank=True),
```
mjfrey: Do you need a new migration?

PR #730, opened 2016-05-12
mattdrayer/remove-default-logo: Use edx-themes logo instead

mattdrayer:
@fredsmith @douglashall -- removing the logo file from the local 'themes' folder, because we are copying over this with an asset from the edx-themes repo during AMI provisioning and it is making Git unhappy.


douglashall:
👍

fredsmith:
Something to think about:  In the case of a default install of this software (like on a sandbox) we'll be copying theme data out of https://github.com/edx/sample-themes.  Will that process put the deleted asset back in the right place?

mattdrayer:
The deleted asset is for a theme override of the base image, which we haven't changed.  At some point we may add a sample theme but that would not be an edx-org asset and it wouldn't touch the base asset -- does that make sense?

clintonb:
How will this affect those of us who don't use edx/edx-themes?

mattdrayer:
@clintonb edx/edx-themes is supplying the edx.org theme for Otto, so if you configure your system to not use themes you'll get the Open Source theme/experience

@douglashall @fredsmith -- I discussed with @saleem-latif just now and we needed to remove a couple more files which are also supplied by edx-themes.

PR #733, opened 2016-05-13
[SOL-1772] Add course query preview endpoint

marjev:
This PR adds a proxy API endpoint that connects ecommerce and Course Catalog(Discovery) app.

We need to get the list of course runs that are a result of the filtering by ES query, done on the Course Catalog(Discovery) side. This will be used during the processes of coupon creation/editing/previewing.

JIRA story: https://openedx.atlassian.net/browse/SOL-1772

@mjfrey already did the 1st review. @clintonb, can you or someone from your team take a look as the 2nd reviewer?

FYI @rlucioni

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -8,6 +8,7 @@ import ddt
 from django.core.urlresolvers import reverse
 from django.db.utils import IntegrityError
 from django.test import RequestFactory
+import httpretty
 from oscar.apps.catalogue.categories import create_from_breadcrumbs
 from oscar.core.loading import get_class, get_model
 from oscar.test import factories
@@ -39,6 +40,7 @@ Voucher = get_model('voucher', 'Voucher')
 COUPONS_LINK = reverse('api:v2:coupons-list')
 
 
+@httpretty.activate
 @ddt.ddt
 class CouponViewSetTest(CouponMixin, CourseCatalogTestMixin, TestCase):
     """Unit tests for creating coupon order."""
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -41,11 +41,7 @@ Voucher = get_model('voucher', 'Voucher')
 
 
 class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
-    """Endpoint for creating coupons.
-
-    Creates a new coupon product, adds it to a basket and creates a
-    new order from that basket.
-    """
+    """ Coupon resource. """
     queryset = Product.objects.filter(product_class__name='Coupon')
     permission_classes = (IsAuthenticated, IsAdminUser)
     filter_backends = (filters.DjangoFilterBackend, )
--- a/ecommerce/tests/mixins.py
+++ b/ecommerce/tests/mixins.py
@@ -279,6 +279,18 @@ class TestServerUrlMixin(object):
         return 'http://{domain}{path}'.format(domain=site.domain, path=path)
 
 
+class ApiMockMixin(object):
+    """ Common Mocks for the API responses. """
+
+    def setUp(self):
+        super(ApiMockMixin, self).setUp()
+
+    def mock_api_error(self, error, url):
+        def callback(request, uri, headers):  # pylint: disable=unused-argument
+            raise error
+        httpretty.register_uri(httpretty.GET, url, body=callback, content_type='application/json')
+
+
 class LmsApiMockMixin(object):
     """ Mocks for the LMS API reponses. """
 
@@ -301,6 +313,36 @@ class LmsApiMockMixin(object):
         course_id = course.id if course else 'course-v1:test+test+test'
         course_url = get_lms_url('api/courses/v1/courses/{}/'.format(course_id))
         httpretty.register_uri(httpretty.GET, course_url, body=course_info_json, content_type='application/json')
+
+
+class CatalogPreviewMockMixin(object):
+    """ Mocks for the Course Discovery responses. """
+
+    def setUp(self):
+        super(CatalogPreviewMockMixin, self).setUp()
+
+    def mock_course_runs_contains_api_response(self, course_run=None, query=None):
+        """ Helper function to register an API endpoint for the course run information. """
+        course_run_info = {
+            'count': 1,
+            'results': [{
+                'key': course_run.id,
+                'title': course_run.name,
+            }] if course_run else [{
+                'key': 'course-v1:test+test+test',
+                'title': 'Test course',
+            }],
+        }
+        course_run_info_json = json.dumps(course_run_info)
+        course_run_url = '{}course_runs/?query={}'.format(
+            settings.COURSE_CATALOG_API_URL,
+            query if query else 'id:course*'
+        )
+        httpretty.register_uri(
+            httpretty.GET, course_run_url,
+            body=course_run_info_json,
+            content_type='application/json'
+        )
 
 
 class CouponMixin(object):
--- a/ecommerce/extensions/api/v2/tests/views/test_catalog.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_catalog.py
@@ -1,10 +1,20 @@
 import json
+import mock
 
+import ddt
+from django.conf import settings
 from django.core.urlresolvers import reverse
+from django.test import RequestFactory
+from edx_rest_api_client.client import EdxRestApiClient
+import httpretty
+from requests.exceptions import ConnectionError, Timeout
 from oscar.core.loading import get_model
+from slumber.exceptions import SlumberBaseException
 
 from ecommerce.extensions.api.serializers import ProductSerializer
+from ecommerce.extensions.api.v2.views.catalog import CatalogViewSet
 from ecommerce.extensions.api.v2.tests.views.mixins import CatalogMixin
+from ecommerce.tests.mixins import ApiMockMixin, CatalogPreviewMockMixin
 from ecommerce.tests.testcases import TestCase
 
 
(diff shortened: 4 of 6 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=733&repo=edx%2Fecommerce

ecommerce/settings/base.py
```diff
@@ -499,7 +499,7 @@
 RECEIPT_PAGE_PATH = '/commerce/checkout/receipt/'
 
 # URL for Course Catalog service
-COURSE_CATALOG_API_URL = 'http://localhost:8008/api/v1/'
+COURSE_CATALOG_API_URL = 'http://10.0.2.2:8010/api/v1'
```
marjev: @mjfrey Here are the latest changes I did. Just change the port here to the port that your Course Discovery service is running on.

ecommerce/settings/base.py
```diff
@@ -499,7 +499,7 @@
 RECEIPT_PAGE_PATH = '/commerce/checkout/receipt/'
 
 # URL for Course Catalog service
-COURSE_CATALOG_API_URL = 'http://localhost:8008/api/v1/'
+COURSE_CATALOG_API_URL = 'http://localhost:8008/api/v1'
```
mjfrey: No need to change this.

ecommerce/core/url_utils.py
```diff
@@ -71,6 +71,7 @@ def get_course_catalog_api_client(site):
         site.siteconfiguration.oauth_settings['SOCIAL_AUTH_EDX_OIDC_SECRET'],
         token_type='jwt'
     )
+
```
mjfrey: No need for this extra line.

ecommerce/tests/mixins.py
```diff
@@ -303,6 +315,36 @@ def mock_course_api_response(self, course=None):
         httpretty.register_uri(httpretty.GET, course_url, body=course_info_json, content_type='application/json')
 
 
+class CourseDiscoveryMockMixin(object):
```
mjfrey: CourseCatalogMockMixin

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -402,6 +402,31 @@ def perform_destroy(self, coupon):
...
+              multiple: false
+        """
+        query = request.GET.get('query', '')
+        if query:
+            try:
+                api = get_course_catalog_api_client(request.site)
+                response = api.course_runs.get(q=query)
+                return Response(response)
+            except (ConnectionError, SlumberBaseException, Timeout):
+                logger.error('Could not contact Course Discovery Service')
```
mjfrey: Unable to connect to Course Catalog service.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -20,7 +26,7 @@
 from ecommerce.extensions.voucher.models import CouponVouchers
 from ecommerce.invoice.models import Invoice
 from ecommerce.tests.factories import ProductFactory, SiteConfigurationFactory, SiteFactory
-from ecommerce.tests.mixins import CouponMixin, ThrottlingMixin
+from ecommerce.tests.mixins import ApiMockMixin, CourseDiscoveryMockMixin, CouponMixin, ThrottlingMixin
```
mjfrey: CourseCatalogMockMixin

ecommerce/settings/base.py
marjev: Done.

ecommerce/core/url_utils.py
marjev: Done.

ecommerce/tests/mixins.py
marjev: Done.

ecommerce/extensions/api/v2/views/coupons.py
marjev: Done.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
marjev: Done.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -402,6 +402,31 @@ def perform_destroy(self, coupon):
         StockRecord.objects.filter(product=coupon).delete()
         coupon.delete()
 
+    @action(is_for_list=True, methods=['get'])
```
vkaracic: Consider using [ `detail_route` ](http://www.django-rest-framework.org/api-guide/routers/#extra-link-and-actions) because this is pending deprecation. (https://github.com/chibisov/drf-extensions/blob/master/rest_framework_extensions/decorators.py#L25)

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -39,8 +45,9 @@
 COUPONS_LINK = reverse('api:v2:coupons-list')
 
 
+@httpretty.activate
 @ddt.ddt
-class CouponViewSetTest(CouponMixin, CourseCatalogTestMixin, TestCase):
+class CouponViewSetTest(CouponMixin, CourseCatalogTestMixin, CourseCatalogMockMixin, ApiMockMixin, TestCase):
```
vkaracic: You don't _need_ to change this but, even though it's the correct name, `CourseCatalogMockMixin` will introduce confusion with `CourseCatalogTestMixin`. Thoughts on renaming one of them?

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -266,6 +273,51 @@ def test_delete_coupon(self):
...
+            settings.COURSE_CATALOG_API_URL,
+            jwt='auth-token'
+        ))
+    )
+    def test_preview_catalog_course_discovery_service_not_available(self, error):
+        """Test catalog query preview when course discovery is not available."""
+        url = '/api/v2/coupons/preview/?query=id:course*'
+        factory = RequestFactory()
+        request = factory.get(url)
+        request.site = self.setup_site_configuration()
```
vkaracic: Perhaps DRY the upper four lines? `prepare_request(url)`

ecommerce/tests/mixins.py
```diff
@@ -303,6 +315,36 @@ def mock_course_api_response(self, course=None):
...
+    def setUp(self):
+        super(CourseCatalogMockMixin, self).setUp()
+
+    def mock_course_runs_contains_api_response(self, course_run=None, query=None):
+        """ Helper function to register an API endpoint for the course run information. """
+        course_run_info = {
+            'count': 1,
+            'results': [{
+                'key': course_run.id,
+                'title': course_run.name
```
vkaracic: Nit: comma at the end.

vkaracic:
I'm done with my review.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
marjev: @mjfrey any suggestions?

ecommerce/extensions/api/v2/views/coupons.py
marjev: That won't work. This has to be a list_route. Using detail_route will set URL to 
`^api/ ^v2/ ^coupons/(?P<pk>[^/.]+)/preview/$ [name='coupons-preview']` and the requirement here is for the URL to be `^api/ ^v2/ ^coupons/preview/$ [name='coupons-preview']`, which is a list_route. In other words, pk shouldn't be required here.

This is a known issue with DRF extensions and I this is a workaround of it. http://stackoverflow.com/questions/26792986/django-rest-framework-list-route-requires-pk

ecommerce/tests/mixins.py
marjev: Done.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
marjev: Done.

ecommerce/extensions/api/v2/views/coupons.py
vkaracic: When is this preview going to be used? If it's a general purpose preview, i.e. when a user enters a query in the UI and hits a preview button the results should be displayed, shouldn't this then be a standalone API view? Creating a simple `APIView` class, binding it in the urls to `r'^preview/$'`, and defining it's `get` method to do as the `preview` does here seems to simpler to me.

marjev: It's not a general purpose preview, but a preview of a coupon catalog content (list of courses). We already have a Catalog view so we could add it there, leave it under `/coupons/` or create a 3rd `APIView`, which again we need to name, which would be pointless to name anything other than `catalog` or `coupon`, if we want to keep things RESTful.

FYI @mjfrey

mjfrey: I don't mind either way.  Maybe moving it to Catalog makes more sense.  And I agree with Marko, this is not a general purpose API but one that is needed to proxy the catalog service when creating / editing coupons.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
mjfrey: can you combine the 2 mixins?

marjev: one is used for Course Catalogs that exist within ecommerce and the other one refers to Course Catalog service which is available via API. In theory, I can try and combine the two, just confirm that we want to do that. In case one is refactored or removed, I would like to avoid any future confusion people might have.

mjfrey: Alright -- do not combine.  Let's think of a better name.

mjfrey: How about,  CoursePreviewMockMixin

ecommerce/extensions/api/v2/views/coupons.py
marjev: Done.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
marjev: Done.

vkaracic:
LGTM

mjfrey:
👍 Please squash

PR #734, opened 2016-05-13
Remove voucher management command

vkaracic:
Revert "Management command to set SINGLE_USE vouchers global_applications"
This reverts commit 5c3260d82d0e6888f303fa9f273ef8976eeb781a.

@mjfrey since the broken vouchers have been fixed, we need to remove the command as it's obsolete.


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=734&repo=edx%2Fecommerce

mjfrey:
👍

PR #735, opened 2016-05-13
Disabled JWT audience claim verification

clintonb:
The value of this claim for JWTs issued by LMS is the OAuth client ID, which we don't have access to here in Otto. Verification is disabled until this issue can be resolved.

ECOM-4414


clintonb:
FYI @jibsheet @rlucioni @bderusha

clintonb:
While it is recommended that the audience claim be verified, it is not required. Additionally, the use of a signature (signed with a secret key) ensures we don't accept unauthorized requests.

rlucioni:
👍

bderusha:
👍

PR #736, opened 2016-05-15
mattdrayer/serialize-prodattr-code: Include 'code' attribute value property in API response

mattdrayer:
@vkaracic @mjfrey -- adding a piece of data to the Order API response for the receipt page to be able to determine the proper course_key value regardless of the product class -- this is used to populate the course name in the "We have received your payment for (course_name)." message.  see https://github.com/edx/edx-platform/pull/12466 for more information and the LMS-side of the fix.

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -66,10 +66,14 @@ class BillingAddressSerializer(serializers.ModelSerializer):
 class ProductAttributeValueSerializer(serializers.ModelSerializer):
     """ Serializer for ProductAttributeValue objects. """
     name = serializers.SerializerMethodField()
+    code = serializers.SerializerMethodField()
     value = serializers.SerializerMethodField()
 
     def get_name(self, instance):
         return instance.attribute.name
+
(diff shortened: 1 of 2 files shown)


vkaracic:
:+1:

clintonb:
This is not right. We already return all attributes.

vkaracic:
Yes but only the name and value of an attribute. The course key attribute has the code `course_key` on both `Seat` and `Enrollment code` product class attributes, while the name is different. This way there is no need for extra matchers on the LMS side.

clintonb:
I don't follow you. The API currently exposes all product attributes as key-value pairs.

vkaracic:
Yes, but the key is the attribute name which is different for the two course key attributes:
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/catalogue/migrations/0002_auto_20150223_1052.py#L30-L31
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/catalogue/migrations/0017_enrollment_code_product_class.py#L26-L27

clintonb:
:+1:

PR #737, opened 2016-05-16
Dynamic Course Catalog Coupons

vkaracic:
This PR introduces coupon support for dynamic course catalog queries.
- `Range` model is modified to store the query and dictionary of supported seat types. For example to create a coupon that supports all the verified course modes from MIT, you'd need to send:

```
catalog_query: 'org:MIT',
course_seat_types: {'verified': 1},
```

within the usual parameters to the coupon API endpoint. 
- In an effort to reduce complexity of the coupon flow, coupon product reusability has been dropped. Before when you would enter the same name and course ID's for a coupon the new vouchers would be appended to the old coupon. From now on a new coupon product is created every time.
- New helper function `get_seats_from_query` that retrieves the seat products from a course catalog query and it's accepted seat types. This is used in the coupon serializer and in the `Range` model.
- Coupon report is altered to display empty values for course-specific columns for dynamic course catalog coupons until there is a decision what exactly needs to displayed.


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=737

ecommerce/extensions/offer/models.py
```diff
@@ -1,13 +1,34 @@
...
+    catalog_query = models.CharField(max_length=255, blank=True, null=True)
+    course_seat_types = JSONField(null=True)
 
     def contains_product(self, product):
-        if self.catalog:
+        if self.catalog_query:
+            try:
+                response = get_course_discovery_client().\
+                    api.v1.course_runs.contains.get(query=self.catalog_query,
+                                                    course_ids=product.course_id,
```
vkaracic: I think this should be `course_run_ids` https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/views.py#L202

ecommerce/extensions/offer/models.py
```diff
@@ -1,13 +1,34 @@
...
+                                                    seat_types=",".join(self.course_seat_types))
```
vkaracic: @mjfrey what was the intent of this part? I don't see this being used in the course-discovery API endpoint.

mjfrey: This API just landed in course discovery.  It needs to be modified a bit as we no longer pass seat_types.  This call checks to see if the passed course runs are contained in the query.  We ask course discovery to do the work since the list of course runs could be quite large.

ecommerce/extensions/offer/models.py
mjfrey: yes it should.

ecommerce/extensions/offer/migrations/0004_auto_20160425_0859.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('offer', '0003_range_course_seats'),
+    ]
+
+    operations = [
+        migrations.RenameField(
```
mjfrey: Re-run the migrations since we do not need both of these.

ecommerce/extensions/offer/models.py
```diff
@@ -1,13 +1,43 @@
...
+    catalog_query = models.CharField(max_length=255, blank=True, null=True)
+    course_seat_types = JSONField(null=True)
+
+    def run_catalog_query(self):
+        """
+        Retrieve the results from running the query contained in catalog_query field.
+        """
+        site = get_current_request().site
+        try:
+            response = get_course_catalog_api_client(site).api.v1.course_runs.contains.get(
```
mjfrey: You can remove the api.vi as the settings for Course Catalog will contain that path.

ecommerce/extensions/offer/models.py
```diff
@@ -1,13 +1,43 @@
...
+                query=self.catalog_query,
+                course_run_ids=product.course_id
+            )
+        except:  # pylint: disable=bare-except
+            logger.error('ERROR: Could not contact Course Discovery Service')
```
mjfrey: Course Catalog service.

ecommerce/settings/test.py
```diff
@@ -124,8 +124,8 @@ def __getitem__(self, item):
 
 
 # Use production settings for asset compression so that asset compilation can be tested on the CI server.
-COMPRESS_ENABLED = True
-COMPRESS_OFFLINE = True
+COMPRESS_ENABLED = False
+COMPRESS_OFFLINE = False
```
mjfrey: Remove

vkaracic:
@mjfrey @marjev this PR is ready for review.

ecommerce/extensions/offer/models.py
```diff
@@ -1,13 +1,46 @@
...
+            )
+        except:  # pylint: disable=bare-except
+            logger.error('Could not contact Course Catalog Service.')
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+        return response
 
     def contains_product(self, product):
-        if self.catalog:
+        if self.catalog_query and self.course_seat_types:
+            if self.course_seat_types.get(product.attr.certificate_type.lower(), False):
```
marjev: Is there a `has` method to test if `course_seat_types` has a `product.attr.certificate_type.lower()` key? It would be easier to read that way.

ecommerce/extensions/api/serializers.py
```diff
@@ -442,9 +445,17 @@ def get_last_edited(self, obj):
...
+        request = self.context['request']
+        if offer.condition.range.catalog:
+            stockrecords = offer.condition.range.catalog.stock_records.all()
+            seats = Product.objects.filter(id__in=[sr.product.id for sr in stockrecords])
+        else:
+            seats = get_seats_from_query(
+                request.site,
+                offer.condition.range.catalog_query,
+                offer.condition.range.course_seat_types
+            )
```
marjev, line 458: nitpick/suggestion: Should we define `range=offer.condition.range` since we refer to it multiple times here?

marjev:
👍 (once my comments are addressed)

ecommerce/extensions/offer/models.py
vkaracic: I changed to use the `in` operator. Should be better now.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -105,21 +107,23 @@ def create(self, request, *args, **kwargs):
             # When a black-listed course mode is received raise an exception.
             # Audit modes do not have a certificate type and therefore will raise
             # an AttributeError exception.
-            seats = Product.objects.filter(stockrecords__id__in=stock_record_ids)
-            for seat in seats:
-                try:
-                    if seat.attr.certificate_type in settings.BLACK_LIST_COUPON_COURSE_MODES:
+            if stock_record_ids:
```
mjfrey, line 113: Should we even support stock_record_ids in creation?  What if we leave catalog support for redemption but now always use a query for all new coupons.  A single seat can be represented with the query "key:course_id".

vkaracic, line 113: I'd rather leave it as is for now, so that we have something to fall back to if the query coupons fail for any reason. Once this is proven to work as expected I'll gladly open a new PR and deprecate the old catalog from coupons.

mjfrey, line 113: We should create a ticket to deprecate the old catalog way.

vkaracic:
@clintonb @rlucioni @bderusha anyone of you guys have time to take a look?

ecommerce/coupons/utils.py
```diff
@@ -0,0 +1,35 @@
...
+        seat_types (JSON): JSON object of accepted seat types
+
+    Returns:
+        List of seat products retrieved from the course catalog query.
+    """
+    response = get_course_catalog_api_client(site).course_runs.get(q=query)
+    accepted_seat_types = [
+        seat_type for seat_type in seat_types if seat_types[seat_type]
+    ]
+    query_products = []
```
mjfrey, line 22: Why not return None?

ecommerce/extensions/api/v2/views/coupons.py
vkaracic, line 113: Ticket created https://openedx.atlassian.net/browse/SOL-1814

ecommerce/coupons/utils.py
```diff
@@ -0,0 +1,35 @@
...
+from oscar.core.loading import get_model
+
+from ecommerce.core.url_utils import get_course_catalog_api_client
+
+Product = get_model('catalogue', 'Product')
+
+
+def get_seats_from_query(site, query, seat_types):
+    """
+    Retrieve seats from a course catalog query.
```
mjfrey: and matching seat types.

ecommerce/coupons/utils.py
vkaracic, line 22: The list is only initialized here, it gets appended for every result in the results below.

mjfrey, line 22: I think my point was why not return None rather than am empty list if there are no results?

vkaracic, line 22: Hmm, I could add something like

```
if query_products:
    return query_products
return False
```

but it seems redundant.

ecommerce/extensions/api/serializers.py
```diff
@@ -442,9 +445,18 @@ def get_last_edited(self, obj):
...
+        _range = offer.condition.range
+        request = self.context['request']
+        if _range.catalog:
+            stockrecords = _range.catalog.stock_records.all()
+            seats = Product.objects.filter(id__in=[sr.product.id for sr in stockrecords])
+        else:
+            seats = get_seats_from_query(
+                request.site,
+                _range.catalog_query,
+                _range.course_seat_types
```
marjev, line 457: This is a list, but `get_seats_from_query` expects a JSON (a Dictionary, to be more precise) in place of `seat_types`. Upon integration, I run into `TypeError: list indices must be integers, not unicode`, since `_range.course_seat_types` has a value of `[u'Verified']`.

marjev, line 457: It looks like stick it mapping for checkboxes is actually a list and not a dictionary, so backbone sends a list instead of a dictionary. I'll customize this in my PR.

ecommerce/extensions/api/serializers.py
```diff
@@ -442,9 +445,18 @@ def get_last_edited(self, obj):
...
         offer = self.retrieve_offer(obj)
-        stockrecords = offer.condition.range.catalog.stock_records.all()
-        seats = Product.objects.filter(id__in=[sr.product.id for sr in stockrecords])
-        serializer = ProductSerializer(seats, many=True, context={'request': self.context['request']})
+        _range = offer.condition.range
+        request = self.context['request']
+        if _range.catalog:
+            stockrecords = _range.catalog.stock_records.all()
+            seats = Product.objects.filter(id__in=[sr.product.id for sr in stockrecords])
+        else:
```
vkaracic: ``elif _range.catalog_query and _range.course_seat_types``?

ecommerce/extensions/catalogue/utils.py
```diff
@@ -1,7 +1,6 @@
 from __future__ import unicode_literals
 
 from hashlib import md5
-
```
vkaracic, line 4: Wrong.

ecommerce/coupons/tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    def setUp(self):
+        super(CouponUtilsTests, self).setUp()
+        self.query = 'key:*'
+        self.seat_types = {'verified': 1}
+        self.course_id = 'course-v1:test+test+test'
+
+    def test_get_seat_from_query(self):
+        """ Verify right seat is returned. """
+        course, seat = self.create_course_and_seat(course_id=self.course_id)
+        self.mock_course_runs_api_response(query=self.query, course_run=course)
```
vkaracic: change name

ecommerce/coupons/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    response = get_course_catalog_api_client(site).course_runs.get(q=query)
+    accepted_seat_types = [
+        seat_type for seat_type in seat_types if seat_types[seat_type]
+    ]
+    query_products = []
+    for result in response['results']:
+        try:
+            product = Product.objects.get(
+                course_id=result['key'],
+                attribute_values__value_text__in=accepted_seat_types
```
vkaracic: Look for attribute name

ecommerce/extensions/offer/models.py
```diff
@@ -1,13 +1,46 @@
...
+from ecommerce.core.url_utils import get_course_catalog_api_client
+from ecommerce.coupons.utils import get_seats_from_query
+
+logger = logging.getLogger(__name__)
 
 
 class Range(AbstractRange):
     catalog = models.ForeignKey('catalogue.Catalog', blank=True, null=True, related_name='ranges')
+    catalog_query = models.CharField(max_length=255, blank=True, null=True)
+    course_seat_types = JSONField(null=True)
```
vkaracic: Change to string.

ecommerce/extensions/offer/models.py
```diff
@@ -1,13 +1,46 @@
...
+            )
+        except:  # pylint: disable=bare-except
+            logger.error('Could not contact Course Catalog Service.')
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+        return response
 
     def contains_product(self, product):
-        if self.catalog:
+        if self.catalog_query and self.course_seat_types:
+            if product.attr.certificate_type.lower() in self.course_seat_types:  # pylint: disable=unsupported-membership-test
```
vkaracic, line 31: Check if attr contains certificate_type

ecommerce/extensions/offer/models.py
```diff
@@ -1,13 +1,46 @@
...
+                response = self.run_catalog_query(product)
+                return ((response['course_runs'][product.course_id]) or
+                        super(Range, self).contains_product(product))  # pylint: disable=bad-super-call
```
vkaracic, line 36: Add a comment why we are calling super here.

ecommerce/extensions/offer/models.py
```diff
@@ -1,13 +1,46 @@
...
+        """
+        request = get_current_request()
+        try:
+            response = get_course_catalog_api_client(request.site).course_runs.contains.get(
+                query=self.catalog_query,
+                course_run_ids=product.course_id
+            )
+        except:  # pylint: disable=bare-except
+            logger.error('Could not contact Course Catalog Service.')
+            return Response(status=status.HTTP_400_BAD_REQUEST)
```
vkaracic: Throw exception.

ecommerce/extensions/offer/models.py
mjfrey, line 31: Need to ensure case is the same on both sides.  course_seat_types is passed with leading capital.  Ex. Verified vs. verified.

ecommerce/coupons/tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        settings.COURSE_CATALOG_API_URL,
+        jwt='auth-token'
+    ))
+)
+class CouponUtilsTests(CouponMixin, CourseCatalogTestMixin, CatalogPreviewMockMixin, TestCase):
+
+    def setUp(self):
+        super(CouponUtilsTests, self).setUp()
+        self.query = 'key:*'
+        self.seat_types = 'verified'
```
clintonb, line 26: This variable is mis-named.

ecommerce/coupons/utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def get_seats_from_query(site, query, seat_types):
+    """
+    Retrieve seats from a course catalog query and matching seat types.
+
+    Arguments:
+        site (Site): current site
+        query (str): course catalog query
+        seat_types (JSON): JSON object of accepted seat types
```
clintonb, line 16: This is incorrect.

PR #739, opened 2016-05-16
mattdrayer/e2etest-receipt-msg: Assert course name appears in thank you message

mattdrayer:
@rlucioni @mjfrey -- I've added a check for a valid "Thank you" message to protect against regressions -- this check should pass after the current RC lands in the staging environment with the course name fix (maybe it already has).

### DIFF
--- a/acceptance_tests/mixins.py
+++ b/acceptance_tests/mixins.py
@@ -320,6 +320,12 @@ class PaymentMixin(object):
         actual = [cell.text for cell in cells]
         self.assertListEqual(actual, expected)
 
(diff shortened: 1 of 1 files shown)


mattdrayer:
Just a quick update -- I've run the e2e tests from my local machine against the staging environment and the check passes, so it looks like we're in good shape.

mattdrayer:
@rlucioni @mjfrey -- unless you guys have any comments/feedback re: this PR I'm going to merge it in the morning.

mjfrey:
👍

PR #740, opened 2016-05-16
Updated JWT decode handler to support multiple signing keys

clintonb:
ECOM-4414

### DIFF
--- a/ecommerce/extensions/api/handlers.py
+++ b/ecommerce/extensions/api/handlers.py
@@ -29,22 +29,30 @@ def jwt_decode_handler(token):
     # JWT_ISSUERS is not one of DRF-JWT's default settings, and cannot be accessed
     # using the `api_settings` object without overriding DRF-JWT's defaults.
     issuers = settings.JWT_AUTH['JWT_ISSUERS']
-    for issuer in issuers:
-        try:
-            return jwt.decode(
-                token,
-                api_settings.JWT_SECRET_KEY,
-                api_settings.JWT_VERIFY,
-                options=options,
-                leeway=api_settings.JWT_LEEWAY,
-                audience=api_settings.JWT_AUDIENCE,
-                issuer=issuer,
-                algorithms=[api_settings.JWT_ALGORITHM]
-            )
-        except jwt.InvalidIssuerError:
-            pass
-        except jwt.InvalidTokenError:
-            logger.exception('JWT decode failed!')
-            raise
+    secret_keys = settings.JWT_AUTH['JWT_SECRET_KEYS'] or (api_settings.JWT_SECRET_KEY,)
 
-    raise jwt.InvalidIssuerError
+    # TODO (CCB): The usage of multiple issuers complicates matters. We should only have one issuer.
+    # Update ecommerce-worker to properly use client credentials, and remove the internal loop. (ECOM-4477)
+    for secret_key in secret_keys:
+        for issuer in issuers:
+            try:
+                return jwt.decode(
+                    token,
+                    secret_key,
+                    api_settings.JWT_VERIFY,
+                    options=options,
+                    leeway=api_settings.JWT_LEEWAY,
+                    audience=api_settings.JWT_AUDIENCE,
+                    issuer=issuer,
+                    algorithms=[api_settings.JWT_ALGORITHM]
+                )
+            except jwt.InvalidIssuerError:
+                # Ignore these errors since we have multiple issuers
+                pass
+            except jwt.DecodeError:
+                # Ignore these errors since we have multiple signing keys
+                pass
+            except jwt.InvalidTokenError:
+                logger.exception('JWT decode failed!')
+
+    raise jwt.InvalidTokenError('All combinations of JWT issuers and secret keys failed to validate the token.')
--- a/ecommerce/extensions/api/tests/test_handlers.py
+++ b/ecommerce/extensions/api/tests/test_handlers.py
@@ -1,0 +1,71 @@
+""" Tests for handler functions. """
+from time import time
+
+from django.conf import settings
+from django.test import TestCase
+import jwt
+import mock
+from oscar.test.factories import UserFactory
+
+from ecommerce.extensions.api.handlers import jwt_decode_handler
+
+ISSUERS = ('test-issuer', 'another-issuer',)
+SIGNING_KEYS = ('insecure-secret-key', 'secret', 'another-secret',)
+
+
+def generate_jwt_token(payload, signing_key=None):
+    """Generate a valid JWT token for authenticated requests."""
+    signing_key = signing_key or settings.JWT_AUTH['JWT_SECRET_KEY']
+    return jwt.encode(payload, signing_key).decode('utf-8')
+
+
+def generate_jwt_payload(user):
+    """Generate a valid JWT payload given a user."""
+    now = int(time())
+    ttl = 5
+    return {
+        'iss': settings.JWT_AUTH['JWT_ISSUERS'][0],
+        'username': user.username,
+        'email': user.email,
+        'iat': now,
+        'exp': now + ttl
+    }
+
+
+class JWTDecodeHandlerTests(TestCase):
+    """ Tests for the `jwt_decode_handler` utility function. """
+
+    def setUp(self):
+        super(JWTDecodeHandlerTests, self).setUp()
+        self.user = UserFactory()
+        self.payload = generate_jwt_payload(self.user)
+        self.jwt = generate_jwt_token(self.payload)
+
+    def test_decode_success(self):
+        self.assertEqual(jwt_decode_handler(self.jwt), self.payload)
+
+    def test_decode_success_with_multiple_issuers(self):
+        settings.JWT_AUTH['JWT_ISSUERS'] = ISSUERS
+
+        for issuer in ISSUERS:
+            self.payload['iss'] = issuer
+            token = generate_jwt_token(self.payload)
+            self.assertEqual(jwt_decode_handler(token), self.payload)
+
+    def test_decode_success_with_multiple_signing_keys(self):
+        settings.JWT_AUTH['JWT_SECRET_KEYS'] = SIGNING_KEYS
+
+        for signing_key in SIGNING_KEYS:
+            token = generate_jwt_token(self.payload, signing_key)
+            self.assertEqual(jwt_decode_handler(token), self.payload)
+
+    def test_decode_error(self):
+        # Update the payload to ensure a validation error
+        self.payload['exp'] = 0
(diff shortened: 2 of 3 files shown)


clintonb:
@rlucioni please review

ecommerce/extensions/api/tests/test_handlers.py
```diff
@@ -0,0 +1,73 @@
...
+    """ Tests for the `jwt_decode_handler` utility function. """
+
+    def setUp(self):
+        super(JWTDecodeHandlerTests, self).setUp()
+        self.user = UserFactory()
+        self.payload = generate_jwt_payload(self.user)
+        self.jwt = generate_jwt_token(self.payload)
+
+    def test_decode_success(self):
+        self.assertDictEqual(jwt_decode_handler(self.jwt), self.payload)
```
rlucioni: You can just use `assertEqual`; it'll do the right thing.

ecommerce/extensions/api/handlers.py
```diff
@@ -29,22 +29,30 @@ def jwt_decode_handler(token):
...
+                    algorithms=[api_settings.JWT_ALGORITHM]
+                )
+            except jwt.InvalidIssuerError:
+                # Ignore these errors since we have multiple issuers
+                pass
+            except jwt.DecodeError:
+                # Ignore these errors since we have multiple signing keys
+                pass
+            except jwt.InvalidTokenError as exc:
+                logger.debug('JWT decode failed: %s', exc)
```
rlucioni: I'd log this at the error level. If this happens in production, we want to know.

ecommerce/extensions/api/tests/test_handlers.py
clintonb: I prefer the explicitness.

ecommerce/extensions/api/handlers.py
clintonb: Not necessarily. With two issuers and two secret keys, we would potentially see an error logged three times for a valid request. That's unnecessarily noisy. Once we get to single issuer, it will make sense to bump this to the error level.

ecommerce/extensions/api/tests/test_handlers.py
rlucioni: Fair enough. FWIW, the unittest [docs](https://docs.python.org/2/library/unittest.html#unittest.TestCase.addTypeEqualityFunc) discourage direct invocation of type-specific methods.

ecommerce/extensions/api/handlers.py
rlucioni: But you're catching the `InvalidIssuerError` and `DecodeError` separately and suppressing logging in those cases. This message would only be logged if other parts of the token's payload are invalid.

clintonb: Ah. Good point.

rlucioni:
👍

vkaracic:
When I go to redeem an enrollment coupon locally, I get this error on ecommerce when the order page gets loaded:

```
2016-05-20 08:05:06,869 ERROR 5977 [django.request] base.py:256 - Internal Server Error: /api/v2/orders/EDX-100004/
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/viewsets.py", line 87, in view
    return self.dispatch(request, *args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/views.py", line 466, in dispatch
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/views.py", line 454, in dispatch
    self.initial(request, *args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/views.py", line 376, in initial
    self.perform_authentication(request)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/views.py", line 310, in perform_authentication
    request.user
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/request.py", line 444, in __getattribute__
    return super(Request, self).__getattribute__(attr)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/request.py", line 219, in user
    self._authenticate()
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/request.py", line 407, in _authenticate
    user_auth_tuple = authenticator.authenticate(self)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/edx_rest_framework_extensions/authentication.py", line 179, in authenticate
    return super(JwtAuthentication, self).authenticate(request)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework_jwt/authentication.py", line 33, in authenticate
    payload = jwt_decode_handler(jwt_value)
  File "/edx/app/ecommerce/ecommerce/ecommerce/extensions/api/handlers.py", line 26, in jwt_decode_handler
    'verify_aud': settings.JWT_AUTH['JWT_VERIFY_AUDIENCE'],
KeyError: 'JWT_VERIFY_AUDIENCE'
```

This happens on the latest ecommerce (with a fresh db) and LMS masters. Do I have to enable/install something?

rlucioni:
@vkaracic look at the error message. A key is missing from your JWT_AUTH dictionary. Check your settings and make sure that the key is present.

vkaracic:
As I said, this is from latest ecommerce and LMS masters, and the only setting that I'm overriding is the address for course catalog. I tried removing all the .pyc files but that doesn't work either. This is the error from the LMS side:

```
HttpServerError: Server Error 500: http://localhost:8002/api/v2/orders/EDX-100007/
2016-05-21 02:58:58,942 ERROR 4977 [django.request] base.py:256 - Internal Server Error: /api/commerce/v1/orders/EDX-100007/
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/rest_framework/views.py", line 466, in dispatch
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/rest_framework/views.py", line 463, in dispatch
  File "/edx/app/edxapp/edx-platform/lms/djangoapps/commerce/api/v1/views.py", line 72, in get
    order = ecommerce_api_client(request.user).orders(number).get()
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/slumber/__init__.py", line 155, in get
    resp = self._request("GET", params=kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/slumber/__init__.py", line 103, in _request
    raise exceptions.HttpServerError("Server Error %s: %s" % (resp.status_code, url), response=resp, content=resp.content)
HttpServerError: Server Error 500: http://localhost:8002/api/v2/orders/EDX-100007/
```

A colleague here is having the same issue.

I can see `JWT_VERIFY_AUDIENCE` in my settings so I'm guess I missed to enable something or?

clintonb:
Testing locally, I need only set values for `JWT_SECRET_KEYS` and `JWT_ISSUERS` in my `private.py` file:

``` python
JWT_AUTH.update({
    'JWT_SECRET_KEYS': ('edx',),
    'JWT_ISSUERS': ('edx',),
})
```

If you are seeing errors, please check all of your settings files—base.py, local.py, private.py—for modifications. Make sure you are calling `JWT_AUTH.update()` instead of reassigning the value.

vkaracic:
That doesn't help me at all. I'm using the exact settings that master has currently and now two colleagues and I are still seeing the same issue, so we're definitely missing something. Can you confirm that you don't get that error when you:
1 - add a verified (can probably be a professional as well) seat to your basket
2 - apply an enrollment code to it
3 - place the order
4 - (automatically) load the receipt page

I'm getting that issue at step 4.

clintonb:
Stage works as expected. You cannot possibly be using "the exact settings that master has" because you wouldn't be able to make payments with those settings. Please share your logs and settings for the ecommerce service.

vkaracic:
Correct, stage works. There is no need for a payment processor on free baskets. Here's my `ecommerce.yml`: https://gist.github.com/vkaracic/dcaeca9480c5ef8e6c4bbe32618e6810
and here's my `JWT_AUTH` in `settings/base.py`:

```
JWT_AUTH = {
    'JWT_SECRET_KEY': None,
    'JWT_ALGORITHM': 'HS256',
    'JWT_VERIFY_EXPIRATION': True,
    'JWT_LEEWAY': 1,
    'JWT_DECODE_HANDLER': 'ecommerce.extensions.api.handlers.jwt_decode_handler',
    # These settings are NOT part of DRF-JWT's defaults.
    'JWT_ISSUERS': (),
    # NOTE (CCB): This is temporarily set to False until we decide what values are acceptable.
    'JWT_VERIFY_AUDIENCE': False,
    'JWT_SECRET_KEYS': (),
}
```

and this is all there is in my `private.py`:

```
COURSE_CATALOG_API_URL = 'http://10.0.2.2:8008/api/v1/'

settings.JWT_AUTH.update({
    'JWT_SECRET_KEYS': ('edx',),
    'JWT_ISSUERS': ('edx',),
})
```

There is no log in my `/edx/var/log/ecommerce/` directory. The full traceback of LMS and ecommerce are posted above.

PR #715, opened 2016-05-05
mattdrayer/override-oscar-from-email: Add new SiteConfiguration field

mattdrayer:
@mjfrey @clintonb -- adding a new "from_email" to the SiteConfiguration, plus a new customer extension and overrides for customer/utils.py and customer/alerts/utils.py

@vkaracic FYI

PR #741, opened 2016-05-17
Add missing customer and voucher migrations.

vkaracic:
On a fresh install from current master these migrations are missing.
The voucher migration removes the blank parameter from orderlinevouchers.vouchers.
The customer migration changes the editable parameter to False.

The voucher migration is on me, because I removed blank parameter and forgot to make migrations. Don't know about the customer one though. @mattdrayer do you have an idea?

FYI @clintonb

### DIFF
--- a/ecommerce/extensions/customer/migrations/0002_auto_20160517_0930.py
+++ b/ecommerce/extensions/customer/migrations/0002_auto_20160517_0930.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import oscar.models.fields.autoslugfield
+import django.core.validators
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('customer', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='communicationeventtype',
+            name='code',
+            field=oscar.models.fields.autoslugfield.AutoSlugField(populate_from=b'name', validators=[django.core.validators.RegexValidator(regex=b'^[a-zA-Z_][0-9a-zA-Z_]*$', message="Code can only contain the letters a-z, A-Z, digits, and underscores, and can't start with a digit.")], editable=False, max_length=128, separator='_', blank=True, help_text='Code used for looking up this event programmatically', unique=True, verbose_name='Code'),
+        ),
+    ]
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=741

mattdrayer:
Hmm, unsure about the CommunicationEventType migration, unless it's a model change that slipped in with our bulk purchasing work and we missed it?

vkaracic:
That's the strange thing, because there are no changes on `models.py` on `customer` app.

mattdrayer:
Hmm, could it be due to an extension of a Django model -- we had to extend several things during the course of the bulk purchasing work -- maybe there's a migration required due to that?

vkaracic:
This is the only change in the `customer` app that I can see:
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/customer/utils.py

mattdrayer:
Could PR #715 be the culprit?  See these pertinent lines:

https://github.com/edx/ecommerce/pull/715/files#diff-c0f325dd8b60d158a9feeb873f2a1bd8R19
https://github.com/edx/ecommerce/pull/715/files#diff-e6a588c4ca7b76cc0a37e6046738799fR1
https://github.com/edx/ecommerce/pull/715/files#diff-5cf31f05647d795d6d580517c507081dR2
https://github.com/edx/ecommerce/pull/715/files#diff-1234957db735399f1fd19d4a3bc401bbR11
https://github.com/edx/ecommerce/pull/715/files#diff-450d651868a8f1ab20c1dad5c5582c68R30

vkaracic:
I think we have our culprit here: https://github.com/edx/ecommerce/pull/715/files#diff-e6a588c4ca7b76cc0a37e6046738799fR21
Interestingly the same thing happens here: https://github.com/django-oscar/django-oscar/pull/1821

clintonb:
Please work to resolve this. It is holding up all deployments to production.

mattdrayer:
@clintonb what is the proper course of action here -- it seems that by extending a particular Oscar model the system is now saying that a migration needs to be run -- should we generate this migration file and submit as a separate PR?

clintonb:
It can be part of this PR.

Clinton Blackburn

mattdrayer:
Adding to the PR doesn't seem like it would unblock production deployments until the PR merges -- is this PR ready to merge?

mattdrayer:
AH, this PR is supposed to merge the missing migrations :)

mattdrayer:
@clintonb I will rebase and merge this PR, which appears to resolve the migration issues, after I get a +1

clintonb:
👍

mattdrayer:
Merged -- please let me know if you continue to experience migration-related issues when shipping the ecommerce project.

vkaracic:
:+1:

PR #744, opened 2016-05-17
Utility to access course discovery via edx-rest-api

mjfrey:
Utility to access course discovery service.  This is needed to support dynamic coupons that use course discovery as the backend range consisting of a query.

@vkaracic @marjev

### DIFF
--- a/ecommerce/core/tests/test_url_utils.py
+++ b/ecommerce/core/tests/test_url_utils.py
@@ -1,7 +1,13 @@
 import mock
 
+from django.conf import settings
+from edx_rest_api_client.auth import SuppliedJwtAuth
+from edx_rest_api_client.client import EdxRestApiClient
+
 from ecommerce.core.exceptions import MissingRequestError
-from ecommerce.core.url_utils import get_ecommerce_url, get_lms_url
+from ecommerce.core.url_utils import(get_course_catalog_api_client,
+                                     get_ecommerce_url, get_lms_url,
+                                     get_oauth2_provider_url)
 from ecommerce.tests.testcases import TestCase
 
 
@@ -15,3 +21,17 @@ class UrlUtilityTests(TestCase):
     def test_get_lms_url_with_no_current_request(self):
         with self.assertRaises(MissingRequestError):
             get_lms_url()
+
+    @mock.patch.object(EdxRestApiClient, 'get_oauth_access_token')
+    def test_get_course_catalog_api_client(self, mock_method):
+        mock_method.return_value = ('auth-token', 'now')
+        api = get_course_catalog_api_client(self.site)
+
+        mock_method.assert_called_with('{root}/access_token'.format(root=get_oauth2_provider_url()),
+                                       self.site.siteconfiguration.oauth_settings['SOCIAL_AUTH_EDX_OIDC_KEY'],
+                                       self.site.siteconfiguration.oauth_settings['SOCIAL_AUTH_EDX_OIDC_SECRET'],
+                                       token_type='jwt')
+        api_session = api._store['session']  # pylint: disable=protected-access
+        self.assertEqual('auth-token', api_session.auth.token)
+        self.assertEqual(SuppliedJwtAuth, type(getattr(api_session, 'auth')))
+        self.assertEqual(settings.COURSE_CATALOG_API_URL, api._store['base_url'])  # pylint: disable=protected-access
--- a/ecommerce/core/url_utils.py
+++ b/ecommerce/core/url_utils.py
@@ -1,5 +1,7 @@
 from urlparse import urljoin
 
+from django.conf import settings
+from edx_rest_api_client.client import EdxRestApiClient
 from threadlocals.threadlocals import get_current_request
 
 from ecommerce.core.exceptions import MissingRequestError
@@ -50,3 +52,27 @@ def get_lms_url(path=''):
 
 def get_oauth2_provider_url():
     return get_lms_url('/oauth2')
+
+
+def get_course_catalog_api_client(site):
+    """
+    Returns an API client to access the Course Catalog service.
+
+    Arguments:
+        site (Site): The site for which to retrieve settings.
+
+    Returns:
+        EdxRestApiClient: The client to access the Course Catalog service.
+    """
+
+    access_token, __ = EdxRestApiClient.get_oauth_access_token(
+        '{root}/access_token'.format(root=get_oauth2_provider_url()),
+        site.siteconfiguration.oauth_settings['SOCIAL_AUTH_EDX_OIDC_KEY'],
+        site.siteconfiguration.oauth_settings['SOCIAL_AUTH_EDX_OIDC_SECRET'],
+        token_type='jwt'
+    )
+    course_catalog_client = EdxRestApiClient(
+        settings.COURSE_CATALOG_API_URL,
+        jwt=access_token
+    )
+    return course_catalog_client
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -498,6 +498,9 @@ SUPPORT_URL = 'SET-ME-PLEASE'
 # Path to the receipt page
 RECEIPT_PAGE_PATH = '/commerce/checkout/receipt/'
 
+# URL for Course Catalog service
+COURSE_CATALOG_API_URL = 'http://localhost:8008/api/v1/'
+
 # Black-listed course modes not allowed to create coupons with
 BLACK_LIST_COUPON_COURSE_MODES = [u'audit', u'honor']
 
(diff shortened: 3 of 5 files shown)


mjfrey:
@rlucioni can you take a look?

ecommerce/core/url_utils.py
```diff
@@ -50,3 +51,27 @@ def get_lms_url(path=''):
...
+
+def get_course_discovery_client():
+    """
+    Returns an API client to access the course discovery service.
+
+    Raises:
+        MissingRequestError: If the current ecommerce site is not in threadlocal storage
+    """
+
+    request = get_current_request()
```
clintonb: Can we stop using this unless absolutely necessary? This sort of dependency injection is prone to issues that we should avoid if possible.

ecommerce/settings/base.py
```diff
@@ -498,6 +498,9 @@
 # Path to the receipt page
 RECEIPT_PAGE_PATH = '/commerce/checkout/receipt/'
 
+# URL root for Course Discovery Service
+COURSE_DISCOVERY_ROOT = 'SET-ME-PLEASE'
```
clintonb: Provide a sane default.

mjfrey: I thought this was the convention.

ecommerce/core/url_utils.py
mjfrey: I am not sure how to get at the request from the range model.

ecommerce/settings/base.py
clintonb: For passwords, maybe. We generally point to localhost for external services.

ecommerce/core/url_utils.py
clintonb: You don't. This function should accept a `Site` argument. How you get the `Site` is up to the caller. If it comes from `get_current_request()`, so be it; but, we should avoid making utility function dependent on dependency injection.

It would be preferable to group this PR with the PR for code that will actually use this function. Breaking down PRs by individual functions in this manner don't give the proper context for reviews.

mjfrey: The reason for this PR is to unblock others needing this utility.  If we could land this we would not have to serialize the PR's

clintonb: Carry on with this PR. The `Site` should be passed as an argument.

ecommerce/core/url_utils.py
```diff
@@ -1,7 +1,8 @@
 from urlparse import urljoin
 
+from django.conf import settings
 from threadlocals.threadlocals import get_current_request
-
+from edx_rest_api_client.client import EdxRestApiClient
```
clintonb, line 4: This is a third-party requirement and should be grouped appropriately.

ecommerce/core/tests/test_url_utils.py
```diff
@@ -15,3 +15,8 @@ def test_get_ecommerce_url_with_no_current_request(self):
     def test_get_lms_url_with_no_current_request(self):
         with self.assertRaises(MissingRequestError):
             get_lms_url()
+
```
clintonb, line 24: You need a test for the positive case.

mjfrey:
@clintonb comments addressed.

ecommerce/settings/base.py
```diff
@@ -498,6 +498,9 @@
 # Path to the receipt page
 RECEIPT_PAGE_PATH = '/commerce/checkout/receipt/'
 
+# URL root for Course Discovery Service
+COURSE_DISCOVERY_ROOT = 'localhost:8003'
```
clintonb: `'http://localhost:8008/api/v1/'`

clintonb: Also, use the new/correct name for the service: `COURSE_CATALOG_API_URL`.

ecommerce/core/url_utils.py
```diff
@@ -50,3 +52,27 @@ def get_lms_url(path=''):
 
 def get_oauth2_provider_url():
     return get_lms_url('/oauth2')
+
+
+def get_course_discovery_client(site):
```
clintonb: `get_course_catalog_api_client`

ecommerce/core/url_utils.py
```diff
@@ -50,3 +52,27 @@ def get_lms_url(path=''):
...
+    """
+    Returns an API client to access the course discovery service.
```
clintonb: Course Catalog service

ecommerce/core/url_utils.py
```diff
@@ -50,3 +52,27 @@ def get_lms_url(path=''):
...
+
+    Arguments:
+        site (Site): The site for which to retrieve settings.
+
+    Returns:
+        course_discovery_client (EdxRestApiClient): The client to access the course discovery service.
```
clintonb: No need to specify the variable name.

`EdxRestApiClient: Authenticated Course Catalog API client.`

ecommerce/core/tests/test_url_utils.py
clintonb, line 24: This has not been addressed. You need to verify the following for `get_course_discovery_client()`:
1. The function makes an HTTP call to the correct endpoint with the correct parameters. (Use `httpretty` to mock the endpoint and the returned token.)
2. The returned object has its root URL set to the correct URL of the catalog service. (Look at `client._store['base_url']`.)
3. Verify the client has the correct authentication class and token. (Confirm the type of `client._store['session'].auth` and the value of it's `token` member.)

mjfrey, line 24: I am using mock already.  Is httpretty necessary?

clintonb, line 24: Ah. No, I didn't see your change.

clintonb, line 24: You should still confirm the mocked method is called with the correct arguments.

mjfrey, line 24: Done

clintonb, line 24: Not quite. You still need to confirm `get_oauth_access_token` was called with the correct arguments, and `api._store['session'].auth.token` is set to the mocked return value.

mjfrey, line 24: Done done.

ecommerce/core/tests/test_url_utils.py
```diff
@@ -15,3 +21,16 @@ def test_get_ecommerce_url_with_no_current_request(self):
...
+    @mock.patch.object(EdxRestApiClient, 'get_oauth_access_token')
+    def test_get_course_catalog_api_client(self, mock_method):
+        mock_method.return_value = ('auth-token', 'now')
+        api = get_course_catalog_api_client(self.site)
+
+        mock_method.assert_called_with('{root}/access_token'.format(root=get_oauth2_provider_url()),
+                                       self.site.siteconfiguration.oauth_settings['SOCIAL_AUTH_EDX_OIDC_KEY'],
+                                       self.site.siteconfiguration.oauth_settings['SOCIAL_AUTH_EDX_OIDC_SECRET'],
+                                       token_type='jwt')
+        self.assertEqual('auth-token', api._store['session'].auth.token)  # pylint: disable=protected-access
```
clintonb: Minor: Create a variable for `api._store['session'].auth` instead of duplicating the call here and below.

clintonb:
👍

ecommerce/core/tests/test_url_utils.py
```diff
@@ -15,3 +21,17 @@ def test_get_ecommerce_url_with_no_current_request(self):
...
+    @mock.patch.object(EdxRestApiClient, 'get_oauth_access_token')
+    def test_get_course_catalog_api_client(self, mock_method):
+        mock_method.return_value = ('auth-token', 'now')
+        api = get_course_catalog_api_client(self.site)
+
+        mock_method.assert_called_with('{root}/access_token'.format(root=get_oauth2_provider_url()),
+                                       self.site.siteconfiguration.oauth_settings['SOCIAL_AUTH_EDX_OIDC_KEY'],
+                                       self.site.siteconfiguration.oauth_settings['SOCIAL_AUTH_EDX_OIDC_SECRET'],
+                                       token_type='jwt')
+        api_session = api._store['session']  # pylint: disable=protected-access
```
clintonb, line 34: `auth = api._store['session'].auth`

ecommerce/core/tests/test_url_utils.py
```diff
@@ -15,3 +21,17 @@ def test_get_ecommerce_url_with_no_current_request(self):
...
+        self.assertEqual('auth-token', api_session.auth.token)
```
clintonb, line 35: `auth.token`

ecommerce/core/tests/test_url_utils.py
```diff
@@ -15,3 +21,17 @@ def test_get_ecommerce_url_with_no_current_request(self):
...
+        self.assertEqual(SuppliedJwtAuth, type(getattr(api_session, 'auth')))
```
clintonb, line 36: `type(auth)`

PR #745, opened 2016-05-17
SOL-1774 Removed blank=False from SiteConfiguration.oauth_settings

douglashall:
SiteConfiguration.oauth_settings should not be a required field because we default back to settings for OAuth configuration for the default site. This field being required was causing issues with saving SiteConfiguration models in Django admin.

@clintonb Please review.

### DIFF
--- a/ecommerce/core/migrations/0016_auto_20160517_2021.py
+++ b/ecommerce/core/migrations/0016_auto_20160517_2021.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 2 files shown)


clintonb:
I disagree. We should be completely multi-tenant. Ultimately, the settings need to go away. Every site should be configured in the same fashion rather than have one site configured via settings and the others via database. Two paths to configuration means two paths that need testing and two potential sources of error.

douglashall:
That's fine with me, though ultimately I think the OAuth secret should not be kept in the database. We should consider that with any future refactoring. Closing this one out.

PR #746, opened 2016-05-18
Return 400 Bad Request for invalid seat type when creating Coupons.

mjfrey:
Do not throw an exception that is not handled as it creates noise in the logs.  Just return a 400 Bad Request instead.


mjfrey:
@clintonb

clintonb:
Tests need to be updated/added.

mjfrey:
Test updated

clintonb:
Squash your commits before merging.
👍

PR #747, opened 2016-05-18
WL-455: Discount Field hidden for bulk seat purchase.

malikshahzad228:


malikshahzad228:
@mattdrayer FYI

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -65,7 +65,7 @@
 
         <div class="total">
             <div class="row">
-            {% if not line_data.enrollment_code %}
+            {% if not formset_lines_data.0.1.enrollment_code %}
```
mattdrayer, line 68: Rather than introducing a direct-index reference (too brittle), we should instead search the array in the view module for an enrollment code item and pass a context parameter switch to the template

malikshahzad228:
@mattdrayer we don't have to merge this now..
Backend was same, So in another PR this is done.

New PR: https://github.com/edx/ecommerce/pull/750

mattdrayer:
Okay, if this PR is a no-op, let's go ahead and close it.

PR #748, opened 2016-05-18
WL-456: Seat Quantity field validation added.

malikshahzad228:
Seat Quantity Filed min value set to 1 instead of 0 (Value passed from views), so html field validation caught this, and prevent the form submitted for the values less then 1.

Changes for MIT Theme: https://github.com/edx/edx-themes/pull/35

### DIFF
--- a/ecommerce/templates/oscar/basket/partials/basket_content.html
+++ b/ecommerce/templates/oscar/basket/partials/basket_content.html
@@ -34,7 +34,7 @@
                             <div class="col-lg-2 col-md-2 col-sm-6 form-inline">
                                 <div class="checkout-quantity form-group">
                                     <div class="input-group  {% if form.errors %}error{% endif %}">
-                                        {% render_field form.quantity class+="quantity form-control" %}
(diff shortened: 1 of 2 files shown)


ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -34,8 +34,8 @@
                             <div class="col-sm-2">
                                 <div class="checkout-quantity">
                                     <div class="input-group  {% if form.errors %}error{% endif %}">
-                                        {% render_field form.quantity class+="quantity" %}
-                                    <button class="btn btn-default" type="submit" data-loading-text="{% trans 'Updating...' %}">{% trans "Update" %}</button>
+                                        {% render_field form.quantity class+="quantity" min="1" %}
```
mattdrayer: the minimum value should not be encoded in the template, it should be provided by the view

mattdrayer:
Looks like we need a rebase, but otherwise 👍 -- @saleem-latif or @asadiqbal08 if one of you can take a look as well, that would be great.

saleem-latif:
👍

PR #749, opened 2016-05-19
MAYN-234: Update Otto Theming app so that COMPREHENSIVE_THEME_DIRS is a list of  directories

saleem-latif:
Hi @mattdrayer , @clintonb  

Kindly review this PR, it contains changes for MAYN-234.

**Description of MAYN-234:**
`COMPREHENSIVE_THEME_DIR` setting is used by theming app to figure out where themes are placed, currently we can specify only one theme directory and all themes must be placed inside this directory. This restriction reduces flexibility during theme setup via configuration. 
Convert `COMPREHENSIVE_THEME_DIR` to a list of directories so that multiple themes directories can be used with one or more theme in each directory.
Updating `COMPREHENSIVE_THEME_DIR` to list will require changes in 
1. Rename `COMPREHENSIVE_THEME_DIR` to `COMPREHENSIVE_THEME_DIRS`
2. Static files finders and storage
3. template loaders
4. template compressors
5. theming helpers
6. management commands for compiling theme sass

**NOTE:**
This PR is dependent on configuration [PR#3065](https://github.com/edx/configuration/pull/3065) (related to [MAYN-281](https://openedx.atlassian.net/browse/MAYN-281)) and should only be merged after configuration changes.


docs/theming.rst
```diff
@@ -103,14 +103,14 @@ It is up to you where to install the theme on the server, but a good default loc
     All themes must reside in the same physical directory.
 
 In order for Otto to use the installed themes, you must specify the location of the theme directory in
-Django settings by setting COMPREHENSIVE_THEME_DIR in your settings file:
+Django settings by setting COMPREHENSIVE_THEME_DIRS in your settings file:
```
clintonb, line 106: `THEME_DIRS`. I still have no clue what "comprehensive" means in this context.

mattdrayer, line 106: I would prefer to keep the terminology consistent with the rest of the platform and documentation.

mattdrayer, line 106: Although I do agree that 'comprehensive' is an interesting choice of terminology.

clintonb, line 106: Rather than continue using a term we both agree is "interesting"/wrong, let's start fixing it now. No sense in being consistently confusing.

mattdrayer, line 106: I don't want to fix it for one IDA -- if we want to change it, that's fine,
but I'd like to see a coordinated effort to do so across ecommerce,
edx-platform, edx-documentation, etc.

On Thu, May 19, 2016 at 1:24 PM, Clinton Blackburn <notifications@github.com

> wrote:
> 
> In docs/theming.rst
> https://github.com/edx/ecommerce/pull/749#discussion_r63919857:
> 
> > @@ -103,14 +103,14 @@ It is up to you where to install the theme on the server, but a good default loc
> >      All themes must reside in the same physical directory.
> > 
> >  In order for Otto to use the installed themes, you must specify the location of the theme directory in
> > -Django settings by setting COMPREHENSIVE_THEME_DIR in your settings file:
> > +Django settings by setting COMPREHENSIVE_THEME_DIRS in your settings file:
> 
> Rather than continue using a term we both agree is "interesting"/wrong,
> let's start fixing it now. No sense in being consistently confusing.
> 
> —
> You are receiving this because you were mentioned.
> https://github.com/edx/ecommerce/pull/749/files/9400f6674c94a498907e07a51c5c122c8a974eaf#r63919857

## 

Matthew Drayer
Solutions Architect
edX | MIT
mattdrayer@edx.org

docs/theming.rst
```diff
@@ -195,9 +195,9 @@ Troubleshooting
 ---------------
 If you have gone through the above procedure and you are not seeing theme overrides, you need to make sure
 
-- ``COMPREHENSIVE_THEME_DIR`` must be path for the directory containing all themes e.g. if your theme is
-  ``/edx/app/ecommerce/ecommerce/themes/my-theme`` then correct value for ``COMPREHENSIVE_THEME_DIR`` is
-  ``/edx/app/ecommerce/ecommerce/themes``.
+- ``COMPREHENSIVE_THEME_DIRS`` must contains path for the directory containing themes e.g. if your theme is
```
mattdrayer: s/contains/contain

docs/theming.rst
```diff
@@ -195,9 +195,9 @@ Troubleshooting
...
+  ``/edx/app/ecommerce/ecommerce/themes/my-theme`` then correct value for ``COMPREHENSIVE_THEME_DIRS`` is
+  ``[/edx/app/ecommerce/ecommerce/themes]``.
```
mattdrayer: do the values require single quotes?

ecommerce/settings/base.py
```diff
@@ -219,6 +219,7 @@
     'social.apps.django_app.middleware.SocialAuthExceptionMiddleware',
     'simple_history.middleware.HistoryRequestMiddleware',
     'threadlocals.middleware.ThreadLocalMiddleware',
+    'ecommerce.theming.middleware.CurrentSiteThemeMiddleware',
```
mattdrayer: Was this not already being specified?  I thought it had been added a while back?

mattdrayer: Ah, I confused this middleware with CurrentSiteWithDefaultMiddleware

mattdrayer:
@saleem-latif -- so far this PR is looking good -- I only have some small nits at present -- I'll take another swing through tomorrow after you've finished your first round and we've got a green build.

saleem-latif:
@mattdrayer , @clintonb I have added remaining tests, build is green. Kindly, take a look

mattdrayer:
@saleem-latif -- really liking how this has turned out, thanks very much for making the updates.  Looking forward to taking advantage of multi-directory support soon!  👍

saleem-latif:
@mattdrayer , @clintonb [Here](https://ecommerce-multi-theme-dirs.sandbox.edx.org) is the sandbox running ecommerce and configuration changes.

ecommerce/theming/compressor.py
```diff
@@ -12,8 +12,8 @@ def offline_context():
     """
 
     for theme in get_themes():
-        main_css = ThemeStorage(prefix=theme.theme_dir).url("css/base/main.css")
-        swagger_css = ThemeStorage(prefix=theme.theme_dir).url("css/base/edx-swagger.css")
+        main_css = ThemeStorage(prefix=theme.theme_dir_name).url("css/base/main.css")
```
mattdrayer: Does it make sense to move these paths into settings?

ecommerce/theming/helpers.py
```diff
@@ -139,89 +133,35 @@ def get_all_theme_template_dirs():
...
-
-        return Path(themes_dir) / theme_dir
-    else:
-        return None
-
-
-def get_base_themes_dir():
+def get_theme_base_dirs():
     """
     Return base directory that contains all the themes.
```
mattdrayer: Need to update this docstring

ecommerce/theming/middleware.py
```diff
@@ -0,0 +1,18 @@
+"""
+Middleware for theming app
+
+Note:
+    This middleware depends on "django_sites_extensions.middleware.CurrentSiteWithDefaultMiddleware" middleware
```
mattdrayer, line 5: See https://github.com/edx/edx-django-sites-extensions/pull/6 and https://github.com/edx/edx-platform/pull/12555

mattdrayer:
Looks like we need to rebase the branch, as well.

ecommerce/theming/middleware.py
saleem-latif, line 5: @mattdrayer I think it would be better to make this change in a separate PR, for better testing. I have created a ticket [WL-495](https://openedx.atlassian.net/browse/WL-495) to track this work.

saleem-latif:
@mattdrayer , @clintonb I have incorporated feedback changes, rebased it with master and tested it on sandbox, it is ready for another review.

ecommerce/theming/template_loaders.py
```diff
@@ -51,4 +51,5 @@ def get_theme_template_sources():
             return get_all_theme_template_dirs()
         else:
```
mattdrayer: Nit: This 'else' condition seems like it will be executed more frequently than the 'if' condition, so consider reversing the order.

ecommerce/theming/helpers.py
```diff
@@ -2,119 +2,113 @@
...
     """
-    Cache site's theme directory.
+    for themes_dir in get_theme_base_dirs():
+        if theme_dir_name in (_dir for _dir in os.listdir(themes_dir) if is_theme_dir(themes_dir / _dir)):
+            return themes_dir
 
-    Example:
-        >> site = Site(domain='red-theme.org', name='Red Theme')
-        >> cache_site_theme_dir(site, 'red-theme')
+    if suppress_error:
```
mattdrayer, line 73: Does it make sense to write a log entry here if we are suppressing the error?  I don't want to create additional noise if it's not useful information.

mattdrayer:
@saleem-latif -- looks like we're in need of a rebase here -- I went through the PR and posted a couple of comments/nits -- nothing major at this point -- @clintonb if you can please take a look (or hand off to a teammate) that would be great -- I'll log a review subtask to keep track of the work on our side.

ecommerce/theming/helpers.py
saleem-latif, line 73: suppress error is being used for static storage class where at certain situations it would okay for a theme to not be found, that is why it won't be useful to log it.

saleem-latif:
@mattdrayer branch is rebased and also incorporated your feedback changes.

mattdrayer:
https://github.com/edx/configuration/pull/3065 has been merged so we are clear to move forward with this PR now

mattdrayer:
I am 👍 on this PR at this point

ecommerce/theming/helpers.py
```diff
@@ -2,119 +2,113 @@
...
 
-    template_paths = [
-        site_theme_dir / 'templates',
-        site_theme_dir / 'templates' / 'oscar',
-    ]
-    return template_paths
+    # return False if theming is disabled on runtime and function is called during request processing
+    if bool(get_current_request()):
+        # check if theming is disabled on runtime
+        if waffle.switch_is_active(settings.DISABLE_THEMING_ON_RUNTIME_SWITCH):
```
schenedx: #nitpick why do we need these two conditions on separate if statements? Why can't we use keyword `and` to test the condition like what your comment has been written?

ecommerce/settings/base.py
```diff
@@ -513,7 +514,9 @@
 DISABLE_THEMING_ON_RUNTIME_SWITCH = "disable_theming_on_runtime"
 
 # Directory that contains all themes
-COMPREHENSIVE_THEME_DIR = DJANGO_ROOT + "/themes"
```
schenedx, line 516: This is a concern for me on backwards compatibility. We basically deleted/retired the existing setting key `COMPREHENSIVE_THEME_DIR` and added a new setting key `COMPREHENSIVE_THEME_DIRS`. For someone who have their own settings file or overrides in private.py, this change will break the theming on their existing site and they will have to debug to find out what to change. Not to mention the name of the settings keys are so similar that is easy to be overlooked.
Why can't we leave the original setting key `COMPREHENSIVE_THEME_DIR` in as default, then for `COMPREHENSIVE_THEME_DIRS`, it just include the existing value in `COMPREHENSIVE_THEME_DIR`?
LIke `COMPREHENSIVE_THEME_DIRS = [COMPREHENSIVE_THEME_DIR]`?

schenedx:
One nitpick and one backward compatibility comment. Rest LGTM

ecommerce/settings/base.py
mattdrayer, line 516: @schenedx I agree with your observation about backwards compatibility -- we have noted this issue in the edx/configuration [changelog](https://github.com/edx/configuration/pull/3065/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR138) for Open edX operators to review when validating the master branch for their own installations.

With this particular feature being only a handful of weeks old and going through such rapid iteration at the moment, we felt that the probability of someone other than edX running the feature in a production setting was unlikely, and thus relatively safe to make the breaking change now before too many people begin to utilize it.

schenedx, line 516: @mattdrayer So what you are saying is that, you don't believe there are some, if any, instances of otto that would even have the setting key `COMPREHENSIVE_THEME_DIR` in their configuration?
If that's the case, I am onboard with this change.

mattdrayer, line 516: That's correct @schenedx -- and we expect those operators who are running on the bleeding edge (like us) to review the edx/configuration changelog prior to deploying anything to their production clusters.

ecommerce/theming/helpers.py
mattdrayer: Make sense to me, @schenedx, I've updated this check as suggested.

schenedx:
👍

PR #751, opened 2016-05-20
mattdrayer/site-aware-fulfillment: Specify site code in worker task

mattdrayer:
@rlucioni @mjfrey -- here is the ecommerce side of making the async fulfillment workflow site aware (ref: https://github.com/edx/ecommerce-worker/pull/16).  This was the only reference I could find to the fulfill_order task, so please let me know if there are other spots that need to be updated.

### DIFF
--- a/ecommerce/extensions/checkout/tests/test_mixins.py
+++ b/ecommerce/extensions/checkout/tests/test_mixins.py
@@ -188,6 +188,7 @@ class EdxOrderPlacementMixinTests(BusinessIntelligenceMixin, RefundTestMixin, Te
         with patch('ecommerce.extensions.checkout.mixins.fulfill_order.delay') as mock_delay:
             EdxOrderPlacementMixin().handle_successful_order(self.order)
             self.assertTrue(mock_delay.called)
+            mock_delay.assert_called_once_with(self.order.number, site_code='edX')
 
     def test_place_free_order(self, __):
         """ Verify an order is placed and the basket is submitted. """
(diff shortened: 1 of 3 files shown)


mattdrayer:
@fredsmith FYI

clintonb:
You need to update the requirements: https://github.com/edx/ecommerce/blob/master/requirements/base.txt#L19

mattdrayer:
Right!  Done :)

rlucioni:
@mattdrayer please update the tests to verify that the task is created with the right arguments: https://github.com/edx/ecommerce/blob/e61b0c238cdadaa13ad775b77cb9ce4da581bb85/ecommerce/extensions/checkout/tests/test_mixins.py#L188

mattdrayer:
Added argument validations as requested

ecommerce/extensions/checkout/tests/test_mixins.py
```diff
@@ -188,6 +188,10 @@ def test_handle_successful_async_order(self, __):
         with patch('ecommerce.extensions.checkout.mixins.fulfill_order.delay') as mock_delay:
             EdxOrderPlacementMixin().handle_successful_order(self.order)
             self.assertTrue(mock_delay.called)
+            delay_args, delay_kwargs = mock_delay.call_args
```
rlucioni: Don't do this by hand. Use [`assert_called_once_with()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with).

mattdrayer: Changed -- that is much nicer, yes.

rlucioni:
@mattdrayer looks good. Tests on this PR will fail until edx-ecommerce-worker 0.4.0 is on PyPI. Once that's been done, please squash your commits. 👍 assuming tests then pass.

mattdrayer:
Thanks much -- next stop, edx/configuration 😄

PR #754, opened 2016-05-25
PayPal Dispute Webhook

clintonb:
This view creates Zendesk tickets when a user files a PayPal dispute.

ECOM-3955


clintonb:
I received no response from edX Support to my requests as to whether this will be useful to them.

PR #755, opened 2016-05-25
change in serializer which is require for building receipt url

ahsan-ul-haq:


clintonb:
You don't need the basket the receipt page uses the order number: https://courses.stage.edx.org/commerce/checkout/receipt/?orderNum=EDX-100666.

clintonb:
@ahsan-ul-haq I clicked the wrong button. I didn't mean to close this PR; although, I don't think it's necessary.

ahsan-ul-haq:
@clintonb It's ok I was also thinking to close this.

PR #756, opened 2016-05-25
Fallback to Django settings if site config is empty.

mjfrey:
Try to gett settings from site config and if not there use the Django settings.


mjfrey:
@douglashall

clintonb:
Why the fallback? I specifically advised against such a fallback to ensure that all sites are configured in the same manner.

douglashall:
Yes, thinking about this more all we need to do is update the SiteConfiguration.oauth_settings field with the correct values for the ecommerce site in stage/prod. I have a story in this sprint to clean up the old settings from configuration and remove the fallback for the Otto login workflow.

mjfrey:
I will work with Doug to update the site configs on stage / prod.

PR #758, opened 2016-05-25
Clear Site cache upon SiteConfiguration change.

mjfrey:
Call to Site clear_cache when SiteConfiguration changes.

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -3,6 +3,7 @@ import logging
 from analytics import Client as SegmentClient
 from django.conf import settings
 from django.contrib.auth.models import AbstractUser
+from django.contrib.sites.models import Site
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.functional import cached_property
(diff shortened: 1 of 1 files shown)


mjfrey:
@clintonb @douglashall

ecommerce/core/models.py
```diff
@@ -187,3 +190,11 @@ def validate_configuration():
     """ Validates all existing SiteConfiguration models """
     for config in SiteConfiguration.objects.all():
         config.clean_fields()
+
+
+@receiver(post_save, sender=SiteConfiguration)
```
clintonb: You can avoid the signal work and simply override `SiteConfiguration.save()`.

mjfrey:
@douglashall @clintonb take a second look please.

ecommerce/core/models.py
```diff
@@ -147,6 +148,13 @@ def clean_fields(self, exclude=None):
     def segment_client(self):
         return SegmentClient(self.segment_key, debug=settings.DEBUG)
 
+    def save(self, *args, **kwargs):
+        """
+        Clear Site cache upon SiteConfiguration changed
```
clintonb: This should simply be a comment above the line, rather than a docstring.

clintonb:
👍

douglashall:
@mjfrey @clintonb Just realized that this will not work in a multi-process environment because Django is caching sites in memory.

clintonb:
https://code.djangoproject.com/ticket/15894

We should not be frequently updating SiteConfiguration. The solution here may be to simply restart the service when changes are made, or find a way to expose a VPN-protected endpoint that can be used to trigger the cache clear on all processes. It's not ideal; but, it works.

PR #759, opened 2016-05-26
update "OrderViewSet" class

ahsan-ul-haq:
On staff users can filter orders by username of any user.
ECOM-4550

### DIFF
--- a/ecommerce/extensions/api/filters.py
+++ b/ecommerce/extensions/api/filters.py
@@ -3,6 +3,7 @@ from django.db.models import Q
 
 from oscar.core.loading import get_model
 
+Order = get_model('order', 'Order')
 Product = get_model('catalogue', 'Product')
 
 
@@ -18,3 +19,13 @@ class ProductFilter(django_filters.FilterSet):
     class Meta(object):
         model = Product
         fields = ('product_class', 'structure', 'title',)
+
+
+class OrderFilter(django_filters.FilterSet):
+    """ Filter orders via query string parameter."""
+
+    username = django_filters.CharFilter(name='user__username')
+
+    class Meta(object):
+        model = Order
+        fields = ('username',)
--- a/ecommerce/extensions/api/v2/tests/views/test_orders.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_orders.py
@@ -5,10 +5,11 @@ import httpretty
 import mock
 from django.contrib.auth.models import Permission
 from django.core.urlresolvers import reverse
-from django.test import override_settings
+from django.test import override_settings, RequestFactory
 from oscar.core.loading import get_model
 from oscar.test import factories
 
+from ecommerce.extensions.api.serializers import OrderSerializer
 from ecommerce.extensions.api.tests.test_authentication import AccessTokenMixin
 from ecommerce.extensions.api.v2.tests.views import OrderDetailViewTestMixin
 from ecommerce.extensions.fulfillment.signals import SHIPPING_EVENT_NAME
@@ -115,6 +116,39 @@ class OrderListViewTests(AccessTokenMixin, ThrottlingMixin, TestCase):
         self.assertEqual(content['results'][0]['number'], unicode(order.number))
         self.assertEqual(content['results'][0]['user']['email'], admin_user.email)
         self.assertEqual(content['results'][0]['user']['username'], admin_user.username)
+
+    def test_username_filter_with_staff(self):
+        """ Verify the staff user can filter data by username."""
+
+        # create two orders for different users
+        order = factories.create_order(user=self.user)
+        other_user = self.create_user()
+        other_order = factories.create_order(user=other_user)
+
+        requester = self.create_user(is_staff=True)
+        self.client.login(email=requester.email, password=self.password)
+
+        self.assert_list_with_username_filter(self.user, order)
+        self.assert_list_with_username_filter(other_user, other_order)
+
+    def test_username_filter_with_non_staff(self):
+        """Non staff users are not allowed to filter on any other username."""
+        requester = self.create_user(is_staff=False)
+        self.client.login(username=requester.username, password=self.password)
+
+        response = self.client.get(self.path, {'username': self.user.username})
+        self.assertEqual(response.status_code, 403)
+
+    def assert_list_with_username_filter(self, user, order):
+        """ Helper method for making assertions. """
+
+        response = self.client.get(self.path, {'username': user.username})
+        self.assertEqual(response.status_code, 200)
+
+        self.assertEqual(
+            response.data['results'][0],
+            OrderSerializer(order, context={'request': RequestFactory(SERVER_NAME=self.site.domain).get('/')}).data
+        )
 
 
 @ddt.ddt
--- a/ecommerce/extensions/api/v2/views/orders.py
+++ b/ecommerce/extensions/api/v2/views/orders.py
@@ -2,15 +2,18 @@
 import logging
 
 from oscar.core.loading import get_model, get_class
-from rest_framework import status, viewsets
+from rest_framework import filters, status, viewsets
 from rest_framework.decorators import detail_route
+from rest_framework.exceptions import PermissionDenied
 from rest_framework.permissions import IsAuthenticated, DjangoModelPermissions
 from rest_framework.response import Response
 
 from ecommerce.extensions.api import serializers
 from ecommerce.extensions.api.constants import APIConstants as AC
+from ecommerce.extensions.api.filters import OrderFilter
 from ecommerce.extensions.api.permissions import IsStaffOrOwner
 from ecommerce.extensions.api.throttles import ServiceUserThrottle
+
 
 logger = logging.getLogger(__name__)
 
@@ -23,13 +26,20 @@ class OrderViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Order.objects.all()
     serializer_class = serializers.OrderSerializer
     throttle_classes = (ServiceUserThrottle,)
+    filter_backends = (filters.DjangoFilterBackend,)
+    filter_class = OrderFilter
 
     def filter_queryset(self, queryset):
         queryset = super(OrderViewSet, self).filter_queryset(queryset)
+
+        username = self.request.query_params.get('username')
         user = self.request.user
 
         # Non-staff users should only see their own orders
         if not user.is_staff:
+            if username and user.username != username:
+                raise PermissionDenied
+
             queryset = queryset.filter(user=user)
 
         return queryset


ahsan-ul-haq:
@clintonb please review

clintonb:
I'd much rather see a username querystring filter (e.g. https://ecommerce.example.com/api/v2/orders/?username=clintonb). DRF has support for this: http://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend.

tasawernawaz:
@clintonb I have add django filter but have some question.
Previously all users can see there own orders. 
After adding filter, user can filter like this `/api/v2/orders?username=<username>`, but that filter is applied only on queryset overridden [here](https://github.com/edx/ecommerce/pull/759/files#diff-eea0521d5c26662ce10e8a69057642b0L27) which is already filtering queryset on requested user.. Please let me know what is expected behavior ?

ahsan-ul-haq:
I would suggest that leave [this](https://github.com/edx/ecommerce/pull/759/files#diff-eea0521d5c26662ce10e8a69057642b0L27) as it was before and add django filter on the resultant data.

tasawernawaz:
Updated code as per your suggestion. @ahsan-ul-haq

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -95,9 +95,10 @@ def test_with_other_users_orders(self):
         (True, False),
     )
     def test_staff_superuser(self, is_staff, is_superuser):
-        """ The view should return all orders for when authenticating as a staff member or superuser. """
```
clintonb, line 98: Revert.

clintonb:
@tasawernawaz your new code needs tests.

The logic for this view should be as follows:
1. If the user is not a staff user, the user only sees his/her own orders regardless of the username filter. It would be preferable to return a 403 status code informing the user he is not allowed to filter by the username of other users as a non-staff user.
2. If the user is a staff user, by default (e.g. no filter) he sees all orders. Adding a filter should limit the orders to those of the requested user.

tasawernawaz:
@clintonb Made suggested changes in view. Please review.

ecommerce/extensions/api/v2/views/orders.py
```diff
@@ -23,16 +25,21 @@ class OrderViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Order.objects.all()
     serializer_class = serializers.OrderSerializer
     throttle_classes = (ServiceUserThrottle,)
+    filter_backends = (filters.DjangoFilterBackend,)
+    filter_class = OrderFilter
 
-    def filter_queryset(self, queryset):
```
clintonb, line 27: Why was this changed?

ecommerce/extensions/api/v2/views/orders.py
```diff
@@ -23,16 +25,21 @@ class OrderViewSet(viewsets.ReadOnlyModelViewSet):
...
-        queryset = super(OrderViewSet, self).filter_queryset(queryset)
+    def list(self, request, *args, **kwargs):
+        username = self.request.query_params.get('username')
         user = self.request.user
 
-        # Non-staff users should only see their own orders
         if not user.is_staff:
-            queryset = queryset.filter(user=user)
+            if username and user.username != username:
+                return Response("You are not allowed to filter on any other username.",
```
clintonb: `raise PermissionDenied`

ecommerce/extensions/api/v2/views/orders.py
tasawernawaz, line 27: I have override `List` view. Do you think this method still required ?

clintonb, line 27: Override `filter_queryset`.

tasawernawaz:
@clintonb Made suggested change. Please review.

ecommerce/extensions/api/v2/views/orders.py
```diff
@@ -23,13 +26,19 @@ class OrderViewSet(viewsets.ReadOnlyModelViewSet):
...
+    filter_backends = (filters.DjangoFilterBackend,)
+    filter_class = OrderFilter
 
     def filter_queryset(self, queryset):
         queryset = super(OrderViewSet, self).filter_queryset(queryset)
+
+        username = self.request.query_params.get('username')
         user = self.request.user
 
-        # Non-staff users should only see their own orders
```
clintonb, line 31: This comment is still relevant.

ecommerce/extensions/api/v2/views/orders.py
```diff
@@ -1,17 +1,20 @@
...
+from rest_framework.exceptions import PermissionDenied
 
 from oscar.core.loading import get_model, get_class
-from rest_framework import status, viewsets
+from rest_framework import filters, status, viewsets
 from rest_framework.decorators import detail_route
 from rest_framework.permissions import IsAuthenticated, DjangoModelPermissions
 from rest_framework.response import Response
 
+from ecommerce.extensions.api.filters import OrderFilter
```
clintonb: Alphabetize.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -116,6 +116,37 @@ def test_user_information(self):
...
+        """ Verify the "staff" user can filter data by username. Non staff
+        users are not allowed to filter on any other username."""
+
+        # create two orders for different users
+        order = factories.create_order(user=self.user)
+        other_user = self.create_user()
+        other_order = factories.create_order(user=other_user)
+
+        requester = self.create_user(is_staff=is_staff)
+        token = self.generate_jwt_token_header(requester)
```
clintonb: You don't need a token. Call `self.client.login`.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -116,6 +116,37 @@ def test_user_information(self):
         self.assertEqual(content['results'][0]['user']['email'], admin_user.email)
         self.assertEqual(content['results'][0]['user']['username'], admin_user.username)
 
+    @ddt.unpack
+    @ddt.data(
+        {'is_staff': True, 'status_code': 200},
+        {'is_staff': False, 'status_code': 403},
+    )
+    def test_username_filter(self, is_staff, status_code):
```
clintonb: Split this into two tests. You can share code with helper methods, but this will be a lot cleaner as two separate tests.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -116,6 +116,37 @@ def test_user_information(self):
...
+        token = self.generate_jwt_token_header(requester)
+
+        self.assert_list_with_username_filter(self.user, order, token, status_code)
+        self.assert_list_with_username_filter(other_user, other_order, token, status_code)
+
+    def assert_list_with_username_filter(self, user, order, token, expected_status_code):
+        response = self.client.get(self.path, data={'username': user.username}, HTTP_AUTHORIZATION=token)
+        self.assertEqual(response.status_code, expected_status_code)
+
+        if expected_status_code == 200:
```
clintonb: Assertion methods shouldn't have if statements. This is a sign that you should have a separate assertion method.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -116,6 +116,37 @@ def test_user_information(self):
...
+            content = json.loads(response.content)
```
clintonb: Use the serializer and look at `response.data`.

ecommerce/extensions/api/filters.py
```diff
@@ -18,3 +19,13 @@ def filter_product_class(self, queryset, value):
     class Meta(object):
         model = Product
         fields = ('product_class', 'structure', 'title',)
+
+
+class OrderFilter(django_filters.FilterSet):
+    """ Filter orders via query string parameter."""
+
+    username = django_filters.CharFilter(name="user__username")
```
zubair-arbi: use single quotes

tasawernawaz:
@clintonb Please review.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -8,7 +8,9 @@
 from django.test import override_settings
 from oscar.core.loading import get_model
 from oscar.test import factories
+from threadlocals.threadlocals import get_current_request
```
clintonb: Get rid of this and avoid using it. It will be going away soon. Use RequestFactory.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -116,6 +118,37 @@ def test_user_information(self):
         self.assertEqual(content['results'][0]['user']['email'], admin_user.email)
         self.assertEqual(content['results'][0]['user']['username'], admin_user.username)
 
+    def test_username_filter_with_staff(self):
+        """ Verify the "staff" user can filter data by username."""
```
clintonb: Staff doesn't need to be in quotes.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -116,6 +118,37 @@ def test_user_information(self):
...
+        requester = self.create_user(is_staff=False)
+        self.client.login(username=requester.username, password=self.password)
+
+        response = self.client.get(self.path, data={'username': self.user.username})
+        self.assertEqual(response.status_code, 403)
+
+    def assert_list_with_username_filter(self, user, order):
+        """ Helper method for making assertions. """
+
+        response = self.client.get(self.path, data={'username': user.username})
```
clintonb: Passing a data kwarg is now how you set a query string.

clintonb:
:+1: (once remaining comments are addressed)

ecommerce/extensions/api/v2/tests/views/test_orders.py
tasawernawaz: Initially I had tried to use `RequestFactory` but `ProductSerializer` return different url from [here](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/serializers.py#L149) what we have in response. Our response contains `testserver.fake` as domain name coming from here.
Using `RequestFactory` how I can override domain name.

clintonb:
``` python
RequestFactory(SERVER_NAME=self.site.domain).get('/')
```

tasawernawaz:
@clintonb I am continuously getting this error on build while running `make validate_js`
`Error: Cannot find module 'gulp-util`. Is this particular related to this PR.

clintonb:
@tasawernawaz given that you aren't dealing with JavaScript, the issue is most likely unrelated. However, since it's preventing you from finishing this, you should investigate.

PR #720, opened 2016-05-09
ECOM-4312 Check added to prevent multiple order issue.

Ayub-Khan:
**ECOM-4312 :**
Add a check to prevent multiple order placement for a single seat

**Description:**
Student often place multiple orders for the same course seat even if they are enrolled in the track they want to. In this pull request i have added a check of enrollment mode in the course for the course seat if student is already enrolled in the course as verified and he is trying to buy verified seat again. This check wont allow him to create basket. 

**Code Reviewers :**
@awaisdar001 and @aamir-khan 
Kindly Review this pr.
FYI: @adampalay

PR #760, opened 2016-05-26
mattdrayer/add-enrollment-active-check: Include "is_active" in enrollment checks

mattdrayer:
@clintonb @Ayub-Khan @douglashall -- I've been investigating bug [WL-487](https://openedx.atlassian.net/browse/WL-487) and it appears there is a scenario where the Enrollment API returns a valid response for an enrollment record request, but the enrollment record itself is actually marked inactive (ie, `{"mode": "verified", "is_active": "False"}`).

See 
- https://github.com/edx/edx-platform/blob/master/common/djangoapps/enrollment/api.py#L112
- https://github.com/edx/edx-platform/blob/master/common/djangoapps/enrollment/data.py#L81

This is a bit different from the list request, which does filter enrollment records according to active status: https://github.com/edx/edx-platform/blob/master/common/djangoapps/enrollment/data.py#L40

Looks like this issue was introduced in https://github.com/edx/ecommerce/pull/720 -- see [here](https://github.com/edx/ecommerce/pull/720/files#diff-398e80deb2588fea52e2551897aac3e2R67) -- so it's currently in production -- after this PR is approved we should probably put together an ecommerce release

### DIFF
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -58,15 +58,31 @@ class BasketSingleItemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockM
         self.catalog = Catalog.objects.create(partner=self.partner)
         self.catalog.stock_records.add(self.stock_record)
 
-    def mock_enrollment_api_success(self, course_id, mode='audit'):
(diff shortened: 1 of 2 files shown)


clintonb:
👍

PR #761, opened 2016-05-26
Add a new management command for honor<->audit conversions.

peter-fogg:
# [ECOM-4262](https://openedx.atlassian.net/browse/ECOM-4262)

@bderusha @clintonb 

I've also removed the old convert_honor_to_audit, since the new command handles both directions. The difference is that the new version updates products in-place, thus making it suitable for converting courses which already have enrollment -- we don't need to update orders.


ecommerce/extensions/catalogue/management/commands/convert_course.py
```diff
@@ -0,0 +1,157 @@
...
+
+        honor_seat.title = honor_seat.title.split(' with honor certificate')[0]
+        honor_seat.attr.certificate_type = ''
+        honor_seat.save()
+
+        stock_record = StockRecord.objects.get(product=honor_seat)
+        stock_record.partner_sku = generate_sku(honor_seat, self.partner)
+        stock_record.save()
+
+    def _install_current_request(self, site_domain):
```
peter-fogg: FYI @douglashall -- this is the approach I'm taking to allow this command to publish courses outside of the context of a request.

douglashall: Nice! How about refactoring this to a command mixin (which could be added as a parent class of any command) that calls this function from `__init__`?

clintonb: I really don't want to make a habit of this. As much as possible, we should update our code to accept site arguments. Relying on a request leads to hacks like this.

bderusha:
LGTM 👍

ecommerce/courses/urls.py
```diff
@@ -4,7 +4,7 @@
 
 urlpatterns = [
     url(r'^migrate/$', views.CourseMigrationView.as_view(), name='migrate'),
-    url(r'^convert_honor_to_audit/$', views.CourseConvertHonorToAuditView.as_view(), name='convert_honor_to_audit'),
+    url(r'^convert_course/$', views.ConvertCourseView.as_view(), name='convert_course'),
```
peter-fogg, line 7: @clintonb @bderusha: do you know if this URL was documented anywhere? I haven't been able to find any instances of it, but if you know of any let me know and I'll update them once this is merged.

clintonb, line 7: I doubt it. I'm still struggling to track down docs for ecommerce since they were moved.

ecommerce/extensions/catalogue/management/commands/convert_course.py
```diff
@@ -0,0 +1,162 @@
...
+
+        self.access_token = options.get('access_token')  # pylint: disable=attribute-defined-outside-init
+        if not self.access_token:
+            logger.error('Cannot convert and publish a course without an access token.')
+            return
+
+        site_domain = options.get('site_domain')
+        self._install_current_request(site_domain)
+
+        partner_code = options.get('partner_code', 'edx')
```
clintonb: This should not have a default. Pass in this value, instead of `site`, and use it to lookup the correct `SiteConfiguration` and `Site` models.

clintonb: P.S. You don't want a default because we don't want to accidentally modify the wrong values. Additionally, you don't want a discrepancy between site and partner.

ecommerce/extensions/catalogue/management/commands/convert_course.py
```diff
@@ -0,0 +1,162 @@
...
+            seat for seat in course.seat_products
+            if getattr(seat.attr, 'certificate_type', '') == ''
+        ]
+        if len(audit_seats) != 1:
+            logger.error('Course [%s] has [%d] audit seats.', course.id, len(audit_seats))
+            raise Exception
+
+        audit_seat = audit_seats[0]
+
+        audit_seat.title += ' with honor certificate'
```
clintonb: Use `course._get_course_seat_name()`, which should be made public, to avoid divergence.

ecommerce/extensions/catalogue/management/commands/convert_course.py
```diff
@@ -0,0 +1,162 @@
...
+
+        Arguments:
+            site_domain (str | None): The domain of the site to set. If not provided, the site is retrieved
+                using settings.DEFAULT_SITE_ID.
+
+        Returns:
+            None
+        """
+        if not site_domain:
+            site = Site.objects.get(id=settings.DEFAULT_SITE_ID)
```
clintonb: No defaults. If the value isn't passed in, the command should fail.

clintonb:
👍  (once comments addressed).

peter-fogg:
@clintonb Comments addressed; I'll merge when tests pass.

PR #764, opened 2016-05-29
Moved coupons-related mixins to appropriate module

clintonb:


ecommerce/coupons/tests/test_utils.py
```diff
@@ -1,12 +1,11 @@
+import httpretty
```
vkaracic, line 1: Isn't this a third party import?

clintonb, line 1: Yes, and it is grouped as such.

vkaracic, line 1: Oh, I always thought mock was part of the standard library.

PR #765, opened 2016-05-29
Removed request dependency injection

clintonb:
Passing Site/SiteConfiguration instances where necessary to avoid the numerous hacks associated with relying on dependency injection.

LEARNER-1283


clintonb:
FYI @edx/ecommerce @mjfrey @douglashall @mattdrayer I'll be testing these changes on stage before merging tomorrow. Please refrain from expanding usage of `django-threadlocals` unless absolutely necessary. It's usage is an anti-pattern that invites hacks.

At present `django-threadlocals` remains in use for the `Range` model and the theming template loader. These usages will be updated once we determine a suitable alternative.

mattdrayer:
@saleem-latif FYI

mjfrey:
@vkaracic FYI

ecommerce/core/constants.py
```diff
@@ -1,17 +1,16 @@
 """Constants core to the ecommerce app."""
 
-ISO_8601_FORMAT = u'%Y-%m-%dT%H:%M:%SZ'
+from __future__ import unicode_literals
 
+ISO_8601_FORMAT = u'%Y-%m-%dT%H:%M:%SZ'
```
peter-fogg: With the `__future__` import above you can leave off the `u`.

ecommerce/core/models.py
```diff
@@ -153,6 +153,73 @@ def save(self, *args, **kwargs):
         Site.objects.clear_cache()
         super(SiteConfiguration, self).save(*args, **kwargs)
 
+    def build_ecommerce_url(self, path=''):
+        """
+        Returns path joined with the appropriate ecommerce URL root for the current site
+
+        Raises:
+            MissingRequestError: If the current ecommerce site is not in threadlocal storage
```
peter-fogg: Update the docstring.

ecommerce/core/constants.py
clintonb: Fixed.

ecommerce/core/context_processors.py
```diff
@@ -1,12 +1,10 @@
...
 
-from ecommerce.core.url_utils import get_lms_dashboard_url, get_lms_url
-
 
 def core(request):
     return {
-        'lms_base_url': get_lms_url(),
-        'lms_dashboard_url': get_lms_dashboard_url(),
+        'lms_base_url': request.site.siteconfiguration.build_lms_url(),
+        'lms_dashboard_url': request.site.siteconfiguration.student_dashboard_url,
```
douglashall, line 7: We need to modify any code that is pulling the SiteConfiguration off of the site to a cached lookup of the SiteConfiguration using the site_id. SiteConfiguration.site should be changed from a OneToOneField to an IntegerField.

clintonb, line 7: I don't follow you. Why is this necessary? We definitely do _not_ want to drop the database relationship between the `Site` and `SiteConfiguration` models.

douglashall, line 7: See our conversation [here](https://github.com/edx/ecommerce/pull/758#issuecomment-222834789). Sounds like you have a good workaround in mind. Can you create a Jira task to track it?

PR #769, opened 2016-05-31
Affiliate window verification CSV

bderusha:
Still working on getting all of the tests to pass.

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -68,6 +68,13 @@ class SiteConfiguration(models.Model):
         null=True,
         blank=True
     )
+    order_attribution_period = models.PositiveIntegerField(
+        verbose_name=_('Order Attribution Period'),
+        help_text=_('Number of days after an order is placed before it can be attributed to an affiliate.'),
+        null=False,
+        blank=False,
+        default=60
+    )
 
     class Meta(object):
         unique_together = ('site', 'partner')
--- a/ecommerce/referrals/urls.py
+++ b/ecommerce/referrals/urls.py
@@ -1,0 +1,11 @@
+from django.conf.urls import url
+
+from ecommerce.referrals import views
+
+urlpatterns = [
+    url(
+        r'^affiliates/(?P<affiliate_partner>\w+)/validation_report/$',
+        views.ValidationReportCsvView.as_view(),
+        name='validation-report'
+    ),
+]
--- a/ecommerce/referrals/views.py
+++ b/ecommerce/referrals/views.py
@@ -1,0 +1,94 @@
+import csv
+from datetime import datetime, timedelta
+import logging
+
+from django.http import Http404, HttpResponse
+from django.utils.text import slugify
+from django.views.generic import View
+from oscar.core.loading import get_model
+from pytz import UTC
+
+from ecommerce.core.httpauth import basic_http_auth
+from ecommerce.referrals.models import Referral
+
+logger = logging.getLogger(__name__)
+Refund = get_model('refund', 'Refund')
+
+
+class ValidationReportCsvView(View):
+    """ Download affiliate attribution validation CSV file view. """
+
+    @basic_http_auth
+    def get(self, request, affiliate_partner):
+        """
+        Creates a CSV referral validation report The structure of the CSV looks like this:
+
+           > Order Reference,Transaction Date,Status,Status Note
+
+           > ORDER-0001,01/01/2016 12:00:00,ACCEPTED,
+           > ORDER-0002,01/01/2016 15:30:00,ACCEPTED,
+           > ORDER-0003,01/01/2016 16:00:00,DECLINED,Order refunded
+
+        Args:
+            request (Request): The GET request
+            affiliate_partner (str): Affiliate partner for the report
+
+        Returns:
+            HttpResponse
+
+        Raises:
+            Http404: When a non-staff user tries to download a CSV for an affiliate other than themselves.
+
+        """
+
+        if request.user.username != affiliate_partner and not request.user.is_staff:
+            raise Http404('No report available for that affiliate.')
+
+        date_param = request.GET.get('date')
+        if date_param:
+            as_of_date = datetime.strptime(date_param, "%Y-%m-%d")
+        else:
+            as_of_date = datetime.utcnow()
+
+        report_date = as_of_date - timedelta(days=request.site.siteconfiguration.order_attribution_period)
+
+        start_period = report_date.replace(hour=0, minute=0, second=0, microsecond=0, tzinfo=UTC)
+        end_period = report_date.replace(hour=23, minute=59, second=59, microsecond=999999, tzinfo=UTC)
+
+        referrals = Referral.objects.select_related('order').prefetch_related('order__refunds').filter(
+            affiliate_id=affiliate_partner,
+            order__date_placed__range=(start_period, end_period)
+        )
+
+        order_info = {}
+        for referral in referrals:
+            order = referral.order
+            is_refunded = order.refunds.count() > 0
+            order_info[order.id] = {
+                'order_number': order.number,
+                'transaction_date': order.date_placed.strftime("%d/%m/%Y %H:%M:%S"),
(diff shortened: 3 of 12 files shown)


bderusha:
@peter-fogg @clintonb please review

ecommerce/referrals/views.py
```diff
@@ -0,0 +1,101 @@
...
+        Raises:
+            PermissionDenied: When a non-staff user tries to download a CSV for an affiliate other than themselves.
+
+        """
+
+        if request.user.username != affiliate_partner and not request.user.is_staff:
+            raise Http404('No report available for that affiliate.')
+
+        as_of_date = datetime.strptime(
+            request.GET.get('date', datetime.utcnow().strftime("%Y-%m-%d")),
```
peter-fogg: It seems a little weird to `strftime` today's date only to `strptime` it back into a proper date object in the case where the user doesn't request a date, but that's more of a nitpick.

ecommerce/referrals/views.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+
+        refunds = Refund.objects.filter(order_id__in=[ref.order_id for ref in referrals])
+
+        order_info = {}
+        for referral in referrals:
+            order = referral.order
+            order_info[order.id] = {
+                'order_number': order.number,
+                'transaction_date': order.date_placed.strftime("%d/%m/%Y %H:%M:%S"),
```
peter-fogg, line 69: Is this something that might be useful to configure per-site?

peter-fogg, line 69: "This" being the date format.

peter-fogg:
👍

ecommerce/referrals/views.py
bderusha: Yea, it's definitely a little weird, but this did seem like the cleanest way to do it.

This would be another option I suppose

``` python
date_param = request.GET.get('date')
if date_param:
    as_of_date = datetime.strptime(date_param, "%Y-%m-%d")
else:
    as_of_date = datetime.utcnow()
```

ecommerce/referrals/views.py
bderusha, line 69: Agreed, that may be useful.  But it's easy enough to add that complexity later, so until we have a requirement to alter the date format I think the YAGNI approach is prudent here

peter-fogg, line 69: 👍

ecommerce/referrals/views.py
bderusha: @peter-fogg on seeing this alternative written out, it's 1 line longer and far easier to hold in your head what's happening.  Change made.

ecommerce/referrals/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.conf.urls import url
+
+from ecommerce.referrals import views
+
+urlpatterns = [
+    url(
+        r'^affiliates/(?P<affiliate_partner>\w+)/validation_report$',
```
clintonb: Add a trailing slash.

ecommerce/referrals/views.py
```diff
@@ -0,0 +1,103 @@
...
+        Returns:
+            HttpResponse
+
+        Raises:
+            PermissionDenied: When a non-staff user tries to download a CSV for an affiliate other than themselves.
+
+        """
+
+        if request.user.username != affiliate_partner and not request.user.is_staff:
+            raise Http404('No report available for that affiliate.')
```
clintonb, line 45: Either raise `PermissionDenied` or update your docstring.

ecommerce/referrals/views.py
```diff
@@ -0,0 +1,103 @@
...
+
+        date_param = request.GET.get('date')
+        if date_param:
+            as_of_date = datetime.strptime(date_param, "%Y-%m-%d")
+        else:
+            as_of_date = datetime.utcnow()
+
+        report_date = as_of_date - timedelta(days=request.site.siteconfiguration.order_attribution_period)
+
+        start_period = datetime(report_date.year, report_date.month, report_date.day, tzinfo=UTC)
```
clintonb: You could also use [`datetime.replace()`](https://docs.python.org/2.7/library/datetime.html#datetime.datetime.replace).

ecommerce/referrals/views.py
```diff
@@ -0,0 +1,103 @@
...
+        end_period = datetime(report_date.year, report_date.month, report_date.day, 23, 59, 59, tzinfo=UTC)
+
+        referrals = Referral.objects.select_related('order').filter(
+            affiliate_id=affiliate_partner,
+            order__date_placed__range=(start_period, end_period)
+        )
+
+        refunds = Refund.objects.filter(order_id__in=[ref.order_id for ref in referrals])
```
clintonb: You be able to use `select_related('order__refund')` above, and get rid of this. If so, you no longer need to loop over refunds. Just check if `order.refund is None`.

ecommerce/referrals/views.py
```diff
@@ -0,0 +1,103 @@
...
+                'transaction_date': order.date_placed.strftime("%d/%m/%Y %H:%M:%S"),
+                'refunded': False,
+                'status_note': '',
+            }
+
+        for refund in refunds:
+            order_info[refund.order_id]['refunded'] = True
+            order_info[refund.order_id]['status_note'] = 'Order refunded'
+
+        file_name = 'Affiliate attribution validation CSV {affiliate} {date}'.format(
```
clintonb: You don't need CSV in the file name, since that's the extension.

ecommerce/referrals/views.py
```diff
@@ -0,0 +1,103 @@
...
+            affiliate=affiliate_partner,
+            date=report_date
+        )
+        file_name = '{filename}.csv'.format(filename=slugify(file_name))
+
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename={filename}'.format(filename=file_name)
+
+        attribution_field_names = ('Order Reference', 'Transaction Date', 'Status', 'Status Note',)
```
clintonb, line 83: Any reason you opted to write all of this on your own instead of using [`djangorestframework-csv`](https://github.com/mjumbewu/django-rest-framework-csv)?

ecommerce/settings/base.py
```diff
@@ -414,6 +414,7 @@
         'edx_rest_framework_extensions.authentication.JwtAuthentication',
         'ecommerce.extensions.api.authentication.BearerAuthentication',
         'rest_framework.authentication.SessionAuthentication',
+        'rest_framework.authentication.BasicAuthentication',
```
clintonb: Your view is not a DRF view, so this has no effect.

ecommerce/core/models.py
```diff
@@ -68,6 +68,13 @@ class SiteConfiguration(models.Model):
         null=True,
         blank=True
     )
+    order_attribution_period = models.PositiveIntegerField(
```
clintonb, line 71: Update the create_or_update_site management command.

clintonb:
This looks fine, but the basic auth portion is incomplete.

ecommerce/referrals/views.py
bderusha, line 83: Mostly because it is not currently utilized in the project and I didn't know about it until just now, so I implemented it similarly to the other CSV endpoint in this repo.

ecommerce/core/models.py
bderusha, line 71: Fixed.

ecommerce/referrals/views.py
bderusha, line 45: Fixed.

ecommerce/referrals/views.py
clintonb, line 83: Between this and the custom basic auth code, this implementation would be a bit cleaner as a DRF view.

clintonb:
👍  (although, I prefer a DRF implementation to take advantage of existing basic auth and CSV implementations)

clintonb:
@bderusha would you close this if it's no longer needed, please?

PR #771, opened 2016-05-31
DRY: Use utility function to get course info from LMS

mjfrey:
Factor out course API calls as a utility.

### DIFF
--- a/ecommerce/courses/tests/test_utils.py
+++ b/ecommerce/courses/tests/test_utils.py
@@ -1,11 +1,19 @@
-import ddt
+import hashlib
 
+import ddt
+import httpretty
+
+from django.core.cache import cache
+
+from ecommerce.core.url_utils import get_lms_url
 from ecommerce.courses.models import Course
-from ecommerce.courses.utils import mode_for_seat
+from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.courses.utils import get_course_info_from_lms, mode_for_seat
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.tests.testcases import TestCase
 
 
+@httpretty.activate
 @ddt.ddt
 class UtilsTests(CourseCatalogTestMixin, TestCase):
     @ddt.unpack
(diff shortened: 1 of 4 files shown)


mjfrey:
@vkaracic

ecommerce/courses/tests/test_utils.py
```diff
@@ -1,11 +1,18 @@
+import hashlib
```
vkaracic, line 1: Watch out for the order of imports, you should add a line here to distinguish between standard library imports and third party ones.

ecommerce/courses/tests/test_utils.py
```diff
@@ -24,3 +31,22 @@ def test_mode_for_seat(self, certificate_type, id_verification_required, mode):
...
+        course_url = get_lms_url('api/courses/v1/courses/{}/'.format(course.id))
+        httpretty.register_uri(httpretty.GET, course_url, status=200, content_type='application/json')
+
+        cache_key = 'courses_api_detail_{}'.format(course.id)
+        cache_hash = hashlib.md5(cache_key).hexdigest()
+        cached_course = cache.get(cache_hash)
+        self.assertIsNone(cached_course)
+
+        response = get_course_info(course.id)
+        self.assertEqual(response, 'HTTPretty :)')
```
vkaracic: You should check the status instead. The default outputs of third party imports (like `HTTPretty :)` here) are subject to change so generally it's better to test for the values that we have a control of, like the status code that you set on line 39, or set a custom body for the registered URI and test it here.

ecommerce/courses/tests/test_utils.py
```diff
@@ -24,3 +31,22 @@ def test_mode_for_seat(self, certificate_type, id_verification_required, mode):
...
+
+        response = get_course_info(course.id)
```
vkaracic: I don't think there is a need to do this twice. I overlooked that there is no `else` part so practically all the lines in the util function will get hit by the test with the first call on line 46.

ecommerce/courses/utils.py
```diff
@@ -8,3 +17,14 @@ def mode_for_seat(seat):
         return 'audit'
 
     return certificate_type
+
+
+def get_course_info(course_key):
```
vkaracic: You should add a docstring explaining what this function does.

vkaracic: Also what do you think about giving this a bit more descriptive name, the current one seems a bit too broad. Something like `get_course_info_from_lms`?

vkaracic:
:+1:

ecommerce/courses/utils.py
```diff
@@ -8,3 +17,15 @@ def mode_for_seat(seat):
         return 'audit'
 
     return certificate_type
+
+
+def get_course_info_from_lms(course_key):
```
clintonb, line 22: get_course_info is actually fine. This data should be coming from the Course Catalog API.

PR #772, opened 2016-05-31
Remove student logout when checking out

JoyceZhu:
SOL-1761: Remove Logout from Checkout

Put the base navbar's logout dropdown in a named block and override it to be empty in student navbars.

Reviewer: @mjfrey


clintonb:
What's the use case/value of this work? Do we actually have data showing that student logouts are leading to a decreased checkout rate? As is, this change affects all pages, not just checkout.

clintonb:
@JoyceZhu if no work is going to be done on this, please close this pull request and delete the branch.

JoyceZhu:
Closing -- Deen has verified this isn't necessary anymore.

PR #773, opened 2016-06-02
Added helper methods to SiteConfiguration model

clintonb:
ECOM-4615

This is a precursor to #765. These helper methods will help ensure that future development can avoid dependency injection.

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -1,18 +1,21 @@
+import datetime
 import logging
+from urlparse import urljoin
 
 from analytics import Client as SegmentClient
 from django.conf import settings
 from django.contrib.auth.models import AbstractUser
 from django.contrib.sites.models import Site
+from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
+from edx_rest_api_client.client import EdxRestApiClient
 from jsonfield.fields import JSONField
 
 from ecommerce.extensions.payment.exceptions import ProcessorNotFoundError
 from ecommerce.extensions.payment.helpers import get_processor_class_by_name, get_processor_class
-
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


clintonb:
FYI @edx/ecommerce @mjfrey @douglashall @mattdrayer

ecommerce/core/models.py
```diff
@@ -153,6 +153,90 @@ def save(self, *args, **kwargs):
...
+        # pylint: disable=unsubscriptable-object
+        if not access_token:
+            access_token, expiration_datetime = EdxRestApiClient.get_oauth_access_token(
+                '{root}/access_token'.format(root=self.oauth2_provider_url),
+                self.oauth_settings['SOCIAL_AUTH_EDX_OIDC_KEY'],
+                self.oauth_settings['SOCIAL_AUTH_EDX_OIDC_SECRET'],
+                token_type='jwt'
+            )
+
+            expires = (expiration_datetime - datetime.datetime.utcnow()).seconds
```
rlucioni, line 225: Alternatively:

``` py
from django.utils import timezone
...
expires = (expiration_datetime - timezone.now()).seconds
```

rlucioni, line 225: Though, if `expiration_datetime` is specified in UTC, we'd want to be subtracting the current time in UTC as well. So maybe using `timezone.now()` isn't the best call.

ecommerce/core/tests/test_models.py
```diff
@@ -199,9 +199,28 @@ def test_get_from_email(self):
...
+        # Verify the value is cached
+        httpretty.reset()
+        self.assertEqual(self.site.siteconfiguration.access_token, token)
+        self.assertFalse(httpretty.has_request())
+
+
+
+    def test_course_catalog_api_client(self):
+        """ Verify the property returns a Course Catalog API client. """
+        self.fail()
```
rlucioni: Need to fill this in?

ecommerce/core/tests/test_models.py
```diff
@@ -199,9 +199,28 @@ def test_get_from_email(self):
         site_config = SiteConfigurationFactory(from_email=expected_from_email, partner__name='TestX')
         self.assertEqual(site_config.get_from_email(), expected_from_email)
 
+    @httpretty.activate
+    def test_access_token(self):
+        """ Verify the property retrieves an access token from the OAuth 2.0 provider. """
+        token, __ = self.mock_access_token_response()
```
rlucioni: `mock_access_token_response` doesn't seem to exist.

mjfrey:
What if any work needs to be done to support this?  Should we create tickets to refactor existing code? Or just use these going forward?

clintonb:
@mjfrey #765 will ultimately clean existing usages. There is no definite timeframe for getting that merged. Going forward, if you are touching code that is calling url_utils directly or adding new code, that code needs to be adapted to use `SiteConfiguration`.

PR #777, opened 2016-06-06
mattdrayer/WL-501: Redirect enrollment code redemptions to LMS dashboard

mattdrayer:
Our QA team discovered an issue with enrollment code (100% discount) redemptions due to updated behavior on the LMS side.  Learners were originally redirected from Otto to the LMS dashboard after redeeming an enrollment code.  This was because  because the default behavior for a login was to send the user to the dashboard.  However after a recent change to the LMS it appears unspecified logins are no longer sent to the dashboard.

The fix seems simple enough -- we just need to specify the LMS dashboard path as the destination for the Otto redirect.

@mjfrey @andy-armstrong FYI

### DIFF
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -221,7 +221,7 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
         else:
             return HttpResponseRedirect(reverse('basket:summary'))
 
-        return HttpResponseRedirect(get_lms_url(''))
+        return HttpResponseRedirect(request.site.siteconfiguration.student_dashboard_url)
 
 
 class EnrollmentCodeCsvView(View):
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -13,7 +13,7 @@ from oscar.test.factories import (
 )
 from oscar.test.utils import RequestFactory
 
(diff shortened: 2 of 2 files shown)


andy-armstrong:
@mattdrayer I'm not familiar with how these redirects should work, so I don't think I can help. Can you find someone from eComm to review this. Do add a unit or acceptance test too.

ecommerce/coupons/views.py
```diff
@@ -221,7 +221,7 @@ def get(self, request):
         else:
             return HttpResponseRedirect(reverse('basket:summary'))
 
-        return HttpResponseRedirect(get_lms_url(''))
+        return HttpResponseRedirect(get_lms_url('/dashboard'))
```
clintonb: Use [`request.site.siteconfiguration.student_dashboard_url`](https://github.com/edx/ecommerce/blob/master/ecommerce/core/models.py#L182).

mattdrayer:
@andy-armstrong just cc'ed you because I believe you were involved in the update to the LMS behavior, but I could be mistaken about that.  I just added the updated tests as well 😄

ecommerce/coupons/views.py
mattdrayer: Cool, good to know this exists -- updated.

andy-armstrong:
@mattdrayer Gotcha. I didn't have anything to do with this change, so I won't be much help. :-)

mjfrey:
:+1:

PR #778, opened 2016-06-07
Added support for single logout

clintonb:


clintonb:
@edx/ecommerce need a volunteer to review.

rlucioni:
@clintonb I'll take a look.

acceptance_tests/mixins.py
```diff
@@ -94,6 +94,18 @@ def register_via_ui(self, course_id=None):
         return username, email, password
 
 
+class LmsLogoutMixin(object):
```
rlucioni: Nitpick: it'd be nice to stay consistent on the capitalization of "LMS."

acceptance_tests/test_auth.py
```diff
@@ -26,3 +24,18 @@ def test_login(self):
...
+
+        self.logout()
+        self.assert_on_lms_logout_page()
+
+    def test_lms_logout(self):
+        """ Verify that logging out of the LMS also logs the user out of Otto. """
+        self.login_with_lms()
+        self.app_login_page.visit()
+        self.logout_via_lms()
+        time.sleep(3)  # Wait for the logout iframes to load
```
rlucioni: The bok-choy [docs](http://bok-choy.readthedocs.io/en/latest/guidelines.html#never-use-time-sleep) strongly discourage use of `time.sleep()`. See the PayPal tests for [an example](https://github.com/edx/ecommerce/blob/7de6446cd3c18766a40fc4dfc4af8f696755bc54/acceptance_tests/mixins.py#L196-L199) of how you can wait for the presence of an iframe more flexibly. You could also use bok-choy's [EmptyPromise](http://bok-choy.readthedocs.io/en/latest/api_reference.html?highlight=promise#bok_choy.promise.EmptyPromise).

rlucioni:
@clintonb looks good. I've left a few suggestions inline. 👍 otherwise.

acceptance_tests/mixins.py
clintonb: Done.

PR #780, opened 2016-06-08
[SOL-1785] Additional invoice fields for coupons.

vkaracic:
FEATURE BRANCH

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -428,6 +428,7 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     num_uses = serializers.SerializerMethodField()
     catalog_query = serializers.SerializerMethodField()
     course_seat_types = serializers.SerializerMethodField()
+    payment_information = serializers.SerializerMethodField()
 
     def retrieve_offer(self, obj):
         """Helper method to retrieve the offer from coupon. """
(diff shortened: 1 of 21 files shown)


ecommerce/extensions/api/serializers.py
```diff
@@ -496,12 +497,21 @@ def get_course_seat_types(self, obj):
         course_seat_types = offer.condition.range.course_seat_types
         return course_seat_types.split(',') if course_seat_types else course_seat_types
 
+    def get_payment_information(self, obj):
+        try:
+            invoice = Invoice.objects.get(order__basket__lines__product=obj)
+            serializer = InvoiceSerializer(invoice)
```
mattdrayer, line 503: It would be best to include only the code being tested within the 'try' block, so you might move serializer and response down to L507

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -412,6 +430,29 @@ def update_coupon_client(self, baskets, client_username):
...
+        Update the invoice data.
+
+        Arguments:
+            coupon (Product): The coupon product with which the invoice is retrieved.
+            data (dict): The request's data from which the invoice data is retrieved
+                         and used for the updated.
+        """
+        invoice = Invoice.objects.filter(order__basket__lines__product=coupon)
+        tds = data.get('tax_deducted_source')
+        update_data = {
```
mattdrayer, line 444: I believe I commented on this already in #797

PR #781, opened 2016-06-08
ECOM-3206 Fix the issuer and make it consistent with LMS

schenedx:
Apparently, if the issuers are not the same between ecommerce and edx-platform, the jwt verification would fail. This change is to make these values consistent between the two platforms
@clintonb @bderusha Please review


clintonb:
👍

PR #782, opened 2016-06-08
mattdrayer/WL-511: Repair invalid SKU publishing/selection workflows

mattdrayer:
@mjfrey @douglashall -- this branch fixes the issue observed in WL-511 -- basically the SKU lookup logic was assuming a single stock record, but in some cases there were multiple stock records to choose from.  I also updated the SKU publishing logic (creates course mode records @ LMS) to only publish bulk SKUs if the product class is one of the allowable types for enrollment codes.

### DIFF
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -77,11 +77,26 @@ def get_basket_switch_data(product):
         switch_link_text = _('Click here to purchase multiple seats in this course')
         structure = 'standalone'
 
-    try:
-        partner_sku = StockRecord.objects.get(
-            product__course_id=product.course_id,
-            product__structure=structure).partner_sku
-    except StockRecord.DoesNotExist:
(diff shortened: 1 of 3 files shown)


ecommerce/extensions/basket/utils.py
```diff
@@ -77,11 +77,25 @@ def get_basket_switch_data(product):
...
 
-    try:
-        partner_sku = StockRecord.objects.get(
+    stock_records = StockRecord.objects.filter(
             product__course_id=product.course_id,
-            product__structure=structure).partner_sku
-    except StockRecord.DoesNotExist:
-        partner_sku = None
-
+            product__structure=structure)
```
douglashall: Should this just be:

```
try:
    partner_sku = StockRecord.objects.get(product=product).partner_sku
    partner_sku = None
```

And then forget the bit below?

mjfrey:
👍

PR #785, opened 2016-06-09
Optimize filtering the courses for Dynamic Catalog Coupons

marjev:


ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -37,10 +38,18 @@ def preview(self, request):
               multiple: false
         """
         query = request.GET.get('query', '')
-        if query:
+        seat_types = request.GET.get('seat_types')
+        if query and seat_types:
```
mjfrey, line 42: Won't query never be None?  Since you are defaulting to ''

marjev, line 42: Values like empty string or empty list are equivalent to None in terms of `if` conditioning. So, if the query is an empty string, it will be treated like None.

But since we don't get anything from setting it to an empty string, I might as well remove the default value.

mjfrey:
👍

PR #786, opened 2016-06-10
mattdrayer/WL-511-seatmode: Consider enrollment code products for mode lookups

mattdrayer:
Enrollment Code product classes were not being considered for product 'course mode' lookups, and one result of this was an invalid enrollment check in certain basket workflows.  Because enrollment code products do not have a 'certificate_type' attribute -- they instead have a 'seat-type' attribute -- we needed to broaden the mode lookup operation, which also required some additional changes/enhancements.

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -190,7 +190,7 @@ class Course(models.Model):
         seat.attr.id_verification_required = id_verification_required
 
         if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and certificate_type in ENROLLMENT_CODE_SEAT_TYPES:
-            self._create_or_update_enrollment_code(certificate_type, partner, price)
+            self._create_or_update_enrollment_code(certificate_type, id_verification_required, partner, price)
 
         if credit_provider:
             seat.attr.credit_provider = credit_provider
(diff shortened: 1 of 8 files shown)


mattdrayer:
@mjfrey FYI

ecommerce/courses/models.py
```diff
@@ -190,7 +190,7 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
         seat.attr.id_verification_required = id_verification_required
 
         if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and certificate_type in ENROLLMENT_CODE_SEAT_TYPES:
-            self._create_or_update_enrollment_code(certificate_type, partner, price)
+            self._create_or_update_enrollment_code(certificate_type, id_verification_required, partner, price)
```
mjfrey: so an enrollment code is always going to be id_verification_required?

ecommerce/courses/models.py
```diff
@@ -273,6 +276,7 @@ def _create_or_update_enrollment_code(self, seat_type, partner, price):
             )
         enrollment_code.attr.course_key = self.id
         enrollment_code.attr.seat_type = seat_type
+        enrollment_code.attr.id_verification_required = id_verification_required
```
mjfrey, line 279: isn't seat_type going to tell you if id_verification is required? For example the seat type is either verified, professional, or no-id-professional.  Can't you infer then from the seat_type rather than having to store in an attribute. Unless I don't understand what seat_type is.

mattdrayer, line 279: The issue I ran into with that was the situation where the cert/seat type was 'professional' but the context was 'no-id-professional' -- it was better to be explicit with the enrollment code.

mattdrayer, line 279: also I'm not actually sure we are setting 'no-id-professional' types on the seat products -- it seems the system is making this decision on the fly based on 'professional' and id_verification_required = False (see ecommerce/courses/utils.py below)

mjfrey, line 279: Take a look in basket/views.py.  We have a routine called _determine_seat_type.  We use this to display a message on the basket page based on if it is professional, professional-no-id, etc...

mjfrey, line 279: Actually I guess you are correct.  It is only professional for seat_type

mjfrey:
:+1:

PR #788, opened 2016-06-10
mattdrayer/ECOM-4683: Return only child seat products in API response

mattdrayer:
@mjfrey @bderusha -- Enrollment Code products were being included in the serialization workflow for the Course Administration Tool course product list.  I've added a filter to return only "child" products for the course (Enrollment Codes are "standalone").

### DIFF
--- a/ecommerce/static/js/models/course_model.js
+++ b/ecommerce/static/js/models/course_model.js
@@ -134,7 +134,7 @@ define([
 
             initialize: function () {
                 this.get('products').on('change:id_verification_required', this.triggerIdVerified, this);
-                this.on('sync', this.removeParentProducts, this);
+                this.on('sync', this.prepareProducts, this);
                 this.on('sync', this.honorModeInit, this);
             },
 
@@ -184,15 +184,19 @@ define([
             },
 
             /**
-             * Removes the parent products from the Product collection.
-             *
-             * This product is never exposed to the user, and should be ignored for all data operations.
+             * Returns the subset of pertinent child seat products from the Product collection.
              */
-            removeParentProducts: function () {
-                var products = this.get('products'),
-                    parents = products.where({structure: 'parent'});
+            prepareProducts: function () {
+                // Create a reference to the current product collection
+                var products = this.get('products');
 
+                // Ignore any parent product models
+                var parents = products.where({structure: 'parent'});
                 products.remove(parents);
+
+                // Ignore any non-Seat models (such as an Enrollment Code)
+                var seats = products.where({product_class: 'Seat'});
+                products.reset(seats);
             },
 
             /**
@@ -203,7 +207,9 @@ define([
             seats: function () {
                 return this.get('products').filter(function (product) {
                     // Filter out parent products since there is no need to display or modify.
-                    return (product instanceof CourseSeat) && product.get('structure') !== 'parent';
+                    return (product instanceof CourseSeat) &&
+                        product.get('structure') !== 'parent' &&
+                        product.get('product_class') === 'Seat';
                 });
             },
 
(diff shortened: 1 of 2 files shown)


clintonb:
There is value to including all products, for example if/when we want to include a link to the enrollment code product on the marketing site. This filtering should be done by the Course Administration Tool.

mattdrayer:
As I understand it, the intent isn't to list the enrollment code directly -- the bulk purchasing variation of any given seat product is indicated by the 'bulk_sku' attribute, which may or may not contain a value.

mattdrayer:
This pull request depends on https://github.com/edx/ecommerce/pull/786

mattdrayer:
Would love to get through this PR today -- @mjfrey @bderusha @clintonb --  do you guys have anything you'd like to see changed or can I move forward?

mjfrey:
LGTM:
👍

clintonb:
This introduces a backwards-incompatible change to the API that necessitates a major version bump. Instead of changing the API, change the client code. Update CAT to only look at course seat child products. Otherwise, we will run into this issue the next time we add a course-related product.

mattdrayer:
I'm not sure I understand why this is being viewed as a backwards-incompatible change -- the API endpoint has not changed, nor the data schema.  In retrospect, the incompatible change was actually the addition of the new Enrollment Code product class, which was not expected by the CAT client and resulted the misbehavior of that tool.  So with this change set we are actually repairing the original semantics of the 'course product list' API operation by introducing this filtering.

ecommerce/static/js/models/course_model.js
```diff
@@ -135,6 +135,7 @@ define([
             initialize: function () {
                 this.get('products').on('change:id_verification_required', this.triggerIdVerified, this);
                 this.on('sync', this.removeParentProducts, this);
+                this.on('sync', this.removeStandaloneProducts, this);
```
mattdrayer: @clintonb -- here's one approach from the client/CAT side that solves the issue -- let me know if it's acceptable and I'll remove the server-side solution from this PR.

clintonb: The approach is correct; however, you should be removing products where `product_class` is not equal to `Seat`.

clintonb:
The server-side change is backwards-incompatible in the sense that it removes data that was previously being provided by the API. We intentionally opted not to limit the data returned in the `products` array to avoid later confusion about what is/is not included.

The client-side change is much more desirable. You just need to update it to filter on product class.

ecommerce/static/js/models/course_model.js
```diff
@@ -184,18 +184,52 @@ define([
...
+                var parents = products.where({structure: 'parent'});
                 products.remove(parents);
             },
 
             /**
+             * Removes "standalone" products from the Product collection.
+             *
+             * This product is never exposed to the user, and should be ignored for all data operations.
+             */
+            removeStandaloneProducts: function () {
```
clintonb: Is this necessary? I'm not aware of any standalone seat products.

mattdrayer: I guess it's not at this point -- "Enrollment Code" product_class products have 'standalone' structure -- I was thinking that maybe the scenario would arise at some point, but no sense having this hanging around to cause problems if it's not needed.

ecommerce/extensions/api/v2/tests/views/test_courses.py
```diff
@@ -49,7 +49,7 @@ def serialize_course(self, course, include_products=False):
         }
 
         if include_products:
-            data['products'] = [self.serialize_product(product) for product in course.products.all()]
+            data['products'] = [self.serialize_product(product) for product in course.seat_products]
```
mattdrayer: This is an artifact from the server-side attempt, I'll remove shortly

clintonb:
👍 (once commits squashed, and comments addressed)

Note that you'll need DevOps to deploy to stage as there are migrations in the latest batch of commits. Feel free to rebase the tag `release-2016-06-14`.

mattdrayer:
Cool -- PR is cleaned up now -- will merge and update the tag after the build goes green

PR #790, opened 2016-06-10
mattdrayer/ECOM-4682: Improve support for Audit seat products in publishing worfklow

mattdrayer:
@bderusha @mjfrey @clintonb -- Enrollment Code products were being included by this seat product serializer in its response, which was causing invalid downstream behavior

### DIFF
--- a/ecommerce/courses/publishers.py
+++ b/ecommerce/courses/publishers.py
@@ -33,11 +33,13 @@ class LMSPublisher(object):
     def serialize_seat_for_commerce_api(self, seat):
         """ Serializes a course seat product to a dict that can be further serialized to JSON. """
         stock_record = seat.stockrecords.first()
+
         bulk_sku = None
-        if seat.attr.certificate_type in ENROLLMENT_CODE_SEAT_TYPES:
+        if getattr(seat.attr, 'certificate_type', '') in ENROLLMENT_CODE_SEAT_TYPES:
             enrollment_code = seat.course.enrollment_code_product
             if enrollment_code:
                 bulk_sku = enrollment_code.stockrecords.first().partner_sku
+
         return {
             'name': mode_for_seat(seat),
             'currency': stock_record.price_currency,
--- a/ecommerce/courses/tests/test_models.py
+++ b/ecommerce/courses/tests/test_models.py
@@ -38,9 +38,10 @@ class CourseTests(CourseCatalogTestMixin, TestCase):
         self.assertEqual(len(course.seat_products), 0)
 
         # Create the seat products
+        toggle_switch(ENROLLMENT_CODE_SWITCH, True)
         seats = [course.create_or_update_seat('honor', False, 0, self.partner),
                  course.create_or_update_seat('verified', True, 50, self.partner)]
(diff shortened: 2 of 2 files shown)


ecommerce/courses/publishers.py
```diff
@@ -32,22 +32,24 @@ def get_course_verification_deadline(self, course):
 
     def serialize_seat_for_commerce_api(self, seat):
         """ Serializes a course seat product to a dict that can be further serialized to JSON. """
-        stock_record = seat.stockrecords.first()
-        bulk_sku = None
-        if seat.attr.certificate_type in ENROLLMENT_CODE_SEAT_TYPES:
-            try:
+        # Exclude enrollment code products
+        if not hasattr(seat.attr, 'seat_type'):
```
clintonb: Use the `seat.get_product_class()` instead of looking at attributes.

mattdrayer:
This PR depends on #786

ecommerce/courses/publishers.py
mattdrayer: Updated!

mattdrayer:
@clintonb @bderusha @mjfrey -- coverage is now 100% on this fix -- let me know if there's anything else you'd like to see -- would be great to merge+ship tomorrow.

ecommerce/courses/publishers.py
```diff
@@ -32,20 +32,24 @@ def get_course_verification_deadline(self, course):
 
     def serialize_seat_for_commerce_api(self, seat):
```
clintonb, line 33: This method should only called for seat products. See https://github.com/edx/ecommerce/blob/41a070d0709120baddd65a76ae14ecf1c6088b99/ecommerce/courses/publishers.py#L94. If there is an issue, it should be corrected in `Course.seat_products`.

mattdrayer, line 33: Sure, that works for me -- I've updated the seat_products property to return only child seat products.

ecommerce/courses/models.py
```diff
@@ -105,7 +106,9 @@ def parent_seat_product(self):
     @property
     def seat_products(self):
         """ Returns a queryset of course seat Products related to this course. """
-        return self.parent_seat_product.children.all().prefetch_related('stockrecords')
+        children = self.parent_seat_product.children.all()
+        child_ids = [child.id for child in children if child.get_product_class().name == SEAT_PRODUCT_CLASS_NAME]
```
clintonb: Why does the parent seat product have children that aren't seats? You're still just treating the symptom of a larger problem here. Some other bit of code is violating the expectation that the parent seat's children are all also seats.

mattdrayer: Alright, we'll keep digging...

mattdrayer:
@clintonb -- the enrollment codes concern appears to have been a red herring -- I've narrowed down the change set to only what seems to be necessary.  The issue stems from the CAT posting back an "Audit" seat model with no 'certificate_type' attribute, and the original check against that attribute to determine the need for an enrollment code lookup was failing.

mjfrey:
👍

clintonb:
:+1:

PR #793, opened 2016-06-13
WL-495: Update edx-django-sites-extensions version to 2.0.1 for ecommerce

saleem-latif:
Hi @mattdrayer , @douglashall , @clintonb 

Kindly review this PR, I have updated the version of edx-django-sites-extensions from `1.0.0` to `2.0.1`.

**Details of Changes:**
With the new version of `edx-django-sites-extensions`, the way default sites are handled is changed, old middleware named `CurrentSiteWithDefaultMiddleware` is now removed and middleware `CurrentSiteMiddleware` is used instead now, which is monkey patched to offer default site functionality.

### DIFF
--- a/ecommerce/extensions/checkout/tests/test_mixins.py
+++ b/ecommerce/extensions/checkout/tests/test_mixins.py
@@ -212,6 +212,7 @@ class EdxOrderPlacementMixinTests(BusinessIntelligenceMixin, RefundTestMixin, Te
         Verify the send confirmation message override functions as expected
         """
         request = RequestFactory()
+        request.get_host = lambda: None
         user = self.create_user()
         user.email = 'test_user@example.com'
         request.user = user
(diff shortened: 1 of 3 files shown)


mattdrayer:
It would be great if we could include a test to validate the end result we are expecting from this middleware -- we do not have to test the middleware implementation itself, but something that would break if the middleware was accidentally disabled or misconfigured would be nice to have.

clintonb:
I cannot approve of unnecessary monkey-patching. The process of creating a custom middleware that checks the request before `SITE_ID` is tedious, but very much doable, especially given that we already did it. Why the change?

douglashall:
As far as I know this upgrade is not required. The edx-django-sites-extensions 1.0 way of allowing for a default site works well with Otto. Just be aware that if you ever run into any third-party code that wants to call `django.contrib.sites.models.SiteManager.get_current()` directly, you may need to upgrade to edx-django-sites-extensions 2.0.1.

mattdrayer:
Okay, I'm closing this PR and the corresponding JIRA ticket as a 'won't do'.

PR #795, opened 2016-06-13
Coupon UI bug fixes.

vkaracic:
- add IDs to input fields and bind them to the label. Using `for` in labels without the accompanying ID is pointless https://developer.mozilla.org/en/docs/Web/HTML/Element/label
- properly trigger change for upper benefit value change. This change was not binded to the right element, so the user could change to a fixed benefit amount, change back to percent and enter any number and the coupon would be saved
- for hidden fields set values to default on change. The entered values would remain on a change event so that a user could select SINGLE_USE type, enter 10 in the `Number of Codes` field, change to MULTI_USE, enter 20 in `Maximum Number of Uses` and thus create 10 codes with 20 max usage.

### DIFF
--- a/ecommerce/static/js/views/coupon_form_view.js
+++ b/ecommerce/static/js/views/coupon_form_view.js
@@ -101,15 +101,7 @@ define([
                     }
                 },
                 'input[name=benefit_type]': {
-                    observe: 'benefit_type',
-                    onGet: function (val) {
-                        if (val === 'Percentage') {
-                            return val;
-                        } else if (val === 'Absolute') {
-                            return 'Absolute';
-                        }
-                        return '';
-                    }
+                    observe: 'benefit_type'
                 },
                 '.benefit-addon': {
                     observe: 'benefit_type',
@@ -204,7 +196,7 @@ define([
             },
 
             changeUpperLimitForBenefitValue: function () {
-                var is_benefit_percentage = this.$el.find('[name=code_type]').val() === 'Percentage',
+                var is_benefit_percentage = this.$el.find('[name=benefit_type]:checked').val() === 'Percentage',
                     max_value = is_benefit_percentage ? '100' : '';
 
                 this.$el.find('[name=benefit_value]').attr('max', max_value);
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=795&repo=edx%2Fecommerce

mjfrey:
👍

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -204,7 +204,7 @@ define([
             },
 
             changeUpperLimitForBenefitValue: function () {
-                var is_benefit_percentage = this.$el.find('[name=code_type]').val() === 'Percentage',
+                var is_benefit_percentage = this.$el.find('[name=benefit_type]:checked').val() === 'Percentage',
```
marjev, line 199: Since we use Backbone.stickit `this.$el.find('[name=benefit_type]:checked').val()` is saved in `this.model.benefit_type`, so for the sake of code readability you might consider using it.

ecommerce/static/templates/coupon_form.html
```diff
@@ -4,17 +4,17 @@
     <div class="fields col-md-6">
         <div class="form-group">
             <label for="title"><%= gettext('Coupon Name') %> *</label>
-            <input type="text" class="form-control" name="title" maxlength="30">
+            <input id="title" type="text" class="form-control" name="title" maxlength="30">
```
marjev, line 7: Do we use id somewhere? I don't see the usage in the PR code.

marjev:
Left a couple of comments, otherwise looks good 👍

ecommerce/static/templates/coupon_form.html
vkaracic, line 7: ID's are needed for the label's `for` attribute.

ecommerce/static/js/views/coupon_form_view.js
vkaracic, line 199: The value of `this.model.benefit_type` is `undefined` when I try using that one.

vkaracic, line 199: Also, while I'm at it, I'm going to remove this function: https://github.com/edx/ecommerce/pull/795/files#diff-bd215f92302c2632268d1d9751012b3fR105
seems pointless now, right?

PR #796, opened 2016-06-14
SOL-1834 Use coupon ID when creating ConditionalOffer models

douglashall:
During the fulfillment step in the Otto bulk purchase workflow, ConditionalOffer models were created with a "None" coupon ID in their name and slug fields. This fixes the issue.

@vkaracic @mjfrey

FYI @mattdrayer

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -211,7 +211,6 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
                 start_datetime=data['start_date'],
                 voucher_type=data['voucher_type'],
                 max_uses=data['max_uses'],
-                coupon_id=coupon_product.id,
                 catalog_query=data['catalog_query'],
                 course_seat_types=data['course_seat_types']
             )
--- a/ecommerce/extensions/voucher/utils.py
+++ b/ecommerce/extensions/voucher/utils.py
@@ -333,7 +333,6 @@ def create_vouchers(
         quantity,
         start_datetime,
         voucher_type,
-        coupon_id=None,
         code=None,
         max_uses=None,
         _range=None,
@@ -380,7 +379,7 @@ def create_vouchers(
         benefit_type=benefit_type,
         benefit_value=benefit_value,
         max_uses=max_uses,
-        coupon_id=coupon_id
+        coupon_id=coupon.id
     )
     for __ in range(quantity):
         voucher = _create_new_voucher(


ecommerce/extensions/voucher/utils.py
```diff
@@ -380,7 +379,7 @@ def create_vouchers(
         benefit_type=benefit_type,
         benefit_value=benefit_value,
         max_uses=max_uses,
-        coupon_id=coupon_id
+        coupon_id=coupon.id
```
mjfrey, line 382: How does this fix the issue,  how is this being called from the bulk purchase flow?  Why is the coupon_id = None.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -211,7 +211,6 @@ def create_coupon_product(self, title, price, data):
                 start_datetime=data['start_date'],
                 voucher_type=data['voucher_type'],
                 max_uses=data['max_uses'],
-                coupon_id=coupon_product.id,
```
mjfrey, line 214: I just don't understand how this is ever None?

ecommerce/extensions/voucher/utils.py
douglashall, line 382: See https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/fulfillment/modules.py#L410. The bulk purchase workflow invocation of `create_vouchers` does not pass the `coupon_id` keyword argument.

ecommerce/extensions/api/v2/views/coupons.py
douglashall, line 214: See the comment below. Was there a reason you decided to include a `coupon_id` keyword argument in addition to the `coupon` keyword argument which was already being sent?

mjfrey, line 214: I can't remember actually.

ecommerce/extensions/voucher/utils.py
mjfrey, line 382: I see.  How has this been working for Matt at all?

mattdrayer, line 382: Wow, great catch @douglashall -- and yeah I agree with @mjfrey -- how has this been working for so long, over so many test runs, with this bug present?  It seems like it should have caused issues right from the beginning, but for some reason it's remained hidden up until today.

ecommerce/extensions/api/v2/views/coupons.py
mattdrayer, line 214: I can see how it is None -- however, I can't see how it was ever NOT None...

mattdrayer:
👍

vkaracic:
I've been meaning to remove that ridiculous `coupon_id` argument, can't remember what I was thinking with that - thanks for fixing this!
The reason why it was working is somewhere in the `get_or_create` method and it "getting" the object instead of creating a new one since the `name` attribute is unique. Although some of the other parameters should be different. I'm happy to investigate if you can point me to, and grant access, to a server/sandbox where you guys are testing this.

Thanks again!
:+1:

PR #797, opened 2016-06-14
[SOL-1785] Backend side for new invoice fields.

vkaracic:


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=797

ecommerce/extensions/api/serializers.py
```diff
@@ -496,12 +497,17 @@ def get_course_seat_types(self, obj):
         course_seat_types = offer.condition.range.course_seat_types
         return course_seat_types.split(',') if course_seat_types else course_seat_types
 
+    def get_payment_information(self, obj):
+        invoice = Invoice.objects.get(order__basket__lines__product=obj)
```
mattdrayer: Need a docstring

mattdrayer: Need to defend against DoesNotExist exceptions

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -101,6 +101,16 @@ def create(self, request, *args, **kwargs):
             catalog_query = request.data.get(CATALOG_QUERY)
             course_seat_types = request.data.get(COURSE_SEAT_TYPES)
 
+            invoice_data = {
+                'number': request.data.get('invoice_number'),
```
mattdrayer: Will it be a problem if any of these parameters return None for their default values, versus a type-specific value?

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -413,6 +429,27 @@ def update_coupon_client(self, baskets, client_username):
             user, __ = User.objects.get_or_create(username=client_username)
             baskets.update(owner=user)
 
+    def update_invoice_data(self, coupon, data):
```
mattdrayer, line 435: Does it make sense to combine this with the code above on L104

ecommerce/extensions/payment/processors/invoice.py
```diff
@@ -18,24 +18,30 @@ class InvoicePayment(BasePaymentProcessor):
...
         event_type, __ = PaymentEventType.objects.get_or_create(
             name=PaymentEventTypeName.PAID)
         event = PaymentEvent(event_type=event_type, processor_name=self.NAME)
 
-        # Create an Invoice.
-        Invoice.objects.create(order=order, business_client=business_client)
-
+        invoice = Invoice.objects.create(order=order, business_client=business_client)
+        if invoice_data:
+            invoice.number = invoice_data.get('number')
```
mattdrayer, line 37: As above, we probably should consider None values and missing keys here.

ecommerce/invoice/models.py
```diff
@@ -10,11 +11,35 @@ class Invoice(TimeStampedModel):
         (NOT_PAID, _('Not Paid')),
         (PAID, _('Paid')),
     )
+    PREPAID, POSTPAID = 'Prepaid', 'Postpaid'
```
mattdrayer: Are these constants that should instead live in ecommerce.core.constants?

ecommerce/invoice/models.py
```diff
@@ -10,11 +11,35 @@ class Invoice(TimeStampedModel):
...
+    type_choices = (
+        (PREPAID, _('Prepaid')),
+        (POSTPAID, _('Postpaid'))
+    )
+    PERCENTAGE, FIXED = 'Percentage', 'Fixed'
```
mattdrayer, line 20: Same with these -- should they move to ecommerce.core.constants?

ecommerce/invoice/models.py
```diff
@@ -10,11 +11,35 @@ class Invoice(TimeStampedModel):
...
+    discount_type_choices = (
+        (PERCENTAGE, _('Percentage')),
+        (FIXED, _('Fixed'))
+    )
     basket = models.ForeignKey('basket.Basket', null=True, blank=True)
     order = models.ForeignKey('order.Order', null=True, blank=False)
     business_client = models.ForeignKey('core.BusinessClient', null=True, blank=False)
     state = models.CharField(max_length=255, default=NOT_PAID, choices=state_choices)
```
mattdrayer, line 28: Should 'NOT_PAID' be included in type_choices above?

ecommerce/invoice/models.py
```diff
@@ -10,11 +11,35 @@ class Invoice(TimeStampedModel):
...
+    type = models.CharField(max_length=255, default=PREPAID, choices=type_choices, null=True, blank=True)
+    invoiced_amount = models.PositiveIntegerField(null=True, blank=True)
+    payment_date = models.DateTimeField(null=True, blank=True)
+    discount_type = models.CharField(
+        max_length=255, default=PERCENTAGE, choices=discount_type_choices, null=True, blank=True
+    )
+    discount_value = models.PositiveIntegerField(null=True, blank=True)
+    tax_deducted_source = models.PositiveIntegerField(
+        validators=[MinValueValidator(1), MaxValueValidator(100)],
+        blank=True, null=True
```
mattdrayer: Nit: all the other model fields define 'null' before 'blank'

ecommerce/extensions/api/serializers.py
vkaracic: Currently, creating a coupon creates an invoice, if an invoice does not exist for a coupon there is a problem.

ecommerce/extensions/api/v2/views/coupons.py
vkaracic: No, that was the intention since there are a few options for invoices each with it's combinations of invoice fields.

ecommerce/extensions/api/v2/views/coupons.py
vkaracic, line 435: Combine how?

ecommerce/extensions/payment/processors/invoice.py
vkaracic, line 37: `get` returns `None` for non-existing keys.

ecommerce/invoice/models.py
vkaracic: No, these are model-specific and I'd like to keep them "Oscar-like" ie. in the model:
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/basket/abstract_models.py#L35-L43
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/voucher/abstract_models.py#L36-L42

ecommerce/invoice/models.py
vkaracic, line 28: No, there is a difference between invoice type and state. For example a postpaid invoice can be paid and not paid.

ecommerce/extensions/api/v2/views/coupons.py
vkaracic, line 435: Ohh, you probably mean the retrieval of the data from the request. Yes, you are quite right!

ecommerce/invoice/models.py
mjfrey, line 28: Right.  There are 2 things.  The type of invoice and the state. Which I should point out -- we are not really using yet.

vkaracic:
@mattdrayer comments have been addressed!

marjev:
LGTM 👍

ecommerce/invoice/models.py
```diff
@@ -10,11 +11,35 @@ class Invoice(TimeStampedModel):
...
+    discount_type_choices = (
+        (PERCENTAGE, _('Percentage')),
+        (FIXED, _('Fixed'))
+    )
     basket = models.ForeignKey('basket.Basket', null=True, blank=True)
     order = models.ForeignKey('order.Order', null=True, blank=False)
     business_client = models.ForeignKey('core.BusinessClient', null=True, blank=False)
     state = models.CharField(max_length=255, default=NOT_PAID, choices=state_choices)
 
+    # New invoice fields.
```
mjfrey: NIT: Maybe this description could be better.  "Additional Invoice fields to support both pre-paid and post-paid invoices."

ecommerce/extensions/payment/processors/invoice.py
mattdrayer, line 37: Right, that's my point -- is it okay with invoice.save() if a value of None is returned for any of these keys?

mattdrayer:
Just one question about None values being acceptable for invoice data fields during the .save() -- if that's not an issue I'm 👍

mjfrey:
:+1: 
Squash commits

ecommerce/extensions/payment/processors/invoice.py
vkaracic, line 37: Yes, we even have an invoice type (N/A) that returns `None` for every of the fields.

PR #798, opened 2016-06-14
Add the current site domain to the email template context

douglashall:
We need to know the current site domain when rendering email templates so that we can properly construct absolute URLs to the HTML email header logo image.

@mattdrayer @saleem-latif 

FYI @mjfrey


mattdrayer:
It would be really great if we could add a test that would protect our theme templates against unexpected changes to contexts -- we have no other way of knowing that these templates will break if attributes such as site_domain are removed in the future.

mattdrayer:
But it's not a functional requirement, I guess -- 👍

PR #800, opened 2016-06-15
Fix course catalog info query and unit test

douglashall:
@JoyceZhu 

Looks like the problem had to do with the way you were trying to use the course catalog api client. Take a look and let me know if you have any questions.

### DIFF
--- a/ecommerce/coupons/tests/mixins.py
+++ b/ecommerce/coupons/tests/mixins.py
@@ -2,12 +2,12 @@ import datetime
 import logging
 import json
 
+import httpretty
 from django.conf import settings
 from django.test import RequestFactory
 from oscar.test import factories
 
 from ecommerce.core.models import BusinessClient
-from ecommerce.core.tests.patched_httpretty import httpretty
 from ecommerce.extensions.api.v2.views.coupons import CouponViewSet
 from ecommerce.extensions.basket.utils import prepare_basket
 from ecommerce.tests.factories import PartnerFactory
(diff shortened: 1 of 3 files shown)


ecommerce/coupons/tests/mixins.py
```diff
@@ -34,12 +34,10 @@ def mock_dynamic_catalog_course_runs_api(self, course_run=None, query=None):
...
+        course_run_url = '{}course_runs/?q={}'.format(
             settings.COURSE_CATALOG_API_URL,
-            # query if query else 'id:course*'
-            course_run.id
+            query if query else 'id:course*'
         )
-        log.debug("URL %s JSON %s", course_run_url, course_run_info_json)
         httpretty.register_uri(
             httpretty.GET, course_run_url,
             body=course_run_info_json,
```
douglashall, line 43: This reverts the changes you had made to this file.

ecommerce/courses/utils.py
```diff
@@ -29,6 +29,10 @@ def get_course_info(site, course_key):
     course = cache.get(cache_hash)
     if not course:  # pragma: no cover
         log.debug('Hitting URL: %s', api.course_runs(course_key)._store['base_url'])
-        course = api.course_runs(course_key).get()
-        cache.set(cache_hash, course, settings.COURSES_API_CACHE_TIMEOUT)
+        response = api.course_runs.get(q='key={}'.format(course_key))
```
JoyceZhu, line 32: Note: looks like the key value needs to be enclosed in quotes to work with the basket view endpoint --

`response = api.course_runs.get(q='key="{}"'.format(course_key))`

JoyceZhu:
Other than the small quote thing I mentioned this works great -- thanks for all the help!

PR #801, opened 2016-06-15
Fixed incorrect login redirect URL view name

clintonb:
ECOM-4714


clintonb:
FYI @mattdrayer @mjfrey

mattdrayer:
:shipit: 👍

PR #802, opened 2016-06-16
[SOL-1785] Frontend changes for new invoice fields.

vkaracic:


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=802

mjfrey:
@AlasdairSwan can you please take a look.

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -84,6 +114,13 @@ define([
                     usage: this.usageLimitation(voucher),
                     category: category,
                     note: note,
+                    invoice_type: invoice_type,
+                    invoice_number: invoice_number,
+                    invoice_payment_date: invoice_payment_date,
+                    invoice_discount_type: invoice_discount_type,
+                    invoice_discount_value: invoice_discount_value,
+                    invoiced_amount: invoiced_amount,
+                    tax_deducted_source_value: tax_deducted_source
```
AlasdairSwan: There is a lot of processing/data manipulation going on in the render function, it would be cleaner to move this to a `formatInvoiceData` function which returns an invoice object that can be passed to the template.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -202,6 +248,24 @@ define([
...
+            },
+
+            toggleDollarPercentIcon: function (val) {
+                if (val === 'Percentage') {
+                    return '%';
+                } else if (val === 'Absolute') {
+                    return '$';
+                } else {
+                    return '';
+                }
```
AlasdairSwan, line 265: I don't like there being so many return statements in a function. It would be easier to read is you set a local icon variable to `''` and then update it in the `if...else if` and return that variable at the end of the function.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -202,6 +248,24 @@ define([
                 this.$el.find('[name=benefit_value]').attr('max', max_value);
             },
 
+            changeUpperLimitForInvoiceDiscountValue: function () {
+                var is_invoice_discount_percentage = this.$el.find(
```
AlasdairSwan: Nitpick: `this.$(` is the same as `this.$el.find(`
Please update the code throughout to use this shorthand convention.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -232,6 +296,47 @@ define([
...
                 }
             },
+ 
+            toggleInvoiceFields: function () {
+                var invoice_type = this.$el.find('[name=invoice_type]:checked').val(),
+                    prepaid_fields = [
+                        '[name=invoice_number]',
+                        '[name=invoice_payment_date]'
+                    ],
+                    self = this;
```
AlasdairSwan: Should not need to alias `this`, you can bind it to each loops.

ecommerce/static/sass/partials/views/_coupon_admin.scss
```diff
@@ -62,6 +72,12 @@
     }
 }
 
+@media(max-width: $screen-md-max) {
+    .invoice-type > div {
+        width: 100% !important;
```
AlasdairSwan: Why do we need the `!important`? It should only be used in edge cases.

ecommerce/static/templates/coupon_form.html
```diff
@@ -81,22 +81,93 @@
...
                 <div class="input-group-addon">$</div>
                 <input id="price" type="number" step="0.01" min="0" class="form-control" name="price">
             </div>
             <p class="help-block"></p>
         </div>
+
+        <div class="form-group">
+            <label for="invoice_payment_date"><%= gettext('Payment Date') %></label>
+            <div class="input-group">
+                <div class="input-group-addon"><i class="fa fa-calendar" aria-hidden="true"></i></div>
```
AlasdairSwan: Should use `span` instead of `i`

ecommerce/static/templates/coupon_form.html
```diff
@@ -81,22 +81,93 @@
             <input id="client" type="text" class="form-control" name="client" maxlength="50">
             <p class="help-block"></p>
         </div>
+ 
+        <hr>
 
         <div class="form-group">
-            <label for="price"><%= gettext('Total to Invoice to Client') %></label>
+            <label><%= gettext('Invoice type') %></label>
```
srpearce: For capitalization consistency, this should be "Invoice Type". I'm fine with keeping the options below in sentence caps.

ecommerce/static/templates/coupon_form.html
```diff
@@ -81,22 +81,93 @@
...
+        </div>
+
+        <div class="form-group">
+            <label for="invoice_number"><%= gettext('Invoice Number') %></label>
+            <input id="invoice_number" class="form-control" type="text" name="invoice_number">
+            <p class="help-block"></p>
+        </div>
+
+        <div class="form-group">
+            <label for="price"><%= gettext('Invoiced amount') %></label>
```
srpearce: Invoiced Amount

ecommerce/static/templates/coupon_form.html
```diff
@@ -81,22 +81,93 @@
...
+        <hr>
+
         <div class="form-group">
             <label for="total_value"><%= gettext('Total Value of Coupon') %></label>
             <div class="input-group">
                 <div class="input-group-addon">$</div>
                 <input id="total_value" type="number" step="0.01" min="0" class="form-control" name="total_value" disabled>
             </div>
-            <p class="help-block"><%= gettext('Number of coupon codes X unit price per seat') %></p>
+            <p class="help-block"><%= gettext('Number of coupon codes * unit price per seat.') %></p>
```
srpearce: Is there no way to include a multiplication sign (×) here? If not, I would prefer a lowercase x to an uppercase X or an asterisk.

I would also delete the period, since we don't use terminal periods elsewhere in the UI.

ecommerce/static/js/views/coupon_form_view.js
vkaracic, line 265: Okay. Is this more of a stylistic change or is there some functional value behind that change?

ecommerce/static/sass/partials/views/_coupon_admin.scss
vkaracic: Sorry, I forgot to remove it.

ecommerce/static/js/views/coupon_form_view.js
vkaracic: I wasn't aware of the `[context]` argument in `each`. Thanks!

ecommerce/static/templates/coupon_form.html
vkaracic: Okay. What's the reasoning behind this change?

ecommerce/static/templates/coupon_form.html
vkaracic: You are quite right :)

ecommerce/static/templates/coupon_form.html
vkaracic: You are right, a lowercase `x` is better. I'm inclined to leave the dot at the end of the sentence, if that's OK?

The CAT might need an update:
screenshot from 2016-06-20 09-33-55

vkaracic:
@AlasdairSwan @srpearce thank you guys for taking the time to review my PR, I appreciate it!
I believe I have addressed all your comments, and left some questions, can you please go over it once more?

ecommerce/static/js/views/coupon_form_view.js
AlasdairSwan, line 265: Stylistic/code readability.

ecommerce/static/templates/coupon_form.html
AlasdairSwan: `<i>` is semantically incorrect, it is for italicizing strings.

vkaracic: Hah, right! Didn't think of that.

ecommerce/static/js/models/coupon_model.js
```diff
@@ -187,6 +188,19 @@ define([
...
+                var payment_information = this.get('payment_information'),
+                    invoice = payment_information.Invoice,
+                    tax_deducted = invoice.tax_deducted_source ? 'Yes' : 'No';
+                this.set('invoice_type', invoice.type);
+                this.set('invoice_discount_type', invoice.discount_type);
+                this.set('invoice_discount_value', invoice.discount_value);
+                this.set('invoice_number', invoice.number);
+                this.set('invoice_payment_date', invoice.payment_date);
+                this.set('tax_deducted_source', invoice.tax_deducted_source);
+                this.set('tax_deduction', tax_deducted);
```
AlasdairSwan: You can set multiple parameters when using set, so it makes sense to just call `this.set` once, passing it an object with the changes.

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -65,12 +65,58 @@ define([
...
+            },
+
+            formatInvoiceData: function() {
+                var invoice_type = this.model.get('invoice_type'),
+                    invoice_number = this.model.get('invoice_number'),
+                    invoice_payment_date = this.model.get('invoice_payment_date'),
+                    invoice_discount_type = this.model.get('invoice_discount_type'),
+                    invoice_discount_value = this.model.get('invoice_discount_value'),
+                    invoiced_amount = this.model.get('invoiced_amount'),
+                    tax_deducted_source = this.model.get('tax_deducted_source');
```
AlasdairSwan, line 86: Not all of these variables are used before the return statement. To make the code easier to read I would only create variables for values you are manipulating and pull the values form the model in the return object (e.g. `invoice_type: this.model.get('invoice_type')`

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -84,6 +130,13 @@ define([
                     usage: this.usageLimitation(voucher),
                     category: category,
                     note: note,
+                    invoice_type: invoice_data.invoice_type,
+                    invoice_number: invoice_data.invoice_number,
+                    invoice_payment_date: invoice_data.invoice_payment_date,
+                    invoice_discount_type: invoice_data.invoice_discount_type,
+                    invoice_discount_value: invoice_data.invoice_discount_value,
+                    invoiced_amount: invoice_data.invoiced_amount,
+                    tax_deducted_source_value: invoice_data.tax_deducted_source
```
AlasdairSwan: You can use `$.extend` to do this in a less verbose way (e.g. extend the invoice_data object with the other values).

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -196,14 +242,34 @@ define([
...
+            },
+
+            toggleDollarPercentIcon: function (val) {
+                var icon;
+                if (val === 'Percentage') {
+                    icon = '%';
+                } else if (val === 'Absolute') {
+                    icon = '$';
+                } else {
+                    icon = '';
```
AlasdairSwan: If you default the value of `icon` to `''` you don't need this last else block.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -367,30 +474,30 @@ define([
...
-                this.$el.find('input[name=catalog_type]').attr('disabled', true);
+                this.$('select[name=code_type]').attr('disabled', true);
+                this.$('select[name=voucher_type]').attr('disabled', true);
+                this.$('input[name=quantity]').attr('disabled', true);
+                this.$('input[name=course_id]').attr('disabled', true);
+                this.$('input[name=code]').attr('disabled', true);
+                this.$('input[name=benefit_type]').attr('disabled', true);
+                this.$('select[name=seat_type]').attr('disabled', true);
+                this.$('input[name=max_uses]').attr('disabled', true);
+                this.$('input[name=catalog_type]').attr('disabled', true);
```
AlasdairSwan: If you add a custom class to non editable fields this function becomes 1 line (e.g. `this.$('.non-editable').attr('disabled', true);`)

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -430,9 +539,13 @@ define([
                     this.model.set('category', defaultCategory[0].id);
                     this.model.set('benefit_type', 'Percentage');
                     this.model.set('catalog_type', 'Single course');
-                    this.$el.find('[name=benefit_value]').attr('max', 100);
-                    this.$el.find('button[type=submit]').html(gettext('Create Coupon'));
-                    this.$el.find('.catalog-query').removeClass('editing');
+                    this.model.set('invoice_discount_type', 'Percentage');
+                    this.model.set('invoice_type', 'Prepaid');
+                    this.model.set('tax_deduction', 'No');
```
AlasdairSwan: You should only call `this.model.set` once, passing it an object containing all changes.

ecommerce/static/sass/partials/views/_coupon_admin.scss
```diff
@@ -1,12 +1,22 @@
...
         font-weight: normal;
     }
+
     .benefit-addon {
-        width: 40px;
+        width: $input-addon-width;
     }
+
+    input[name="invoice_discount_value"],
     input[name="benefit_value"] {
```
AlasdairSwan: Classes should be used for styling instead of elements and names.

ecommerce/static/sass/partials/views/_coupon_admin.scss
```diff
@@ -109,7 +125,14 @@
...
         .catalog-query,
-        .seat-types {
+        .seat-types,
+        .invoice-type,
+        .invoice-number,
+        .invoiced-amount,
+        .invoice-discount-type,
+        .invoice-discount-value,
+        .invoice-payment-date,
+        .tax-deducted-source-value {
```
AlasdairSwan: It would be better to use a single, re-usable class to keep the code DRY and reduce the size of the compiled CSS.

ecommerce/static/templates/coupon_form.html
```diff
@@ -81,22 +81,82 @@
...
+
+        <div class="form-group hidden">
+            <div class="input-group tax_deducted_source_value">
+                <input id="tax_deducted_source_value" type="number" step="1" min="1" max="100" class="form-control" name="tax_deducted_source_value">
+                <div class="input-group-addon">%</div>
+                <p class="help-block"></p>
             </div>
-            <p class="help-block"><%= gettext('Number of coupon codes X unit price per seat') %></p>
         </div>
     </div>
```
AlasdairSwan, line 162: Throughout this file you use underscores for ids and classes (e.g. `invoice_discount_value`). The codebase follows the convention of using dashes instead (e.g. `invoice-discount-value`). Please update this pr to use dashes for classes and ids.

AlasdairSwan:
@vkaracic second review complete, let me know when it is ready for another look.

vkaracic:
@AlasdairSwan I addressed all your comments in the last commit.

ecommerce/static/sass/partials/views/_coupon_admin.scss
```diff
@@ -1,12 +1,22 @@
 .coupon-create-edit-view {
-    .benefit-type label {
+    .benefit-type label,
+    .invoice-type label,
+    .invoice-discount-type label,
+    .tax-deduction label {
```
AlasdairSwan: Elements should not be used for styling. Please remove `label`, or preferably just use a more generic class that is attached to the label.

AlasdairSwan:
Once the label styling has been updated to use classes it's good to go 👍

ecommerce/static/sass/partials/views/_coupon_admin.scss
vkaracic: I've added a new class for all the labels that are listed here and used that here instead.

PR #804, opened 2016-06-16
mattdrayer/WL-530: Allow Enrollment Code purchases for enrolled users

mattdrayer:
@mjfrey @douglashall 

The enrollment check during basket additions was matching the mode of the Enrollment Code product to the mode of the Seat product for an enrolled user.  This change set cleans up the view workflow and skips over the enrollment check if the product being added is an Enrollment Code.

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -46,19 +46,45 @@ class BasketSingleItemView(View):
 
         try:
             product = StockRecord.objects.get(partner=partner, partner_sku=sku).product
+        except StockRecord.DoesNotExist:
+            return HttpResponseBadRequest(_('SKU [{sku}] does not exist.').format(sku=sku))
+
+        # If the product isn't available then there's no reason to continue with the basket addition
+        purchase_info = request.strategy.fetch_for_product(product)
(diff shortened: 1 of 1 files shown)


douglashall:
👍 I'm guessing you will need to add a test case due to the automated coverage check.

ecommerce/extensions/basket/views.py
```diff
@@ -46,19 +46,45 @@ def get(self, request):
...
         try:
             product = StockRecord.objects.get(partner=partner, partner_sku=sku).product
+        except StockRecord.DoesNotExist:
+            return HttpResponseBadRequest(_('SKU [{sku}] does not exist.').format(sku=sku))
+
+        # If the product isn't available then there's no reason to continue with the basket addition
+        purchase_info = request.strategy.fetch_for_product(product)
+        if not purchase_info.availability.is_available_to_buy:
+            msg = _('Product [{product}] not available to buy.').format(product=product.title)
+            return HttpResponseBadRequest(msg)
```
douglashall, line 56: Nice work moving this check to before we make the LMS call!

mattdrayer:
Green build!  Merging now...

PR #807, opened 2016-06-17
Added validation for verification deadline on course publication endpoint.

waheedahmed:
[ECOM-2381](https://openedx.atlassian.net/browse/ECOM-2381)

Validation of course verification deadlines before attempting publication to the LMS:
1. Only courses with a verified seat (i.e., corresponding to "verified" or "professional" (with verification!) course mode) are allowed to have a verification deadline.
2. For courses with a verified seat, the verification deadline is optional.
3. If a course has both an upgrade and verification deadline set, then the verification deadline must be after the upgrade deadline.

### DIFF
--- a/ecommerce/courses/publishers.py
+++ b/ecommerce/courses/publishers.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 import json
 import logging
 
+from dateutil.parser import parse
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 from edx_rest_api_client.client import EdxRestApiClient
@@ -94,6 +95,17 @@ class LMSPublisher(object):
         name = course.name
         verification_deadline = self.get_course_verification_deadline(course)
         modes = [self.serialize_seat_for_commerce_api(seat) for seat in course.seat_products]
+
+        if verification_deadline:
+            is_verification_deadline_invalid = any(
+                parse(mode['expires']) > parse(verification_deadline) if mode['expires'] else False for mode in modes
+            )
+            if is_verification_deadline_invalid:
+                logger.error(
+                    u'Failed to publish commerce data for [%s] to LMS. The verification deadline must occur '
+                    u'after the upgrade deadline.', course_id
+                )
+                return _(u'The verification deadline must occur after the upgrade deadline.')
 
         has_credit = 'credit' in [mode['name'] for mode in modes]
         if has_credit:
--- a/ecommerce/courses/tests/test_publishers.py
+++ b/ecommerce/courses/tests/test_publishers.py
@@ -3,6 +3,7 @@ import json
 
 import ddt
 from django.test import override_settings
+from django.utils import timezone
 import httpretty
 import mock
 from oscar.core.loading import get_model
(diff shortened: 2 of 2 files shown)


waheedahmed:
@awais786 Please review.

waheedahmed:
@rlucioni Please review.

tasawernawaz:
LGTM 👍

ecommerce/courses/publishers.py
```diff
@@ -95,7 +96,18 @@ def publish(self, course, access_token=None):
         verification_deadline = self.get_course_verification_deadline(course)
         modes = [self.serialize_seat_for_commerce_api(seat) for seat in course.seat_products]
 
-        has_credit = 'credit' in [mode['name'] for mode in modes]
+        is_deadline_invalid = any(
+            [parse(mode['expires']) > course.verification_deadline if mode['expires'] else False for mode in modes]
```
rlucioni: You don't need the outer brackets here.

ecommerce/courses/publishers.py
```diff
@@ -95,7 +96,18 @@ def publish(self, course, access_token=None):
...
+        )
+        if is_deadline_invalid:
+            logger.error(
+                u'Failed to publish commerce data for [%s] to LMS. The verification deadline must occur '
+                u'AFTER the upgrade deadline.', course_id
```
rlucioni: Nitpick: please lowercase "after," here and below.

ecommerce/courses/publishers.py
```diff
@@ -118,10 +130,12 @@ def publish(self, course, access_token=None):
         data = {
             'id': course_id,
             'name': name,
-            'verification_deadline': verification_deadline,
             'modes': modes,
         }
 
+        if 'verified' in mode_names or 'professional' in mode_names:
+            data['verification_deadline'] = verification_deadline
```
rlucioni: Why is this necessary? `verification_deadline` defaults to None if the course doesn't have one.

rlucioni: I don't think we want the key to go missing from the returned data like this.

ecommerce/courses/publishers.py
rlucioni: You have a bug here if the verification deadline is `None`. Any value is greater than `None`, but that doesn't necessarily mean that the configuration is invalid.

The goal of this story is to add validation of verification deadlines, not validation of upgrade deadlines. This check should only occur if `verification_deadline` is not None. Instead of accessing `course.verification_deadline` directly, use `verification_deadline` from above. Refer to the validation logic from edx-platform for an example: https://github.com/edx/edx-platform/blob/master/lms/djangoapps/commerce/api/v1/serializers.py#L67-L88.

Also, please name this variable more clearly: `is_verification_deadline_invalid`.

ecommerce/courses/publishers.py
waheedahmed: OK got you'r second point we don't want the key to go missing but for first point what if `verification_deadline` is not None for audit mode? Like you mentioned on the ticket in first point "Only courses with a verified seat (i.e., corresponding to "verified" or "professional" (with verification!) course mode) are allowed to have a verification deadline."

waheedahmed:
@rlucioni does it looks good now? And can you please tell why code coverage is failing while my test is covering all my changes.

rlucioni:
@waheedahmed I spoke with @clintonb about this. ECOM-2381 was created almost a year ago, before verification deadlines were validated by serializers on the LMS. A lot has changed since then. We don't need multiple points of validation. That being the case, this PR and the corresponding ticket should be closed.

@schenedx please make sure that the team re-verifies the need for tickets older than three months before starting to work on them.

PR #808, opened 2016-06-21
Fix voucher offers API endpoint

marjev:
@mjfrey Here's a PR for the voucher offers API endpoint fix. For a single course coupon, Course Catalog won't be called. Instead, course info will be fetched from LMS API.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_vouchers.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_vouchers.py
@@ -1,14 +1,17 @@
 from __future__ import unicode_literals
 
 import json
+import mock
 
 import ddt
 import httpretty
 from django.core.urlresolvers import reverse
 from opaque_keys.edx.keys import CourseKey
 from oscar.core.loading import get_model
-from oscar.test.factories import ConditionalOfferFactory, ProductFactory, RangeFactory, VoucherFactory
+from oscar.test.factories import ConditionalOfferFactory, RangeFactory, VoucherFactory
+from requests.exceptions import ConnectionError, Timeout
 from rest_framework.test import APIRequestFactory
+from slumber.exceptions import SlumberBaseException
 
 from ecommerce.core.tests.decorators import mock_course_catalog_api_client
 from ecommerce.coupons.tests.mixins import CatalogPreviewMockMixin
@@ -16,14 +19,17 @@ from ecommerce.coupons.views import get_voucher_and_products_from_code
 from ecommerce.extensions.api import serializers
 from ecommerce.extensions.api.v2.views.vouchers import VoucherViewSet
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.extensions.partner.models import StockRecord
 from ecommerce.extensions.test.factories import prepare_voucher
 from ecommerce.coupons.tests.mixins import CouponMixin
-from ecommerce.tests.mixins import Catalog
+from ecommerce.courses.models import Course
+from ecommerce.tests.mixins import Catalog, LmsApiMockMixin
 from ecommerce.tests.testcases import TestCase
 
 COUPON_CODE = 'COUPONCODE'
 
 
+Range = get_model('offer', 'Range')
 StockRecord = get_model('partner', 'StockRecord')
 
 
@@ -60,7 +66,13 @@ class VoucherViewSetTests(TestCase):
 
 @ddt.ddt
 @httpretty.activate
-class VoucherViewOffersEndpointTests(CatalogPreviewMockMixin, CouponMixin, CourseCatalogTestMixin, TestCase):
+class VoucherViewOffersEndpointTests(
+        CatalogPreviewMockMixin,
+        CouponMixin,
+        CourseCatalogTestMixin,
+        LmsApiMockMixin,
+        TestCase
+):
     """ Tests for the VoucherViewSet offers endpoint. """
 
     def setUp(self):
@@ -85,31 +97,47 @@ class VoucherViewOffersEndpointTests(CatalogPreviewMockMixin, CouponMixin, Cours
 
         self.assertEqual(response.status_code, 400)
 
-    @mock_course_catalog_api_client
-    def test_voucher_offers_listing_with_range_catalog(self):
-        """ Verify the endpoint returns offers data when range has a catalog. """
-        self.mock_dynamic_catalog_course_runs_api()
-        product = ProductFactory(stockrecords__price_excl_tax=100)
-        new_range = RangeFactory(products=[product, ])
+    def test_voucher_offers_listing_for_a_single_course_voucher(self):
+        """ Verify the endpoint returns offers data when a single product is in voucher range. """
+        course, seat = self.create_course_and_seat()
+        self.mock_course_api_response(course=course)
+        new_range = RangeFactory(products=[seat, ])
         new_range.catalog = Catalog.objects.create(partner=self.partner)
-        new_range.catalog.stock_records.add(StockRecord.objects.get(product=product))
+        new_range.catalog.stock_records.add(StockRecord.objects.get(product=seat))
         voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
         request = self.prepare_offers_listing_request(voucher.code)
         response = self.endpointView(request)
 
         self.assertEqual(response.status_code, 200)
 
-        # If no product is associated with the voucher range, Bad Request status should be returned
-        new_range.remove_product(product)
+        # If no seat is associated with the voucher range, Bad Request status should be returned
+        new_range.remove_product(seat)
         response = self.endpointView(request)
         self.assertEqual(response.status_code, 400)
+
+    @ddt.data((ConnectionError,), (Timeout,), (SlumberBaseException,))
+    @ddt.unpack
+    def test_voucher_offers_listing_api_exception_caught(self, exception):
+        """ Verify the endpoint returns status 400 Bad Request when ConnectionError occurs """
+        with mock.patch(
+            'ecommerce.extensions.api.v2.views.vouchers.VoucherViewSet.get_offers',
+            mock.Mock(side_effect=exception)
+        ):
+            __, seat = self.create_course_and_seat()
+            new_range = RangeFactory(products=[seat, ])
+            voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
+            request = self.prepare_offers_listing_request(voucher.code)
+            response = self.endpointView(request)
+
+            self.assertEqual(response.status_code, 400)
 
     @mock_course_catalog_api_client
     def test_voucher_offers_listing_product_found(self):
         """ Verify the endpoint returns offers data for single product range. """
-        self.mock_dynamic_catalog_course_runs_api()
-        product = ProductFactory(stockrecords__price_excl_tax=100)
-        new_range = RangeFactory(products=[product, ])
+        course, seat = self.create_course_and_seat()
+        self.mock_course_api_response(course=course)
+
+        new_range = RangeFactory(products=[seat, ])
         voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
         request = self.prepare_offers_listing_request(voucher.code)
         response = self.endpointView(request)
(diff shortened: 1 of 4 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=808&repo=edx%2Fecommerce

mjfrey:
@mattdrayer

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -80,36 +88,79 @@ def get_offers(self, products, request, voucher):
...
-            q=benefit.range.catalog_query
-        )['results']
 
-        course_ids = [product.course_id for product in products]
-        courses = Course.objects.filter(id__in=course_ids)
-        contains_verified_course = next((False for course in courses if course.type != 'verified'), True)
+        if catalog_query:
+            query_results = request.site.siteconfiguration.course_catalog_api_client.course_runs.get(
+                q=catalog_query
+            )['results']
```
mattdrayer: What if an error is returned by the server during the GET operation?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -80,36 +88,79 @@ def get_offers(self, products, request, voucher):
...
+                        benefit=benefit,
+                        course=course,
+                        course_info=course_catalog_data,
+                        is_verified=contains_verified_course,
+                        offers=offers,
+                        stock_record=stock_record,
+                        voucher=voucher
+                    )
+        else:
+            product = products[0]
```
mattdrayer: If there are no products this will raise an IndexError

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -80,36 +88,79 @@ def get_offers(self, products, request, voucher):
...
+            course_id = product.course_id
+            stock_record = StockRecord.objects.get(product__id=product.id)
```
mattdrayer, line 100: Do we need to defend against StockRecord.DoesNotExist here?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -80,36 +88,79 @@ def get_offers(self, products, request, voucher):
...
+
+            course_info = get_course_info_from_lms(course_id)
+            course = Course.objects.get(id=course_id)
```
mattdrayer: Do we need to defend against Course.DoesNotExist here?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -80,36 +88,79 @@ def get_offers(self, products, request, voucher):
...
+                    course=course,
+                    course_info=course_info,
+                    is_verified=(course.type == 'verified'),
+                    offers=offers,
+                    stock_record=stock_record,
+                    voucher=voucher
+                )
         return offers
+
+    def get_course_offer_data(self, benefit, course, course_info, is_verified, offers, stock_record, voucher):
```
mattdrayer: This method is named 'get' but it's actually modifying the `offers` container -- imho the method should have a different name.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -80,36 +88,79 @@ def get_offers(self, products, request, voucher):
...
-                    'course_start_date': course_catalog_data['start'],
-                    'id': course.id,
-                    'image_url': course_catalog_data['image']['src'],
-                    'organization': CourseKey.from_string(course.id).org,
-                    'seat_type': course.type,
-                    'stockrecords': serializers.StockRecordSerializer(stock_record).data,
-                    'title': course.name,
-                    'voucher_end_date': voucher.end_datetime,
-                })
+                course = courses.get(id=course_id)
```
mattdrayer: Why are we doing courses.get() here, but Courses.objects.get() on line 125?  Either way we are getting a course object -- seems like we could simplify this by consolidating the work up to the top of the products loop?

mattdrayer:
@marjev @mjfrey -- made a pass through -- let me know when you're ready for another look.

mjfrey:
👍

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -145,3 +204,28 @@ def test_get_offers(self):
...
+        voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
+        voucher, products = get_voucher_and_products_from_code(voucher.code)
+        request = self.prepare_offers_listing_request(voucher.code)
+        self.mock_course_api_response(course=course)
+        offers = VoucherViewSet().get_offers(products=products, request=request, voucher=voucher)
+
+        with mock.patch('ecommerce.extensions.api.v2.views.vouchers.Course.objects.get') as mock_get:
+            mock_get.side_effect = Course.DoesNotExist
+            offers = VoucherViewSet()._get_offers_for_product(products[0], voucher)  # pylint:disable=protected-access
+            self.assertEqual(len(offers), 0)
```
marjev, line 220: Don't mock this, delete the course. This is the edge case, which would indicate a deeper problem. Do we want to present the end user with empty page when no Course is found in ecommerce?

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -145,3 +204,28 @@ def test_get_offers(self):
...
+
+        with mock.patch('ecommerce.extensions.api.v2.views.vouchers.StockRecord.objects.get') as mock_get:
+            mock_get.side_effect = StockRecord.DoesNotExist
+            offers = VoucherViewSet()._get_offers_for_product(products[0], voucher)  # pylint:disable=protected-access
+            self.assertEqual(len(offers), 0)
```
marjev, line 225: Don't mock this, delete the stock record. This is the edge case, which would indicate a deeper problem. Do we want to present the end user with empty page when no Stock Record is found in ecommerce?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+                offers = self.get_offers(products, request, voucher)
+            except (ConnectionError, SlumberBaseException, Timeout):
+                logger.error('Could not get course information.')
+                return Response(status=status.HTTP_400_BAD_REQUEST)
             cache.set(cache_hash, offers, settings.COURSES_API_CACHE_TIMEOUT)
 
         page = self.paginate_queryset(offers)
         return self.get_paginated_response(page)
 
+    def _get_offers_for_product(self, product, voucher, course_info=None, offers=None):
```
marjev, line 89: Why is `offers` an optional parameter when the whole point of this method is appending a dictionary value to it?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+        """
+        Retrieves product-specific offers based on the course and benefit information
+        """
+        course_id = product.course_id
+        try:
+            course = Course.objects.get(id=course_id)
+        except Course.DoesNotExist:
+            return offers if offers else []
```
marjev, line 97: In this case, we have a deeper problem in the database. If Course is not found, we should let the user know what type of error occurred. This way, an empty page would be a result and I'm not sure we want that.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+
+        try:
+            stock_record = StockRecord.objects.get(product__id=product.id)
+        except StockRecord.DoesNotExist:
+            return offers if offers else []
```
marjev, line 102: Same comment as above.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
         page = self.paginate_queryset(offers)
         return self.get_paginated_response(page)
 
+    def _get_offers_for_product(self, product, voucher, course_info=None, offers=None):
+        """
+        Retrieves product-specific offers based on the course and benefit information
+        """
+        course_id = product.course_id
+        try:
+            course = Course.objects.get(id=course_id)
```
marjev, line 95: Why are we making separate queries to the DB for each course (in case multiple courses are associated with the voucher)? This will definitely slow things down.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+        """
+        catalog_query = self._get_voucher_catalog_query(voucher)
+        client = request.site.siteconfiguration.course_catalog_api_client
+        try:
+            response = client.course_runs.get(
+                q=catalog_query,
+                page_size=DEFAULT_CATALOG_PAGE_SIZE,
+                limit=DEFAULT_CATALOG_PAGE_SIZE
+            )
+        except (ConnectionError, SlumberBaseException, Timeout):  # pragma: no cover
```
marjev, line 141: This is redundant. Line 81 catches these exceptions. Same exceptions can happen in line 105.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+            logger.error('Could not get course run information.')
+            return Response(status=status.HTTP_400_BAD_REQUEST)
```
marjev, line 143: This Response object will be returned to the `offers` method, but since the Response object can't be paginated, error will occur.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+                return Response(status=status.HTTP_400_BAD_REQUEST)
             cache.set(cache_hash, offers, settings.COURSES_API_CACHE_TIMEOUT)
 
         page = self.paginate_queryset(offers)
         return self.get_paginated_response(page)
 
+    def _get_offers_for_product(self, product, voucher, course_info=None, offers=None):
+        """
+        Retrieves product-specific offers based on the course and benefit information
+        """
```
marjev, line 92: Nitpick: Docstring is missing parameters and return value description.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+                'title': course.name,
+                'voucher_end_date': voucher.end_datetime,
+            })
+
+        return offers
+
+    def _get_offers_for_catalog_query(self, request, products, voucher):
+        """
+        Batch workflow to retrieve product offers for a given catalog query
+        """
```
marjev, line 132: Nitpick: Docstring is missing parameters and return value description.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+        for product in products:
+            course_id = product.course_id
+            course_info = next((result for result in query_results if result['key'] == course_id), None)
+            offers = self._get_offers_for_product(product, voucher, course_info, offers)
+        return offers
+
+    def _get_voucher_benefit(self, voucher):
+        """
+        Retrieves a benefit for the first offer from the specified voucher
+        """
```
marjev, line 159: Nitpick: Docstring is missing parameters and return value description.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+        return voucher.offers.first().benefit
```
marjev, line 160: Since we don't have a voucher model in our project to make this a property in, we should place this somewhere in the Voucher application (probably utils).

If we want to DRY this, it should be available everywhere in the project. This has nothing to do with the Voucher API.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+
+    def _get_voucher_catalog_query(self, voucher):
+        """
+        Retrieves the catalog query, from the range, from the benefit, from the offer, from the specified voucher...
+        """
```
marjev, line 165: Nitpick: Docstring missing parameters and return value description.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+        benefit = self._get_voucher_benefit(voucher)
+        return benefit.range.catalog_query
```
marjev, line 167: Consider placing this somewhere in the Offers application or wherever the Benefit model lives in the Oscar. This has nothing to do with the API endpoint, so it shouldn't be a private method encapsulated here.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -90,37 +176,14 @@ def get_offers(self, products, request, voucher):
         Returns:
             List: List of course offers where each offer is represented by a dictionary
         """
-        benefit = voucher.offers.first().benefit
-        offers = []
-        query_results = request.site.siteconfiguration.course_catalog_api_client.course_runs.get(
-            q=benefit.range.catalog_query, page_size=DEFAULT_CATALOG_PAGE_SIZE, limit=DEFAULT_CATALOG_PAGE_SIZE
-        )['results']
+        # If a voucher was not provided, return an empty offer set
+        if not voucher or not products:
```
marjev, line 180: This is handled by the `offers` method and seems redundant.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,12 +76,96 @@ def offers(self, request):
...
+
+        benefit = self._get_voucher_benefit(voucher)
+
+        if benefit and course_info:
+
+            if offers is None:
+                offers = []
+            offers.append({
+                'benefit': serializers.BenefitSerializer(benefit).data,
+                'contains_verified': (course.type == 'verified'),
```
marjev, line 116: Important: What is the use case for this?

Currently, this causes a bug on the offers page. This definitely doesn't describe the entire offers list but instead sets this flag for one offer only (and that's the only case this will work for, as well as the case when one of the pages consists of the verified seats).

PR #809, opened 2016-06-21
[WIP] [SOL-1814] Deprecate the old catalog from coupons

vkaracic:
Some of the changes are adopted from https://github.com/edx/ecommerce/pull/765 FYI @clintonb 

Coupon report tests are skipped because the report needs to be revamped for multi-course coupons, which I guess blocks this PR, @mjfrey ?

https://openedx.atlassian.net/browse/SOL-1814

### DIFF
--- a/ecommerce/coupons/tests/mixins.py
+++ b/ecommerce/coupons/tests/mixins.py
@@ -100,7 +100,7 @@ class CouponMixin(object):
 
         return pc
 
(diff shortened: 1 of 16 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=809&repo=edx%2Fecommerce

vkaracic:
Deprecation discontinued.

PR #811, opened 2016-06-21
[SOL-1819] Multi offer coupon.

vkaracic:
Enable ONCE_PER_CUSTOMER and MULTI_USE coupon types to have max_uses and quantity > 1 at the same time. This requires each of the vouchers to have individual offers associated with them.

https://openedx.atlassian.net/browse/SOL-1819

### DIFF
--- a/ecommerce/extensions/voucher/utils.py
+++ b/ecommerce/extensions/voucher/utils.py
@@ -220,7 +220,9 @@ def generate_coupon_report(coupon_vouchers):
     return field_names, rows
 
 
-def _get_or_create_offer(product_range, benefit_type, benefit_value, coupon_id=None, max_uses=None):
+def _get_or_create_offer(
+        product_range, benefit_type, benefit_value, coupon_id=None, max_uses=None, offer_number=None
+):
     """
     Return an offer for a catalog with condition and benefit.
 
@@ -248,9 +250,8 @@ def _get_or_create_offer(product_range, benefit_type, benefit_value, coupon_id=N
     )
 
     offer_name = "Coupon [{}]-{}-{}".format(coupon_id, offer_benefit.type, offer_benefit.value)
-    if max_uses:
-        # Offer needs to be unique for each multi-use coupon.
-        offer_name = "{} (Coupon [{}] - max_uses:{})".format(offer_name, coupon_id, max_uses)
+    if offer_number:
+        offer_name = "{} [{}]".format(offer_name, offer_number)
 
     offer, __ = ConditionalOffer.objects.get_or_create(
         name=offer_name,
@@ -359,6 +360,7 @@ def create_vouchers(
     """
     logger.info("Creating [%d] vouchers product [%s]", quantity, coupon.id)
     vouchers = []
+    offers = []
 
     if _range:
         # Enrollment codes use a custom range.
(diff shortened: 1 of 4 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=811&repo=edx%2Fecommerce

ecommerce/extensions/voucher/utils.py
```diff
@@ -374,18 +375,24 @@ def create_vouchers(
...
-        benefit_type=benefit_type,
-        benefit_value=benefit_value,
-        max_uses=max_uses,
-        coupon_id=coupon.id
-    )
-    for __ in range(quantity):
+    num_of_offers = quantity if multi_offer else 1
+
+    for num in range(num_of_offers):
+        offer = _get_or_create_offer(
```
mjfrey, line 386: Why are you creating more than one offer?  For this case it is multiple vouchers for the same offer of just dynamic type.

vkaracic, line 386: A user should be able to redeem the coupon the number of `max_uses` times. If vouchers share the same offer they share the same `max_global_applications` value. In other terms if you create 2 coupons with 2 `max_uses`, if the same offer is shared, the first user can redeem all the usages which renders the second one unable to redeem it's coupon at all.

ecommerce/extensions/voucher/utils.py
```diff
@@ -220,7 +220,7 @@ def generate_coupon_report(coupon_vouchers):
     return field_names, rows
 
 
-def _get_or_create_offer(product_range, benefit_type, benefit_value, coupon_id=None, max_uses=None):
+def _get_or_create_offer(product_range, benefit_type, benefit_value, coupon_id=None, max_uses=None, number=None):
```
mjfrey: change number to quantity

ecommerce/extensions/voucher/utils.py
```diff
@@ -248,9 +248,8 @@ def _get_or_create_offer(product_range, benefit_type, benefit_value, coupon_id=N
     )
 
     offer_name = "Coupon [{}]-{}-{}".format(coupon_id, offer_benefit.type, offer_benefit.value)
-    if max_uses:
-        # Offer needs to be unique for each multi-use coupon.
-        offer_name = "{} (Coupon [{}] - max_uses:{})".format(offer_name, coupon_id, max_uses)
+    if number:
```
mjfrey: quantity

ecommerce/extensions/voucher/utils.py
vkaracic: Why? This not quantity, but the number of the offer.

vkaracic:
@iivic @marjev please review this one as well.

ecommerce/extensions/voucher/utils.py
marjev: Perhaps naming it `offer_number` will help to avoid any confusion, I guess.

marjev:
LGTM 👍

mjfrey:
👍

iivic:
👍

vkaracic:
@mattdrayer please review this PR.

ecommerce/extensions/voucher/utils.py
mattdrayer, line 386: I agree that it is a bit confusing -- it would be great if we could explain what is going on here with some detailed code comments.

mattdrayer:
@vkaracic my only request is to add some more comments to the updated implementation clarifying the logic for future developers -- otherwise LGTM -- 👍

PR #812, opened 2016-06-21
[SOL-1831] Show maximum uses in coupon detail page

iivic:
@marjev @mjfrey @vkaracic  Someone please review.

### DIFF
--- a/ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
+++ b/ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
@@ -112,6 +112,7 @@ define([
                 expect(view.$el.find('.invoiced-amount > .value').text()).toEqual(
                     _s.sprintf('$%s', model.get('price'))
                 );
+                expect(view.$el.find('.max-uses > .value').text()).toEqual(model.get('max_uses'));
                 expect(view.renderVoucherTable).toHaveBeenCalled();
                 expect(view.$el.find('.invoice-type > .value').text()).toEqual(model.get('invoice_type'));
                 expect(view.$el.find('.invoice-number > .value').text()).toEqual(model.get('invoice_number'));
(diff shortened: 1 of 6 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=812&repo=edx%2Fecommerce

ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
```diff
@@ -152,6 +152,14 @@ define([
                 expect(e.preventDefault).toHaveBeenCalled();
                 expect(window.open).toHaveBeenCalledWith(url, '_blank');
             });
+
```
vkaracic, line 187: Try including `max_uses` in the model used in test on line 81 in this spec, and a line something like:

```
expect(view.$el.find('.max-uses > .value').text()).toEqual(model.get('max_uses'));
```

In your test below you are testing your function, but nowhere are you testing if that function is called on render.

vkaracic:
After the one comment is addressed :+1:

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -66,6 +66,14 @@ define([
                 return '';
             },
 
+            setMaxUses: function(voucher) {
+               if (voucher.usage === 'Single use') {
+                   return 1;
```
marjev: Change this in the model. So, if voucher.usage is 'Single use', set model max_uses to 1. This way, we let model handle the logic and we display model's data on the template.

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -84,6 +92,7 @@ define([
                     usage: this.usageLimitation(voucher),
                     category: category,
                     note: note,
+                    maxUses: this.setMaxUses(voucher)
```
marjev: Once you address this comment https://github.com/edx/ecommerce/pull/812/files#r67881116, you won't need this line.

marjev:
Please provide a screenshot of the details page so that we can see the SASS/template changes.

iivic:
Comments addressed. Here is the ss screen shot 2016-06-23 at 13 24 42

marjev:
👍

vkaracic:
:+1:

PR #813, opened 2016-06-21
Add support for course catalog pagenated results

mjfrey:
Get all paginated results from course catalog.

### DIFF
--- a/ecommerce/coupons/tests/mixins.py
+++ b/ecommerce/coupons/tests/mixins.py
@@ -23,6 +23,7 @@ class CatalogPreviewMockMixin(object):
         """ Helper function to register a dynamic course catalog API endpoint for the course run information. """
         course_run_info = {
             'count': 1,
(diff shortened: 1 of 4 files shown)


mjfrey:
@marjev @vkaracic

mjfrey:
@bderusha

mjfrey:
@mattdrayer

mattdrayer:
These `while` loops seem unnecessary to me -- in a server-to-server context, I see no problem with simply bumping up the size of the response payload to include all records, using a value of say, 10000.  We can get what we need in a single network request, and doing so eliminates the need to introduce the new logic, which seems like it would be better to locate in the EdxRestApiClient, anyway.

mjfrey:
ok, I will close this PR out in favor of just bumping up the limit in the API call to the Catalog Service.

bderusha:
Sounds good

PR #815, opened 2016-06-22
Bump up page_size when calling course catalog

mjfrey:
Get all results from course catalog.

### DIFF
--- a/ecommerce/coupons/utils.py
+++ b/ecommerce/coupons/utils.py
@@ -1,6 +1,8 @@
 """ Coupon related utility functions. """
 from oscar.core.loading import get_model
 
+from ecommerce.core.constants import DEFAULT_CATALOG_PAGE_SIZE
+
 Product = get_model('catalogue', 'Product')
 
 
@@ -16,7 +18,9 @@ def get_seats_from_query(site, query, seat_types):
     Returns:
         List of seat products retrieved from the course catalog query.
     """
-    response = site.siteconfiguration.course_catalog_api_client.course_runs.get(q=query)
+    response = site.siteconfiguration.course_catalog_api_client.course_runs.get(q=query,
+                                                                                page_size=DEFAULT_CATALOG_PAGE_SIZE,
+                                                                                limit=DEFAULT_CATALOG_PAGE_SIZE)
     query_products = []
     for result in response['results']:
         try:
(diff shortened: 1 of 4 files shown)


mjfrey:
@mattdrayer @bderusha

ecommerce/coupons/utils.py
```diff
@@ -16,7 +16,7 @@ def get_seats_from_query(site, query, seat_types):
     Returns:
         List of seat products retrieved from the course catalog query.
     """
-    response = site.siteconfiguration.course_catalog_api_client.course_runs.get(q=query)
+    response = site.siteconfiguration.course_catalog_api_client.course_runs.get(q=query, page_size=10000, limit=10000)
```
bderusha: Can we put this value into a shared configuration?

At the very least it would be good to have this be a shared constant

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -42,7 +42,8 @@ def preview(self, request):
         if query and seat_types:
             seat_types = seat_types.split(',')
             try:
-                results = request.site.siteconfiguration.course_catalog_api_client.course_runs.get(q=query)['results']
+                results = request.site.siteconfiguration.course_catalog_api_client.\
```
bderusha: NIT: if we are going to break this into two lines anyway, would prefer 

```
client = request.site.siteconfiguration.course_catalog_api_client
results = client.course_runs.get(q=query, page_size=10000, limit=10000)['results']
```

ecommerce/coupons/utils.py
```diff
@@ -3,6 +3,8 @@
 
 Product = get_model('catalogue', 'Product')
 
+CATALOG_PAGE_SIZE = 10000
```
mattdrayer: I hate to nitpick on this, but (imho) a better location for this constant is either in `ecommerce.core.constants.py` or 'ecommerce.extensions.catalogue.utils.py'

bderusha:
👍

ecommerce/coupons/utils.py
mattdrayer: I'd also consider renaming it to 'DEFAULT_CATALOG_PAGE_SIZE' or something along the lines of "EXPANDED_", "MAX_", etc.

mjfrey:
@mattdrayer comments addressed.

mattdrayer:
👍 time to 🚢

PR #817, opened 2016-06-22
Only cache query results not single courses

mjfrey:
Only cache offers for queries not single courses.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_vouchers.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_vouchers.py
@@ -145,6 +145,22 @@ class VoucherViewOffersEndpointTests(
         self.assertEqual(response.status_code, 200)
 
     @mock_course_catalog_api_client
+    def test_voucher_offers_listing_catalog_query(self):
+        """ Verify the endpoint returns offers data for single product range. """
+        course, seat = self.create_course_and_seat()
+        self.mock_dynamic_catalog_course_runs_api(query='*:*', course_run=course)
+        new_range, __ = Range.objects.get_or_create(catalog_query='*:*')
+        new_range.add_product(seat)
+        voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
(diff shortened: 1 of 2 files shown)


mjfrey:
@mattdrayer

mattdrayer:
why do we want to only cache the offers for the catalog query scenario?  Seems like we would still want to cache and avoid making a backend call in either case?  Maybe the hash key could just be different depending on which mode we're operating in?

mattdrayer:
something like:

```
cache_key = 'voucher_offers_{}'.format(voucher.id)

query = voucher.offers.first().benefit.range.catalog_query
if query:
    cache_key = 'voucher_offers_{}'.format(query)

cache_hash = hashlib.md5(cache_key).hexdigest()
offers = cache.get(cache_hash)

if not offers:  # pragma: no cover
    try:
        offers = self.get_offers(products, request, voucher)

cache.set(cache_hash, offers, settings.COURSES_API_CACHE_TIMEOUT)
```

mjfrey:
@mattdrayer good idea.

mattdrayer:
Green build -- clear to merge -- 👍

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -145,6 +145,22 @@ def test_voucher_offers_listing_product_found(self):
...
+        new_range, __ = Range.objects.get_or_create(catalog_query='*:*')
+        new_range.add_product(seat)
+        voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
+        voucher, __ = get_voucher_and_products_from_code(voucher.code)
+        request = self.prepare_offers_listing_request(voucher.code)
+        response = self.endpointView(request)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertGreater(len(response.data), 0)
+        # More assertions here
```
marjev, line 161: Do we need this comment?

marjev:
This can be closed due to the existence of the #819. I included these changes there to avoid having numerous commits in history regarding this API endpoint.

PR #819, opened 2016-06-23
Fix voucher offers API endpoint

marjev:


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=819

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -72,118 +76,118 @@ def offers(self, request):
...
-            query_results = response['results']
-
-        offers = []
-        for product in products:
+            )['results']
+
+            course_ids = [product.course_id for product in products]
+            courses = Course.objects.filter(id__in=course_ids)
+            stock_records = StockRecord.objects.filter(product__in=products)
+            contains_verified_course = 'verified' == benefit.range.course_seat_types
```
mjfrey: this is a list.  Will this logic work?

marjev: This is always comma-separated string. That's how it's saved in the database.

mjfrey: ok so == will work if the string is "verified, professional"?

marjev: it should be False, in that case. True is returned when offers contains only verified courses.

mjfrey: 👍 Great

mjfrey:
@mattdrayer please take another look at this.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -6,23 +6,24 @@
...
+from ecommerce.core.url_utils import get_lms_url
 from ecommerce.coupons.tests.mixins import CatalogPreviewMockMixin
 from ecommerce.coupons.views import get_voucher_and_products_from_code
+from ecommerce.courses.models import Course
 from ecommerce.extensions.api import serializers
 from ecommerce.extensions.api.v2.views.vouchers import VoucherViewSet
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
-from ecommerce.extensions.partner.models import StockRecord
 from ecommerce.extensions.test.factories import prepare_voucher
 from ecommerce.coupons.tests.mixins import CouponMixin
```
vkaracic: Can you move this one too, please?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -72,118 +76,120 @@ def offers(self, request):
             return Response(status=status.HTTP_400_BAD_REQUEST)
 
         query = voucher.offers.first().benefit.range.catalog_query
-        cache_key = 'catalog_query_{}'.format(query)
+        if query:
+            cache_key = 'catalog_query_{}'.format(query)
```
mattdrayer: I think it's important that the cache key prefix reflect the information actually being stored in the cache, which in this case is voucher offers.  If we prefix the cache key with "catalog_query", really there could be anything stored in the cache and another workflow could run into problems if it attempts to use the same key prefix in a different context.  So let's go with 'voucher_offers_' for both conditions here.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -72,118 +76,120 @@ def offers(self, request):
...
         offers = cache.get(cache_hash)
-        if not offers:  # pragma: no cover
+        if not offers:
             try:
                 offers = self.get_offers(products, request, voucher)
             except (ConnectionError, SlumberBaseException, Timeout):
                 logger.error('Could not get course information.')
                 return Response(status=status.HTTP_400_BAD_REQUEST)
+            except Http404:
+                logger.error('Could not get product information.')
```
mattdrayer: Can we add more context to the log message here?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -72,118 +76,120 @@ def offers(self, request):
...
-            logger.error('Could not get course run information.')
-            return Response(status=status.HTTP_400_BAD_REQUEST)
-
-        query_results = []
-        if 'results' in response:
-            query_results = response['results']
-
-        offers = []
-        for product in products:
+            )['results']
```
mattdrayer, line 119: Is it possible that no results key will be available in the response?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -72,118 +76,120 @@ def offers(self, request):
...
+
+            for product in products:
+                course_id = product.course_id
+                course_catalog_data = next((result for result in query_results if result['key'] == course_id), None)
+
+                try:
+                    stock_record = stock_records.get(product__id=product.id)
+                except StockRecord.DoesNotExist:
+                    stock_record = None
+                    logger.error('Stock Record for a product %s not found.', product.id)
```
mattdrayer: "...for product %s..."

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -72,118 +76,120 @@ def offers(self, request):
...
+                        voucher=voucher
+                    ))
+        else:
+            product = products[0]
             course_id = product.course_id
-            course_info = next((result for result in query_results if result['key'] == course_id), None)
-            offers = self._get_offers_for_product(product, voucher, course_info, offers)
+            course_info = get_course_info_from_lms(course_id)
+            stock_record = get_object_or_404(StockRecord, product__id=product.id)
+            course = get_object_or_404(Course, id=course_id)
```
mattdrayer, line 154: I think it would be better to perform this database lookup before making the network call to the LMS for course_info above

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -72,118 +76,120 @@ def offers(self, request):
...
+
+            if course_info and course:
```
mattdrayer: if there is no course, will the value be None or will the 404 condition be triggered?  ie, does it even make sense to check the status of course here?

mattdrayer:
@marjev looking good!  I left some comments and also it looks like the branch needs a rebase -- let me know when you're ready for a final pass.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
marjev: Why? It's used in `VoucherViewOffersEndpointTests`.

ecommerce/extensions/api/v2/views/vouchers.py
marjev, line 119: It's a paginated API endpoint, so results will be empty or contains course run info in a form of a list.

ecommerce/extensions/api/v2/views/vouchers.py
marjev: Done.

ecommerce/extensions/api/v2/views/vouchers.py
marjev: Done. I added product title to provide more context, but since this is an edge case that should never happen and in case it happens, this logger should help us indicate that data in DB got corrupted.

ecommerce/extensions/api/v2/views/vouchers.py
marjev: Done.

ecommerce/extensions/api/v2/views/vouchers.py
marjev, line 154: Done.

ecommerce/extensions/api/v2/views/vouchers.py
marjev: Done.

marjev:
@mattdrayer I addressed the comments you left in the previous review. Can you please take another look? Thanks!

Let me know if you have any suggestions or questions.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
vkaracic: Move up, as in C comes before E :)

mattdrayer:
changes LGTM -- nice!  👍

PR #820, opened 2016-06-23
Added ellipsis to long offer page course title

iivic:
@marjev  @vkaracic  @mjfrey  A short PR to review. 
screen shot 2016-06-23 at 14 30 38

### DIFF
--- a/ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
+++ b/ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
@@ -112,7 +112,7 @@ define([
                 expect(view.$el.find('.invoiced-amount > .value').text()).toEqual(
                     _s.sprintf('$%s', model.get('price'))
                 );
-                expect(view.$el.find('.max-uses > .value').text()).toEqual(model.get('max_uses'));
+                expect(view.$('.max-uses > .value').text()).toEqual(model.get('max_uses'));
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=820

vkaracic:
Course names in the cards should be links to the course in case a customer wants a more detailed description of the course before enrolling/redeeming the coupon.

Since this is not part of this PR, :+1:

ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
```diff
@@ -112,7 +112,7 @@ define([
                 expect(view.$el.find('.invoiced-amount > .value').text()).toEqual(
                     _s.sprintf('$%s', model.get('price'))
                 );
-                expect(view.$el.find('.max-uses > .value').text()).toEqual(model.get('max_uses'));
+                expect(parseInt(view.$('.max-uses > .value').text())).toEqual(parseInt(model.get('max_uses')));
```
marjev: HTML element value should be an integer value and so should the model attribute `max_uses` (since we use Backbone StickIt).

Make sure that this passes without parseInt.

marjev:
LGTM 👍 once my comment is addressed

PR #821, opened 2016-06-23
[WIP] [SOL-1847] Change not-selected values to null on toggling catalog type.

vkaracic:
When toggling between single and multi course catalog types, the input fields are hidden but the values in them remain, hence it can happen that a single and multi course coupon is created.
This PR contains code that resets the values of non-selected input fields on catalog type toggle.

https://openedx.atlassian.net/browse/SOL-1847

### DIFF
--- a/ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
+++ b/ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
@@ -219,6 +219,23 @@ define([
                     view.updateCourseSeatTypes();
                     expect(view.dynamic_catalog_view.seat_types).toEqual(model.get('course_seat_types'));
                 });
+
+                it('should remove dynamic catalog values from fields when toggled to single course', function() {
+                    var catalog_query = '*:*',
+                        course_seat_types = ['verified'];
+
+                    model.set('catalog_query', catalog_query);
+                    model.set('course_seat_types', course_seat_types);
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=821

mjfrey:
👍

marjev:
👍 once you try using model.set() instead of changing HTML value.

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
```diff
@@ -219,6 +219,23 @@ define([
...
+                        course_seat_types = ['verified'];
+
+                    model.set('catalog_query', catalog_query);
+                    model.set('course_seat_types', course_seat_types);
+                    view.updateCatalogQuery();
+                    view.updateCourseSeatTypes();
+                    expect(view.dynamic_catalog_view.query).toEqual(catalog_query);
+                    expect(view.dynamic_catalog_view.seat_types).toEqual(course_seat_types);
+
+                    view.$el.find('#single-course').prop('checked', true);
```
iivic: Short notation for $el.find() is just $().

iivic:
👍

PR #823, opened 2016-06-24
Fix for new offer landing page.

mjfrey:
New offer landing page is a tad slower so we need to wait for the elements on the page.


mjfrey:
@mattdrayer

mattdrayer:
👍, assuming the IDs included in this PR correctly match those in the template!

mjfrey:
No ID's changed.  Just the page take a bit longer to load.

mattdrayer:
While this is definitely an acceptable approach, I think there's a better way to wait for an element to appear using Promises, EmptyPromises, and BrokenPromises (sounds like a bad country song)

PR #824, opened 2016-06-27
Enable patch editing of the coupons

marjev:


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=824

mattdrayer:
Overall the code LGTM -- we'll need to add both python tests and javascript tests to ensure all of the new logical branches are properly covered -- also looks like a rebase is required.

marjev:
@vkaracic @mattdrayer I finished writing additional tests for this. Could you please take another look at the code?

This should:
1) Save a coupon attributes before editing form renders
2) Check the difference between server state and current model state and create a patch dictionary containing only those values that were edited
3) Make a patch update if possible

ecommerce/static/js/models/coupon_model.js
```diff
@@ -158,6 +159,12 @@ define([
                 return course_id ? course_id.value : '';
             },
 
+            updateCategory: function() {
+                var categoryID = this.get('categories')[0].id;
```
vkaracic, line 183: Why the zero index category?

marjev, line 183: Because it's a list of categories ... that's how it's serialized in the API, so I have to get it that way

vkaracic, line 183: List of one category? Because I don't understand why should the category always be updated with the first in the list.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -80,6 +80,27 @@ def setup_site_configuration(self):
...
+            'invoice_number': 'INV-00055',
+            'invoice_payment_date': datetime.datetime(2016, 1, 1, tzinfo=pytz.UTC).isoformat(),
+            'invoice_type': Invoice.PREPAID,
+            'tax_deducted_source': None
+        }
+
+        invoice_data = CouponViewSet().retrieve_invoice_data(request_data)
+
+        self.assertDictEqual(invoice_data, {
+            'discount_type': Invoice.PERCENTAGE,
```
vkaracic: Nit:
`'discount_type': request_data['invoice_discount_type']`

ecommerce/static/js/models/coupon_model.js
marjev, line 183: Yeah, a list including a category is exposed via coupon API.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
marjev: Done.

marjev:
@mjfrey can you please review this?

ecommerce/static/js/test/specs/views/coupon_edit_view_spec.js
```diff
@@ -96,6 +96,17 @@ define([
...
+                it('should patch save the model when form is in editing mode and has editable attributes', function () {
+                    var formView = view.formView;
+                    spyOn(formView.model, 'save');
+                    spyOn(formView.model, 'isValid').and.returnValue(true);
+
+                    expect(formView.modelServerState).toEqual(model.pick(formView.editableAttributes));
+                    formView.model.set('title', 'Test Title');
+                    formView.submit($.Event('click'));
+                    expect(model.save).toHaveBeenCalled();
+                });
```
vkaracic, line 109: How does patch save differ from normal/put save?

marjev, line 109: The main difference that I'm trying to get advantage of here is that only some attributes will be sent to the server and not the entire model represented as a JSON.

vkaracic, line 109: Okay. Should you verify that the `isValid` method is called as well?

marjev, line 109: I can do that, but that's testing backbone and not our code.

vkaracic:
:+1:

PR #825, opened 2016-06-27
[SOL-1859] Omit unavailable seats in offer landing page.

vkaracic:
If a dynamic coupon would contain one or more seats that are unavailable (expired) an error message would be displayed to the user trying to redeem some of the seats, so one expired seat could render the coupon for 1000 seats useless.
This PR contains changes that omit unavailable seats for the coupon offer landing page, displaying to the user only the seats that are available to her.

The offer page for single-course coupons works are before, displaying an error when the seat is unavailable.

https://openedx.atlassian.net/browse/SOL-1859

### DIFF
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -90,10 +90,10 @@ def voucher_is_valid(voucher, products, request):
             voucher_msg = msg.replace('voucher', 'coupon')
             return False, voucher_msg
 
-    for product in products:
-        purchase_info = request.strategy.fetch_for_product(product)
+    if len(products) == 1:
+        purchase_info = request.strategy.fetch_for_product(products[0])
         if not purchase_info.availability.is_available_to_buy:
-            return False, _('Product [{product}] not available for purchase.'.format(product=product))
+            return False, _('Product [{product}] not available for purchase.'.format(product=products[0]))
 
     # If the voucher's number of applications exceeds it's limit.
     offer = voucher.offers.first()
--- a/ecommerce/extensions/api/v2/views/vouchers.py
+++ b/ecommerce/extensions/api/v2/views/vouchers.py
@@ -110,7 +110,6 @@ class VoucherViewSet(NonDestroyableModelViewSet):
         benefit = voucher.offers.first().benefit
         catalog_query = benefit.range.catalog_query
         offers = []
-
         if catalog_query:
             query_results = request.site.siteconfiguration.course_catalog_api_client.course_runs.get(
                 q=catalog_query,
@@ -124,6 +123,11 @@ class VoucherViewSet(NonDestroyableModelViewSet):
             contains_verified_course = (benefit.range.course_seat_types == 'verified')
 
             for product in products:
+                # Omit unavailable seats from the offer results so that one seat does not cause an
+                # error message for every seat in the query result.
+                if not request.strategy.fetch_for_product(product).availability.is_available_to_buy:
+                    logger.info('%s is unavailable to buy. Omitting it from the results.', product)
+                    continue
                 course_id = product.course_id
                 course_catalog_data = next((result for result in query_results if result['key'] == course_id), None)
 
--- a/ecommerce/coupons/tests/mixins.py
+++ b/ecommerce/coupons/tests/mixins.py
@@ -19,22 +19,23 @@ class CatalogPreviewMockMixin(object):
     def setUp(self):
         super(CatalogPreviewMockMixin, self).setUp()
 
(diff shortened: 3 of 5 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=825

vkaracic:
Reviewers:
- [x] @marjev 
- [x] @mattdrayer

ecommerce/coupons/views.py
```diff
@@ -90,10 +90,10 @@ def voucher_is_valid(voucher, products, request):
             voucher_msg = msg.replace('voucher', 'coupon')
             return False, voucher_msg
 
-    for product in products:
-        purchase_info = request.strategy.fetch_for_product(product)
+    if len(products) == 1:
+        purchase_info = request.strategy.fetch_for_product(products[0])
         if not purchase_info.availability.is_available_to_buy:
```
marjev, line 95: Why is this check done only when one product is associated with a voucher?

vkaracic, line 95: For a single-course coupon this check is required. Multi-course goes through the same check and needs to skip this one.

marjev, line 95: So, I guess there's no way to make a check somewhere in the process for both types of coupons? Just for the sake of maintenance?

marjev:
Generally LGTM :+1:  (if possible, perhaps we should consider making the check only once for all types of coupons, but this can be done at a later time)

ecommerce/coupons/views.py
vkaracic, line 95: Not that I'm aware of. The issue is that this particular check is done at the offer landing page view, while the seats for the multi-course coupon are prepared in a voucher API endpoint (from which backbone retrieves and displays them). There is some redundancy there that we could/should address.

marjev, line 95: Yeah, we should go through all the code and create cleanup tickets once we get through the post-release tickets.

vkaracic, line 95: Couldn't agree more.

On Mon, Jun 27, 2016 at 4:44 PM, Marko Jevtić notifications@github.com
wrote:

> In ecommerce/coupons/views.py
> https://github.com/edx/ecommerce/pull/825#discussion_r68588806:
> 
> > @@ -90,10 +90,10 @@ def voucher_is_valid(voucher, products, request):
> >              voucher_msg = msg.replace('voucher', 'coupon')
> >              return False, voucher_msg
> > -    for product in products:
> > -        purchase_info = request.strategy.fetch_for_product(product)
> > -    if len(products) == 1:
> > -        purchase_info = request.strategy.fetch_for_product(products[0])
> >        if not purchase_info.availability.is_available_to_buy:
> 
> Yeah, we should go through all the code and create cleanup tickets once we
> get through the post-release tickets.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> https://github.com/edx/ecommerce/pull/825/files/42ad808ef8f58bf0a3c0ec37b8b2e79c06d48fca#r68588806,
> or mute the thread
> https://github.com/notifications/unsubscribe/ACrZHIzss_lnUJ5e_1VM8BKVT9dlhMq8ks5qP-HHgaJpZM4I_ICB
> .

## 

Vedran Karačić

Mail:     vedran@edx.org
             vkaracic@extensionengine.com
Github: vkaracic https://github.com/vkaracic
Skype:  karacicvedran

mattdrayer, line 95: Not a requirement to merge this PR, but let's go ahead and log the meta-ticket to do the discovery work for post-launch cleanup

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -124,6 +123,10 @@ def get_offers(self, products, request, voucher):
             contains_verified_course = (benefit.range.course_seat_types == 'verified')
 
             for product in products:
+                # Omit unavailable seats from the offer results so that one seat does not cause an
+                # error message for every seat in the query result.
+                if not request.strategy.fetch_for_product(product).availability.is_available_to_buy:
+                    continue
```
mattdrayer, line 130: Would it be valuable to write out a log entry here?

vkaracic, line 130: Could be, I'll add one.

ecommerce/coupons/views.py
vkaracic, line 95: Created: https://openedx.atlassian.net/browse/SOL-1867

mattdrayer:
LGTM -- 👍 after checks pass

PR #827, opened 2016-06-27
[SOL-1861] Include max_uses when editing a coupon.

vkaracic:
When editing a coupon, `max_uses` attribute was not accounted for causing the coupon to default to 10000 max uses when edited.
This PR contains changes that account for the `max_uses` attribute when updating a coupon.

Reviewers:
- [x] @marjev 
- [x] @mattdrayer 

https://openedx.atlassian.net/browse/SOL-1861

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -394,7 +394,8 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
             offer=voucher_offer,
             benefit_value=benefit_value,
             benefit_type=voucher_offer.benefit.type,
-            coupon=coupon
+            coupon=coupon,
+            max_uses=voucher_offer.max_global_applications
         )
         for voucher in vouchers.all():
             voucher.offers.clear()
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -480,14 +480,17 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CatalogPr
     def test_update_coupon_benefit_value(self):
         coupon = Product.objects.get(title='Test coupon')
         vouchers = coupon.attr.coupon_vouchers.vouchers.all()
+        max_uses = vouchers[0].offers.first().max_global_applications
+        benefit_value = Decimal(54)
+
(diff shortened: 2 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=827

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -340,7 +340,12 @@ def update(self, request, *args, **kwargs):
 
         benefit_value = request.data.get(AC.KEYS.BENEFIT_VALUE, '')
         if benefit_value:
-            self.update_coupon_benefit_value(benefit_value=benefit_value, vouchers=vouchers, coupon=coupon)
+            self.update_coupon_benefit_value(
+                benefit_value=benefit_value,
+                vouchers=vouchers,
+                coupon=coupon,
+                max_uses=request.data.get('max_uses')
```
marjev: Don't do this here. Check to see if request data contains max_uses. Create a similar thing like it's done for voucher dates and range data above. This will be important for editing fix I'm currently working on.

marjev:
Let me know once the change I suggested is implemented so that I might take another look.

ecommerce/extensions/api/v2/views/coupons.py
vkaracic: I don't fully understand, but `max_uses` is not editable now if you mean to add it to an updateable fields list?

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -378,7 +383,7 @@ def create_update_data_dict(self, request_data, request_data_key, update_dict, u
             update_dict[update_dict_key] = prepare_course_seat_types(value) \
                 if update_dict_key == COURSE_SEAT_TYPES else value
 
-    def update_coupon_benefit_value(self, benefit_value, coupon, vouchers):
+    def update_coupon_benefit_value(self, benefit_value, coupon, vouchers, max_uses=None):
```
mattdrayer: What does it mean for max_uses to be None in this context?

vkaracic: `None` is the default value for this attribute in the offer condition. When no `max_uses` is defined the coupon will fall back to the default 10000 maximum usages.

mattdrayer:
👍

marjev:
👍 write a test to make sure that max_uses value remains the same after the update.

PR #828, opened 2016-06-27
[SOL-1863] Check if course seat types attribute exists before using it

vkaracic:
Old coupons (created before addition of dynamic coupons) do not contain `course_seat_types` attribute which causes error in the coupon details page.

This PR contains changes that fix that error in addition with a test verifying that the dynamic coupon specific attributes are empty. There is a test for when they exist: https://github.com/edx/ecommerce/blob/61bb1a0dcd34a5a660bb5b8ca750e3ea9be91c1e/ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js#L145-L169

Reviewers:
- [x] @marjev 
- [x] @mattdrayer 

https://openedx.atlassian.net/browse/SOL-1863

### DIFF
--- a/ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
+++ b/ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
@@ -128,6 +128,8 @@ define([
                 expect(view.$el.find('.tax-deducted-source-value > .value').text()).toEqual(
                     view.taxDeductedSource(model.get('tax_deducted_source'))
                 );
+                expect(view.$('.seat-types > .value').text()).toEqual('');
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=828

mattdrayer:
Code LGTM, but appears you need more test coverage.

mattdrayer:
👍 after checks pass

mattdrayer:
Looks like you just need to hit the null return condition

marjev:
👍

PR #829, opened 2016-06-27
[SOL-1835] Add guideline link to catalog query string.

vkaracic:
screenshot from 2016-06-27 18-08-57

---

https://openedx.atlassian.net/browse/SOL-1835
Bullet points 1, 2 and 3 have been addressed in previous work:

---

screenshot from 2016-06-27 18-11-38

---

screenshot from 2016-06-27 18-11-32

Reviewers:
- [x] @srpearce 
- [x] Roderick Morales

### DIFF
--- a/ecommerce/static/templates/coupon_form.html
+++ b/ecommerce/static/templates/coupon_form.html
@@ -181,7 +181,7 @@
             <p class="help-block"></p>
         </div>
         <div class="form-group catalog-query">
-            <label for="catalog-query"><%= gettext('Valid for:') %> *</label>
+            <label for="catalog-query"><%= gettext('Query string:') %> * <a href="https://stage-edx-discovery.edx.org/" class="normal-font-weight">(query guidelines)</a></label>
             <textarea id="catalog-query" class="form-control" name="catalog_query" rows="10"></textarea>
             <p class="help-block"></p>
         </div>


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=829

ecommerce/static/templates/coupon_form.html
```diff
@@ -181,7 +181,7 @@
             <p class="help-block"></p>
         </div>
         <div class="form-group catalog-query">
-            <label for="catalog-query"><%= gettext('Valid for:') %> *</label>
+            <label for="catalog-query"><%= gettext('Query string:') %> * (<a href="https://stage-edx-discovery.edx.org/" class="normal-font-weight">guideline</a>)</label>
```
srpearce: guideline --> query hints

I would also make the parentheses around "query hints" part of the link - it looks a little funny to have the parentheses in bold dark gray and the link text in non-bold blue.

vkaracic: Here's an updated version:
screenshot from 2016-06-28 09-58-23

roderickmorales:
@srpearce I don't want to get too picky but is "hints" the right word here? Isn't more like "query syntax documentation"... "hint" sounds like a clue instead of a guide.

srpearce:
How about "query guidelines"? (I didn't want to have just "guidelines" because it didn't specify what the guidelines were for, and I was thinking in terms of screen readers.)

vkaracic:
`query guidelines` it is!

roderickmorales:
👍

PR #832, opened 2016-06-29
Updated edx-drf-extensions

clintonb:
This update ensures we properly decode JWTs and create staff users when retrieving requests from other services.

ECOM-4814


clintonb:
@edx/ecommerce please review

rlucioni:
👍

PR #835, opened 2016-06-29
[SOL-1830] Update Voucher Report for Dynamic Catalog Coupon

marjev:
Here's a PR for updating Voucher Report to include more information for Dynamic Catalog Coupons.

JIRA ticket describing the requirements can be found here: https://openedx.atlassian.net/browse/SOL-1830

Requirements are:
1: Add a column to display the query and course seat types
2: If a coupon has been redeemed add the course id for which it was used.

@mattdrayer, @vkaracic can you please review this?

### DIFF
--- a/ecommerce/extensions/voucher/utils.py
+++ b/ecommerce/extensions/voucher/utils.py
@@ -58,6 +58,26 @@ def _get_voucher_status(voucher, offer):
     return status
 
 
+def _get_discount_info(discount_data):
+    """Retrieve voucher discount info.
+
+    Arguments:
+        discount_data (dict)
+
+    Returns
+        coupon_type (str)
+        discount_amount (str)
+        discount_percentage (str)
+    """
+    if discount_data:
+        coupon_type = _('Discount') if discount_data['is_discounted'] else _('Enrollment')
+        discount_percentage = _("{percentage} %").format(percentage=discount_data['discount_percentage'])
+        discount_amount = currency(discount_data['discount_value'])
+        return coupon_type, discount_percentage, discount_amount
+    logger.warning('Unable to get voucher discount info. Discount data not provided.')
+    return None, None, None
+
+
 def _get_info_for_coupon_report(coupon, voucher):
     offer = voucher.offers.all().first()
     if offer.condition.range.catalog:
@@ -70,27 +90,19 @@ def _get_info_for_coupon_report(coupon, voucher):
         discount_data = get_voucher_discount_info(offer.benefit, seat_stockrecord.price_excl_tax)
     else:
         # Note (multi-courses): Need to account for multiple seats.
+        catalog_query = offer.condition.range.catalog_query
+        course_seat_types = offer.condition.range.course_seat_types
+        course_id = None
         coupon_stockrecord = None
         invoiced_amount = None
         seat_stockrecord = None
-        course_id = None
         course_organization = None
         price = None
         discount_data = None
 
     history = coupon.history.first()
-
-    if discount_data:
-        coupon_type = _('Discount') if discount_data['is_discounted'] else _('Enrollment')
-        discount_percentage = _("{percentage} %").format(percentage=discount_data['discount_percentage'])
-        discount_amount = currency(discount_data['discount_value'])
-    else:
-        coupon_type = None
-        discount_percentage = None
-        discount_amount = None
-
+    coupon_type, discount_percentage, discount_amount = _get_discount_info(discount_data)
     status = _get_voucher_status(voucher, offer)
-
     path = '{path}?code={code}'.format(path=reverse('coupons:offer'), code=voucher.code)
     url = get_ecommerce_url(path)
     author = history.history_user.full_name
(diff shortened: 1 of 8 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=835

marjev:
@mjfrey can you take a look as well?

vkaracic:
Before merging this, I'd like to use this opportunity to get these tickets addressed:
https://openedx.atlassian.net/browse/SOL-1782
https://openedx.atlassian.net/browse/SOL-1757
@mjfrey

marjev:
What we should do is try and make a single story for each milestone so that a single PR is opened for reporting purposes per milestone.

ecommerce/extensions/voucher/utils.py
```diff
@@ -58,6 +58,25 @@ def _get_voucher_status(voucher, offer):
...
+    Arguments:
+        discount_data (dict)
+
+    Returns
+        coupon_type (str)
+        discount_amount (str)
+        discount_percentage (str)
+    """
+    if discount_data:
+        coupon_type = _('Discount') if discount_data['is_discounted'] else _('Enrollment')
```
mattdrayer, line 73: We moved product names into constant values to avoid scattering the literal strings all over the place -- I'm wondering if it makes sense to move the coupon type strings into constant values as well?

ecommerce/extensions/voucher/utils.py
```diff
@@ -58,6 +58,25 @@ def _get_voucher_status(voucher, offer):
...
+        discount_percentage = _("{percentage} %").format(percentage=discount_data['discount_percentage'])
```
mattdrayer, line 74: Do we need to worry about None values here, or will there always be a value provided?

ecommerce/extensions/voucher/utils.py
```diff
@@ -58,6 +58,25 @@ def _get_voucher_status(voucher, offer):
...
+        discount_amount = currency(discount_data['discount_value'])
```
mattdrayer, line 75: Same as above -- is it possible for a None value to find its way into this reference?

ecommerce/extensions/voucher/utils.py
```diff
@@ -70,27 +89,19 @@ def _get_info_for_coupon_report(coupon, voucher):
...
-        coupon_type = _('Discount') if discount_data['is_discounted'] else _('Enrollment')
-        discount_percentage = _("{percentage} %").format(percentage=discount_data['discount_percentage'])
-        discount_amount = currency(discount_data['discount_value'])
-    else:
-        coupon_type = None
-        discount_percentage = None
-        discount_amount = None
-
+    coupon_type, discount_percentage, discount_amount = _get_discount_info(discount_data)
     status = _get_voucher_status(voucher, offer)
```
mattdrayer, line 105: If we receive None,None,None from the above helper method here, is there any point in continuing?  Should we write an entry to the log if this happens?  Is there a graceful failure mode we could enter?

ecommerce/extensions/voucher/utils.py
marjev, line 73: Yes, it does. There are tasks opened for this cleanup, which is currently underway. Is it ok if we do that outside of this PR, just that we can track the change globally?

ecommerce/extensions/voucher/utils.py
marjev, line 74: It will always be provided. The utility method that formats the discount_data will return 
`{ 'discount_percentage': 0.00, 'discount_value': 0.00, 'is_discounted': False }` in case benefit doesn't exist or the product (seat) doesn't have a price.

ecommerce/extensions/voucher/utils.py
marjev, line 75: I provided the explanation above.

ecommerce/extensions/voucher/utils.py
marjev, line 105: Since discount data is always formatted by the utility method (default values are returned in edge cases), `None,None,None` values will be returned only if the expected argument is an empty dictionary/None/empty list or an empty string. In this instance, that will never happen.

We just wanted to make sure that sending unexpected values returns something. Should we handle that integrity edge case by logging some message?

ecommerce/extensions/voucher/utils.py
mattdrayer, line 73: It's okay with me as long as we have a ticket logged to perform this cleanup work, yep.

ecommerce/extensions/voucher/utils.py
mattdrayer, line 105: I'm always okay with logging a message if it's useful information for debugging an unexpected scenario

marjev, line 105: I've added a warning message. Please see 5c93debaceab548deb7e9378e344ce65cff89644

mattdrayer:
👍 -- please squash commits before merging

PR #837, opened 2016-06-30
[SOL-1729] UI issue in create coupon page.

iivic:
Few margins changed to line up fields. @marjev @vkaracic 
screen shot 2016-06-30 at 12 27 40
screen shot 2016-06-30 at 12 27 53
screen shot 2016-06-30 at 12 28 13

### DIFF
--- a/ecommerce/static/sass/partials/views/_coupon_admin.scss
+++ b/ecommerce/static/sass/partials/views/_coupon_admin.scss
@@ -16,8 +16,13 @@
         width: $input-addon-width + $small-input-width;
     }
 
+    .hidden.help-block {
+        display: block !important; // !important to override hidden class
+    }
+
     .help-block {
         margin: 0;
+        height: 23px;
     }
 
     .catalog-type {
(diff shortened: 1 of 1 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=837

ecommerce/static/sass/partials/views/_coupon_admin.scss
```diff
@@ -18,6 +18,8 @@
 
     .help-block {
         margin: 0;
+        height: 23px;
+        display: block !important; // !important to override hidden class
```
marjev: Did you try applying this style when .help-block has .hidden class? We usually try to avoid using !important.

iivic: Yes i tried. This is the only way i was able to apply this change.

iivic: 👍

marjev:
👍

PR #838, opened 2016-07-01
[SOL-1856] User is taken back to coupons home on clicking next button…

iivic:
When clicked next, previous or active button nothing should happen. @marjev @vkaracic

### DIFF
--- a/ecommerce/static/js/views/dynamic_catalog_view.js
+++ b/ecommerce/static/js/views/dynamic_catalog_view.js
@@ -15,13 +15,18 @@ define(['jquery',
             template: _.template(DynamicCatalogButtons),
 
             events: {
-                'click [name=preview_catalog]': 'previewCatalog'
+                'click [name=preview_catalog]': 'previewCatalog',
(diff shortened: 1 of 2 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=838&repo=edx%2Fecommerce

vkaracic:
You are missing a test for this.

iivic:
Comment addressed.

ecommerce/static/js/views/dynamic_catalog_view.js
```diff
@@ -24,6 +25,10 @@ define(['jquery',
                 this._super();
             },
 
+            stopEventPropagation: function(event) {
```
marjev, line 28: You never call this function in tests.

iivic, line 28: Done.

vkaracic:
:+1:

marjev:
👍

PR #839, opened 2016-07-04
[SOL-1820] Invoice management command

vkaracic:
Management command for creating invoices for coupon orders that do not have them.

### DIFF
--- a/ecommerce/invoice/management/commands/create_invoices.py
+++ b/ecommerce/invoice/management/commands/create_invoices.py
@@ -1,0 +1,38 @@
+""" This command invoices for orders that don't have them. """
+import logging
+
+from django.core.management import BaseCommand
+
+from ecommerce.core.models import BusinessClient
+from ecommerce.extensions.basket.models import Basket
+from ecommerce.extensions.catalogue.models import Product
+from ecommerce.extensions.order.models import Order
+from ecommerce.extensions.payment.processors.invoice import InvoicePayment
+from ecommerce.invoice.models import Invoice
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Create invoices for orders that don't have them.
+    At some point in the past we decided to use the BusinessClient from invoices,
+    before that it was used from basket objects, and the orders back then didn't have
+    invoices. This command creates invoices for those orders.
+    """
+
+    def handle(self, *args, **options):
+        for coupon in Product.objects.filter(product_class__name='Coupon'):
+            try:
+                basket = Basket.objects.get(lines__product=coupon, status=Basket.SUBMITTED)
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=839

vkaracic:
@mattdrayer please review.

ecommerce/invoice/management/commands/create_invoices.py
```diff
@@ -0,0 +1,38 @@
...
+
+class Command(BaseCommand):
+    """
+    Create invoices for orders that don't have them.
+    At some point in the past we decided to use the BusinessClient from invoices,
+    before that it was used from basket objects, and the orders back then didn't have
+    invoices. This command creates invoices for those orders.
+    """
+
+    ch = logging.StreamHandler()
```
mattdrayer: sorry, what does 'ch' stand for?  Maybe we can name it better?

ecommerce/invoice/management/commands/create_invoices.py
```diff
@@ -0,0 +1,38 @@
...
+    ch.setLevel(logging.DEBUG)
+    logger.addHandler(ch)
+
+    def handle(self, *args, **options):
+        for coupon in Product.objects.filter(product_class__name='Coupon'):
+            basket = Basket.objects.get(lines__product=coupon, status=Basket.SUBMITTED)
```
mattdrayer: Basket.DoesNotExist?

ecommerce/invoice/management/commands/create_invoices.py
```diff
@@ -0,0 +1,38 @@
...
+            order = Order.objects.get(basket=basket)
```
mattdrayer: Order.DoesNotExist?

ecommerce/invoice/management/commands/create_invoices.py
vkaracic: It's part of most of the management commands, and I think it's not really needed so I'll remove it altogether.

ecommerce/invoice/management/commands/create_invoices.py
vkaracic: The coupon, basket and order are created within an atomic transaction meaning that every coupon must have a basket and order, or otherwise it won't be created.
That being said, it's always possible for data to be bugged (like the one course on stage with 2 seat types) so I'll add loggers in case these don't exist.

vkaracic:
@mattdrayer comments addressed!

mattdrayer:
👍

marjev:
LGTM 👍

iivic:
👍

PR #836, opened 2016-06-29
[SOL-1835] Coupon UI fixes

vkaracic:
Invoice data validation; required field asteriks; routing of external links; render invoice data on coupon details page.

https://openedx.atlassian.net/browse/SOL-1835

PR #840, opened 2016-07-04
Changed asterisk on coupon creation form page.

vkaracic:
https://openedx.atlassian.net/browse/SOL-1835
Partially addressed in https://github.com/edx/ecommerce/pull/836 but I missed a couple of fields. This PR contains the last changes for the workflow. Screenshots below.

screenshot from 2016-07-04 14-00-53
screenshot from 2016-07-04 14-01-32

### DIFF
--- a/ecommerce/static/js/views/coupon_form_view.js
+++ b/ecommerce/static/js/views/coupon_form_view.js
@@ -479,8 +479,8 @@ define([
                 course.fetch({data: {include_products: true}});
             },
 
-            /*
-             * Update price field and model.stockrecords
+            /**
+             * Update price field and model.stockrecords.
              */
             changeSeatType: function () {
                 var seatType = this.getSeatType(),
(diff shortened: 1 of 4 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=840

vkaracic:
@srpearce @roderickmorales please review this.

roderickmorales:
👍

srpearce:
👍  for this. How do I request a change to the error message? I would recommend deleting "Error!" (which sounds a bit abrupt) and just leaving it as "Please complete all required fields."

vkaracic:
That part is used by both the coupon form and course form, and I'm not sure if we're allowed to change it on the course side as well.

vkaracic:
Disregard that. Here's the latest:
screenshot from 2016-07-07 08-08-46

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -527,15 +529,39 @@ define([
...
                 e.stopPropagation();
                 window.open(e.currentTarget.href);
             },
 
+
+            /**
+             * Override default renderAlert to display a custom message.
+             */
+            renderAlert: function (level, message) {
+                if (!this.model.isValid(true)) {
```
marjev, line 545: You can DRY this by applying polymorphism. Common code should reside in the base function.

srpearce:
👍

marjev:
👍

PR #841, opened 2016-07-05
[SOL-1907] Set coupon custom code lower bound limit to 1.

vkaracic:
There are legacy codes from ShoppingCart that need to be moved to Otto which have their code length lower than 8, and Otto has a lower bound limit of 8 for custom codes.

This PR changes the lower bound limit from 8 to 1 and adds a check if the code already exists.

https://openedx.atlassian.net/browse/SOL-1907

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -6,7 +6,6 @@ from decimal import Decimal
 import dateutil.parser
 from django.conf import settings
 from django.db import transaction
-from django.db.utils import IntegrityError
 from django.http import Http404
 from django.shortcuts import get_object_or_404
 from oscar.core.loading import get_model
@@ -106,6 +105,8 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
         Returns:
             200 if the order was created successfully; the basket ID is included in the response
                 body along with the order ID and payment information.
+            400 if a custom code is received that already exists,
+                if a course mode is selected that is not supported.
             401 if an unauthenticated request is denied permission to access the endpoint.
             429 if the client has made requests at a rate exceeding that allowed by the configured rate limit.
             500 if an error occurs when attempting to create a coupon.
@@ -129,6 +130,16 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
             max_uses = request.data.get('max_uses')
             catalog_query = request.data.get(CATALOG_QUERY)
             course_seat_types = request.data.get(COURSE_SEAT_TYPES)
+
+            if code:
+                try:
+                    Voucher.objects.get(code=code)
+                    return Response(
+                        'A coupon with code {code} already exists.'.format(code=code),
+                        status=status.HTTP_400_BAD_REQUEST
+                    )
+                except Voucher.DoesNotExist:
+                    pass
 
             invoice_data = self.retrieve_invoice_data(request.data)
 
@@ -218,10 +229,6 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
 
         Returns:
             A coupon product object.
-
-        Raises:
-            IntegrityError: An error occured when create_vouchers method returns
-                            an IntegrityError exception
         """
 
         product_class = ProductClass.objects.get(slug='coupon')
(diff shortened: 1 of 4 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=841&repo=edx%2Fecommerce

ecommerce/static/js/models/coupon_model.js
```diff
@@ -93,8 +93,8 @@ define([
                 },
                 code: {
                     required: false,
-                    rangeLength: [8, 16],
-                    msg: gettext('Code field must be empty or between 8 and 16 characters')
+                    rangeLength: [1, 16],
+                    msg: gettext('Code field must be empty or between 1 and 16 characters')
```
marjev: So, if I understand correctly, the only validation here is that the code must have max 16 characters? Should the message be changed? If so, tag someone who can provide such a message.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -130,6 +131,13 @@ def create(self, request, *args, **kwargs):
             catalog_query = request.data.get(CATALOG_QUERY)
             course_seat_types = request.data.get(COURSE_SEAT_TYPES)
 
+            if code:
+                if code in Voucher.objects.values_list('code', flat=True):
```
marjev: You can simplify this. Try `Voucher.objects.get('code')` and return the Response. If the exception is raised, simply continue (pass) the execution.

marjev:
Let me know when you've addressed my comments so that I can take another look.

vkaracic:
Comments addressed!

marjev:
👍

vkaracic:
@mattdrayer please review.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -106,6 +105,8 @@ def create(self, request, *args, **kwargs):
         Returns:
             200 if the order was created successfully; the basket ID is included in the response
                 body along with the order ID and payment information.
+            400 if an custom code is received that already exists
```
mattdrayer: Nit: "if a custom code..."

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -130,6 +131,16 @@ def create(self, request, *args, **kwargs):
             catalog_query = request.data.get(CATALOG_QUERY)
             course_seat_types = request.data.get(COURSE_SEAT_TYPES)
 
+            if code:
+                try:
+                    Voucher.objects.get(code=code)
+                    return Response(
+                        'A coupon with code {code} already exist.'.format(code=code),
```
mattdrayer: Nit: "exists"

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -231,25 +238,21 @@ def create_coupon_product(self, title, price, data):
...
-                start_datetime=data['start_date'],
-                voucher_type=data['voucher_type'],
-                max_uses=data['max_uses'],
-                catalog_query=data['catalog_query'],
-                course_seat_types=data['course_seat_types']
-            )
-        except IntegrityError:
-            logger.exception('Failed to create vouchers for [%s] coupon.', coupon_product.title)
-            raise
+        create_vouchers(
```
mattdrayer, line 241: Is it no longer necessary to wrap this call with a try/except of any sort?  Great if that is the case, but I always question when protection is removed.

ecommerce/static/js/models/coupon_model.js
```diff
@@ -93,8 +93,8 @@ define([
                 },
                 code: {
                     required: false,
-                    rangeLength: [8, 16],
-                    msg: gettext('Code field must be empty or between 8 and 16 characters')
+                    rangeLength: [1, 16],
+                    msg: gettext('Code field must be empty or up to 16 characters long.')
```
mattdrayer: Can we just say "Code field must be 0 to 16 characters long." ?

ecommerce/extensions/api/v2/views/coupons.py
vkaracic, line 241: Try try-except block was in place to protect against creating a voucher with a custom code that already exists. That test has been moved further up.

mattdrayer, line 241: Yes, but here we were catching an IntegrityError, and above it is a DoesNotExist ?

vkaracic, line 241: An integrity error is raised when a new model object is created with a unique restriction violation. In the code above I'm not creating a new voucher, but rather trying to retrieve one - if it's retrieved that means that a voucher with that code already exists, meaning that there will be an integrity error triggered here on creation, hence the 400 response.

ecommerce/static/js/models/coupon_model.js
vkaracic: It's best to involve @srpearce here.

vkaracic: @srpearce

srpearce: Optional. If you enter a code, it must contain 1-16 alphanumeric characters.

vkaracic:
Guys, all comments have been addressed. Do you mind taking a second look?

marjev:
Looks good! 👍

mattdrayer:
👍 although looks like you need to rebase

PR #843, opened 2016-07-05
Add edx-django-release-util

efischer19:
[TE-1500](https://openedx.atlassian.net/browse/TE-1500)


efischer19:
@mattdrayer Can you give this a quick review? It's adding tools that aren't used currently, but could help the platform/deployment team in the future.

mattdrayer:
LGTM -- 👍 -- but we should probably have @rlucioni take a look as well

rlucioni:
👍

PR #845, opened 2016-07-07
[SOL-1821] Set redemption count for single use vouchers to 1 if used.

vkaracic:
Since the vouchers use the same offer, when redeeming them the `num_application` increments for all vouchers making the `Redemption Count` column inconsistent in the coupon report.

This PR contains changes that change the redemption count to the number of orders for single use vouchers.

https://openedx.atlassian.net/browse/SOL-1821


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=845

iivic:
Do you need a test to check redemption_count is set properly??

vkaracic:
Test added!

marjev:
LGTM 👍

iivic:
👍

PR #846, opened 2016-07-07
[SOL-1837] Internationalize the coupon fields

vkaracic:
Characters like `čćšđž` in various coupon fields would trigger encoding errors.
This PR contains changes that account for non-ASCII characters in coupons fields:
- coupon name
- note
- code
- client
- invoice number

https://openedx.atlassian.net/browse/SOL-1837

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -1,4 +1,6 @@
 """Serializers for data manipulated by ecommerce API endpoints."""
+from __future__ import unicode_literals
+
 from decimal import Decimal
 import logging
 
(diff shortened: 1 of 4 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=846

iivic:
LGTM 👍

marjev:
Perhaps you could use `from __future__ import unicode_literals` to avoid the need to specify which literals contain a non-ASCII character.

vkaracic:
Comment addressed!

marjev:
LGTM 👍

PR #848, opened 2016-07-07
[SOL-1916] Category ID's not sent during coupon creation

iivic:
@vkaracic  @marjev  Category_ids were not set on init. Should be fixed now.

### DIFF
--- a/ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
+++ b/ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
@@ -99,6 +99,10 @@ define([
                     expect(visible('[name=benefit_value]')).toBe(false);
                     expect(visible('[name=code]')).toBe(false);
                 });
+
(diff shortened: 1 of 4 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=848

vkaracic:
Why was this happening in the first place?

iivic:
Category_ids were not set when the page rendered. So when the save button is pressed there is nothing to send/save on that field. Adding this line 'category_ids': [defaultCategory[0].id] in the render function fixed the problem.

marjev:
This LGTM 👍 I'm currently working on cleaning up this mess, so that won't be a problem for long. We can use this fix until then to make sure that users can create coupons without problems.

vkaracic:
LGTM

PR #852, opened 2016-07-08
Fill in required invoice fields for coupon acceptance tests.

mjfrey:
Need to add new required invoice fields to coupon creation for acceptance tests.


mjfrey:
@vkaracic

vkaracic:
The reason why this needs to be done is because invoice fields are made required. If these fields are not needed, you can just select `N/A` as the invoice type without needing to fill out any additional invoice field.
If you want to leave it like this, I suggest filling out the `price` field as well which would cover all required fields.

mjfrey:
@vkaracic added price.

vkaracic:
:+1:

PR #856, opened 2016-07-11
Pwattenberger/sailthru

PaulWattenberger:
Won't build until changes to ecommerce-worker merged

### DIFF
--- a/ecommerce/sailthru/signals.py
+++ b/ecommerce/sailthru/signals.py
@@ -41,6 +41,7 @@ def process_checkout_complete(sender, order=None, request=None, user=None, **kwa
     # pass event to ecommerce_worker.sailthru.v1.tasks to handle asynchronously
     update_course_enrollment.delay(user.email, course_url, False, product.attr.certificate_type,
                                    unit_cost=price, course_id=course_id, currency=order.currency,
+                                   site_code=request.site.siteconfiguration.partner.short_code,
                                    message_id=request.COOKIES.get('sailthru_bid'))
 
 
@@ -73,12 +74,13 @@ def process_basket_addition(sender, product=None, request=None, user=None, **kwa
     # pass event to ecommerce_worker.sailthru.v1.tasks to handle asynchronously
     update_course_enrollment.delay(user.email, course_url, True, product.attr.certificate_type,
                                    unit_cost=price, course_id=course_id, currency=currency,
+                                   site_code=request.site.siteconfiguration.partner.short_code,
                                    message_id=request.COOKIES.get('sailthru_bid'))
 
 
 def _build_course_url(course_id):
-    """
-    Build a course url from a course id and the host
+    """Build a course url from a course id and the host
+
     :param request:
     :param course_id:
     :return:
--- a/ecommerce/sailthru/signals.py
+++ b/ecommerce/sailthru/signals.py
@@ -29,7 +29,8 @@ def process_checkout_complete(sender, order=None, request=None, user=None, **kwa
 
     # loop through lines in order
     #  If multi product orders become common it may be worthwhile to pass an array of
-    #  orders to the worker in one call to save overhead
+    #  orders to the worker in one call to save overhead, however, that would be difficult
+    #  because of the fact that there are different templates for free enroll versus paid enroll
     for line in order.lines.all():
 
         # get product
--- a/ecommerce/sailthru/signals.py
+++ b/ecommerce/sailthru/signals.py
@@ -1,8 +1,8 @@
 import logging
 
-from django.conf import settings
 from django.dispatch import receiver
 from oscar.core.loading import get_class
+import waffle
 
 from ecommerce_worker.sailthru.v1.tasks import update_course_enrollment
 from ecommerce.extensions.analytics.utils import silence_exceptions
@@ -19,10 +19,11 @@ basket_addition = get_class('basket.signals', 'basket_addition')
 def process_checkout_complete(sender, order=None, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
     """Tell Sailthru when payment done.
 
-    Parameters described at http://django-oscar.readthedocs.io/en/releases-1.1/ref/signals.html
+    Arguments:
+            Parameters described at http://django-oscar.readthedocs.io/en/releases-1.1/ref/signals.html
     """
 
-    if not settings.SAILTHRU_ENABLE:
+    if not waffle.switch_is_active('sailthru_enable'):
         return
 
     # get product (should only be 1)
@@ -30,7 +31,7 @@ def process_checkout_complete(sender, order=None, request=None, user=None, **kwa
 
     # return if no price, since enrolls are handled by lms
     price = order.total_excl_tax
-    if not price or price == 0:
+    if not price:
         return
 
     course_id = product.course_id
@@ -46,14 +47,15 @@ def process_checkout_complete(sender, order=None, request=None, user=None, **kwa
 
 
 @receiver(basket_addition)
-@silence_exceptions("Failed to call Sailthru upon backet addition.")
+@silence_exceptions("Failed to call Sailthru upon basket addition.")
 def process_basket_addition(sender, product=None, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
     """Tell Sailthru when payment started.
 
-    Parameters described at http://django-oscar.readthedocs.io/en/releases-1.1/ref/signals.html
+    Arguments:
+            Parameters described at http://django-oscar.readthedocs.io/en/releases-1.1/ref/signals.html
     """
 
-    if not settings.SAILTHRU_ENABLE:
+    if not waffle.switch_is_active('sailthru_enable'):
         return
 
     course_id = product.course_id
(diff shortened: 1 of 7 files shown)


PaulWattenberger:
@rlucioni 
@cpennington

cpennington:
:+1:, once the tests pass.

ecommerce/settings/base.py
```diff
@@ -541,3 +546,6 @@
 
 # Affiliate cookie key
 AFFILIATE_COOKIE_KEY = 'affiliate_id'
+
+# Settings for Sailthru email marketting integration
+SAILTHRU_ENABLE = False
```
rlucioni: Use a django-waffle switch instead of a Django setting.

ecommerce/settings/base.py
```diff
@@ -479,9 +482,11 @@
 # See http://celery.readthedocs.org/en/latest/configuration.html#celery-imports.
 CELERY_IMPORTS = (
     'ecommerce_worker.fulfillment.v1.tasks',
+    'ecommerce.sailthru.tasks',
 )
 
-CELERY_ROUTES = {'ecommerce_worker.fulfillment.v1.tasks.fulfill_order': {'queue': 'fulfillment'}}
+CELERY_ROUTES = {'ecommerce_worker.fulfillment.v1.tasks.fulfill_order': {'queue': 'fulfillment'},
+                 'ecommerce_worker.sailthru.v1.tasks.update_course_enrollment': {'queue': 'fulfillment'}}
```
rlucioni: Please don't use the fulfillment queue for this. Use a separate named queue (e.g., 'email_marketing'). If you do this, you'll need to update https://github.com/edx/configuration/blob/master/playbooks/roles/ecomworker/templates/edx/app/ecomworker/ecomworker.sh.j2.

ecommerce/settings/base.py
```diff
@@ -479,9 +482,11 @@
 # See http://celery.readthedocs.org/en/latest/configuration.html#celery-imports.
 CELERY_IMPORTS = (
     'ecommerce_worker.fulfillment.v1.tasks',
+    'ecommerce.sailthru.tasks',
```
rlucioni: Account for task versioning here (v1 is missing).

ecommerce/sailthru/signals.py
```diff
@@ -0,0 +1,95 @@
...
+    """Tell Sailthru when payment done.
+
+    Parameters described at http://django-oscar.readthedocs.io/en/releases-1.1/ref/signals.html
+    """
+
+    if not settings.SAILTHRU_ENABLE:
+        return
+
+    # get product (should only be 1)
+    product = order.lines.first().product
```
rlucioni: Is there a need to limit this to one product? Is Sailthru capable of dealing with a mulit-product order?

ecommerce/sailthru/signals.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # return if no price, since enrolls are handled by lms
```
rlucioni: I don't think this comment is very helpful.

ecommerce/sailthru/signals.py
```diff
@@ -0,0 +1,95 @@
...
+    price = order.total_excl_tax
+    if not price or price == 0:
```
rlucioni: More concisely: `if not price`.

ecommerce/sailthru/signals.py
```diff
@@ -0,0 +1,95 @@
...
+                                   site_code=request.site.siteconfiguration.partner.short_code,
+                                   message_id=request.COOKIES.get('sailthru_bid'))
+
+
+def _build_course_url(course_id):
+    """Build a course url from a course id and the host
+
+    :param request:
+    :param course_id:
+    :return:
```
rlucioni: Please keep your docstrings consistent with those used in this repo.

ecommerce/sailthru/signals.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    host = get_lms_url()
+    # hack for integration testing since Sailthru rejects urls without a valid domain
+    if host.startswith('http://127.0.0.1'):
+        host = 'http://courses.edx.org'
```
rlucioni: Please don't do this. This URL should come from settings and will need to change on a per-environment basis. This would be a great place to use edx-platform's configuration models. In their absence, consider https://github.com/lazybird/django-solo.

ecommerce/sailthru/signals.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # pass event to ecommerce_worker.sailthru.v1.tasks to handle asynchronously
+    update_course_enrollment.delay(user.email, course_url, False, product.attr.certificate_type,
+                                   unit_cost=price, course_id=course_id, currency=order.currency,
+                                   site_code=request.site.siteconfiguration.partner.short_code,
+                                   message_id=request.COOKIES.get('sailthru_bid'))
+
+
+@receiver(basket_addition)
+@silence_exceptions("Failed to call Sailthru upon backet addition.")
```
rlucioni: s/backet/basket

ecommerce/sailthru/signals.py
PaulWattenberger: In theory, but it goes beyond the project requirements

ecommerce/settings/base.py
PaulWattenberger: @clintonb Clinton, is django-waffle the standard for these settings in ecommerce?

clintonb: Yes.

rlucioni: http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/ecommerce/additional_features/gate_ecommerce.html

PaulWattenberger: ok, thx

ecommerce/settings/base.py
PaulWattenberger: This doesn't appear to be needed here since it is a worker config option.  The whole CELERY_IMPORTS could probably be removed

rlucioni: Good call, yeah.

PaulWattenberger:
@rlucioni

rlucioni:
@PaulWattenberger you shouldn't merge/deploy these changes until the necessary configuration changes are made. Otherwise, these tasks will pile up in the queue without being processed by any workers.

PaulWattenberger:
@rlucioni I'll leave sailthru_enable off on ecommerce until everything is set up on ecommerce_worker.  Nothing will be sent to the queue

ecommerce/sailthru/signals.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+logger = logging.getLogger(__name__)
+post_checkout = get_class('checkout.signals', 'post_checkout')
+basket_addition = get_class('basket.signals', 'basket_addition')
+
+
+@receiver(post_checkout)
+@silence_exceptions("Failed to call Sailthru upon order completion.")
+def process_checkout_complete(sender, order=None, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
```
rlucioni, line 20: Are order, request, and user really keyword arguments? I don't see anything built in below to handle cases in which they aren't provided. For example, this will blow up if an order is not provided.

ecommerce/sailthru/signals.py
```diff
@@ -0,0 +1,90 @@
...
+        # pass event to ecommerce_worker.sailthru.v1.tasks to handle asynchronously
+        update_course_enrollment.delay(user.email, course_url, False, mode_for_seat(product),
+                                       unit_cost=price, course_id=course_id, currency=order.currency,
+                                       site_code=request.site.siteconfiguration.partner.short_code,
+                                       message_id=request.COOKIES.get('sailthru_bid'))
+
+
+@receiver(basket_addition)
+@silence_exceptions("Failed to call Sailthru upon basket addition.")
+def process_basket_addition(sender, product=None, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
```
rlucioni, line 56: Again, these arguments are treated below as if they were positional. There's no handling for cases in which they aren't provided. Are they really optional?

ecommerce/sailthru/signals.py
```diff
@@ -0,0 +1,90 @@
...
+    """Tell Sailthru when payment started.
```
rlucioni, line 57: The attached signal doesn't necessarily indicate payment has begun, just that a product has been added to a basket.

PaulWattenberger, line 57: That's the goal, added to shopping cart

rlucioni:
@PaulWattenberger leaving the switch off sounds good. This looks fine, I'd just like to get my questions around optional arguments sorted out.

ecommerce/sailthru/signals.py
PaulWattenberger, line 20: I can test for that, but other part of ecommerce make the assumption, e.g. https://github.com/edx/ecommerce/blob/163fb716ec5148ce39ccfeae03db7776974f7ea7/ecommerce/extensions/fulfillment/signals.py

rlucioni, line 20: Hm, I wonder why that's the case. Regardless, the `silence_exceptions` decorator should mitigate the effects if one of those kwargs is missing.

rlucioni:
👍 once commits squashed.

PaulWattenberger:
@rlucioni thanks!

PR #858, opened 2016-07-13
Added view exposing feature flags

clintonb:
ECOM-4854

### DIFF
--- a/ecommerce/features/tests.py
+++ b/ecommerce/features/tests.py
@@ -1,0 +1,28 @@
+from __future__ import unicode_literals
+
+import json
+
+from django.conf import settings
+from django.test import TestCase
+from django.test.utils import override_settings
+from rest_framework.reverse import reverse
+from waffle.models import Flag
+
+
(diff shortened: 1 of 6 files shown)


clintonb:
@bderusha @AlasdairSwan @Dillon-Dumesnil please review

bderusha:
The code looks fine to me.  But I am wondering why it lives in ecommerce.  Wouldn't it make more sense for this to be controlled by the Catalog service?

clintonb:
It lives here because this is the only IDA actively using Waffle. Additionally, I don't want to deal with the overhead of putting this in LMS.

Dillon-Dumesnil:
Looks good minus failing the travis builds

clintonb:
Related: https://github.com/edx/configuration/pull/3247

bderusha:
👍

clintonb:
Drat! Thanks for the review @bderusha; but, we've decided it's probably not a good idea to make a request to Otto for nearly every Drupal page load.

PR #859, opened 2016-07-14
[WIP] [SOL-1938]

vkaracic:
Enrollment code settings set in a white label site are re-used in the edX site which created enrollment codes for courses that it should not. 
This PR contains code that adds a new site configuration flag for enabling creating enrollment codes for toggling the enrollment codes for specific sites, and a checkbox in the CAT for toggling it for individual courses. 
Because the checkbox needs to be disabled for when creating enrollment codes is disabled, a new API endpoint for site configuration was implemented so that the Backbone application can get that information.

https://openedx.atlassian.net/browse/SOL-1938
- [ ] adjust documentation: https://github.com/edx/edx-documentation/pull/1125

Creating enabled:
screenshot from 2016-07-18 14-41-11

Creating disabled:
screenshot from 2016-07-18 14-41-44

### DIFF
--- a/ecommerce/core/management/commands/create_or_update_site.py
+++ b/ecommerce/core/management/commands/create_or_update_site.py
@@ -88,6 +88,12 @@ class Command(BaseCommand):
                             type=str,
                             required=True,
                             help='from email')
+        parser.add_argument('--enable-enrollment-codes',
+                            action='store',
+                            dest='enable_enrollment_codes',
+                            type=bool,
+                            required=False,
+                            help='Enable the creation of enrollment codes.')
 
     def handle(self, *args, **options):
         site_id = options.get('site_id')
@@ -100,6 +106,7 @@ class Command(BaseCommand):
         client_secret = options.get('client_secret')
         segment_key = options.get('segment_key')
         from_email = options.get('from_email')
+        enable_enrollment_codes = True if options.get('enable_enrollment_codes') else False
 
         try:
             site = Site.objects.get(id=site_id)
@@ -120,7 +127,7 @@ class Command(BaseCommand):
             partner.save()
             logger.info('Partner created with code %s', partner_code)
 
-        _, _ = SiteConfiguration.objects.get_or_create(
+        SiteConfiguration.objects.update_or_create(
             site=site,
             defaults={
                 'partner': partner,
@@ -129,6 +136,7 @@ class Command(BaseCommand):
                 'payment_processors': options['payment_processors'],
                 'segment_key': segment_key,
                 'from_email': from_email,
+                'enable_enrollment_codes': enable_enrollment_codes,
                 'oauth_settings': {
                     'SOCIAL_AUTH_EDX_OIDC_URL_ROOT': '{lms_url_root}/oauth2'.format(lms_url_root=lms_url_root),
                     'SOCIAL_AUTH_EDX_OIDC_KEY': client_id,
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -119,10 +119,33 @@ class Course(models.Model):
 
         return name
 
-    def create_or_update_seat(self, certificate_type, id_verification_required, price, partner,
-                              credit_provider=None, expires=None, credit_hours=None, remove_stale_modes=True):
+    def create_or_update_seat(
+            self,
+            certificate_type,
+            id_verification_required,
+            price,
+            partner,
+            credit_provider=None,
+            expires=None,
+            credit_hours=None,
+            remove_stale_modes=True,
+            create_enrollment_code=False
+    ):
         """
         Creates course seat products.
+
+        Arguments:
+            certificate_type(str): The seat type.
+            id_verification_required(bool): Whether an ID verification is required.
+            price(int): Price of the seat.
+            partner(Partner): Site partner.
+
+        Optional arguments:
+            credit_provider(str): Name of the organization that provides the credit
+            expires(Datetime): Date when the seat type expires.
+            credit_hours(int): Number of credit hours provided.
+            remove_stale_modes(bool): Remove stale modes.
+            create_enrollment_code(bool): Whether an enrollment code is created in additon to the seat.
 
         Returns:
             Product:  The seat that has been created or updated.
@@ -188,8 +211,9 @@ class Course(models.Model):
         seat.attr.certificate_type = certificate_type
         seat.attr.course_key = course_id
         seat.attr.id_verification_required = id_verification_required
-
-        if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and certificate_type in ENROLLMENT_CODE_SEAT_TYPES:
+        if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and \
+                certificate_type in ENROLLMENT_CODE_SEAT_TYPES and \
+                create_enrollment_code:
             self._create_or_update_enrollment_code(certificate_type, id_verification_required, partner, price)
 
         if credit_provider:
@@ -261,7 +285,6 @@ class Course(models.Model):
         Returns:
             Enrollment code product.
         """
-
         enrollment_code_product_class = ProductClass.objects.get(name=ENROLLMENT_CODE_PRODUCT_CLASS_NAME)
         enrollment_code = self.enrollment_code_product
         if not enrollment_code:
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -14,6 +14,7 @@ from rest_framework.reverse import reverse
 import waffle
 
 from ecommerce.core.constants import ISO_8601_FORMAT, COURSE_ID_REGEX
+from ecommerce.core.models import Site, SiteConfiguration
 from ecommerce.core.url_utils import get_ecommerce_url
 from ecommerce.courses.models import Course
 from ecommerce.coupons.utils import get_seats_from_query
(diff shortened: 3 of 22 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=859

vkaracic:
@mattdrayer this is the PR for enrollment code checkbox. It should appear at the bottom of the form for creating a new seat for a course. I haven't added a disable for when the flag for enrollment codes is off, yet.

ecommerce/courses/models.py
```diff
@@ -120,7 +120,8 @@ def get_course_seat_name(self, certificate_type, id_verification_required):
         return name
 
     def create_or_update_seat(self, certificate_type, id_verification_required, price, partner,
-                              credit_provider=None, expires=None, credit_hours=None, remove_stale_modes=True):
+                              credit_provider=None, expires=None, credit_hours=None,
+                              remove_stale_modes=True, bulk_enrollment=False):
```
mattdrayer: Please add argument definitions in this method's docstring

mattdrayer: Would it be better to name this argument something like "create_enrollment_code", instead of "bulk_enrollment"?

ecommerce/courses/models.py
```diff
@@ -188,8 +189,9 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
         seat.attr.certificate_type = certificate_type
         seat.attr.course_key = course_id
         seat.attr.id_verification_required = id_verification_required
-
-        if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and certificate_type in ENROLLMENT_CODE_SEAT_TYPES:
+        if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and \
+                certificate_type in ENROLLMENT_CODE_SEAT_TYPES and \
+                bulk_enrollment:
```
mattdrayer: As noted above, I think this argument might be better named "create_enrollment_code"

ecommerce/extensions/api/serializers.py
```diff
@@ -275,6 +275,7 @@ def save(self):
         course_name = self.validated_data['name']
         course_verification_deadline = self.validated_data.get('verification_deadline')
         products = self.validated_data['products']
+
```
mattdrayer: Why the extra line here?

ecommerce/extensions/api/serializers.py
```diff
@@ -299,6 +300,7 @@ def save(self):
 
                     # Extract arguments required for Seat creation, deserializing as necessary.
                     certificate_type = attrs.get('certificate_type', '')
+                    bulk_enrollment = product['course'].get('bulk_enrollment')
```
mattdrayer: 'create_enrollment_code'

ecommerce/extensions/api/serializers.py
```diff
@@ -317,6 +319,7 @@ def save(self):
                         expires=expires,
                         credit_provider=credit_provider,
                         credit_hours=credit_hours,
+                        bulk_enrollment=bulk_enrollment
```
mattdrayer: 'create_enrollment_code'

ecommerce/extensions/api/v2/tests/views/test_publication.py
```diff
@@ -41,7 +41,15 @@ def setUp(self):
                             'name': 'id_verification_required',
                             'value': False
                         }
-                    ]
+                    ],
+                    'course': {
+                        'bulk_enrollment': 'true',
```
mattdrayer: 'create_enrollment_code'

mattdrayer:
@vkaracic -- I took a first pass through -- overall this appears to be in-line with what has been discussed so far in terms of the solution.  We do need to consider the site-level feature flag as well, but I know you are aware of that so I'll just keep an eye on the PR for when that comes along.

ecommerce/courses/models.py
```diff
@@ -188,8 +189,9 @@ def create_or_update_seat(self, certificate_type, id_verification_required, pric
         seat.attr.certificate_type = certificate_type
         seat.attr.course_key = course_id
         seat.attr.id_verification_required = id_verification_required
-
-        if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and certificate_type in ENROLLMENT_CODE_SEAT_TYPES:
+        if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and \
```
mattdrayer, line 214: Do we need the site flag here?

vkaracic, line 214: Yes, but I think this resolves it: https://github.com/edx/ecommerce/pull/859/files#diff-852a6b45e4fa1255461cf3fa6e5ceda7R303

ecommerce/core/management/commands/create_or_update_site.py
```diff
@@ -100,6 +106,7 @@ def handle(self, *args, **options):
         client_secret = options.get('client_secret')
         segment_key = options.get('segment_key')
         from_email = options.get('from_email')
+        enable_enrollment_codes = options.get('enable_enrollment_codes') is not None
```
mattdrayer: Good chance that I'm not following this line correctly, but it seems to be that if options.get('enable_enrollment_codes') returns False, then the "is not None" check will return a True, which will be used for the value of enable_enrollment_codes?

mattdrayer:
LGTM @vkaracic!  I see there is one spot in the JS in need of test coverage and that should put you over the threshold -- nice work with a fast turnaround -- 👍

ecommerce/core/management/commands/create_or_update_site.py
vkaracic: You're right, that is a mistake! Here I have to make sure that the model gets either a `False` or a `True`, and not a `None`, since `None != False`.

ecommerce/courses/models.py
```diff
@@ -120,10 +120,21 @@ def get_course_seat_name(self, certificate_type, id_verification_required):
         return name
 
     def create_or_update_seat(self, certificate_type, id_verification_required, price, partner,
-                              credit_provider=None, expires=None, credit_hours=None, remove_stale_modes=True):
+                              credit_provider=None, expires=None, credit_hours=None,
+                              remove_stale_modes=True, create_enrollment_code=False):
```
marjev: nitpicks: list the parameters one in each line and alphabetize. That way it will be easier to preview parameters.

ecommerce/courses/models.py
```diff
@@ -120,10 +120,21 @@ def get_course_seat_name(self, certificate_type, id_verification_required):
...
+        Arguments:
+            certificate_type(str): The seat type.
+            id_verification_required(bool): Whether an ID verification is required.
+            price(int): Price of the seat.
+            partner(Partner): Site partner.
+            credit_provider(str): Name of the organization that provides the credit
+            expires(Datetime): Date when the seat type expires.
+            credit_hours(int): Number of credit hours provided.
+            remove_stale_modes(bool): Remove stale modes.
+            create_enrollment_code(bool): Whether an enrollment code is created in additon to the seat.
```
marjev, line 148: nitpick: alphabetize

ecommerce/extensions/api/v2/urls.py
```diff
@@ -8,7 +8,8 @@
                                                products as product_views, courses as course_views,
                                                publication as publication_views, partners as partner_views,
                                                catalog as catalog_views, stockrecords as stockrecords_views,
-                                               coupons as coupon_views, vouchers as voucher_views)
+                                               coupons as coupon_views, vouchers as voucher_views,
+                                               siteconfiguration as siteconfiguration_views)
```
marjev: nitpick: alphabetize or since this grows, perhaps we should import `from ecommerce.extensions.api.v2 import views` and then reference the views like `views.baskets` etc. But, it's up to you.

ecommerce/extensions/api/v2/views/siteconfiguration.py
```diff
@@ -0,0 +1,12 @@
+"""API endpoint for site configuration."""
+from rest_framework import viewsets
+from rest_framework.permissions import IsAuthenticated, IsAdminUser
```
marjev: nitpick: alphabetize the right side.

marjev:
Regarding these flags, what's the default value and is it set for development environments. We have a lot of configuration steps and I think that we should strive to allow an effortless installation of our application, at least for development purposes. I'm not sure if you should provide some value in `devstack.py` settings.

marjev:
besides nitpicks, it LGTM 👍

ecommerce/courses/models.py
vkaracic: I've moved each parameter into a new line, but this can't be alphabetized for several reasons. `self` has to be the first parameter, you can't mix parameters with keyword parameters, and if I alphabetize here I'd need to through every call to this method in the codebase and adjust the order of arguments there too.

ecommerce/extensions/api/v2/urls.py
vkaracic: I alphabetized this and moved each module to a new line. Changing to `import views` and then using `views.baskets` for example, gives out import errors for some reason.

ecommerce/static/templates/course_form.html
```diff
@@ -68,11 +68,21 @@
...
+                <%- gettext('Include an Honor Seat with this course') %>
+            </span>
+        </div>
+    </div>
+</div>
+
+<div class="fields">
+    <div class="form-group create-enrollment-code hidden checkbox">
+        <label for="bulkEnrollment">
+        <input type="checkbox" id="bulkEnrollment" name="create_enrollment_code" value="true"> <%- gettext('Include Enrollment Coupon') %>
```
mattdrayer: Should this be "Include Enrollment Code" ?

mattdrayer: Maybe even "Include ENROLLMENT_CODE_PRODUCT_CLASS_NAME"

vkaracic: Updated! I'd rather not include a constant for consistency sake since constants aren't included anywhere here.

mattdrayer:
Cool, I'm good w/ this PR now -- 👍 :shipit:

marjev:
LGTM 👍

vkaracic:
Thanks for the reviews, gents!

PR #860, opened 2016-07-18
End-to-end test causes Jenkins job to hang

ahsan-ul-haq:
ECOM-4823

### DIFF
--- a/acceptance_tests/mixins.py
+++ b/acceptance_tests/mixins.py
@@ -79,8 +79,7 @@ class LogistrationMixin(LmsUserMixin):
         email = email or LMS_EMAIL
         password = password or LMS_PASSWORD
 
-        # Note: We use Selenium directly here (as opposed to bok-choy) to avoid issues with promises being broken.
-        self.lms_login_page.browser.get(self.lms_login_page.url(course_id))  # pylint: disable=not-callable
+        self.lms_login_page.visit(course_id)
         self.lms_login_page.login(email, password)
 
     def register_via_ui(self, course_id=None):
@@ -110,8 +109,7 @@ def login_with_lms(self, email=None, password=None, course_id=None):
     email = email or LMS_EMAIL
     password = password or LMS_PASSWORD
 
-    # Note: We use Selenium directly here (as opposed to bok-choy) to avoid issues with promises being broken.
-    self.lms_login_page.browser.get(self.lms_login_page.url(course_id))  # pylint: disable=not-callable
+    self.lms_login_page.visit(course_id)
     self.lms_login_page.login(email, password)
 
 
@@ -124,6 +122,11 @@ class OttoAuthenticationMixin(object):
         """ Start the login process via Otto's login page (which should redirect to LMS.) """
         email = email or LMS_EMAIL
         password = password or LMS_PASSWORD
+
+        # Note: We have to visit LMS Login page to provide basic auth as otto login page redirect to LMS login page
+        # which requires basic auth and not providing that could cause the test to hang.
+        lms_login_page = LMSLoginPage(self.browser)
+        lms_login_page.visit()
 
         self.otto_login_page.visit()
         submit_lms_login_form(self.otto_login_page, email, password)
--- a/acceptance_tests/pages/ecommerce.py
+++ b/acceptance_tests/pages/ecommerce.py
@@ -36,6 +36,5 @@ class EcommerceLoginPage(LMSLoginPage):
     that page and it's properties.
     """
 
-    @property
-    def url(self):  # pylint: disable=arguments-differ
+    def url(self, course_id=None):  # pylint: disable=arguments-differ,unused-argument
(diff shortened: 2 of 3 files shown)


ahsan-ul-haq:
@schenedx please review.

schenedx:
@clintonb Please help review.
@ahsan-ul-haq Can you explain why we need to log into LMS first to log into Otto? What is the actual root cause of the test hang?

ahsan-ul-haq:
@schenedx I am not actually logging into LMS I am just visiting the page to provide basic auth. The actual root cause of test hang is that when we click on otto login button it redirect to LMS login page and it require basic auth to provide. So if we just visit page in start of test and provide basic auth then we can login from otto and it will not hang the test.

schenedx:
@ahsan-ul-haq If that's the root cause is that when otto redirect to lms, the basic auth window shows, why can't we implement the handle of basic auth in the `self. login_via_otto()` function? By visit LMS page first looked like a hack to me.

schenedx:
@Dillon-Dumesnil You maybe interested in this PR as well

acceptance_tests/test_auth.py
```diff
@@ -46,6 +47,8 @@ def test_lms_logout(self):
 
     def test_login_redirection(self):
         """ Verify the user is redirected to the Otto dashboard after logging in. """
+        lms_login_page = LMSLoginPage(self.browser)
+        lms_login_page.browser.get(lms_login_page.url(None))  # pylint: disable=not-callable
```
clintonb: `lms_login_page.visit()`

acceptance_tests/test_auth.py
```diff
@@ -46,6 +47,8 @@ def test_lms_logout(self):
 
     def test_login_redirection(self):
         """ Verify the user is redirected to the Otto dashboard after logging in. """
+        lms_login_page = LMSLoginPage(self.browser)
```
clintonb: Add a comment explaining why this is necessary.

acceptance_tests/mixins.py
```diff
@@ -125,6 +125,12 @@ def login_via_otto(self, email=None, password=None):
         email = email or LMS_EMAIL
         password = password or LMS_PASSWORD
 
+        # Note: We have to visit LMS Login page to provide basic auth as otto login page redirect to LMS login page
+        # which requires basic auth and not providing that could cause the test to hang.
+        lms_login_page = LMSLoginPage(self.browser)
+        # Note: We use Selenium directly here (as opposed to bok-choy) to avoid issues with promises being broken.
+        lms_login_page.browser.get(lms_login_page.url(None))  # pylint: disable=not-callable
```
ahsan-ul-haq: @clintonb we can't use `lms_login_page.visit()` here as `visit` expect the url to be a string or property but in this case url is a function and using visit would cause `'function' object has no attribute 'find'` error.

clintonb: In that case, please update the comment about using Selenium to be relevant. I wrote it about a year ago, and it makes absolutely no sense. Alternatively you could override the `visit` method of `LMSLoginPage` to encapsulate this logic. It doesn't make much sense to copy this logic in three different locations.

ahsan-ul-haq:
@clintonb please review

acceptance_tests/mixins.py
```diff
@@ -110,8 +109,7 @@ def login_with_lms(self, email=None, password=None, course_id=None):
     email = email or LMS_EMAIL
     password = password or LMS_PASSWORD
 
-    # Note: We use Selenium directly here (as opposed to bok-choy) to avoid issues with promises being broken.
-    self.lms_login_page.browser.get(self.lms_login_page.url(course_id))  # pylint: disable=not-callable
+    self.lms_login_page.visit(course_id)  # pylint: disable=not-callable
```
schenedx: Are you sure we still need the `pylint` comment here?

acceptance_tests/pages/ecommerce.py
```diff
@@ -36,6 +36,5 @@ class EcommerceLoginPage(LMSLoginPage):
     that page and it's properties.
     """
 
-    @property
-    def url(self):  # pylint: disable=arguments-differ
+    def url(self, course_id=None):  # pylint: disable=arguments-differ,unused-argument
```
schenedx, line 39: Why do we need a course_id passed in at all?

schenedx:
Minor comments, rest LGTM

acceptance_tests/pages/ecommerce.py
ahsan-ul-haq, line 39: As this page is being inherited from `LMSLoginPage` so instead of default `visit` method, the method that we override is being called which would expect a `course_id` in url method. So either we can do this or check the instance in `visit` method and call appropriate `url` method accordingly.

schenedx, line 39: Got it. Thanks for the explanation.

clintonb:
👍 once pylint comment addressed

PR #861, opened 2016-07-18
[SOL-1935] Remove voucher listing from detail page as it takes too lo…

iivic:
Details page was slow if huge amount of vouchers created. Coupon api endpoint changed and vouchers removed from details page. @marjev @vkaracic

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -6,6 +6,7 @@ import logging
 
 from dateutil.parser import parse
 from django.db import transaction
+from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
 from django.contrib.auth import get_user_model
 from oscar.core.loading import get_model, get_class
@@ -429,32 +430,122 @@ class CouponListSerializer(serializers.ModelSerializer):
 
 class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     """ Serializer for Coupons. """
+    benefit_type = serializers.SerializerMethodField()
+    benefit_value = serializers.SerializerMethodField()
     coupon_type = serializers.SerializerMethodField()
-    last_edited = serializers.SerializerMethodField()
-    seats = serializers.SerializerMethodField()
-    client = serializers.SerializerMethodField()
-    vouchers = serializers.SerializerMethodField()
-    categories = CategorySerializer(many=True, read_only=True)
-    note = serializers.SerializerMethodField()
-    max_uses = serializers.SerializerMethodField()
-    num_uses = serializers.SerializerMethodField()
     catalog_query = serializers.SerializerMethodField()
+    categories = CategorySerializer(many=True, read_only=True)
+    client = serializers.SerializerMethodField()
+    code = serializers.SerializerMethodField()
+    code_status = serializers.SerializerMethodField()
     course_seat_types = serializers.SerializerMethodField()
+    end_date = serializers.SerializerMethodField()
+    last_edited = serializers.SerializerMethodField()
+    max_uses = serializers.SerializerMethodField()
+    note = serializers.SerializerMethodField()
+    num_uses = serializers.SerializerMethodField()
     payment_information = serializers.SerializerMethodField()
+    quantity = serializers.SerializerMethodField()
+    seats = serializers.SerializerMethodField()
+    start_date = serializers.SerializerMethodField()
+    voucher_type = serializers.SerializerMethodField()
+
+    def retrieve_benefit(self, obj):
+        """Helper method to retrieve the benefit from voucher. """
+        return self.retrieve_voucher(obj).benefit
+
+    def retrieve_end_date(self, obj):
+        """Helper method to retrieve the voucher end datetime. """
+        return self.retrieve_voucher(obj).end_datetime
 
     def retrieve_offer(self, obj):
         """Helper method to retrieve the offer from coupon. """
-        return obj.attr.coupon_vouchers.vouchers.first().offers.first()
+        return self.retrieve_voucher(obj).offers.first()
+
+    def retrieve_start_date(self, obj):
+        """Helper method to retrieve the voucher start datetime. """
+        return self.retrieve_voucher(obj).start_datetime
+
+    def retrieve_voucher(self, obj):
+        """Helper method to retrieve the first voucher from coupon. """
+        return obj.attr.coupon_vouchers.vouchers.first()
+
+    def retrieve_voucher_usage(self, obj):
+        """Helper method to retrieve usage from voucher. """
+        return self.retrieve_voucher(obj).usage
+
+    def retrieve_quantity(self, obj):
+        """Helper method to retrieve number from vouchers. """
+        return obj.attr.coupon_vouchers.vouchers.count()
+
+    def get_benefit_type(self, obj):
+        return self.retrieve_benefit(obj).type
+
+    def get_benefit_value(self, obj):
+        return self.retrieve_benefit(obj).value
 
     def get_coupon_type(self, obj):
+        benefit = self.retrieve_benefit(obj)
+        if benefit.type == Benefit.PERCENTAGE and benefit.value == 100:
+            return _('Enrollment code')
+        return _('Discount code')
+
+    def get_catalog_query(self, obj):
         offer = self.retrieve_offer(obj)
-        if offer.benefit.type == Benefit.PERCENTAGE and offer.benefit.value == 100:
-            return 'Enrollment code'
-        return 'Discount code'
+        return offer.condition.range.catalog_query
+
+    def get_client(self, obj):
+        return Invoice.objects.get(order__basket__lines__product=obj).business_client.name
+
+    def get_code(self, obj):
+        if self.retrieve_quantity(obj) == 1:
+            return self.retrieve_voucher(obj).code
+
+    def get_code_status(self, obj):
+        start_date = self.retrieve_start_date(obj)
+        end_date = self.retrieve_end_date(obj)
+        current_datetime = timezone.now()
+        in_time_interval = (start_date < current_datetime) and (end_date > current_datetime)
+        return _('ACTIVE') if in_time_interval else _('INACTIVE')
+
+    def get_course_seat_types(self, obj):
+        offer = self.retrieve_offer(obj)
+        course_seat_types = offer.condition.range.course_seat_types
+        return course_seat_types.split(',') if course_seat_types else course_seat_types
+
+    def get_end_date(self, obj):
+        return self.retrieve_end_date(obj)
 
     def get_last_edited(self, obj):
         history = obj.history.latest()
         return history.history_user.username, history.history_date
+
+    def get_max_uses(self, obj):
+        offer = self.retrieve_offer(obj)
+        return offer.max_global_applications
+
+    def get_note(self, obj):
+        try:
+            return obj.attr.note
+        except AttributeError:
+            return None
+
+    def get_num_uses(self, obj):
+        offer = self.retrieve_offer(obj)
+        return offer.num_applications
+
+    def get_payment_information(self, obj):
+        """
+        Retrieve the payment information.
+        Currently only invoices are supported, in the event of adding another
+        payment processor append it to the response dictionary.
+        """
+        invoice = Invoice.objects.filter(order__basket__lines__product=obj).first()
+        response = {'Invoice': InvoiceSerializer(invoice).data}
+        return response
+
+    def get_quantity(self, obj):
+        return self.retrieve_quantity(obj)
 
     def get_seats(self, obj):
         offer = self.retrieve_offer(obj)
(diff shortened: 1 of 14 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=861&repo=edx%2Fecommerce

iivic:
This is how the details page looks like now. Notice the download button.
screen shot 2016-07-18 at 12 15 01

ecommerce/extensions/api/serializers.py
```diff
@@ -509,12 +540,53 @@ def get_payment_information(self, obj):
...
+                _range.course_seat_types
+            )
+        serializer = ProductSerializer(seats, many=True, context={'request': request})
+        return serializer.data
+
+    def get_start_date(self, obj):
+        return self.retrieve_start_date(obj)
+
+    def get_voucher_code(self, obj):
+        return self.retrieve_voucher(obj).code
```
vkaracic: Why would you want to retrieve one code?

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -318,6 +318,35 @@ def get_response_json(self, method, path, data=None):
...
 
+    def test_create_serializer_data(self):
+        """Test if coupon serializer creates data for details page"""
+        details_response = self.get_response_json(
+            'GET',
+            reverse('api:v2:coupons-detail', args=[self.coupon.id]),
+            data=self.data
+        )
+
+        self.assertEqual(details_response['coupon_type'], 'Enrollment code')
```
vkaracic, line 328: Have you tried comparing the response to `CouponSerializer(self.coupon).data` ?

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -318,6 +318,35 @@ def get_response_json(self, method, path, data=None):
...
+        )
+        self.assertEqual(details_response['usage_limitation'], 'Can be used once by multiple customers')
+
+        self.coupon.attr.coupon_vouchers.vouchers.all().update(usage=Voucher.MULTI_USE)
+        details_response = self.get_response_json(
+            'GET',
+            reverse('api:v2:coupons-detail', args=[self.coupon.id]),
+            data=self.data
+        )
+        self.assertEqual(details_response['usage_limitation'], 'Can be used multiple times by multiple customers')
```
vkaracic: You can factor this test out in another one and use `ddt` for the repetitions.

ecommerce/static/js/test/mock_data/coupons.js
```diff
@@ -161,7 +161,11 @@ define([], function(){
                 'name': 'TESTCAT'
             }
         ],
-        'vouchers': [percentageDiscountCodeVoucher]
+        'start_date': '2015-01-01T00:00:00Z',
+        'end_date': '2016-01-01T00:00:00Z',
+        'voucher_type': 'Single use',
+        benefit_type: 'Percentage',
```
vkaracic: Why are there no quotes around `benefit_type` and `benefit_value`?

ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
```diff
@@ -77,17 +65,8 @@ define([
...
-
-                valueDiscountCodeVoucher.usage = 'Multi-use';
-                expect(view.usageLimitation(valueDiscountCodeVoucher)).toBe(
-                    'Can be used multiple times by multiple customers'
-                );
-
-                valueDiscountCodeVoucher.usage = '';
-                expect(view.usageLimitation(valueDiscountCodeVoucher)).toBe('');
+            it('should get usage limitation from coupon model', function () {
+                expect(view.model.get('usage_limitation')).toBe(model.get('usage_limitation'));
```
vkaracic: Would it be easier to compare the two models right away, instead of in separate tests?

ecommerce/extensions/api/serializers.py
iivic: This is not needed I forgot to delete it.

vkaracic:
I have completed my first pass.

ecommerce/static/js/test/mock_data/coupons.js
iivic: Copy/Paste at its finest. I already fixed it.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
iivic, line 328: Yes. It doesn't work that way.

ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
iivic: 👍

ecommerce/extensions/api/serializers.py
marjev: We need this method for custom codes, you're just missing an `if` statement.

iivic: You are right we do need this method but it needs to be called get_code otherwise it will have no effect.

ecommerce/extensions/api/serializers.py
```diff
@@ -425,80 +426,110 @@ class Meta(object):
...
-        if offer.benefit.type == Benefit.PERCENTAGE and offer.benefit.value == 100:
-            return 'Enrollment code'
-        return 'Discount code'
+        course_seat_types = offer.condition.range.course_seat_types
+        return course_seat_types.split(',') if course_seat_types else course_seat_types
+
+    def get_discount_value(self, obj):
+        benefit = self.retrieve_benefit(obj)
+        benefit_type = benefit.type
+        benefit_value = str(benefit.value).rstrip('0').rstrip('.')
```
marjev: Don't do this here. API should present the data as it is. Front end should do all the necessary formatting. Alternatively, if you insist on doing the formatting here, don't do it like this.
You're changing the type as well as hacking into the value. Check `normalize()` or some similar solution.

ecommerce/extensions/api/serializers.py
```diff
@@ -509,12 +540,53 @@ def get_payment_information(self, obj):
...
+
+    def get_usage_limitation(self, obj):
+        voucher_usage = self.retrieve_voucher_usage(obj)
+        if voucher_usage == 'Single use':
+            return _('Can be used once by one customer')
+        elif voucher_usage == 'Multi-use':
+            return _('Can be used multiple times by multiple customers')
+        elif voucher_usage == 'Once per customer':
+            return _('Can be used once by multiple customers')
+        return ''
```
marjev: Do this mapping in BackboneJS, too. It's easier to work with values like `Single use` than it is with values like `Can be used once by one customer`. You should add a mapping for this in Backbone and show appropriate message.

marjev:
Finished my first pass.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
iivic: Thank you for the info. I will move this data to backbone so that the test is not needed here.

ecommerce/extensions/api/serializers.py
iivic: 👍

ecommerce/extensions/api/serializers.py
iivic: 👍

iivic:
Can you check it one more time?

vkaracic:
LGTM

mjfrey:
@iivic I would like to see the approach for the download report optimization before we merge this.

iivic:
@mjfrey  I'm looking into it. Once I figure out how to improve the download speed I'll let you know so we can discuss it. This branch will not be merged until the optimization is done.

iivic:
Guys review the report download improvements when you get the time.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -57,7 +57,7 @@ def setUp(self):
         self.catalog.stock_records.add(self.stock_record)
 
         self.coupon = self.create_coupon(
-            title='Test product',
+            title='Tešt product'.decode('utf-8'),
```
marjev: Why are you decoding only this value here? Seems unnecessary.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -263,8 +263,14 @@ def test_create_discount_coupon(self):
...
             )
 
-    def assert_report_row(self, row, coupon, voucher):
-        """ Verify that the row fields contain the right data. """
+    def assert_report_first_row(self, row, coupon, voucher):
+        """
+        Verify that the first row fields contain the right data.
+        Args:
+            row (list): First row in report
+            coupon: self.coupon
```
marjev: coupon (Product): Coupon for which the report is generate.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -279,16 +285,29 @@ def assert_report_row(self, row, coupon, voucher):
...
-            get_ecommerce_url() + self.REDEMPTION_URL.format(voucher.code)
-        )
         self.assertEqual(row['Note'], coupon.attr.note)
         self.assertEqual(row['Created By'].decode('utf-8'), coupon.history.first().history_user.full_name)
         self.assertEqual(row['Create Date'], coupon.history.latest().history_date.strftime("%b %d, %y"))
         self.assertEqual(row['Coupon Start Date'], voucher.start_datetime.strftime("%b %d, %y"))
         self.assertEqual(row['Coupon Expiry Date'], voucher.end_datetime.strftime("%b %d, %y"))
 
+    def assert_report_row(self, row, voucher):
+        """ Verify that the row fields contain the right data. """
```
marjev: To keep things consistent, write the docstring the same way you did for the method above.

ecommerce/extensions/voucher/tests/test_utils.py
iivic: For some reason this value couldn't be processed hence the tests failed. Decoding and encoding it fixed the issue.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -263,8 +263,14 @@ def test_create_discount_coupon(self):
...
 
-    def assert_report_row(self, row, coupon, voucher):
-        """ Verify that the row fields contain the right data. """
+    def assert_report_first_row(self, row, coupon, voucher):
+        """
+        Verify that the first row fields contain the right data.
+        Args:
+            row (list): First row in report
+            coupon: self.coupon
+            voucher (Voucher): voucher got through Coupon Name
```
marjev: Voucher associated with the Coupon

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -331,34 +350,38 @@ def test_generate_coupon_report(self):
             'Coupon Expiry Date',
         ])
 
+        voucher = Voucher.objects.get(name=rows[0]['Coupon Name'])
+        self.assert_report_first_row(rows.pop(0), self.coupon, voucher)
```
marjev: This will always pass. Get `voucher` from `vouchers` variable found on line 318.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -331,34 +350,38 @@ def test_generate_coupon_report(self):
...
             voucher_type=Voucher.SINGLE_USE
         )
 
         __, rows = generate_coupon_report(self.coupon_vouchers)
 
-        inactive_coupon_row = rows[1]
-        self.assertEqual(inactive_coupon_row['Coupon Name'], coupon_name)
-        self.assertEqual(inactive_coupon_row['Status'], _('Inactive'))
+        # The data in row[0] is the same for all rows
+        # The data from row[1] is unique for all vouchers
```
vkaracic: This is contradictory. How is the data from the first row the same for every other row? Is the first row just copy pasted a number of times?

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -331,34 +350,38 @@ def test_generate_coupon_report(self):
...
+        self.assertEqual(rows[0]['Coupon Name'], coupon_title.encode('utf-8'))
```
marjev: Unnecessary encoding.

ecommerce/extensions/voucher/utils.py
```diff
@@ -77,44 +77,66 @@ def _get_discount_info(discount_data):
 
 
 def _get_info_for_coupon_report(coupon, voucher):
-    offer = voucher.offers.all().first()
+    history = coupon.history.first()
+    author = history.history_user.full_name.encode('utf8')
```
marjev: Why are you encoding this value?

ecommerce/extensions/voucher/utils.py
```diff
@@ -77,44 +77,66 @@ def _get_discount_info(discount_data):
...
+    category_name = ProductCategory.objects.get(product=coupon).category.name
+
+    try:
+        note = coupon.attr.note.encode('utf8')
```
marjev: Same question as above?

ecommerce/extensions/voucher/utils.py
```diff
@@ -204,24 +207,26 @@ def generate_coupon_report(coupon_vouchers):
     for coupon_voucher in coupon_vouchers:
         coupon = coupon_voucher.coupon
         client = Invoice.objects.get(order__basket__lines__product=coupon).business_client.name
+        rows.append(_get_info_for_coupon_report(coupon, coupon_voucher.vouchers.first()))
+        rows[0]['Client'] = client.encode('utf8')
```
marjev: Same as above?

ecommerce/extensions/voucher/tests/test_utils.py
marjev: Use `from __future__ import unicode_literals` should fix this. It's bad to do this everywhere in the code where we're using the string literals.

ecommerce/extensions/voucher/utils.py
```diff
@@ -77,44 +77,66 @@ def _get_discount_info(discount_data):
...
-        seat_stockrecord = None
         course_organization = None
         price = None
-        discount_data = None
+        coupon_type, discount_percentage, discount_amount = None, None, None
 
-    history = coupon.history.first()
-    coupon_type, discount_percentage, discount_amount = _get_discount_info(discount_data)
+    coupon_data = {
+        'Code': 'This row applies to all vouchers',
```
vkaracic, line 110: This is an interesting code :smile:

ecommerce/extensions/voucher/tests/test_utils.py
iivic: This doesn't work.

ecommerce/extensions/api/serializers.py
```diff
@@ -513,12 +544,39 @@ def get_payment_information(self, obj):
...
+
+    def get_seats(self, obj):
+        offer = self.retrieve_offer(obj)
+        _range = offer.condition.range
+        request = self.context['request']
+        if _range.catalog:
+            stockrecords = _range.catalog.stock_records.all()
+            seats = Product.objects.filter(id__in=[sr.product.id for sr in stockrecords])
+        else:
+            seats = get_seats_from_query(
```
mjfrey, line 558: We should not show the seats for dynamic coupons.  There could be too many.  We should only show the course seat if the coupon was redeemed which we get from the order.

ecommerce/extensions/voucher/tests/test_utils.py
iivic: 👍

ecommerce/extensions/voucher/tests/test_utils.py
iivic: 👍

ecommerce/extensions/voucher/tests/test_utils.py
iivic: 👍

ecommerce/extensions/voucher/tests/test_utils.py
iivic: I will try to explain it better :)

ecommerce/extensions/voucher/utils.py
iivic, line 110: The purpose of this 'Code' is to explain the new change in the report.
The first row contains data that is not required in every row hence the text.
Since the Code column is the first column and there will be no data in it (since the unique voucher data starts showing from row 2 it makes sense to show the info text here.

ecommerce/extensions/api/serializers.py
iivic, line 558: Yes, this is how the download report process works.

ecommerce/extensions/voucher/tests/test_utils.py
marjev: Use `UnicodeWriter` or a similar pattern for this.
See https://docs.python.org/2/library/csv.html

marjev:
Once DRY is applied to those encoding/decoding pieces, 👍

iivic:
That should be it guys. One quick look on the last commit changes?

marjev:
👍

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -268,8 +270,8 @@ def assert_report_first_row(self, row, coupon, voucher):
         Verify that the first row fields contain the right data.
         Args:
             row (list): First row in report
-            coupon: self.coupon
-            voucher (Voucher): voucher got through Coupon Name
+            coupon (Product): Coupon for which the report is generate.
```
vkaracic: Grammar. :smile:

vkaracic:
:+1:

PR #862, opened 2016-07-18
ECOM-4800 Django management command to find out orders with multiple payments

Ayub-Khan:
[ECOM-4800](https://openedx.atlassian.net/browse/ECOM-4800)

**Description :**
Created a Django management command which checks orders from order_order ecommerce table and Payment responses in PaymentProcessorResponse to find out orders which have been paid multiple times. 

**Reviewers :**
- [ ]  @Shrhawk 
- [x]  @attiyaIshaque
- [ ] @clintonb  
  kindly check the box after review

fyi: @adampalay

### DIFF
--- a/ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
+++ b/ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
@@ -1,0 +1,74 @@
+import datetime
+from django.core.management.base import BaseCommand, CommandError
+from ecommerce.extensions.payment.models import PaymentProcessorResponse
+from ecommerce.extensions.order.models import Order
+
+
+class Command(BaseCommand):
+
+    help = """Prints order number of all orders which were paid multiple times between start date and end date.
+    Supported action:
+        [start_date] [end_date]       'Prints all orders which were paid multiple times in provided time range.'.
+    """
+
+    args = "start_date end_date [start_date] [end_date]"
+
+    @staticmethod
+    def _get_argument(args, variable_name):
+        """
+        DRY helper.  Tries to pop topmost value from `args` and raises a CommandError
+        with a formatted message in case of failure.  This function mutates `args` in place.
+        """
+        try:
+            return datetime.datetime.strptime(args.pop(0), '%d-%m-%Y')
+        except ValueError:
+            raise CommandError("{} was not specified or specified correctly in args 'd-m-y'.".format(variable_name))
+
+    def handle(self, *args, **options):
+        """
+        Main dispatch.
+        """
+        args = list(args)
+        if len(args) < 2:
+            raise CommandError("Required arguments `Start Date` and `End Date` are missing.")
+        start_date = self._get_argument(args, 'Start Date')
+        end_date = self._get_argument(args, 'End Date')
+        if start_date >= end_date:
+            raise CommandError("Argument `Start Date` must be less than `End Date`.")
+
+        orders = Order.objects.filter(date_placed__range=(start_date, end_date), total_incl_tax__gt=0)
+        for order in orders:
+            if self.number_of_payments_for_order(order) > 1:
+                print str(order.number)
+
+    @staticmethod
+    def number_of_payments_for_order(order):
+        """
+        Gets all PaymentProcessorResponses against the order and finds how many times order was paid.
+        Args:
+            order: order model object
+
+        Returns: No of payments which have not been refunded (Total Number of Payments - Total Number of refunds)
+
+        """
+        number_of_payments = 0
+        payment_processor_responses = PaymentProcessorResponse.objects.filter(basket_id=order.basket_id)
+        if len(payment_processor_responses) == 1:
+            return 1
+        for payment_processor_response in payment_processor_responses:
+            response = str(payment_processor_response.response)
+            if 'paypal' in response:
+                # checks if it's a refund response
+                if 'update_time' and 'refund' in response:
+                    number_of_payments -= 1
+                # it's not a update or message response its a payment response
+                elif ('update_time' or 'message') not in response:
+                    number_of_payments += 1
+            else:
+                # check if its a cybersource payment response
+                if 'sale' in response:
+                    number_of_payments += 1
+                # check if its a cybersource refund response
+                elif 'reconciliation' in response:
+                    number_of_payments -= 1
+        return number_of_payments
--- a/ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
+++ b/ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
@@ -1,0 +1,140 @@
+import datetime
+import ddt
+from factory.django import mute_signals
+from django.core.urlresolvers import reverse
+from django.core.management import call_command
+from django.core.management.base import CommandError
+
+from oscar.test import factories
+from oscar.core.loading import get_class, get_model
+from oscar.test.contextmanagers import mock_signal_receiver
+from ecommerce.core.tests.patched_httpretty import httpretty
+
+from ecommerce.extensions.fulfillment.status import ORDER
+from ecommerce.extensions.payment.processors.cybersource import Cybersource
+from ecommerce.extensions.payment.processors.paypal import Paypal
+from ecommerce.extensions.payment.tests.mixins import PaymentEventsMixin, CybersourceMixin, PaypalMixin
+from ecommerce.extensions.payment.management.commands.orders_with_multiple_payments import Command
+from ecommerce.tests.testcases import TestCase
+
+Basket = get_model('basket', 'Basket')
+Order = get_model('order', 'Order')
+PaymentEvent = get_model('order', 'PaymentEvent')
+PaymentEventType = get_model('order', 'PaymentEventType')
+PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
+SourceType = get_model('payment', 'SourceType')
+
+post_checkout = get_class('checkout.signals', 'post_checkout')
+
+
+@ddt.ddt
+class TestOrdersWithMultiplePaymentsCommand(PaypalMixin, CybersourceMixin, PaymentEventsMixin, TestCase):
+    """ Tests processing of orders_with_multiple_payments command for cybersouce and paypal orders"""
+
+    def setUp(self):
+        super(TestOrdersWithMultiplePaymentsCommand, self).setUp()
+        self.user = factories.UserFactory()
+        self.billing_address = self.make_billing_address()
+
+        self.basket = factories.create_basket()
+        self.basket.owner = self.user
+        self.basket.freeze()
+
+        self.cybersource_processor = Cybersource()
+        self.cybersource_processor_name = self.cybersource_processor.NAME
+
+        self.paypal_processor = Paypal()
+        self.paypal_processor_name = self.paypal_processor.NAME
+
+    def test_command_handler_without_args(self):
+        """
+        Tests the processing of main command handler with out the arguments.
+        """
+        with self.assertRaises(CommandError) as exception:
+            call_command('orders_with_multiple_payments')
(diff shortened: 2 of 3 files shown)


attiyaIshaque:
@Ayub-Khan How much time is this taking to run against replicas or live databases?

Ayub-Khan:
@attiyaIshaque  Depends on number of orders placed in given time range.

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
```diff
@@ -0,0 +1,72 @@
...
+            response = str(payment_processor_response.response)
+            if 'paypal' in response:
+                if 'update_time' and 'refund' in response:
+                    number_of_payments -= 1
+                elif 'update_time' or 'message' in response:
+                    continue
+                else:
+                    number_of_payments += 1
+            else:
+                if 'sale' in response:
```
Shrhawk, line 69: @Ayub-Khan why this is inner if ? this can be put directly in elif part.

Ayub-Khan, line 69: @Shrhawk  the outer if differentiates between the banks. If there is paypal key words in response it is paypal transection else it is cybersource. Inner ifs differentiate between the transaction types like refund , payment or info payment processor response status update. This could be done without outer ifs but it will be confusing . this is batter for understanding.  Outer ifs for differentiating between banks and inner between transaction types.

attiyaIshaque:
Once the tests pass. :+1:

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        args = list(args)
+        if len(args) < 2:
+            raise CommandError("Required arguments `Start Date` and `End Date` are missing.")
+        start_date = self._get_argument(args, 'Start Date')
+        end_date = self._get_argument(args, 'End Date')
+        if start_date >= end_date:
+            raise CommandError("Argument `Start Date` must be less than `End Date`.")
+
+        orders = Order.objects.filter(date_placed__range=(start_date, end_date)).filter(total_incl_tax__gt=0)
```
attiyaIshaque: @Ayub-Khan This looks good, but I would suggest also try using `classMethod` here.

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        Gets all PaymentProcessorResponses against the order and finds how many times order was paid.
+        Args:
+            order: order model object
+
+        Returns: No of payments which have not been refunded (Total Number of Payments - Total Number of refunds)
+
+        """
+        number_of_payments = 0
+        payment_processor_responses = PaymentProcessorResponse.objects.all().filter(basket_id=order.basket_id)
```
attiyaIshaque: It's better to use like this`PaymentProcessorResponse.objects.filter(basket_id=order.basket_id)`

ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
```diff
@@ -0,0 +1,125 @@
...
+        # Test method returns 0 if payment has been refunded
+        self.assertEquals(Command.number_of_payments_for_order(order), 0)
+        # Remove 'Refund' fields from the second response.
+        payment_response = PaymentProcessorResponse.objects.all()[1]
+        del payment_response.response[u'transactions'][0][u'related_resources'][0][u'sale'][u'links'][1]
+        payment_response.save()
+        # Test method returns no of payments equals to 1
+        self.assertEquals(Command.number_of_payments_for_order(order), 1)
+        # Change one Response to refund from payment
+        payment_response = PaymentProcessorResponse.objects.all()[0]
```
Shrhawk: we can use `first()` method of django queryset like `PaymentProcessorResponse.objects.first()`

ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
```diff
@@ -0,0 +1,125 @@
...
+PaymentEvent = get_model('order', 'PaymentEvent')
+PaymentEventType = get_model('order', 'PaymentEventType')
+PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
+SourceType = get_model('payment', 'SourceType')
+
+post_checkout = get_class('checkout.signals', 'post_checkout')
+
+
+class TestOrdersWithMultiplePaymentsCommand(PaypalMixin, CybersourceMixin, PaymentEventsMixin, TestCase):
+    """ Test processing of orders_with_multiple_payments command """
```
attiyaIshaque: Doc string needs to be updated.

ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
```diff
@@ -0,0 +1,125 @@
...
+        # Genrate Payment Responses for Paypal
+        self.mock_oauth2_response()
+        self.mock_payment_creation_response(self.basket)
+        self.paypal_processor.get_transaction_parameters(self.basket, request=self.request)
+        self.mock_payment_creation_response(self.basket, find=True)
+        self.mock_payment_execution_response(self.basket)
+        self.client.get(reverse('paypal_execute'), self.RETURN_DATA)
+        # Check order has been created after the payment
+        order = Order.objects.get(basket_id=self.basket.id)
+        self.assertIsNotNone(order, 'No order was created for the basket after payment.')
```
Shrhawk: what are we comparing here ?

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
```diff
@@ -0,0 +1,70 @@
...
+            if 'paypal' in response:
+                if 'update_time' and 'refund' in response:
+                    number_of_payments -= 1
+                elif ('update_time' or 'message') not in response:
+                    number_of_payments += 1
+            else:
+                if 'sale' in response:
+                    number_of_payments += 1
+                elif 'reconciliation' in response:
+                    number_of_payments -= 1
```
Shrhawk, line 73:
```
if 'paypal' in response:
                if 'update_time' and 'refund' in response:
                    number_of_payments -= 1
                elif ('update_time' or 'message') not in response:
                    number_of_payments += 1
 elif 'sale' in response:
               number_of_payments += 1
 elif 'reconciliation' in response:
              number_of_payments -= 1
```

@Ayub-Khan is this seems cleaner to you ?

ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
```diff
@@ -0,0 +1,125 @@
...
+        # Remove 'update_time' as it's never a part of real first paypal response.
+        payment_response = PaymentProcessorResponse.objects.all()[0]
```
attiyaIshaque: Adding a `.first()`method to Django's QuerySet.

ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        with self.assertRaises(CommandError) as exception:
+            call_command('orders_with_multiple_payments')
+            self.assertEqual(exception.message, "Required arguments `Start Date` and `End Date` are missing.")
+        with self.assertRaises(CommandError) as exception:
+            call_command('orders_with_multiple_payments', '01-02-2016', '01-01-2016')
+            self.assertEqual(exception.message, "Argument `Start Date` must be less than `End Date`.")
+        with self.assertRaises(CommandError) as exception:
+            call_command('orders_with_multiple_payments', '2016-02-01', '01-01-2016')
+            self.assertEqual(exception.message, "Start Date was not specified or specified correctly in args 'd-m-y'.")
```
Shrhawk: we can use  @ddt.data to avoid reputation of code from line no **50-58**
**Example**: https://github.com/edx/ecommerce/blob/37cf4c5f8a68df3cf8695d396c7a8768a89b1aac/ecommerce/courses/tests/test_utils.py#L24

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
Ayub-Khan: In the Next if im checking the length of object this will return so it will be fine if there are more than 1 PaymentProcessorResponses but if there is only 1 it will fail . this way it will return list for both cases even its one.

ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
Ayub-Khan: oh yeah.

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
attiyaIshaque: Using the `.filter` also, returns a list event if there are matching records. Note it will not fail event If there are not any matching records. So feel free to use.

Ayub-Khan: if there is only one object .filter does not return it in the list.

ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
Ayub-Khan: that order is not None

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
Shrhawk: @Ayub-Khan we can use single filter here 
`Order.objects.filter(date_placed__range=(start_date, end_date), total_incl_tax__gt=0)`

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
Ayub-Khan, line 73: yeah. i will add some comments there.

ecommerce/extensions/payment/management/tests/test_orders_with_multiple_payments.py
Shrhawk: then this is enough to compare the order as:
`self.assertIsNotNone(order)`

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
attiyaIshaque: `.filter` always return a list either it has object or not.

attiyaIshaque:
@Ayub-Khan Only one comment; once you address all these comments.

attiyaIshaque:
Once commits squashed 👍

Shrhawk:
👍

ecommerce/extensions/payment/management/commands/orders_with_multiple_payments.py
```diff
@@ -0,0 +1,74 @@
...
+            raise CommandError("Required arguments `Start Date` and `End Date` are missing.")
+        start_date = self._get_argument(args, 'Start Date')
+        end_date = self._get_argument(args, 'End Date')
+        if start_date >= end_date:
+            raise CommandError("Argument `Start Date` must be less than `End Date`.")
+
+        orders = Order.objects.filter(date_placed__range=(start_date, end_date), total_incl_tax__gt=0)
+        for order in orders:
+            if self.number_of_payments_for_order(order) > 1:
+                print str(order.number)
```
adampalay, line 42: we should probably be logging this instead of printing it

adampalay:
@Ayub-Khan , please get a review from someone on the ecom team too before merging this

Ayub-Khan:
@clintonb Kindly Review.

clintonb:
What's the use case and rollout plan for this? Who is expected to run this command, and with what frequency?

PR #865, opened 2016-07-20
Added Adyen template overrides

douglashall:
Initial implementation of an approach which would allow payment processor code to override default Otto/Oscar templates.


douglashall:
FYI @zubair-arbi

ecommerce/settings/base.py
```diff
@@ -197,6 +199,11 @@
         }
     },
 ]
+
+# Payment processor template override directories
+PAYMENT_PROCESSOR_TEMPLATE_DIRS = [
```
zubair-arbi: @douglashall is this setting still necessary?

zubair-arbi:
@douglashall this approach overall looks goods. I have a few questions:
1. We are moving `adyen` processor configuration in a new app `adyen`, shouldn't this be in `ecommerce/extensions/payment/processors` with other payment processors?
2. In template loaders configuration we are adding templates by getting active processors list and since for now we have custom theme for `adyen` only so its ok but should we worry about the priority  in case there will be multiple custom processor's themes?

douglashall:
Closing in favor of https://github.com/edx/ecommerce/pull/883

PR #866, opened 2016-07-20
[SOL-1936] Load courses asynchronously on dynamic coupon preview.

vkaracic:
The preview button on dynamic coupon creation tool used to fetch all the results for a specific query and then display it in the data table. This is slow and prone to timeouts for queries that contained a lot of courses.
This PR contains changes that make the preview button load courses asynchronously with a set timeout of 500 ms between AJAX calls to the course discovery service.

https://openedx.atlassian.net/browse/SOL-1936

_nevermind the branch name_

### DIFF
--- a/ecommerce/extensions/api/v2/views/catalog.py
+++ b/ecommerce/extensions/api/v2/views/catalog.py
@@ -41,21 +41,30 @@ class CatalogViewSet(NestedViewSetMixin, ReadOnlyModelViewSet):
         """
         query = request.GET.get('query')
         seat_types = request.GET.get('seat_types')
+        offset = request.GET.get('offset')
+        limit = request.GET.get('limit', DEFAULT_CATALOG_PAGE_SIZE)
+
         if query and seat_types:
             seat_types = seat_types.split(',')
             try:
-                results = get_range_catalog_query_results(
-                    limit=DEFAULT_CATALOG_PAGE_SIZE,
+                response = get_range_catalog_query_results(
+                    limit=limit,
                     query=query,
-                    site=request.site
-                )['results']
+                    site=request.site,
+                    offset=offset
+                )
+                results = response['results']
                 course_ids = [result['key'] for result in results]
                 courses = serializers.CourseSerializer(
                     Course.objects.filter(id__in=course_ids),
                     many=True,
                     context={'request': request}
                 ).data
-                return Response(data=[course for course in courses if course['type'] in seat_types])
+                data = {
+                    'next': response['next'],
+                    'courses': [course for course in courses if course['type'] in seat_types]
+                }
+                return Response(data=data)
             except (ConnectionError, SlumberBaseException, Timeout):
                 logger.error('Unable to connect to Course Catalog service.')
                 return Response(status=status.HTTP_400_BAD_REQUEST)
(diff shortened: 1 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=866

ecommerce/static/js/views/dynamic_catalog_view.js
```diff
@@ -38,22 +38,10 @@ define(['jquery',
...
-                        query : this.query,
-                        seat_types: this.seat_types.join()
-                    },
-                    success: this.onSuccess
-                });
-            },
-
-            onSuccess: function(data) {
-                this.$el.find('#coursesTable').DataTable({
+                this.table = this.$('#coursesTable').DataTable({
```
mjfrey: The table needs to be cleared each time otherwise you keep appending to the existing list of courses.

mjfrey:
How do you know when there are no more pages left to get?

ecommerce/static/js/views/dynamic_catalog_view.js
vkaracic: I don't understand that logic, can you clarify what the use/benefit of that is?

vkaracic:
This line: https://github.com/edx/ecommerce/pull/866/files#diff-5d2f8c42b4a1b2fa8c7d3832380dc426R85 checks to see if the data returned contains anything, if not it stops the process.

ecommerce/static/js/views/dynamic_catalog_view.js
vkaracic: Oh, you probably meant when the query gets changed and previewed again, good point!

mjfrey:
But what if a page of data does not contain matches for our query, but there are more in others? I think you are going to need to use the logic I pasted in the ticket.  Check for the existence of more pages.

vkaracic:
Comment have been addressed, can you take another look?

vkaracic:
@marjev

ecommerce/core/constants.py
```diff
@@ -18,7 +18,7 @@
 ENROLLMENT_CODE_SEAT_TYPES = ['verified', 'professional', 'no-id-professional']
 
 # Course Catalog constants
-DEFAULT_CATALOG_PAGE_SIZE = 10000
+DEFAULT_CATALOG_PAGE_SIZE = 10
```
marjev: Will this affect any other pieces of code? I think that every API call to the Course Discovery uses this, but I'm not sure.

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -40,19 +40,30 @@ def preview(self, request):
         """
         query = request.GET.get('query')
         seat_types = request.GET.get('seat_types')
+        offset = request.GET.get('offset')
```
marjev, line 44: What if offset is not provided? Should we check for that or set a default value?

marjev:
Except for a couple of questions, LGTM 👍

ecommerce/extensions/api/v2/views/catalog.py
vkaracic, line 44: If `offset` is not provided in the request it will default to `None` which, in the API call, will result in the first page being loaded.

ecommerce/core/constants.py
vkaracic: You're quite right! Several other apps are using this.

mjfrey:
:+1: 
Please squash.

PR #868, opened 2016-07-20
WL-499 Remove duplicate analytics initialization code

douglashall:
This PR eliminates a JS console error ("Segment snippet included twice.") that was being thrown on several page renders. The issue was that we were calling `AnalyticsUtils.analyticsSetUp();` from both the [common module](https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/common.js#L25) which is loaded on every page and from page specific modules. I have removed the duplicate calls from the page specific modules.

I have also removed duplicate definitions of backbone model data that is rendered within script tags in templates.

@vkaracic @mjfrey @hasnain-naveed

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -7,14 +7,12 @@ define([
         'underscore',
         'underscore.string',
         'utils/utils',
-        'utils/analytics_utils',
         'js-cookie'
     ],
     function ($,
               _,
               _s,
               Utils,
-              AnalyticsUtils,
               Cookies
     ) {
         'use strict';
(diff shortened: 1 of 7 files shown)


vkaracic:
Are the same user behaviors still getting recorded?

douglashall:
@vkaracic Yes, analytics initialization is handled [here](https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/common.js#L25).

vkaracic:
I was thinking about this part https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/utils/analytics_utils.js#L29-L32
since you are removing `initModelData` initializations from a couple of templates, and if that would affect user behavior recordings?

douglashall:
@vkaracic Yes, we set up the model data in the base template [here](https://github.com/edx/ecommerce/blob/master/ecommerce/templates/edx/base.html#L44). We do not need to do it again in templates that inherit from `edx/base.html`, right?

vkaracic:
In that case :+1:

mjfrey:
👍

PR #874, opened 2016-07-29
SOL-1926: Add org filter on courses

malikshahzad228:
Jira link: [Course Administration (CAT): Course listings view should reflect White Label experience](https://openedx.atlassian.net/browse/SOL-1926)

### DIFF
--- a/ecommerce/core/management/commands/create_or_update_site.py
+++ b/ecommerce/core/management/commands/create_or_update_site.py
@@ -46,6 +46,12 @@ class Command(BaseCommand):
                             type=str,
                             default='',
                             help='Partner name to select/create and associate with site.')
+        parser.add_argument('--partner-orgs',
+                            action='store',
+                            dest='partner_orgs',
+                            type=str,
+                            default='',
+                            help='Organization filter for partner course listings.')
         parser.add_argument('--lms-url-root',
                             action='store',
                             dest='lms_url_root',
@@ -101,6 +107,7 @@ class Command(BaseCommand):
         site_name = options.get('site_name')
         partner_code = options.get('partner_code')
         partner_name = options.get('partner_name')
+        partner_orgs = options.get('partner_orgs')
         lms_url_root = options.get('lms_url_root')
         client_id = options.get('client_id')
         client_secret = options.get('client_secret')
@@ -121,11 +128,11 @@ class Command(BaseCommand):
         site.save()
 
         partner, partner_created = Partner.objects.get_or_create(code=partner_code)
-        if partner_created:
-            partner.name = partner_name
-            partner.short_code = partner_code
-            partner.save()
-            logger.info('Partner created with code %s', partner_code)
+        partner.name = partner_name
+        partner.short_code = partner_code
+        partner.organization_list = partner_orgs
+        partner.save()
+        logger.info('Partner %s with code %s', 'created' if partner_created else 'updated', partner_code)
 
         SiteConfiguration.objects.update_or_create(
             site=site,
--- a/ecommerce/core/tests/test_commands.py
+++ b/ecommerce/core/tests/test_commands.py
@@ -26,6 +26,7 @@ class CreateOrUpdateSiteCommandTests(TestCase):
         self.client_secret = 'ecommerce-secret'
         self.segment_key = 'test-segment-key'
         self.from_email = 'site_from_email@example.com'
+        self.partner_orgs = 'test-orgs'
 
     def _check_site_configuration(self, site, partner):
         site_configuration = site.siteconfiguration
@@ -41,7 +42,7 @@ class CreateOrUpdateSiteCommandTests(TestCase):
 
     def _call_command(self, site_domain, partner_code, lms_url_root, client_id, client_secret, from_email,
                       site_id=None, site_name=None, partner_name=None, theme_scss_path=None,
-                      payment_processors=None, segment_key=None, enable_enrollment_codes=False):
+                      payment_processors=None, segment_key=None, enable_enrollment_codes=False, partner_orgs=None):
         """
         Internal helper method for interacting with the create_or_update_site management command
         """
@@ -74,6 +75,8 @@ class CreateOrUpdateSiteCommandTests(TestCase):
             command_args.append('--enable-enrollment-codes={enable_enrollment_codes}'.format(
                 enable_enrollment_codes=enable_enrollment_codes
             ))
+        if partner_orgs:
+            command_args.append('--partner-orgs={partner_orgs}'.format(partner_orgs=partner_orgs))
         call_command(self.command_name, *command_args)
 
     def test_create_site(self):
@@ -127,6 +130,67 @@ class CreateOrUpdateSiteCommandTests(TestCase):
         self.assertEqual(site.name, updated_site_name)
         self._check_site_configuration(site, partner)
         self.assertTrue(site.siteconfiguration.enable_enrollment_codes)
+
+    def test_create_partner(self):
+        """ Verify the command creates Site, Partner, and SiteConfiguration. """
+        site_domain = 'ecommerce-fake1.server'
+        new_partner = 'fake2'
+        partner_orgs = 'test-org'
+
+        self.assertFalse(Partner.objects.filter(code=new_partner))
+
+        self._call_command(
+            site_domain=site_domain,
+            partner_code=new_partner,
+            lms_url_root=self.lms_url_root,
+            theme_scss_path=self.theme_scss_path,
+            payment_processors=self.payment_processors,
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            segment_key=self.segment_key,
+            from_email=self.from_email,
+            partner_orgs=partner_orgs
+        )
+
+        partner = Partner.objects.get(code=new_partner)
+        self.assertEqual(partner.short_code, new_partner)
+        self.assertEqual(partner.organization_list, partner_orgs)
+
+    def test_update_partner_org(self):
+        """ Verify the command updates partner organization """
+        site_domain = 'ecommerce-fake2.server'
+        partner_org1 = 'test-org1'
+        partner_org2 = 'test-org2'
+
+        self._call_command(
+            site_domain=site_domain,
+            partner_code=self.partner,
+            lms_url_root=self.lms_url_root,
+            theme_scss_path=self.theme_scss_path,
+            payment_processors=self.payment_processors,
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            segment_key=self.segment_key,
+            from_email=self.from_email,
+            partner_orgs=partner_org1
+        )
+
+        self.assertEqual(Partner.objects.get(code=self.partner).organization_list, partner_org1)
+
+        self._call_command(
+            site_domain=site_domain,
+            partner_code=self.partner,
+            lms_url_root=self.lms_url_root,
+            theme_scss_path=self.theme_scss_path,
+            payment_processors=self.payment_processors,
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            segment_key=self.segment_key,
+            from_email=self.from_email,
+            partner_orgs=partner_org2
+        )
+
+        self.assertEqual(Partner.objects.get(code=self.partner).organization_list, partner_org2)
 
     @data(
         ['--site-id=1'],
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -36,6 +36,7 @@ class Course(models.Model):
     )
     history = HistoricalRecords()
     thumbnail_url = models.URLField(null=True, blank=True)
+    organization = models.CharField(max_length=100, null=False, blank=False)
 
     def __unicode__(self):
         return unicode(self.id)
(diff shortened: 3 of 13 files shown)


malikshahzad228:
@mattdrayer Build is failing currently:

**Error Message:**
Segmentation fault (core dumped)
make: **\* [validate_python] Error 139

The command "make validate_python" exited with 2.

Debugging the cause of this. but change is final. Any thoughts from you?

ecommerce/extensions/partner/models.py
```diff
@@ -12,6 +12,7 @@ class StockRecord(AbstractStockRecord):
 class Partner(AbstractPartner):
     # short_code is the unique identifier for the 'Partner'
     short_code = models.CharField(max_length=8, unique=True, null=False, blank=False)
+    org = models.CharField(max_length=30, unique=True, null=True, blank=True)
```
mattdrayer: @malikshahzad228 -- this line is the cause of the SegmentationFault that we are seeing in the unit tests.  When I remove all attributes except `max_length=30` the tests pass.  When I add back `unique=True` only, I observe `IntegrityError: column org is not unique`.  Adding `null=True` along with the unique attribute results in the SegmentationFault.  

I'm not sure why this is the behavior, but at least we understand what is going on now and can do something about it 😄

vkaracic: From my experience, segfaults happen in the sql test database. Separating the unit tests into two different classes usually helps.

ecommerce/extensions/partner/models.py
```diff
@@ -12,6 +12,7 @@ class StockRecord(AbstractStockRecord):
 class Partner(AbstractPartner):
     # short_code is the unique identifier for the 'Partner'
     short_code = models.CharField(max_length=8, unique=True, null=False, blank=False)
+    org = models.CharField(max_length=30, blank=True)
```
mattdrayer: this should be 'organization', and if it is going to contain more than one entry (ie, a list) it should be 'organizations' or 'organization_list'

ecommerce/extensions/api/v2/tests/views/test_publication.py
```diff
@@ -290,6 +290,21 @@ def test_invalid_course_id(self):
...
         self.assert_course_does_not_exist(self.course_id)
 
+    def test_invalid_course_org(self):
+        """Verify that attempting to save a course with a bad ORG yields a 500."""
+        partner = self.partner
+        dummy_org_name = 'bad-org'
+        partner.org = dummy_org_name
+        partner.save()
+        error_msg = 'Course [{course_id}] is not saved. ' \
+                    'Courses only form organization {partner_org}' \
```
mattdrayer: 'form' or 'from'?

ecommerce/extensions/partner/models.py
malikshahzad228: Renamed to 'organization'.
As discussed with Kashif and Zeeshan, one course only could be from one organization. And usually one partner is against one organization.

If you can think of any other scenarios then kindly let me know.

ecommerce/core/management/commands/create_or_update_site.py
```diff
@@ -46,6 +46,12 @@ def add_arguments(self, parser):
                             type=str,
                             default='',
                             help='Partner name to select/create and associate with site.')
+        parser.add_argument('--partner-org',
+                            action='store',
+                            dest='partner_org',
+                            type=str,
+                            default='',
+                            help='Partner org to restrrict course from that organization.')
```
mattdrayer: "Organization filter for partner course listings."

ecommerce/core/management/commands/create_or_update_site.py
```diff
@@ -121,11 +128,22 @@ def handle(self, *args, **options):
...
-            partner.name = partner_name
-            partner.short_code = partner_code
-            partner.save()
-            logger.info('Partner created with code %s', partner_code)
+        partner.name = partner_name
+        partner.short_code = partner_code
+
+        if partner_org:
+            # ensuring uniqueness of partner
+            # adding constraint to database causing segmentation fault error
```
mattdrayer: The segmentation fault isn't actually the error -- it is an IntegrityError that is being caused by multiple 'null' values being added in the unit test.  The SegementationFault was being raised on top of the IntegrityError because of the "null=True" attribute on the model.

ecommerce/core/management/commands/create_or_update_site.py
```diff
@@ -121,11 +128,22 @@ def handle(self, *args, **options):
...
         partner, partner_created = Partner.objects.get_or_create(code=partner_code)
-        if partner_created:
-            partner.name = partner_name
-            partner.short_code = partner_code
-            partner.save()
-            logger.info('Partner created with code %s', partner_code)
+        partner.name = partner_name
+        partner.short_code = partner_code
+
+        if partner_org:
```
mattdrayer: I feel like we should perform the partner_org existence check before trying to create a new partner at all.  Can we move this ahead of the get_or_create call on line 130?   I think that would help to simplify the check -- basically if a partner_org value is specified but it's already in use, we should cancel the workflow.  And we should note that we cannot leverage the automatic Django ORM exception framework here, because there is a unique constraint on the nullable field.

ecommerce/extensions/api/serializers.py
```diff
@@ -278,16 +278,26 @@ def save(self):
         course_verification_deadline = self.validated_data.get('verification_deadline')
         products = self.validated_data['products']
         partner = self.get_partner()
+        partner_org = partner.org
 
         try:
             if not waffle.switch_is_active('publish_course_modes_to_lms'):
+                # if waffle switch is not on from Django admin
```
mattdrayer: This comment doesn't introduce any more context from the code above -- I'm all for code comments, but they should add value.

ecommerce/extensions/api/serializers.py
```diff
@@ -278,16 +278,26 @@ def save(self):
...
                 message = _(
                     u'Course [{course_id}] was not published to LMS '
                     u'because the switch [publish_course_modes_to_lms] is disabled. '
                     u'To avoid ghost SKUs, data has not been saved.'
                 ).format(course_id=course_id)
 
                 raise Exception(message)
+            elif partner_org and course_id.find(partner_org) == -1:
```
mattdrayer: we can't parse the course_id value to determine the org, unfortunately -- we need to treat course keys as "opaque", because they can change at any time -- for example, a course key could change to "course-v2:09s8-09-sdfs-23e2323423423sdfs" at some point in the future, and there is no org available now.  We'll need to determine the org by converting the course_id string into a CourseKey instance.

ecommerce/extensions/partner/migrations/0009_partner_org.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
mattdrayer: We'll need to regenerate this migration file after changing the attribute from "org" to "organization"

ecommerce/extensions/partner/models.py
mattdrayer: I think it's possible in the near future that we may have the need for a partner to combine more than one organization -- for example, a site which lists both MITx and MITProfessionalX courses together.

malikshahzad228: Then I believe one site could also belongs to multiple partners. So no need to verify site uniqueness for partners. Like
- foo lists courses of MITx
- bar lists courses of MITx and MITProfessionalX

mattdrayer: I'm not sure we can consider partner and organization to be equivalent in terms of integrity/constraint.  The case you describe above is a good example.  Partner "foo" has site "foo.com" which lists MITx courses.  Partner "bar" has site "bar.com" which lists MITx and MITProfessionalX courses.  I do see the potential for the site:partner relationship to become sites:partners, but we do not have to alter that relationship now?

malikshahzad228: I feel that relationship between site and partner will remain unchanged. And only partner should have multiple organizations.
And field will be "organization_list".

mattdrayer: Okay -- it would be nice if there was a corresponding Organization model/class in the ecommerce system, but I guess there is not, so organization_list should suffice for now.

mattdrayer:
LGTM -- 👍 -- @mjfrey or @vkaracic can you take a look as well

mjfrey:
@mattdrayer This is not filtering the list shown in the course admin tool?

mattdrayer:
HAH, you are right @mjfrey -- this PR at the moment only covers adding the organization list -- still need to filter out the courses from the view.  Sorry, I got ahead of myself with the review 😄

ecommerce/core/tests/test_commands.py
```diff
@@ -128,6 +131,102 @@ def test_update_site(self):
...
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            segment_key=self.segment_key,
+            from_email=self.from_email,
+            partner_orgs=partner_orgs
+        )
+
+        self.assertEqual(Partner.objects.get(code=self.partner).short_code, self.partner)
+
+        self._call_command(
```
vkaracic: Why do you call the same command twice?

malikshahzad228: @vkaracic 1st command updated. This was meant to create partner without org.

ecommerce/courses/management/commands/populate_model.py
```diff
@@ -0,0 +1,18 @@
+""" This command populate organization in existing courses."""
```
mattdrayer: We need to ensure that command modules are named as specifically as possible -- so in this case we might name the module "add_course_organization.py".

ecommerce/courses/models.py
```diff
@@ -36,6 +36,7 @@ class Course(models.Model):
     )
     history = HistoricalRecords()
     thumbnail_url = models.URLField(null=True, blank=True)
+    organization = models.CharField(max_length=100, null=False, blank=False)
```
mattdrayer: organization, or organization_list?

ecommerce/extensions/api/serializers.py
```diff
@@ -277,6 +278,10 @@ def save(self):
         course_verification_deadline = self.validated_data.get('verification_deadline')
         products = self.validated_data['products']
         partner = self.get_partner()
+        organization_list = partner.organization_list
+        course_org = CourseKey.from_string(course_id).org
+
+        from pdb import set_trace; set_trace()
```
mattdrayer: Need to remove this debug statement

mattdrayer:
Take 2:  👍 👍 for me -- @mjfrey and/or @vkaracic, this PR is ready for another look.

ecommerce/core/tests/test_commands.py
```diff
@@ -128,6 +131,104 @@ def test_update_site(self):
...
+            from_email=self.from_email,
+            partner_orgs=partner_orgs
+        )
+
+        partner = Partner.objects.get(code=new_partner)
+        self.assertEqual(partner.short_code, new_partner)
+        self.assertEqual(partner.organization_list, partner_orgs)
+
+    def test_update_partner(self):
+        """ Verify the command updates Site and creates Partner, and SiteConfiguration """
```
vkaracic: Says here this updates Site, but from the code below I'd say that this updates the organizations list?

ecommerce/core/tests/test_commands.py
```diff
@@ -128,6 +131,104 @@ def test_update_site(self):
...
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            segment_key=self.segment_key,
+            from_email=self.from_email,
+            partner_orgs=partner_orgs
+        )
+
+        self.assertEqual(Partner.objects.get(code=self.partner).organization_list, partner_orgs)
+
+    def test_update_partner_org(self):
```
vkaracic, line 159: Unless I'm missing something, this test does the same thing as the one above, so you can remove the above one.

ecommerce/courses/models.py
vkaracic, line 39: You can get the organization from the course key, I don't think field is necessary. Also in https://github.com/edx/ecommerce/pull/874/files#diff-852a6b45e4fa1255461cf3fa6e5ceda7R282 you're getting the organization from the key instead of this field.

vkaracic, line 39: OK, never mind that link, I saw now that you're saving the organization on line 307 in serializer. But I'm still curious why this field is necessary?

vkaracic, line 39: OK, never mind # 2 (I should really start to look through the code before asking). I see you're using this in the queryset filter. I guess filtering like that is faster.

malikshahzad228, line 39: @vkaracic - yes, you got it right. this field was necessary or else I have to fetch all the courses and then filter them while getting org from there ids.

mattdrayer:
It seems to me that @vkaracic since you have not noted any major issues I think we should move forward with the merge.

vkaracic:
Yup! Good work :+1:

clintonb:
We should chat about this before it gets merged again. This greatly muddies the multi-tenant waters unnecessarily. We already have the `Partner` model. Associate `Courses` with that model, not this new concept.

PR #877, opened 2016-08-02
Display the basket bulk enrollment purchase if it's enabled in the settings.

vkaracic:
If creating enrollment codes was enabled then the switch link in the basket would not be visible if the basket did not contain enrollment codes, so switching from bulk enrollment to individual seat and then back wouldn't be possible. 
This PR contains changes that fix that behavior. Now the switch link is present for courses that have an enrollment code associated with them.

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -113,15 +113,15 @@ class BasketSummaryView(BasketView):
             except (ConnectionError, SlumberBaseException, Timeout):
                 logger.exception('Failed to retrieve data from Course API for course [%s].', course_key)
 
-            # Set to true if any course in basket has bulk purchase scenario
-            if line.product.get_product_class().name == ENROLLMENT_CODE_PRODUCT_CLASS_NAME and \
-                    self.request.site.siteconfiguration.enable_enrollment_codes:
-                is_bulk_purchase = True
-                # Iterate on message storage so all messages are marked as read
-                list(messages.get_messages(self.request))
-
-                # Get variables for alternative basket view
+            if self.request.site.siteconfiguration.enable_enrollment_codes:
+                # Get variables for the switch link that toggles from enrollment codes and seat.
                 switch_link_text, partner_sku = get_basket_switch_data(line.product)
+                if line.product.get_product_class().name == ENROLLMENT_CODE_PRODUCT_CLASS_NAME:
+                    is_bulk_purchase = True
+                    # Iterate on message storage so all messages are marked as read.
+                    # This will hide the success messages when a user updates the quantity
+                    # for an item in the basket.
+                    list(messages.get_messages(self.request))
 
             if line.has_discount:
                 benefit = self.request.basket.applied_offers().values()[0].benefit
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=877

ecommerce/extensions/basket/views.py
```diff
@@ -113,15 +113,14 @@ def get_context_data(self, **kwargs):
...
                 logger.exception('Failed to retrieve data from Course API for course [%s].', course_key)
 
-            # Set to true if any course in basket has bulk purchase scenario
-            if line.product.get_product_class().name == ENROLLMENT_CODE_PRODUCT_CLASS_NAME and \
-                    self.request.site.siteconfiguration.enable_enrollment_codes:
-                is_bulk_purchase = True
-                # Iterate on message storage so all messages are marked as read
-                list(messages.get_messages(self.request))
-
+            if self.request.site.siteconfiguration.enable_enrollment_codes:
```
mattdrayer, line 116: Can we add some better comments for this new block to describe exactly what is going on?  Most of the comments here simply restate what the code is doing, which isn't really helping anyone.

ecommerce/extensions/basket/views.py
```diff
@@ -113,15 +113,13 @@ def get_context_data(self, **kwargs):
...
-                # Iterate on message storage so all messages are marked as read
-                list(messages.get_messages(self.request))
-
-                # Get variables for alternative basket view
+            if self.request.site.siteconfiguration.enable_enrollment_codes:
+                # Get variables for the switch link that toggles from enrollment codes and seat.
                 switch_link_text, partner_sku = get_basket_switch_data(line.product)
+                if line.product.get_product_class().name == ENROLLMENT_CODE_PRODUCT_CLASS_NAME:
+                    is_bulk_purchase = True
+                    # Iterate on message storage so all messages are marked as read
```
mattdrayer: @malikshahzad228 -- can you provide some information about what this line of code does?  We are trying to determine if it should be moved out of the containing "if" block

malikshahzad228: @mattdrayer This checks if any item in loop from basket is for bulk purchase, if yes it marks the 'is_bulk_purchase' flag as true. later this flag passed to context to hide some functionality(links, and alert messages).

mattdrayer: Okay, so should L122 only be executed if enrollment codes is enabled, and the current line.product is an enrollment code product?

malikshahzad228: Yes, we could say that. But more appropriate words will be 'L122 only be executed if enrollment codes is enabled, and the any item in loop/basket (line.product) is an enrollment code product'

(This fix was for hiding success messages when user updates the quantity for item in basket.)

vkaracic:
@mattdrayer PR updated with an additional test and I added to the comment about the success messages being read. Should all be fine now.

mattdrayer:
Great, thanks guys for working on this together!  👍 after checks pass

PR #878, opened 2016-08-03
Hide empty fields in coupon details view.

vkaracic:
Empty coupon fields `Note` and `Maximum Uses` would appear in the coupon detail view. Usually empty fields are hidden on the details page so this PR contains code that hides those two fields if they are empty.

https://openedx.atlassian.net/browse/SOL-1985
@mjfrey @iivic

### DIFF
--- a/ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
+++ b/ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
@@ -33,6 +33,16 @@ define([
                 percentageDiscountCodeVoucher = Mock_Coupons.percentageDiscountCodeVoucher;
                 valueDiscountCodeVoucher = Mock_Coupons.valueDiscountCodeVoucher;
                 verifiedSeat = Mock_Coupons.verifiedSeat;
+
+                jasmine.addMatchers({      
+                    toExist: function () {
+                        return {
+                            compare: function (actual) {
+                                return { pass: $.contains(document.documentElement, $(actual)[0]) };
+                            }
+                        };
+                    }
+                });
             });
 
             it('should compare view.model with model sent', function () {
@@ -181,8 +191,8 @@ define([
                 expect(view.$('.invoice-payment-date .value').text()).toEqual(
                     view.formatDateTime(model.get('invoice_payment_date'))
                 );
-                expect(SpecUtils.visibleElement(view, '.invoice_discount_type', '.info-item')).toBe(false);
-                expect(SpecUtils.visibleElement(view, '.invoice_discount_value', '.info-item')).toBe(false);
+                expect(view.$('.invoice_discount_type')).not.toExist();
+                expect(view.$('.invoice_discount_value')).not.toExist();
             });
 
             it('should render postpaid invoice data.', function() {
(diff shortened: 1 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=878&repo=edx%2Fecommerce

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -141,6 +142,23 @@ define([
...
+                    note.addClass('hidden');
+                }
+
+                if (this.model.get('max_uses')) {
+                    max_uses.removeClass('hidden');
+                } else {
+                    max_uses.addClass('hidden');
+                }
+            },
+
```
iivic: Would it be possible to make this function universal to all model fields instead of this two fields?

ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
```diff
@@ -138,6 +138,25 @@ define([
...
+                view.render();
+                expect(SpecUtils.visibleElement(view, '.note', '.info-item')).toBe(true);
+                expect(SpecUtils.visibleElement(view, '.max-uses', '.info-item')).toBe(true);
+
+                view.model.unset('max_uses');
+                view.render();
+                expect(SpecUtils.visibleElement(view, '.max-uses', '.info-item')).toBe(false);
+                
+                view.model.unset('note');
+                view.render();
```
iivic: Why is it necessary to call render() two times? Can you just unset both fields and call render() one time?

ecommerce/static/js/views/coupon_detail_view.js
vkaracic: Not really, because the attribute names in the model are different in different ways than the field names.

ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
vkaracic: So that I can test that each field is not dependent on the other.

iivic:
👍

vkaracic:
@AlasdairSwan can you please review this?

ecommerce/static/js/views/coupon_detail_view.js
AlasdairSwan: @vkaracic will these fields ever be shown after the page is rendered? If not then it would be cleaner to not render them at all (instead of hiding them after already rendering them). It would make sense to have this conditional logic in the template.

vkaracic: @AlasdairSwan good idea! I implemented a filter on the template side, can you please review?

ecommerce/static/templates/coupon_detail.html
```diff
@@ -29,21 +29,25 @@ <h3 class="hd-3">
             <div class="heading"><%= gettext('Category') %></div>
             <div class="value"><%= category %></div><br />
         </div>
+        <% if(coupon['note']) {%>
```
AlasdairSwan: Dot notation is preferable (i.e. `coupon.note`).

ecommerce/static/templates/coupon_detail.html
```diff
@@ -29,21 +29,25 @@ <h3 class="hd-3">
...
         </div>
+        <%}%>
         <div class="info-item grid-item discount-value">
             <div class="heading"><%= gettext('Discount Value:') %></div>
             <div class="value"><%= discountValue %></div>
         </div>
         <div class="info-item grid-item course-info">
             <div class="heading"><%= gettext('Valid for courses:') %></div>
         </div>
+        <% if(coupon['catalog_query']) {%>
```
AlasdairSwan: Dot notation is preferable (i.e. `coupon.catalog_query`).

ecommerce/static/templates/coupon_detail.html
```diff
@@ -54,19 +58,23 @@ <h3 class="hd-3">
...
             <div class="heading"><%= gettext('Seat Types:') %></div>
             <div class="value"><%= courseSeatType %></div>
             <div class="catalog_buttons"></div>
         </div>
+        <%}%>
         <div class="info-item grid-item usage-limitations">
             <div class="heading"><%= gettext('Usage Limitations:') %></div>
             <div class="value"><%= usage %></div>
         </div>
+        <% if(coupon['max_uses']) {%>
```
AlasdairSwan: Dot notation is preferable (i.e. `coupon.max_uses`).

AlasdairSwan:
Once the nitpicks about dot notation are addressed and the branch is rebased it's good to go 👍

PR #884, opened 2016-08-08
[SOL-1989] Remove duplicate invoices

vkaracic:
While coupons could've been re-used, a new invoice would be created for each re-use, resulting in a coupon having more than one invoice.
Since the new invoice fields are introduced after the coupon re-used was removed, the duplicates have pretty much the same values (except for date created, ID, etc.), so this command removes all but the first invoice created for each coupon that has duplicate invoices.

https://openedx.atlassian.net/browse/SOL-1989
@mjfrey

### DIFF
--- a/ecommerce/invoice/management/commands/squash_duplicate_invoices.py
+++ b/ecommerce/invoice/management/commands/squash_duplicate_invoices.py
@@ -1,0 +1,26 @@
+import logging
+
+from django.core.management import BaseCommand
+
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=884&repo=edx%2Fecommerce

mjfrey:
👍

PR #885, opened 2016-08-08
Exclude unused URLs from core Oscar apps

rlucioni:
@clintonb @mjfrey @schenedx @adampalay @e0d


clintonb:
👍

mjfrey:
👍

e0d:
@mjfrey I'm assuming your review includes thoughts about the whitelabel cases?

PR #886, opened 2016-08-08
Add partner code to catalog cache key

mjfrey:
Add the partner short code to the cache key so we do not collide.


mjfrey:
@mattdrayer

mattdrayer:
👍

PR #887, opened 2016-08-09
Remove vouchers when switching to bulk enrollment code basket.

vkaracic:
When switching to a basket that contains enrollment codes, applying a voucher to that product returns an error, and it shouldn't be possible to do that anyway. 
The code in this PR makes sure that no voucher is applied if an enrollment code is in the basket.

https://openedx.atlassian.net/browse/SOL-2003

### DIFF
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -19,9 +19,9 @@ def prepare_basket(request, product, voucher=None):
     Create or get the basket, add the product, apply a voucher, and record referral data.
 
     Existing baskets are merged. The specified product will
-    be added to the remaining open basket. If a voucher is passed, all existing
-    ones added to the basket are removed because we allow only one voucher per
-    basket after the Voucher is applied to the basket.
+    be added to the remaining open basket. If voucher is passed, all existing
+    vouchers added to the basket are removed because we allow only one voucher per basket.
+    Vouchers are not applied if an enrollment code product is in the basket.
 
     Arguments:
         request (Request): The request object made to the view.
@@ -34,12 +34,13 @@ def prepare_basket(request, product, voucher=None):
     basket = Basket.get_basket(request.user, request.site)
     basket.flush()
     basket.add_product(product, 1)
-    if voucher:
+    if voucher or product.get_product_class().name == ENROLLMENT_CODE_PRODUCT_CLASS_NAME:
         for v in basket.vouchers.all():
             basket.vouchers.remove(v)
-        basket.vouchers.add(voucher)
-        Applicator().apply(basket, request.user, request)
-        logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
+        if voucher:
+            basket.vouchers.add(voucher)
+            Applicator().apply(basket, request.user, request)
+            logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
 
     affiliate_id = request.COOKIES.get(settings.AFFILIATE_COOKIE_KEY)
     if affiliate_id:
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=887

ecommerce/extensions/basket/utils.py
```diff
@@ -34,9 +34,9 @@ def prepare_basket(request, product, voucher=None):
     basket = Basket.get_basket(request.user, request.site)
     basket.flush()
     basket.add_product(product, 1)
-    if voucher:
-        for v in basket.vouchers.all():
-            basket.vouchers.remove(v)
+    for v in basket.vouchers.all():
+        basket.vouchers.remove(v)
+    if product.get_product_class().name != ENROLLMENT_CODE_PRODUCT_CLASS_NAME and voucher:
```
mattdrayer: Please update the docstring to reflect the new workflow semantics

ecommerce/extensions/basket/utils.py
```diff
@@ -34,9 +34,9 @@ def prepare_basket(request, product, voucher=None):
     basket = Basket.get_basket(request.user, request.site)
     basket.flush()
     basket.add_product(product, 1)
-    if voucher:
-        for v in basket.vouchers.all():
-            basket.vouchers.remove(v)
+    for v in basket.vouchers.all():
```
mattdrayer: It seems to me that we only want to wipe out the basket voucher if we're working in the enrollment code context?

mjfrey: Only remove all vouchers from basket if the product class is enrollment_code.

vkaracic: No, this part of the code was here before enrollment codes were introduced. It's here to ensure that you can't apply more than one voucher per basket.

mattdrayer: but if voucher is None and there is already a voucher on the basket, do we want to remove the existing voucher?

vkaracic: I moved the check for voucher back.

ecommerce/extensions/basket/utils.py
```diff
@@ -37,9 +37,10 @@ def prepare_basket(request, product, voucher=None):
     if voucher:
         for v in basket.vouchers.all():
             basket.vouchers.remove(v)
-        basket.vouchers.add(voucher)
-        Applicator().apply(basket, request.user, request)
-        logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
+        if product.get_product_class().name != ENROLLMENT_CODE_PRODUCT_CLASS_NAME:
```
mattdrayer: So hopefully I am not leading you around in circles on this one, but I think we need to do the enrollment code check first.  IF we are in an enrollment code context, remove any/all existing vouchers, do not apply a new voucher, if somehow specified.  If we are NOT in an enrollment code context, we remove any/all vouchers IF a new voucher is specified.

Unless I am crazy, which is possible.

vkaracic: I'm confused, how is that different from what we have now?

mattdrayer: Right now we have:

```
IF new voucher:
   clear all vouchers
   if not enrollment code, add new voucher
```

I think it should be:

```
if enrollment code:
    clear all vouchers
elif new voucher:
    clear all vouchers
    add new voucher
else:
    do nothing
```

mattdrayer: But maybe that is faulty logic, I am open to changing my mind!

mattdrayer:
Squash and merge!!!  👍

PR #888, opened 2016-08-10
Fixes

vkaracic:
Instead of creating two separate PR's I'm doing this in one with two separate commits. This PR contains:
- adding seats back only for single course coupons, because that seat is displayed in the details page
- a couple of fixes to the https://github.com/edx/ecommerce/commit/877e7b4ad3aa918645c5d4a14a8478447a23aca2 changes:
  - changes to the test, because the current one won't work for running tests with `DISABLE_MIGRATIONS`
  - changes to the `prepare_basket` logic, now if a product that is an enrollment code and a voucher are passed to that function, the voucher is not applied.

@mjfrey @mattdrayer 
(sorry for not finishing this yesterday, but I didn't have any more time to work on it)

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -447,6 +447,7 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     quantity = serializers.SerializerMethodField()
     start_date = serializers.SerializerMethodField()
     voucher_type = serializers.SerializerMethodField()
+    seats = serializers.SerializerMethodField()
 
     def retrieve_benefit(self, obj):
         """Helper method to retrieve the benefit from voucher. """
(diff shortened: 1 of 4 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=888&repo=edx%2Fecommerce

ecommerce/extensions/api/serializers.py
```diff
@@ -447,6 +447,7 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     quantity = serializers.SerializerMethodField()
     start_date = serializers.SerializerMethodField()
     voucher_type = serializers.SerializerMethodField()
+    seat = serializers.SerializerMethodField()
```
mjfrey: Is this a problem that this is called seat and not seats?

vkaracic: Fixed!

mjfrey:
👍

PR #890, opened 2016-08-10
Add missing basket_addition call.  Handle call to post_checkout with no user=

PaulWattenberger:
Add missing basket_addition call.  
Handle signal send to post_checkout with no user= or request=

### DIFF
--- a/ecommerce/sailthru/signals.py
+++ b/ecommerce/sailthru/signals.py
@@ -17,7 +17,8 @@ basket_addition = get_class('basket.signals', 'basket_addition')
 
 @receiver(post_checkout)
 @silence_exceptions("Failed to call Sailthru upon order completion.")
-def process_checkout_complete(sender, order=None, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
+def process_checkout_complete(sender, order=None, user=None, request=None,  # pylint: disable=unused-argument
+                              response=None, **kwargs):  # pylint: disable=unused-argument
     """Tell Sailthru when payment done.
 
     Arguments:
@@ -26,6 +27,10 @@ def process_checkout_complete(sender, order=None, request=None, user=None, **kwa
 
     if not waffle.switch_is_active('sailthru_enable'):
         return
+
+    message_id = None
+    if request:
+        message_id = request.COOKIES.get('sailthru_bid')
 
     # loop through lines in order
     #  If multi product orders become common it may be worthwhile to pass an array of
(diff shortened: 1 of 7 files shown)


PaulWattenberger:
@clintonb Clinton, can you review.  When I ran on stage, i discovered that we aren't calling the basket_addition oscar signal so I had to add a call

ecommerce/sailthru/signals.py
```diff
@@ -44,16 +45,19 @@ def process_checkout_complete(sender, order=None, request=None, user=None, **kwa
         # figure out course url
         course_url = _build_course_url(course_id)
 
+        # get current request
+        request = threadlocals.get_current_request()
```
clintonb: `threadlocals` is evil and I wish it never existed. Why is it being used here?

PaulWattenberger: Copied from https://github.com/edx/ecommerce/blob/e61b0c238cdadaa13ad775b77cb9ce4da581bb85/ecommerce/extensions/checkout/signals.py

Actually, the request is supposed to be passed in the post_checkout signal call, but a couple of the calls we added, don't include it and I didn't want to change them if I didn't have to...

clintonb: Let's do this the right way, without `threadlocals`.

PaulWattenberger:
@clintonb

ecommerce/sailthru/signals.py
```diff
@@ -27,6 +27,12 @@ def process_checkout_complete(sender, order=None, request=None, user=None, **kwa
     if not waffle.switch_is_active('sailthru_enable'):
         return
 
+    site_code = None
+    message_id = None
+    if request:
+        site_code = request.site.siteconfiguration.partner.short_code
```
clintonb: Use `order.site` instead of `request.site`.

ecommerce/sailthru/signals.py
```diff
@@ -17,7 +17,7 @@
 
 @receiver(post_checkout)
 @silence_exceptions("Failed to call Sailthru upon order completion.")
-def process_checkout_complete(sender, order=None, request=None, user=None, **kwargs):  # pylint: disable=unused-argument
+def process_checkout_complete(sender, order=None, request=None, **kwargs):  # pylint: disable=unused-argument
```
clintonb:
``` python
post_checkout = Signal(providing_args=["order", "user", "request", "response"])
```

This signal should list each of the above args as kwargs.

clintonb:
👍 (once comments addressed, commits squashed). Remember to delete your branch after merging.

PR #891, opened 2016-08-11
SOL-1991: Create Management Command for adding Site and SiteThemes

saleem-latif:
Hi @mattdrayer
This PR is ready for review, It adds a management command for adding Site and SiteTheme. I have also moved tests for management commands into `management/commands/tests` directory as suggested by @mattdrayer .

@clintonb , @douglashall -- FYI

### DIFF
--- a/docs/theming.rst
+++ b/docs/theming.rst
@@ -130,6 +130,50 @@ Disabling a Theme
 -----------------
 Theme can be disabled by removing its corresponding ``Site Theme`` entry using django admin.
 
+---------------------------------------
+Creating or Updating Site and SiteTheme
+---------------------------------------
+If you have already setup ``COMPREHENSIVE_THEME_DIRS`` then you can use management command for adding
+``Site`` and ``SiteTheme`` directly from the terminal.
+
+.. code-block:: Bash
+
+    python manage.py create_or_update_site_theme --site-domain=localhost:8002 --site-name=localhost:8002 --site-theme=my-theme
+
+``create_or_update_site_theme`` accepts the following optional arguments
+
+    :--settings: settings file to use, ``default: ecommerce.settings.devstack``
+
+        .. code-block:: Bash
+
+            python manage.py create_or_update_site_theme --settings=ecommerce.settings.production
+
+    :--site-id: id of the site that you want to update
+
+        .. code-block:: Bash
+
+            # update domain of the site with id 1 and add a new theme ``my-theme`` for this site
+            python manage.py create_or_update_site_theme --site-id=1 --site-domain=my-theme.localhost:8002 --site-name=my-theme.localhost:8002 --site-theme=my-theme
+
+    :--site-domain: domain of the site to be created,
+
+        .. code-block:: Bash
+
+            python manage.py create_or_update_site_theme --site-domain=localhost:8002 --site-theme=my-theme
+
+    :--site-name: Name of the site to be created, ``default: ''``
+
+        .. code-block:: Bash
+
+            python manage.py create_or_update_site_theme --site-domain=localhost:8002 --site-name=localhost:8002 --site-theme=my-theme
+
+    :--site-theme: theme dir for the new theme,
+
+        .. code-block:: Bash
+
+            python manage.py create_or_update_site_theme --site-domain=localhost:8002 --site-name=localhost:8002 --site-theme=my-theme
+
+
 --------------------
 Compiling Theme Sass
 --------------------
--- a/ecommerce/theming/management/commands/create_or_update_site_theme.py
+++ b/ecommerce/theming/management/commands/create_or_update_site_theme.py
@@ -1,0 +1,68 @@
+"""
+Creates or updates a Site and Site Theme.
+"""
+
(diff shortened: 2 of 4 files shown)


ecommerce/theming/tests/test_commands.py
```diff
@@ -6,9 +6,11 @@
 from django.conf import settings
 from django.test import TestCase, override_settings
 from django.core.management import call_command, CommandError
```
mattdrayer, line 8: I've observed two different patterns with test locations when it comes to management commands -- this one, where a single `test_commands.py` module is included in the main `tests` directory for the app, and the one that I prefer which is to include a `tests` folder in each module directory, with dedicated test modules for each command, eg: `theming/management/commands/tests/test_create_or_update_site_theme.py` -- so let's go with the latter.

ecommerce/theming/tests/test_commands.py
```diff
@@ -155,3 +157,74 @@ def test_skip_collect(self):
...
+        self.assertEqual(site_theme.theme_dir_name, "test")
+
+    def test_update_site(self):
+        """
+        Test that site is updated properly if site-id belongs to an existing site.
+        """
+        # Create a site to update
+        site = Site.objects.create(domain="test.localhost", name="Test Site")
+        call_command(
+            "update_assets",
```
mattdrayer: If this is a test of the "update_assets" command it should be moved into its own `test_update_assets.py` module per the above comment.

ecommerce/theming/management/commands/create_or_update_site_theme.py
```diff
@@ -0,0 +1,67 @@
...
+                            help='Name of the theme to apply to the site.')
+
+    def handle(self, *args, **options):
+        site_id = options.get('site_id')
+        site_domain = options.get('site_domain')
+        site_name = options.get('site_name')
+        site_theme = options.get('site_theme')
+
+        try:
+            site = Site.objects.get(id=site_id)
```
mattdrayer, line 51: If site_id is not specified as an argument, will it be treated as a None value?  If this is the case, will the exception be DoesNotExist or something else?  I'm not sure of Django's exception handling behavior when it comes to `objects.get(id=None)`

ecommerce/theming/management/commands/create_or_update_site_theme.py
```diff
@@ -0,0 +1,67 @@
...
+            site, site_created = Site.objects.get_or_create(domain=site_domain)
+            if site_created:
+                logger.info('Site created with domain %s', site_domain)
+
+        site.domain = site_domain
+        if site_name:
+            site.name = site_name
+        site.save()
+
+        SiteTheme.objects.update_or_create(
```
mattdrayer: Does it matter if the specified theme does not exist?  Seems like we should perform a check?

ecommerce/theming/management/commands/create_or_update_site_theme.py
saleem-latif, line 51: @mattdrayer `Site.objects.get(id=None)` raises `DoesNotExist` error.

```
>>> Site.objects.get(id=None)
  File "<console>", line 1, in <module>
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/db/models/query.py", line 334, in get
    self.model._meta.object_name
DoesNotExist: Site matching query does not exist.
```

ecommerce/theming/management/commands/create_or_update_site_theme.py
saleem-latif: @mattdrayer if theme does not exist, then it will be created, I have also added logging message to show whether theme was created or updated. Is that what you had in mind ?

saleem-latif:
@mattdrayer I have incorporated your feedback, kindly take another look

clintonb:
Update the documentation in the docs directory.

saleem-latif:
@clintonb I have updated documentation, kindly take a look

docs/theming.rst
```diff
@@ -130,6 +130,50 @@ Disabling a Theme
 -----------------
 Theme can be disabled by removing its corresponding ``Site Theme`` entry using django admin.
 
+---------------------------------------
+Creating or Updating Site and SiteTheme
```
clintonb, line 134: @srpearce please review.

mattdrayer:
PR looks good!  Thanks very much for reworking the `test_commands.py` module -- it is much better now.  👍

mattdrayer:
@saleem-latif looks like this PR is now in need of a rebase.  We're also still waiting on a parse from @srpearce and final comments from @clintonb or @douglashall

mattdrayer:
@saleem-latif please rebase this PR and merge -- I think we've let it hang around for comments long enough

saleem-latif:
Going ahead with the merge now.

PR #894, opened 2016-08-12
Add a list of email domains to an offer for validation at redemption.

vkaracic:
This PR introduces a new way of restricting coupons by email domains of users. A new input field is added in the coupon creation form:
screenshot from 2016-08-12 15-19-25

`ConditionalOffer` model has been extended to add the `email_domain` character field, and to extend the `is_condition_satisfied` method to check if the basket owners email domain is in the email domain list for the offer. This is usually done by proxying the `Condition` model and extending its `is_satisfied` method, but, for some reason, the proxy needs to be unique: https://github.com/django-oscar/django-oscar/blob/1.1.1/src/oscar/apps/offer/abstract_models.py#L640-L641 and we can't have a unique proxy because every condition is different for coupons (different range or products.) Therefor I decided on doing it this way.

https://openedx.atlassian.net/browse/SOL-1967

### DIFF
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -177,6 +177,9 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
         if not valid_voucher:
             return render(request, template_name, {'error': msg})
 
+        if not voucher.offers.first().is_email_valid(request.user.email):
+            return render(request, template_name, {'error': _('You are not eligible to use this coupon.')})
+
         if request.user.is_user_already_enrolled(request, product):
             return render(request, template_name, {'error': _('You are already enrolled in the course.')})
 
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -4,10 +4,31 @@ import hashlib
 from django.conf import settings
 from django.core.cache import cache
 from django.db import models
-from oscar.apps.offer.abstract_models import AbstractRange
+from oscar.apps.offer.abstract_models import AbstractConditionalOffer, AbstractRange
 from threadlocals.threadlocals import get_current_request
 
 from ecommerce.coupons.utils import get_seats_from_query
+
+
+class ConditionalOffer(AbstractConditionalOffer):
+    email_domains = models.CharField(max_length=255, blank=True, null=True)
+
+    def is_email_valid(self, email):
+        """
+        Check if the email is within the email_domains if email_domains are set,
+        else return True.
+        """
+        domain = email.split('@')[1]
+        return domain in self.email_domains if self.email_domains else True  # pylint: disable=unsupported-membership-test
+
+    def is_condition_satisfied(self, basket):
+        """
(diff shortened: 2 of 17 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=894&repo=edx%2Fecommerce

ecommerce/coupons/views.py
```diff
@@ -177,6 +177,11 @@ def get(self, request):
         if not valid_voucher:
             return render(request, template_name, {'error': msg})
 
+        user_email_domain = request.user.email.split('@')[1]
+        email_domains = voucher.offers.first().email_domains
+        if email_domains and user_email_domain not in email_domains:
+            return render(request, template_name, {'error': _('You are not eligible for this coupon.')})
```
mjfrey: to use this coupon.

mjfrey:
@marjev please take a look.

ecommerce/coupons/views.py
```diff
@@ -177,6 +177,11 @@ def get(self, request):
         if not valid_voucher:
             return render(request, template_name, {'error': msg})
 
+        user_email_domain = request.user.email.split('@')[1]
+        email_domains = voucher.offers.first().email_domains
+        if email_domains and user_email_domain not in email_domains:
```
marjev: Should we move this logic to Conditional Offer so that an instance of that class returns whether or not an email domain is valid or not? That way we'll hide the business logic from views and simply call it here, something like this:
`if not voucher.offers.first().IsEmailValid(request.user.email)`

ecommerce/extensions/offer/models.py
```diff
@@ -4,12 +4,28 @@
...
+class ConditionalOffer(AbstractConditionalOffer):
+    email_domains = models.CharField(max_length=255, blank=True, null=True)
+
+    def is_condition_satisfied(self, basket):
+        """
+        In addition to Oscar's check to see if the condition is satisfied,
+        a check for if basket owners email domain is within the allowed email domains.
+        """
+        if self.email_domains:
+            user_email_domain = basket.owner.email.split('@')[1]
```
marjev: Consider having a utility method for this. DRY, since you're doing this in another place.

ecommerce/extensions/voucher/utils.py
```diff
@@ -119,7 +119,8 @@ def _get_info_for_coupon_report(coupon, voucher):
         'Discount Amount': discount_amount,
         'Invoiced Amount': invoiced_amount,
         'Note': note,
-        'Price': price
+        'Price': price,
+        'Email Domains': offer.email_domains
```
marjev: nitpick: alphabetize

marjev:
I finished my first pass. Tag me again once you've addressed the comments.

vkaracic:
@marjev comments addressed!

marjev:
👍

mjfrey:
👍

PR #899, opened 2016-08-16
Move the security fix into Eucalyptus

schenedx:
@nedbat @rlucioni Please review. This change will move the security fix into Eucalyptus
@adampalay FYI


nedbat:
:+1: looks the same as the change in the security announcement :)

rlucioni:
👍

PR #901, opened 2016-08-16
Remove attempts to reverse catalogue URLs from templates

rlucioni:
We don't use URLs from the catalogue app. ECOM-5204.

@clintonb please review.


clintonb:
👍

PR #903, opened 2016-08-18
[SOL-2033] Change voucher list to raw ID list in basket details admin page.

vkaracic:
The basket details page in Django admin was displaying all the vouchers in a list, where the selected one meant it was the one applied, which caused a large load-time because of the large number of vouchers.
The code in this PR changes the `vouchers` field to a `raw_id_field`, changing the list of all vouchers to a list of ID's of only the applied vouchers effectively speeding up the load.

https://openedx.atlassian.net/browse/SOL-2033
@clintonb (thanks for tip)

### DIFF
--- a/ecommerce/extensions/basket/admin.py
+++ b/ecommerce/extensions/basket/admin.py
@@ -19,6 +19,7 @@ class PaymentProcessorResponseInline(admin.TabularInline):
 
 @admin.register(Basket)
 class BasketAdminExtended(BasketAdmin):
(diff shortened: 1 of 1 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=903&repo=edx%2Fecommerce

clintonb:
👍

PR #906, opened 2016-08-22
[SOL-2004] Display seats instead of courses in catalog preview API endpoint.

vkaracic:
The dynamic catalog preview in the coupon form would display courses instead of seats which means that it would not display a credit course which has a verified seat so there would be a discrepancy between the seats displayed in the offer landing page (with the verified seat from credit course) and the preview window (without that seat).
This PR contains code that changes the preview to show those seats instead of courses.

https://openedx.atlassian.net/browse/SOL-2004
_Yet again don't mind the name of the branch_


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=906&repo=edx%2Fecommerce

vkaracic:
@mjfrey please review.

marjev:
LGTM 👍

mjfrey:
👍

PR #907, opened 2016-08-23
Rename start and end datetime parameters; Refactor update dictionary creation method;

marjev:
2nd cleanup PR:

API endpoints should take start_datetime and end_datetime and add a 'name' key with title value in request data so that we can remove UPDATABLE_VOUCHER_FIELDS. We also don't need `retrieve_invoice_data` method since we can reuse `create_update_data_dict`.

@mjfrey, @vkaracic, @iivic can you guys please take a look at this?

This is a part of [SOL-1890](https://openedx.atlassian.net/browse/SOL-1890).


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=907

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -394,7 +370,7 @@ def test_update_title(self):
         """Test updating a coupon's title."""
         data = {
             'id': self.coupon.id,
-            'title': 'New title'
+            'name': 'New title'
```
iivic: Why did you rename this field into name?

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -371,19 +340,24 @@ def update(self, request, *args, **kwargs):
...
-        if request_data_key in request_data:
-            value = request_data.get(request_data_key)
-            update_dict[update_dict_key] = prepare_course_seat_types(value) \
-                if update_dict_key == 'course_seat_types' else value
+        update_dict = {}
+
+        for field in fields:
+            if field in data:
+                value = prepare_course_seat_types(data.get(field)) if field == 'course_seat_types' else data.get(field)
+                update_dict[field.replace('invoice_', '')] = value
```
iivic, line 360: Why is field.replace('invoice_', '') required here?

marjev: Backbone uses `invoice_` prefix to mark the Invoice model fields. So, removing that prefix is necessary here to ensure that we can easily update associated Invoice object.

ecommerce/static/js/models/coupon_model.js
```diff
@@ -237,11 +237,17 @@ define([
                     options.data = JSON.stringify(this.toJSON());
                 } else {
                     if (_.has(attributes, 'start_date')) {
-                        attributes.start_date = moment.utc(attributes.start_date);
+                        attributes.start_datetime = moment.utc(attributes.start_date);
+                        delete attributes.start_date;
```
vkaracic: Why is this being deleted?

ecommerce/extensions/api/v2/tests/views/test_coupons.py
marjev: If the data contains `title`, it will only update title of the coupon product, since Product has title field. Since this is testing updating the voucher names, we have to provide name attribute for this to work most naturally. I will, however, update the docstring and name of the test method since it's not describing the real test scenario.

iivic: """Test updating a coupon's title.""" That's why it didn't make sense to me. Once you've done the update 👍

ecommerce/static/js/models/coupon_model.js
marjev: I'll remove that line. TBH, I don't remember why I placed that there, since this was done so long ago.

vkaracic:
:+1:

PR #908, opened 2016-08-23
Add an OEP-2 compliant openedx.yaml file

cpennington:
This adds an `openedx.yaml` file, as described by OEP-2:
http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html

The data in this file was transformed from the contents of
edx/repo-tools-data:repos.yaml

### DIFF
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -1,0 +1,9 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: ecom
+oeps: {}
+openedx-release: {ref: master}
+owner: edx/ecommerce
+tags: [ecom]
+track-pulls: true


cpennington:
@edx/ecommerce: Another one for you folks.

openedx.yaml
```diff
@@ -0,0 +1,9 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: ecom
+oeps: {}
+openedx-release: {ref: master}
+owner: edx/ecommerce
+tags: [dedx]
```
rlucioni: Any reason to keep this name around? I think we'd prefer use of an "ecom" tag.

openedx.yaml
```diff
@@ -0,0 +1,9 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: ecom
```
rlucioni, line 4: We use the name "ecommerce" everywhere, so that seems like a more appropriate nickname.

rlucioni:
👍 please squash commits before merging.

PR #909, opened 2016-08-23
Use catalog service for course info.

mjfrey:
Migrate single course coupons to pull course info from catalog service rather than LMS.

### DIFF
--- a/ecommerce/courses/utils.py
+++ b/ecommerce/courses/utils.py
@@ -3,9 +3,6 @@ import hashlib
 from django.conf import settings
 from django.core.cache import cache
 from django.utils.translation import ugettext_lazy as _
-from edx_rest_api_client.client import EdxRestApiClient
-
-from ecommerce.core.url_utils import get_lms_url
 
 
 def mode_for_seat(product):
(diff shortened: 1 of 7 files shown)


mjfrey:
@vkaracic @marjev @mattdrayer Please review.

ecommerce/courses/utils.py
```diff
@@ -23,14 +20,15 @@ def mode_for_seat(product):
...
 
 
-def get_course_info_from_lms(course_key):
-    """ Get course information from LMS via the course api and cache """
-    api = EdxRestApiClient(get_lms_url('api/courses/v1/'))
-    cache_key = 'courses_api_detail_{}'.format(course_key)
+def get_course_info_from_catalog(site, course_key):
+    """ Get course information from catalog service and cache """
+    api = site.siteconfiguration.course_catalog_api_client
+    partner = site.siteconfiguration.partner.short_code
```
mattdrayer: I'd rename this to `partner_short_code`

ecommerce/extensions/basket/views.py
```diff
@@ -106,12 +106,13 @@ def get_context_data(self, **kwargs):
...
-                course = get_course_info_from_lms(course_key)
-                image_url = get_lms_url(course['media']['course_image']['uri'])
-                short_description = course['short_description']
-                course_name = course['name']
+                course = get_course_info_from_catalog(self.request.site, course_key)
+                description = course['short_description']
+                image_url = course['image']['src']
+                short_description = description if description else ''
+                course_name = course['title']
             except (ConnectionError, SlumberBaseException, Timeout):
```
mattdrayer, line 114: The try/except should cover only what needs to be protected -- in this case `get_course_info_from_catalog`

ecommerce/courses/utils.py
```diff
@@ -23,14 +20,15 @@ def mode_for_seat(product):
...
+    cache_key = 'courses_api_detail_{}{}'.format(course_key, partner)
     cache_hash = hashlib.md5(cache_key).hexdigest()
     course = cache.get(cache_hash)
     if not course:  # pragma: no cover
-        course = api.courses(course_key).get()
+        course = api.course_runs(course_key).get(partner=partner)
```
mattdrayer: Is this a course or a course run?  I see we are calling the variable `course` but loading the data from the `course_runs` api

mattdrayer:
Just a couple comments @mjfrey -- overall the changes LGTM -- nice work!  👍

ecommerce/courses/utils.py
mjfrey: we get the course_run for the course.  That is what is purchasable.

mattdrayer: Right, so maybe we should rename the course variable to course_run here

vkaracic:
:+1:

marjev:
👍

PR #910, opened 2016-08-24
[SOL-2038] Add a check if email_domains are in coupon update data before updating it

vkaracic:
Email domains would get deleted if they are not altered during an update because Backbone would not send email domains as a parameter in the update.
This PR contains code that checks if the email domains are in the update request parameters before updating them.

https://openedx.atlassian.net/browse/SOL-2038
@mjfrey @marjev please review.


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=910&repo=edx%2Fecommerce

marjev:
👍

mjfrey:
👍

PR #912, opened 2016-08-25
Fixed course save button not redirecting

iivic:
@mjfrey  @marjev  @vkaracic 
Easy fix. The problem was detecting the issue.
I had to remove the store.reset() from form_view since the coupon model didn't work correctly. I moved that line in the coupon_model save function so the DuplicateId error wouldn't appear when creating a new coupon. 
I didn't encounter DuplicateId error when creating a new course so this fix should be sufficient. However further testing is needed in both cases to be sure.
Please review and test creating and editing courses and coupons I did but just in case :)

### DIFF
--- a/ecommerce/static/js/models/course_model.js
+++ b/ecommerce/static/js/models/course_model.js
@@ -301,7 +301,7 @@ define([
              * to external systems, in an atomic fashion. This is desirable to
              * avoid synchronization issues across systems.
              */
-            save: function (options) {
+            save: function (attributes, options) {
                 var verificationDeadline,
                     honorMode,
                     honorSeatClass,
--- a/ecommerce/static/js/models/coupon_model.js
+++ b/ecommerce/static/js/models/coupon_model.js
@@ -219,6 +219,9 @@ define([
             },
 
             save: function (attributes, options) {
+                // Remove all saved models from store, which prevents Duplicate id errors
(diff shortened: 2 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=912

ecommerce/static/js/models/course_model.js
```diff
@@ -301,7 +301,7 @@ define([
              * to external systems, in an atomic fashion. This is desirable to
              * avoid synchronization issues across systems.
              */
-            save: function (options) {
+            save: function (attributes, options) {
```
vkaracic, line 304: Why this change?

iivic: Because the save call is sending two parameters: attributes that should be changed and options ( success, error, etc. ). Without this change the course save button doesn't call any callback function which should be called.

vkaracic: So this is the reason why there was no redirect after course save?

iivic: Exactly.

vkaracic:
:+1:

marjev:
👍

PR #915, opened 2016-08-29
Remove remaining large calls to catalog service.

mjfrey:
Remove last remaining calls to catalog service that could be very large.  We now page the results.


mjfrey:
@vkaracic @marjev

marjev:
👍

vkaracic:
Why is this needed to be removed?

mjfrey:
It is not used.  Plus we should not be returning all products for a catalog query.  This can be a very large set and is not needed.

vkaracic:
Then how should all the products of an offer be fetched?

mjfrey:
We won't for an offer with a query.  That function is not able to handle async responses. so we just return none for query style offers.

vkaracic:
So the offer landing page won't show any products for query coupons?

mjfrey:
From what I understand -- offer landing page does not use this.

vkaracic:
Offer landing page gets its information from `/api/v2/vouchers/offers/` and that endpoint gets its products from a function that uses range's `all_products()`: https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/v2/views/vouchers.py#L69

I'm not sure if there is a fallback that kicks in when the products are missing. Have you tested this with a query coupon? Does the information show up on the offer landing page even though the seats are not retrieved from the query?

PR #916, opened 2016-08-29
Check for top level image null

mjfrey:
Do not assume a top level image attribute exits in the JSON.


mjfrey:
@mattdrayer

mattdrayer:
Stupid magic tuple... 👍

PR #917, opened 2016-08-29
Mjfrey/revert catalog info

mjfrey:
Revert use of catalog service for all course info.  Not all sites are using catalog service and breaks offers/basket pages.


mjfrey:
@mattdrayer

mattdrayer:
👍

PR #918, opened 2016-08-30
Check existence of JSON data from both catalog and LMS course API

mjfrey:
Always check if a JSON attribute is avail before accessing it. We can not assume the data is always correct.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_vouchers.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_vouchers.py
@@ -352,3 +352,24 @@ class VoucherViewOffersEndpointTests(
             'title': course.name,
             'voucher_end_date': voucher.end_datetime
         })
+
+    def test_get_course_offer_verify_null_fields(self):
+        """ Verify that the course offers data is properly formatted. """
+        benefit = BenefitFactory()
+        course, seat = self.create_course_and_seat()
+        course_info = {
+        }
+        stock_record = seat.stockrecords.first()
+        voucher = VoucherFactory()
+
+        offer = VoucherViewSet().get_course_offer_data(
+            benefit=benefit,
+            course=course,
+            course_info=course_info,
+            is_verified=True,
+            stock_record=stock_record,
+            voucher=voucher
+        )
+
+        self.assertEqual(offer['image_url'], '')
+        self.assertEqual(offer['course_start_date'], '')
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -397,3 +397,18 @@ class BasketSummaryViewTests(CourseCatalogTestMixin, LmsApiMockMixin, ApiMockMix
         self.assertEqual(response.status_code, 200)
         cached_course_after = cache.get(cache_hash)
         self.assertEqual(cached_course_after['name'], self.course.name)
+
+    def test_empty_course_api_response(self):
+        """ Check to see if we can handle empty response from the course api """
+        seat = self.create_seat(self.course)
+        self.create_basket_and_add_product(seat)
+        course_id = 'course-v1:test+test+test'
+        course_url = get_lms_url('api/courses/v1/courses/{}/'.format(course_id))
+        course_info_json = json.dumps('{}')
+        httpretty.register_uri(httpretty.GET, course_url, body=course_info_json, content_type='application/json')
+
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 200)
+        line_data = response.context['formset_lines_data'][0][1]
+        self.assertEqual(line_data.get('image_url'), None)
+        self.assertEqual(line_data.get('course_short_description'), None)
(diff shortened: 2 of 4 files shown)


mjfrey:
@mattdrayer

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -352,3 +352,37 @@ def test_get_course_offer_data(self):
             'title': course.name,
             'voucher_end_date': voucher.end_datetime
         })
+
+    def test_get_course_offer_null_image_data(self):
```
mattdrayer: Since we're also testing the null start date scenario here, it probably makes sense to rename this method to something more generic like `test_get_course_offer_verify_null_fields` or something along those lines.  Then you can just add tack an in-line comment onto the end of each assertion.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -352,3 +352,37 @@ def test_get_course_offer_data(self):
...
+
+        self.assertDictEqual(offer, {
+            'benefit': {
+                'type': benefit.type,
+                'value': benefit.value
+            },
+            'contains_verified': True,
+            'course_start_date': '',
+            'id': course.id,
+            'image_url': '',
```
mattdrayer: I'd append a comment to this line to highlight that this is one of the pertinent checks.  Another way to do this would be to have specific assertions for each field we're concerned with -- it's valid to do the full-on dictionary check but then it's not really clear what we're asserting.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -397,3 +397,18 @@ def test_cached_course(self):
         self.assertEqual(response.status_code, 200)
         cached_course_after = cache.get(cache_hash)
         self.assertEqual(cached_course_after['name'], self.course.name)
+
+    def test_empty_course_api_response(self):
+        seat = self.create_seat(self.course)
+        self.create_basket_and_add_product(seat)
+        course_info_json = json.dumps('{}')
```
mattdrayer: Nit: Move below the course_url setup on line 406 so it's closer to where it's being used.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -397,3 +397,18 @@ def test_cached_course(self):
...
+        course_id = 'course-v1:test+test+test'
+        course_url = get_lms_url('api/courses/v1/courses/{}/'.format(course_id))
```
mattdrayer, line 406: Is this the same as self.path below?

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -397,3 +397,18 @@ def test_cached_course(self):
         self.assertEqual(response.status_code, 200)
         cached_course_after = cache.get(cache_hash)
         self.assertEqual(cached_course_after['name'], self.course.name)
+
+    def test_empty_course_api_response(self):
```
mattdrayer, line 401: It's not clear to me what is being validated based on the name of this test -- if the name is correct maybe add a docstring describing the use case(s) being covered?

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -397,3 +397,18 @@ def test_cached_course(self):
...
+        seat = self.create_seat(self.course)
+        self.create_basket_and_add_product(seat)
+        course_info_json = json.dumps('{}')
+        course_id = 'course-v1:test+test+test'
+        course_url = get_lms_url('api/courses/v1/courses/{}/'.format(course_id))
+        httpretty.register_uri(httpretty.GET, course_url, body=course_info_json, content_type='application/json')
+
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.context['formset_lines_data']), 1)
```
mattdrayer: Is it really important for this particular test to verify that only one row is being returned?  Does it matter?

ecommerce/extensions/basket/views.py
```diff
@@ -107,8 +107,8 @@ def get_context_data(self, **kwargs):
             short_description = None
             try:
                 course = get_course_info_from_lms(course_key)
-                image_url = get_lms_url(course['media']['course_image']['uri'])
-                short_description = course['short_description']
+                image_url = course.get('media', '{}').get('image', '{}').get('raw', '')
+                short_description = course.get('short_description', '')
                 course_name = course['name']
             except (ConnectionError, SlumberBaseException, Timeout):
```
mattdrayer, line 113: The try is only protecting against errors stemming from `get_course_info_from_lms`, so the except should be moved up to that line -- the rest can all come after.

mattdrayer:
A couple clarification requests and some nits, otherwise LGTM -- 👍

ecommerce/extensions/basket/tests/test_views.py
mjfrey: no

PR #919, opened 2016-08-30
Really fix missing JSON data.

mjfrey:
Don't get fancy and check if JSON attributes exist before trying to get them.

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -107,7 +107,10 @@ class BasketSummaryView(BasketView):
             short_description = None
             try:
                 course = get_course_info_from_lms(course_key)
-                image_url = course.get('media', '{}').get('image', '{}').get('raw', '')
+                try:
+                    image_url = course['media']['image']['raw']
+                except (KeyError, TypeError):
+                    image_url = ''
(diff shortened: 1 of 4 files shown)


mjfrey:
@douglashall

ecommerce/extensions/basket/views.py
```diff
@@ -107,7 +107,12 @@ def get_context_data(self, **kwargs):
             short_description = None
             try:
                 course = get_course_info_from_lms(course_key)
-                image_url = course.get('media', '{}').get('image', '{}').get('raw', '')
+                image = course.get('media')
+                if image:
+                    image = image.get('image')
+                    if image:
+                        image = image.get('raw', '')
+                image_url = image
```
douglashall: Should `image_url` get set even if you don't fall all the way through the if hierarchy above?

For instance, do you ever want `image_url` to get set to the result of `course.get('media')`?

douglashall: I would favor something like:

```
try:
    image_url = course['media']['image']['raw']
except (KeyError, TypeError):
    image_url = ''
```

douglashall:
👍 Code looks good. You could probably use some tests to verify the exceptions catch on image lookup if your timetable permits.

PR #921, opened 2016-08-31
[SOL-2046] Remove unnecessary call to Course Catalog Service

marjev:


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=921

marjev:
@mjfrey, @vkaracic Here's a PR that proposes a removal of `all_products()` method call that is made on the voucher offers API endpoint.

This is WIP because I'll have to fix the tests and increase the coverage. I tagged you so that you may take a look at the general idea of this fix and propose any additional changes that we can do to further increase the performance of this endpoint.

vkaracic:
LGTM

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -75,12 +76,12 @@ def offers(self, request):
             return Response(status=status.HTTP_400_BAD_REQUEST)
 
         try:
-            offers_data = self.get_offers(products, request, voucher)
+            offers_data = self.get_offers(request, voucher)
```
mjfrey, line 75: This is still the same issue.  The call to get_offers is using the DEFAULT_CATALOG_PAGE_SIZE.  This is a problem.  We want to only get the first page.

marjev:
@mjfrey Can you take a final look before this is merged?

mjfrey:
@marjev do we want to add the removal of the call to the catalog service in all_products and reduce the default page size?

marjev:
@mjfrey can you simply reopen your PR?

mjfrey:
:+1:

PR #922, opened 2016-08-31
Update basket deletion management command

rlucioni:
Avoid deletion of invoiced baskets, and sleep between each batch deletion. Also do a better job explaining why we used batched deletes. In service of ECOM-5386.

@clintonb @jibsheet please review. FYI @edx/ecommerce.

### DIFF
--- a/ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
+++ b/ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
@@ -4,6 +4,9 @@ Management command that deletes baskets associated with orders.
 These baskets don't have much value once the order is placed, and unnecessarily take up space.
 """
 from __future__ import unicode_literals
+
+import time
+
 from django.core.management import BaseCommand
 from django.db import transaction
 from oscar.core.loading import get_model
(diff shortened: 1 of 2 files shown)


clintonb:
👍

jibsheet:
👍 and I now actually understand how your "delete between" logic works.

ecommerce/extensions/basket/management/commands/delete_ordered_baskets.py
```diff
@@ -15,34 +18,50 @@ class Command(BaseCommand):
...
         parser.add_argument('--commit',
                             action='store_true',
                             dest='commit',
                             default=False,
                             help='Actually delete the baskets.')
 
     def handle(self, *args, **options):
-        queryset = Basket.objects.filter(order__isnull=False)
+        # Only select those baskets linked to an order, and those not linked to an invoice.
+        # TODO: Simplify this query when the foreign key to Basket is removed from Invoice.
```
vkaracic, line 43: @rlucioni is the Basket foreign key ever going to be removed from the Invoice model?

rlucioni: @vkaracic why do you ask? I would touch base with @mjfrey. If it's causing problems, we should talk about how to balance removing it with our other priorities.

vkaracic: No, no problems. I'm just going over all the TODO's left in the code and I don't think that the Basket FK will be removed so I'm thinking on removing the TODO.

rlucioni: I see. What harm is this TODO causing? Consider leaving it here. We've been talking about ways we can use baskets and the ecommerce service more effectively, and I think that this change could be made as part of work we decide to do in that area.

PR #924, opened 2016-08-31
Use Babel directly to format benefit value.

mjfrey:
Calling oscar's currency on the benefit value returns a format that contains line breaks.  This is not what we want and is broken for all locales but english.
Instead use Babel directly with our own format.

### DIFF
--- a/ecommerce/extensions/offer/utils.py
+++ b/ecommerce/extensions/offer/utils.py
@@ -1,9 +1,11 @@
 """Offer Utility Methods. """
 from decimal import Decimal
 
-from django.utils.translation import ugettext_lazy as _
+from babel.numbers import format_currency
+
+from django.conf import settings
+from django.utils.translation import get_language, to_locale, ugettext_lazy as _
 from oscar.core.loading import get_model
-from oscar.templatetags.currency_filters import currency
 
 Benefit = get_model('offer', 'Benefit')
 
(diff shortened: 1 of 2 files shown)


mjfrey:
@rlucioni @jibsheet Please review.  This is a fix for the unicode basket error.  We where using oscar's built in currency function which in turn uses Babel.  But the format that is used is wrong for us.  I changed this to use Babel directly.

ecommerce/extensions/offer/utils.py
```diff
@@ -35,5 +36,8 @@ def format_benefit_value(benefit):
     if benefit.type == Benefit.PERCENTAGE:
         benefit_value = _('{benefit_value}%'.format(benefit_value=benefit_value))
     else:
-        benefit_value = _('{benefit_value}'.format(benefit_value=currency(benefit_value)))
+        converted_benefit = babel.numbers.format_currency(
+            Decimal(benefit.value), 'USD', format=u'#,##0.00',
```
vkaracic: `USD` or any currency for that matter shouldn't be hardcoded anywhere, it's better to use this setting: https://github.com/edx/ecommerce/blob/master/ecommerce/settings/_oscar.py#L97

mjfrey:
@vkaracic

vkaracic:
:+1:

PR #925, opened 2016-08-31
Upgrade gunicorn to 19.6.0

rlucioni:
Includes a note about why gevent has been left at 1.0.2. ECOM-5386.

I've confirmed that this upgrade won't break CI: http://jenkins.edx.org:8080/job/ansible-provision-continuous-integration/31475/.

@jibsheet @edx/ecommerce


jibsheet:
👍

PR #926, opened 2016-09-02
Fixes for coupon form.

vkaracic:
If you select discount as a fixed amount the upper limit is still there and you wouldn't be able to set more than $100 discount. Same in the postpaid invoice.
For the query coupons the field `Valid of courses:` appears even though there is no information for it.

@marjev @mjfrey


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=926

marjev:
LGTM 👍

mjfrey:
👍

PR #928, opened 2016-09-02
Get info from order directly not via the basket

mjfrey:
Go directly to the order lines rather than via basket to get a match.

### DIFF
--- a/ecommerce/extensions/voucher/utils.py
+++ b/ecommerce/extensions/voucher/utils.py
@@ -208,7 +208,7 @@ def generate_coupon_report(coupon_vouchers):
 
     for coupon_voucher in coupon_vouchers:
         coupon = coupon_voucher.coupon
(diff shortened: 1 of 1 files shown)


mjfrey:
FYI @vkaracic

mjfrey:
@mattdrayer @rlucioni can you take a look?

mattdrayer:
This seems like a valid change to me -- I can't see why we would want to traverse the basket model to get from the order to the line, but perhaps @vkaracic has some insight to lend here.

+1 for me and good to merge once all checks/tests are passing.

rlucioni:
👍

vkaracic:
@mattdrayer I too just today found out that you can access the lines in an order directly, can't remember why this was implemented that way. Oscar is a mysterious beast.

:+1:

PR #930, opened 2016-09-05
Change order-basket-lines queries to order-lines.

vkaracic:
There is no need to traverse through baskets when getting the invoice via a line product criteria.
Also the `squash_duplicate_invoices` command is removed because it was ran and not necessary anymore. (was added here https://github.com/edx/ecommerce/pull/884)

@mjfrey please review.


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=930

mjfrey:
:+1:

PR #934, opened 2016-09-08
Remove last remaining bulk calls to catalog service.

mjfrey:
No longer need to call the catalog service and get all results back for a query.  This is done via page requests.


mjfrey:
@vkaracic @marjev

marjev:
👍

vkaracic:
The changes look good to me, I'm just curious if we should leave a comment in the `all_products` that it doesn't include seats from the course discovery?

:+1:

marjev:
@mjfrey I've added all the changes to my PR so that we have all this in a single commit.

Can you verify that this PR can be closed in favor of #937?

PR #849, opened 2016-07-07
[SOL-1953] New receipt endpoint

JoyceZhu:
@mjfrey , current state of the new receipt endpoint

PR #935, opened 2016-09-09
[SOL-2058] Get provider data from LMS

marjev:


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=935

marjev:
@mjfrey Here's the PR for getting provider data from the LMS API. Can you please review this?

@vkaracic this PR modifies already existing method for getting provider data because the new receipt page will rely on this data. We will use provider data in Django view that loads the template managed by Backbone. Please take a look as the 2nd reviewer.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -70,7 +71,7 @@ def test_post_checkout_callback(self):
                 course_title=order.lines.first().product.title,
                 full_name=self.user.get_full_name(),
                 credit_hours=2,
-                credit_provider='Hogwarts',
```
vkaracic, line 73: Awww :cry:

ecommerce/extensions/checkout/utils.py
```diff
@@ -1,37 +1,32 @@
...
-        else:
-            logger.error(
-                'Failed retrieve provider information for %s provider. Provider API returned status code %d. Error: %s',
-                provider_id, response.status_code, response.text)
-            return None
+        return EdxRestApiClient(
+            get_lms_url('api/credit/v1/'),
+            oauth_access_token=access_token
+        ).providers(provider_id).get()
+    except SlumberHttpBaseException:
```
vkaracic: Could these exceptions be grouped together, since the message is pretty much the same?

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -33,8 +33,9 @@ def test_post_checkout_callback(self):
         to fulfill the newly-placed order and send receipt email.
         """
         httpretty.register_uri(
-            httpretty.GET, get_lms_url('api/credit/v1/providers/ASU'),
-            body='{"display_name": "Hogwarts"}',
+            httpretty.GET,
+            get_lms_url('api/credit/v1/providers/ASU/'),
+            body='{"display_name": "Arizona State University"}',
             content_type="application/json"
```
vkaracic: Nit: single quotes.

ecommerce/extensions/checkout/utils.py
```diff
@@ -1,37 +1,32 @@
...
-        response = requests.get(provider_info_url, headers=headers, timeout=timeout)
-        if response.status_code == 200:
-            return response.json()
-        else:
-            logger.error(
-                'Failed retrieve provider information for %s provider. Provider API returned status code %d. Error: %s',
-                provider_id, response.status_code, response.text)
-            return None
+        return EdxRestApiClient(
+            get_lms_url('api/credit/v1/'),
```
vkaracic: https://github.com/edx/ecommerce/blob/master/ecommerce/core/url_utils.py#L11-L13 so there's that.
Perhaps send the request and construct the URL with: https://github.com/edx/ecommerce/blob/master/ecommerce/core/models.py#L183

vkaracic (commented):
I've finished my first pass.

marjev:
@vkaracic Comments have been addressed.

vkaracic (approved):
LGTM

marjev:
I would like someone from @edx/ecommerce to review this as well.

I made a change to the method that calls LMS API endpoint to get the provider data. It's used only on a couple of places and I think I applied the changes correctly.

The reason why I need the provider data is because the new receipt page that's been implemented in #849 requires that. The plan is to have a method for fetching Provider Data and use it to get the data in the Receipt Page Django View to provide that data directly.

Thanks!

ecommerce/extensions/checkout/signals.py
```diff
@@ -75,7 +75,12 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
                 )
                 return
             elif product.get_product_class().name == 'Seat':
-                provider_data = get_provider_data(provider_id)
+                request = threadlocals.get_current_request()
```
clintonb: Please stop using `threadlocals`. Get the site from the order ALWAYS to avoid brand leakage. Pass the user to signal call, and you don't need a request.

ecommerce/extensions/checkout/utils.py
```diff
@@ -1,40 +1,32 @@
...
+from requests.exceptions import ConnectionError, Timeout
+from slumber.exceptions import SlumberHttpBaseException
 
 
 logger = logging.getLogger(__name__)
 
 
-def get_provider_data(provider_id):
+def get_provider_data(access_token, provider_id, site_configuration):
     """Get the provider information for provider id provider.
```
clintonb: Huh? What information are you retrieving?

ecommerce/extensions/checkout/utils.py
```diff
@@ -1,40 +1,32 @@
...
+from edx_rest_api_client.client import EdxRestApiClient
+from requests.exceptions import ConnectionError, Timeout
+from slumber.exceptions import SlumberHttpBaseException
 
 
 logger = logging.getLogger(__name__)
 
 
-def get_provider_data(provider_id):
+def get_provider_data(access_token, provider_id, site_configuration):
```
clintonb: `get_credit_provider_details`

ecommerce/extensions/checkout/utils.py
```diff
@@ -1,40 +1,32 @@
...
     """Get the provider information for provider id provider.
 
     Args:
+        access_token(str): JWT access token
```
clintonb: Add spaces after the variable names.

ecommerce/extensions/checkout/utils.py
```diff
@@ -1,40 +1,32 @@
...
-        logger.exception('Failed to retrieve data for %s provider, connection timeout', provider_id)
+        return EdxRestApiClient(
+            site_configuration.build_lms_url('api/credit/v1/'),
+            oauth_access_token=access_token
+        ).providers(provider_id).get()
+    except (ConnectionError, SlumberHttpBaseException, Timeout) as ex:
+        logger.exception(
+            'Failed to retrieve credit provider details for provider [%s], Because of [%s]',
+            provider_id,
+            ex,
```
clintonb: You don't need to pass the exception when using `logger.exception`.

ecommerce/extensions/checkout/utils.py
```diff
@@ -1,40 +1,32 @@
...
 logger = logging.getLogger(__name__)
 
 
-def get_provider_data(provider_id):
+def get_provider_data(access_token, provider_id, site_configuration):
     """Get the provider information for provider id provider.
 
     Args:
+        access_token(str): JWT access token
         provider_id(str): Identifier for the provider
```
clintonb: credit_provider_id

ecommerce/extensions/checkout/tests/test_signals.py
clintonb: Why was this changed? We occasionally use MIT/Harvard in tests because they are founding partners. ASU is not. If our partnership with them ever dissolves, we may have to remove them entirely from our code base. Avoid this problem now, and use a fake university name (e.g. ACME).

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -33,9 +32,10 @@ def test_post_checkout_callback(self):
         to fulfill the newly-placed order and send receipt email.
         """
         httpretty.register_uri(
-            httpretty.GET, get_lms_url('api/credit/v1/providers/ASU'),
-            body='{"display_name": "Hogwarts"}',
-            content_type="application/json"
+            httpretty.GET,
+            self.site_configuration.build_lms_url('api/credit/v1/providers/ASU/'),
+            body='{"display_name": "Arizona State University"}',
```
clintonb: Use a variable that can be shared below.

ecommerce/extensions/checkout/tests/test_utils.py
```diff
@@ -17,11 +16,16 @@ def test_get_provider_data(self):
         Check if correct data returns on the full filled request.
         """
         httpretty.register_uri(
-            httpretty.GET, get_lms_url('api/credit/v1/providers/ASU'),
+            httpretty.GET,
+            self.site.siteconfiguration.build_lms_url('api/credit/v1/providers/ASU/'),
```
clintonb: Use a variable for the provider ID.

ecommerce/extensions/checkout/tests/test_utils.py
```diff
@@ -17,11 +16,16 @@ def test_get_provider_data(self):
...
             body='{"display_name": "Arizona State University"}',
```
clintonb: Use a variable for the display name.

ecommerce/extensions/checkout/tests/test_utils.py
```diff
@@ -17,11 +16,16 @@ def test_get_provider_data(self):
...
             content_type="application/json"
         )
-        provider_data = get_provider_data('ASU')
+        provider_data = get_provider_data(
+            access_token=self.access_token,
```
clintonb: You can pass all of these as `args` instead of `kwargs`.

clintonb (changes requested):
`threadlocals` needs to go away. The current usage (on master) is "okay"; but, we should not be expanding its usage.

marjev:
@clintonb Thanks for the review! Please take another look at my PR.

ecommerce/extensions/checkout/signals.py
```diff
@@ -68,14 +67,18 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
...
+            credit_provider_id = getattr(product.attr, 'credit_provider', None)
+            if not credit_provider_id:
                 logger.error(
                     'Failed to send credit receipt notification. Credit seat product [%s] has not provider.', product.id
                 )
                 return
             elif product.get_product_class().name == 'Seat':
-                provider_data = get_provider_data(provider_id)
+                provider_data = get_credit_provider_details(
+                    access_token=order.user.access_token,
```
clintonb: This could be problematic due to the fact that the email may be sent (in the case of a re-send) after the user's access token has expired. You should use `order.site.siteconfiguration.access_token`. See https://github.com/edx/ecommerce/blob/master/ecommerce/core/models.py#L223-L250.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -1,63 +1,81 @@
...
+
+        Returns:
+            Order
+        """
+        course = CourseFactory()
+        seat = course.create_or_update_seat(seat_type, False, 50, self.partner, credit_provider_id, None, 2)
+        basket = BasketFactory()
+        basket.add_product(seat, 1)
+        order = factories.create_order(number=1, basket=basket, user=self.user)
+        order.site = self.site
```
clintonb: This should be set on the basket. `BasketFactory(site=self.site)` should work. If not, use `basket.site = self.site`.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -1,63 +1,81 @@
...
+            credit_provider_id (str): Credit provider associated with the course seat.
+
+        Returns:
+            Order
+        """
+        course = CourseFactory()
+        seat = course.create_or_update_seat(seat_type, False, 50, self.partner, credit_provider_id, None, 2)
+        basket = BasketFactory()
+        basket.add_product(seat, 1)
+        order = factories.create_order(number=1, basket=basket, user=self.user)
```
clintonb: Unset `number`, and one will be generated for you.

ecommerce/extensions/checkout/tests/test_utils.py
clintonb: The comments in this file still need to be addressed.

marjev:
@clintonb I think I addressed all the comments you left yesterday + the ones left on the `test_utils.py`. Could you please take another look?

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -70,8 +85,24 @@ def test_post_checkout_callback(self):
...
+
+    def test_post_checkout_callback_no_credit_provider(self):
+        order = self.prepare_order('verified')
+        with LogCapture(LOGGER_NAME) as l:
+            send_course_purchase_email(None, user=self.user, order=order)
+            l.check(
+                (
+                    LOGGER_NAME,
+                    'ERROR',
+                    'Failed to send credit receipt notification. Credit seat product [{}] has not provider.'.format(
```
clintonb: Nit: s/not/no

clintonb (approved):
Remember to squash commits before merging

PR #936, opened 2016-09-12
Check for user account activations status before redeeming a coupon code.

vkaracic:
User that did not activate their account should not be able to redeem a coupon. This PR contains a new method in the user model that returns the account details for a user from LMS.

https://openedx.atlassian.net/browse/SOL-1866

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -309,7 +309,7 @@ class User(AbstractUser):
             status = api.enrollment(','.join([self.username, course_key])).get()
         except (ConnectionError, SlumberBaseException, Timeout) as ex:
             log.exception(
-                'Failed to retrieve enrollment details for [%s] in course [%s], Because of [%s]',
+                'Failed to retrieve enrollment details for [%s] in course [%s], because of [%s]',
                 self.username,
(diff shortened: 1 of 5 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=936&repo=edx%2Fecommerce

ecommerce/core/models.py
```diff
@@ -315,6 +315,36 @@ def is_user_already_enrolled(self, request, seat):
...
+            response = api.accounts(self.username).get()
+            return response
+        except (ConnectionError, SlumberBaseException, Timeout) as ex:
+            log.exception(
+                'Failed to retrieve account details for [%s], because of [%s]',
+                self.username,
+                ex,
+            )
+            raise ex
+        return False  # pragma: no cover
```
iivic: Will this line be executed?

iivic (approved):
One question but looks good.

vkaracic:
@mjfrey please review.

marjev (approved):

PR #937, opened 2016-09-12
Change all_products implementation for Multi-course Coupons

marjev:


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=937

marjev:
@mjfrey Here's the PR I was talking about during our standup. Please take a look. I added the change to the constants as you suggested to reduce the DEFAULT_CATALOG_PAGE_SIZE.

ecommerce/extensions/offer/models.py
```diff
@@ -7,8 +7,6 @@
 from oscar.apps.offer.abstract_models import AbstractConditionalOffer, AbstractRange
 from threadlocals.threadlocals import get_current_request
 
-from ecommerce.coupons.utils import get_seats_from_query
```
mjfrey, line 10: I think we should also remove get_seats_from_query since this was the last use of it.  I think.

marjev: Done. Take a look at 6edb52ddebcd7fe96ef86166fa8bb8a2286b2f96.

marjev:
@vkaracic

vkaracic:
LGTM

mjfrey:
👍

ecommerce/coupons/views.py
```diff
@@ -129,10 +128,6 @@ def get_context_data(self, **kwargs):
                 return {
                     'error': _('The voucher is not applicable to your current basket.'),
                 }
-            except (ConnectionError, SlumberBaseException, Timeout):
-                return {
-                    'error': _('Coupon information not available at this time. Please try again later.'),
-                }
```
marjev, line 135: @mjfrey @vkaracic @iivic I removed this catch because we don't call Course Catalog anymore in `get_voucher_and_products_from_code`. See the changes I did in this PR for that method. I also removed the tests for that case since we no longer need them.

mjfrey:
👍

PR #938, opened 2016-09-13
[SOL-2007] Coupon Report do not shows the coupon details.

iivic:
Added invoiced_amount, discount_percentage and coupon_type (in case when benefit type is percentage based) to dynamic coupon report.
@marjev @vkaracic @mjfrey

### DIFF
--- a/ecommerce/extensions/voucher/tests/test_utils.py
+++ b/ecommerce/extensions/voucher/tests/test_utils.py
@@ -277,13 +277,22 @@ class UtilTests(CouponMixin, CourseCatalogMockMixin, CourseCatalogTestMixin, Lms
             voucher (Voucher): Voucher associated with the Coupon
         """
         offer = voucher.offers.first()
-        discount_data = get_voucher_discount_info(
-            offer.benefit,
-            offer.condition.range.catalog.stock_records.first().price_excl_tax
-        )
-        coupon_type = _('Discount') if discount_data['is_discounted'] else _('Enrollment')
-        discount_percentage = _("{percentage} %").format(percentage=discount_data['discount_percentage'])
-        discount_amount = currency(discount_data['discount_value'])
+        if offer.condition.range.catalog:
+            discount_data = get_voucher_discount_info(
+                offer.benefit,
+                offer.condition.range.catalog.stock_records.first().price_excl_tax
+            )
+            coupon_type = _('Discount') if discount_data['is_discounted'] else _('Enrollment')
+            discount_percentage = _('{percentage} %').format(percentage=discount_data['discount_percentage'])
+            discount_amount = currency(discount_data['discount_value'])
+        else:
+            if offer.benefit.type == Benefit.PERCENTAGE:
+                coupon_type = _('Discount') if offer.benefit.value < 100 else _('Enrollment')
+            else:
+                coupon_type = None
+            discount_amount = None
+            discount_percentage = _('{percentage} %').format(
+                percentage=offer.benefit.value) if offer.benefit.type == Benefit.PERCENTAGE else None
 
         self.assertEqual(row['Coupon Type'], coupon_type)
         self.assertEqual(row['Category'], ProductCategory.objects.get(product=coupon).category.name)
@@ -373,6 +382,24 @@ class UtilTests(CouponMixin, CourseCatalogMockMixin, CourseCatalogTestMixin, Lms
         self.assertNotIn('Course Seat Types', field_names)
         self.assertNotIn('Redeemed For Course ID', field_names)
 
+    def test_report_for_dynamic_coupon_with_fixed_benefit_type(self):
+        """ Verify the coupon report contains correct data for coupon with fixed benefit type. """
+        dynamic_coupon = self.create_coupon(
+            benefit_type=Benefit.FIXED,
+            benefit_value=50,
+            catalog_query='*:*',
+            course_seat_types='verified',
+            max_uses=1,
+            note='Tešt note',
+            quantity=1,
+            title='Tešt product'
+        )
+        dynamic_coupon.history.all().update(history_user=self.user)
+        coupon_voucher = CouponVouchers.objects.get(coupon=dynamic_coupon)
+        __, rows = generate_coupon_report([coupon_voucher])
+        voucher = coupon_voucher.vouchers.first()
+        self.assert_report_first_row(rows[0], dynamic_coupon, voucher)
+
     def test_report_for_inactive_coupons(self):
         """ Verify the coupon report show correct status for inactive coupons. """
         coupon_title = self.coupon.title
@@ -406,10 +433,7 @@ class UtilTests(CouponMixin, CourseCatalogMockMixin, CourseCatalogTestMixin, Lms
         field_names, rows = generate_coupon_report([query_coupon.attr.coupon_vouchers])
 
         empty_fields = (
-            'Coupon Type',
             'Discount Amount',
-            'Discount Percentage',
-            'Invoiced Amount',
             'Price',
         )
         for field in empty_fields:
--- a/ecommerce/extensions/voucher/utils.py
+++ b/ecommerce/extensions/voucher/utils.py
@@ -87,10 +87,10 @@ def _get_info_for_coupon_report(coupon, voucher):
         note = ''
 
     offer = voucher.offers.first()
+    coupon_stockrecord = StockRecord.objects.get(product=coupon)
+    invoiced_amount = currency(coupon_stockrecord.price_excl_tax)
     if offer.condition.range.catalog:
         seat_stockrecord = offer.condition.range.catalog.stock_records.first()
-        coupon_stockrecord = StockRecord.objects.get(product=coupon)
-        invoiced_amount = currency(coupon_stockrecord.price_excl_tax)
         course_id = seat_stockrecord.product.attr.course_key
         course_organization = CourseKey.from_string(course_id).org
         price = currency(seat_stockrecord.price_excl_tax)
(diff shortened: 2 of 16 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=938&repo=edx%2Fecommerce

ecommerce/extensions/voucher/utils.py
```diff
@@ -87,10 +87,10 @@ def _get_info_for_coupon_report(coupon, voucher):
         note = ''
 
     offer = voucher.offers.first()
+    coupon_stockrecord = StockRecord.objects.get(product=coupon)
```
marjev, line 90: what happens if StockRecord object is not found?

ecommerce/extensions/voucher/utils.py
```diff
@@ -100,11 +100,21 @@ def _get_info_for_coupon_report(coupon, voucher):
         # Note (multi-courses): Need to account for multiple seats.
         catalog_query = offer.condition.range.catalog_query
         course_seat_types = offer.condition.range.course_seat_types
+        if offer.benefit.type == Benefit.PERCENTAGE:
+            discount_percentage = _("{percentage} %").format(percentage=float(offer.benefit.value))
```
marjev: why do we need a float of `offer.benefit.value`?

ecommerce/extensions/voucher/utils.py
```diff
@@ -100,11 +100,21 @@ def _get_info_for_coupon_report(coupon, voucher):
...
+        else:
+            discount_percentage = None
```
marjev: try making this a one-liner.

ecommerce/extensions/voucher/utils.py
```diff
@@ -100,11 +100,21 @@ def _get_info_for_coupon_report(coupon, voucher):
...
+        if offer.benefit.type == Benefit.PERCENTAGE:
+            if float(offer.benefit.value) < 100:
```
marjev: In case we need the float of benefit value, DRY it.

ecommerce/extensions/voucher/utils.py
```diff
@@ -100,11 +100,21 @@ def _get_info_for_coupon_report(coupon, voucher):
...
+                coupon_type = _('Discount')
+            else:
+                coupon_type = _('Enrollment')
```
marjev: One liner.

ecommerce/extensions/voucher/utils.py
```diff
@@ -454,6 +464,7 @@ def create_vouchers(
     return vouchers
 
 
+
```
marjev: This empty line is unnecessary.

marjev:
I finished my first pass. Let me know once the comments have been addressed.

ecommerce/extensions/voucher/utils.py
iivic: There must be a StockRecord for coupon.

ecommerce/extensions/voucher/utils.py
iivic: I'm not sure I saw that float is used this way in get_voucher_discount_info() but removing it changes nothing. I will remove the float ...

ecommerce/extensions/voucher/utils.py
iivic: Removed float.

iivic:
Comments addressed.

ecommerce/extensions/voucher/utils.py
marjev: Is there a backend validation that supports that use-case?

iivic: We can't create a coupon without stock record. Is this sufficient?

iivic:
Test edited review once more.

ecommerce/extensions/voucher/utils.py
marjev: Does deleting a StockRecord deletes the associated Product? If not, either catch the possible Exception here or enforce the necessary logic.

iivic: No. It doesn't delete the associated Product. I'll catch the exception ...

ecommerce/extensions/voucher/utils.py
```diff
@@ -87,10 +88,13 @@ def _get_info_for_coupon_report(coupon, voucher):
         note = ''
 
     offer = voucher.offers.first()
+    try:
+        coupon_stockrecord = StockRecord.objects.get(product=coupon)
+    except StockRecord.DoesNotExist:
+        raise Http404('StockRecord not found.')
```
marjev: This can be turned into a one-liner. https://docs.djangoproject.com/en/1.10/topics/http/shortcuts/#get-object-or-404

Also, in the case of the exception, we want to log the error message. This is the edge case and we want to be aware of this unexpected behaviour.

If this exception is raised, something needs to catch it.

ecommerce/extensions/voucher/utils.py
marjev: @mjfrey, I'm in doubt of what's acceptable here.

Here's the thing. Coupon not having a StockRecord is something that should never happen, if the ecommerce is used like we expect it. Problem will occur, however, if there's no StockRecord object associated with the Coupon. Django will throw `ObjectNotFound` exception.

It should never happen, and should indicate the bigger issue, but should we simply ignore the possible trap here? Should we simply log a message and throw the Exception anyway, thus having a broken UI? Alternatively, we can catch the Exception, log it and show some kind of the error message on the details page, if clicking `Download Voucher Report` button fails (assuming that we don't want to allow a user to download a report for a broken Coupon, the one that has no StockRecord object associated with it).

I'm just thinking about this because this is an edge case that could happen if data in our ecommerce database gets corrupted for any reason.

mjfrey: @marjev I think we should catch the error -- log it and show an error page to the user.  You are correct that this would show an underlying data issue that we should flag and fix.

iivic:
This is the error message when the download fails
screen shot 2016-09-20 at 14 22 21

ecommerce/extensions/voucher/utils.py
```diff
@@ -216,9 +216,9 @@ def generate_coupon_report(coupon_vouchers):
         client = Invoice.objects.get(order__lines__product=coupon).business_client.name
         try:
             rows.append(_get_info_for_coupon_report(coupon, coupon_voucher.vouchers.first()))
-        except StockRecord.DoesNotExist as ex:
+        except StockRecord.DoesNotExist:
```
marjev: You'll have to move this in the top-most method to avoid sending generic values around. Right now you're returning `[]`.

ecommerce/extensions/voucher/views.py
```diff
@@ -29,6 +29,9 @@ def get(self, request, coupon_id):  # pylint: disable=unused-argument
 
         field_names, rows = generate_coupon_report(coupons_vouchers)
```
marjev: Move try-except here. Catch the exception and return HttpResponse with status=404 and custom message.

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -159,11 +159,26 @@ define([
...
+                $.ajax({
+                    url: url,
+                    type: 'GET',
+                    success: function(){
+                        event.preventDefault();
+                        window.open(url, '_blank');
+                        return _this;
+                    },
+                    error: function() {
+                        $alerts.removeClass('hidden').children().text(gettext('StockRecord not found. Download canceled.'));
```
marjev: This message will be shown for every error that might occur here. Send the message from the backend and simply show the sent message on the template.

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -159,11 +159,26 @@ define([
...
+                        $('body').animate({
+                            scrollTop: $alerts.offset().top
+                        }, 500);
+                        $alerts.focus();
```
marjev: Move this to utils and use the newly created function here and https://github.com/edx/ecommerce/blob/d7fc9529919704a187a741f105deebd7f11208c5/ecommerce/static/js/views/form_view.js#L65 and in any other place this is currently done.

ecommerce/static/templates/coupon_detail.html
```diff
@@ -17,6 +17,9 @@ <h3 class="hd-3">
             <span class="coupon-type"><%= coupon.coupon_type %></span>
         </h3>
     </div>
+    <div class="alerts hidden" tabindex="-1" aria-live="polite">
+        <div class="alert alert-danger" role="alert"></div>
+    </div>
```
marjev: DRY this if multiple templates are using the same div.
Also, do you need the inner div element?

ecommerce/extensions/voucher/utils.py
marjev: @mjfrey There's a screenshot containing the error message. Can you take a look to see if the message is appropriate?

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -373,6 +382,23 @@ def test_generate_coupon_report(self):
         self.assertNotIn('Course Seat Types', field_names)
         self.assertNotIn('Redeemed For Course ID', field_names)
 
+        # Test dynamic coupon with fixed benefit type
+        dynamic_coupon = self.create_coupon(
```
vkaracic, line 387: Can this be moved to a new test method?

ecommerce/static/js/utils/alert_utils.js
```diff
@@ -0,0 +1,49 @@
...
+        'jquery',
+        'views/alert_view'],
+    function ($,
+              AlertView) {
+        'use strict';
+
+        return {
+             /**
+             * Remove all alerts currently on display.
+              *
```
vkaracic, line 11: Nit: arrange the stars in a line.

ecommerce/static/js/utils/alert_utils.js
```diff
@@ -0,0 +1,49 @@
...
+             * @param {View} context - View which removes the alerts.
```
vkaracic: Should this say `View for which the alerts are removed`?

vkaracic (approved):
Just a couple of nits.

ecommerce/extensions/voucher/views.py
```diff
@@ -27,7 +31,11 @@ def get(self, request, coupon_id):  # pylint: disable=unused-argument
 
         filename = "{}.csv".format(slugify(filename))
 
-        field_names, rows = generate_coupon_report(coupons_vouchers)
+        try:
+            field_names, rows = generate_coupon_report(coupons_vouchers)
+        except StockRecord.DoesNotExist:
+            logger.exception(u'Failed to find StockRecord for Coupon [%d].', coupon.id)
+            return HttpResponse(_('StockRecord not found. Download canceled.'), status=404)
```
mjfrey: "Failed to find a matching StockRecord for Coupon, Report download canceled"

marjev (approved):

PR #942, opened 2016-09-17
ECOM-5486 pass sailthru campaign id to Sailthru for enrolls

PaulWattenberger:


PaulWattenberger:
@clintonb code to retain sailthru cookie when handling audit/honor enrolls with async fulfillment enabled

ecommerce/sailthru/signals.py
```diff
@@ -115,11 +111,15 @@ def process_basket_addition(sender, product=None, user=None, request=None, baske
...
 
-        # pass event to ecommerce_worker.sailthru.v1.tasks to handle asynchronously
-        update_course_enrollment.delay(user.email, _build_course_url(course_id), True, mode_for_seat(product),
-                                       unit_cost=price, course_id=course_id, currency=currency,
-                                       site_code=partner.short_code,
-                                       message_id=message_id)
+        # inform sailthru if there is a price, no need to tell sailthru about adding enrolls to shopping cart since
+        # abandoned cart is not possible
+        if price:
+
```
clintonb: Remove the unnecessary break.

ecommerce/sailthru/signals.py
```diff
@@ -115,11 +111,15 @@ def process_basket_addition(sender, product=None, user=None, request=None, baske
...
+            # pass event to ecommerce_worker.sailthru.v1.tasks to handle asynchronously
```
clintonb: This comment doesn't add any value to the reader, and should be removed.

ecommerce/sailthru/signals.py
```diff
@@ -115,11 +111,15 @@ def process_basket_addition(sender, product=None, user=None, request=None, baske
                 value_text=message_id
             )
 
-        # pass event to ecommerce_worker.sailthru.v1.tasks to handle asynchronously
-        update_course_enrollment.delay(user.email, _build_course_url(course_id), True, mode_for_seat(product),
-                                       unit_cost=price, course_id=course_id, currency=currency,
-                                       site_code=partner.short_code,
-                                       message_id=message_id)
+        # inform sailthru if there is a price, no need to tell sailthru about adding enrolls to shopping cart since
```
clintonb: What does the second half of this comment mean? Why does abandonment matter here? If the user has enrolled, Sailthru should be notified.

clintonb (approved):

ecommerce/sailthru/signals.py
PaulWattenberger: Sailthru needs to be told when enrollment complete, but not when the enroll is added to the basket.  That is only needed for abandoned cart detection...

clintonb: Please update the comment to be clearer:

```
We only need to notify Sailthru when there is the possibility of an abandoned cart. Only paid orders have the possibility of being abandoned, so we only notify Sailthru for paid orders.
```

clintonb (approved):

PR #945, opened 2016-09-20
Add select_related to courses api endpoint

bderusha:
[ECOM-5701](https://openedx.atlassian.net/browse/ECOM-5701)

Decrease number of individual database queries by 66%

@rlucioni @cpennington please review 

**http://localhost:8002/api/v2/courses/?include_products=true | Master**
_DB time_: 28.04 ms (68 queries including 66 duplicates )
_Request time_: 954.053 msec

**http://localhost:8002/api/v2/courses/?include_products=true | Branch**
_DB time_: 10.49 ms (25 queries including 20 duplicates )
_Request time_: 584.325 msec

### DIFF
--- a/ecommerce/extensions/api/v2/views/courses.py
+++ b/ecommerce/extensions/api/v2/views/courses.py
@@ -1,8 +1,10 @@
 """HTTP endpoints for interacting with courses."""
+from django.db.models import Prefetch
 from rest_framework import status
 from rest_framework.decorators import detail_route
 from rest_framework.permissions import IsAuthenticated, IsAdminUser
 from rest_framework.response import Response
+from oscar.core.loading import get_model
 import waffle
 
 from ecommerce.core.constants import COURSE_ID_REGEX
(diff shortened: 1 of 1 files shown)


bderusha:
@rlucioni @cpennington please review

rlucioni:
@bderusha consider [asserting on query count](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries) in tests to cover your work here.

ecommerce/extensions/api/v2/views/courses.py
```diff
@@ -1,19 +1,35 @@
...
 from ecommerce.courses.models import Course
 from ecommerce.extensions.api.v2.views import NonDestroyableModelViewSet
 from ecommerce.extensions.api import serializers
 
+Product = get_model('catalogue', 'Product')
+ProductAttributeValue = get_model('catalogue', 'ProductAttributeValue')
+
 
 class CourseViewSet(NonDestroyableModelViewSet):
+    product_attribute_value_prefetch = Prefetch(
```
rlucioni, line 20: @cpennington and I were pretty happy with a pattern we used in the catalog for this kind of prefetching, and I wonder if we should try standardizing on it here. The idea is that you include a classmethod on serializers which returns the queryset that serializer needs. The [`ProgramSerializer`](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/serializers.py#L470) is a good example.

ecommerce/extensions/api/v2/views/courses.py
```diff
@@ -1,19 +1,35 @@
...
+        queryset=ProductAttributeValue.objects.select_related('attribute').all()
+    )
+
+    products_prefetch = Prefetch(
+        'products',
+        queryset=Product.objects.select_related('parent__product_class').all()
+    )
     lookup_value_regex = COURSE_ID_REGEX
-    queryset = Course.objects.all()
+    queryset = Course.objects.prefetch_related(
```
rlucioni: I'm surprised to see this working here. What happens when you load the browsable API twice in a row with this? I had to define a [`get_queryset`](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/views.py#L384) method in the catalog to prevent prefetches from compounding when using the `queryset` property.

rlucioni:
@bderusha just curious, what kind of response time/query count change are we seeing as a result of this? Does the debug toolbar show any remaining queries scaling linearly in the size of the response?

clintonb (approved):
Please quantify the improvements made. What's the current query count/response time? How do these changes affect those numbers?

bderusha:
@clintonb @rlucioni added data for the expected gains

bderusha:
Merging this now to unblock data refreshing, will continue to refactor code and add tests.

PR #947, opened 2016-09-21
Add ability to create coupons for credit seats.

vkaracic:
This PR contains changes that allow the creation of coupons for credit seats.
screenshot from 2016-09-21 14-45-42
screenshot from 2016-09-21 14-45-47

Also additional custom jasmine matches are added and test updated accordingly.

### DIFF
--- a/ecommerce/extensions/offer/tests/test_models.py
+++ b/ecommerce/extensions/offer/tests/test_models.py
@@ -3,6 +3,7 @@ import hashlib
 import httpretty
 import mock
 from django.core.cache import cache
+from django.core.exceptions import ValidationError
 from django.test import RequestFactory
 from oscar.core.loading import get_model
 from oscar.test import factories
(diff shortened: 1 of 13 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=947

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -106,6 +106,11 @@ def create(self, request, *args, **kwargs):
             invoice_data = self.create_update_data_dict(data=request.data, fields=Invoice.UPDATEABLE_INVOICE_FIELDS)
 
             if course_seat_types:
+                if len(course_seat_types) > 1 and 'credit' in course_seat_types:
```
marjev: Add this as a custom validator for course_seat_types Range field. If it fails there, ValidationError will be raised instead.

vkaracic:
@mjfrey please review.

vkaracic:
Comments addressed.

marjev (approved):

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -122,6 +124,20 @@ def test_query_range_all_products(self):
         self.range.course_seat_types = 'verified'
         self.assertEqual(len(self.range.all_products()), 0)
 
+    def test_credit_seat_type_validator(self):
+        """
+        Verify the validator raises error for combination of credit and another seat type.
+        """
+        ver_pro_types = 'verified,professional'
+        credit_type = 'credit'
+        invalid_type = 'credit,verified'
```
iivic: Are these variables used only once? If so why not use the strings directly?

vkaracic:
Comments addressed.

iivic (approved):

mjfrey:
👍

PR #948, opened 2016-09-21
HON-93 Add options to the create_or_update_site command to populate SiteConfiguration fields

douglashall:
@vkaracic @mjfrey 

Associated edx-documentation PR is [here](https://github.com/edx/edx-documentation/pull/1258).

### DIFF
--- a/ecommerce/core/management/commands/create_or_update_site.py
+++ b/ecommerce/core/management/commands/create_or_update_site.py
@@ -94,6 +94,18 @@ class Command(BaseCommand):
                             type=bool,
                             required=False,
                             help='Enable the creation of enrollment codes.')
+        parser.add_argument('--payment-support-email',
+                            action='store',
+                            dest='payment_support_email',
+                            type=str,
+                            required=False,
+                            help='Email address displayed to user for payment support')
+        parser.add_argument('--payment-support-url',
+                            action='store',
+                            dest='payment_support_url',
+                            type=str,
+                            required=False,
+                            help='URL displayed to user for payment support')
 
     def handle(self, *args, **options):
         site_id = options.get('site_id')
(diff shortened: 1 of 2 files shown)


ecommerce/core/management/commands/create_or_update_site.py
```diff
@@ -94,6 +94,18 @@ def add_arguments(self, parser):
...
+                            action='store',
+                            dest='payment_support_email',
+                            type=str,
+                            required=False,
+                            help='Email address displayed to user for payment support')
+        parser.add_argument('--payment-support-url',
+                            action='store',
+                            dest='payment_support_url',
+                            type=str,
+                            required=False,
```
vkaracic, line 107: I think this is required. A site configuration without this value gives out errors everywhere (that it's missing).

douglashall: It is not required. The model allows for this field to be [blank](https://github.com/edx/ecommerce/blob/master/ecommerce/core/models.py#L94).

This field is only used in the basket [template](https://github.com/edx/ecommerce/blob/282ebf64b0c7b1ed0cd73d7a0f34936bd548c88a/ecommerce/templates/oscar/basket/partials/basket_content.html#L204) when the basket is empty.

douglashall:
@vkaracic Tests now passing on this one.

ecommerce/core/management/commands/create_or_update_site.py
vkaracic: You are correct. With a new database I don't get any errors, I'm not sure why I kept getting them on my old database.

vkaracic (approved):
LGTM
Thanks for adding this!

PR #949, opened 2016-09-21
Move basket page to use catalog service for meta-data

mjfrey:
Make the basket view use catalog data instead of LMS (again).

### DIFF
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -22,9 +22,10 @@ from testfixtures import LogCapture
 from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME, ENROLLMENT_CODE_SWITCH
 from ecommerce.core.models import SiteConfiguration
 from ecommerce.core.tests import toggle_switch
+from ecommerce.core.tests.decorators import mock_course_catalog_api_client
 from ecommerce.core.url_utils import get_lms_enrollment_api_url
 from ecommerce.core.url_utils import get_lms_url
-from ecommerce.coupons.tests.mixins import CouponMixin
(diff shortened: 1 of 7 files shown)


mjfrey:
@marjev @vkaracic 
Trying to release this again now that all WL sites are in the catalog.  Can you take a look.

marjev (approved):

marjev:
LGTM

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -387,39 +389,50 @@ def test_line_item_discount_data(self):
...
-        course_info_json = json.dumps(course_info)
-        httpretty.register_uri(httpretty.GET, course_url, body=course_info_json, content_type='application/json')
+        course_run_info_json = json.dumps(course_info)
+        course_run_url = '{}course_runs/{}/?partner={}'.format(
+            settings.COURSE_CATALOG_API_URL,
+            self.course.id,
+            self.site.siteconfiguration.partner.short_code
+        )
 
+        httpretty.register_uri(
```
vkaracic: Instead of registering a new URI, can you call `mock_dynamic_catalog_single_course_runs_api`?

vkaracic (approved):
One comment, otherwise :+1:

ecommerce/extensions/basket/tests/test_views.py
mjfrey: I am overwriting the return JSON.  That is why I am not using the built in Mock.

vkaracic: You can save yourself the additional code here by adding an additional keyword argument in that mock. Something like this combination: https://github.com/edx/ecommerce/blob/d4bb441536c25c494d63881ab0841da32139e16a/ecommerce/coupons/tests/mixins.py#L50-L52

PR #951, opened 2016-09-27
Improved performance of order admin page

clintonb:
- Using raw_id fields for order line stock records and products
- Added select_related to improve query performance

These changes greatly decrease the query count and load time of this page; but, the use of inlines poses a risk due to the fact that any choice field will result in query for each inline displaying that field. Some form of database query caching might help alleviate this issue.

ECOM-5767


clintonb:
@edx/ecommerce please review
@bderusha see the commit comment as it may pose issues with your publisher work

schenedx (approved):
👍

PR #952, opened 2016-09-28
Display error message when no offers exist for a coupon.

vkaracic:
The offer landing page for a dynamic coupon that has no offers (no seats to display) would display an empty page. This PR contains changes that make the offer landing page display an error message instead of the empty page.
screenshot from 2016-09-28 12-39-47

https://openedx.atlassian.net/browse/SOL-2090

### DIFF
--- a/ecommerce/static/js/collections/offer_collection.js
+++ b/ecommerce/static/js/collections/offer_collection.js
@@ -13,10 +13,19 @@ define([
             url: '/api/v2/vouchers/offers/',
 
             initialize: function() {
+                this.empty = false;
                 this.page = 1;
                 this.perPage = 6;
                 this.updateLimits();
                 this.on('update', this.updateNumberOfPages);
+            },
+
+            parse: function(response) {
+                this._super(response);
+                if (response.results.length === 0) {
+                    this.empty = true;
+                }
+                return response.results;
             },
 
             updateNumberOfPages: function() {
--- a/ecommerce/static/js/test/specs/collections/offer_collection_spec.js
+++ b/ecommerce/static/js/test/specs/collections/offer_collection_spec.js
@@ -1,5 +1,6 @@
 define([
-        'collections/offer_collection'
+        'collections/offer_collection',
+        'backbone.super'
     ],
     function(OfferCollection) {
         'use strict';
@@ -7,6 +8,7 @@ define([
             response = {
                 count: 4,
                 page: 1,
+                empty: false,
                 next: null,
                 previous: null,
                 results: [
@@ -29,6 +31,12 @@ define([
 
             it('should return the results list in the response', function() {
                 expect(collection.parse(response)).toEqual(response.results);
+            });
+
+            it('should set the collection is empty when no results are fetched', function() {
+                response.results = [];
+                collection.parse(response);
+                expect(collection.empty).toBeTruthy();
             });
 
             it('should fetch the next page of results', function() {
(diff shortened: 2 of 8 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=952

ecommerce/static/js/views/offer_view.js
```diff
@@ -43,7 +46,9 @@ define([
             },
 
             render: function() {
-                if (this.collection.length > 0) {
+                if (this.collection.empty) {
```
iivic, line 50: Why can't you just check if length == 0 since the fetch isn't async? Empty looks unnecessary to me.

vkaracic: I've returned the async because we probably want to do this asynchronously.

iivic (approved):

ecommerce/static/js/collections/offer_collection.js
```diff
@@ -13,12 +13,23 @@ define([
             url: '/api/v2/vouchers/offers/',
 
             initialize: function() {
+                this.empty = false;
                 this.page = 1;
                 this.perPage = 6;
+                this.empty = false;
```
marjev: Redundant.

ecommerce/static/js/collections/offer_collection.js
```diff
@@ -13,12 +13,23 @@ define([
...
                 this.updateLimits();
                 this.on('update', this.updateNumberOfPages);
             },
 
+            parse: function(response) {
+                this._super(response);
+                if (response.results.length === 0) {
+                    this.empty = true;
+                    return false;
```
marjev: Do you need to return false?

ecommerce/static/js/test/specs/collections/offer_collection_spec.js
```diff
@@ -31,6 +33,12 @@ define([
                 expect(collection.parse(response)).toEqual(response.results);
             });
 
+            it('should set it is empty when no results are fetched', function() {
```
marjev: what is `it` in the test description?

ecommerce/static/js/views/offer_view.js
```diff
@@ -62,6 +67,12 @@ define([
                 return this;
             },
 
+            emptyOffer: function() {
```
marjev: Functions/methods names should usually describe what's been done. I suggest renaming this to `showEmptyOfferErrorMessage` or something similar.

marjev (changes requested):

ecommerce/static/js/collections/offer_collection.js
vkaracic: Yes.

vkaracic:
Comments addressed!

ecommerce/static/js/collections/offer_collection.js
marjev: Can't you return the response.results? I thought you're only using `this.empty` flag.

vkaracic: Displays an empty page without this return instead of the error message.

marjev: Can you double check that? If I remove `return false;` locally, it works.

marjev (approved):
LGTM

mjfrey:
👍

PR #953, opened 2016-09-28
[SOL-2083] Check if user is enrolled in course only if the request arrives from the Offer Landing Page

marjev:
@mjfrey Here's a PR for [SOL-2083](https://openedx.atlassian.net/browse/SOL-2083).

Please take a look.

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -52,9 +52,10 @@ class BasketSingleItemView(View):
             msg = _('Product [{product}] not available to buy.').format(product=product.title)
             return HttpResponseBadRequest(msg)
 
-        # If the product is not an Enrollment Code, we check to see if the user is already
-        # enrolled to prevent double-enrollment and/or accidental coupon usage
-        if product.get_product_class().name != ENROLLMENT_CODE_PRODUCT_CLASS_NAME:
+        # If the product is not an Enrollment Code and this is a Coupon Redemption request,
+        # we check to see if the user is already enrolled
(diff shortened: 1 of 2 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=953

ecommerce/extensions/basket/views.py
```diff
@@ -54,7 +54,10 @@ def get(self, request):
 
         # If the product is not an Enrollment Code, we check to see if the user is already
         # enrolled to prevent double-enrollment and/or accidental coupon usage
-        if product.get_product_class().name != ENROLLMENT_CODE_PRODUCT_CLASS_NAME:
+        print request.GET
```
mjfrey: remove the prints

marjev: Prints removed.

ecommerce/extensions/basket/views.py
```diff
@@ -54,7 +54,7 @@ def get(self, request):
 
         # If the product is not an Enrollment Code, we check to see if the user is already
         # enrolled to prevent double-enrollment and/or accidental coupon usage
```
mjfrey: Please update this comment to say we will only check if this is a coupon redemption.

marjev: I've updated the comment. Please take a look to see if its meaning is understandable enough.

mjfrey:
👍

PR #954, opened 2016-09-28
Upgrade Django to 1.8.15

rlucioni:
Django 1.8.15 fixes a security issue in 1.8.14. Release notes are at https://docs.djangoproject.com/en/1.10/releases/1.8.15.

@edx/ecommerce FYI.


clintonb:
👍

PR #955, opened 2016-09-29
Add email validation for coupon form.

vkaracic:
It was possible to add anything in the email domain input field. This PR contains pattern validation for the email domain field.

https://openedx.atlassian.net/browse/SOL-2034

### DIFF
--- a/ecommerce/static/js/models/coupon_model.js
+++ b/ecommerce/static/js/models/coupon_model.js
@@ -108,7 +108,11 @@ define([
                         return Backbone.Validation.messages.seat_types;
                     }
                 },
-                email_domains: {required: false},
+                email_domains: {
+                    pattern: 
+                    /^((xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,}(,((xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,})*$/,
+                    required: false
+                },
                 start_date: function (val) {
                     var startDate,
                         endDate;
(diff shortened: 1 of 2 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=955&repo=edx%2Fecommerce

iivic (approved):
A---....a.bf
This is a valid domain and the only questionable case I could find. If it's acceptable I approve.

ecommerce/static/js/models/coupon_model.js
```diff
@@ -108,7 +108,10 @@ define([
                         return Backbone.Validation.messages.seat_types;
                     }
                 },
-                email_domains: {required: false},
+                email_domains: {
+                    pattern: /^[\w\d][\w\d-\.]*[\w\d]\.[a-z]{2,11}(,[\w\d][\w\d-\.]*[\w\d]\.[a-z]{2,11})*$/,
```
marjev: what about special characters in the domain name? Should domain name start with a number?

Check this: http://regexlib.com/Search.aspx?k=email&c=-1&m=5&ps=20

vkaracic: Domain names can only have dots and hyphens as special characters.
Domains can start with a number. https://www.000webhost.com/

vkaracic: I ended up using the one from here: https://www.safaribooksonline.com/library/view/regular-expressions-cookbook/9781449327453/ch08s15.html

marjev (approved):

PR #957, opened 2016-09-29
Upgrade Django to 1.8.15 (#954)

rlucioni:
@jibsheet @mjfrey @edx/ecommerce FYI. This is a branch off of the latest release tag, release-2016-09-23, with only the Django upgrade on it. I'll be building an AMI off of this hash so we can get the upgrade out.


rlucioni:
Hm, GitHub seems to have closed this on my behalf. Investigating.

rlucioni:
I mistakenly rebased onto master. I will open a separate PR.

PR #960, opened 2016-10-03
Disallow showing the note about verified certificate in baskets with credit seats.

vkaracic:
This PR sets the `is_verification_required` to `False` if the product is a credit seat, and subsequently, the basket does not show the verification message in that case.

https://openedx.atlassian.net/browse/SOL-2093

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -151,7 +151,8 @@ class BasketSummaryView(BasketView):
 
             # Check product attributes to determine if ID verification is required for this basket
             try:
-                is_verification_required = is_verification_required or line.product.attr.id_verification_required
+                is_verification_required = line.product.attr.id_verification_required \
+                    and line.product.attr.certificate_type != 'credit'
             except AttributeError:
                 pass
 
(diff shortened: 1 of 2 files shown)


gsong:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=960

iivic (approved):

ecommerce/extensions/basket/views.py
```diff
@@ -151,7 +151,8 @@ def get_context_data(self, **kwargs):
 
             # Check product attributes to determine if ID verification is required for this basket
             try:
-                is_verification_required = is_verification_required or line.product.attr.id_verification_required
+                is_verification_required = (is_verification_required or line.product.attr.id_verification_required) \
+                    and line.product.attr.certificate_type != 'credit'
```
marjev, line 155:
```
is_verification_required = line.product.attr.id_verification_required and line.product.attr.certificate_type != 'credit'
```

marjev (approved):

PR #961, opened 2016-10-03
[SOL-2042] Change the email validation in offer models to disregard sub domains.

vkaracic:
https://openedx.atlassian.net/browse/SOL-2042

### DIFF
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -1,5 +1,6 @@
 # noinspection PyUnresolvedReferences
 import hashlib
+import re
 
 from django.conf import settings
 from django.core.cache import cache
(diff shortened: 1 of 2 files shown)


gsong:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=961

vkaracic:
@gsong who are you and what have you done to the edX PR bot? :)

gsong:
@vkaracic I'm part of the OS team at edX. I'm doing more work on the integration between JIRA and GitHub right now, and apparently I'm the bot for now. 😜

Looks like it's a problem that the bot even commented on this issue!

ecommerce/extensions/offer/models.py
```diff
@@ -15,10 +15,42 @@ class ConditionalOffer(AbstractConditionalOffer):
...
+                invalid user email domains:
+                    'other.com' etc.
+
+            2)
+                email_domains value: 'sub.example.com'
+                valid user email domain:
+                    'sub.example.com'
+                invalid user email domains:
+                    'sub1.example.com', 'example.com' etc.
+
```
marjev, line 40: Do we want more standardized docstring here, with a short description, list of arguments and return value description.

ecommerce/extensions/offer/models.py
```diff
@@ -15,10 +15,42 @@ class ConditionalOffer(AbstractConditionalOffer):
...
+            for domain in self.email_domains.split(','):
+                # If the domain contains sub domains
+                if len(domain.split('.')) > 2:
+                    if full_user_email_domain == domain:
+                        return True
+                else:
+                    if domain == user_email_domain:
+                        return True
+            return False
+        return True
```
marjev, line 55: Did you try using regex here? Something like `<any_set_of_valid_username_characters>@<any_set_of_valid_domain_characters><domain>`.
If a match is found return True, otherwise check for another domain. This way, you won't have to do all this splitting and joining, which is very confusing.

marjev (changes requested):

ecommerce/extensions/offer/models.py
vkaracic: Involving regex, if anything, adds to the complexity. I'll try with different variable names and/or comments.

vkaracic: I've made some changes, see if it makes more sense now.

marjev: Try replacing

```
# Example user email: username@sub.example.com
full_user_email_domain = email.split('@')[1]  # 'sub.example.com' part
second_and_top_level_domain = full_user_email_domain.split('.')[-2:]
user_email_domain = '.'.join(second_and_top_level_domain)  # 'example.com' part

for domain in self.email_domains.split(','):
    # If the domain contains sub domain(s)
    if len(domain.split('.')) > 2:
        if full_user_email_domain == domain:
            return True
        else:
            if domain == user_email_domain:
                return True
return False
```

with this:

```
for domain in self.email_domains.split(','):
    pattern=r'(?P<username>\w+)@(?P<subdomain>\w+\.)*{domain}'.format(domain=domain)
    match = re.match(pattern, email)
    if match and match.group(0) == email:
        return True
return False
```

I didn't do a lot of testing but it seems to work.

vkaracic: That works and is much better than I had. Thanks!

marjev (approved):
LGTM

iivic (approved):

PR #963, opened 2016-10-03
[SOL-2073] Change the redirect after enrollment code redemption to the receipt page.

vkaracic:
So far when a user clicks 'Enroll' on an enrollment code's offer landing page, the user gets redirect to the dashboard. This PR changes that part of the code so that when a user clicks on that button it gets redirected to the receipt page.


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=963&repo=edx%2Fecommerce

ecommerce/coupons/views.py
```diff
@@ -161,12 +162,15 @@ def get(self, request):
 
         basket = prepare_basket(request, product, voucher)
         if basket.total_excl_tax == 0:
-            self.place_free_order(basket)
+            order = self.place_free_order(basket)
+            if order.status == 'Complete':
```
mjfrey: This won't work with async fulfillment.  The order status will never be 'Complete' this fast.  Need a way from the receipt page or some view to wait for fulfillment to complete.

mjfrey (changes requested):

ecommerce/coupons/views.py
vkaracic: What do you suggest?

ecommerce/coupons/views.py
```diff
@@ -161,12 +162,15 @@ def get(self, request):
...
+                redirect_path = get_receipt_page_url(self.request.site.siteconfiguration, order.number)
+            else:
+                redirect_path = reverse('checkout:error')
```
iivic: This can be rewritten like
```
if order.status == 'Complete':
    return HttpResponseRedirect(get_receipt_page_url(self.request.site.siteconfiguration, order.number))
return HttpResponseRedirect(reverse('checkout:error'))
```

ecommerce/coupons/views.py
```diff
@@ -161,12 +162,15 @@ def get(self, request):
...
+            return HttpResponseRedirect(redirect_path)
         else:
```
iivic: This else can be removed.

iivic (changes requested):

vkaracic:
Here's how I intended it:
out-5

On localhost this will finish instantly so I've paused the fulfillment so that you can see the waiting process.

vkaracic:
@iivic @marjev @mjfrey comments addressed, please take another look.

ecommerce/coupons/views.py
```diff
@@ -161,11 +164,22 @@ def get(self, request):
...
+            # Every 2 seconds the order status is checked. If after three checks
+            # it's not complete or the status is an error, user is redirected
+            # to the checkout error page.
+            for __ in range(3):  # pragma: no cover
+                order.refresh_from_db()
+                if order.status == ORDER.COMPLETE:
+                    return HttpResponseRedirect(get_receipt_page_url(self.request.site.siteconfiguration, order.number))
+                elif order.status == ORDER.FULFILLMENT_ERROR:
+                    break
+                time.sleep(2)  # pragma: no cover
```
iivic, line 179: Why 2 seconds?

ecommerce/static/js/views/offer_view.js
```diff
@@ -24,14 +24,27 @@ define([
             events: {
                 'click .prev': 'previous',
                 'click .next': 'next',
-                'click .page-number': 'goToPage'
+                'click .page-number': 'goToPage',
+                'click #RedeemEnrollment': 'disableReedemEnrollmentBtn'
```
iivic: This is not part of this PR but for a multi-course coupon on offer landing page there will be multiple links with the id="RedeemEnrollment" or "PurchaseCredit". The id should be unique. This is something we should take care of ....

ecommerce/coupons/views.py
```diff
@@ -161,11 +164,22 @@ def get(self, request):
...
+            # Waiting for the order to finish its async fulfillment.
+            # Every 2 seconds the order status is checked. If after three checks
+            # it's not complete or the status is an error, user is redirected
+            # to the checkout error page.
+            for __ in range(3):  # pragma: no cover
+                order.refresh_from_db()
+                if order.status == ORDER.COMPLETE:
+                    return HttpResponseRedirect(get_receipt_page_url(self.request.site.siteconfiguration, order.number))
+                elif order.status == ORDER.FULFILLMENT_ERROR:
+                    break
```
marjev: why not `return HttpResponseRedirect(reverse('checkout:error'))`?

ecommerce/coupons/views.py
marjev: Are the 2 seconds a constant somewhere? It seems to me like this might change so we'll have to do it here as well. Using and importing the constant here might prevent potential future headaches.

ecommerce/templates/oscar/checkout/error.html
```diff
@@ -4,23 +4,27 @@
 
 {% block title %}
     {% trans "Checkout Error" %}
-{% endblock title %}
+{% endblock %}
 
 {% block navbar %}
     {% include 'edx/partials/_student_navbar.html' %}
 {% endblock navbar %}
```
marjev: nitpick: not your doing, but can we remove the navbar part from the endblock as well.

marjev (changes requested):

ecommerce/coupons/views.py
vkaracic: The two seconds are arbitrary, we can add any value here, 2 seems sufficient.
You mean saving this as a constant in the settings? When will this change?

marjev: If it's related to the business logic of a model, it makes more sense to me to place it under that model where the 2 seconds rule is enforced.
Also, as you said, 2 seconds are arbitrary, which means it can change at any time, e. g. when a problem occurs. When that happens, hypothetically speaking, we should minimize the work that needs to be done by DRY-ing this.

marjev (approved):

vkaracic:
@clintonb can you please take a look? Specifically at the way I've implemented the waiting for the order to fulfill?

ecommerce/core/models.py
```diff
@@ -277,7 +277,7 @@ def student_dashboard_url(self):
     @property
     def enrollment_api_url(self):
         """ Returns the URL for the root of the Enrollment API. """
-        return self.build_lms_url('/api/enrollment/v1/')
+        return self.build_lms_url('/api/enrollment/v1/enrollment')
```
clintonb: Why the change? We should always point to the API root, not a specific resource.

ecommerce/extensions/order/models.py
```diff
@@ -21,6 +21,9 @@ def contains_coupon(self):
         return any(line.product.get_product_class().name == 'Coupon' for line in
                    self.basket.all_lines())
 
+    class Meta(object):
+        get_latest_by = 'date_placed'
```
clintonb, line 25: I believe this requires a migration.

vkaracic:
Comments addressed!

iivic (approved):

clintonb (approved):

mjfrey (approved):

PR #966, opened 2016-10-07
Updated i18n-tool version, .travis.yml and Makefile files changed

iivic:
@clintonb  Are those the required changes?


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=966&repo=edx%2Fecommerce

Makefile
```diff
@@ -111,6 +112,9 @@ push_translations:
 
 update_translations: pull_translations fake_translations
 
+changed_translations:
+	cd ecommerce && i18n-tool changed
+
```
clintonb, line 121: Add a new target, `validate_translations`, that calls `fake_translations` followed by `changed_translations`.

.travis.yml
```diff
@@ -34,5 +34,8 @@ script:
     - make static
     - make validate_python
     - make validate_js
+
+    # Check if translation files are up-to-date
```
clintonb: Call this before asset compilation. We want to fail on the easy stuff early so that devs can quickly get a fix pushed.

Makefile
```diff
@@ -111,6 +112,9 @@ push_translations:
 
 update_translations: pull_translations fake_translations
 
+changed_translations:
```
clintonb: How about `detect_missing_source_translations`, or something that better reflects what we want to achieve?
Also, add a comment mentioning the need to run the extract command first.

clintonb (changes requested):

.travis.yml
```diff
@@ -29,13 +29,13 @@ script:
     - cd docs && make html
     - cd ..
 
+    # Check if translation files are up-to-date
+    - make detect_changed_source_translations
```
iivic: Is this command name more descriptive? @clintonb

iivic:
Is detect_changed_source_translations more descriptive? 
I also updated en source files and committed the eo dummy files but I'm not sure if I should have done that and when those files need to be committed.

clintonb (approved):

clintonb:
Looks good. I don't think we have rules for the fake translations. However, it makes sense to include them so that we can test for missing translations in non-local environments. Also, remembers to squash commits before merging.

iivic:
I added `pip install edx-i18n-tools` to the .travis.yml file so that the tests pass. Any objections?

clintonb:
Big objection: https://github.com/edx/ecommerce/blob/master/requirements/local.txt#L9. 

Please ensure you are testing locally by setting up a fresh virtualenv for each project, and running `make requirements`.

ecommerce/conf/locale/en/LC_MESSAGES/django.po
```diff
@@ -29,131 +29,142 @@ msgstr ""
 msgid "Important dates"
 msgstr ""
 
-#: ecommerce/core/models.py:37
+#: ecommerce/core/models.py:38
```
nedbat, line 32: It's annoying that the line numbers here change so often and clutter the diffs.  In i18n-tools, the line numbers are removed during the "generate" step which merges .po files together.  You don't need merge, so you aren't getting the line number scrubbing.  Maybe we should change where that happens (to extract) so that other repos get the benefit.

iivic:
@clintonb  Travis doesn't install edx-i18n-tools because the requirements from local.txt file are not installed. Travis calls `make requirements.js` perhaps `make requirements` should be called instead of 
`pip install -r requirements/test.txt
  pip install -r requirements/docs.txt
  make requirements.js`
because it will install all of that and i18n requirements or should I just move the i18n tools to another requirements file?

ecommerce/conf/locale/en/LC_MESSAGES/django.po
iivic: I'm not sure what you're trying to say and what needs to be done here.

clintonb: Nothing needs to be done now. What @nedbat mentioned is a feature request/enhancement that is out-of-scope for your work @iivic.

clintonb:
@iivic yes, please update `.travis.yml` to run `make requirements`. I'm not sure what that wasn't the case before.

iivic:
Please take a quick look just to make sure I didn't forget something.

.travis.yml
```diff
@@ -21,14 +21,15 @@ before_install:
...
-    - pip install -r requirements/docs.txt
-    - make requirements.js
+    - make requirements
 script:
     # Ensure documentation can be compiled
     - cd docs && make html
     - cd ..
 
+    # Check if translation files are up-to-date
+    - make detect_changed_source_translations
```
clintonb: `make validate_translations`

iivic:
Updated. Good to go?

clintonb (approved):

PR #968, opened 2016-10-07
Add UTM tracking for order referrals

bderusha:
[ECOM-5539](https://openedx.atlassian.net/browse/ECOM-5539)

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -94,6 +94,20 @@ class SiteConfiguration(models.Model):
         max_length=255,
         blank=True
     )
+    utm_cookie_name = models.CharField(
+        verbose_name=_('UTM Cookie Name'),
+        help_text=_('Name of cookie storing UTM data.'),
+        max_length=255,
+        blank=True,
+        default="",
+    )
+    affiliate_cookie_name = models.CharField(
+        verbose_name=_('Affiliate Cookie Name'),
+        help_text=_('Name of cookie storing affiliate data.'),
+        max_length=255,
+        blank=True,
+        default="",
+    )
 
     class Meta(object):
         unique_together = ('site', 'partner')
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -15,6 +15,7 @@ from rest_framework.response import Response
 from ecommerce.extensions.analytics.utils import audit_log
 from ecommerce.extensions.api import data as data_api, exceptions as api_exceptions
 from ecommerce.extensions.api.serializers import OrderSerializer
+from ecommerce.extensions.basket.utils import attribute_cookie_data
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
 from ecommerce.extensions.payment import exceptions as payment_exceptions
 from ecommerce.extensions.payment.helpers import (get_default_processor_class, get_processor_class_by_name)
@@ -132,6 +133,8 @@ class BasketCreateView(EdxOrderPlacementMixin, generics.CreateAPIView):
         with transaction.atomic():
             basket = Basket.create_basket(request.site, request.user)
             basket_id = basket.id
+
+            attribute_cookie_data(basket, request)
 
             requested_products = request.data.get('products')
             if requested_products:
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -1,12 +1,17 @@
+import datetime
+import json
+import mock
+
 import ddt
 from django.test import RequestFactory
 from oscar.core.loading import get_model
-from oscar.test.factories import ProductFactory, RangeFactory, VoucherFactory
+from oscar.test.factories import BasketFactory, ProductFactory, RangeFactory, VoucherFactory
+import pytz
 
 from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME, ENROLLMENT_CODE_SWITCH
 from ecommerce.core.tests import toggle_switch
 from ecommerce.courses.tests.factories import CourseFactory
-from ecommerce.extensions.basket.utils import prepare_basket
+from ecommerce.extensions.basket.utils import prepare_basket, attribute_cookie_data
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.extensions.partner.models import StockRecord
 from ecommerce.extensions.test.factories import prepare_voucher
@@ -29,6 +34,7 @@ class BasketUtilsTests(CourseCatalogTestMixin, TestCase):
         self.request.COOKIES = {}
         self.request.user = self.create_user()
         site_configuration = SiteConfigurationFactory(partner__name='Tester')
+        site_configuration.utm_cookie_name = 'test.edx.utm'
         self.request.site = site_configuration.site
 
     def test_prepare_basket_with_voucher(self):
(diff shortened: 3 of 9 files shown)


bderusha:
@edx/ecommerce Please review

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -120,15 +133,142 @@ def test_prepare_basket_affiliate_cookie_lifecycle(self):
...
+            'utm_term': utm_term,
+            'utm_content': utm_content,
+            'created_at': utm_created_at,
+        }
+
+        self.request.COOKIES['test.edx.utm'] = json.dumps(utm_cookie)
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        attribute_cookie_data(basket, self.request)
+
+        # test affiliate id from cookie saved in referral
```
schenedx: I feel this is a different test case. Why put these scenarios into a single test case?

ecommerce/extensions/basket/utils.py
```diff
@@ -91,3 +87,65 @@ def get_basket_switch_data(product):
             partner_sku = stock_record.partner_sku
             break
     return switch_link_text, partner_sku
+
+
+def attribute_cookie_data(basket, request):
+    try:
+        try:
```
schenedx: This looks a bit weird. Maybe it's the first time I am seeing it. That said, maybe we can wrap this second try into a new function like `_get_referral_of_basket()`?

ecommerce/referrals/models.py
```diff
@@ -9,6 +9,23 @@
...
+        'affiliate_id',
+        'utm_source',
+        'utm_medium',
+        'utm_campaign',
+        'utm_term',
+        'utm_content',
+        'utm_created_at',
+    )
+
+    site = models.ForeignKey('sites.Site', null=False, blank=False)
```
schenedx, line 22: Does this mean the referral is now multi-tenancy ready?

schenedx (approved):
All minor comments. 👍

ecommerce/extensions/basket/tests/test_utils.py
bderusha: Because each subsequent test would involve doing the all of the steps before it. 

create cookie data
assert cookie data saved

create cookie data
assert cookie data saved
update cookie
assert new data saved

create cookie data
assert cookie data saved
delete cookie
assert cookie data deleted

It seemed more straightforward to test the entire lifecycle in a single test.

ecommerce/referrals/models.py
bderusha: For ecommerce, yes.  For LMS it is not yet

ecommerce/extensions/basket/utils.py
bderusha: Agreed that it looks weird.  Will update.

ecommerce/core/models.py
```diff
@@ -94,6 +94,20 @@ class SiteConfiguration(models.Model):
         max_length=255,
         blank=True
     )
+    utm_cookie_name = models.CharField(
+        verbose_name=_('UTM Cookie Name'),
+        help_text=_('Name of cookie storing UTM data.'),
+        max_length=255,
+        null=True,
```
rlucioni: From the [Django docs](https://docs.djangoproject.com/en/1.10/ref/models/fields/#null):

> Avoid using null on string-based fields such as CharField and TextField because empty string values will always be stored as empty strings, not as NULL. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -133,6 +134,8 @@ def create(self, request, *args, **kwargs):
             basket = Basket.create_basket(request.site, request.user)
             basket_id = basket.id
 
+            attribute_cookie_data(basket, request)
```
rlucioni, line 137: @mjfrey would know better, but I think you may also need to update the checkout view.

ecommerce/extensions/basket/utils.py
```diff
@@ -91,3 +87,70 @@ def get_basket_switch_data(product):
...
+        # Save the record if any attribution attributes are set on it.
+        if any([getattr(referral, attribute) for attribute in Referral.ATTRIBUTION_ATTRIBUTES]):
+            referral.save()
+        # Clean up the record if no attribution attributes are set and it exists in the DB.
+        elif referral.pk:
+            referral.delete()
+        # Otherwise we can ignore the instantiated but unsaved referral
+
+    # Don't let attribution errors prevent users from creating baskets
+    except Exception:  # pylint: disable=broad-except
```
rlucioni: You could just use a bare except here.

ecommerce/extensions/basket/utils.py
```diff
@@ -91,3 +87,70 @@ def get_basket_switch_data(product):
...
+        referral = Referral(basket=basket, site=site)
+    return referral
+
+
+def _record_affiliate_basket_attribution(referral, request):
+    """
+      Attribute this user's basket to the referring affiliate, if applicable.
+    """
+
+    # TODO: update this line to use site configuration once config in production (2016-10-04)
```
rlucioni, line 125: Is there a JIRA ticket to track this?

ecommerce/extensions/basket/utils.py
```diff
@@ -91,3 +87,70 @@ def get_basket_switch_data(product):
...
+
+def _record_utm_basket_attribution(referral, request):
+    """
+      Attribute this user's basket to UTM data, if applicable.
+    """
+    utm_cookie_name = request.site.siteconfiguration.utm_cookie_name
+    utm_cookie = request.COOKIES.get(utm_cookie_name, "{}")
+    utm = json.loads(utm_cookie)
+
+    referral.utm_source = utm.get('utm_source')
```
rlucioni: More concisely:

``` py
for param in ['utm_source', ...]:
    setattr(referral, param, utm.get(param))
```

ecommerce/referrals/models.py
```diff
@@ -9,6 +9,23 @@
...
     basket = models.OneToOneField('basket.Basket', null=True, blank=True)
     order = models.OneToOneField('order.Order', null=True, blank=True)
+    affiliate_id = models.CharField(_('Affiliate ID'), null=True, blank=True, default=None, max_length=255)
+    utm_source = models.CharField(_('UTM Source'), null=True, blank=True, default=None, max_length=255)
+    utm_medium = models.CharField(_('UTM Medium'), null=True, blank=True, default=None, max_length=255)
+    utm_campaign = models.CharField(_('UTM Campaign'), null=True, blank=True, default=None, max_length=255)
```
rlucioni: Same comment from before about null and blank on a CharField applies here.

rlucioni (changes requested):

ecommerce/extensions/api/v2/views/baskets.py
mjfrey: If you want to track cancel/errors you might want to add tracking to the checkout views.

ecommerce/extensions/basket/utils.py
bderusha: 🐻 fixed

ecommerce/extensions/basket/utils.py
bderusha: [ECOM-6008](https://openedx.atlassian.net/browse/ECOM-6008)

ecommerce/extensions/api/v2/views/baskets.py
bderusha: Just need to track basket creation.

bderusha:
@rlucioni changes made please rereview

ecommerce/core/models.py
```diff
@@ -94,6 +94,20 @@ class SiteConfiguration(models.Model):
...
+        max_length=255,
+        blank=True,
+        default="",
+    )
+    affiliate_cookie_name = models.CharField(
+        verbose_name=_('Affiliate Cookie Name'),
+        help_text=_('Name of cookie storing affiliate data.'),
+        max_length=255,
+        blank=True,
+        default="",
```
rlucioni, line 109: Did you re-create migrations after making these model changes? If so, this default doesn't seem to have made it into the migrations.

bderusha: Good catch!!!  Updated

rlucioni (approved):

ecommerce/referrals/models.py
clintonb: This should be `null=True`. This is currently breaking on stage. For whatever reason, Django has assigned a default of 6.

ecommerce/referrals/migrations/0002_auto_20161011_1728.py
```diff
@@ -0,0 +1,56 @@
...
+    dependencies = [
+        ('sites', '0001_initial'),
+        ('referrals', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='referral',
+            name='site',
+            field=models.ForeignKey(default=6, to='sites.Site'),
```
clintonb, line 18: The source of many problems.

PR #969, opened 2016-10-11
Fixed Sailthru signal bug

clintonb:
ECOM-5955

### DIFF
--- a/ecommerce/sailthru/signals.py
+++ b/ecommerce/sailthru/signals.py
@@ -106,7 +106,7 @@ def process_basket_addition(sender, product=None, user=None, request=None, baske
             BasketAttribute.objects.update_or_create(
                 basket=basket,
                 attribute_type=get_basket_attribute_type(),
-                value_text=message_id
+                defaults={'value_text': message_id}
             )
 
         # inform sailthru if there is a price.  The purpose of this call is to tell Sailthru when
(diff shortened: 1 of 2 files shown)


clintonb:
@edx/ecommerce please review

ecommerce/sailthru/signals.py
```diff
@@ -127,10 +125,10 @@ def _build_course_url(course_id):
...
-    except BasketAttributeType.DoesNotExist:
-        attribute_type = BasketAttributeType.objects.create(name=SAILTHRU_CAMPAIGN)
-    return attribute_type
+def get_basket_attribute_type():
+    """ Returns `BasketAttributeType` for Sailthru campaign ID.
+
+    Returns:
+        BasketAttributeType
+    """
+    return BasketAttributeType.objects.get(name=SAILTHRU_CAMPAIGN)
```
rlucioni, line 134: Why is it that we can get rid of the exception handler? Do we always expect there to be a matching `BasketAttributeType`?

rlucioni (approved):
Thank you very much for addressing this.

ecommerce/sailthru/signals.py
clintonb: The handler seems to have been added before we had a migration creating the `BasketAttributeType`.

rlucioni: Got it, thanks.

PR #972, opened 2016-10-17
Added verbosity to pip requirements installation

clintonb:
I have removed the quiet argument so that we can see exactly what is installed when this command is called. This is primarily useful for knowing what is installed by Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ cache:
     - node_modules
     - ecommerce/static/bower_components
 addons:
+    firefox: latest
     apt:
         packages:
             - lcov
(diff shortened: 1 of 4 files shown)


clintonb:
FYI @edx/ecommerce @marjev

marjev:
@clintonb Check this out: https://gist.github.com/tristandostaler/2f8b28f2bf503db4422a5549e8fed538

Not sure if that will help, but might be a good starting point. If we plan to update the Selenium and Firefox versions, looks like a way to go.
Also, official documentation: https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette/WebDriver

clintonb:
@marjev nice find! My latest fixup commit incorporates that new knowledge. Here's hoping it resolves the issue!

marjev:
🎆 It works!! 🍾

PR #975, opened 2016-10-20
[SOL-1823] Add acceptance test for creating dynamic coupon.

vkaracic:
This PR contains:
- fixes for the professional education enrollment test for which the logic is different when checkout on ecommerce is enabled.
- new acceptance test for creating a dynamic coupon. Since the `tearDown` method in coupon admin tests sends requests to the ecommerce server after every test case to delete the created coupons, adding a new one resulted in the server returning 429 error, so I added a sleep period to circumvent that.

https://openedx.atlassian.net/browse/SOL-1823


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=975&repo=edx%2Fecommerce

vkaracic:
@marjev @iivic @mjfrey please review.

acceptance_tests/mixins.py
```diff
@@ -378,6 +379,9 @@ def get_and_delete_coupons(self, partial_title):
         Get all coupons created in the acceptance tests and delete those
         and all associated stock records and vouchers.
         """
+        # Tear-down's after every test cause the server to return 429 errors,
+        # Therefor there needs to be a cooling period between them.
+        time.sleep(10)
```
marjev, line 384: http://bok-choy.readthedocs.io/en/latest/guidelines.html#never-use-time-sleep

vkaracic:
@edx/ecommerce can anyone please review this?

marjev (commented):
Just one comment. More of a suggestion.

Perhaps you should tag someone from ecommerce or testeng team to take a look at this. They should be more helpful.

acceptance_tests/mixins.py
vkaracic: I'm aware of that, but this is not a `PageObject` class.

acceptance_tests/mixins.py
```diff
@@ -378,6 +379,9 @@ def get_and_delete_coupons(self, partial_title):
         Get all coupons created in the acceptance tests and delete those
         and all associated stock records and vouchers.
         """
+        # Tear-down's after every test cause the server to return 429 errors,
```
iivic: Tear-down's ?

acceptance_tests/pages/coupons.py
```diff
@@ -43,26 +49,32 @@ def fill_create_coupon_form(self, is_discount):
...
-        self.q(css=course_id_input).fill(VERIFIED_COURSE_ID)
-        self.wait_for_ajax()
 
-        wait = WebDriverWait(self.browser, 2)
-        verified_option_present = EC.presence_of_element_located(
-            (By.CSS_SELECTOR, 'select[name="seat_type"] option[value="Verified"]')
-        )
-        wait.until(verified_option_present)
+        if is_dynamic:
+            self.q(css='input#multiple-courses').click()
```
iivic, line 54: Can you just select by the id #multiple-courses?

acceptance_tests/pages/coupons.py
```diff
@@ -43,26 +49,32 @@ def fill_create_coupon_form(self, is_discount):
...
+            self.q(css='textarea#catalog-query').fill('org:TestX')
```
iivic, line 55: Same as mentioned.

acceptance_tests/pages/coupons.py
```diff
@@ -43,26 +49,32 @@ def fill_create_coupon_form(self, is_discount):
...
+            self.q(css='input#professional').click()
```
iivic, line 56: Same as mentioned.

acceptance_tests/mixins.py
marjev: The problem with this is that we can't really say why the sleep has to last 10s.

iivic (changes requested):
Few changes before approving.

acceptance_tests/pages/coupons.py
vkaracic: I could but the convention for this part of the code seems to be the selector that I used.

iivic (approved):

acceptance_tests/config.py
```diff
@@ -50,3 +50,5 @@
 ENABLE_COUPON_ADMIN_TESTS = str2bool(os.environ.get('ENABLE_COUPON_ADMIN_TESTS', False))
 
 BULK_PURCHASE_SKU = os.environ.get('BULK_PURCHASE_SKU')
+
+CHECKOUT_ON_ECOMMERCE = os.environ.get('CHECKOUT_ON_ECOMMERCE')
```
mjfrey: just make this always true and remove this flag.  Shopping cart is now deprecated.

vkaracic: In that case should there be a ticket for remove the shopping cart code?

vkaracic:
@edx/ecommerce again, can someone please review this?

acceptance_tests/mixins.py
```diff
@@ -378,6 +379,9 @@ def get_and_delete_coupons(self, partial_title):
         Get all coupons created in the acceptance tests and delete those
         and all associated stock records and vouchers.
         """
+        # Teardowns after every test cause the server to return 429 errors,
+        # Therefor there needs to be a cooling period between them.
+        time.sleep(10)
```
clintonb, line 384: The current rate limit is 50 requests per minute. How many requests are you making, that this is necessary?

acceptance_tests/test_coupon_admin.py
```diff
@@ -20,8 +22,8 @@ def setUp(self):
         self.coupons_list_page = CouponsListPage(self.browser)
         self.login_with_lms()
 
-    def create_coupon(self):
-        """ Create a coupon via UI. """
+    def create_coupon(self, is_dynamic=False):
+        """ Create a single-course and dynamic coupon via UI. """
```
clintonb: Aren't those the only two variants of a coupon? `"Create a coupon via UI."`

acceptance_tests/test_profed_enrollment.py
```diff
@@ -25,10 +26,12 @@ def test_payment_required(self):
             # Click the first enroll button on the page to take the browser to the track selection page,
             # and allow it to load.
             course_about_page.q(css='.js-enroll-btn').first.click()
-            WebDriverWait(self.browser, 10).until(EC.presence_of_element_located((By.CLASS_NAME, 'basket')))
+            WebDriverWait(self.browser, 10).until(on_basket_page)
         else:
-            # Visit the course mode page (where auto-enrollment normally occurs)
-            LMSCourseModePage(self.browser, PROFESSIONAL_COURSE_ID).visit()
+            # Clicking on the enroll button the user gets redirected to the ecommerce basket.
```
clintonb: s/the user gets redirected/redirects the user

acceptance_tests/mixins.py
vkaracic: I doubt it's more than 50 but without this I get 429 status code errors. This could be due to my local set-up. Is there a way to run the acceptance tests on stage? If not I guess I could run them on a sandbox and test there.

clintonb:

PR #976, opened 2016-10-21
[SOL-2109] Change the default values of max_uses field in coupon form.

vkaracic:
When the user selects 'Can be used multiple times by multiple users'
usage limitation option, the default value would be 1 which doesn't
make sense for multi-use vouchers. Instead now its default is blank
which, if left that way, will default to 10000 which is the arbitrary
limit that Oscar sets for this attribute, and the minimum limit is set
to 2. Values of other usage limitations are left the same.

https://openedx.atlassian.net/browse/SOL-2109

### DIFF
--- a/ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
+++ b/ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
@@ -41,11 +41,11 @@ define([
 
                 beforeEach(function () {
                     jasmine.clock().install();
-                    seatType = view.$el.find('[name=seat_type]');
+                    seatType = view.$('[name=seat_type]');
                     spyOn($, 'ajax').and.callFake(function (options) {
                         options.success(courseData);
                     });
-                    view.$el.find('[name=course_id]').val(courseId).trigger('input');
+                    view.$('[name=course_id]').val(courseId).trigger('input');
                     // event is debounced, override _.now to trigger immediately
                     spyOn(_, 'now').and.returnValue(Date.now() + 110);
                     jasmine.clock().tick(110);
@@ -79,7 +79,7 @@ define([
 
             describe('enrollment code', function () {
                 beforeEach(function () {
-                    view.$el.find('[name=code_type]').val('enrollment').trigger('change');
+                    view.$('[name=code_type]').val('enrollment').trigger('change');
                 });
 
                 it('should show the price field', function () {
@@ -127,7 +127,7 @@ define([
                 ];
 
                 beforeEach(function () {
-                    view.$el.find('[name=code_type]').val('Discount code').trigger('change');
+                    view.$('[name=code_type]').val('Discount code').trigger('change');
                 });
 
                 it('should show the discount field', function () {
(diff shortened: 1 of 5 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=976

mjfrey:
👍

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
```diff
@@ -172,11 +172,22 @@ define([
                     expect(SpecUtils.formGroup(view, '[name=code]')).toBeVisible();
                 });
 
-                it('should show the usage number field only for once-per-customer vouchers', function () {
+                it('should show the max_uses field only for once-per-customer and multi-use vouchers', function () {
                     view.$el.find('[name=voucher_type]').val('Single use').trigger('change');
                     expect(SpecUtils.formGroup(view, '[name=max_uses]')).not.toBeVisible();
                     view.$el.find('[name=voucher_type]').val('Once per customer').trigger('change');
                     expect(SpecUtils.formGroup(view, '[name=max_uses]')).toBeVisible();
+                    view.$el.find('[name=voucher_type]').val('Multi-use').trigger('change');
```
iivic: Nitpick: use shortcut for view.$el.find (view.$)

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
```diff
@@ -172,11 +172,22 @@ define([
...
+                    expect(SpecUtils.formGroup(view, '[name=max_uses]')).toBeVisible();
+                });
+
+                it('should set different values for max_uses field for different voucher types', function() {
+                    view.$el.find('[name=voucher_type]').val('Once per customer').trigger('change');
```
iivic: Nitpick: use shortcut for view.$el.find (view.$)

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
```diff
@@ -172,11 +172,22 @@ define([
...
+                    expect(view.$('[name=max_uses]').val()).toBe('1');
+                    expect(view.$('[name=max_uses]').attr('min')).toBe('1');
+                    view.$el.find('[name=voucher_type]').val('Multi-use').trigger('change');
```
iivic: Nitpick: use shortcut for view.$el.find (view.$)

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -421,21 +421,32 @@ define([
...
-                    this.formGroup('[name=max_uses]').removeClass(this.hiddenClass);
-                } else {
-                    this.hideField('[name=max_uses]', 1);
+                    this.formGroup(maxUsesFieldSelector).removeClass(this.hiddenClass);
+                    /* For coupons that can be used multiple times by multiple users, the max_uses
+                     * field needs to be empty by default and the minimum can not be less than 2.
+                     */
+                    if (voucherType === 'Multi-use') {
+                        this.model.set('max_uses', null);
+                        this.$(maxUsesFieldSelector).attr('min', 2);
```
iivic, line 445: Maybe we should do the validation with backbone validation instead of this?

iivic (changes requested):

vkaracic:
Comments addressed.

iivic (approved):

PR #978, opened 2016-10-24
[SOL-1954, SOL-1955, SOL-1956] New Receipt Page - Frontend

marjev:
@mjfrey Here's a frontend part of the Receipt Page PR. I don't want to split this any further because I'm afraid we might lose the bigger picture that this PR offers. Any suggested changes should be applied, if needed, across the entire PR.

@vkaracic please review as well.

Acceptance tests will be the 3rd PR.

- [x] UX review
- [x] Text review

Latest screenshot:

image


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=978&repo=edx%2Fecommerce

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,161 @@ def get_context_data(self, **kwargs):
...
+                    'error_text': _('Order {order_number} not found.').format(order_number=order_number),
+                    'for_help_text': '',
+                    'is_payment_complete': False,
+                    'page_title': _('Order not found')
+                })
+                return self.render_to_response(context)
+
+            seat = order.lines.first().product
+            site_configuration = self.request.site.siteconfiguration
+            course = site_configuration.course_catalog_api_client.course_runs(seat.attr.course_key).get()
```
mjfrey: Don't you need to catch an exception here?  What if we do not find the course from the catalog?

mjfrey: actually -- why don't we just use the course info already in otto.  No need to call the catalog service.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,161 @@ def get_context_data(self, **kwargs):
...
+                )
+
+            order_data = OrderSerializer(order, context={'request': self.request}).data
+            discount_value = float(order_data['discount'])
+            total_cost = float(order_data['total_excl_tax'])
+            original_cost = discount_value + total_cost
+
+            try:
+                processor_response = PaymentProcessorResponse.objects.filter(basket=order.basket).latest()
+                if processor_response and processor_response.processor_name == 'paypal':
```
mjfrey: This logic should be part of the ProcessorResponse class as a method.  get_recipient_info.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,161 @@ def get_context_data(self, **kwargs):
...
+        payment_support_link = '<a href="mailto:{email}">{email}</a>'.format(email=payment_support_email)
+
+        context.update({
+            'dashboard': get_lms_url('/dashboard'),
+            'error_summary': _('An error occurred while creating your receipt.'),
+            'error_text': None,
+            'for_help_text': _(
+                "If your course does not appear on your dashboard, contact {payment_support_link}."
+            ).format(payment_support_link=payment_support_link),
+            'is_payment_complete': True,
```
mjfrey: is this ever False?

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,161 @@ def get_context_data(self, **kwargs):
...
+            'lms_url': get_lms_url(),
+            'name': '{} {}'.format(self.request.user.first_name, self.request.user.last_name),
+            'nav_hidden': True,
```
mjfrey: same here

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -122,3 +145,112 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        self.assertEqual(response.status_code, 200)
+        self.assertDictContainsSubset(receipt_data, response.context_data['receipt'])
+        self.assertDictContainsSubset(context_data, response.context_data)
+
+    @httpretty.activate
+    def post_to_receipt_page(self, post_data):
+        httpretty.register_uri(httpretty.POST, self.path, body=post_data)
+
+        response = self.client.post(self.path, params={'basket_id': 1}, data=post_data)
+        self.assertEqual(response.status_code, 200)
```
mjfrey, line 277: this is a helper function not a test.  You should check for this in the actual test that calls it.

mjfrey:
@clintonb can you take a look?

ecommerce/extensions/checkout/app.py
```diff
@@ -7,12 +7,14 @@ class CheckoutApplication(app.CheckoutApplication):
...
     cancel_checkout = get_class('checkout.views', 'CancelCheckoutView')
     checkout_error = get_class('checkout.views', 'CheckoutErrorView')
+    receipt_response = get_class('checkout.views', 'ReceiptResponseView')
 
     def get_urls(self):
         urls = [
             url(r'^free-checkout/$', self.free_checkout.as_view(), name='free-checkout'),
             url(r'^cancel-checkout/$', self.cancel_checkout.as_view(), name='cancel-checkout'),
             url(r'^error/', self.checkout_error.as_view(), name='error'),
+            url(r'^receipt/', self.receipt_response.as_view(), name='receipt'),
```
clintonb, line 17: Why not use the existing Oscar view, `thank-you`? (Example URL: https://ecommerce.edx.org/checkout/thank-you/?order_number=EDX-1234)

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -122,3 +145,112 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        response = self.client.post(self.path)
+        self.assertEqual(response.status_code, 302)
+
+    def test_get_receipt_for_nonexisting_order(self):
+        order_number = 'ABC123'
+        response = self.client.get('{path}?order_number={order_number}'.format(
+            order_number=order_number,
+            path=self.path
+        ))
+        self.assertEqual(response.status_code, 200)
```
clintonb: The status code should be 404.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -122,3 +145,112 @@ def test_post_returns_payment_support_email_in_context(self):
...
+
+    def setUp(self):
+        super(ReceiptViewTests, self).setUp()
+        self.user = self.create_user()
+        self.client.login(username=self.user.username, password=self.password)
+
+    def test_login_required(self):
+        """ The view should redirect to the login page if the user is not logged in. """
+        self.client.logout()
+        response = self.client.post(self.path)
```
clintonb: What about `GET` operations?

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -122,3 +145,112 @@ def test_post_returns_payment_support_email_in_context(self):
...
+            'provider_data': None,
+            'verified': seat.attr.certificate_type == 'verified'
+        }
+
+        self.assertEqual(response.status_code, 200)
+        self.assertDictContainsSubset(receipt_data, response.context_data['receipt'])
+        self.assertDictContainsSubset(context_data, response.context_data)
+
+    @httpretty.activate
+    def post_to_receipt_page(self, post_data):
```
clintonb: Add a docstring.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -122,3 +145,112 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        return response
+
+    @ddt.data('ACCEPT', 'REJECT', 'ERROR')
+    def test_cybersource_decision(self, decision):
+        """
+        Ensure that when Cybersource sends a response back to the view, it renders the page title appropriately
+        depending on the decision code provided in response data.
+        """
+        post_data = {'decision': decision, 'reason_code': '200', 'signed_field_names': 'dummy'}
+        expected_pattern = r"<title>(\s+)Receipt" if decision == 'ACCEPT' else r"<title>(\s+)Payment Failed"
```
clintonb: Split this into two tests—one for accept, the other for reject/error. A helper function will help reduce duplication.

For future reference: An `if-else` in a test is a sign that there should be a separate test case.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -122,3 +145,112 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        self.assertRegexpMatches(response.content, expected_pattern)
+
+    def test_hide_nav_header(self):
+        """
+        Verify that the header navigation links are hidden for the edx.org version
+        """
+        post_data = {'decision': 'ACCEPT', 'reason_code': '200', 'signed_field_names': 'dummy'}
+        response = self.post_to_receipt_page(post_data)
+
+        self.assertNotContains(response, "How it Works")
```
clintonb: I don't see these strings anywhere in the codebase. Why is this test necessary?

ecommerce/extensions/checkout/views.py
```diff
@@ -1,25 +1,28 @@
 """ Checkout related views. """
 from __future__ import unicode_literals
-
```
clintonb, line 3: Restore.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+                    site_configuration=site_configuration
+                )
+
+            order_data = OrderSerializer(order, context={'request': self.request}).data
+            discount_value = float(order_data['discount'])
+            total_cost = float(order_data['total_excl_tax'])
+            original_cost = discount_value + total_cost
+
+            try:
+                processor_response = PaymentProcessorResponse.objects.filter(basket=order.basket).latest()
```
clintonb: The `PaymentProcessorResponse` was never intended to be used for any purpose outside of debugging, hence our not parsing any of the responses. If you need address info, it needs to be retrieved from the `Order` model.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+                        'country': response.get('req_bill_to_address_country')
+                    }
+                else:
+                    recipient_info = None
+            except (PaymentProcessorResponse.DoesNotExist, KeyError):
+                recipient_info = None
+
+            receipt = {
+                'billed_to': recipient_info,
+                'currency': settings.OSCAR_DEFAULT_CURRENCY,
```
clintonb: Currency should come from the order.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+            except Order.DoesNotExist:
+                context.update({
+                    'error_text': _('Order {order_number} not found.').format(order_number=order_number),
+                    'for_help_text': '',
+                    'is_payment_complete': False,
+                    'page_title': _('Order not found')
+                })
+                return self.render_to_response(context)
+
+            seat = order.lines.first().product
```
clintonb: Let's stop doing this now. We are quickly approaching a world where orders contain multiple items. We should not port the single item order paradigm to this page.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+                'is_refunded': False,
+                'items': [{
+                    'description': line['description'],
+                    'cost': add_currency(float(line['line_price_excl_tax'])),
+                    'quantity': line['quantity']
+                } for line in order_data['lines']],
+                'order_number': order.number,
+                'original_cost': add_currency(original_cost),
+                'payment_processor': order_data['payment_processor'],
+                'purchased_datetime': dateutil.parser.parse(order_data['date_placed']).strftime('%d. %B %Y'),
```
clintonb: This needs to be localized. See http://stackoverflow.com/a/6991918/592820 for an example implementation.

ecommerce/extensions/checkout/views.py
```diff
@@ -1,25 +1,28 @@
...
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import RedirectView, TemplateView
 from oscar.apps.checkout.views import *  # pylint: disable=wildcard-import, unused-wildcard-import
 from oscar.core.loading import get_class, get_model
 
-from ecommerce.core.url_utils import get_lms_url
+from ecommerce.extensions.api.serializers import OrderSerializer
 from ecommerce.extensions.checkout.exceptions import BasketNotFreeError
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
+from ecommerce.extensions.checkout.utils import add_currency, get_credit_provider_details, get_receipt_page_url
```
clintonb: `add_currency` has a couple issues:
1. The name is confusing. `format_price` is less confusing.
2. It defaults to the currency from settings. While we may only use a single currency today, that will not always be the case. Correcting this to accept both a currency and an amount will prevent a lot of headaches in the future.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+                        'last_name': response.get('req_bill_to_surname'),
+                        'city': response.get('req_bill_to_address_city'),
+                        'state': response.get('req_bill_to_address_state'),
+                        'postal_code': response.get('req_bill_to_address_postal_code'),
+                        'country': response.get('req_bill_to_address_country')
+                    }
+                else:
+                    recipient_info = None
+            except (PaymentProcessorResponse.DoesNotExist, KeyError):
+                recipient_info = None
```
clintonb: If you were to keep this code (which I recommend against, since you should not be using `PaymentProcessorResponse`), you should log the exception alongside a helpful message. Otherwise, debugging will be painful.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+                'vouchers': order_data['vouchers']
+            }
+
+            context.update({
+                'course': course,
+                'is_verification_required': seat.attr.id_verification_required,
+                'lms_url': order.site.siteconfiguration.lms_url_root,
+                'provider_data': provider_data,
+                'receipt': receipt,
+                'verified': seat.attr.certificate_type == 'verified'
```
clintonb: This is redundant given that you have `is_verification_required`.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -122,3 +145,112 @@ def test_post_returns_payment_support_email_in_context(self):
...
+                }
+            },
+            'is_verification_required': seat.attr.id_verification_required,
+            'lms_url': order.site.siteconfiguration.lms_url_root,
+            'provider_data': None,
+            'verified': seat.attr.certificate_type == 'verified'
+        }
+
+        self.assertEqual(response.status_code, 200)
+        self.assertDictContainsSubset(receipt_data, response.context_data['receipt'])
```
clintonb: Add the data to `context_data` and remove this assertion.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
             'payment_support_email': self.request.site.siteconfiguration.payment_support_email,
         })
         return context
+
+
+class ReceiptResponseView(TemplateView):
```
clintonb: This view needs to validate access to the requested order. Orders should be accessible if either of the following conditions is met:
1. The user is a staff user.
2. The user placed the order (e.g. `order.user == request.user`)

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+                ).format(payment_support_link=context['payment_support_link'])
+            })
+        return self.render_to_response(context)
+
+    def get_context_data(self, **kwargs):
+        context = super(ReceiptResponseView, self).get_context_data(**kwargs)
+
+        site_configuration = self.request.site.siteconfiguration
+        payment_support_email = site_configuration.payment_support_email
+        payment_support_link = '<a href="mailto:{email}">{email}</a>'.format(email=payment_support_email)
```
clintonb: No HTML in views. Return the email address only. The template should handle the rendering. Also, this should probably be part of the core context processor since it already contains the support URL.

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+
+        context.update({
+            'dashboard': site_configuration.build_lms_url('/dashboard'),
+            'error_summary': _('An error occurred while creating your receipt.'),
+            'error_text': None,
```
clintonb: Is this necessary? If this field is not present, will the page still render?

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+            'for_help_text': _(
```
clintonb: Why isn't this in the template?

ecommerce/extensions/checkout/views.py
```diff
@@ -110,3 +114,162 @@ def get_context_data(self, **kwargs):
...
+                "If your course does not appear on your dashboard, contact {payment_support_link}."
+            ).format(payment_support_link=payment_support_link),
+            'is_payment_complete': True,
+            'lms_url': site_configuration.lms_url_root,
+            'name': '{} {}'.format(self.request.user.first_name, self.request.user.last_name),
+            'nav_hidden': True,
+            'page_title': _('Receipt'),
+            'payment_support_email': payment_support_email,
+            'payment_support_link': payment_support_link,
+            'platform_name': settings.SITE_NAME,
```
clintonb: This is part of the core context processor.

ecommerce/static/js/routers/receipt_router.js
```diff
@@ -0,0 +1,24 @@
...
+            // Base/root path of the app
+            root: '/checkout/',
+
+            routes: {
+                'receipt/?order_number=:orderNumber': 'showReceiptPage'
+            },
+
+            showReceiptPage: function(orderNumber) {
+                var page = new ReceiptPage({orderNumber: orderNumber});
+                this.currentView = page;
```
clintonb: Merge with the line above. No need for the variable if it's not being used.

ecommerce/templates/oscar/checkout/receipt.html
```diff
@@ -0,0 +1,224 @@
...
+                    </div>
+                </div>
+
+                <div class="wrapper-report">
+                  <div class="order-headers order-details">{% trans "Order Details" %}</div>
+                  <div class="table-responsive">
+                  <table class="table">
+                    <thead class="order-columns">
+                      <tr>
+                        <th scope="col" >{% trans "Qty" %}.</th> <!-- Translators: Abbreviation for quantity.-->
```
clintonb: This comment should be above the line marked for translation.

ecommerce/templates/oscar/checkout/receipt.html
```diff
@@ -0,0 +1,224 @@
...
+                            </div>
+                            {% for voucher in receipt.vouchers %}
+                                <div class="voucher-info">{% trans "Coupon applied" %}:
+                                     <span class="voucher-code">{{ voucher.code }}
+                                     {{ receipt.discount_percentage }}% off</span>
+                                     <span class="discount">-{{ receipt.discount }}</span>
+                               </div>
+                           {% endfor %}
+                            </td>
+                            {% if receipt.discount %}
```
clintonb: Move the if-statement inside the div, and de-dupe.

ecommerce/static/js/apps/receipt_app.js
```diff
@@ -0,0 +1,14 @@
+require([
```
clintonb: The app and router seem unnecessary for a single page.

ecommerce/extensions/checkout/views.py
clintonb: I have a similar question. I see no usage of this data aside from `course.key`, which you already have.

clintonb (changes requested):
I'd like to see some changes:
1. Assume all orders have multiple items. If you're making calls to the Catalog API, get the info for all course runs at the same time. Use `/course_runs/?keys=A,B,C`.
2. Extend the existing Oscar view.
3. There is a lot of HTML/text munging in the view. This should be in the template.
4. A lot of the issues I have with the view are due to the fact that you're using the same view to handle CyberSource and render the receipt. Ideally the receipt view should _only_ render the receipt. Use another interstitial view to wait for CyberSource processing.

ecommerce/extensions/checkout/views.py
marjev: Good catch. This can be removed entirely. This was added because the Old Receipt Page design had Course name on the page.

I'll remove this.

ecommerce/extensions/checkout/views.py
marjev: yes, it is, when the payment is not complete, which is some use-case with CyberSource I'm not too familiar with and didn't have time to go into (would take me additional days just to figure that one out).

We'll have to figure out how we want to handle CyberSource use-case in our code. Clinton suggested what seems to be a solid solution to this problem. I'll try to implement it that way.

ecommerce/extensions/checkout/views.py
marjev: I'll remove this. It seems to have no purpose at all.

ecommerce/extensions/checkout/app.py
marjev: Should I remove this `url(r'thank-you/$', self.thankyou_view.as_view(), name='thank-you'),`?

I will extend the `ThankYouView` provided by Oscar, but I'm not sure if we want to expose ThankYouView via `thank-you` URL.

clintonb: Yeah, I would replace that thank-you URL with receipt.

ecommerce/extensions/checkout/tests/test_views.py
marjev: I don't know. My guess is this was copy/pasted from the LMS.

I'll remove this.

ecommerce/extensions/checkout/views.py
marjev: I decided to do what you recommended.

ecommerce/static/js/apps/receipt_app.js
marjev: App simply initializes the router, which initializes the page and passes the Order number to it. Receipt Backbone View uses the Order number to trigger Analytics events that are a requirement for [SOL-1956](https://openedx.atlassian.net/browse/SOL-1956).

If you know a way to do this Django side, without using Backbone Analytics Model, please let me know.

marjev:
@mjfrey @clintonb Most of the comments have been addressed.

Regarding the `post` method that handles the CyberSource Payments, should I simply add a redirect to the receipt page if the CyberSource Payment didn't fail? I was thinking of using [this view](https://github.com/edx/ecommerce/commit/89ff8dc18dd97523da1132f83075c1612ef29b66#diff-3a39a1e895cafeeed9c6a0c0dffd2dceR37). Let me know what you think.

ecommerce/static/js/apps/receipt_app.js
marjev: I removed all Backbone.js stuff for now. I'll rethink this. I found what I was looking for and I think that using Backbone for this is definitely an overcomplication.

ecommerce/extensions/checkout/signals.py
```diff
@@ -57,6 +58,32 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
     )
 
 
+@receiver(post_payment, dispatch_uid='tracking.post_payment_callback')
```
marjev: @mjfrey I decided to use `post_payment` signal as a trigger for the Segment.io event `Completed Purchase`.

Can you review this code? Do I need to send `post_payment` signal in `handle_processor_response` method for each PaymentProcessor or is that already being done somewhere in Otto/Oscar code?

marjev: FYI @clintonb

marjev:
@clintonb While I was trying to address the 4th comment you left on the PR, I began to wonder. Since the LMS Receipt Page View has the CyberSource handling logic deeply integrated within, will removing that logic from the ecommerce Receipt Page View require additional changes to the CyberSource handling code?
I still don't know which piece of code sends the POST request to the Receipt Page View (endpoint) to begin with or at least handles to which URL the request will be made to?

FYI @mjfrey

.gitignore
```diff
@@ -41,6 +41,7 @@ htmlcov
 nosetests.xml
 unittests.xml
 acceptance_tests/xunit.xml
+geckodriver.log
```
clintonb: `*.log`

bower.json
```diff
@@ -11,23 +11,25 @@
   ],
   "dependencies": {
     "backbone": "~1.2.1",
+    "backbone-relational": "~0.9.0",
```
clintonb, line 14: Since you're touching these, please update them: https://www.versioneye.com/user/projects/5818a6aa89f0a91daa7e6da0?child=summary.

ecommerce/core/context_processors.py
```diff
@@ -6,5 +6,6 @@ def core(request):
         'lms_base_url': get_lms_url(),
         'lms_dashboard_url': get_lms_dashboard_url(),
         'platform_name': request.site.name,
+        'support_email': request.site.siteconfiguration.payment_support_email,
```
clintonb: This isn't needed by every template. Move it to the context processor for the view that actually uses it.

ecommerce/extensions/checkout/app.py
```diff
@@ -36,12 +38,10 @@ def get_urls(self):
             url(r'payment-details/$',
                 self.payment_details_view.as_view(), name='payment-details'),
 
-            # Preview and thankyou
+            # Preview
             url(r'preview/$',
                 self.payment_details_view.as_view(preview=True),
                 name='preview'),
-            url(r'thank-you/$', self.thankyou_view.as_view(),
```
clintonb, line 43: Delete my recently-committed template override, too. :(

ecommerce/extensions/checkout/signals.py
```diff
@@ -12,6 +12,7 @@
 
 logger = logging.getLogger(__name__)
 post_checkout = get_class('checkout.signals', 'post_checkout')
+post_payment = get_class('checkout.signals', 'post_payment')
```
clintonb: Remove this since it is not used.

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,144 @@ def get_context_data(self, **kwargs):
...
+        return self.render_to_response(context)
+
+    def get_context_data(self, **kwargs):
+        return {
+            'is_payment_complete': True,
+            'name': '{} {}'.format(self.request.user.first_name, self.request.user.last_name),
+            'page_title': _('Receipt')
+        }
+
+    def get_object(self):
```
clintonb:
``` python
def get_object(self):
    kwargs = {
        'number': self.request.GET['order_number'],
    }

    user = self.request.user
    if not user.is_staff:
        kwargs['user'] = user

    return get_object_or_404(Order, **kwargs)
```

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,144 @@ def get_context_data(self, **kwargs):
...
+        try:
+            return Order.objects.get(number=self.request.GET['order_number'])
+        except Order.DoesNotExist:
+            return None
+
+    def update_context_with_order_data(self, context, order):
```
clintonb: We can be smarter about this. Rather than do all of this on the backend, push it to the templates. Use partials and/or template tags to clean this up.

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,144 @@ def get_context_data(self, **kwargs):
...
+            'error_text': _('The receipt that you specified does not exist in this location. '
+                            'Make sure that the URL is correct and try again.'),
+            'is_payment_complete': False,
+            'order_not_found': True,
+            'order_history_url': request.site.siteconfiguration.build_lms_url('account/settings'),
+            'page_title': _('Order not found')
+        })
+        return self.render_to_response(context=context, status=404)
+
+    def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
```
clintonb: Let's move this to a special page for CyberSource.

ecommerce/extensions/offer/utils.py
```diff
@@ -60,6 +60,6 @@ def format_benefit_value(benefit):
     if benefit.type == Benefit.PERCENTAGE:
         benefit_value = _('{benefit_value}%'.format(benefit_value=benefit_value))
     else:
-        converted_benefit = add_currency(Decimal(benefit.value))
-        benefit_value = _('${benefit_value}'.format(benefit_value=converted_benefit))
+        # TODO: Fix this; Currency should be a dynamic value;
```
clintonb: Are you fixing this?

ecommerce/static/js/config.js
```diff
@@ -12,8 +12,11 @@ require.config({
         'collections': 'js/collections',
         'datatables.net': 'bower_components/datatables/media/js/jquery.dataTables',
         'dataTablesBootstrap': 'bower_components/datatables/media/js/dataTables.bootstrap',
+        'date-utils': 'bower_components/date-utils/lib/date-utils.min',
         'ecommerce': 'js/shims/ecommerce',
+        'edx-ui-toolkit': '../../node_modules/edx-ui-toolkit/src/js',
```
clintonb: Does this actually work? If so, we can move many of our bower-based dependencies to npm.

ecommerce/static/js/pages/receipt_page.js
```diff
@@ -0,0 +1,33 @@
...
+    ) {
+        'use strict';
+
+        var onReady = function() {
+            var order_id = $.url('?order_number') || null;
+            if(order_id){
+                trackPurchase(order_id);
+            }
+        },
+        trackPurchase = function(order_id) {
```
clintonb: We should find a way to do this only the first time this loads. Subsequent views of this page should not fire events.

ecommerce/templates/edx/base.html
```diff
@@ -25,6 +25,7 @@
         {% block stylesheets %}
         {% endblock %}
     {% endcompress %}
+    {% block header_extras %} {% endblock header_extras %}
```
clintonb: This is not used, and should be removed.

ecommerce/extensions/checkout/utils.py
```diff
@@ -54,18 +54,18 @@ def get_receipt_page_url(site_configuration, order_number=None):
...
         amount (Decimal): Price amount
+        currency (str): Price currency
 
     Returns:
         str: Formatted price with currency.
     """
     return format_currency(
         amount,
-        settings.OSCAR_DEFAULT_CURRENCY,
-        format=u'#,##0.00',
```
mjfrey, line 69: we need this format.  Why did you remove it?  This will break when the browser is in some languages.

ecommerce/extensions/offer/utils.py
marjev: Yes. There are tickets already created for this cleanup work.

ecommerce/extensions/checkout/utils.py
marjev: If this format is in place, the $ is not visible on the page (when the currency is set to 'USD'). Removing the format seems to fix the issue.

mjfrey: you need add it.  without this -- the basket page breaks for some locales.

bower.json
marjev: Only `backbone-relational` or all of them? 
If you meant all of the dependencies, I'm worried about updating the jQuery from 1.12.2 to 3.1.1. That's a rather drastic change and perhaps should be done in a separate PR.

ecommerce/static/js/config.js
marjev: This is a leftover from Joyce's original commit. I wasn't able to use variables like `$baseline` so I guess it's not working. I'll remove this.

clintonb:
Remember to incorporate these changes: https://github.com/edx/edx-platform/pull/12470

clintonb:
I started a Verification API at https://github.com/edx/edx-platform/commits/clintonb/verification-status-api.

ecommerce/static/js/pages/receipt_page.js
marjev: Let me think about this some more.

marjev:
@clintonb @mjfrey Comments have been addressed. Vedran commented on one because we currently don't have a clear idea on how to get a basket URL for the thawed basket (in case CyberSource payment has failed). Also, I'll have to think about how to trigger the analytics event only once.

bower.json
marjev: I'll hold off from doing that because updating backbone-relational caused an unrelated test failure. I would like to keep this PR as clean and simple as possible.

ecommerce/static/js/pages/receipt_page.js
marjev: Addressed. Please take a look at 0224d38

marjev: Actually, this isn't doing what I was hoping it would do, so I'll have to rethink this further, but hopefully I'll manage to send bool value from the PaymentProcessorView with the redirect.

ecommerce/templates/oscar/checkout/cybersource_error.html
```diff
@@ -0,0 +1,37 @@
+{% extends 'edx/base.html' %}
```
mjfrey: Do we want to call this payment_error and not use the cyber source name?

marjev: If I understood correctly, this template is only for CyberSource usecase, because sometimes CyberSource Payment might "fail".

I didn't notice PayPal having this issue. There should be a subtask created on JIRA that describes this case. This template is only being called by CyberSourceInterstitialView.

ecommerce/extensions/api/serializers.py
```diff
@@ -199,44 +201,79 @@ class Meta(object):
 
 class LineSerializer(serializers.ModelSerializer):
     """Serializer for parsing line item data."""
+    formatted_price = serializers.SerializerMethodField()
```
clintonb: Why not do this client-side?

ecommerce/extensions/payment/views.py
```diff
@@ -272,6 +272,38 @@ def post(self, request):
...
+        cybersource_response = request.POST.dict()
+        order_number = cybersource_response['req_reference_number']
+        if Order.objects.filter(number=order_number).exists():
+            receipt_page_url = get_receipt_page_url(
+                order_number=cybersource_response.get('req_reference_number'),
+                site_configuration=self.request.site.siteconfiguration
+            )
+            return self.redirect_to_receipt_page_on_success(request, receipt_page_url)
+        else:
+            context = self.get_context_data()
```
clintonb: The purpose of this view is to process the response from CyberSource, which is why you're inheriting `CybersourceNotifyView`. As is, all you're doing is redirecting the user. Ultimately, we want to get rid of `CybersourceNotifyView` (or at least minimize our reliance on it) and complete all transactions via `CybersourceInterstitialView`.

Given the complexity, this should be split into a separate PR.

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,92 @@ def get_context_data(self, **kwargs):
...
+                                'course_key': seat.attr.course_key,
+                            })
+                            providers.append(provider_data)
+                else:
+                    if hasattr(seat.attr, 'id_verification_required') and seat.attr.id_verification_required:
+                        verification_data[seat.attr.course_key] = seat.attr.id_verification_required
+
+        order_data = OrderSerializer(order, context={'request': self.request}).data
+
+        payment_successful = self.request.session.pop('payment_successful', None)
```
clintonb: Let's be more specific in the naming here. `payment_successful` is too broad. `fire_tracking_events` seems more appropriate. Also, the code can be simplified:

``` python
context['fire_tracking_evnets'] = self.request.session.pop('fire_tracking_evnets', False)
```

ecommerce/templates/oscar/checkout/cybersource_error.html
```diff
@@ -0,0 +1,37 @@
...
+{% load i18n %}
+{% load staticfiles %}
```
clintonb: You don't appear to be using this. It should be removed.

ecommerce/templates/oscar/checkout/cybersource_error.html
```diff
@@ -0,0 +1,37 @@
...
+{% block content %}
+  <div class="container content-wrapper receipt-cancel-error">
+    <h1>{% trans "Payment Failed" %}</h1>
+    <p>
+      {% blocktrans %}
+        A system error occurred while processing your payment. <b>You have not been charged.</b>
+      {% endblocktrans %}</p>
+
+    <p>
+      {% with "<a class='nav-link' href='mailto:"|add:payment_support_email|add:"'>"|safe as start_link %}
```
clintonb: For future reference, we have a custom tag, `captureas`, that will simplify this.

ecommerce/extensions/payment/views.py
```diff
@@ -272,6 +272,38 @@ def post(self, request):
...
+            basket_id = OrderNumberGenerator().basket_id(order_number)
+            basket = self._get_basket(basket_id)
+            if basket:
+                basket.thaw()
+            return self.render_to_response(context=context, status=500)
```
clintonb: `502 Bad Gateway` may be a more appropriate status code.

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,92 @@ def get_context_data(self, **kwargs):
...
+                        if provider_data:
+                            provider_data.update({
+                                'course_key': seat.attr.course_key,
+                            })
+                            providers.append(provider_data)
+                else:
+                    if hasattr(seat.attr, 'id_verification_required') and seat.attr.id_verification_required:
+                        verification_data[seat.attr.course_key] = seat.attr.id_verification_required
+
+        order_data = OrderSerializer(order, context={'request': self.request}).data
```
clintonb: Why bother with serialization? Django templates can use accept an `Order` object.

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,92 @@ def get_context_data(self, **kwargs):
...
+
+        payment_successful = self.request.session.pop('payment_successful', None)
+        if payment_successful:
+            context['payment_successful'] = payment_successful
+
+        context.update({
+            'order_data': order_data,
+            'providers': providers,
+            'purchased_datetime': dateutil.parser.parse(order_data['date_placed']),
```
clintonb: This goes along with my earlier question about serialization. We have an Order object. Use it, and do any special formatting in the template.

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,92 @@ def get_context_data(self, **kwargs):
...
+        """
+        site_configuration = self.request.site.siteconfiguration
+
+        verification_data = {}
+        providers = []
+        if self.request.user == order.user:
+            for line in order.lines.all():
+                seat = line.product
+                if self.request.user.is_verified(self.request):
+                    if hasattr(seat.attr, 'certificate_type') and seat.attr.certificate_type == 'credit':
```
clintonb: The attribute you actually care about is `credit_provider`. Just check for it.

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,92 @@ def get_context_data(self, **kwargs):
...
+            order (Order): Order for which the Receipt Page is being rendered.
+        """
+        site_configuration = self.request.site.siteconfiguration
+
+        verification_data = {}
+        providers = []
+        if self.request.user == order.user:
+            for line in order.lines.all():
+                seat = line.product
+                if self.request.user.is_verified(self.request):
```
clintonb: If tests aren't failing on this call, they should be because this method does not accept any arguments.

Also, @vkaracic we should be caching this API response. It makes no sense to re-fetch verification data if the verification expires in a year.

ecommerce/static/js/pages/receipt_page.js
```diff
@@ -0,0 +1,34 @@
...
+define([
+        'jquery',
+        'jquery-url'
+    ],
+    function ($
+    ) {
+        'use strict';
+
+        var onReady = function() {
+            var order_id = $.url('?order_number') || null,
```
clintonb: `orderId`

clintonb: Why bother with jQuery URL? Pass the order number as a data attribute.

ecommerce/static/js/config.js
```diff
@@ -12,8 +12,10 @@ require.config({
         'collections': 'js/collections',
         'datatables.net': 'bower_components/datatables/media/js/jquery.dataTables',
         'dataTablesBootstrap': 'bower_components/datatables/media/js/dataTables.bootstrap',
+        'date-utils': 'bower_components/date-utils/lib/date-utils.min',
```
clintonb: This doesn't seem to be used anywhere.

bower.json
```diff
@@ -11,23 +11,25 @@
...
+    "backbone.stickit": "~0.9.2",
     "bootstrap-sass": "~3.3.4",
-    "jquery": "~1.11.3",
-    "js-cookie": "~2.1.2",
-    "requirejs": "~2.1.15",
-    "underscore": "~1.8.2",
     "bootstrapaccessibilityplugin": "~1.0.4",
     "datatables": "1.10.10",
-    "fontawesome": "~4.3.0",
+    "date-utils": "1.2.16",
```
clintonb: This doesn't seem to be used anywhere.

ecommerce/static/js/apps/receipt_app.js
clintonb: Skip the middle man. We don't need this app file. Just load the page JS directly. Also, update `build.js`.

ecommerce/static/js/pages/receipt_page.js
```diff
@@ -0,0 +1,34 @@
...
+            window.analytics.track('Completed Purchase', {
+                orderId: order_id,
+                total: el.data('total-amount'),
+                currency: el.data('currency')
+            });
+        };
+
+        return {
+            onReady: onReady,
+            trackPurchase: trackPurchase
```
clintonb: This is only called by `onReady`. There is no need to expose it. If you get rid of `receipt_app.js`, there is no need to expose anything.

ecommerce/templates/oscar/checkout/_credit_provider_data.html
```diff
@@ -0,0 +1,38 @@
...
+
+    <a id="credit-button"
+      class="complete-course btn-neutral"
+      data-course-key="{{ provider.course_key }}"
+      data-provider="{{ provider.id }}"
+      data-track-category="credit"
+      data-track-event="edx.bi.credit.clicked_complete_credit"
+      data-track-label="{{ provider.course_key }}"
+      data-track-type="click"
+      data-username="{{ name }}"
```
clintonb: Why is this not username?

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,92 @@ def get_context_data(self, **kwargs):
...
+        except Http404:
+            context.update({
+                'order_history_url': request.site.siteconfiguration.build_lms_url('account/settings'),
+                'page_title': _('Order not found')
+            })
+            return self.render_to_response(context=context, status=404)
+
+    def get_context_data(self, **kwargs):
+        return {
+            'name': '{} {}'.format(self.request.user.first_name, self.request.user.last_name),
```
clintonb: It seems the only place where this is used, you actually need the username of the person who placed the order. This should be `order.user.username`. If you passed the order to context, this line can go away entirely.

ecommerce/extensions/checkout/views.py
clintonb: Also, this shouldn't necessarily be nested in the check for verification. Say a user has purchased a credit seat but his/her verification expired. We still want to show details for credit retrieval.

ecommerce/extensions/checkout/views.py
clintonb: We should only check for verification status if a seat requiring verification is in the order.

ecommerce/templates/oscar/checkout/receipt.html
```diff
@@ -0,0 +1,148 @@
...
+
+                  {% for provider in providers %}
+                    {% include 'oscar/checkout/_credit_provider_data.html' %}
+                  {% endfor %}
+                </div>
+              </div>
+
+              <div class="nav-wizard row">
+                {% if verification_data %}
+                  {% for key, value in verification_data.items %}
```
clintonb: We only need to show one verification link. Although the URLs are course run-dependent, verification is not actually tied to the course run, just the user.

@mjfrey this needs to be updated on LMS.

ecommerce/extensions/checkout/views.py
clintonb: Also, @vkaracic it seems we are re-raising HTTP errors. I may have suggested this, in which case it was a bad idea. We should log the exceptions (which you are doing), but not raise them.

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,92 @@ def get_context_data(self, **kwargs):
...
+    @method_decorator(login_required)
+    def dispatch(self, *args, **kwargs):
+        """
+        Customers should only be able to view their receipts when logged in. To avoid blocking responses
+        from payment processors which POST back to the page, the view must be CSRF-exempt.
+        """
+        return super(ReceiptResponseView, self).dispatch(*args, **kwargs)
+
+    def get(self, request, *args, **kwargs):
+        context = self.get_context_data(**kwargs)
```
clintonb: You've done a bit too much here, and made this slightly more complicated than it needs to be. I'll post a simpler version shortly.

clintonb (changes requested):

ecommerce/extensions/checkout/views.py
vkaracic: Okay, I'll rework it tomorrow. Also the `is_verified` method accepts the request so it can construct the LMS URL.

ecommerce/templates/oscar/checkout/receipt.html
mjfrey: @clintonb can you say more?  What needs to change in LMS?

ecommerce/extensions/checkout/views.py
mjfrey: you can only be eligible for credit if you have already verified.  Credit link will never been shown if not.

clintonb: Not quite. You are considered credit-eligible if you have a row in the credit_eligibility table. Given that credit eligibility is good for a year, it is feasible that someone could verify today, get credit eligibility tomorrow, and return a year from today with credit eligibility but in an unverified state.

By the time the user has reached the receipt page for a credit seat, Otto has already confirmed credit eligibility. We should display the credit information _regardless_ of verification status since, at this point, verification doesn't matter.

mjfrey: yes I see.

ecommerce/extensions/checkout/views.py
clintonb: I must have missed the addition of the `request` argument. We should avoid making our model methods dependent on `Request` objects. We should instead pass exactly the data we need (e.g. a `Site` object). This will allow them to be used outside of the scope of requests, if necessary.

ecommerce/extensions/api/serializers.py
clintonb: By which I mean in the template, not via JavaScript.

ecommerce/templates/oscar/checkout/receipt.html
clintonb: Verification URLs currently follow the following format: `/verify_student/start-flow/<course-id>/`. This somewhat (although, not really) made sense when payment and verification were heavily integrated. Since the rollout of Otto, the two actions are completely separate. From a model standpoint, verifications have never been linked to courses from a model standpoint.

Ideally, verification should be independent of a specific course/enrollment. If we are worried about costs, we can set broad rules for when a user is allowed to verify:

The user is enrolled in a course run that requires verification AND either of the following is true:
1. The  user is currently unverified.
2. The user's verification is expiring in the next 7 days. (Again, this is broad. We could tighten this date to be relative to the earliest unexpired verification deadline of a user's enrollments.)

Aside: Personally, I'm starting to question the need for upgrade/verification deadlines; but, that is a separate conversation.

The URLs can be simplified: `/verify_student/start_flow/`.

mjfrey: I understand, however we should do this optimization after this is moved from LMS.  The current receipt page shows this link for each purchase (course_run).

ecommerce/extensions/checkout/views.py
```diff
@@ -113,3 +116,92 @@ def get_context_data(self, **kwargs):
...
+class ReceiptResponseView(ThankYouView):
+    """ Handles behavior needed to display an order receipt. """
+    template_name = 'checkout/receipt.html'
+
+    @method_decorator(csrf_exempt)
+    @method_decorator(login_required)
+    def dispatch(self, *args, **kwargs):
+        """
+        Customers should only be able to view their receipts when logged in. To avoid blocking responses
+        from payment processors which POST back to the page, the view must be CSRF-exempt.
```
clintonb: Payment processors should no longer be posting to this page.

ecommerce/templates/oscar/checkout/receipt.html
clintonb: I don't necessarily consider this an optimization as much as it is paying down debt. Punting this now is just transferring the debt from one system to another.

At a minimum, the verification message should be displayed only once, and debt tickets should be created.

mjfrey: I will create 2 tickets.  One for re-working the LMS verify flow as stated above and the other for this page/view to use the new endpoint.

ecommerce/extensions/checkout/views.py
clintonb: @vkaracic I explored this a bit more. I can see a case where it would be useful to know the difference between the API call failing and the user not actually being verified. For example, in the event of a failure, instead of misleading the user to the verification page, we could inform the user that we were unable to check verification status and provide a link to verify in case the user is not verified. This difference is subtle; but, user experience is improved because the user is not confused if we erroneously take him/her to the verification page and he/she is already verified.

Thus, it makes sense to raise a custom error (e.g. `VerificationStatusError`) that wraps up all errors except `HttpNotFoundError`. This way callers have a single error to handle, and can choose how to inform the user.

ecommerce/templates/oscar/checkout/_credit_provider_data.html
```diff
@@ -0,0 +1,38 @@
...
+      href="{{ provider.url }}">
```
clintonb: This won't work as expected in all cases. Particularly, for providers with which we have a full integration (e.g. we send signed data), this will fail. _sigh_ I have no good alternatives at the moment. Unfortunately, a resolution to this issue (due to signing key security concerns) may not come quickly.

FYI @mjfrey

clintonb: Hmm...you could clean this up by forcing all credit requests to be initiated from the student dashboard rather than the receipt page. We already support that behavior. I'm not certain why we ever put this functionality on the receipt page.

ecommerce/extensions/checkout/views.py
clintonb: I had an idea, which lead to more ideas, which lead to https://github.com/edx/ecommerce/commits/clintonb/receipt-temp. Feel free to incorporate these changes.

ecommerce/extensions/api/serializers.py
marjev: touché

ecommerce/extensions/payment/views.py
marjev: I'll remove this from the PR and open a new one. I agree with your statement.

Instead of inheriting logic from `CybersourceNotifyView`, I'll move the logic that handles the Cybersource response. We can go from there in a new PR.

marjev:
@clintonb @mjfrey All the tests passed. Green coverage as well. I think I addressed all the comments. Please take another look.

I would like to keep this as simple as this and open a new PR for the responsive design once we're sure the current changes (text, design) on the Receipt Page is what we want.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -121,3 +145,166 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        self.assertEqual(payment_method, '{} {}'.format(source.card_type, source.label))
+
+    @httpretty.activate
+    def test_get_receipt_for_existing_order(self):
+        """
+        Staff user and Order owner should be able to see the Receipt Page.
+        All other users should get the 404 status.
+        """
+        staff_user = self.create_user(is_staff=True)
+        other_user = self.create_user()
```
mjfrey: use ddt here to pass in whether or not to be a staff user.

marjev: I would rather split this into two test case scenarios to avoid branching.

marjev: Done.

ecommerce/extensions/offer/utils.py
marjev: Done.

ecommerce/core/models.py
```diff
@@ -485,6 +485,7 @@ def is_verified(self, site):
                     cache.set(cache_key, verification, cache_timeout)
             return verification
         except HttpNotFoundError:
+            log.info('No verification data found for [%s]', self.username)
```
clintonb: Use `log.debug`to avoid littering Splunk.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -368,7 +368,7 @@ def test_response_success(self, benefit_type, benefit_value):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.context['formset_lines_data']), 1)
         line_data = response.context['formset_lines_data'][0][1]
-        self.assertEqual(line_data['benefit_value'], format_benefit_value(benefit))
+        self.assertEqual(line_data['benefit_value'], format_benefit_value(benefit, settings.OSCAR_DEFAULT_CURRENCY))
```
clintonb: `basket.currency`

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -121,3 +145,178 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        """
+        self.mock_verification_status_api(
+            self.site,
+            user,
+            status=200,
+            is_verified=False
+        )
+        return self.create_order(credit=credit)
+
+    def test_login_required_post_request(self):
```
clintonb: The view should now be `GET` only, since the CyberSource code has been removed from the receipt page.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -121,3 +145,178 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        """ The view should redirect to the login page if the user is not logged in. """
+        self.client.logout()
+        response = self.client.post(self.path)
+        self.assertEqual(response.status_code, 302)
+
+    def test_login_required_get_request(self):
```
clintonb, line 211: Remove.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -121,3 +145,178 @@ def test_post_returns_payment_support_email_in_context(self):
...
+            status=200,
+            is_verified=False
+        )
+        return self.create_order(credit=credit)
+
+    def test_login_required_post_request(self):
+        """ The view should redirect to the login page if the user is not logged in. """
+        self.client.logout()
+        response = self.client.post(self.path)
+        self.assertEqual(response.status_code, 302)
```
clintonb:
```python
testserver_login_url = self.get_full_url(reverse(settings.LOGIN_URL))
expected_url = '{path}?next={next}'.format(path=testserver_login_url, next=urllib.quote(self.path))
self.assertRedirects(response, expected_url, target_status_code=302)
```

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -121,3 +145,178 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        self.client.logout()
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 302)
+
+    def test_get_receipt_for_nonexisting_order(self):
+        """ The view should return 404 status if the Order is not found. """
+        order_number = 'ABC123'
+        response = self.client.get('{path}?order_number={order_number}'.format(
+            order_number=order_number,
+            path=self.path
```
clintonb: Nit: Order the kwargs as they appear in the format string.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -121,3 +145,178 @@ def test_post_returns_payment_support_email_in_context(self):
...
+
+    def test_get_payment_method_no_source(self):
+        """ Payment method should be None when an Order has no Payment source. """
+        order = self.create_order()
+        payment_method = ReceiptResponseView().get_payment_method(order)
+        self.assertEqual(payment_method, None)
+
+    def test_get_payment_method_payment_processor(self):
+        """
+        Payment Processor name should be displayed as the Payment method
```
clintonb: We should not be displaying the payment processor name to consumers. "CyberSource" means nothing to them. Display the card/payment type.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -121,3 +145,178 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        payment_method = ReceiptResponseView().get_payment_method(order)
+        self.assertEqual(payment_method, source.source_type.name)
+
+    def test_get_payment_method_credit_card_purchase(self):
+        """
+        Credit card type and Source label should be displayed as the Payment method
+        when a Credit card was used to purchase a product.
+        """
+        order = self.create_order()
+        source = factories.SourceFactory(order=order)
```
clintonb: `source = factories.SourceFactory(order=order, card_type='Dummy Card', label='Test')`

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -121,3 +145,178 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        Arguments:
+            order (Order): Order for which the Receipt Page should be opened.
+            user (User): User that's logging in.
+
+        Returns:
+            response (Response): Response object that's returned by a ReceiptResponseView
+        """
+        self.client.logout()
+        self.client.login(username=user.username, password=self.password)
+        return self.client.get('{path}?order_number={order_number}'.format(
```
clintonb: This bit of code is duplicated numerous times. Create a helper method:

```python
def get_receipt_response(self, order):
  url = '{path}?order_number={order_number}'.format(path=self.path, order_number=order.number)
  return self.client.get(url)
```

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -121,3 +145,178 @@ def test_post_returns_payment_support_email_in_context(self):
...
+        self.mock_verification_status_api(
+            self.site,
+            self.user,
+            status=200,
+            is_verified=True
+        )
+        seat = order.lines.first().product
+        body = {'display_name': 'Hogwarts'}
+
+        httpretty.register_uri(
```
clintonb: We are no longer calling the Credit API, so this should go.

ecommerce/templates/edx/checkout/receipt.html
```diff
@@ -0,0 +1,155 @@
...
+
+{% block content %}
+  <div class="receipt">
+  <div class="container content-container">
+    <div id="receipt-container"
+         class="pay-and-verify"
+         data-currency="{{ order.currency }}"
+         data-fire-tracking-events="{{ fire_tracking_events|yesno:"true,false" }}"
+         data-order-id="{{ order.number }}"
+         data-total-amount="{{ order.total_excl_tax }}">
```
clintonb: `order.total_incl_tax`

ecommerce/templates/edx/checkout/receipt.html
```diff
@@ -0,0 +1,155 @@
...
+
+{% block navbar %}
+  {% include 'edx/partials/_student_navbar.html' %}
+{% endblock navbar %}
+
+{% block content %}
+  <div class="receipt">
+  <div class="container content-container">
+    <div id="receipt-container"
+         class="pay-and-verify"
```
clintonb: Remove this class. This seems to be leftover from the move from LMS.

ecommerce/templates/edx/checkout/receipt.html
```diff
@@ -0,0 +1,155 @@
...
+{% block javascript %}
+  <script src="{% static 'js/pages/receipt_page.js' %}"></script>
+{% endblock javascript %}
+
+{% block navbar %}
+  {% include 'edx/partials/_student_navbar.html' %}
+{% endblock navbar %}
+
+{% block content %}
+  <div class="receipt">
```
clintonb: There is a lot of nesting going on here. Can we combine these three levels of `<div>`?

ecommerce/templates/edx/checkout/receipt.html
```diff
@@ -0,0 +1,155 @@
...
+                  <span class="order-payment-heading">{% trans "Payment Method" %}</span>
+                  <div class="order-info">{{ payment_method }}</div>
+                {% endif %}
+
+                <span class="order-date-heading">{% trans "Order Date" %}</span>
+                <div class="order-info">{{ order.date_placed|date:"DATE_FORMAT" }}</div>
+              </div>
+            </div>
+
+            {#            TODO: Make this table responsive.#}
```
clintonb: This needs to be done.

ecommerce/templates/edx/checkout/receipt_not_found.html
```diff
@@ -0,0 +1,45 @@
...
+      <div class="error-message container">
+          <h3 class="title">
+            <span class="sr">{% blocktrans %} {{ error_summary }} {% endblocktrans %}</span>
+            {{ error_summary }}
+          </h3>
+          <div class="copy">
+            <p>
+              {% blocktrans %}
+                The specified order could not be located.
+                Please ensure that the URL is correct, and try again.
```
clintonb: Nit: this can go on one line.

ecommerce/templates/oscar/checkout/_credit_provider_data.html
```diff
@@ -0,0 +1,38 @@
+{% load i18n %}
```
clintonb: This is not used, and should be deleted.

ecommerce/templates/oscar/checkout/error.html
```diff
@@ -22,6 +22,6 @@
             {{ payment_support_email }}{{ end_link }}.
             {% endblocktrans %}
         {% endwith %}</p>
-    
+
```
clintonb: This empty line is not necessary.

ecommerce/static/sass/partials/views/_receipt.scss
```diff
@@ -0,0 +1,274 @@
...
+        font-size: 13px;
+    }
+
+    .old-price {
+        color: #7c7c7c;
+        font-weight: 600;
+        text-decoration: line-through;
+    }
+
+    .order-amount {
```
clintonb: Why is this necessary? Why not let the browser fit it all?

ecommerce/extensions/offer/utils.py
```diff
@@ -46,12 +46,13 @@ def get_discount_value(discount_percentage, product_price):
     return discount_percentage * product_price / 100.0
 
 
-def format_benefit_value(benefit):
+def format_benefit_value(benefit, currency):
```
clintonb: Are these changes still necessary?

ecommerce/extensions/checkout/utils.py
```diff
@@ -2,7 +2,7 @@
 
 from babel.numbers import format_currency
 from django.conf import settings
-from django.utils.translation import get_language, to_locale
+from django.utils.translation import get_language, to_locale, ugettext_lazy as _
```
clintonb: Are any changes to this file still necessary?

clintonb (changes requested):

ecommerce/extensions/checkout/tests/test_views.py
marjev: I'll leave this test method and remove the `post` one.

ecommerce/extensions/checkout/tests/test_views.py
marjev: Here's the InVision design:
https://projects.invisionapp.com/share/7D76K432C#/screens/206952822_Receipt_Whole_Cart_Coupon
Looking at it, I came to a conclusion that we want to display Payment Processor name. 
What should be displayed in case payment processor was used?

ecommerce/templates/edx/checkout/receipt.html
marjev: Let's do this in the separate PR. We'll have to make the entire page to be responsive.

Also, I want to make sure that texts and the "original" design is done properly before I focus on the responsive design.

ecommerce/static/sass/partials/views/_receipt.scss
marjev: I'll remove that. I don't have an explanation for most of these changes because I pulled them from Joyce's original PR. :/

ecommerce/templates/edx/checkout/receipt.html
clintonb: If responsiveness or accessibility support are not happening on this PR, this PR cannot be merged to master. We just met our DOJ obligations regarding accessibility, and we aren't going to risk running afoul of them again. 25% of our users are on mobile, and we will not be alienating them.

ecommerce/extensions/checkout/tests/test_views.py
clintonb: PayPal is an exception since it is both a payment processor class and a payment type. You should be displaying payment/card type from the `Source` model.

ecommerce/templates/edx/checkout/receipt.html
marjev: Ok, understood.

What about CyberSource Interstitial View PR? Should that be merged before/after this one?

clintonb:
That can be merged once it's ready. It is independent of the receipt page.

marjev:
@clintonb Comments addressed.

Here are two screenshots showing desktop and mobile design, respectively.

image

image

ecommerce/core/templatetags/core_extras.py
```diff
@@ -38,6 +41,34 @@ def do_captureas(parser, token):
...
+    Returns:
+        str: Course organization.
+    """
+    return CourseKey.from_string(course_key).org
+
+
+@register.filter(name='benefit_discount')
+def benefit_discount(benefit):
+    """
+    Format benefit value.
```
clintonb: Format it how? Please provide an example in the docstring.

ecommerce/core/templatetags/core_extras.py
```diff
@@ -38,6 +41,34 @@ def do_captureas(parser, token):
     return CaptureasNode(nodelist, args)
 
 
+@register.filter(name='organization')
```
clintonb: course_organization

ecommerce/core/templatetags/core_extras.py
```diff
@@ -38,6 +41,34 @@ def do_captureas(parser, token):
...
+def course_organization(course_key):
```
clintonb, line 43: This is not used anywhere. It, and its tests, should be deleted.

ecommerce/core/templatetags/core_extras.py
```diff
@@ -38,6 +41,34 @@ def do_captureas(parser, token):
...
+    """
+    Format benefit value.
+
+    Arguments:
+        benefit (Benefit): Voucher's Benefit.
+
+    Returns:
+        str: String value containing formatted benefit value and type.
+    """
+    return format_benefit_value(benefit)
```
clintonb: This method is implemented incorrectly for fixed price discounts. It will always return a string formatted for USD.  This will cause issues for our open source users who do not use USD. This isn't a show-stopper for this PR; but, please file a ticket to fix this, @mjfrey.

ecommerce/core/templatetags/core_extras.py
```diff
@@ -38,6 +41,34 @@ def do_captureas(parser, token):
...
+        course_key (str): Course key.
+
+    Returns:
+        str: Course organization.
+    """
+    return CourseKey.from_string(course_key).org
+
+
+@register.filter(name='benefit_discount')
+def benefit_discount(benefit):
```
clintonb: This doesn't belong in `core`. `core` is intended for components that will be used throughout the entire project. Move this tag to an `offer_extras` or `offer_tags` in the `offer` app.

clintonb (changes requested):
I may have misinterpreted your message about the responsive design as an indication that this is ready for review; however, it doesn't seem my earlier comments have been addressed. Let me know when this is ready.

ecommerce/core/templatetags/core_extras.py
marjev: https://github.com/edx/ecommerce/pull/978/files#diff-cfe5821c0f7876d15964704fe72e0218R85

ecommerce/core/templatetags/core_extras.py
marjev: @mjfrey I think the reason behind this implementation is the locale error that you told me about. It would be great if the ticket has a concrete description of the problem.

marjev:
@clintonb I apologize for pinging you before, without checking what I've pushed. Some of the changes were stashed locally and I forgot to apply those before.

I think all the comments are addressed now.

Latest screenshots:
Desktop version

Mobile version

marjev:
@clintonb anything else needed on this PR?

clintonb (approved):

PR #979, opened 2016-10-25
[SOL-2101, SOL-2102] Increase text length for dynamic coupons

vkaracic:
The catalog query field needs to be able to store more than 255 characters, therefor it's being changed to a Django TextField field. Additionally a field was added to indicate the length of the query.

https://openedx.atlassian.net/browse/SOL-2101
https://openedx.atlassian.net/browse/SOL-2102

### DIFF
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -75,7 +75,7 @@ class Range(AbstractRange):
         'course_seat_types',
     ]
     catalog = models.ForeignKey('catalogue.Catalog', blank=True, null=True, related_name='ranges')
-    catalog_query = models.CharField(max_length=255, blank=True, null=True)
+    catalog_query = models.TextField(blank=True, null=True)
     course_seat_types = models.CharField(
         max_length=255,
         validators=[validate_credit_seat_type],
--- a/ecommerce/static/js/views/coupon_form_view.js
+++ b/ecommerce/static/js/views/coupon_form_view.js
@@ -224,6 +224,7 @@ define([
             events: {
                 'input [name=course_id]': 'fillFromCourse',
                 'input [name=quantity]': 'changeTotalValue',
+                'input [name=catalog_query]': 'updateCatalogQueryLength',
 
                 // catch value after autocomplete
                 'blur [name=course_id]': 'fillFromCourse',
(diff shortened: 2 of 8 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=979&repo=edx%2Fecommerce

vkaracic:
@clintonb do you have time to review this, please?

ecommerce/extensions/offer/models.py
```diff
@@ -75,7 +75,7 @@ class Range(AbstractRange):
         'course_seat_types',
     ]
     catalog = models.ForeignKey('catalogue.Catalog', blank=True, null=True, related_name='ranges')
-    catalog_query = models.CharField(max_length=255, blank=True, null=True)
+    catalog_query = models.TextField(blank=True, null=True)
```
iivic, line 78: Should catalog queries have length limit?

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -279,6 +280,11 @@ define([
                 this._super();
             },
 
+            updateCatalogQueryLength: function() {
+                var query_length = this.$('textarea[name=catalog_query]').val().length;
+                this.$('span.query_length').text(query_length);
```
iivic: Unnecessary long selector why not just `.query_length`?

iivic (commented):
Can you provide a ss since ui changed?

clintonb:
Please provide examples of queries that are currently not allowed?

clintonb (approved):

ecommerce/extensions/offer/models.py
vkaracic: I can't find anywhere a limit on haystack queries and it's not indicated on the ticket so I guess not.
@clintonb is there a limit on the query length the course discovery can handle?

clintonb: The only limits are those of URLs/query strings between Otto and the Catalog Service. I haven't found any limits for Python/nginx. The browser limits range anywhere from 2000 to 200000 characters.

vkaracic:
The ticket indicates that queries longer than 255 are a problem. Changing the field to `TextField` should fix this.

vkaracic:
@iivic 
out-1

iivic (approved):

mjfrey (approved):

PR #980, opened 2016-10-25
Updated to Django Oscar 1.3 

clintonb:
ECOM-7095

### DIFF
--- a/ecommerce/extensions/basket/migrations/0009_auto_20170215_2229.py
+++ b/ecommerce/extensions/basket/migrations/0009_auto_20170215_2229.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import oscar.models.fields.slugfield
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('basket', '0008_auto_20170215_2224'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='line',
+            name='line_reference',
+            field=oscar.models.fields.slugfield.SlugField(max_length=128, verbose_name='Line Reference'),
+        ),
+    ]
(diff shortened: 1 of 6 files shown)


clintonb:
@edx/ecommerce please review

ecommerce/coupons/tests/mixins.py
```diff
@@ -117,7 +117,7 @@ class CouponMixin(object):
 
     def setUp(self):
         super(CouponMixin, self).setUp()
-        self.category = factories.CategoryFactory()
+        self.category = factories.CategoryFactory(path='1000')
```
mikedikan: Curious about this addition - do you think it's worthy of a comment?

ecommerce/extensions/catalogue/tests/mixins.py
```diff
@@ -22,13 +23,19 @@ class CourseCatalogTestMixin(object):
...
     useful when running tests without database migrations (which normally create these objects).
     """
+
     def setUp(self):
         super(CourseCatalogTestMixin, self).setUp()
 
         # Force the creation of a seat ProductClass
         self.seat_product_class  # pylint: disable=pointless-statement
         self.enrollment_code_product_class  # pylint: disable=pointless-statement
-        self.category, _created = Category.objects.get_or_create(name='Seats', defaults={'depth': 1})
```
rlucioni: What was wrong with the get_or_create?

ecommerce/extensions/dashboard/app.py
```diff
@@ -1,7 +1,13 @@
+from auth_backends.urls import auth_urlpatterns
 from django.conf.urls import url, include
 from oscar.apps.dashboard import app
 from oscar.core.loading import get_class
 
+from ecommerce.core.views import LogoutView
+
+# NOTE: This should match AUTH_URLS in ecommerce/urls.py.
+AUTH_URLS = [url(r'^logout/$', LogoutView.as_view(), name='logout'), ] + auth_urlpatterns
```
rlucioni: Just curious, why does this need to appear here in addition to ecommerce/urls.py? Can you import the variable and use it here?

rlucioni (approved):

rlucioni:
@clintonb have you taken a look at the migrations being added to see if any will be problematic with our data? I see some which touch the basket and order tables.

clintonb:
@rlucioni the migration changes are mostly Django-related (e.g. modify `verbose_name`). They will pose no problem for out data.

PR #982, opened 2016-10-25
Update the exception logging on filfullment modules to lower the seve…

schenedx:
…rity now that we have retry

ECOM-5976

@edx/ecommerce Please review


schenedx:
Will update splunk alerts search instead

PR #983, opened 2016-10-26
Created view to generate signed CyberSource parameters

clintonb:
This view signs the user-input billing details along with the server-generated basket details, and returns the complete parameter JSON object. The resulting data can be forwarded to CyberSource to complete a transaction.

SOL-2116

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -54,6 +54,13 @@ class SiteConfiguration(models.Model):
         null=False,
         blank=False
     )
+    client_side_payment_processor = models.CharField(
+        verbose_name=_('Payment processors'),
+        help_text=_('Processor that will be used for client-side payments'),
+        max_length=255,
+        null=True,
+        blank=True
+    )
     oauth_settings = JSONField(
         verbose_name=_('OAuth settings'),
         help_text=_('JSON string containing OAuth backend settings.'),
@@ -150,6 +157,25 @@ class SiteConfiguration(models.Model):
                 )
                 raise ValidationError(exc.message)
 
+    def _clean_client_side_payment_processor(self):
+        """
+        Validates the client_side_payment_processor field value.
+
+
+        Raises:
+            ValidationError: If the field contains the name of a payment processor NOT found in the
+            payment_processors field list.
+        """
+        value = (self.client_side_payment_processor or '').strip()
+        if value and value not in self.payment_processors_set:
+            raise ValidationError('Processor [{processor}] must be in the payment_processors field in order to '
+                                  'be configured as a client-side processor.'.format(processor=value))
+
+    def _all_payment_processors(self):
+        """ Returns all processor classes declared in settings. """
+        all_processors = [get_processor_class(path) for path in settings.PAYMENT_PROCESSORS]
+        return all_processors
+
     def get_payment_processors(self):
         """
         Returns payment processor classes enabled for the corresponding Site
@@ -157,7 +183,7 @@ class SiteConfiguration(models.Model):
         Returns:
             list[BasePaymentProcessor]: Returns payment processor classes enabled for the corresponding Site
         """
-        all_processors = [get_processor_class(path) for path in settings.PAYMENT_PROCESSORS]
+        all_processors = self._all_payment_processors()
         all_processor_names = {processor.NAME for processor in all_processors}
 
         missing_processor_configurations = self.payment_processors_set - all_processor_names
@@ -171,6 +197,21 @@ class SiteConfiguration(models.Model):
             processor for processor in all_processors
             if processor.NAME in self.payment_processors_set and processor.is_enabled()
         ]
+
+    def get_client_side_payment_processor_class(self):
+        """ Returns the payment processor class to be used for client-side payments.
+
+        If no processor is set, returns None.
+
+         Returns:
+             BasePaymentProcessor
+        """
+        if self.client_side_payment_processor:
+            for processor in self._all_payment_processors():
+                if processor.NAME == self.client_side_payment_processor:
+                    return processor
+
+        return None
 
     def get_from_email(self):
         """
@@ -186,6 +227,9 @@ class SiteConfiguration(models.Model):
         """ Validates model fields """
         if not exclude or 'payment_processors' not in exclude:
             self._clean_payment_processors()
+
+        if not exclude or 'client_side_payment_processor' not in exclude:
+            self._clean_client_side_payment_processor()
 
     @cached_property
     def segment_client(self):
--- a/ecommerce/extensions/api/v2/tests/views/test_checkout.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_checkout.py
@@ -18,7 +18,7 @@ class DummyProcessorWithUrl(DummyProcessor):
     """ Dummy payment processor class that has a test payment page url. """
     NAME = 'dummy_with_url'
 
-    def get_transaction_parameters(self, basket, request=None):
+    def get_transaction_parameters(self, basket, request=None, use_client_side_checkout=False, **kwargs):
         dummy_values = {
             'payment_page_url': 'test_processor.edx',
             'transaction_param': 'test_trans_param'
--- a/ecommerce/extensions/payment/exceptions.py
+++ b/ecommerce/extensions/payment/exceptions.py
@@ -1,10 +1,16 @@
 """Exceptions and error messages used by payment processors."""
+from __future__ import unicode_literals
+
 from django.utils.translation import ugettext_lazy as _
 from oscar.apps.payment.exceptions import GatewayError, PaymentError
 
-
-PROCESSOR_NOT_FOUND_DEVELOPER_MESSAGE = u"Lookup for a payment processor with name [{name}] failed"
+PROCESSOR_NOT_FOUND_DEVELOPER_MESSAGE = "Lookup for a payment processor with name [{name}] failed"
 PROCESSOR_NOT_FOUND_USER_MESSAGE = _("We don't support the payment option you selected.")
+
+
+class ProcessorMisconfiguredError(Exception):
+    """ Raised when a payment processor has invalid/missing settings. """
+    pass
 
 
 class ProcessorNotFoundError(Exception):
@@ -24,4 +30,12 @@ class InvalidCybersourceDecision(GatewayError):
 
 class PartialAuthorizationError(PaymentError):
     """The amount authorized by the payment processor differs from the requested amount."""
+    pass
+
+
+class PCIViolation(PaymentError):
+    """ Raised when a payment request violates PCI compliance.
+
+    If we are raising this exception BAD things are happening, and the service MUST be taken offline IMMEDIATELY!
+    """
     pass
--- a/ecommerce/extensions/payment/forms.py
+++ b/ecommerce/extensions/payment/forms.py
@@ -1,0 +1,73 @@
+from __future__ import unicode_literals
+
+import logging
+
+import pycountry
+from django import forms
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext as _
+from oscar.core.loading import get_model
+
+logger = logging.getLogger(__name__)
+Basket = get_model('basket', 'Basket')
+
+
+def country_choices():
+    """ Returns a tuple of tuples, each containing an ISO 3166 country code. """
+    return ((country.alpha2, country.name) for country in pycountry.countries)
+
+
+class PaymentForm(forms.Form):
+    """
+    Payment form with billing details.
+
+    This form captures the data necessary to complete a payment transaction. The current field constraints pertain
+    to CyberSource Silent Order POST, but should work nicely with other payment providers.
+    """
+
+    def __init__(self, user, *args, **kwargs):
+        super(PaymentForm, self).__init__(*args, **kwargs)
+        self.fields['basket'].queryset = self.fields['basket'].queryset.filter(owner=user)
+
+    basket = forms.ModelChoiceField(queryset=Basket.objects.all(), widget=forms.HiddenInput())
+    first_name = forms.CharField(max_length=60, label=_('First Name'))
+    last_name = forms.CharField(max_length=60, label=_('Last Name'))
+    address_line1 = forms.CharField(max_length=29, label=_('Address'))
+    address_line2 = forms.CharField(max_length=29, required=False, label=_('Address (continued)'))
+    city = forms.CharField(max_length=32, label=_('City'))
+    state = forms.CharField(max_length=60, required=False, label=_('State/Province'))
+    postal_code = forms.CharField(max_length=10, required=False, label=_('Zip/Postal Code'))
+    country = forms.ChoiceField(choices=country_choices, label=_('Country'))
+
+    def clean(self):
+        cleaned_data = super(PaymentForm, self).clean()
+
+        # Perform specific validation for the United States and Canada
+        country = cleaned_data.get('country')
+        if country in ('US', 'CA'):
+            state = cleaned_data.get('state')
+
+            # Ensure that a valid 2-character state/province code is specified.
+            if not state:
+                raise ValidationError({'state': _('This field is required.')})
+
+            code = '{country}-{state}'.format(country=country, state=state)
+
+            try:
+                pycountry.subdivisions.get(code=code)
+            except KeyError:
+                msg = _('{state} is not a valid state/province in {country}.').format(state=state, country=country)
+                logger.debug(msg)
+                raise ValidationError({'state': msg})
+
+            # Ensure the postal code is present, and limited to 9 characters
+            postal_code = cleaned_data.get('postal_code')
+            if not postal_code:
+                raise ValidationError({'postal_code': _('This field is required.')})
+
+            if len(postal_code) > 9:
+                raise ValidationError(
+                    {'postal_code': _(
+                        'Postal codes for the U.S. and Canada are limited to nine (9) characters.')})
+
+        return cleaned_data
--- a/ecommerce/extensions/payment/processors/__init__.py
+++ b/ecommerce/extensions/payment/processors/__init__.py
@@ -1,8 +1,12 @@
+from __future__ import unicode_literals
+
 import abc
 
 import waffle
 from django.conf import settings
 from oscar.core.loading import get_model
+
+from ecommerce.extensions.payment.constants import CLIENT_SIDE_CHECKOUT_FLAG_NAME
 
 PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
 
@@ -23,16 +27,17 @@ class BasePaymentProcessor(object):  # pragma: no cover
         self.site = site
 
     @abc.abstractmethod
-    def get_transaction_parameters(self, basket, request=None):
+    def get_transaction_parameters(self, basket, request=None, use_client_side_checkout=False, **kwargs):
         """
         Generate a dictionary of signed parameters required for this processor to complete a transaction.
 
         Arguments:
+            use_client_side_checkout:
             basket (Basket): The basket of products being purchased.
-
-        Keyword Arguments:
-            request (Request): A Request object which can be used to construct an absolute URL in
+            request (Request, optional): A Request object which can be used to construct an absolute URL in
                 cases where one is required.
+            use_client_side_checkout (bool, optional): Determines if client-side checkout should be used.
+            **kwargs: Additional parameters.
 
         Returns:
             dict: Payment processor-specific parameters required to complete a transaction. At a minimum,
@@ -69,6 +74,18 @@ class BasePaymentProcessor(object):  # pragma: no cover
         """
         partner_short_code = self.site.siteconfiguration.partner.short_code
         return settings.PAYMENT_PROCESSOR_CONFIG[partner_short_code.lower()][self.NAME.lower()]
+
+    @property
+    def client_side_payment_url(self):
+        """
+        Returns the URL to which payment data, collected directly from the payment page, should be posted.
+
+        If the payment processor does not support client-side payments, ``None`` will be returned.
+
+        Returns:
+            str
+        """
+        return None
 
     def record_processor_response(self, response, transaction_id=None, basket=None):
         """
--- a/ecommerce/extensions/payment/processors/cybersource.py
+++ b/ecommerce/extensions/payment/processors/cybersource.py
@@ -18,8 +18,10 @@ from ecommerce.core.url_utils import get_ecommerce_url
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.order.constants import PaymentEventTypeName
 from ecommerce.extensions.payment.constants import CYBERSOURCE_CARD_TYPE_MAP
-from ecommerce.extensions.payment.exceptions import (InvalidSignatureError, InvalidCybersourceDecision,
-                                                     PartialAuthorizationError)
+from ecommerce.extensions.payment.exceptions import (
+    InvalidSignatureError, InvalidCybersourceDecision, PartialAuthorizationError, PCIViolation,
+    ProcessorMisconfiguredError
+)
 from ecommerce.extensions.payment.helpers import sign
 from ecommerce.extensions.payment.processors import BasePaymentProcessor
 from ecommerce.extensions.payment.transport import RequestsTransport
@@ -41,8 +43,8 @@ class Cybersource(BasePaymentProcessor):
     For reference, see
     http://apps.cybersource.com/library/documentation/dev_guides/Secure_Acceptance_WM/Secure_Acceptance_WM.pdf.
     """
-
     NAME = 'cybersource'
+    PCI_FIELDS = ('card_cvn', 'card_expiry_date', 'card_number', 'card_type',)
 
     def __init__(self, site):
         """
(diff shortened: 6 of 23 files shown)


clintonb:
@mjfrey @iivic @marjev @vkaracic  please review
FYI @edx/ecommerce

ecommerce/extensions/api/v2/tests/views/test_checkout.py
```diff
@@ -18,7 +18,7 @@ class DummyProcessorWithUrl(DummyProcessor):
     """ Dummy payment processor class that has a test payment page url. """
     NAME = 'dummy_with_url'
 
-    def get_transaction_parameters(self, basket, request=None):
+    def get_transaction_parameters(self, basket, request=None, **kwargs):
```
marjev: Should this have a docstring?

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -97,22 +111,13 @@ def get_transaction_parameters(self, basket, request=None):
...
-        # be removed.
-        single_seat = self.get_single_seat(basket)
-        if single_seat:
-            parameters['merchant_defined_data1'] = single_seat.attr.course_key
-            parameters['merchant_defined_data2'] = getattr(single_seat.attr, 'certificate_type', '')
-
         # Level 2/3 details
         if self.send_level_2_3_details:
-            parameters['amex_data_taa1'] = '{}'.format(get_current_request().site.name)
+            parameters['amex_data_taa1'] = '{}'.format(site.name)
```
marjev: Why not simply `site.name` without format?

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -314,6 +314,27 @@ def issue_credit(self, source, amount, currency):
...
+class CybersourceSOP(Cybersource):
+    """ CyberSource Silent Order POST processor. """
+    NAME = 'cybersource_sop'
+    PCI_FIELDS = ('card_type', 'card_number', 'card_expiry_date', 'card_cvn')
+
+    def __init__(self):
+        super(CybersourceSOP, self).__init__()
+        self.sop_url = self.configuration['sop_url']
+
+    def _generate_parameters(self, basket, kwargs):
```
marjev: Do we need a docstring here?

ecommerce/extensions/payment/tests/mixins.py
```diff
@@ -76,12 +82,13 @@ def assert_payment_source_exists(self, basket, source_type, reference, label):
 class CybersourceMixin(object):
     """ Mixin with helper methods for testing CyberSource notifications. """
     DEFAULT_CARD_TYPE = 'visa'
+    FROZEN_NOW = datetime.datetime(2015, 3, 14, 9, 26, 53)
 
     def generate_signature(self, secret_key, data):
         """ Generate a signature for the given data dict. """
```
marjev, line 87: should we consider making this docstring more descriptive by adding info about args and return value?

ecommerce/extensions/payment/forms.py
```diff
@@ -0,0 +1,73 @@
...
+
+            # Ensure the postal code is present, and limited to 9 characters
+            postal_code = cleaned_data.get('postal_code')
+            if not postal_code:
+                raise ValidationError({'postal_code': _('This field is required.')})
+
+            if len(postal_code) > 9:
+                raise ValidationError(
+                    {'postal_code': _(
+                        'Postal codes for the U.S. and Canada are limited to nine (9) characters.')})
```
marjev, line 71: Should we consider making this more abstract, by adding a separate validation method for the United States and Canada transactions? This way, it would be easier to extend the validation for other countries.

Just a thought. Let me know what you think.

ecommerce/extensions/payment/tests/mixins.py
```diff
@@ -101,43 +108,43 @@ def make_billing_address(self, overrides=None):
         kwargs.update(overrides or {})
         return newfactories.BillingAddressFactory(**kwargs)
 
-    def generate_notification(self, secret_key, basket, decision=u'ACCEPT', billing_address=None, auth_amount=None,
+    def generate_notification(self, secret_key, basket, decision='ACCEPT', billing_address=None, auth_amount=None,
                               **kwargs):
         """ Generates a dict containing the API reply fields expected to be received from CyberSource. """
```
marjev, line 111: More descriptive docstring (args and return value)?

ecommerce/extensions/payment/tests/mixins.py
```diff
@@ -172,34 +179,99 @@ def runTransaction(self, **kwargs):  # pylint: disable=unused-argument
...
+                expected['item_{}_unit_price'.format(index)] = str(line.unit_price_incl_tax)
+
+        # Add the extra parameters
+        expected.update(kwargs.get('extra_parameters', {}))
+
+        signed_field_names = expected.keys() + ['transaction_uuid']
+        expected['signed_field_names'] = ','.join(sorted(signed_field_names))
+
+        # Copy the UUID value so that we can properly generate the signature.
+        expected['transaction_uuid'] = transaction_uuid
```
marjev: Why not add this somewhere between lines 195-212, when the expected is initialized? That way, `signed_field_names` can simply be expected.keys() because `transaction__uuid` will be the key, so you won't have to add it explicitly on line 238.

This way, you should be able to do something like 
`expected['signed_field_names'] = ','.join(sorted(expected.keys()))`

ecommerce/extensions/payment/tests/processors/test_cybersource.py
```diff
@@ -44,79 +39,26 @@ def setUp(self):
...
-        with mock.patch.object(Cybersource, 'utcnow', return_value=self.PI_DAY):
+        with mock.patch.object(Cybersource, 'utcnow', return_value=self.FROZEN_NOW):
             # NOTE (CCB): Instantiate a new processor object to ensure we reload any overridden settings.
-            actual = self.processor_class().get_transaction_parameters(self.basket)
+            actual = self.processor_class().get_transaction_parameters(self.basket, **kwargs)
 
-        expected = self.get_expected_transaction_parameters(actual['transaction_uuid'], include_level_2_3_details)
+        expected = self.get_expected_transaction_parameters(self.basket, actual['transaction_uuid'],
+                                                            include_level_2_3_details,
+                                                            processor=self.processor, **kwargs)
```
marjev: Nitpick: Break this long line like it's suggested [here in documentation](http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#breaking-long-lines). It will be easier to read which values are sent in this call.

ecommerce/extensions/payment/tests/test_views.py
```diff
@@ -600,3 +606,110 @@ def test_command_exception(self, mock_call_command):
         mock_call_command.side_effect = Exception("oof")
         self.get_response(True, 500, {"action": "list"})
         self.assertTrue(mock_call_command.called)
+
+
+@ddt.ddt
+class CybersourceSubmitViewTests(CybersourceMixin, TestCase):
+    path = reverse('cybersource_submit')
+    maxDiff = None
```
marjev: Do we need `maxDiff`?

marjev (commented):
Mostly questions.

ecommerce/extensions/payment/exceptions.py
```diff
@@ -25,3 +26,11 @@ class InvalidCybersourceDecision(GatewayError):
 class PartialAuthorizationError(PaymentError):
     """The amount authorized by the payment processor differs from the requested amount."""
     pass
+
+
+class PCIViolation(PaymentError):
+    """ Raised when a payment request violates PCI compliance.
+
+    If we are raising this exception BAD things are happen, and the service MUST be taken offline IMMEDIATELY!
```
vkaracic: There's a grammar error here ("are happen").

ecommerce/extensions/payment/forms.py
```diff
@@ -0,0 +1,73 @@
...
+            if not state:
+                raise ValidationError({'state': _('This field is required.')})
+
+            code = '{country}-{state}'.format(country=country, state=state)
+
+            try:
+                pycountry.subdivisions.get(code=code)
+            except KeyError:
+                msg = _('{state} is not a valid state/province in {country}.').format(state=state, country=country)
+                logger.debug(msg)
```
vkaracic, line 60: Shouldn't the logger level be `error` here, since an exception is raised afterwards?

ecommerce/extensions/payment/processors/__init__.py
```diff
@@ -1,9 +1,11 @@
+from __future__ import unicode_literals
+
 import abc
 
+import waffle
```
vkaracic, line 5: Just to confirm, direct imports should be over `from ... import ..` imports? Unrelated to this PR, but I think we could benefit from running a package like [isort](http://timothycrosley.github.io/isort/) once on the whole project directory and fix the import orders in all files?

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -68,20 +67,35 @@ def __init__(self):
...
+
+        # Sign all fields
+        parameters['signed_field_names'] = ','.join(sorted(parameters.keys()))
+        parameters['signature'] = self._generate_signature(parameters)
+
+        parameters['payment_page_url'] = self.payment_page_url
+
+        return parameters
+
+    def _generate_parameters(self, basket, kwargs):
```
vkaracic: Add docstring?

ecommerce/extensions/payment/tests/test_views.py
```diff
@@ -600,3 +606,110 @@ def test_command_exception(self, mock_call_command):
...
+        error_msg = 'Your basket may have been modified or already purchased. Refresh the page to try again.'
+        self._assert_basket_error(basket.id, error_msg)
+
+    def test_valid_request(self):
+        """ Verify the view returns the CyberSource parameters if the request is valid. """
+        basket = factories.create_basket()
+        basket.owner = self.user
+        basket.strategy = Selector().strategy()
+        basket.site = self.site
+        basket.thaw()
```
vkaracic: The basket is not frozen at this point, what you want to do is call `save()` instead of `thaw()`.

ecommerce/extensions/payment/tests/test_views.py
```diff
@@ -600,3 +606,110 @@ def test_command_exception(self, mock_call_command):
...
+            response = self.client.post(self.path, data)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response['content-type'], 'application/json')
+
+        actual = json.loads(response.content)['form_fields']
+        transaction_uuid = actual['transaction_uuid']
+        extra_parameters = {
+            'payment_method': 'card',
+            'unsigned_field_names': 'card_cvn,card_expiry_date,card_number,card_type',
```
vkaracic, line 689: Is it possible to DRY this up, in the form of `','join(CybersourceSOP.PCI_FIELDS)` ?

ecommerce/extensions/payment/views.py
```diff
@@ -34,8 +41,94 @@
...
+            error_msg = _('Your basket may have been modified or already purchased. Refresh the page to try again.')
+            return self._basket_error_response(error_msg)
+
+        basket.strategy = request.strategy
+        Applicator().apply(basket, user, self.request)
+
+        # Add extra parameters for Silent Order POST
+        extra_parameters = {
+            'payment_method': 'card',
+            'unsigned_field_names': 'card_cvn,card_expiry_date,card_number,card_type',
```
vkaracic: Again, perhaps DRY this with `','.join(CybersourceSOP.PCI_FIELDS)`?

ecommerce/extensions/payment/views.py
```diff
@@ -34,8 +41,94 @@
...
+            'bill_to_email': user.email,
+            'device_fingerprint_id': request.session.session_key,
+            # 'override_custom_receipt_page': 'http://localhost:4567/receipt',
```
vkaracic: Is this still necessary?

vkaracic (approved):
A few comments, but overall LGTM.

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -314,6 +314,27 @@ def issue_credit(self, source, amount, currency):
...
+        self.sop_url = self.configuration['sop_url']
+
+    def _generate_parameters(self, basket, kwargs):
+        parameters = super(CybersourceSOP, self)._generate_parameters(basket, kwargs)
+
+        # Mitigate PCI compliance issues
+        signed_field_names = parameters.keys()
+        if any(pci_field in signed_field_names for pci_field in self.PCI_FIELDS):
+            raise PCIViolation('One or more PCI-related fields is contained in the payment parameters. '
+                               'This service is NOT PCI-compliant! Deactivate this service immediately!')
```
mjfrey: We should add the name of the service in this exception message.

ecommerce/extensions/payment/forms.py
```diff
@@ -0,0 +1,73 @@
...
+
+    basket = forms.ModelChoiceField(queryset=Basket.objects.all(), widget=forms.HiddenInput())
+    first_name = forms.CharField(max_length=60, label=_('First Name'))
+    last_name = forms.CharField(max_length=60, label=_('Last Name'))
+    address_line1 = forms.CharField(max_length=29, label=_('Address'))
+    address_line2 = forms.CharField(max_length=29, required=False, label=_('Address (continued)'))
+    city = forms.CharField(max_length=32, label=_('City'))
+    state = forms.CharField(max_length=60, required=False, label=_('State/Province'))
+    postal_code = forms.CharField(max_length=10, required=False, label=_('Zip/Postal Code'))
+    country = forms.ChoiceField(choices=country_choices, label=_('Country'))
```
iivic, line 40: Nitpick: alphabetize

ecommerce/extensions/payment/tests/processors/test_cybersource.py
```diff
@@ -14,14 +13,11 @@
 from oscar.apps.payment.exceptions import UserCancelled, TransactionDeclined, GatewayError
 from oscar.core.loading import get_model
 from oscar.test import factories
-from threadlocals.threadlocals import get_current_request
 
-from ecommerce.core.constants import ISO_8601_FORMAT
-from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.payment.exceptions import (
-    InvalidSignatureError, InvalidCybersourceDecision, PartialAuthorizationError
+    InvalidSignatureError, InvalidCybersourceDecision, PartialAuthorizationError, PCIViolation
```
iivic: Nitpick: alphabetize

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -18,8 +18,9 @@
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.order.constants import PaymentEventTypeName
 from ecommerce.extensions.payment.constants import CYBERSOURCE_CARD_TYPE_MAP
-from ecommerce.extensions.payment.exceptions import (InvalidSignatureError, InvalidCybersourceDecision,
-                                                     PartialAuthorizationError)
+from ecommerce.extensions.payment.exceptions import (
+    InvalidSignatureError, InvalidCybersourceDecision, PartialAuthorizationError, PCIViolation
```
iivic: Nitpick: alphabetize

ecommerce/extensions/payment/tests/processors/test_cybersource.py
```diff
@@ -14,14 +13,11 @@
...
 )
-from ecommerce.extensions.payment.processors.cybersource import Cybersource, suds_response_to_dict
+from ecommerce.extensions.payment.processors.cybersource import Cybersource, suds_response_to_dict, CybersourceSOP
```
iivic: Nitpick: alphabetize

iivic (approved):
LGTM just a few Nitpicks.

ecommerce/extensions/api/v2/tests/views/test_checkout.py
clintonb: Nope. It's a dummy class solely used for testing. Additionally, the base class already has a docstring for this method.

ecommerce/extensions/payment/processors/cybersource.py
clintonb: Fixed.

ecommerce/extensions/payment/processors/cybersource.py
clintonb: Updated.

ecommerce/extensions/payment/tests/mixins.py
clintonb: I'd rather not. If you're modifying CyberSource code and don't understand the signature generation, you probably shouldn't be touching the code. That's not meant to be insulting/snarky; it's just an admission of the prerequisites for touching this code.

ecommerce/extensions/payment/tests/mixins.py
clintonb: Nope. See my comment above.

ecommerce/extensions/payment/tests/mixins.py
clintonb: Updated.

ecommerce/extensions/payment/forms.py
clintonb: This is solely meant to be a prototype for testing this functionality. Additionally, CyberSource doesn't require any additional validation for countries besides the United States or Canada.

Future iterations (using other payment processors) will not even require these fields, so it's not worth investing heavily in validation here.

ecommerce/extensions/payment/tests/processors/test_cybersource.py
clintonb: Updated.

ecommerce/extensions/payment/tests/test_views.py
clintonb: No, but it doesn't hurt to leave it. Removed.

ecommerce/extensions/payment/exceptions.py
clintonb: Updated.

ecommerce/extensions/payment/forms.py
clintonb: Nope. That logging is literally for debugging (e.g. local). The raised `ValidationError` will be added to the forms `errors` dict, which should be parsed and displayed to the user (or, in this case, returned in the response).

ecommerce/extensions/payment/processors/__init__.py
clintonb: I use PyCharm's default ordering.

ecommerce/extensions/payment/tests/test_views.py
clintonb: `thaw()` calls save. This ensures that the basket is in the correct state regardless of the state set by the factory.

ecommerce/extensions/payment/views.py
clintonb: Done.

ecommerce/extensions/payment/tests/test_views.py
clintonb: Nope. Doing that means screw-up in the constant isn't caught by tests.

ecommerce/extensions/payment/views.py
clintonb: Only for my local testing. Removed.

ecommerce/extensions/payment/processors/cybersource.py
clintonb: It's the service whose logs you're reading. The name is in the log path.

ecommerce/extensions/payment/forms.py
clintonb: The fields? No. They are in the order that I and, I assume, others think about completing the form.

ecommerce/extensions/payment/tests/processors/test_cybersource.py
clintonb: I see no major benefit to that ordering.

ecommerce/extensions/payment/processors/cybersource.py
clintonb: I see no major benefit to that ordering.

ecommerce/extensions/payment/tests/processors/test_cybersource.py
clintonb: I see no major benefit to that ordering.

ecommerce/extensions/payment/processors/cybersource.py
clintonb: Done.

ecommerce/extensions/payment/forms.py
marjev: Ok. Thanks for the explanation.

marjev (approved):

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -243,7 +243,13 @@ def _checkout(self, basket, payment_processor, request=None):
             # returned by this endpoint, simply returning the order number will suffice for now.
             response_data['order'] = {'number': order.number}
         else:
-            parameters = payment_processor.get_transaction_parameters(basket, request=self.request)
+            # FIXME Pass get_transaction_parameters flag
```
mjfrey: How did we say we where going to get this again? Didn't we also discuss the need to change the payment notification endpoint?  To get the correct profile.

clintonb: This is going to get tricky with a single processor. The original intent of the `use_sop_profile` kwarg was to instruct the method to generate parameters with the SOP profile. However, in the case of checkout, the intent is less of an instruction and more a suggestion. As in, I would like to use SOP, if it's configured, otherwise, go with the normal flow.

There is no great solution here. One option is to assume the site uses SOP if (a) client-side checkout is enabled (as determined via Waffle) and (b) the site has a client-side processor defined (in the `SiteConfiguration` model). However, our payment processors now have a dependency on Waffle.

Alternatively, we can push that logic to a helper function...

clintonb:
This is dependent on https://github.com/edx/ecommerce/pull/990 (which covers the cleanup work).

ecommerce/core/models.py
```diff
@@ -54,6 +54,13 @@ class SiteConfiguration(models.Model):
         null=False,
         blank=False
     )
+    client_side_payment_processor = models.CharField(
```
mjfrey, line 57: Rather than this, can we just add a field to the BasePaymentProcessor.
use_client_side

Then the processor implements this.  This was we can support more than 1 processor that implements client side. For example PayPal and Cybersource.

The context sent to the basket page would set this field so the buttons would know how to behave.

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -64,31 +67,82 @@ def __init__(self):
...
-            request (Request): A Request object which could be used to construct an absolute URL; not
-                used by this method.
+            extra_parameters (dict): Additional signed parameters that should be included in the signature
+                and returned dict. Note that these parameters will override any default values.
 
         Returns:
             dict: CyberSource-specific parameters required to complete a transaction, including a signature.
         """
+        sop_config_values = (self.sop_access_key, self.sop_payment_page_url, self.sop_profile_id, self.sop_secret_key,)
+        if use_client_side_checkout and not all(sop_config_values):
```
mjfrey, line 104: use_client_side_checkout, if stored in the processor itself rather than the site_config would not have to be passed around.

ecommerce/extensions/api/v2/views/checkout.py
```diff
@@ -21,6 +23,11 @@ class CheckoutView(APIView):
     """
     permission_classes = (IsAuthenticated,)
 
+    def _use_client_side_checkout(self, payment_processor):
+        request = self.request
+        return request.site.siteconfiguration.client_side_payment_processor == payment_processor.NAME \
```
mjfrey: This could just check the payment_process flag to see if client_side is enabled.

mjfrey:
@clintonb 
What if we encapsulate the client side logic to the processor implementation itself?  Rather than the siteconfig defining this.  
So the processor can support both modes if it wants and describes which method to use.

The context can then see that and inform the basket page.

ecommerce/core/models.py
clintonb: We do not want to support multiple client-side processors. Look at this as client-side _credit card_ processor. This does not exclude non-credit card processors. When we move to PayPal Express, we won't touch this.

This work would be infinitely simpler if we moved all tenants at once. I'm still not convinced of the need to roll out this functionality by tenant.

ecommerce/extensions/payment/processors/cybersource.py
clintonb: Not quite. The functionality needs to be enabled on both a site and user level (for ramped rollout). This bit should come from the caller so that the caller knows exactly what information to pass.

For example, the Basket API makes a call to this method. Today, it passes none of the extra parameters required for Silent Order POST. If you move the `use_client_side_checkout` determination into the processor, and failed to update the clients, you will end up in a scenario where Waffle thinks client-side payments should be active, but the API client has not passed the data necessary for SOP. Payments will fail.

ecommerce/extensions/payment/tests/test_views.py
vkaracic: Alright, but if any of the actions up there case the state of the basket to change to non-OPEN then there is an issue with the basket.

clintonb:
@mjfrey what you are describing will break existing consumers of the basket and checkout API endpoints unless we create a new version of the API. What was _supposed_ to be a discovery prototype has quickly grown beyond that scope. If we don't want to modify the `SiteConfiguration` model, we need to drop support for enabling this functionality per-site. I would love to do this since I see no good reason for a per-site rollout.

mjfrey:
@clintonb Thanks for the explanation. I am just iterating over design ideas hence this being used as a "prototype". I am just exploring all our options is all.

ecommerce/extensions/payment/processors/cybersource.py
mjfrey: I guess what I was thinking was a property on the processor (is_client_enabled) and if so, the API's would check this and pass extra info if needed.

mjfrey (approved):

PR #984, opened 2016-10-26
[SOL-2074] Omit expired professional courses from offer landing page.

vkaracic:
In the coupon offer landing page omit the professional courses which
have a set enrollment end datetime which has passed.

https://openedx.atlassian.net/browse/SOL-2074

According to @griffresch in the ticket:

> edX and Partners often change the enrollment end-date of a course (any type, not just ProfEd).

This might cause problems in the following scenario:
1. Course discovery service (CDS) picks up the enrollment end date metadata for a course.
2. Course enrollment end date is edited to a passed date, to prematurely close the enrollment for example, and because of that the course is no longer listed in the courses API from which the CDS retrieves the course data which means CDS has the old date stored.
3. Even though the course enrollment date has passed, we still get the old date from CDS and show that course is the offer landing page.

I don't know what settings are for the course discovery on stage and production, but that described issue happens to me locally.

Upon further investigation, setting this value (https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L2696) to `see_in_catalog` displays the course who's enrollment end date has passed. So if that setting is set so on stage and production, I could change the code to check directly on the LMS API the enrollment end date instead of CDS. @mjfrey who can ask for the values of these settings?

### DIFF
--- a/ecommerce/extensions/api/v2/views/vouchers.py
+++ b/ecommerce/extensions/api/v2/views/vouchers.py
@@ -2,8 +2,10 @@
 import logging
 from urlparse import urlparse
 
-from django.shortcuts import get_object_or_404
 import django_filters
+from dateutil import parser
+from django.shortcuts import get_object_or_404
+from django.utils.timezone import now
 from opaque_keys.edx.keys import CourseKey
 from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
@@ -91,7 +93,8 @@ class VoucherViewSet(NonDestroyableModelViewSet):
 
     def retrieve_course_objects(self, results, course_seat_types):
         """ Helper method to retrieve all the courses, products and stock records
-        from course IDs in course catalog response results.
+        from course IDs in course catalog response results. Professional courses
+        which have a set enrollment end date and which has passed are omitted.
 
         Args:
             results(dict): Course catalog response results.
@@ -100,12 +103,22 @@ class VoucherViewSet(NonDestroyableModelViewSet):
         Returns:
             Querysets of products and stock records retrieved from results.
         """
-        course_ids = [result['key'] for result in results]
-        products = Product.objects.filter(
-            course_id__in=course_ids,
-            attributes__name='certificate_type',
-            attribute_values__value_text__in=course_seat_types.split(',')
-        )
+        all_course_ids = []
+        nonexpired_course_ids = []
+        for result in results:
+            all_course_ids.append(result['key'])
+            if not result['enrollment_end'] or (
+                    result['enrollment_end'] and parser.parse(result['enrollment_end']) > now()
+            ):
+                nonexpired_course_ids.append(result['key'])
+
+        products = []
+        for seat_type in course_seat_types.split(','):
+            products.extend(Product.objects.filter(
+                course_id__in=nonexpired_course_ids if seat_type == 'professional' else all_course_ids,
+                attributes__name='certificate_type',
+                attribute_values__value_text=seat_type
+            ))
         stock_records = StockRecord.objects.filter(product__in=products)
         return products, stock_records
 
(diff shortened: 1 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=984&repo=edx%2Fecommerce

mjfrey:
@vkaracic you might want to ask Bill D.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -100,12 +103,29 @@ def retrieve_course_objects(self, results, course_seat_types):
...
         """
-        course_ids = [result['key'] for result in results]
-        products = Product.objects.filter(
-            course_id__in=course_ids,
-            attributes__name='certificate_type',
-            attribute_values__value_text__in=course_seat_types.split(',')
-        )
+        all_course_ids = []
+        nonexpired_course_ids = []
+        for result in results:
```
mjfrey, line 108: can you combine the 2 loops.  Look up the product in here?  If course_seat_types has professional then check the enrollment_end.

vkaracic: If I would lookup the product in this loop, that would mean I would have to do a database call for every iteration which I'd like to avoid.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -2,7 +2,9 @@
 import logging
 from urlparse import urlparse
 
+from dateutil import parser
 from django.shortcuts import get_object_or_404
+from django.utils.timezone import now
 import django_filters
```
marjev: nitpick: Not your doing, but perhaps you could just sort these imports properly.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -100,12 +103,29 @@ def retrieve_course_objects(self, results, course_seat_types):
...
+                    course_id__in=nonexpired_course_ids,
+                    attributes__name='certificate_type',
+                    attribute_values__value_text=seat_type
+                ))
+            else:
+                products.extend(Product.objects.filter(
+                    course_id__in=all_course_ids,
+                    attributes__name='certificate_type',
+                    attribute_values__value_text=seat_type
+                ))
```
marjev: this can be dried like:

```
products.extend(Product.objects.filter(
    course_id__in=nonexpired_course_ids if seat_type == 'professional' else all_course_ids,
    attributes__name='certificate_type',
    attribute_values__value_text=seat_type
 ))
```

Also, looks to me like the loop is unnecessary. You can check if `course_seat_types == 'professional'`. Each other scenario will extend the products with all_course_ids seats.

So, I'd suggest something like:

```
products = Product.objects.filter(
    course_id__in=nonexpired_course_ids if course_seat_types == 'professional' else all_course_ids,
    attributes__name='certificate_type',
    attribute_values__value_text=seat_type
 )
```

ecommerce/extensions/api/v2/views/vouchers.py
marjev: In a comment below, you can see my suggestion on how the 2nd loop could be removed. Let me know if I'm missing anything there.

marjev (changes requested):

ecommerce/extensions/api/v2/views/vouchers.py
vkaracic: So if I understand you correctly, for a scenario where the seat types are `verified,professional` the expired professional seats would be included in the products with your suggestion, right?

vkaracic:
Comments addressed!

mjfrey (approved):

vkaracic:
@marjev comments addressed.

marjev (approved):

PR #985, opened 2016-10-26
[SOL-1975] Code | Offer Page - Display in order of Course Launch Date

iivic:
This PR sorts the offers by the course start date but we limited the api call to 50 results from the course discovery service so additional work is required on that side. Once I've done the sorting on course discovery I'll propose a PR in the course discovery repo.
@mjfrey @vkaracic @marjev 

https://openedx.atlassian.net/browse/SOL-1975

### DIFF
--- a/ecommerce/coupons/utils.py
+++ b/ecommerce/coupons/utils.py
@@ -8,7 +8,7 @@ from oscar.core.loading import get_model
 Product = get_model('catalogue', 'Product')
 
 
-def get_range_catalog_query_results(limit, query, site, offset=None):
+def get_range_catalog_query_results(limit, query, site, offset=None, ordering=None):
     """
     Get catalog query results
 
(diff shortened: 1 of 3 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=985&repo=edx%2Fecommerce

vkaracic (approved):
Good job.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -189,7 +189,8 @@ def get_offers_from_query(self, request, voucher, catalog_query):
                     voucher=voucher
                 ))
 
-        return offers, next_page
+        sorted_offers = sorted(offers, key=lambda k: k['course_start_date'])
+        return sorted_offers, next_page
```
marjev: why not: `return sorted(offers, key=lambda k: k['course_start_date']), next_page`

ecommerce/coupons/utils.py
```diff
@@ -17,20 +17,22 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
         offset (int): Page offset
+        ordering (str): Course Discovery results ordering (only start is enabled for now)
 
     Returns:
-        dict: Query seach results received from Course Catalog API
+        dict: Query search results received from Course Catalog API
     """
     partner_code = site.siteconfiguration.partner.short_code
-    cache_key = 'course_runs_{}_{}_{}_{}'.format(query, limit, offset, partner_code)
+    cache_key = 'course_runs_{}_{}_{}_{}_{}'.format(query, limit, offset, partner_code, ordering)
```
vkaracic, line 26: Is the `ordering` necessary here?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -109,13 +109,14 @@ def retrieve_course_objects(self, results, course_seat_types):
...
 
-    def get_offers_from_query(self, request, voucher, catalog_query):
+    def get_offers_from_query(self, request, voucher, catalog_query, ordering=None):
         """ Helper method for collecting offers from catalog query.
 
         Args:
             request (WSGIRequest): Request data.
             voucher (Voucher): Oscar Voucher for which the offers are returned.
             catalog_query (str): The query for the Course Discovery.
+            ordering (str): Course Discovery results ordering (only start is enabled for now)
```
vkaracic, line 119: Ascending or descending?

iivic: If you set the ordering to start it will be ascending but if you add - before start (ordering=-start) the ordering will be descending. The API endpoint uses OrderingFilter http://www.django-rest-framework.org/api-guide/filtering/#orderingfilter

ecommerce/coupons/utils.py
iivic: It is not a must but the cache_key should be unique so it doesn't hurt. Besides that all parameters are included I just followed that pattern.

mjfrey (approved):

PR #988, opened 2016-10-27
Fixed referrals migration

clintonb:
Allow site column to be nullable, with no default. Related to #968.

ECOM-6230

### DIFF
--- a/ecommerce/referrals/migrations/0002_auto_20161011_1728.py
+++ b/ecommerce/referrals/migrations/0002_auto_20161011_1728.py
@@ -15,7 +15,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='referral',
             name='site',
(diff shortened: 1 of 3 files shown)


clintonb:

bderusha (approved):

ecommerce/referrals/migrations/0002_auto_20161011_1728.py
```diff
@@ -15,7 +15,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='referral',
             name='site',
-            field=models.ForeignKey(default=6, to='sites.Site'),
```
clintonb, line 18: It seems Django got too smart and used the latest value from @bderusha's machine...which is 6.

jibsheet: "smart"

PR #989, opened 2016-10-31
Fix receipt page for free checkouts.

mjfrey:
Generate correct receipt page URL for free purchases.


mjfrey:
@clintonb please review.

clintonb (approved):

PR #992, opened 2016-11-01
Update the paypal payment creation code to be more defensive and introduce retry

schenedx:
@edx/ecommerce Please review
ECOM-5891

We are adding a retry to the Paypal create payment method so we don't fail out of 1 fragile attempt

### DIFF
--- a/ecommerce/extensions/payment/processors/paypal.py
+++ b/ecommerce/extensions/payment/processors/paypal.py
@@ -127,23 +127,37 @@ class Paypal(BasePaymentProcessor):
         except PaypalWebProfile.DoesNotExist:
             pass
 
-        payment = paypalrestsdk.Payment(data, api=self.paypal_api)
-        payment.create()
+        available_attempts = 1
+        if waffle.switch_is_active('PAYPAL_RETRY_ATTEMPTS'):
+            available_attempts = self.retry_attempts
 
-        # Raise an exception for payments that were not successfully created. Consuming code is
-        # responsible for handling the exception.
-        if not payment.success():
-            error = self._get_error(payment)
-            # pylint: disable=unsubscriptable-object
-            entry = self.record_processor_response(error, transaction_id=error['debug_id'], basket=basket)
-
-            logger.error(
-                "Failed to create PayPal payment for basket [%d]. PayPal's response was recorded in entry [%d].",
-                basket.id,
-                entry.id
-            )
-
-            raise GatewayError(error)
+        for i in range(1, available_attempts + 1):
+            payment = paypalrestsdk.Payment(data, api=self.paypal_api)
+            payment.create()
+            # Raise an exception for payments that were not successfully created. Consuming code is
+            # responsible for handling the exception.
+            if not payment.success():
+                if i < available_attempts:
+                    logger.warning(
+                        u"Creating paypal payment for basket [%d] unsuccessful. Will retry",
+                        basket.id,
+                        exc_info=True
+                    )
+                else:
+                    error = self._get_error(payment)
+                    # pylint: disable=unsubscriptable-object
+                    entry = self.record_processor_response(
+                        error,
+                        transaction_id=error['debug_id'],
+                        basket=basket
+                    )
+                    logger.error(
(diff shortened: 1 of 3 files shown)


ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -127,23 +127,38 @@ def get_transaction_parameters(self, basket, request=None):
...
+            if not payment.success():
+                if i < available_attempts:
+                    logger.warning(
+                        u"Creating paypal payment for basket [%d] unsuccessful. Will retry",
+                        basket.id,
+                        exc_info=True
+                    )
+                else:
+                    error = self._get_error(payment)
+                    error = self._get_error(payment)
```
MatthewPiatetsky: why is this line here twice?

schenedx: Good catch, will fix

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -127,23 +127,38 @@ def get_transaction_parameters(self, basket, request=None):
...
-            logger.error(
-                "Failed to create PayPal payment for basket [%d]. PayPal's response was recorded in entry [%d].",
-                basket.id,
-                entry.id
-            )
-
-            raise GatewayError(error)
+        available_attempts = 1
+        if waffle.switch_is_active('PAYPAL_RETRY_ATTEMPTS'):
+            available_attempts = self.retry_attempts
```
MatthewPiatetsky, line 132: where is self.retry_attempts set?

schenedx: It's in the constructor of this class. Look at line 51 of this file

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -127,23 +127,38 @@ def get_transaction_parameters(self, basket, request=None):
...
+
+        for i in range(1, available_attempts + 1):
+            payment = paypalrestsdk.Payment(data, api=self.paypal_api)
+            payment.create()
+            # Raise an exception for payments that were not successfully created. Consuming code is
+            # responsible for handling the exception.
+            if not payment.success():
+                if i < available_attempts:
+                    logger.warning(
+                        u"Creating paypal payment for basket [%d] unsuccessful. Will retry",
```
MatthewPiatetsky, line 142: Is this warning necessary? Alternatively, you could consider having one log for all the unsuccessful attempts.

mikedikan (approved):

MatthewPiatetsky:
@schenedx 👍 minor comments

ecommerce/extensions/payment/processors/paypal.py
schenedx: Yes, I would say so. It is a good trace to leave because each attempt may result in a different error. We should capture the instability of Paypal, so we can better prepare ourselves.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -127,23 +127,37 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
+        available_attempts = 1
+        if waffle.switch_is_active('PAYPAL_RETRY_ATTEMPTS'):
+            available_attempts = self.retry_attempts
+
+        for i in range(1, available_attempts + 1):
+            payment = paypalrestsdk.Payment(data, api=self.paypal_api)
+            payment.create()
+            # Raise an exception for payments that were not successfully created. Consuming code is
+            # responsible for handling the exception.
+            if not payment.success():
```
rlucioni, line 139: @schenedx I don't think this is quite what we want. If I'm remembering correctly, the problem we're trying to deal with here is an exception raised when calling `payment.create()`. The code you have here won't give us any additional protection in that case. I think you want to be using an exception handler.

rlucioni (commented):
Looks like I got here just a minute too late, but I don't think this is done.

ecommerce/extensions/payment/processors/paypal.py
schenedx: @rlucioni I reviewed the stacktrace of the splunk alert we are trying to handle. I think you are right. I'll make a new PR

PR #995, opened 2016-11-02
Fixed display of payment button labels

clintonb:
Django templates were attempting to instantiate the payment processors. This fix disables that functionality for the payment processor classes since (a) we don't need to instantiate them and (b) they cannot be instantiated without a Site object.

ECOM-6277 and SOL-2114


clintonb:
FYI @edx/ecommerce

rlucioni:
👍

PR #996, opened 2016-11-03
Catch exceptions during paypal payment creation and introduce retry

schenedx:
ECOM-5891
@edx/ecommerce Please review
This is to take on @rlucioni's feedback on the same ticket. Introduced another test case as well

### DIFF
--- a/ecommerce/extensions/payment/models.py
+++ b/ecommerce/extensions/payment/models.py
@@ -2,6 +2,7 @@ from django.db import models
 from django.utils.translation import ugettext_lazy as _
 from jsonfield import JSONField
 from oscar.apps.payment.abstract_models import AbstractSource
+from solo.models import SingletonModel
 
 from ecommerce.extensions.payment.constants import CARD_TYPE_CHOICES
 
@@ -32,5 +33,18 @@ class PaypalWebProfile(models.Model):
     name = models.CharField(max_length=255, unique=True)
 
 
+class PaypalProcessorConfiguration(SingletonModel):
+    """ This is a configuration model for PayPal Payment Processor"""
+    retry_attempts = models.PositiveSmallIntegerField(
+        default=0,
+        verbose_name=_(
+            'Number of times to retry failing Paypal client actions (e.g., payment creation, payment execution)'
+        )
+    )
+
+    class Meta(object):
+        verbose_name = "Paypal Processor Configuration"
+
+
 # noinspection PyUnresolvedReferences
-from oscar.apps.payment.models import *  # noqa pylint: disable=wildcard-import,unused-wildcard-import,wrong-import-position,wrong-import-order
+from oscar.apps.payment.models import *  # noqa pylint: disable=ungrouped-imports, wildcard-import,unused-wildcard-import,wrong-import-position,wrong-import-order
--- a/ecommerce/extensions/payment/processors/paypal.py
+++ b/ecommerce/extensions/payment/processors/paypal.py
@@ -14,7 +14,7 @@ from oscar.core.loading import get_model
 
 from ecommerce.core.url_utils import get_ecommerce_url
 from ecommerce.extensions.order.constants import PaymentEventTypeName
-from ecommerce.extensions.payment.models import PaypalWebProfile
+from ecommerce.extensions.payment.models import PaypalWebProfile, PaypalProcessorConfiguration
 from ecommerce.extensions.payment.processors import BasePaymentProcessor
 from ecommerce.extensions.payment.utils import middle_truncate
 
@@ -48,7 +48,7 @@ class Paypal(BasePaymentProcessor):
         super(Paypal, self).__init__(site)
 
         # Number of times payment execution is retried after failure.
-        self.retry_attempts = self.configuration.get('retry_attempts', 1)
+        self.retry_attempts = PaypalProcessorConfiguration.get_solo().retry_attempts
 
     @cached_property
     def paypal_api(self):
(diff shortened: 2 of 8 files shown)


MatthewPiatetsky:
👍

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -132,32 +132,45 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
+                            entry.id,
+                            exc_info=True
+                        )
+                        raise GatewayError(error)
+
+            except:  # pylint: disable=bare-except
                 if i < available_attempts:
                     logger.warning(
-                        u"Creating paypal payment for basket [%d] unsuccessful. Will retry",
+                        u"Creating paypal payment for basket [%d] result in an exception. Will retry",
```
mikedikan: nitpick: change wording here to say 'resulted'

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -132,32 +132,45 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
-            # Raise an exception for payments that were not successfully created. Consuming code is
-            # responsible for handling the exception.
-            if not payment.success():
+            try:
+                payment = paypalrestsdk.Payment(data, api=self.paypal_api)
+                payment.create()
+                if payment.success():
+                    break
+                else:
+                    if i < available_attempts:
```
bderusha: Should this be `<=` ?

bderusha (approved):

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -132,32 +132,45 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
+                        logger.warning(
+                            u"Creating paypal payment for basket [%d] unsuccessful. Will retry",
```
rlucioni: > Creating PayPal payment for basket [%d] was unsuccessful. Will retry.

ecommerce/extensions/payment/processors/paypal.py
rlucioni: > Creating PayPal payment for basket [%d] resulted in an exception. Will retry.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -132,32 +132,45 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
-                        basket=basket
-                    )
-                    logger.error(
-                        u"Failed to create PayPal payment for basket [%d]. PayPal's response recorded in entry [%d].",
-                        basket.id,
-                        entry.id,
-                        exc_info=True
-                    )
-                    raise GatewayError(error)
+                    raise
```
rlucioni, line 178: To ease the process of investigating these failures, you may want to log something at the ERROR level here explaining that retries have been exhausted before re-raising.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -132,32 +132,45 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
             available_attempts = self.retry_attempts
```
rlucioni, line 132: Have you verified that `retry_attempts` is actually defined in our production PayPal config? It may not be (this code would still work if it was missing, but no retries would occur).

I'd also encourage you to take this opportunity to clean up the way `self.retry_attempts` is assigned a value. All we really need, on line 51 at class initialization, is:

``` py
is_retry_enabled = waffle.switch_is_active('PAYPAL_RETRY_ATTEMPTS')
self.retry_attempts = self.configuration.get('retry_attempts', 1) if is_retry_enabled else 1
```

schenedx: Good catch. In https://github.com/edx/configuration/blob/master/playbooks/roles/ecommerce/defaults/main.yml#L105, we never really specified the retry_attempts. Which makes this code useless. I'll make the PR against configuration repo. That said. I don't really see the point of putting the class variable "retry_attempts" behind the waffle switch. What benefit does it have?

rlucioni: @schenedx is the configuration repo really the right place for this value? Would we benefit from being able to tune this without redeploying the entire service? If so, the config repo is probably not the best place for this.

schenedx:
@rlucioni I took in your feedback and created a `PaymentProcessorsConfiguration` singleton model to capture the retry_attempts value. 
@edx/ecommerce Please review again

clintonb (commented):
How frequently will the retry value change? If we aren't changing this on a weekly basis, I see no need for a new model. Just set the value in settings.

rlucioni:
@clintonb, @schenedx and I were just having that conversation. We think this has the potential to require tuning, and we'd rather not re-deploy the app to tweak the value. I suggested a model in anticipation of this need. It's true that this can reside just as happily in settings, but it won't be as easy to change from there.

schenedx:
@clintonb I agree with @rlucioni that it would be more beneficial to us if we can put the retry_attempts value in a DB based config model. This way, we can quickly react to needs for a change. I do agree with you in the immediate future, after we selected a value, we don't need to change the value anymore. This change do not impose negative impact in that scenario.

schenedx:
@rlucioni Do you have more comments for this change?

ecommerce/extensions/payment/models.py
```diff
@@ -32,5 +33,15 @@ class PaypalWebProfile(models.Model):
     name = models.CharField(max_length=255, unique=True)
 
 
+class PaymentProcessorsConfiguration(SingletonModel):
+    """ This is a configuration model for all Payment Processors"""
+    retry_attempts = models.PositiveSmallIntegerField(
```
rlucioni, line 38: This is specifically for PayPal. I hesitate to make it seem applicable to "all payment processors." We may want the flexibility to adjust this by payment processor in the future.

ecommerce/extensions/payment/models.py
```diff
@@ -32,5 +33,15 @@ class PaypalWebProfile(models.Model):
...
+        default=0,
+        verbose_name=_('Number of Retry Attempts for Payment Processor Requests'))
```
rlucioni: > Number of times to retry failing PayPal client actions (e.g., payment creation, payment execution).

rlucioni (changes requested):

ecommerce/extensions/payment/models.py
schenedx: OK. I will update the name to be `Paypal` specific

ecommerce/extensions/payment/models.py
```diff
@@ -32,5 +33,15 @@ class PaypalWebProfile(models.Model):
     name = models.CharField(max_length=255, unique=True)
 
 
+class PaypalProcessorConfiguration(SingletonModel):
+    """ This is a configuration model for PayPal Payment Processor"""
+    retry_attempts = models.PositiveSmallIntegerField(
+        default=0,
+        verbose_name=_('Number of times to retry failing Paypal client actions'))
```
rlucioni: I'd modify this provide the examples I quoted before. The "client actions" referred to here are clear to you and me, but may not be to a future maintainer.

rlucioni (approved):

ecommerce/extensions/payment/models.py
schenedx: Ok

ecommerce/extensions/payment/models.py
```diff
@@ -32,5 +33,18 @@ class PaypalWebProfile(models.Model):
     name = models.CharField(max_length=255, unique=True)
 
 
+class PaypalProcessorConfiguration(SingletonModel):
+    """ This is a configuration model for PayPal Payment Processor"""
+    retry_attempts = models.PositiveSmallIntegerField(
+        default=0,
+        verbose_name=_(
+            'Number of times to retry failing Paypal client actions (e.g., payment creation, payment execution)'
```
clintonb, line 41: @schenedx please regenerate a new migration with the updated `verbose_name` value.

schenedx: Good catch. I created PR for the new `verbose_name` update migration

PR #999, opened 2016-11-04
i18n updates

clintonb:


clintonb:
FYI @edx/ecommerce

Makefile
```diff
@@ -94,14 +94,13 @@ accept:
 	nosetests --with-ignore-docstrings -v acceptance_tests --with-xunit --xunit-file=acceptance_tests/xunit.xml
 
 extract_translations:
-	python manage.py makemessages -l en -v1 -d django --ignore="docs/*" --ignore="src/*" --ignore="i18n/*" --ignore="assets/*" --ignore="node_modules/*" --ignore="ecommerce/static/bower_components/*" --ignore="ecommerce/static/build/*"
-	python manage.py makemessages -l en -v1 -d djangojs --ignore="docs/*" --ignore="src/*" --ignore="i18n/*" --ignore="assets/*" --ignore="node_modules/*" --ignore="ecommerce/static/bower_components/*" --ignore="ecommerce/static/build/*"
+	cd ecommerce && i18n_tool extract -v
```
rlucioni, line 97: What about cd'ing back up a level, so that we're not left inside the ecommerce project directory? These commands all assume we're operating from the root of the repo, so returning there would make sense.

rlucioni:
@clintonb the test failure looks unrelated to your changes, looks like someone introduced a flaky test.

Makefile
clintonb: `make` targets essentially open in a new shell. Once this command exits, subsequent commands will start from the directory where command is executed.

rlucioni: Good to know, thanks.

rlucioni (approved):

clintonb:
I no longer remember why I made these changes. :(

PR #1001, opened 2016-11-07
[SOL-2124] Add command for assigning category to category-less coupons.

vkaracic:
There are coupon in production that were created before categories were introduce
which now causes issues in the coupon listings. This command will assign a default
Support-Other category to all coupons without category.
https://openedx.atlassian.net/browse/SOL-2124

@mjfrey

### DIFF
--- a/ecommerce/coupons/management/commands/populate_coupon_categories.py
+++ b/ecommerce/coupons/management/commands/populate_coupon_categories.py
@@ -1,0 +1,29 @@
+import logging
+
+from django.core.management import BaseCommand
+from oscar.core.loading import get_model
+
+Category = get_model('catalogue', 'Category')
+Product = get_model('catalogue', 'Product')
+ProductCategory = get_model('catalogue', 'ProductCategory')
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """ Populate the category field for coupons that do not have a category. """
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.DEBUG)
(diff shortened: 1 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1001

vkaracic:
@mattdrayer please review.

ecommerce/coupons/management/commands/populate_coupon_categories.py
```diff
@@ -0,0 +1,29 @@
...
+class Command(BaseCommand):
+    """ Populate the category field for coupons that do not have a category. """
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.DEBUG)
+    logger.addHandler(ch)
+
+    def handle(self, *args, **options):
+        support_other_cat = Category.objects.get(name='Support-Other')
+        for coupon in Product.objects.filter(product_class__name='Coupon'):
+            try:
```
mattdrayer, line 22: Can you use get_or_create here?

vkaracic: Not quite. A `get_or_create` will have to check for the category as well, and if I send category as a parameter it will almost always create a new object because coupons can have different categories.

mjfrey (approved):

mattdrayer:
👍

vkaracic:
Thanks for reviewing this, gents!

PR #1004, opened 2016-11-09
Payments Cleanup

clintonb:


clintonb:
@edx/ecommerce please review

rlucioni:
@clintonb where'd all these new strings come from? Shouldn't the translation validation run by Travis prevent others from merging string changes without extracting them for translation?

clintonb:
@rlucioni nothing new. For whatever reason, the extraction has moved them around.

rlucioni (approved):

PR #1005, opened 2016-11-09
ECOM-6217 Added logs and try catch on order creation

Ayub-Khan:
ECOM-6217

**Description :**
Added try catch around the order.save and logged exception to check if it faces any exception on order.save and cause Missing Order issue. 

**Reviewers :** 
- [ ]  @noraiz-anwar 
- [ ]  @Qubad786

### DIFF
--- a/ecommerce/extensions/order/utils.py
+++ b/ecommerce/extensions/order/utils.py
@@ -99,8 +99,11 @@ class OrderCreator(OscarOrderCreator):
             order_data['status'] = status
         if extra_order_fields:
             order_data.update(extra_order_fields)
-        order = Order(**order_data)
-        order.save()
+        try:
+            order = Order(**order_data)
+            order.save()
+        except Exception:
+            logger.exception('Failed to save Order [%s]', order_number)
 
         try:
             referral = Referral.objects.get(basket=basket)
--- a/ecommerce/extensions/order/utils.py
+++ b/ecommerce/extensions/order/utils.py
@@ -102,7 +102,7 @@ class OrderCreator(OscarOrderCreator):
         try:
             order = Order(**order_data)
             order.save()
-        except Exception:
+        except Exception:  # pylint: disable=broad-except
             logger.exception('Failed to save Order [%s]', order_number)
 
         try:


ecommerce/extensions/order/utils.py
```diff
@@ -99,8 +99,11 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
             order_data['status'] = status
         if extra_order_fields:
             order_data.update(extra_order_fields)
-        order = Order(**order_data)
-        order.save()
+        try:
+            order = Order(**order_data)
+            order.save()
+        except Exception:
```
noraiz-anwar: I don't think quality test will pass here. It is too broad exception handling and quality check script don't like it.

noraiz-anwar (dismissed):

noraiz-anwar (commented):
Need to add tests

schenedx:
@clintonb @rlucioni Please review. This will have an impact to the splunk alerts as well

schenedx (approved):
👍

ecommerce/extensions/order/utils.py
```diff
@@ -99,8 +99,11 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
             order_data['status'] = status
         if extra_order_fields:
             order_data.update(extra_order_fields)
-        order = Order(**order_data)
-        order.save()
+        try:
+            order = Order(**order_data)
+            order.save()
+        except Exception:  # pylint: disable=broad-except
```
clintonb, line 105: You don't need `Exception` here.

clintonb (approved):

ecommerce/extensions/order/utils.py
```diff
@@ -99,8 +99,11 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
...
+            logger.exception('Failed to save Order [%s]', order_number)
```
rlucioni, line 106: Should the exception be re-raised here? Can this method safely continue if saving the order fails?

clintonb: Yes, the exception should be raised. Now that I think about it, any exceptions that previously occurred should already be logged...unless they are caught, squashed, and not logged.

rlucioni: Is the goal of this work simply to log something when orders fail to save? If that's the case, to @clintonb's point, this PR seems unnecessary. It only makes sense if we want this method to proceed in spite of the order being unsaved.

clintonb (changes requested):
Rescinding approval until we determine the goals and necessity of this logging.

ecommerce/extensions/order/utils.py
adampalay: We wanted to add more nuanced logging to help us get to the root cause; it seems to me that adding `raise` after we log the exception is the way to go. @rlucioni , the PR is necessary to get that extra information and to make it easier to find in splunk.

rlucioni: @adampalay there should already be stacktraces being logged to Splunk. I'm not convinced this change adds any value. It should already be easy to find instances of the exception in the logs.

Ayub-Khan: @adampalay  Sorry my mistake it's not going to help us in any way. @rlucioni  is right if there is exception its already logged and we can check it using file name search.

adampalay: I see. What about having the order number in the logs? Would that be useful?

Ayub-Khan: @adampalay No its same.

PR #1006, opened 2016-11-09
[SOL-2126] Add a check for user verification status

vkaracic:
@clintonb this is the ecommerce part of https://github.com/edx/edx-platform/pull/13942, mind reviewing this, please?

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -372,14 +372,13 @@ class User(AbstractUser):
                 append_slash=False
             )
             status = api.enrollment(','.join([self.username, course_key])).get()
-        except (ConnectionError, SlumberBaseException, Timeout) as ex:
+        except (ConnectionError, SlumberBaseException, Timeout):
             log.exception(
-                'Failed to retrieve enrollment details for [%s] in course [%s], because of [%s]',
+                'Failed to retrieve enrollment details for [%s] in course [%s]',
                 self.username,
-                course_key,
-                ex,
+                course_key
             )
-            raise ex
+            raise
 
         seat_type = mode_for_seat(seat)
         if status and status.get('mode') == seat_type and status.get('is_active'):
@@ -407,13 +406,12 @@ class User(AbstractUser):
             )
             response = api.accounts(self.username).get()
             return response
-        except (ConnectionError, SlumberBaseException, Timeout) as ex:
+        except (ConnectionError, SlumberBaseException, Timeout):
             log.exception(
-                'Failed to retrieve account details for [%s], because of [%s]',
-                self.username,
-                ex,
+                'Failed to retrieve account details for [%s]',
+                self.username
             )
-            raise ex
+            raise
 
     def is_eligible_for_credit(self, course_key):
         """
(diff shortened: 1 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1006

ecommerce/core/models.py
```diff
@@ -443,14 +445,44 @@ def is_eligible_for_credit(self, course_key):
             response = api.eligibility().get(**query_strings)
         except (ConnectionError, SlumberBaseException, Timeout) as ex:  # pragma: no cover
             log.exception(
-                'Failed to retrieve eligibility details for [%s] in course [%s], Because of [%s]',
+                'Failed to retrieve eligibility details for [%s] in course [%s], because of [%s]',
```
clintonb: Get rid of the "because of" part. `log.exception` already includes the exception details.

ecommerce/core/models.py
```diff
@@ -443,14 +445,44 @@ def is_eligible_for_credit(self, course_key):
...
                 self.username,
                 course_key,
                 ex,
             )
             raise ex
         return response
 
+    def is_verified(self):
+        """
+        Check if a user has verified hers identity.
```
clintonb: his/her

ecommerce/core/models.py
```diff
@@ -443,14 +445,44 @@ def is_eligible_for_credit(self, course_key):
...
+            api = EdxRestApiClient(
+                get_lms_url('api/user/v1/'),
+                oauth_access_token=self.access_token
+            )
+            response = api.accounts(self.username).verification_status().get()
+            expires = dateutil.parser.parse(response['expires'])
+            if response['status'] == 'approved' and expires > now():
+                return True
+            return False
+        except (ConnectionError, SlumberBaseException, Timeout) as ex:  # pragma: no cover
```
clintonb: What about other exceptions?

ecommerce/core/models.py
```diff
@@ -443,14 +445,44 @@ def is_eligible_for_credit(self, course_key):
...
+            log.exception(
+                'Failed to retrieve verification status details for [%s] because of [%s]',
```
clintonb: As above, get rid of "because of".

ecommerce/core/tests/test_models.py
```diff
@@ -122,6 +125,26 @@ def test_user_is_not_eligible(self):
         user, course_key = self.prepare_credit_eligibility_info(eligible=False)
         self.assertFalse(user.is_eligible_for_credit(course_key))
 
+    @httpretty.activate
+    def verification_status(self, status='approved', expires=None):
```
clintonb: Turn this into an assertion, and pass in `expected_status`.

clintonb (approved):
See my comment on the API PR. That should help decrease some workload here and with other clients.

ecommerce/core/models.py
vkaracic: Which other exceptions?

vkaracic:
@clintonb Comments addressed!

ecommerce/core/models.py
```diff
@@ -443,14 +441,39 @@ def is_eligible_for_credit(self, course_key):
...
+            establishing a connection with the LMS verification status API endpoint.
+        """
+        try:
+            api = EdxRestApiClient(
+                get_lms_url('api/user/v1/'),
+                oauth_access_token=self.access_token
+            )
+            response = api.accounts(self.username).verification_status().get()
+            return response.get('is_verified', False)
+        except (ConnectionError, SlumberBaseException, Timeout) as ex:  # pragma: no cover
```
clintonb: Why no coverage? Also, no need for the `ex` variable. Simply calling `raise` will raise the most-recent exception. Regarding my previous comment, these exceptions are fine.

clintonb (approved):

ecommerce/core/models.py
vkaracic: Because codecov demands coverage of the except part. Should I write a test that the logger has been called?

clintonb (approved):

marjev (approved):

PR #1008, opened 2016-11-11
[SOL-2111] Add Backend Validation for Coupon Creation - Benefit Type

iivic:
@clintonb  @mjfrey  @vkaracic  Oscar is not validating the benefit type so I added this validation. Please note that I extend the oscar AbstractBenefit class and override the save method to do the validation before the model instance is saved. There will be more backend validation for the coupon form fields so confirm this is the right way to do it or provide another solution.
https://openedx.atlassian.net/browse/SOL-2111

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -4,11 +4,12 @@ import logging
 
 import dateutil.parser
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.db import transaction
 from django.http import Http404
 from django.shortcuts import get_object_or_404
 from oscar.core.loading import get_model
-from rest_framework import filters, generics, status, viewsets
+from rest_framework import filters, generics, serializers, status, viewsets
 from rest_framework.permissions import IsAdminUser, IsAuthenticated
 from rest_framework.response import Response
 
@@ -20,6 +21,7 @@ from ecommerce.extensions.api.serializers import CategorySerializer, CouponSeria
 from ecommerce.extensions.basket.utils import prepare_basket
 from ecommerce.extensions.catalogue.utils import create_coupon_product, get_or_create_catalog
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
+from ecommerce.extensions.offer.models import VALID_BENEFIT_TYPES
 from ecommerce.extensions.payment.processors.invoice import InvoicePayment
 from ecommerce.extensions.voucher.models import CouponVouchers
 from ecommerce.extensions.voucher.utils import update_voucher_offer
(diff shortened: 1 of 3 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1008"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev:
LGTM

marjev (approved):

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -79,90 +80,93 @@ def create(self, request, *args, **kwargs):
...
+
+                # Create an order now since payment is handled out of band via an invoice.
+                client, __ = BusinessClient.objects.get_or_create(name=request.data.get('client'))
+                invoice_data = self.create_update_data_dict(data=request.data, fields=Invoice.UPDATEABLE_INVOICE_FIELDS)
+                response_data = self.create_order_for_invoice(
+                    basket, coupon_id=coupon_product.id, client=client, invoice_data=invoice_data
+                )
+
+                return Response(response_data, status=status.HTTP_200_OK)
+        except ValidationError as e:
```
clintonb, line 169: 1. Where is this actually being raised? Move the try-except there.
2. Log the exception with `logger.exception()`.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -79,90 +80,93 @@ def create(self, request, *args, **kwargs):
...
+            return HttpResponseBadRequest(e)
```
clintonb: This is an API recall. You want to use DRF to properly serialize your error responses. See http://www.django-rest-framework.org/api-guide/exceptions/#validationerror.

ecommerce/extensions/offer/models.py
```diff
@@ -6,10 +6,27 @@
...
+        self.clean()
+        super(Benefit, self).save(*args, **kwargs)  # pylint: disable=bad-super-call
+
+    def clean(self):
+        self.clean_type()
+        super(Benefit, self).clean()  # pylint: disable=bad-super-call
+
+    def clean_type(self):
+        if self.type not in VALID_BENEFIT_TYPES:
+            raise ValidationError("Unrecognised benefit type {type}".format(type=self.type))
```
clintonb: 1. Single quotes.
2. Mark for translation.

clintonb (changes requested):

ecommerce/extensions/api/v2/views/coupons.py
iivic: 1. It's raised in the benefit model https://github.com/edx/ecommerce/pull/1008/files#diff-49b6e1e79e2754be1784cc6dc9a1fe84R27. The documentation suggests that we should avoid catching exceptions inside atomic https://docs.djangoproject.com/en/1.9/topics/db/transactions/#controlling-transactions-explicitly that's why I do it here.

ecommerce/extensions/api/v2/views/coupons.py
iivic: I don't understand what's wrong here. The documentation didn't help...

clintonb: What you are returning will return a simple HTTP 400 with no body. [DRF's default error handler](http://www.django-rest-framework.org/api-guide/exceptions/#exception-handling-in-rest-framework-views) will serialize error responses with detail messages. We aren't necessarily using them today; but, we should stay in the habit ensuring all of our responses, success and error, are JSON and handled by DRF.

ecommerce/extensions/api/v2/views/coupons.py
clintonb: Got it. The exception should still be logged.

iivic:
@clintonb Comments addressed. Make sure I understood you correctly.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -166,7 +166,8 @@ def create(self, request, *args, **kwargs):
 
                 return Response(response_data, status=status.HTTP_200_OK)
         except ValidationError as e:
-            return HttpResponseBadRequest(e)
+            logger.exception(e.message)
+            return Response({'error': e.message}, status=status.HTTP_400_BAD_REQUEST)
```
clintonb: Raise [`ValidationError`](http://www.django-rest-framework.org/api-guide/exceptions/#validationerror), and DRF will handle the rest.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -166,7 +166,8 @@ def create(self, request, *args, **kwargs):
 
                 return Response(response_data, status=status.HTTP_200_OK)
         except ValidationError as e:
-            return HttpResponseBadRequest(e)
+            logger.exception(e.message)
```
clintonb: `logger.exception()` already includes the original exception message. The message should here should be more informative.

clintonb (approved):

PR #1010, opened 2016-11-11
[SOL-2134] Add error message template.

vkaracic:
An increasing number of apps are displaying error message by copying
the structure of the coupon error message. This commit introduces a
general template for displaying error regardless of application.

https://openedx.atlassian.net/browse/SOL-2134

@iivic @marjev @mjfrey please review.

### DIFF
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -12,6 +12,7 @@ from oscar.test.factories import (
     ConditionalOfferFactory, OrderFactory, OrderLineFactory, RangeFactory, VoucherFactory
 )
 from oscar.test.utils import RequestFactory
+from rest_framework import status
 
 from ecommerce.core.url_utils import get_lms_url
 from ecommerce.coupons.tests.mixins import CouponMixin
@@ -263,7 +264,6 @@ class CouponOfferViewTests(ApiMockMixin, CouponMixin, CourseCatalogTestMixin, Lm
 
 class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin, TestCase):
     redeem_url = reverse('coupons:redeem')
-    HTTP_MOVED = 301
 
     def setUp(self):
         super(CouponRedeemViewTests, self).setUp()
@@ -317,7 +317,7 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin
         response = self.client.get(url_without_sku)
         self.assertEqual(response.context['error'], _('SKU not provided.'))
 
-    def test_invalid_voucher(self):
+    def test_invalid_code_voucher(self):
         """ Verify an error is returned when voucher does not exist. """
         code = 'DOESNTEXIST'
         url = self.redeem_url + '?code={}&sku={}'.format(code, self.stock_record.partner_sku)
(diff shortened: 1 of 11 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1010"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/templates/edx/partials/_error.html
```diff
@@ -1,19 +1,23 @@
 {% extends 'edx/base.html' %}
 {% load i18n %}
 
-{% block title %}{% trans "Redeem" %}{% endblock %}
+{% block title %}
+    {% blocktrans %}
```
marjev: I think you can remove the blocktrans.

ecommerce/coupons/tests/test_views.py
```diff
@@ -317,7 +317,7 @@ def test_sku_not_provided(self):
         response = self.client.get(url_without_sku)
         self.assertEqual(response.context['error'], _('SKU not provided.'))
 
-    def test_invalid_voucher(self):
+    def test_invalid_code_voucher(self):
```
marjev: nitpick: test_invalid_voucher_code

ecommerce/coupons/tests/test_views.py
```diff
@@ -332,6 +332,17 @@ def test_no_product(self):
         response = self.client.get(url)
         self.assertEqual(response.context['error'], _('The product does not exist.'))
 
+    def test_invalid_voucher(self):
```
marjev: if you're already renaming the test method above, you can name this something like:
`test_expired_voucher`

ecommerce/coupons/views.py
```diff
@@ -89,32 +89,29 @@ def get_context_data(self, **kwargs):
...
-            return {
-                'error': msg,
-            }
-        return {
-            'error': _('This coupon code is invalid.'),
-        }
+            error_context.update({'error': msg})
+            return error_context
+        error_context.update({'error': _('This coupon code is invalid.')})
+        return error_context
```
marjev: Perhaps we can dry this up a bit:

```
if code is None:
    # set error message
else:
   # do everything currently under if code is not None
   # set error messages accordingly

# update error context with error message only once
# return error context after it's sure the error template will be rendered
```

ecommerce/coupons/views.py
```diff
@@ -130,40 +127,47 @@ def get(self, request):
         then applies the voucher and if the basket total is FREE places the order and
         enrolls the user in the course.
         """
-        template_name = 'coupons/_offer_error.html'
+        template_name = 'edx/partials/_error.html'
```
marjev: No. You should still use offer_error template and extend the "partial" template from there.

ecommerce/coupons/views.py
```diff
@@ -130,40 +127,47 @@ def get(self, request):
...
-            return render(request, template_name, {
-                'error': _('You need to activate your account in order to redeem this coupon.')
-            })
+            error_context.update({'error': _('You need to activate your account in order to redeem this coupon.')})
+            return render(request, template_name, error_context)
 
         if request.user.is_user_already_enrolled(request, product):
-            return render(request, template_name, {'error': _('You are already enrolled in the course.')})
+            error_context.update({'error': _('You are already enrolled in the course.')})
+            return render(request, template_name, error_context)
```
marjev, line 161: same comment as above: I think we can dry this a bit.

Set error message for each case. Update context and render once everything is ready to rock and roll.

ecommerce/templates/edx/partials/_error.html
```diff
@@ -1,19 +1,23 @@
...
-            <div class="depth depth-2 message-error">
+            <div class="depth depth-2 message-error-content">
                 <h3>{{ error }}</h3>
                 {% trans "If you need assistance, contact edX support." %}
             </div>
         </div>
     </div>
-{% endblock %}
\ No newline at end of file
+{% endblock %}
```
marjev: This looks to me like a base template rather than a partial template.
Partials are injected in other templates when certain conditions are met.
This will be a base template for each error-specific template.

You should extend the base.html in this one, and extend this template in other error templates.

marjev (changes requested):

vkaracic:
@marjev I've changed the file structure as you said. There is now a base error template and the offer one extends it and adds a custom title to it, this way there is no need to send anything more than the error message in the context which makes the code simpler. Take another look please.

iivic (approved):

marjev (approved):

vkaracic:
Forgot to squash commits, sorry! :disappointed:

PR #1011, opened 2016-11-11
Change is_verified to account for not found verifications.

vkaracic:
This commit fixes a bug where an exception is raised when no
verifications where found on LMS for a specific user, that was
not caught before.

https://openedx.atlassian.net/browse/SOL-2135

@marjev @clintonb please review.
Now I know what you meant with this  https://github.com/edx/ecommerce/pull/1006#discussion_r87180862 :smiley:

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -14,7 +14,7 @@ from django.utils.translation import ugettext_lazy as _
 from edx_rest_api_client.client import EdxRestApiClient
 from jsonfield.fields import JSONField
 from requests.exceptions import ConnectionError, Timeout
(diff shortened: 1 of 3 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1011"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev (approved):
LGTM

Perhaps you should double check to make sure that you covered all the exceptions that might be raised by the method you're calling.

iivic (approved):

PR #1012, opened 2016-11-11
[SOL-2121] Implement SDN check

vkaracic:
We want to forbid buying courses for learners who are on the SDN list.
https://openedx.atlassian.net/browse/SOL-2121

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -131,6 +131,27 @@ class SiteConfiguration(models.Model):
         blank=True,
         default=False
     )
+    enable_sdn_check = models.BooleanField(
+        verbose_name=_('Enable SDN check'),
+        help_text=_('Enable SDN check at checkout.'),
+        default=False
+    )
+    sdn_api_url = models.CharField(
+        verbose_name=_('US Treasury SDN API URL'),
+        max_length=255,
+        blank=True
+    )
+    sdn_api_key = models.CharField(
+        verbose_name=_('US Treasury SDN API key'),
+        max_length=255,
+        blank=True
+    )
+    sdn_api_list = models.CharField(
+        verbose_name=_('SDN lists'),
+        help_text=_('A comma-seperated list of Treasury OFAC lists to check against.'),
+        max_length=255,
+        blank=True
+    )
 
     class Meta(object):
         unique_together = ('site', 'partner')
@@ -349,6 +370,16 @@ class SiteConfiguration(models.Model):
         """
         return EdxRestApiClient(settings.ENTERPRISE_API_URL, jwt=self.access_token)
 
+    @cached_property
+    def user_api_client(self):
+        """
+        Returns the API client to access the user API endpoint on LMS.
+
+        Returns:
+            EdxRestApiClient: The client to access the LMS user API service.
+        """
+        return EdxRestApiClient(self.build_lms_url('/api/user/v1/'), jwt=self.access_token)
+
 
 class User(AbstractUser):
     """Custom user model for use with OIDC."""
@@ -425,8 +456,8 @@ class User(AbstractUser):
         try:
             api = EdxRestApiClient(
                 request.site.siteconfiguration.build_lms_url('/api/user/v1'),
-                jwt=request.site.siteconfiguration.access_token,
-                append_slash=False
+                append_slash=False,
+                jwt=request.site.siteconfiguration.access_token
             )
             response = api.accounts(self.username).get()
             return response
@@ -511,6 +542,26 @@ class User(AbstractUser):
             msg = 'Failed to retrieve verification status details for [{username}]'.format(username=self.username)
             log.exception(msg)
             raise VerificationStatusError(msg)
+
+    def deactivate_account(self, site_configuration):
+        """Deactive the user's account.
+
+        Args:
+            site_configuration (SiteConfiguration): The site configuration
+                from which the LMS account API endpoint is created.
+
+        Returns:
+            Response from the deactivation API endpoint.
+        """
+        try:
+            api = site_configuration.user_api_client
+            return api.accounts(self.username).deactivate().post()
+        except:  # pylint: disable=bare-except
+            log.exception(
+                'Failed to deactivate account for user [%s]',
+                self.username
+            )
+            raise
 
 
 class Client(User):
--- a/ecommerce/extensions/api/v2/tests/views/test_sdn.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_sdn.py
@@ -1,0 +1,49 @@
+import json
+import mock
+
+import ddt
+from django.core.urlresolvers import reverse
+from rest_framework import status
+
+from ecommerce.core.models import User
+from ecommerce.extensions.api.v2.tests.views import JSON_CONTENT_TYPE
+from ecommerce.extensions.payment.utils import SDNClient
+from ecommerce.tests.testcases import TestCase
+
+
+@ddt.ddt
+class SDNCheckViewSetTests(TestCase):
+    PATH = reverse('api:v2:sdn:search')
+
+    def setUp(self):
+        super(SDNCheckViewSetTests, self).setUp()
+        user = self.create_user()
+        self.client.login(username=user.username, password=self.password)
+        self.site.siteconfiguration.enable_sdn_check = True
+        self.site.siteconfiguration.save()
+
+    def make_request(self):
+        """Make a POST request to the endpoint."""
+        return self.client.post(
+            self.PATH,
+            data=json.dumps({
+                'name': 'Tester',
+                'country': 'TE'
+            }),
+            content_type=JSON_CONTENT_TYPE
+        )
+
+    def test_authentication_required(self):
+        """Verify only authenticated users can access endpoint."""
+        self.client.logout()
+        self.assertEqual(self.make_request().status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_sdn_check_match(self):
+        """Verify the endpoint returns the number of hits SDN check made."""
+        with mock.patch.object(SDNClient, 'search') as sdn_validator_mock:
+            with mock.patch.object(User, 'deactivate_account') as deactivate_account_mock:
+                sdn_validator_mock.return_value = {'total': 1}
+                deactivate_account_mock.return_value = True
+                response = self.make_request()
+                self.assertEqual(json.loads(response.content)['hits'], 1)
+                self.assertTrue(sdn_validator_mock.called)
--- a/ecommerce/extensions/api/v2/urls.py
+++ b/ecommerce/extensions/api/v2/urls.py
@@ -15,6 +15,7 @@ from ecommerce.extensions.api.v2.views import (
     providers as provider_views,
     publication as publication_views,
     refunds as refund_views,
+    sdn as sdn_views,
     siteconfiguration as siteconfiguration_views,
     stockrecords as stockrecords_views,
     vouchers as voucher_views
@@ -70,6 +71,10 @@ PROVIDER_URLS = [
     url(r'^$', provider_views.ProviderViewSet.as_view(), name='list_providers')
 ]
 
+SDN_URLS = [
+    url(r'^search/$', sdn_views.SDNCheckViewSet.as_view(), name='search')
+]
+
 urlpatterns = [
     url(r'^baskets/', include(BASKET_URLS, namespace='baskets')),
     url(r'^checkout/$', include(CHECKOUT_URLS, namespace='checkout')),
@@ -78,6 +83,7 @@ urlpatterns = [
     url(r'^providers/', include(PROVIDER_URLS, namespace='providers')),
     url(r'^publication/', include(ATOMIC_PUBLICATION_URLS, namespace='publication')),
     url(r'^refunds/', include(REFUND_URLS, namespace='refunds')),
+    url(r'^sdn/', include(SDN_URLS, namespace='sdn'))
 ]
 
 router = ExtendedSimpleRouter()
--- a/ecommerce/extensions/api/v2/views/sdn.py
+++ b/ecommerce/extensions/api/v2/views/sdn.py
@@ -1,0 +1,41 @@
+"""API endpoint for performing an SDN check on users."""
+from rest_framework.views import APIView
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from ecommerce.extensions.payment.utils import SDNClient
+
+
+class SDNCheckViewSet(APIView):
+    """Performs an SDN check for a given user."""
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request):
+        """
+        POST handler for the view. User data is posted to this handler
+        which performs an SDN check and returns whether the user passed
+        or failed.
+        """
+        name = request.data['name']
+        country = request.data['country']
+        hits = 0
+
+        site_configuration = request.site.siteconfiguration
+        if site_configuration.enable_sdn_check:
+            sdn_check = SDNClient(
+                api_url=site_configuration.sdn_api_url,
+                api_key=site_configuration.sdn_api_key,
+                sdn_list=site_configuration.sdn_api_list
+            )
+            response = sdn_check.search(name, country)
+            hits = response['total']
+            if hits > 0:
+                sdn_check.deactivate_user(
+                    request.user,
+                    request.site.siteconfiguration,
+                    name,
+                    country,
+                    response
+                )
+
+        return Response({'hits': hits})
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -279,6 +279,7 @@ class BasketSummaryView(BasketView):
                 unicode(course_key)
             ),
             'enable_client_side_checkout': False,
+            'sdn_check': site_configuration.enable_sdn_check
         })
 
         payment_processors = site_configuration.get_payment_processors()
--- a/ecommerce/extensions/payment/admin.py
+++ b/ecommerce/extensions/payment/admin.py
@@ -5,6 +5,8 @@ from oscar.apps.payment.admin import *  # noqa pylint: disable=wildcard-import,u
 from oscar.core.loading import get_model
 from solo.admin import SingletonModelAdmin
 
+from ecommerce.extensions.payment.models import SDNCheckFailure
+
 PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
 PaypalProcessorConfiguration = get_model('payment', 'PaypalProcessorConfiguration')
 
@@ -34,5 +36,18 @@ class PaymentProcessorResponseAdmin(admin.ModelAdmin):
 
     formatted_response.allow_tags = True
 
+
+@admin.register(SDNCheckFailure)
+class SDNCheckFailureAdmin(admin.ModelAdmin):
+    search_fields = ('username', 'full_name')
+    list_display = ('username', 'full_name', 'country')
+    fields = ('username', 'full_name', 'country', 'formatted_response')
+    readonly_fields = ('username', 'full_name', 'country', 'formatted_response')
+
+    def formatted_response(self, obj):
+        pretty_response = pformat(obj.sdn_check_response)
+
+        # Use format_html() to escape user-provided inputs, avoiding an XSS vulnerability.
+        return format_html('<br><br><pre>{}</pre>', pretty_response)
 
 admin.site.register(PaypalProcessorConfiguration, SingletonModelAdmin)
--- a/ecommerce/extensions/payment/models.py
+++ b/ecommerce/extensions/payment/models.py
@@ -1,7 +1,10 @@
+from __future__ import unicode_literals
+
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 from jsonfield import JSONField
 from oscar.apps.payment.abstract_models import AbstractSource
+from django_extensions.db.models import TimeStampedModel
 from solo.models import SingletonModel
 
 from ecommerce.extensions.payment.constants import CARD_TYPE_CHOICES
@@ -45,6 +48,21 @@ class PaypalProcessorConfiguration(SingletonModel):
     class Meta(object):
         verbose_name = "Paypal Processor Configuration"
 
+
+class SDNCheckFailure(TimeStampedModel):
+    """ Record of SDN check failure. """
+    full_name = models.CharField(max_length=255)
+    username = models.CharField(max_length=255)
+    country = models.CharField(max_length=2)
+    sdn_check_response = JSONField()
+
+    def __unicode__(self):
+        return 'SDN check failure [{username}]'.format(
+            username=self.username
+        )
+
+    class Meta(object):
+        verbose_name = 'SDN Check Failure'
 
 # noinspection PyUnresolvedReferences
 from oscar.apps.payment.models import *  # noqa pylint: disable=ungrouped-imports, wildcard-import,unused-wildcard-import,wrong-import-position,wrong-import-order
--- a/ecommerce/extensions/payment/tests/test_models.py
+++ b/ecommerce/extensions/payment/tests/test_models.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from ecommerce.tests.testcases import TestCase
+from ecommerce.extensions.payment.models import SDNCheckFailure
+
+
+class SDNCheckFailureTests(TestCase):
+    def setUp(self):
+        self.full_name = 'Keyser Söze'
+        self.username = 'UnusualSuspect'
+        self.country = 'US'
+        self.sdn_check_response = {'description': 'Looks a bit suspicious.'}
+
+    def test_unicode(self):
+        """ Verify the __unicode__ method returns the correct value. """
+        basket = SDNCheckFailure.objects.create(
+            full_name=self.full_name,
+            username=self.username,
+            country=self.country,
+            sdn_check_response=self.sdn_check_response
+        )
+        expected = 'SDN check failure [{username}]'.format(
+            username=self.username
+        )
+
+        self.assertEqual(unicode(basket), expected)
--- a/ecommerce/extensions/payment/tests/test_utils.py
+++ b/ecommerce/extensions/payment/tests/test_utils.py
@@ -1,4 +1,16 @@
-from ecommerce.extensions.payment.utils import clean_field_value, middle_truncate
+import json
+import time
+from urllib import urlencode
+
+import mock
+import httpretty
+from django.conf import settings
+from django.test import override_settings
+from requests.exceptions import HTTPError, Timeout
+
+from ecommerce.core.models import User
+from ecommerce.extensions.payment.models import SDNCheckFailure
+from ecommerce.extensions.payment.utils import clean_field_value, middle_truncate, SDNClient
 from ecommerce.tests.testcases import TestCase
 
 
@@ -22,3 +34,100 @@ class UtilsTests(TestCase):
         """ Verify the passed value is cleaned of specific special characters. """
         value = 'Some^text:\'test-value'
         self.assertEqual(clean_field_value(value), 'Sometexttest-value')
+
+
+class SDNCheckTests(TestCase):
+    """ Tests for the SDN check function. """
+
+    def setUp(self):
+        super(SDNCheckTests, self).setUp()
+        self.name = 'Dr. Evil'
+        self.country = 'Evilland'
+        self.user = self.create_user(full_name=self.name)
+        self.site_configuration = self.site.siteconfiguration
+        self.site_configuration.enable_sdn_check = True,
+        self.site_configuration.sdn_api_url = 'http://sdn-test.fake/'
+        self.site_configuration.sdn_api_key = 'fake-key'
+        self.site_configuration.sdn_api_list = 'SDN,TEST'
+        self.site_configuration.save()
+
+        self.sdn_validator = SDNClient(
+            self.site_configuration.sdn_api_url,
+            self.site_configuration.sdn_api_key,
+            self.site_configuration.sdn_api_list
+        )
+
+    def mock_sdn_response(self, response, status_code=200):
+        """ Mock the SDN check API endpoint response. """
+        params = urlencode({
+            'sources': self.site_configuration.sdn_api_list,
+            'api_key': self.site_configuration.sdn_api_key,
+            'type': 'individual',
+            'name': self.name,
+            'countries': self.country
+        })
+        sdn_check_url = '{api_url}?{params}'.format(
+            api_url=self.site_configuration.sdn_api_url,
+            params=params
+        )
+
+        httpretty.register_uri(
+            httpretty.GET,
+            sdn_check_url,
+            status=status_code,
+            body=response,
+            content_type='application/json'
+        )
+
+    def assert_sdn_check_failure_recorded(self, response):
+        """ Assert an SDN check failure is logged and has the correct values. """
+        self.assertEqual(SDNCheckFailure.objects.count(), 1)
+        sdn_object = SDNCheckFailure.objects.first()
+        self.assertEqual(sdn_object.full_name, self.name)
+        self.assertEqual(sdn_object.country, self.country)
+        self.assertEqual(sdn_object.sdn_check_response, response)
+
+    @httpretty.activate
+    @override_settings(SDN_CHECK_REQUEST_TIMEOUT=0.1)
+    def test_sdn_check_timeout(self):
+        """Verify SDN check logs an exception if the request times out."""
+        def mock_timeout(_request, _uri, headers):
+            time.sleep(settings.SDN_CHECK_REQUEST_TIMEOUT + 0.1)
+            return (200, headers, {'total': 1})
+
+        self.mock_sdn_response(mock_timeout, status_code=200)
+        with self.assertRaises(Timeout):
+            with mock.patch('ecommerce.extensions.payment.utils.logger.exception') as mock_logger:
+                self.sdn_validator.search(self.name, self.country)
+                self.assertTrue(mock_logger.called)
+
+    @httpretty.activate
+    def test_sdn_check_connection_error(self):
+        """ Verify the check logs an exception in case of a connection error. """
+        self.mock_sdn_response(json.dumps({'total': 1}), status_code=400)
+        with self.assertRaises(HTTPError):
+            with mock.patch('ecommerce.extensions.payment.utils.logger.exception') as mock_logger:
+                self.sdn_validator.search(self.name, self.country)
+                self.assertTrue(mock_logger.called)
+
+    @httpretty.activate
+    def test_sdn_check_match(self):
+        """ Verify the SDN check returns the number of matches and records the match. """
+        sdn_response = {'total': 1}
+        self.mock_sdn_response(json.dumps(sdn_response))
+        response = self.sdn_validator.search(self.name, self.country)
+        self.assertEqual(response, sdn_response)
+
+    def test_deactivate_user(self):
+        """ Verify an SDN failure is logged. """
+        response = {'description': 'Bad dude.'}
+        self.assertEqual(SDNCheckFailure.objects.count(), 0)
+        with mock.patch.object(User, 'deactivate_account') as deactivate_account:
+            deactivate_account.return_value = True
+            self.sdn_validator.deactivate_user(
+                self.user,
+                self.site_configuration,
+                self.name,
+                self.country,
+                response)
+            self.assert_sdn_check_failure_recorded(response)
--- a/ecommerce/extensions/payment/urls.py
+++ b/ecommerce/extensions/payment/urls.py
@@ -1,7 +1,7 @@
 """ Payment-related URLs """
 from django.conf.urls import url
 
-from ecommerce.extensions.payment.views import cybersource, PaymentFailedView
+from ecommerce.extensions.payment.views import cybersource, PaymentFailedView, SDNFailure
 from ecommerce.extensions.payment.views.paypal import PaypalPaymentExecutionView, PaypalProfileAdminView
 
 urlpatterns = [
@@ -9,6 +9,7 @@ urlpatterns = [
     url(r'^cybersource/redirect/$', cybersource.CybersourceInterstitialView.as_view(), name='cybersource_redirect'),
     url(r'^cybersource/submit/$', cybersource.CybersourceSubmitView.as_view(), name='cybersource_submit'),
     url(r'^error/$', PaymentFailedView.as_view(), name='payment_error'),
+    url(r'^sdn/failure/$', SDNFailure.as_view(), name='sdn_failure'),
     url(r'^paypal/execute/$', PaypalPaymentExecutionView.as_view(), name='paypal_execute'),
     url(r'^paypal/profiles/$', PaypalProfileAdminView.as_view(), name='paypal_profiles'),
 ]
--- a/ecommerce/extensions/payment/utils.py
+++ b/ecommerce/extensions/payment/utils.py
@@ -1,6 +1,17 @@
+import json
+import logging
 import re
+from urllib import urlencode
 
+import requests
+from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
+from oscar.core.loading import get_model
+
+from ecommerce.extensions.payment.models import SDNCheckFailure
+
+logger = logging.getLogger(__name__)
+Basket = get_model('basket', 'Basket')
 
 
 def middle_truncate(string, chars):
@@ -53,3 +64,73 @@ def clean_field_value(value):
         A cleaned string.
     """
     return re.sub(r'[\^:"\']', '', value)
+
+
+class SDNClient(object):
+    """A utility class that handles SDN related operations."""
+    def __init__(self, api_url, api_key, sdn_list):
+        self.api_url = api_url
+        self.api_key = api_key
+        self.sdn_list = sdn_list
+
+    def search(self, name, country):
+        """
+        Searches the OFAC list for an individual with the specified details.
+        The check returns zero hits if:
+            * request to the SDN API times out
+            * SDN API returns a non-200 status code response
+            * user is not found on the SDN list
+
+        Args:
+            name (str): Individual's full name.
+            country (str): ISO 3166-1 alpha-2 country code where the individual is from.
+        Returns:
+            dict: SDN API response.
+        """
+
+        params = urlencode({
+            'sources': self.sdn_list,
+            'api_key': self.api_key,
+            'type': 'individual',
+            'name': name,
+            'countries': country
+        })
+        sdn_check_url = '{api_url}?{params}'.format(
+            api_url=self.api_url,
+            params=params
+        )
+
+        try:
+            response = requests.get(sdn_check_url, timeout=settings.SDN_CHECK_REQUEST_TIMEOUT)
+        except requests.exceptions.Timeout:
+            logger.exception('Connection to US Treasury SDN API timed out for [%s].', name)
+            raise
+
+        if response.status_code != 200:
+            logger.exception(
+                'Unable to connect to US Treasury SDN API for [%s]. Status code [%d] with message: [%s]',
+                name, response.status_code, response.content
+            )
+            raise requests.exceptions.HTTPError('Unable to connect to SDN API')
+
+        return json.loads(response.content)
+
+    def deactivate_user(self, user, site_configuration, name, country, search_results):
+        """ Deactivates a user account.
+
+        Args:
+            user (User): User whose account should be deactivated.
+            site_configuration (SiteConfiguration): The current site's configuration.
+            name (str): The user's name.
+            country (str): ISO 3166-1 alpha-2 country code where the individual is from.
+            search_results (dict): Results from a call to `search` that will
+                be recorded as the reason for the deactivation.
+        """
+        SDNCheckFailure.objects.create(
+            full_name=name,
+            username=user.username,
+            country=country,
+            sdn_check_response=search_results
+        )
+        logger.warning('SDN check failed for user [%s]', name)
+        user.deactivate_account(site_configuration)
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -244,6 +244,8 @@ CREDIT_PROVIDER_CACHE_TIMEOUT = 600
 
 VOUCHER_CACHE_TIMEOUT = 10  # Value is in seconds.
 
+SDN_CHECK_REQUEST_TIMEOUT = 5  # Value is in seconds.
+
 # APP CONFIGURATION
 DJANGO_APPS = [
     'django.contrib.admin',
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -47,7 +47,7 @@ define([
         onFail = function(){
             var message = gettext('Problem occurred during checkout. Please contact support');
             $('#messages').empty().append(
-                _s.sprintf('<div class="error">%s</div>', message)
+                _s.sprintf('<div class="alert alert-error">%s</div>', message)
             );
         },
         onSuccess = function (data) {
@@ -170,6 +170,33 @@ define([
 
         isCardTypeSupported = function (cardType) {
             return $.inArray(cardType, ['amex', 'discover', 'mastercard', 'visa']) > -1;
+        },
+
+        sdnCheck = function(event) {
+            var first_name = $('input[name=first_name]').val(),
+                last_name = $('input[name=last_name]').val(),
+                country = $('select[name=country]').val();
+
+            $.ajax({
+                url: '/api/v2/sdn/search/',
+                method: 'POST',
+                contentType: 'application/json; charset=utf-8',
+                dataType: 'json',
+                headers: {
+                    'X-CSRFToken': Cookies.get('ecommerce_csrftoken')
+                },
+                data: JSON.stringify({
+                    'name': _s.sprintf('%s %s', first_name, last_name),
+                    'country': country
+                }),
+                async: false,
+                success: function(data) {
+                    if (data.hits > 0) {
+                        event.preventDefault();
+                        Utils.redirect('/payment/sdn/failure/');
+                    }
+                }
+            });
         },
 
         onReady = function() {
@@ -304,6 +331,9 @@ define([
                 }
                 cardInfoValidation(e);
                 cardHolderInfoValidation(e);
+                if ($('input[name=sdn-check]').val() === 'enabled') {
+                    sdnCheck(e);
+                }
             });
 
             $paymentButtons.find('.payment-button').click(function (e) {
@@ -360,6 +390,7 @@ define([
             onFail: onFail,
             onReady: onReady,
             onSuccess: onSuccess,
+            sdnCheck: sdnCheck,
             showVoucherForm: showVoucherForm,
         };
     }
--- a/ecommerce/static/js/payment_processors/cybersource.js
+++ b/ecommerce/static/js/payment_processors/cybersource.js
@@ -70,6 +70,9 @@ require([
                                     field.focus();
                                 }
                             }
+                        } else {
+                            // Unhandled error types should default to the general payment error page.
+                            window.location.href = '/payment/error/';
                         }
                     }
                 }
--- a/ecommerce/templates/oscar/checkout/sdn_failure.html
+++ b/ecommerce/templates/oscar/checkout/sdn_failure.html
@@ -1,0 +1,26 @@
+{% extends 'edx/base.html' %}
+{% load i18n %}
+{% load staticfiles %}
+
+{% block title %}SDN Check Failure{% endblock %}
+
+{% block navbar %}
+    {% include 'edx/partials/_student_navbar.html' %}
+{% endblock navbar %}
+
+{% block content %}
+    <div id="error-message">
+        <div class="container">
+            <div class="depth depth-2 message-error-content">
+                <h3>SDN Check Failure</h3>
+                <p>{% with "<a class='nav-link' href='mailto:ofac.reconsideration@treasury.gov'>"|safe as ofac_email_link %}
+                    {% blocktrans with end_link="</a>"|safe  %}
+                    Unfortunately, your account profile or payment information appears to match one or more records on a U.S. Treasury Department sanctions list. This means we cannot complete your transaction or provide you with services and must suspend your learner account.
+
+                    If you have questions regarding clearing a match, please contact {{ ofac_email_link }}ofac.reconsideration@treasury.gov{{ end_link }} for information about options for clearing a match. Your account will be suspended until this matter is resolved satisfactorily.
+                    {% endblocktrans %}
+                {% endwith %}</p>
+            </div>
+        </div>
+    </div>
+{% endblock %}
--- a/ecommerce/core/migrations/0024_auto_20170208_1520.py
+++ b/ecommerce/core/migrations/0024_auto_20170208_1520.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0023_siteconfiguration_send_refund_notifications'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='siteconfiguration',
+            name='enable_sdn_check',
+            field=models.BooleanField(default=False, help_text='Enable SDN check at checkout.', verbose_name='Enable SDN check'),
+        ),
+        migrations.AddField(
+            model_name='siteconfiguration',
+            name='sdn_api_key',
+            field=models.CharField(max_length=255, verbose_name='US Treasury SDN API key', blank=True),
+        ),
+        migrations.AddField(
+            model_name='siteconfiguration',
+            name='sdn_api_list',
+            field=models.CharField(help_text='A comma seperated list of Treasury OFAC lists to check against.', max_length=255, verbose_name='SDN lists', blank=True),
+        ),
+        migrations.AddField(
+            model_name='siteconfiguration',
+            name='sdn_api_url',
+            field=models.CharField(max_length=255, verbose_name='US Treasury SDN API URL', blank=True),
+        ),
+    ]
--- a/ecommerce/core/tests/test_models.py
+++ b/ecommerce/core/tests/test_models.py
@@ -1,3 +1,5 @@
+import json
+
 import ddt
 import httpretty
 import mock
@@ -174,6 +176,26 @@ class UserTests(CourseCatalogTestMixin, LmsApiMockMixin, TestCase):
         httpretty.disable()
         with self.assertRaises(VerificationStatusError):
             user.is_verified(self.site)
+
+    @httpretty.activate
+    def test_deactivation(self):
+        """Verify the deactivation endpoint is called for the user."""
+        user = self.create_user()
+        expected_response = {'user_deactivated': True}
+        self.mock_deactivation_api(self.request, user.username, response=json.dumps(expected_response))
+        self.assertEqual(user.deactivate_account(self.request.site.siteconfiguration), expected_response)
+
+    def test_deactivation_exception_handling(self):
+        """Verify an error is logged if an exception happens."""
+        def callback(*args):  # pylint: disable=unused-argument
+            raise ConnectionError
+        user = self.create_user()
+        self.mock_deactivation_api(self.request, user.username, response=callback)
+
+        with self.assertRaises(ConnectionError):
+            with mock.patch('ecommerce.core.models.log.exception') as mock_logger:
+                user.deactivate_account(self.request.site.siteconfiguration)
+                self.assertTrue(mock_logger.called)
 
 
 class BusinessClientTests(TestCase):
--- a/ecommerce/extensions/payment/migrations/0013_sdncheckfailure.py
+++ b/ecommerce/extensions/payment/migrations/0013_sdncheckfailure.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.utils.timezone
+import jsonfield.fields
+import django_extensions.db.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('payment', '0012_auto_20161109_1456'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='SDNCheckFailure',
+            fields=[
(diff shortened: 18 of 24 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1012"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/core/models.py
```diff
@@ -121,10 +121,50 @@ class SiteConfiguration(models.Model):
...
+        verbose_name=_('SDN lists'),
+        help_text=_('A comma seperated list of Treasury OFAC lists to check against.'),
+        max_length=255,
+        blank=True
+    )
 
     class Meta(object):
         unique_together = ('site', 'partner')
 
+    def sdn_check_url(self, full_name):
```
mjfrey: need to add &address={query} to this.  Search on both full name and billing address.

ecommerce/extensions/basket/views.py
```diff
@@ -96,6 +98,19 @@ def _determine_seat_type(self, product):
             seat_type = get_certificate_type_display_value(product.attr.seat_type)
         return seat_type
 
+    def get(self, *args, **kwargs):
```
mjfrey: Not going to call this here.  Will be called from the unified basket page once we have billing name and address.

ecommerce/core/models.py
vkaracic: Users don't have an address. We get the billing address for the payment processor after the payment has been made.

ecommerce/extensions/basket/views.py
vkaracic: Oh. So I should wait for that unified basket to land in order to see what information I can use from it then.

ecommerce/core/models.py
mjfrey: We will have the billing address from the client side payment page.  We need to use the purchaser name / address not the learner account name.

ecommerce/extensions/payment/models.py
```diff
@@ -46,5 +46,16 @@ class Meta(object):
         verbose_name = "Paypal Processor Configuration"
 
 
+class SDNCheckFailure(models.Model):
+    """ Record of SDN check failure. """
+    full_name = models.CharField(max_length=255)
+    address = models.CharField(max_length=255)
+    sdn_check_response = JSONField()
```
mjfrey: I think we are going to want to pull out specific info from the response rather than the entire response.  Let me check on this.

ecommerce/extensions/payment/utils.py
```diff
@@ -37,3 +48,37 @@ def middle_truncate(string, chars):
...
+        address(str): User's address.
+    Returns:
+        result (Bool): Whether or not there is a match.
+    """
+    site_config = request.site.siteconfiguration
+    basket = Basket.get_basket(request.user, request.site)
+    response = requests.get(site_config.sdn_check_url(full_name, address))
+
+    if response.status_code != status.HTTP_200_OK:
+        logger.info(
```
mjfrey: Make this a logger.error so we get a Splunk alert.  Also log the Full Name, basket id, and course trying to be purchased.

ecommerce/core/models.py
```diff
@@ -125,10 +125,58 @@ class SiteConfiguration(models.Model):
...
+            full_name(str): Full name of the user who is checked.
+            address(str): User's address.
+            country(str): User's country.
+
+        Returns:
+            URL constructed from saved SDN values and the user's full name.
+        """
+        return (
+            '{sdn_api}/?sources={sdn_list}&api_key={sdn_key}&type=individual'
+            '&name={full_name}&address={address}&country={country}'
```
mjfrey: it is countries not country.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +64,40 @@ def clean_field_value(value):
...
+    Returns:
+        result (Bool): Whether or not there is a match.
+    """
+    site_config = request.site.siteconfiguration
+    basket = Basket.get_basket(request.user, request.site)
+    response = requests.get(site_config.sdn_check_url(full_name, address, country))
+
+    if response.status_code != status.HTTP_200_OK:
+        logger.info(
+            'Unable to connect to US Treasury SDN API. Status code [%d] with message: %s',
```
mjfrey: We need to log the basket id when this happens so we have a record for which transaction.

ecommerce/extensions/payment/urls.py
```diff
@@ -1,14 +1,15 @@
...
-from ecommerce.extensions.payment.views import cybersource, PaymentFailedView
+from ecommerce.extensions.payment.views import cybersource, PaymentFailedView, SdnFailure
 from ecommerce.extensions.payment.views.paypal import PaypalPaymentExecutionView, PaypalProfileAdminView
 
 urlpatterns = [
     url(r'^cybersource/notify/$', cybersource.CybersourceNotifyView.as_view(), name='cybersource_notify'),
     url(r'^cybersource/redirect/$', cybersource.CybersourceInterstitialView.as_view(), name='cybersource_redirect'),
     url(r'^cybersource/submit/$', cybersource.CybersourceSubmitView.as_view(), name='cybersource_submit'),
     url(r'^error/$', PaymentFailedView.as_view(), name='payment_error'),
+    url(r'^sdn-failure/$', SdnFailure.as_view(), name='sdn-failure'),
```
mjfrey: SDNFailure.as_view
sdn_failure

ecommerce/templates/oscar/checkout/sdn_failure.html
```diff
@@ -0,0 +1,24 @@
...
+{% endblock navbar %}
+
+{% block content %}
+    <div id="error-message">
+        <div class="container">
+            <div class="depth depth-2 message-error-content">
+                <h3>SDN check failure.</h3>
+                <p>{% with "<a class='nav-link' href='mailto:ofac.reconsideration@treasury.gov'>"|safe as ofac_email_link %}
+                    {% blocktrans with end_link="</a>"|safe  %}
+                    "Unfortunately, your account profile or payment information appears to match one or more records on a U.S. Treasury Department sanctions list. This means we cannot complete your transaction or provide you with services and must suspend your learner account. If you have questions regarding clearing a match, please contact {{ ofac_email_link }}ofac.reconsideration@treasury.gov{{ end_link }} for information about options for clearing a match. Your account will be suspended until this matter is resolved satisfactorily."
```
mjfrey: No need for " around this message.

ecommerce/extensions/payment/models.py
```diff
@@ -46,5 +46,23 @@ class Meta(object):
...
+class SDNCheckFailure(models.Model):
+    """ Record of SDN check failure. """
+    full_name = models.CharField(max_length=255)
+    country = models.CharField(max_length=255)
+    sdn_check_response = JSONField()
+    basket = models.ForeignKey('basket.Basket')
+    created = models.DateTimeField(auto_now_add=True)
+
+    def __unicode__(self):
+        return _(u'{full_name} [{country}] - basket [{basket_id}]').format(
```
marjev: Import `from __future__ import unicode_literals`

ecommerce/extensions/payment/tests/test_models.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class SDNCheckFailureTests(TestCase):
+    def setUp(self):
+        self.full_name = 'Darth Vader'
+        self.country = 'Galactic Empire'
+        self.sdn_check_response = {'description': 'One bad dude'}
+        self.basket = factories.BasketFactory()
+
+        self.failure_object = SDNCheckFailure.objects.create(
```
marjev: Create a factory

ecommerce/extensions/payment/tests/test_models.py
```diff
@@ -0,0 +1,29 @@
...
+            full_name=self.full_name,
+            country=self.country,
+            sdn_check_response=self.sdn_check_response,
+            basket=self.basket
+        )
+
+    def test_unicode_representation(self):
+        """Verify the __unicode__ method returns the correct value."""
+        expected = u'{full_name} [{country}] - basket [{basket_id}]'.format(
```
marjev: `from __future__ import unicode_literals`

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -163,8 +181,21 @@ def test_field_error(self):
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response['content-type'], JSON)
 
-        errors = json.loads(response.content)['field_errors']
-        self.assertIn(field, errors)
+        response_content = json.loads(response.content)
+        self.assertNotIn('sdn_failure', response_content.keys())
```
marjev: No need to call `keys`

ecommerce/extensions/payment/utils.py
```diff
@@ -1,6 +1,17 @@
 import re
 
+import json
+import logging
```
marjev: Why is there an empty row between `re` and `json`?

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +64,38 @@ def clean_field_value(value):
...
+
+    if response.status_code != status.HTTP_200_OK:
+        logger.info(
+            'Unable to connect to US Treasury SDN API for basket [%d]. Status code [%d] with message: %s',
+            basket.id, response.status_code, response.content
+        )
+        return True
+    elif json.loads(response.content)['total'] == 0:
+        return True
+    else:
```
marjev: no need for else

marjev (approved):
Nitpick comments. LGTM

ecommerce/extensions/payment/models.py
iivic: Nitpick: I don't see how this can be translated.

iivic (approved):

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -92,8 +88,17 @@ def form_valid(self, form):
...
+            return JsonResponse(
+                {'error': _('Your basket may have been modified or already purchased. Refresh the page to try again.')},
+                status=400
+            )
+
+        full_name = '{first_name} {last_name}'.format(
+            first_name=data['first_name'],
+            last_name=data['last_name']
+        )
+        if not sdn_check(request, full_name, data['country']):
```
mjfrey: let's make this a waffle sample.  That way we can ramp up the percentage.  Like we did for the basket page.  This way we can monitor it before we go to 100%.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +64,38 @@ def clean_field_value(value):
...
+        request (Request): The request object made to the view.
+        full_name(str): Full name of the user who is checked.
+        country(str): User's country.
+
+    Returns:
+        result (Bool): Whether or not there is a match.
+    """
+    site_config = request.site.siteconfiguration
+    basket = Basket.get_basket(request.user, request.site)
+    response = requests.get(site_config.sdn_check_url(full_name, country))
```
mjfrey: We should specify a timeout on this request.  We do not want to wait around and hangup the checkout process if the API is slow.

mjfrey:
@douglashall FYI

mjfrey:
@clintonb Please take a look.

vkaracic:
Since deactivating an account can only be done by the ecommerce worker, and since I have 0 experience with the ecommerce worker, to avoid multiple deployments to stage and reverts, I'd like to ask @clintonb to take a look at the last commit to see if it can be done that way. Basically what needs to happen is a POST request to the LMS account deactivation endpoint, made by the ecommerce worker. Can the ecom worker be authenticated to make the request with ``oauth_access_token`` or should I do it this way: https://github.com/edx/ecommerce-worker/blob/master/ecommerce_worker/utils.py#L60 ?

ecommerce/core/models.py
```diff
@@ -126,10 +126,56 @@ class SiteConfiguration(models.Model):
...
+
+        Args:
+            full_name(str): Full name of the user who is checked.
+            country(str): User's country.
+
+        Returns:
+            URL constructed from saved SDN values and the user's full name and country.
+        """
+        return (
+            '{sdn_api}/?sources={sdn_list}&api_key={sdn_key}&type=individual'
```
clintonb: Use `urlencode` and `urllib.parse`.

ecommerce/core/models.py
```diff
@@ -493,6 +539,35 @@ def is_verified(self, site):
             log.exception(msg)
             raise VerificationStatusError(msg)
 
+    def deactivate(self, site):
+        """Deactive the user's account. Ecommerce worker is the only that has the necessary
```
clintonb: Remove this talk about ecommerce worker. That is not actually relevant.

ecommerce/core/models.py
```diff
@@ -493,6 +539,35 @@ def is_verified(self, site):
...
+
+        Returns:
+            Response from the deactivation API endpoint.
+
+        Raises:
+            ConnectionError, SlumberBaseException and Timeout for failures in
+            establishing a connection with the LMS account deactivation API endpoint.
+        """
+        try:
+            ecom_worker = User.objects.get(username=settings.ECOMMERCE_SERVICE_WORKER_USERNAME)
```
clintonb:
```python
user_api_client = EdxRestApiClient(site_configuration.build_lms_url('/api/user/v1'), jwt=site_configuration.access_token)
```

ecommerce/core/models.py
```diff
@@ -493,6 +539,35 @@ def is_verified(self, site):
...
+            api = EdxRestApiClient(
+                site.siteconfiguration.build_lms_url('/api/user/v1'),
+                oauth_access_token=ecom_worker.access_token
```
clintonb: We should be using JWT authentication for all of our APIs. If the User API doesn't support JWT authentication, please update the API.

ecommerce/core/models.py
```diff
@@ -493,6 +539,35 @@ def is_verified(self, site):
...
+            )
+            response = api.accounts(self.username).deactivate().post()
+            return response
+        except (ConnectionError, SlumberBaseException, Timeout):
```
clintonb: The exception type doesn't matter, especially since you are re-raising it.

ecommerce/core/models.py
```diff
@@ -493,6 +539,35 @@ def is_verified(self, site):
...
+            log.exception(
+                'Failed to retrieve account details for [%s]',
```
clintonb: `'Failed to deactivate account for user [%s]'`

ecommerce/extensions/payment/models.py
```diff
@@ -46,5 +48,23 @@ class Meta(object):
         verbose_name = "Paypal Processor Configuration"
 
 
+class SDNCheckFailure(models.Model):
+    """ Record of SDN check failure. """
+    full_name = models.CharField(max_length=255)
+    country = models.CharField(max_length=255)
+    sdn_check_response = JSONField()
+    basket = models.ForeignKey('basket.Basket')
```
clintonb: This field needs to support the case when we delete old baskets. `on_delete=models.SET_NULL`

ecommerce/extensions/payment/models.py
```diff
@@ -46,5 +48,23 @@ class Meta(object):
...
+
+    def __unicode__(self):
+        return '{full_name} [{country}] - basket [{basket_id}]'.format(
+            full_name=self.full_name,
+            country=self.country,
+            basket_id=self.basket.id
+        )
+
+    class Meta(object):
+        verbose_name = "SDN Check Failure"
```
clintonb: Single quotes.

ecommerce/extensions/payment/models.py
```diff
@@ -46,5 +48,23 @@ class Meta(object):
...
+class SDNCheckFailure(models.Model):
+    """ Record of SDN check failure. """
+    full_name = models.CharField(max_length=255)
+    country = models.CharField(max_length=255)
+    sdn_check_response = JSONField()
+    basket = models.ForeignKey('basket.Basket')
+    created = models.DateTimeField(auto_now_add=True)
+
+    def __unicode__(self):
+        return '{full_name} [{country}] - basket [{basket_id}]'.format(
```
clintonb: What's the use case for this? This is a lot of info. I worry that it will be too much when displayed in Django admin.

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -22,3 +34,87 @@ def test_clean_field_value(self):
...
+    """ Tests for the SDN check function. """
+
+    def setUp(self):
+        super(SDNCheckTests, self).setUp()
+        self.request = RequestFactory()
+        self.request.COOKIES = {}
+        self.username = 'Dr. Evil'
+        self.country = 'Evilland'
+        self.request.user = self.create_user(full_name=self.username)
+        site_configuration = SiteConfigurationFactory(
```
clintonb: Why not simply update `self.site.siteconfiguration`? Alternatively, you can update `SiteConfigurationFactory`.

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -22,3 +34,87 @@ def test_clean_field_value(self):
...
+        """ Mock the SDN check API endpoint response. """
+        httpretty.register_uri(
+            httpretty.GET,
+            self.request.site.siteconfiguration.sdn_check_url(self.username, self.country),
+            status=status_code,
+            body=response,
+            content_type='application/json'
+        )
+
+    def assert_sdn_check_failure(self, basket, response):
```
clintonb: `assert_sdn_check_failure_recorded`

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -45,6 +48,9 @@ def setUp(self):
         self.user = self.create_user()
         self.client.login(username=self.user.username, password=self.password)
         self.site.siteconfiguration.enable_otto_receipt_page = True
+        self.site.siteconfiguration.sdn_api_url = 'http://sdn.check'
```
clintonb: Consider updating `SiteConfigurationFactory`.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -112,11 +118,25 @@ def test_invalid_basket_status(self, status):
         error_msg = 'Your basket may have been modified or already purchased. Refresh the page to try again.'
         self._assert_basket_error(basket.id, error_msg)
 
+    def mock_sdn_check_api_response(self, results):
```
clintonb: Why is there SDN code in the CyberSource test module? The two should be independent, so I would not expect to see any SDN code here. If there needs to be some form of integration test, the code should be refactored to avoid the co-mingling.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +66,52 @@ def clean_field_value(value):
...
+    if waffle.flag_is_active(request, SDN_CHECK_FLAG_NAME):
+        site_config = request.site.siteconfiguration
+        basket = Basket.get_basket(request.user, request.site)
+        try:
+            response = requests.get(
+                site_config.sdn_check_url(full_name, country),
+                timeout=settings.SDN_CHECK_REQUEST_TIMEOUT
+            )
+        except requests.exceptions.Timeout:
+            logger.info('Connection to US Treasury SDN API timed out for basket [%d].', basket.id)
```
clintonb: `logger.warning`

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +66,52 @@ def clean_field_value(value):
...
+            return True
+
+        if response.status_code != status.HTTP_200_OK:
+            logger.info(
```
clintonb: `logger.warning`

ecommerce/settings/base.py
```diff
@@ -244,6 +244,8 @@
 
 VOUCHER_CACHE_TIMEOUT = 10  # Value is in seconds.
 
+SDN_CHECK_REQUEST_TIMEOUT = 5
```
clintonb: Include a comment denoting the units.

ecommerce/static/js/payment_processors/cybersource.js
```diff
@@ -70,6 +70,11 @@ require([
                                     field.focus();
                                 }
                             }
+                        } else if (error.sdn_failure) {
```
clintonb: We definitely should not have SDN-related JS here. The ultimate goal is to support additional payment processors. That doesn't scale if they all need to know about SDN.

ecommerce/core/models.py
```diff
@@ -126,10 +126,56 @@ class SiteConfiguration(models.Model):
...
+        verbose_name=_('SDN lists'),
+        help_text=_('A comma seperated list of Treasury OFAC lists to check against.'),
+        max_length=255,
+        blank=True
+    )
 
     class Meta(object):
         unique_together = ('site', 'partner')
 
+    def sdn_check_url(self, full_name, country):
```
clintonb: This doesn't belong here. It would be ideal if the SDN code is confined to its own class:

```python
class SDNValidator(object):
  def __init__(self, api_url, api_key, api_list):
    ...

  def search(self, name, country):
     ...
```

ecommerce/templates/oscar/checkout/sdn_failure.html
```diff
@@ -0,0 +1,24 @@
+{% extends 'edx/base.html' %}
+{% load i18n %}
+{% load staticfiles %}
+
+{% block title %}{% endblock %}
```
clintonb: You need a title.

ecommerce/templates/oscar/checkout/sdn_failure.html
```diff
@@ -0,0 +1,24 @@
...
+{% endblock navbar %}
+
+{% block content %}
+    <div id="error-message">
+        <div class="container">
+            <div class="depth depth-2 message-error-content">
+                <h3>SDN check failure.</h3>
+                <p>{% with "<a class='nav-link' href='mailto:ofac.reconsideration@treasury.gov'>"|safe as ofac_email_link %}
+                    {% blocktrans with end_link="</a>"|safe  %}
+                    Unfortunately, your account profile or payment information appears to match one or more records on a U.S. Treasury Department sanctions list. This means we cannot complete your transaction or provide you with services and must suspend your learner account. If you have questions regarding clearing a match, please contact {{ ofac_email_link }}ofac.reconsideration@treasury.gov{{ end_link }} for information about options for clearing a match. Your account will be suspended until this matter is resolved satisfactorily.
```
clintonb: Use a line break to make this legible for development.

ecommerce/templates/oscar/checkout/sdn_failure.html
```diff
@@ -0,0 +1,24 @@
...
+{% load staticfiles %}
+
+{% block title %}{% endblock %}
+
+{% block navbar %}
+    {% include 'edx/partials/_student_navbar.html' %}
+{% endblock navbar %}
+
+{% block content %}
+    <div id="error-message">
```
clintonb, line 12: Nit: this should be a class, not an ID.

ecommerce/templates/oscar/checkout/sdn_failure.html
```diff
@@ -0,0 +1,24 @@
...
+        <div class="container">
+            <div class="depth depth-2 message-error-content">
+                <h3>SDN check failure.</h3>
```
clintonb: SDN Check Failure

clintonb (changes requested):

ecommerce/extensions/payment/models.py
vkaracic: To have a quick overview of all the matches, should be easier to find someone instead of searching through each object. I don't think it will be too much, the country is two characters and basket_id just a number. Perhaps I should change the full name to username?

vkaracic:
@clintonb I believe I have addressed all the comments. Can you take another look please?

ecommerce/core/models.py
```diff
@@ -507,6 +531,30 @@ def is_verified(self, site):
...
+        Returns:
+            Response from the deactivation API endpoint.
+        """
+        try:
+            api = EdxRestApiClient(
+                site_configuration.build_lms_url('/api/user/v1'),
+                jwt=site_configuration.access_token
+            )
+            response = api.accounts(self.username).deactivate().post()
+            return response
```
clintonb: Nit: single line

ecommerce/core/models.py
```diff
@@ -507,6 +531,30 @@ def is_verified(self, site):
             log.exception(msg)
             raise VerificationStatusError(msg)
 
+    def deactivate(self, site_configuration):
```
mjfrey: deactivate_user_account

ecommerce/extensions/api/v2/views/sdn_check.py
```diff
@@ -0,0 +1,30 @@
+"""API endpoint for performing an SDN check on users."""
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from rest_framework.permissions import IsAuthenticated
+
+from ecommerce.extensions.payment.utils import SDNValidator
+
+
+class SDNCheckViewSet(APIView):
```
mjfrey: The only problem I have with this is any logged in user can call this endpoint.  Do we want that? Can this be part of the payment form validation on the server?  Have it be a mixin so it is not payment processor specific?

mjfrey (changes requested):

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +67,70 @@ def clean_field_value(value):
         A cleaned string.
     """
     return re.sub(r'[\^:"\']', '', value)
+
+
+class SDNValidator(object):
```
clintonb: `SDNClient`

ecommerce/extensions/payment/models.py
```diff
@@ -46,5 +48,25 @@ class Meta(object):
         verbose_name = "Paypal Processor Configuration"
 
 
+class SDNCheckFailure(models.Model):
+    """ Record of SDN check failure. """
+    full_name = models.CharField(max_length=255)
+    username = models.CharField(max_length=255)
+    country = models.CharField(max_length=255)
+    sdn_check_response = JSONField()
+    basket = models.ForeignKey('basket.Basket', blank=True, null=True, on_delete=models.SET_NULL)
```
clintonb: How/why is the basket relevant?

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +67,70 @@ def clean_field_value(value):
...
+                )
+                return True
+            elif json.loads(response.content)['total'] == 0:
+                return True
+
+            request.user.deactivate(request.site.siteconfiguration)
+            SDNCheckFailure.objects.create(
+                full_name=name,
+                username=request.user.username,
+                account_deactivated=request.user.has_usable_password(),
```
clintonb: This is not the correct method to use. At no point do we set the user's password on Otto.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +67,70 @@ def clean_field_value(value):
...
+            if response.status_code != status.HTTP_200_OK:
+                logger.warning(
+                    'Unable to connect to US Treasury SDN API for basket [%d]. Status code [%d] with message: [%s]',
+                    basket.id, response.status_code, response.content
+                )
+                return True
+            elif json.loads(response.content)['total'] == 0:
+                return True
+
+            request.user.deactivate(request.site.siteconfiguration)
```
clintonb: You should record the response before attempting to deactivate the account. If the account deactivation fails, you will have no record of the SDN failure.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +67,70 @@ def clean_field_value(value):
...
+
+
+class SDNValidator(object):
+    """A utility class that handles SDN related operations."""
+    def __init__(self, api_url, api_key, sdn_list):
+        self.api_url = api_url
+        self.api_key = api_key
+        self.sdn_list = sdn_list
+
+    def search(self, request, name, country):
```
clintonb: Pass in the basket, if it's actually needed, and pull the user and site from it. Avoid relying on `Request` objects, as that makes testing difficult and limits the usefulness of the class.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +67,70 @@ def clean_field_value(value):
...
+                params=params
+            )
+            basket = Basket.get_basket(request.user, request.site)
+            try:
+                response = requests.get(sdn_check_url, timeout=settings.SDN_CHECK_REQUEST_TIMEOUT)
+            except requests.exceptions.Timeout:
+                logger.warning('Connection to US Treasury SDN API timed out for basket [%d].', basket.id)
+                return True
+
+            if response.status_code != status.HTTP_200_OK:
```
clintonb: Use `200`. There is no need to introduce a DRF dependency here.

ecommerce/extensions/api/v2/views/sdn_check.py
```diff
@@ -0,0 +1,30 @@
...
+        name = request.data['name']
+        country = request.data['country']
+
+        site_configuration = request.site.siteconfiguration
+        sdn_check = SDNValidator(
+            api_url=site_configuration.sdn_api_url,
+            api_key=site_configuration.sdn_api_key,
+            sdn_list=site_configuration.sdn_api_list
+        )
+        if not sdn_check.search(request, name, country):
```
clintonb: A bit more future-proofing. We made the mistake with the Refunds API of making assumption about our responses. I would make `sdn_check.search` return the number of hits, instead of a simple boolean. This allows the API to later be expanded, if necessary. Additionally, your client(s) can make decisions on their own regarding how to interpret the data. In this specific case that may not be useful; but, we should still maintain good API practices.

`return Response({'hits': sdn_check.search(request, name, country)})`

ecommerce/extensions/api/v2/urls.py
```diff
@@ -78,6 +79,7 @@
     url(r'^providers/', include(PROVIDER_URLS, namespace='providers')),
     url(r'^publication/', include(ATOMIC_PUBLICATION_URLS, namespace='publication')),
     url(r'^refunds/', include(REFUND_URLS, namespace='refunds')),
+    url(r'^sdn-check/$', sdn_check_views.SDNCheckViewSet.as_view(), name='sdn_check')
```
clintonb: Future-proof this with name-spacing. This should be at `/sdn/search/`.

ecommerce/extensions/api/v2/views/sdn_check.py
```diff
@@ -0,0 +1,30 @@
+"""API endpoint for performing an SDN check on users."""
```
clintonb: `views/sdn.py`

ecommerce/extensions/api/v2/views/sdn_check.py
```diff
@@ -0,0 +1,30 @@
...
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from rest_framework.permissions import IsAuthenticated
```
clintonb: Alphabetize.

clintonb (changes requested):

ecommerce/extensions/api/v2/urls.py
mjfrey: @clintonb Do you think this is ok to have this public endpoint?  Meaning anyone logged into Otto can call this directly.

ecommerce/extensions/payment/models.py
vkaracic: @mjfrey is the basket relevant?

mjfrey: No.

ecommerce/extensions/api/v2/urls.py
clintonb: Not really, no. The issue I see with leaving this open is that the learner can potentially lock him/herself out of an account if the wrong name is entered or country selected. That's a problem, whether this is open or locked down, so I'll leave it to you to ponder/decide.

Did we do any analysis of how many existing users this might affect? If we look at every user who has placed an order in the past year, how many might be locked out by this?

mjfrey: No, not yet.  Good idea.

ecommerce/extensions/payment/utils.py
vkaracic: It seems that the basket is not necessary, but in this case the request is needed to check if the waffle flag is active.

vkaracic:
@clintonb comments addressed! Please don't mind the failing JS test, I'm still working on it. Can you please take another look?

ecommerce/extensions/api/v2/views/sdn.py
```diff
@@ -0,0 +1,28 @@
...
+    def post(self, request):
+        """
+        POST handler for the view. User data is posted to this handler
+        which performs an SDN check and returns whether the user passed
+        or failed.
+        """
+        name = request.data['name']
+        country = request.data['country']
+
+        site_configuration = request.site.siteconfiguration
```
clintonb: Do your waffle checking here.

ecommerce/extensions/payment/utils.py
clintonb: If the flag isn't on, we shouldn't even call this method. Do the switching in the code that calls this method.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +66,73 @@ def clean_field_value(value):
...
+            * request to the SDN API times out
+            * SDN API returns a non-200 status code response
+            * user is not found on the SDN list
+
+        Args:
+            request (WSGIRequest): The request from which the site configuration is retrieved.
+            name (str): User's full name.
+            country (str): ISO alpha-2 country code where the user is from.
+        Returns:
+            The number of matches that have been found on the SDN list for the user.
```
clintonb: int: The number....

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +66,73 @@ def clean_field_value(value):
...
+                logger.warning('Connection to US Treasury SDN API timed out for user [%d].', user.id)
+                return None
+
+            if response.status_code != 200:
+                logger.warning(
+                    'Unable to connect to US Treasury SDN API for user [%d]. Status code [%d] with message: [%s]',
+                    user.id, response.status_code, response.content
+                )
+                return None
+            elif json.loads(response.content)['total'] == 0:
```
clintonb: This can be cleaned up...

```python
response = response.json()
total = response['total']

if total > 0:
  # Log SDN failure

return total
```

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +66,73 @@ def clean_field_value(value):
...
+                return None
+
+            sdn_failure = SDNCheckFailure.objects.create(
```
clintonb: After further thought, this should be a separate method. `search` should only worry about search. A separate method, `deactivate_user`, should worry about user deactivation. Why? 

1. This separation keeps the code cleaner, by ensuring that each method does one thing well.
2. This separation allows us to use the `search` method later (e.g. for debugging) without fear of deactivating a user.


```python


class SDNClient(object):
    def __init__(self, api_url, api_key, sdn_list):
        self.api_url = api_url
        self.api_key = api_key
        self.sdn_list = sdn_list
        self.timeout = settings.SDN_CHECK_REQUEST_TIMEOUT

    def search(self, name, country):
        """ Searches the OFAC list for an individual with the specified details.

        Args:
            name (str): Full name.
            country (str): ISO 3166-1 alpha-2 country code.

        Returns:
            dict: Search results.

        Raises:
            Timeout: if the request to the external API times out.
        """
        pass

    def deactivate_user(self, user, search_results):
        """ Deactivates a user account.

        Args:
            user (User): User whose account should be deactivated.
            search_results (dict): Results from a call to `search` that will
                be recorded as the reason for the deactivation.
        """
        pass

```

ecommerce/static/js/pages/basket_page.js
```diff
@@ -172,6 +172,38 @@ define([
             return $.inArray(cardType, ['amex', 'discover', 'mastercard', 'visa']) > -1;
         },
 
+        redirectPage = function(path) {
+            console.log('got here');
```
clintonb: Remove.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +66,73 @@ def clean_field_value(value):
...
+            * user is not found on the SDN list
+
+        Args:
+            request (WSGIRequest): The request from which the site configuration is retrieved.
+            name (str): User's full name.
+            country (str): ISO alpha-2 country code where the user is from.
+        Returns:
+            The number of matches that have been found on the SDN list for the user.
+        """
+        if waffle.flag_is_active(request, SDN_CHECK_FLAG_NAME):
```
clintonb: Actually...why is `waffle` necessary? Use `SiteConfiguration.enable_sdn_check`.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -304,6 +336,7 @@ define([
                 }
                 cardInfoValidation(e);
                 cardHolderInfoValidation(e);
+                sdnCheck(e);
```
clintonb: This should only be called if SDN validation is active for the site.

ecommerce/extensions/payment/urls.py
```diff
@@ -1,14 +1,15 @@
...
-from ecommerce.extensions.payment.views import cybersource, PaymentFailedView
+from ecommerce.extensions.payment.views import cybersource, PaymentFailedView, SDNFailure
 from ecommerce.extensions.payment.views.paypal import PaypalPaymentExecutionView, PaypalProfileAdminView
 
 urlpatterns = [
     url(r'^cybersource/notify/$', cybersource.CybersourceNotifyView.as_view(), name='cybersource_notify'),
     url(r'^cybersource/redirect/$', cybersource.CybersourceInterstitialView.as_view(), name='cybersource_redirect'),
     url(r'^cybersource/submit/$', cybersource.CybersourceSubmitView.as_view(), name='cybersource_submit'),
     url(r'^error/$', PaymentFailedView.as_view(), name='payment_error'),
+    url(r'^sdn-failure/$', SDNFailure.as_view(), name='sdn_failure'),
```
clintonb: `sdn/failure`

ecommerce/extensions/payment/models.py
```diff
@@ -46,5 +48,24 @@ class Meta(object):
...
 
 
+class SDNCheckFailure(models.Model):
+    """ Record of SDN check failure. """
+    full_name = models.CharField(max_length=255)
+    username = models.CharField(max_length=255)
+    country = models.CharField(max_length=255)
+    sdn_check_response = JSONField()
+    account_deactivated = models.BooleanField()
+    created = models.DateTimeField(auto_now_add=True)
```
clintonb: Inherit `TimeStampedModel`, and this can go away.

ecommerce/extensions/payment/admin.py
```diff
@@ -36,3 +38,4 @@ def formatted_response(self, obj):
 
 
 admin.site.register(PaypalProcessorConfiguration, SingletonModelAdmin)
+admin.site.register(SDNCheckFailure)
```
clintonb: Put yourself in the shoes of someone tracking down an SDN failure. We're going to need search—on user.username and user.email—and a more useful list view—without the search results.

General rule of thumb: if a model is worth adding to the admin (and most are), customize it to make it truly useful.

ecommerce/extensions/api/v2/tests/views/test_sdn.py
```diff
@@ -0,0 +1,45 @@
...
+            data=json.dumps({
+                'name': 'Tester',
+                'country': 'TE'
+            }),
+            content_type=JSON_CONTENT_TYPE
+        )
+
+    def test_authentication_required(self):
+        """Verify only authenticated users can access endpoint."""
+        self.assertEqual(self.make_request().status_code, status.HTTP_200_OK)
```
clintonb: Without mocking, this call will simply wait for the API call to fail. This will further increase the time it takes to run tests. Remove this line since this test only wants to validate that unauthenticated users are turned away.

ecommerce/extensions/api/v2/tests/views/test_sdn.py
```diff
@@ -0,0 +1,45 @@
...
+        self.client.logout()
+        self.assertEqual(self.make_request().status_code, status.HTTP_401_UNAUTHORIZED)
+
+    @ddt.data(0, 1)
+    def test_sdn_check_match(self, sdn_hits):
+        """Verify the endpoint returns the number of hits SDN check made."""
+        with mock.patch.object(SDNClient, 'search') as sdn_validator_mock:
+            sdn_validator_mock.return_value = sdn_hits
+            response = self.make_request()
+            self.assertEqual(json.loads(response.content)['hits'], sdn_hits)
```
clintonb: Add: `self.assertTrue(sdn_validator_mock.called)`

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -22,3 +37,105 @@ def test_clean_field_value(self):
...
+        with mock.patch.object(User, 'deactivate_account') as deactivate_account:
+            deactivate_account.return_value = True
+            self.assertEqual(self.sdn_validator.search(self.request, self.name, self.country), sdn_response['total'])
+            self.assert_sdn_check_failure_recorded(json.dumps(sdn_response))
+
+    @httpretty.activate
+    def test_sdn_check_pass(self):
+        """ Verify the SDN check returns None if the user passed and no failure is saved. """
+        self.mock_sdn_response(json.dumps({'total': 0}))
+        BasketFactory(owner=self.request.user, site=self.request.site)
```
clintonb: This is not needed.

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -22,3 +37,105 @@ def test_clean_field_value(self):
...
+        self.assertEqual(SDNCheckFailure.objects.count(), 0)
+        self.assertIsNone(self.sdn_validator.search(self.request, self.name, self.country))
+
+    @httpretty.activate
+    def test_sdn_check_match(self):
+        """ Verify the SDN check returns the number of matches and records the match. """
+        sdn_response = {'total': 1}
+        self.mock_sdn_response(json.dumps(sdn_response))
+        BasketFactory(owner=self.request.user, site=self.request.site)
```
clintonb: This is not needed.

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -1,4 +1,19 @@
-from ecommerce.extensions.payment.utils import clean_field_value, middle_truncate
+import json
+import time
+from urllib import urlencode
+
+import mock
+import httpretty
+from django.conf import settings
+from django.test import override_settings, RequestFactory
+from oscar.test.factories import BasketFactory
```
clintonb: This is not needed.

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -22,3 +37,105 @@ def test_clean_field_value(self):
...
         self.assertEqual(clean_field_value(value), 'Sometexttest-value')
+
+
+@override_flag(SDN_CHECK_FLAG_NAME, active=True)
+class SDNCheckTests(TestCase):
+    """ Tests for the SDN check function. """
+
+    def setUp(self):
+        super(SDNCheckTests, self).setUp()
+        self.request = RequestFactory()
```
clintonb: You should not need to make any changes to `self.request`.

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -22,3 +37,105 @@ def test_clean_field_value(self):
...
+        self.name = 'Dr. Evil'
+        self.country = 'Evilland'
+        self.request.user = self.create_user(full_name=self.name)
+        self.site_configuration = SiteConfigurationFactory(
+            enable_sdn_check=True,
+            sdn_api_url='http://sdn-test.fake/',
+            sdn_api_key='fake-key',
+            sdn_api_list='SDN,TEST'
+        )
+        self.request.site = self.site_configuration.site
```
clintonb: Remove, along with the other changes to `self.request`.

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -22,3 +37,105 @@ def test_clean_field_value(self):
...
+    """ Tests for the SDN check function. """
+
+    def setUp(self):
+        super(SDNCheckTests, self).setUp()
+        self.request = RequestFactory()
+        self.request.COOKIES = {}
+        self.name = 'Dr. Evil'
+        self.country = 'Evilland'
+        self.request.user = self.create_user(full_name=self.name)
+        self.site_configuration = SiteConfigurationFactory(
```
clintonb: You're duplicating a lot of work done by the base `TestCase` class:

```python
self.site_configuration = self.site.siteconfiguration
self.site_configuration. enable_sdn_check = True
...
self.site_configuration.save()
```

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -22,3 +37,105 @@ def test_clean_field_value(self):
         """ Verify the passed value is cleaned of specific special characters. """
         value = 'Some^text:\'test-value'
         self.assertEqual(clean_field_value(value), 'Sometexttest-value')
+
+
+@override_flag(SDN_CHECK_FLAG_NAME, active=True)
```
clintonb: This should not be needed if we're relying on `SiteConfiguration`.

clintonb (changes requested):

ecommerce/extensions/payment/utils.py
vkaracic: @mjfrey should we drop the waffle requirement?

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +66,73 @@ def clean_field_value(value):
...
+            params = urlencode({
+                'sources': self.sdn_list,
+                'api_key': self.api_key,
+                'type': 'individual',
+                'name': name,
+                'country': country
```
mjfrey: This should be countries not country.

ecommerce/extensions/payment/utils.py
mjfrey: Product wants to do a percentage based rollout so we can monitor and make sure we are not getting false positives.

ecommerce/static/js/pages/basket_page.js
vkaracic: A GET endpoint on the SDN API that serves information if SDN is active or not, or a context variable in the basket view to send to the JS part?

clintonb: Context variable.

ecommerce/extensions/payment/utils.py
vkaracic: Should the ``search`` method raise ``Timeout``? The idea was to silently fail and log the failure. That way if the Treasury API goes down it won't cause all payments to fail on our side.

vkaracic: Actually is this class needed? I think I could move all of this to the SDN API endpoint.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +63,76 @@ def clean_field_value(value):
...
+        The check returns zero hits if:
+            * request to the SDN API times out
+            * SDN API returns a non-200 status code response
+            * user is not found on the SDN list
+
+        Args:
+            name (str): Individual's full name.
+            country (str): ISO 3166-1 alpha-2 country code where the individual is from.
+        Returns:
+            json: SDN API response.
```
clintonb: Return the `dict`. There is no need to make the callers have to convert JSON to a `dict`.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +63,76 @@ def clean_field_value(value):
...
+            params=params
+        )
+
+        try:
+            response = requests.get(sdn_check_url, timeout=settings.SDN_CHECK_REQUEST_TIMEOUT)
+        except requests.exceptions.Timeout:
+            logger.warning('Connection to US Treasury SDN API timed out for [%s].', name)
+            return None
+
+        if response.status_code != 200:
```
clintonb: Ideally, you should raise an exception here and above. We should avoid special return values where possible. In the past I've established a base exception class (e.g. `SDNClientError`) and caught that. Given that the response to a timeout and a server error will be the same, you can differentiate with an error message rather than subclasses. If, however, we chose to retry a timeout, you'd want to make an appropriate exception subclass.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +63,76 @@ def clean_field_value(value):
...
+        """
+        sdn_failure = SDNCheckFailure.objects.create(
+            full_name=search_results['name'],
+            username=user.username,
+            account_deactivated=False,
+            country=search_results['country'],
+            sdn_check_response=search_results['response']
+        )
+        user.deactivate_account(site_configuration)
+        # TODO: Check if the user has actually been deactivated
```
davec-edx: Does this TODO still need to be done?

mjfrey (approved):
Once last couple comments addressed.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +63,73 @@ def clean_field_value(value):
...
+        SDNCheckFailure.objects.create(
+            full_name=search_results['name'],
+            username=user.username,
+            account_deactivated=False,
+            country=search_results['country'],
+            sdn_check_response=search_results['response']
+        )
+        user.deactivate_account(site_configuration)
+
+        logger.warning('SDN check failed for user [%s]', search_results['name'])
```
davec-edx: I'm not following what's happening here.  user.deactivate_account is being called, which returns a value if successful or throws an error if not.  Then we're always logging a warning about the SDN check failing.

If this is for handling the SDN error, please update the docstring, but also add the try/except, won't we need it?

mjfrey: We always will log a positive SDN match regardless of user deactivation.

davec-edx: Let me rephrase:

Primarily: If we call call into deactivate_account and it throws an error we will never hit the log message so perhaps the logging (line 135) should come before the call to deactivate.

Secondarily, if we call deactivate_account and it doesn't throw an error, we're never doing anything with the results of the POST (such as checking the return code for success); will that cause an inconsistency anyplace?  Will that appear to be a successful deactivation when in reality it wasn't?

ecommerce/extensions/api/v2/views/sdn.py
```diff
@@ -0,0 +1,43 @@
...
+        POST handler for the view. User data is posted to this handler
+        which performs an SDN check and returns whether the user passed
+        or failed.
+        """
+        name = request.data['name']
+        country = request.data['country']
+        hits = 0
+
+        site_configuration = request.site.siteconfiguration
+        if waffle.flag_is_active(request, SDN_CHECK_FLAG_NAME) and site_configuration.enable_sdn_check:
```
clintonb: Ditch `waffle`. It makes no sense. We are either using SDN, or we are not. FYI @mjfrey

ecommerce/extensions/payment/models.py
```diff
@@ -46,5 +49,22 @@ class Meta(object):
         verbose_name = "Paypal Processor Configuration"
 
 
+class SDNCheckFailure(TimeStampedModel):
+    """ Record of SDN check failure. """
+    full_name = models.CharField(max_length=255)
+    username = models.CharField(max_length=255)
+    country = models.CharField(max_length=255)
```
clintonb: Country codes should be 2 characters.

ecommerce/extensions/payment/utils.py
```diff
@@ -53,3 +64,72 @@ def clean_field_value(value):
...
+        """ Deactivates a user account.
+
+        Args:
+            user (User): User whose account should be deactivated.
+            site_configuration (SiteConfiguration): The current site's configuration.
+            search_results (dict): Results from a call to `search` that will
+                be recorded as the reason for the deactivation.
+        """
+        SDNCheckFailure.objects.create(
+            full_name=search_results['name'],
```
clintonb: These should be separate arguments. Don't overload the search response argument in this manner.

ecommerce/extensions/api/v2/views/sdn.py
```diff
@@ -0,0 +1,43 @@
...
+            sdn_check = SDNClient(
+                api_url=site_configuration.sdn_api_url,
+                api_key=site_configuration.sdn_api_key,
+                sdn_list=site_configuration.sdn_api_list
+            )
+            response = json.loads(sdn_check.search(name, country))
```
clintonb: No need for `json.loads()` here.

clintonb (changes requested):

ecommerce/extensions/api/v2/views/sdn.py
vkaracic: Since this decision is a bit over my rank I'll let you guys (+ @edx/product) discuss it.

mjfrey: @clintonb @vkaracic We can remove the waffle switch.  We will just turn on / off as needed.

vkaracic:
@clintonb comments addressed, please take another look.

ecommerce/core/models.py
```diff
@@ -131,6 +131,30 @@ class SiteConfiguration(models.Model):
         blank=True,
         default=False
     )
+    enable_sdn_check = models.BooleanField(
+        verbose_name=_('Enable SDN check'),
+        help_text=_('Enable SDN check at checkout.'),
+        blank=True,
```
clintonb: `blank` is not applicable here.

ecommerce/core/models.py
```diff
@@ -131,6 +131,30 @@ class SiteConfiguration(models.Model):
...
+    )
+    sdn_api_url = models.CharField(
+        verbose_name=_('US Treasury SDN API URL'),
+        help_text=_('US Treasury SDN API URL.'),
+        max_length=255,
+        blank=True
+    )
+    sdn_api_key = models.CharField(
+        verbose_name=_('US Treasury SDN API key'),
+        help_text=_('US Treasury SDN API key.'),
```
clintonb: If the verbose name and help text are the same, the help text needs to go.

ecommerce/core/models.py
```diff
@@ -131,6 +131,30 @@ class SiteConfiguration(models.Model):
...
     )
+    enable_sdn_check = models.BooleanField(
+        verbose_name=_('Enable SDN check'),
+        help_text=_('Enable SDN check at checkout.'),
+        blank=True,
+        default=False
+    )
+    sdn_api_url = models.CharField(
+        verbose_name=_('US Treasury SDN API URL'),
+        help_text=_('US Treasury SDN API URL.'),
```
clintonb: If the verbose name and help text are the same, the help text needs to go.

ecommerce/core/models.py
```diff
@@ -512,6 +536,29 @@ def is_verified(self, site):
...
+
+        Args:
+            site_configuration (SiteConfiguration): The site configuration
+                from which the LMS account API endpoint is created.
+
+        Returns:
+            Response from the deactivation API endpoint.
+        """
+        try:
+            api = EdxRestApiClient(
```
clintonb: Nit: Move this to `SiteConfiguration.user_api_client`.

ecommerce/core/models.py
```diff
@@ -512,6 +536,29 @@ def is_verified(self, site):
...
+                site_configuration.build_lms_url('/api/user/v1'),
+                jwt=site_configuration.access_token
+            )
+            return api.accounts(self.username).deactivate().post()
+        except:  # pragma: no cover pylint: disable=bare-except
```
clintonb: `# pragma: no cover` is not acceptable here. You need to test the failure case.

ecommerce/extensions/api/v2/tests/views/test_sdn.py
```diff
@@ -0,0 +1,48 @@
...
+        self.client.logout()
+        self.assertEqual(self.make_request().status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_sdn_check_match(self):
+        """Verify the endpoint returns the number of hits SDN check made."""
+        with mock.patch.object(SDNClient, 'search') as sdn_validator_mock:
+            with mock.patch.object(User, 'deactivate_account') as deactivate_account_mock:
+                sdn_validator_mock.return_value = {'total': 1}
+                deactivate_account_mock.return_value = True
+                self.make_request()
```
clintonb: You need to validate the view's response.

ecommerce/extensions/payment/models.py
```diff
@@ -46,5 +49,20 @@ class Meta(object):
...
 
+class SDNCheckFailure(TimeStampedModel):
+    """ Record of SDN check failure. """
+    full_name = models.CharField(max_length=255)
+    username = models.CharField(max_length=255)
+    country = models.CharField(max_length=2)
+    sdn_check_response = JSONField()
+
+    def __unicode__(self):
+        return 'SDN check failure [{username}]'.format(  # pragma: no cover
```
clintonb: This is a very easy to test to write. Also, it should return a Unicode string.

ecommerce/extensions/payment/utils.py
clintonb: This needs to be addressed.

ecommerce/static/js/pages/basket_page.js
clintonb: This needs to be addressed.

clintonb (changes requested):

ecommerce/core/models.py
vkaracic: Test that a log entry was created?

clintonb: Yes. If it's worth logging, it's worth testing.

clintonb: Otherwise, drop the exception handler.

ecommerce/extensions/payment/models.py
vkaracic: I understand that it's not necessary to prefix a string with ``u`` if ``from __future__ import unicode_literals`` is imported?

ecommerce/extensions/payment/utils.py
vkaracic: I'm inclined not to subclass new exception classes. For the above case a ``Timeout`` exception could be raised. For this one I'm thinking maybe ``HTTPError('Unable to connect to SDN API')``?

vkaracic:
@clintonb comments addressed.

ecommerce/extensions/api/v2/views/sdn.py
```diff
@@ -0,0 +1,41 @@
...
+
+        site_configuration = request.site.siteconfiguration
+        if site_configuration.enable_sdn_check:
+            sdn_check = SDNClient(
+                api_url=site_configuration.sdn_api_url,
+                api_key=site_configuration.sdn_api_key,
+                sdn_list=site_configuration.sdn_api_list
+            )
+            response = sdn_check.search(name, country)
+            hits = response['total']
```
mjfrey, line 31: You might want to use response.get in case response does not contain 'total'

clintonb: Given this is from the Treasury API, there should always be a `'total'` key. Otherwise, they've violated their API contract. The government would *never* do that!

ecommerce/core/models.py
```diff
@@ -131,6 +131,27 @@ class SiteConfiguration(models.Model):
...
+        blank=True
+    )
+    sdn_api_key = models.CharField(
+        verbose_name=_('US Treasury SDN API key'),
+        max_length=255,
+        blank=True
+    )
+    sdn_api_list = models.CharField(
+        verbose_name=_('SDN lists'),
+        help_text=_('A comma seperated list of Treasury OFAC lists to check against.'),
```
clintonb: Nit: comma-separated

ecommerce/core/models.py
```diff
@@ -349,6 +370,16 @@ def enterprise_api_client(self):
...
 
+    @cached_property
+    def user_api_client(self):
+        """
+        Returns the API client to access the user API endpoint on LMS.
+
+        Returns:
+            EdxRestApiClient: The client to access the LMS user API service.
+        """
+        return EdxRestApiClient(self.build_lms_url('/api/user/v1/'), jwt=self.access_token)
```
clintonb, line 381: Does this need `append_slash=False`?

ecommerce/core/models.py
```diff
@@ -425,8 +456,8 @@ def account_details(self, request):
         try:
             api = EdxRestApiClient(
```
clintonb, line 457: `request.site_configuration.user_api_client`

ecommerce/extensions/api/v2/views/sdn.py
vkaracic: b5af8c5a51915aa14b9bcf695fd01e954610f9dbdd162beea1e58201b0e1b253

clintonb: /s

ecommerce/extensions/payment/admin.py
```diff
@@ -35,4 +37,18 @@ def formatted_response(self, obj):
     formatted_response.allow_tags = True
 
 
+@admin.register(SDNCheckFailure)
+class SDNCheckFailureAdmin(admin.ModelAdmin):
+    search_fields = ('username', 'full_name')
+    list_display = ('username', 'full_name', 'country')
+    fields = ('username', 'full_name', 'country', 'formatted_response')
+    readonly_fields = ('username', 'full_name', 'country', 'formatted_response')
+    show_full_result_count = False
```
clintonb: Nit: I highly doubt we will get enough hits—order of hundreds of thousands—for this to matter.

ecommerce/extensions/payment/models.py
clintonb: Get rid of the `# pragma: no cover`.

ecommerce/extensions/basket/views.py
```diff
@@ -279,6 +279,7 @@ def get_context_data(self, **kwargs):
                 unicode(course_key)
             ),
             'enable_client_side_checkout': False,
+            'sdn_check': 'enabled' if site_configuration.enable_sdn_check else 'disabled'
```
clintonb: Just use `True` or `False`. Avoid the special strings.

clintonb (approved):

clintonb (approved):

ecommerce/core/models.py
vkaracic: That depends. For account details endpoint there is no trailing slash, but for the deactivation endpoint there is.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -304,6 +331,9 @@ define([
                 }
                 cardInfoValidation(e);
                 cardHolderInfoValidation(e);
+                if ($('input[name=sdn-check]').val() === 'enabled') {
```
clintonb, line 334: This needs to be updated since the value was changed to `True`/`False`.

vkaracic: I've changed the template as well:
https://github.com/edx/ecommerce/pull/1012/files#diff-67f74e748e826f7ceb2290e1f960a1f1R139

PR #1014, opened 2016-11-15
Fixes the CSS of the error page.

vkaracic:
The padding was applied to the wrong selector. Basically now we have this:
screenshot from 2016-11-15 15-40-27

And we want this:
screenshot from 2016-11-15 15-39-20

@iivic please review.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1014"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

iivic (approved):
LGTM

PR #1016, opened 2016-11-16
Add custom verification status exception.

vkaracic:
@clintonb something like this?

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -16,6 +16,7 @@ from jsonfield.fields import JSONField
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
 
+from ecommerce.core.exceptions import VerificationStatusError
 from ecommerce.core.url_utils import get_lms_url
 from ecommerce.courses.utils import mode_for_seat
 from ecommerce.extensions.payment.exceptions import ProcessorNotFoundError
@@ -448,13 +449,13 @@ class User(AbstractUser):
             raise
         return response
 
-    def is_verified(self, request):
+    def is_verified(self, site):
         """
         Check if a user has verified his/her identity.
         Calls the LMS verification status API endpoint and returns the verification status information.
 
         Args:
-            request (WSGIRequest): The request from which the LMS account API endpoint is created.
+            site (Site): The site object from which the LMS account API endpoint is created.
 
         Returns:
             True if the user is verified, false otherwise.
@@ -465,19 +466,17 @@ class User(AbstractUser):
         """
         try:
             api = EdxRestApiClient(
-                request.site.siteconfiguration.build_lms_url('api/user/v1/'),
+                site.siteconfiguration.build_lms_url('api/user/v1/'),
                 oauth_access_token=self.access_token
             )
             response = api.accounts(self.username).verification_status().get()
             return response.get('is_verified', False)
         except HttpNotFoundError:
             return False
-        except (ConnectionError, SlumberBaseException, Timeout):  # pragma: no cover
-            log.exception(
-                'Failed to retrieve verification status details for [%s]',
-                self.username
-            )
-            raise
+        except (ConnectionError, SlumberBaseException, Timeout):
+            msg = 'Failed to retrieve verification status details for [{username}]'.format(username=self.username)
+            log.exception(msg)
+            raise VerificationStatusError(msg)
 
 
 class Client(User):
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -1,8 +1,10 @@
 import datetime
+import hashlib
 import logging
 from urlparse import urljoin
 
 from analytics import Client as SegmentClient
+from dateutil.parser import parse
 from django.conf import settings
 from django.contrib.auth.models import AbstractUser
 from django.contrib.sites.models import Site
@@ -10,6 +12,7 @@ from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.functional import cached_property
+from django.utils.timezone import now
 from django.utils.translation import ugettext_lazy as _
 from edx_rest_api_client.client import EdxRestApiClient
 from jsonfield.fields import JSONField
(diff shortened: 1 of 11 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1016"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb (approved):
Beautiful! Can you also take on caching in another commit/PR? If we get back data indicating the user has been verified, we should cache that response until that verification has expired. Errors/404s should not be cached.

vkaracic:
Will that take up too much cache space? The default verification expiration is +1 year, so in that year there might be a lot of verification information stored in cache.

clintonb:
@vkaracic good point. A cursory glance at Google shows that memcached (our cache of choice) uses a last recently used algorithm to evict items when space is full. Based on that knowledge, it seems memcached will do "the right thing" and get rid of stale verification statuses when necessary. 

@edx/devops @edx/ecommerce @cpennington @efagin thoughts on how long we should cache this data?

ecommerce/core/models.py
```diff
@@ -464,20 +469,25 @@ def is_verified(self, request):
...
+            response = cache.get(cache_hash)
+            if not response:
+                api = EdxRestApiClient(
+                    site.siteconfiguration.build_lms_url('api/user/v1/'),
+                    oauth_access_token=self.access_token
+                )
+                response = api.accounts(self.username).verification_status().get()
+
+                cache_timeout = int((parse(response.get('expiration_datetime')) - now()).total_seconds())
+                cache.set(cache_hash, response, cache_timeout)
```
clintonb: Since we only care about `response['is_verified']`, that should be all we store.

ecommerce/core/models.py
```diff
@@ -464,20 +469,25 @@ def is_verified(self, request):
             establishing a connection with the LMS verification status API endpoint.
         """
         try:
-            api = EdxRestApiClient(
-                request.site.siteconfiguration.build_lms_url('api/user/v1/'),
-                oauth_access_token=self.access_token
-            )
-            response = api.accounts(self.username).verification_status().get()
+            cache_key = 'verification_status_{username}'.format(username=self.username)
+            cache_hash = hashlib.md5(cache_key).hexdigest()
```
rlucioni: This doesn't strike me as necessary given the key function defined in common/djangoapps/util/memcache.py.

vkaracic: So something like `safe_key(<username>, 'verification_status', 1)` ?

rlucioni: No, you should be able to use the value of `cache_key` you have defined above. The `safe_key` function is called under the hood, having been specified as the `KEY_FUNCTION` for the configured memcache backend.

rlucioni: For example, `cache.set(cache_key, ...)`.

vkaracic: Okay well that's for edx-platform. I can't find any mention of `KEY_FUNCTION` in ecommerce settings.

ecommerce/core/models.py
```diff
@@ -464,20 +469,26 @@ def is_verified(self, request):
...
+            verification = cache.get(cache_hash)
+            if not verification:
+                api = EdxRestApiClient(
+                    site.siteconfiguration.build_lms_url('api/user/v1/'),
+                    oauth_access_token=self.access_token
+                )
+                response = api.accounts(self.username).verification_status().get()
+
+                verification = response.get('is_verified', False)
+                cache_timeout = int((parse(response.get('expiration_datetime')) - now()).total_seconds())
```
clintonb: We shouldn't waste space storing data when `'is_verified'` is `False`. We are just going to call the API again.

ecommerce/core/models.py
rlucioni: You are very right, I didn't realize what repo we were in. What you had is fine, unless we want to define a `safe_key` function and hook it up as necessary.

vkaracic: Probably not in this PR, as this logic is used everywhere where we cache something. In case we want to define `safe_key` (@clintonb thoughts?) that should be a different PR which would change all the places where we do this.

clintonb: I agree that `safe_key` is out of scope for these changes. MD5 is sufficient. If we need to centralize hashing of cache keys, we can do so in the future.

ecommerce/core/models.py
jibsheet: Does ecommerce actually have an elasticache deployed?  I see no sign of that in configuration.  It appears you're caching other things, and they're going in a local memory cache on each independent server in production.  You're also not pinning, so people can get a machine without their cache.  You don't need to pin, but you shouldn't have machine local caches.

You can cache for a year, just know that things will evict.  You should also provide a sense of scale so that the infrastructure can be deployed with an appropriate size.  There will need to be an OPS ticket for infrastructure.

clintonb:
@jibsheet DEVOPS-5072 filed to setup a cache server.

jibsheet:
@clintonb bigger than a DEVOPS ticket.  I'll move it into OPS for grooming.

ecommerce/core/models.py
```diff
@@ -464,20 +469,27 @@ def is_verified(self, request):
...
         """
         try:
-            api = EdxRestApiClient(
-                request.site.siteconfiguration.build_lms_url('api/user/v1/'),
-                oauth_access_token=self.access_token
-            )
-            response = api.accounts(self.username).verification_status().get()
-            return response.get('is_verified', False)
+            cache_key = 'verification_status_{username}'.format(username=self.username)
+            cache_hash = hashlib.md5(cache_key).hexdigest()
```
clintonb: I can totally see someone coming through in the future and using `cache_key` in the call to `cache.set()` instead of `cache_hash`. We can avoid this issue by renaming `cache_hash` to `cache_key`.

clintonb (approved):

ecommerce/core/models.py
vkaracic: True, I found an instance where I missed it even. I've added another commit changing this in every place we save caches this way.

clintonb: Nice catch!

clintonb (approved):
I'll leave it up to you if you want to squash your last two commits. Either way, you are clear to merge (or rebase, which is my personal favorite).

vkaracic:
I have enough un-squashed commits in master as is, I'll squash the last two :)
Thanks guys for the reviews!

PR #1017, opened 2016-11-16
[SOL-2137] Add Backend Validation for Coupon Creation - Benefit Value

iivic:
Values for benefit value that are not allowed: string, empty, negative numbers, when benefit type 'Percentage' max value is 100. https://openedx.atlassian.net/browse/SOL-2137
@vkaracic @mjfrey @marjev  Please review.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -262,21 +262,29 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         self.assertEqual(details_response['coupon_type'], 'Enrollment code')
         self.assertEqual(details_response['code_status'], 'ACTIVE')
 
+    def test_create_coupon_product_invalid_category_data(self):
+        """Test creating coupon when provided category data is invalid."""
+        self.data.update({'category': {'id': 10000, 'name': 'Category Not Found'}})
+        response_data = self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        self.assertEqual(response_data.status_code, status.HTTP_404_NOT_FOUND)
+
     @ddt.data(
-        ({'category': {'a': 'a', 'b': 'b'}}, status.HTTP_400_BAD_REQUEST),
-        ({'category': {'id': 10000, 'name': 'Category Not Found'}}, status.HTTP_404_NOT_FOUND)
+        ('benefit_type', ['', 'Incorrect benefit type']),
+        ('benefit_value', ['', 'Incorrect benefit value', -1, 101]),
+        ('category', [{'a': 'a', 'b': 'b'}]),
     )
     @ddt.unpack
-    def test_create_coupon_product_invalid_category_data(self, category_data, expected_status_code):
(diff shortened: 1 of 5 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1017"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/offer/models.py
```diff
@@ -1,4 +1,5 @@
 # noinspection PyUnresolvedReferences
+import logging
```
vkaracic: Alphabetize this please. :)

ecommerce/extensions/offer/models.py
```diff
@@ -21,12 +22,23 @@ def save(self, *args, **kwargs):
...
 
     def clean_type(self):
         if self.type not in VALID_BENEFIT_TYPES:
+            logger.exception(
+                'Failed to create Benefit. Benefit type must be one of the following %s.', VALID_BENEFIT_TYPES
+            )
             raise ValidationError(_('Unrecognised benefit type {type}'.format(type=self.type)))
 
+    def clean_value(self):
+        if self.value < 0:
```
vkaracic, line 36: And what happens if it's > 100?

ecommerce/extensions/offer/models.py
```diff
@@ -21,12 +22,23 @@ def save(self, *args, **kwargs):
...
+            logger.exception(
+                'Failed to create Benefit. Benefit value may not be a negative number.'
```
vkaracic: This and the logger message above can fit in one line.

vkaracic (changes requested):

ecommerce/extensions/offer/models.py
iivic: Oscar validates this case if the type is Percentage oscar raises a ValidationError if the type is 'Absolute' nothing happens.

iivic:
@clintonb

vkaracic (approved):
LGTM

clintonb (approved):

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -262,21 +262,29 @@ def test_create_serializer_data(self):
         self.assertEqual(details_response['coupon_type'], 'Enrollment code')
         self.assertEqual(details_response['code_status'], 'ACTIVE')
 
-    @ddt.data(
-        ({'category': {'a': 'a', 'b': 'b'}}, status.HTTP_400_BAD_REQUEST),
```
marjev, line 266: Why is this removed? This should still fail and is there as something that can happen and should return at least 400 BAD REQUEST status or something better.

marjev (changes requested):

ecommerce/extensions/api/v2/tests/views/test_coupons.py
iivic: It's not removed I moved it to `test_create_coupon_product_invalid_data` test method.

marjev (approved):

PR #1019, opened 2016-11-16
Wrap attribute_cookie_data as atomic_transaction

bderusha:
This prevents errors in this method from rolling back all
basket-related transactions.

ECOM-6366

### DIFF
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -3,7 +3,8 @@ import json
 import mock
 
 import ddt
-from django.test import RequestFactory
+from django.db import transaction
+from django.test import RequestFactory, TransactionTestCase
 from oscar.core.loading import get_model
 from oscar.test.factories import BasketFactory, ProductFactory, RangeFactory, VoucherFactory
 import pytz
@@ -13,6 +14,7 @@ from ecommerce.core.tests import toggle_switch
 from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.extensions.basket.utils import prepare_basket, attribute_cookie_data
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.tests.mixins import UserMixin
 from ecommerce.extensions.partner.models import StockRecord
 from ecommerce.extensions.test.factories import prepare_voucher
 from ecommerce.referrals.models import Referral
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/basket/utils.py
```diff
@@ -91,18 +92,20 @@ def get_basket_switch_data(product):
...
-        _record_utm_basket_attribution(referral, request)
-
-        # Save the record if any attribution attributes are set on it.
-        if any([getattr(referral, attribute) for attribute in Referral.ATTRIBUTION_ATTRIBUTES]):
-            referral.save()
-        # Clean up the record if no attribution attributes are set and it exists in the DB.
-        elif referral.pk:
-            referral.delete()
-        # Otherwise we can ignore the instantiated but unsaved referral
+        # Required to prevent rolling back other atomic commits if exception is found
```
rlucioni, line 95: > Required to prevent rolling back the outer transaction if an exception is raised. This method may be called within a view protected by ATOMIC_REQUESTS.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -273,3 +275,25 @@ def test_attribute_cookie_data_multiple_cookies(self):
...
+    def setUp(self):
+        super(BasketUtilsTranactionTests, self).setUp()
+        self.request = RequestFactory()
+        self.request.COOKIES = {}
+        self.request.user = self.create_user()
+        site_configuration = SiteConfigurationFactory(partner__name='Tester')
+        site_configuration.utm_cookie_name = 'test.edx.utm'
+        self.request.site = site_configuration.site
+
+    def test_attribution_atomic_transation(self):
```
rlucioni, line 290: s/transation/transaction

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -273,3 +275,25 @@ def test_attribute_cookie_data_multiple_cookies(self):
...
+        with transaction.atomic():
+            basket = BasketFactory(owner=self.request.user, site=self.request.site)
+            StockRecord.objects.create(partner_sku="TEST123", product_id=1, partner_id=1)
+            with mock.patch('ecommerce.extensions.basket.utils._referral_from_basket_site') as mock_get_referral:
+                mock_get_referral.side_effect = AttributeError()
+                attribute_cookie_data(basket, self.request)
```
rlucioni, line 296: Can you confirm that `_referral_from_basket_site` is actually being mocked as you expect it to be? That could be a reason that this doesn't fail as you expect it to without your fix.

schenedx: Why mock `_referral_from_basket_site`? I believe the production issue fails on `referral.save()` method. Shouldn't we tailor the test case to fail on `referral.save()`?

rlucioni: @schenedx you're right about where the failure occurs in production, but in the context of this test, an exception raised anywhere in the atomic block should be treated the same.

bderusha: @rlucioni confirmed.  AttributeError is raised and the exception is caught.

@schenedx in this case we are testing any exception case so I had this blow up on the first method called to avoid any side effects or assumptions about how we eventually call `.save()`.  I suppose it also speeds up the test case (but not in any real way)

schenedx: Then can we write another test case to mock a Integrity error exception during `referral.save()`? I think there is value to understand even in that case, the code is working as designed

clintonb:

clintonb:
Closed due to inactivity.

PR #1021, opened 2016-11-17
Add management command for adding site to referrals without sites

bderusha:
@schenedx please review


ecommerce/referrals/management/commands/add_site_to_referrals.py
```diff
@@ -0,0 +1,44 @@
+""" Adds a Site to Referrals that do not already have one. """
```
schenedx: `Adds a Site to Referrals that have it set to null`

schenedx (approved):
👍

PR #1022, opened 2016-11-17
This prevents errors in attribute_cookie_data from rolling back all basket-related transactions.

schenedx:
@edx/ecommerce Please review
This is to fix the issue in ECOM-6366 for the long term

### DIFF
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -3,7 +3,8 @@ import json
 import mock
 
 import ddt
-from django.test import RequestFactory
+from django.db import transaction
+from django.test import RequestFactory, TransactionTestCase
 from oscar.core.loading import get_model
 from oscar.test.factories import BasketFactory, ProductFactory, RangeFactory, VoucherFactory
 import pytz
@@ -13,6 +14,7 @@ from ecommerce.core.tests import toggle_switch
 from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.extensions.basket.utils import prepare_basket, attribute_cookie_data
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.tests.mixins import UserMixin
 from ecommerce.extensions.partner.models import StockRecord
 from ecommerce.extensions.test.factories import prepare_voucher
 from ecommerce.referrals.models import Referral
@@ -120,12 +122,18 @@ class BasketUtilsTests(CourseCatalogTestMixin, TestCase):
             mock_attr_method.assert_called_with(basket, self.request)
 
     def test_attribute_cookie_data_affiliate_cookie_lifecycle(self):
-        """ Verify a basket is returned and referral captured. """
-        affiliate_id = 'test_affiliate'
-        self.request.COOKIES['affiliate_id'] = affiliate_id
+        """ Verify a basket is returned and referral captured if there is cookie info """
+
+        # If there is no cookie info, verify no referral is created.
         basket = BasketFactory(owner=self.request.user, site=self.request.site)
         attribute_cookie_data(basket, self.request)
+        with self.assertRaises(Referral.DoesNotExist):
+            Referral.objects.get(basket=basket)
 
+        # If there is cookie info, verify a referral is captured
+        affiliate_id = 'test_affiliate'
+        self.request.COOKIES['affiliate_id'] = affiliate_id
+        attribute_cookie_data(basket, self.request)
         # test affiliate id from cookie saved in referral
         referral = Referral.objects.get(basket_id=basket.id)
         self.assertEqual(referral.affiliate_id, affiliate_id)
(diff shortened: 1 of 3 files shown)


ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -120,12 +122,19 @@ def test_prepare_basket_calls_attribution_method(self):
             mock_attr_method.assert_called_with(basket, self.request)
 
     def test_attribute_cookie_data_affiliate_cookie_lifecycle(self):
-        """ Verify a basket is returned and referral captured. """
-        affiliate_id = 'test_affiliate'
-        self.request.COOKIES['affiliate_id'] = affiliate_id
+        """ Verify a basket is returned and referral captured if there are cookie info """
```
rlucioni: "if there is no cookie info"

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -120,12 +122,19 @@ def test_prepare_basket_calls_attribution_method(self):
...
+
+        # If there are no cookie info, no referral captured
```
rlucioni: "If there is no cookie info, verify no referral is created."

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -120,12 +122,19 @@ def test_prepare_basket_calls_attribution_method(self):
...
         basket = BasketFactory(owner=self.request.user, site=self.request.site)
         attribute_cookie_data(basket, self.request)
+        # Make sure no referral is created
```
rlucioni: This comment adds no value.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -120,12 +122,19 @@ def test_prepare_basket_calls_attribution_method(self):
...
+        with self.assertRaises(Referral.DoesNotExist):
+            Referral.objects.get(basket=basket)
 
+        # If there are cookie info, the referral is captured
```
rlucioni: "If there is cookie info, verify a referral is created."

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -273,3 +282,56 @@ def test_attribute_cookie_data_multiple_cookies(self):
...
+        }
+
+        affiliate_id = 'affiliate'
+
+        self.request.COOKIES['test.edx.utm'] = json.dumps(utm_cookie)
+        self.request.COOKIES['affiliate_id'] = affiliate_id
+
+    def test_attribution_atomic_transation(self):
+        """
+            Verify db error raised from attributing cookie to basket
```
rlucioni: Nitpick: let's try to keep our docstrings consistent, please. We don't typically indent like this.

``` py
"""
Verify that an IntegrityError raised while creating a referral
does not prevent a basket from being created.
"""
```

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -273,3 +282,56 @@ def test_attribute_cookie_data_multiple_cookies(self):
...
+        self._setup_request_cookie()
+        product = ProductFactory()
+        existing_basket = Basket.get_basket(self.request.user, self.request.site)
+        existing_referral = Referral(basket=existing_basket, site=self.request.site)
+        # Let's save an existing referral object to force the duplication happen in database
+        existing_referral.save()
+
+        with transaction.atomic():
+            with mock.patch('ecommerce.extensions.basket.utils._referral_from_basket_site') as mock_get_referral:
+                # Mock to return a duplicated referral object, so when saved, a DB integrity error is raised
```
rlucioni, line 326: Please add a comment about how mocking an exception as a side effect doesn't trigger a rollback. That's valuable knowledge we don't want to lose.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -273,3 +282,56 @@ def test_attribute_cookie_data_multiple_cookies(self):
         # test referral record is deleted when no cookies are set
         with self.assertRaises(Referral.DoesNotExist):
             Referral.objects.get(basket_id=basket.id)
+
+
+class BasketUtilsTranactionTests(UserMixin, TransactionTestCase):
```
rlucioni: BasketUtilsTransactionTests

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -273,3 +282,56 @@ def test_attribute_cookie_data_multiple_cookies(self):
...
+            'utm_content': utm_content,
+            'created_at': utm_created_at,
+        }
+
+        affiliate_id = 'affiliate'
+
+        self.request.COOKIES['test.edx.utm'] = json.dumps(utm_cookie)
+        self.request.COOKIES['affiliate_id'] = affiliate_id
+
+    def test_attribution_atomic_transation(self):
```
rlucioni: test_attribution_atomic_transaction

ecommerce/extensions/basket/utils.py
```diff
@@ -91,18 +92,20 @@ def get_basket_switch_data(product):
 
 def attribute_cookie_data(basket, request):
     try:
-        referral = _referral_from_basket_site(basket, request.site)
+        with transaction.atomic():
+            # Required to prevent rolling back other atomic commits if exception is found
```
rlucioni: You can add a little more detail here:

``` py
# If an exception is raised below, this nested atomic block prevents the
# outer transaction created by ATOMIC_REQUESTS from being rolled back.
```

rlucioni (changes requested):

requirements/test.txt
```diff
@@ -2,7 +2,7 @@
 -r base.txt
 
 bok-choy==0.4.7
-coverage==4.0.2
+coverage==4.2
```
MatthewPiatetsky, line 5: is there a reason we're not specifying the patch version here?

rlucioni: @MatthewPiatetsky the coverage package doesn't specify one: https://pypi.python.org/pypi/coverage.

ecommerce/extensions/basket/utils.py
```diff
@@ -91,18 +92,20 @@ def get_basket_switch_data(product):
...
+            # Save the record if any attribution attributes are set on it.
+            if any([getattr(referral, attribute) for attribute in Referral.ATTRIBUTION_ATTRIBUTES]):
+                referral.save()
+            # Clean up the record if no attribution attributes are set and it exists in the DB.
+            elif referral.pk:
+                referral.delete()
+            # Otherwise we can ignore the instantiated but unsaved referral
 
     # Don't let attribution errors prevent users from creating baskets
     except:  # pylint: disable=broad-except, bare-except
```
MatthewPiatetsky, line 112: Do we want to differentiate an IntegrityError due to the atomic condition vs other unrelated errors?

rlucioni: Any exceptions raised here will trigger a rollback of the inner block.

ecommerce/extensions/basket/tests/test_utils.py
schenedx: @rlucioni should be `if there is cookie info`. We should not have referral captured if there is no cookie info

schenedx:
@rlucioni Could you please review again?

ecommerce/extensions/basket/utils.py
bderusha: @MatthewPiatetsky I don't think we do.  The point is to log, but not raise any issues that happen during attribution

bderusha (approved):
👍

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -273,3 +281,60 @@ def test_attribute_cookie_data_multiple_cookies(self):
...
+        # Let's save an existing referral object to force the duplication happen in database
+        existing_referral.save()
+
+        with transaction.atomic():
+            with mock.patch('ecommerce.extensions.basket.utils._referral_from_basket_site') as mock_get_referral:
+                # Mock to return a duplicated referral object, so when saved, a DB integrity error is raised
+                # Mocking to raise random exception will not roll back the transaction
+                # Django db would not be aware of the mocked exception,
+                # since we actually would handle the exception in the attribute_cookie_data method.
+                # This would result in the transaction being committed
```
rlucioni: Thanks for adding this. However, are we sure that what you've written is entirely true? We don't want to mislead future maintainers with false assumptions. We know that using a mock side effect to raise an exception isn't enough to trigger a rollback, and we know that forcing a genuine IntegrityError is. Unless we know why that is, I wouldn't go any further here.

rlucioni (commented):
@schenedx looks good, just one last small thing.

schenedx:
@rlucioni Took in your last feedback. Can you review and comment?

rlucioni (approved):

PR #1023, opened 2016-11-18
Updated payment processors

clintonb:
The creation/modification of Source and PaymentEvent objects has been moved to the locations in the code base responsible for triggering the processing of payment processor responses and issuing credits. This move ensures that payment processors are responsible solely for processing payments. All administrative work will be handled outside of the payment processor classes.

SOL-2125

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,4 @@ override.cfg
 private.py
 docs/_build
 *.backup
+*.log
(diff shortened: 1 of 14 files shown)


clintonb:
@edx/ecommerce @mjfrey @vkaracic @iivic @marjev please review

mjfrey (approved):
Great.  This nicely reduces the work put on the processor code.  Also removes duplicate code. Would also be nice to break out the payment processor specific code out of the payment/views.py.  Having all the processor views in a single file is getting a bit un-managable.

clintonb:
@rlucioni please review

rlucioni (approved):
Nice.

PR #1024, opened 2016-11-18
Updated payment processor switch signal receiver

clintonb:
- Using `cache.delete` rather than `caches['default']`
- Added tests


clintonb:
@edx/ecommerce please review

clintonb:
FYI @jibsheet

rlucioni (changes requested):
@clintonb these switches don't appear to actually be used on the newer checkout page hosted by Otto. Even if we were to update the page so that these switches are respected, I don't think this caching makes much sense anymore given Waffle's caching of the switch values.

rlucioni (approved):
However, 👍 for fixing the code itself, even if it's dead. Could you at least add a TODO comment so that our conversation here isn't lost? You could even link to this PR.

PR #1025, opened 2016-11-20
[SOL-1955] Add interstitial for CyberSource payments

marjev:
@clintonb Here's a separate PR for Interstitial CyberSource View. We can continue discussion here.

Related story: https://openedx.atlassian.net/browse/SOL-1955

FYI
@mjfrey @vkaracic

### DIFF
--- a/ecommerce/extensions/payment/tests/views/test_cybersource.py
+++ b/ecommerce/extensions/payment/tests/views/test_cybersource.py
@@ -7,284 +7,33 @@ import ddt
 import mock
 from django.conf import settings
 from django.core.urlresolvers import reverse
-from factory.django import mute_signals
 from freezegun import freeze_time
-from oscar.apps.order.exceptions import UnableToPlaceOrder
 from oscar.apps.payment.exceptions import PaymentError, UserCancelled, TransactionDeclined
 from oscar.core.loading import get_class, get_model
 from oscar.test import factories
-from oscar.test.contextmanagers import mock_signal_receiver
-from testfixtures import LogCapture
 
-from ecommerce.extensions.fulfillment.status import ORDER
-from ecommerce.extensions.payment.processors.cybersource import Cybersource
-from ecommerce.extensions.payment.tests.mixins import PaymentEventsMixin, CybersourceMixin
-from ecommerce.extensions.payment.views.cybersource import CybersourceNotifyView
+from ecommerce.extensions.payment.exceptions import InvalidSignatureError, InvalidBasketError
+from ecommerce.extensions.payment.tests.mixins import CybersourceMixin, CybersourceNotificationTestsMixin
+from ecommerce.extensions.payment.views.cybersource import CybersourceInterstitialView, CybersourceNotifyView
 from ecommerce.tests.testcases import TestCase
 
 JSON = 'application/json'
 
 Basket = get_model('basket', 'Basket')
 Order = get_model('order', 'Order')
+OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
 PaymentEvent = get_model('order', 'PaymentEvent')
-PaymentEventType = get_model('order', 'PaymentEventType')
 PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
 Selector = get_class('partner.strategy', 'Selector')
-SourceType = get_model('payment', 'SourceType')
 
 post_checkout = get_class('checkout.signals', 'post_checkout')
 
 
 @ddt.ddt
-class CybersourceNotifyViewTests(CybersourceMixin, PaymentEventsMixin, TestCase):
+class CybersourceNotifyViewTests(CybersourceNotificationTestsMixin, TestCase):
     """ Test processing of CyberSource notifications. """
-
-    def setUp(self):
-        super(CybersourceNotifyViewTests, self).setUp()
-
-        self.toggle_ecommerce_receipt_page(True)
-
-        self.user = factories.UserFactory()
-        self.billing_address = self.make_billing_address()
-
-        self.basket = factories.create_basket()
-        self.basket.owner = self.user
-        self.basket.freeze()
-
-        self.processor = Cybersource(self.site)
-        self.processor_name = self.processor.NAME
-
-    def _assert_payment_data_recorded(self, notification):
-        """ Ensure PaymentEvent, PaymentProcessorResponse, and Source objects are created for the basket. """
-
-        # Ensure the response is stored in the database
-        self.assert_processor_response_recorded(self.processor_name, notification['transaction_id'], notification,
-                                                basket=self.basket)
-
-        # Validate a payment Source was created
-        reference = notification['transaction_id']
-        source_type = SourceType.objects.get(code=self.processor_name)
-        label = notification['req_card_number']
-        self.assert_payment_source_exists(self.basket, source_type, reference, label)
-
-        # Validate that PaymentEvents exist
-        paid_type = PaymentEventType.objects.get(code='paid')
-        self.assert_payment_event_exists(self.basket, paid_type, reference, self.processor_name)
-
-    def _assert_processing_failure(self, notification, status_code, error_message, log_level='ERROR'):
-        """Verify that payment processing operations fail gracefully."""
-        logger_name = 'ecommerce.extensions.payment.views.cybersource'
-        with LogCapture(logger_name) as l:
-            response = self.client.post(reverse('cybersource_notify'), notification)
-
-            self.assertEqual(response.status_code, status_code)
-
-            self.assert_processor_response_recorded(
-                self.processor_name,
-                notification[u'transaction_id'],
-                notification,
-                basket=self.basket
-            )
-
-            l.check(
-                (
-                    logger_name,
-                    'INFO',
-                    'Received CyberSource merchant notification for transaction [{transaction_id}], '
-                    'associated with basket [{basket_id}].'.format(
-                        transaction_id=notification[u'transaction_id'],
-                        basket_id=self.basket.id
-                    )
-                ),
-                (logger_name, log_level, error_message)
-            )
-
-    # Disable the normal signal receivers so that we can verify the state of the created order.
-    @mute_signals(post_checkout)
-    def test_accepted(self):
-        """
-        When payment is accepted, the following should occur:
-            1. The response is recorded and PaymentEvent/Source objects created.
-            2. An order for the corresponding basket is created.
-            3. The order is fulfilled.
-        """
-
-        # The basket should not have an associated order if no payment was made.
-        self.assertFalse(Order.objects.filter(basket=self.basket).exists())
-
-        notification = self.generate_notification(self.basket, billing_address=self.billing_address)
-
-        with mock_signal_receiver(post_checkout) as receiver:
-            response = self.client.post(reverse('cybersource_notify'), notification)
-
-            # Validate that a new order exists in the correct state
-            order = Order.objects.get(basket=self.basket)
-            self.assertIsNotNone(order, 'No order was created for the basket after payment.')
-            self.assertEqual(order.status, ORDER.OPEN)
-
-            # Validate the order's line items
-            self.assertListEqual(list(order.lines.values_list('product__id', flat=True)),
-                                 list(self.basket.lines.values_list('product__id', flat=True)))
-
-            # Verify the post_checkout signal was emitted
-            self.assertEqual(receiver.call_count, 1)
-            __, kwargs = receiver.call_args
-            self.assertEqual(kwargs['order'], order)
-
-        # The view should always return 200
-        self.assertEqual(response.status_code, 200)
-
-        # Validate the payment data was recorded for auditing
-        self._assert_payment_data_recorded(notification)
-
-        # The basket should be marked as submitted. Refresh with data from the database.
-        basket = Basket.objects.get(id=self.basket.id)
-        self.assertTrue(basket.is_submitted)
-        self.assertIsNotNone(basket.date_submitted)
-
-    @ddt.data('CANCEL', 'DECLINE', 'ERROR', 'blah!')
-    def test_not_accepted(self, decision):
-        """
-        When payment is NOT accepted, the processor's response should be saved to the database. An order should NOT
-        be created.
-        """
-
-        notification = self.generate_notification(self.basket, decision=decision)
-        response = self.client.post(reverse('cybersource_notify'), notification)
-
-        # The view should always return 200
-        self.assertEqual(response.status_code, 200)
-
-        # The basket should not have an associated order if no payment was made.
-        self.assertFalse(Order.objects.filter(basket=self.basket).exists())
-
-        # Ensure the response is stored in the database
-        self.assert_processor_response_recorded(self.processor_name, notification[u'transaction_id'], notification,
-                                                basket=self.basket)
-
-    @ddt.data(
-        (PaymentError, 200, 'ERROR', 'CyberSource payment failed for basket [{basket_id}]. '
-                                     'The payment response was recorded in entry [{response_id}].'),
-        (UserCancelled, 200, 'INFO', 'CyberSource payment did not complete for basket [{basket_id}] because '
-                                     '[UserCancelled]. The payment response was recorded in entry [{response_id}].'),
-        (TransactionDeclined, 200, 'INFO', 'CyberSource payment did not complete for basket [{basket_id}] because '
-                                           '[TransactionDeclined]. The payment response was recorded in entry '
-                                           '[{response_id}].'),
-        (KeyError, 500, 'ERROR', 'Attempts to handle payment for basket [{basket_id}] failed.')
-    )
-    @ddt.unpack
-    def test_payment_handling_error(self, error_class, status_code, log_level, error_message):
-        """
-        Verify that CyberSource's merchant notification is saved to the database despite an error handling payment.
-        """
-        notification = self.generate_notification(
-            self.basket,
-            billing_address=self.billing_address,
-        )
-        with mock.patch.object(CybersourceNotifyView, 'handle_payment', side_effect=error_class) as fake_handle_payment:
-            self._assert_processing_failure(
-                notification,
-                status_code,
-                error_message.format(basket_id=self.basket.id, response_id=1),
-                log_level
-            )
-            self.assertTrue(fake_handle_payment.called)
-
-    def test_unable_to_place_order(self):
-        """ When payment is accepted, but an order cannot be placed, log an error and return HTTP 200. """
-
-        notification = self.generate_notification(
-            self.basket,
-            billing_address=self.billing_address,
-        )
-
-        # Verify that anticipated errors are handled gracefully.
-        with mock.patch.object(CybersourceNotifyView, 'handle_order_placement',
-                               side_effect=UnableToPlaceOrder) as fake_handle_order_placement:
-            error_message = 'Payment was received, but an order for basket [{basket_id}] could not be placed.'.format(
-                basket_id=self.basket.id,
-            )
-            self._assert_processing_failure(notification, 500, error_message)
-            self.assertTrue(fake_handle_order_placement.called)
-
-        # Verify that unanticipated errors are also handled gracefully.
-        with mock.patch.object(CybersourceNotifyView, 'handle_order_placement',
-                               side_effect=KeyError) as fake_handle_order_placement:
-            error_message = 'Payment was received, but an order for basket [{basket_id}] could not be placed.'.format(
-                basket_id=self.basket.id,
-            )
-            self._assert_processing_failure(notification, 500, error_message)
-            self.assertTrue(fake_handle_order_placement.called)
-
-    def test_invalid_basket(self):
-        """ When payment is accepted for a non-existent basket, log an error and record the response. """
-        order_number = '{}-{}'.format(self.partner.short_code.upper(), 101986)
-
-        notification = self.generate_notification(
-            self.basket,
-            billing_address=self.billing_address,
-            req_reference_number=order_number,
-        )
-        response = self.client.post(reverse('cybersource_notify'), notification)
-
-        self.assertEqual(response.status_code, 400)
-        self.assert_processor_response_recorded(self.processor_name, notification[u'transaction_id'], notification)
-
-    @ddt.data(('line2', 'foo'), ('state', 'bar'))
-    @ddt.unpack
-    @mock.patch('ecommerce.extensions.payment.views.cybersource.CybersourceNotifyView.handle_order_placement')
-    def test_optional_fields(self, field_name, field_value, mock_placement_handler):
-        """ Ensure notifications are handled properly with or without keys/values present for optional fields. """
-
-        def check_notification_address(notification, expected_address):
-            response = self.client.post(reverse('cybersource_notify'), notification)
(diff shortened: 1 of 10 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1025&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -182,6 +183,17 @@ def place_free_order(self, basket, request=None):
 
         return order
 
+    def redirect_to_receipt_page_on_success(self, request, receipt_page_url):
+        """ Adds data that indicates payment success to the request session object.
+        Arguments:
+            request(HttpRequest): Request made to a Payment Processor View.
+            receipt_page_url(str): Receipt Page URL
+        Returns:
+            response(HttpRedirectResponse): Redirect Response.
```
clintonb: Redirect to the specified receipt page URL.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -182,6 +183,17 @@ def place_free_order(self, basket, request=None):
 
         return order
 
+    def redirect_to_receipt_page_on_success(self, request, receipt_page_url):
+        """ Adds data that indicates payment success to the request session object.
+        Arguments:
+            request(HttpRequest): Request made to a Payment Processor View.
+            receipt_page_url(str): Receipt Page URL
```
clintonb: Lower case "page".

ecommerce/extensions/checkout/mixins.py
```diff
@@ -182,6 +183,17 @@ def place_free_order(self, basket, request=None):
 
         return order
 
+    def redirect_to_receipt_page_on_success(self, request, receipt_page_url):
+        """ Adds data that indicates payment success to the request session object.
+        Arguments:
```
clintonb: Add a newline before this.

ecommerce/extensions/payment/tests/test_views.py
```diff
@@ -725,3 +726,36 @@ def test_field_error(self):
 
         errors = json.loads(response.content)['field_errors']
         self.assertIn(field, errors)
+
+
+class CybersourceInterstitialViewTests(RefundTestMixin, TestCase):
```
clintonb: This view has nothing to do with refunds, so that mixin should not be included. It might be simpler to inherit `CybersourceNotifyViewTests`.

ecommerce/extensions/payment/views.py
```diff
@@ -272,6 +272,38 @@ def post(self, request):
             return HttpResponse(status=500)
 
 
+class CybersourceInterstitialView(CybersourceNotifyView, TemplateView):
+    """ Interstitial view for Cybersource Payments. """
+    template_name = 'checkout/cybersource_error.html'
+
+    def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
```
clintonb: You need to take advantage of the logic of `CybersourceNotifyView` here to actually process the response from CyberSource.

clintonb (changes requested):
I have updated the ticket to better describe the intent of these changes. Notably, you are not processing transaction responses in the new view.

ecommerce/extensions/payment/tests/test_views.py
marjev: I included that mixin because it had order creation method.
I'll create a new mixin for that purpose and make the necessary changes.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -182,6 +183,19 @@ def place_free_order(self, basket, request=None):
 
         return order
 
+    def redirect_to_receipt_page_on_success(self, request, receipt_page_url):
```
clintonb: This method is only used by `CybersourceInterstitialView`. Move it there.

ecommerce/extensions/offer/tests/mixins.py
```diff
@@ -0,0 +1,35 @@
...
+
+from oscar.test.factories import create_order
+from oscar.test.newfactories import BasketFactory
+
+from ecommerce.courses.models import Course
+from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.extensions.fulfillment.status import ORDER
+
+
+class OfferTestMixin(CourseCatalogTestMixin):
```
clintonb: This has nothing to do with offers, and does not belong here. Now that I see your use case, this mixin is not necessary. Oscar has an `OrderFactory`. Use it for your new tests and leave `RefundTestMixin` alone.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -414,3 +416,36 @@ def test_field_error(self):
...
+    """ Test interstitial view for Cybersource Payments. """
+
+    def setUp(self):
+        super(CybersourceInterstitialViewTests, self).setUp()
+        self.toggle_ecommerce_receipt_page(True)
+        self.user = factories.UserFactory()
+
+    def test_redirect_to_receipt_page_path(self):
+        """ Successful CyberSource Payments should be redirected to the receipt page. """
+        order = self.create_order()
```
clintonb: Use Oscar's `OrderFactory`.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -261,3 +263,35 @@ def post(self, request):
...
             logger.exception(self.order_placement_failure_msg, basket.id)
             return HttpResponse(status=500)
+
+
+class CybersourceInterstitialView(CybersourceNotifyView, TemplateView):
+    """ Interstitial view for Cybersource Payments. """
+    template_name = 'checkout/cybersource_error.html'
+
+    def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
+        # CyberSource responses will indicate whether a payment failed due to a transaction on their end. In this case,
```
clintonb: This implementation is still wrong. The purpose of subclassing `CybersourceNotifyView` is to take advantage of its ability to create orders. Let's pretend we got rid of the `/payment/cybersource/notify/` endpoint. Under that scenario I expect this view (`CybersourceInterstitialView`) at `/payment/cybersource/redirect/` to be able to process the CyberSource response. That's the end goal of the interstitial: do the work of `CybersourceNotifyView` and redirect the user to the receipt page.

clintonb (changes requested):

ecommerce/extensions/checkout/mixins.py
marjev: It should also be used in the PayPal scenarios, right?

ecommerce/extensions/payment/views/cybersource.py
marjev: I still haven't got to this. I'm working on fixing the tests and increasing coverage.

ecommerce/extensions/checkout/mixins.py
clintonb: No. This is only necessary for CyberSource. We already do something similar for PayPal, and we do not want to mix payment processing code.

marjev:
@clintonb, could you please take another look? I think I addressed all comments and I think that Interstitial View now has everything we need.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -414,3 +418,15 @@ def test_field_error(self):
...
+@ddt.ddt
+class CybersourceInterstitialViewTests(CybersourceNotifyViewTests, TestCase):
+    """ Test interstitial view for Cybersource Payments. """
+    path = reverse('cybersource_redirect')
+
+    def setUp(self):
+        super(CybersourceInterstitialViewTests, self).setUp()
+
+        self.go_to_receipt_page_status_code = 302
+        self.cybersource_processing_failure_status_code = 502
```
clintonb: This should not differ from the value returned by `CybersourceNotifyViewTests`.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -414,3 +418,15 @@ def test_field_error(self):
 
         errors = json.loads(response.content)['field_errors']
         self.assertIn(field, errors)
+
+
+@ddt.ddt
+class CybersourceInterstitialViewTests(CybersourceNotifyViewTests, TestCase):
```
clintonb: `CybersourceNotifyViewTests` already inherits `TestCase`.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -54,6 +56,9 @@ def setUp(self):
         self.processor = Cybersource(self.site)
         self.processor_name = self.processor.NAME
 
+        self.go_to_receipt_page_status_code = 200
```
clintonb: `CybersourceNotifyView` doesn't know/care about receipts. `successful_payment_status_code` is more appropriate.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -261,3 +264,136 @@ def post(self, request):
...
+        Returns:
+            response(HttpRedirectResponse): Redirect to the specified receipt page URL.
+        """
+        request.session['fire_tracking_events'] = True
+        return redirect(receipt_page_url)
+
+    def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
+        """Process a CyberSource merchant notification and place an order for paid products as appropriate."""
+
+        # Note (CCB): Orders should not be created until the payment processor has validated the response's signature.
```
clintonb: This code appears to be copied verbatim from `CybersourceNotifyView`. Don't copy code. Extract it to a mixin.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -261,3 +264,136 @@ def post(self, request):
         except:  # pylint: disable=bare-except
             logger.exception(self.order_placement_failure_msg, basket.id)
             return HttpResponse(status=500)
+
+
+class CybersourceInterstitialView(CybersourceNotifyView, TemplateView):
```
clintonb: Keep this view simple. It should only be redirecting. In the case of a successful payment, redirect to the receipt page. In the case of a failed payment, redirect to an error page. The error page doesn't need to be processor-specific (e.g. /payment/error/). Just remember to thaw the basket so the user can attempt payment again.
image

ecommerce/templates/oscar/checkout/cybersource_error.html
```diff
@@ -0,0 +1,37 @@
...
+{% block navbar %}
+    {% include 'edx/partials/_student_navbar.html' %}
+{% endblock navbar %}
+
+{% block content %}
+  <div class="container content-wrapper receipt-cancel-error">
+    <h1>{% trans "Payment Failed" %}</h1>
+    <p>
+      {% blocktrans %}
+        A system error occurred while processing your payment. <b>You have not been charged.</b>
```
clintonb: Use `<strong>` instead of `<b>`

ecommerce/tests/mixins.py
```diff
@@ -284,6 +284,12 @@ def mock_access_token_response(self, status=200):
 
         return token
 
+    def toggle_ecommerce_receipt_page(self, enable_otto_receipt_page):
```
clintonb, line 287: Given that this is only used in one place, I'm not sure how useful it is as a method.

ecommerce/extensions/payment/tests/views/test_cybersource.py
clintonb: Also, at no point are you testing the rendering of the error page.

clintonb (changes requested):
The shared code must be extracted to a mixin, not duplicated.

ecommerce/extensions/payment/views/cybersource.py
marjev: Is this ok:
```
def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
        """Process a CyberSource merchant notification and place an order for paid products as appropriate."""

        cybersource_response = request.POST.dict()
        order_number = cybersource_response.get('req_reference_number')

        if cybersource_response.get('decision') != 'ACCEPT':
            basket_id = OrderNumberGenerator().basket_id(order_number)
            basket = self._get_basket(basket_id)
            if basket:
                basket.thaw()
                redirect(reverse('checkout:error'))
        else:
            shipping_charge = shipping_method.calculate(basket)
            self.handle_order_placement(
                order_number,
                basket.owner,
                basket,
                None,
                NoShippingRequired(),
                shipping_charge,
                self._get_billing_address(cybersource_response),
                OrderTotalCalculator().calculate(basket, shipping_charge),
                request=request
            )
            receipt_page_url = get_receipt_page_url(
                order_number=order_number,
                site_configuration=self.request.site.siteconfiguration
            )
            self._redirect_to_receipt_page_on_success(request, receipt_page_url)
```

I assumed that txn is the same thing that's used on the LMS receipt page. If it's not, I need an explanation of what it stands for, how it's used and where we use it.

clintonb: Not quite. See https://github.com/edx/ecommerce/commits/clintonb/payment-explain.

ecommerce/tests/mixins.py
marjev: This is shared with #978

marjev:
@clintonb Can you please take another look?

ecommerce/extensions/checkout/utils.py
```diff
@@ -42,15 +43,15 @@ def get_receipt_page_url(site_configuration, order_number=None):
...
-        return site_configuration.build_ecommerce_url('{base_url}{order_number}'.format(
-            base_url=settings.RECEIPT_PAGE_PATH,
-            order_number=order_number if order_number else ''
-        ))
-    return site_configuration.build_lms_url(
-        '{base_url}{order_number}'.format(
-            base_url='/commerce/checkout/receipt',
-            order_number='?orderNum={}'.format(order_number) if order_number else ''
-        )
+        base_url = site_configuration.build_ecommerce_url(settings.RECEIPT_PAGE_PATH)
```
mjfrey: Do not use settings.RECEIPT_PAGE_PATH but rather make this a constant within the processor itself.  We can avoid changing settings files as well as have an endpoint per processor.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -256,8 +250,72 @@ def post(self, request):
...
+
+        try:
+            self.create_order(request, basket, notification)
             return HttpResponse()
         except:  # pylint: disable=bare-except
-            logger.exception(self.order_placement_failure_msg, basket.id)
             return HttpResponse(status=500)
+
+
+class CybersourceInterstitialView(CybersourceNotificationMixin, View):
```
mjfrey, line 289: you are missing @method_decorator(csrf_exempt)
This throws an error when cybersource tries to POST back.

mjfrey (changes requested):

ecommerce/extensions/checkout/utils.py
clintonb: Agreed. This PR shouldn't deal with receipts at all aside from redirecting the user there. I understand this was extracted from the receipt page PR; but, let's keep the two separated to avoid confusion and maintain independence.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -162,35 +165,20 @@ def test_not_accepted(self, decision):
         self.assert_processor_response_recorded(self.processor_name, notification[u'transaction_id'], notification,
                                                 basket=self.basket)
 
-    @ddt.data(
```
clintonb, line 165: Why was all of this removed?

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -162,35 +165,20 @@ def test_not_accepted(self, decision):
...
-                status_code,
-                error_message.format(basket_id=self.basket.id, response_id=1),
-                log_level
-            )
-            self.assertTrue(fake_handle_payment.called)
+        with mock.patch.object(CybersourceNotifyView, 'handle_payment', side_effect=KeyError):
+            response = self.client.post(self.path, notification)
+            self.assertEqual(response.status_code, 500)
 
-    def test_unable_to_place_order(self):
```
clintonb, line 193: Why is this change necessary?

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -414,3 +396,74 @@ def test_field_error(self):
 
         errors = json.loads(response.content)['field_errors']
         self.assertIn(field, errors)
+
+
+@ddt.ddt
+class CybersourceInterstitialViewTests(CybersourceMixin, TestCase):
```
clintonb: Similar to the view being tested, you should extract the appropriate tests from `CybersourceNotifyViewTests` to a mixin shared with `CybersourceInterstitialViewTests` in order to fully test the view.

ecommerce/extensions/payment/urls.py
```diff
@@ -1,12 +1,14 @@
...
 from django.conf.urls import url
 
-from ecommerce.extensions.payment.views.cybersource import CybersourceNotifyView, CybersourceSubmitView
+from ecommerce.extensions.payment.views import cybersource, PaymentFailedView
 from ecommerce.extensions.payment.views.paypal import PaypalPaymentExecutionView, PaypalProfileAdminView
 
 urlpatterns = [
-    url(r'^cybersource/notify/$', CybersourceNotifyView.as_view(), name='cybersource_notify'),
-    url(r'^cybersource/submit/$', CybersourceSubmitView.as_view(), name='cybersource_submit'),
+    url(r'^payment/error/$', PaymentFailedView.as_view(), name='payment_error'),
```
clintonb: Keep these alphabetized. See `lms/urls.py` on edx/edx-platform to understand my desire for order.

clintonb: Also, no need for the payment prefix on the path. That's already taken care of.

ecommerce/extensions/payment/views/cybersource.py
clintonb: This can be resolved by moving the `dispatch` method to `CybersourceNotificationMixin`.

ecommerce/templates/oscar/checkout/payment_error.html
```diff
@@ -0,0 +1,37 @@
...
+{% block navbar %}
+    {% include 'edx/partials/_student_navbar.html' %}
+{% endblock navbar %}
+
+{% block content %}
+  <div class="container content-wrapper receipt-cancel-error">
+    <h1>{% trans "Payment Failed" %}</h1>
+    <p>
+      {% blocktrans %}
+        A system error occurred while processing your payment. <b>You have not been charged.</b>
```
clintonb: Use `<strong>` instead of `<b>`. See http://stackoverflow.com/a/271776/592820.

ecommerce/templates/oscar/checkout/payment_error.html
```diff
@@ -0,0 +1,37 @@
...
+    <p>
+      {% with "<a class='nav-link' href='mailto:"|add:payment_support_email|add:"'>"|safe as start_link %}
+        {% blocktrans with end_link="</a>"|safe %}
+          Please wait a few minutes and then try again. For help, contact {{ start_link }}{{ payment_support_email }}{{ end_link }}.
+        {% endblocktrans %}
+      {% endwith %}
+    </p>
+
+    <p>
+      {% with "<a class='nav-link' href='|add:basket_url|add:'>"|safe as start_link %}
```
clintonb, line 30: `basket_url` is not defined in the view...which is not tested.

ecommerce/extensions/checkout/utils.py
marjev: I moved this into the Receipt Page PR: https://github.com/edx/ecommerce/pull/978/commits/ca2ab7f1569b3f3979011e9f1ae6c41f7c12020b

marjev:
@mjfrey @clintonb Comments addressed.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -7,284 +7,35 @@
...
 PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
 Selector = get_class('partner.strategy', 'Selector')
-SourceType = get_model('payment', 'SourceType')
 
 post_checkout = get_class('checkout.signals', 'post_checkout')
 
 
 @ddt.ddt
-class CybersourceNotifyViewTests(CybersourceMixin, PaymentEventsMixin, TestCase):
+class CybersourceNotifyViewTests(CybersourceNotificationTestsMixin, PaymentEventsMixin, TestCase):
```
clintonb: Update `CybersourceMixin` to inherit `PaymentEventsMixin` and remove `PaymentEventsMixin` from here and the class below.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -414,3 +165,62 @@ def test_field_error(self):
...
+
+    @ddt.data(InvalidSignatureError, InvalidBasketError, KeyError)
+    def test_invalid_payment_error(self, error_class):
+        """ Verify the view redirects to Payment error page when a Payment error occurred. """
+        notification = self.generate_notification(
+            self.basket,
+            billing_address=self.billing_address,
+        )
+        with mock.patch.object(CybersourceInterstitialView, 'validate_notification', side_effect=error_class):
+            response = self.client.post(self.path, notification, follow=True)
```
clintonb: Don't follow. Verify the the location of the redirect with `self.assertRedirects()`.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -414,3 +165,62 @@ def test_field_error(self):
...
+            self.assertEqual(response.status_code, 200)
+            self.assertDictContainsSubset(
+                {
+                    'basket_url': reverse('basket:summary'),
+                    'payment_support_email': self.request.site.siteconfiguration.payment_support_email
+                },
+                dict(response.context)
```
clintonb: Is the casting actually necessary? If not, get rid of it.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -414,3 +165,62 @@ def test_field_error(self):
...
+            self.basket,
+            billing_address=self.billing_address,
+        )
+        with mock.patch.object(CybersourceInterstitialView, 'validate_notification', side_effect=error_class):
+            response = self.client.post(self.path, notification, follow=True)
+            self.assertEqual(response.status_code, 200)
+            self.assertDictContainsSubset(
+                {
+                    'basket_url': reverse('basket:summary'),
+                    'payment_support_email': self.request.site.siteconfiguration.payment_support_email
```
clintonb: There is no need for `self.request`. Use `self.siteconfiguration.payment_support_email`.

clintonb: Where is `self.request` coming from? We shouldn't have this if it's not used.

clintonb (approved):

marjev:
@mjfrey anything else needed in this PR?

mjfrey (approved):

PR #1026, opened 2016-11-21
Test ecommerce via the container built in configuration.

feanil:
This allows us to further test that the build on the correct version of the correct OS.

### DIFF
--- a/.travis/docker-compose-travis.yml
+++ b/.travis/docker-compose-travis.yml
@@ -1,0 +1,16 @@
+# This is only meant to be used for testing on travis-ci.org
+version: "2"
+
+services:
+  ecommerce:
+    image: edxops/ecommerce:latest
+
+    container_name: ecommerce_testing
+    volumes:
+      - ..:/edx/app/ecommerce/ecommerce
+      - $HOME/.cache/pip:/edx/app/ecommerce/.cache/pip
+    # The docker container produced from configuration.git does not currently
+    # run and expose a port for ecommerce.  This means that we need to run some
+    # command that keeps the ecommerce container alive while we run tests on it.
+    # We have not yet standardized on an init replacement which could be used instead.
+    command: tail -f /dev/null
--- a/.travis/run_tests.sh
+++ b/.travis/run_tests.sh
@@ -23,7 +23,7 @@ export DJANGO_SETTINGS_MODULE=ecommerce.settings.test
 make validate_translations
 
 # Compile assets and run validation
-xvfb-run make clean_static
-xvfb-run make static
+make clean_static
+make static
 xvfb-run make validate_python
 xvfb-run make validate_js
--- a/.travis/run_tests.sh
+++ b/.travis/run_tests.sh
@@ -1,0 +1,30 @@
+#!/bin/bash -xe
+. /edx/app/ecommerce/venvs/ecommerce/bin/activate
+. /edx/app/ecommerce/nodeenvs/ecommerce/bin/activate
+
+apt update
+apt install -y xvfb firefox gettext
+
+cd /edx/app/ecommerce/ecommerce
+
+pip install -U pip wheel
+# Make it so bower can run without sudo.
+# https://github.com/GeoNode/geonode/pull/1070
+echo '{ "allow_root": true }' > /root/.bowerrc
+
+make requirements
+
+# Ensure documentation can be compiled
+cd docs && make html
+cd ..
+
+export DJANGO_SETTINGS_MODULE=ecommerce.settings.test
+
+# Check if translation files are up-to-date
+make validate_translations
+
+# Compile assets and run validation
+xvfb-run make clean_static
+xvfb-run make static
+xvfb-run make validate_python
+xvfb-run make validate_js
--- a/.travis/run_tests.sh
+++ b/.travis/run_tests.sh
@@ -7,7 +7,6 @@ apt install -y xvfb firefox gettext
 
 cd /edx/app/ecommerce/ecommerce
 
-pip install -U pip wheel
 # Make it so bower can run without sudo.
 # https://github.com/GeoNode/geonode/pull/1070
 echo '{ "allow_root": true }' > /root/.bowerrc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,40 +1,30 @@
 language: python
 python:
     - "2.7"
-sudo: false
 
-# Cache the pip directory. "cache: pip" doesn't work due to install override. See https://github.com/travis-ci/travis-ci/issues/3239.
+sudo: required
+services:
+  - docker
+
 cache:
-  - bundler
+  - pip
   - directories:
-    - $HOME/.cache/pip
     - node_modules
     - ecommerce/static/bower_components
+
 addons:
     firefox: latest
     apt:
         packages:
             - lcov
+
 before_install:
-    - "export DISPLAY=:99.0"
-    - "sh -e /etc/init.d/xvfb start"
-    - "export DJANGO_SETTINGS_MODULE=ecommerce.settings.test"
-install:
-    - pip install -U codecov
-    - pip install -U pip wheel
-    - make requirements
+    - docker-compose -f ./.travis/docker-compose-travis.yml up -d
+
 script:
-    # Ensure documentation can be compiled
-    - cd docs && make html
-    - cd ..
+    - docker exec ecommerce_testing /edx/app/ecommerce/ecommerce/.travis/run_tests.sh
 
-    # Check if translation files are up-to-date
-    - make validate_translations
-
-    # Compile assets and run validation
-    - make clean_static
-    - make static
-    - make validate_python
-    - make validate_js
 after_success:
+    - pip install -U codecov
+    - docker exec ecommerce_testing /edx/app/ecommerce/ecommerce/.travis/run_coverage.sh
     - codecov
--- a/.travis/run_coverage.sh
+++ b/.travis/run_coverage.sh
@@ -1,0 +1,4 @@
+#!/bin/bash -xe
+. /edx/app/ecommerce/venvs/ecommerce/bin/activate
+cd /edx/app/ecommerce/ecommerce
+coverage xml


feanil:
@rlucioni @clintonb @mjfrey 

FYI @jibsheet

.travis/docker-compose-travis.yml
```diff
@@ -0,0 +1,16 @@
+# This is only meant to be used for testing on travis-ci.org
```
rlucioni, line 1: I'd like to keep the structure of these files consistent across our repos. In programs, we have a `docker-compose-travis.yml` at the root of the repo. Here, the file lives in a `.travis` directory. Either is fine with me as long as we're consistent. Looks like a `.travis` directory was introduced to programs as well, so it could just be a matter of moving the `docker-compose-travis.yml` file on that end.

Related: these files should be added to https://github.com/edx/cookiecutter-django-ida/.

.travis/docker-compose-travis.yml
```diff
@@ -0,0 +1,16 @@
...
+
+    container_name: ecommerce_testing
+    volumes:
+      - ..:/edx/app/ecommerce/ecommerce
+      - $HOME/.cache/pip:/edx/app/ecommerce/.cache/pip
+    # The docker container produced from configuration.git does not currently
+    # run and expose a port for ecommerce.  This means that we need to run some
+    # command that keeps the ecommerce container alive while we run tests on it.
+    # We have not yet standardized on an init replacement which could be used instead.
+    command: tail -f /dev/null
```
rlucioni, line 16: Thoughts on using https://github.com/krallin/tini for this?

.travis/docker-compose-travis.yml
feanil: Agreed, @jibsheet thoughts on the location of this file.  I felt that since the other travis files were hidden, this should be as well.

jibsheet: I think moving programs into .travis/ is a good idea. Docker compose was a late breaking change so I didn't think about location as much for that file (other than trying to make it clear that it was travis specific in the name).

.travis/docker-compose-travis.yml
jibsheet: There are several competing "tiny initd" projects.  I like tini, but other folks have other opinions.  This is why i had settled on the hack for this project.  If Feanil wants to tackle switching, great, but that's the reason I hadn't done tini last time.

feanil: Seems fine for when we actually care what process is running in the container.  Thinking about it a bit more, it seems like even if we did have an init replacement, we would still run the tail here to keep the container running.

.travis/run_tests.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash -xe
+. /edx/app/ecommerce/venvs/ecommerce/bin/activate
+. /edx/app/ecommerce/nodeenvs/ecommerce/bin/activate
+
+apt update
+apt install -y xvfb firefox gettext
```
clintonb, line 6: We should install this as part of image creation. `gettext` is definitely needed for developing translations. `firefox` is needed for local runs of Selenium.

.travis/run_tests.sh
```diff
@@ -0,0 +1,30 @@
...
+
+cd /edx/app/ecommerce/ecommerce
+
+pip install -U pip wheel
```
clintonb: It would be nice to install these in the image.

.travis/run_tests.sh
```diff
@@ -0,0 +1,30 @@
...
+# Make it so bower can run without sudo.
+# https://github.com/GeoNode/geonode/pull/1070
+echo '{ "allow_root": true }' > /root/.bowerrc
```
clintonb, line 12: This would also be useful in the image.

.travis/run_tests.sh
feanil: The image is created from the default settings which are the production settings(no dev requirements either).  I don't think it makes sense to install these in production which is why I put them here.  Are you suggesting making this a part of the dockerfile definiton?  @jibsheet  @e0d thoughts on this.  I think that since the Dockerfiles are meant for development, this is fine but it will mean one more thing to deal with if we want to separate out test/dev settings from production settings.  One option may be to add to the ansible-overrides.yml for ecommerce to add these packages.

clintonb: Correct, I'm advocating for putting these in `Dockerfile`.

.travis/run_tests.sh
feanil: Waiting on responses to the question [above](https://github.com/edx/ecommerce/pull/1026#discussion_r89206094) which will likely inform how we deal with this as well.

.travis/run_tests.sh
feanil: Waiting on responses to the question [above](https://github.com/edx/ecommerce/pull/1026#discussion_r89206094) which will likely inform how we deal with this as well.

.travis/run_tests.sh
jibsheet: I admit I do not have strong feelings (obviously I installed gettext into the programs Dockerfile) although much of me wants to avoid having dev tools installed on production images.

If we can do it with ansible-overrides.yml I would be +1 on that, and would swap programs around.

It'll lead to some sort of bug later where we don't install something in prod, but I think it's a good compromise between dev utility and prod security.

.travis/run_tests.sh
feanil: I think adding this to the Dockerfile directly makes sense since it's specific to running bower in docker.

.travis/run_tests.sh
feanil: I believe pip is already installed as a part of the standard requirements, this bit was sort of cargo-culted, I think it makes more sense to just use the version that is in the container, objections to just removing this line?

.travis/docker-compose-travis.yml
feanil: Per this discussion: https://github.com/edx/programs/pull/179

.travis/run_tests.sh
clintonb: `pip` is installed; but, it's usually an older version (7.x) instead of the latest (9.x). I haven't noticed any performance improvements; but, `pip` will show an upgrade warning every time it is invoked. `wheel` definitely leads to decreased install times, and should be included in the image.

.travis/run_tests.sh
e0d: Is the correct thing to do here to extend the container via a test specific Dockerfile adding testing related tools?  We could either pre-build that "enhanced" container or do it on the fly I would think.

clintonb: > test specific

Isn't everything we are doing test-specific? Are there plans to start using Docker in production?

feanil: @clintonb If we start by keeping test and potentially production separate, we maintain the flexibility of being able to use docker for production in the future, pretty cheaply.  It's generally much harder to separate test and prod after the fact if we do decide we want to use docker in production.

@e0d my thought was that for adding a few more apt packages, we can just do it via the current overrides file which I understood to be test specific ansible overrides.

.travis/run_tests.sh
feanil: Thinking more about this, the correct way to do this would be to actualy upgard the pip in the container that is installed via ansible.  As for `wheel`, why no install this as a part of 'make requirements' which is dev specific anyway?

clintonb: `wheel` needs to be installed before any other dependencies so that we can take advantage of wheel for those dependencies that use it. We generally alphabetize our imports, so `wheel` would be one of the last dependencies installed. Also, `wheel` isn't actually *needed* by any of the software we produce. It is more of an operational/development convenience. I'm surprised we aren't using `wheel` everywhere given the potential to speed up build/development time.

feanil: Why not make a target that just installs wheel, and make that a pre-requisite of `make requirements` ?

clintonb: We can do that; but, what's the harm in doing this at the system level (e.g. install `wheel` after we install the virtualenv)?

feanil: Fair question. This seems like a completly different conversation from the goal of this PR which is to get docker based test parity with what we were doing before.

clintonb: Agreed.

feanil: Also, upon further investigation, virtualenv has been installing wheel by default for a while now: https://virtualenv.pypa.io/en/latest/changes/#id19  If this is not working, then it is probably a bug in ecommerce.  I quickly checked edxapp on a sandbox and it does have `wheel` installed and we don't do anything special there.  I'd like to remove this line unless you have further objections @clintonb

clintonb: (facepalm) This line was pulled from `.travis.yml`. Travis, to my knowledge, doesn't use virtualenvs, hence our adding this line. If we are running tests in containers, this is not needed.

.travis/run_tests.sh
e0d: @clintonb I think the answers are no and no in that order ;)

Even without specific plans, I don't see why we wouldn't establish a solid pattern for separation of concerns, especially given the facility of layering changes on containers available in docker.

feanil: @clintonb https://github.com/edx/configuration/pull/3496#issuecomment-263346759

.travis/run_tests.sh
feanil: This needs to be left as is because adding it to the root Dockerfile has the same issue as adding other test related config to it.  This is not needed for the basic build because that is run by ansible and runs as the ecommerce user.  The trouble is that we are running these tests using volume mounting which has various permissions issues when we try to use it in conjunction with ansible.

feanil:
@clintonb @rlucioni @jibsheet can I get a final review when you get the chance.

.travis/run_tests.sh
rlucioni: So, is this necessary? Sounds like no, if we're using virtualenvs to run tests now. Though I don't see a virtualenv being sourced anywhere.

rlucioni (approved):
Just one note about wheel.

.travis/run_tests.sh
feanil: @rlucioni you can see the venv being sourced at the top of run_tests.sh and run_coverage.sh

I'll remove this.

jibsheet (approved):

PR #1028, opened 2016-11-22
[Sol-2131] Enable single course to dynamic query switching on coupon edit page

iivic:
https://openedx.atlassian.net/browse/SOL-2131
Switching from single course to catalog query is enabled.
Switching from catalog query to single course is disabled.
@mjfrey @vkaracic @marjev Please review.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -37,6 +37,7 @@ Order = get_model('order', 'Order')
 Product = get_model('catalogue', 'Product')
 ProductCategory = get_model('catalogue', 'ProductCategory')
 ProductClass = get_model('catalogue', 'ProductClass')
+Range = get_model('offer', 'Range')
 StockRecord = get_model('partner', 'StockRecord')
 Voucher = get_model('voucher', 'Voucher')
 
@@ -251,6 +252,18 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         elif method == 'PUT':
             response = self.client.put(path, json.dumps(data), 'application/json')
         return json.loads(response.content)
+
+    def _get_voucher_range_with_updated_dynamic_catalog_values(self):
+        """Helper method for updating dynamic catalog values."""
+        path = reverse('api:v2:coupons-detail', kwargs={'pk': self.coupon.id})
+        data = {
+            'catalog_query': '*:*',
+            'course_seat_types': ['verified'],
+        }
+        self.client.put(path, json.dumps(data), 'application/json')
+        new_coupon = Product.objects.get(id=self.coupon.id)
+        vouchers = new_coupon.attr.coupon_vouchers.vouchers
+        return vouchers.first().offers.first().benefit.range, data
 
     def test_create_serializer_data(self):
         """Test if coupon serializer creates data for details page"""
@@ -477,6 +490,21 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
 
         new_coupon = Product.objects.get(id=self.coupon.id)
         self.assertEqual(new_coupon.attr.note, note)
+
+    def test_update_dynamic_range_values(self):
+        """ Verify dynamic range values are updated in case range has no catalog. """
+        voucher_range = self.coupon.attr.coupon_vouchers.vouchers.first().offers.first().benefit.range
+        Range.objects.filter(id=voucher_range.id).update(**{'catalog': None})
+        voucher_range, data = self._get_voucher_range_with_updated_dynamic_catalog_values()
+        self.assertEqual(voucher_range.catalog_query, data['catalog_query'])
+        self.assertEqual(voucher_range.course_seat_types, data['course_seat_types'][0])
+
+    def test_update_catalog_type(self):
+        """Test updating dynamic range values deletes catalog."""
+        voucher_range, data = self._get_voucher_range_with_updated_dynamic_catalog_values()
+        self.assertEqual(voucher_range.catalog, None)
+        self.assertEqual(voucher_range.catalog_query, data['catalog_query'])
+        self.assertEqual(voucher_range.course_seat_types, data['course_seat_types'][0])
 
     def test_update_coupon_benefit_value(self):
         vouchers = self.coupon.attr.coupon_vouchers.vouchers.all()
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -211,7 +211,7 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
         return response_data
 
     def update(self, request, *args, **kwargs):
-        """Update start and end dates of all vouchers associated with the coupon."""
+        """Update coupon depending on request data sent."""
         super(CouponViewSet, self).update(request, *args, **kwargs)
 
         coupon = self.get_object()
@@ -226,6 +226,9 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
 
         if range_data:
             voucher_range = vouchers.first().offers.first().benefit.range
+            # Remove catalog if switching from single course to dynamic query
+            if voucher_range.catalog:
+                range_data['catalog'] = None
             Range.objects.filter(id=voucher_range.id).update(**range_data)
 
         benefit_value = request.data.get('benefit_value')
@@ -288,7 +291,6 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
             benefit_value (Decimal): Benefit value associated with a new offer
             coupon (Product): Coupon product associated with vouchers
             vouchers (ManyRelatedManager): Vouchers associated with the coupon to be updated
-            coupon (Product): Coupon product associated with vouchers
         """
         voucher_offers = vouchers.first().offers
         voucher_offer = voucher_offers.first()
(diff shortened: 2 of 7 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1028"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -226,6 +226,9 @@ def update(self, request, *args, **kwargs):
 
         if range_data:
             voucher_range = vouchers.first().offers.first().benefit.range
+            # Remove catalog if switching from single course to dynamic query
+            if voucher_range.catalog:
+                range_data['catalog'] = None
```
vkaracic, line 231: Mind updating the docstring on this method, please?

ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
```diff
@@ -166,6 +166,12 @@ define([
 
                 expect(view.$('.catalog-query .value').text()).toEqual('id:*');
                 expect(view.$('.seat-types .value').text()).toEqual('verified, professional');
+
+                // Unset all attributes set here in order that other tests work as planned
```
vkaracic: Huh?

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
```diff
@@ -306,6 +306,20 @@ define([
                     expect(view.$('span.query_length').text()).toEqual(String(query_2.length));
                 });
             });
+
+            describe('single course coupon', function () {
+                it('should reset catalog type and related fields when cancelButtonClicked called', function() {
```
vkaracic: I believe this can be in ``it`` in the upper ``describe`` section.

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
```diff
@@ -306,6 +306,20 @@ define([
...
+                    view.model.set({
+                        'catalog_type': 'multy',
+                        'catalog_query': 'query',
+                        'course_seat_types': ['seat1','seat2']
+                    });
+                    view.cancelButtonClicked();
+                    expect(view.model.get('catalog_type')).toEqual(undefined);
+                    expect(view.model.get('catalog_query')).toEqual(undefined);
+                    expect(view.model.get('course_seat_types')).toEqual([]);
```
vkaracic: How come this equals ``[]`` here and ``undefined`` on line 295?

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -283,6 +294,18 @@ define([
...
+                if (this.initCatalogType !== this.model.get('catalog_type')) {
+                    this.model.set({
+                        'catalog_type': this.initCatalogType,
+                        'catalog_query': this.initCatalogQuery,
+                        'course_id': this.initCourseId,
+                        'course_seat_types': this.initCourseSeatTypes,
+                        'seat_type': this.initSeatType
+                    });
+                }
+            },
```
vkaracic, line 297: Something like [this](http://stackoverflow.com/a/33400454) might be easier. Clone all the attributes when someone opens the edit view and when cancel is clicked just apply the saved initial variables back to the model. This way all the variables are rolled back instead of just the ones you used.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -594,11 +622,12 @@ define([
...
-                        this.$('#credit').attr('checked', true);
-                        this.$('.non-credit-seats').addClass(this.hiddenClass);
-                    } else {
-                        this.$('#non-credit').attr('checked', true);
+                    this.$('#non-credit').attr('checked', true);
+                    if (this.model.get('course_seat_types')) {
+                        if (this.model.get('course_seat_types')[0] === 'credit') {
+                            this.$('#credit').attr('checked', true);
+                            this.$('.non-credit-seats').addClass(this.hiddenClass);
+                        }
```
vkaracic, line 619: What is happening here?

vkaracic (changes requested):

ecommerce/extensions/api/v2/views/coupons.py
iivic: Since you asked nicely.

ecommerce/static/js/test/specs/views/coupon_detail_view_spec.js
iivic: While I was writing my test I discovered that the values set in this test remained in the model. This caused issues so after hours of debugging I found the source and it works. If you have a better idea please share it with me.

vkaracic: In that case you could move these actions to a ``afterEach`` method.

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
iivic: On model init course_seat_types is set to [],  and on view init this value is being saved so when calling `view.cancelButtonClicked()` the value is reset to [] while on line 295 the value is undefined because toggling single-course radio button will unset catalog_query and course_seat_types.

ecommerce/static/js/views/coupon_form_view.js
iivic: The only issue with the example is that it makes a shallow copy what I need is a deep copy.

ecommerce/static/js/views/coupon_form_view.js
iivic: `this.model.get('course_seat_types')[0]` would result with an error because course_seat_types can be undefined that is why I added the if statement `if (this.model.get('course_seat_types'))`.

ecommerce/extensions/api/v2/views/coupons.py
marjev, line 231: Write a test on the backend side to make sure this works.

marjev (changes requested):

marjev (approved):

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -551,6 +570,15 @@ def test_dynamic_catalog_coupon(self):
...
         self.assertEqual(detail['course_seat_types'], course_seat_types)
 
+        updated_catalog_query = '*:*'
+        data = {'catalog_query': updated_catalog_query}
+        path = reverse('api:v2:coupons-detail', kwargs={'pk': self.coupon.id})
+        self.client.put(path, json.dumps(data), 'application/json')
+        new_coupon = Product.objects.get(id=self.coupon.id)
+        vouchers = new_coupon.attr.coupon_vouchers.vouchers
+        voucher_range = vouchers.first().offers.first().benefit.range
+        self.assertEqual(voucher_range.catalog_query, data['catalog_query'])
```
vkaracic: Why are you doing this here?

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -265,6 +266,12 @@ define([
                         'title',
                         'email_domains'
                     ];
+
+                    /**
+                     * Remember init attributes (deep copy).
+                     * In case cancel is clicked reset values in cancelButtonClicked function
```
vkaracic: ``Store initial model attribute values in order to revert to them when cancel button is clicked.``

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -418,6 +434,10 @@ define([
                     this.formGroup('[name=seat_type]').addClass(this.hiddenClass);
                     this.model.unset('course_id');
                     this.model.unset('seat_type');
+
+                    if (!this.model.get('course_seat_types')) {
+                        this.model.set('course_seat_types', []);
```
vkaracic, line 436: Shouldn't you unset this value?

vkaracic (commented):
A few questions.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
iivic: I need to test updating a multiple courses coupon without catalog. The multi course coupon is already created in this test but if I write another test I would have to copy all this again so I figured it's better to do it here ...

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -478,6 +479,22 @@ def test_update_note(self):
         new_coupon = Product.objects.get(id=self.coupon.id)
         self.assertEqual(new_coupon.attr.note, note)
 
+    def test_update_catalog_type(self):
```
vkaracic, line 502: Add a docstring explaining this test, please.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
vkaracic: It might be more time consuming, but the better way would be to extract all the operations that are needed for this new test into a shared helper method, write the new test and use that method there and in this one here.

ecommerce/static/js/views/coupon_form_view.js
iivic: No. In case of multi course coupon `course_seat_types` must be set for validation to kick in. In case that `course_seat_types` is unset the validation would not kick in for empty `course_seat_types` which we currently don't allow.

iivic:
Ready for another review.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -252,6 +253,17 @@ def get_response_json(self, method, path, data=None):
             response = self.client.put(path, json.dumps(data), 'application/json')
         return json.loads(response.content)
 
+    def _get_voucher_range_with_updated_dynamic_catalog_values(self):
```
vkaracic, line 256: Nit: add docstring, please.

vkaracic (approved):
LGTM

PR #1032, opened 2016-11-28
[SOL-2140] Add Backend Validation for Catalog Query

iivic:
@marjev @vkaracic @mjfrey I made the backend validation for `catalog_query` parameter and since it's tied closely with the `course_seat_types` field some validation is included for this field as well but there will be a separate PR for it with more validation.
https://openedx.atlassian.net/browse/SOL-2140

### DIFF
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -111,6 +111,28 @@ class Range(AbstractRange):
         null=True
     )
 
+    def save(self, *args, **kwargs):
+        self.clean()
+        super(Range, self).save(*args, **kwargs)  # pylint: disable=bad-super-call
+
+    def clean(self):
+        """ Validation for model fields. """
+        if self.catalog and (self.catalog_query or self.course_seat_types):
+            logger.exception(
+                'Failed to create Range. Catalog and dynamic catalog fields may not be set in the same range.'
+            )
+            raise ValidationError(_('Catalog and dynamic catalog fields may not be set in the same range.'))
+
+        # Both catalog_query and course_seat_types must be set or empty
+        exception_msg = 'Failed to create Range. If catalog_query is set course_seat_types must be set as well.'
+        validation_error_msg = _('Both catalog_query and course_seat_types fields must be set.')
+        if self.catalog_query and not self.course_seat_types:
+            logger.exception(exception_msg)
+            raise ValidationError(validation_error_msg)
+        elif self.course_seat_types and not self.catalog_query:
+            logger.exception(exception_msg)
+            raise ValidationError(validation_error_msg)
+
     def run_catalog_query(self, product):
         """
         Retrieve the results from running the query contained in catalog_query field.
--- a/ecommerce/extensions/offer/tests/test_models.py
+++ b/ecommerce/extensions/offer/tests/test_models.py
@@ -1,5 +1,6 @@
 import hashlib
 
+import ddt
 import httpretty
 import mock
 from django.core.cache import cache
@@ -16,8 +17,10 @@ from ecommerce.tests.testcases import TestCase
 
 Catalog = get_model('catalogue', 'Catalog')
 ConditionalOffer = get_model('offer', 'ConditionalOffer')
+Range = get_model('offer', 'Range')
 
 
+@ddt.ddt
 class RangeTests(CouponMixin, CourseCatalogTestMixin, CourseCatalogMockMixin, TestCase):
     def setUp(self):
         super(RangeTests, self).setUp()
@@ -72,6 +75,7 @@ class RangeTests(CouponMixin, CourseCatalogTestMixin, CourseCatalogMockMixin, Te
             non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
         """
         self.range.catalog_query = large_query
+        self.range.course_seat_types = ['verified', ]
         self.range.save()
         self.assertEqual(self.range.catalog_query, large_query)
 
@@ -147,6 +151,38 @@ class RangeTests(CouponMixin, CourseCatalogTestMixin, CourseCatalogMockMixin, Te
 
         with self.assertRaises(ValidationError):
             validate_credit_seat_type('credit,verified')
+
+    @ddt.data(
+        {'catalog_query': '*:*'},
+        {'catalog_query': '', 'course_seat_types': ['verified']},
+        {'course_seat_types': ['verified']},
+    )
+    def test_creating_range_with_wrong_data(self, data):
+        """Verify creating range without catalog_query or catalog_seat_types raises ValidationError."""
+        with self.assertRaises(ValidationError):
+            Range.objects.create(**data)
+
+    @ddt.data(
+        {'catalog_query': '*:*'},
+        {'catalog_query': '*:*', 'course_seat_types': ['verified']},
+        {'course_seat_types': ['verified']}
+    )
+    def test_creating_range_with_catalog_and_dynamic_fields(self, data):
+        """Verify creating range with catalog and dynamic fields set will raise exception."""
+        data.update({'catalog': self.catalog})
+        with self.assertRaises(ValidationError):
+            Range.objects.create(**data)
+
+    def test_creating_dynamic_range(self):
+        """Verify creating range with catalog_query or catalog_seat_types creates range with those values."""
+        data = {
+            'catalog_query': 'id:testquery',
+            'course_seat_types': ['verified', 'professional']
+        }
+        new_range = Range.objects.create(**data)
+        self.assertEqual(new_range.catalog_query, data['catalog_query'])
+        self.assertEqual(new_range.course_seat_types, data['course_seat_types'])
+        self.assertEqual(new_range.catalog, None)
 
 
 class ConditionalOfferTests(TestCase):


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1032&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

iivic:
I'm done with the changes. Please review.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -92,7 +94,19 @@ def create(self, request, *args, **kwargs):
...
+                        raise ValidationError(
+                            _('Multiple courses coupon can not be created without catalog_query parameter.')
+                        )
+                    if not course_seat_types:
+                        logger.exception(
+                            'Failed to create Coupon. Multiple courses coupon must have course_seat_types.'
+                        )
+                        raise ValidationError(
+                            _('Multiple courses coupon can not be created without course_seat_types parameter.')
+                        )
```
marjev: Why do you need validation here as well as in `Range` model? Shouldn't one be more than enough?

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -92,7 +94,19 @@ def create(self, request, *args, **kwargs):
                     except Voucher.DoesNotExist:
                         pass
 
-                if course_seat_types:
+                if request.data.get('catalog_type') == 'Multiple courses':
```
marjev: Do we actually need the `catalog_type`? As I recall, it was only a front-end attribute.

marjev (changes requested):

ecommerce/extensions/api/v2/views/coupons.py
iivic: I'm checking `catalog_type` to be sure `catalog_query` and `course_seat_types` are required.
Maybe checking if `catalog` is sent would be a good replacement??

ecommerce/extensions/api/v2/views/coupons.py
iivic: I assumed that we use the Range model in various places and not only in the coupon creation method. Notice that the model validation will pass if both of the attributes are set (catalog_query and course_seat_types) or non of them since there can be only the catalog.

ecommerce/extensions/offer/models.py
```diff
@@ -111,6 +111,30 @@ class Range(AbstractRange):
         null=True
     )
 
+    def save(self, *args, **kwargs):
+        self.clean()
+        super(Range, self).save(*args, **kwargs)  # pylint: disable=bad-super-call
+
+    def clean(self):
+        # If catalog_query is sent course_seat_types must be sent as well and vice versa
```
vkaracic: Convert these to docstrings.

ecommerce/extensions/offer/models.py
```diff
@@ -111,6 +111,30 @@ class Range(AbstractRange):
...
+        if self.course_seat_types and not self.catalog_query:
+            logger.exception(
+                'Failed to create Range. Both course_seat_types and catalog_query must be set or none of them.'
+            )
+            raise ValidationError(_('If course_seat_types is set, catalog_query must be set as well.'))
+        elif self.catalog_query and not self.course_seat_types:
```
vkaracic: You could combine these into something like:
```
if not (self.catalog_query and self.course_seat_types)
```

ecommerce/extensions/offer/models.py
```diff
@@ -111,6 +111,30 @@ class Range(AbstractRange):
...
+            logger.exception(
+                'Failed to create Range. Both course_seat_types and catalog_query must be set or none of them.'
+            )
+            raise ValidationError(_('If catalog_query is set, course_seat_types must be set as well.'))
+        self.clean_catalog_query()
+
+    def clean_catalog_query(self):
```
vkaracic: I don't think this was necessary to extract to a separate method.

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -148,6 +152,17 @@ def test_credit_seat_type_validator(self):
         with self.assertRaises(ValidationError):
             validate_credit_seat_type('credit,verified')
 
+    @ddt.data(
+        {'catalog_query': '*:*'},
+        {'catalog_query': ''},
+        {'catalog_query': '', 'course_seat_types': ['verified']},
+        {'course_seat_types': ['verified']}
+    )
+    def test_creating_range_without_required_data(self, data):
```
vkaracic: Create another unit test that verifies when the data passes without a validation error.

ecommerce/extensions/api/v2/views/coupons.py
vkaracic: The validation should kick in on saving the model and bubble up to the view. There will be more validations added later and this could easily become redundant. The more business logic in the model layer the better.

vkaracic (changes requested):

ecommerce/extensions/api/v2/views/coupons.py
iivic: @vkaracic  Yes but we need the `catalog_type` in order to know which fields are required or like I mentioned already would it be vise to check for the catalog instead??

ecommerce/extensions/offer/models.py
iivic: No. This will require catalog_query and course_seat_types to be set on every range but there is the case when only the catalog is set.

ecommerce/extensions/offer/models.py
iivic: I think this is good practice. There will be more validation for other fields and if the validation is for one field only it's better to separate them instead of putting them into one method.

vkaracic: In that case you can move the existing code in ``clean()`` to a method of it's own as well, like [here](https://github.com/edx/ecommerce/blob/6640de3848f3ee077bf403f6d6147fac94e63677/ecommerce/extensions/offer/models.py#L23), because this now seems like a half-measure.

ecommerce/extensions/offer/models.py
vkaracic:
```
if (self.catalog_query or self.course_seat_types) and not (self.catalog_query and self.course_seat_types)
```
but this is complicating it too much. Leave as is :+1:

ecommerce/extensions/offer/models.py
iivic: I was thinking about that as well but this validation includes two fields so it makes sense to me keeping them in this "common" clean method. What makes more sense to you?

ecommerce/extensions/api/v2/views/coupons.py
vkaracic: IMHO model constraining should happen at the model layer, not the view/controller layer.

iivic: I will remove this validation and redo the model validation.

ecommerce/extensions/offer/models.py
vkaracic: Not really, it's one segment defining the business logic of the model, doesn't matter if it includes two fields.

iivic:
Lots of tests failed after the model validation. Please review.

ecommerce/coupons/tests/test_views.py
```diff
@@ -9,7 +9,7 @@
 from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_class, get_model
 from oscar.test.factories import (
-    ConditionalOfferFactory, OrderFactory, OrderLineFactory, RangeFactory, VoucherFactory
+    BenefitFactory, ConditionalOfferFactory, OrderFactory, OrderLineFactory, RangeFactory, VoucherFactory
 )
```
marjev, line 13: A lot of Factories are being used from `oscar.test.factories`. It makes no sense to list all of them like this.
Just import the factories:
`from oscar.test import factories` and apply the change to all the tests.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -349,7 +354,7 @@ def test_voucher_offers_listing_catalog_query(self):
     def test_get_offers_for_single_course_voucher(self):
         """ Verify that the course offers data is returned for a single course voucher. """
         course, seat = self.create_course_and_seat()
-        new_range = RangeFactory(products=[seat, ])
+        new_range = RangeFactory(products=[seat, ], catalog=self.catalog)
```
marjev: You can DRY this.
```
course, seat = self.create_course_and_seat()
new_range = RangeFactory(products=[seat, ], catalog=self.catalog)
```
These two lines are almost in every test method.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -295,7 +300,7 @@ def test_voucher_offers_listing_product_found(self):
         course, seat = self.create_course_and_seat()
         self.mock_dynamic_catalog_single_course_runs_api(course)
 
-        new_range = RangeFactory(products=[seat, ])
+        new_range = RangeFactory(products=[seat, ], catalog=self.catalog)
```
marjev: Why do we need both products and catalog in Range?

If not, you should leave only one of them (which is necessary) and update the entire PR with the necessary change.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -452,7 +458,8 @@ def test_get_course_offer_data(self):
 
     def test_get_course_offer_verify_null_fields(self):
         """ Verify that the course offers data is properly formatted. """
-        benefit = BenefitFactory()
+        _range = RangeFactory(catalog=self.catalog)
+        benefit = BenefitFactory(range=_range)
```
marjev: DRY this as well.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -404,7 +405,7 @@ def fulfill_product(self, order, lines):
                 attributes__name='certificate_type',
                 attribute_values__value_text=line.product.attr.seat_type
             )
-            _range, created = Range.objects.get_or_create(name=name)
+            _range, created = Range.objects.get_or_create(name=name, catalog=CatalogFactory(name='test'))
```
marjev: why do you need to set a name for the catalog here?
I didn't notice you doing that in any of the previous tests.

ecommerce/extensions/offer/models.py
```diff
@@ -111,6 +111,33 @@ class Range(AbstractRange):
         null=True
     )
 
+    def save(self, *args, **kwargs):
+        self.clean()
+        super(Range, self).save(*args, **kwargs)  # pylint: disable=bad-super-call
```
marjev, line 116: My OCD senses are tingling. We're adding `# pylint: disable=bad-super-call` together with `super` without giving much thought to it.

Can we replace `Range` with `self.__class__` to avoid disabling Pylint's bad-super-call quality issue?
@clintonb any objections to my suggestion? 
I tried replacing `Range` with `type(self)` but it didn't work.

ecommerce/extensions/offer/models.py
```diff
@@ -111,6 +111,33 @@ class Range(AbstractRange):
...
+                    _('Multiple courses coupon can not be created without catalog_query parameter.')
+                )
+            if not self.course_seat_types:
+                logger.exception(
+                    'Failed to create Coupon. Multiple courses coupon must have course_seat_types.'
+                )
+                raise ValidationError(
+                    _('Multiple courses coupon can not be created without course_seat_types parameter.')
+                )
+
```
marjev, line 135: All of these messages are mentioning `Coupon` even though this is a Range validation.
Do we want to couple the two that tightly?

In future, we or someone from the Open Source Community might decide to create a Range for a new type of products (that aren't Coupons).

I think we shouldn't make an assumption that Ranges will be created solely for the purpose of Coupon creation.

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -12,52 +13,49 @@
...
         self.range.add_product(self.product)
-        self.range_with_catalog.catalog = self.catalog
         self.stock_record = factories.create_stockrecord(self.product, num_in_stock=2)
         self.catalog.stock_records.add(self.stock_record)
 
     def test_range_contains_product(self):
         """
         contains_product(product) should return Boolean value
         """
-        self.assertTrue(self.range.contains_product(self.product))
```
marjev, line 40: If we have a line `self.range.add_product(self.product)`, shouldn't this assertion remain to make sure we didn't break existing Django Oscar functionality?

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -12,52 +13,49 @@
...
         self.assertTrue(self.range_with_catalog.contains_product(self.product))
 
         not_in_range_product = factories.create_product()
-        self.assertFalse(self.range.contains_product(not_in_range_product))
-        self.assertFalse(self.range.contains_product(not_in_range_product))
```
marjev, line 45: One should remain. Other one is redundant.

ecommerce/extensions/offer/models.py
```diff
@@ -111,6 +111,33 @@ class Range(AbstractRange):
...
+            if not self.catalog_query:
+                logger.exception('Failed to create Coupon. Multiple courses coupon must have catalog_query.')
+                raise ValidationError(
+                    _('Multiple courses coupon can not be created without catalog_query parameter.')
+                )
+            if not self.course_seat_types:
+                logger.exception(
+                    'Failed to create Coupon. Multiple courses coupon must have course_seat_types.'
+                )
+                raise ValidationError(
```
vkaracic: How about:
```
if not (self.catalog_query and self.course_seat_types)
```

vkaracic (approved):
I left you one nitpick, up to you if you want to apply it.
The ``RangeFactory`` should ideally be able to be called without any required arguments. I'm not sure if that is in the scope of this PR/ticket so I'll leave that decision to you as well. If you decide not to change that at least create a tech debt ticket for creating a custom ``RangeFactory`` (possibly assign to me, I'd like to dabble with factories :smiley: )

ecommerce/extensions/offer/tests/test_models.py
marjev: A lot of these tests assumed that someone might add a `Product` to a `Range` using `add_product` method which is inherited from Django Oscar's `AbstractBenefit` class.

We should extend the existing code with additional features, without breaking any existing code.

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -12,52 +13,49 @@
...
+        self.assertFalse(self.range_with_catalog.contains_product(not_in_range_product))
 
     def test_range_number_of_products(self):
         """
         num_products() should return number of num_of_products
         """
-        self.assertEqual(self.range.num_products(), 1)
+        self.assertEqual(self.range.num_products(), 0)
```
marjev: Should be 1 if we used `add_product()` to add a product to the `Range`.

ecommerce/tests/factories.py
```diff
@@ -16,6 +16,16 @@ class Meta(object):
     short_code = factory.SelfAttribute('name')
 
 
+class CatalogFactory(factory.DjangoModelFactory):
+    """Factory for Catalog objects."""
```
marjev: This docstring can be removed. 
It's not adding anything that the name isn't already describing.

marjev (changes requested):

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
iivic: I created a new RangeFactory it will automatically create a catalog. The products parameter is optional but is needed in some tests because it's somehow involved in the test. I went through all the tests and removed the products parameter where not needed.

ecommerce/extensions/fulfillment/modules.py
iivic: Since this is get_or_create if the catalog with name test exists it will be found otherwise a Range instance will be created every time. Is there something wrong with this logic?

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
marjev: My point is that the catalog is currently empty so `new_range` isn't benefiting from it. Also, if I'm not mistaken, the catalog is an optional field.

A Range can have:
1) Catalog query and course_seat_types
2) Products (list of products)
3) Catalog (list of stock records -- I think)

iivic: The new validation requires either catalog or catalog_query and course_seat_types to be set. If none of them is set an exception will be raised.

ecommerce/extensions/fulfillment/modules.py
marjev: Factories should be used only in tests.

Also, this catalog doesn't contain any StockRecords and is useless. Catalog, therefore, should be an optional field and added to Range only if it contains StockRecords.

Products are added via `add_products` method and will be found in the `products` list.

ecommerce/extensions/offer/models.py
iivic: http://stackoverflow.com/questions/18208683/when-calling-super-in-a-derived-class-can-i-pass-in-self-class
It looks like a bad idea.

marjev (approved):

PR #1033, opened 2016-11-28
Ignore translations generated for test themes.

feanil:
Files are generated under this directory when translations testing runs. They should be ignored.

@edx/ecommerce

### DIFF
--- a/ecommerce/tests/themes-dir-2/test-theme-3/conf/locale/.gitignore
+++ b/ecommerce/tests/themes-dir-2/test-theme-3/conf/locale/.gitignore
@@ -1,1 +1,3 @@
-# this file is to add conf/local dir to git, currently there is no file to ignore.
+# this file is to add conf/local dir to git
+# it ignores translations generated for the test themes
+*


rlucioni:
Do we nest .gitignore files like this anywhere else? It seems odd.

@feanil I know you're not adding this file, but maybe we can take this opportunity to clean up. Why not put this in the .gitignore at the root of the repo?

feanil:
@rlucioni this got lost in the shuffle.  Coming back to it now.  The reason this file is here is because otherwise, this directory is not created.  I'm not super familiar with the translations logic but this sort of pattern is used to make sure a directory exists as a part of the checkout.  I'm happy to put the ignore at the toplevel but since it's just ignoring everything int his folder, I thought it made sense here.  Either way we should not remove the gitignore file.

rlucioni:
@feanil I see, thanks for the explanation. We typically use an empty `.gitkeep` file for the purpose you're describing.

clintonb:

feanil:
@rlucioni  to be clear, it was previously only for making sure the directory exists but now it actually ignores the translations under here which is useful. Feel free to close if you guys don't want it but I found it useful when I was working on dockerizing the tests.

rlucioni:
@feanil I'm going to close this for now. If we were to merge this, there should be corresponding changes made to the `.gitignore` files used by the other test themes. That said, there doesn't seem to be a need for this right now.

PR #1035, opened 2016-11-30
Updated basket page button handling

clintonb:
Pulling the processor name from the data attribute instead of the value attribute. It is unclear if the value attribute works as expected for button elements. Given that we know data attributes definitely work, we will use it.

ECOM-6155


clintonb:
@edx/ecommerce please review

mjfrey:
@clintonb could this be why we sometimes see empty payment processors in Splunk?

clintonb:
@mjfrey yes, this is meant to address that issue (ECOM-6155).

mjfrey (approved):
Excellent.

schenedx (approved):
👍

PR #1036, opened 2016-11-30
Updated production requirements

clintonb:


clintonb:

schenedx (approved):
Is there a need to have the library updated? Or this is a good practice in general?

clintonb:
@schenedx we primarily want to update New Relic to get bug fixes. In general, I like to keep dependencies updated for bug fixes and to avoid falling too far behind.

PR #1037, opened 2016-12-01
[SOL-2151] Allow max number of uses to be increased on a multi-use coupon

iivic:
@vkaracic @marjev @mjfrey While enabling this feature I noticed a bug: on editing once per customer coupon the max_uses field value was always 1. Now this is fixed. The max_uses field editing is only enabled for multi-use and not for once per customer coupons. Please review.

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -29,6 +29,7 @@ from ecommerce.invoice.models import Invoice
 Basket = get_model('basket', 'Basket')
 Catalog = get_model('catalogue', 'Catalog')
 Category = get_model('catalogue', 'Category')
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
 logger = logging.getLogger(__name__)
 Order = get_model('order', 'Order')
 Product = get_model('catalogue', 'Product')
@@ -253,12 +254,9 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
             coupon.attr.note = note
             coupon.save()
 
-        if 'email_domains' in request.data:
-            email_domains = request.data.get('email_domains')
-            # Need to update for individual vouchers because in case of multiple
-            # multi-use voucher each voucher will have individual offer.
-            for voucher in vouchers.all():
-                voucher.offers.update(email_domains=email_domains)
+        offer_data = self.create_update_data_dict(data=request.data, fields=ConditionalOffer.UPDATABLE_OFFER_FIELDS)
+        if offer_data:
+            ConditionalOffer.objects.filter(vouchers__in=vouchers.all()).update(**offer_data)
 
         self.update_invoice_data(coupon, request.data)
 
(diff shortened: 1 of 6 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1037&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

iivic:
screen shot 2016-12-01 at 12 35 04

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -248,6 +248,10 @@ def update(self, request, *args, **kwargs):
         if coupon_price:
             StockRecord.objects.filter(product=coupon).update(price_excl_tax=coupon_price)
 
+        if 'max_uses' in request.data:
+            for voucher in vouchers.all():
+                voucher.offers.update(max_global_applications=request.data.get('max_uses'))
```
marjev: update `UPDATEABLE_VOUCHER_FIELDS` with `max_uses` and do the bulk update on line 223.

ecommerce/static/js/test/specs/views/coupon_edit_view_spec.js
```diff
@@ -170,6 +170,14 @@ define([
                     expect(view.model.get('course_id')).toBe(formView._initAttributes.course_id);
                     expect(view.model.get('seat_type')).toBe(formView._initAttributes.seat_type);
                 });
+
+                it('should verify editing of max_uses field on multi-use voucher is possible', function (){
+                    expect(view.$('[name=max_uses]').attr('disabled')).toBe(undefined);
```
marjev: https://api.jquery.com/has-attribute-selector/

marjev (changes requested):

ecommerce/extensions/api/v2/views/coupons.py
iivic: max_uses is an offer field so I don't see how this will work or maybe I misunderstood you?

marjev: combine the `max_global_applications` with `email_domains` update so we might do the bulk_update instead of two separate updates.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -253,12 +254,12 @@ def update(self, request, *args, **kwargs):
             coupon.attr.note = note
             coupon.save()
 
-        if 'email_domains' in request.data:
-            email_domains = request.data.get('email_domains')
+        if [field for field in ConditionalOffer.UPDATABLE_OFFER_FIELDS if field in request.data]:
+            offer_data = self.create_update_data_dict(data=request.data, fields=ConditionalOffer.UPDATABLE_OFFER_FIELDS)
```
marjev:
```
offer_data = self.create_update_data_dict(data=request.data, fields= ConditionalOffer. UPDATABLE_OFFER_FIELDS):
    if offer_data:
        ConditionalOffer.objects.filter(vouchers__in=vouchers.all()).update(**offer_data)
```

marjev (changes requested):

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -468,10 +469,17 @@ define([
                      */
                     if (voucherType === 'Multi-use') {
                         this.model.set('max_uses', multiUseMaxUsesValue);
-                        this.$(maxUsesFieldSelector).attr('min', 2);
+                        if (this.editing) {
+                            this.$(maxUsesFieldSelector).removeAttr('disabled');
```
vkaracic: When exactly is this field disabled then?

marjev (approved):

ecommerce/static/js/views/coupon_form_view.js
iivic: It is disable when voucherType is Once per customer on edit page.

vkaracic: It's hidden for once-per-customer on edit page. Anyway, all you should do is remove the ``non-editable`` class from the HTML template so that this is not triggered: https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/views/coupon_form_view.js#L572

You can also remove the test that checks if this field has disabled attribute.

iivic: It's not hidden but I will make it editable.

iivic:
Comments addressed and commits squashed.

vkaracic (approved):
LGTM

PR #1039, opened 2016-12-05
[SOL-2153] When editing coupon and save without actualy changing anything invoice amount changes

iivic:
@mjfrey  @vkaracic @marjev It's just one line but still review and check if the bug is fixed and if other bugs occurred because of the fix I already did but one can not be careful enough.

### DIFF
--- a/ecommerce/static/js/views/coupon_form_view.js
+++ b/ecommerce/static/js/views/coupon_form_view.js
@@ -531,9 +531,7 @@ define([
                         }
                     });
                     // update field
-                    this.$('[name=seat_type]')
-                        .html(this.seatTypes)
-                        .trigger('change');
+                    this.$('[name=seat_type]').html(this.seatTypes).trigger('change');
 
                     if (this.editing && seatType) {
                         this.$('[name=seat_type]').val(_s.capitalize(seatType));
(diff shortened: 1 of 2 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1039"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

vkaracic (approved):
LGTM

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -552,7 +550,8 @@ define([
 
                 this.model.set('seat_type', seatType);
 
-                if (seatType) {
+                // The price should not change when editing
+                if (seatType && !this.editing) {
```
marjev, line 554: Do we have a test case for this?

Usually, when a bug is caught, it would be great if we could have a test for that case?

So, I guess we have the one for the automatic update of the invoice amount. Can we add one for editing use-case, just to make sure that someone doesn't break it in the future?

marjev (changes requested):

ecommerce/static/js/views/coupon_form_view.js
iivic: Done.

marjev (approved):

PR #1040, opened 2016-12-05
[SOL-2149] Pagination is not showing correct range of coupons on the last page.

iivic:
@mjfrey @vkaracic @marjev Quick review.
https://openedx.atlassian.net/browse/SOL-2149

### DIFF
--- a/ecommerce/static/js/collections/offer_collection.js
+++ b/ecommerce/static/js/collections/offer_collection.js
@@ -36,7 +36,7 @@ define([
 
             updateLimits: function() {
                 this.lowerLimit = (this.page - 1) * this.perPage;
-                this.upperLimit = this.page * this.perPage;
+                this.upperLimit = this.onLastPage() ? this.length : this.page * this.perPage;
             },
 
             goToPage: function(pageNumber) {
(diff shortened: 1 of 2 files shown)


iivic:
screen shot 2016-12-05 at 16 53 20

vkaracic (approved):
LGTM

ecommerce/static/js/collections/offer_collection.js
```diff
@@ -36,7 +36,7 @@ define([
 
             updateLimits: function() {
                 this.lowerLimit = (this.page - 1) * this.perPage;
-                this.upperLimit = this.page * this.perPage;
+                this.upperLimit = this.onLastPage() ? this.length : this.page * this.perPage;
```
marjev, line 39: Can you write additional test that would catch this bug, in case this breaks in the future?

marjev (changes requested):

ecommerce/static/js/collections/offer_collection.js
iivic: Done.

marjev (approved):

PR #1041, opened 2016-12-07
changed acceptance_tests to e2e for oep-8 compliance

estute:
@jzoldak @benpatterson 
for TE-1636


benpatterson:
Overall looks good. I see a reference to an acceptance environment variable in this file: `ecommerce/extensions/dashboard/orders/tests.py` Not sure if that's important, so I'm pointing it out.

Otherwise LGTM

jzoldak:
Yeah, @benpatterson @estute I'd vote that we leave the system environment variable DISABLE_ACCEPTANCE_TESTS as-is, because we may not know all the places that it is set and/or used.
👍  (after rebase)

estute:
FYI @edx/ecommerce

rlucioni (approved):

PR #1042, opened 2016-12-08
[SOL-2162] Add Backend Validation for Client

iivic:
@mjfrey @vkaracic @marjev The client model doesn't allow empty values for name field so if a post request without the name parameter is sent to our coupon api it will break. This ensures the error is handled properly.

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -499,10 +499,16 @@ class Client(User):
 class BusinessClient(models.Model):
     """The model for the business client."""
 
-    name = models.CharField(_('Name'), unique=True, max_length=255, blank=False, null=False)
+    name = models.CharField(_('Name'), unique=True, max_length=255)
 
     def __str__(self):
         return self.name
+
+    def save(self, *args, **kwargs):
+        if not self.name:
+            log.exception('Failed to create BusinessClient. BusinessClient name may not be empty.')
+            raise ValidationError(_('BusinessClient name must be set.'))
+        super(BusinessClient, self).save(*args, **kwargs)
 
 
 def validate_configuration():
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1042&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

vkaracic (approved):
Nitpick: remove ``blank`` and ``null`` arguments here https://github.com/edx/ecommerce/pull/1042/files#diff-c35ba6c955b6de8aba48f3a0f17be18eR502

ecommerce/core/models.py
```diff
@@ -499,11 +499,17 @@ class Client(User):
 class BusinessClient(models.Model):
     """The model for the business client."""
 
-    name = models.CharField(_('Name'), unique=True, max_length=255, blank=False, null=False)
+    name = models.CharField(_('Name'), unique=True, max_length=255)
```
marjev, line 502: I think this requires a new migration.

marjev (changes requested):

ecommerce/core/models.py
iivic: Those are the default values so removing them changes nothing hence no migration is needed.

marjev (approved):

ecommerce/core/models.py
davec-edx: Will they be noticed during a make migrations?  The next commit that will require it might pick up on that and raise a question about it.

iivic: Removing the values doesn't change the model because the values are already set by default so the change will not be picked up by django makemigrations. I just tried to make sure there shouldn't be any problems. You are welcome to double check 😄

PR #1043, opened 2016-12-09
Updated configuration for Docker-based devstack

clintonb:
ECOM-6569


clintonb:
@rlucioni @davec-edx @jmbowman please review
FYI @edx/devops

Makefile
```diff
@@ -27,6 +27,7 @@ help:
 	@echo '    make static                       compile and compress static assets               		'
 	@echo '    make detect_changed_source_translations    check if translation files are up-to-date		'
 	@echo '    make validate_translations        install fake translations and check if translation files are up-to-date'
+	@echo '    make production-requirements      install requirements for production                    '
```
davec-edx, line 30: Possible duplicate comment:  switch to _ instead of -

Makefile
```diff
@@ -36,6 +37,9 @@ requirements.js:
 requirements: requirements.js
 	pip install -r requirements/local.txt --exists-action w
 
+production-requirements: requirements.js
```
davec-edx, line 40: Change target to use _ instead of -

Makefile
```diff
@@ -122,4 +126,4 @@ validate_translations: fake_translations detect_changed_source_translations
 # Targets in a Makefile which do not produce an output file with the same name as the target name
 .PHONY: help requirements migrate serve clean validate_python quality validate_js validate html_coverage accept \
 	extract_translations dummy_translations compile_translations fake_translations pull_translations \
-	push_translations update_translations fast_validate_python clean_static
+	push_translations update_translations fast_validate_python clean_static production-requirements
```
davec-edx, line 129: _ instead of -

(thought I was in a review, but not...oops)

ecommerce/settings/base.py
```diff
@@ -1,6 +1,8 @@
 """Common settings and globals."""
 import datetime
 import os
+import platform
+from logging.handlers import SysLogHandler
```
davec-edx, line 5: Is this just needed so we can run in devstack?  If so, should we wrap it in DEBUG so production isn't bringing in unnecessary code unless it is in DEBUG mode? (more possible paths)

ecommerce/settings/base.py
```diff
@@ -285,32 +287,74 @@
...
-            'handlers': ['mail_admins'],
-            'level': 'ERROR',
+            'handlers': ['console', 'local'],
             'propagate': True,
+            'level': 'WARNING'
+        },
+        '': {
+            'handlers': ['console', 'local'],
+            'level': 'DEBUG',
+            'propagate': False
```
davec-edx, line 357: It may go against convention, but should we eliminate the  formatting in the file so it is tabular.  It might make it easier to quickly scan and make sure all logging levels are where they're supposed to be and not accidentally left  at the wrong level.

*MIGHT* want a unit test that will warn if target is production and a log level is at the wrong level in production so we don't flood the logs due to a statement being missed.

ecommerce/settings/base.py
davec-edx: Do we have any flags that say where an IDA is running (PROD, STAGE, DEVSTACK)?

davec-edx (approved):

ecommerce/settings/base.py
rlucioni: @davec-edx the syslog handler is used in production, too. We don't have flags telling us which environment we're in other than Django's DEBUG. Again, @jmbowman's work may help here. That aside, the hope is to have development settings mirror production settings as closely as possible.

ecommerce/settings/base.py
rlucioni: Again, I think we'll want to address any changes here separately.

rlucioni (approved):

ecommerce/settings/base.py
davec-edx: @rlucioni Agreed, and I didn't want to write a book about it in a comment. So consider it just a note :-)

PR #1044, opened 2016-12-09
prototype for enterprise entitlements flow

zubair-arbi:
ENT-84

This is a **sample code** for the prototype of enterprise entitlements.
Please ignore the print statements or commented code.
@asadiqbal08 @saleem-latif @mattdrayer @kashifch

* add settings variable `ENTERPRISE_API_URL` to locate enterprise service API
* add method `enterprise_api_client` in `core/models.py`
* add enterprise utils file to fetch data for enterprises
* update basket view to display enterprise name and logo

**Related edx-enterprise PR for enterprise entitlements:** https://github.com/edx/edx-enterprise/pull/21

### Not Added (pending)
* validate the coupon record in Ecom
* add flow in case learner hasn't signed consent
* apply enterprise coupon in basket view before rendering in case of valid and eligible enterprise learner

### Basket page Screen shots
screen shot 2016-12-09 at 2 50 28 pm

### DIFF
--- a/ecommerce/coupons/enterprise_utils.py
+++ b/ecommerce/coupons/enterprise_utils.py
@@ -1,0 +1,46 @@
+"""
+Enterprise related utility functions.
+"""
+import logging
+
+from django.conf import settings
+from django.core.cache import cache
+
+
+log = logging.getLogger(__name__)
+
+
+def get_user_enterprise_info(user, site):
+    """
+    Get the Enterprise for the given learner along with the entitlement info.
+
+    Arguments:
+        user (django.contrib.auth.User): django auth user
+        site (django.contrib.sites.Site): site instance
+    """
+    # TODO: Fetch learner related enterprise info from Enterprise API against a specific site
+    # response = {}
+    # try:
+    #     response = site.siteconfiguration.enterprise_api_client.course_runs.get(user.username, site.domain)
+    # except (ConnectionError, SlumberBaseException, Timeout):
+    #     log.exception(
+    #         'Failed to retrieve enterprise info for the learner [%s]',
+    #         user.username
+    #     )
+    #     raise
+
+    # create dummy data for testing
+    response = {
+        'username': user.username,
+        'enterprise': {
+            'name': 'Arbisoft',
+            'catalog': 1,
+            'site_domain': site.domain,
+            'enable_data_sharing_consent': True,
+            'enforce_data_sharing_consent': 'At Enrollment',
+            'logo': 'https://www.edx.org/sites/default/files/homepage/banner/promo/bkg/onethird-539x290-csedweek.jpg',
+            'user_data_sharing_consent': True
+        }
+    }
+    # TODO: Cache the response from enterprise API in case of 200 status
+    return response
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -18,6 +18,7 @@ from oscar.core.loading import get_class, get_model
 from ecommerce.core.url_utils import get_ecommerce_url
 from ecommerce.core.views import StaffOnlyMixin
 from ecommerce.coupons.decorators import login_required_for_credit
+from ecommerce.coupons.enterprise_utils import get_user_enterprise_info
 from ecommerce.extensions.api import exceptions
 from ecommerce.extensions.basket.utils import prepare_basket
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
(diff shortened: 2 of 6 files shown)


mattdrayer:
@brittneyexline FYI on this proto-PR

ecommerce/extensions/basket/views.py
```diff
@@ -106,6 +107,34 @@ def get_context_data(self, **kwargs):
         basket = self.request.basket
         site = self.request.site
         site_configuration = site.siteconfiguration
+        user = self.request.user
+
+        # TODO: Add Enterprise related entitlements code here
```
mattdrayer, line 112: We should wrap all of the enterprise code up into its own function(s)

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -11,6 +11,20 @@
...
 {% if not basket.is_empty %}
     <div id="content-inner">
+        {% if enterprise_customer %}
+            <h3>
+                <span class="pull-left">
+                    <img class="thumbnail" height="50px" width="50px" src="{{ enterprise_customer.logo }}"/>
+                </span>
+                <span class="pull-left">
+                     {% blocktrans with enterprise_name=enterprise_customer.name %}
+                        You are eligible for the enterprise entitlements from {{ enterprise_name }}.
```
mattdrayer, line 21: @srpearce -- this is just a prototype at the moment, but just wanted to point out a piece of learner-facing copy here.

ecommerce/templates/oscar/basket/partials/basket_content.html
```diff
@@ -11,6 +11,20 @@
 
 {% if not basket.is_empty %}
     <div id="content-inner">
+        {% if enterprise_customer %}
```
mattdrayer, line 14: We should be more specific here about the intended outcome and its corresponding switch -- that is, we probably should not trigger UI behavior based simply on the presence of an object.  If we want to display a logo we should pass a logo, or at least check for the presence of a logo.  If we want to display a line of eligibility copy we should base this on whether or not the user is actually eligible for the entitlement(s).

ecommerce/extensions/basket/views.py
```diff
@@ -106,6 +107,34 @@ def get_context_data(self, **kwargs):
...
+            # proceed to the checkout
+            logger.info(
+                'Learner [%s] is not associated with any enterprise, proceeding to checkout.',
+                self.request.user.username
+            )
+        else:
+            # learner is registered with an enterprise
+            enable_data_sharing_consent = enterprise_customer.get('enable_data_sharing_consent')
+            user_data_sharing_consent = enterprise_customer.get('user_data_sharing_consent')
+            if enable_data_sharing_consent and not user_data_sharing_consent:
```
mattdrayer, line 126: this might read better if the parameter was named `enterprise_data_sharing_consent_is_required`, and the subsequent parameter `user_data_sharing_consent_is_granted`.

ecommerce/coupons/views.py
```diff
@@ -100,6 +101,9 @@ def get_context_data(self, **kwargs):
                 return {'error': _('Coupon does not exist')}
             except exceptions.ProductNotFoundError:
                 return {'error': _('The voucher is not applicable to your current basket.')}
+
+            # TODO: Add Enterprise related entitlements code here
```
mattdrayer, line 105: Let's be sure to package up this (and all) enterprise-related code into its own function(s) which can be independently covered with tests.

ecommerce/coupons/enterprise_utils.py
```diff
@@ -0,0 +1,46 @@
...
+import logging
+
+from django.conf import settings
+from django.core.cache import cache
+
+
+log = logging.getLogger(__name__)
+
+
+def get_user_enterprise_info(user, site):
```
mattdrayer, line 13: Let's reword all of our enterprise functions to 'get_enterprise...', etc.  For example this one would be `get_enterprise_user_info(user, site)`  I also think that we should position `site` as the first parameter when it is needed.

zubair-arbi:
@mattdrayer I have addressed your feedback in a new commit against the @saleem-latif [Ecom PR for entitlements](https://github.com/edx/ecommerce/pull/1046/) in an attempt to centralize the prototype changes:

Please review the changes: https://github.com/edx/ecommerce/pull/1046/commits/450b8f42d626cbfb9a231024192ea4fb1ac7a44e

mattdrayer:
Supports work being done in #1061 -- perhaps we should combine these two change sets.

zubair-arbi:
Closing this PR as its no longer needed.

PR #1045, opened 2016-12-09
[SOL-2166] Add Backend Validation for Code

iivic:
The following cases are covered: 
1. Voucher can not be created without code
2. Voucher can only be created with alphanumeric code value
3. Enrollment coupon can not be created if code is sent (random code will be created)
4. Coupon will not be created if quantity is greater than one and code sent (random codes will be created depending of quantity value).

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -714,6 +714,23 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         self.data['category'] = 'String type'
         self.assert_post_response_status(self.data)
 
+    def test_creating_enrollment_coupon_with_code(self):
+        """ Test that bad request status is returned when creating enrollment coupon with code parameter set. """
+        self.data.update({'code': 'test'})
+        self.assert_post_response_status(self.data)
+
+    def test_creating_coupon_with_code_and_quantity_greater_than_one(self):
+        """
+        Test that bad request status is returned when creating discount coupon
+        with code set and quantity greater than one.
+        """
+        self.data.update({
+            'benefit_value': 90,
+            'code': 'test',
+            'quantity': 5
+        })
+        self.assert_post_response_status(self.data)
+
 
 class CouponCategoriesListViewTests(TestCase):
     """ Tests for the coupon category list view. """
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -5,7 +5,7 @@ import logging
 import dateutil.parser
 from django.conf import settings
 from django.core.exceptions import ValidationError
-from django.db import transaction
+from django.db import IntegrityError, transaction
 from django.http import Http404
 from django.shortcuts import get_object_or_404
 from oscar.core.loading import get_model
@@ -136,25 +136,28 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
                 else:
                     coupon_catalog = None
 
-                coupon_product = create_coupon_product(
-                    benefit_type=request.data.get('benefit_type'),
-                    benefit_value=request.data.get('benefit_value'),
-                    catalog=coupon_catalog,
-                    catalog_query=request.data.get('catalog_query'),
-                    category=category,
-                    code=code,
-                    course_seat_types=course_seat_types,
-                    email_domains=request.data.get('email_domains'),
-                    end_datetime=dateutil.parser.parse(request.data.get('end_datetime')),
-                    max_uses=max_uses,
-                    note=request.data.get('note'),
-                    partner=partner,
-                    price=request.data.get('price'),
-                    quantity=request.data.get('quantity'),
-                    start_datetime=dateutil.parser.parse(request.data.get('start_datetime')),
-                    title=request.data.get('title'),
-                    voucher_type=voucher_type
-                )
+                try:
+                    coupon_product = create_coupon_product(
+                        benefit_type=request.data.get('benefit_type'),
+                        benefit_value=request.data.get('benefit_value'),
+                        catalog=coupon_catalog,
+                        catalog_query=request.data.get('catalog_query'),
+                        category=category,
+                        code=code,
+                        course_seat_types=course_seat_types,
+                        email_domains=request.data.get('email_domains'),
+                        end_datetime=dateutil.parser.parse(request.data.get('end_datetime')),
+                        max_uses=max_uses,
+                        note=request.data.get('note'),
+                        partner=partner,
+                        price=request.data.get('price'),
+                        quantity=request.data.get('quantity'),
+                        start_datetime=dateutil.parser.parse(request.data.get('start_datetime')),
+                        title=request.data.get('title'),
+                        voucher_type=voucher_type
+                    )
+                except IntegrityError as e:
+                    return Response(str(e), status=status.HTTP_400_BAD_REQUEST)
 
                 basket = prepare_basket(request, coupon_product)
 
--- a/ecommerce/extensions/catalogue/tests/test_utils.py
+++ b/ecommerce/extensions/catalogue/tests/test_utils.py
@@ -111,12 +111,13 @@ class CouponCreationTests(CouponMixin, TestCase):
         super(CouponCreationTests, self).setUp()
         self.catalog = Catalog.objects.create(partner=self.partner)
 
-    def create_custom_coupon(self, code='', max_uses=None, note=None, quantity=1, title='Tešt Čoupon'):
+    def create_custom_coupon(self, benefit_value=100, code='', max_uses=None, note=None, quantity=1,
+                             title='Tešt Čoupon'):
         """Create a custom test coupon product."""
 
         return create_coupon_product(
             benefit_type=Benefit.PERCENTAGE,
-            benefit_value=100,
+            benefit_value=benefit_value,
             catalog=self.catalog,
             catalog_query=None,
             category=self.category,
@@ -137,12 +138,14 @@ class CouponCreationTests(CouponMixin, TestCase):
     def test_custom_code_integrity_error(self):
         """Test custom coupon code duplication."""
         self.create_custom_coupon(
+            benefit_value=90,
             code='CUSTOMCODE',
             title='Custom coupon',
         )
 
         with self.assertRaises(IntegrityError):
             self.create_custom_coupon(
+                benefit_value=90,
                 code='CUSTOMCODE',
                 title='Coupon with integrity issue',
             )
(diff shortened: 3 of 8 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1045&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -696,6 +696,23 @@ def test_not_update_email_domains(self):
...
+    def test_creating_coupon_with_code_and_quantity_greater_than_one(self):
+        """
+        Test that an exception is raised when creating discount coupon with code set and quantity greater than one.
+        """
+        self.data.update({
+            'benefit_value': 90,
+            'code': 'test'
+        })
+        response = self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
davec-edx: I'm unclear what this is testing. It looks like this is testing a val of 90 with a code, but the docstring says it is testing a quantity.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -72,16 +74,28 @@ def create(self, request, *args, **kwargs):
...
         course_seat_types = request.data.get('course_seat_types')
         max_uses = request.data.get('max_uses')
         partner = request.site.siteconfiguration.partner
+        quantity = request.data.get('quantity')
         stock_record_ids = request.data.get('stock_record_ids')
         voucher_type = request.data.get('voucher_type')
 
         try:
             with transaction.atomic():
+                if benefit_type == Benefit.PERCENTAGE and benefit_value == 100 and code:
```
davec-edx: Is this being tested?

ecommerce/extensions/api/v2/tests/views/test_coupons.py
iivic: The data is set in the setUp method (quantity=2). The value is changed to 90 to get a discount coupon if the value is 100 and type percentage it will be an enrollment coupon and that is not what I need. So if quantity > 1 and code sent it should return 400 status.

ecommerce/extensions/api/v2/views/coupons.py
iivic: https://github.com/edx/ecommerce/pull/1045/files/13a6e5e7e5788379acaafb94c25e6636050c3289#diff-3bdba75be46de6780d19faac1f58ded3R699

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -696,6 +696,23 @@ def test_not_update_email_domains(self):
         self.get_response_json('PUT', path, self.data)
         self.assertEqual(offer.email_domains, email_domains)
 
+    def test_creating_enrollment_coupon_with_code(self):
+        """ Test that an exception is raised when creating enrollment coupon with code parameter set. """
```
marjev: Actually, you're testing that the `bad request` response is returned, not that an exception is raised.

Change the docstring so it doesn't confuse any code readers.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -696,6 +696,23 @@ def test_not_update_email_domains(self):
...
+        self.data.update({'code': 'test'})
+        response = self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_creating_coupon_with_code_and_quantity_greater_than_one(self):
+        """
+        Test that an exception is raised when creating discount coupon with code set and quantity greater than one.
```
marjev: same comment I left for the previous test.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -72,16 +74,28 @@ def create(self, request, *args, **kwargs):
...
+                    return Response(_('Code may not be set for Enrollment Coupon.'), status=status.HTTP_400_BAD_REQUEST)
```
marjev: Move this to the utility method that calls the `Voucher.objects.create()` for now.

This is not the ideal solution, and we'll have to create a `Voucher` model later when we get to Vouchers app cleanup, but for now, this is the best we can do.

Again, backend validation should not reside within API endpoint.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -72,16 +74,28 @@ def create(self, request, *args, **kwargs):
...
+
+                if quantity > 1 and code:
```
marjev: Is this actually needed?

I thought Django Oscar placed the unique restriction on the code field. If you remove this and catch the Integrity error, I think you'll be able to remove this.

Just pass the exception message to the Response and leave the same status.

ecommerce/extensions/catalogue/tests/test_utils.py
```diff
@@ -158,7 +158,7 @@ def test_coupon_note(self):
 
     def test_custom_code_string(self):
         """Test creating a coupon with custom voucher code."""
-        custom_code = 'ČUSTOMĆODE'
+        custom_code = 'CUSTOMCODE'
```
marjev, line 164: You don't need to change this.

ecommerce/extensions/voucher/models.py
```diff
@@ -16,5 +24,16 @@ class OrderLineVouchers(models.Model):
...
-# noinspection PyUnresolvedReferences
+
+class Voucher(AbstractVoucher):
+    def save(self, *args, **kwargs):
+        if not self.code:
+            logger.exception('Failed to create Voucher. Voucher code must be set.')
+            raise ValidationError(_('Voucher code must be set.'))
+        if not re.match('^[a-zA-Z0-9]+$', self.code):
+            logger.exception('Failed to create Voucher. Voucher code must contain only alphanumeric characters.')
+            raise ValidationError(_('Voucher code must contain only alphanumeric characters.'))
```
marjev, line 37: make a `clean` method and call it here instead.

marjev (changes requested):

ecommerce/extensions/api/v2/views/coupons.py
iivic: You are right. The code field is unique and IntegrityError is raised.

ecommerce/extensions/voucher/models.py
```diff
@@ -16,5 +24,16 @@ class OrderLineVouchers(models.Model):
...
     vouchers = models.ManyToManyField('voucher.Voucher', related_name='order_line_vouchers')
 
-# noinspection PyUnresolvedReferences
+
+class Voucher(AbstractVoucher):
+    def save(self, *args, **kwargs):
+        if not self.code:
+            logger.exception('Failed to create Voucher. Voucher code must be set.')
+            raise ValidationError(_('Voucher code must be set.'))
+        if not re.match('^[a-zA-Z0-9]+$', self.code):
```
vkaracic: ``if not self.code.isalnum()``

ecommerce/extensions/catalogue/tests/test_utils.py
iivic: The code must contain only alphanumeric values. Diacritic characters are not allowed. That's why this will not pass the validation.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
vkaracic: Nevertheless there is no harm in explicitly setting the quantity to be more than two.

vkaracic (changes requested):

iivic:
All comments addressed please review the changes.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -707,6 +707,25 @@ def test_update_max_uses_field(self):
         response = self.get_response_json('PUT', path, self.data)
         self.assertEqual(response['max_uses'], max_uses)
 
+    def test_creating_enrollment_coupon_with_code(self):
+        """ Test that bad request status is returned when creating enrollment coupon with code parameter set. """
+        self.data.update({'code': 'test'})
+        response = self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
vkaracic: Perhaps wait for this to merge: https://github.com/edx/ecommerce/pull/1050/files#diff-3bdba75be46de6780d19faac1f58ded3R246

ecommerce/extensions/voucher/models.py
```diff
@@ -1,5 +1,12 @@
 # noinspection PyUnresolvedReferences
```
vkaracic: Is this some PyCharm crap? :smile:

ecommerce/extensions/voucher/utils.py
```diff
@@ -362,6 +362,9 @@ def _create_new_voucher(code, coupon, end_datetime, name, offer, start_datetime,
     Returns:
         Voucher
     """
+    if offer.benefit.type == Benefit.PERCENTAGE and offer.benefit.value == 100 and code:
+        logger.exception('Failed to create Voucher. Code may not be set for enrollment coupon.')
+        raise ValidationError(_('Voucher can not be created when code is set in enrollment coupon.'))
```
vkaracic, line 367: I would rather see this in the voucher model validation method than here.

ecommerce/extensions/voucher/models.py
iivic: It seems so...

ecommerce/extensions/voucher/utils.py
iivic: I don't see how this will work. A Voucher must have a code field set so this validation makes no sense in the model it would disallow creating enrollment codes.

vkaracic: Right. My mistake.

marjev (approved):

ecommerce/extensions/voucher/tests/test_models.py
```diff
@@ -0,0 +1,18 @@
...
+class VoucherTests(TestCase):
+    def test_no_code_raises_exception(self):
+        """ Verify creating voucher without code set raises exception. """
+        with self.assertRaises(ValidationError):
+            Voucher.objects.create()
+
+    def test_wrong_code_data_raises_exception(self):
+        """ Verify creating voucher with code value that contains spaces (non alphanumeric value) raises exception. """
+        with self.assertRaises(ValidationError):
+            Voucher.objects.create(code='Only alphanumeric without spaces')
```
vkaracic, line 33: Add a test case which should pass and create a voucher.

vkaracic (changes requested):

vkaracic (approved):

PR #1046, opened 2016-12-09
WIP -- Entitlements for enterprise learners

saleem-latif:
@mattdrayer , @asadiqbal08 , @zubair-arbi  FYI

### DIFF
--- a/ecommerce/enterprise/exceptions.py
+++ b/ecommerce/enterprise/exceptions.py
@@ -1,0 +1,9 @@
+"""
+Exception classes for enterprise app.
+"""
+
+
+class RequiresDataSharingError(Exception):
+    """
+    Exception is raised when a learner needs to consent for data sharing before getting the entitlements
+    """
--- a/ecommerce/enterprise/exceptions.py
+++ b/ecommerce/enterprise/exceptions.py
@@ -3,7 +3,7 @@ Exception classes for enterprise app.
 """
 
 
-class RequiresDataSharingError(Exception):
+class RequiresDataSharingConsentError(Exception):
     """
     Exception is raised when a learner needs to consent for data sharing before getting the entitlements
     """
--- a/ecommerce/enterprise/helpers.py
+++ b/ecommerce/enterprise/helpers.py
@@ -1,0 +1,92 @@
+"""
+Helper methods for enterprise app.
+"""
+import logging
+from edx_rest_api_client.client import EdxRestApiClient
+from requests.exceptions import ConnectionError, Timeout
+from slumber.exceptions import SlumberBaseException
+
+log = logging.getLogger(__name__)
+
+
+def get_learner_info(user, site):
+    """
+    Fetch user and its enterprise data from lms.
+
+    Args:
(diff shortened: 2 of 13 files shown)


ecommerce/enterprise/helpers.py
```diff
@@ -0,0 +1,152 @@
+"""
+Helper methods for enterprise app.
```
mattdrayer, line 2: We'll need to add an "is_feature_enabled" check at some point

ecommerce/enterprise/exceptions.py
```diff
@@ -0,0 +1,9 @@
+"""
+Exception classes for enterprise app.
+"""
+
+
+class RequiresDataSharingError(Exception):
```
mattdrayer: RequiresDataSharingConsentError

ecommerce/settings/base.py
```diff
@@ -599,3 +602,12 @@
...
 CRISPY_TEMPLATE_PACK = 'bootstrap3'
+
+
+# Enterprise settings
+# enable or disable enterprise features
+ENABLE_ENTERPRISES = True
+
+# name for waffle switch to use for disabling enterprise features on runtime.
+# Note: management command ignore this switch
+DISABLE_ENTERPRISES_ON_RUNTIME_SWITCH = "disable_theming_on_runtime"
```
mattdrayer: looks like this needs to be fixed

ecommerce/extensions/basket/views.py
```diff
@@ -74,6 +75,10 @@ def get(self, request):
                 msg = _('An error occurred while retrieving enrollment details. Please try again.')
                 return HttpResponseBadRequest(msg)
 
+        # If learner is eligible for entitlements, apply entitlements
+        if voucher is None and enterprise_helpers.is_learner_eligible_for_entitlements(request.user, request.site):
+            voucher = enterprise_helpers.get_entitlement_voucher(request, product)
```
mattdrayer, line 80: maybe we can move the eligibility check inside of get_entitlement_voucher?

mattdrayer:
Overall this prototype makes sense to me, although I see two possible approaches:

1. Introduce the enterprise concept into each service where pertinent workflow orchestration is required.  Keep the orchestration logic contained inside of the enterprise app.  Minimize dependencies and completely refuse/reject direct coupling to application constructs.

2) Keep ALL enterprise orchestration logic within the enterprise app currently hosted in the LMS.  Execute workflows in other services via API calls from the enterprise app to each service.

I think we may want to go the latter route here, versus adding a "slice" of enterprise orchestration to each service where it is needed.  Thoughts?

saleem-latif:
@mattdrayer If we are not doing a lot of customizations (branding wise of behavior wise) in ecommerce then we can go with the second approach. However, for customizing ecommerce, e.g. customized basket page for enterprise learners or having learners prompted for data sharing consent in ecommerce, we will have to add enterprise concept there.

Second approach, however, does seem a better way to go and I think for current use cases will be a good way forward.

clintonb:
Please delete your branches after closing PRs.

PR #1047, opened 2016-12-09
Add logging to help identify affected learner and basket for debugging

schenedx:
@edx/ecommerce Please review

### DIFF
--- a/ecommerce/extensions/checkout/views.py
+++ b/ecommerce/extensions/checkout/views.py
@@ -42,7 +42,12 @@ class FreeCheckoutView(EdxOrderPlacementMixin, RedirectView):
             # Need to re-apply the voucher to the basket.
             Applicator().apply(basket, self.request.user, self.request)
             if basket.total_incl_tax != Decimal(0):
-                raise BasketNotFreeError("Basket is not free.")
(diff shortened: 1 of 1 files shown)


ecommerce/extensions/checkout/views.py
```diff
@@ -42,7 +42,12 @@ def get_redirect_url(self, *args, **kwargs):
             # Need to re-apply the voucher to the basket.
             Applicator().apply(basket, self.request.user, self.request)
             if basket.total_incl_tax != Decimal(0):
-                raise BasketNotFreeError("Basket is not free.")
+                raise BasketNotFreeError(
+                    "Basket [{}] is not free. User affected [{}]".format(
+                        basket.id,
+                        self.request.user.email
```
clintonb: Use `self.request.user.id` to avoid exposing PII.

However, `basket.owner` should already be set to `self.request.uer`.

mikedikan (approved):

clintonb (changes requested):

schenedx:
@clintonb Took in your feedback. Please review again

clintonb (approved):

PR #1048, opened 2016-12-09
Client side checkout basket

vkaracic:
This PR consists of:

- Generalizing basket page: changes to the current basket so that it accepts different type of products and can accept multiple products. In the basket view each line is 'handled' based on the product class of that line's product. If no handler is specified it falls back to 'regular type' product handler. Screenshots: https://drive.google.com/drive/folders/0ByBQUYJ4LMyAb3lDTl96ZVpleXM?usp=sharing

- Client side checkout basket: Redesign of the new client side checkout basket with keeping in mind backwards compatibility with old/current basket. Screenshot available here: https://drive.google.com/drive/folders/0ByBQUYJ4LMyANll2VmtNNDRHY28?usp=sharing

- Credit card recognition and validation: credit cards are recognized automatically on user input (if the credit card information [regex pattern, CyberSource type ID, CVV number length] is available in the ``getCreditCardType`` function).
The credit card regex matches are created according to [these ranges](https://en.wikipedia.org/wiki/Payment_card_number#Issuer_identification_number_.28IIN.29). The credit card number is validated with [Luhn's algorithm](https://en.wikipedia.org/wiki/Luhn_algorithm). CVN inputted length is checked if it matches the inputted credit card type's CVN length. The expiry month and year validation fires if a non-number is sent or the selected combination of month and year is less than current month and year.

- Factoring out payment processor specifics: namely the CyberSource specific field signing code in ``basket_app.js``. This way different payment processors can be plugged in.

- [ ] Documentation update

https://openedx.atlassian.net/browse/SOL-2113


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1048&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/basket/utils.py
```diff
@@ -57,6 +57,8 @@ def prepare_basket(request, product, voucher=None):
 
 def get_basket_switch_data(product):
     product_class_name = product.get_product_class().name
+    structure = product.structure
```
mjfrey, line 60: what are these used for?

ecommerce/static/js/apps/basket_app.js
```diff
@@ -1,9 +1,11 @@
 require([
         'jquery',
-        'pages/basket_page'
+        'pages/basket_page',
+        'payment_processors/cybersource'
     ],
     function ($,
-              BasketPage) {
+              BasketPage,
+              CyberSource) {
```
mjfrey: Is there anyway we can dynamically pull these in based on the client side processor setting?

ecommerce/extensions/basket/views.py
```diff
@@ -96,105 +97,134 @@ def _determine_seat_type(self, product):
...
+            payment_processor_class = site_configuration.get_client_side_payment_processor_class()
+
+            if payment_processor_class:
+                payment_processor = payment_processor_class(site)
+
+                today = datetime.today()
+                context.update({
+                    'enable_client_side_checkout': True,
+                    'client_side_payment_processor_name': payment_processor.NAME,
+                    'paypal_enabled': len(payment_processors) > 1,
```
mjfrey: How does this ensure PayPal?  Need to check if it is in the list.

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -16,10 +17,34 @@
 {% endblock %}
 
 {% block content %}
-    <div class="basket">
+    <div class="basket {% if enable_client_side_checkout %}basket-new{% endif %}">
```
mjfrey: basket-client-side instead of basket-new

mjfrey (changes requested):

ecommerce/static/js/pages/basket_page.js
```diff
@@ -67,11 +66,51 @@ define([
...
+                $(errorMsg).empty();  // Clear existing validation error messages.
+            });
+            cardType = Utils.getCreditCardType(cardNumber);
+
+            if (!Utils.isValidCardNumber(cardNumber)) {
+                appendValidationErrorMsg(event, cardNumberField, 'Invalid card number');
+            } else if (typeof cardType === 'undefined') {
+                appendValidationErrorMsg(event, cardNumberField, 'Unsupported card type');
+            } else if (cvnNumber.length !== cardType.cvnLength || !Number.isInteger(Number(cvnNumber))) {
+                appendValidationErrorMsg(event, cvnNumberField, 'Wrong CVN number');
```
mjfrey: Invalid CVN

ecommerce/extensions/payment/forms.py
```diff
@@ -27,13 +32,31 @@ class PaymentForm(forms.Form):
...
+            Div('city', css_class='col-md-6'),
+            Div('state', css_class='col-md-6'),
+            Div('country', css_class='col-md-6'),
+            Div('postal_code', css_class='col-md-6')
+        )
         self.fields['basket'].queryset = self.fields['basket'].queryset.filter(owner=user)
+        for bound_field in self:
+            # https://www.w3.org/WAI/tutorials/forms/validation/#validating-required-input
+            if hasattr(bound_field, 'field') and bound_field.field.required:
+                self.fields[bound_field.name].label = _('{label} (required)'.format(label=bound_field.label))
```
mjfrey, line 86: Can we change all the (required) to * to be consistent.

vkaracic: No. The ticket specifically stated to use ``(required)`` instead of *.

vkaracic:
@clintonb can you please review this? Each commit is _mostly_ a separate change so you can go through it commit by commit if it's easier.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -195,6 +195,11 @@ def post(self, request):
             if not basket:
                 logger.error('Received payment for non-existent basket [%s].', basket_id)
                 return HttpResponse(status=400)
+
+            if cybersource_response.get('transaction_id').lower() != 'accept':
+                basket.thaw()
+                logger.info('Received a non-accept decision. Basket [%s] thawed.', basket.id)
```
mjfrey: Can we add to the log if this was using the client or hosted page?  Will be nice to understand failures for each.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -0,0 +1,176 @@
...
+                        <p class="validation-error"></p>
+                    </div>
+                    <input type="hidden" name="card_expiry_date" class="pci-field">
+                </div>
+
+                <button id="payment-button" type="submit" class="btn btn-primary btn-large col-sm-offset-6 col-sm-6 col-xs-12"
+                        data-track-type="click"
+                        data-track-event="edx.bi.ecommerce.basket.payment_selected"
+                        data-track-category="checkout"
+                        data-processor-name="{{ client_side_payment_processor_name }}"
```
mjfrey: Let's add an additional attribute that specifies this is from the client-side checkout page.

mjfrey:
@clintonb please take a look.

ecommerce/extensions/basket/views.py
```diff
@@ -96,105 +97,143 @@ def _determine_seat_type(self, product):
...
+                line_data = self._get_course_data(line.product)
+                show_voucher_form = False
+            else:
+                line_data = {
+                    'product_title': line.product.title,
+                    'image_url': None,
+                    'product_description': line.product.description
+                }
 
+            # TODO: handle these links for multi-line baskets.
```
clintonb, line 161: Handle what links? Is this going to be done?

ecommerce/extensions/basket/views.py
```diff
@@ -96,105 +97,143 @@ def _determine_seat_type(self, product):
...
-                except (KeyError, TypeError):
-                    image_url = ''
-                short_description = course.get('short_description', '')
-                course_name = course.get('title', '')
-            except (ConnectionError, SlumberBaseException, Timeout):
-                logger.exception('Failed to retrieve data from Catalog Service for course [%s].', course_key)
+            product_class_name = line.product.get_product_class().name
+            if product_class_name == 'Seat':
+                line_data = self._get_course_data(line.product)
+                if (hasattr(line.product.attr, 'id_verification_required') and
```
clintonb: `if getattr(line.product.attr, 'id_verification_required', False) and line.product.attr.certificate_type != 'credit':`

ecommerce/extensions/basket/views.py
```diff
@@ -96,105 +97,143 @@ def _determine_seat_type(self, product):
...
-            user = self.request.user
-            context.update({
-                'analytics_data': prepare_analytics_data(
-                    user,
-                    self.request.site.siteconfiguration.segment_key,
-                    unicode(course_key)
-                ),
-                'enable_client_side_checkout': False,
-            })
+        course_key = lines_data[0].get('course_key') if len(lines) == 1 else None
```
clintonb, line 180: Nope! Any code that is still assuming we are using single-item baskets needs to be updated or removed.

ecommerce/extensions/payment/forms.py
```diff
@@ -8,13 +8,18 @@
...
+
 logger = logging.getLogger(__name__)
 Basket = get_model('basket', 'Basket')
 
 
 def country_choices():
     """ Returns a tuple of tuples, each containing an ISO 3166 country code. """
-    return ((country.alpha2, country.name) for country in pycountry.countries)
+    countries = [(country.alpha2, country.name) for country in pycountry.countries]
+    countries.insert(0, ('', '<Choose country>'))
```
clintonb: Mark for translation.

ecommerce/extensions/payment/forms.py
clintonb: Add a translators comment on the line above explaining how this should be translated.

ecommerce/extensions/payment/views/cybersource.py
clintonb: What's the plan here? Why is the payment processor code being modified?

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -158,6 +165,10 @@ def test_not_accepted(self, decision):
         # The basket should not have an associated order if no payment was made.
         self.assertFalse(Order.objects.filter(basket=self.basket).exists())
 
+        # The basket should be in open.
```
clintonb: The basket should be in the OPEN state.

ecommerce/static/js/utils/utils.js
```diff
@@ -162,6 +162,81 @@ define([
                         el.setAttribute('datepicker-initialized', 'true');
                     }
                 });
+            },
+
+            /**
```
clintonb: Move this to `ecommerce/static/js/utils/credit_card.js`.

ecommerce/static/js/apps/basket_app.js
clintonb: We must pull these in dynamically to support other payment processors. It seems I never documented/shared the plan for payment processors. Here it is:

> The idea is that we maintain the existing backend architecture. Each client-side processor will have an associated Django template that will load JS. (I don't think there is a need for anything else.) We will build  the payment form. The processor-specific JS can add hidden fields, loads Stripe.js, etc. Essentially the payment form, and its element IDs, is our contract.

We will need to do the following:
1. Create `templates/payment/_cybersource.html` which loads the CyberSource JS.
2. Add a variable to the processor, `template_name`, and set it to `'payment/_cybersource.html'`.
2. Update the checkout view/template to render this template.

I realize this is a bit of extra work. I can take this on after this lands. For now, please load the CyberSource file on its own, separately from `basket_app.js`.

ecommerce/static/js/payment_processors/cybersource.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * CyberSource payment processor specific actions.
+ */
+define(['jquery'], function($) {
+        'use strict';
+
+    var signFields = function(event, signingUrl, $paymentForm) {
+        var $signedFields = $('input,select', $paymentForm).not('.pci-field');
+        // Format the date to a format that CyberSource accepts (MM-YYYY).
```
clintonb: Add a new line above this to make the code a bit more legible.

ecommerce/static/js/payment_processors/cybersource.js
```diff
@@ -0,0 +1,52 @@
...
+                for (var key in formData) {
+                    if (formData.hasOwnProperty(key)) {
+                        $paymentForm.append(
+                            '<input type="hidden" name="' + key + '" value="' + formData[key] + '" />'
+                        );
+                    }
+                }
+            },
+            error: function (jqXHR, textStatus, errorThrown) {
+                // TODO Handle errors. Ideally the form should be validated in JavaScript
```
clintonb: This needs to be handled.

ecommerce/templates/oscar/basket/partials/hosted_checkout_basket.html
```diff
@@ -0,0 +1,190 @@
...
+{% load purchase_info_tags %}
+{% load widget_tweaks %}
+
+{% if not is_bulk_purchase %}
+    {% include 'partials/alert_messages.html' %}
+{% endif %}
+
+<div id="content-inner">
+    {% block basket_form_main %}
+        <form action="." method="post" class="basket_summary" id="basket_formset">
```
clintonb: This ID attribute is unused, and should be removed.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -0,0 +1,196 @@
...
+{% load currency_filters %}
+{% load crispy_forms_tags %}
+{% load widget_tweaks %}
+
+{% if show_voucher_form %}
+    {% include 'partials/alert_messages.html' %}
+{% endif %}
+
+<div id="summary" class="col-sm-5">
+        <form action="." method="post" class="basket_summary row" id="basket_formset">
```
clintonb: This ID attribute is unused, and should be removed.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -0,0 +1,196 @@
...
+                {% if paypal_enabled %}
+                <legend>{% trans "SELECT PAYMENT METHOD" %}</legend>
+                {% endif %}
+                <div class="col-sm-12 payment-methods">
+                    <div id="payment-method" class="col-md-4 col-sm-7 col-xs-6">
+                        <a href="#payment-information" id="payment-method-image" aria-label="Pay with a Credit Card" role="button"><img aria-hidden="true" src="/static/images/credit_card_options.png"></a>
+                    </div>
+                    <div id="payment-processor" class="payment-buttons col-md-3 col-sm-5 col-xs-6" data-basket-id="{{ basket.id }}">
+                    {% if paypal_enabled %}
+                        <button aria-label="Pay with paypal"
```
clintonb: Mark this, and all of your labels, for translation.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -0,0 +1,196 @@
...
+                                data-track-type="click"
+                                data-track-event="edx.bi.ecommerce.basket.payment_selected"
+                                data-track-category="checkout"
+                                data-processor-name="paypal"
+                                data-course-id="{{ course_key }}"
```
clintonb, line 120: Nope!

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -0,0 +1,196 @@
...
+                                class="btn payment-button payment-processor-paypal"
+                                value="paypal"
+                                id="paypal"
```
clintonb: This ID is unused, and should be removed.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -0,0 +1,196 @@
...
+            </fieldset>
+        </div>
+        {% endif %}
+        <div id="billing-information" class="form-input-elements placeholder">
+            <fieldset>
+                {% if not free_basket %}
+                <legend aria-label="Billing information">{% trans "BILLING INFORMATION" %}</legend>
+                <div class="pci-fields row">
+                    <div id="card-number" class="form-group col-sm-6">
+                        <label for="card-number-input" class="control-label">{% trans "Card Number (required)" %}<span class="sr">, Secure</span></label>
```
clintonb: Mark the screen reader text for translation. The same goes for the other text you have in this template.

ecommerce/templates/oscar/basket/partials/hosted_checkout_basket.html
```diff
@@ -0,0 +1,190 @@
...
+                    {% endfor %}
+                {% endif %}
+            </div>
+        </div>
+    </div>
+
+    <div class="row verification-note">
+        <div class="col-sm-12">
+            <span><i class="fa fa-exclamation-circle"></i></span>
+            {# Translators: <b></b> tags will bold the text within. Keep the tags and translate the text within. #}
```
clintonb: Use `<strong>` instead of `<b>`. It provides better semantics.

ecommerce/templates/oscar/basket/partials/hosted_checkout_basket.html
```diff
@@ -0,0 +1,190 @@
...
+            {% endif %}
+        </div>
+    </div>
+    <div class="row help">
+        <div class="col-sm-12">
+            <p><b>{% trans "Have questions?" %}</b></p>
+            <a href="{{ support_url }}">{% trans "Please read our FAQs to view common questions about our certificates." %}</a>
+        </div>
+    </div>
+</div>
```
clintonb: Add a new line.

clintonb:
Where can I see this in action?

ecommerce/extensions/basket/views.py
vkaracic: The links for switching between bulk enrollment codes and individual seats. AFAIK only white label sites are using this and when they decide to introduce multi-line baskets the whole logic for handling these use cases needs to be changed. I thought I could change it in this PR but it's too much out of scope and the PR is already too big so I just left a note.

ecommerce/extensions/basket/views.py
vkaracic: I think we'll need to talk to the analytics team about how they'd want to handle multi-line basket analytics. This line was just for backwards compatibility.

ecommerce/extensions/payment/views/cybersource.py
vkaracic: If a user makes a mistake, for example inputs the wrong ZIP code, or there is a problem with the payment processor servers and proceeds to payment, CyberSource will decline the transaction and the basket will be frozen and as such cannot be accessed again so the user in that case would have to add the item(s) to a new basket all over again. This will thaw the basket in that case and the user can fix the mistake and try again.

@mjfrey we'll need to think about adding a 'remove item' button to the form.

vkaracic:
@clintonb comments addressed, please take another look.

ecommerce/extensions/basket/utils.py
clintonb: I have the same question.

ecommerce/extensions/basket/views.py
clintonb: Has this conversation been started?

ecommerce/extensions/payment/tests/mixins.py
```diff
@@ -306,7 +306,7 @@ def setUp(self):
 
         self.toggle_ecommerce_receipt_page(True)
 
-        self.user = factories.UserFactory()
+        self.user = factories.UserFactory(password=self.password)
```
clintonb: I don't think this will do what you intend.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -171,14 +171,18 @@ class CybersourceInterstitialViewTests(CybersourceNotificationTestsMixin, TestCa
     path = reverse('cybersource_redirect')
     view = CybersourceInterstitialView
 
+    def setUp(self):
```
clintonb: Why are any changes necessary to this file?

clintonb (approved):
Please start the conversation with Analytics regarding reporting. We must NOT continue assuming we have single item baskets. FYI @mulby

ecommerce/extensions/payment/tests/views/test_cybersource.py
vkaracic: Because ``self.client`` needs to have a user in order to retrieve and use that user's basket. If the user is not authenticated the basket in the request(s) is an empty basket object instead of the basket created in the ``CybersourceNotificationTestsMixin`` mixin, which would be fine if the basket summary view didn't need the actual (saved to the database) basket object which it does in the  ``test_payment_cancelled_error`` tests.

Without the changes in this file and the added password in the mixin file above, this error would occur: https://gist.github.com/vkaracic/5a4dab7d944f3a58c1ff72ad425307c7

ecommerce/extensions/basket/utils.py
vkaracic: These are used to construct the link for switching between a seat and bulk enrollment coupon purchase (BEC). I've added default values for these because, if you look a few lines below, these are set only for BEC and seats and in case of a 'regular' product, ie. not one of those two, this would break.

ecommerce/extensions/basket/views.py
vkaracic: @mjfrey please advise on how / when are we going to address multi-line basket analytics questions.

ecommerce/extensions/payment/tests/views/test_cybersource.py
clintonb: You don't need to modify as much as you have. Set `fetch_redirect_response=False` on the call to [`assertRedirects`](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.SimpleTestCase.assertRedirects).

vkaracic: You are correct :+1:

ecommerce/extensions/basket/views.py
mjfrey: I will speak with Gabe and see how they want to handle this.

mjfrey:
@vkaracic  please squash commits.

mjfrey (approved):

clintonb:
@vkaracic please add the ticket number, SOL-2113, to your final commit(s).

vkaracic:
@mjfrey this now leaves the question of documentation. Do we write the documentation for this or notify the documentation team?

clintonb:
Even if the Doc Team were fully-staffed, you should write your own first draft of documentation. As the implementer of the component you have the most knowledge to write the most knowledgeable documentation. 



On Tue, Dec 20, 2016 at 4:44 AM -0500, "Vedran Karačić" <notifications@github.com> wrote:












@mjfrey this now leaves the question of documentation. Do we write the documentation for this or notify the documentation team?

mulby:
@mjfrey @clintonb @vkaracic - Does this change allow us to process multiple items per transaction?

We built our audit process to handle this case (in theory), however, it has not be vetted very thoroughly since there have not been very many ways this could happen.

If this is going to become more commonplace, some subset of our two teams should probably figure out what the impact will be on the audit.

mulby:
@katymyw @stroilova @brianhw FYI

vkaracic:
Adding an item to the basket in our current flow does not support adding more than one product, but changes in this PR allow the basket to support multiple lines. The product adding logic is going to change at some point and we want everything else to be ready for supporting multiple lines, among which are analytics.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -148,6 +153,227 @@ define([
...
+                            '/static/images/credit_cards/' + card.name + '.png'
+                        );
+                        expect($('input[name=card_type]').val()).toEqual(card.type);
+                    });
+                });
+            });
+
+            describe('clientSideCheckoutValidation', function() {
+                    var currentYear = moment().year(),
+                        lastMonth = moment().subtract(1, 'month').month();
```
saleem-latif, line 189: @vkaracic For some reason `moment().subtract(1, 'month').month()` is giving wrong number.

```
> moment().year()
 2017
> moment().month()
 0
> moment().subtract(1, 'month').month()
 11
```
I think this is the correct approach
```
> moment().subtract(1, 'month').year()
2016
> moment().subtract(1, 'month').month()
11
```
I am updating this here https://github.com/edx/ecommerce/pull/1071, please let me know if you have any concerns.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -67,11 +68,74 @@ define([
...
+            if (!CreditCardUtils.isValidCardNumber(cardNumber)) {
+                appendValidationErrorMsg(event, cardNumberField, 'Invalid card number');
+            } else if (typeof cardType === 'undefined') {
+                appendValidationErrorMsg(event, cardNumberField, 'Unsupported card type');
+            } else if (cvnNumber.length !== cardType.cvnLength || !Number.isInteger(Number(cvnNumber))) {
+                appendValidationErrorMsg(event, cvnNumberField, 'Invalid CVN');
+            }
+
+            if (!Number.isInteger(Number(cardExpiryMonth)) ||
+                Number(cardExpiryMonth) > 12 || Number(cardExpiryMonth) < 1) {
```
saleem-latif, line 123: @vkaracic Upon further discovery I tracked another test failure to this place, in Javascript monthly numbering starts from 0 i.e. (Jan is represented by a 0, Feb with 1 and so on. so this check alsi needs to be updated.) I think it would be better to create a JIRA ticker for this to track the work.
@mattdrayer FYI

mattdrayer: https://openedx.atlassian.net/browse/ENT-127

PR #1050, opened 2016-12-13
[SOL-2169] Add Backend Validation for Category

iivic:
I covered the case when no category or category with wrong value type (must be dict) is sent to the api.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -243,6 +243,10 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         self.response = self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
         self.coupon = Product.objects.get(title=self.data['title'])
 
+    def assert_post_response_status(self, data, expected_status=status.HTTP_400_BAD_REQUEST):
+        response = self.client.post(COUPONS_LINK, json.dumps(data), 'application/json')
+        self.assertEqual(response.status_code, expected_status)
+
     def get_response_json(self, method, path, data=None):
         """Helper method for sending requests and returning JSON response content."""
         if method == 'GET':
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1050"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -707,6 +707,12 @@ def test_update_max_uses_field(self):
         response = self.get_response_json('PUT', path, self.data)
         self.assertEqual(response['max_uses'], max_uses)
 
+    def test_create_coupon_without_category(self):
+        """ Verify creating coupon without category returns bad request. """
+        del self.data['category']
+        response = self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
```
vkaracic, line 716: Can you add another test case where the category is other than a dictionary type?

iivic: Sure.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -707,6 +707,18 @@ def test_update_max_uses_field(self):
...
+    def test_create_coupon_with_category_not_dict(self):
+        """ Verify creating coupon with category not being a dictionary returns bad request. """
+        self.data['category'] = 'String type'
+        response = self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
vkaracic: You could factor out these two lines into a assert method, something like ``assert_post_response_status()``. There are 6-7 places where this is repeated in this test class.

vkaracic (approved):

marjev (approved):

PR #1052, opened 2016-12-14
[SOL-2168] Add Backend Validation for Course Seat Types

iivic:
The following cases are covered:
1. Value must be string or unicode (in range model)
2. 'credit' type can't be paired with other types (in range model)
3. Model values must be subset of ['credit', 'professional', 'verified'] (in range model)
4. On coupon creation course seat types must be a list or the method that joins the values will return a TypeError exception. I caught and handled the exception.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -731,6 +731,21 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         })
         self.assert_post_response_status(self.data)
 
+    @ddt.data(8, 'string', {'dict': 'type'})
+    def test_creating_coupon_with_wrong_course_seat_types(self, course_seat_types):
+        """ Verify creating coupon with course seat types not a list returns bad request. """
+        self.data.update({'course_seat_types': course_seat_types})
+        self.assert_post_response_status(self.data)
+
+    def test_creating_coupon_with_course_seat_types(self):
+        """ Verify creating coupon with course seat types list creates coupon. """
+        self.data.update({
+            'catalog_query': 'test',
+            'course_seat_types': ['verified'],
+            'stock_record_ids': None
+        })
+        self.assert_post_response_status(self.data, status.HTTP_200_OK)
+
 
 class CouponCategoriesListViewTests(TestCase):
     """ Tests for the coupon category list view. """
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -94,9 +94,25 @@ class ConditionalOffer(AbstractConditionalOffer):
         return super(ConditionalOffer, self).is_condition_satisfied(basket)  # pylint: disable=bad-super-call
 
 
-def validate_credit_seat_type(value):
-    if len(value.split(',')) > 1 and 'credit' in value:
+def validate_credit_seat_type(course_seat_types, allowed_seat_types):
+    if not isinstance(course_seat_types, basestring):
+        logger.exception('Failed to create Range. Credit seat types must be type str or unicode.')
+        raise ValidationError(_('Credit seat types must be type str or unicode.'))
+
+    course_seat_types_list = course_seat_types.split(',')
+
+    if len(course_seat_types_list) > 1 and 'credit' in course_seat_types_list:
+        logger.exception('Failed to create Range. Credit seat type cannot be paired with other seat types.')
         raise ValidationError('Credit seat types cannot be paired with other seat types.')
+
+    if not set(course_seat_types_list).issubset(set(allowed_seat_types)):
+        logger.exception(
+            'Failed to create Range. Not allowed course seat types %s. Allowed values for course seat types are %s',
+            course_seat_types_list, allowed_seat_types
+        )
+        raise ValidationError(_(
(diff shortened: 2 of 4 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1052&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -714,6 +714,12 @@ def test_create_coupon_with_category_not_dict(self):
         self.data['category'] = 'String type'
         self.assert_post_response_status(self.data)
 
+    @ddt.data(8, 'string', {'dict': 'type'})
+    def test_creating_coupon_with_wrong_course_seat_types(self, course_seat_types):
+        """ Verify creating coupon with course seat types not a list returns bad request. """
+        self.data.update({'course_seat_types': course_seat_types})
+        self.assert_post_response_status(self.data)
+
```
davec-edx, line 739: Should there be a corresponding positive test to confirm that list does work?

iivic: It doesn't hurt to add one 😄

ecommerce/extensions/offer/migrations/0006_auto_20161025_1446.py
```diff
@@ -15,6 +15,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='range',
             name='course_seat_types',
-            field=models.CharField(blank=True, max_length=255, null=True, validators=[ecommerce.extensions.offer.models.validate_credit_seat_type]),
```
vkaracic, line 18: Don't edit past migrations, create new one with the changes applied.

ecommerce/extensions/offer/models.py
```diff
@@ -135,6 +130,28 @@ def clean(self):
...
+            raise ValidationError(_('Credit seat types must be type str or unicode.'))
+
+        course_seat_types_list = self.course_seat_types.split(',')
+        if len(course_seat_types_list) > 1 and 'credit' in course_seat_types_list:
+            logger.exception('Failed to create Range. Credit seat type cannot be paired with other seat types.')
+            raise ValidationError(_('Credit seat type cannot be paired with other seat types.'))
+
+        if not set(course_seat_types_list).issubset(set(self.COURSE_SEAT_TYPES_VALUES)):
+            logger.exception(
+                'Failed to create Range. Allowed values for course seat types are %s',
```
vkaracic: Nit: maybe indicate what exactly failed.

ecommerce/extensions/offer/models.py
```diff
@@ -94,21 +94,16 @@ def is_condition_satisfied(self, basket):
...
-    if len(value.split(',')) > 1 and 'credit' in value:
-        raise ValidationError('Credit seat types cannot be paired with other seat types.')
-
-
 class Range(AbstractRange):
     UPDATABLE_RANGE_FIELDS = [
         'catalog_query',
         'course_seat_types',
     ]
+    COURSE_SEAT_TYPES_VALUES = ['credit', 'professional', 'verified']
```
vkaracic: Nit: ``ALLOWED_SEAT_TYPES`` sounds better.

vkaracic (changes requested):

ecommerce/extensions/offer/migrations/0006_auto_20161025_1446.py
iivic: That will not work. This is the only way I managed to remove it.

iivic:
Review once more.

vkaracic (approved):

ecommerce/extensions/offer/migrations/0006_auto_20161025_1446.py
marjev: create a new migration. Either generate it or do it manually.
In that migration, alter the field.

marjev (changes requested):

iivic:
@marjev

marjev (approved):

ecommerce/extensions/offer/models.py
```diff
@@ -94,16 +94,33 @@ def is_condition_satisfied(self, basket):
         return super(ConditionalOffer, self).is_condition_satisfied(basket)  # pylint: disable=bad-super-call
 
 
-def validate_credit_seat_type(value):
-    if len(value.split(',')) > 1 and 'credit' in value:
+def validate_credit_seat_type(course_seat_types, allowed_seat_types):
```
saleem-latif, line 97: @iivic While trying to update a range, I am getting this error on django admin
```
TypeError at /admin/offer/range/3/
validate_credit_seat_type() takes exactly 2 arguments (1 given)
```
`validate_credit_seat_type` is used as a validator [here](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/offer/models.py#L191) which expects the method to require only one argument. So, I think this method should either not have a second arguments or second argument should have a default value.

iivic: @saleem-latif I agree with you. The second argument is not required. I created a PR with the fix please review and test if it works for you (it did for me). https://github.com/edx/ecommerce/pull/1129

PR #1053, opened 2016-12-14
WL-862 Multi-source analytics event tracking

douglashall:
**Jira Story**:  https://openedx.atlassian.net/browse/WL-862
**Sandbox Testing**
https://edx.sandbox.edx.org
https://mitpe.sandbox.edx.org
See me for access to the Segment sources and GA properties that these sites are connected to.

This PR allows us to configure multiple Segment write keys for analytics tracking. Additionally, it allows us to configure zero or more GA property tracking IDs (this is needed to overcome a Segment limitation which prevents us from adding more than one GA integration per-segment source). Each site can have it's own analytics tracking configuration.

For server-side tracking, we send tracking events to each Segment source configured for the current site. For client-side tracking, we send tracking events to the default Segment source and any GA properties that are configured for the current site. This gives us the flexibility to set up Segment integrations with GA in such a way that we can send tracking events to multiple GA properties from both client and server.

A typical multi-GA property tracking setup would include the following:

- Segment Source A
- Segment Source B
- GA Property A
- GA Property B
- Add GA Property A as an integration under Segment Source A
- Add GA Property B as an integration under Segment Source B
- Configure Segment Source A's write key as the default write key in SiteConfiguration
- Configure Segment Source B's write key as an additional write key in SiteConfiguration
- Configure GA Property B's tracking ID in SiteConfiguration

Under this configuration, server-side tracking events will flow through Segment Source A/B to their respective GA Properties. Client-side tracking events will flow through Segment Source A to GA Property A and directly to GA Property B.

For now, I have left the deprecated `SiteConfiguration.segment_key` field so that we can deploy this change to production with no effect on edx.org, populate the new `SiteConfiguration.analytics_configuration` field for edx.org and WL sites, and then follow up with removal of the segment_key field. I have created a follow on [PR](https://github.com/edx/ecommerce/pull/1069) for this work.

The https://edx.sandbox.edx.org sandbox site referenced above is configured with the `segment_key` field populated only to verify the rollout plan.

### DIFF
--- a/ecommerce/core/management/commands/create_or_update_site.py
+++ b/ecommerce/core/management/commands/create_or_update_site.py
@@ -1,13 +1,14 @@
 """ Creates or updates a Site including Partner and SiteConfiguration data. """
 
 from __future__ import unicode_literals
+from copy import deepcopy
 import logging
 
 from django.contrib.sites.models import Site
 from django.core.management import BaseCommand
 from oscar.core.loading import get_model
 
-from ecommerce.core.models import SiteConfiguration
+from ecommerce.core.models import DEFAULT_ANALYTICS_CONFIGURATION, SiteConfiguration
 
 logger = logging.getLogger(__name__)
 Partner = get_model('partner', 'Partner')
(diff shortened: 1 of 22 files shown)


douglashall:
@mulby @bderusha @clintonb Please review. Thanks.

douglashall:
@rlucioni Would you be able to review this?

clintonb (changes requested):
I see no need to include GA directly. We are already using Segment to contact GA. If multiple GA integrations are needed, you can either [add them to Segment](https://segment.com/docs/integrations/google-analytics/#multiple-trackers) or [use Google Tag Manager](http://www.kristaseiden.com/step-by-step-adding-a-second-ga-property-via-google-tag-manager/). 
Introducing multiplexing of this sort is something I'd rather outsource to companies that already do it, and keep our analytics footprint to a minimum.

douglashall:
@clintonb This PR implements the Segment-recommended way of including multiple GA trackers on a page which you have linked in your comment.

We could use GTM for sending page views to multiple GA properties, but we wouldn't be able to track other events that we may want to send client-side.

douglashall:
@clintonb Also, this can totally wait until next week. I have no plans to merge this today. No reason to cut your vacation short. :)

douglashall:
No longer needed.

PR #1055, opened 2016-12-19
Fix cybersource receipt page override

marjev:
@mjfrey Here's a fix for the receipt page override that supports the Old Receipt Page.

### DIFF
--- a/ecommerce/extensions/payment/processors/cybersource.py
+++ b/ecommerce/extensions/payment/processors/cybersource.py
@@ -16,6 +16,7 @@ from suds.wsse import Security, UsernameToken
 
 from ecommerce.core.constants import ISO_8601_FORMAT
 from ecommerce.core.url_utils import get_ecommerce_url
+from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.payment.constants import CYBERSOURCE_CARD_TYPE_MAP
 from ecommerce.extensions.payment.exceptions import (
     InvalidSignatureError, InvalidCybersourceDecision, PartialAuthorizationError, PCIViolation,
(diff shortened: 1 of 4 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1055"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -139,6 +139,16 @@ def _generate_parameters(self, basket, use_sop_profile, **kwargs):
             access_key = self.sop_access_key
             profile_id = self.sop_profile_id
 
+        if site.siteconfiguration.enable_otto_receipt_page:
```
mjfrey: can we just modify get_receipt_page_url to take a URL override since we already check the siteconfig flag there.That way you can just call it with the cyber source_redirect override and if the otto hosted page is enabled use it otherwise ignore.

mjfrey (changes requested):

ecommerce/extensions/payment/processors/cybersource.py
marjev: The method `get_receipt_page_url` should return either old or new receipt page URL.

The logic here should set the receipt page URL between the old receipt page URL or the Interstitial View URL.

marjev: comment addressed.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -39,11 +39,14 @@ class CybersourceNotifyViewTests(CybersourceNotificationTestsMixin, TestCase):
 @ddt.ddt
 class CybersourceSubmitViewTests(CybersourceMixin, TestCase):
     path = reverse('cybersource_submit')
+    maxDiff = None
```
mjfrey: We don't need this.

mjfrey (changes requested):

ecommerce/extensions/payment/tests/views/test_cybersource.py
marjev: fixed

mjfrey:
@clintonb can you take a look.  This is to account for the case when both the notify and the interstitial end points get called.

clintonb (approved):
Remember to test locally/on a sandbox to understand how this will actually behave. We may end up needing to create a Waffle flag to toggle one or the other.

mjfrey (approved):

PR #1060, opened 2016-12-21
[LEARNER-373] Fix Order API endpoint filtering

marjev:
@mjfrey:

Here's the fix for the Order API endpoint. Let's discuss this after standup.
I would like to discuss the use-case and how this behaves when the old Receipt page is turned on.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_orders.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_orders.py
@@ -14,6 +14,8 @@ from ecommerce.extensions.api.tests.test_authentication import AccessTokenMixin
 from ecommerce.extensions.api.v2.tests.views import OrderDetailViewTestMixin
 from ecommerce.extensions.fulfillment.signals import SHIPPING_EVENT_NAME
 from ecommerce.extensions.fulfillment.status import ORDER
+from ecommerce.extensions.test.factories import create_order
+from ecommerce.tests.factories import SiteFactory
 from ecommerce.tests.mixins import ThrottlingMixin
 from ecommerce.tests.testcases import TestCase
 
@@ -58,17 +60,21 @@ class OrderListViewTests(AccessTokenMixin, ThrottlingMixin, TestCase):
         self.assert_empty_result_response(response)
 
     def test_with_orders(self):
-        """ The view should return a list of the user's orders, sorted reverse chronologically. """
-        order = factories.create_order(user=self.user)
+        """
+        The view should return a list of the user's orders, sorted reverse chronologically, filtered by current site.
+        """
+        order = create_order(site=self.site, user=self.user)
+        create_order(site=SiteFactory(), user=self.user)
         response = self.client.get(self.path, HTTP_AUTHORIZATION=self.token)
         self.assertEqual(response.status_code, 200)
         content = json.loads(response.content)
 
+        self.assertEqual(Order.objects.count(), 2)
         self.assertEqual(content['count'], 1)
         self.assertEqual(content['results'][0]['number'], unicode(order.number))
 
         # Test ordering
-        order_2 = factories.create_order(user=self.user)
+        order_2 = create_order(site=self.site, user=self.user)
         response = self.client.get(self.path, HTTP_AUTHORIZATION=self.token)
         self.assertEqual(response.status_code, 200)
         content = json.loads(response.content)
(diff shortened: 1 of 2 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1060"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev:
@mjfrey Is this still needed? If so, I'll write a test before we merge this.

Also, is there a ticket for this work? I didn't find any but perhaps I looked on the wrong board.

marjev:
@edx/helio can someone review this?

We want to filter orders by site and show only those Orders fulfilled on the current ecommerce site.

vkaracic (approved):

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -58,13 +59,17 @@ def test_no_orders(self):
         self.assert_empty_result_response(response)
 
     def test_with_orders(self):
-        """ The view should return a list of the user's orders, sorted reverse chronologically. """
+        """
+        The view should return a list of the user's orders, sorted reverse chronologically, filtered by current site.
+        """
         order = factories.create_order(user=self.user)
+        factories.create_order(site=SiteFactory())
```
clintonb: This test isn't doing what you think it is. Both of the orders are associated with different sites. I suspect this is only passing because `order.site` and `request.site` are `None`.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -58,13 +59,17 @@ def test_no_orders(self):
...
         response = self.client.get(self.path, HTTP_AUTHORIZATION=self.token)
         self.assertEqual(response.status_code, 200)
         content = json.loads(response.content)
 
         self.assertEqual(content['count'], 1)
+        self.assertEqual(Order.objects.count(), 2)
```
clintonb: Move this up a line. It seems strange to put it in the middle of assertions for `content`.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -58,13 +59,17 @@ def test_no_orders(self):
         self.assert_empty_result_response(response)
 
     def test_with_orders(self):
-        """ The view should return a list of the user's orders, sorted reverse chronologically. """
+        """
+        The view should return a list of the user's orders, sorted reverse chronologically, filtered by current site.
+        """
         order = factories.create_order(user=self.user)
```
clintonb:
```python
from ecommerce.extensions.test.factories import create_order
order = create_order(site=self.site)
```

clintonb (changes requested):

ecommerce/extensions/api/v2/tests/views/test_orders.py
marjev: I think not. I've added two additional assertions to make sure that's not the case.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -58,17 +60,23 @@ def test_no_orders(self):
         self.assert_empty_result_response(response)
 
     def test_with_orders(self):
-        """ The view should return a list of the user's orders, sorted reverse chronologically. """
-        order = factories.create_order(user=self.user)
+        """
+        The view should return a list of the user's orders, sorted reverse chronologically, filtered by current site.
+        """
+        order = create_order(user=self.user)
```
clintonb: Please explicitly associate all orders in this test with a `Site`. This only works because `OrderCreator` relies on `threadlocals`: https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/order/utils.py#L80-L81. In the future, I want to get rid of that library. Just as it lead to confusion here, it can lead to confusion in production. Our actions should always be explicit.

clintonb (changes requested):

ecommerce/extensions/api/v2/tests/views/test_orders.py
marjev: Done.

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -58,17 +60,21 @@ def test_no_orders(self):
...
         self.assertEqual(response.status_code, 200)
         content = json.loads(response.content)
 
+        self.assertEqual(Order.objects.count(), 2)
         self.assertEqual(content['count'], 1)
         self.assertEqual(content['results'][0]['number'], unicode(order.number))
 
         # Test ordering
-        order_2 = factories.create_order(user=self.user)
+        order_2 = create_order(user=self.user)
```
clintonb: `site=self.site`

clintonb (approved):

clintonb (approved):

PR #1062, opened 2016-12-22
Client side basket checkout fixes.

vkaracic:
* Display form errors from the server on the UI
* Mobile design alignment fix
* Alt attributes for all images
* Catch the ValueError for when a user with empty basket
  opens the basket page. A basket is carried in user
  sessions as a SimpleLazyObject, if it's filled it will
  be commited to the database, but when it's empty, our
  check for vouchers will return a ValueError instead.

@clintonb I knew I forgot something. This PR contains error handling of the django payment form errors.
screenshot from 2016-12-22 13-29-50

@mjfrey this PR contains a fix for the form element alignments.
screenshot from 2016-12-22 13-36-51


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1062"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mjfrey (approved):

PR #1066, opened 2016-12-26
ENT-110: Add Catalog selection option to Coupon create/edit form

zubair-arbi:
ENT-110

There were two ways to add catalogs selection functionality:

1. Provide catalogs lists from view and pass the list as an argument all the way to `coupon_form_view.js` from `coupon_app.js`
2. Get catalogs list through `ajax` while rendering the page `coupon_form.html`.

I followed second approach which is currently being used for populating the `catagories` field and also seems more suitable for our current requirement.

We also needed a new field to save the catalog id returned from the discovery service to display the selected catalog on edit coupon view and also to display catalog courses on offer redeem page. Saving the catalog id on our end will abstract us away from any changes (name, query etc) being made in the service `discovery`. 

> _**Note:** We will need to create a follow up story to update the `Redeem` coupon page to use the new field `course_catalog` for displaying the courses against some coupon._

## Completed tasks:
* Add new field `course_catalog` to save the id for the related catalog in coupon range model
* Add migration for the `offer` model `Range`
* Update coupon form to display the new course type selection `Catalog`
* Update coupon form to add a new field `course_catalog` to allow catalogs selection while creating coupons
* Update backend code to save the `course_catalog` value in the model `Range`
* Add method `get_course_catalogs` to recursively get all catalogs or a single course catalog from the service `discovery`
* Update serializer `CouponSerializer` to include `course_catalog` field
* Display course catalog name on the coupon detail page in case of `Catalog`
* Update the `Coupon edit page` for course catalogs
* Use existing settings variable `COURSES_API_CACHE_TIMEOUT` for caching the Course Catalogs API response which has default value of `3600  # Value is in seconds`.
* Refactor the method `create` of class `CouponViewSet` and move all cleaning/validation logic to separate class method `clean_voucher_request_data`.
* Add unit tests
* Add js tests (https://github.com/edx/ecommerce/pull/1071)

### Coupon create page
screen shot 2016-12-30 at 4 15 46 pm

### Coupon detail page
screen shot 2016-12-30 at 4 16 07 pm

### Coupon edit page
screen shot 2016-12-30 at 4 16 33 pm

## Related stories to this feature (future):
* Update coupon redemption logic to handle course runs based on the catalog fetched on the basis of new `Range` model field `course_catalog`
* Update coupon report generation method to include course catalogs

### DIFF
--- a/ecommerce/coupons/tests/mixins.py
+++ b/ecommerce/coupons/tests/mixins.py
@@ -149,7 +149,7 @@ class CouponMixin(object):
     def create_coupon(self, benefit_type=Benefit.PERCENTAGE, benefit_value=100, catalog=None,
                       catalog_query=None, client=None, code='', course_seat_types=None, email_domains=None,
                       max_uses=None, note=None, partner=None, price=100, quantity=5, title='Test coupon',
-                      voucher_type=Voucher.SINGLE_USE):
+                      voucher_type=Voucher.SINGLE_USE, course_catalog=None):
         """Helper method for creating a coupon.
 
         Arguments:
@@ -157,12 +157,18 @@ class CouponMixin(object):
             benefit_value(int): The voucher benefit value
             catalog(Catalog): Catalog of courses for which the coupon applies
             catalog_query(str): Course query string
+            client (BusinessClient):  Optional business client object
             code(str): Custom coupon code
+            course_catalog (int): Course catalog id from Catalog Service
             course_seat_types(str): A string of comma-separated list of seat types
             email_domains(str): A comma seperated list of email domains
+            max_uses (int): Number of Voucher max uses
+            note (str): Coupon note.
             partner(Partner): Partner used for creating a catalog
             price(int): Price of the coupon
+            quantity (int): Number of vouchers to be created and associated with the coupon
             title(str): Title of the coupon
+            voucher_type (str): Voucher type
 
         Returns:
             coupon (Coupon)
@@ -184,6 +190,7 @@ class CouponMixin(object):
             catalog_query=catalog_query,
             category=self.category,
             code=code,
+            course_catalog=course_catalog,
             course_seat_types=course_seat_types,
             email_domains=email_domains,
             end_datetime=datetime.datetime(2020, 1, 1),
--- a/ecommerce/courses/tests/test_utils.py
+++ b/ecommerce/courses/tests/test_utils.py
@@ -1,9 +1,8 @@
 import hashlib
 
 import ddt
-import httpretty
-
 from django.core.cache import cache
+import httpretty
 
 from ecommerce.core.constants import ENROLLMENT_CODE_SWITCH
 from ecommerce.core.tests import toggle_switch
@@ -11,8 +10,9 @@ from ecommerce.core.tests.decorators import mock_course_catalog_api_client
 from ecommerce.coupons.tests.mixins import CourseCatalogMockMixin
 from ecommerce.courses.models import Course
 from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.courses.tests.mixins import CourseCatalogServiceMockMixin
 from ecommerce.courses.utils import (
-    get_certificate_type_display_value, get_course_info_from_catalog, mode_for_seat
+    get_certificate_type_display_value, get_course_info_from_catalog, mode_for_seat, get_course_catalogs
 )
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.tests.testcases import TestCase
@@ -74,3 +74,110 @@ class UtilsTests(CourseCatalogTestMixin, CourseCatalogMockMixin, TestCase):
     def test_cert_display_assertion(self):
         """ Verify assertion for invalid cert type """
         self.assertRaises(ValueError, lambda: get_certificate_type_display_value('junk'))
+
+
+@ddt.ddt
+@httpretty.activate
+class GetCourseCatalogUtilTests(CourseCatalogServiceMockMixin, TestCase):
+
+    def tearDown(self):
+        # Reset HTTPretty state (clean up registered urls and request history)
+        httpretty.reset()
+
+    def _assert_num_requests(self, count):
+        """
+        DRY helper for verifying request counts.
+        """
+        self.assertEqual(len(httpretty.httpretty.latest_requests), count)
+
+    def _assert_get_course_catalogs(self, catalog_name_list):
+        """
+        Helper method to validate the response from the method
+        "get_course_catalogs".
+        """
+        cache_key = '{}.catalog.api.data'.format(self.request.site.domain)
+        cache_key = hashlib.md5(cache_key).hexdigest()
+        cached_course_catalogs = cache.get(cache_key)
+        self.assertIsNone(cached_course_catalogs)
+
+        response = get_course_catalogs(self.request.site)
+
+        self.assertEqual(len(response), len(catalog_name_list))
+        for catalog_index, catalog in enumerate(response):
+            self.assertEqual(catalog['name'], catalog_name_list[catalog_index])
+
+        cached_course = cache.get(cache_key)
+        self.assertEqual(cached_course, response)
+
+    @mock_course_catalog_api_client
+    def test_get_course_catalogs_for_single_catalog_with_id(self):
+        """
+        Verify that method "get_course_catalogs" returns proper response for a
+        single catalog by its id.
+        """
+        self.mock_course_discovery_api_for_catalog_by_resource_id()
+
+        catalog_id = 1
+        cache_key = '{}.catalog.api.data.{}'.format(self.request.site.domain, catalog_id)
+        cache_key = hashlib.md5(cache_key).hexdigest()
+        cached_course_catalog = cache.get(cache_key)
+        self.assertIsNone(cached_course_catalog)
+
+        response = get_course_catalogs(self.request.site, catalog_id)
+
+        self.assertEqual(response['count'], 1)
+        self.assertEqual(response['results'][0]['name'], 'Catalog {}'.format(catalog_id))
+
+        cached_course = cache.get(cache_key)
+        self.assertEqual(cached_course, response)
+
+        # Verify the API was actually hit (not the cache)
+        self._assert_num_requests(1)
+
+    @mock_course_catalog_api_client
+    @ddt.data(
+        ['Catalog 1'],
+        ['Catalog 1', 'Catalog 2'],
+    )
+    def test_get_course_catalogs_for_single_page_api_response(self, catalog_name_list):
+        """
+        Verify that method "get_course_catalogs" returns proper response for
+        single page Course Discovery API response and uses cache to return data
+        in case of same API request.
+        """
+        self.mock_course_discovery_api_for_catalogs(catalog_name_list)
+
+        self._assert_get_course_catalogs(catalog_name_list)
+
+        # Verify the API was hit once
+        self._assert_num_requests(1)
+
+        # Now fetch the catalogs again and there should be no more actual call
+        # to Course Discovery API as the data will be fetched from the cache
+        get_course_catalogs(self.request.site)
+        self._assert_num_requests(1)
+
+    @mock_course_catalog_api_client
+    def test_get_course_catalogs_for_paginated_api_response(self):
+        """
+        Verify that method "get_course_catalogs" returns all catalogs for
+        paginated Course Discovery API response for multiple catalogs.
+        """
+        catalog_name_list = ['Catalog 1', 'Catalog 2', 'Catalog 3']
+        self.mock_course_discovery_api_for_paginated_catalogs(catalog_name_list)
+
+        self._assert_get_course_catalogs(catalog_name_list)
+
+        # Verify the API was hit for each catalog page
+        self._assert_num_requests(len(catalog_name_list))
+
+    @mock_course_catalog_api_client
+    def test_get_course_catalogs_for_failure(self):
+        """
+        Verify that method "get_course_catalogs" raises exception in case
+        the Course Discovery API fails to return data.
+        """
+        self.mock_course_discovery_api_for_failure()
+
+        with self.assertRaises(Exception):
+            get_course_catalogs(self.request.site)
--- a/ecommerce/courses/utils.py
+++ b/ecommerce/courses/utils.py
@@ -1,4 +1,5 @@
 import hashlib
+from urlparse import parse_qs, urlparse
 
 from django.conf import settings
 from django.core.cache import cache
@@ -31,6 +32,67 @@ def get_course_info_from_catalog(site, course_key):
         course_run = api.course_runs(course_key).get(partner=partner_short_code)
         cache.set(cache_key, course_run, settings.COURSES_API_CACHE_TIMEOUT)
     return course_run
+
+
+def get_course_catalogs(site, resource_id=None):
+    """
+    Get details related to course catalogs from Catalog Service.
+
+    Arguments:
+        site (Site): Site object containing Site Configuration data
+        resource_id (int or str): Identifies a specific resource to be retrieved
+
+    Returns:
+        dict: Course catalogs received from Course Catalog API
+
+    """
+    resource = 'catalogs'
+    base_cache_key = '{}.catalog.api.data'.format(site.domain)
+
+    cache_key = '{}.{}'.format(base_cache_key, resource_id) if resource_id else base_cache_key
+    cache_key = hashlib.md5(cache_key).hexdigest()
+    cached = cache.get(cache_key)
+    if cached:
+        return cached
+
+    api = site.siteconfiguration.course_catalog_api_client
+    endpoint = getattr(api, resource)
+    response = endpoint(resource_id).get()
+
+    if resource_id:
+        results = response
+    else:
+        results = traverse_pagination(response, endpoint)
+
+    cache.set(cache_key, results, settings.COURSES_API_CACHE_TIMEOUT)
+    return results
+
+
+def traverse_pagination(response, endpoint):
+    """
+    Traverse a paginated API response.
+
+    Extracts and concatenates "results" (list of dict) returned by DRF-powered
+    APIs.
+
+    Arguments:
+        response (Dict): Current response dict from service API
+        endpoint (slumber Resource object): slumber Resource object from edx-rest-api-client
+
+    Returns:
+        list of dict.
+
+    """
+    results = response.get('results', [])
+
+    next_page = response.get('next')
+    while next_page:
+        querystring = parse_qs(urlparse(next_page).query, keep_blank_values=True)
+        response = endpoint.get(**querystring)
+        results += response.get('results', [])
+        next_page = response.get('next')
+
+    return results
 
 
 def get_certificate_type_display_value(certificate_type):
--- a/ecommerce/extensions/api/v2/views/catalog.py
+++ b/ecommerce/extensions/api/v2/views/catalog.py
@@ -13,6 +13,7 @@ from slumber.exceptions import SlumberBaseException
 from ecommerce.core.constants import DEFAULT_CATALOG_PAGE_SIZE
 from ecommerce.coupons.utils import get_range_catalog_query_results
 from ecommerce.extensions.api import serializers
+from ecommerce.courses.utils import get_course_catalogs
 
 
 Catalog = get_model('catalogue', 'Catalog')
@@ -73,3 +74,20 @@ class CatalogViewSet(NestedViewSetMixin, ReadOnlyModelViewSet):
                 logger.error('Unable to connect to Course Catalog service.')
                 return Response(status=status.HTTP_400_BAD_REQUEST)
         return Response(status=status.HTTP_400_BAD_REQUEST)
+
+    @action(is_for_list=True, methods=['get'])
+    def course_catalogs(self, request):
+        """
+        Returns response with all course catalogs in the format:
+        ["results": {"id": 1, "name": "Dummy Catalog"}]
+        """
+        try:
+            results = get_course_catalogs(site=request.site)
+        except:  # pylint: disable=bare-except
+            logger.exception('Failed to retrieve course catalogs data from the Course Discovery API.')
+            results = []
+
+        # Create catalogs list with sorting by name
+        catalogs = [{'id': catalog['id'], 'name': catalog['name']} for catalog in results]
+        data = {'results': sorted(catalogs, key=lambda catalog: catalog.get('name', '').lower())}
+        return Response(data=data)
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -7,7 +7,6 @@ from django.core.exceptions import ValidationError
 from django.db import IntegrityError, transaction
 from django.http import Http404
 from django.shortcuts import get_object_or_404
-from django.utils.translation import ugettext as _
 from oscar.core.loading import get_model
 from rest_framework import filters, generics, serializers, status, viewsets
 from rest_framework.permissions import IsAdminUser, IsAuthenticated
@@ -73,95 +72,34 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
             429 if the client has made requests at a rate exceeding that allowed by the configured rate limit.
             500 if an error occurs when attempting to create a coupon.
         """
-        category_data = request.data.get('category')
-        code = request.data.get('code')
-        course_seat_types = request.data.get('course_seat_types')
-        max_uses = request.data.get('max_uses')
-        partner = request.site.siteconfiguration.partner
-        stock_record_ids = request.data.get('stock_record_ids')
-        voucher_type = request.data.get('voucher_type')
-
         try:
             with transaction.atomic():
-                if code:
-                    try:
-                        Voucher.objects.get(code=code)
-                        return Response(
-                            'A coupon with code {code} already exists.'.format(code=code),
-                            status=status.HTTP_400_BAD_REQUEST
-                        )
-                    except Voucher.DoesNotExist:
-                        pass
-
-                if course_seat_types:
-                    try:
-                        course_seat_types = prepare_course_seat_types(course_seat_types)
-                    except (AttributeError, TypeError) as e:
-                        logger.exception('Failed to create coupon. Invalid course seat types data: %s', e.message)
-                        return Response(
-                            _('Invalid course seat types data: {}').format(e.message),
-                            status=status.HTTP_400_BAD_REQUEST
-                        )
-
                 try:
-                    category = Category.objects.get(name=category_data['name'])
-                except Category.DoesNotExist:
-                    return Response(
-                        'Category {category_name} not found.'.format(category_name=category_data['name']),
-                        status=status.HTTP_404_NOT_FOUND
-                    )
-                except (KeyError, TypeError):
-                    return Response('Invalid Coupon Category data.', status=status.HTTP_400_BAD_REQUEST)
-
-                # Maximum number of uses can be set for each voucher type and disturb
-                # the predefined behaviours of the different voucher types. Therefor
-                # here we enforce that the max_uses variable can't be used for SINGLE_USE
-                # voucher types.
-                if max_uses and voucher_type != Voucher.SINGLE_USE:
-                    max_uses = int(max_uses)
-                else:
-                    max_uses = None
-
-                # When a black-listed course mode is received raise an exception.
-                # Audit modes do not have a certificate type and therefore will raise
-                # an AttributeError exception.
-                if stock_record_ids:
-                    seats = Product.objects.filter(stockrecords__id__in=stock_record_ids)
-                    for seat in seats:
-                        try:
-                            if seat.attr.certificate_type in settings.BLACK_LIST_COUPON_COURSE_MODES:
-                                return Response('Course mode not supported', status=status.HTTP_400_BAD_REQUEST)
-                        except AttributeError:
-                            return Response('Course mode not supported', status=status.HTTP_400_BAD_REQUEST)
-
-                    stock_records_string = ' '.join(str(id) for id in stock_record_ids)
-                    coupon_catalog, __ = get_or_create_catalog(
-                        name='Catalog for stock records: {}'.format(stock_records_string),
-                        partner=partner,
-                        stock_record_ids=stock_record_ids
-                    )
-                else:
-                    coupon_catalog = None
+                    cleaned_voucher_data = self.clean_voucher_request_data(request)
+                except ValidationError as error:
+                    logger.exception('Failed to create coupon. %s', error.message)
+                    return Response(error.message, status=error.code)
 
                 try:
                     coupon_product = create_coupon_product(
-                        benefit_type=request.data.get('benefit_type'),
-                        benefit_value=request.data.get('benefit_value'),
-                        catalog=coupon_catalog,
-                        catalog_query=request.data.get('catalog_query'),
-                        category=category,
-                        code=code,
-                        course_seat_types=course_seat_types,
-                        email_domains=request.data.get('email_domains'),
-                        end_datetime=request.data.get('end_datetime'),
-                        max_uses=max_uses,
-                        note=request.data.get('note'),
-                        partner=partner,
-                        price=request.data.get('price'),
-                        quantity=request.data.get('quantity'),
-                        start_datetime=request.data.get('start_datetime'),
-                        title=request.data.get('title'),
-                        voucher_type=voucher_type
+                        benefit_type=cleaned_voucher_data.get('benefit_type'),
+                        benefit_value=cleaned_voucher_data.get('benefit_value'),
+                        catalog=cleaned_voucher_data.get('coupon_catalog'),
+                        catalog_query=cleaned_voucher_data.get('catalog_query'),
+                        category=cleaned_voucher_data.get('category'),
+                        code=cleaned_voucher_data.get('code'),
+                        course_catalog=cleaned_voucher_data.get('course_catalog'),
+                        course_seat_types=cleaned_voucher_data.get('course_seat_types'),
+                        email_domains=cleaned_voucher_data.get('email_domains'),
+                        end_datetime=cleaned_voucher_data.get('end_datetime'),
+                        max_uses=cleaned_voucher_data.get('max_uses'),
+                        note=cleaned_voucher_data.get('note'),
+                        partner=cleaned_voucher_data.get('partner'),
+                        price=cleaned_voucher_data.get('price'),
+                        quantity=cleaned_voucher_data.get('quantity'),
+                        start_datetime=cleaned_voucher_data.get('start_datetime'),
+                        title=cleaned_voucher_data.get('title'),
+                        voucher_type=cleaned_voucher_data.get('voucher_type')
                     )
                 except IntegrityError as e:
                     return Response(str(e), status=status.HTTP_400_BAD_REQUEST)
@@ -178,6 +116,106 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
                 return Response(response_data, status=status.HTTP_200_OK)
         except ValidationError as e:
             raise serializers.ValidationError(e.message)
+
+    @classmethod
+    def clean_voucher_request_data(cls, request):
+        """
+        Helper method to return cleaned request data for voucher creation or
+        raise validation error with error code.
+
+        Arguments:
+            request (HttpRequest): request with voucher data
+
+        """
+        category_data = request.data.get('category')
+        code = request.data.get('code')
+        course_catalog_data = request.data.get('course_catalog')
+        course_seat_types = request.data.get('course_seat_types')
+        max_uses = request.data.get('max_uses')
+        partner = request.site.siteconfiguration.partner
+        stock_record_ids = request.data.get('stock_record_ids')
+        voucher_type = request.data.get('voucher_type')
+
+        if code and Voucher.does_exist(code):
+            return Response(
+                'A coupon with code {code} already exists.'.format(code=code),
+                status=status.HTTP_400_BAD_REQUEST
+            )
+
+        if course_seat_types:
+            try:
+                course_seat_types = prepare_course_seat_types(course_seat_types)
+            except (AttributeError, TypeError) as exception:
+                validation_message = 'Invalid course seat types data: {}'.format(exception.message)
+                raise ValidationError(validation_message, code=status.HTTP_400_BAD_REQUEST)
+
+        try:
+            category = Category.objects.get(name=category_data['name'])
+        except Category.DoesNotExist:
+            validation_message = 'Category "{category_name}" not found.'.format(category_name=category_data['name'])
+            raise ValidationError(validation_message, code=status.HTTP_404_NOT_FOUND)
+        except (KeyError, TypeError):
+            validation_message = 'Invalid Coupon Category data.'
+            raise ValidationError(validation_message, code=status.HTTP_400_BAD_REQUEST)
+
+        try:
+            course_catalog = course_catalog_data['id']
+        except (KeyError, TypeError):
+            validation_message = 'Unexpected catalog data format received for coupon.'
+            raise ValidationError(validation_message, code=status.HTTP_400_BAD_REQUEST)
+
+        # Maximum number of uses can be set for each voucher type and disturb
+        # the predefined behaviours of the different voucher types. Therefor
+        # here we enforce that the max_uses variable can't be used for SINGLE_USE
+        # voucher types.
+        if max_uses and voucher_type != Voucher.SINGLE_USE:
+            max_uses = int(max_uses)
+        else:
+            max_uses = None
+
+        # When a black-listed course mode is received raise an exception.
+        # Audit modes do not have a certificate type and therefore will raise
+        # an AttributeError exception.
+        if stock_record_ids:
+            seats = Product.objects.filter(stockrecords__id__in=stock_record_ids)
+            for seat in seats:
+                try:
+                    if seat.attr.certificate_type in settings.BLACK_LIST_COUPON_COURSE_MODES:
+                        validation_message = 'Course mode not supported'
+                        raise ValidationError(validation_message, code=status.HTTP_400_BAD_REQUEST)
+                except AttributeError:
+                    validation_message = 'Course mode not supported'
+                    raise ValidationError(validation_message, code=status.HTTP_400_BAD_REQUEST)
+
+            stock_records_string = ' '.join(str(id) for id in stock_record_ids)
+            coupon_catalog, __ = get_or_create_catalog(
+                name='Catalog for stock records: {}'.format(stock_records_string),
+                partner=partner,
+                stock_record_ids=stock_record_ids
+            )
+        else:
+            coupon_catalog = None
+
+        return {
+            'benefit_type': request.data.get('benefit_type'),
+            'benefit_value': request.data.get('benefit_value'),
+            'catalog': coupon_catalog,
+            'catalog_query': request.data.get('catalog_query'),
+            'category': category,
+            'code': code,
+            'course_catalog': course_catalog,
+            'course_seat_types': course_seat_types,
+            'email_domains': request.data.get('email_domains'),
+            'end_datetime': request.data.get('end_datetime'),
+            'max_uses': max_uses,
+            'note': request.data.get('note'),
+            'partner': partner,
+            'price': request.data.get('price'),
+            'quantity': request.data.get('quantity'),
+            'start_datetime': request.data.get('start_datetime'),
+            'title': request.data.get('title'),
+            'voucher_type': voucher_type,
+        }
 
     def create_order_for_invoice(self, basket, coupon_id, client, invoice_data=None):
         """Creates an order from the basket and invokes the invoice payment processor."""
@@ -240,6 +278,13 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
             # Remove catalog if switching from single course to dynamic query
             if voucher_range.catalog:
                 range_data['catalog'] = None
+
+            course_catalog_data = request.data.get('course_catalog')
+            if course_catalog_data:
+                # make sure course catalog is None if its empty
+                course_catalog = course_catalog_data.get('id') if course_catalog_data.get('id') else None
+                range_data['course_catalog'] = course_catalog
+
             Range.objects.filter(id=voucher_range.id).update(**range_data)
 
         benefit_value = request.data.get('benefit_value')
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -119,10 +119,16 @@ class Range(AbstractRange):
     UPDATABLE_RANGE_FIELDS = [
         'catalog_query',
         'course_seat_types',
+        'course_catalog',
     ]
     ALLOWED_SEAT_TYPES = ['credit', 'professional', 'verified']
     catalog = models.ForeignKey('catalogue.Catalog', blank=True, null=True, related_name='ranges')
     catalog_query = models.TextField(blank=True, null=True)
+    course_catalog = models.PositiveIntegerField(
+        help_text=_('Course catalog id from the Catalog Service.'),
+        null=True,
+        blank=True
+    )
     course_seat_types = models.CharField(
         max_length=255,
         validators=[validate_credit_seat_type],
--- a/ecommerce/extensions/voucher/tests/test_utils.py
+++ b/ecommerce/extensions/voucher/tests/test_utils.py
@@ -137,6 +137,13 @@ class UtilTests(CouponMixin, CourseCatalogMockMixin, CourseCatalogTestMixin, Lms
             course_seat_types=course_seat_types
         )
 
+    def create_course_catalog_coupon(self, coupon_title, quantity, course_catalog):
+        return self.create_coupon(
+            title=coupon_title,
+            quantity=quantity,
+            course_catalog=course_catalog,
+        )
+
     def use_voucher(self, order_num, voucher, user):
         """
         Mark voucher as used by provided users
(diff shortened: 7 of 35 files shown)


ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,55 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+    no_data = []
+    resource = 'catalogs'
+    api = site.siteconfiguration.course_catalog_api_client
+    try:
+        endpoint = getattr(api, resource)
+        resource_id = None
+        querystring = {}
+        response = endpoint(resource_id).get(**querystring)
+
+        if resource_id:
```
saleem-latif: wouldn't `resource_id` alway be `None` ?

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -73,3 +73,18 @@ def preview(self, request):
...
                 return Response(status=status.HTTP_400_BAD_REQUEST)
         return Response(status=status.HTTP_400_BAD_REQUEST)
+
+    @action(is_for_list=True, methods=['get'])
+    def all_catalogs(self, request):
+        """
+        Get all course catalogs.
+        """
+        try:
+            results = get_all_course_catalogs(site=request.site)
```
saleem-latif: I think it would a good idea to add caching here, (We also discussed this in person)

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -601,6 +630,31 @@ define([
...
             },
 
+            fetchCourseCatalogsData: function() {
+                this.limit = 10;
+                this.offset = 0;
+
+                Backbone.ajax({
+                    context: this,
+                    type: 'GET',
+                    url: window.location.origin + '/api/v2/catalogs/all_catalogs/',
```
saleem-latif: why are we using `window.location.origin + '/api/v2/catalogs/all_catalogs/'` instead of using absolute url path `'/api/v2/catalogs/all_catalogs/'` ?

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -601,6 +630,31 @@ define([
...
+                    success: this.onSuccess
+                });
+            },
+
+            onSuccess: function(data) {
+                var courseCatalogElement = this.$('#course-catalog'),
+                    option;
+
+                data.forEach(function (catalog) {
+                    option = document.createElement('option');
```
saleem-latif: nit: I think it would be readable to create `option` element using jquery.
```
option = $("<option></option>").val(catalog.id).text(catalog.name)
```

saleem-latif (commented):
@zubair-arbi Some changes after going through the code.

ecommerce/static/templates/coupon_form.html
```diff
@@ -204,6 +206,11 @@
             </div>
             <div class="catalog_buttons"></div>
         </div>
+        <div class="form-group course-catalog">
+            <label for="course-catalog"><%= gettext('Select from course catalogs:') %> *</label>
```
asadiqbal08, line 210: mark for translations  ?

ecommerce/static/templates/coupon_form.html
```diff
@@ -164,6 +164,8 @@
                 <label for="single-course"><%= gettext('Single course') %></label>
                 <input id="multiple-courses" type="radio" name="catalog_type" value="Multiple courses">
                 <label for="multiple-courses"><%= gettext('Multiple courses') %></label>
+                <input id="catalog" type="radio" name="catalog_type" value="Course catalog">
+                <label for="catalog"><%= gettext('Course catalog') %></label>
```
asadiqbal08: mark for translation.

ecommerce/coupons/utils.py
```diff
@@ -1,10 +1,13 @@
 """ Coupon related utility functions. """
 import hashlib
+import logging
 
 from django.conf import settings
 from django.core.cache import cache
 from oscar.core.loading import get_model
 
+
+log = logging.getLogger(__name__)
```
asadiqbal08: nit: LOGGER instead of log.

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,55 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+        else:
+            results = _traverse_pagination(response, endpoint, querystring, no_data)
+    except:  # pylint: disable=bare-except
+        log.exception('Failed to retrieve data from the Discovery API.')
+        return no_data
+
+    return results
+
+
+def _traverse_pagination(response, endpoint, querystring, no_data):
```
asadiqbal08: what represent `endpoint` , `querystring` here? should include description in method docstring.

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -73,3 +73,18 @@ def preview(self, request):
                 logger.error('Unable to connect to Course Catalog service.')
                 return Response(status=status.HTTP_400_BAD_REQUEST)
         return Response(status=status.HTTP_400_BAD_REQUEST)
+
+    @action(is_for_list=True, methods=['get'])
+    def all_catalogs(self, request):
```
asadiqbal08: it is returning catalogs. Name should be descriptive. get_all_catalogs ?

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -73,3 +73,18 @@ def preview(self, request):
...
+        """
+        Get all course catalogs.
```
asadiqbal08: nit: return response with all course catalogs.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -152,6 +152,7 @@ def create(self, request, *args, **kwargs):
                         catalog_query=request.data.get('catalog_query'),
                         category=category,
                         code=code,
+                        course_catalog=request.data.get('catalog'),
```
asadiqbal08: is `catalog` here referring to only one catalog ?

ecommerce/static/js/models/coupon_model.js
```diff
@@ -57,6 +57,11 @@ define([
                         return this.get('catalog_type') === 'Multiple courses';
                     }
                 },
+                course_catalog: {
+                    required: function () {
+                        return this.get('catalog_type') === 'Course catalog';
```
asadiqbal08: can we use JS dict / hash here that can contain the course types in it  instead of matching like that ? 
e.g. 
CATALOG_TYPES = {
  discount_code: 'Discount code', 
  . 
  . 
}

ecommerce/static/js/models/coupon_model.js
```diff
@@ -185,9 +190,16 @@ define([
 
             updateSeatData: function () {
                 var seat_data,
-                    seats = this.get('seats');
+                    seats = this.get('seats'),
+                    catalog_type;
 
-                this.set('catalog_type', this.has('catalog_query') ? 'Multiple courses': 'Single course');
+                if (this.has('catalog_query')) {
```
asadiqbal08: better to have `course_catalog_query` name here .

ecommerce/coupons/utils.py
mattdrayer: Likewise for `querystring` -- it would always be `{}`

ecommerce/coupons/utils.py
mattdrayer: While this method needs some tuning, it seems like a useful piece of functionality that could be added to https://github.com/edx/edx-rest-api-client.  This way other workflows could leverage the same capability.

mattdrayer: +1 on that

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,55 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+    """Traverse a paginated API response.
+
+    Extracts and concatenates "results" (list of dict) returned by DRF-powered APIs.
+    """
+    results = response.get('results', no_data)
```
mattdrayer: I guess I don't really see the point of passing around this `no_data` container -- is the re-use just an attempt to minimize the memory footprint, or is there a logical reason to share this structure that I am not seeing?

ecommerce/extensions/api/v2/views/catalog.py
mattdrayer: if this is a wrapper around `get_all_course_catalogs` then I would name it the same and reference the imported method using its module namespace.

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -73,3 +73,18 @@ def preview(self, request):
...
+        """
+        try:
+            results = get_all_course_catalogs(site=request.site)
+        except (ConnectionError, SlumberBaseException, Timeout):
+            logger.error('Unable to connect to Course Catalog service.')
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+
+        # Order catalogs by name
+        catalogs = sorted(results, key=lambda catalog: catalog.get('name', '').lower())
```
mattdrayer: it would be nice if we could send the sort order to `get_all_course_catalogs`, instead of doing it here.

ecommerce/extensions/offer/models.py
```diff
@@ -121,6 +121,7 @@ class Range(AbstractRange):
         'course_seat_types',
     ]
     ALLOWED_SEAT_TYPES = ['credit', 'professional', 'verified']
+    course_catalog = models.PositiveIntegerField(help_text=_('Course catalog id.'), null=True, blank=True)
```
mattdrayer: Wait, I'm confused -- we are adding `course_catalog`, but there is already a `catalog` -- what is the difference between these two fields?  We will need to be very clear about the use for each.

ecommerce/extensions/voucher/utils.py
```diff
@@ -393,30 +393,38 @@ def create_vouchers(
...
-            coupon (Coupon): Coupon entity associated with vouchers.
-            end_datetime (datetime): End date for voucher offer.
-            name (str): Voucher name.
-            quantity (int): Number of vouchers to be created.
-            start_datetime (datetime): Start date for voucher offer.
-            voucher_type (str): Type of voucher.
-            code (str): Code associated with vouchers. Defaults to None.
-            email_domains (str): List of email domains to restrict coupons. Defaults to None.
+    Arguments:
+        benefit_type (str): Type of benefit associated with vouchers.
```
mattdrayer, line 428: It seems this list follows the order of the arguments in the method signature.  I'm not sure if that is a standard/convention, but IMHO we should instead alphabetize this list.

ecommerce/static/js/views/coupon_form_view.js
mattdrayer: I actually think the current proposal reads easier than the jQuery version, so I'm okay keeping it as-is.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -641,6 +694,7 @@ define([
                     this.fillFromCourse();
                 } else {
                     this.model.set({
+                        'course_catalog': this.catalogData[0].value,
```
mattdrayer: Do we need to protect against an empty array here?

ecommerce/coupons/utils.py
zubair-arbi: @asadiqbal08 there are two different conventions implemented in ecom for the logging variable name these are `log` and `logger`. Also the [python docs also gives example of using `logger` as variable name](https://docs.python.org/2/howto/logging.html#advanced-logging-tutorial).
I will  change the name to `logger`.

ecommerce/coupons/utils.py
zubair-arbi: @mattdrayer I agree that this is very useful functionality and in LMS we are already using it for `programs`, `credentials`, `catalogs` etc.
Would it make sense to have a separate story to add this functionality in `edx-rest-api-client` and update this code once we have new version available?

ecommerce/coupons/utils.py
zubair-arbi: @mattdrayer I am reusing the code from LMS. I believe the idea is to have a default in case the API we are using, does not provide the desired format causing exception or don't return anything.

ecommerce/extensions/api/v2/views/catalog.py
zubair-arbi: @mattdrayer Since this name become part of url so we cannot name it `get_all_course_catalogs`. Also I think we can change it to `course_catalogs`, this way the url will become: `/api/v2/catalogs/course_catalogs/`

ecommerce/extensions/api/v2/views/coupons.py
zubair-arbi: yes. I am updating it to `course_catalog`.

ecommerce/extensions/offer/models.py
zubair-arbi: @mattdrayer The thing is that while creating the coupon the admin can provide a catalog query by selecting the `Multiple courses` option on `Coupon Admin Tool` page. Now interesting thing is that the catalog query is saved in the `Range` model instead of creating a catalog in the `discovery` service.

In our case we are managing the course catalogs provided by the `discovery`. So we had 2 possible solutions:

1. When admin will select a course catalog from discovery in CAT tool, use the existing `catalog_query` to save the query against the catalog. But then while viewing the coupon details and editing a coupon we will just have a query for a course catalog but we will not be able to tell which catalog (name of the catalog) it was in the discovery and this will be become problematic in case there are multiple catalog in discovery with same query or if the catalog query changes in the `discovery` service.
2. Add a new field `course_catalog` and just save the catalog id from the discovery service and then we wont have to worry if the name or the query change in the `discovery` service we will always be able to fetch fresh data from `discovery` this will increase the inter-services API calls but gives us flexibility. 

After some thinking we started following the second solution. Please let me know if this make sense or we can have a better approach.

ecommerce/static/templates/coupon_form.html
zubair-arbi: this is the default format for translation.

ecommerce/static/templates/coupon_form.html
zubair-arbi: already marked. `gettext` handles it.

ecommerce/static/js/models/coupon_model.js
zubair-arbi: this is existing form field so don't want to change it.

ecommerce/extensions/voucher/utils.py
zubair-arbi: It looks like we are following your suggested pattern in some other place in ecommerce and I think this would also  increase the readability so I will make the changes.

zubair-arbi:
@saleem-latif @mattdrayer @asadiqbal08 I have addressed almost all of your feedback. Please take another look.

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,66 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+        resource_id (int or str): Identifies a specific resource to be retrieved
+
+    Returns:
+        dict: Course catalogs received from Course Catalog API
+    """
+    no_data = []
+    resource = 'catalogs'
+    cache_key = 'catalog.api.data'
+
+    if cache_key:
```
mattdrayer: It seems like there will always be a value for `cache_key` here.

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,66 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+        cache_key = '{}.{}'.format(cache_key, resource_id) if resource_id else cache_key
+
+        cached = cache.get(cache_key)
```
mattdrayer: Caching and cache control is another area that we should consider moving into the https://github.com/edx/edx-rest-api-client library.  Caching should be invisible to the service, although it should be possible to configure aspects of the caching mechanism (eg, TTL) in order to maintain control over performance and resource utilization.

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,66 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+
+        if resource_id:
+            results = response
+        else:
+            results = _traverse_pagination(response, endpoint, no_data)
+    except:  # pylint: disable=bare-except
+        logger.exception('Failed to retrieve data from the Discovery API.')
+        return no_data
+
+    if cache_key:
```
mattdrayer: As with above, it appears there will always be a value for cache_key

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,66 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
     return response
 
 
+def get_course_catalogs(site, resource_id=None):
```
mattdrayer: This entire function seems like it should be moved into a "catalog service helpers" module.

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,66 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+        logger.exception('Failed to retrieve data from the Discovery API.')
+        return no_data
+
+    if cache_key:
+        cache.set(cache_key, results, settings.COURSES_API_CACHE_TIMEOUT)
+
+    return results
+
+
+def _traverse_pagination(response, endpoint, no_data):
```
mattdrayer: If `get_course_catalogs` is moved into a "helpers" module, this should accompany it.

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -73,3 +73,21 @@ def preview(self, request):
...
+
+    @action(is_for_list=True, methods=['get'])
+    def course_catalogs(self, request):
+        """
+        Returns response with all course catalogs in the format:
+        ["results": {"id": 1, "name": "Dummy Catalog"}]
+        """
+        try:
+            results = get_course_catalogs(site=request.site)
+        except (ConnectionError, SlumberBaseException, Timeout):
```
mattdrayer: It looks like this except may not be hit because `get_course_catalogs` has a catch-all `except` handler that swallows all exceptions and returns `no_data`

ecommerce/extensions/offer/models.py
mattdrayer: Thank you for the explanation -- the second approach described above is the proper solution.

ecommerce/extensions/voucher/utils.py
```diff
@@ -393,30 +393,37 @@ def create_vouchers(
         start_datetime,
         voucher_type,
         code=None,
-        max_uses=None,
-        _range=None,
         catalog_query=None,
+        course_catalog=None,
```
mattdrayer: Is it safe to reorder these arguments?  We need to be careful when modifying method signatures.

ecommerce/extensions/voucher/utils.py
```diff
@@ -393,30 +393,37 @@ def create_vouchers(
...
+        coupon (Coupon): Coupon entity associated with vouchers.
+        course_catalog (int): Course catalog id from discovery service. Defaults to None.
+        course_seat_types (str): Comma-separated list of course seat types.
+        email_domains (str): List of email domains to restrict coupons. Defaults to None.
+        end_datetime (datetime): End date for voucher offer.
+        name (str): Voucher name.
+        quantity (int): Number of vouchers to be created.
+        start_datetime (datetime): Start date for voucher offer.
+        voucher_type (str): Type of voucher.
+        max_uses (int): Number of Voucher max uses. Defaults to None.
```
mattdrayer: Looks like `max_uses` and `_range` missed out on the alphabetization party

ecommerce/static/js/models/coupon_model.js
```diff
@@ -185,9 +193,17 @@ define([
...
+                    catalog_type = 'Multiple courses'
+                } else if (this.has('course_catalog') && this.get('course_catalog').id != '') {
+                    catalog_type = 'Catalog course'
```
mattdrayer: I think this should just be "Catalog"

ecommerce/static/templates/coupon_form.html
mattdrayer: Let's be sure we are consistent with capitalization -- above we have `Course Catalog`

zubair-arbi: @mattdrayer existing values have only first word in camel case, e.g, `Single course`, `Multiple courses`.

ecommerce/static/js/models/coupon_model.js
zubair-arbi: I will update this

ecommerce/extensions/voucher/utils.py
zubair-arbi: Aaah, I will fix it.

ecommerce/coupons/utils.py
zubair-arbi: @mattdrayer We don't have any catalog service app available in the Ecom, so should we create `catalog_service_helpers.py` in `coupons` app?

ecommerce/extensions/voucher/utils.py
zubair-arbi: I am not sure, I will revert the change to method signature.

ecommerce/extensions/voucher/utils.py
zubair-arbi: Oho, I have updated the order for `max_uses` but I am not sure where to put `_range` in this order.
[This site](http://alphabetizer.flap.tv) suggests moving it at the top of list but it seems more appropriate to put it at the end.

ecommerce/static/js/models/coupon_model.js
zubair-arbi: Sure, I am going to add a dict:
```javascriptcatalogTypes: {
    single_course: 'Single course',
    multiple_courses: 'Multiple courses',
    catalog: 'Catalog'
},
```

ecommerce/static/js/models/coupon_model.js
zubair-arbi: I will update the field

ecommerce/coupons/utils.py
mattdrayer: I'm okay doing that to start, although it seems like a better place would be a `utils` package.

mattdrayer: Actually I would prefer introducing a `services` package, but I haven't seen that convention anywhere else.  It makes sense to me, however.

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,62 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+
+        if resource_id:
+            results = response
+        else:
+            results = _traverse_pagination(response, endpoint, no_data)
+    except:  # pylint: disable=bare-except
+        logger.exception('Failed to retrieve data from the Discovery API.')
+        return no_data
+
+    cache.set(cache_key, results, settings.COURSES_API_CACHE_TIMEOUT)
```
mattdrayer: We should introduce a new setting for the catalog service, which I think is different (at least semantically) from the "courses API" -- eg `CATALOG_API_CACHE_TIMEOUT`

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -114,6 +115,11 @@ def create(self, request, *args, **kwargs):
                 except (KeyError, TypeError):
                     return Response('Invalid Coupon Category data.', status=status.HTTP_400_BAD_REQUEST)
 
+                try:
+                    course_catalog = course_catalog_data['id']
+                except (KeyError, TypeError):
+                    return Response('Invalid coupon Course Catalog data.', status=status.HTTP_400_BAD_REQUEST)
```
mattdrayer: "Unexpected catalog data format received for coupon."

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -241,6 +248,11 @@ def update(self, request, *args, **kwargs):
             # Remove catalog if switching from single course to dynamic query
             if voucher_range.catalog:
                 range_data['catalog'] = None
+
+            course_catalog_data = request.data.get('course_catalog')
+            if course_catalog_data:
+                range_data['course_catalog'] = course_catalog_data.get('id')
```
mattdrayer: Do we care about an unexpected catalog data schema here, as on line 119?

ecommerce/extensions/offer/models.py
```diff
@@ -119,10 +119,12 @@ class Range(AbstractRange):
     UPDATABLE_RANGE_FIELDS = [
         'catalog_query',
         'course_seat_types',
+        'course_catalog',
     ]
     ALLOWED_SEAT_TYPES = ['credit', 'professional', 'verified']
     catalog = models.ForeignKey('catalogue.Catalog', blank=True, null=True, related_name='ranges')
     catalog_query = models.TextField(blank=True, null=True)
+    course_catalog = models.PositiveIntegerField(help_text=_('Course catalog id.'), null=True, blank=True)
```
mattdrayer: Maybe we can add something to the help text about being from the Catalog/Discovery service.

mattdrayer:
Looking good @zubair-arbi!  I think we are almost there with this PR.  Just a few more items to address.

ecommerce/extensions/offer/models.py
zubair-arbi: @mattdrayer Can you please suggest some text? I am thinking of `Course catalog id from catalog service`.

ecommerce/coupons/utils.py
zubair-arbi: Where do you think we be best place for that package? At the same level of main apps, e.g. `core`, `coupons` etc or in `ecommerce/extensions`? I believe `extensions` dir is for overriding oscar apps so we can create at main apps level.

ecommerce/coupons/utils.py
zubair-arbi: Would it be sufficient to set the default value at `3600  # Value is in seconds`?
If we add this new settings variable then we will also need to make sure that the devops can update this setting through configuration.

ecommerce/extensions/offer/models.py
mattdrayer: That works for me, although I would capitalize "Catalog Service"

ecommerce/coupons/utils.py
mattdrayer: Yes, that is the default value for the COURSES API setting, so I think it makes sense.  Agreed that we should add support for this setting to https://github.com/edx/configuration/blob/master/playbooks/roles/ecommerce/defaults/main.yml but I do not see anything in that repository for `COURSES_API_CACHE_TIMEOUT` (see https://github.com/edx/configuration/search?utf8=%E2%9C%93&q=COURSES_API_CACHE_TIMEOUT)

mattdrayer: @edx/devops  FYI ^^ -- do you have a recommendation for what to do here?

mattdrayer:
@vkaracic FYI on this PR

ecommerce/static/templates/coupon_form.html
asadiqbal08: ah. I forgot it.

ecommerce/static/js/collections/catalog_collection.js
```diff
@@ -0,0 +1,15 @@
+define([
+        'collections/paginated_collection',
+        'models/catalog_model'
+    ],
+    function (PaginatedCollection,
+              Catalog) {
```
asadiqbal08, line 6: why breaking the line here ?

ecommerce/static/js/collections/catalog_collection.js
```diff
@@ -0,0 +1,15 @@
...
+        'use strict';
+
+        return PaginatedCollection.extend({
+                model: Catalog,
```
asadiqbal08, line 10: Is the collection here keeping the list of catalog models ?

ecommerce/static/js/collections/catalog_collection.js
asadiqbal08: would it be `catalog` or `catalogs` ?

ecommerce/static/js/models/catalog_model.js
```diff
@@ -0,0 +1,11 @@
+define([
+        'backbone.relational'
+    ],
+    function () {
+        'use strict';
+
+        return Backbone.RelationalModel.extend({
```
asadiqbal08, line 7: you are making a relational model here. can you specify to which model it is related ?

ecommerce/extensions/api/v2/views/coupons.py
zubair-arbi: I don't think we need same logic here, since on coupon creation we are doing all stuff in `transaction.atomic` so any error will roll back the transaction. But here we are only updating individual fields so it would make sense to let the exception raise and this is the current behavior too.

ecommerce/static/js/collections/catalog_collection.js
zubair-arbi: 1. We are breaking line after each param and this is convention we are following.
2. It would be Catalog (as its a model not multiple objects of a model)

ecommerce/static/js/collections/catalog_collection.js
zubair-arbi: That's correct.

ecommerce/static/js/models/catalog_model.js
zubair-arbi: Catalog. The data will be fetched from `course_catalogs` api.

ecommerce/static/js/collections/catalog_collection.js
asadiqbal08: ok. I meant, collection contains the multiple objects of model

asadiqbal08:
Beside some @mattdrayer comments and jenkins status. changes overall LGTM 👍

ecommerce/static/templates/coupon_detail.html
```diff
@@ -50,6 +50,12 @@ <h3 class="hd-3">
             <div class="value"><%= coupon.catalog_query %></div>
         </div>
         <%}%>
+        <% if(courseCatalogName) {%>
+        <div class="info-item grid-item catalog-query">
```
saleem-latif: nit change `catalog-query` to something like `catalog-name`

ecommerce/coupons/utils.py
mattdrayer: I think the main apps level is the proper place for this new package.  You are correct in that `extensions` is intended for overriding the base oscar apps.

ecommerce/static/js/models/catalog_model.js
mattdrayer: If it's not clear we could add a code comment.

mattdrayer:
So very very close!  If we can address the refactoring concern above (services package), rebase, and see some green checks, I'm a 👍 .  It would be great if @vkaracic could take a spin through as well.

zubair-arbi:
@mattdrayer I have added `services` package, Please let me know about your thoughts on this.

ecommerce/services/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+Ecommerce support for edX services.
+
+This package will be used as a wrapper for interacting with the edX services
```
mattdrayer: Here I would say "This package contains clients and related tools for communicating with other Open edX services."

ecommerce/settings/base.py
```diff
@@ -282,6 +282,8 @@
 
     # Sailthru email marketing integration
     'ecommerce.sailthru',
+
+    'ecommerce.services',
```
mattdrayer: This is interesting -- I didn't realize we had to declare this package as a "local app" -- I guess I don't have a problem with doing so, more just a TIL (Today I Learned) thing.

mattdrayer:
@zubair-arbi -- overall this is what I had in mind.  Just one small change request, and we'll need to add/move pertinent tests as well.

ecommerce/services/course_discovery.py
```diff
@@ -0,0 +1,50 @@
...
+
+        if resource_id:
+            results = response
+        else:
+            results = traverse_pagination(response, endpoint, no_data)
+    except:  # pylint: disable=bare-except
+        logger.exception('Failed to retrieve data from the Course Discovery API.')
+        return no_data
+
+    cache.set(cache_key, results, settings.CATALOG_API_CACHE_TIMEOUT)
```
jibsheet: @mattdrayer I'm not sure what devops input you were looking for here.
As long as it follows the patterns set out elsewhere in ecommerce (which @clintonb recently cleaned up) I believe it will be fine.

Remember that ecommerce does not currently have a memcache deployed (because it was never requested).  There's a ticket to build and deploy one, but until then, this will use whatever the django default cache is for things (per process I believe) - so be certain your code will perform in that use case.

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -73,3 +74,17 @@ def preview(self, request):
...
+        """
+        Returns response with all course catalogs in the format:
+        ["results": {"id": 1, "name": "Dummy Catalog"}]
+        """
+        results = get_course_catalogs(site=request.site)
+
+        # Create catalogs list with sorting by name
+        catalogs = [{'id': catalog['id'], 'name': catalog['name']} for catalog in results]
+        sorted_catalogs = sorted(catalogs, key=lambda catalog: catalog.get('name', '').lower())
+        data = {'results': sorted_catalogs}
```
vkaracic: Nit: You could just add the dict directly to the response, no need for initializing another variable.

ecommerce/services/utils.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Traverse a paginated API response.
+
+    Extracts and concatenates "results" (list of dict) returned by DRF-powered
+    APIs.
+
+    Arguments:
+        response (Dict): Current response dict from service API
+        endpoint (slumber Resource object): slumber Resource object from edx-rest-api-client
+        no_data (list): Response in case of empty API result
```
vkaracic: Is this parameter relevant? When/why will ``no_data`` be anything than an empty list?

ecommerce/services/tests/test_course_discovery.py
```diff
@@ -0,0 +1,108 @@
...
+        self.assertIsNone(cached_course_catalogs)
+
+        response = get_course_catalogs(self.request.site)
+
+        self.assertEqual(len(response), len(catalog_name_list))
+        self.assertEqual(response[0]['name'], catalog_name_list[0])
+        self.assertEqual(response[1]['name'], catalog_name_list[1])
+
+        cached_course = cache.get(cache_key)
+        self.assertEqual(cached_course, response)
```
vkaracic: These two tests, this one and the one above, are pretty much the same. You could combine them with ``ddt``.

ecommerce/services/tests/test_course_discovery.py
```diff
@@ -0,0 +1,108 @@
...
+
+        response = get_course_catalogs(self.request.site)
+
+        self.assertEqual(len(response), len(catalog_name_list))
+        self.assertEqual(response[0]['name'], catalog_name_list[0])
+        self.assertEqual(response[1]['name'], catalog_name_list[1])
+        self.assertEqual(response[2]['name'], catalog_name_list[2])
+
+        cached_course = cache.get(cache_key)
+        self.assertEqual(cached_course, response)
```
vkaracic: You could DRY these lines. Factor out lines 96-98 to a ``assert_empty_cache(cache_key)`` method and the rest of the code below that to another method, maybe ``verify_catalog_response()``.

ecommerce/static/js/models/coupon_model.js
```diff
@@ -28,11 +30,18 @@ define([
         });
         _.extend(Backbone.Model.prototype, Backbone.Validation.mixin);
 
+        var CATALOG_TYPES = {
```
vkaracic, line 34: Out of curiosity, have you tried using ``const``?

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -629,6 +669,21 @@ define([
                             this.$('.non-credit-seats').addClass(this.hiddenClass);
                         }
                     }
+
+                    if (this.model.has('course_catalog')) {
+                        if (typeof this.model.get('course_catalog') === 'number') {
+                            catalogId = this.model.get('course_catalog');
+                        } else if (!$.isEmptyObject(this.model.get('course_catalog'))) {
```
vkaracic, line 677: I'm curious, when will ``course_catalog`` be an empty object?

vkaracic (commented):
Mostly questions and nits. Overall looks good.

ecommerce/services/utils.py
zubair-arbi: For our current implementation the default response `no_data` is always set to empty list. This is to ensure the default value remains same in case we get empty response while traversing the paginated API responses.

ecommerce/static/js/models/coupon_model.js
zubair-arbi: Nope. Nice suggestion to use [`const`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/const). I will make the change.

ecommerce/static/js/views/coupon_form_view.js
zubair-arbi: If there are not catalogs and default empty value is selected.

ecommerce/services/utils.py
vkaracic: I get that, but I don't see when will this ever be anything than an empty list.

zubair-arbi: You are right, So what do you suggest?

vkaracic: Just remove the ``no_data`` parameter and change line 29 here to user ``[]`` instead of ``no_data``.

mattdrayer:
Welcome back, @mjfrey!  Here's the PR we've discussed -- please feel free to take a walk thru.

ecommerce/coupons/tests/mixins.py
```diff
@@ -149,20 +149,26 @@ def coupon_product_class(self):
...
         """Helper method for creating a coupon.
 
         Arguments:
             benefit_type(str): The voucher benefit type
             benefit_value(int): The voucher benefit value
             catalog(Catalog): Catalog of courses for which the coupon applies
             catalog_query(str): Course query string
+            client (BusinessClient):  Optional business client object
             code(str): Custom coupon code
+            course_catalog (int): Course catalog id from discovery service
```
mjfrey: Course catalog id from Catalog Service

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -13,6 +13,7 @@
 from ecommerce.core.constants import DEFAULT_CATALOG_PAGE_SIZE
 from ecommerce.coupons.utils import get_range_catalog_query_results
 from ecommerce.extensions.api import serializers
+from ecommerce.services.course_discovery import get_course_catalogs
```
mjfrey: No need to put this in a new app.  We already call the catalog service from courses/utils.  I would re-locate this there.

ecommerce/services/course_discovery.py
```diff
@@ -0,0 +1,51 @@
+"""
```
mjfrey: move this function to courses/utils.py

ecommerce/services/course_discovery.py
```diff
@@ -0,0 +1,51 @@
...
+    cache_key = '{}.{}'.format(base_cache_key, resource_id) if resource_id else base_cache_key
+    cache_key = hashlib.md5(cache_key).hexdigest()
+    cached = cache.get(cache_key)
+    if cached:
+        return cached
+
+    api = site.siteconfiguration.course_catalog_api_client
+    try:
+        endpoint = getattr(api, resource)
+        response = endpoint(resource_id).get()
```
mjfrey: Are there permissions we need to care about?  For example this is pulling all catalogs for a site.

mjfrey: Do catalogs contains course_runs or courses?  Coupons / Vouchers expect course_runs so we should be careful.  I also do not see the logic that checks for this upon redemption.  Is that not here yet?

mjfrey (changes requested):
A quick first pass.

ecommerce/services/course_discovery.py
zubair-arbi: @mjfrey We need to get all catalogs for a site.
We have a separate story to update the redemption logic to handle course runs based on the course catalog id.

zubair-arbi:
@mattdrayer Hurray the build is green now :)
@mjfrey I have addressed your feedback, Please have another look.

ecommerce/courses/utils.py
```diff
@@ -33,6 +38,71 @@ def get_course_info_from_catalog(site, course_key):
...
+        resource_id (int or str): Identifies a specific resource to be retrieved
+
+    Returns:
+        dict: Course catalogs received from Course Catalog API
+
+    """
+    resource = 'catalogs'
+    base_cache_key = 'catalog.api.data'
+
+    cache_key = '{}.{}'.format(base_cache_key, resource_id) if resource_id else base_cache_key
```
mjfrey, line 52: I think we want to add the site in the key so we do not get a cached version for a different site.

ecommerce/courses/utils.py
```diff
@@ -33,6 +38,71 @@ def get_course_info_from_catalog(site, course_key):
...
+    try:
+        endpoint = getattr(api, resource)
+        response = endpoint(resource_id).get()
+
+        if resource_id:
+            results = response
+        else:
+            results = traverse_pagination(response, endpoint)
+    except:  # pylint: disable=bare-except
+        logger.exception('Failed to retrieve data from the Course Discovery API.')
```
mjfrey: This method should throw an exception not log an error.  The caller should try / catch this call and do what is appropriate.  See get_course_info_from_catalog.

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -73,3 +74,16 @@ def preview(self, request):
...
                 return Response(status=status.HTTP_400_BAD_REQUEST)
         return Response(status=status.HTTP_400_BAD_REQUEST)
+
+    @action(is_for_list=True, methods=['get'])
+    def course_catalogs(self, request):
+        """
+        Returns response with all course catalogs in the format:
+        ["results": {"id": 1, "name": "Dummy Catalog"}]
+        """
+        results = get_course_catalogs(site=request.site)
```
mjfrey: this should be in a try / catch around get_course_catalogs.

ecommerce/settings/base.py
```diff
@@ -237,6 +237,9 @@
 # Cache course info from course API.
 COURSES_API_CACHE_TIMEOUT = 3600  # Value is in seconds
 
+# Cache catalog info from Course Discovery API.
+CATALOG_API_CACHE_TIMEOUT = 3600  # Value is in seconds
```
mjfrey: Do we really need another setting or can we just share COURSES_API_CACHE_TIMEOUT?

mjfrey (changes requested):

ecommerce/settings/base.py
mattdrayer: We could share it but semantically speaking I think these two things are different

mjfrey: COURSES_API_CACHE_TIMEOUT is the setting we use today for the TTL of the results from the catalog service. The name is probably not the best.

mattdrayer: Okay, we'll just use that, then -- @zubair-arbi FYI ^^

zubair-arbi:
@mjfrey Thanks for providing valuable feedback. I have addressed your comments. Please give another review.

ecommerce/courses/tests/test_utils.py
```diff
@@ -74,3 +74,110 @@ def test_cert_display(self, cert_type, cert_display):
...
+        """
+        Verify that method "get_course_catalogs" returns proper response for
+        single page Course Discovery API response and uses cache to return data
+        in case of same API request.
+        """
+        self.mock_course_discovery_api_for_catalogs(catalog_name_list)
+
+        self._assert_get_course_catalogs(catalog_name_list)
+
+        # Verify the API was hit for once
```
mikedikan: nit: "was hit once"

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -52,7 +52,7 @@ def get_serializer_class(self):
             return CouponListSerializer
         return CouponSerializer
 
-    def create(self, request, *args, **kwargs):
+    def create(self, request, *args, **kwargs):  # pylint: disable=too-many-statements
```
mikedikan: I don't think this is the right response to this pylint error - this method should be broken up with some private helpers.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -242,6 +249,39 @@ def test_create_discount_coupon(self):
...
+        course_catalog = 1
+
+        course_catalog_coupon = self.create_course_catalog_coupon(
+            coupon_title=coupon_title,
+            quantity=quantity,
+            course_catalog=course_catalog,
+        )
+        self.assertEqual(course_catalog_coupon.title, coupon_title)
+
+        course_catalog_vouchers = course_catalog_coupon.attr.coupon_vouchers.vouchers.all()
```
mikedikan, line 267: this whole method has a lot of chaining objects/methods.  (this.that.other_thing.other_other_thing.first().final_thing).  If it wasn't a test method, i'd be concerned about testing for null conditions to avoid exceptions.  I didn't see any occurences in the actual code of this kind of chaining, but It might be valuable to add a helper method to the coupon and voucher objects to make this a little cleaner.

ecommerce/static/js/test/spec-runner.js
```diff
@@ -20,6 +20,8 @@ if (!window.gettext) {
 // Establish the global namespace
 window.ecommerce = window.ecommerce || {};
 window.ecommerce.coupons = window.ecommerce.coupons || {};
+window.ecommerce.catalogs = window.ecommerce.catalogs || {};
+window.ecommerce.catagories = window.ecommerce.catagories || {};
```
mikedikan: nit: catagories -> categories

ecommerce/static/js/test/specs/models/coupon_model_spec.js
```diff
@@ -81,12 +81,39 @@ define([
...
                     model.set('course_seat_types', ['verified']);
                     model.validate();
                     expect(model.isValid()).toBeTruthy();
                 });
 
+                it('should validate catalog query and course seat types for type Catalog', function () {
+                    model.set('catalog_type', 'Catalog');
+                    model.set('course_catalog', '');
+                    model.validate();
+                    expect(model.isValid()).toBeFalsy();
```
mikedikan: curious - do we not know exactly if this is false?  why tobefalsy?

mikedikan (commented):
Overall, this looks pretty good.  I couldn't see any big architectural problems, just a few questions, some code style suggestions, and some nitpicks.

mikedikan:
I've asked @AlasdairSwan to do a quick sanity check (this afternoon) on the backbone app changes.  They look fine to me, but I bow to his superior knowledge.

ecommerce/static/js/apps/coupon_admin_app.js
```diff
@@ -20,6 +22,11 @@ require([
             ecommerce.coupons.categories = new CategoryCollection();
             ecommerce.coupons.categories.url = '/api/v2/coupons/categories/';
             ecommerce.coupons.categories.fetch({ async: false });
+
+            ecommerce.coupons.catalogs = new CatalogCollection();
+            ecommerce.coupons.catalogs.url = '/api/v2/catalogs/course_catalogs/';
```
AlasdairSwan: This url is already being set in the Collection itself, why is it being added here too?

ecommerce/static/js/models/coupon_model.js
```diff
@@ -24,15 +26,23 @@ define([
             required: gettext('This field is required.'),
             number: gettext('This value must be a number.'),
             date: gettext('This value must be a date.'),
-            seat_types: gettext('At least one seat type must be selected.'),
+            seat_types: gettext('At least one seat type must be selected.')
         });
         _.extend(Backbone.Model.prototype, Backbone.Validation.mixin);
 
+        /* jshint esnext: true */
+        const CATALOG_TYPES = {
```
AlasdairSwan: This is ES2015, are you running Babel to ensure it doesn't break legacy browsers?

ecommerce/static/js/test/specs/models/coupon_model_spec.js
AlasdairSwan: Yeah, this seems very loose. I would prefer a check to determine that ```model.isValid()``` is ```false```.

mjfrey (approved):

AlasdairSwan (changes requested):

ecommerce/extensions/voucher/tests/test_utils.py
zubair-arbi: @mikedikan You are right about doing lot of chaining here but since this is a test and if some exception or null condition occurs then test would fail and that's what we want to happen. There are other tests where we are chaining like here.

ecommerce/static/js/models/coupon_model.js
zubair-arbi: @AlasdairSwan I haven't used Babel for legacy browsers but the tests are running fine for Firefox 42 which is default for the ecommerce devstack. Here are the browser compatibility stats:
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/const#Browser_compatibility
The desktop version of browsers seen supporting this but the mobile version don't seen sure about this constant. So I am going to use var instead to avoid any issues.

ecommerce/static/js/apps/coupon_admin_app.js
zubair-arbi: This is redundant, I am removing this line.

ecommerce/static/js/test/specs/models/coupon_model_spec.js
zubair-arbi: From your discussion what I understood is to modify the statement like:
```javascript
expect(model.isValid()).toBe(false);
```
Reference: http://stackoverflow.com/questions/32615713/tobetrue-vs-tobetruthy-vs-tobetrue

zubair-arbi:
@mikedikan @AlasdairSwan I have addressed your feedback, please review again.

mikedikan (approved):

AlasdairSwan (approved):

mattdrayer:
@zubair-arbi -- looks like we have at least 27 +1's now 😉 -- please rebase, squash, and merge after final Jenkins checks are passing.  Nice job sticking with this one!

PR #1067, opened 2016-12-28
ziafazal/WL-632: fixed horizontal scrolling on basked page when viewed in mobile device

ziafazal:
This PR has change to fix horizontal scrolling on basked page when viewed in mobile device.
[WL-632](https://openedx.atlassian.net/browse/WL-632) has more details on the issue.
@douglashall could you please review?

### DIFF
--- a/ecommerce/static/sass/partials/components/_buttons.scss
+++ b/ecommerce/static/sass/partials/components/_buttons.scss
@@ -8,6 +8,7 @@
 
 // Use the Pattern Library's decoration (border-bottom)
 .btn-link {
+    white-space: normal;
     &:focus, &:hover {
         text-decoration: none;
     }


douglashall:
@ziafazal This does not seem to solve the issue. What if we just remove the `btn` and `btn-link` classes from the basket switch `<a>` in the template? Note we will also have to do this for each theme in the edx-themes repo.

ziafazal:
@douglashall removing `btn` and `btn-link` classes would remove the styling and that would give us unexpected UI. This issue is not related to any of the WL sites theming it is a valid issue for edx.org too so ideally we should fix it inside the ecommerce app as this PR does.

douglashall:
@ziafazal Removing the `btn` and `btn-link` classes did not seem to affect styling for me. Does the change you have here fix the issue for you? It did not seem to fix it for me.

douglashall:
@ziafazal Now that I think about it though, this will no longer be an issue when we move to the new single page checkout that another team is working on. Please close out this PR and I'll close the Jira issue.

ziafazal:
Changes no longer needed since we a new checkout page.

PR #1068, opened 2016-12-28
[ECOM-6693] Fix payment error; Zip code is an optional field

marjev:
@clintonb, @mjfrey :

Here's a fix for the https://openedx.atlassian.net/browse/ECOM-6693 bug.
Please review once you're back. Thanks!


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1068&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

PR #1069, opened 2016-12-28
WL-895 Remove deprecated SiteConfiguration.segment_key field

douglashall:
This is a post-deployment follow up to https://github.com/edx/ecommerce/pull/1053 which removes the deprecated segment_key field from the SiteConfiguration model. Once this change is deployed, we can follow up with https://github.com/edx/ecommerce/pull/1075 to remove the column from the database.

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -81,14 +81,6 @@ class SiteConfiguration(models.Model):
         blank=False,
         default={}
     )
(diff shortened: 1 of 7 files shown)


douglashall:
@clintonb @bderusha @rlucioni

bderusha (approved):

douglashall:
No longer needed.

PR #1070, opened 2016-12-29
[ECOM-6698] Add padding between header and receipt page content

marjev:
Reviewers:
@douglashall 
@mjfrey 
Please take a look. Thanks!
Below is the screenshot showing the change.

JIRA ticket:
https://openedx.atlassian.net/browse/ECOM-6698

Screenshot:
image

### DIFF
--- a/ecommerce/static/sass/partials/views/_receipt.scss
+++ b/ecommerce/static/sass/partials/views/_receipt.scss
@@ -1,6 +1,8 @@
 .receipt {
     margin-bottom: 20px;
 
+    padding-top: 20px;
+
     background-color: #ffffff;
     font-family: $font-family-sans-serif;
     line-height: 1em;


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1070&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/static/sass/partials/views/_receipt.scss
```diff
@@ -1,6 +1,8 @@
 .receipt {
     margin-bottom: 20px;
 
+    padding-top: 20px;
```
douglashall, line 4: Can you remove the blank lines above and below this line?

douglashall:
👍

ecommerce/static/sass/partials/views/_receipt.scss
marjev: I've added those because of the concentric CSS guideline that the UX team agreed upon. This way, we're splitting the outer and inner CSS rules by grouping them and making scss files more readable.

If you disagree, I can remove those.

douglashall: Ah, I get it. Makes sense. Thanks for the clarification.

PR #1072, opened 2016-12-29
[ECOM-6693] Make postal code field optional.

vkaracic:


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1072&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/payment/forms.py
```diff
@@ -93,8 +93,8 @@ def __init__(self, user, *args, **kwargs):
     address_line1 = forms.CharField(max_length=29, label=_('Address'))
     address_line2 = forms.CharField(max_length=29, required=False, label=_('Suite/Apartment Number'))
     city = forms.CharField(max_length=32, label=_('City'))
-    state = forms.CharField(max_length=60, required=False, label=_('State/Province'))
-    postal_code = forms.CharField(max_length=10, required=False, label=_('Zip/Postal Code'))
+    state = forms.CharField(max_length=60, label=_('State/Province'))
```
clintonb: These should only be required for buyers in the U.S. and Canada. See the clean method below.

clintonb (changes requested):
Margining this as-is will cause issues for users outside of North America.

ecommerce/extensions/payment/forms.py
vkaracic: Okay, so that then leaves the question do we need the state and postal code information for non-US/CA buyers, @mjfrey? Just the address is enough for the SDN check.

mjfrey: Why are these only required for US/CA buyers?  Are there counties that do not have these and can still purchase things with a CC?

mjfrey: I see now that there are many countries without postal codes.  So yes @vkaracic just move to the clean method for US/CA learners and we will use what we have for address SDN checking.

vkaracic:
I've changed this PR to make the postal code field optional. Please take another look.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -152,7 +152,8 @@ def _get_billing_address(self, cybersource_response):
 
             # Oscar uses line4 for city
             line4=cybersource_response['req_bill_to_address_city'],
-            postcode=cybersource_response['req_bill_to_address_postal_code'],
+            # Postal code is optional
```
mjfrey, line 155: I still think we need to not have this optional for US / CA.  Add a check in the clean method.

mjfrey (changes requested):

ecommerce/extensions/payment/views/cybersource.py
vkaracic: Actually that's in an older PR: https://github.com/edx/ecommerce/pull/1062 I'd appreciate if you could review it though.

mjfrey: So what happens if I am in CA / US and I leave out the postal code.

mjfrey (approved):

ecommerce/extensions/payment/views/cybersource.py
vkaracic: This happens in that case.
screenshot from 2017-01-05 16-54-48

vkaracic:
@clintonb can you please take a second look?

clintonb (approved):

PR #1073, opened 2016-12-29
Fix design of credit card field.

vkaracic:
Changes the CSS of the credit card field on the
client-side basket checkout. The credit card
icon used to overlay the credit card numbers.
These changes fix that.

Screenshots are available here:
https://drive.google.com/drive/folders/0ByBQUYJ4LMyAVUFJeWE2UW1neTg?usp=sharing

https://openedx.atlassian.net/browse/ECOM-6707

### DIFF
--- a/ecommerce/static/sass/partials/views/_basket.scss
+++ b/ecommerce/static/sass/partials/views/_basket.scss
@@ -367,17 +367,17 @@
 
                     .fa-lock {
                         position: relative;
-                        float: left;
+                        float: right;
                         top: -25px;
-                        left: 92%;
+                        right: 10%;
                         margin-right: -5%;
                     }
 
                     .card-type-icon {
                         position: relative;
-                        float: left;
+                        float: right;
                         top: -30px;
-                        left: 70%;
+                        right: 28%;
                         margin-right: -15%;
                     }
                 }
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1073&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/static/sass/partials/views/_basket.scss
```diff
@@ -487,6 +487,14 @@
             }
         }
     }
+
+    .basket-client-side .container {
+        .payment-form #payment-information {
+            .form-input-elements .card-type-icon {
+                left: 75%;
```
douglashall: How about using pixels here instead of a percentage? This way the icon won't float around as the browser window width changes.

douglashall:
👍

ecommerce/static/sass/partials/views/_basket.scss
vkaracic: Not really because the input field itself changes and it should follow that change as close as possible, which is not possible with fixed pixels.

douglashall: I was thinking that since the credit card icon and lock icon are relatively positioned, you could float them right instead of left and then set the right property instead of the left property using pixels instead of percentage. This would keep the positioning consistent regardless of browser window size.

vkaracic: I've changed the float direction, but I'll stick with percentages because when I tried with pixels the icon positions were less consistent.

mjfrey (approved):

PR #1074, opened 2016-12-29
[ECOM-6699] Add dropdown for US and CA country states.

vkaracic:
Note: This should not be merged before https://github.com/edx/ecommerce/pull/1062 as that PR allows displaying these validation errors and stopping the checkout process.

Only a combination of country code and state code (2 character strings)
was allowed. With this change full names of states are allowed as well.

https://openedx.atlassian.net/browse/ECOM-6699

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -168,6 +168,107 @@ define([
                 hideVoucherForm();
             });
 
+            $('select[name=country]').on('change', function() {
+                var country = $('select[name=country]').val(),
+                    inputDiv = $('#div_id_state .controls'),
+                    states = {
+                        'US': {
+                            'Alabama': 'AL',
+                            'Alaska': 'AK',
+                            'American': 'AS',
+                            'Arizona': 'AZ',
+                            'Arkansas': 'AR',
+                            'California': 'CA',
+                            'Colorado': 'CO',
+                            'Connecticut': 'CT',
+                            'Delaware': 'DE',
+                            'Dist. of Columbia': 'DC',
+                            'Florida': 'FL',
+                            'Georgia': 'GA',
+                            'Guam': 'GU',
+                            'Hawaii': 'HI',
+                            'Idaho': 'ID',
+                            'Illinois': 'IL',
+                            'Indiana': 'IN',
+                            'Iowa': 'IA',
+                            'Kansas': 'KS',
+                            'Kentucky': 'KY',
+                            'Louisiana': 'LA',
+                            'Maine': 'ME',
+                            'Maryland': 'MD',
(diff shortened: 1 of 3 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1074&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

douglashall:
Is it not possible to display a list of countries to choose from instead of a text field?

vkaracic:
Something like this: http://www.cssscript.com/generic-country-state-dropdown-list-countries-js/ ?

douglashall:
@vkaracic I'm not sure countries.js specifically is the right solution, but yes, it would be great if there was a dropdown with a list of countries that the user could select from.

vkaracic:
@douglashall do you have any other package in mind? CountryJS seems pretty convenient: https://www.npmjs.com/package/countryjs

douglashall:
@vkaracic I would check in with @clintonb on how we might implement the country list.

clintonb:
The country list should simply be a choice field sourced from `pycountry`. States/provinces should be the same, if applicable. I'll leave it to you to decide how that gets rendered client-side. Keep in mind that this form will be used for other payment providers (e.g. Stripe). To ensure compatibility, please ensure the field values are the two-digit codes for the countries and states.

Also, have a look at CyberSource's default payment form. I believe they change the state field to a text field if you select certain countries. We may want to do something similar to avoid issues with missing states.

vkaracic:
I've made changes that add a dropdown only for CA and US (the only ones we actually need the correct state codes. Here's a short clip of how it works:
out-1

ecommerce/static/js/pages/basket_page.js
```diff
@@ -168,6 +168,106 @@ define([
...
+                        'aria-required="true" required></select>'
+                    );
+                    _.each(states[country], function(value, key) {
+                        $('#id_state').append($('<option>', {value: value, text: key}));
+                    });
+                } else {
+                    $(inputDiv).empty();
+                    $(inputDiv).append(
+                        '<input class="textinput textInput form-control" id="id_state"' +
+                        'maxlength="60" name="state" type="text">'
```
clintonb, line 267: What is dictating this max-length value?

clintonb: I see this comes from the Django form (which is ultimately driven by CyberSource). Please add comments informing future developers of the need to update max-length in both locations, or pull this value from the Django-generated form.

vkaracic: Comments added.

clintonb (approved):

mjfrey (approved):

PR #1075, opened 2016-12-29
WL-895 Migration for removing SiteConfiguration.segment_key field

douglashall:
This is a follow on to https://github.com/edx/ecommerce/pull/1069 which adds the migration for removing the SiteConfiguration.segment_key column from the database.


douglashall:
@clintonb @bderusha @rlucioni

bderusha (approved):

douglashall:
No longer needed.

PR #1076, opened 2016-12-29
Increase address input field length.

vkaracic:
There are no known measures for limiting addresses so I'm leaving the max length.

ECOM-6705


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1076&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mjfrey (approved):

PR #1077, opened 2016-12-30
[WIP] ENT-112: Automatic Enrollment Code Applied for Enterprise Learners

mattdrayer:
@zubair-arbi @saleem-latif @asadiqbal08 @brittneyexline FYI -- also it looks like @zubair-arbi's branch is in need of a rebase.


zubair-arbi:
@mattdrayer It looks like we are moving in right direction.

PR #1078, opened 2017-01-03
Upgrade edx-auth-backends to 0.5.3

rlucioni:
This version of the package limits the installed version of PSA to <0.3.0. 0.3.0 introduced a major breaking change, effectively gutting PSA and moving its Django components to a new social-auth-app-django package.

@edx/ecommerce @jibsheet FYI. This is follow-up to https://github.com/edx/auth-backends/pull/27 and should fix CI builds.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -15,7 +15,7 @@ django-waffle==0.11.1
 djangorestframework==3.2.3
 djangorestframework-jwt==1.8.0
 drf-extensions==0.2.8
-edx-auth-backends==0.5.2
+edx-auth-backends==0.5.3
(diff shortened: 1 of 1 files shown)


clintonb (approved):

rlucioni:
@vkaracic @mjfrey are the build failures on this PR familiar to either of you? They don't appear to be tied to my change; see https://github.com/edx/ecommerce/pull/1079, on which the builds also fail.

vkaracic:
@rlucioni failing build fixed in https://github.com/edx/ecommerce/pull/1080 - not yet merged though.

PR #1080, opened 2017-01-04
Fix CC expired month JS test.

vkaracic:
@saleem-latif @mattdrayer this should fix the failing JS test in a similar way that we freeze time in python tests. I know you wanted to include this in one of your PR's but this needs to be merged ASAP as it causes all the CI builds to fail and blocks any work on ecomm.

### DIFF
--- a/ecommerce/static/js/test/specs/pages/basket_page_spec.js
+++ b/ecommerce/static/js/test/specs/pages/basket_page_spec.js
@@ -7,8 +7,7 @@ define([
         'models/tracking_model',
         'models/user_model',
         'views/analytics_view',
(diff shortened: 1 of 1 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1080"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

saleem-latif:
@vkaracic LGTM for unblocking ecommerce releases 👍

rlucioni (approved):
Let's get this merged so we can get builds passing again.

PR #1081, opened 2017-01-04
Limit basket checkout address length.

vkaracic:
According to http://apps.cybersource.com/library/documentation/dev_guides/Verification_Svcs_SO_API/Verification_Svcs_SO_API.pdf the address field needs to be 60 characters long.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1081"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mjfrey (approved):

PR #1084, opened 2017-01-05
[WL-910] Fix basket quantity update element design.

vkaracic:
Screenshots here: https://drive.google.com/open?id=0ByBQUYJ4LMyAV0hjYzJCamtseU0

https://openedx.atlassian.net/browse/WL-910

### DIFF
--- a/ecommerce/static/sass/partials/views/_basket.scss
+++ b/ecommerce/static/sass/partials/views/_basket.scss
@@ -409,6 +409,30 @@
                 margin: 0;
             }
 
+            .quantity-group {
+                padding-left: 0;
+            }
+
+            .quantity-label {
+                display: inline-block;
+                padding-right: 10px;
+            }
+
+            .checkout-quantity {
+                display: inline-block;
+            }
+
+            .spinner {
+                width: 70px;
+                float: left;
+                padding-right: 10px;
+            }
+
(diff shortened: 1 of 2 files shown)


douglashall:
@vkaracic This is what I'm seeing now in the smallest breakpoint:
screen shot 2017-01-11 at 9 11 43 am

I think we want the text input and update button to remain the same size across all breakpoints, we just don't want them to wrap.

Taking a step back and thinking about this more holistically, from a usability perspective it would be great if we could get rid of the update button and have the basket auto update when the quantity field is changed.

vkaracic:
It was the point to have the update button larger for smaller devices which is easier to use. Changing the quantity has to refresh the page every time (because quantity needs to be saved in Oscar), I doubt it would be friendly for users that want more than 2 quantity to refresh their page each time.

douglashall:
Okay, I would do a quick check in with Liz on this decision.

Why do we need to refresh the page on a quantity change? Why not just make an XHR?

vkaracic:
That's how Oscar works. Even with an ajax call for change the quantity the price would stay the same. I guess it can be changed to work like that but that needs to be a different ticket.

douglashall:
Sounds good. It would be great if we could make that happen some day.

PR #1085, opened 2017-01-05
Fix Firefox autofill credit card bug.

vkaracic:
Firefox has a bug when autofilling input fields no event would be triggered:
https://bugzilla.mozilla.org/show_bug.cgi?id=87943
In our case that's when the credit card information is autofilled and thus the card type would not be registered. In this PR I've moved the card detection code to a separate function and called it when payment button is clicked so that it performs a credit card detection even when no credit card input field related events are triggered.

https://openedx.atlassian.net/browse/ECOM-6741

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -132,12 +132,31 @@ define([
                 appendCardValidationErrorMsg(event, cardExpiryMonthField, 'Card expired');
             }
         },
+
+        detectCreditCard = function() {
+            var cardNumber = $('#card-number-input').val().replace(/\s+/g, ''),
+                card,
+                iconPath = '/static/images/credit_cards/';
+
+            if (cardNumber.length > 12) {
+                card = CreditCardUtils.getCreditCardType(cardNumber);
+
+                if (typeof card !== 'undefined') {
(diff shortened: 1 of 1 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1085&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

vkaracic:
@douglashall mind taking a look?

mjfrey (approved):

ecommerce/static/js/pages/basket_page.js
```diff
@@ -150,28 +169,16 @@ define([
...
-                cardNumber = $('#card-number-input').val().replace(/\s+/g, '');
-
-                if (cardNumber.length > 12) {
-                    card = CreditCardUtils.getCreditCardType(cardNumber);
-
-                    if (typeof card !== 'undefined') {
-                        $('.card-type-icon').attr(
-                            'src',
-                            iconPath + card.name + '.png'
-                        ).removeClass('hidden');
```
douglashall, line 162: @vkaracic @mjfrey Heads up. Looks like we lost the credit card icon display with this commit.

PR #1086, opened 2017-01-06
ENT-127: Fix JS test failure for basket page tests

saleem-latif:
Hi @mattdrayer , @asadiqbal08 , @vkaracic , @zubair-arbi 

Kindly take a look at this PR, it contains changes for ENT-127.

__Description of Changes:__
Months in Javascript start from 0 (i.e. JAN is 0, FEB is 1 etc.) however months displayed in the dropdown for credit card expiry months start from 1 (i.e. JAN is 1, FEB is 0 etc.). So, we get an off by 1 error if we compare months coming dropdown and month accessed from js `Date` type. In order to resolve the issue we must add 1 to the month number accessed via js `Date` type.

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -102,7 +102,10 @@ define([
 
         cardInfoValidation = function (event) {
             var cardType,
-                currentMonth = new Date().getMonth(),
+                // We are adding 1 here because month in js style date-time starts with 0
+                // i.e. 0 for JAN, 1 for FEB etc. However, credit card expiry months start with 1
+                // i.e 1 for JAN, 2 for FEB etc.
+                currentMonth = new Date().getMonth() + 1,
                 currentYear = new Date().getFullYear(),
                 cardNumber = $('input[name=card_number]').val(),
                 cvnNumber = $('input[name=card_cvn]').val(),
--- a/ecommerce/static/js/test/specs/pages/basket_page_spec.js
+++ b/ecommerce/static/js/test/specs/pages/basket_page_spec.js
@@ -7,7 +7,8 @@ define([
         'models/tracking_model',
         'models/user_model',
         'views/analytics_view',
-        'js-cookie'
+        'js-cookie',
+        'moment'
     ],
     function ($,
               _,
@@ -17,7 +18,9 @@ define([
               TrackingModel,
               UserModel,
               AnalyticsView,
-              Cookies) {
+              Cookies,
+              moment
+    ) {
         'use strict';
 
         describe('Basket Page', function () {
@@ -208,8 +211,21 @@ define([
             });
 
             describe('clientSideCheckoutValidation', function() {
-                    var currentYear = 2016,
-                        lastMonth = 11;
+            var cc_expiry_months = {
+                    JAN: '01',
+                    FEB: '02',
+                    MAR: '03',
+                    APR: '04',
+                    MAY: '05',
+                    JUN: '06',
+                    JUL: '07',
+                    AUG: '08',
+                    SEP: '09',
+                    OCT: '10',
+                    NOV: '11',
+                    DEC: '12'
+                };
+
 
                 beforeEach(function() {
                     $(
@@ -232,8 +248,6 @@ define([
                         '<div><input name="card_cvn">' +
                         '<p class="help-block"></p></div>' +
                         '<div><select name="card_expiry_month">' +
-                        '<option value="01">01</option>' +
-                        '<option value="12">12</option>' +
                         '<option value="99">99</option>' +
                         '</select>' +
                         '<p class="help-block"></p></div>' +
@@ -244,11 +258,10 @@ define([
                         '<button id="payment-button">Pay</button>'
                     ).appendTo('body');
 
-                    $('select[name=card_expiry_year]').append('<option value="' +
-                        currentYear + '">' + currentYear + '</option>'
-                    );
-                    $('select[name=card_expiry_month]').append('<option value="' +
-                        lastMonth + '">' + lastMonth + '</option>'
+                    $('select[name=card_expiry_month]').append(
+                        _.reduce(_.toArray(cc_expiry_months), function(memo, value){
+                            return memo + '<option value="' + value + '">' + value + '</option>';
+                        }, '')
                     );
 
                     $('input[name=first_name]').val('Joey');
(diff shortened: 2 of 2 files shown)


ecommerce/static/js/pages/basket_page.js
```diff
@@ -102,7 +102,10 @@ define([
 
         cardInfoValidation = function (event) {
             var cardType,
-                currentMonth = new Date().getMonth(),
+                // We are adding 1 here because month in js style date-time starts with 0
+                // i.e. 0 for JAN, 1 for FEB etc. However, credit card expiry months start with 1
```
mattdrayer, line 106: Do we know if credit card expiry months ALWAYS start with 1?  ie, is it the same across all CC vendors (Visa, MasterCard, etc.)

saleem-latif: @mattdrayer this is where months are coming from https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/views.py#L204 and here it the template code that is creating dropdown https://github.com/edx/ecommerce/blob/master/ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html#L166-L168

mattdrayer: Cool, thanks for the additional info -- I'm good with this PR -- @vkaracic if you can also take a look that would be swell -- 👍

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -232,14 +249,10 @@ define([
...
 
-                    // Freeze time to 12-2016.
-                    spyOn(Date.prototype, 'getFullYear').and.returnValue(currentYear);
-                    spyOn(Date.prototype, 'getMonth').and.returnValue(lastMonth + 1);
-
                     BasketPage.onReady();
                 });
 
-                describe('cardHolderInformationValidation', function() {
+                describe('card HolderInformationValidation', function() {
```
vkaracic: Why this change?

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -376,6 +395,39 @@ define([
                         expect($('select[name=card_expiry_month] ~ .help-block').has('span').length).toEqual(0);
                     });
                 });
+
+                describe('Validate card expiration date.', function() {
```
vkaracic: How does this test differ from the one on line 384?

vkaracic (commented):
Overall it's OK, I just have a couple of questions.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
saleem-latif: @vkaracic it was an accidental change, I have reverted this.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
saleem-latif: @vkaracic this test verifies the scenario where card expiry is previous month (e.g. if current month is 'FEB' and card expires in 'JAN').
The test  on line 384 uses hardcoded values of month, while this tests tries to simulate the behavior of month dropdown for credit card expiry month. As the original issue will only be recreated if month is coming from dropdown and is not hard coded.

vkaracic: In essence it's pretty much the same test and it's unnecessary in my opinion.

saleem-latif: @vkaracic It is this test that exposes the bug, test will fail if we remove the fix, and pass only if the fix is there. Other tests will pass in all cases. That is why I think we need this test also.

vkaracic: In that case I think the problem lies here: https://github.com/edx/ecommerce/pull/1086/files#diff-d434c36b5a7b587cb874b84c75a1c9c3R321 since it freezes the month to 12 which does in exists in a 0-11 month scale. I suggest fixing that test instead of writing a new one.

saleem-latif:
@vkaracic I have removed those tests cases and updated existing tests, kindly take a look.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -17,12 +18,28 @@ define([
...
+              Cookies,
+              moment
+    ) {
         'use strict';
 
         describe('Basket Page', function () {
             var data,
-                form;
+                form,
+                cc_expiry_months = {
```
vkaracic: Move this variable where it's used.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -244,11 +257,10 @@ define([
                         '<button id="payment-button">Pay</button>'
                     ).appendTo('body');
 
-                    $('select[name=card_expiry_year]').append('<option value="' +
-                        currentYear + '">' + currentYear + '</option>'
-                    );
-                    $('select[name=card_expiry_month]').append('<option value="' +
-                        lastMonth + '">' + lastMonth + '</option>'
+                    $('select[name=card_expiry_month]').append(
+                        _.reduce(_.toArray(cc_expiry_months), function(memo, value){
```
vkaracic, line 262: What's "memo"?

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -325,7 +333,20 @@ define([
...
+                        enRouteCardNumber = '201401173701274', // Unsupported type (Dec, 2016)
+                        today = moment(),
+                        cardExpirationMonth = 'FEB',  // Card Expires in February
+                        thisMonth = moment().month('MAR').month(); // Let's say this month is March
+
+                    beforeEach (function () {
+                        $('select[name=card_expiry_year]').append('<option value="' +
+                            today.year() + '">' + today.year() + '</option>'
+                        );
+                        // Freeze Month to March.
```
vkaracic: Lowercase "month".

vkaracic (approved):
The comments I left are all nits. Overall LGTM.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
saleem-latif: @vkaracic "memo" contains the result of previous iteration of `reduce` call. ref: http://underscorejs.org/#reduce

saleem-latif: I used the same naming convention as underscore, let me know if some other name sounds better.

vkaracic: No, no, it's fine, carry on :+1:

PR #1087, opened 2017-01-06
Reduce impact store dashboard can have on site performance

rlucioni:
Oscar's default store dashboard doesn't scale well for sites with millions of orders. To limit the impact this page can have on large systems, all queries against orders are now limited to those placed in the last 24 hours. Order's date_placed field is indexed for more efficient queries. I've also removed fields that aren't useful to us (e.g., stock alerts).

@edx/ecommerce


clintonb (approved):

schenedx:
@rlucioni If I want to view orders outside of the 24 hours back, with your change here, can I still view then in UI?

rlucioni:
@schenedx yes. These changes are limited to the store dashboard.

schenedx (approved):
👍

PR #1088, opened 2017-01-09
[ECOM-6746] Editing Single Course Coupon Code to Multi-Course Redeems…

iivic:
https://openedx.atlassian.net/browse/ECOM-6746
On coupon edit page when saving changes in case of single-use vouchers the max_uses field is set to 1. This will set the global_max_applications to 1 which means that the coupon can be redeemed only once. The max_uses field should not be set in case of single-use vouchers.
@mjfrey @vkaracic @marjev

### DIFF
--- a/ecommerce/extensions/offer/tests/test_commands.py
+++ b/ecommerce/extensions/offer/tests/test_commands.py
@@ -1,0 +1,58 @@
+from __future__ import unicode_literals
+from StringIO import StringIO
+
+from django.core.management import call_command
+from oscar.core.loading import get_model
+from oscar.test import factories
+
+from ecommerce.tests.testcases import TestCase
+
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
+Voucher = get_model('voucher', 'Voucher')
+
+
+class SetMaxApplicationsToNoneCommandTest(TestCase):
+    command = 'set_max_applications_to_none'
+    filter_condition = {
+        'max_global_applications': 1,
+        'vouchers__usage': Voucher.SINGLE_USE
+    }
+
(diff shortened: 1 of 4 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1088"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

vkaracic:
Have you tested that a single-use coupon can be redeemed only once now?

vkaracic (approved):

iivic:
@vkaracic Yes I did.

marjev (approved):

iivic:
Please review one more time I included the management command.

mjfrey (approved):

ecommerce/extensions/offer/tests/test_commands.py
```diff
@@ -0,0 +1,44 @@
...
+        call_command(self.command, stdout=output)
+        actual_output = output.getvalue().strip()
+
+        self.assertTrue(actual_output.startswith(
+            'Setting max_global_applications field to None for ConditionalOffer [{}]...'.format(offer)
+        ))
+        self.assertTrue(actual_output.endswith('Done.'))
+        self.assertEqual(ConditionalOffer.objects.filter(**self.filter_condition).count(), 0)
+
+    def test_command_without_sample(self):
```
vkaracic, line 42: Please add a unit test to verify vouchers other than the ones within filter conditions are remain unaffected after running this command.

vkaracic (changes requested):

vkaracic (approved):

PR #1090, opened 2017-01-10
[SOL-2175] Add util for cleaning CyberSource fields.

vkaracic:
As per http://apps.cybersource.com/library/documentation/dev_guides/Getting_Started_SO/Getting_Started_SO_API.pdf?searchid=1468415087832 (page 19), colons and carets are not allowed in ``item_#`` fields. The same cleanup is added for ``bill_to`` fields.

https://openedx.atlassian.net/browse/SOL-2175

### DIFF
--- a/ecommerce/extensions/payment/utils.py
+++ b/ecommerce/extensions/payment/utils.py
@@ -1,3 +1,5 @@
+import re
+
 from django.utils.translation import ugettext_lazy as _
 
 
@@ -37,3 +39,17 @@ def middle_truncate(string, chars):
     truncated = u'{start}{indicator}{end}'.format(start=start, indicator=indicator, end=end)
 
     return truncated
+
+
+def clean_field_value(value):
+    """Strip the value of any special characters.
+
+    Currently strips caret(^), colon(:) and quote(" ') characters from the value.
+
+    Args:
+        value (str): The original value.
+
+    Returns:
+        A cleaned string.
+    """
+    return re.sub(r'[\^:"\']', '', value)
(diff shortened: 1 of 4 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1090"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev (approved):

iivic (approved):

ecommerce/extensions/payment/utils.py
```diff
@@ -37,3 +39,17 @@ def middle_truncate(string, chars):
...
+
+    Currently strips caret(^), colon(:) and quote(" ') characters from the value.
+
+    Args:
+        value (str): The original value.
+
+    Returns:
+        A cleaned string.
+    """
+    return re.sub(r'[\^:"\']', '', value)
```
mjfrey, line 55: Do we need to look for other chars than this?

vkaracic: I don't know, you only told me from your email exchange with their support that quotes need to be removed.
I've contacted Cybersource support and from the bunch of documentation they sent me I could only find the one instance in the documentation in this PR's header, where any non-allowed characters are mentioned.

mjfrey: Right, different answers each time you ask.

mjfrey (approved):

ecommerce/extensions/payment/utils.py
vkaracic: We can add/remove characters in this function if we find out more.

PR #1092, opened 2017-01-13
Revert "ENT-110: Add Catalog selection option to Coupon create/edit form"

zubair-arbi:
Reverts edx/ecommerce#1066
The create and edit pages seem broken with the changes so reverting the PR.

@asadiqbal08 @saleem-latif @mattdrayer

PR #1071, opened 2016-12-29
ENT-126:  Add js tests for catalog dropdown in coupon edit view

saleem-latif:
@mattdrayer , @zubair-arbi , @asadiqbal08  -- FYI

Currently, This PR only fixes the current failures. I will be adding more tests to this PR.

PR #1093, opened 2017-01-13
ENT-110: Add Catalog selection option to Coupon create/edit form

zubair-arbi:
ENT-110

This is a rework of adding catalog selection functionality on coupon create/edit form which I merged today "13-Jan-2016" and had to revert since the field `catalog` in `Range` model was passed as empty value.
## Branch History:
**Original PR:** https://github.com/edx/ecommerce/pull/1066
**Reverted PR:** https://github.com/edx/ecommerce/pull/1092

There were two ways to add catalogs selection functionality:

1. Provide catalogs lists from view and pass the list as an argument all the way to `coupon_form_view.js` from `coupon_app.js`
2. Get catalogs list through `ajax` while rendering the page `coupon_form.html`.

I followed second approach which is currently being used for populating the `catagories` field and also seems more suitable for our current requirement.

We also needed a new field to save the catalog id returned from the discovery service to display the selected catalog on edit coupon view and also to display catalog courses on offer redeem page. Saving the catalog id on our end will abstract us away from any changes (name, query etc) being made in the service `discovery`. 

> _**Note:** We will need to create a follow up story to update the `Redeem` coupon page to use the new field `course_catalog` for displaying the courses against some coupon._

## Completed tasks:
* Add new field `course_catalog` to save the id for the related catalog in coupon range model
* Add migration for the `offer` model `Range`
* Update coupon form to display the new course type selection `Catalog`
* Update coupon form to add a new field `course_catalog` to allow catalogs selection while creating coupons
* Update backend code to save the `course_catalog` value in the model `Range`
* Add method `get_course_catalogs` to recursively get all catalogs or a single course catalog from the service `discovery`
* Update serializer `CouponSerializer` to include `course_catalog` field
* Display course catalog name on the coupon detail page in case of `Catalog`
* Update the `Coupon edit page` for course catalogs
* Use existing settings variable `COURSES_API_CACHE_TIMEOUT` for caching the Course Catalogs API response which has default value of `3600  # Value is in seconds`.
* Refactor the method `create` of class `CouponViewSet` and move all cleaning/validation logic to separate class method `clean_voucher_request_data`.
* Add unit tests
* Add js tests (https://github.com/edx/ecommerce/pull/1071)

### Coupon create page
screen shot 2016-12-30 at 4 15 46 pm

### Coupon detail page
screen shot 2016-12-30 at 4 16 07 pm

### Coupon edit page
screen shot 2016-12-30 at 4 16 33 pm

## Related stories to this feature (future):
* Update coupon redemption logic to handle course runs based on the catalog fetched on the basis of new `Range` model field `course_catalog`
* Update coupon report generation method to include course catalogs

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -7,7 +7,6 @@ from django.core.exceptions import ValidationError
 from django.db import IntegrityError, transaction
 from django.http import Http404
 from django.shortcuts import get_object_or_404
-from django.utils.translation import ugettext as _
 from oscar.core.loading import get_model
 from rest_framework import filters, generics, serializers, status, viewsets
 from rest_framework.permissions import IsAdminUser, IsAuthenticated
@@ -73,98 +72,37 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
             429 if the client has made requests at a rate exceeding that allowed by the configured rate limit.
             500 if an error occurs when attempting to create a coupon.
         """
-        category_data = request.data.get('category')
-        code = request.data.get('code')
-        course_seat_types = request.data.get('course_seat_types')
-        max_uses = request.data.get('max_uses')
-        partner = request.site.siteconfiguration.partner
-        stock_record_ids = request.data.get('stock_record_ids')
-        voucher_type = request.data.get('voucher_type')
-
         try:
             with transaction.atomic():
-                if code:
-                    try:
-                        Voucher.objects.get(code=code)
-                        return Response(
-                            'A coupon with code {code} already exists.'.format(code=code),
-                            status=status.HTTP_400_BAD_REQUEST
-                        )
-                    except Voucher.DoesNotExist:
-                        pass
-
-                if course_seat_types:
-                    try:
-                        course_seat_types = prepare_course_seat_types(course_seat_types)
-                    except (AttributeError, TypeError) as e:
-                        logger.exception('Failed to create coupon. Invalid course seat types data: %s', e.message)
-                        return Response(
-                            _('Invalid course seat types data: {}').format(e.message),
-                            status=status.HTTP_400_BAD_REQUEST
-                        )
-
                 try:
-                    category = Category.objects.get(name=category_data['name'])
-                except Category.DoesNotExist:
-                    return Response(
-                        'Category {category_name} not found.'.format(category_name=category_data['name']),
-                        status=status.HTTP_404_NOT_FOUND
-                    )
-                except (KeyError, TypeError):
-                    return Response('Invalid Coupon Category data.', status=status.HTTP_400_BAD_REQUEST)
-
-                # Maximum number of uses can be set for each voucher type and disturb
-                # the predefined behaviours of the different voucher types. Therefor
-                # here we enforce that the max_uses variable can't be used for SINGLE_USE
-                # voucher types.
-                if max_uses and voucher_type != Voucher.SINGLE_USE:
-                    max_uses = int(max_uses)
-                else:
-                    max_uses = None
-
-                # When a black-listed course mode is received raise an exception.
-                # Audit modes do not have a certificate type and therefore will raise
-                # an AttributeError exception.
-                if stock_record_ids:
-                    seats = Product.objects.filter(stockrecords__id__in=stock_record_ids)
-                    for seat in seats:
-                        try:
-                            if seat.attr.certificate_type in settings.BLACK_LIST_COUPON_COURSE_MODES:
-                                return Response('Course mode not supported', status=status.HTTP_400_BAD_REQUEST)
-                        except AttributeError:
-                            return Response('Course mode not supported', status=status.HTTP_400_BAD_REQUEST)
-
-                    stock_records_string = ' '.join(str(id) for id in stock_record_ids)
-                    coupon_catalog, __ = get_or_create_catalog(
-                        name='Catalog for stock records: {}'.format(stock_records_string),
-                        partner=partner,
-                        stock_record_ids=stock_record_ids
-                    )
-                else:
-                    coupon_catalog = None
+                    cleaned_voucher_data = self.clean_voucher_request_data(request)
+                except ValidationError as error:
+                    logger.exception('Failed to create coupon. %s', error.message)
+                    return Response(error.message, status=error.code)
 
                 try:
                     coupon_product = create_coupon_product(
-                        benefit_type=request.data.get('benefit_type'),
-                        benefit_value=request.data.get('benefit_value'),
-                        catalog=coupon_catalog,
-                        catalog_query=request.data.get('catalog_query'),
-                        category=category,
-                        code=code,
-                        course_seat_types=course_seat_types,
-                        email_domains=request.data.get('email_domains'),
-                        end_datetime=request.data.get('end_datetime'),
-                        max_uses=max_uses,
-                        note=request.data.get('note'),
-                        partner=partner,
-                        price=request.data.get('price'),
-                        quantity=request.data.get('quantity'),
-                        start_datetime=request.data.get('start_datetime'),
-                        title=request.data.get('title'),
-                        voucher_type=voucher_type
+                        benefit_type=cleaned_voucher_data['benefit_type'],
+                        benefit_value=cleaned_voucher_data['benefit_value'],
+                        catalog=cleaned_voucher_data['coupon_catalog'],
+                        catalog_query=cleaned_voucher_data['catalog_query'],
+                        category=cleaned_voucher_data['category'],
+                        code=cleaned_voucher_data['code'],
+                        course_catalog=cleaned_voucher_data['course_catalog'],
+                        course_seat_types=cleaned_voucher_data['course_seat_types'],
+                        email_domains=cleaned_voucher_data['email_domains'],
+                        end_datetime=cleaned_voucher_data['end_datetime'],
+                        max_uses=cleaned_voucher_data['max_uses'],
+                        note=cleaned_voucher_data['note'],
+                        partner=cleaned_voucher_data['partner'],
+                        price=cleaned_voucher_data['price'],
+                        quantity=cleaned_voucher_data['quantity'],
+                        start_datetime=cleaned_voucher_data['start_datetime'],
+                        title=cleaned_voucher_data['title'],
+                        voucher_type=cleaned_voucher_data['voucher_type'],
                     )
-                except IntegrityError as e:
-                    return Response(str(e), status=status.HTTP_400_BAD_REQUEST)
+                except (KeyError, IntegrityError) as error:
+                    return Response(str(error), status=status.HTTP_400_BAD_REQUEST)
 
                 basket = prepare_basket(request, coupon_product)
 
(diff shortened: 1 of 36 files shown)


ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -199,7 +199,7 @@ def clean_voucher_request_data(cls, request):
         return {
             'benefit_type': request.data.get('benefit_type'),
             'benefit_value': request.data.get('benefit_value'),
-            'catalog': coupon_catalog,
+            'coupon_catalog': coupon_catalog,
```
mattdrayer, line 200: Good catch on this issue, @zubair-arbi -- let's be sure to cover it with a test, and then I am good with the fix.

zubair-arbi: @mattdrayer I am now using the key lookup instead of using dict get method. This way if the key in not available we will get the `KeyError`. For our current case the tests will throw:
```python
KeyError: u'coupon_catalog'
```

zubair-arbi:
@mattdrayer @mikedikan @mjfrey I would really appreciate if you guys can do another review. We had to revert the [original PR|https://github.com/edx/ecommerce/pull/1066] due to the `catalog` field being set to `None`.

The reason for `catalog` field being set to `None` was a typo in the keys returned by the new helper method `clean_voucher_request_data` which I created in order to refactor the main `create` method of class `CouponViewSet`. Now I have updated to code to throw `KeyError` if a key is missing or wrong in the response dict returned by the method `clean_voucher_request_data`.

ecommerce/extensions/api/v2/views/coupons.py
mattdrayer: okay -- I think we should have @vkaracic take a look at this change as well -- if I understand this change correctly then we are now requiring all of these parameters to be provided in `request.data`, whereas using `.get()` the parameters could be passed through as null values if not present.

vkaracic: @iivic is working with validations, he'll be of more use here.

iivic (changes requested):
A few things:
- on coupon creation when changing number of codes field a TypeError appears in the console (edited: this is a legacy issue I will fix this bug you can ignore it.)
screen shot 2017-01-16 at 11 47 02
- when the default value is left for the catalog another type error appears on details page 
screen shot 2017-01-16 at 11 47 56
- when catalog changed it creates the coupon and displays the details page however the edit page didn't appear
screen shot 2017-01-16 at 11 53 15

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -630,6 +670,21 @@ define([
...
+                    if (this.model.has('course_catalog')) {
+                        if (typeof this.model.get('course_catalog') === 'number') {
+                            catalogId = this.model.get('course_catalog');
+                        } else if (!$.isEmptyObject(this.model.get('course_catalog'))) {
+                            catalogId = this.model.get('course_catalog').id;
+                        }
+                    }
+
+                    if (catalogId !== '') {
+                        this.model.set('course_catalog', ecommerce.coupons.catalogs[catalogId]);
```
iivic: ecommerce.coupons.catalogs will return an object. I don't see how `ecommerce.coupons.catalogs[catalogId]` fits in.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -417,22 +441,37 @@ define([
...
+                    this.model.unset('course_id');
+                    this.model.unset('seat_type');
+                    this.model.unset('stock_record_ids');
+                    this.model.unset('catalog_query');
+                    this.model.unset('course_seat_types');
+                    this.formGroup('[name=catalog_query]').addClass(this.hiddenClass);
+                    this.formGroup('[name=course_seat_types]').addClass(this.hiddenClass);
+                    this.formGroup('[name=course_id]').addClass(this.hiddenClass);
+                    this.formGroup('[name=seat_type]').addClass(this.hiddenClass);
+                    this.formGroup('[name=course_catalog]').removeClass(this.hiddenClass);
```
iivic, line 463: You should include `this.$('[name=seat_type] option').remove();` When course_id entered and switched to catalog the seat_type would remain in the field this will remove it.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -73,95 +72,34 @@ def create(self, request, *args, **kwargs):
...
-                    stock_records_string = ' '.join(str(id) for id in stock_record_ids)
-                    coupon_catalog, __ = get_or_create_catalog(
-                        name='Catalog for stock records: {}'.format(stock_records_string),
-                        partner=partner,
-                        stock_record_ids=stock_record_ids
-                    )
-                else:
-                    coupon_catalog = None
+                    cleaned_voucher_data = self.clean_voucher_request_data(request)
+                except ValidationError as error:
```
iivic, line 79: You don't have to catch the exception here. There is a try except outside of the atomic transaction for this purpose just add the logger there.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -73,95 +72,34 @@ def create(self, request, *args, **kwargs):
...
+                        email_domains=cleaned_voucher_data['email_domains'],
+                        end_datetime=cleaned_voucher_data['end_datetime'],
+                        max_uses=cleaned_voucher_data['max_uses'],
+                        note=cleaned_voucher_data['note'],
+                        partner=cleaned_voucher_data['partner'],
+                        price=cleaned_voucher_data['price'],
+                        quantity=cleaned_voucher_data['quantity'],
+                        start_datetime=cleaned_voucher_data['start_datetime'],
+                        title=cleaned_voucher_data['title'],
+                        voucher_type=cleaned_voucher_data['voucher_type'],
```
iivic, line 102: You should catch the KeyError.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -179,6 +117,104 @@ def create(self, request, *args, **kwargs):
...
+        code = request.data.get('code')
+        course_catalog_data = request.data.get('course_catalog')
+        course_seat_types = request.data.get('course_seat_types')
+        max_uses = request.data.get('max_uses')
+        partner = request.site.siteconfiguration.partner
+        stock_record_ids = request.data.get('stock_record_ids')
+        voucher_type = request.data.get('voucher_type')
+
+        if code and Voucher.does_exist(code):
+            validation_message = 'A coupon with code {code} already exists.'.format(code=code)
```
iivic, line 140: I noticed you removed the translations from the exception messages. Is there a reason for that?

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -240,6 +276,19 @@ def update(self, request, *args, **kwargs):
             # Remove catalog if switching from single course to dynamic query
             if voucher_range.catalog:
                 range_data['catalog'] = None
+
+            course_catalog_data = request.data.get('course_catalog')
+            if course_catalog_data:
+                # make sure course catalog is None if its empty
+                course_catalog = course_catalog_data.get('id') if course_catalog_data.get('id') else None
```
iivic: I don't understand this. If course_catalog_data is empty it will not pass the if statement. Please explain and change the comment to cover what is really happening if I misunderstood.

ecommerce/static/js/models/coupon_model.js
```diff
@@ -24,15 +26,23 @@ define([
...
+            multiple_courses: 'Multiple courses',
+            catalog: 'Catalog'
+        };
+
         return Backbone.RelationalModel.extend({
             urlRoot: '/api/v2/coupons/',
 
             defaults: {
                 category: {id: 3, name: 'Affiliate Promotion'},
+                course_catalog: '',
```
iivic: Why is course_catalog set by default? It makes more sense to me to set it when it's actually selected this will save you unnecessary work on the backend.

ecommerce/static/js/models/coupon_model.js
```diff
@@ -185,18 +202,39 @@ define([
...
-                if (this.get('catalog_type') === 'Single course') {
+                if (this.get('catalog_type') === this.catalogTypes.single_course) {
                     if (seats[0]) {
                         seat_data = seats[0].attribute_values;
 
                         this.set('seat_type', this.getCertificateType(seat_data));
                         this.set('course_id', this.getCourseID(seat_data));
                         this.updateTotalValue(this.getSeatPrice());
                     }
+                    this.set('course_catalog', this.defaults.course_catalog);
```
iivic: This is what I mentioned. Why is course_catalog set when catalog_type is not 'Catalog'?

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -114,6 +116,8 @@ define([
             render: function () {
                 var html,
                     category = this.model.get('category').name,
+                    courseCatalog = this.model.get('course_catalog'),
```
iivic: What are you expecting courseCatalog to be and why can it differ (number or something else)?

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -123,9 +127,16 @@ define([
                     price = _s.sprintf('$%s', this.model.get('price'));
                 }
 
+                if (typeof courseCatalog === 'number') {
```
iivic: This check is why I asked https://github.com/edx/ecommerce/pull/1093/files#r96374710. The value shouldn't differ from time to time instead it should be always the same.

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -630,6 +672,21 @@ define([
...
+
+                    if (this.model.has('course_catalog')) {
+                        if (typeof this.model.get('course_catalog') === 'number') {
+                            catalogId = this.model.get('course_catalog');
+                        } else if (!$.isEmptyObject(this.model.get('course_catalog'))) {
+                            catalogId = this.model.get('course_catalog').id;
+                        }
+                    }
+
+                    if (catalogId !== '') {
```
iivic: https://github.com/edx/ecommerce/pull/1093/files#r96374710

iivic (changes requested):

ecommerce/extensions/api/v2/views/coupons.py
zubair-arbi: @iivic the outside `ValidationError` exception handling is for the serializers. This is for our custom validation and its consistent with our previous code logic.

ecommerce/extensions/api/v2/views/coupons.py
zubair-arbi: We aren't displaying these messages to the user instead we are just using these for logs. There were just a few places where we used translations and all other places we didn't so for consistency I removed all translation markings. I you think we should mark these messages for translation then let me know.

ecommerce/extensions/api/v2/views/coupons.py
iivic: The outside ValidationError is used to catch model validation exceptions (and other ValidationErrors) and to handle them appropriately. It will catch this exception as well so this code is unnecessary since it does the same thing. serializers.ValidationError will return a response with status code "400 Bad Request" and the message provided.
My point is you do the same thing twice.

ecommerce/extensions/api/v2/views/coupons.py
iivic: I think you are right. Tag someone with more experience who could enlighten us what to do here.

ecommerce/static/js/views/coupon_detail_view.js
zubair-arbi: This id due to the fact that if there is selected value then the model will return a dict like `{id: '1', name: 'Dummy catalog'}` but if we are editing a coupon then from server we get only course catalog id `1`.

ecommerce/static/js/views/coupon_form_view.js
zubair-arbi: added

ecommerce/extensions/api/v2/views/coupons.py
zubair-arbi: Done

ecommerce/extensions/api/v2/views/coupons.py
zubair-arbi: @mjfrey can you please provide your thoughts on this?

ecommerce/static/js/views/coupon_detail_view.js
zubair-arbi: yes `number` and `dict` as described in my comment below

zubair-arbi:
@iivic I addressed your feedback. Please review again.

ecommerce/extensions/api/v2/views/coupons.py
zubair-arbi: @iivic I have tried changing this but there are lots of tests failing due to this change as this way we are not returning the errors with custom code. I have tried providing error code to `serializers.ValidationError` but it gives error `TypeError: __init__() takes exactly 2 arguments (3 given)`.

iivic: I just confirmed in the django rest framework source code. It looks like we are using an older version of the rest framework which doesn't support the code parameter. What you should do is replace the serializers.ValidationError outside with your code.

ecommerce/static/js/views/coupon_detail_view.js
iivic: That makes sense but this is the coupon details page. Why is it changing on this page?

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -630,6 +672,19 @@ define([
                             this.$('.non-credit-seats').addClass(this.hiddenClass);
                         }
                     }
+
+                    if (this.model.has('course_catalog')) {
+                        if (typeof this.model.get('course_catalog') === 'number') {
```
iivic: Your comment on this matter that says it will be a number on edit page makes me conclude that this will always be the case since this is inside `if (this.editing) {`

ecommerce/static/js/views/coupon_form_view.js
```diff
@@ -630,6 +672,19 @@ define([
...
+                            catalogId = this.model.get('course_catalog');
+                        }
+                    }
+
+                    if (catalogId !== '') {
+                        this.model.set('course_catalog', ecommerce.coupons.catalogs.get(catalogId));
+                    } else {
+                        this.model.set('course_catalog', this.model.defaults.course_catalog);
```
iivic: Once again why are you setting course_catalog if the catalog_type != 'Catalog'?

iivic (changes requested):

ecommerce/extensions/api/v2/views/coupons.py
zubair-arbi: @iivic @mattdrayer I have created a separate story ENT-146 to handle further refactoring as the scope of this story is already big enough.

ecommerce/static/js/views/coupon_detail_view.js
zubair-arbi: @iivic I am going to update this line to check if `courseCatalog` in not empty.

zubair-arbi:
@iivic I have addressed your remaining feedback. Please review.

iivic (approved):
Good job.
Don't forget to squash the commits before merging.

zubair-arbi:
@mattdrayer I have rebased and squashed the changes. Please have a look after that I can merge it.

mattdrayer:
👍 :shipit:

PR #1095, opened 2017-01-16
Revert "Change enrollment code redemption to the receipt page."

vkaracic:
This reverts commit 42527ff34e1fe8c597552f5fe39e35fd3012d5d0.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1095&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mjfrey (approved):

PR #1096, opened 2017-01-17
[SOL-2192] Add number pattern to number input fields in basket checkout.

vkaracic:
This will trigger the numeric keyboard on iOS devices.
Android devices trigger numeric keyboard only on number type input fields, but these fields need to be text fields.
https://openedx.atlassian.net/browse/SOL-2192

### DIFF
--- a/ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
+++ b/ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
@@ -143,7 +143,7 @@
                 <div class="pci-fields row">
                     <div id="card-number" class="form-group col-md-6">
                         <label for="card-number-input" class="control-label">{% trans "Card Number (required)" %}<span class="sr">, {% trans "Secure" %}</span></label>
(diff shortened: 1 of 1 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1096&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

schenedx:
@vkaracic Does this solution prevent a mobile user to input non-numeric characters? If not, LGTM

vkaracic:
I think it does. At least it does for an online iOS emulator: https://appetize.io/demo?device=iphone5s&scale=75&orientation=portrait&osVersion=9.3
I'll remove the pattern from the ZIP/Postal code field since there are postal codes that contain letters, but the other two (CC number and CVN number) can only be numbers so I don't think there is a problem with only the numeric key being available for those.

vkaracic:
@schenedx ^

schenedx (approved):
👍 Glad I asked the question.

PR #1097, opened 2017-01-17
[SOL-2194] Bug fix TypeError: seatData is undefined

iivic:
Upon changing quantity field on coupon creation form console would log an error because the seatData variable is undefined and that happened because the seat_type field is not set. This bug did not interrupt our creation flow but it was annoying so I fixed it.
https://openedx.atlassian.net/browse/SOL-2194
@vkaracic @marjev

### DIFF
--- a/ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
+++ b/ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
@@ -80,6 +80,21 @@ define([
                     expect(view.updateTotalValue).toHaveBeenCalled();
                     expect(view.getSeatData).toHaveBeenCalled();
                 });
+
+                it('should call updateTotalValue when catalog_type "Single course"', function () {
+                    spyOn(view, 'updateTotalValue');
+                    view.changeTotalValue();
+                    expect(view.updateTotalValue).toHaveBeenCalled();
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1097&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
```diff
@@ -80,6 +80,19 @@ define([
                     expect(view.updateTotalValue).toHaveBeenCalled();
                     expect(view.getSeatData).toHaveBeenCalled();
                 });
+
+                it('should call updateTotalValue when catalog_type "Single course"', function () {
+                    spyOn(view, 'updateTotalValue');
+                    view.changeTotalValue();
```
vkaracic, line 86: Add a line that asserts the model's ``catalog_type`` is indeed set to ``Single course``. Same for the unit test below, that it is not set that is.

vkaracic (changes requested):

iivic:
@vkaracic Comment addressed.

vkaracic (approved):

marjev (approved):

PR #1098, opened 2017-01-17
Require login for BasketSummaryView

mjfrey:
Do not allow anonymous access to the basket summary view.

### DIFF
--- a/ecommerce/extensions/basket/app.py
+++ b/ecommerce/extensions/basket/app.py
@@ -11,7 +11,7 @@ class BasketApplication(app.BasketApplication):
 
     def get_urls(self):
         urls = [
-            url(r'^$', self.summary_view.as_view(), name='summary'),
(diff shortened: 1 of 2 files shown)


mjfrey:
@clintonb please take a look.

ecommerce/extensions/basket/views.py
```diff
@@ -86,6 +88,10 @@ class BasketSummaryView(BasketView):
     Display basket contents and checkout/payment options.
     """
 
+    @method_decorator(login_required)
```
clintonb: Implement this in `basket/app.py` similar to what we've done at https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/app.py#L14.

clintonb (approved):
Update the implementation. Also, for future reference, there is a `LoginRequiredMixin` in django-braces.

clintonb (approved):

PR #1100, opened 2017-01-18
Unload basket get_context() into multiple methods.

vkaracic:
``BasketSummaryView``'s ``get_context_data`` method is too big, as in if you add one new line pylint would complain that it has too many statements. In this PR I'm breaking it off to a few private methods.

@marjev @iivic please review.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1100"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev (approved):

iivic (approved):
Looking good. Just make sure it works.

PR #1102, opened 2017-01-20
Using edX Python Social Auth strategy

clintonb:
This strategy incorporates all of our custom settings into a single, shared, location rather than duplicating them in every IDA.

ECOM-6842


clintonb:
@edx/ecommerce please review

rlucioni (approved):

PR #1103, opened 2017-01-20
Fix checkout button analytics attribute.

vkaracic:
The new analytics attribute to track if a checkout happend on the hosted or client side basket wasn't working because of a mistake I did in naming the attribute. This PR fixes that.

Segment data before:
screenshot from 2017-01-20 13-46-45

Segment data after update:
screenshot from 2017-01-20 13-46-13


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1103"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

iivic (approved):

PR #1104, opened 2017-01-23
ENT-121: Automated entitlements applied for enterprise learners

zubair-arbi:
ENT-121

Automatically find and apply active entitlements (coupons) on the order basket, for an enterprise learner.
Added setting `ENTERPRISE_API_CACHE_TIMEOUT = 3600  # Value is in seconds` for enterprise API response cache timeout.

**Q. How to activate this feature?**
**A.** Add active waffle switch with name `enable_enterprise_on_runtime`  to enable this feature

_**NOTE:** This code is written with the assumption that for now every enterprise will have only one entitlement/voucher._

**Related PR for enterprise API (In-progress):** https://github.com/edx/edx-enterprise/pull/30

**Related Config PR:** Add a related [PR in edx configuration repo for overriding the production value for the setting `ENTERPRISE_API_URL`](https://github.com/edx/configuration/pull/3652)

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -330,6 +330,20 @@ class SiteConfiguration(models.Model):
 
         return EdxRestApiClient(settings.COURSE_CATALOG_API_URL, jwt=self.access_token)
 
+    @cached_property
+    def enterprise_api_client(self):
+        """
+        Constructs a Slumber-based REST API client for the provided site.
+
+        Example:
+            site.siteconfiguration.enterprise_api_client.enterprise-learner(learner.username).get()
+
+        Returns:
+            EdxRestApiClient: The client to access the Enterprise service.
+
+        """
+        return EdxRestApiClient(settings.ENTERPRISE_API_URL, jwt=self.access_token)
+
 
 class User(AbstractUser):
     """Custom user model for use with OIDC."""
--- a/ecommerce/coupons/tests/mixins.py
+++ b/ecommerce/coupons/tests/mixins.py
@@ -50,12 +50,17 @@ class CourseCatalogMockMixin(object):
             content_type='application/json'
         )
 
-    def mock_dynamic_catalog_course_runs_api(self, course_run=None, query=None, course_run_info=None):
-        """ Helper function to register a dynamic course catalog API endpoint for the course run information. """
+    def mock_dynamic_catalog_course_runs_api(self, course_run=None, partner_code=None, query=None,
+                                             course_run_info=None):
+        """
+        Helper function to register a course catalog API endpoint for getting
+        course runs information.
+        """
         if not course_run_info:
             course_run_info = {
                 'count': 1,
-                'next': 'path/to/next/page',
+                'next': None,
+                'previous': None,
                 'results': [{
                     'key': course_run.id,
                     'title': course_run.name,
@@ -78,6 +83,18 @@ class CourseCatalogMockMixin(object):
         httpretty.register_uri(
             httpretty.GET,
             course_run_url_with_query,
+            body=course_run_info_json,
+            content_type='application/json'
+        )
+
+        course_run_url_with_query_and_partner_code = '{}course_runs/?q={}&partner={}'.format(
+            settings.COURSE_CATALOG_API_URL,
+            partner_code if partner_code else 'edx',
+            query if query else 'id:course*'
+        )
+        httpretty.register_uri(
+            httpretty.GET,
+            course_run_url_with_query_and_partner_code,
             body=course_run_info_json,
             content_type='application/json'
         )
--- a/ecommerce/coupons/utils.py
+++ b/ecommerce/coupons/utils.py
@@ -5,34 +5,91 @@ from django.conf import settings
 from django.core.cache import cache
 from oscar.core.loading import get_model
 
+from ecommerce.courses.utils import traverse_pagination
+
+
 Product = get_model('catalogue', 'Product')
 
 
-def get_range_catalog_query_results(limit, query, site, offset=None):
+def get_catalog_course_runs(site, query, limit=None, offset=None):
     """
-    Get catalog query results
+    Get course runs for a site on the basis of provided query from the Course
+    Catalog API.
+
+    This method will get all course runs by recursively retrieving API
+    next urls in the API response if no limit is provided.
 
     Arguments:
         limit (int): Number of results per page
+        offset (int): Page offset
         query (str): ElasticSearch Query
         site (Site): Site object containing Site Configuration data
-        offset (int): Page offset
 
+    Example:
+        >>> get_catalog_course_runs(site, query, limit=1)
+        {
+            "count": 1,
+            "next": "None",
+            "previous": "None",
+            "results": [{
+                "key": "course-v1:edX+DemoX+Demo_Course",
+                "title": edX Demonstration Course,
+                "start": "2016-05-01T00:00:00Z",
+                "image": {
+                    "src": "path/to/the/course/image"
+                },
+                "enrollment_end": None
+            }],
+        }
     Returns:
-        dict: Query seach results received from Course Catalog API
+        dict: Query search results for course runs received from Course
+            Catalog API
+
+    Raises:
+        ConnectionError: requests exception "ConnectionError"
+        SlumberBaseException: slumber exception "SlumberBaseException"
+        Timeout: requests exception "Timeout"
+
     """
+    api_resource_name = 'course_runs'
     partner_code = site.siteconfiguration.partner.short_code
-    cache_key = 'course_runs_{}_{}_{}_{}'.format(query, limit, offset, partner_code)
+    cache_key = '{site_domain}_{partner_code}_{resource}_{query}_{limit}_{offset}'.format(
+        site_domain=site.domain,
+        partner_code=partner_code,
+        resource=api_resource_name,
+        query=query,
+        limit=limit,
+        offset=offset
+    )
     cache_key = hashlib.md5(cache_key).hexdigest()
+
     response = cache.get(cache_key)
     if not response:
-        response = site.siteconfiguration.course_catalog_api_client.course_runs.get(
-            limit=limit,
-            offset=offset,
-            q=query,
-            partner=partner_code
-        )
+        api = site.siteconfiguration.course_catalog_api_client
+        endpoint = getattr(api, api_resource_name)
+
+        if limit:
+            response = endpoint().get(
+                partner=partner_code,
+                q=query,
+                limit=limit,
+                offset=offset
+            )
+        else:
+            response = endpoint().get(
+                partner=partner_code,
+                q=query
+            )
+            all_response_results = traverse_pagination(response, endpoint)
+            response = {
+                'count': len(all_response_results),
+                'next': 'None',
+                'previous': 'None',
+                'results': all_response_results,
+            }
+
         cache.set(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
+
     return response
 
 
--- a/ecommerce/enterprise/__init__.py
+++ b/ecommerce/enterprise/__init__.py
@@ -1,0 +1,4 @@
+"""
+This package contains all workflows and communications related to the Open edX
+Enterprise service.
+"""
--- a/ecommerce/enterprise/entitlements.py
+++ b/ecommerce/enterprise/entitlements.py
@@ -1,0 +1,278 @@
+"""
+Helper methods for getting site based enterprise entitlements against the
+learners.
+
+Enterprise learners can get coupons, offered by their respective Enterprise
+customers with which they are affiliated. The coupon product id's for the
+enterprise entitlements are provided by the Enterprise Service on the basis
+of the learner's enterprise eligibility criterion.
+"""
+import hashlib
+import logging
+
+from django.conf import settings
+from django.core.cache import cache
+from oscar.core.loading import get_model
+from requests.exceptions import ConnectionError, Timeout
+from slumber.exceptions import SlumberBaseException
+
+from ecommerce.coupons.utils import get_catalog_course_runs
+from ecommerce.coupons.views import voucher_is_valid
+from ecommerce.courses.utils import get_course_catalogs
+from ecommerce.enterprise.utils import is_enterprise_feature_enabled
+from ecommerce.extensions.api.serializers import retrieve_voucher
+
+
+logger = logging.getLogger(__name__)
+Product = get_model('catalogue', 'Product')
+Voucher = get_model('voucher', 'Voucher')
+
+
+def get_entitlement_voucher(request, product):
+    """
+    Returns entitlement voucher for the given product against an enterprise
+    learner.
+
+    Arguments:
+        request (HttpRequest): request with voucher data
+        product (Product): A product that has course_key as attribute (seat or
+            bulk enrollment coupon)
+
+    """
+    if not is_enterprise_feature_enabled():
+        return None
+
+    vouchers = get_vouchers_for_learner(request.site, request.user)
+    if vouchers:
+        entitlement_voucher = get_available_voucher_for_product(request, product, vouchers)
+        return entitlement_voucher
+
+    return None
+
+
+def get_vouchers_for_learner(site, user):
+    """
+    Get vouchers against the list of all enterprise entitlements for the
+    provided learner.
+
+    Arguments:
+        site: (django.contrib.sites.Site) site instance
+        user: (django.contrib.auth.User) django auth user
+
+    Returns:
+        list of Voucher class objects
+
+    """
+    entitlements = get_entitlements_for_learner(site, user)
+    if not entitlements:
+        return None
+
+    vouchers = []
+    for entitlement in entitlements:
+        try:
+            coupon_product = Product.objects.filter(product_class__name='Coupon').get(id=entitlement['entitlement_id'])
+        except Product.DoesNotExist:
+            logger.exception(
+                'There was an error getting coupon product with the entitlement id %s',
+                entitlement['entitlement_id']
+            )
+            return None
+
+        entitlement_voucher = retrieve_voucher(coupon_product)
+        vouchers.append(entitlement_voucher)
+
+    return vouchers
+
+
+def get_entitlements_for_learner(site, user):
+    """
+    Get entitlements for the provided learner if the provided learner is
+    affiliated with an enterprise.
+
+    Arguments:
+        site: (django.contrib.sites.Site) site instance
+        user: (django.contrib.auth.User) django auth user
+
+    """
+    try:
+        enterprise_learner_data = get_enterprise_learner_data(site, user)['results']
+    except (ConnectionError, SlumberBaseException, Timeout, KeyError, TypeError):
+        logger.exception(
+            'Failed to retrieve enterprise info for the learner [%s]',
+            user.username
+        )
+        return None
+
+    if not enterprise_learner_data:
+        logger.info('Learner with username [%s] in not affiliated with any enterprise', user.username)
+        return None
+
+    try:
+        entitlements = enterprise_learner_data[0]['enterprise_customer']['enterprise_customer_entitlements']
+    except KeyError:
+        logger.error('Invalid structure for enterprise learner API response for the learner [%s]', user.username)
+        return None
+
+    return entitlements
+
+
+def get_enterprise_learner_data(site, user):
+    """
+    Fetch information related to enterprise and its entitlements according to
+    the eligibility criterion for the provided learners from the Enterprise
+    Service.
+
+    Example:
+        get_enterprise_learner_data(site, user)
+
+    Arguments:
+        site: (django.contrib.sites.Site) site instance
+        user: (django.contrib.auth.User) django auth user
+
+    Returns:
+        dict: {
+            "enterprise_api_response_for_learner": {
+                "count": 1,
+                "num_pages": 1,
+                "current_page": 1,
+                "results": [
+                    {
+                        "enterprise_customer": {
+                            "uuid": "cf246b88-d5f6-4908-a522-fc307e0b0c59",
+                            "name": "TestShib",
+                            "catalog": 2,
+                            "active": true,
+                            "site": {
+                                "domain": "example.com",
+                                "name": "example.com"
+                            },
+                            "enable_data_sharing_consent": true,
+                            "enforce_data_sharing_consent": "at_login",
+                            "enterprise_customer_users": [
+                                1
+                            ],
+                            "branding_configuration": {
+                                "enterprise_customer": "cf246b88-d5f6-4908-a522-fc307e0b0c59",
+                                "logo": "https://open.edx.org/sites/all/themes/edx_open/logo.png"
+                            },
+                            "enterprise_customer_entitlements": [
+                                {
+                                    "enterprise_customer": "cf246b88-d5f6-4908-a522-fc307e0b0c59",
+                                    "entitlement_id": 69
+                                }
+                            ]
+                        },
+                        "user_id": 5,
+                        "user": {
+                            "username": "staff",
+                            "first_name": "",
+                            "last_name": "",
+                            "email": "staff@example.com",
+                            "is_staff": true,
+                            "is_active": true,
+                            "date_joined": "2016-09-01T19:18:26.026495Z"
+                        },
+                        "data_sharing_consent": [
+                            {
+                                "user": 1,
+                                "state": "enabled",
+                                "enabled": true
+                            }
+                        ]
+                    }
+                ],
+                "next": null,
+                "start": 0,
+                "previous": null
+            }
+        }
+
+    Raises:
+        ConnectionError: requests exception "ConnectionError"
+        SlumberBaseException: slumber exception "SlumberBaseException"
+        Timeout: requests exception "Timeout"
+
+    """
+    api_resource_name = 'enterprise-learner'
+    partner_code = site.siteconfiguration.partner.short_code
+    cache_key = '{site_domain}_{partner_code}_{resource}_{username}'.format(
+        site_domain=site.domain,
+        partner_code=partner_code,
+        resource=api_resource_name,
+        username=user.username
+    )
+    cache_key = hashlib.md5(cache_key).hexdigest()
+
+    response = cache.get(cache_key)
+    if not response:
+        api = site.siteconfiguration.enterprise_api_client
+        endpoint = getattr(api, api_resource_name)
+        querystring = {'username': user.username}
+        response = endpoint().get(**querystring)
+        cache.set(cache_key, response, settings.ENTERPRISE_API_CACHE_TIMEOUT)
+
+    return response
+
+
+def get_available_voucher_for_product(request, product, vouchers):
+    """
+    Get first active entitlement from a list of vouchers for the given
+    product.
+
+    Arguments:
+        product (Product): A product that has course_key as attribute (seat or
+            bulk enrollment coupon)
+        request (HttpRequest): request with voucher data
+        vouchers: (List) List of voucher class objects for an enterprise
+
+    """
+    for voucher in vouchers:
+        is_valid_voucher, __ = voucher_is_valid(voucher, [product], request)
+        if is_valid_voucher:
+            voucher_course_ids = get_course_ids_from_voucher(request.site, voucher)
+            if product.course_id in voucher_course_ids:
+                return voucher
+
+
+def get_course_ids_from_voucher(site, voucher):
+    """
+    Get site base list of course run keys from the provided voucher object.
+
+    Arguments:
+        site: (django.contrib.sites.Site) site instance
+        voucher (Voucher): voucher class object
+
+    Returns:
+        list of course ids
+
+    """
+    voucher_offer = voucher.offers.first()
+    offer_range = voucher_offer.condition.range
+    if offer_range.course_catalog:
+        try:
+            course_catalog = get_course_catalogs(site=site, resource_id=offer_range.course_catalog)
+        except (ConnectionError, SlumberBaseException, Timeout):
+            logger.error('Unable to connect to Course Catalog service for course catalogs.')
+            return None
+
+        try:
+            course_runs = get_catalog_course_runs(site, course_catalog.get('query'))['results']
+        except (ConnectionError, SlumberBaseException, Timeout, KeyError):
+            logger.error('Unable to get course runs from Course Catalog service.')
+            return None
+
+        voucher_course_ids = [course_run.get('key') for course_run in course_runs if course_run.get('key')]
+    elif offer_range.catalog_query:
+        try:
+            course_runs = get_catalog_course_runs(site, offer_range.catalog_query)['results']
+        except (ConnectionError, SlumberBaseException, Timeout, KeyError):
+            logger.error('Unable to get course runs from Course Catalog service.')
+            return None
+
+        voucher_course_ids = [course_run.get('key') for course_run in course_runs if course_run.get('key')]
+    else:
+        stock_records = offer_range.catalog.stock_records.all()
+        seats = Product.objects.filter(id__in=[sr.product.id for sr in stock_records])
+        voucher_course_ids = [seat.course_id for seat in seats]
+
+    return voucher_course_ids
--- a/ecommerce/enterprise/utils.py
+++ b/ecommerce/enterprise/utils.py
@@ -1,0 +1,22 @@
+"""
+Helper methods for enterprise app.
+"""
+from django.conf import settings
+import waffle
+
+
+def is_enterprise_feature_enabled():
+    """
+    Returns boolean indicating whether enterprise feature is enabled or
+    disabled.
+
+    Example:
+        >> is_enterprise_feature_enabled()
+        True
+
+    Returns:
+         (bool): True if enterprise feature is enabled else False
+
+    """
+    is_enterprise_enabled = waffle.switch_is_active(settings.ENABLE_ENTERPRISE_ON_RUNTIME_SWITCH)
+    return is_enterprise_enabled
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -16,12 +16,14 @@ from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME, SEAT_PR
 from ecommerce.core.exceptions import SiteConfigurationError
 from ecommerce.core.url_utils import get_lms_url
 from ecommerce.courses.utils import get_certificate_type_display_value, get_course_info_from_catalog, mode_for_seat
+from ecommerce.enterprise.entitlements import get_entitlement_voucher
 from ecommerce.extensions.analytics.utils import prepare_analytics_data
 from ecommerce.extensions.basket.utils import get_basket_switch_data, prepare_basket
 from ecommerce.extensions.offer.utils import format_benefit_value
 from ecommerce.extensions.partner.shortcuts import get_partner_for_site
 from ecommerce.extensions.payment.constants import CLIENT_SIDE_CHECKOUT_FLAG_NAME
 from ecommerce.extensions.payment.forms import PaymentForm
+
 
 Benefit = get_model('offer', 'Benefit')
 logger = logging.getLogger(__name__)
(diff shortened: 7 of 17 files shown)


ecommerce/enterprise/__init__.py
```diff
@@ -0,0 +1 @@
+
```
mattdrayer: We'll need to add some documentation here describing the package.

ecommerce/core/models.py
```diff
@@ -330,6 +330,17 @@ def course_catalog_api_client(self):
 
         return EdxRestApiClient(settings.COURSE_CATALOG_API_URL, jwt=self.access_token)
 
+    @cached_property
+    def enterprise_api_client(self):
+        """
+        Returns an API client to access the Enterprise service.
```
mattdrayer: This line is redundant, so we should revise to something more explanatory -- perhaps describe the intended use, how to create a client, any gotchas to watch out for, exception conditions, etc.

ecommerce/enterprise/tmp/__init__.py
```diff
@@ -0,0 +1 @@
+
```
mattdrayer: Hmm, after looking at this directory...it's sort of a weird sub-package -- is there a better name to use versus `tmp`?  Can it be entirely eliminated?  If it *must* remain, we'll need to be very clear with our documentation about its intended purpose/use.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,116 @@
...
+    product.
+
+    Arguments:
+        product (Product): A product that has course_key as attribute (seat or
+            bulk enrollment coupon)
+        request (HttpRequest): request with voucher data
+        vouchers: (List) List of voucher class objects for an enterprise
+
+    """
+    # TODO: Handle multiple entitlements/vouchers for an enterprise WRT some criterion
```
mattdrayer: Will this TODO be addressed in this PR?  If not, I recommend logging a ticket versus keeping track of future development work in the code.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    Fetch data related to enterprise learners.
+
+    Arguments:
+        learner: (django.contrib.auth.User) django auth user
+        site: (django.contrib.sites.Site) site instance
+
+    """
+    response = site.siteconfiguration.enterprise_api_client.enterprise-learner(learner.username).get()
+    # TODO: Cache the response from enterprise API in case of 200 status
```
mattdrayer: Will this TODO be addressed in this PR? If not, I recommend logging a ticket versus keeping track of future development work in the code.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,116 @@
+"""
+Helper methods for enterprise entitlements.
```
mattdrayer: This line doesn't really explain the purpose of this module to an edx/ecommerce project developer.  What is an "enterprise entitlement"?  How do I use it?

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,116 @@
...
+        learner: (django.contrib.auth.User) django auth user
+        site: (django.contrib.sites.Site) site instance
+
+    """
+    vouchers = []
+    entitlements = get_entitlements_for_learner(site, learner)
+    for entitlement in entitlements:
+        try:
+            voucher = Voucher.objects.get(id=entitlement['entitlement_id'])
+            vouchers.append(voucher)
```
mattdrayer: It looks like this line should be relocated to below the `try/except`, because it's not the actual work requiring protection.  We could add a `continue` to the `except` in order to skip to the next iteration in the case of a failure. and perform the append after we have ascertained the `Voucher` call was successful.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Arguments:
+        learner: (django.contrib.auth.User) django auth user
+        site: (django.contrib.sites.Site) site instance
+
+    """
+    entitlements = []
+    try:
+        enterprise_learner_data = get_enterprise_learner_data(site, learner)['results']
+    except:  # pylint: disable=bare-except
```
mattdrayer: Is it not feasible to be explicit about the possible exceptions here?

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,116 @@
...
+    Get entitlements for the provided learner if the provided learner is
+    affiliated with an enterprise.
+
+    Arguments:
+        learner: (django.contrib.auth.User) django auth user
+        site: (django.contrib.sites.Site) site instance
+
+    """
+    entitlements = []
+    try:
```
mattdrayer: I think it would be useful to describe the expected structure of the enterprise_learner_data in a code comment here.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,116 @@
...
+        return entitlements
+
+    if len(enterprise_learner_data) == 0:
+        logger.info('Learner with username %s in not affiliated with any enterprise', learner.username)
+        return entitlements
+
+    entitlements = enterprise_learner_data[0]['enterprise_customer']['entitlements']
+    return entitlements
+
+@utils.dummy_data("enterprise_api_response_for_learner")
```
mattdrayer: I guess I'm not really understanding the whole `tmp/utils` thing that is going on here.  Is this a supporting mechanism for the test harness?  It seems like there should be a more straightforward way of accomplishing what is needed.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,116 @@
...
+def get_enterprise_learner_data(site, learner):
+    """
+    Fetch data related to enterprise learners.
```
mattdrayer: What data?  What does it look like?  What is the failure mode?  All good information to include.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Arguments:
+        learner: (django.contrib.auth.User) django auth user
+        site: (django.contrib.sites.Site) site instance
+
+    """
+    response = site.siteconfiguration.enterprise_api_client.enterprise-learner(learner.username).get()
```
mattdrayer: It feels like we should have a `try/except` here, or at least a gate on the `learner` object

ecommerce/enterprise/tmp/utils.py
```diff
@@ -0,0 +1,16 @@
+import json
+import os
+
+from django.contrib.auth.models import User
+
+
+def dummy_data(key):
```
mattdrayer: I know that this is a decorator based on my code review, and the style/layout of the function, but I still think a code comment or docstring would be good to have here.

ecommerce/enterprise/entitlements.py
zubair-arbi: @mattdrayer This `tmp\utils` code is just for temporary purpose. I will remove it once I have the functionality working.

ecommerce/enterprise/tmp/__init__.py
zubair-arbi: @mattdrayer don't worry, this will be entirely eliminated :)

ecommerce/enterprise/entitlements.py
zubair-arbi: @mattdrayer I have added this as an output from the method `get_enterprise_learner_data` as it makes more sense to put it there.

ecommerce/enterprise/entitlements.py
zubair-arbi: @mattdrayer I have added a ticket https://openedx.atlassian.net/browse/ENT-159 for this TODO.

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,39 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
     return response
 
 
+def get_all_range_catalog_query_results(site, query):
```
saleem-latif: nit: wouldn't it be better to have a name something like `get_catalog_course_runs`

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,230 @@
...
+        site: (django.contrib.sites.Site) site instance
+
+    """
+    vouchers = []
+    entitlements = get_entitlements_for_learner(site, learner)
+    for entitlement in entitlements:
+        try:
+            coupon_product = Product.objects.filter(product_class__name='Coupon').get(id=entitlement['entitlement_id'])
+        except Product.DoesNotExist:
+            logger.warning(
```
saleem-latif: It would be better to use `logger.exception` as it will also log error trace

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,230 @@
...
+    affiliated with an enterprise.
+
+    Arguments:
+        learner: (django.contrib.auth.User) django auth user
+        site: (django.contrib.sites.Site) site instance
+
+    """
+    entitlements = []
+    try:
+        enterprise_learner_data = get_enterprise_learner_data(site, learner)['results']
```
saleem-latif: Please handle `KeyError` here

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,230 @@
...
+            'Failed to retrieve enterprise info for the learner [%s]',
+            learner.username
+        )
+        return entitlements
+
+    if len(enterprise_learner_data) == 0:
+        logger.info('Learner with username %s in not affiliated with any enterprise', learner.username)
+        return entitlements
+
+    entitlements = enterprise_learner_data[0]['enterprise_customer']['entitlements']
```
saleem-latif: Add `KeyError` exception handling.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,230 @@
...
+
+    """
+    resource = 'enterprise-learner'
+    partner_code = site.siteconfiguration.partner.short_code
+    cache_key = '{}_{}_{}.api.data'.format(site.domain, partner_code, resource)
+    cache_key = hashlib.md5(cache_key).hexdigest()
+
+    response = cache.get(cache_key)
+    if not response:
+        response = site.siteconfiguration.enterprise_api_client.enterprise-learner(learner.username).get()
```
saleem-latif: Add network error handling for this API call.

ecommerce/enterprise/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
+Client app for or containing all enterprise related functionality and
+communication with the Enterprise Service.
```
asadiqbal08: can you make it better? specially `... and communication with the Enterprise Service.`

ecommerce/extensions/basket/views.py
```diff
@@ -51,6 +53,15 @@ def get(self, request):
         except StockRecord.DoesNotExist:
             return HttpResponseBadRequest(_('SKU [{sku}] does not exist.').format(sku=sku))
 
+        if voucher is None:
```
saleem-latif, line 56: Can we update it to something like this.
```
if voucher is None:
    voucher  = get_entitlement_voucher(request, product)
```

ecommerce/enterprise/utils.py
```diff
@@ -0,0 +1,27 @@
...
+        True
+
+    Returns:
+         (bool): True if enterprise feature is enabled else False
+
+    """
+    # Return True if we are currently processing a request and enterprise
+    # feature is enabled via the waffle switch
+    if waffle.switch_is_active(settings.ENABLE_ENTERPRISE_ON_RUNTIME_SWITCH):
+        return True
```
asadiqbal08: make it simple 
`return waffle.switch_is_active(settings.ENABLE_ENTERPRISE_ON_RUNTIME_SWITCH)`
OR 
```
is_enterprise_enabled = waffle.switch_is_active(settings.ENABLE_ENTERPRISE_ON_RUNTIME_SWITCH)
return is_enterprise_enabled
```

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,39 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+        site (Site): Site object containing Site Configuration data
+
+    Returns:
+        dict: Query search results received from Course Catalog API
+
+    """
+    resource = 'course_runs'
+    partner_code = site.siteconfiguration.partner.short_code
+    cache_key = '{}_{}_{}_{}.api.data'.format(site.domain, partner_code, resource, query)
+    cache_key = hashlib.md5(cache_key).hexdigest()
```
saleem-latif: I think you forgot to import `hashlib` and `cache`, I am getting this error
```
  File "/edx/app/ecommerce/ecommerce/ecommerce/extensions/basket/views.py", line 58, in get
    enterprise_voucher = get_entitlement_voucher(request, product)
  File "/edx/app/ecommerce/ecommerce/ecommerce/enterprise/entitlements.py", line 43, in get_entitlement_voucher
    vouchers = get_vouchers_for_learner(request.site, request.user)
  File "/edx/app/ecommerce/ecommerce/ecommerce/enterprise/entitlements.py", line 59, in get_vouchers_for_learner
    entitlements = get_entitlements_for_learner(site, learner)
  File "/edx/app/ecommerce/ecommerce/ecommerce/enterprise/entitlements.py", line 87, in get_entitlements_for_learner
    enterprise_learner_data = get_enterprise_learner_data(site, learner)['results']
  File "/edx/app/ecommerce/ecommerce/ecommerce/enterprise/entitlements.py", line 176, in get_enterprise_learner_data
    cache_key = hashlib.md5(cache_key).hexdigest()
NameError: global name 'hashlib' is not defined
```

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,230 @@
...
+        cache.set(cache_key, response, settings.ENTERPRISE_API_CACHE_TIMEOUT)
+
+    return response
+
+def get_available_voucher_for_product(request, product, vouchers):
```
saleem-latif: there should be 2 blank lines between module level functions

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,230 @@
...
+
+    """
+    for voucher in vouchers:
+        is_valid_voucher, __ = voucher_is_valid(voucher, [product], request)
+        if is_valid_voucher:
+            voucher_course_ids = get_course_ids_from_voucher(request.site, voucher)
+            if product.course_id in voucher_course_ids:
+                return voucher
+
+def get_course_ids_from_voucher(site, voucher):
```
saleem-latif: 2 blank line instead of 1, other functions in the module also need to be fixed.

ecommerce/coupons/utils.py
zubair-arbi: Thanks for pointing this out. The imports were missing in other file `enterprise/entitlements.py`

ecommerce/enterprise/entitlements.py
zubair-arbi: This is the desired flow. This method will raise exceptions and these exception will be handled in the method `get_entitlements_for_learner` where we are calling this method.

ecommerce/extensions/basket/views.py
zubair-arbi: updated

zubair-arbi:
@mattdrayer @iivic @mjfrey This PR is ready for review. I will start adding unittests by tomorrow. You cannot run it on local machine as it depends on an Enterprise API endpoint which is in progress.

ecommerce/core/models.py
```diff
@@ -330,6 +330,20 @@ def course_catalog_api_client(self):
 
         return EdxRestApiClient(settings.COURSE_CATALOG_API_URL, jwt=self.access_token)
 
+    @cached_property
+    def enterprise_api_client(self):
+        """
+        Returns an Slumber based REST API client for the provided site.
```
mattdrayer: "Returns a Slumber-based..."

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,39 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
     return response
 
 
+def get_catalog_course_runs(site, query):
+    """
+    Get all catalogs on the basis of provided query from the Course Catalog
```
mattdrayer: Are we getting catalogs or course runs?

ecommerce/core/models.py
mattdrayer: Actually instead of "Returns" here you might say "Constructs"

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,39 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+        query (str): ElasticSearch Query
+        site (Site): Site object containing Site Configuration data
+
+    Returns:
+        dict: Query search results received from Course Catalog API
+
+    """
+    resource = 'course_runs'
+    partner_code = site.siteconfiguration.partner.short_code
+    cache_key = '{}_{}_{}_{}.api.data'.format(site.domain, partner_code, resource, query)
```
mattdrayer: This cache key format follows a similar pattern to other cache keys found in Otto, yes?

ecommerce/enterprise/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
+Client app for or containing all enterprise related functionality and
```
mattdrayer: This sentence needs to be revised -- how about something like:

"This package contains all workflows and communications related to the Open edX Enterprise service.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,244 @@
+"""
+Helper methods for getting site based enterprise entitlements against the
+learners.
+
+Enterprise learners can get coupons, offered by their respective Enterprise
+customers with which they are affiliated. The coupon product id's for the
+enterprise entitlements are provided by the Enterprise Service on the bases
```
mattdrayer: bases -> basis (of)

mattdrayer:
This is looking good so far, @zubair-arbi -- just a couple small nits from me at this point since you've already addressed the big things from my earlier review -- 👍 from my side.  @mjfrey @iivic @vkaracic if one of you guys can take a pass through this PR as well that would be great.

mattdrayer:
FYI, the PR for the Enterprise service API is https://github.com/edx/edx-enterprise/pull/30

ecommerce/coupons/utils.py
zubair-arbi: correct, I rearranged the key structure a bit for better readability and their precedence, e.g. putting site domain at start of key for site base uniqueness.

ecommerce/enterprise/__init__.py
zubair-arbi: Sounds good

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,244 @@
...
+    learner.
+
+    Arguments:
+        request (HttpRequest): request with voucher data
+        product (Product): A product that has course_key as attribute (seat or
+            bulk enrollment coupon)
+
+    """
+    if not is_enterprise_feature_enabled():
+        return None
```
asadiqbal08, line 43: `logger.info(` here ? `enterprise feature is not enabled`

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,244 @@
...
+        get_enterprise_learner_data(site, learner)
+
+    Arguments:
+        learner: (django.contrib.auth.User) django auth user
+        site: (django.contrib.sites.Site) site instance
+
+    Returns:
+        dict: {
+            "enterprise_api_response_for_learner": {
+                "count": 1,
```
asadiqbal08, line 135: is the api always return count 1 ?

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,244 @@
...
+    try:
+        enterprise_learner_data = get_enterprise_learner_data(site, learner)['results']
+    except (ConnectionError, SlumberBaseException, Timeout, KeyError, TypeError):
+        logger.exception(
+            'Failed to retrieve enterprise info for the learner [%s]',
+            learner.username
+        )
+        return []
+
+    if len(enterprise_learner_data) == 0:
```
asadiqbal08: `if not enterprise_learner_data:`

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,244 @@
...
+    Arguments:
+        learner: (django.contrib.auth.User) django auth user
+        site: (django.contrib.sites.Site) site instance
+
+    """
+    vouchers = []
+    entitlements = get_entitlements_for_learner(site, learner)
+    for entitlement in entitlements:
+        try:
+            coupon_product = Product.objects.filter(product_class__name='Coupon').get(id=entitlement['entitlement_id'])
```
asadiqbal08: can you add a comment line that indicate the structure of coupon_product after executing the query ?

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,244 @@
...
+    vouchers = get_vouchers_for_learner(request.site, request.user)
+    entitlement_voucher = get_available_voucher_for_product(request, product, vouchers)
+
+    return entitlement_voucher
+
+
+def get_vouchers_for_learner(site, learner):
+    """
+    Get vouchers against the list of all enterprise entitlements for the
+    provided learner.
```
asadiqbal08: kindly add the vouchers structure that is going to be returned.

ecommerce/enterprise/entitlements.py
zubair-arbi: Since this feature is enabled explicitly by a waffle switch I don't think it will help much in production.

ecommerce/enterprise/entitlements.py
zubair-arbi: @asadiqbal08 Clean code suggests to avoid comments as much as possible and use expressive names for readability, since here we can check the model for fields so its unnecessary to explicitly add Product class structure here. The basic reason for this is that code is ever evolving and being refactored by the developers but this is not true for the comments specifically if they are representing the data structure. If the related class/model is changed then there will be no error to change the class/model structure in the comments and it would miss guide the new developer.

ecommerce/enterprise/entitlements.py
zubair-arbi: Yes, In case the learner is affiliated with an enterprise.

ecommerce/enterprise/entitlements.py
asadiqbal08: I was not saying to add the whole Product class structure here :)

ecommerce/enterprise/entitlements.py
zubair-arbi: @asadiqbal08 I have add `Returns` statement.

ecommerce/enterprise/entitlements.py
zubair-arbi: @asadiqbal08 I am sorry if I miss understood. If you were referring to the entitlement structure then I have added a complete response structure of enterprise learner API response in the method `get_enterprise_learner_data` and I think I will be a duplicate to add a portion of that structure here again.

zubair-arbi:
@asadiqbal08 I have addressed you feedback.
@mattdrayer @saleem-latif @mjfrey @iivic I have added unittests too. Please review.

mattdrayer:
I'm good with the changes -- 👍 -- @mjfrey @iivic @vkaracic @marjev please jump in when available.

ecommerce/coupons/tests/mixins.py
```diff
@@ -112,6 +112,41 @@ def mock_dynamic_catalog_contains_api(self, course_run_ids, query):
             content_type='application/json'
         )
 
+    def mock_get_catalog_course_runs_api(self, course_run, query, partner_code):
```
mjfrey: we should just augment mock_dynamic_catalog_single_course_runs_api rather than adding a new mock.

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,39 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
     return response
 
 
+def get_catalog_course_runs(site, query):
```
mjfrey: why this new method? We already have get_range_catalog_query_results

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,247 @@
...
+        api = site.siteconfiguration.enterprise_api_client
+        endpoint = getattr(api, resource)
+        querystring = {'username': learner.username}
+        response = endpoint().get(**querystring)
+        cache.set(cache_key, response, settings.ENTERPRISE_API_CACHE_TIMEOUT)
+
+    return response
+
+
+def get_available_voucher_for_product(request, product, vouchers):
```
mjfrey, line 217: there is already get_voucher_and_products_from_code that does basically what you are trying to do here.  If it does not do exactly what you want -- we should update it.

mjfrey (changes requested):
quick first pass.  Seems that we can re-use a bunch of logic we already have regarding Voucher / catalog lookup.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,247 @@
...
+    return entitlement_voucher
+
+
+def get_vouchers_for_learner(site, learner):
+    """
+    Get vouchers against the list of all enterprise entitlements for the
+    provided learner.
+
+    Arguments:
+        learner: (django.contrib.auth.User) django auth user
```
marjev: nitpick: I think `User` will suffice, mostly for consistency because I think we used only `User` in other docstrings. 
Same think with the `Site`.

ecommerce/enterprise/entitlements.py
marjev: Can we optimize this by doing the following:
```
entitlement_ids = [entitlement['entitlement_id'] for entitlement in entitlements]
coupon_products = Product.objects.filter(product_class__name='Coupon').filter(id__in=entitlement_ids)
```

This way, only a single query is made to the database. You get a set of coupon_products in a single call. Use a loop to iterate over the results. No need to exception handling either, just check to see if the result set is empty or not.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,247 @@
...
+    partner_code = site.siteconfiguration.partner.short_code
+    cache_key = '{}_{}_{}.api.data'.format(site.domain, partner_code, resource)
+    cache_key = hashlib.md5(cache_key).hexdigest()
+
+    response = cache.get(cache_key)
+    if not response:
+        api = site.siteconfiguration.enterprise_api_client
+        endpoint = getattr(api, resource)
+        querystring = {'username': learner.username}
+        response = endpoint().get(**querystring)
```
marjev: If think you should handle API exceptions here and not in the place where you call `get_enterprise_learner_data()`.
The reasons why I say this are:
1. If this is ever reused, it would require another set of exception handling. It's basically duplicating code. Someone can, also, forget to add exception handling, as well, which means this method would not handle the exception it caused.
2. This is open-source code, so we should either handle the exceptions our code stumbles upon or document in docstring that we're raising/passing an exception.

I'm suggesting moving the exception handling from https://github.com/edx/ecommerce/pull/1104/files#diff-50449a66b32057163f5619acfe7bc17eR94 to this line, since this line might cause an exception to be raised.

ecommerce/coupons/utils.py
```diff
@@ -36,6 +39,39 @@ def get_range_catalog_query_results(limit, query, site, offset=None):
...
+    resource = 'course_runs'
+    partner_code = site.siteconfiguration.partner.short_code
+    cache_key = '{}_{}_{}_{}.api.data'.format(site.domain, partner_code, resource, query)
+    cache_key = hashlib.md5(cache_key).hexdigest()
+
+    response = cache.get(cache_key)
+    if not response:
+        api = site.siteconfiguration.course_catalog_api_client
+        endpoint = getattr(api, resource)
+        response = endpoint().get(
```
marjev: If this method is not removed, as suggested by Michael in a comment above, you should handle API exceptions this might raise.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,247 @@
+"""
```
marjev, line 1: @zubair-arbi @mjfrey There are a lot of methods in this file that do something when a voucher is sent as an argument.

I'm wondering if we should move those to voucher utils or Voucher model. Subjectively, I would never think of looking for such methods in this file, or even this app (`enterprise`). I would expect something like that to be within the `voucher` app.

Let me know what are your thoughts on this matter.

marjev (changes requested):

ecommerce/coupons/utils.py
zubair-arbi: @mjfrey  the main difference is that this method gets all catalog course runs by recursively retrieving the results. I will see if I can refactor the the method `get_range_catalog_query_results` to include this functionality.

ecommerce/coupons/tests/mixins.py
zubair-arbi: Since I need to mock the course_runs api so I will use the mock `mock_dynamic_catalog_course_runs_api`

ecommerce/coupons/utils.py
```diff
@@ -5,34 +5,92 @@
...
 from oscar.core.loading import get_model
 
+from ecommerce.courses.utils import traverse_pagination
+
+
 Product = get_model('catalogue', 'Product')
 
 
-def get_range_catalog_query_results(limit, query, site, offset=None):
+def get_catalog_course_runs(site, query, limit=None, offset=None):
```
zubair-arbi, line 14: @mjfreyn @marjev I have refactored the method `get_range_catalog_query_results` into `get_catalog_course_runs` to incorporate the needs for existing usage and for enterprise usage. Please review that if this is what you were suggesting.

ecommerce/enterprise/entitlements.py
zubair-arbi: @mjfrey @marjev @mattdrayer these method are kept separate since we wanted to isolate the enterprise logic to the existing flow. Please note that the enterprise service returns the coupon product id (`entitlement_id`), from that we fetch the coupon product then we check if the course in the user basket exists in the coupon voucher range and if that's the case we just return the first active voucher code which is automatically applied for the enterprise learners.

ecommerce/coupons/utils.py
marjev: LGTM

ecommerce/enterprise/entitlements.py
marjev: Ok, sounds reasonable enough.

marjev (approved):
LGTM

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,267 @@
...
+
+    """
+    voucher_offer = voucher.offers.first()
+    offer_range = voucher_offer.condition.range
+    if offer_range.course_catalog:
+        try:
+            course_catalog = get_course_catalogs(site=site, resource_id=offer_range.course_catalog)
+        except (ConnectionError, SlumberBaseException, Timeout):
+            logger.error('Unable to connect to Course Catalog service for course catalogs.')
+            return []
```
mjfrey: Why return an empty list?  We should return None.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,267 @@
...
+
+        try:
+            course_runs = get_catalog_course_runs(site, course_catalog.get('query'))['results']
+        except (ConnectionError, SlumberBaseException, Timeout, KeyError):
+            logger.error('Unable to get course runs from Course Catalog service.')
+            return []
```
mjfrey: Same here

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,267 @@
...
+
+        voucher_course_ids = [course_run.get('key') for course_run in course_runs if course_run.get('key')]
+    elif offer_range.catalog_query:
+        try:
+            course_runs = get_catalog_course_runs(site, offer_range.catalog_query)['results']
+        except (ConnectionError, SlumberBaseException, Timeout, KeyError):
+            logger.error('Unable to get course runs from Course Catalog service.')
+            return []
```
mjfrey: and here.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,267 @@
...
+    entitlements = get_entitlements_for_learner(site, user)
+    for entitlement in entitlements:
+        try:
+            coupon_product = Product.objects.filter(product_class__name='Coupon').get(id=entitlement['entitlement_id'])
+        except Product.DoesNotExist:
+            logger.exception(
+                'There was an error getting coupon product with the entitlement id %s',
+                entitlement['entitlement_id']
+            )
+            return []
```
mjfrey: We should return None if no Vouchers where found.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,267 @@
...
+        request (HttpRequest): request with voucher data
+        product (Product): A product that has course_key as attribute (seat or
+            bulk enrollment coupon)
+
+    """
+    if not is_enterprise_feature_enabled():
+        return None
+
+    vouchers = get_vouchers_for_learner(request.site, request.user)
+    entitlement_voucher = get_available_voucher_for_product(request, product, vouchers)
```
mjfrey: you should check if you have any vouchers before calling this.  Why call this if you have no vouchers.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,267 @@
...
+
+    """
+    try:
+        enterprise_learner_data = get_enterprise_learner_data(site, user)['results']
+    except (ConnectionError, SlumberBaseException, Timeout, KeyError, TypeError):
+        logger.exception(
+            'Failed to retrieve enterprise info for the learner [%s]',
+            user.username
+        )
+        return []
```
mjfrey: None

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,267 @@
...
+
+    if not enterprise_learner_data:
+        logger.info('Learner with username [%s] in not affiliated with any enterprise', user.username)
+        return []
```
mjfrey: None

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,267 @@
...
+
+    try:
+        entitlements = enterprise_learner_data[0]['enterprise_customer']['entitlements']
+    except KeyError:
+        logger.error('Invalid structure for enterprise learner API response for the learner [%s]', user.username)
+        return []
```
mjfrey: None

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,267 @@
...
+    learner.
+
+    Arguments:
+        request (HttpRequest): request with voucher data
+        product (Product): A product that has course_key as attribute (seat or
+            bulk enrollment coupon)
+
+    """
+    if not is_enterprise_feature_enabled():
+        return None
```
mjfrey, line 43: you return None here but empty set below.  Always return None if nothing is found.

mjfrey (changes requested):

ecommerce/enterprise/entitlements.py
zubair-arbi: @mjfrey thanks for pointing this out.

zubair-arbi:
@mjfrey  I have addressed your feedback. Please take another look.

ecommerce/enterprise/entitlements.py
```diff
@@ -0,0 +1,269 @@
...
+            logger.exception(
+                'There was an error getting coupon product with the entitlement id %s',
+                entitlement['entitlement_id']
+            )
+            return None
+
+        entitlement_voucher = retrieve_voucher(coupon_product)
+        vouchers.append(entitlement_voucher)
+
+    return vouchers
```
mjfrey, line 84: What happens if there are no entitlements?  Do we return []?

zubair-arbi: @mjfrey yes.

mjfrey: it should be None right?

zubair-arbi: I think it will be redundant to check if vouchers list is empty then return None, since we are populating the vouchers list on runtime so it makes sense if the response is empty list.

zubair-arbi:
@mjfrey can you please give a final review?

mjfrey (approved):

saleem-latif:
👍 once all checks are passing

zubair-arbi:
Ignoring the VersionEye failure.
```
 punycode 	1.4.1 	2.1.0 	UNKNOWN 	0 
```

PR #1105, opened 2017-01-24
Change enrollment code redemption to the receipt page.

vkaracic:
Reopening this PR: https://github.com/edx/ecommerce/pull/963
Only change is the increase from three to five checks for order completion. https://github.com/edx/ecommerce/compare/vkaracic/SOL-2073?expand=1#diff-c797ef6707ae3dfb1ee8828592244a9eR170

SOL-2073


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1105&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/order/models.py
```diff
@@ -21,6 +21,9 @@ def contains_coupon(self):
         return any(line.product.get_product_class().name == 'Coupon' for line in
                    self.basket.all_lines())
 
+    class Meta(object):
+        get_latest_by = 'date_placed'
```
mjfrey: Why this change?

vkaracic: I can't remember, probably because of the test. I'll revert it back.

mjfrey (approved):
Squash commits.

PR #1110, opened 2017-01-26
Updated to edx-drf-extensions 1.2.2

clintonb:
This will ensure we have the most up-to-date version when we later modify the JWT_ISSUERS configuration.

ECOM-6569 and ECOM-6539


clintonb:
FYI @edx/ecommerce

rlucioni (approved):

PR #1113, opened 2017-01-27
[SOL-2197] Do not allow unsupported CCs from new basket page.

iivic:
@mjfrey @vkaracic @marjev https://openedx.atlassian.net/browse/SOL-2197
The following credit card types are supported: American express, discover, mastercard, visa.
I also included a card number check before the card image is displayed that means the image will not be displayed if the card number doesn't pass the validation.
For unsupported card types a message will appear
screen shot 2017-01-27 at 13 29 56

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -124,7 +124,7 @@ define([
 
             if (!CreditCardUtils.isValidCardNumber(cardNumber)) {
                 appendCardValidationErrorMsg(event, cardNumberField, 'Invalid card number');
-            } else if (typeof cardType === 'undefined') {
+            } else if (typeof cardType === 'undefined' || !isCardTypeSupported(cardType.name)) {
                 appendCardValidationErrorMsg(event, cardNumberField, 'Unsupported card type');
             } else if (cvnNumber.length !== cardType.cvnLength || !Number.isInteger(Number(cvnNumber))) {
                 appendCardValidationErrorMsg(event, cvnNumberField, 'Invalid CVN');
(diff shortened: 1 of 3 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1113&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev (approved):

vkaracic (approved):
I guess you could it do this way. But you could also just remove the unsupported credit card types from here: https://github.com/edx/ecommerce/blob/49e89aa3af5ff0c928d3a4c1c5ac23da52c0de53/ecommerce/static/js/utils/credit_card.js#L31
That way we don't have code that's not used anywhere.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -124,7 +124,7 @@ define([
 
             if (!CreditCardUtils.isValidCardNumber(cardNumber)) {
                 appendCardValidationErrorMsg(event, cardNumberField, 'Invalid card number');
-            } else if (typeof cardType === 'undefined') {
+            } else if (typeof cardType === 'undefined' || !isCardTypeSupported(cardType.name)) {
                 appendCardValidationErrorMsg(event, cardNumberField, 'Unsupported card type');
```
mjfrey, line 128: we might want a better message like:
"edX accepts Visa, MasterCard, American Express, and Discover credit cards"

mjfrey: @iivic just leave "Unsupported card type"

mjfrey (approved):

PR #1114, opened 2017-01-31
Basket checkbox for enrollment codes.

vkaracic:
out

Locally this works as expected for me, but I still would like to try it out on a WL site. @douglashall and/or @mattdrayer do you have any WL sandbox on which this can be tested out?

https://openedx.atlassian.net/browse/SOL-2176

### DIFF
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -9,16 +9,18 @@ from oscar.core.loading import get_model
 from oscar.test.factories import BasketFactory, ProductFactory, RangeFactory, VoucherFactory
 import pytz
 
-from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME, ENROLLMENT_CODE_SWITCH
+from ecommerce.core.constants import ENROLLMENT_CODE_SWITCH
 from ecommerce.core.tests import toggle_switch
 from ecommerce.courses.tests.factories import CourseFactory
-from ecommerce.extensions.basket.utils import prepare_basket, attribute_cookie_data
+from ecommerce.extensions.basket.utils import (
+    attribute_cookie_data, get_enrollment_code, get_seat_from_enrollment_code, prepare_basket
+)
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
-from ecommerce.tests.mixins import UserMixin
 from ecommerce.extensions.partner.models import StockRecord
 from ecommerce.extensions.test.factories import prepare_voucher
 from ecommerce.referrals.models import Referral
 from ecommerce.tests.factories import SiteConfigurationFactory
+from ecommerce.tests.mixins import UserMixin
 from ecommerce.tests.testcases import TestCase
 
 Benefit = get_model('offer', 'Benefit')
@@ -35,9 +37,9 @@ class BasketUtilsTests(CourseCatalogTestMixin, TestCase):
         self.request = RequestFactory()
         self.request.COOKIES = {}
         self.request.user = self.create_user()
-        site_configuration = SiteConfigurationFactory(partner__name='Tester')
-        site_configuration.utm_cookie_name = 'test.edx.utm'
-        self.request.site = site_configuration.site
+        self.site_configuration = SiteConfigurationFactory(partner__name='Tester')
+        self.site_configuration.utm_cookie_name = 'test.edx.utm'
+        self.request.site = self.site_configuration.site
 
     def test_prepare_basket_with_voucher(self):
         """ Verify a basket is returned and contains a voucher and the voucher is applied. """
@@ -61,10 +63,7 @@ class BasketUtilsTests(CourseCatalogTestMixin, TestCase):
 
     def test_prepare_basket_enrollment_with_voucher(self):
         """Verify the basket does not contain a voucher if enrollment code is added to it."""
-        course = CourseFactory()
-        toggle_switch(ENROLLMENT_CODE_SWITCH, True)
-        course.create_or_update_seat('verified', False, 10, self.partner, create_enrollment_code=True)
-        enrollment_code = Product.objects.get(product_class__name=ENROLLMENT_CODE_PRODUCT_CLASS_NAME)
+        __, __, enrollment_code = self.create_course_seat_and_enrollment_code()
         voucher, product = prepare_voucher()
 
         basket = prepare_basket(self.request, product, voucher)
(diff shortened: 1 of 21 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1114"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

vkaracic:
@douglashall @mattdrayer this is ready for review. Since it concerns enrollment codes I'd like you or someone from your team's to review this, please.

douglashall:
@mjfrey and team should be able to handle the code review for you.

Please also reach out to Janaky and Liz Cohen for final approval of the UX. Thanks!

lizcohen:
UX looks good 👍

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -282,6 +281,50 @@ def test_attribute_cookie_data_multiple_cookies(self):
         with self.assertRaises(Referral.DoesNotExist):
             Referral.objects.get(basket_id=basket.id)
 
+    def prepare_seat_with_enrollment_code(self):
```
marjev: nitpick: perhaps moving this between setup and test methods would be better.

Having all test classes have the following format would increase readability:
```
TestClass
    # override methods
    # utility methods
    # test methods
```

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -282,6 +281,50 @@ def test_attribute_cookie_data_multiple_cookies(self):
...
+        """Helper function for creating a seat with an enrollment code."""
```
marjev: Should this docstring also contain information about return values?

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -257,6 +264,21 @@ def create_and_apply_benefit_to_basket(self, basket, product, benefit_type, bene
         basket.vouchers.add(voucher)
         Applicator().apply(basket)
 
+    def prepare_course_seat_and_enrollment_code(self):
```
marjev: This looks really similar to a method also available in this PR. Perhaps it's time to consider using a mixin.

ecommerce/extensions/basket/utils.py
```diff
@@ -159,3 +161,32 @@ def _record_utm_basket_attribution(referral, request):
...
+    Returns:
+        Either the enrollment code product or None if:
+            * enrollment codes are globally disabled
+            * enrollment codes are disabled for the passed site configuration
+            * passed seat does not have an enrollment code
+    """
+    if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and siteconfiguration.enable_enrollment_codes:
+        course = Course.objects.get(id=seat.attr.course_key)
+        return course.enrollment_code_product
+    else:
```
marjev: you don't need `else`

ecommerce/extensions/basket/utils.py
```diff
@@ -159,3 +161,32 @@ def _record_utm_basket_attribution(referral, request):
...
+        siteconfiguration (SiteConfiguration): The configuration of the site from which the request came.
+        seat (Product): The seat for which the enrollment code is retreived.
+    Returns:
+        Either the enrollment code product or None if:
+            * enrollment codes are globally disabled
+            * enrollment codes are disabled for the passed site configuration
+            * passed seat does not have an enrollment code
+    """
+    if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and siteconfiguration.enable_enrollment_codes:
+        course = Course.objects.get(id=seat.attr.course_key)
```
marjev, line 179: what if `Course` object doesn't exist? Should we return None in that case?

ecommerce/extensions/basket/utils.py
```diff
@@ -159,3 +161,32 @@ def _record_utm_basket_attribution(referral, request):
...
+        return course.enrollment_code_product
+    else:
+        return None
+
+
+def get_seat_from_enrollment_code(enrollment_code):
+    """Retrieves the seat from the passed enrollment code."""
+    course = Course.objects.get(id=enrollment_code.attr.course_key)
```
marjev, line 186: same question as in line 189.

ecommerce/extensions/basket/views.py
```diff
@@ -309,6 +319,26 @@ def get_context_data(self, **kwargs):
...
         return context
 
+    def post(self, *args, **kwargs):
+        """POST request handler."""
+        basket = self.request.basket
+        if self._client_side_checkout_enabled() and basket.lines.count() == 1:
+            line = basket.lines.first()
+            add_enrollment_code = self.request.POST.get('add-enrollment-code') == 'checked'
+
+            if line.product.get_product_class().name == SEAT_PRODUCT_CLASS_NAME and add_enrollment_code:
```
marjev: We could DRY this `line.product.get_product_class().name == SEAT_PRODUCT_CLASS_NAME` by implementing a property in Product class named `is_seat_product`. Similar thing might be done on other places that do similar thing.

So, Product can have properties for each product type. That will shorten lines like this. This line, for example, would be shortened to `if line.product.is_seat_product and add_enrollment_code`.

ecommerce/extensions/basket/views.py
```diff
@@ -309,6 +319,26 @@ def get_context_data(self, **kwargs):
...
+                enrollment_code = get_enrollment_code(self.request.site.siteconfiguration, line.product)
+                basket.flush()
+                basket.add(enrollment_code)
+                return HttpResponseRedirect(reverse('basket:summary'))
+            elif line.product.get_product_class().name == ENROLLMENT_CODE_PRODUCT_CLASS_NAME and \
```
marjev: same comment as in line 329.

marjev (changes requested):

ecommerce/extensions/basket/tests/test_utils.py
vkaracic: I am for keeping placements uniform throughout the codebase, but I'm more for [keeping case and effect close together](https://testing.googleblog.com/2017/01/testing-on-toilet-keep-cause-and-effect.html).

ecommerce/extensions/basket/utils.py
vkaracic: Honestly if a seat does not have a course tied to it I'd much rather have the ``DoesNotExist`` exception raised here. Although I believe in that case it would never get to this point, an exception would be raised somewhere before since this would be a major database defect.

vkaracic:
@marjev comments addressed.

ecommerce/extensions/basket/views.py
```diff
@@ -211,9 +212,12 @@ def _process_basket_lines(self, lines):
 
             line_data.update({
                 'benefit_value': benefit_value,
-                'enrollment_code': line.product.get_product_class().name == ENROLLMENT_CODE_PRODUCT_CLASS_NAME,
+                'enrollment_code': product.is_enrollment_code_product,
+                'has_enrollment_code': True if get_enrollment_code(
+                    self.request.site.siteconfiguration, product
+                ) else False,
```
marjev: You don't need if-else. `get_enrollment_code(self.request.site.siteconfiguration, product)` is enough.

marjev (approved):
Other than one additional comment, LGTM!

iivic (approved):
Everything is alphabetized 👍

vkaracic:
Closing this because we don't want the checkbox anymore.

PR #1116, opened 2017-02-01
Display order details for seats that do not require id verification.

vkaracic:
The order details message should be displayed for all seats and enrollment codes.
This PR additionally contains a new test helper method for easier enrollment code creation.

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -179,9 +179,19 @@ class BasketSummaryView(BasketView):
             product_class_name = line.product.get_product_class().name
             if product_class_name == 'Seat':
                 line_data = self._get_course_data(line.product)
-                if (getattr(line.product.attr, 'id_verification_required', False) and
-                        line.product.attr.certificate_type != 'credit'):
+                certificate_type = line.product.attr.certificate_type
+
+                if getattr(line.product.attr, 'id_verification_required', False) and certificate_type != 'credit':
                     display_verification_message = True
+
+                if certificate_type == 'verified':
+                    order_details_msg = _(
+                        'You will be automatically enrolled in the verified track'
+                        ' of the course upon completing your order.'
+                    )
+                elif certificate_type == 'credit':
+                    order_details_msg = _('You will receive your credit upon completing your order.')
+                else:
                     order_details_msg = _(
                         'You will be automatically enrolled in the course upon completing your order.'
                     )
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -257,6 +257,24 @@ class BasketSummaryViewTests(CourseCatalogTestMixin, CourseCatalogMockMixin, Lms
         basket.vouchers.add(voucher)
         Applicator().apply(basket)
 
+    def prepare_course_seat_and_enrollment_code(self, seat_type='verified', id_verification=False):
+        """Helper function that creates a new course, enables enrollment codes and creates a new
+        seat and enrollment code for it.
+
+        Args:
+            seat_type (str): Seat/certification type.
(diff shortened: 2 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1116"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/basket/views.py
```diff
@@ -156,9 +156,9 @@ def _process_basket_lines(self, lines):
                 if (getattr(line.product.attr, 'id_verification_required', False) and
                         line.product.attr.certificate_type != 'credit'):
                     display_verification_message = True
-                    order_details_msg = _(
-                        'You will be automatically enrolled in the course upon completing your order.'
-                    )
+                order_details_msg = _(
+                    'You will be automatically enrolled in the course upon completing your order.'
```
douglashall: I'm not sure it is appropriate for this message to show for verified seat upgrades since the user would already be enrolled in the course at this time. Please check with Janaky and @mjfrey on this.

vkaracic: It is the text that was requested in: https://openedx.atlassian.net/browse/SOL-2195

douglashall: It most certainly is. I think you want to double-check with Janaky and Michael though because the ask in SOL-2195 doesn't make sense for verified seat upgrades.

jsubramanian: In our current flow, we don't actually let the learner know that they are already enrolled when they enter the payment flow, so I think this message is OK to have.

douglashall: @jsubramanian I disagree. The user could decide to upgrade to a verified seat after interacting with the course for some time.

jsubramanian: @douglashall, good point. This message works in the initial purchase flow but not in the subsequent upgrade flow. Let me think about this and provide an updated message later today.

douglashall: @jsubramanian @vkaracic I think we have the ability here to distinguish verified seats from professional using the product attributes, no? Should we just show the current message only for professional? We can add a different message for verified seats later.

jsubramanian: @douglashall , if we know the type of seat being purchased it would be great to different between PE/WL and VC purchases. Can we do that?

douglashall: I think we can. @vkaracic can confirm.

vkaracic: Each seat has an attribute describing it's type (professional, verified, credit, audit), although it has no information about it being a WL seat or edX.org one.

jsubramanian: @vkaracic right now all WL courses are PE so our PE message will work for WL courses. We can keep the current message that we have for PE courses. I'll get you a new message for VC purchases and credit. Audit will never see this page.

jsubramanian: @vkaracic and @douglashall  am proposing these messages:

For VC purchases - "You will be automatically enrolled in the verified track of the course upon completing your order"

For Credit Purchases - "You will receive your credit upon completing your order"

Ran this by Lauren as well for the credit piece.

vkaracic:
Message(s) updated! Here's how the order details display now for different seat types:

screenshot from 2017-02-03 16-37-25

---
screenshot from 2017-02-03 16-37-54

---
screenshot from 2017-02-03 16-40-57

---
screenshot from 2017-02-03 16-41-17

@douglashall @jsubramanian please review.

vkaracic:
@douglashall @jsubramanian please review this.

mjfrey (approved):

PR #1119, opened 2017-02-01
WL-964 Add course start and end dates to basket view context

douglashall:
**Summary**
A WL partner has made a request to display the course start/end dates on the checkout page. EdX does not wish to display the course start/end dates on the checkout page. This PR adds the course start/end dates to the basket view context so that themes which override the checkout page template can make use of them as they wish.

**Sandbox**
https://mitpe.sandbox.edx.org/courses/course-v1:MITProfessionalX+IOTx+2016_T1/about

This can be tested by clicking the "Enroll Now" button on the page above.

### DIFF
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -515,6 +515,27 @@ class BasketSummaryViewTests(CourseCatalogTestMixin, CourseCatalogMockMixin, Lms
         expected_url = '{path}?next={next}'.format(path=testserver_login_url, next=urllib.quote(self.path))
         self.assertRedirects(response, expected_url, target_status_code=302)
 
+    @ddt.data(
+        (None, None),
+        ('invalid-date', None),
+        ('2017-02-01T00:00:00', datetime.datetime(2017, 2, 1)),
+    )
+    @ddt.unpack
+    @mock_course_catalog_api_client
+    @override_settings(PAYMENT_PROCESSORS=['ecommerce.extensions.payment.tests.processors.DummyProcessor'])
+    def test_context_data_contains_course_dates(self, date_string, expected_result):
+        seat = self.create_seat(self.course)
+        self.create_basket_and_add_product(seat)
+        self.mock_dynamic_catalog_single_course_runs_api(self.course, {
+            'start': date_string,
+            'end': date_string
+        })
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 200)
+        for _, line_data in response.context['formset_lines_data']:
+            self.assertEqual(line_data['course_start'], expected_result)
+            self.assertEqual(line_data['course_end'], expected_result)
+
 
 class VoucherAddMessagesViewTests(TestCase):
     """ VoucherAddMessagesView view tests. """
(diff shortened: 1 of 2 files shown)


douglashall:
@edx/ecommerce Can someone please review this? Thanks.

ecommerce/extensions/basket/views.py
```diff
@@ -127,6 +145,8 @@ def _get_course_data(self, product):
             'course_key': course_key,
             'image_url': image_url,
             'product_description': short_description,
+            'course_start': course_start,
```
davec-edx, line 154: I know some of this is older code, but what are we returning if we just hit the exception above?  It looks to me like we could fail for a few reasons, unable to get from the catalog for example, but then log it and return a bad structure...

One a second note, if we do fail in date conversion, is it safe to return None?

douglashall: If we hit one of the exceptions above, we will return a dict with a bunch of `None` values. This is safe. Do you see a better way to handle the exception cases?

The data provided here to the upstream code is purely for informational purposes and has no effect on the integrity of the system.

davec-edx: "better" is a bit subjective.  I prefer to use exception handling and throw a custom error up through the stack to the api level.  No sense in continuing to process data that doesn't exist.  But I don't know the use case for this either.

douglashall: I definitely agree the approach you describe is the right way to go for many use cases. For this one though, I don't think we would want to display an error to the user when they were trying to pay for something just because we couldn't grab all the course details from the catalog service.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -229,6 +230,8 @@ def setUp(self):
         self.client.login(username=self.user.username, password=self.password)
         self.course = CourseFactory(name='BasketSummaryTest')
         site_configuration = self.site.siteconfiguration
+        self.request = RequestFactory().request()
```
clintonb: This is already setup for you in `TestCase`.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -515,6 +518,32 @@ def test_login_required_basket_summary(self):
...
+
+        course_data = view._get_course_data(product)  # pylint: disable=protected-access
+
+        self.assertIn('course_start', course_data)
+        self.assertIn('course_end', course_data)
+
+    @ddt.data(
+        (None, None),
+        ('invalid-date', None),
+        ('2017-02-01T00:00:00', dateutil.parser.parse('2017-02-01T00:00:00')),
```
clintonb: Nit: Use a `datetime` object instead of calling `dateutil`. This protects us in the event of a bug in `dateutil`.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -515,6 +518,32 @@ def test_login_required_basket_summary(self):
         expected_url = '{path}?next={next}'.format(path=testserver_login_url, next=urllib.quote(self.path))
         self.assertRedirects(response, expected_url, target_status_code=302)
 
+    def test_get_course_data_returns_course_dates(self):
```
clintonb: It would be ideal if these two test cases were combined, and you looked at the rendered context. This is more inline with the other tests in this class.

douglashall:
Thanks @clintonb and @davec-edx. I have addressed comments. Good to go?

clintonb (approved):

PR #1121, opened 2017-02-01
Send identify events to Segment before page events from ecommerce

mikedikan:
ECOM-6650

Calling the logging of the user to segment before calling the tracking function, so events are attributed to the correct user.  Also, small cleanup of settings file.

@edx/ecommerce for review

### DIFF
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -429,13 +429,6 @@ EXTRA_SCOPE = ['permissions']
 # END AUTHENTICATION
 
 
-# ANALYTICS
-# Specify a key to emit events to the corresponding Segment project. `None` disables tracking.
-# See: https://segment.com/docs/libraries/python/
-SEGMENT_KEY = None
-# END ANALYTICS
-
-
 # DJANGO REST FRAMEWORK
 REST_FRAMEWORK = {
     'DEFAULT_AUTHENTICATION_CLASSES': (
--- a/ecommerce/static/js/views/analytics_view.js
+++ b/ecommerce/static/js/views/analytics_view.js
@@ -42,6 +42,7 @@ define([
                     // kick off segment
                     this.initSegment(trackId);
                     this.logUser();
+                    this.recordPageData();
 
                     // now segment has been loaded, we can track events
                     this.listenTo(this.model, 'segment:track', this.track);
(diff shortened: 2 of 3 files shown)


ecommerce/static/js/views/analytics_view.js
```diff
@@ -65,20 +65,19 @@ define([
...
-             * Get data for initializing segment.
+             * Get data for initializing segment and records the initial page load.
              */
-            getSegmentPageData: function () {
+            recordPageData: function () {
+                var pageData = {};
+
                 if (this.options.courseModel.get('courseId')) {
-                    return this.buildCourseProperties();
+                    pageData= this.buildCourseProperties();
```
AlasdairSwan: Nitpick: missing space between ```pageData``` and ```=```.

AlasdairSwan (changes requested):
Just 1 nit.

ecommerce/settings/base.py
```diff
@@ -426,13 +426,6 @@
 # END AUTHENTICATION
 
 
-# ANALYTICS
-# Specify a key to emit events to the corresponding Segment project. `None` disables tracking.
-# See: https://segment.com/docs/libraries/python/
-SEGMENT_KEY = None
```
rlucioni, line 435: If this is unused, it should also be removed from [edx/configuration](https://github.com/edx/configuration/blob/master/playbooks/roles/ecommerce/defaults/main.yml#L70-L71); you may want to consult with Devops to see if any secure overrides need to be cleaned up.

Related: it's probably worth documenting how this is meant to be set up now.

rlucioni (approved):

ecommerce/settings/base.py
mikedikan: Good call, thanks @renzo for the suggestion

ecommerce/static/js/views/analytics_view.js
mikedikan: nit accepted

ecommerce/settings/base.py
mikedikan: @rlucioni https://openedx.atlassian.net/browse/DEVOPS-5416

mikedikan:
@AlasdairSwan please re-review

AlasdairSwan (approved):
:shipit:

PR #1122, opened 2017-02-02
ENT-115: Link EnterpriseCourseEnrollment to EnterpriseCustomer

asadiqbal08:
This PR is dependent on [ENT-136](https://github.com/edx/edx-enterprise/pull/36). 

@saleem-latif @zubair-arbi @mtyaka @mattdrayer fyi

### DIFF
--- a/ecommerce/enterprise/entitlements.py
+++ b/ecommerce/enterprise/entitlements.py
@@ -276,3 +276,30 @@ def get_course_ids_from_voucher(site, voucher):
         voucher_course_ids = [seat.course_id for seat in seats]
 
     return voucher_course_ids
+
+
+def link_course_enrollment_to_enterprise_customer(site, course_id, enterprise_customer_user_id):
+    """
+    Make a link of course enrollment with enterprise customer
+
+    Arguments:
+        site: (django.contrib.sites.Site) site instance
+        course_id: course id
+        user: (django.contrib.auth.User) django auth user
+
+    """
+    api_resource_name = 'enterprise-course-enrollment'
+    api = site.siteconfiguration.enterprise_api_client
+    endpoint = getattr(api, api_resource_name)
+    data = {
+        'course_id': course_id,
+        'enterprise_customer_user': enterprise_customer_user_id
+    }
+    try:
+        endpoint.post(data)
+    except (ConnectionError, SlumberBaseException, Timeout) as exc:
+        logger.exception(
+            'Linking of course "%s" with enterprise customer user is failed with message: %s',
+            course_id, exc.response.text
+        )
+        raise exc
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -25,6 +25,7 @@ from ecommerce.extensions.fulfillment.status import LINE
 from ecommerce.extensions.voucher.models import OrderLineVouchers
 from ecommerce.extensions.voucher.utils import create_vouchers
 from ecommerce.notifications.notifications import send_notification
+import ecommerce.enterprise.entitlements as enterprise_api
 
 Benefit = get_model('offer', 'Benefit')
 Product = get_model('catalogue', 'Product')
@@ -213,6 +214,14 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
 
                 if response.status_code == status.HTTP_200_OK:
                     line.set_status(LINE.COMPLETE)
+                    learner_data = enterprise_api.get_enterprise_learner_data(site=order.site, user=order.user)
+
+                    if learner_data['results']:
+                        enterprise_api.link_course_enrollment_to_enterprise_customer(
+                            site=order.site,
+                            course_id=course_key,
+                            enterprise_customer_user_id=learner_data['results'][0]['id']
+                        )
 
                     audit_log(
                         'line_fulfilled',
--- a/ecommerce/extensions/fulfillment/tests/test_modules.py
+++ b/ecommerce/extensions/fulfillment/tests/test_modules.py
@@ -11,6 +11,7 @@ from oscar.test import factories
 from oscar.test.newfactories import UserFactory, BasketFactory
 from requests.exceptions import ConnectionError, Timeout
 from testfixtures import LogCapture
+from edx_rest_api_client import exceptions
 
 from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME, ENROLLMENT_CODE_SWITCH
 from ecommerce.core.tests import toggle_switch
@@ -28,6 +29,8 @@ from ecommerce.extensions.fulfillment.tests.mixins import FulfillmentTestMixin
 from ecommerce.extensions.voucher.models import OrderLineVouchers
 from ecommerce.extensions.voucher.utils import create_vouchers
 from ecommerce.tests.testcases import TestCase
+from ecommerce.core.tests.decorators import mock_enterprise_api_client
+from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
 
 JSON = 'application/json'
 LOGGER_NAME = 'ecommerce.extensions.analytics.utils'
@@ -44,7 +47,8 @@ Voucher = get_model('voucher', 'Voucher')
 
 @ddt.ddt
 @override_settings(EDX_API_KEY='foo')
-class EnrollmentFulfillmentModuleTests(CourseCatalogTestMixin, FulfillmentTestMixin, TestCase):
+class EnrollmentFulfillmentModuleTests(EnterpriseServiceMockMixin, CourseCatalogTestMixin, FulfillmentTestMixin,
+                                       TestCase):
     """Test course seat fulfillment."""
 
     course_id = 'edX/DemoX/Demo_Course'
@@ -88,6 +92,10 @@ class EnrollmentFulfillmentModuleTests(CourseCatalogTestMixin, FulfillmentTestMi
 
     @httpretty.activate
     @mock.patch('ecommerce.extensions.fulfillment.modules.parse_tracking_context')
+    @mock.patch(
+        'ecommerce.extensions.fulfillment.modules.enterprise_api.get_enterprise_learner_data',
+        mock.Mock(return_value={'results': []})
+    )
     def test_enrollment_module_fulfill(self, parse_tracking_context):
         """Happy path test to ensure we can properly fulfill enrollments."""
         httpretty.register_uri(httpretty.POST, get_lms_enrollment_api_url(), status=200, body='{}', content_type=JSON)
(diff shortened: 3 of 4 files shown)


ecommerce/enterprise/entitlements.py
```diff
@@ -276,3 +276,24 @@ def get_course_ids_from_voucher(site, voucher):
...
+    """
+    Make a link of course enrollment with enterprise customer
+
+    Arguments:
+        site: (django.contrib.sites.Site) site instance
+        course_id: course id
+        user: (django.contrib.auth.User) django auth user
+
+    """
+    api_resource_name = 'enterprise-customer-entitlement'
```
mattdrayer: It seems odd that we are posting to the `enterprise-customer-entitlement` resource/endpoint in order to update the customer-enrollment relationship -- are we sure this is the correct path to follow?

asadiqbal08: @mattdrayer Yes, I have made some changes locally, that I will push today. It has been updated. Thank !

ecommerce/enterprise/entitlements.py
```diff
@@ -276,3 +277,30 @@ def get_course_ids_from_voucher(site, voucher):
...
+        'enterprise_customer_user': enterprise_customer_user_id
+    }
+    try:
+        endpoint.post(data)
+    except exceptions.HttpClientError as exc:
+            logger.exception(
+                'Linking of course "%s" with enterprise customer user is failed with message: %s',
+                course_id, exc.response.text
+            )
+            raise exc
```
asadiqbal08: Q: should we just log OR raise the exception at that point ?

ecommerce/enterprise/entitlements.py
```diff
@@ -276,3 +277,30 @@ def get_course_ids_from_voucher(site, voucher):
...
+    api_resource_name = 'enterprise-course-enrollment'
+    api = site.siteconfiguration.enterprise_api_client
+    endpoint = getattr(api, api_resource_name)
+    data = {
+        'course_id': course_id,
+        'enterprise_customer_user': enterprise_customer_user_id
+    }
+    try:
+        endpoint.post(data)
+    except exceptions.HttpClientError as exc:
```
saleem-latif: I think we also need to log other network exceptions like `SlumberBaseException` (that is the base class for all exceptions raised by slumber), `ConnectionError` and `Timeout` exceptions from requests.

ecommerce/enterprise/entitlements.py
```diff
@@ -276,3 +277,30 @@ def get_course_ids_from_voucher(site, voucher):
...
+        logger.exception(
+            'Linking of course "%s" with enterprise customer user is failed with message: %s',
+            course_id, exc.response.text
+        )
+        raise exc
```
saleem-latif, line 305: Do we want to interrupt enrollment flow or fail silently here ? @mattdrayer your thoughts on this.

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -335,6 +347,67 @@ def test_credit_enrollment_module_fulfill(self):
         }
         self.assertEqual(actual, expected)
 
+    @httpretty.activate
+    @mock_enterprise_api_client
```
saleem-latif, line 351: I think it would be good to add positive test case for the scenario when current user is not an enterprise learner. This would verify that our additions do not affect existing flow.

saleem-latif (commented):
@asadiqbal08 A couple of comments, otherwise PR looks good.

saleem-latif:
Seems like we also need to rebase

ecommerce/extensions/fulfillment/tests/test_modules.py
asadiqbal08: There is a test in which we do not have any enterprise customer that is passing successfully.

ecommerce/enterprise/entitlements.py
asadiqbal08: I think, HttpClientError is handling the `Timeout`. not ?

ecommerce/extensions/fulfillment/tests/test_modules.py
saleem-latif: I talking about test case where we have an enterprise customer and there are some enterprise learner associated with it. but the user trying to enroll in the course is not associated with any enterprise customer.

ecommerce/enterprise/entitlements.py
saleem-latif: No, `HttpClientError` is defined by `slumber` package while `ConnectionError` and `Timeout` are defined by `requests` package so they are mutually exclusive exceptions.

ecommerce/extensions/fulfillment/tests/test_modules.py
asadiqbal08: @saleem-latif actually, we are mocking data as you can see under `mock_enterprise_api_client`, 
the code --> `enterprise_api.get_enterprise_learner_data(site=order.site, user=order.user)` -- > will return the mock data rather then fetching actual data from the database. I think other existing tests are validating the enrollment scenario that you are saying above.

saleem-latif: If existing tests are verifying that than it is good 👍

ecommerce/enterprise/entitlements.py
mattdrayer: I think we need to interrupt -- if we fail silently then the learner will end up with an enrollment that would not be linked to the enterprise customer and would not be included in the invoice/report for enterprise enrollments.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -213,6 +214,14 @@ def fulfill_product(self, order, lines):
 
                 if response.status_code == status.HTTP_200_OK:
                     line.set_status(LINE.COMPLETE)
+                    learner_data = enterprise_api.get_enterprise_learner_data(site=order.site, user=order.user)
```
mattdrayer, line 217: Is it possible to combine these two `enterprise_api' operations into a single network call?

asadiqbal08: @mattdrayer one call is a `GET` and later it is a `POST` call. 
firstly, We are determining the user who is placing the order either belongs to enterprise customer If yes then we create a association of course enrollment with enterprise-customer-user.

mattdrayer: If we just call the `POST`, and this workflow simply performs the verification internally, then we can eliminate the `GET` call.  We are already calling the service for every transaction.  Here's an example, given 50% enterprise learners:

100 ecommerce transactions
100 GET requests for learner data
50 POST requests for enrollment linking
50 enterprise enrollment links created
150 total requests to enterprise service

OR

100 ecommerce transactions
100 POST requests to link enrollments (with internal verification of enterprise learner status)
50 enterprise enrollment links created
100 total requests to enterprise service

Unless I am missing something?

ecommerce/enterprise/entitlements.py
mattdrayer: @asadiqbal08 just a bump on this comment ^^

asadiqbal08: Yes. We should interrupt here.

ecommerce/extensions/fulfillment/modules.py
asadiqbal08: Ok. @bdero We should make the `POST` [api-PR#40](https://github.com/edx/edx-enterprise/pull/40/files) to handle @mattdrayer concern above.

bdero:
This PR heavily overlaps with stuff in #1127, but it's more general, so I'm going to rebase that PR onto this one. FYI @mattdrayer 

@asadiqbal08 Can you rebase this branch again when you get the chance?

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -213,6 +214,14 @@ def fulfill_product(self, order, lines):
...
+
+                    if learner_data['results']:
+                        enterprise_api.link_course_enrollment_to_enterprise_customer(
+                            site=order.site,
+                            course_id=course_key,
+                            enterprise_customer_user_id=learner_data['results'][0]['id']
```
mattdrayer, line 223: Note that here we are assuming that there is only one enterprise linked to the learner -- we should log a ticket to update this spot to support multiple enterprises in the not-too-distant-future

bdero:
@asadiqbal08 I went ahead and did a quick rebase on this to make rebasing https://github.com/edx/ecommerce/pull/1127 on top of it easier (and so the build can turn green now that the dependency license problem is solved)

asadiqbal08:
@bdero I can see the same logic in #1127. So should I close that PR ?
@mattdrayer ^^

bdero:
@asadiqbal08 I rebased #1127 on top of this PR, but it shouldn't be closed because it also implements the course-specific consent gate for coupons.

ecommerce/extensions/fulfillment/modules.py
bdero, line 223: Doesn't this need to create an EnterpriseCustomerUser if one doesn't already exist?

bdero: On second thought, maybe I can add this to the view in https://github.com/edx/edx-enterprise/pull/40/files?

bdero: Nevermind, in this context, and EnterpriseCustomerUser must already exist (I'm guessing this is how entitlements work). It's not how coupons work, so I'll just need to make sure the call to create an EnterpriseCustomerUser happens before this block executes when a coupon has an associated EnterpriseCustomer in #1127.

asadiqbal08: why we need to create a EnterpriseCustomerUser if it does not exist ? According to me, We need to create an association of course-enrollment to enterprise-customer if the user who is placing the order, is an enterprise-learner/user.

bdero: For coupons associated with an EnterpriseCustomer, the redeeming user may or may not already be an EnterpriseCustomerUser.

mattdrayer:
@mjfrey

ecommerce/extensions/fulfillment/modules.py
bdero: I'm going to change the EnterpriseCourseEnrollment POST to do the following in order to satisfy this:

- Input: "username", "course_id", "consent_granted"

-  Try to find the EnterpriseCustomerUser associated with User (since we're assuming a 1-1 relationship for MVP - if/when this is no longer an assumption, we need to also require the EnterpriseCustomer UUID)
    -  If it doesn't exist, throw an error
    -  If it does, use the associated EnterpriseCustomer to create the new EnterpriseCourseEnrollment

bdero:
Since all of this is going to be downstreamed to be done in the Enrollment view in the LMS, should this PR be closed? @asadiqbal08 @mattdrayer

mattdrayer:
@asadiqbal08 please make a final evaluation to determine if any pieces must be kept, but I think we should be safe to close this PR, yes.  We should keep the JIRA ticket open, however, to ensure the use case is properly implemented and verified.

asadiqbal08:
Closing this PR as changes have been moved into #1127

bdero:
FYI, I'll be opening an LMS PR soon, after I finish cleaning up the ENT-134/ENT-162 PRs @asadiqbal08 @mattdrayer 

The two PRs are:
* The https://github.com/edx/edx-enterprise/pull/40 PR is all set to go, and is the highest piece in the dependency chain.
* Also https://github.com/edx/ecommerce/pull/1127 is all set too, just needs more tests.

PR #1123, opened 2017-02-02
ENT-169: Update offer's Range model to use course catalog query

zubair-arbi:
…nterprise coupon

ENT-169

* Update the method `contains_product` of `Range` model to get the catalog query from the linked catalog in the course catalog service.
* Update the method `run_catalog_query` of `Range` model to use optional parameter `query` for custom catalog course run queries (in case of course catalog coupons)
* Add method `is_course_in_enterprise_courses` for verification that the basket product (course run) exists in the learner's related enterprise catalog.
* Remove the custom entitlement method `get_course_ids_from_voucher` for enterprise and use the updated method `contains_product` of `Range` model for checking the product in the enterprise entitlement/coupon course runs
* Fix bug [ENT-174](https://openedx.atlassian.net/browse/ENT-174): For enterprise entitlements, first verify that the basket product (course run) exists in the related enterprise catalog before checking the product in the related coupon course runs list
* Update auto entitlements to be able to use multiple coupon codes
* Add/Update unittests

## Testing Steps:
**Sandbox:** https://zubair-arbi.sandbox.edx.org/

As a staff user setup an enterprise with entitlements:
1. Add a course catalog in Discovery django admin: https://discovery-enterprise.sandbox.edx.org/admin/catalogs/catalog/
2. Add an enterprise using LMS django admin with course catalog created above: https://zubair-arbi.sandbox.edx.org/admin/enterprise/enterprisecustomer/
3. Use CAT tool in ECOM to add a catalog coupon (enrollment or discount) with with enterprise course catalog: https://ecommerce-zubair-arbi.sandbox.edx.org/coupons/
4. Add the newly created coupon id (get it from the coupon url, e.g. 65 in "https://ecommerce-zubair-arbi.sandbox.edx.org/coupons/65/") and add it in the section "Enterprise Customer Entitlements" in LMS django admin for enterprise.
5. Link some user/learner with the enterprise using LMS django admin: https://zubair-arbi.sandbox.edx.org/admin/enterprise/enterprisecustomeruser/add/

As an enterprise learner:
1. Try to enroll in the course available in the enterprise catalog and verify that the enterprise entitlement coupon voucher is automatically applied.
2. Remove the auto applied enterprise coupon.
3. Try to enroll in the course which is not available in the enterprise catalog and verify that the enterprise entitlement coupon voucher is not applied.

_**Note:** For testing you can use the enterprise learner "`zubair@edx.org`" and an existing course "https://zubair-arbi.sandbox.edx.org/courses/course-v1:edX+DemoX+Demo_Course/about" which are linked with the enterprise "`Arbisoft ENT`" which has an enrollment discount coupon "https://ecommerce-zubair-arbi.sandbox.edx.org/coupons/65" available._

### DIFF
--- a/ecommerce/enterprise/entitlements.py
+++ b/ecommerce/enterprise/entitlements.py
@@ -16,11 +16,10 @@ from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import SlumberBaseException
 
-from ecommerce.coupons.utils import get_catalog_course_runs
 from ecommerce.coupons.views import voucher_is_valid
 from ecommerce.courses.utils import get_course_catalogs
 from ecommerce.enterprise.utils import is_enterprise_feature_enabled
-from ecommerce.extensions.api.serializers import retrieve_voucher
+from ecommerce.extensions.api.serializers import retrieve_all_vouchers
 
 
 logger = logging.getLogger(__name__)
@@ -34,28 +33,29 @@ def get_entitlement_voucher(request, product):
     learner.
 
     Arguments:
-        request (HttpRequest): request with voucher data
-        product (Product): A product that has course_key as attribute (seat or
+        product (Product): A product that has course_id as attribute (seat or
             bulk enrollment coupon)
+        request (HttpRequest): request with voucher data
 
     """
     if not is_enterprise_feature_enabled():
         return None
 
-    vouchers = get_vouchers_for_learner(request.site, request.user)
-    if vouchers:
-        entitlement_voucher = get_available_voucher_for_product(request, product, vouchers)
-        return entitlement_voucher
+    vouchers = get_course_vouchers_for_learner(request.site, request.user, product.course_id)
+    if not vouchers:
+        return None
 
-    return None
+    entitlement_voucher = get_available_voucher_for_product(request, product, vouchers)
+    return entitlement_voucher
 
 
-def get_vouchers_for_learner(site, user):
+def get_course_vouchers_for_learner(site, user, course_id):
     """
     Get vouchers against the list of all enterprise entitlements for the
-    provided learner.
+    provided learner and course id.
 
     Arguments:
+        course_id (str): The course ID.
         site: (django.contrib.sites.Site) site instance
         user: (django.contrib.auth.User) django auth user
 
@@ -63,7 +63,7 @@ def get_vouchers_for_learner(site, user):
         list of Voucher class objects
 
     """
-    entitlements = get_entitlements_for_learner(site, user)
+    entitlements = get_course_entitlements_for_learner(site, user, course_id)
     if not entitlements:
         return None
 
@@ -78,18 +78,19 @@ def get_vouchers_for_learner(site, user):
             )
             return None
 
-        entitlement_voucher = retrieve_voucher(coupon_product)
-        vouchers.append(entitlement_voucher)
+        entitlement_voucher = retrieve_all_vouchers(coupon_product)
+        vouchers.extend(entitlement_voucher)
 
     return vouchers
 
 
-def get_entitlements_for_learner(site, user):
+def get_course_entitlements_for_learner(site, user, course_id):
     """
-    Get entitlements for the provided learner if the provided learner is
-    affiliated with an enterprise.
+    Get entitlements for the provided learner against the provided course id
+    if the provided learner is affiliated with an enterprise.
 
     Arguments:
+        course_id (str): The course ID.
         site: (django.contrib.sites.Site) site instance
         user: (django.contrib.auth.User) django auth user
 
@@ -108,12 +109,88 @@ def get_entitlements_for_learner(site, user):
         return None
 
     try:
+        enterprise_catalog_id = enterprise_learner_data[0]['enterprise_customer']['catalog']
         entitlements = enterprise_learner_data[0]['enterprise_customer']['enterprise_customer_entitlements']
     except KeyError:
-        logger.error('Invalid structure for enterprise learner API response for the learner [%s]', user.username)
+        logger.exception('Invalid structure for enterprise learner API response for the learner [%s]', user.username)
+        return None
+
+    # Before returning entitlements verify that the provided course exists in
+    # the enterprise course catalog
+    if not is_course_in_enterprise_catalog(site, course_id, enterprise_catalog_id):
         return None
 
     return entitlements
+
+
+def is_course_in_enterprise_catalog(site, course_id, enterprise_catalog_id):
+    """
+    Verify that the provided course id exists in the site base list of course
+    run keys from the provided enterprise course catalog.
+
+    Arguments:
+        course_id (str): The course ID.
+        site: (django.contrib.sites.Site) site instance
+        enterprise_catalog_id (Int): Course catalog id of enterprise
+
+    Returns:
+        Boolean
+
+    """
+    try:
+        enterprise_course_catalog = get_course_catalogs(site=site, resource_id=enterprise_catalog_id)
+    except (ConnectionError, SlumberBaseException, Timeout):
+        logger.exception('Unable to connect to Course Catalog service for course catalogs.')
+        return None
+
+    if is_course_in_catalog_query(site, course_id, enterprise_course_catalog.get('query')):
+        return True
+
+    return False
+
+
+def is_course_in_catalog_query(site, course_id, enterprise_catalog_query):
+    """
+    Find out if the provided course exists in list of courses against the
+    enterprise course catalog query.
+
+    Arguments:
+        site: (django.contrib.sites.Site) site instance
+        course_id (Int): Course catalog id of enterprise
+        enterprise_catalog_query (Str): Enterprise course catalog query
+
+    Returns:
+        Boolean
+
+    """
+    partner_code = site.siteconfiguration.partner.short_code
+    cache_key = hashlib.md5(
+        '{site_domain}_{partner_code}_catalog_query_contains_{course_id}_{query}'.format(
+            site_domain=site.domain,
+            partner_code=partner_code,
+            course_id=course_id,
+            query=enterprise_catalog_query
+        )
+    ).hexdigest()
+    response = cache.get(cache_key)
+    if not response:
+        try:
+            response = site.siteconfiguration.course_catalog_api_client.course_runs.contains.get(
+                query=enterprise_catalog_query,
+                course_run_ids=course_id,
+                partner=partner_code
+            )
+            cache.set(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
+        except (ConnectionError, SlumberBaseException, Timeout):
+            logger.exception('Unable to connect to Course Catalog service for course runs.')
+            return False
+
+    try:
+        is_course_in_course_runs = response['course_runs'][course_id]
+    except KeyError:
+        return False
+
+    return is_course_in_course_runs
 
 
 def get_enterprise_learner_data(site, user):
@@ -220,7 +297,7 @@ def get_available_voucher_for_product(request, product, vouchers):
     product.
 
     Arguments:
-        product (Product): A product that has course_key as attribute (seat or
+        product (Product): A product that has course_id as attribute (seat or
             bulk enrollment coupon)
         request (HttpRequest): request with voucher data
         vouchers: (List) List of voucher class objects for an enterprise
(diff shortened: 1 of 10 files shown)


ecommerce/enterprise/entitlements.py
```diff
@@ -108,6 +112,15 @@ def get_entitlements_for_learner(site, user):
         return None
 
     try:
+        enterprise_catalog_id = enterprise_learner_data[0]['enterprise_customer']['catalog']
+    except KeyError:
+        logger.error('Invalid structure for enterprise learner API response for the learner [%s]', user.username)
```
saleem-latif: @zubair-arbi This exception handling seems to be covering the case when API got updated and this code did not, in those scenarios we would like to let the error propagate instead of suppressing it. So, I think it would be better to raise the exception here after logging it using `logger.exception`

ecommerce/enterprise/entitlements.py
```diff
@@ -116,6 +129,40 @@ def get_entitlements_for_learner(site, user):
     return entitlements
 
 
+def is_product_in_enterprise_courses(site, product, enterprise_catalog_id):
```
saleem-latif: Since, we only want to check if given course is present in the given course catalog, I think it would be better to abstract away some details here. e.g. we can abstract away terms like `product` and `enterprise` from this method. If we could divide it into smaller functions into something like
```
catalog = fetch_catalog_courses(site, catalog_id)
def is_course_in_catalog(site, course_id, catalog)
```

ecommerce/enterprise/entitlements.py
```diff
@@ -116,6 +129,40 @@ def get_entitlements_for_learner(site, user):
...
+    """
+    try:
+        enterprise_course_catalog = get_course_catalogs(site=site, resource_id=enterprise_catalog_id)
+    except (ConnectionError, SlumberBaseException, Timeout):
+        logger.error('Unable to connect to Course Catalog service for course catalogs.')
+        return False
+
+    try:
+        enterprise_course_runs = get_catalog_course_runs(site, enterprise_course_catalog.get('query'))['results']
+    except (ConnectionError, SlumberBaseException, Timeout, KeyError):
```
saleem-latif: Again we would not want to suppress exceptions like `KeyError` that hide API changes.

ecommerce/extensions/offer/models.py
```diff
@@ -214,21 +217,30 @@ def clean(self):
         if self.course_seat_types:
             validate_credit_seat_type(self.course_seat_types)
 
-    def run_catalog_query(self, product):
+    def run_catalog_query(self, product, query=None):
```
saleem-latif, line 226: I think we can abstract `product` here also and directly pass in `course_id`

ecommerce/extensions/offer/models.py
```diff
@@ -240,7 +252,19 @@ def contains_product(self, product):
...
+            request = get_current_request()
+            try:
+                course_catalog = get_course_catalogs(site=request.site, resource_id=self.course_catalog)
+            except (ConnectionError, SlumberBaseException, Timeout):
+                raise Exception('Unable to connect to Course Catalog service for course catalogs.')
+
+            response = self.run_catalog_query(product, course_catalog.get('query'))
+            # Range can have a catalog query and 'regular' products in it,
+            # therefor an OR is used to check for both possibilities.
+            return ((response['course_runs'][product.course_id]) or
```
saleem-latif, line 273: I think `response['course_runs'][product.course_id]` would raise a `KeyError` instead of running False if `product.course_id ` is not present in response.

ecommerce/enterprise/entitlements.py
```diff
@@ -84,12 +86,14 @@ def get_vouchers_for_learner(site, user):
     return vouchers
 
 
-def get_entitlements_for_learner(site, user):
+def get_product_entitlements_for_learner(site, user, product):
```
saleem-latif: I think it would be better if we can change `product` to `course_id` as entitlement flow depends on course id and not product.

saleem-latif (changes requested):

ecommerce/extensions/offer/models.py
zubair-arbi: @saleem-latif this is the desired behavior and its consistent with our dynamic range `self.catalog_query and self.course_seat_types`. The response is expected to be like:
```javascript
{u'course_runs': {u'course-v1:test-org+course+TZAzXRxXaIdA': True}}
```

ecommerce/extensions/offer/models.py
zubair-arbi: @saleem-latif I think we should avoid changes the existing naming convention for coupons as we have to update/add the related tests for validation.

ecommerce/enterprise/entitlements.py
zubair-arbi: @saleem-latif since we are stubbing the existing basket checkout flow with the enterprise entitlements logic so it would make sense to log the exceptions instead of raising it to avoid basket page crash. I think the entitlements error handling should be contained in the enterprise app. Also we can create custom entitlements alerts on `Splunk` if we think this logging might get missed.

ecommerce/enterprise/entitlements.py
zubair-arbi: I will update the methods/tests for `course_id`

zubair-arbi:
@saleem-latif I have addressed your feedback.
@mattdrayer Please review.

zubair-arbi:
@iivic @mjfrey I would appreciate if one of you review the changes.

saleem-latif:
@zubair-arbi overall code looks good to me, can you create a sandbox with these changes and share testing details ?

zubair-arbi:
@saleem-latif here is the sandbox with current changes: https://zubair-arbi.sandbox.edx.org/

iivic (approved):
LGTM

ecommerce/enterprise/entitlements.py
```diff
@@ -34,36 +34,38 @@ def get_entitlement_voucher(request, product):
     learner.
 
     Arguments:
-        request (HttpRequest): request with voucher data
         product (Product): A product that has course_key as attribute (seat or
```
mattdrayer: `course_key` or `course_id`?  I see `course_id` being referenced below?

ecommerce/enterprise/entitlements.py
```diff
@@ -34,36 +34,38 @@ def get_entitlement_voucher(request, product):
...
 
     """
     if not is_enterprise_feature_enabled():
         return None
 
-    vouchers = get_vouchers_for_learner(request.site, request.user)
-    if vouchers:
-        entitlement_voucher = get_available_voucher_for_product(request, product, vouchers)
-        return entitlement_voucher
+    course_id = product.course_id
```
mattdrayer: Nit: I think you can avoid this particular initialization -- it appears to be referenced only once in this function, and we're not doing any sort of validation or guarding prior to its use.

ecommerce/enterprise/entitlements.py
```diff
@@ -34,36 +34,38 @@ def get_entitlement_voucher(request, product):
...
+    vouchers = get_course_vouchers_for_learner(request.site, request.user, course_id)
+    if not vouchers:
+        return None
 
-    return None
+    entitlement_voucher = get_available_voucher_for_product(request, product, vouchers)
+    return entitlement_voucher
```
mattdrayer, line 49: Is there any sort of protection or validation we need to consider for this workflow prior to the return?

ecommerce/enterprise/entitlements.py
```diff
@@ -229,50 +290,7 @@ def get_available_voucher_for_product(request, product, vouchers):
     for voucher in vouchers:
         is_valid_voucher, __ = voucher_is_valid(voucher, [product], request)
         if is_valid_voucher:
-            voucher_course_ids = get_course_ids_from_voucher(request.site, voucher)
-            if product.course_id in voucher_course_ids:
+            voucher_offer = voucher.offers.first()
```
mattdrayer, line 309: Will there always be offers available?

ecommerce/enterprise/entitlements.py
```diff
@@ -229,50 +290,7 @@ def get_available_voucher_for_product(request, product, vouchers):
...
+            offer_range = voucher_offer.condition.range
```
mattdrayer, line 310: Will there always be `condition` and/or `range` attributes?

ecommerce/enterprise/entitlements.py
```diff
@@ -229,50 +290,7 @@ def get_available_voucher_for_product(request, product, vouchers):
...
+            if offer_range.contains_product(product):
```
mattdrayer, line 311: is it okay for this function to terminate without a catch-all `return` at the end?

ecommerce/extensions/offer/models.py
```diff
@@ -214,21 +217,30 @@ def clean(self):
         if self.course_seat_types:
             validate_credit_seat_type(self.course_seat_types)
 
-    def run_catalog_query(self, product):
+    def run_catalog_query(self, product, query=None):
         """
         Retrieve the results from running the query contained in catalog_query field.
         """
+        if not query:
+            query = self.catalog_query
```
mattdrayer, line 231: Do we need to verify that a catalog_query attribute is available and/or has a value?

saleem-latif (approved):

ecommerce/enterprise/entitlements.py
zubair-arbi: @mattdrayer we are already verifying the vouchers in the method `get_course_vouchers_for_learner` and `get_available_voucher_for_product`. We only return the valid and active voucher or `None` in case no vocher is found. 

**These validations include:**
1. Enterprise feature is enabled
2. User is enterprise learner
3. Basket course is available for enterprise catalog
4. Enterprise has entitlements
5. Get active vouchers for the enterprise
6. Return only voucher which has the basket course

Please let me know if I have missed something.

ecommerce/enterprise/entitlements.py
zubair-arbi: Yes, according to my understanding there is always going to be an offer for a voucher. @mjfrey can confirm.

ecommerce/enterprise/entitlements.py
zubair-arbi: Yes, @mjfrey can you please confirm.

ecommerce/enterprise/entitlements.py
zubair-arbi: I am going to explicitly add a `return None` statement

ecommerce/extensions/offer/models.py
zubair-arbi: This should always have some value (existing behavior for dynamic coupons).

mattdrayer:
Thanks for addressing all of my concerns (thru changes and/or clarifications) -- 👍

ecommerce/enterprise/entitlements.py
```diff
@@ -108,14 +110,72 @@ def get_entitlements_for_learner(site, user):
...
+    Arguments:
+        site: (django.contrib.sites.Site) site instance
+        enterprise_catalog_id (Int): Course catalog id of enterprise
+
+    Returns:
+        List of course runs
+
+    """
+    try:
+        enterprise_course_catalog = get_course_catalogs(site=site, resource_id=enterprise_catalog_id)
```
mjfrey: Few things here:
1: I would add an endpoint to the catalog service to return course_runs for a given catalog_id. This way you avoid multiple network trips.

ecommerce/enterprise/entitlements.py
```diff
@@ -108,14 +110,72 @@ def get_entitlements_for_learner(site, user):
...
+    except (ConnectionError, SlumberBaseException, Timeout):
+        logger.exception('Unable to connect to Course Catalog service for course catalogs.')
+        return None
+
+    try:
+        enterprise_course_runs = get_catalog_course_runs(site, enterprise_course_catalog.get('query'))['results']
```
mjfrey: 2: The results of this call could be many pages of results which I can not see being handled.  
3: There should be caching.

mjfrey (changes requested):

ecommerce/enterprise/entitlements.py
zubair-arbi: @mjfrey @mattdrayer  I have added a new ticket [ENT-201](https://openedx.atlassian.net/browse/ENT-201) to implement this new end point.

ecommerce/enterprise/entitlements.py
zubair-arbi: @mjfrey we are using the method [get_catalog_course_runs](https://github.com/edx/ecommerce/blob/master/ecommerce/coupons/utils.py#L14) which provide all course runs if no limit is provided by traversing pagination recursively. Also this method caches the result too. Please let me know if this clarify things.

zubair-arbi:
@mjfrey Please let me know if you are okay with merging this PR.

mjfrey:
I would rather see the new endpoint in catalog service before this is merged.  Why merge if it is not what we want?

zubair-arbi:
@mjfrey @mattdrayer I have updated the entitlements code to use catalog course runs contain endpoint to verify if a course exists against the enterprise catalog query. So now we don't need to get all course runs recursively from catalog service.
Since this PR is required for @saleem-latif work with consent based entitlements so we can focus on ENT-201 after this PR is merged.

mattdrayer:
@zubair-arbi please update this branch and when green you can merge in order to unblock @saleem-latif with the expectation that we will continue the work to revise this implementation to reflect the changes discussed with @mjfrey represented by ENT-201.

zubair-arbi:
@mattdrayer @mjfrey @saleem-latif  I am merging this PR and will address suggested improvements as part of ENT-201.

PR #1124, opened 2017-02-02
[WL-962] Prevent CyberSource Notify View from creating an order when the Otto Receipt Page is turned on

marjev:
@mjfrey Here's the PR that adds the fix you proposed yesterday. Please take a look.

JIRA ticket: https://openedx.atlassian.net/browse/WL-962

### DIFF
--- a/ecommerce/extensions/payment/tests/views/test_cybersource.py
+++ b/ecommerce/extensions/payment/tests/views/test_cybersource.py
@@ -35,6 +35,21 @@ class CybersourceNotifyViewTests(CybersourceNotificationTestsMixin, TestCase):
     path = reverse('cybersource_notify')
     view = CybersourceNotifyView
 
+    def setUp(self):
+        super(CybersourceNotifyViewTests, self).setUp()
+        self.site.siteconfiguration.enable_otto_receipt_page = False
+        self.site.siteconfiguration.save()
+
+    def test_otto_receipt_page_enabled(self):
+        """
+        Verify that the Notify view returns HTTP response with 200 status
+        when the Otto hosted receipt page is enabled.
+        """
+        self.site.siteconfiguration.enable_otto_receipt_page = True
+        self.site.siteconfiguration.save()
+        response = self.client.post(self.path)
+        self.assertEqual(response.status_code, 200)
+
 
 @ddt.ddt
 class CybersourceSubmitViewTests(CybersourceMixin, TestCase):
--- a/ecommerce/extensions/payment/views/cybersource.py
+++ b/ecommerce/extensions/payment/views/cybersource.py
@@ -271,6 +271,11 @@ class CybersourceNotifyView(CybersourceNotificationMixin, View):
     def post(self, request):
         """Process a CyberSource merchant notification and place an order for paid products as appropriate."""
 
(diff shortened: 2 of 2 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1124&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -35,6 +35,25 @@ class CybersourceNotifyViewTests(CybersourceNotificationTestsMixin, TestCase):
...
+        when the Otto hosted receipt page is enabled.
+        """
+        self.site.siteconfiguration.enable_otto_receipt_page = True
+        self.site.siteconfiguration.save()
+        notification = self.generate_notification(
+            self.basket,
+            billing_address=self.billing_address,
+        )
+        response = self.client.post(self.path, notification)
+        self.assertEqual(response.status_code, 200)
```
vkaracic: Maybe I don't have enough context on this fix, but this test passes regardless if the client side basket page is enabled, or not.

marjev: Good catch! Comment addressed.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -271,6 +271,11 @@ class CybersourceNotifyView(CybersourceNotificationMixin, View):
     def post(self, request):
         """Process a CyberSource merchant notification and place an order for paid products as appropriate."""
 
+        # If the Otto receipt page is enabled, CyberSource Interstitial View
+        # should handle the CyberSource response
+        if request.site.siteconfiguration.enable_otto_receipt_page:
+            return HttpResponse()
```
vkaracic, line 277: How is the interstitial view called from here?

marjev: It's not. Both are called as they're both set as callbacks (until we remove Notification View entirely or partially, which is the plan, I guess).

We just want to say "Don't do anything, Interstitial View got this."

vkaracic (approved):

ecommerce/extensions/payment/views/cybersource.py
mjfrey: To be clear, the interstitial view is not used unless the new receipt page is turned on.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -271,6 +271,11 @@ class CybersourceNotifyView(CybersourceNotificationMixin, View):
     def post(self, request):
         """Process a CyberSource merchant notification and place an order for paid products as appropriate."""
 
+        # If the Otto receipt page is enabled, CyberSource Interstitial View
+        # should handle the CyberSource response
+        if request.site.siteconfiguration.enable_otto_receipt_page:
```
mjfrey: I think we also want to add a check to make sure client side checkout is enabled since they are connected.

mjfrey (approved):

PR #1126, opened 2017-02-03
Sending notification after refund approval

clintonb:
An email is now sent to learners after a credit is issued for a refund. This functionality can be controlled for each site, and is only applicable for refunds of a course seat.

ECOM-6975

### DIFF
--- a/ecommerce/extensions/checkout/utils.py
+++ b/ecommerce/extensions/checkout/utils.py
@@ -1,14 +1,13 @@
 import logging
 import urllib
 
-from babel.numbers import format_currency
+from babel.numbers import format_currency as default_format_currency
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.utils.translation import get_language, to_locale
 from edx_rest_api_client.client import EdxRestApiClient
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import SlumberHttpBaseException
-
 
 logger = logging.getLogger(__name__)
 
@@ -60,6 +59,18 @@ def get_receipt_page_url(site_configuration, order_number=None, override_url=Non
     )
 
 
+def format_currency(currency, amount, format=None, locale=None):  # pylint: disable=redefined-builtin
+    locale = locale or to_locale(get_language())
+    format = format or getattr(settings, 'OSCAR_CURRENCY_FORMAT', None)
+
+    return default_format_currency(
+        amount,
+        currency,
+        format=format,
+        locale=locale
+    )
+
+
 def add_currency(amount):
     """ Adds currency to the price amount.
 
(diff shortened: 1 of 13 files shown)


clintonb:
@edx/ecommerce please review

schenedx (approved):
👍

ecommerce/extensions/checkout/utils.py
```diff
@@ -69,9 +80,4 @@ def add_currency(amount):
     Returns:
         str: Formatted price with currency.
     """
-    return format_currency(
-        amount,
-        settings.OSCAR_DEFAULT_CURRENCY,
-        format=u'#,##0.00',
-        locale=to_locale(get_language())
-    )
+    return format_currency(settings.OSCAR_DEFAULT_CURRENCY, amount, u'#,##0.00')
```
rlucioni, line 83: I know this isn't new, but do you know what the hash marks are for?

ecommerce/extensions/refund/tests/test_models.py
```diff
@@ -326,6 +342,80 @@ def test_deny_wrong_state(self, status):
...
+        order = create_order(basket=basket, user=user)
+        order_url = get_receipt_page_url(site_configuration, order.number)
+
+        refund = Refund.create_with_lines(order, order.lines.all())
+
+        with LogCapture(REFUND_MODEL_LOGGER_NAME) as l:
+            refund._notify_purchaser()  # pylint: disable=protected-access
+
+        msg = 'Course refund notification scheduled for Refund [{}].'.format(refund.id)
+        l.check(
```
rlucioni, line 372: This is admittedly a bit philosophical, but I'd like to share regardless: tests like these where we assert on the exact value of the log message seem like [change detector tests](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html). This test will break if the message is change slightly, even if the original intent is retained. I tend to only assert that the logger was called. If it's critical that, say, some identifier appear in the logged message, I'd check for its presence alone.

ecommerce/tests/factories.py
```diff
@@ -28,6 +28,7 @@ class Meta(object):
     lms_url_root = factory.LazyAttribute(lambda obj: "http://lms.testserver.fake")
     site = factory.SubFactory(SiteFactory)
     partner = factory.SubFactory(PartnerFactory)
+    send_refund_notifications = False
```
rlucioni, line 31: Should this be added to the management command which creates new site configs?

rlucioni (approved):

ecommerce/extensions/checkout/utils.py
clintonb: Formats for comma-delimited versus decimal-delimited currencies.

ecommerce/extensions/refund/tests/test_models.py
clintonb: In this case the identifier for alerts is the majority of the message. Nevertheless, I agree that change detector tests are non-ideal.

ecommerce/tests/factories.py
clintonb: Done.

rlucioni (approved):

PR #1101, opened 2017-01-20
ENT-135: Add EnterpriseCustomer field to coupons

haikuginger:
**Description:**

This pull request adds the ability for an ecommerce administrator to create coupon codes that are linked to a particular Enterprise Customer object.

**JIRA:** [ENT-135](https://openedx.atlassian.net/browse/ENT-135)

**Dependencies:** Depends on [ENT-112](https://github.com/edx/edx-enterprise/pull/30); in addition, this cannot be tested effectively until changes are made to that PR to enable OAuth API integration.

**Merge deadline:** 30 January 2017

**Installation instructions:**

To do manual testing, you must have edx-enterprise from the above-linked branch installed in your LMS. In addition, if the changes there haven't been made yet, you must modify the `EnterpriseCustomerViewSet` in that PR to use the `OAuth2Authentication` class, as [here](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/course_structure_api/v0/views.py#L32) in `edx-platform`.

**Screenshot:**

(Notice the *Enterprise Customer* field in the right-side column)
Coupon screenshot

**Testing instructions:**

1. With the ecommerce stack and LMS running, go to http://localhost:8002/login; grant access to the ecommerce stack. You must be using a staff user or above.
2. Go to http://localhost:8002/coupons/new.
    - Note that when using a course selection mode that is not a single catalog selection, an Enterprise Customer dropdown is visible, and it shows a list of the Enterprise Customer objects installed in your LMS.
3. Create a new coupon using one of your Enterprise Customers; go to the Edit view and observe that the dropdown still shows the Enterprise Customer you chose.
    - Change the name of the EC in the LMS and refresh this page; observe that the name changes here as well.

**Reviewers:**
- [ ] @bradenmacdonald 
- [ ] edX TBD

**Author concerns:**

I discovered while implementing this that it does not appear to be possible to edit existing coupons; based on checking against master, I do not believe this regression was introduced in this code, nor did I want to spend the time to properly diagnose and fix the issue.

PR #1127, opened 2017-02-03
ENT-134: Add consent gate when redeeming coupons from enterprise customers

bdero:
This PR is based on [#1101](https://github.com/edx/ecommerce/pull/1101), which needs to be merged first. For now, view the commit directly to see the diff of this PR.

This PR implements consent gating for coupon use by redirecting the user to the LMS course-specific consent UI when an enterprise customer is associated with a coupon. However, the coupon is not consumed nor does the user have to be associated with an Enterprise Customer until after consent is given and the order has been fulfilled.

This is achieved by a [change in edx-enterprise](https://github.com/edx/edx-enterprise/pull/40), where an option "enrollment_deferred" is added to allow removing the assumption that a user has an EnterpriseCourseEnrollment. For validating consent, this change leverages the redirect URL passed via the "next" GET parameter to the enterprise consent view.

- Say the user visits a coupon redeem URL (e.g. `/coupons/redeem/?code=IYO3SSQVEMHY2R5S&sku=BB3ECA9`), the view checks for the presence of a `consent_token` in the GET parameters (`content_token` has three degrees of uniqueness; the current OAuth2 access token (as the hmac key), the voucher ID, and the Enterprise Customer UUID).
- Because there isn't a `consent_token` in the URL, the view constructs a redirect to the enterprise consent page in the LMS, but with `enrollment_deferred` enabled and a `next` parameter containing an absolute URL that looks something like this (notice it's the same URL as before, but with the added `content_token` parameter):
    ```
    http://localhost:8002/coupons/redeem/?code=IYO3SSQVEMHY2R5S&sku=BB3ECA9&consent_token=524684104aecccb220ee362e4fccb57ce72f03fb1359970b6556a9f7d808695c
    ```
- So after the user provides consent and presses "Submit", they're redirected to the URL provided in the `next` parameter, which provides a `consent_token`.
- Now that the `consent_token` has been given, the coupon redeem view validates that it's correct before processing the coupon and order.

The security lies in the fact that constructing the valid `consent_token` requires knowledge of the user's current OAuth2 access token with the LMS. So for a third party adversary, constructing a URL that directly enrolls a user without asking for consent isn't feasible.

**JIRA tickets**: ENT-134.

**Dependencies**: #1101 and https://github.com/edx/edx-enterprise/pull/40

**Testing instructions**:

Initial setup:
1. Setup a devstack.
1. Override the edx-enterprise version with the [correct branch](https://github.com/edx/edx-enterprise/pull/40):
   ```
   /edx/bin/pip.edxapp install -e git+https://github.com/open-craft/edx-enterprise.git@bdero/optional-consent-enrollment#egg=edx-enterprise==optional-consent-enrollment
   /edx/bin/manage.edxapp lms migrate --settings=devstack
   ```
1. Checkout this branch (open-craft:bdero/coupon-enrollment-consent) for ecommerce.
1. Connect [ecommerce to the LMS](https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/ecommerce/install_ecommerce.html#configure-edx-openid-connect-oidc).
1. (Optionally) setup the [course discovery](https://open-edx-course-catalog.readthedocs.io/en/latest/getting_started.html) service and [connect it to the LMS](https://open-edx-course-catalog.readthedocs.io/en/latest/oidc.html).
1. Create an Enterprise Customer in the LMS: http://localhost:8000/admin/enterprise/enterprisecustomer/add/
    1. Give it any name.
    1. Check the **Enable data sharing consent** box.
    1. For **Enforce data sharing consent**, select **At Enrollment**.
1. Add a new Course in Ecommerce: http://localhost:8002/courses/new/
    1. For the course ID, use "course-v1:edX+DemoX+Demo_Course".
    1. Name it "edX Demonstration Course".
    1. Choose "Professional Education" for the Course Type.
1. Create a new Coupon in Ecommerce: http://localhost:8002/coupons/new
    1. Give it any Coupon Name.
    1. Under Code Type choose "Enrollment Code".
    1. Give it a valid date range that includes today.
    1. For Client, type any name. E.g. "Some Client".
    1. Select the "Single course" radio button.
    1. For the Course ID, use "course-v1:edX+DemoX+Demo_Course"
    1. The only available Seat Type should be automatically populated: "Professional"
    1. For the Enterprise Customer, select the name of the customer created in step 6.

Testing the PR:
1. Navigate to the [coupon list](http://localhost:8002/coupons/) and click "Download Coupon Report" to retrieve a CSV containing the coupon URL.
    coupon report button screenshot
1. Log out of the LMS/ecommerce.
1. The coupon report CSV should contain an offer URL that looks like this:
    ```
    http://localhost:8002/coupons/offer/?code=EGGZYLXDKNZK6W7C
    ```
    Navigate to that URL.
1. The first screen is the offer landing page.
    offer landing page screenshot
    Click "Enroll Now".
1. Next, you should be redirected to the LMS to login.
    login page screenshot
    Login using a user that's not already enrolled in the Demo course.
1. Upon logging in, you should be redirected to the course-specific Data Sharing Consent screen. The correct Enterprise Customer name and Course name should show up in the dialog.
    data sharing consent screenshot
    **Leave the "I agree" checkbox unchecked** and click Submit.
1. An alert dialog should show up alerting you that if you don't agree, you won't be given access to the course. Click Ok and note that you're redirected to the dashboard.
1. Navigate back to the offer URL (from step 3). This time, **check the "I agree" box** and click Submit. You should be redirected to the dashboard and now be enrolled in the course.
1. Navigate to the offer URL (from step 3) again. This time, there should be an error saying "This coupon has already been used."
    coupon error screenshot
1. Verify that there is a new EnterpriseCustomerUser (se the [admin view](http://localhost:8000/admin/enterprise/enterprisecustomeruser/)) that links the EnterpriseCustomer and User.

**Reviewers**
- [ ] OpenCraft: @mtyaka 
- [ ] edX: @mattdrayer

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -283,6 +283,15 @@ class SiteConfiguration(models.Model):
         """
         return urljoin(self.lms_url_root, path)
 
+    def build_enterprise_service_url(self, path=''):
+        """
+        Returns path joined with the appropriate Enterprise service URL root for the current site.
+
+        Returns:
+            str
+        """
+        return urljoin(settings.ENTERPRISE_SERVICE_URL, path)
+
     @property
     def commerce_api_url(self):
         """ Returns the URL for the root of the Commerce API (hosted by LMS). """
@@ -311,7 +320,12 @@ class SiteConfiguration(models.Model):
     @property
     def enterprise_api_url(self):
         """ Returns the URL for the Enterprise service. """
-        return self.build_lms_url('/enterprise/api/v1')
+        return settings.ENTERPRISE_API_URL
+
+    @property
+    def enterprise_grant_data_sharing_url(self):
+        """ Returns the URL for the Enterprise data sharing permission view. """
+        return self.build_enterprise_service_url('grant_data_sharing_permissions')
 
     @property
     def access_token(self):
@@ -365,7 +379,7 @@ class SiteConfiguration(models.Model):
             EdxRestApiClient: The client to access the Enterprise service.
 
         """
-        return EdxRestApiClient(settings.ENTERPRISE_API_URL, jwt=self.access_token)
+        return EdxRestApiClient(self.enterprise_api_url, jwt=self.access_token)
 
     @cached_property
     def user_api_client(self):
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -19,6 +19,12 @@ from ecommerce.core.url_utils import get_ecommerce_url
 from ecommerce.core.views import StaffOnlyMixin
 from ecommerce.coupons.decorators import login_required_for_credit
 from ecommerce.extensions.api import exceptions
+from ecommerce.enterprise.exceptions import EnterpriseDoesNotExist
+from ecommerce.enterprise.utils import (
+    get_enterprise_course_consent_url,
+    get_enterprise_customer_data_sharing_consent_token,
+    get_enterprise_customer_from_voucher,
+)
 from ecommerce.extensions.basket.utils import prepare_basket
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
 from ecommerce.extensions.voucher.utils import get_voucher_and_products_from_code
@@ -26,6 +32,7 @@ from ecommerce.extensions.voucher.utils import get_voucher_and_products_from_cod
 Applicator = get_class('offer.utils', 'Applicator')
 Basket = get_model('basket', 'Basket')
 Benefit = get_model('offer', 'Benefit')
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
 logger = logging.getLogger(__name__)
 OrderLineVouchers = get_model('voucher', 'OrderLineVouchers')
 Order = get_model('order', 'Order')
@@ -149,7 +156,7 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
         if not voucher.offers.first().is_email_valid(request.user.email):
             return render(request, template_name, {'error': _('You are not eligible to use this coupon.')})
 
-        if not request.user.account_details(request)['is_active']:
+        if not request.user.account_details(request).get('is_active'):
             return render(
                 request,
                 template_name,
@@ -158,6 +165,47 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
 
         if request.user.is_user_already_enrolled(request, product):
             return render(request, template_name, {'error': _('You are already enrolled in the course.')})
+
+        try:
+            enterprise_customer = get_enterprise_customer_from_voucher(request.site, request.user.access_token, voucher)
+        except EnterpriseDoesNotExist as e:
+            # If an EnterpriseException is caught while pulling the EnterpriseCustomer, that means there's no
+            # corresponding EnterpriseCustomer in the Enterprise service (which should never happen).
+            logger.exception(e.message)
+            return render(
+                request,
+                template_name,
+                {'error': _('Couldn\'t find a matching Enterprise Customer for this coupon.')}
+            )
+
+        if enterprise_customer is not None and enterprise_customer.get('enable_data_sharing_consent'):
+            consent_token = get_enterprise_customer_data_sharing_consent_token(
+                request.user.access_token,
+                product.course.id,
+                enterprise_customer['id']
+            )
+            received_consent_token = request.GET.get('consent_token')
+            if received_consent_token:
+                # If the consent token is set, then the user is returning from the consent view. Render out an error
+                # if the computed token doesn't match the one received from the redirect URL.
+                if received_consent_token != consent_token:
+                    return render(
+                        request,
+                        template_name,
+                        {'error': _('Invalid data sharing consent token provided.')}
+                    )
+            else:
+                # The user hasn't been redirected to the interstitial consent view to collect consent, so
+                # redirect them now.
+                redirect_url = get_enterprise_course_consent_url(
+                    request.site,
+                    code,
+                    sku,
+                    consent_token,
+                    product.course.id,
+                    enterprise_customer['id']
+                )
+                return HttpResponseRedirect(redirect_url)
 
         basket = prepare_basket(request, product, voucher)
         if basket.total_excl_tax == 0:
--- a/ecommerce/enterprise/utils.py
+++ b/ecommerce/enterprise/utils.py
@@ -1,8 +1,21 @@
 """
 Helper methods for enterprise app.
 """
+import hashlib
+import hmac
+from urllib import urlencode
+
 from django.conf import settings
+from django.core.urlresolvers import reverse
+from edx_rest_api_client.client import EdxRestApiClient
+from oscar.core.loading import get_model
 import waffle
+from slumber.exceptions import HttpNotFoundError
+
+from ecommerce.courses.utils import traverse_pagination
+from ecommerce.enterprise.exceptions import EnterpriseDoesNotExist
+
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
 
 
 def is_enterprise_feature_enabled():
@@ -20,3 +33,134 @@ def is_enterprise_feature_enabled():
     """
     is_enterprise_enabled = waffle.switch_is_active(settings.ENABLE_ENTERPRISE_ON_RUNTIME_SWITCH)
     return is_enterprise_enabled
+
+
+def get_enterprise_customer(site, token, uuid):
+    """
+    Return a single enterprise customer
+    """
+    client = EdxRestApiClient(
+        site.siteconfiguration.enterprise_api_url,
+        oauth_access_token=token
+    )
+    path = ['enterprise-customer', str(uuid)]
+    client = reduce(getattr, path, client)
+
+    try:
+        response = client.get()
+    except HttpNotFoundError:
+        return None
+    return {
+        'name': response['name'],
+        'id': response['uuid'],
+        'enable_data_sharing_consent': response['enable_data_sharing_consent'],
+    }
+
+
+def get_enterprise_customers(site, token):
+    resource = 'enterprise-customer'
+    client = EdxRestApiClient(
+        site.siteconfiguration.enterprise_api_url,
+        oauth_access_token=token
+    )
+    endpoint = getattr(client, resource)
+    response = endpoint.get()
+    return [
+        {
+            'name': each['name'],
+            'id': each['uuid'],
+        }
+        for each in traverse_pagination(response, endpoint)
+    ]
+
+
+def get_or_create_enterprise_customer_user(site, enterprise_customer_uuid, username):
+    """
+    Create a new EnterpriseCustomerUser on the enterprise service if one doesn't already exist.
+    Return the EnterpriseCustomerUser data.
+    """
+    data = {
+        'enterprise_customer': str(enterprise_customer_uuid),
+        'username': username,
+    }
+    api_resource_name = 'enterprise-learner'
+    api = site.siteconfiguration.enterprise_api_client
+    endpoint = getattr(api, api_resource_name)
+
+    get_response = endpoint.get(**data)
+    if get_response.get('count') == 1:
+        result = get_response['results'][0]
+        return result
+
+    response = endpoint.post(data)
+    return response
+
+
+def get_enterprise_customer_from_voucher(site, token, voucher):
+    """
+    Given a Voucher, find the associated Enterprise Customer and retrieve data about
+    that customer from the Enterprise service. If there is no Enterprise Customer
+    associated with the Voucher, `None` is returned.
+    """
+    try:
+        offer = voucher.offers.get(benefit__range__enterprise_customer__isnull=False)
+    except ConditionalOffer.DoesNotExist:
+        # There's no Enterprise Customer associated with this voucher.
+        return None
+
+    # Get information about the enterprise customer from the Enterprise service.
+    enterprise_customer_uuid = offer.benefit.range.enterprise_customer
+    enterprise_customer = get_enterprise_customer(site, token, enterprise_customer_uuid)
+    if enterprise_customer is None:
+        raise EnterpriseDoesNotExist(
+            'Enterprise customer with UUID {uuid} does not exist in the Enterprise service.'.format(
+                uuid=enterprise_customer_uuid
+            )
+        )
+
+    return enterprise_customer
+
+
+def get_enterprise_course_consent_url(site, code, sku, consent_token, course_id, enterprise_customer_uuid):
+    """
+    Construct the URL that should be used for redirecting the user to the Enterprise service for
+    collecting consent. The URL contains a specially crafted "next" parameter that will result
+    in the user being redirected back to the coupon redemption view with the verified consent token.
+    """
+    callback_url = '{protocol}://{domain}{resource}?{params}'.format(
+        protocol=settings.PROTOCOL,
+        domain=site.domain,
+        resource=reverse('coupons:redeem'),
+        params=urlencode({
+            'code': code,
+            'sku': sku,
+            'consent_token': consent_token,
+        })
+    )
+    request_params = {
+        'course_id': course_id,
+        'enterprise_id': enterprise_customer_uuid,
+        'enrollment_deferred': True,
+        'next': callback_url,
+    }
+    redirect_url = '{base}?{params}'.format(
+        base=site.siteconfiguration.enterprise_grant_data_sharing_url,
+        params=urlencode(request_params)
+    )
+    return redirect_url
+
+
+def get_enterprise_customer_data_sharing_consent_token(access_token, course_id, enterprise_customer_uuid):
+    """
+    Generate a sha256 hmac token unique to an end-user Access Token, Course, and
+    Enterprise Customer combination.
+    """
+    consent_token_hmac = hmac.new(
+        str(access_token),
+        '{course_id}_{enterprise_uuid}'.format(
+            course_id=course_id,
+            enterprise_uuid=enterprise_customer_uuid,
+        ),
+        digestmod=hashlib.sha256,
+    )
+    return consent_token_hmac.hexdigest()
--- a/ecommerce/extensions/api/v2/tests/views/test_enterprise.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_enterprise.py
@@ -23,7 +23,7 @@ class TestEnterpriseCustomerView(TestCase):
         ]
     }
 
-    @mock.patch('ecommerce.extensions.api.v2.views.enterprise.EdxRestApiClient')
+    @mock.patch('ecommerce.enterprise.utils.EdxRestApiClient')
     def test_get_customers(self, mock_client):
         instance = mock_client.return_value
         setattr(
--- a/ecommerce/extensions/api/v2/views/enterprise.py
+++ b/ecommerce/extensions/api/v2/views/enterprise.py
@@ -1,9 +1,8 @@
-from edx_rest_api_client.client import EdxRestApiClient
 from rest_framework import generics
 from rest_framework.permissions import IsAuthenticated, IsAdminUser
 from rest_framework.response import Response
 
-from ecommerce.courses.utils import traverse_pagination
+from ecommerce.enterprise.utils import get_enterprise_customers
 
 
 class EnterpriseCustomerViewSet(generics.GenericAPIView):
@@ -13,20 +12,3 @@ class EnterpriseCustomerViewSet(generics.GenericAPIView):
     def get(self, request):
         site = request.site
         return Response(data={'results': get_enterprise_customers(site, token=request.user.access_token)})
-
-
-def get_enterprise_customers(site, token):
-    resource = 'enterprise-customer'
-    client = EdxRestApiClient(
-        site.siteconfiguration.enterprise_api_url,
-        oauth_access_token=token
-    )
-    endpoint = getattr(client, resource)
-    response = endpoint.get()
-    return [
-        {
-            'name': each['name'],
-            'id': each['uuid'],
-        }
-        for each in traverse_pagination(response, endpoint)
-    ]
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -19,6 +19,7 @@ from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME, SEAT_PR
 from ecommerce.core.url_utils import get_lms_enrollment_api_url
 from ecommerce.courses.models import Course
 from ecommerce.courses.utils import mode_for_seat
+from ecommerce.enterprise.utils import get_or_create_enterprise_customer_user
 from ecommerce.extensions.analytics.utils import audit_log, parse_tracking_context
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.fulfillment.status import LINE
@@ -209,6 +210,25 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
                     }
                 )
             try:
+                # Collect the EnterpriseCustomer UUID from the coupon, if any.
+                enterprise_customer_uuid = None
+                for discount in order.discounts.all():
+                    enterprise_customer_uuid = discount.voucher.benefit.range.enterprise_customer
+                    if enterprise_customer_uuid is not None:
+                        data['enterprise_course_consent'] = True
+                        break
+
+                # If an EnterpriseCustomer UUID is associated with the coupon, create an EnterpriseCustomerUser
+                # on the Enterprise service if one doesn't already exist.
+                if enterprise_customer_uuid is not None:
+                    get_or_create_enterprise_customer_user(
+                        order.site,
+                        enterprise_customer_uuid,
+                        order.user.username
+                    )
+
+                # Post to the Enrollment API. The LMS will take care of posting a new EnterpriseCourseEnrollment to
+                # the Enterprise service if the user+course has a corresponding EnterpriseCustomerUser.
                 response = self._post_to_enrollment_api(data, user=order.user)
 
                 if response.status_code == status.HTTP_200_OK:
--- a/ecommerce/settings/local.py
+++ b/ecommerce/settings/local.py
@@ -1,5 +1,6 @@
 """Development settings and globals."""
 from __future__ import absolute_import
+from urlparse import urljoin
 
 from ecommerce.settings.base import *
 
@@ -120,3 +121,5 @@ ENABLE_AUTO_AUTH = True
 # Lastly, see if the developer has any local overrides.
 if os.path.isfile(join(dirname(abspath(__file__)), 'private.py')):
     from .private import *  # pylint: disable=import-error
+
+ENTERPRISE_API_URL = urljoin(ENTERPRISE_SERVICE_URL, 'api/v1/')
--- a/ecommerce/settings/production.py
+++ b/ecommerce/settings/production.py
@@ -1,5 +1,6 @@
 """Production settings and globals."""
 from os import environ
+from urlparse import urljoin
 
 # Normally you should not import ANYTHING from Django directly
 # into your settings, but ImproperlyConfigured is an exception.
@@ -9,6 +10,9 @@ import yaml
 
 from ecommerce.settings.base import *
 
+
+# Protocol used for construcing absolute callback URLs
+PROTOCOL = 'https'
 
 # Enable offline compression of CSS/JS
 COMPRESS_ENABLED = True
@@ -79,3 +83,5 @@ for __, configs in PAYMENT_PROCESSOR_CONFIG.iteritems():
             'error_path': PAYMENT_PROCESSOR_ERROR_PATH,
         })
 # END PAYMENT PROCESSOR OVERRIDES
+
+ENTERPRISE_API_URL = urljoin(ENTERPRISE_SERVICE_URL, 'api/v1/')
--- a/ecommerce/settings/test.py
+++ b/ecommerce/settings/test.py
@@ -1,11 +1,13 @@
 from __future__ import absolute_import
 
 from path import Path
+from urlparse import urljoin
 
 from ecommerce.settings.base import *
 
 
 SITE_ID = 1
+PROTOCOL = 'http'
 
 # TEST SETTINGS
 INSTALLED_APPS += (
@@ -142,3 +144,5 @@ COMPREHENSIVE_THEME_DIRS = [
 ]
 
 DEFAULT_SITE_THEME = "test-theme"
+
+ENTERPRISE_API_URL = urljoin(ENTERPRISE_SERVICE_URL, 'api/v1/')
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -18,12 +18,18 @@ from ecommerce.core.url_utils import get_lms_url
 from ecommerce.coupons.tests.mixins import CouponMixin
 from ecommerce.coupons.views import voucher_is_valid
 from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
+from ecommerce.enterprise.utils import (
+    get_enterprise_course_consent_url,
+    get_enterprise_customer_data_sharing_consent_token,
+)
 from ecommerce.extensions.api import exceptions
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.extensions.test.factories import prepare_voucher
 from ecommerce.extensions.voucher.utils import get_voucher_and_products_from_code
 from ecommerce.tests.mixins import ApiMockMixin, LmsApiMockMixin
 from ecommerce.tests.testcases import TestCase
+
 
 Applicator = get_class('offer.utils', 'Applicator')
 Basket = get_model('basket', 'Basket')
@@ -38,6 +44,7 @@ VoucherApplication = get_model('voucher', 'VoucherApplication')
 
 CONTENT_TYPE = 'application/json'
 COUPON_CODE = 'COUPONTEST'
+ENTERPRISE_CUSTOMER = 'cf246b88-d5f6-4908-a522-fc307e0b0c59'
 
 
 class CouponAppViewTests(TestCase):
@@ -264,7 +271,8 @@ class CouponOfferViewTests(ApiMockMixin, CouponMixin, CourseCatalogTestMixin, Lm
         self.assertEqual(response.status_code, 200)
 
 
-class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin, TestCase):
+class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin, EnterpriseServiceMockMixin,
+                            TestCase):
     redeem_url = reverse('coupons:redeem')
 
     def setUp(self):
@@ -283,27 +291,40 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin
         self.catalog.stock_records.add(StockRecord.objects.get(product=self.seat))
         self.student_dashboard_url = get_lms_url(self.site.siteconfiguration.student_dashboard_url)
 
-    def redeem_url_with_params(self, code=COUPON_CODE):
+    def redeem_url_with_params(self, code=COUPON_CODE, consent_token=None):
         """ Constructs the coupon redemption URL with the proper string query parameters. """
-        return self.redeem_url + '?code={}&sku={}'.format(code, self.stock_record.partner_sku)
+        params = {
+            'code': code,
+            'sku': self.stock_record.partner_sku,
+        }
+        if consent_token is not None:
+            params['consent_token'] = consent_token
+        return '{base}?{params}'.format(base=self.redeem_url, params=urllib.urlencode(params))
 
-    def create_and_test_coupon_and_return_code(self, benefit_value=90, code=COUPON_CODE, email_domains=None):
+    def create_and_test_coupon_and_return_code(
+            self,
+            benefit_value=90,
+            code=COUPON_CODE,
+            email_domains=None,
+            enterprise_customer=None
+    ):
         """ Creates coupon and returns code. """
         coupon = self.create_coupon(
             benefit_value=benefit_value,
             catalog=self.catalog,
             code=code,
-            email_domains=email_domains
+            email_domains=email_domains,
+            enterprise_customer=enterprise_customer
         )
         coupon_code = coupon.attr.coupon_vouchers.vouchers.first().code
         self.assertEqual(Voucher.objects.filter(code=coupon_code).count(), 1)
         return coupon_code
 
-    def assert_redemption_page_redirects(self, expected_url, target=200, code=COUPON_CODE):
+    def assert_redemption_page_redirects(self, expected_url, target=200, code=COUPON_CODE, consent_token=None):
         """ Verify redirect from redeem page to expected page. """
         self.request.user = self.user
         self.mock_enrollment_api(self.request, self.user, self.course.id, is_active=False, mode=self.course_mode)
-        response = self.client.get(self.redeem_url_with_params(code=code))
+        response = self.client.get(self.redeem_url_with_params(code=code, consent_token=consent_token))
         self.assertRedirects(response, expected_url, status_code=302, target_status_code=target)
 
     def test_login_required(self):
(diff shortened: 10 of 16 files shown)


openedx-webhooks:
Thanks for the pull request, @bdero! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1127&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

bdero:
@mtyaka Agreed - I went ahead and added an alternative method to collect a single enterprise customer after verifying the endpoint is built with django rest framework.

ecommerce/core/models.py
```diff
@@ -296,6 +296,16 @@ def oauth2_provider_url(self):
...
 
     @property
+    def enterprise_api_url(self):
+        """ Returns the URL for the Enterprise service. """
+        return self.build_lms_url('/enterprise/api/v1')
+
+    @property
+    def enterprise_grant_data_sharing_url(self):
+        """ Returns the URL for the Enterprise data sharing permission view. """
+        return self.build_lms_url('/enterprise/grant_data_sharing_permissions')
```
mattdrayer: Looks like we might be letting a verb-oriented API sneak in here?

ecommerce/coupons/views.py
```diff
@@ -159,6 +165,69 @@ def get(self, request):
         if request.user.is_user_already_enrolled(request, product):
             return render(request, template_name, {'error': _('You are already enrolled in the course.')})
 
+        try:
+            offer = voucher.offers.get(benefit__range__enterprise_customer__isnull=False)
+        except ConditionalOffer.DoesNotExist:
+            # There's no Enterprise Customer associated with this voucher, so there's no need for consent gating.
+            pass
+        else:
+            # Get information about the enterprise customer from the LMS.
```
mattdrayer: Not really getting it from the LMS -- we're getting it from the Enterprise service, which at the moment is hosted in the LMS

ecommerce/coupons/views.py
```diff
@@ -159,6 +165,69 @@ def get(self, request):
...
+            enterprise_customer_uuid = offer.benefit.range.enterprise_customer
```
mattdrayer: It would be great if we could move this customer data retrieval logic into the `enterprise` module

ecommerce/coupons/views.py
```diff
@@ -159,6 +165,69 @@ def get(self, request):
...
+                request.user.access_token,
+                enterprise_customer_uuid
+            )
+            if not enterprise_customer:
+                raise Exception('Enterprise customer with UUID {uuid} does not exist in the LMS.'.format(
+                    uuid=enterprise_customer_uuid
+                ))
+
+            if enterprise_customer['enable_data_sharing_consent']:
+                consent_token_hmac = hmac.new(
```
mattdrayer: Along the lines of above, I think it would be good to move this consent logic out of the coupon view, if possible.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -249,6 +267,38 @@ def create_order_for_invoice(self, basket, coupon_id, client, invoice_data=None)
 
         return response_data
 
+    def update_range_data(self, request, vouchers):
+        """
+        Update the range data for a particular request.
+        """
+        range_data = self.create_update_data_dict(data=request.data, fields=Range.UPDATABLE_RANGE_FIELDS)
+
+        if range_data:
```
mattdrayer: Can we flip this `if` around to a guard clause?  ie `if not range_data: ...`

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -249,6 +267,38 @@ def create_order_for_invoice(self, basket, coupon_id, client, invoice_data=None)
...
+            else:
+                range_data['course_catalog'] = None
+
+            enterprise_customer_data = request.data.get('enterprise_customer')
+            if enterprise_customer_data:
+                range_data['enterprise_customer'] = enterprise_customer_data.get('id')
+            else:
+                range_data['enterprise_customer'] = None
+
+            Range.objects.filter(id=voucher_range.id).update(**range_data)
```
mattdrayer: Is there a `return` missing here?

mattdrayer: Why use `filter` here?  Can there be more than one `Range`?  What if no Range objects are retrieved?

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -0,0 +1,46 @@
+from edx_rest_api_client.client import EdxRestApiClient
+from rest_framework import generics
+from rest_framework.permissions import IsAuthenticated, IsAdminUser
+from rest_framework.response import Response
+
+from ecommerce.courses.utils import traverse_pagination
+
+
+class EnterpriseCustomerViewSet(generics.GenericAPIView):
+
```
mattdrayer: Need a docstring

ecommerce/extensions/api/v2/views/coupons.py
bdero: @mattdrayer This particular bit of code is not part of this PR, it's part of https://github.com/edx/ecommerce/pull/1101, which looks like it's already been reviewed twice. Maybe there's more context there?

bdero: Also, looking at the [base PR](https://github.com/edx/ecommerce/pull/1101), this new function contains little new code, it's mostly factoring out existing stuff as seen in [this diff](https://github.com/edx/ecommerce/pull/1127/files#diff-c9a451f0afeebe0481e4cee1c16b7c94L285).

ecommerce/extensions/api/v2/views/coupons.py
bdero: @mattdrayer Same comments as https://github.com/edx/ecommerce/pull/1127/files#r99751380!

ecommerce/core/models.py
bdero: @mattdrayer I'm not sure I have the complete context to address this. This URL is used when constructing the destination of a 301 sent to the client browser. Would it make more sense to construct this URL somewhere else?

ecommerce/coupons/views.py
bdero: Fixed, thanks.

ecommerce/coupons/views.py
bdero: Sounds good, I separated this out.

ecommerce/coupons/views.py
bdero: Will there be another place we reuse deferred enrollment consent stuff in ecommerce? I'm having trouble thinking of a reasonable abstraction for this that wouldn't add complexity since the consent_token and URL construction is so tightly coupled with the context of coupons and specifically this view.

bdero:
@mattdrayer I've responded to your comments and this is ready for another look. Note that none of the code in f91f023 is actually part of this PR, it's #1101.

ecommerce/coupons/views.py
```diff
@@ -159,6 +165,52 @@ def get(self, request):
         if request.user.is_user_already_enrolled(request, product):
             return render(request, template_name, {'error': _('You are already enrolled in the course.')})
 
+        enterprise_customer = get_enterprise_customer_from_voucher(request.site, request.user.access_token, voucher)
+        if enterprise_customer is not None and enterprise_customer['enable_data_sharing_consent']:
+            consent_token_hmac = hmac.new(
```
e-kolpakov: @bdero I'd suggest moving the `if .. else ..` branches into methods.

ecommerce/enterprise/enrollments.py
```diff
@@ -0,0 +1,119 @@
...
+def get_enterprise_customer(site, token, uuid):
+    """
+    Return a single enterprise customer
+    """
+    client = EdxRestApiClient(
+        site.siteconfiguration.enterprise_api_url,
+        oauth_access_token=token
+    )
+    for nav in ['enterprise-customer', str(uuid)]:
+        client = getattr(client, nav)
```
e-kolpakov: @bdero I''m not suggesting to actually do that here, but if you'd like to show off:

    parts = ['enterprise-customer', str(uuid)]
    reduce(operator.attrgetter, parts, client)

Yes, now I'm showing off :)

ecommerce/enterprise/enrollments.py
```diff
@@ -0,0 +1,119 @@
...
+    endpoint = create_enterprise_service_client(site, 'auth-user')
+
+    result = endpoint.get(username=username)
+    if result.get('count') == 1:
+        return result['results'][0]
+
+    return None
+
+
+def get_or_post_enterprise_customer_user(site, enterprise_customer_uuid, username):
```
e-kolpakov: @bdero nit: I don't feel like this is a good idea, or at least a good name.

Options to address mine "I have a baaaad feeling about this":

1. Totally ignore :)
2. Rename to something like `get_or_create_enterprise_customer_user`.
3. Better yet: split  into get and create and let caller do the logic (of creating if it does not exist).

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -0,0 +1,62 @@
...
+        self.assertJSONEqual(
+            result.content.decode('utf-8'),
+            {
+                'results': [
+                    {
+                        'name': 'Starfleet Academy',
+                        'id': '5113b17bf79f4b5081cf3be0009bc96f'
+                    },  # Note that the private information from the API has been stripped
+                    {
+                        'name': 'Millennium Falcon',
```
e-kolpakov: @bdero nit: How's entire academy get in one list with just one ship? Or do you claim that Millenium Falcon is an "organization" on it's own (i.e. travelling band)? :)

e-kolpakov: @bdero Oh, right it's not from this PR. Question to you then @haikuginger.

ecommerce/core/models.py
mattdrayer: I guess I'm trying to understand what resource we're working with here -- it's not clear from the URL what entity we are modifying on the server / retrieving for the user.

mattdrayer: also does this URL need an `/api/v1` URI stem?

ecommerce/coupons/views.py
```diff
@@ -159,6 +165,52 @@ def get(self, request):
         if request.user.is_user_already_enrolled(request, product):
             return render(request, template_name, {'error': _('You are already enrolled in the course.')})
 
+        enterprise_customer = get_enterprise_customer_from_voucher(request.site, request.user.access_token, voucher)
+        if enterprise_customer is not None and enterprise_customer['enable_data_sharing_consent']:
```
mattdrayer: Can we count on `enable_data_sharing_consent` to always be present?  ie, do we need to think about `KeyError`?

ecommerce/coupons/views.py
mattdrayer: I still feel like this would be useful as its own little utility function, but if we're not going to use it anywhere else then I will yield to the YAGNI priniciple 😄

ecommerce/coupons/views.py
```diff
@@ -159,6 +165,52 @@ def get(self, request):
...
+                '{course_id}_{enterprise_uuid}'.format(
+                    course_id=product.course.id,
+                    enterprise_uuid=enterprise_customer['uuid'],
+                ),
+                digestmod=hashlib.sha256,
+            )
+            consent_token = consent_token_hmac.hexdigest()
+            received_consent_token = request.GET.get('consent_token')
+            if received_consent_token:
+                # If the content token is set, then the user is returning from the consent view. Validate
```
mattdrayer: "content" -> "consent"

ecommerce/coupons/views.py
```diff
@@ -159,6 +165,52 @@ def get(self, request):
...
+                # that the computed token is the same as the one received from the redirect URL.
```
mattdrayer: semantics, but the comment logic is the inverse of the actual check

ecommerce/enterprise/enrollments.py
```diff
@@ -0,0 +1,119 @@
...
+        site.siteconfiguration.enterprise_api_url,
+        oauth_access_token=token
+    )
+    for nav in ['enterprise-customer', str(uuid)]:
+        client = getattr(client, nav)
+    response = client.get()
+    return response
+
+
+def get_auth_user(site, username):
```
mattdrayer: should we include "enterprise" in the name of this method?

ecommerce/enterprise/enrollments.py
```diff
@@ -0,0 +1,119 @@
...
+        'consent_granted': True,
+        'course_id': course_id
+    }
+    endpoint = create_enterprise_service_client(site, 'enterprise-course-enrollment')
+
+    response = endpoint.post(data)
+    return response
+
+
+def create_enterprise_service_client(site, resource):
```
mattdrayer: It would help with the context for this module if we moved this particular function to the top of the file, since many of the other functions refer to it.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -207,6 +195,36 @@ def clean_voucher_request_data(cls, request):
...
+                    validation_message = 'Course mode not supported'
+                    raise ValidationError(validation_message, code=status.HTTP_400_BAD_REQUEST)
+
+            stock_records_string = ' '.join(str(id) for id in stock_record_ids)
+            coupon_catalog, __ = get_or_create_catalog(
+                name='Catalog for stock records: {}'.format(stock_records_string),
+                partner=partner,
+                stock_record_ids=stock_record_ids
+            )
+        else:
```
mattdrayer: Can we convert this into a guard:

```
if not stock_record_ids:
    return None

[rest of code]
```

mattdrayer: ah, perhaps this is the same deal as below...

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -0,0 +1,35 @@
...
+    permission_classes = (IsAuthenticated, IsAdminUser,)
+
+    def get(self, request):
+        site = request.site
+        return Response(data={'results': get_enterprise_customers(site, token=request.user.access_token)})
+
+
+def get_enterprise_customers(site, token):
```
mattdrayer: I was going to suggest moving this to `ecommerce/enterprise/enrollments.py` but realized that this module is semantically scoped to enrollments.  It might be a good time to introduce an "enterprise_helpers.py" (or "enterprise/api.py" for all other parts of the ecommerce system to interact with the Enterprise service.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -14,11 +14,13 @@
 from rest_framework import status
 import requests
 from requests.exceptions import ConnectionError, Timeout
+from threadlocals.threadlocals import get_current_request
 
 from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME, SEAT_PRODUCT_CLASS_NAME
 from ecommerce.core.url_utils import get_lms_enrollment_api_url
 from ecommerce.courses.models import Course
 from ecommerce.courses.utils import mode_for_seat
+from ecommerce.enterprise.enrollments import post_enterprise_course_enrollment
```
mattdrayer: yeah, seeing this now I'm thinking that having an `api.py` module to hide away the implementation specifics for interacting with the Enterprise service would be a good thing

ecommerce/coupons/views.py
bdero: Good catch, this really should be doing a `.get` instead. I changed it, thanks!

ecommerce/coupons/views.py
bdero: Alrighty. Thanks for introducing "YAGNI" into my vocab. :)

ecommerce/coupons/views.py
bdero: Fixed, thanks!

ecommerce/coupons/views.py
bdero: Fixed.

ecommerce/enterprise/enrollments.py
bdero: Seems reasonable, added it in.

ecommerce/enterprise/enrollments.py
bdero: Yeah, I went ahead and moved it to the top.

ecommerce/extensions/fulfillment/modules.py
bdero: @mattdrayer I'm actually considering refactoring it into an `EnterpriseApiClient` as suggested by @e-kolpakov (similar to the [CourseCatalogApiClient](https://github.com/edx/edx-enterprise/pull/29/files#diff-c59e08ddfc189e27b159a7151e1a87a3R27) in edx-enterprise).

ecommerce/enterprise/enrollments.py
```diff
@@ -0,0 +1,119 @@
...
+        'user_id': user['id'],
+    }
+    endpoint = create_enterprise_service_client(site, 'enterprise-learner')
+
+    get_response = endpoint.get(**data)
+    if get_response.get('count') == 1:
+        result = get_response['results'][0]
+        return result['id']
+
+    response = endpoint.post(data)
```
asadiqbal08: Why is it required to create an enterprise customer user explicitly here ?

ecommerce/enterprise/enrollments.py
```diff
@@ -0,0 +1,119 @@
...
+    return response['id']
+
+
+def post_enterprise_course_enrollment(site, enterprise_customer_uuid, username, course_id):
+    """
+    Post an EnterpriseCourseEnrollment to the Enterprise service by first fetching the corresponding
+    EnterpriseCustomerUser.
+    """
+    enterprise_customer_user = get_or_post_enterprise_customer_user(site, enterprise_customer_uuid, username)
```
asadiqbal08: What happened if API raise an exception while creating a `enterprise_customer_user` ?

ecommerce/enterprise/enrollments.py
```diff
@@ -0,0 +1,119 @@
...
+
+    data = {
+        'enterprise_customer_user': enterprise_customer_user,
+        'consent_granted': True,
+        'course_id': course_id
+    }
+    endpoint = create_enterprise_service_client(site, 'enterprise-course-enrollment')
+
+    response = endpoint.post(data)
```
asadiqbal08: Here we are making two network calls line:84, 93. I think, @mattdrayer has some concern on that.

asadiqbal08: We should handle any possible exceptions here too. Or a parent block is handling it ??

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -209,10 +212,32 @@ def fulfill_product(self, order, lines):
...
                 )
             try:
+                enterprise_customer_uuid = None
+                for discount in order.discounts.all():
+                    enterprise_customer_uuid = discount.voucher.benefit.range.enterprise_customer
+                    if enterprise_customer_uuid is not None:
+                        break
+                if enterprise_customer_uuid is not None:
+                    request = get_current_request()
+                    post_enterprise_course_enrollment(
```
asadiqbal08: eager to know, What is the actual difference here and [LINE#236](https://github.com/edx/ecommerce/pull/1127/files#diff-c9f38bb98fd490cc1127bb1cbdd272a3R236) ? If the code is replicated, then should not we make the call when the user is successfully enrolled in the course ? referring [here#231](https://github.com/edx/ecommerce/pull/1127/files#diff-c9f38bb98fd490cc1127bb1cbdd272a3R231)

ecommerce/enterprise/enrollments.py
bdero: Users that are not already an EnterpriseCustomerUser may redeem coupons associated with an EnterpriseCustomer. Since the coupons need to result in EnterpriseCourseEnrollments being created, we need to make sure an EnterpriseCustomerUser exists for the user.

ecommerce/extensions/fulfillment/modules.py
bdero: I just pushed the changes I made since rebasing, which involves removal of the duplicate parts, so this is now just making a call to create an Enterprise Customer User if one doesn't exist.

ecommerce/enterprise/enrollments.py
bdero: Since this function does the same thing as another in the base PR, I removed it.

ecommerce/enterprise/enrollments.py
bdero: Since this function does the same thing as another in the base PR, I removed it.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -209,10 +211,30 @@ def fulfill_product(self, order, lines):
                     }
                 )
             try:
+                enterprise_customer_uuid = None
+                for discount in order.discounts.all():
+                    enterprise_customer_uuid = discount.voucher.benefit.range.enterprise_customer
+                    if enterprise_customer_uuid is not None:
+                        break
+                if enterprise_customer_uuid is not None:
+                    get_or_post_enterprise_customer_user(
```
asadiqbal08: I still need to understand the purpose of these same multiple network calls  to the same model(s) [HERE#220](https://github.com/edx/ecommerce/pull/1127/files#diff-c9f38bb98fd490cc1127bb1cbdd272a3R220) and [HERE#233](https://github.com/edx/ecommerce/pull/1127/files#diff-c9f38bb98fd490cc1127bb1cbdd272a3R233).

asadiqbal08: @mattdrayer

bdero: @asadiqbal08 They're not the same. `get_or_post_enterprise_customer_user` create's an EnterpriseCustomerUser, `enterprise_api.link_course_enrollment_to_enterprise_customer` creates an EnterpriseCourseEnrollment.

But there is one additional route for round trip saving that I just implemented here: If there's a coupon being redeemed with an associated EnterpriseCustomer, we have to create an EnterpriseCustomerUser, which returns the ID. So I refactored this logic to not fetch the EnterpriseCustomerUser again when creating the EnterpriseCourseEnrollment in the event that we already know the EnterpriseCustomerUser ID.

asadiqbal08: I know, these are not the same but I have the concern to fetch enterprise-customer-learner data multiple times but I can see you have updated the code and move it under a condition. Thanks for the clarification.

mattdrayer:
@mjfrey FYI

ecommerce/core/models.py
```diff
@@ -314,6 +314,11 @@ def enterprise_api_url(self):
         return self.build_lms_url('/enterprise/api/v1')
 
     @property
+    def enterprise_grant_data_sharing_url(self):
+        """ Returns the URL for the Enterprise data sharing permission view. """
+        return self.build_lms_url('/enterprise/grant_data_sharing_permissions')
```
mattdrayer: Instead of `build_lms_url` here we should instead introduce a `build_enterprise_service_url` method -- remember that we are secretly leeching off the LMS for hosting :)

ecommerce/extensions/fulfillment/modules.py
bdero: @asadiqbal08 Ah, gotcha. I'm glad I addressed the concern, your comment prompted me to notice, thanks! :)

ecommerce/enterprise/enrollments.py
```diff
@@ -0,0 +1,93 @@
...
+
+    return None
+
+
+def get_or_post_enterprise_customer_user(site, enterprise_customer_uuid, username):
+    """
+    Create a new EnterpriseCustomerUser on the enterprise service if one doesn't already exist.
+    Return the ID of the EnterpriseCustomerUser.
+    """
+    user = get_enterprise_auth_user(site, username)
```
mattdrayer: Seems like the enterprise service should be able to do a lookup of a user using the `username` during the POST to the `enterprise-learner` endpoint instead of having to call `auth-user` first just to get an identifier which we then turn right around and send back to the enterprise service...

ecommerce/enterprise/enrollments.py
```diff
@@ -0,0 +1,93 @@
+"""
```
mattdrayer: None of the functions here appear to be specific to "enrollments", so they should probably instead be moved to a `utils.py` module

ecommerce/enterprise/entitlements.py
```diff
@@ -276,3 +276,30 @@ def get_course_ids_from_voucher(site, voucher):
         voucher_course_ids = [seat.course_id for seat in seats]
 
     return voucher_course_ids
+
+
+def link_course_enrollment_to_enterprise_customer(site, course_id, enterprise_customer_user_id):
```
mattdrayer: the name of this function leads me to believe it should be moved to the `enrollments.py` module

mattdrayer: In addition -- and I feel like we may have already discussed so please forgive my amnesia -- the name of this function doesn't really fit with the arguments it's asking for.  That is, we're not giving it a course enrollment, and we're not giving it an enterprise customer.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -1,10 +1,13 @@
...
 from rest_framework import generics
 from rest_framework.permissions import IsAuthenticated, IsAdminUser
 from rest_framework.response import Response
 
 from ecommerce.courses.utils import traverse_pagination
 
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
+
 
 class EnterpriseCustomerViewSet(generics.GenericAPIView):
```
mattdrayer: The `get_enterprise_customers` workflow in this view should probably call some internal logic such vs. connecting directly to the enterprise service.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -214,6 +216,37 @@ def fulfill_product(self, order, lines):
...
+
+                    # If we don't already have the user's associated EnterpriseCustomerUser, try fetching it from the
+                    # Enterprise service.
+                    if enterprise_customer_user_id is None:
+                        learner_data = enterprise_api.get_enterprise_learner_data(site=order.site, user=order.user)
+                        if learner_data['results']:
+                            enterprise_customer_user_id = learner_data['results'][0]['id']
+
+                    if enterprise_customer_user_id is not None:
+                        enterprise_api.link_course_enrollment_to_enterprise_customer(
```
mattdrayer: So I still feel like we should be able to accomplish all of this simply by passing the course enrollment and the enterprise customer information to `link_course_enrollment_to_enterprise_customer`  -- there's really no need to do all the EnterpriseCustomerUser back-and-forth.

Let's take a further step back from that, even.  I discussed this particular part of the PR with @mjfrey today, and it actually looks like the wrong place to be handling this work.  Instead, the EnterpriseCourseEnrollment record should be added during he CourseEnrollment workflow in LMS, not here in Otto.  The LMS is perfectly capable of determining if this is an enterprise context, and if so hitting the enterprise service after the course enrollment record has been written -- Otto can remain ignorant of the entire thing.

ecommerce/enterprise/entitlements.py
asadiqbal08: is it make sense -- `link_course_with_enterprise_learner` ?

ecommerce/extensions/fulfillment/modules.py
bdero: Alright, after we discussed, I setup this PR for such a change by dropping the ENT-115 from the dependencies, leaving only the creation of the EnterpriseCustomerUser in the case of coupons if need be, and I will be opening a new PR to add the new EnterpriseCourseEnrollment post to the LMS enrollment view.

ecommerce/core/models.py
bdero: Ok, this required a setting change. I replaced ENTERPRISE_API_URL with ENTERPRISE_SERVICE_URL because the full API URL is too specific for this case: https://github.com/edx/ecommerce/pull/1127/commits/a7ccd054fff65edcb551c313b25c83cd1c4252a0

ecommerce/enterprise/enrollments.py
bdero: Done. I removed `get_enterprise_auth_user` and implemented this view functionality here: https://github.com/edx/edx-enterprise/pull/40/commits/e78e793cdb525a22572ffb1e2911a06148ed8faa

ecommerce/enterprise/entitlements.py
bdero: This was dropped from this PR since we're downstreaming this call to the LMS enrollment view.

ecommerce/enterprise/enrollments.py
bdero: Done.

ecommerce/extensions/api/v2/views/enterprise.py
bdero: Ok, I moved get_enterprise_customers to the enterprise module

ecommerce/coupons/views.py
```diff
@@ -159,6 +165,52 @@ def get(self, request):
         if request.user.is_user_already_enrolled(request, product):
             return render(request, template_name, {'error': _('You are already enrolled in the course.')})
 
+        enterprise_customer = get_enterprise_customer_from_voucher(request.site, request.user.access_token, voucher)
+        if enterprise_customer is not None and enterprise_customer.get('enable_data_sharing_consent'):
+            consent_token_hmac = hmac.new(
```
mattdrayer: ...still think we should move this consent token generator into a function like `get_enterprise_customer_data_sharing_consent_token`...

ecommerce/core/models.py
```diff
@@ -311,7 +320,12 @@ def oauth2_provider_url(self):
     @property
     def enterprise_api_url(self):
         """ Returns the URL for the Enterprise service. """
-        return self.build_lms_url('/enterprise/api/v1')
+        return settings.ENTERPRISE_API_URL
```
mattdrayer, line 323: I'm slightly confused with why we need both ENTERPRISE_API_URL and ENTERPRISE_SERVICE_URL -- can we just have SERVICE?

mattdrayer:
This PR looks great now -- @mjfrey will be happy 😉

ecommerce/core/models.py
mattdrayer: Discussed offline -- I get it now

ecommerce/core/models.py
```diff
@@ -283,6 +283,15 @@ def build_lms_url(self, path=''):
...
         return urljoin(self.lms_url_root, path)
 
+    def build_enterprise_service_url(self, path=''):
+        """
+        Returns path joined with the appropriate Enterprise service URL root for the current site.
+
+        Returns:
+            str
+        """
+        return urljoin(settings.ENTERPRISE_SERVICE_URL, path)
```
mtyaka, line 293: It looks like you have to be very careful with slash characters when using `urljoin` - the way this is currently set up is not working for me in devstack. This is how `urljoin` works with different combinations of slashes:

```python
urljoin('http://localhost:8000/enterprise/', '/grant_data_sharing_permission') 
# -> 'http://localhost:8000/grant_data_sharing_permissions'

urljoin('http://localhost:8000/enterprise', 'grant_data_sharing_permission') 
# -> 'http://localhost:8000/grant_data_sharing_permissions'

urljoin('http://localhost:8000/enterprise/', 'grant_data_sharing_permission') 
# -> 'http://localhost:8000/enterprise/grant_data_sharing_permissions'
```

The last combination is the only combination that works as you'd expect.

ecommerce/core/models.py
```diff
@@ -311,7 +320,12 @@ def oauth2_provider_url(self):
     @property
     def enterprise_api_url(self):
         """ Returns the URL for the Enterprise service. """
-        return self.build_lms_url('/enterprise/api/v1')
+        return settings.ENTERPRISE_API_URL
+
+    @property
+    def enterprise_grant_data_sharing_url(self):
+        """ Returns the URL for the Enterprise data sharing permission view. """
+        return self.build_enterprise_service_url('/grant_data_sharing_permissions')
```
mtyaka: Does not build correct URL for me - the leading slash character should be removed.

ecommerce/settings/local.py
```diff
@@ -124,3 +125,5 @@
 # Lastly, see if the developer has any local overrides.
 if os.path.isfile(join(dirname(abspath(__file__)), 'private.py')):
     from .private import *  # pylint: disable=import-error
+
+ENTERPRISE_API_URL = urljoin(ENTERPRISE_SERVICE_URL, '/api/v1/')
```
mtyaka: Does not work correctly because of the leading slash.

ecommerce/settings/production.py
```diff
@@ -79,3 +83,5 @@ def get_env_setting(setting):
             'error_path': PAYMENT_PROCESSOR_ERROR_PATH,
         })
 # END PAYMENT PROCESSOR OVERRIDES
+
+ENTERPRISE_API_URL = urljoin(ENTERPRISE_SERVICE_URL, '/api/v1/')
```
mtyaka: Does not work correctly because of the leading slash.

ecommerce/settings/test.py
```diff
@@ -142,3 +143,5 @@ def __getitem__(self, item):
 ]
 
 DEFAULT_SITE_THEME = "test-theme"
+
+ENTERPRISE_API_URL = urljoin(ENTERPRISE_SERVICE_URL, '/api/v1/')
```
mtyaka: Does not work correctly because of the leading slash.

ecommerce/enterprise/utils.py
```diff
@@ -20,3 +27,81 @@ def is_enterprise_feature_enabled():
...
+    api = site.siteconfiguration.enterprise_api_client
+    endpoint = getattr(api, api_resource_name)
+
+    get_response = endpoint.get(**data)
+    if get_response.get('count') == 1:
+        result = get_response['results'][0]
+        return result['id']
+
+    response = endpoint.post(data)
+    return response['id']
```
mtyaka: This is possibly due to some kind of misconfiguration on my part, but I can't figure it out :/
This fails for me on this line because the `'id'` attribute is not present in the response. I logged the response that I get and it looks like this:

```python
{u'username': u'mtyaka', u'enterprise_customer': u'46f5f645-af20-4f56-ad4c-6d8705fcf8e9'}
```

The `EnterpriseCustomerUser` does not get created. Any idea why?

mtyaka (changes requested):
@bdero I've run into some issues while testing this. See inline comments for more info.

ecommerce/enterprise/utils.py
bdero: Ah, sorry, you are right, this endpoint is no longer be returning an ID (we don't need the ID anymore). I fixed this and it should be all set now.

ecommerce/settings/test.py
bdero: Fixed, thanks

ecommerce/settings/production.py
bdero: Fixed, thanks

ecommerce/settings/local.py
bdero: Fixed, thanks

ecommerce/core/models.py
bdero: Fixed, thanks

ecommerce/core/models.py
bdero: Yeah, sorry - I had just made this change quickly without manually testing it, and this code path isn't covered yet in tests. I've fixed it. Thanks for figuring out that odd urljoin nuance. It looks urljoin might be distinguising between absolute vs relative URLs, although I've got no idea what's going on with that second example. ;)

ecommerce/coupons/views.py
bdero: Alright, I factored this out, and I also factored out the redirect URL construction code, which simplified test building for these code paths.

bdero:
@mtyaka This is ready for your continued review. :)

mtyaka:
:+1: It all works for me now!

- [x] I tested this: I followed the testing procedure in the PR with several variations.
- [x] I read through the code
- [x] I checked for accessibility issues
- [x] Includes documentation

mtyaka (approved):

bdero:
Thanks @mtyaka!

@mattdrayer Could you let me know if you have any more comments for this?

mattdrayer:
@bdero I think the PR looks really great now -- thanks so much for sticking with this one as it evolved -- you're a trooper 😄     👍 :shipit:

bdero:
@mattdrayer Great! Before merging this, I think we need to wait until https://github.com/edx/edx-platform/pull/14542 is merged too, since it will contain the edx-enterprise version bump to `0.23.0` (https://github.com/edx/edx-enterprise/pull/40).

ecommerce/coupons/views.py
```diff
@@ -159,6 +166,47 @@ def get(self, request):
...
+        try:
+            enterprise_customer = get_enterprise_customer_from_voucher(request.site, request.user.access_token, voucher)
+        except EnterpriseException as e:
+            # If an EnterpriseException is caught while pulling the EnterpriseCustomer, that means there's no
+            # corresponding EnterpriseCustomer in the Enterprise service (which should never happen).
+            logger.exception(e.message)
+            return render(
+                request,
+                template_name,
+                {'error': _('An unexpected error occured while redeeming this coupon.')}
```
e-kolpakov: @bdero Might make sense to be more specific here - if I understand right this message is shown to users, so they will have little to no understanding of what exactly went wrong. Dev team will be able to debug it (since log record is added), but it would require slightly less development involvement if user-facing message could be more specific, i.e. "Was unable to redeem the coupon due to {enteprise customer not found / disabled / we don't like it / we don't like you}".

ecommerce/coupons/views.py
```diff
@@ -159,6 +166,47 @@ def get(self, request):
...
+            )
+
+        if enterprise_customer is not None and enterprise_customer.get('enable_data_sharing_consent'):
```
e-kolpakov: @bdero nit: IDE should have complained that you're referencing potentially uninitialized `enterpise_customer` here. Not a big deal, but I recall I once had some trouble refactoring such a flow - details are fuzzy though. Might be good to add something like `enterprise_customer = None` above the try.

ecommerce/enterprise/utils.py
```diff
@@ -20,3 +33,129 @@ def is_enterprise_feature_enabled():
...
+    for nav in ['enterprise-customer', str(uuid)]:
+        client = getattr(client, nav)
+    try:
+        response = client.get()
+    except HttpNotFoundError:
+        return None
+    return response
+
+
+def get_enterprise_customers(site, token):
```
e-kolpakov: @bdero may I suggest enforcing uniform Enterprise Customer structure across singular GET (`get_enterprise_customer`) and plural GET (`get_enterprise_customers`) requests? I mean, return value of `get_enterprise_customers` should contain a list of exactly same results as if you just called `get_enterprise_customer` multiple times.

Having different structure for plurals and singulars is a mine - one wouldn't be able to tell it exists without looking at the code or facing an error in runtime.

ecommerce/enterprise/utils.py
```diff
@@ -20,3 +33,129 @@ def is_enterprise_feature_enabled():
...
+    return [
+        {
+            'name': each['name'],
+            'id': each['uuid'],
+        }
+        for each in traverse_pagination(response, endpoint)
+    ]
+
+
+def get_or_post_enterprise_customer_user(site, enterprise_customer_uuid, username):
```
e-kolpakov: @bdero nit: `get_or_create_` ?

ecommerce/enterprise/utils.py
```diff
@@ -20,3 +33,129 @@ def is_enterprise_feature_enabled():
...
+    api_resource_name = 'enterprise-learner'
+    api = site.siteconfiguration.enterprise_api_client
+    endpoint = getattr(api, api_resource_name)
+
+    get_response = endpoint.get(**data)
+    if get_response.get('count') == 1:
+        result = get_response['results'][0]
+        return result
+
+    response = endpoint.post(data)
```
e-kolpakov, line 95: @bdero Might be good to have the same return value structure as with GET methods. Maybe even to an extent of doing a "get by id" request additionally.

ecommerce/enterprise/utils.py
```diff
@@ -20,3 +33,129 @@ def is_enterprise_feature_enabled():
...
+        offer = voucher.offers.get(benefit__range__enterprise_customer__isnull=False)
+    except ConditionalOffer.DoesNotExist:
+        # There's no Enterprise Customer associated with this voucher.
+        return None
+
+    # Get information about the enterprise customer from the Enterprise service.
+    enterprise_customer_uuid = offer.benefit.range.enterprise_customer
+    enterprise_customer = get_enterprise_customer(site, token, enterprise_customer_uuid)
+    if enterprise_customer is None:
+        raise EnterpriseException(
```
e-kolpakov: @bdero might be a good idea to throw a differentiated exception type - i.e. dedicated exception type for this scenario, inherited from `EnterpriseException`

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -1,9 +1,11 @@
-from edx_rest_api_client.client import EdxRestApiClient
+from oscar.core.loading import get_model
 from rest_framework import generics
 from rest_framework.permissions import IsAuthenticated, IsAdminUser
 from rest_framework.response import Response
 
-from ecommerce.courses.utils import traverse_pagination
+from ecommerce.enterprise.utils import get_enterprise_customers
+
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
```
e-kolpakov: @bdero Seems to not be used in this file - please explain (is it here for side-effects?) or remove

e-kolpakov (commented):
I don't want it to become a blocker here, but I'd  strongly suggest to make get_enterprise_customer(s) methods to return data of identical structure (with ..._customers returning an array of course).

ecommerce/extensions/api/v2/views/enterprise.py
bdero: Yup, you're right, I removed this.

ecommerce/enterprise/utils.py
bdero: I actually made this exception for this specific scenario, and this is actually the only place it's being raised, so I feel like it might be better to leave it as-is for now rather than make child exceptions. :) It could be that I'm just missing the point, but what do you think?

ecommerce/enterprise/utils.py
bdero: I feel we shouldn't add new requests to this, since there are actually some operations compressed into the corresponding edx-enterprise view in order to minimize extra request overhead for e.g. large orders. (See previous addressed comments in this PR for details)

ecommerce/enterprise/utils.py
bdero: Changed it, thanks.

ecommerce/enterprise/utils.py
bdero: Alright, I made it so they return the same structure for an Enterprise Customer.

ecommerce/enterprise/enrollments.py
bdero: Alright, I renamed this to `get_or_create_enterprise_customer_user`.

ecommerce/enterprise/enrollments.py
bdero: Ah, true. I changed it to this:

```python
    path = ['enterprise-customer', str(uuid)]
    reduce(operator.attrgetter, path, client)
```

Thanks!

ecommerce/coupons/views.py
bdero: Since the except path returns, this is safe. I believe that's why pycharm nor pylint is complaining about it. :)

ecommerce/coupons/views.py
bdero: Ok, I changed it to be more specific, thanks!

```python
_('Couldn\'t find a matching Enterprise Customer for this coupon.')
```

ecommerce/coupons/views.py
bdero: This was refactored a bit and separated a bit more into functions since the diff that this comment is on.

bdero:
All comments addressed @e-kolpakov !

ecommerce/enterprise/enrollments.py
bdero: Had to use `getattr` since `operator.attrgetter` returns a callable rather than the attribute value itself.

ecommerce/enterprise/utils.py
haikuginger: I agree with @bdero. If we don't actually make use of the extra data right now, performing an extra request seems like a bad idea.

ecommerce/enterprise/utils.py
haikuginger: `EnterpriseException` does feel a bit overbroad as a naming scheme. I don't think it's necessary to create an entire hierarchy of exceptions based around it, but I might rename it to something like `EnterpriseDoesNotExist` (nit, so feel free to ignore).

bdero: Ok, I changed it to EnterpriseDoesNotExist

bdero:
After a little offline discussion, I will merge this now since it has already had two approvals, currently has no explicit disapprovals, this PR has gone through extensive internal and external review in general, and the blocking edx-platform PR is now merged! :)

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -209,6 +210,25 @@ def fulfill_product(self, order, lines):
                     }
                 )
             try:
+                # Collect the EnterpriseCustomer UUID from the coupon, if any.
+                enterprise_customer_uuid = None
+                for discount in order.discounts.all():
+                    enterprise_customer_uuid = discount.voucher.benefit.range.enterprise_customer
```
clintonb, line 216: The assumption that all discounts are linked to vouchers is wrong. This breaks fulfillment of orders that include a conditional offer not linked to a voucher (e.g. site-wide discount).

mattdrayer: @clintonb would it be an acceptable band-aid to add a check here for the existence of a linked voucher until the new benefit-based approach has been implemented?

clintonb: That's fine for now, sure. But, we need to fix this functionality soon. Fulfillment should not be concerned with vouchers. I am concerned about other pitfalls that may await us.

PR #1130, opened 2017-02-06
Get account details with site configuration access token

iivic:
@clintonb Did I understand your comment correctly https://openedx.atlassian.net/browse/SOL-2163?focusedCommentId=241447&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-241447
Instead of using the users access token we should use site configuration access token on every server-to-server API call?
If that's the case can you tell me what I'm missing because the change doesn't work. The 401 HttpClientError is still getting raised.

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -425,7 +425,7 @@ class User(AbstractUser):
         try:
             api = EdxRestApiClient(
                 request.site.siteconfiguration.build_lms_url('/api/user/v1'),
-                oauth_access_token=self.access_token,
+                jwt=request.site.siteconfiguration.access_token,
                 append_slash=False
             )
             response = api.accounts(self.username).get()
--- a/ecommerce/core/tests/test_models.py
+++ b/ecommerce/core/tests/test_models.py
@@ -98,6 +98,17 @@ class UserTests(CourseCatalogTestMixin, LmsApiMockMixin, TestCase):
         self.mock_account_api(self.request, user.username, data=user_details)
         self.assertDictEqual(user.account_details(self.request), user_details)
 
+    def test_user_details_uses_jwt(self):
+        """Verify user_details uses jwt from site configuration to call EdxRestApiClient."""
+        user = self.create_user()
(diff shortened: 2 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1130&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/core/models.py
```diff
@@ -425,7 +425,7 @@ def account_details(self, request):
         try:
             api = EdxRestApiClient(
                 request.site.siteconfiguration.build_lms_url('/api/user/v1'),
-                oauth_access_token=self.access_token,
+                oauth_access_token=request.site.siteconfiguration.access_token,
```
clintonb: `jwt`

https://github.com/edx/ecommerce/blob/master/ecommerce/core/models.py#L319

clintonb (changes requested):
This needs to be tested.

clintonb:
@iivic this change is exactly what I was referring to. If you're still seeing a 401 after fixing the call, you're going to need to share more information. Particularly, I need to know what your `JWT_AUTH` settings are for both services.

ecommerce/core/models.py
iivic: It works after changing the argument to jwt.

iivic:
@clintonb 
I was thinking about mocking the EdxRestApiClient response but then I realized it makes no sense because it will not cover any real issue. There is a test for account_details response [test](https://github.com/edx/ecommerce/blob/master/ecommerce/core/tests/test_models.py#L94) so I didn't add one.
I tested the fix locally and it works fine. Is this PR good to go?

iivic:
FYI @mjfrey

clintonb (approved):
It would be nice to update the test to verify a JWT is sent; but, that is not crucial.

iivic:
@clintonb I made a test to verify the jwt is sent. Take a quick look to make sure it's done properly.

clintonb (approved):

PR #1131, opened 2017-02-06
Setting Celery router for refund email task

clintonb:
ECOM-6975

### DIFF
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -499,8 +499,11 @@ CELERY_IMPORTS = (
     'ecommerce_worker.fulfillment.v1.tasks',
 )
 
-CELERY_ROUTES = {'ecommerce_worker.fulfillment.v1.tasks.fulfill_order': {'queue': 'fulfillment'},
-                 'ecommerce_worker.sailthru.v1.tasks.update_course_enrollment': {'queue': 'email_marketing'}}
+CELERY_ROUTES = {
+    'ecommerce_worker.fulfillment.v1.tasks.fulfill_order': {'queue': 'fulfillment'},
+    'ecommerce_worker.sailthru.v1.tasks.update_course_enrollment': {'queue': 'email_marketing'},
+    'ecommerce_worker.sailthru.v1.tasks.send_course_refund_email': {'queue': 'email_marketing'},
+}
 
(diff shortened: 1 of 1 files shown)


clintonb:
FYI @edx/ecommerce 

@edx/devops I suspect this may be why I'm not seeing the refund email task get picked up in stage. Based on http://docs.celeryproject.org/en/3.1/userguide/routing.html#automatic-routing, I have the following questions:

1. What is the value of `CELERY_CREATE_MISSING_QUEUES`?
2. What is the value of `CELERY_QUEUES`?

rlucioni:
@clintonb `CELERY_CREATE_MISSING_QUEUES` is useful if you don't want to explicitly specify `CELERY_QUEUES`. You'll only want to specify `CELERY_QUEUES` if you need to configure advanced routing beyond what the default `kombu.Queue` objects created by Celery give you.

clintonb:
@rlucioni agreed. Given that setting is supposed to be enabled by default, I'd like to work with DevOps to understand why it has been disabled.

rlucioni:
@clintonb where are you seeing that this has been disabled? I don't see any evidence of that in the base config or in the config repo defaults.

clintonb:
@rlucioni correction...we are not using a default queue and/or automatic routing. If we had been, I would not have needed to make this change.

rlucioni:
@clintonb we are using both the default queue and automatic routing. The ecomworker only consumes tasks from the `fulfillment` and `email_marketing` queues; see the end of [this line](https://github.com/edx/configuration/blob/master/playbooks/roles/ecomworker/templates/edx/app/ecomworker/ecomworker.sh.j2#L18). If you route something to the default queue, the worker isn't going to pick it up. Specifying `CELERY_ROUTES` is how you use automatic routing. For example, designating the `fulfillment` queue for `fulfill_order` tasks is all we need to do for Celery to create that queue for us.

clintonb:
@rlucioni I think the disconnect for me is with the word "automatic". If routing is automated, I don't expect to need to manually declare the task in settings.

rlucioni:
@clintonb I see. I've noticed nuances like these hidden throughout the Celery docs. Technically, if the ecomworker were processing the default `celery` queue, you wouldn't have had to declare the task anywhere. You had to make a change because we use named (but otherwise automatically created) queues.

PR #1133, opened 2017-02-07
ENT-136: Update coupon redemption page to account for the catalog id

saleem-latif:
@mattdrayer , @zubair-arbi , @asadiqbal08 --FYI

### DIFF
--- a/ecommerce/coupons/utils.py
+++ b/ecommerce/coupons/utils.py
@@ -1,14 +1,19 @@
 """ Coupon related utility functions. """
+import logging
 import hashlib
 
 from django.conf import settings
 from django.core.cache import cache
 from oscar.core.loading import get_model
+from slumber.exceptions import HttpNotFoundError
 
 from ecommerce.courses.utils import traverse_pagination
+from ecommerce.core.utils import get_cache_key
 
 
 Product = get_model('catalogue', 'Product')
+
+logger = logging.getLogger(__name__)
 
 
 def get_catalog_course_runs(site, query, limit=None, offset=None):
@@ -106,3 +111,57 @@ def prepare_course_seat_types(course_seat_types):
     if course_seat_types:
         return ','.join(seat_type.lower() for seat_type in course_seat_types)
     return None
+
+
+def fetch_course_catalog(site, catalog_id):
+    """
+    Fetch course catalog for the given catalog id.
+
+    This method will fetch catalog for given catalog id, if there is no catalog with the given
+    catalog id, method will return `None`.
+
+    Arguments:
+        site (Site): Instance of the current site.
+        catalog_id (int): An integer specifying the primary key value of catalog to fetch.
+
+    Example:
+        >>> fetch_course_catalog(site, catalog_id=1)
+        {
+            "id": 1,
+            "name": "All Courses",
+            "query": "*:*",
+            ...
+        }
+    Returns:
+        (dict): A dictionary containing key/value pairs corresponding to catalog attribute/values.
+
+    Raises:
+        ConnectionError: requests exception "ConnectionError", raised if if ecommerce is unable to connect
+            to enterprise api server.
+        SlumberBaseException: base slumber exception "SlumberBaseException", raised if API response contains
+            http error status like 4xx, 5xx etc.
+        Timeout: requests exception "Timeout", raised if enterprise API is taking too long for returning
+            a response. This exception is raised for both connection timeout and read timeout.
+
+    """
+    api_resource = 'catalogs'
+
+    cache_key = get_cache_key(
+        site_domain=site.domain,
+        resource=api_resource,
+        catalog_id=catalog_id,
+    )
+
+    response = cache.get(cache_key)
+    if not response:
+        api = site.siteconfiguration.course_catalog_api_client
+        endpoint = getattr(api, api_resource)
+
+        try:
+            response = endpoint(catalog_id).get()
+        except HttpNotFoundError:
+            logger.exception("Catalog '%s' not found.", catalog_id)
+            raise
+
+    cache.set(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
+    return response
--- a/ecommerce/extensions/api/v2/tests/views/test_vouchers.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_vouchers.py
@@ -13,6 +13,7 @@ from opaque_keys.edx.keys import CourseKey
 from oscar.core.loading import get_model
 from oscar.test.factories import BenefitFactory, OrderLineFactory, OrderFactory, RangeFactory
 from requests.exceptions import ConnectionError, Timeout
+from rest_framework import status
 from rest_framework.test import APIRequestFactory
 from slumber.exceptions import SlumberBaseException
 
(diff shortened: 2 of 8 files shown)


ecommerce/coupons/utils.py
```diff
@@ -106,3 +110,53 @@ def prepare_course_seat_types(course_seat_types):
     if course_seat_types:
         return ','.join(seat_type.lower() for seat_type in course_seat_types)
     return None
+
+
+def fetch_catalog(site, catalog_id):
+    """
+    Fetch course catalog for the given catalog id.
```
mattdrayer: I think we can just say "catalog" and not "course catalog" here

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -77,6 +77,9 @@ def offers(self, request):
         except (ConnectionError, SlumberBaseException, Timeout):
             logger.error('Could not get course information.')
```
mattdrayer: Does "course information" properly describe what we are retrieving here?

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -77,6 +77,9 @@ def offers(self, request):
...
             return Response(status=status.HTTP_400_BAD_REQUEST)
+        except HttpNotFoundError:
+            logger.error('Could not get catalog info from catalog-id for voucher code %s.', code)
```
mattdrayer: We are not dealing with "catalogs" here, per se, we are dealing with "offers"

mattdrayer:
Overall approach LGTM, @saleem-latif -- branch needs a rebase + tests but the implementation makes sense.  I see we are adding some network communications in this PR, so @mjfrey (or someone from his team) is probably going to want to take a look as well.

mattdrayer:
@haikuginger @bdero here is the change set for ENT-136 -- please let me know if this is blocking anything having to do with ENT-134 / ENT-162

bdero:
@mattdrayer AFAIK it's not, since the catalog is at no point used in the checks for whether an EnterpriseCustomerUser or EnterpriseCourseEnrollment will be created in https://github.com/edx/edx-enterprise/pull/40 nor https://github.com/edx/ecommerce/pull/1127.

ecommerce/coupons/utils.py
```diff
@@ -106,3 +110,53 @@ def prepare_course_seat_types(course_seat_types):
     if course_seat_types:
         return ','.join(seat_type.lower() for seat_type in course_seat_types)
     return None
+
+
+def fetch_catalog(site, catalog_id):
+    """
+    Fetch catalog for the given catalog id.
```
vkaracic: I don't think you are fetching a [``catalog``](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/catalogue/models.py#L19) here but a ``course catalog``. You should be mindful of the distinction since catalogs are still used for single-course coupons.

ecommerce/coupons/utils.py
```diff
@@ -106,3 +110,53 @@ def prepare_course_seat_types(course_seat_types):
...
+        catalog_id=catalog_id,
+    )
+
+    response = cache.get(cache_key)
+    if not response:
+        api = site.siteconfiguration.course_catalog_api_client
+        endpoint = getattr(api, api_resource)
+
+        try:
+            response = endpoint(catalog_id).get()
```
vkaracic, line 161: You should be able to do this with ``api.catalogs(catalog_id).get()``

ecommerce/coupons/utils.py
```diff
@@ -106,3 +110,53 @@ def prepare_course_seat_types(course_seat_types):
...
+            "query": "*:*",
+            ...
+        }
+    Returns:
+        (dict): A dictionary containing key/value pairs corresponding to catalog attribute/values.
+
+    Raises:
+        ConnectionError: requests exception "ConnectionError"
+        SlumberBaseException: slumber exception "SlumberBaseException"
+        Timeout: requests exception "Timeout"
```
vkaracic: This part should describe when and why an exception is raised.

ecommerce/utils.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+import hashlib
+
+
+def get_cache_key(**kwargs):
+    """
+    Get MD5 encoded cache key for given arguments.
+
+    Here is the format of key before MD5 encryption.
+        kay1:value1_key2:value2 ...
```
vkaracic: ``key1``

ecommerce/utils.py
```diff
@@ -0,0 +1,28 @@
...
+        # "site_domain:example.com_resource:catalogs"
+        a54349175618ff1659dee0978e3149ca
+
+    Arguments:
+        **kwargs: Key word arguments that need to be present in cache key.
+
+    Returns:
+         An MD5 encoded key uniquely identified by the key word arguments.
+    """
+    key = "_".join(["{}:{}".format(item, value) for item, value in kwargs.iteritems()])
```
vkaracic: Please import ``six`` and use ``six.iteritems(kwargs)``. This will help transition to Python 3 where ``iteritems`` was replaced by ``items``.

ecommerce/extensions/voucher/utils.py
```diff
@@ -635,7 +635,7 @@ def get_voucher_and_products_from_code(code):
     voucher_range = voucher.offers.first().benefit.range
     products = voucher_range.all_products()
```
vkaracic, line 636: Will this work for the course catalog case? Since it's not accounted for in [``all_products()``](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/offer/models.py#L299-L306).

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -75,10 +75,13 @@ def offers(self, request):
         try:
             offers_data = self.get_offers(request, voucher)
         except (ConnectionError, SlumberBaseException, Timeout):
-            logger.error('Could not get course information.')
+            logger.error('Could not connect to offers service.')
```
vkaracic: Technically it's ``Could not connect to course catalog service.``

ecommerce/coupons/utils.py
```diff
@@ -106,3 +110,53 @@ def prepare_course_seat_types(course_seat_types):
...
+
+def fetch_catalog(site, catalog_id):
+    """
+    Fetch catalog for the given catalog id.
+
+    This method will fetch catalog for given catalog id, if there is no catalog with the given
+    catalog id, method will return `None`.
+
+    Arguments:
+        site (django.contrib.sites.models.Site): Instance to the current site.
```
vkaracic: Just ``site (Site)`` will suffice. Also ``Instance of the current site``.

vkaracic (changes requested):

ecommerce/coupons/utils.py
saleem-latif: @vkaracic I did this to adhere to current coding pattern as is done [here](https://github.com/edx/ecommerce/blob/master/ecommerce/coupons/utils.py#L54) and [here](https://github.com/edx/ecommerce/blob/master/ecommerce/courses/utils.py#L54)

ecommerce/extensions/voucher/utils.py
saleem-latif: Good catch, I will update it and tests for this.

saleem-latif:
@vkaracic I have incorporated your feedback changes, however, I still need to add tests for the change set. Feel free to give another pass on the PR and share your thoughts before the tests are added.

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -74,10 +74,13 @@ def offers(self, request):
         try:
             offers_data = self.get_offers(request, voucher)
         except (ConnectionError, SlumberBaseException, Timeout):
-            logger.error('Could not get course information.')
+            logger.error('Could not connect to course catalog service.')
             return Response(status=status.HTTP_400_BAD_REQUEST)
+        except HttpNotFoundError:
```
vkaracic: Pylint reports ``E: 79,15: Bad except clauses order (SlumberBaseException is an ancestor class of HttpNotFoundError) (bad-except-order)`` so you don't need this.

saleem-latif:
@vkaracic kindly take a look now.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -474,3 +516,71 @@ def test_get_course_offer_verify_null_fields(self):
...
+
+        # Populate database for the test case.
+        course, seat = self.create_course_and_seat()
+        new_range, __ = Range.objects.get_or_create(course_catalog=catalog_id, course_seat_types='verified')
+        new_range.add_product(seat)
+        voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
+
+        # Mock network calls
+        self.mock_dynamic_catalog_course_runs_api(query=catalog_query, course_run=course)
+        self.mock_fetch_course_catalog_with_failure(catalog_id=catalog_id, expected_status=status.HTTP_404_NOT_FOUND)
```
vkaracic: The lines above this one are almost exactly the same in all tests you've added. Please factor them out into a single method and use that instead, i.e. [don't repeat yourself](http://deviq.com/don-t-repeat-yourself/).

vkaracic (approved):
One comment, otherwise LGTM.

mattdrayer:
Looks like another rebase is needed -- 👍 assuming all checks pass after doing that.

PR #1135, opened 2017-02-07
Only send consumer_id for hosted payment page.

mjfrey:
When using SOP profile, do not send optional consumer_id.


mjfrey:
@clintonb please review.

clintonb (approved):

PR #1137, opened 2017-02-08
ECOM-6928 : Revert [SOL-2083] Because it enables student to create multiple orders for same course seat

Ayub-Khan:
Reverts edx/ecommerce#953

**ECOM-6928** :
In SOL-2083 changes enrollment check to work only for coupon baskets, thus enabling other students to place multiple orders for same course(aka Multiple Order Issue). 

**Steps to Reproduce:**
1. Enroll in a verified course on stage.
2. Open Basket page from dashboard in new tab and pay for the course.
3. Open Basket page from same dashboard again and pay for the course.

**Expected** : 
Student should not be able to pay multiple times for single course seat.
**Actual** : 
Student can pay for a course multiple times. 

fyi @mjfrey

### DIFF
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -168,13 +168,12 @@ class BasketSingleItemViewTests(CouponMixin, CourseCatalogTestMixin, CourseCatal
         """
         course = CourseFactory()
         self.mock_enrollment_api_success_enrolled(course.id, mode=mode)
-        product = course.create_or_update_seat(mode, id_verification, 0, self.partner)
+        product = course.create_or_update_seat(mode, id_verification, 0, self.partner, create_enrollment_code=False)
         stock_record = StockRecordFactory(product=product, partner=self.partner)
         catalog = Catalog.objects.create(partner=self.partner)
         catalog.stock_records.add(stock_record)
-        self.create_coupon(catalog=catalog, code=COUPON_CODE, benefit_value=5)
 
-        url = '{path}?sku={sku}&code={code}'.format(path=self.path, sku=stock_record.partner_sku, code=COUPON_CODE)
+        url = '{path}?sku={sku}'.format(path=self.path, sku=stock_record.partner_sku)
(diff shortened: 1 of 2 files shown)


clintonb (changes requested):
This is not doing what you described in the description.

Ayub-Khan:
@clintonb sorry i did not get your point. Can you explain. There are two checks 
- "product.get_product_class().name != ENROLLMENT_CODE_PRODUCT_CLASS_NAME:"
- "request.user.is_user_already_enrolled(request, product)" (inner if)

First one makes sure that second one does not work for enrollment code orders. While it works for all other baskets. Second check makes sure that if a student is already in verified mode for a course he should not be able to place any other order for that. But adding coupon code condition to first one makes sure that second one only runs when there is a coupon in the basket. Thus enabling students to create multiple orders for the course.

clintonb (approved):

Ayub-Khan:
@mjfrey can you review ?

mjfrey (approved):
I think you will need to add some tests.

Ayub-Khan:
@mjfrey tests are already there.

Ayub-Khan:
jenkins run all

clintonb:
@Ayub-Khan @adampalay this needs to be revisited. The error page seen when a user is already enrolled is a simple line of text on an otherwise empty page. This is poor user experience. I am inclined to revert this until a proper UI is in place.

Ayub-Khan:
@clintonb I agree this is a bad user experience but as compare to letting user pay multiple times its bearable. Also for user to place multiple orders and payments and than getting refund for those would be worse. In my opinion you should let this be there and create a ticket to create or update response page when this occurs, there is no need to revert this as this blocks 43 average multiple payment attempts every day. 
fyi @adampalay

rlucioni:
This change appears to be tied to auth-related errors preventing users from enrolling and should be reverted until the issue is better understood. See [Splunk](https://splunk.edx.org/en-US/app/search/search?sid=1488257147.4684184).

PR #1140, opened 2017-02-10
Add Backend Validation for Note

iivic:
SOL-2204
The only requirement is that note must be type str or unicode (basestring).
I changed the ValidationError handling in the update method so notice that as well.
@vkaracic @marjev  @mjfrey

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -357,6 +357,12 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         vouchers = new_coupon.attr.coupon_vouchers.vouchers
         return vouchers.first().offers.first().benefit.range, data
 
+    def _create_and_get_coupon_details(self):
+        """Helper method that creates and returns coupon details."""
+        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        coupon = Product.objects.get(title=self.data['title'])
+        return self.get_response_json('GET', reverse('api:v2:coupons-detail', args=[coupon.id]))
+
     def test_create_serializer_data(self):
         """Test if coupon serializer creates data for details page"""
         details_response = self.get_response_json(
@@ -483,13 +489,7 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
             'quantity': 1,
             'title': 'Tešt čoupon 2'
         })
-        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
-        coupon = Product.objects.get(title=self.data['title'])
-        details_response = self.get_response_json(
-            'GET',
-            reverse('api:v2:coupons-detail', args=[coupon.id]),
-            data=self.data
-        )
+        details_response = self._create_and_get_coupon_details()
         self.assertEqual(details_response['code'], self.data['code'])
         self.assertEqual(details_response['coupon_type'], 'Discount code')
         self.assertEqual(details_response['enterprise_customer'], self.data['enterprise_customer']['id'])
@@ -890,13 +890,11 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
             'title': 'Max uses update',
             'voucher_type': Voucher.MULTI_USE
         })
-        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
-        coupon = Product.objects.get(title=self.data['title'])
-        details = self.get_response_json('GET', reverse('api:v2:coupons-detail', args=[coupon.id]))
+        details = self._create_and_get_coupon_details()
         self.assertEqual(details['max_uses'], self.data['max_uses'])
 
         self.data['max_uses'] = 5
-        response = self.get_response_json('PUT', reverse('api:v2:coupons-detail', args=[coupon.id]), self.data)
+        response = self.get_response_json('PUT', reverse('api:v2:coupons-detail', args=[details['id']]), self.data)
         self.assertEqual(response['max_uses'], self.data['max_uses'])
 
     def test_create_coupon_without_category(self):
@@ -988,14 +986,36 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
             'title': 'Coupon update with max uses {}'.format(max_uses),
             'voucher_type': Voucher.MULTI_USE
         })
-        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
-        coupon = Product.objects.get(title=self.data['title'])
-        details = self.get_response_json('GET', reverse('api:v2:coupons-detail', args=[coupon.id]))
+        details = self._create_and_get_coupon_details()
         self.assertEqual(details['max_uses'], 2)
 
         self.data.update({'max_uses': max_uses})
         response = self.client.put(
-            reverse('api:v2:coupons-detail', args=[coupon.id]),
+            reverse('api:v2:coupons-detail', args=[details['id']]),
+            json.dumps(self.data),
+            'application/json'
+        )
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    @ddt.data(1, {'some': 'dict'}, ['array'])
+    def test_create_coupon_with_invalid_note(self, invalid_note):
+        """Verify creating coupon with invalid note field returns 400 response."""
+        self.data['note'] = invalid_note
+        self.assert_post_response_status(self.data)
+
+    @ddt.data(1, {'some': 'dict'}, ['array'])
+    def test_update_coupon_with_invalid_note(self, invalid_note):
+        """Verify updating coupon with invalid note field returns 400 response."""
+        self.data.update({
+            'note': 'Some valid note.',
+            'title': 'Coupon update with note {}'.format(invalid_note),
+        })
+        details = self._create_and_get_coupon_details()
+        self.assertEqual(details['note'], self.data['note'])
+
+        self.data['note'] = invalid_note
+        response = self.client.put(
+            reverse('api:v2:coupons-detail', args=[details['id']]),
             json.dumps(self.data),
             'application/json'
         )
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -347,7 +347,12 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
             serializer = self.get_serializer(coupon)
             return Response(serializer.data)
         except ValidationError as error:
-            raise serializers.ValidationError(error.message)
+            error_message = 'Failed to update Coupon [{coupon_id}]. {msg}'.format(
+                coupon_id=coupon.id,
+                msg=error.message
+            )
+            logger.exception(error_message)
+            raise serializers.ValidationError(error_message)
 
     def create_update_data_dict(self, data, fields):
         """
@@ -439,10 +444,7 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
                     if offer_data['max_global_applications'] < 1:
                         raise ValueError
                 except ValueError:
-                    log_message_and_raise_validation_error(
-                        'Failed to update Coupon [{coupon_id}]. '
-                        'max_global_applications field must be a positive number.'.format(coupon_id=coupon_id)
-                    )
+                    raise ValidationError('max_global_applications field must be a positive number.')
             ConditionalOffer.objects.filter(vouchers__in=vouchers.all()).update(**offer_data)
 
     def destroy(self, request, pk):  # pylint: disable=unused-argument
--- a/ecommerce/extensions/catalogue/models.py
+++ b/ecommerce/extensions/catalogue/models.py
@@ -1,4 +1,3 @@
-# noinspection PyUnresolvedReferences
 from django.db import models
 from django.db.models.signals import post_init, post_save
 from django.dispatch import receiver
@@ -9,6 +8,7 @@ from simple_history.models import HistoricalRecords
 from ecommerce.core.constants import (
     COUPON_PRODUCT_CLASS_NAME, ENROLLMENT_CODE_PRODUCT_CLASS_NAME, SEAT_PRODUCT_CLASS_NAME
 )
+from ecommerce.core.utils import log_message_and_raise_validation_error
 
 
 class Product(AbstractProduct):
@@ -18,7 +18,6 @@ class Product(AbstractProduct):
     expires = models.DateTimeField(null=True, blank=True,
                                    help_text=_('Last date/time on which this product can be purchased.'))
     history = HistoricalRecords()
-
     original_expires = None
 
     @property
@@ -32,6 +31,16 @@ class Product(AbstractProduct):
     @property
     def is_coupon_product(self):
         return self.get_product_class().name == COUPON_PRODUCT_CLASS_NAME
+
+    def save(self, *args, **kwargs):
+        try:
+            if not isinstance(self.attr.note, basestring) and self.attr.note is not None:
+                log_message_and_raise_validation_error(
+                    'Failed to create Product. Product note value must be of type string'
+                )
+        except AttributeError:
+            pass
+        super(Product, self).save(*args, **kwargs)  # pylint: disable=bad-super-call
 
 
 @receiver(post_init, sender=Product)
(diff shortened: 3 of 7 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1140&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/catalogue/models.py
```diff
@@ -4,13 +4,28 @@
...
     course = models.ForeignKey('courses.Course', null=True, blank=True, related_name='products')
     expires = models.DateTimeField(null=True, blank=True,
                                    help_text=_('Last date/time on which this product can be purchased.'))
     history = HistoricalRecords()
 
+    def save(self, *args, **kwargs):
+        self.clean()
+        super(Product, self).save(*args, **kwargs)  # pylint: disable=bad-super-call
+
+    def clean(self):
```
vkaracic: ``Product`` already has a [clean](https://github.com/django-oscar/django-oscar/blob/1.1.1/src/oscar/apps/catalogue/abstract_models.py#L367) method. This is how it works for ``note`` attribute (or any [text](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/catalogue/migrations/0016_coupon_note_attribute.py#L18) type attributes):
```
p = Product.objects.filter(product_class__name='Coupon').first()
p  # <Product: Test1>
p.attr.note = 123
p.clean()
ValidationError: [u"note attribute [u'Must be str or unicode']"]
```

Try to either re-use the already existing validation, or adjust it to your needs but don't override ``clean`` like this.

vkaracic (changes requested):

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -910,6 +910,32 @@ def test_update_coupon_with_invalid_max_uses(self, max_uses):
...
+    @ddt.data(1, {'some': 'dict'}, ['array'])
+    def test_update_coupon_with_invalid_note(self, invalid_note):
+        """Verify updating coupon with invalid note field returns 400 response."""
+        self.data.update({
+            'note': 'Some valid note.',
+            'title': 'Coupon update with note {}'.format(invalid_note),
+        })
+        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        coupon = Product.objects.get(title=self.data['title'])
+        details = self.get_response_json('GET', reverse('api:v2:coupons-detail', args=[coupon.id]))
```
marjev: Lines 927 and 928 are called consecutively in almost all test methods for updating coupon fields. It would be a good time to refactor this.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -312,7 +312,12 @@ def update(self, request, *args, **kwargs):
             serializer = self.get_serializer(coupon)
             return Response(serializer.data)
         except ValidationError as error:
-            raise serializers.ValidationError(error.message)
+            error_message = 'Failed to update Coupon [{coupon_id}]. {msg}'.format(
+                coupon_id=coupon.id,
+                msg=error.message
+            )
+            logger.exception(error_message)
+            raise serializers.ValidationError(error_message)
```
marjev, line 355: why not use `log_message_and_raise_validation_error `?

ecommerce/extensions/catalogue/models.py
```diff
@@ -4,13 +4,28 @@
...
+        try:
+            if not isinstance(self.attr.note, basestring) and self.attr.note is not None:
+                log_message_and_raise_validation_error(
+                    'Failed to create Product. Product note value must be type string'
```
marjev: nitpick: `must be of type string`.

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -0,0 +1,39 @@
+import ddt
+
```
marjev: remove this space.

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -0,0 +1,39 @@
...
+class ProductTests(TestCase):
+    def test_create_product_with_note(self):
+        """Verify creating a product with valid note value creates product."""
+        note = 'Some test note 2.'
+        product_class = ProductClass.objects.create(name='Coupon', slug='coupon')
+        coupon_product = Product.objects.create(
+            id=1,
+            title='Test title',
+            product_class=product_class
+        )
```
marjev: Don't we have factories for `Product_class` and `Product`? If not, it's time to create those or update existing if, for any reason, you're not able to use those to create test objects.

marjev (changes requested):

ecommerce/extensions/api/v2/views/coupons.py
iivic: This is not the same [ValidationError](http://www.django-rest-framework.org/api-guide/exceptions/#validationerror) as the django ValidationError.

iivic:
Ready for second review.

marjev (approved):

ecommerce/extensions/catalogue/models.py
```diff
@@ -1,16 +1,28 @@
...
                                    help_text=_('Last date/time on which this product can be purchased.'))
     history = HistoricalRecords()
 
+    def save(self, *args, **kwargs):
+        if hasattr(self.attr, 'note'):
+            try:
+                self.clean()
+            except ValidationError:
+                log_message_and_raise_validation_error(
+                    'Failed to create Product. Product note value must be of type string.'
```
vkaracic: ``clean()`` does a validation of all the product attributes (among other properties, [see table](https://github.com/django-oscar/django-oscar/blob/1.1.1/src/oscar/apps/catalogue/abstract_models.py#L367)). If one of the other than ``note`` fails you will log this.
Lose the attribute check and just do this:
```
try:
    self.clean()
except ValidationError as e:
    log_message_and_raise_validation_error(e.message)
```

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -0,0 +1,42 @@
...
+        coupon_product = factories.ProductFactory(
+            title=self.COUPON_PRODUCT_TITLE,
+            product_class=coupon_product_class
+        )
+        factories.ProductAttributeFactory(
+            code='note',
+            name='Note',
+            product_class=coupon_product_class,
+            type='text'
+        )
```
vkaracic: You could re-use this: https://github.com/edx/ecommerce/blob/master/ecommerce/coupons/tests/mixins.py#L146

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -944,13 +943,37 @@ def test_update_coupon_with_invalid_max_uses(self, max_uses):
...
+
+    @ddt.data(1, {'some': 'dict'}, ['array'])
+    def test_update_coupon_with_invalid_note(self, invalid_note):
+        """Verify updating coupon with invalid note field returns 400 response."""
+        self.data.update({
+            'note': 'Some valid note.',
+            'title': 'Coupon update with note {}'.format(invalid_note),
+        })
+        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        details = self._get_coupon_details(self.data['title'])
```
vkaracic: Nit: the POST call above is above every call of this abstraction, maybe factor it in as well?

vkaracic (changes requested):

iivic:
Ready for another review.

ecommerce/extensions/catalogue/models.py
```diff
@@ -1,16 +1,28 @@
...
+
 
 class Product(AbstractProduct):
     course = models.ForeignKey('courses.Course', null=True, blank=True, related_name='products')
     expires = models.DateTimeField(null=True, blank=True,
                                    help_text=_('Last date/time on which this product can be purchased.'))
     history = HistoricalRecords()
 
+    def save(self, *args, **kwargs):
+        if hasattr(self.attr, 'note'):
```
vkaracic: Why would you call ``clean`` only for coupons with ``note``? Shouldn't everything be validated?

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -0,0 +1,42 @@
...
+class ProductTests(TestCase, CouponMixin):
+    COUPON_PRODUCT_TITLE = 'Some test title.'
+
+    def _create_coupon_product_with_note_attribute(self, note):
+        """Helper method that creates a coupon product with note attribute set."""
+        coupon_product = factories.ProductFactory(
+            title=self.COUPON_PRODUCT_TITLE,
+            product_class=self.coupon_product_class
+        )
+        voucher = factories.VoucherFactory(code='TESTVOUCHER')
```
vkaracic: Do you need the ``code`` parameter?

vkaracic (approved):
A couple of nits. Going forward we should definitely consider creating a coupon factory.

ecommerce/extensions/catalogue/models.py
iivic: Because the validation will fail if the note isn't set for a coupon product. I'm validating the field from the coupon creation/edit form. If you want to make the validation for everything I wish you good luck.

vkaracic: Because it isn't set? That's odd since it's set to not be required: https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/catalogue/migrations/0016_coupon_note_attribute.py#L19
What's the error message that you get?

iivic: It is not required when not calling clean() on Product but when clean is called it will require all the attributes to be set (title, coupon_vouchers, note). The title and coupon_vouchers will be set on every coupon but the note field can be empty and that's why I first check if the attribute is set.
The message is "note attribute cannot be blank"

ecommerce/extensions/catalogue/tests/test_models.py
iivic: No copy/paste leftovers.

iivic:
Ditto on the coupon factory. It will make testing easier.

ecommerce/extensions/catalogue/models.py
vkaracic: I found why that is happening and a couple of ways to fix it, BUT I'm not telling :smiley: 
It's a fun problem to tinker about (at least it was fun for me) and since ``note`` validation is not of crucial importance right now, I'd like to see what you'll come up with. Hopefully by the end of this exercise you'll be a bit more familiar by the inner workings of Django and Oscar.
To be clear, the assignment is to find a way to call ``clean()`` in the ``save()`` method successfully.
Have fun! :)
If necessary, I can give hints in exchange for free candy.

ecommerce/extensions/catalogue/utils.py
```diff
@@ -68,11 +68,17 @@ def create_coupon_product(
...
     """
     product_class = ProductClass.objects.get(slug='coupon')
-    coupon_product = Product.objects.create(title=title, product_class=product_class)
+    coupon_product = Product(title=title, product_class=product_class)
+    coupon_product.attr.coupon_vouchers = CouponVouchers()
+    coupon_product.attr.note = note
+    coupon_product.save()
+    coupon_product.attr.coupon_vouchers.coupon = coupon_product
+    coupon_product.attr.coupon_vouchers.save()
+
```
vkaracic: I would instantiate CouponVouchers separately:
```
coupon_vouchers = CouponVouchers()
...
coupon_vouchers.save()
...
vouchers = create_vouchers(...)
coupon_vouchers.add(*vouchers)
```

iivic:
I removed the oscar clean method call before save because a lot of tests need to be rewritten (around 520 tests failed). Now just the note type s being tested. I also moved the coupon_vouchers from create_vouchers method since it makes no sense keeping it there.

iivic:
@marjev  @vkaracic  Please take a quick look before I merge this.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -357,6 +357,12 @@ def _get_voucher_range_with_updated_dynamic_catalog_values(self):
         vouchers = new_coupon.attr.coupon_vouchers.vouchers
         return vouchers.first().offers.first().benefit.range, data
 
+    def _create_and_get_coupon_details(self):
+        """Helper method that creates and returns coupon details."""
+        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        coupon = Product.objects.get(title=self.data['title'])
+        return self.get_response_json('GET', reverse('api:v2:coupons-detail', args=[coupon.id]))
```
vkaracic, line 364: Since this returns only details, is it possible to create the coupon directly without sending a POST request to the coupons endpoint and use that to get the details? That will save up some testing time.

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -0,0 +1,42 @@
...
+        coupon_vouchers.vouchers.add(voucher)
+        coupon_product.attr.coupon_vouchers = coupon_vouchers
+        coupon_product.attr.note = note
+        coupon_product.save()
+        return coupon_product
+
+    def test_create_product_with_note(self):
+        """Verify creating a product with valid note value creates product."""
+        note = 'Some test note.'
+        self._create_coupon_product_with_note_attribute(note)
```
vkaracic: This method should return the coupon product back, so you can use that one instead of the making another db call below.

vkaracic (changes requested):

ecommerce/extensions/api/v2/tests/views/test_coupons.py
iivic: I guess a Product could be created but we want to test if the api endpoint works as expected because of the validation so not calling the api would make the tests useless. If you have a better solution feel free to advice me.

iivic:
@vkaracic Comment addressed the other needs further explanation.

vkaracic (approved):

ecommerce/extensions/catalogue/models.py
```diff
@@ -1,9 +1,10 @@
-# noinspection PyUnresolvedReferences
```
marjev, line 1: Why is this being removed?

ecommerce/extensions/catalogue/models.py
```diff
@@ -13,6 +14,16 @@ class Product(AbstractProduct):
...
     history = HistoricalRecords()
 
+    def save(self, *args, **kwargs):
+        try:
+            if not isinstance(self.attr.note, basestring) and self.attr.note is not None:
+                log_message_and_raise_validation_error(
+                    'Failed to create Product. Product note value must be of type string.'
+                )
+        except AttributeError:
+            pass
```
marjev: Instead of overriding `save`, can you use this:
https://github.com/django-oscar/django-oscar/blob/31ea8dcbcce4ccb6bcbdbb08c3fe9b01103fae2c/src/oscar/apps/catalogue/product_attributes.py#L36

If not, you can use:
https://docs.djangoproject.com/en/1.10/ref/signals/#pre-save

ecommerce/extensions/catalogue/models.py
```diff
@@ -30,6 +41,4 @@ def __unicode__(self):
             catalog_name=self.name
         )
 
-
-# noinspection PyUnresolvedReferences
```
marjev, line 86: Again, why exactly is this being removed?

ecommerce/extensions/voucher/utils.py
```diff
@@ -346,7 +346,7 @@ def _generate_code_string(length):
     return voucher_code
 
 
-def _create_new_voucher(code, coupon, end_datetime, name, offer, start_datetime, voucher_type):
```
marjev, line 349: Why remove `coupon`?

If you need to remove it, however, change the docstring.

marjev (changes requested):

ecommerce/extensions/catalogue/models.py
iivic: As far as I know this is PyCharm specific and I can't see it's use since we are disabling pylint where it's necessary.

ecommerce/extensions/catalogue/models.py
iivic: This statement was [redundant](https://github.com/edx/ecommerce/pull/1140/files/f4e3a67f16162b4dd834f860eb817889c17932a8#diff-115852c6fffaa39425a3d451b1b81285L1).

ecommerce/extensions/voucher/utils.py
iivic: I concluded that the coupon is not needed here since the functions name implies that only a voucher is created.  I moved the voucher adding to coupon_voucher into `create_coupon_product` function.
I will remove the docstring.

ecommerce/extensions/catalogue/models.py
iivic: `validate_attributes` will make a bunch of tests fail (around 560) so I decided to validate the field directly instead of fixing every test.

Overriding the save method is consistent throughout the code. If you think pre-save is the way to go we should use it in the future as well.

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -1,10 +1,36 @@
+import ddt
 from django.utils.timezone import now, timedelta
+from django.core.exceptions import ValidationError
```
vkaracic: Alphabetize please.

vkaracic (changes requested):

marjev (approved):

vkaracic (approved):

PR #1143, opened 2017-02-12
Updated npm dependencies

clintonb:
ECOM-7141


clintonb:
@edx/ecommerce please review

ecommerce/static/js/pages/basket_page.js
```diff
@@ -391,7 +389,7 @@ define([
             onReady: onReady,
             onSuccess: onSuccess,
             sdnCheck: sdnCheck,
-            showVoucherForm: showVoucherForm,
+            showVoucherForm: showVoucherForm
         };
     }
 );
```
AlasdairSwan, line 397: @clintonb it would be great if we could move this file to Backbone to get some better structure in place for it (although probably out of scope of this PR).

clintonb: It took a lot for me not to do that yesterday. This is one of about three rabbit holes I fell down while working on Stripe payments.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -15,42 +15,22 @@ define([
...
+        function hideVoucherForm() {
             $('#voucher_form_container').hide();
             $('#voucher_form_link').show();
-        },
-        onFail = function(){
+        }
+
+        function onFail() {
             var message = gettext('Problem occurred during checkout. Please contact support');
             $('#messages').empty().append(
```
AlasdairSwan: Why not just ```$('#messages').html(```?

AlasdairSwan (approved):
It would be great to schedule the work to move this to Backbone and there's a nit about how a DOM element is being updated but it's a judgement call so 👍

rlucioni (approved):

ecommerce/static/js/pages/basket_page.js
clintonb: Your guess is as good as mine. Fixed.

clintonb:
@AlasdairSwan please review the latest commit: https://github.com/edx/ecommerce/pull/1143/commits/83b38bd4ee5f4afa78e84c2487c1f64ed5e748d7

AlasdairSwan (approved):
:shipit:

PR #1144, opened 2017-02-13
Skipping broken test

clintonb:
This test consistently fails on Travis, but not locally. Skipping it.

### DIFF
--- a/ecommerce/extensions/dashboard/orders/tests.py
+++ b/ecommerce/extensions/dashboard/orders/tests.py
@@ -1,4 +1,5 @@
 import os
+from unittest import skipIf
 
 from django.contrib.messages import constants as MSG
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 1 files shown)


ecommerce/extensions/dashboard/orders/tests.py
```diff
@@ -76,9 +77,6 @@ def login(self, path):
 
     def retry_fulfillment(self):
         """ Click the retry fulfillment button and wait for the AJAX call to finish. """
-        # Wait for the button to appear, otherwise calling tests may fail on Travis
```
rlucioni, line 79: Any leads on why this was failing on Travis only?

clintonb: (shrug) It makes no sense. Without a screenshot or HTML output, I have no leads.

PR #1145, opened 2017-02-13
SDN Cleanup

clintonb:
- Added missing translations
- Cleaned up template text to make it simpler to translate

screen shot 2017-02-13 at 12 40 03 am


clintonb:
@mjfrey please review

mjfrey (approved):

PR #1146, opened 2017-02-13
Updated Travis build

clintonb:
The run_tests.sh script is replaced with individual steps in the Travis configuration file. This allows us to better measure the execution time of individual steps of the build, and actually fail the build when individual steps fail.

Additionally, the apt packages that were once installed by run_tests.sh are now installed in the Docker image by default.


clintonb:
This is coupled with https://github.com/edx/configuration/pull/3692.

@edx/devops @edx/ecommerce please review

.travis.yml
```diff
@@ -13,18 +13,25 @@ cache:
...
 before_install:
-    - docker-compose -f ./.travis/docker-compose-travis.yml up -d
+  - docker-compose -f ./.travis/docker-compose-travis.yml up -d
+
+install:
+  - docker exec -t ecommerce_testing  bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ && make requirements'
 
 script:
-    - docker exec ecommerce_testing /edx/app/ecommerce/ecommerce/.travis/run_tests.sh
+  - docker exec -t ecommerce_testing  bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/docs && make html'
```
rlucioni, line 27: Can we please keep `run_tests.sh`? I'd like to keep this repo consistent with others using Travis for testing, if possible. Keeping the script in a separate file made it easy to read. It also avoids the need to repeatedly duplicate `docker exec -t ecommerce_testing  bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ && DJANGO_SETTINGS_MODULE=ecommerce.settings.test`.

Also, take a look at the [Make targets](https://github.com/edx/course-discovery/blob/master/.travis/docker.mk) I added for this in discovery. Doing the same here would be an easy way to clean this up.

rlucioni: I just read your commit message. If we're going to go this route, can we at least tie the steps to a Make target so it's easy to run locally when necessary?

clintonb: `run_tests.sh` needs to go or be fixed. Multiple PRs have passed with missing migrations. I find it easier to use `.travis.yml` as it has been for a while than deal with Bash scripts.

Running these steps manually (from the host machine, as opposed to the container's shell) is an exceptional case to which I see no need to cater. Adding Make targets to cut down on duplication doesn't make much sense given that every line of the script is calling a Make target. If you want to run this locally, it's pretty easy to open a shell into the container and run the Make targets directly.

rlucioni: That's fair. I didn't realize builds were failing silently.

rlucioni (approved):

PR #1149, opened 2017-02-13
Change enrollment code redemption to the receipt page.

vkaracic:
Reopening https://github.com/edx/ecommerce/pull/963
SOL-2073

Either the order fulfillment takes more than 10 seconds to finish, or this code somehow does not work. @clintonb thoughts?


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1149&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/coupons/views.py
```diff
@@ -161,11 +164,22 @@ def get(self, request):
...
-
-        return HttpResponseRedirect(request.site.siteconfiguration.student_dashboard_url)
+            order = self.place_free_order(basket)
+
+            # Waiting for the order to finish its async fulfillment.
+            # Every 2 seconds the order status is checked. If after five checks
+            # it's not complete or the status is an error, user is redirected
+            # to the checkout error page.
+            for __ in range(5):  # pragma: no cover
+                order.refresh_from_db()
```
vkaracic, line 174: This should work, but one idea would be to change it to
``order = Order.objects.get(id=order.id)``

PR #1152, opened 2017-02-13
Removed configuration validation on startup

clintonb:
The false-positives raised by this validation for tests has lead us (me, at least) to ignore this validation. When developing locally, bad data prevents the application from starting, which prevents me from easily fixing that data via Django admin. This check costs more than it's worth. It's time for it to go.

ECOM-7108


clintonb:
@edx/ecommerce please review
FYI @edx/devops

rlucioni:
Thanks @clintonb, this will stop us from seeing useless errors related to this in the logs.

clintonb:
@rlucioni yes. Yes it will. In theory, this check was a good idea. In practice, it was ignored. :|

PR #1154, opened 2017-02-14
Upgraded to Django 1.9.12

clintonb:
ECOM-7096


clintonb:
@edx/ecommerce please review the final commit. Most of the changes are resolving Django 1.10 deprecation warnings.

rlucioni (approved):
Nice.

ecommerce/core/migrations/0026_auto_20170215_2234.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('core', '0025_auto_20170214_0003'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='user',
+            name='username',
+            field=models.CharField(error_messages={'unique': 'A user with that username already exists.'}, help_text='Required. 30 characters or fewer. Letters, digits and @/./+/-/_ only.', max_length=30, unique=True, validators=[django.core.validators.RegexValidator('^[\\w.@+-]+$', 'Enter a valid username. This value may contain only letters, numbers and @/./+/-/_ characters.')], verbose_name='username'),
```
clintonb, line 19: This migration *should not* affect the database. The username length is currently 30 characters and unique. This seems to be a result of Django making migrations for non-database-related model field changes (e.g. `help_text`).

saleem-latif:
@clintonb We are getting `404` error while accessing /api/v2/catalogs/course_catalogs/ . Error is present on all `list_route` for [catalogs](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/v2/views/catalog.py#L23-L92). Error is present on all commits after [966cedea42efcd795dad2b2b1b572742cfae7a2d](https://github.com/edx/ecommerce/commit/966cedea42efcd795dad2b2b1b572742cfae7a2d). 

@mattdrayer , @zubair-arbi -- FYI

clintonb:
This, folks, is why we write tests that actually go through the URL routing instead of calling the view directly. I'll submit a patch shortly.

FYI @zubair-arbi @vkaracic @marjev ^

PR #1155, opened 2017-02-14
Running Travis tests with MySQL

clintonb:
ECOM-7111

### DIFF
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -6,7 +6,7 @@ import pytz
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.utils.timezone import now
-from django.utils.translation import ugettext_lazy as _
+from factory.fuzzy import FuzzyText
 from oscar.core.loading import get_class, get_model
 from oscar.test.factories import (
     ConditionalOfferFactory, OrderFactory, OrderLineFactory, RangeFactory, VoucherFactory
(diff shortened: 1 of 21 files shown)


ecommerce/coupons/tests/test_views.py
```diff
@@ -6,7 +6,7 @@
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.utils.timezone import now
-from django.utils.translation import ugettext_lazy as _
+from factory.fuzzy import FuzzyText
```
rlucioni, line 9: Remember that these fuzzers are deprecated. It'd be great to use Faker instead.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -67,7 +67,7 @@ def setUp(self):
         factories.ProductFactory(
             structure='child',
             parent=self.base_product,
-            title=u'𝐋𝐏 𝟓𝟕𝟎-𝟒 𝐒𝐮𝐩𝐞𝐫𝐥𝐞𝐠𝐠𝐞𝐫𝐚',
+            title='LP 570-4 Superleggra',
```
rlucioni: s/Superleggra/[Superleggera](https://i.ytimg.com/vi/D_iyckPiD9g/maxresdefault.jpg)

rlucioni (approved):
Thanks for slogging through this, I know it can be rough.

ecommerce/coupons/tests/test_views.py
clintonb: We are relying on `factory_boy` from `django-oscar`. I am hesitant to upgrade in case the fuzzers get removed. Future Oscar versions may resolve this.

PR #1156, opened 2017-02-15
Updated to Django Oscar 1.2.2

clintonb:
ECOM-7095


clintonb:
@edx/ecommerce please review

clintonb:
FYI @mjfrey note the removal of the custom messages.

ecommerce/extensions/basket/views.py
```diff
@@ -317,62 +315,3 @@ def get_context_data(self, **kwargs):
...
-    View that applies a voucher to basket.
-    We change default messages oscar returns.
-    """
-
-    def form_valid(self, form):
-        super(VoucherAddMessagesView, self).form_valid(form)
-
-        code = form.cleaned_data['code']
-
-        for msg in list(messages.get_messages(self.request)):
```
mjfrey: We need these messages. Why are they removed?

clintonb: @mjfrey here are the default views now being used:
* https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/basket/views.py#L302
* https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/basket/views.py#L384

Eh...looks like they are still using the word voucher.

.travis.yml
```diff
@@ -13,21 +13,28 @@ cache:
     - ecommerce/static/bower_components
 
 addons:
-    firefox: latest
```
MatthewPiatetsky: why remove this?

clintonb: See https://github.com/edx/ecommerce/pull/1146. Ignore the first commit.

Makefile
```diff
@@ -32,7 +32,8 @@ help:
 
 requirements.js:
 	npm install
-	$(NODE_BIN)/bower install
+	# Allow root for Docker
```
MatthewPiatetsky: what does this mean?

ecommerce/extensions/basket/views.py
```diff
@@ -319,60 +318,75 @@ def get_context_data(self, **kwargs):
...
-                messages.error(
-                    self.request,
-                    _("Coupon code '{code}' is invalid.").format(code=code)
-                )
-
+                self.apply_voucher_to_basket(voucher)
         return redirect_to_referrer(self.request, 'basket:summary')
-
-
-class VoucherRemoveMessagesView(VoucherRemoveView):
```
mjfrey: Unless this is fixed in the latest Oscar we still need this as we will start getting Server errors in the case that triggers this.

mjfrey (changes requested):
Can we also see this in a sandbox or anywhere we can run some tests?

ecommerce/extensions/basket/views.py
clintonb: Under what conditions do these errors manifest? Django itself supports string lookups. I've never seen these issues.

mjfrey: To reproduce:
1: Add item to basket, apply voucher.
2: Open second tab go to basket page.
3: First tab remove voucher.
4: Second tab remove voucher.
Server error will happen in Oscar.

clintonb: Was an issue filed at https://github.com/django-oscar/django-oscar/?

mjfrey: Not that I am aware of.

clintonb: https://github.com/django-oscar/django-oscar/pull/2241

clintonb: I'll restore this view for now; although, I don't think we should support multi-tab use cases.

mjfrey: I agree, however we where getting Splunk alerts from this.

clintonb:
@mjfrey sandbox building. Any other feedback?

mjfrey (approved):

PR #1157, opened 2017-02-15
Updated to Django Oscar 1.4

clintonb:
ECOM-7095


clintonb:
@edx/ecommerce please review the final commit

rlucioni (approved):

PR #1159, opened 2017-02-15
Add site to SDN Failure model.

vkaracic:
Not sure if the ``site`` field should be ``null=True, blank=True``-ed. In case it's not I can create a management command to populate ``site`` attribute of existing SDN Failure entries on edX and WL(s ?) and remove the ``null=True, blank=True`` parameter.

https://openedx.atlassian.net/browse/SOL-2205


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1159&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mjfrey (approved):
We can let them be blank, since this has not been deployed yet.

PR #1161, opened 2017-02-15
Fixed payment refactor issues

clintonb:
- Corrected height of the card number input field
- Removed compress tags from CyberSource template to avoid loading issues
- Using default verbosity for collectstatic/compress commands to aid future debugging

ECOM-7141


clintonb:
@mjfrey please review

FYI @edx/ecommerce

mjfrey (approved):

PR #1163, opened 2017-02-16
Updated base template

clintonb:
These changes incorporate changes from edx/edx-themes, allowing us to remove the duplicate templates in that repository.

ECOM-7141


clintonb:
FYI @mjfrey @douglashall @edx/ecommerce

mjfrey (approved):
I assume a follow on PR to remove these files from the edx theme.

PR #1164, opened 2017-02-16
Disabled console logging for tests

clintonb:
Our test logs are so large Travis doesn't display them on the default build summary page. Most of the information logged is useless. In the event of an error, the logs will displayed alongside a stack trace thanks to nose. This change disables console logging, removing the extraneous log output.

ECOM-3054


clintonb:
@edx/ecommerce please review

rlucioni (approved):

PR #1165, opened 2017-02-16
Add base client side checkout basket template.

vkaracic:
https://openedx.atlassian.net/browse/ECOM-7248


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1165&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

douglashall:
@vkaracic I think you want to leave `client_side_checkout_basket.html`, create `seat_type.html`, and include it in `client_side_checkout_basket.html`.

vkaracic:
You're right, that is better.

clintonb (approved):

mjfrey (approved):

PR #1166, opened 2017-02-16
Add address and products to SDN failure model.

vkaracic:
The SDN failure entry needs to contain additional information about the user's address and the products that the user tried to buy. I've implemented products as a M2M field to support multi-line baskets.

https://openedx.atlassian.net/browse/SOL-2205
https://openedx.atlassian.net/browse/ECOM-7251


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1166&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -169,6 +169,7 @@ define([
         function sdnCheck(event) {
             var first_name = $('input[name=first_name]').val(),
                 last_name = $('input[name=last_name]').val(),
+                address = $('input[name=address_line1]').val(),
```
mjfrey, line 172: I think we want the city / state as well? This will just give us the street.

ecommerce/extensions/payment/utils.py
```diff
@@ -93,6 +94,7 @@ def search(self, name, country):
             'api_key': self.api_key,
             'type': 'individual',
```
mjfrey, line 95: We also need to change this to individual,entity

ecommerce/static/js/pages/basket_page.js
vkaracic: I don't know. Not even sure if address should be sent at all since the only example in the documentation uses city for the address.

vkaracic: @mjfrey for now I'll join the two (address and city) with a comma, but if you could I'd still like to have an answer for this. An address query string parameter would indicated that the address should be sent. The documentation's only example uses just city for address, and obviously we can't get them to answer a question about it. It probably doesn't matter but still I would like to finish this and possibly not go back and forth with the parameters again.

mjfrey (approved):

PR #1167, opened 2017-02-17
Fixed checkout acceptance tests

clintonb:
ECOM-7099


clintonb:
@edx/ecommerce please review

PR #1170, opened 2017-02-23
Option to activate/deactivate enrollment codes

vkaracic:
This is a different version of https://github.com/edx/ecommerce/pull/1151. Here ``Product.expires`` was used to indicate if an enrollment code is active or inactive.

@clintonb is this what you suggested in that other PR? Another thing that needs to be addressed here is when the verification is closed for a seat, i.e. the seat product is set to be expired, the accompanying enrollment code should be disabled as well.

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -4,8 +4,9 @@ import logging
 from django.conf import settings
 from django.db import models, transaction
 from django.db.models import Q, Count
+from django.utils.timezone import now, timedelta
 from django.utils.translation import ugettext_lazy as _
-from oscar.core.loading import get_model
+from oscar.core.loading import get_class, get_model
 from simple_history.models import HistoricalRecords
 import waffle
 
@@ -24,6 +25,7 @@ Partner = get_model('partner', 'Partner')
 Product = get_model('catalogue', 'Product')
 ProductCategory = get_model('catalogue', 'ProductCategory')
 ProductClass = get_model('catalogue', 'ProductClass')
+Selector = get_class('partner.strategy', 'Selector')
 StockRecord = get_model('partner', 'StockRecord')
 
 
@@ -108,6 +110,16 @@ class Course(models.Model):
         """ Returns a queryset of course seat Products related to this course. """
         return self.parent_seat_product.children.all().prefetch_related('stockrecords')
 
+    @property
+    def enrollment_code_product(self):
+        """Returns this course's enrollment code if it exists and is active."""
+        enrollment_code = self.get_enrollment_code()
+        if enrollment_code:
+            info = Selector().strategy().fetch_for_product(enrollment_code)
+            if info.availability.is_available_to_buy:
+                return enrollment_code
+        return None
+
     def get_course_seat_name(self, certificate_type, id_verification_required):
         """ Returns the name for a course seat. """
         name = u'Seat in {}'.format(self.name)
@@ -143,7 +155,7 @@ class Course(models.Model):
 
         Optional arguments:
             credit_provider(str): Name of the organization that provides the credit
-            expires(Datetime): Date when the seat type expires.
+            expires(datetime): Date when the seat type expires.
             credit_hours(int): Number of credit hours provided.
             remove_stale_modes(bool): Remove stale modes.
             create_enrollment_code(bool): Whether an enrollment code is created in additon to the seat.
@@ -215,7 +227,7 @@ class Course(models.Model):
         if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and \
                 certificate_type in ENROLLMENT_CODE_SEAT_TYPES and \
                 create_enrollment_code:
-            self._create_or_update_enrollment_code(certificate_type, id_verification_required, partner, price)
+            self._create_or_update_enrollment_code(certificate_type, id_verification_required, partner, price, expires)
 
         if credit_provider:
             seat.attr.credit_provider = credit_provider
@@ -261,8 +273,7 @@ class Course(models.Model):
 
         return seat
 
-    @property
-    def enrollment_code_product(self):
+    def get_enrollment_code(self):
         """ Returns an enrollment code Product related to this course. """
         try:
             # Current use cases dictate that only one enrollment code product exists for a given course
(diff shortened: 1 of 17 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1170&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/courses/models.py
```diff
@@ -107,6 +108,15 @@ def seat_products(self):
         """ Returns a queryset of course seat Products related to this course. """
         return self.parent_seat_product.children.all().prefetch_related('stockrecords')
 
+    @property
+    def enrollment_code_product(self):
+        """Returns this course's enrollment code if it exists and is active."""
+        enrollment_code = self.get_enrollment_code()
+        if enrollment_code:
+            if enrollment_code.expires is None or (enrollment_code.expires and enrollment_code.expires > now()):
```
clintonb: Use the `Selector` and strategy. See https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/serializers.py#L192-L194.

ecommerce/courses/models.py
```diff
@@ -281,12 +290,14 @@ def _create_or_update_enrollment_code(self, seat_type, id_verification_required,
             seat_type (str): Seat type.
             partner (Partner): Seat provider set in the stock record.
             price (Decimal): Price of the seat.
+            expires (Datetime): Date when the enrollment code expires.
```
clintonb: `datetime` (uncapitalized)

clintonb (commented):
Yes, this is more inline with what I was thinking.

vkaracic:
In that case I believe this is ready for reviews.

marjev (approved):

ecommerce/templates/courses/course_app.html
```diff
@@ -62,6 +62,9 @@
 
 {% block content %}
     <div id="app" class="container" data-credit-providers="{{ credit_providers }}"></div>
+    <script>
+        window.bulkEnrollmentCodesEnabled = {% if bulk_enrollment_codes_enabled %}true{% else %}false{% endif %};
```
iivic, line 66: Why did you include this in the template?

iivic (approved):
LGTM. Just one question.

ecommerce/templates/courses/course_app.html
vkaracic: To know if bulk enrollment codes are enabled or disabled.

vkaracic:
Guys, can you please review the last commit too?

vkaracic:
@clintonb regarding the last commit here, you've said that you change the seat's expires date to disable their purchases, I've updated the ``Product.save`` method to deactivate the seat's enrollment code as well in that case. Other than that case, how often and how are seat expires dates changed? So that I can account for any additional cases as well.

clintonb:
I'm on vacation. See the Course Administration Tool, and its code, for more information on how we handle this with courses.

Clinton Blackburn




On Tue, Feb 28, 2017 at 6:37 AM -0500, "Vedran Karačić" <notifications@github.com> wrote:












@clintonb regarding the last commit here, you've said that you change the seat's expires date to disable their purchases, I've updated the Product.save method to deactivate the seat's enrollment code as well in that case. Other than that case, how often and how are seat expires dates changed? So that I can account for any additional cases as well.

ecommerce/extensions/catalogue/models.py
```diff
@@ -29,6 +31,23 @@ def is_enrollment_code_product(self):
...
+    def __init__(self, *args, **kwargs):
+        super(Product, self).__init__(*args, **kwargs)  # pylint: disable=bad-super-call
+        self.__original_expires = self.expires
+
+    def save(self, *args, **kwargs):
+        """ Overriding the default save method.
+        If the seat expires field is updated, the accompanying enrollment code's expires
+        field needs to be updated too.
+        """
+        if self.get_product_class().name == SEAT_PRODUCT_CLASS_NAME and (self.expires != self.__original_expires):
```
marjev: `if self.is_seat_product`

ecommerce/extensions/catalogue/models.py
```diff
@@ -29,6 +31,23 @@ def is_enrollment_code_product(self):
     def is_coupon_product(self):
         return self.get_product_class().name == COUPON_PRODUCT_CLASS_NAME
 
+    def __init__(self, *args, **kwargs):
+        super(Product, self).__init__(*args, **kwargs)  # pylint: disable=bad-super-call
+        self.__original_expires = self.expires
```
marjev: Can we do this using `post_init` signal?
Code example: https://github.com/edx/edx-platform/blob/master/cms/djangoapps/course_creators/models.py#L52

ecommerce/extensions/catalogue/models.py
```diff
@@ -29,6 +31,23 @@ def is_enrollment_code_product(self):
...
+        If the seat expires field is updated, the accompanying enrollment code's expires
+        field needs to be updated too.
+        """
+        if self.get_product_class().name == SEAT_PRODUCT_CLASS_NAME and (self.expires != self.__original_expires):
+            enrollment_code = self.course.get_enrollment_code()
+            if enrollment_code:
+                enrollment_code.expires = self.expires
+                enrollment_code.save()
+        super(Product, self).save(*args, **kwargs)  # pylint: disable=bad-super-call
+        self.__original_expires = self.expires
```
marjev: This looks to be the same thing that's done `post_init` so perhaps `post_save` should point to the same callback, if you decide to use signals.

ecommerce/extensions/catalogue/models.py
```diff
@@ -29,6 +31,23 @@ def is_enrollment_code_product(self):
...
+    def save(self, *args, **kwargs):
+        """ Overriding the default save method.
+        If the seat expires field is updated, the accompanying enrollment code's expires
+        field needs to be updated too.
+        """
+        if self.get_product_class().name == SEAT_PRODUCT_CLASS_NAME and (self.expires != self.__original_expires):
+            enrollment_code = self.course.get_enrollment_code()
+            if enrollment_code:
+                enrollment_code.expires = self.expires
+                enrollment_code.save()
```
marjev: Again, if you decide to use Django signals, this can be done in the `pre_save` callback.

marjev (changes requested):

vkaracic:
@marjev I've implemented signals, see if that's what you were thinking.

marjev:
If it works, yeah, that's it. LGTM.

marjev (approved):

iivic (approved):

PR #1171, opened 2017-02-23
Django 1.9 fixes

clintonb:
ECOM-7096

### DIFF
--- a/ecommerce/extensions/test/factories.py
+++ b/ecommerce/extensions/test/factories.py
@@ -1,7 +1,12 @@
 from datetime import datetime
 
+import factory
 from django.utils.timezone import now
 from oscar.test.factories import *  # pylint:disable=wildcard-import,unused-wildcard-import
+from oscar.test.factories import (
+    ConditionalOfferFactory as BaseConditionalOfferFactory,
+    VoucherFactory as BaseVoucherFactory,
+)
 
 Benefit = get_model('offer', 'Benefit')
 Catalog = get_model('catalogue', 'Catalog')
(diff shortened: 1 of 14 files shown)


ecommerce/extensions/test/factories.py
```diff
@@ -1,7 +1,12 @@
 from datetime import datetime
 
+import factory
 from django.utils.timezone import now
 from oscar.test.factories import *  # pylint:disable=wildcard-import,unused-wildcard-import
+from oscar.test.factories import (
```
davec-edx, line 6: Please add a comment about this test module overloading those two classes to avoid later confusion.

clintonb:
@edx/ecommerce please review
FYI @zubair-arbi @vkaracic @marjev @iivic

ecommerce/extensions/test/factories.py
clintonb: Where is the confusion? What information should the comment relay?

davec-edx: You just had an import of \_\_all\_\_, then are re-importing a couple classes with the 'as' clause.  Then in the code referencing the original names.  I wondered why, until at the bottom I saw that you had overloaded the methods.  It was not obvious at the point it was being used so I'm merely asking for a comment along the lines of:

`# These classes are being overloaded but preserving the originals for testing`

clintonb: That comment is not quite accurate. We aren't preserving the originals at all. They are completely overwritten, although inherited.

davec-edx: So, see the confusion?  You are bringing in the original class and renaming it, then the overloading classes inherit from it.  Yes.  Bottom line is I found it confusing as a reader.

clintonb: I don't see the confusion, perhaps because I'm too close to the code and we have similar imports throughout the repository. I appreciate the feedback, but I believe the momentary confusion would be resolved almost immediately by looking further down the file.

PR #1173, opened 2017-02-24
Updated e2e tests

clintonb:
- Using JWT access token
- Fixed issue with state field on checkout form

ECOM-7099

### DIFF
--- a/e2e/mixins.py
+++ b/e2e/mixins.py
@@ -2,6 +2,7 @@ import logging
 import time
 import uuid
 
+import datetime
 import requests
 from edx_rest_api_client.client import EdxRestApiClient
 from selenium.common.exceptions import TimeoutException
@@ -155,7 +156,7 @@ class EcommerceApiMixin(object):
     @property
     def ecommerce_api_client(self):
         access_token, __ = get_access_token()
-        return EdxRestApiClient(ECOMMERCE_API_URL, oauth_access_token=access_token)
+        return EdxRestApiClient(ECOMMERCE_API_URL, jwt=access_token)
 
     def assert_order_created_and_completed(self):
         orders = self.ecommerce_api_client.orders.get()['results']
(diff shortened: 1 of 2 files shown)


clintonb:
FYI @edx/ecommerce

e2e/mixins.py
```diff
@@ -297,7 +284,29 @@ def checkout_with_credit_card(self, address):
             'card-cvn': '123'
         }
 
+        country = address['country']
+        state = address['state'] or ''
+
+        select_fields = (
+            ('#id_country', country),
+            ('#card-expiry-month', '12'),
+            ('#card-expiry-year', '2030')
```
rlucioni: I know it'll be 13 years until this is a problem, but I feel compelled to say something anyway. Since you're in here. you can future-proof it with `datetime.datetime.now().year + 3`.

e2e/mixins.py
```diff
@@ -297,7 +284,29 @@ def checkout_with_credit_card(self, address):
...
+        )
+
+        if country in ('US', 'CA',):
+            select_fields += (('#id_state', state,),)
```
rlucioni: Since you need to mutate it, can you just make `select_fields` a list? Then you can simply `.append()`.

e2e/mixins.py
```diff
@@ -297,7 +284,29 @@ def checkout_with_credit_card(self, address):
...
+
+        # Select the appropriate <option> elements
+        for selector, value in select_fields:
+            if value:
+                select = Select(self.browser.find_element_by_css_selector(selector))
+                select.select_by_value(value)
+
+        # Fill in the text fields
         for field, value in billing_information.items():
+            print(field, value)
```
rlucioni: Leftover from debugging?

rlucioni (approved):

PR #1177, opened 2017-03-01
ENT-231 Course seat types working correctly for Catalog coupon

asadiqbal08:
[ENT-231](https://openedx.atlassian.net/browse/ENT-231)

### DIFF
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -263,20 +263,23 @@ class Range(AbstractRange):
         """
         Assert if the range contains the product.
         """
-        if self.course_catalog:
-            request = get_current_request()
-            try:
-                course_catalog = get_course_catalogs(site=request.site, resource_id=self.course_catalog)
-            except (ConnectionError, SlumberBaseException, Timeout):
-                raise Exception(
-                    'Unable to connect to Course Catalog service for catalog with id [%s].' % self.course_catalog
-                )
+        # course_catalog is associated with course_seat_types.
+        if self.course_catalog and self.course_seat_types:
+            # Product certificate type should belongs to range seat types.
+            if product.attr.certificate_type.lower() in self.course_seat_types:  # pylint: disable=unsupported-membership-test
+                request = get_current_request()
+                try:
+                    course_catalog = get_course_catalogs(site=request.site, resource_id=self.course_catalog)
+                except (ConnectionError, SlumberBaseException, Timeout):
+                    raise Exception(
+                        'Unable to connect to Course Catalog service for catalog with id [%s].' % self.course_catalog
+                    )
 
-            response = self.run_catalog_query(product, course_catalog.get('query'))
-            # Range can have a catalog query and 'regular' products in it,
-            # therefor an OR is used to check for both possibilities.
-            return ((response['course_runs'][product.course_id]) or
-                    super(Range, self).contains_product(product))  # pylint: disable=bad-super-call
+                response = self.run_catalog_query(product, course_catalog.get('query'))
+                # Range can have a catalog query and 'regular' products in it,
+                # therefor an OR is used to check for both possibilities.
+                return ((response['course_runs'][product.course_id]) or
+                        super(Range, self).contains_product(product))  # pylint: disable=bad-super-call
         elif self.catalog_query and self.course_seat_types:
             if product.attr.certificate_type.lower() in self.course_seat_types:  # pylint: disable=unsupported-membership-test
                 response = self.run_catalog_query(product)
--- a/ecommerce/extensions/offer/tests/test_models.py
+++ b/ecommerce/extensions/offer/tests/test_models.py
@@ -187,6 +187,37 @@ class RangeTests(CouponMixin, CourseCatalogServiceMockMixin, CourseCatalogTestMi
 
     @httpretty.activate
     @mock_course_catalog_api_client
+    @ddt.data(
+        (1, 'verified,professional', True),
+        (1, 'verified', True),
+        (2, 'professional', False),
+    )
+    @ddt.unpack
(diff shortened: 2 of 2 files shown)


asadiqbal08:
@mattdrayer FYI here. Build is failing with unknown reason. 
@zubair-arbi @saleem-latif  can you guys please review the change set ?

mattdrayer:
@asadiqbal08 I have restarted the Travis CI build -- let's see what happens

ecommerce/extensions/offer/models.py
```diff
@@ -263,20 +263,21 @@ def contains_product(self, product):
...
-                raise Exception(
-                    'Unable to connect to Course Catalog service for catalog with id [%s].' % self.course_catalog
-                )
-
-            response = self.run_catalog_query(product, course_catalog.get('query'))
-            # Range can have a catalog query and 'regular' products in it,
-            # therefor an OR is used to check for both possibilities.
-            return ((response['course_runs'][product.course_id]) or
-                    super(Range, self).contains_product(product))  # pylint: disable=bad-super-call
+        if self.course_catalog and self.course_seat_types:
```
mattdrayer: It would be helpful if we could describe either in the docstring or in a code comment the use case being covered by this block.

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -187,6 +187,36 @@ def test_course_catalog_query_range_contains_product_for_failure(self):
 
     @httpretty.activate
     @mock_course_catalog_api_client
+    @ddt.data(
+        {'course_catalog': 1, 'course_seat_types': 'verified,professional'},
+        {'course_catalog': 1, 'course_seat_types': 'verified'},
+        {'course_catalog': 2, 'course_seat_types': 'professional'},
+    )
+    def test_course_catalog_range_contains_product(self, data):
```
zubair-arbi: IMHO we should pass unpacked data along with expected response `is_product_in_range` instead of manually getting the values from a combined dict `data`:
```python
@ddt.data(
    (1, 'verified', True),
    (1, 'verified,professional', True),
    (2, 'professional', False),
)
@ddt.unpack
def test_course_catalog_range_contains_product(self, course_catalog, course_seat_types, is_product_in_range):
```

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -187,6 +187,36 @@ def test_course_catalog_query_range_contains_product_for_failure(self):
...
+        self.mock_dynamic_catalog_contains_api(query=catalog_query, course_run_ids=[course.id])
+
+        course_catalog_id = 1
+        self.mock_course_discovery_api_for_catalog_by_resource_id(
+            catalog_id=course_catalog_id, catalog_query=catalog_query
+        )
+        range = Range.objects.create(**data)
+        response = range.contains_product(seat)
+
+        #course_seat_types 'professional' does not exist.
```
zubair-arbi: you can remove this comment and the `if`, `else` conditions after you have used unpacked data with expected response `is_product_in_range`

asadiqbal08: @zubair-arbi Actually, I was passing the `**data` for creating Range object `Range.objects.create(**data)` that's why I chosen the `if` condition

zubair-arbi: I just wanted to point out that in tests we normally avoid if/else clauses, if we cannot avoid that then we should consider creating separate tests for each clauses.

asadiqbal08:
@zubair-arbi kindly take another look

saleem-latif:
@asadiqbal08 there is a quality failure in `ecommerce/extensions/offer/models.py` at line 269,

```
Module ecommerce.extensions.offer.models
E:269,56: Value 'self.course_seat_types' doesn't support membership test (unsupported-membership-test)
```

saleem-latif (approved):
👍  Once quality Failure is fixed,

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -187,6 +187,40 @@ def test_course_catalog_query_range_contains_product_for_failure(self):
...
+            catalog_id=course_catalog_id, catalog_query=catalog_query
+        )
+
+        self.range.course_seat_types = course_seat_types
+        self.range.course_catalog = course_catalog
+        self.range.save()
+
+        response = self.range.contains_product(seat)
+
+        if is_product_in_range:
```
zubair-arbi: We can replace this if/else with
```
self.assertEqual(is_product_in_range, response)
```

ecommerce/extensions/offer/models.py
```diff
@@ -263,20 +263,23 @@ def contains_product(self, product):
...
-                )
-
-            response = self.run_catalog_query(product, course_catalog.get('query'))
-            # Range can have a catalog query and 'regular' products in it,
-            # therefor an OR is used to check for both possibilities.
-            return ((response['course_runs'][product.course_id]) or
-                    super(Range, self).contains_product(product))  # pylint: disable=bad-super-call
+        # course_catalog is associated with course_seat_types.
+        if self.course_catalog and self.course_seat_types:
+            # Entitlement should be applied to those products whom type belongs to range seat types.
```
zubair-arbi: My opinion would be to remove/update the `Entitlement` part from this comment, since this is a generic Range method.

zubair-arbi (approved):
Changes look good now 👍

asadiqbal08:
@mattdrayer Zeshan has tested the changes on the sandbox. As per him, working as expected. Share your final thoughts on the PR.

mattdrayer:
Great -- thanks very much for adding the comments -- glad to see testing went well -- :+1:

PR #1178, opened 2017-03-01
ECOM-7349 Learners cannot get enrollment status on Ecommerce SingleBasketItemView

Ayub-Khan:
**ECOM-7349** 
Learners cannot get enrollment status on Ecommerce SingleBasketItemView because of user oauth token expiration. 
Updated the Oauth access token for enrolment check on basket page from learner Oauth token to Service User Oauth token.

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -17,7 +17,7 @@ from django.utils.translation import ugettext_lazy as _
 from edx_rest_api_client.client import EdxRestApiClient
 from jsonfield.fields import JSONField
 from requests.exceptions import ConnectionError, Timeout
-from slumber.exceptions import HttpNotFoundError, SlumberBaseException
+from slumber.exceptions import HttpNotFoundError, SlumberBaseException, HttpClientError
 
 from ecommerce.core.exceptions import VerificationStatusError
 from ecommerce.core.url_utils import get_lms_url
@@ -434,11 +434,11 @@ class User(AbstractUser):
         try:
             api = EdxRestApiClient(
                 request.site.siteconfiguration.build_lms_url('/api/enrollment/v1'),
-                oauth_access_token=self.access_token,
+                jwt=request.site.siteconfiguration.access_token,
                 append_slash=False
             )
             status = api.enrollment(','.join([self.username, course_key])).get()
-        except (ConnectionError, SlumberBaseException, Timeout):
+        except (ConnectionError, SlumberBaseException, Timeout, HttpClientError):
             log.exception(
                 'Failed to retrieve enrollment details for [%s] in course [%s]',
                 self.username,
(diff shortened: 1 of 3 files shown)


Ayub-Khan:
jenkins run bokchoy

ecommerce/core/models.py
```diff
@@ -434,11 +434,11 @@ def is_user_already_enrolled(self, request, seat):
         try:
             api = EdxRestApiClient(
                 request.site.siteconfiguration.build_lms_url('/api/enrollment/v1'),
-                oauth_access_token=self.access_token,
+                oauth_access_token=request.site.siteconfiguration.access_token,
```
mjfrey: While we are at it, we should probably switch this to use JWT.

mjfrey (changes requested):

ecommerce/core/models.py
Ayub-Khan: @mjfrey What about changing in whole class and removing access_token [property](https://github.com/edx/ecommerce/blob/74a41b365adf7b7385d09c8bf325c5b87226d369/ecommerce/core/models.py#L401)?

mjfrey: yes.

mjfrey: actually you might not want to change the property on the User model, but rather just change this method.  Not sure what else relies on getting the User token.

schenedx (changes requested):
I believe I wrote in the ticket, we should just switch the auth to use JWT

Ayub-Khan:
@schenedx so we should not catch the possible exception or change the other calls and let them fail?

schenedx:
@Ayub-Khan That's not what I meant. Your change on catching exception is good.
I meant we should use JWT instead of using oauth token. Rest you should update as you see fit. Sorry about the confusion.

ecommerce/core/models.py
Ayub-Khan: @mjfrey  already changed it .

Ayub-Khan:
@schenedx still confused. am i doing it the wrong way ?. i used siteconfiguration's jwt tokken .

ecommerce/core/models.py
mjfrey: @Ayub-Khan to use JWT you need.
jwt=request.site.siteconfiguration.access_token
that is still using oauth.

Ayub-Khan:
@mjfrey @schenedx sorry my bad updated. should i remove other changes or keep them.

schenedx:
@Ayub-Khan I think you can have other changes to stay. However, I am wondering how you tested this PR? Where you able to run these changes through a sandbox? Or your local machine? And ensure the user sees the basket page no problem?

schenedx (approved):
👍 Code change here LGTM.
Please make sure it is properly tested

mjfrey (approved):

schenedx:
@Ayub-Khan We are merging this PR and let it go to stage. We will test this in stage. The sandbox was too difficult to figure out.

schenedx:
@Ayub-Khan This PR did not solve the problem in stage. It may have made the issue worse in stage. I will create a PR to revert both this one and the commit in https://github.com/edx/ecommerce/pull/1137

PR #1181, opened 2017-03-02
Revert to correct the Learners getting errors on SingleBasketItemView

schenedx:
@rlucioni @mjfrey Please review

Revert "[SOL-2083] Check if user is enrolled in course only if the request arrives from the Offer Landing Page" and "ECOM-7349 Learners cannot get enrollment status on Ecommerce SingleBasketItemView"

### DIFF
--- a/docs/additional_features/change_processors.rst
+++ b/docs/additional_features/change_processors.rst
@@ -1,0 +1,33 @@
+.. _Changing Payment Processors:
+
+#############################
+Changing Payment Processors
+#############################
+
+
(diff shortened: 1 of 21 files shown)


rlucioni (commented):
@schenedx this doesn't appear to revert the problematic change from https://github.com/edx/ecommerce/pull/1137.

schenedx:
@rlucioni Good catch. Added the missing revert

rlucioni (approved):

mjfrey (approved):

PR #1186, opened 2017-03-02
Moved docs from RTD to ecommerce repo

davec-edx:
Moving docs from RTD over to repo per: 

ECOM-7190

@rlucioni @schenedx


schenedx (approved):
👍

rlucioni (approved):

PR #1187, opened 2017-03-03
Synchronous order fulfillment for free baskets.

vkaracic:
It doesn't make sense to keep a user waiting for an async process to finish just so the result from that process can be used to determine where to send the user. That whole use case is synchronous vulgaris and should be executed as such.
Not to mention that the _enrollment code applied > checkout > receipt page_ flow would work with this as opposed to the [last few tries](https://github.com/edx/ecommerce/pull/1149).


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1187"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb (changes requested):
We added asynchronous fulfillment to avoid tying up all of our gunicorn workers on LMS with two requests for the same enrollment call.

What problem is this change trying to resolve?

This will need to be load-tested (before and after this change) to ensure this does not lead to performance degradation.

mjfrey (changes requested):
I do not think we should do this.  During large coupon based campaigns this will be a problem bypassing the async order fulfillment.  We need to figure out why the receipt page approach was not working.

vkaracic:
@clintonb the problem it's trying to solve is redirecting a user to a receipt page if the order fulfillment was successful or the checkout error page if it wasn't. The previous tries (https://github.com/edx/ecommerce/pull/1149) don't work on stage, they all timeout (event the last 10 seecond timeout).

clintonb:
If there is an issue creating the order, go to the checkout error page. After an order has been created, go to the receipt page. If the order has not been fulfilled, display a message to the user and poll the system until the order is fulfilled or there is an error. If there is a fulfillment error, display a message to that effect.

PR #1189, opened 2017-03-07
ENT-213 Consent Declined message on Offers page

bdero:
This change adds:

- A generic info message that can be used by passing an `info` dict to the context of any templates extending `edx/base.html`.
- A `failure_url` callback as a parameter to the consent URL.
- A consent failure message informing the user that they weren't enrolled in the course. It shows up when `consent_failure={SKU}` is passed to the coupon offer view.

**JIRA tickets**: Implements [ENT-213](https://openedx.atlassian.net/browse/ENT-213).

**Dependencies**:
This PR depends on this `edx-enterprise` PR: https://github.com/edx/edx-enterprise/pull/66

This change will come into effect with an Enterprise PR that @haikuginger will be making for ENT-185/ENT-212, where he will be including the `failure_url` call. Everything in this PR can be tested without it, though.

**Screenshots**:

*Desktop:*
desktop screenshot

*Mobile:*
mobile screenshot

**Sandbox URL**: TBD.

**Testing instructions**:

Initial setup:
1. Setup a devstack.
1. Checkout this branch (open-craft:bdero/ent-213) for ecommerce.
1. Override the edx-enterprise version with the [correct branch](https://github.com/edx/edx-enterprise/pull/66):
   ```
   /edx/bin/pip.edxapp install -e git+https://github.com/open-craft/edx-enterprise.git@bdero/add-contact-email#egg=edx-enterprise==add-contact-email
   /edx/bin/manage.edxapp lms migrate --settings=devstack
   ```
1. Connect [ecommerce to the LMS](https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/ecommerce/install_ecommerce.html#configure-edx-openid-connect-oidc).
1. Setup the [course discovery](https://open-edx-course-catalog.readthedocs.io/en/latest/getting_started.html) service and [connect it to the LMS](https://open-edx-course-catalog.readthedocs.io/en/latest/oidc.html).
1. Create an Enterprise Customer in the LMS: http://localhost:8000/admin/enterprise/enterprisecustomer/add/
    1. Give it any name.
    1. Check the **Enable data sharing consent** box.
    1. For **Enforce data sharing consent**, select **At Enrollment**.
    1. Leave the **Contact email** field blank for now.
1. Add a new Course in Ecommerce: http://localhost:8002/courses/new/
    1. For the course ID, use "course-v1:edX+DemoX+Demo_Course".
    1. Name it "edX Demonstration Course".
    1. Choose "Professional Education" for the Course Type.
1. Create a new Coupon in Ecommerce: http://localhost:8002/coupons/new
    1. Give it any Coupon Name.
    1. Under Code Type choose "Enrollment Code".
    1. Give it a valid date range that includes today.
    1. For Client, type any name. E.g. "Some Client".
    1. Select the "Single course" radio button.
    1. For the Course ID, use "course-v1:edX+DemoX+Demo_Course"
    1. The only available Seat Type should be automatically populated: "Professional"
    1. For the Enterprise Customer, select the name of the customer created in step 6.

Testing the PR:
1. Navigate to the [coupon list](http://localhost:8002/coupons/) and click "Download Coupon Report" to retrieve a CSV containing the coupon URL.
    coupon report button screenshot
1. The coupon report CSV should contain an offer URL that looks like this:
    ```
    http://localhost:8002/coupons/offer/?code=EGGZYLXDKNZK6W7C
    ```
    Navigate to that URL.
1. Click on the "Enroll Now" button.
    offer landing page screenshot
    You should be redirected to the course-specific Data Sharing Consent screen.
    data sharing consent screenshot
1. Inspect the URL of the page to find that there's a `failure_url` parameter that looks something like this: `&failure_url=http%3A%2F%2Flocalhost%3A8002%2Fcoupons%2Foffer%2F%3Fcode%3DHV7SMPXDYJFTGRU2%26consent_failed%3DBB3ECA9`
1. URL decode the parameter.
    For example:
    ```
    python2 -c 'import urllib; print urllib.unquote_plus("http%3A%2F%2Flocalhost%3A8002%2Fcoupons%2Foffer%2F%3Fcode%3DHV7SMPXDYJFTGRU2%26consent_failed%3DBB3ECA9")'
    ```
    Which yields: http://localhost:8002/coupons/offer/?code=HV7SMPXDYJFTGRU2&consent_failed=BB3ECA9
1. Navigate to your decoded URL.
1. You should arrive back on the offers page. Observe that there is a blue info message area that looks similar to this:
    info message screenshot
1. Navigate back to the Enterprise Customer list in the LMS and edit the Enterprise Customer you created during the initial setup: http://localhost:8000/admin/enterprise/enterprisecustomer/
    1. Fill in the **Contact email** field with "contact@example.com".
1. Repeat steps 3 through 6.
1. Observe that the content failure message now contains the contact email:
   screenshot with contact email

**Author notes and concerns**:

The image attached in the [original ticket](https://openedx.atlassian.net/browse/ENT-213) wants to show "custom contact info". We're not currently storing this information anywhere AFAICT, so I've excluded it from this PR for the time being. CC @bradenmacdonald @haikuginger 

**Reviewers**
- [ ] @pomegranited 
- [ ] edX reviewer[s] TBD

### DIFF
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -1,6 +1,7 @@
 import datetime
 import urllib
 
+import ddt
 import httpretty
 import pytz
 from django.conf import settings
@@ -183,9 +184,12 @@ class GetVoucherTests(CourseCatalogTestMixin, TestCase):
         self.assert_error_messages(voucher, product, user, error_msg)
 
 
+@ddt.ddt
 @httpretty.activate
-class CouponOfferViewTests(ApiMockMixin, CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin, TestCase):
+class CouponOfferViewTests(ApiMockMixin, CouponMixin, CourseCatalogTestMixin, EnterpriseServiceMockMixin,
+                           LmsApiMockMixin, TestCase):
     path = reverse('coupons:offer')
+    credit_seat = None
 
     def setUp(self):
         super(CouponOfferViewTests, self).setUp()
@@ -238,17 +242,25 @@ class CouponOfferViewTests(ApiMockMixin, CouponMixin, CourseCatalogTestMixin, Lm
         response = self.client.get(url)
         self.assertEqual(response.context['error'], 'The voucher is not applicable to your current basket.')
 
-    def prepare_url_for_credit_seat(self, code='CREDIT'):
+    def prepare_url_for_credit_seat(self, code='CREDIT', enterprise_customer=None):
         """Helper method for creating a credit seat and construct the URL to its offer landing page.
 
         Returns:
             URL to its offer landing page.
         """
         __, credit_seat = self.create_course_and_seat(seat_type='credit')
+        self.credit_seat = credit_seat
         # Make sure to always pair `course_seat_types` and `catalog_query` parameters because
         # if one of them is missing it could result in a SEGFAULT error when running tests
         # with migrations enabled.
-        _range = RangeFactory(products=[credit_seat, ], course_seat_types='credit', catalog_query='*:*')
+        range_kwargs = {
+            'products': [credit_seat, ],
+            'course_seat_types': 'credit',
+            'catalog_query': '*:*',
+        }
+        if enterprise_customer:
+            range_kwargs['enterprise_customer'] = enterprise_customer
+        _range = RangeFactory(**range_kwargs)
         prepare_voucher(code=code, _range=_range)
 
         url = '{path}?code={code}'.format(path=self.path, code=code)
@@ -269,6 +281,50 @@ class CouponOfferViewTests(ApiMockMixin, CouponMixin, CourseCatalogTestMixin, Lm
         url = self.prepare_url_for_credit_seat()
         response = self.client.get(url)
         self.assertEqual(response.status_code, 200)
+
+    def test_consent_failed_invalid(self):
+        """ Verify that an error is returned if the consent_failed parameter is not a valid SKU. """
+        url = '{}&consent_failed={}'.format(self.prepare_url_for_credit_seat(), 'INVALID')
+        response = self.client.get(url)
+        self.assertEqual(response.context['error'], 'SKU INVALID does not exist.')
+
+    def test_consent_failed_no_enterprise_customer(self):
+        """ Verify that an error is returned if the voucher has no associated EnterpriseCustomer. """
+        base_url = self.prepare_url_for_credit_seat(enterprise_customer=None)
+        sku = self.credit_seat.stockrecords.first().partner_sku
+        url = '{}&consent_failed={}'.format(base_url, sku)
+        response = self.client.get(url)
+        self.assertEqual(
+            response.context['error'],
+            'There is no Enterprise Customer associated with SKU {sku}.'.format(sku=sku)
+        )
+
+    @ddt.data(
+        ('', 'If you have concerns about sharing your data, please contact your administrator at TestShib.'),
+        (
+            'contact@example.com',
+            'If you have concerns about sharing your data, please contact your administrator at TestShib at '
+            'contact@example.com.',
+        ),
+    )
+    @ddt.unpack
+    def test_consent_failed_message(self, contact_email, expected_response):
+        """ Verify that the consent failure message shows up when the consent_failed parameter is a valid SKU. """
+        self.mock_specific_enterprise_customer_api(
+            ENTERPRISE_CUSTOMER,
+            name='TestShib',
+            contact_email=contact_email
+        )
+        base_url = self.prepare_url_for_credit_seat(enterprise_customer=ENTERPRISE_CUSTOMER)
+        sku = self.credit_seat.stockrecords.first().partner_sku
+        url = '{}&consent_failed={}'.format(base_url, sku)
+        response = self.client.get(url)
+        self.assertContains(
+            response,
+            'Enrollment in {course_name} was not complete.'.format(course_name=self.credit_seat.course.name),
+            status_code=200
+        )
+        self.assertContains(response, expected_response, status_code=200)
 
 
 class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin, EnterpriseServiceMockMixin,
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -22,12 +22,14 @@ from ecommerce.extensions.api import exceptions
 from ecommerce.enterprise.exceptions import EnterpriseDoesNotExist
 from ecommerce.enterprise.utils import (
     get_enterprise_course_consent_url,
+    get_enterprise_customer_consent_failed_context_data,
     get_enterprise_customer_data_sharing_consent_token,
     get_enterprise_customer_from_voucher,
 )
 from ecommerce.extensions.basket.utils import prepare_basket
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
 from ecommerce.extensions.voucher.utils import get_voucher_and_products_from_code
+
 
 Applicator = get_class('offer.utils', 'Applicator')
 Basket = get_model('basket', 'Basket')
@@ -100,20 +102,26 @@ class CouponOfferView(TemplateView):
 
     def get_context_data(self, **kwargs):
         code = self.request.GET.get('code', None)
-        if code is not None:
-            try:
-                voucher, products = get_voucher_and_products_from_code(code=code)
-            except Voucher.DoesNotExist:
-                return {'error': _('Coupon does not exist')}
-            except exceptions.ProductNotFoundError:
-                return {'error': _('The voucher is not applicable to your current basket.')}
-            valid_voucher, msg = voucher_is_valid(voucher, products, self.request)
-            if valid_voucher:
-                self.template_name = 'coupons/offer.html'
-                return
+        if code is None:
+            return {'error': _('This coupon code is invalid.')}
 
+        try:
+            voucher, products = get_voucher_and_products_from_code(code=code)
+        except Voucher.DoesNotExist:
+            return {'error': _('Coupon does not exist')}
+        except exceptions.ProductNotFoundError:
+            return {'error': _('The voucher is not applicable to your current basket.')}
+        valid_voucher, msg = voucher_is_valid(voucher, products, self.request)
+        if not valid_voucher:
             return {'error': msg}
-        return {'error': _('This coupon code is invalid.')}
+
+        context_data = super(CouponOfferView, self).get_context_data(**kwargs)
+        context_data.update(get_enterprise_customer_consent_failed_context_data(self.request, voucher))
+
+        if context_data and 'error' not in context_data:
+            self.template_name = 'coupons/offer.html'
+
+        return context_data
 
     @method_decorator(login_required_for_credit)
     def get(self, request, *args, **kwargs):
--- a/ecommerce/enterprise/utils.py
+++ b/ecommerce/enterprise/utils.py
@@ -1,12 +1,14 @@
 """
 Helper methods for enterprise app.
 """
+from collections import OrderedDict
 import hashlib
 import hmac
 from urllib import urlencode
 
 from django.conf import settings
 from django.core.urlresolvers import reverse
+from django.utils.translation import ugettext as _
 from edx_rest_api_client.client import EdxRestApiClient
 from oscar.core.loading import get_model
 import waffle
@@ -16,6 +18,8 @@ from ecommerce.courses.utils import traverse_pagination
 from ecommerce.enterprise.exceptions import EnterpriseDoesNotExist
 
 ConditionalOffer = get_model('offer', 'ConditionalOffer')
+StockRecord = get_model('partner', 'StockRecord')
+CONSENT_FAILED_PARAM = 'consent_failed'
 
 
 def is_enterprise_feature_enabled():
@@ -54,6 +58,7 @@ def get_enterprise_customer(site, token, uuid):
         'name': response['name'],
         'id': response['uuid'],
         'enable_data_sharing_consent': response['enable_data_sharing_consent'],
+        'contact_email': response.get('contact_email', ''),
     }
 
 
@@ -72,6 +77,56 @@ def get_enterprise_customers(site, token):
         }
         for each in traverse_pagination(response, endpoint)
     ]
+
+
+def get_enterprise_customer_consent_failed_context_data(request, voucher):
+    """
+    Get the template context to display a message informing the user that they were not enrolled in the course
+    due to not consenting to data sharing with the Enterprise Customer.
+
+    If the `consent_failed` GET param is defined and it's not set to a valid SKU, return an error context that
+    says the given SKU doesn't exist.
+    """
+    consent_failed_sku = request.GET.get(CONSENT_FAILED_PARAM)
+    if consent_failed_sku is None:
+        # If no SKU was supplied via the consent failure param, then don't display any messages.
+        return {}
+
+    # The user is redirected to this view with `consent_failed` defined (as the product SKU) when the
+    # user doesn't consent to data sharing.
+    try:
+        product = StockRecord.objects.get(partner_sku=consent_failed_sku).product
+    except StockRecord.DoesNotExist:
+        return {'error': _('SKU {sku} does not exist.').format(sku=consent_failed_sku)}
+
+    # Return the view with an info message informing the user that the enrollment didn't complete.
+    enterprise_customer = get_enterprise_customer_from_voucher(
+        request.site,
+        request.user.access_token,
+        voucher
+    )
+    if not enterprise_customer:
+        return {'error': _('There is no Enterprise Customer associated with SKU {sku}.').format(
+            sku=consent_failed_sku
+        )}
+
+    contact_info = enterprise_customer['contact_email']
+
+    # Use two full messages instead of using a computed string, so that translation utilities will pick up on both
+    # strings as unique.
+    message = _('If you have concerns about sharing your data, please contact your administrator at {enterprise}.')
+    if contact_info:
+        message = _(
+            'If you have concerns about sharing your data, please contact your administrator at {enterprise} at '
+            '{contact_info}.'
+        )
+
+    return {
+        'info': {
+            'title': _('Enrollment in {course_name} was not complete.').format(course_name=product.course.name),
+            'message': message.format(enterprise=enterprise_customer['name'], contact_info=contact_info,)
+        }
+    }
 
 
 def get_or_create_enterprise_customer_user(site, enterprise_customer_uuid, username):
@@ -127,9 +182,12 @@ def get_enterprise_course_consent_url(site, code, sku, consent_token, course_id,
     collecting consent. The URL contains a specially crafted "next" parameter that will result
     in the user being redirected back to the coupon redemption view with the verified consent token.
     """
-    callback_url = '{protocol}://{domain}{resource}?{params}'.format(
+    base_url = '{protocol}://{domain}'.format(
         protocol=settings.PROTOCOL,
         domain=site.domain,
+    )
+    callback_url = '{base}{resource}?{params}'.format(
+        base=base_url,
         resource=reverse('coupons:redeem'),
         params=urlencode({
             'code': code,
@@ -137,11 +195,20 @@ def get_enterprise_course_consent_url(site, code, sku, consent_token, course_id,
             'consent_token': consent_token,
         })
     )
+    failure_url = '{base}{resource}?{params}'.format(
+        base=base_url,
+        resource=reverse('coupons:offer'),
+        params=urlencode(OrderedDict([
+            ('code', code),
+            (CONSENT_FAILED_PARAM, sku),
+        ])),
+    )
     request_params = {
         'course_id': course_id,
         'enterprise_id': enterprise_customer_uuid,
         'enrollment_deferred': True,
         'next': callback_url,
+        'failure_url': failure_url,
     }
     redirect_url = '{base}?{params}'.format(
         base=site.siteconfiguration.enterprise_grant_data_sharing_url,
--- a/ecommerce/static/js/routers/offer_router.js
+++ b/ecommerce/static/js/routers/offer_router.js
@@ -11,8 +11,11 @@ define([
             // Base/root path of the app
             root: '/coupons/',
 
-            routes: {
-                'offer/?code=:code': 'showOfferPage'
+            initialize: function(options) {
+                // This is where views will be rendered
+                this.$el = options.$el;
+
+                this.route(/^offer\/\?code=(\w+)/, 'showOfferPage');
             },
 
             showOfferPage: function(code) {
--- a/ecommerce/templates/edx/partials/_info_message.html
+++ b/ecommerce/templates/edx/partials/_info_message.html
@@ -1,0 +1,11 @@
+{% if info %}
+    <div id="info-message">
+        <div class="container">
+            <i class="fa fa-info-circle" aria-hidden="true"></i>
+            <div class="depth depth message-info-content">
+                <h3>{{ info.title }}</h3>
+                {{ info.message }}
+            </div>
+        </div>
+    </div>
+{% endif %}
--- a/ecommerce/enterprise/tests/mixins.py
+++ b/ecommerce/enterprise/tests/mixins.py
@@ -20,13 +20,13 @@ class EnterpriseServiceMockMixin(object):
         super(EnterpriseServiceMockMixin, self).setUp()
         cache.clear()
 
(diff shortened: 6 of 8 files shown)


openedx-webhooks:
Thanks for the pull request, @bdero! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1189"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/coupons/views.py
```diff
@@ -109,6 +111,38 @@ def get_context_data(self, **kwargs):
                 return {'error': _('The voucher is not applicable to your current basket.')}
             valid_voucher, msg = voucher_is_valid(voucher, products, self.request)
             if valid_voucher:
+                consent_failed = self.request.GET.get('consent_failed', None)
```
pomegranited: `'consent_failed'` is an important magic string.  Can you set a variable somewhere that this function (and `get_enterprise_course_consent_url`) can share?

Up for discussion.

ecommerce/static/js/routers/offer_router.js
```diff
@@ -11,11 +11,15 @@ define([
...
-                'offer/?code=:code': 'showOfferPage'
+            initialize: function(options) {
+                // This is where views will be rendered
+                this.$el = options.$el;
+
+                this.route(/^offer\/\?code=(\w+)/, 'showOfferPage');
             },
 
             showOfferPage: function(code) {
+                console.log(code);
```
pomegranited: nit: remove logging.

ecommerce/coupons/views.py
```diff
@@ -109,6 +111,38 @@ def get_context_data(self, **kwargs):
...
+                if consent_failed is not None:
+                    # The user is redirected to this view with `consent_failed` defined (as the product SKU) when the
+                    # user doesn't consent to data sharing.
+                    try:
+                        product = StockRecord.objects.get(partner_sku=consent_failed).product
+                    except StockRecord.DoesNotExist:
+                        return {'error': _('SKU {sku} does not exist.').format(sku=consent_failed)}
+
+                    self.template_name = 'coupons/offer.html'
```
pomegranited: Line duplicated from original [L146](https://github.com/edx/ecommerce/pull/1189/files#diff-c797ef6707ae3dfb1ee8828592244a9eR146).

ecommerce/coupons/views.py
pomegranited: I know we try to tread lightly through the code we modify, but this method is getting very hard to parse.  Would you consider a slight rewrite, to keep the error messages towards the top?  Also removes the duplicate line noted below.

```python
    def get_context_data(self, **kwargs):
        code = self.request.GET.get('code', None)
        if code is None:
            return {'error': _('This coupon code is invalid.')}

        try:
            voucher, products = get_voucher_and_products_from_code(code=code)
        except Voucher.DoesNotExist:
            return {'error': _('Coupon does not exist')}
        except exceptions.ProductNotFoundError:
            return {'error': _('The voucher is not applicable to your current basket.')}
        valid_voucher, msg = voucher_is_valid(voucher, products, self.request)
        if not valid_voucher:
            return {'error': msg}

        context_data = None
        consent_failed = self.request.GET.get('consent_failed', None)
        if consent_failed is not None:
            # The user is redirected to this view with `consent_failed` defined (as the product SKU) when the
            # user doesn't consent to data sharing.
            try:
                product = StockRecord.objects.get(partner_sku=consent_failed).product
                return {'error': _('SKU {sku} does not exist.').format(sku=consent_failed)}

            # Return the view with an info message informing the user that the enrollment didn't complete
            request = get_current_request()
            enterprise_customer = get_enterprise_customer_from_voucher(
                request.site,
                request.user.access_token,
                voucher
            )
            context_data = {
                'info': {
                    'title': _('Enrollment in {course_name} was not complete.').format(
                        course_name=product.course.name
                    ),
                    'message': _(
                        'If you have concerns about sharing your data, please contact your administrator '
                        'at {enterprise}.'
                    ).format(
                        enterprise=enterprise_customer['name']
                    )
                }
            }

        self.template_name = 'coupons/offer.html'
        return context_data
```

pomegranited (changes requested):
Works as described!  Thank you for the detailed test instructions.

Once my comments are addressed or refuted, I'm happy to approve.

- [x] I tested this on my devstack, using Chrome and the mobile emulator.
- [x] I read through the code, and left some comments.
- [x] I checked for accessibility issues.  The added info bar is a static part of the HTML template, and so screen readers should have no trouble reading it.
- [x] Includes documentation - docstrings.

ecommerce/coupons/views.py
bdero: Alrighty, I've renamed the variable to `consent_failed_sku`, since that's all it is at the end of the day. The reason we need to SKU to be passed for this info message is that we need a way to identify which product (which has the course info) the user denied consent for in the list of offers.

I've kept the name `consent_failed` for the GET param though, since I think it makes sense to say that consent gathering failed for a specified SKU. Does this seem sound?

Let me know if this addresses your concern!

bdero: Totally agree, I'm all for this and added these changes in. Thanks!

ecommerce/coupons/views.py
bdero: All set, thanks!

ecommerce/static/js/routers/offer_router.js
bdero: Removed, thanks!

bdero:
@pomegranited Thank you for the review! I've addressed all comments.

ecommerce/coupons/views.py
bdero: Ah, nevermind, I just got what you were saying here. I added `CONSENT_FAILED_PARAM` to `ecommerce.enterprise.utils`, which is where `get_enterprise_course_consent_url` lives.

pomegranited: +1 Yep that's what I meant.  Sorry for being unclear!

pomegranited (approved):

bdero:
FYI @mattdrayer @asadiqbal08 @saleem-latif, this PR along with https://github.com/edx/edx-enterprise/pull/66 and is ready for your review!

mattdrayer:
LGTM @bdero!  Definitely would like @asadiqbal08 to take a look here -- also we should see if @vkaracic or @marjev can take a swing through as well since this is an ecommerce mod.

ecommerce/enterprise/utils.py
```diff
@@ -54,6 +55,7 @@ def get_enterprise_customer(site, token, uuid):
         'name': response['name'],
         'id': response['uuid'],
         'enable_data_sharing_consent': response['enable_data_sharing_consent'],
+        'contact_email': response['contact_email'],
```
pomegranited: I'd change this to reduce the dependency on https://github.com/edx/edx-enterprise/pull/66:

```python
'contact_email': response.get('contact_email', ''),
```

ecommerce/coupons/views.py
```diff
@@ -100,20 +103,58 @@ class CouponOfferView(TemplateView):
...
+        except Voucher.DoesNotExist:
+            return {'error': _('Coupon does not exist')}
+        except exceptions.ProductNotFoundError:
+            return {'error': _('The voucher is not applicable to your current basket.')}
+        valid_voucher, msg = voucher_is_valid(voucher, products, self.request)
+        if not valid_voucher:
             return {'error': msg}
-        return {'error': _('This coupon code is invalid.')}
+
+        context_data = None
```
pomegranited: At the risk of being annoying and asking for another refactor.. I think the functionality from here on down should be removed to a new method in `ecommerce.enterprise.utils`, e.g.

```python
        context_data = get_enterprise_customer_context_data(self.request, voucher)
```

`get_enterprise_customer_context_data` can return None if there's nothing to say.

pomegranited (changes requested):
I've got one small nit, and another impertinent refactor request..  🐑

ecommerce/enterprise/utils.py
bdero: Great idea! Thanks!

ecommerce/coupons/views.py
bdero: Yeah, I think this is a good idea - this seems like it should live in the Enterprise module. I went ahead and made `get_enterprise_customer_consent_failed_context_data` in there. It's kind of a mouthful, but I wasn't sure how to make it shorter while still being true to what it does. Thanks!

bdero:
Thanks for the review @pomegranited and thanks for taking a look @mattdrayer! I've addressed all comments.

ecommerce/coupons/tests/test_views.py
```diff
@@ -238,17 +242,25 @@ def test_no_product(self):
...
         # Make sure to always pair `course_seat_types` and `catalog_query` parameters because
         # if one of them is missing it could result in a SEGFAULT error when running tests
         # with migrations enabled.
-        _range = RangeFactory(products=[credit_seat, ], course_seat_types='credit', catalog_query='*:*')
+        range_kwargs = {
+            'products': [credit_seat, ],
+            'course_seat_types': 'credit',
+            'catalog_query': '*:*',
+        }
+        if enterprise_customer is not None:
```
vkaracic: ``if enterprise_customer:``

ecommerce/coupons/views.py
```diff
@@ -100,20 +102,25 @@ class CouponOfferView(TemplateView):
...
+        except exceptions.ProductNotFoundError:
+            return {'error': _('The voucher is not applicable to your current basket.')}
+        valid_voucher, msg = voucher_is_valid(voucher, products, self.request)
+        if not valid_voucher:
             return {'error': msg}
-        return {'error': _('This coupon code is invalid.')}
+
+        context_data = get_enterprise_customer_consent_failed_context_data(self.request, voucher)
+
+        if context_data is not None and 'error' not in context_data.keys():
```
vkaracic: ``if context_data and 'error' not in context_data:``

ecommerce/enterprise/utils.py
```diff
@@ -74,6 +78,53 @@ def get_enterprise_customers(site, token):
...
 
+def get_enterprise_customer_consent_failed_context_data(request, voucher):
+    """
+    Get the template context to display a message informing the user that they were not enrolled in the course
+    due to not consenting to data sharing with the Enterprise Customer.
+
+    If the `consent_failed` GET param is defined and it's not set to a valid SKU, return an error context that
+    says the given SKU doesn't exist.
+    """
+    consent_failed_sku = request.GET.get(CONSENT_FAILED_PARAM, None)
```
vkaracic: You don't need the additional ``None`` parameter when doing ``dict.get()``

ecommerce/enterprise/utils.py
```diff
@@ -74,6 +78,53 @@ def get_enterprise_customers(site, token):
...
+    except StockRecord.DoesNotExist:
+        return {'error': _('SKU {sku} does not exist.').format(sku=consent_failed_sku)}
+
+    # Return the view with an info message informing the user that the enrollment didn't complete
+    enterprise_customer = get_enterprise_customer_from_voucher(
+        request.site,
+        request.user.access_token,
+        voucher
+    )
+    if len(enterprise_customer['contact_email']) > 0:
```
vkaracic: ``if enterprise_customer['contact_email']: ``

ecommerce/enterprise/utils.py
```diff
@@ -127,21 +178,32 @@ def get_enterprise_course_consent_url(site, code, sku, consent_token, course_id,
...
         params=urlencode({
             'code': code,
             'sku': sku,
             'consent_token': consent_token,
         })
     )
+    failure_url = '{base}{resource}?code={code}&{consent_failed_param_name}={sku}'.format(
+        base=base_url,
+        resource=reverse('coupons:offer'),
+        code=code,
```
vkaracic: Please use ``urlencode`` for encoding query string parameters.

ecommerce/static/js/routers/offer_router.js
```diff
@@ -11,8 +11,11 @@ define([
             // Base/root path of the app
             root: '/coupons/',
 
-            routes: {
-                'offer/?code=:code': 'showOfferPage'
+            initialize: function(options) {
```
vkaracic, line 14: Why this change?

ecommerce/enterprise/utils.py
```diff
@@ -74,6 +78,53 @@ def get_enterprise_customers(site, token):
     ]
 
 
+def get_enterprise_customer_consent_failed_context_data(request, voucher):
```
marjev, line 82: Will this method ever be used outside the `CouponOfferView`? If not, this method should perhaps live within the view itself.

ecommerce/templates/edx/partials/_info_message.html
```diff
@@ -0,0 +1,11 @@
+{% if info != None %}
```
marjev: I think only `if info` will do the trick.

ecommerce/coupons/views.py
```diff
@@ -100,20 +102,25 @@ class CouponOfferView(TemplateView):
...
+        except Voucher.DoesNotExist:
+            return {'error': _('Coupon does not exist')}
+        except exceptions.ProductNotFoundError:
+            return {'error': _('The voucher is not applicable to your current basket.')}
+        valid_voucher, msg = voucher_is_valid(voucher, products, self.request)
+        if not valid_voucher:
             return {'error': msg}
-        return {'error': _('This coupon code is invalid.')}
+
+        context_data = get_enterprise_customer_consent_failed_context_data(self.request, voucher)
```
vkaracic: Make sure you're not overriding default context data. First get the default context data with ``context = super(CouponOfferView, self).get_context_data(**kwargs)`` and then append your data to that dictionary.

ecommerce/coupons/tests/test_views.py
```diff
@@ -270,6 +282,40 @@ def test_credit_seat_response(self):
...
+    @ddt.unpack
+    def test_consent_failed_message(self, contact_email, expected_response):
+        """ Verify that the consent failure message shows up when the consent_failed parameter is a valid SKU. """
+        self.mock_specific_enterprise_customer_api(
+            ENTERPRISE_CUSTOMER,
+            name='TestShib',
+            contact_email=contact_email
+        )
+        base_url = self.prepare_url_for_credit_seat(enterprise_customer=ENTERPRISE_CUSTOMER)
+        product = self.credit_seat.rangeproduct_set.first().product
```
vkaracic: Isn't the seat the product? You can do ``sku = self.credit_seat.stockrecords.first().partner_sku`` directly.

vkaracic (changes requested):
I have a few questions.

marjev (changes requested):

pomegranited (approved):
:+1: when tests pass.  Thanks for your patience, @bdero !

- [x] I tested this on my devstack using the lovely detailed instructions provided.
- [x] I read through the code.
- [x] I checked for accessibility issues.
- [x] Includes documentation.

ecommerce/coupons/tests/test_views.py
bdero: Done. I believe I've gotten comments on other edX PRs asking me to explicitly do checks against `None` instead of doing implicit truthy checks (I might not be remembering correctly, though). I will change this for now, but let me know if you still think this is the right way to go.

ecommerce/coupons/tests/test_views.py
bdero: It's not just grabbing the `product` for the SKU, the test also requires the course name for verification of the view message.

Given the product line wouldn't go away, would you still prefer `self.credit_seat.stockrecords.first().partner_sku` over `sku = product.stockrecords.first().partner_sku` in this case?

ecommerce/coupons/views.py
bdero: Done, thanks!

ecommerce/coupons/views.py
bdero: Done.

ecommerce/enterprise/utils.py
bdero: Fixed, thanks.

ecommerce/enterprise/utils.py
bdero: Done.

ecommerce/enterprise/utils.py
bdero: Done.

ecommerce/static/js/routers/offer_router.js
bdero: This change is necessary because any URL containing more params than just `code`, e.g. `offer/?code=SOME_CODE&consent_failed=SOME_SKU`, would result in those params getting parsed as the code before being passed to the backbone view.

I think it's safe to assume that `codes` will at least never have non-alphanumeric characters, so my fix was to more accurately parse out the code param as `\w+`.

ecommerce/enterprise/utils.py
bdero: I highly doubt it's usable outside the context of `CouponOfferView`, but the general trend seems to be that most Enterprise stuff is pulled into the Enterprise module whether that's the case or not. What are your thoughts about this @pomegranited @mattdrayer?

ecommerce/templates/edx/partials/_info_message.html
bdero: Done.

bdero:
@pomegranited Thanks!
@vkaracic @marjev Thanks for the reviews, I've addressed all your comments!

ecommerce/enterprise/utils.py
pomegranited: My thoughts were along the same lines as @bdero, but I see your point, since this logic is specific to this view. Your call, @marjev.

mattdrayer:
@asadiqbal08 if you can do a review of this PR, I think @bdero will be ready to wrap things up.

ecommerce/enterprise/utils.py
```diff
@@ -74,6 +79,58 @@ def get_enterprise_customers(site, token):
...
+        voucher
+    )
+    if not enterprise_customer:
+        return {'error': _('There is no Enterprise Customer associated with SKU {sku}.').format(
+            sku=consent_failed_sku
+        )}
+
+    if enterprise_customer['contact_email']:
+        contact_info = ' at {contact_email}'.format(contact_email=enterprise_customer['contact_email'])
+    else:
```
asadiqbal08: nit: We can remove the extra `else`
```
contact_info = ''
if enterprise_customer['contact_email']:
   contact_info = ' at {contact_email}'.format(contact_email=enterprise_customer['contact_email'])

```

ecommerce/enterprise/utils.py
```diff
@@ -74,6 +79,58 @@ def get_enterprise_customers(site, token):
...
+        contact_info = ''
```
asadiqbal08: a question: If we could not find out the enterprise_customer any `contact_email` then any other default one to show instead of empty sting ?

ecommerce/enterprise/utils.py
```diff
@@ -74,6 +79,58 @@ def get_enterprise_customers(site, token):
...
+
+    return {
+        'info': {
+            'title': _('Enrollment in {course_name} was not complete.').format(
+                course_name=product.course.name
+            ),
+            'message': _(
+                'If you have concerns about sharing your data, please contact your administrator '
+                'at {enterprise}{contact_info}.'
```
asadiqbal08: A mock-up on JIRA shows a message `If you have concerns about sharing your data, please contact your administrator at {enterprise} at {contact_info}.`

asadiqbal08:
@bdero few questions above. rest of the changes LGTM 👍 . branch need to be rebase with latest

ecommerce/enterprise/utils.py
bdero: No, on the upstream ticket we decided for it to be an optional contact email field, so we need to gracefully handle the case when it's blank.

ecommerce/enterprise/utils.py
bdero: Hmm yea, I think this should actually be refactored to be two separate translation strings - one for each case.

The first looks like this:
> If you have concerns about sharing your data, please contact your administrator at {enterprise} at {contact_info}.

And the second looks like this:
> If you have concerns about sharing your data, please contact your administrator at {enterprise}.

The way I wrote it is incorrect because the "at" part of the string isn't going through i18n.

ecommerce/enterprise/utils.py
marjev: IMHO, `get_enterprise_customer_from_voucher` seems to be related to the Enterprise stuff from the business logic perspective, so it should live in the Enterprise module, but since everything else in this method is about formatting data specifically for the `CouponOfferView`, I think it should be close to the View itself.
But, I'll leave it to @bdero to decide what he thinks is best in the context of this PR, to keep things consistent. It's not a showstopper.

marjev (approved):

ecommerce/enterprise/utils.py
bdero: Thanks. I don't have a terribly strong opinion here, but I'll leave this in `enterprise/utils.py` for now just because it seems nice to have e.g. the enterprise-related errors bundled up and the docstring the in this case. The way I had it embedded in the view before was starting to become quite hard to parse through. :)

ecommerce/enterprise/utils.py
bdero: Done.

bdero:
All comments addressed @asadiqbal08, thanks for the review!

@vkaracic When you get a chance, can you give this another look over and approve if you have no more requests?

ecommerce/coupons/tests/test_views.py
vkaracic: I'm trying to tell you that ``self.credit_seat`` is the product. In this case this should be true:
```
self.credit_seat.rangeproduct_set.first().product == self.credit_seat
```

ecommerce/coupons/views.py
```diff
@@ -100,20 +102,26 @@ class CouponOfferView(TemplateView):
...
+        except Voucher.DoesNotExist:
+            return {'error': _('Coupon does not exist')}
+        except exceptions.ProductNotFoundError:
+            return {'error': _('The voucher is not applicable to your current basket.')}
+        valid_voucher, msg = voucher_is_valid(voucher, products, self.request)
+        if not valid_voucher:
             return {'error': msg}
-        return {'error': _('This coupon code is invalid.')}
+
+        context_data = super(CouponOfferView, self).get_context_data(**kwargs) or {}
```
vkaracic: No need for the ``or``, context will always be a dictionary.

ecommerce/coupons/views.py
```diff
@@ -100,20 +102,26 @@ class CouponOfferView(TemplateView):
...
+        context_data.update(get_enterprise_customer_consent_failed_context_data(self.request, voucher))
+
+        if context_data and 'error' not in context_data.keys():
```
vkaracic: No need for ``.keys()``.

ecommerce/static/js/routers/offer_router.js
vkaracic: Have you tried ``'offer/?code=:code/': 'showOfferPage'`` ?

ecommerce/coupons/tests/test_views.py
vkaracic: In my experience it is. Although Python's ethos is ``explicit is better than implicit``, I believe saying ``if something-exists`` is explicit enough since ``enterprise_customer`` is not a boolean value. Even if it were I doubt you would want to proceed if it's ``False``.

vkaracic (approved):
A couple of nitpicks left.

ecommerce/coupons/views.py
bdero: Done.

ecommerce/coupons/views.py
bdero: Done, thanks.

ecommerce/static/js/routers/offer_router.js
bdero: I gave this a try, but this would only match against a URL containing a slash at the end of the code:
```
offer/?code=HV7SMPXDYJFTGRU2/
```

I don't think the routing library backbone uses is intended for grabbing stuff out of get params. I'd call this a generally bad pattern. An example of another downside of writing route keys this way is that now the order of the get params is needlessly important.

ecommerce/coupons/tests/test_views.py
bdero: Gotcha, fixed. Thanks!

bdero:
Thanks @vkaracic !

vkaracic:
You're very welcome!

PR #1190, opened 2017-03-07
Fix enrollment code product class slug.

vkaracic:
So ``slugify`` does not work as I expect it to, could've sworn that I tested that function, obviously not.
This PR contains the management command approach to updating the product class and enrollment codes, [this branch](https://github.com/edx/ecommerce/tree/vkaracic/ECOM-7393b) contains the migration approach. Not sure which one is more appropriate. Thoughts, @clintonb ?

https://openedx.atlassian.net/browse/ECOM-7393

### DIFF
--- a/ecommerce/extensions/catalogue/management/commands/remove_wrong_product_class.py
+++ b/ecommerce/extensions/catalogue/management/commands/remove_wrong_product_class.py
@@ -1,0 +1,37 @@
+import logging
+
+from django.core.management import BaseCommand
+from oscar.core.loading import get_model
+
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1190"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb:
I would use a migration if I had to do this. Given that we don't (to my knowledge) rely on the `slug` field anywhere, why does this even matter?

vkaracic:
For consistency sake.

PR #1191, opened 2017-03-07
Add migration to update enrollment code product class.

vkaracic:
https://openedx.atlassian.net/browse/ECOM-7393

### DIFF
--- a/ecommerce/extensions/catalogue/migrations/0024_fix_enrollment_code_slug.py
+++ b/ecommerce/extensions/catalogue/migrations/0024_fix_enrollment_code_slug.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from oscar.core.loading import get_model
+
+ProductClass = get_model("catalogue", "ProductClass")
+
+WRONG_SLUG = 'enrollment-code'
+RIGHT_SLUG = 'enrollment_code'
+
+
+def fix_enrollment_code_slug(apps, schema_editor):
+    """Update the faulty product class."""
+    try:
+        product_class = ProductClass.objects.get(slug=WRONG_SLUG)
+        product_class.slug = RIGHT_SLUG
+        product_class.save()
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1191"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb (commented):
This is unnecessary. There is nothing to be consistent with given that we don't use slugs. Also, the value in production is `enrollment_code`.

vkaracic:
Yes, but the value in every sandbox from here on out is going to be ``enrollment-code`` which might cause confusion if someone decides to use the slug attribute.

Not to mention that it's probably going to be like that in any possible open source contribution.

vkaracic:
AND in every new fresh installation of ecommerce.

ecommerce/extensions/catalogue/migrations/0024_fix_enrollment_code_slug.py
```diff
@@ -0,0 +1,47 @@
...
+Product = get_model("catalogue", "Product")
+ProductClass = get_model("catalogue", "ProductClass")
+
+WRONG_SLUG = 'enrollment-code'
+RIGHT_SLUG = 'enrollment_code'
+
+
+def fix_enrollment_code_slug(apps, schema_editor):
+    """Update or remove the faulty product class and update any enrollment codes with faulty product class."""
+    faulty_product_class = ProductClass.objects.filter(slug=WRONG_SLUG).first()
```
clintonb: Keep it simple. We don't want to touch every product, or delete anything.

```python
try:
  product_class = ProductClass.objects.get(slug=WRONG_SLUG)
  product_class.slug = RIGHT_SLUG
  product_class.save()
except DoesNotExist:
  # If it's not broken, there is nothing to fix.
  pass
```

The revert can be the same, but swap the constant names.

clintonb (approved):

marjev (approved):

PR #1192, opened 2017-03-08
Change SDN check to be called only on valid form submissions.

vkaracic:
ECOM-7419

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -66,6 +66,7 @@ define([
             field.find('~.help-block').append('<span>' + msg + '</span>');
             field.focus();
             event.preventDefault();
+            $('.payment-form').attr('data-has-error', true);
         }
 
         function appendCardHolderValidationErrorMsg(field, msg) {
@@ -91,6 +92,7 @@ define([
                 if ($(field).val() === '') {
                     event.preventDefault();
                     appendCardHolderValidationErrorMsg($(field), gettext('This field is required'));
+                    $('.payment-form').attr('data-has-error', true);
                 }
             });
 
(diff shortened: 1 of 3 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1192"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -66,6 +66,7 @@ define([
             field.find('~.help-block').append('<span>' + msg + '</span>');
             field.focus();
             event.preventDefault();
+            $('.payment-form').addClass('has-error');
```
clintonb: Use a `data-*` attribute instead of a class to ensure we don't run into errors if someone ever uses that class name for visual styling.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -229,7 +229,7 @@ define([
 
                     $('input[name=first_name]').val('Joey');
                     $('input[name=last_name]').val('Tribbiani');
-                    $('input[name=address_line1]').val('Central Park');
+                    $('input[name=address_line1]').val('Central Perk');
```
clintonb, line 232: ???

clintonb (approved):
Thanks for the quick turnaround.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
vkaracic: It's the cafe from Friends: http://friends.wikia.com/wiki/Central_Perk

clintonb: Never seen it...or *Seinfeld*.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -66,6 +66,7 @@ define([
             field.find('~.help-block').append('<span>' + msg + '</span>');
             field.focus();
             event.preventDefault();
+            $('.payment-form').attr('data-has-error', true);
```
clintonb, line 69: Use [`$.data()`](https://api.jquery.com/data/).

clintonb: What is present works, but using `$.data()` is cleaner.

vkaracic: That depends. If we want to use ``data-*`` attribute on DOM elements this is the way to set the value, because setting values with ``$.data()`` will set a value on the DOM node, not the element, so it's an either-or.

PR #1193, opened 2017-03-08
Updated pycountry and pytz

clintonb:
Django Oscar 1.3+ dropped support for older versions of pycountry.

ECOM-7422


clintonb:

rlucioni (approved):

PR #1194, opened 2017-03-08
Make sure we save the basket before adding anything back.

mjfrey:
Save basket after a flush to prevent multiple quantity items.


schenedx (approved):
👍 
Good catch.

PR #1195, opened 2017-03-09
ENT-201 use catalog contains endpoint for varifying the course seats against …

zubair-arbi:
…a course catalog range
@saleem-latif @asadiqbal08 @mattdrayer 

Previously we were using two API calls for checking if a course run exists for a catalog by first retrieving the catalog and then using the course_runs contains endpoint for the provided course run, from discovery service.

Now we are using the updated catalog contains endpoint which tells us if a course run is available in its query.

**Depends on:** [add support for course run key in catalog contains endpoint](https://github.com/edx/course-discovery/pull/721)

### DIFF
--- a/ecommerce/courses/tests/mixins.py
+++ b/ecommerce/courses/tests/mixins.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 import json
 
 import httpretty
@@ -35,6 +37,28 @@ class CourseCatalogServiceMockMixin(object):
         httpretty.register_uri(
             method=httpretty.GET,
             uri=single_catalog_uri,
+            body=course_discovery_api_response_json,
+            content_type='application/json'
+        )
+
+    def mock_course_discovery_api_for_catalog_contains(self, catalog_id=1, course_run_ids=None):
+        """
+        Helper function to register course catalog contains API endpoint.
+        """
+        course_run_ids = course_run_ids or []
+        courses = {course_run_id: True for course_run_id in course_run_ids}
+
+        course_discovery_api_response = {
+            'courses': courses
+        }
+        course_discovery_api_response_json = json.dumps(course_discovery_api_response)
+        catalog_contains_uri = '{}{}/contains/?course_run_id={}'.format(
+            self.COURSE_DISCOVERY_CATALOGS_URL, catalog_id, ','.join(course_run_ids)
+        )
+
+        httpretty.register_uri(
+            method=httpretty.GET,
+            uri=catalog_contains_uri,
             body=course_discovery_api_response_json,
             content_type='application/json'
         )
--- a/ecommerce/enterprise/entitlements.py
+++ b/ecommerce/enterprise/entitlements.py
@@ -7,7 +7,6 @@ customers with which they are affiliated. The coupon product id's for the
 enterprise entitlements are provided by the Enterprise Service on the basis
 of the learner's enterprise eligibility criterion.
 """
-import hashlib
 import logging
 
 from django.conf import settings
@@ -17,8 +16,8 @@ from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import SlumberBaseException
 
 from ecommerce.core.constants import COUPON_PRODUCT_CLASS_NAME
+from ecommerce.core.utils import get_cache_key
 from ecommerce.coupons.views import voucher_is_valid
-from ecommerce.courses.utils import get_course_catalogs
 from ecommerce.enterprise import api as enterprise_api
 from ecommerce.enterprise.utils import is_enterprise_feature_enabled
 from ecommerce.extensions.api.serializers import retrieve_all_vouchers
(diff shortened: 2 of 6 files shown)


ecommerce/extensions/offer/models.py
```diff
@@ -259,6 +252,32 @@ def run_catalog_query(self, product, query=None):
 
         return response
 
+    def catalog_contains_product(self, product):
```
saleem-latif, line 257: I think it would be best to define a helper method in `ecommerce/coupons/utils.py` that makes network calls to discovery API, and use that method here. This way will have all procedures that perform API calls in a single place.
Name of the utility function can be something like `catalog_contains_course(self, course)` or something better.

saleem-latif: Also, in my opinion I think it would be more readable if `catalog_contains_product ` here returns a boolean if it is possible. and we changes function signature accordingly. thoughts ?

ecommerce/extensions/offer/models.py
```diff
@@ -259,6 +252,32 @@ def run_catalog_query(self, product, query=None):
...
+        """
+        Retrieve the results from using the catalog contains endpoint for
+        catalog service for the catalog id contained in field "course_catalog".
+        """
+        request = get_current_request()
+        partner_code = request.site.siteconfiguration.partner.short_code
+        cache_key = get_cache_key(
+            site_domain=request.site.domain,
```
saleem-latif, line 265: It would be good to also add `resource` key-value pair in cache key to make sure cache key does not collide with any other.

ecommerce/extensions/offer/models.py
```diff
@@ -259,6 +252,32 @@ def run_catalog_query(self, product, query=None):
...
+            partner_code=partner_code,
+            course_id=product.course_id,
+            catalog_id=self.course_catalog
+        )
+        response = cache.get(cache_key)
+        if not response:
+            course_catalog_api_client = request.site.siteconfiguration.course_catalog_api_client
+            try:
+                response = course_catalog_api_client.catalogs(self.course_catalog).contains.get(
```
saleem-latif: nit: please add a comment here explaining what would be the final URL for the API call, may be something like

```
# GET: /api/v1/catalogs/{catalog_id}/contains?course_run_ids={course_run_id}
```

saleem-latif (changes requested):
Overall looks good, however, I have added a couple of comments.

ecommerce/extensions/offer/models.py
zubair-arbi: added

ecommerce/extensions/offer/models.py
zubair-arbi: @saleem-latif @mattdrayer I would recommend a separate story for refactoring/moving the range model API calls to the `ecommerce/coupons/utils.py` file as we have to update both this method and `run_catalog_query` method. Also we have to add separate tests for the utils too. The scope of current story already seems a bit extended and its involves working with 3 repos.

zubair-arbi:
@mattdrayer @mjfrey please review.

ecommerce/enterprise/entitlements.py
```diff
@@ -161,60 +160,32 @@ def is_course_in_enterprise_catalog(site, course_id, enterprise_catalog_id):
...
             return False
 
     try:
-        is_course_in_course_runs = response['course_runs'][course_id]
+        is_course_in_catalog = response['courses'][course_id]
     except KeyError:
         return False
 
-    return is_course_in_course_runs
+    return is_course_in_catalog
```
mattdrayer: Looks like this can be moved to the `try` block:
```
try: 
    return response['courses'][course_id]
except KeyError:
    return False
```

mattdrayer:
One change, otherwise 👍

ecommerce/extensions/offer/models.py
mattdrayer: I'm okay with logging (another) follow-up story -- we've put in a fair amount of effort on this job and we have some higher priorities that we'll need to address next.

zubair-arbi: I have logged a ticket [ENT-257](https://openedx.atlassian.net/browse/ENT-257) to address this feedback.

saleem-latif (approved):

saleem-latif:
@zubair-arbi commits need to be squashed otherwise LGTM 👍

mattdrayer:
@vkaracic and/or @marjev may want to take a quick look at this one.

ecommerce/courses/tests/mixins.py
```diff
@@ -39,6 +39,30 @@ def mock_course_discovery_api_for_catalog_by_resource_id(self, catalog_id=1, cat
             content_type='application/json'
         )
 
+    def mock_course_discovery_api_for_catalog_contains(self, catalog_id=1, course_run_ids=None):
+        """
+        Helper function to register course catalog contains API endpoint.
+        """
+        courses = {}
+        course_run_ids = course_run_ids or []
```
marjev: Can you set `[]` as the default value in line 42? That way you won't need this line.

ecommerce/courses/tests/mixins.py
```diff
@@ -39,6 +39,30 @@ def mock_course_discovery_api_for_catalog_by_resource_id(self, catalog_id=1, cat
...
+        for course_run_id in course_run_ids:
+            courses.update({course_run_id: True})
```
marjev: You can do this during the initialization like this:
`courses = {course_run_id: True for course_run_id in course_run_ids}`

ecommerce/courses/tests/mixins.py
```diff
@@ -39,6 +39,30 @@ def mock_course_discovery_api_for_catalog_by_resource_id(self, catalog_id=1, cat
...
+
+        course_discovery_api_response = {
+            u'courses': courses
```
marjev: You can import `unicode_literals` like this: https://github.com/edx/ecommerce/blob/master/ecommerce/coupons/views.py#L1

Then, you won't need to mark the string as Unicode.

ecommerce/courses/tests/mixins.py
zubair-arbi: @marjev This would result in pylint error ["dangerous-default-value"](http://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python):
```ecommerce@precise64:~/ecommerce$ make quality
pylint --rcfile=pylintrc ecommerce e2e
************* Module ecommerce.courses.tests.mixins
W: 42, 4: Dangerous default value [] as argument (dangerous-default-value)
```

zubair-arbi:
@marjev I have addressed your feedback, please have another look.

marjev (approved):
LGTM

clintonb (dismissed):
If this feature relies on the Catalog Service, please add a feature flag to disable it until the Catalog Service is released.

zubair-arbi:
@clintonb we already have a waffle switch `enable_enterprise_on_runtime` which [enables enterprise functionality in Ecom](https://github.com/edx/ecommerce/blob/master/ecommerce/enterprise/entitlements.py#L42) and we can use this feature flag instead of adding a new one. 
@mattdrayer what is your suggestion?

mattdrayer:
We're sort of in moot-point territory now because the updated Catalog service has already deployed.  However, for this particular PR, I think I would have recommended that we avoid binding the catalog query logic to the enterprise feature flag.  That flag is for enterprise features, and while they might *use* the updated query logic, the actual code was more of a core ecommerce change.  So a new feature flag probably should have been introduced if we were expecting to merge both PRs to their respective masters and subsequently manage the releases.

TBH I much prefer simply merging+deploying the updated Catalog service, ensuring no rollback was required, and then merging+deploying the updated Ecommerce service at a later point in time (which is what we're doing now).  It's easier to manage and we don't have to worry about juggling feature flags.

PR #1196, opened 2017-03-09
Reactivated the offers app to fix the offers dashboard

clintonb:


clintonb:
FYI @edx/ecommerce

rlucioni (commented):
@clintonb reactivating this app will expose pages which aren't part of the dashboard (read: not behind a login page). We've had problems with search engines indexing this content in the past. It seems like the `robots.txt` file in this repo should prevent that, but it doesn't seem to be used. See https://ecommerce.edx.org/robots.txt.

Without fixing that first, I'm concerned that reactivating this app will result in search engines indexing this content again.

clintonb:
@rlucioni the offers pages being exposed are not linked anywhere for a search index to crawl. If they were to somehow find the pages, they wouldn't load because the catalogue app is disabled.

PR #1198, opened 2017-03-12
Added isort for import order validation

clintonb:


clintonb:
@edx/ecommerce please review

rlucioni (approved):

PR #1200, opened 2017-03-13
Log warnings for SDN API exceptions

mjfrey:
Change log exceptions to warnings for SDN API.

### DIFF
--- a/ecommerce/extensions/payment/utils.py
+++ b/ecommerce/extensions/payment/utils.py
@@ -104,11 +104,11 @@ class SDNClient(object):
         try:
             response = requests.get(sdn_check_url, timeout=settings.SDN_CHECK_REQUEST_TIMEOUT)
         except requests.exceptions.Timeout:
-            logger.exception('Connection to US Treasury SDN API timed out for [%s].', name)
+            logger.warning('Connection to US Treasury SDN API timed out for [%s].', name)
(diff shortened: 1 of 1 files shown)


ecommerce/extensions/payment/utils.py
```diff
@@ -104,11 +104,11 @@ def search(self, name, address, country):
         try:
             response = requests.get(sdn_check_url, timeout=settings.SDN_CHECK_REQUEST_TIMEOUT)
         except requests.exceptions.Timeout:
-            logger.exception('Connection to US Treasury SDN API timed out for [%s].', name)
+            logger.warning('Connection to US Treasury SDN API timed out for [%s].', name)
```
davec-edx, line 107: Is a failure retried?  Should there be a threshold put on this such that after _x_ failures it logs an exception instead?  In other words, if there is a real problem, don't want to hide it by accident.

mjfrey: I am going to setup a special Splunk alert for these.  Rather than have it be in the catch all that exceptions go.

davec-edx: 👍

mjfrey:
@rlucioni please review.

rlucioni (approved):

PR #1202, opened 2017-03-14
[WIP] Change receipt page to be a Backbone view

vkaracic:
@clintonb this is very much still a work in progress but please take a look and let me know if I'm on the right track, and if this is inline with your [comment](https://github.com/edx/ecommerce/pull/1187#issuecomment-284720747).


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1202&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb:
You should not redirect to the receipt page unless there is an order. What's the waiting for? The only payment processor for which we have to wait is CyberSource. There should be no waiting there because we should be using the interstitial from https://github.com/edx/ecommerce/pull/1025.

I am personally not a fan of client-side rendering, especially for static pages like this. Why does the page need to be converted to Backbone? (FYI we are switching to React according to https://github.com/edx/open-edx-proposals/pull/32)

vkaracic:
There is always an order, this is for when a user applies an enrollment code in which case it circumvents any payment and goes directly to the receipt page (currently it redirects the user to the dashboard, but the intention was to redirect to the receipt page). Since the order is fulfilled asynchronously it could take some time before we know if it's complete or there was a fulfillment error, and since an approach like [this one](https://github.com/edx/ecommerce/pull/1149) does not work, this is the alternative I've came up with.

clintonb:
I offer a simpler approach that doesn't involve overhauling the receipt page unnecessarily. As with Cybersource, redirect to an interstitial that waits for the order to reach certain states. Once those states are reach redirect to the final page in the flow—receipt or error.

vkaracic:
Can you point out to me where exactly does the Cybersource interstitial view _wait_ for something?

clintonb:
The wait is not an asynchronous poll. The data that gets POSTed by CyberSource is used to create the order synchronously. After it's created, the user is redirected.

vkaracic:
The order is created synchronously, but fulfilled asynchronously, same as it is here.

clintonb:
Back up. We don't necessarily care about fulfillment here. Free orders are created immediately, so it should be safe to go straight to the receipt.

Why are we treating enrollment code orders differently from any other order?

vkaracic:
So we can display the receipt page even if the order has not been fulfilled (in this case the user enrolled in the course) or there was a fulfillment error?

clintonb:
Yes. The receipt page doesn't care about the order's status.

mjfrey:
But there are cases where an enrollment code was used but not successful.  For that reason, checking the order status is necessary.

clintonb:
Define "not successful". If the enrollment code cannot be redeemed, an order should not be created. If the order is created, but not fulfilled, that is a separate issue.

mjfrey:
Yes I am talking about the later.  There are times when an enrollment code (free order) is used for a professional course for example.  But fulfillment can not happen.  Today The learner is just taken to the dashboard with no indication of success or failure.

PR #1203, opened 2017-03-14
ECOM-7410 Update course seat validation check to prevent product title from having HTML

Ayub-Khan:
**ECOM-7410**

**Description :**
Students were unable to checkout using cybersource when product title had HTML in it. Updated the course seat creation validation check for product title to check for HTML.

**Testing :**
Locally tested changes Working fine. 
All tests are passing locally.


clintonb (changes requested):
We should fix this at the source—CAT. Don't allow courses to be created with invalid text.

Also, please include the ticket number on the last line of your commit message.

attiyaIshaque (changes requested):
Code changes look good. Could you add a regression test for this?

ecommerce/static/js/utils/validation_patterns.js
```diff
@@ -14,5 +14,13 @@ define([
         _.extend(Backbone.Validation.messages, {
             courseId: gettext('The course ID is invalid.')
         });
+
+        _.extend(Backbone.Validation.patterns, {
+            productName: /^((?!&\w+;|<\/*\w+>).)*$/
+        });
+
+        _.extend(Backbone.Validation.messages, {
+            productName: gettext('The product name is invalid.')
```
clintonb: Provide helpful text that explains what is considered valid. Also, run `make fake_translations` to update the translations.

clintonb (changes requested):
This needs tests.

Ayub-Khan:
@clintonb @attiyaIshaque   updated

ecommerce/static/js/utils/validation_patterns.js
```diff
@@ -14,5 +14,13 @@ define([
         _.extend(Backbone.Validation.messages, {
             courseId: gettext('The course ID is invalid.')
         });
+
+        _.extend(Backbone.Validation.patterns, {
+            productName: /^((?!&\w+;|<\/*\w+>).)*$/
+        });
+
+        _.extend(Backbone.Validation.messages, {
+            productName: gettext('The product name contains HTML in it.')
```
clintonb: The product name cannot contain HTML.

clintonb (approved):
Remember to squash your commits, and move the ticket number to the last line of your commit message.

attiyaIshaque (approved):
Once commits squashed 👍

PR #1204, opened 2017-03-15
ENT-264: Fix invalid use of request.user.access_token for enterprise customer lookups

mattdrayer:
Addresses bug reported in ENT-264: List of Enterprise Customers is not appearing in the "Enterprise Customer" dropdown of the Coupon Administration Tool

Explanation: 
The `get_enterprise_customers` call included in `EnterpriseCustomerViewSet.get` currently specifies the `request.user.access_token` field, which results in an incorrect value.  This was not detected on devstack or in sandboxes because all of the testing is typically done using the 'staff' user, which has elevated permissions.  It became clear in the staging environment using a regular user's access token which does not translate to the same level of permissions as the site's access token.

Solution:
Update reference to instead use the site's token, via `site.siteconfiguration.access_token`

@mjfrey, FYI

### DIFF
--- a/ecommerce/core/utils.py
+++ b/ecommerce/core/utils.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import hashlib
 import logging
+from urlparse import parse_qs, urlparse
 
 import six
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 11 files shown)


ecommerce/core/utils.py
```diff
@@ -45,3 +46,30 @@ def get_cache_key(**kwargs):
     key = '__'.join(['{}:{}'.format(item, value) for item, value in six.iteritems(kwargs)])
 
     return hashlib.md5(key).hexdigest()
+
+
+def traverse_pagination(response, endpoint):
```
mattdrayer, line 51: Moved `traverse_pagination` to this location to eliminate the coupling between `enterprise` and `courses`

mattdrayer: A nice side effect of this is that we also eliminated the coupling between `coupons` and `courses` -- that is included in this PR as well.

mattdrayer:
@mjfrey @vkaracic @marjev -- assuming we see another green build after pushing the squashed version, this PR should be ready for review.

marjev (approved):

vkaracic (approved):
Can you please remove ``TestCase`` from ``ecommerce/enterprise/tests/test_api.py:EnterpriseAPITests`` as well?

mjfrey (approved):

mattdrayer:
Thanks all -- @vkaracic I've made your recommended change and pushed an updated branch -- will merge after checks are green.

PR #1205, opened 2017-03-16
[ECOM-7447] Logout user if SDN matched.

vkaracic:


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1205&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/api/v2/views/sdn.py
```diff
@@ -50,4 +51,5 @@ def post(self, request):
                 # the user is allowed to make the purchase.
                 pass
 
-        return Response({'hits': hits})
+        logout(request)
+        return Response({'hits': 1})
```
vkaracic: Changed to always return a hit and logout for debugging purposes.

ecommerce/static/js/pages/sdn_failure_page.js
```diff
@@ -0,0 +1,16 @@
+define([], function () {
+    return {
+        onReady: function () {
+            function delete_cookie( name, path, domain ) {
```
vkaracic: Should probably remove this and do this in basket_page.js

ecommerce/extensions/api/v2/views/sdn.py
```diff
@@ -45,6 +46,7 @@ def post(self, request):
                         country,
                         response
                     )
+                    logout(request)
```
vkaracic, line 49: Not sure if this one is still necessary.

e2e/mixins.py
```diff
@@ -12,11 +12,9 @@
 from selenium.webdriver.support.ui import WebDriverWait
 
 from e2e.api import EnrollmentApiClient, get_access_token
-from e2e.config import (
-    BASIC_AUTH_PASSWORD, BASIC_AUTH_USERNAME, ECOMMERCE_API_URL, ECOMMERCE_URL_ROOT, LMS_AUTO_AUTH, LMS_EMAIL,
-    LMS_HTTPS, LMS_PASSWORD, LMS_URL_ROOT, LMS_USERNAME, MARKETING_URL_ROOT, MAX_COMPLETION_RETRIES, PAYPAL_EMAIL,
-    PAYPAL_PASSWORD
-)
+from e2e.config import (BASIC_AUTH_PASSWORD, BASIC_AUTH_USERNAME, ECOMMERCE_API_URL, ECOMMERCE_URL_ROOT, LMS_AUTO_AUTH,
```
clintonb: Looks like you ran `isort` on everything, and we never configured `multi_line_output`. Please set `multi_line_output=3` in `.isort.cfg` and re-run the command to fix these. See https://github.com/timothycrosley/isort#multi-line-output-modes for details.

ecommerce/extensions/api/v2/tests/views/test_sdn.py
```diff
@@ -40,17 +41,29 @@ def test_authentication_required(self):
         self.client.logout()
         self.assertEqual(self.make_request().status_code, status.HTTP_401_UNAUTHORIZED)
 
-    @ddt.data(0, 1)
-    def test_sdn_check_match(self, hits):
-        """Verify the endpoint returns the number of hits SDN check made."""
+    @contextmanager
+    def sdn_mock(self, hits):
```
clintonb: `mock_sdnclient_search`

ecommerce/extensions/api/v2/tests/views/test_sdn.py
```diff
@@ -40,17 +41,29 @@ def test_authentication_required(self):
...
+        """Mock the SDNClient.search and User.deactivate_account methods."""
         with mock.patch.object(SDNClient, 'search') as sdn_validator_mock:
             with mock.patch.object(User, 'deactivate_account') as deactivate_account_mock:
```
clintonb: This code will be cleaner if this is moved to its own mock method.

ecommerce/extensions/api/v2/tests/views/test_sdn.py
```diff
@@ -40,17 +41,29 @@ def test_authentication_required(self):
...
                 sdn_validator_mock.return_value = {'total': hits}
                 deactivate_account_mock.return_value = True
```
clintonb: These can be one-liners:

```python
mock.patch.object(User, 'deactivate_account', return_value=True) 
```

ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -5,7 +5,7 @@
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 from django.db.utils import IntegrityError
-from paypalrestsdk import WebProfile    # pylint: disable=ungrouped-imports
```
clintonb: You're touching *way* more code than you should be. Please revert all changes that are not actually needed for SDN/logout. In the future, if you need to run `isort, run it against the specific directory/files you are editing.

ecommerce/extensions/payment/tests/views/test_sdn_view.py
```diff
@@ -0,0 +1,19 @@
+from django.core.urlresolvers import reverse
+
+from ecommerce.tests.testcases import TestCase
+
+
+class SDNViewsTests(TestCase):
+    failure_path = reverse('sdn_failure')
+
+    def setUp(self):
```
clintonb: Remove. There is no need to have this here if you aren't actually overriding the default.

ecommerce/static/js/apps/sdn_failure_app.js
```diff
@@ -0,0 +1,13 @@
+require([
```
clintonb: Yeah...this is overkill.

ecommerce/templates/oscar/checkout/sdn_logout.html
```diff
@@ -0,0 +1,13 @@
+{% extends 'edx/base.html' %}
```
clintonb: You've made this a bit more complicated than it needs to be. Keep the existing page. The only change necessary is to load the LMS logout page in an invisible iframe. No need for JS at all since there is no redirection taking place.

ecommerce/extensions/api/v2/views/sdn.py
clintonb: Keep it. If the account is deactivated, you don't want the user believing they can perform actions they cannot.

clintonb (changes requested):
Simplify.

ecommerce/extensions/payment/management/commands/paypal_profile.py
vkaracic: But technically all these changes are correct.

... once I run ``isort`` with the correct flags.

ecommerce/extensions/api/v2/views/sdn.py
vkaracic: Yes, but shouldn't the iframe take care of logging the user out?

ecommerce/templates/oscar/checkout/sdn_logout.html
vkaracic: If you mean adding something like:
```
<div id="iframeContainer" style="visibility: hidden">
  <iframe src="http://localhost:8000/logout"></iframe>
</div>
```
in the SDN failure template - that doesn't work. That code redirects the user the dashboard and if I add ``sandbox="allow-forms allow-same-origin allow-scripts" `` to the iframe it doesn't redirect (it load the iframe just fine) but does not log out the user, at least it doesn't work for me locally.

ecommerce/extensions/payment/management/commands/paypal_profile.py
clintonb: They may be technically correct but they are out-of-scope for this PR. It's unnecessary noise. If you want to clean the entire repo, do it in another PR.

vkaracic:
@clintonb ready for a second review!

ecommerce/extensions/api/v2/tests/views/test_sdn.py
```diff
@@ -40,16 +41,26 @@ def test_authentication_required(self):
         self.client.logout()
         self.assertEqual(self.make_request().status_code, status.HTTP_401_UNAUTHORIZED)
 
+    @contextmanager
+    def mock_sdnclient_search(self, hits):
+        """Mock the SDNClient.search and User.deactivate_account methods."""
+        with mock.patch.object(SDNClient, 'search', return_value={'total': hits}) as sdn_validator_mock:
+            with mock.patch.object(User, 'deactivate_account', return_value=True):
+                yield
+                self.assertTrue(sdn_validator_mock.called)
```
clintonb: You're doing far too much in this method. Break this up into two mock methods. If the assertion is really important, it too can be a method.

ecommerce/templates/oscar/checkout/sdn_logout.html
clintonb: Have you looked at the logs to determine why the client is redirected to the dashboard rather than logged out?

clintonb (commented):
I'd like more info on why a simple iframe doesn't work.

ecommerce/extensions/api/v2/tests/views/test_sdn.py
vkaracic: Sorry, can you elaborate a bit more? I've done this to somewhat DRY the lower two test cases and if I'm going to create mock methods for each individual mock and call them in both tests it kinda beats that purpose.

ecommerce/templates/oscar/checkout/sdn_logout.html
vkaracic: Some update: The redirect to the dashboard happens because of this:
https://github.com/edx/edx-platform/blob/master/lms/templates/main.html#L45-L52
But omitting the ``allow-top-navigation`` from the iframe sandbox flags prevents that redirect, hence the second scenario that I mentioned above.
Still trying to figure out why it's not logging the user out.

clintonb: Simple solution: update the logout view, setting `allow_iframing` to `True`.

vkaracic:
@clintonb I believe I have addressed your comments. Also I've verified that this works locally and on a sandbox.

ecommerce/templates/oscar/checkout/sdn_failure.html
```diff
@@ -36,4 +36,8 @@
       </div>
     </div>
   </div>
+
+  <div id="iframeContainer" style="visibility: hidden">
```
clintonb: This `<div>` is unnecessary. Apply the style to the `<iframe>`.

ecommerce/extensions/payment/tests/views/test_sdn_view.py
```diff
@@ -0,0 +1,13 @@
+from django.core.urlresolvers import reverse
+
+from ecommerce.tests.testcases import TestCase
+
+
+class SDNViewsTests(TestCase):
```
clintonb: The filename should be `test_sdn.py`. This class name should be `SDNFailureTests`.

ecommerce/extensions/payment/tests/views/test_sdn_view.py
```diff
@@ -0,0 +1,13 @@
...
+    failure_path = reverse('sdn:failure')
+
+    def test_sdn_logout_context(self):
+        """SDN logout view needs to have the logout URL in its context."""
+        logout_url = self.site.siteconfiguration.build_lms_url('/logout')
```
clintonb: No need for the slash.

ecommerce/extensions/api/v2/tests/views/test_sdn.py
```diff
@@ -42,18 +42,24 @@ def test_authentication_required(self):
 
     @ddt.data(0, 1)
     def test_sdn_check_match(self, hits):
-        """Verify the endpoint returns the number of hits SDN check made."""
```
clintonb: Why did any of this need to be touched?

clintonb (approved):

ecommerce/extensions/api/v2/tests/views/test_sdn.py
vkaracic: Scoped it in by mistake.

vkaracic:
@edx/helio somebody else please review.

ecommerce/templates/oscar/checkout/sdn_failure.html
```diff
@@ -36,4 +36,6 @@
       </div>
     </div>
   </div>
+
+  <iframe style="visibility: hidden" sandbox="allow-forms allow-same-origin allow-scripts" src="{{ logout_url }}"></iframe>
```
iivic, line 40: Why inline style?

vkaracic: Because it only needs to be hidden from the user, it doesn't and never will require any styling other than that, so there is no need to include id/class and SASS/CSS changes for this.

iivic (approved):

PR #1206, opened 2017-03-16
Fixing a11y issues on new basket page

iivic:
@cptvitamin You can check which ticket is done and for those who are not done properly please leave a comment.
@vkaracic Please review from a more technical point.

The following tickets have been addressed:
- [ ] ECOM-7506
- [ ] ECOM-7515
- [ ] ECOM-7514
- [ ] ECOM-7513
- [ ] ECOM-7521
- [ ] ECOM-7509

### DIFF
--- a/ecommerce/templates/coupons/coupon_app.html
+++ b/ecommerce/templates/coupons/coupon_app.html
@@ -22,7 +22,7 @@
                     <span aria-hidden="true" class="icon-bar"></span>
                 </button>
                 <a class="navbar-brand" href="/coupons/">
-                    <div class="navbar-brand-logo" alt="{{ platform_name }}"></div>
+                    <img class="navbar-brand-logo" alt="{{ platform_name }} logo" src="{% static 'images/default-logo.png' %}"/>
                     <div class="navbar-brand-app">{% trans "E-Commerce Code Administration" %}</div>
                 </a>
             </div>
--- a/ecommerce/templates/courses/course_app.html
+++ b/ecommerce/templates/courses/course_app.html
@@ -22,7 +22,7 @@
                     <span aria-hidden="true" class="icon-bar"></span>
                 </button>
                 <a class="navbar-brand" href="/courses/">
-                    <div class="navbar-brand-logo" alt="{{ platform_name }}"></div>
+                    <img class="navbar-brand-logo" alt="{{ platform_name }} logo" src="{% static 'images/default-logo.png' %}"/>
                     <div class="navbar-brand-app">{% trans "E-Commerce Administration" %}</div>
                 </a>
             </div>
--- a/ecommerce/templates/edx/partials/_base_navbar.html
+++ b/ecommerce/templates/edx/partials/_base_navbar.html
@@ -1,19 +1,19 @@
 {% load i18n %}
+{% load staticfiles %}
 
-<nav class="navbar navbar-default" aria-label="Account">
+<header class="navbar navbar-default">
   <div class="container">
     <!-- Brand and toggle get grouped for better mobile display -->
     <div class="navbar-header">
-      <button type="button" class="navbar-toggle collapsed" data-toggle="collapse"
-              data-target="#main-navbar-collapse" aria-expanded="false">
+      <button id="hamburger-button" class="navbar-toggle collapsed" type="button" aria-expanded="false">
         <span class="sr-only">{% trans "Toggle navigation" %}</span>
         <span aria-hidden="true" class="icon-bar"></span>
         <span aria-hidden="true" class="icon-bar"></span>
         <span aria-hidden="true" class="icon-bar"></span>
       </button>
-      <div class="navbar-brand">
-        <div class="navbar-brand-logo" alt="{{ platform_name }}"></div>
-      </div>
+      <h1 class="navbar-brand">
+        <img class="navbar-brand-logo" alt="{{ platform_name }} logo" src="{% static 'images/default-logo.png' %}"/>
+      </h1>
     </div>
 
     <div class="collapse navbar-collapse" id="main-navbar-collapse">
@@ -45,4 +45,4 @@
       </ul>
     </div>
   </div>
-</nav>
+</header>
--- a/ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
+++ b/ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
@@ -106,20 +106,19 @@
                 <h2 id="payment-method-region" class="title">{% trans "SELECT PAYMENT METHOD" %}</h2>
                 <div class="col-sm-12 payment-methods">
                     <div id="payment-method" class="col-md-4 col-sm-7 col-xs-6">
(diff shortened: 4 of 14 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1206&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/templates/edx/partials/_base_navbar.html
```diff
@@ -1,6 +1,6 @@
 {% load i18n %}
 
-<nav class="navbar navbar-default" aria-label="Account">
+<header class="navbar navbar-default" aria-label="Account">
```
cptvitamin: @iivic The aria-label for this element doesn't quite make sense anymore. Let's make it "Shopping Cart" since it is the header for the shopping cart page.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -104,20 +104,20 @@
                 <legend>{% trans "SELECT PAYMENT METHOD" %}</legend>
                 <div class="col-sm-12 payment-methods">
                     <div id="payment-method" class="col-md-4 col-sm-7 col-xs-6">
-                        <a href="#payment-information" id="payment-method-image" aria-label={% trans "Pay with a Credit Card" %} role="button"><img aria-hidden="true" src="/static/images/credit_card_options.png" alt="{% trans "Credit cards" %}"></a>
+                        <a href="#payment-information" id="payment-method-image" aria-label="{% trans "Pay with a Credit Card" %}" role="button"><img aria-hidden="true" src="/static/images/credit_card_options.png" alt=""></a>
```
cptvitamin: Let's convert the aria-label to a title attribute.  title reaches a broader audience and has better implementation support.  Screen readers treat both the same on button elements (or links with role=button)

Let's remove the aria-hidden=true on the image here (its an invalid attr and will cause test failures).  The proper way to hide an image is with an empty alt attribute, which we now have.  Alternatively, you could use role=presentation, but its redundant and not necessary.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -104,20 +104,20 @@
...
                     </div>
                     <div id="payment-processor" class="payment-buttons col-md-3 col-sm-5 col-xs-6" data-basket-id="{{ basket.id }}">
                         {# Translators: Do NOT translate the name PayPal. #}
-                        <button aria-label={% trans "Pay with PayPal" %}
+                        <button aria-label="{% trans "Pay with PayPal" %}"
```
cptvitamin: let's remove the aria-label all together since we have a title attr on line 119 below.  This will make it consistent with the CreditCard button.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -162,19 +162,21 @@
...
                             {% endfor %}
                         </select>
                         <p class="help-block"></p>
                     </div>
                     <div class="form-group col-sm-6 col-xs-5">
-                        <select id="card-expiry-year" class="form-control pci-field" required aria-required="true">
-                            <option>{% trans "(Year)" %}</option>
+                        <select id="card-expiry-year" class="form-control pci-field" required 
+                        aria-labelledby="year-label expiration-label">
+                            <option id=year-label>{% trans "Year" %}</option>
```
cptvitamin: nit, can you quote this id value?

cptvitamin (changes requested):
Some small requests and modifications based on implementation.

ecommerce/templates/edx/partials/_base_navbar.html
iivic: This header is used on other pages as well so "Shopping Cart" will not describe it accurately. There will maybe be more pages in the future so a more generic name or leaving it as is would be better in this case.

vkaracic: I think the label is irrelevant on this element. ``Account`` label should be added where the account is displayed, not the whole header.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -162,19 +162,21 @@
                         <i class="fa fa-lock" aria-hidden="true"></i>
                         <p class="help-block"></p>
                     </div>
-                    <label id="expiration-label" for="card-expiry-month" class="control-label">{% trans "Expiration (required)" %}</label>
+                    <p id="expiration-label" for="card-expiry-month" class="control-label">{% trans "Expiration (required)" %}</p>
                     <div class="form-group col-sm-6 col-xs-7">
-                        <select id="card-expiry-month" class="form-control pci-field" required aria-required="true">
-                            <option>{% trans "(Month)" %}</option>
+                        <select id="card-expiry-month" class="form-control pci-field" required 
+                        aria-labelledby="month-label expiration-label">
```
vkaracic, line 169: Is ``aria-labelledby="month-label expiration-label"`` a combination of the two labels?

iivic: According to [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) it is.

iivic:
@cptvitamin You can review the latest commit
- [ ] ECOM-7508
- [ ] ECOM-7517

cptvitamin:
I'm not seeing ```<section class="footer-main container">``` converted to a footer element

iivic:
@cptvitamin  I can't locate the mentioned section. 
For more info look at the comment section https://openedx.atlassian.net/browse/ECOM-7518

ecommerce/templates/coupons/coupon_app.html
```diff
@@ -22,7 +22,7 @@
                     <span aria-hidden="true" class="icon-bar"></span>
                 </button>
                 <a class="navbar-brand" href="/coupons/">
-                    <div class="navbar-brand-logo" alt="{{ platform_name }}"></div>
+                    <img class="navbar-brand-logo" alt="Institution logo" src="{% static 'images/default-logo.png' %}"/>
```
cptvitamin: @iivic the alt for the log should be dynamic alt="{{ platform_name }} logo">

ecommerce/templates/courses/course_app.html
```diff
@@ -22,7 +22,7 @@
                     <span aria-hidden="true" class="icon-bar"></span>
                 </button>
                 <a class="navbar-brand" href="/courses/">
-                    <div class="navbar-brand-logo" alt="{{ platform_name }}"></div>
+                    <img class="navbar-brand-logo" alt="Institution logo" src="{% static 'images/default-logo.png' %}"/>
```
cptvitamin: @iivic the alt for the log should be dynamic alt="{{ platform_name }} logo">

cptvitamin (changes requested):
Just a small change WRT alt and then 👍

iivic:
Merged into https://github.com/edx/ecommerce/pull/1213.

PR #1209, opened 2017-03-20
[ECOM-7533] Catch verification error on new receipt page

marjev:
JIRA ticket: https://openedx.atlassian.net/browse/ECOM-7533

Since the custom Exception is used only on the new receipt page, I removed it.
If Verification endpoint is unavailable, `False` is assumed.

@mjfrey, @vkaracic: Please review. Thanks!

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -19,7 +19,6 @@ from jsonfield.fields import JSONField
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
 
-from ecommerce.core.exceptions import VerificationStatusError
 from ecommerce.core.url_utils import get_lms_url
 from ecommerce.core.utils import log_message_and_raise_validation_error
 from ecommerce.courses.utils import mode_for_seat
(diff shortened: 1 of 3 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1209"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/core/models.py
```diff
@@ -552,7 +551,7 @@ def is_verified(self, site):
         except (ConnectionError, SlumberBaseException, Timeout):
             msg = 'Failed to retrieve verification status details for [{username}]'.format(username=self.username)
             log.exception(msg)
-            raise VerificationStatusError(msg)
+            return False
```
vkaracic, line 550: Maybe it would be better to catch the exception where it's used? It makes more sense to raise an exception here instead of assuming that the user is verified if the endpoint times out.

Either that or keep this and change the docstring on this method.

mjfrey: I am fine with not raising the exception if we never use it.  Just change the docstring.

marjev: I changed the docstring.

mjfrey (approved):

marjev:
@mjfrey Should I tag someone else to take a look or is it ready to merge?

vkaracic (approved):

mjfrey:
@marjev merge

PR #1210, opened 2017-03-20
Add redirect to receipt page after redeeming a free basket.

vkaracic:
When applying an enrollment coupon to a basket and redeeming it, the user would be redirected to the student dashboard. This PR changes that so that the user is redirected to the receipt page or the error page if something went wrong with placing the order (does NOT include fulfillment, this should be it's own ticket).
I've also changed the checkout error page design to be more inline with other error pages:
screenshot from 2017-03-20 14-28-02


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1210"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/coupons/views.py
```diff
@@ -214,11 +215,15 @@ def get(self, request):
...
-            return HttpResponseRedirect(reverse('basket:summary'))
-
-        return HttpResponseRedirect(request.site.siteconfiguration.student_dashboard_url)
+            try:
+                order = self.place_free_order(basket)
+                return HttpResponseRedirect(get_receipt_page_url(self.request.site.siteconfiguration, order.number))
+            except Exception as e:  # pylint: disable=broad-except
+                # If any error happens when creating an order the user is redirected
+                # to the checkout error page.
+                logger.exception('Failed to create an order with message: %s', e.message)
```
clintonb: There is no need to log the message when using `log.exception()`. That is already done for you. I would update this to `logger.exception('Failed to create free order for basket [%d].', basket.id)`.

ecommerce/templates/oscar/checkout/error.html
```diff
@@ -4,23 +4,27 @@
...
-            Please try to submit your payment again. If this problem persists, contact {{ start_link }}
-            {{ payment_support_email }}{{ end_link }}.
-            {% endblocktrans %}
-        {% endwith %}</p>
+    <div id="error-message">
+        <div class="container">
+            <div class="depth depth-2 message-error-content">
+                <h1>{% trans "Checkout Error" %}</h1>
+                <h3>{{ error }}</h3>
+                <p>{% blocktrans %} An error has occurred with your payment. <b>You have not been charged.</b>
```
clintonb: Avoid including HTML in strings marked for translation. This should be split: `{% trans "An error has occurred with your payment." %} <b>{% trans "You have not been charged." %}</b>`

ecommerce/coupons/tests/test_views.py
```diff
@@ -373,12 +378,25 @@ def create_and_test_coupon_and_return_code(
...
+    def assert_redemption_page_redirects(
+            self, expected_url, target=200, code=COUPON_CODE, consent_token=None, query_string=''
+    ):
         """ Verify redirect from redeem page to expected page. """
         self.request.user = self.user
         self.mock_enrollment_api(self.request, self.user, self.course.id, is_active=False, mode=self.course_mode)
         response = self.client.get(self.redeem_url_with_params(code=code, consent_token=consent_token))
-        self.assertRedirects(response, expected_url, status_code=302, target_status_code=target)
+
+        if query_string:
```
clintonb: If statements in test methods are a bit of a code smell for me. This would be better off as a separate assertion method, or build `expected_url` outside of this method.

ecommerce/coupons/tests/test_views.py
```diff
@@ -436,20 +454,47 @@ def test_basket_redirect_discount_code(self):
...
+    @ddt.unpack
+    def test_basket_redirect_enrollment_code(self, ecomm_receipt, query_string):
+        """ Verify the view redirects to the receipt page when an enrollment code is provided. """
+        self.toggle_ecommerce_receipt_page(ecomm_receipt)
         code = self.create_and_test_coupon_and_return_code(benefit_value=100, code='')
-        httpretty.register_uri(httpretty.GET, self.student_dashboard_url, status=status.HTTP_301_MOVED_PERMANENTLY)
+        receipt_page_url = get_receipt_page_url(self.site.siteconfiguration)
+        enrollment_api_url = '{root}enrollment'.format(root=self.site.siteconfiguration.enrollment_api_url)
+        httpretty.register_uri(httpretty.POST, enrollment_api_url, status=status.HTTP_200_OK)
+        httpretty.register_uri(httpretty.GET, receipt_page_url, status=status.HTTP_301_MOVED_PERMANENTLY)
```
clintonb: This is not necessary. Update your call to `assertRedirects()` to *not* follow the redirect.

ecommerce/coupons/tests/test_views.py
```diff
@@ -436,20 +454,47 @@ def test_basket_redirect_discount_code(self):
...
+    )
+    @ddt.unpack
+    def test_basket_redirect_enrollment_code(self, ecomm_receipt, query_string):
+        """ Verify the view redirects to the receipt page when an enrollment code is provided. """
+        self.toggle_ecommerce_receipt_page(ecomm_receipt)
         code = self.create_and_test_coupon_and_return_code(benefit_value=100, code='')
-        httpretty.register_uri(httpretty.GET, self.student_dashboard_url, status=status.HTTP_301_MOVED_PERMANENTLY)
+        receipt_page_url = get_receipt_page_url(self.site.siteconfiguration)
+        enrollment_api_url = '{root}enrollment'.format(root=self.site.siteconfiguration.enrollment_api_url)
+        httpretty.register_uri(httpretty.POST, enrollment_api_url, status=status.HTTP_200_OK)
```
clintonb: Why do you need to mock the Enrollment API?

ecommerce/coupons/tests/test_views.py
```diff
@@ -541,29 +595,40 @@ def test_enterprise_customer_successful_redemption(self):
...
             consent_token=consent_token,
+            query_string=query_string
         )
         last_request = httpretty.last_request()
         self.assertEqual(last_request.path, '/api/enrollment/v1/enrollment')
         self.assertEqual(last_request.method, 'POST')
 
     @httpretty.activate
-    def test_multiple_vouchers(self):
+    @ddt.data(
```
clintonb: I would update these tests to assume we are using Otto's receipt page since that is what we will be doing long-term. There isn't much value to testing the LMS-based receipt page since it's just a matter of changing the redirect URL.

clintonb (changes requested):

ecommerce/coupons/tests/test_views.py
vkaracic: Because order fulfillment sends a POST request to enroll the user.

vkaracic: Never mind.

ecommerce/coupons/views.py
```diff
@@ -214,11 +215,15 @@ def get(self, request):
...
         if basket.total_excl_tax == 0:
-            self.place_free_order(basket)
-        else:
-            return HttpResponseRedirect(reverse('basket:summary'))
-
-        return HttpResponseRedirect(request.site.siteconfiguration.student_dashboard_url)
+            try:
+                order = self.place_free_order(basket)
+                return HttpResponseRedirect(get_receipt_page_url(self.request.site.siteconfiguration, order.number))
+            except Exception as e:  # pylint: disable=broad-except
```
clintonb: `e` is not used and should be removed.

ecommerce/coupons/views.py
```diff
@@ -214,11 +215,15 @@ def get(self, request):
...
+                # If any error happens when creating an order the user is redirected
```
clintonb: This comment adds no value.

ecommerce/templates/oscar/checkout/error.html
```diff
@@ -4,23 +4,26 @@
...
-            {% blocktrans with end_link="</a>"|safe %}
-            Please try to submit your payment again. If this problem persists, contact {{ start_link }}
-            {{ payment_support_email }}{{ end_link }}.
-            {% endblocktrans %}
-        {% endwith %}</p>
+    <div id="error-message">
+        <div class="container">
+            <div class="depth depth-2 message-error-content">
+                <h1>{% trans "Checkout Error" %}</h1>
+                <h3>{{ error }}</h3>
```
clintonb: Where is this coming from? Is it translated?

clintonb (commented):
One last comment about the error message before approving.

ecommerce/templates/oscar/checkout/error.html
vkaracic: It's not used. It's used in the coupon error page and I've must've forgotten to remove it here.

vkaracic:
@clintonb comments addressed!

marjev (approved):

clintonb (approved):

vkaracic:
Forgot to squash commits... AGAIN! :rage: Sorry!

PR #1211, opened 2017-03-20
mattdrayer: Change from OAuth2 to JWT for API tokens

mattdrayer:
This PR changes the token type from OAuth2 (Bearer) to JWT for communication with the Open edX Enterprise Service

@clintonb, FYI


clintonb (approved):
DRY: Create a function, `get_enterprise_api_client(site)`, that contains this logic.

mattdrayer:
Good call -- I've made that change locally, and will push after checks are complete to see if anything else needs to be addressed

PR #1215, opened 2017-03-21
Remove the check of the LMS heartbeat

jibsheet:
We know that ecommerce won't fall out of the loadbalancer, but it still
puts load on the LMS (800 of the roughly 4,000 checks we see).  We're
also ratcheting down the frequency of heartbeats based on the discussion
here
https://openedx.atlassian.net/wiki/display/EdxOps/Design+for+scaling+White+Labels+and+not+drowning+in+health+checks

We forsee in the future a "is gunicorn alive" and "detailed heartbeat"
as discussed there.  At that time it might make sense to revert this and
then refactor it, but this is a much smaller/simpler change.

### DIFF
--- a/ecommerce/core/tests/test_views.py
+++ b/ecommerce/core/tests/test_views.py
@@ -8,76 +8,36 @@ from django.contrib.auth import get_user_model
 from django.core.urlresolvers import reverse
 from django.db import DatabaseError
 from django.test.utils import override_settings
-from requests import Response
-from requests.exceptions import RequestException
 from rest_framework import status
-from testfixtures import LogCapture
 
-from ecommerce.core.constants import Status, UnavailabilityMessage
+from ecommerce.core.constants import Status
 from ecommerce.tests.testcases import TestCase
 
 LOGGER_NAME = 'ecommerce.core.views'
 User = get_user_model()
 
 
-@mock.patch('requests.get')
 class HealthTests(TestCase):
     """Tests of the health endpoint."""
 
     def setUp(self):
         super(HealthTests, self).setUp()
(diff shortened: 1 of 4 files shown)


jibsheet:
@clintonb This came out of some @edx/devops design work around heartbeats (link in the PR message).  I spoke to @rlucioni and I know that ecom doesn't fall out of the load balancer when the LMS is down, but we'd like to remove the requests altogether.  Longer explanation in the PR message.

ecommerce/core/tests/test_views.py
```diff
@@ -86,8 +46,7 @@ def _assert_health(self, status_code, overall_status, database_status, lms_statu
         expected_data = {
             'overall_status': overall_status,
             'detailed_status': {
-                'database_status': database_status,
-                'lms_status': lms_status
+                'database_status': database_status
```
clintonb: Nit: trailing comma

clintonb (approved):

ecommerce/core/tests/test_views.py
jibsheet: I don't see it (from the diff database_status, => database_status) so should be valid json

clintonb: I meant add one. Again, it's a nit, so no big deal.

feanil: This part is a python dict and the output does a dict comparison of the parsed json response.  Since it's python we should leave the trailing comma.

jibsheet: doh, got it.  I'm so used to "trailing comma" meaning "JS(ON) is going to break now".

PR #1216, opened 2017-03-21
Removed systems checks for the update_assets management command

clintonb:
System checks require database access, which is not available when creating Docker images. Given that we don't need the checks, they have been removed.


clintonb:
FYI @edx/ecommerce @douglashall @saleem-latif

clintonb:
Related to https://github.com/edx/devstack/issues/68

PR #1219, opened 2017-03-23
Only log a warning if verification status failed.

mjfrey:
No need to log an exception when verification status fails calling LMS.  Changed to a warning.


mjfrey:
@vkaracic @marjev

vkaracic (approved):

PR #1220, opened 2017-03-27
[ECOM-7519] Make CVV help info accessible

marjev:
### JIRA ticket:
[No access to CVV help info](https://openedx.atlassian.net/browse/ECOM-7519)

### Reviewers:

- [x] @mjfrey (Code review)
- [x] @cptvitamin (a11y review)

### Screenshot:
screen shot 2017-03-28 at 13 20 09

### Sandbox:
https://ecommerce-coupons.sandbox.edx.org/basket/

You can log in using default staff user credentials.


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1220&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

cptvitamin:
@marjev is there any way we can get this to activate on button press only rather than on focus?  Also, the button itself needs some sort of accessible text.  I recommend adding a title attribute:
```title="Help with CVV"```

**Comment by marjev** [comment]
@cptvitamin I'll use Bootstrap popover instead, since Bootstrap tooltip uses the button's title as its content.

Bootstrap popover, however, is only triggered on click and doesn't rely solely on the title attribute. I'll ping you once the sandbox is updated so that you might preview the changes.

**Comment by marjev** [comment]
@cptvitamin Sandbox is updated. You can use it to preview the changes.

**Comment by cptvitamin** [comment]
@marjev this is still triggering on focus.  If this is going to be a button (which it should be) it should only trigger on click/keypress.  The button also still has not title attribute value, which serves as the accessible name.  Right now, this is a blank button.  The SR user would have no idea what it does.

**Comment by cptvitamin** [comment]
@marjev there are some technical deficiencies here that are likely due to using out of the box bootstrap features (it's not known for being accessible.)  So we either have to fix the bootstrap implementation or roll our own.  This gets even trickier since there is no good guidance or agreement on how to implement tooltip functionality.  I am going to rely on what I know to be tried and true WCAG conformance requirements, despite what the ARIA Practices guide says:

1. The visible tooltip should ONLY be triggered on mouse hover, or button press, NOT focus.
2. The visible tooltip should not exist in the DOM until mouse hover or button press.

I suggest the following:

```
<button aria-haspopup=true aria-controls="cvvtooltip" aria-expanded=false> <!-- the aria here goes against a few rules, but for all intents and purposes, it is the best compromise we have -->
  <i class="fa fa-question-circle" aria-hidden="true"></i>
  <span class="sr">Help with CVV</span> <!-- this is the accessible name for the button. It can also be expressed as a title attribute on the button (not data-title as you have it now)
</button>
<div id="cvvtooltip">The three last digits in the signature area on the back of your card. For American Express, it is the four digits on the front of the card.
</div><!-- this is hidden by default, but made visible on mouse hover or button press (not focus). Whenever this is visible, the aria-expanded attribute on the button needs to be toggled to true, and vice-versa -->
```

marjev:
@cptvitamin I did what you suggested in your last comment. Could you please take another look? I've updated the sandbox.

cptvitamin:
@marjev awesome!  This is great. I have one tiny little request and then you can merge it. Can you close the popup on blur?  If the user opens it with the keyboard and TABs away from it, it should close since its not a modal dialog box.  That's it.  Going to give you thumbs since that should be pretty straight forward to do without further testing 👍

cptvitamin (approved):
It would be great if you could close the popup on blur, but other than that, this is good to go.

marjev:
@mjfrey thumbs?

mjfrey (approved):

PR #1221, opened 2017-03-27
Demote duplicate PayPal payment message to a warning

rlucioni:
We can't prevent PayPal from sending us these duplicate notifications. We're already doing the right thing by ignoring duplicate notifications instead of trying to execute them.

@mikedikan we should be making more changes like these as we find errors that are really warnings.


mikedikan (approved):

schenedx (approved):
👍

PR #1222, opened 2017-03-28
Change SDN check address parameter to city only.

vkaracic:
The Treasury Dept. SDN check endpoint accepts as the address
parameter only the address or the city, but not both. Since
their documentation mentions only an example where the address
parameter is a city name, our SDN check call is changed to
send only the city as the address parameter accordingly.

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -171,7 +171,6 @@ define([
         function sdnCheck(event) {
             var first_name = $('input[name=first_name]').val(),
                 last_name = $('input[name=last_name]').val(),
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1222"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mjfrey (approved):

davec-edx:
Perhaps I'm missing the big picture.  But looking at the sdn.py file, it looks like it is using the address NOT the city.  However, shouldn't the js be sending everything and let the back end decide?

PR #1223, opened 2017-03-28
Change SDN interface address to city.

vkaracic:
Since we are using the city as an address parameter in the SDN check, it makes more sense semantically the parameter ``city`` throughout our code except for when making the call to the Treasury endpoint.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1223"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/payment/migrations/0017_auto_20170328_1445.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('payment', '0016_auto_20170227_1402'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
clintonb: You will need to manually call RenameField.

clintonb (changes requested):

vkaracic:
@clintonb comment addressed!

ecommerce/extensions/payment/migrations/0017_auto_20170328_1445.py
```diff
@@ -12,11 +12,12 @@ class Migration(migrations.Migration):
...
     operations = [
-        migrations.RemoveField(
+        migrations.RenameField(
             model_name='sdncheckfailure',
-            name='address',
+            old_name='address',
+            new_name='city',
         ),
-        migrations.AddField(
+        migrations.AlterField(
```
vkaracic: Not sure if this part is needed though. The ``city`` length on the form is 32.

clintonb (approved):

vkaracic:
I'll try :smile:

mjfrey (approved):

ecommerce/extensions/payment/utils.py
```diff
@@ -93,7 +93,7 @@ def search(self, name, address, country):
             'api_key': self.api_key,
             'type': 'individual',
             'name': unicode(name).encode('utf-8'),
-            'address': unicode(address).encode('utf-8'),
+            'address': unicode(city).encode('utf-8'),
```
davec-edx: add comment to explain why they 'address' key contains the city.  I know we've talked about this, but who will remember in 6 months to fill the next person in on it?

davec-edx:
I'd like to see a comment in the view to inform a future dev why it's address=city (commented on the exact line) but otherwise, looks good.

PR #1225, opened 2017-03-28
Updated verification URL on receipt page

clintonb:
The receipt page now points to the reverification page, which is not
course-specific. This is one step in making our receipt page more
applicable to orders containing multiple course seats.

LEARNER-278

### DIFF
--- a/ecommerce/extensions/checkout/tests/test_views.py
+++ b/ecommerce/extensions/checkout/tests/test_views.py
@@ -8,7 +8,6 @@ from django.core.urlresolvers import reverse
 from oscar.core.loading import get_model
 from oscar.test import newfactories as factories
 
-from ecommerce.core.url_utils import get_lms_url
 from ecommerce.coupons.tests.mixins import CourseCatalogMockMixin
 from ecommerce.extensions.checkout.exceptions import BasketNotFreeError
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
@@ -257,9 +256,7 @@ class ReceiptResponseViewTests(CourseCatalogMockMixin, LmsApiMockMixin, RefundTe
             'payment_method': None,
             'fire_tracking_events': False,
             'display_credit_messaging': False,
-            'verification_url': get_lms_url(
-                'verify_student/verify-now/{course_id}'.format(course_id=self.course.id)
-            ),
+            'verification_url': self.site.siteconfiguration.build_lms_url('verify_student/reverify'),
         }
 
         self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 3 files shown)


clintonb:
@edx/helio please review

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -257,9 +257,7 @@ def test_get_receipt_for_existing_order(self):
             'payment_method': None,
             'fire_tracking_events': False,
             'display_credit_messaging': False,
-            'verification_url': get_lms_url(
-                'verify_student/verify-now/{course_id}'.format(course_id=self.course.id)
-            ),
+            'verification_url': get_lms_url('verify_student/reverify'),
```
vkaracic: Perhaps build the LMS URL with the ``site_configuration`` object. 
https://github.com/edx/ecommerce/blob/master/ecommerce/core/url_utils.py#L66
https://github.com/edx/ecommerce/blob/master/ecommerce/core/url_utils.py#L15

ecommerce/extensions/checkout/views.py
```diff
@@ -193,11 +193,7 @@ def get_order_verification_context(self, order):
 
             if verified_course_id:
                 context.update({
-                    'verification_url': get_lms_url(
-                        'verify_student/verify-now/{course_id}'.format(
-                            course_id=verified_course_id
-                        )
-                    ),
+                    'verification_url': get_lms_url('verify_student/reverify'),
```
vkaracic: Same here.

vkaracic (changes requested):

ecommerce/extensions/checkout/tests/test_views.py
clintonb: Updated.

ecommerce/extensions/checkout/views.py
clintonb: Updated.

ecommerce/extensions/checkout/views.py
```diff
@@ -37,13 +36,16 @@ def dispatch(self, *args, **kwargs):
         return super(FreeCheckoutView, self).dispatch(*args, **kwargs)
 
     def get_redirect_url(self, *args, **kwargs):
-        basket = Basket.get_basket(self.request.user, self.request.site)
+        request = self.request
```
vkaracic, line 39: Why this change?

clintonb: Simplify access to the `request` object. Also, DRY.

vkaracic (approved):

marjev (approved):

PR #1226, opened 2017-03-28
Removed course-centric event tracking

clintonb:
Prior to this change course ID was treated special compared to other
event properties. This change treats the course ID as any other
property. Given that we have no course-centric pages, course ID will
almost never be included in the event properties going forward.

LEARNER-278

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -87,21 +87,17 @@ def audit_log(name, **kwargs):
     logger.info(message)
 
 
-def prepare_analytics_data(user, segment_key, course_id=None):
+def prepare_analytics_data(user, segment_key):
     """ Helper function for preparing necessary data for analytics.
 
     Arguments:
         user (User): The user making the request.
         segment_key (str): Segment write/API key.
-        course_id (str): The course ID.
 
     Returns:
         str: JSON object with the data for analytics.
     """
     data = {
-        'course': {
-            'courseId': course_id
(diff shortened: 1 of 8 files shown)


clintonb:
@edx/ecommerce @edx/helio please review

ecommerce/extensions/analytics/utils.py
```diff
@@ -87,7 +87,7 @@ def audit_log(name, **kwargs):
     logger.info(message)
 
 
-def prepare_analytics_data(user, segment_key, course_id=None):
+def prepare_analytics_data(user, segment_key):
     """ Helper function for preparing necessary data for analytics.
```
vkaracic, line 91: Please remove ``course_id`` from the docstring as well.

ecommerce/static/js/views/analytics_view.js
```diff
@@ -102,31 +89,16 @@ define([
...
-             * to Segment.
-             *
-             * @param eventType String event type.
-             */
+          /**
+           * Catch 'segment:track' events and create events and send
+           * to Segment.
+           *
+           * @param eventType String event type.
+           * @param properties Event properties
```
vkaracic: Nit: dot at the end.

vkaracic (approved):

ecommerce/static/js/views/analytics_view.js
clintonb: Not a sentence. No need for a period.

PR #1228, opened 2017-03-29
ENT-281: Update Account Activation message in Coupon flow

saleem-latif:
Hi @mattdrayer , @asadiqbal08 , @zubair-arbi ,

Kindly take a look at this PR, I have updated account activation message. I have updated account activation message in coupon redemption page.

__JIRA Ticket:__ [ENT-281](https://openedx.atlassian.net/browse/ENT-281)

screen shot 2017-03-29 at 4 03 50 pm

### DIFF
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -566,8 +566,8 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin
         self.mock_access_token_response()
 
         response = self.client.get(self.redeem_url_with_params())
-        msg = 'You need to activate your account in order to redeem this coupon.'
-        self.assertEqual(response.context['error'], msg)
+        self.assertEqual(response.context['course_name'], self.course.name)
+        self.assertEqual(response.context['user_email'], self.user.email)
 
 
 class EnrollmentCodeCsvViewTests(TestCase):
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -169,8 +169,11 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
         if not request.user.account_details(request).get('is_active'):
             return render(
                 request,
-                template_name,
-                {'error': _('You need to activate your account in order to redeem this coupon.')}
+                'edx/email_confirmation_required.html',
+                {
+                    'course_name': product.course and product.course.name,
+                    'user_email': request.user and request.user.email,
+                }
             )
 
         if request.user.is_user_already_enrolled(request, product):
--- a/ecommerce/static/sass/partials/views/_error.scss
+++ b/ecommerce/static/sass/partials/views/_error.scss
@@ -36,3 +36,38 @@
 .message-info-content {
   margin-left: 40px;
 }
+
+.account-activation-message {
+
+  .enrollment-message-container {
+    margin-bottom: 30px;
+  }
+
+  h2 {
+    font-size: 1rem;
+    font-weight: bold;
+  }
+  .course-title {
+    display: block;
+    font-weight: 100;
+    font-size: 24px;
+  }
+
+  .instruction-info {
+    font-size: 24px;
+    font-weight: 100;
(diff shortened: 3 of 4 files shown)


mattdrayer:
@chris-mike FYI on this PR

ecommerce/templates/edx/email_confirmation_required.html
```diff
@@ -0,0 +1,33 @@
...
+    {% include 'edx/partials/_student_navbar.html' %}
+{% endblock navbar %}
+
+
+{% block content %}
+    <div id="error-message" class="account-activation-message">
+        <div class="container">
+            <div class="enrollment-message-container">
+                <h2 class="enrollment-message">
+                  {% trans "You are enrolling in :" %}
```
mattdrayer: relocate space character --> `"You are enrolling in: "`

ecommerce/coupons/tests/test_views.py
```diff
@@ -566,8 +566,8 @@ def test_inactive_user_rejection(self):
         self.mock_access_token_response()
 
         response = self.client.get(self.redeem_url_with_params())
-        msg = 'You need to activate your account in order to redeem this coupon.'
-        self.assertEqual(response.context['error'], msg)
+        self.assertEqual(response.context['user_email'], self.user.email)
```
mattdrayer: Super Nit: Reorder these checks to align with parameter definition in views.py

ecommerce/coupons/views.py
```diff
@@ -169,8 +169,11 @@ def get(self, request):
         if not request.user.account_details(request).get('is_active'):
             return render(
                 request,
-                template_name,
-                {'error': _('You need to activate your account in order to redeem this coupon.')}
+                'edx/email_confirmation_required.html',
+                {
+                    'course_name': product.course and product.course.name,
```
mattdrayer, line 174: This may be a dumb question, but I don't understand the use of `and` here.  It's not being used as a string concatenator, right?  What we are essentially saying here is "If product.course is True and product.course.name is True", which should result in a value of True for the "course_name" parameter, right?  I must be missing something...

saleem-latif: @mattdrayer `and` here is a guard against `None` values of `product.course` and `request.user`. These statements are equivalent too
```
{
    'course_name': product.course.name if product.course else None,
    'user_email': request.user.email if request.user else None,
}
```

mattdrayer:
LGTM -- looks like a rebase is required -- one more +1 from @zubair-arbi or @asadiqbal08 and we should be good to merge after the rebase/rebuild.  👍

chris-mike:
👍  Looks good. Makes sense.

ecommerce/static/sass/partials/views/_error.scss
```diff
@@ -36,3 +36,38 @@
...
+  .fa.fa-envelope-o {
+    color: #b92167;
+    font-size: 26px;
+    margin-right: 8px;
+  }
+
+  .activate {
+    font-size: 1rem;
+    font-weight: bold;
+    margin: 0 0 0 0;
```
zubair-arbi: nit: use just `0` here.

ecommerce/coupons/views.py
zubair-arbi: @saleem-latif IMHO Its better to use the more readable format as the equivalent statements you have provided.

ecommerce/templates/edx/email_confirmation_required.html
```diff
@@ -0,0 +1,33 @@
+{% extends 'edx/base.html' %}
+{% load i18n %}
+{% load staticfiles %}
```
zubair-arbi: I think we can remove `staticfiles` template tag since we are not using any static files here

zubair-arbi:
looks good 👍

ecommerce/coupons/views.py
saleem-latif: @zubair-arbi We talked about this offline, just putting here for future reference, I am using `product.course and product.course.name` as it is more compact and acts as a guard against `None` values.

PR #1230, opened 2017-03-31
LEARNER-324 New basket endpoint to accept multiple SKUs

marjev:
### JIRA task:
[new basket endpoint to accept a list of SKUs to add to the basket](https://openedx.atlassian.net/browse/LEARNER-324)

### to-do:

- [x] write tests
- [x] update translation files


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1230&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/basket/app.py
```diff
@@ -16,6 +17,7 @@ def get_urls(self):
             url(r'^vouchers/(?P<pk>\d+)/remove/$', self.remove_voucher_view.as_view(), name='vouchers-remove'),
             url(r'^saved/$', login_required(self.saved_view.as_view()), name='saved'),
             url(r'^single-item/$', login_required(self.single_item_view.as_view()), name='single-item'),
+            url(r'^multiple-items/$', login_required(self.multiple_items_view.as_view()), name='multiple-items'),
```
mjfrey: We might want to just call this add-items

ecommerce/extensions/basket/views.py
```diff
@@ -86,7 +86,31 @@ def get(self, request):
...
+    def get(self, request):
+        partner = get_partner_for_site(request)
+
+        skus = request.GET.get('skus').split(',')
+        code = request.GET.get('code', None)
+
+        if not skus:
+            return HttpResponseBadRequest(_('No SKUs provided.'))
+
+        products = [sr.product for sr in StockRecord.objects.filter(partner=partner, partner_sku__in=skus)]
```
clintonb: `products = Product.objects.filter(stockrecords__partner=partner, stockrecords__partner_sku__in=skus)`

ecommerce/extensions/basket/views.py
```diff
@@ -86,7 +86,31 @@ def get(self, request):
...
+class BasketMultipleItemsView(View):
+    """
+    View that adds multiple products to a user's basket.
+    An additional coupon code can be supplied so the offer is applied to the basket.
+    """
+
+    def get(self, request):
+        partner = get_partner_for_site(request)
+
+        skus = request.GET.get('skus').split(',')
```
clintonb: `skus = request.GET.getlist('sku')`

The URL looks like `http://example.com/basket/add/?sku=8CF08E5&sku=26CD722`

ecommerce/extensions/basket/app.py
clintonb: `url(r'^add/$', login_required(self.multiple_items_view.as_view()), name='add-multi'),`

clintonb (changes requested):

ecommerce/extensions/basket/app.py
```diff
@@ -16,6 +17,7 @@ def get_urls(self):
             url(r'^vouchers/(?P<pk>\d+)/remove/$', self.remove_voucher_view.as_view(), name='vouchers-remove'),
             url(r'^saved/$', login_required(self.saved_view.as_view()), name='saved'),
             url(r'^single-item/$', login_required(self.single_item_view.as_view()), name='single-item'),
```
clintonb, line 19: Mark this view as deprecated. Either Oscar or Django has a decorator. Alternatively, just log the warning in the post() method.

marjev:
@clintonb @mjfrey 
Comments addressed. Can you please take another look?

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -256,6 +257,71 @@ def test_enrollment_api_failure(self, error):
...
+@ddt.ddt
+class BasketMultipleItemsViewTests(CourseCatalogTestMixin, TestCase):
+    """ BasketMultipleItemsView view tests. """
+    path = reverse('basket:add-multi')
+
+    def setUp(self):
+        super(BasketMultipleItemsViewTests, self).setUp()
+        self.user = self.create_user()
+        self.client.login(username=self.user.username, password=self.password)
+        self.products = []
```
clintonb: You do a lot of unnecessary work for every test despite every test not needing all of this data. Move these calls to the tests that need them. `setUp()` should be quite light.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -256,6 +257,71 @@ def test_enrollment_api_failure(self, error):
...
+        self.course = CourseFactory()
+        for __ in range(10):
+            self.create_product_with_stockrecord()
+        self.total_price = sum(product.stockrecords.first().price_excl_tax for product in self.products)
+
+    def create_product_with_stockrecord(self):
+        """ Helper method for creating a product and a stock record. """
+        product = ProductFactory(stockrecords__partner=self.partner)
+        self.products.append(product)
+        self.data.append(('sku', StockRecordFactory(product=product, partner=self.partner).partner_sku))
```
clintonb: `StockRecordFactory` creates a product. `ProductFactory` creates a stock record. There is no need to use both. I would use `StockRecordFactory` since you ultimately care about stock records' SKUs.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -256,6 +257,71 @@ def test_enrollment_api_failure(self, error):
...
+
+    def test_add_multiple_products_to_basket(self):
+        """ Verify the basket accepts multiple products. """
+        response = self.client.get(self.path, data=self.data)
```
clintonb: All of the overhead in `create_product_with_stockrecord` can be simplified here.

```python
stock_records = StockRecordFactory.create_batch(3, partner=self.partner)
qs = urllib.urlencode({'sku': [stock_record.partner_sku for stock_record in stock_records]}, True)
url = '{root}?{qs}'.format(root=self.path, qs=qs)
```

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -256,6 +257,71 @@ def test_enrollment_api_failure(self, error):
...
+        self.assertEqual(response.status_code, 303)
+
+        basket = Basket.objects.get(owner=self.user, site=self.site)
+        basket.strategy = strategy.Default()
```
clintonb: Why is this necessary?

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -256,6 +257,71 @@ def test_enrollment_api_failure(self, error):
...
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.content, 'No SKUs provided.')
+
+    def test_add_multiple_products_no_available_products(self):
+        """ Verify the Bad request exception is thrown when no skus are provided. """
+        response = self.client.get(self.path, data=[('sku', 1), ('sku', 2)])
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.content, 'Products with [1, 2] do not exist.')
+
+    @ddt.data(Voucher.SINGLE_USE, Voucher.MULTI_USE)
```
clintonb: The usage type shouldn't matter. Just use the default of `VoucherFactory`.

ecommerce/extensions/basket/views.py
```diff
@@ -124,7 +127,31 @@ def get(self, request):
...
+
+        skus = request.GET.getlist('sku')
+        code = request.GET.get('code', None)
+
+        if not skus:
+            return HttpResponseBadRequest(_('No SKUs provided.'))
+
+        products = Product.objects.filter(stockrecords__partner=partner, stockrecords__partner_sku__in=skus)
+        if not products:
+            return HttpResponseBadRequest(_('Products with [{skus}] do not exist.').format(skus=', '.join(skus)))
```
clintonb: Products with SKU(s) [] do not exist.

clintonb (commented):
Clean up the tests, and this will be good to go.

ecommerce/extensions/basket/tests/test_views.py
marjev: Basket middleware adds the strategy to the basket, but since this basket didn't go through the middleware, the strategy is missing. 
Because of that, an exception is thrown when I try to get the total price.
One way is to add the strategy manually.

The end goal was to check if the basket total price is equal to the sum of product prices.

ecommerce/extensions/basket/tests/test_views.py
marjev: I tried create_batch before, but ran into this error:
`IntegrityError: UNIQUE constraint failed: partner_partner.short_code`

I'll leave the method I created, but use some of the things you suggested in this comment.

marjev:
Comments addressed.

mjfrey (approved):

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -256,6 +256,62 @@ def test_enrollment_api_failure(self, error):
...
+
+    def test_add_multiple_products_to_basket(self):
+        """ Verify the basket accepts multiple products. """
+        products = ProductFactory.create_batch(3, stockrecords__partner=self.partner)
+        qs = urllib.urlencode({'sku': [product.stockrecords.first().partner_sku for product in products]}, True)
+        url = '{root}?{qs}'.format(root=self.path, qs=qs)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 303)
+
+        basket = response.wsgi_request.basket
```
clintonb, line 277: Interesting.

clintonb (approved):

PR #1231, opened 2017-04-01
Updated receipt page

clintonb:
- Updated verification message, removing reference to single course
- Receipt page now displays discounts for non-voucher-based offers

LEARNER-332


clintonb:
Prior to this change the receipt page did not display all conditional offer discounts, only those associated with vouchers.

**Voucher Discount**
screencapture-localhost-18130-checkout-receipt-1491059825132

**Conditional Offer**
screencapture-localhost-18130-checkout-receipt-1491059815590

clintonb:
@edx/helio please review

mjfrey (approved):

vkaracic:
Should UX review this as well?

clintonb:
Review for one new line item? Nah. I feel confident in pushing this as-is. If we need to iterate later, we can.

vkaracic (approved):
Coolio

PR #1233, opened 2017-04-03
ENT-270: Add enterprise message to offers page

zubair-arbi:
ENT-270

Add/update the existing offers page heading message from "`Select one of the following courses.`" to "`Welcome to edX. Please choose from the courses selected by your organization to start learning.`" for **all users**.

screen shot 2017-04-05 at 5 30 21 pm

**Previously:** 
screen shot 2017-03-28 at 3 25 37 pm

### DIFF
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -120,6 +120,11 @@ class CouponOfferView(TemplateView):
         context_data.update(get_enterprise_customer_consent_failed_context_data(self.request, voucher))
 
         if context_data and 'error' not in context_data:
+            context_data.update({
+                'offer_app_page_heading': _('Welcome to edX'),
+                'offer_app_page_heading_message': _('Please choose from the courses selected by your '
+                                                    'organization to start learning.')
+            })
             self.template_name = 'coupons/offer.html'
 
         return context_data
--- a/ecommerce/static/templates/_offer_course_list.html
+++ b/ecommerce/static/templates/_offer_course_list.html
@@ -1,7 +1,10 @@
 <section class="col-xs-12 col-md-8 <% if (courses.length == 1) { %>col-lg-8 col-lg-offset-2<% }else { %>col-lg-12<% } %>">
 
     <% if (courses.length > 1) { %>
-    <h3 class="number-of-products-header col-xs-12"><%- gettext('Select one of the following courses.') %></h3>
+        <div class="offer-page-heading">
+            <h3 class="heading"><%= pageHeading %></h3>
+            <p><%= pageHeadingMessage %></p>
+        </div>
     <% } %>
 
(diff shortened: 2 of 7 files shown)


ecommerce/static/templates/_offer_course_list.html
```diff
@@ -1,7 +1,9 @@
 <section class="col-xs-12 col-md-8 <% if (courses.length == 1) { %>col-lg-8 col-lg-offset-2<% }else { %>col-lg-12<% } %>">
 
     <% if (courses.length > 1) { %>
-    <h3 class="number-of-products-header col-xs-12"><%- gettext('Select one of the following courses.') %></h3>
+        <h3 class="offer-message-bar">
+            <%- gettext("Welcome to edX. Please choose from the courses selected by your organization to start learning.") %>
```
mattdrayer: I'd prefer to set this string in the view and pass it via the context.  Then we can write a test to cover the string and easily add conditional logic in the future without cluttering the template code.

zubair-arbi:
@mattdrayer I have updated the PR to provide offer page heading from offer view and added a js test for it. Please let me know if this is what you were expecting. Also I am waiting for the Griff's suggestion on the layout/design for the message box.

mattdrayer:
Thanks -- I would also like to see a python test checking for the presence of the parameter in the context -- thus will help to enforce the implicit contract between view and template

asadiqbal08:
👍 kindly rebase it

zubair-arbi:
@mattdrayer I have addressed your feedback, please review again.

mattdrayer:
Looks like another rebase is in order -- otherwise the PR LGTM -- 👍

@vkaracic and/or @marjev care to take a look?

ecommerce/coupons/views.py
```diff
@@ -120,6 +120,11 @@ def get_context_data(self, **kwargs):
         context_data.update(get_enterprise_customer_consent_failed_context_data(self.request, voucher))
 
         if context_data and 'error' not in context_data:
+            context_data.update({
+                'offer_app_page_heading': _('Welcome to edX'),
+                'offer_app_page_heading_message': _('Please choose from the courses selected by your '
+                                                    'organization to start learning.')
+            })
```
marjev, line 127: Can you add this text in the template directly?
You can show it if there wasn't any errors in the context data.

marjev (changes requested):

ecommerce/coupons/views.py
zubair-arbi: @marjev This is intentional. I have made this change as pr suggestion by @mattdrayer, this way we have covered the change in unittest and we can easily add conditional logic in the future without cluttering the template code.

marjev (approved):

PR #1234, opened 2017-04-03
Upgraded to jQuery 3.2.1

clintonb:
This upgrade resolves a security vulnerability. The base template has
been added/modified to ensure that all Oscar and edX pages use the same
version of jQuery (installed via Bower).

LEARNER-455


clintonb:
FYI @edx/fedx-team

PR #1218, opened 2017-03-22
[WIP] ENT-282: Use custom Benefit model for Enterprise Customer linking via coupons

mattdrayer:
This pull request revises the current implementation for linking Enterprise Customers to Coupons from a Range-based approach to a custom Benefit-based approach.  See http://django-oscar.readthedocs.io/en/releases-1.4/howto/how_to_create_a_custom_benefit.html for more information

JIRA: [ENT-282](https://openedx.atlassian.net/browse/ENT-282)

Because this is a first-time effort with expertise and context distributed across multiple teams, our plan is to construct the implementation as a series of commits on this PR to allow for fast feedback cycles and asynchronous discussion.  The first commit is a placeholder to get the conversation started.  We will add to the change set as new things are learned.

@mjfrey @clintonb @haikuginger @zubair-arbi @saleem-latif @asadiqbal08 

Please tag additional people as needed.

PR #1236, opened 2017-04-04
Confirm that a discount has a corresponding voucher for enterprise scenarios

mattdrayer:
@clintonb @mjfrey -- as discussed in https://github.com/edx/ecommerce/pull/1127#discussion_r109728419 I've added a check for a voucher object on the discount.  Let me know if this will address the short term need and we'll continue the work on https://github.com/edx/ecommerce/pull/1218.

### DIFF
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -213,10 +213,11 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
                 # Collect the EnterpriseCustomer UUID from the coupon, if any.
                 enterprise_customer_uuid = None
                 for discount in order.discounts.all():
-                    enterprise_customer_uuid = discount.voucher.benefit.range.enterprise_customer
-                    if enterprise_customer_uuid is not None:
-                        data['enterprise_course_consent'] = True
-                        break
+                    if discount.voucher:
+                        enterprise_customer_uuid = discount.voucher.benefit.range.enterprise_customer
+                        if enterprise_customer_uuid is not None:
+                            data['enterprise_course_consent'] = True
+                            break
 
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/fulfillment/modules.py
```diff
@@ -213,10 +213,11 @@ def fulfill_product(self, order, lines):
                 # Collect the EnterpriseCustomer UUID from the coupon, if any.
                 enterprise_customer_uuid = None
                 for discount in order.discounts.all():
-                    enterprise_customer_uuid = discount.voucher.benefit.range.enterprise_customer
-                    if enterprise_customer_uuid is not None:
-                        data['enterprise_course_consent'] = True
-                        break
+                    if getattr(discount, "voucher"):
```
clintonb: I believe `discount.voucher` works here.

clintonb (approved):

ecommerce/extensions/fulfillment/modules.py
mattdrayer: I actually was going to revise to `hasattr()`.  I figured that the `discount.voucher` in a no-voucher scenario would raise an `AttributeError`?

clintonb: `discount.voucher` will either be a `Voucher` or `None. See https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/order/abstract_models.py#L1133-L1139.

mattdrayer: Ah, nice!  Didn't realize it was a defined property -- TIL 😄 -- okay, will revise, then figure out how to write a test featuring a discount w/ no voucher.

mattdrayer:
@clintonb one more look please and then I'll merge

clintonb (approved):

clintonb:
@mattdrayer please rebase and merge.

PR #1238, opened 2017-04-05
[LEARNER-326] update analytics to support multi line basket

marjev:
Reviewers:
@edx/helio 

JIRA ticket:
[update analytics to support multi line basket](https://openedx.atlassian.net/browse/LEARNER-326)

### DIFF
--- a/ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
+++ b/ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
@@ -114,7 +114,6 @@
                                 data-track-event="edx.bi.ecommerce.basket.payment_selected"
                                 data-track-category="checkout"
                                 data-processor-name="paypal"
-                                data-course-id="{{ course_key }}"
                                 class="btn payment-button payment-processor-paypal"
                                 value="paypal"
                                 title="{% trans "Pay with PayPal" %}"
@@ -199,7 +198,6 @@
                        data-track-type="click"
                        data-track-event="edx.bi.ecommerce.basket.free_checkout"
                        data-track-category="checkout"
-                       data-course-id="{{ course_key }}"
                        class="btn btn-primary btn-large col-sm-12 col-xs-12">
                         {% trans "Place Order" %}
                     </a>
@@ -209,7 +207,6 @@
                             data-track-event="edx.bi.ecommerce.basket.payment_selected"
                             data-track-category="checkout"
                             data-processor-name="{{ client_side_payment_processor.NAME }}"
-                            data-course-id="{{ course_key }}"
                             data-track-checkout-type="client_side">
                       {% trans "Place Order" %}
                     </button>
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -354,7 +354,6 @@ class BasketSummaryView(BasketView):
         context_updates, lines_data = self._process_basket_lines(lines)
         context.update(context_updates)
 
-        course_key = lines_data[0].get('course_key') if len(lines) == 1 else None
         user = self.request.user
         context.update({
             'analytics_data': prepare_analytics_data(
@@ -383,7 +382,6 @@ class BasketSummaryView(BasketView):
             total_benefit = None
 
         context.update({
-            'course_key': course_key,
             'formset_lines_data': zip(formset, lines_data),
             'free_basket': context['order_total'].incl_tax == 0,
             'homepage_url': get_lms_url(''),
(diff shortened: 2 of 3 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1238"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -113,7 +113,7 @@ <h2 id="payment-method-region" class="title">{% trans "select payment method" %}
                                 data-track-event="edx.bi.ecommerce.basket.payment_selected"
                                 data-track-category="checkout"
                                 data-processor-name="paypal"
-                                data-course-id="{{ course_key }}"
+                                data-course-ids="{{ course_keys }}"
```
clintonb: Nothing in our code base actually uses this attribute. All of this should go away.

clintonb (changes requested):

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
clintonb: "All of this" includes the data attributes and the context variable.

marjev:
@clintonb I removed unused code.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -110,12 +110,7 @@ <h2 id="payment-method-region" class="title">{% trans "select payment method" %}
                     </div>
                     <div id="payment-processor" class="payment-buttons col-md-3 col-sm-5 col-xs-6" data-basket-id="{{ basket.id }}">
                         {# Translators: Do NOT translate the name PayPal. #}
-                        <button data-track-type="click"
-                                data-track-event="edx.bi.ecommerce.basket.payment_selected"
-                                data-track-category="checkout"
-                                data-processor-name="paypal"
-                                data-course-id="{{ course_key }}"
```
clintonb: The tracking is used. It's only this data attribute that is not used.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -196,21 +191,11 @@ <h2 id="billing-information-region" class="title">{% trans "billing information"
                 <div class="payment-button col-sm-6 col-sm-offset-6 col-xs-12">
                 {% if free_basket %}
                     <a href="{% url 'checkout:free-checkout' %}"
-                       data-track-type="click"
-                       data-track-event="edx.bi.ecommerce.basket.free_checkout"
-                       data-track-category="checkout"
-                       data-course-id="{{ course_key }}"
```
clintonb: This is the only line that should be removed.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -196,21 +191,11 @@ <h2 id="billing-information-region" class="title">{% trans "billing information"
...
                        class="btn btn-primary btn-large col-sm-12 col-xs-12">
                         {% trans "Place Order" %}
                     </a>
                 {% else %}
-                    <button id="payment-button" type="submit" class="btn btn-primary btn-large col-sm-12 col-xs-12"
-                            data-track-type="click"
-                            data-track-event="edx.bi.ecommerce.basket.payment_selected"
-                            data-track-category="checkout"
-                            data-processor-name="{{ client_side_payment_processor.NAME }}"
-                            data-course-id="{{ course_key }}"
```
clintonb: This is the only line that should be removed.

clintonb (changes requested):
Whoa. *Tracking* is used. We simply don't use the `data-course-id` attribute.

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
vkaracic: Are you sure about that? I vaguely remember ``course-id`` being a part of the analytics event on Segment.

clintonb: Show me. We have no code that accesses this attribute. I have found no references in Google Tag Manager.

clintonb: I any case, this is not the way we should be adding properties to events. Just as we should never assume a basket contains one item, we should never assume it contains a specific type of product.

marjev:
I removed `data-course-id` attribute and `course_key` context variable.
Everything else is back.

clintonb (approved):
Squash your commits.

marjev:
@vkaracic thumbs?

vkaracic (approved):

PR #1240, opened 2017-04-05
Updated Order and Refund Events

clintonb:


clintonb:
We should use Segment's semantic events to avoid future confusion:

* https://segment.com/docs/spec/ecommerce/v2/
* https://segment.com/docs/integrations/google-analytics/

clintonb:
@edx/ecommerce @edx/helio please review

clintonb:
Google Tag Manager has been updated to recognize the "Order Completed" event in addition to "Completed Order" and "Completed Purchase".

ecommerce/static/js/pages/receipt_page.js
```diff
@@ -9,7 +9,7 @@ define([
         'use strict';
 
         function trackPurchase(order_id, total_amount, currency) {
-            window.analytics.track('Completed Order', {
+            window.analytics.track('Order Completed', {
```
mjfrey, line 12: Do we need to tell Adley about this change?

mjfrey (approved):

ecommerce/static/js/pages/receipt_page.js
clintonb: I have no idea.

mjfrey: I would just let BA know about the change in case it affects the reporting.

douglashall:
@clintonb What needs to be updated in GA/GTM in order to support this change? There are WL containers and properties in play that we need to consider.

ecommerce/static/js/pages/receipt_page.js
clintonb: Agreed. I have notified them via HipChat.

clintonb:
@douglashall I made the following changes in GTM:

1. Renamed the "Completed Purchase" trigger to "Order Completed" (not required, but good for consistency).
2. Added a regex for the event name: `Completed (Order|Purchase)|Order Completed`

PR #1241, opened 2017-04-05
Upgrade Django to 1.9.13

rlucioni:
@edx/learner FYI

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,5 +1,5 @@
 analytics-python==1.1.0
-Django==1.9.12
+django==1.9.13
 django-compressor==2.1.1
 django-crispy-forms==1.6.1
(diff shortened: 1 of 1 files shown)


clintonb (approved):

rlucioni:
Since when are ecommerce tests so flaky?

clintonb:
@rlucioni which test? I recently un-skipped a refund test.

rlucioni:
@clintonb I should have made a note; I don't remember which exact tests were failing. One was a JS test. I restarted the build twice and the history seems to be lost. If I see it again I'll make a ticket.

rlucioni:
@clintonb [LEARNER-491](https://openedx.atlassian.net/browse/LEARNER-491).

PR #1242, opened 2017-04-06
Sort countries alphabetically by their full name.

vkaracic:
After [updating](https://github.com/edx/ecommerce/commit/b78a6cbf60502c2f96d6feb5d53f58f68050d2fd) the ``pycountry`` package the country list wasn't in alphabetical order:
screenshot from 2017-04-06 13-27-18

This PR sorts the countries back again:
screenshot from 2017-04-06 13-35-50

### DIFF
--- a/ecommerce/extensions/payment/tests/test_forms.py
+++ b/ecommerce/extensions/payment/tests/test_forms.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import ddt
+import pycountry
 from oscar.test import factories
 
 from ecommerce.extensions.payment.forms import PaymentForm
@@ -8,9 +9,9 @@ from ecommerce.tests.testcases import TestCase
 
 
 @ddt.ddt
-class CyberSourceSubmitFormTests(TestCase):
+class PaymentFormTests(TestCase):
     def setUp(self):
-        super(CyberSourceSubmitFormTests, self).setUp()
+        super(PaymentFormTests, self).setUp()
         self.user = self.create_user()
         self.basket = factories.create_basket()
         self.basket.owner = self.user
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1242&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb (commented):
Test?

vkaracic:
@clintonb test added!

ecommerce/extensions/payment/tests/test_forms.py
```diff
@@ -1,14 +1,19 @@
 from __future__ import unicode_literals
 
+import copy
```
clintonb: This should be separated, since it's part of Python.

ecommerce/extensions/payment/tests/test_forms.py
```diff
@@ -1,14 +1,19 @@
...
+import pycountry
 from oscar.test import factories
 
-from ecommerce.extensions.payment.forms import PaymentForm
+from ecommerce.extensions.payment.forms import PaymentForm, country_choices
 from ecommerce.tests.testcases import TestCase
 
 
 @ddt.ddt
 class CyberSourceSubmitFormTests(TestCase):
```
clintonb: There is no class named `CyberSourceSubmitForm`. This should be named `PaymentFormTests`.

ecommerce/extensions/payment/tests/test_forms.py
```diff
@@ -92,3 +97,25 @@ def test_postal_code_validation_outside_north_america(self):
 
         invalid_postal_code = ''.join(['a' for __ in range(0, 11)])
         self.assert_form_not_valid(country='IN', postal_code=invalid_postal_code)
+
+    def tearDown(self):
```
clintonb: This is unnecessarily complicated. Try this:

```python
    def test_countries_sorting(self):
        """ Verify the country choices are sorted by country name. """
        data = self._generate_data()
        form = PaymentForm(user=self.user, data=data)
        expected = sorted([(country.alpha_2, country.name) for country in pycountry.countries], key=lambda x: x[1])
        actual = list(form.fields['country'].choices)
        actual.pop(0)   # Remove the "Choose country" placeholder
        self.assertEqual(actual, expected)

```

clintonb (changes requested):

vkaracic:
@clintonb comments addressed!

clintonb (approved):

PR #1247, opened 2017-04-11
[ENT-303] Allow coupon API to be filtered by EnterpriseCustomer UUID; allow catalog-based coupons to be linked to EnterpriseCustomers

haikuginger:
**Description:**

This pull request adds two feature:

First, the ability to link a catalog-based coupon to an EnterpriseCustomer. We had initially disallowed this, but it's become apparent that this is a necessary use case, so this pull request removes the Javascript that disallowed it.

Second, the ability to filter, in the API, coupons by the EnterpriseCustomer that they're linked to. This is accomplished by passing an `enterprise_customer` query parameter containing the UUID of the EnterpriseCustomer in question.

**JIRA:** [ENT-303](https://openedx.atlassian.net/browse/ENT-303)

**Dependencies:** None

**Merge deadline:** 24 April 2017

**Installation instructions:** Standard ecommerce installation

**Testing instructions:**

1. In the coupon creation view, verify that you can create a coupon tied to a catalog that is also tied to an EnterpriseCustomer.
2. In the coupon list API view, verify that adding an `enterprise_customer` query parameter that is a valid UUID filters the coupons down to only those coupons that are linked to that EnterpriseCustomer.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -476,38 +476,49 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         response = self.client.get(COUPONS_LINK)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         coupon_data = json.loads(response.content)['results'][0]
-        self.assertEqual(coupon_data['title'], self.data['title'])
-        self.assertEqual(coupon_data['category']['name'], self.data['category']['name'])
-        self.assertEqual(coupon_data['client'], self.data['client'])
+        self.assert_coupon_data(coupon_data)
 
     def test_list_coupons_by_enterprise_customer(self):
         """
         Test the behavior of the coupon list view when filtering by EnterpriseCustomer UUID
         """
         ec_uuid = 'ec1f642c-3e5f-4e30-bdba-2d683d7bcba9'
+
         # First, test that when we filter by a UUID that isn't present, we don't get a result.
         response = self.client.get(COUPONS_LINK, {'enterprise_customer': ec_uuid})
         self.assertEqual(len(json.loads(response.content)['results']), 0)
+        
         # Next, test that when we don't pass a real UUID, we don't attempt to filter.
         response = self.client.get(COUPONS_LINK, {'enterprise_customer': 'fake_uuid'})
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(json.loads(response.content)['results']), 1)
+        # Validate that the contents of that single coupon are what we expect
         coupon_data = json.loads(response.content)['results'][0]
-        self.assertEqual(coupon_data['title'], self.data['title'])
-        self.assertEqual(coupon_data['category']['name'], self.data['category']['name'])
-        self.assertEqual(coupon_data['client'], self.data['client'])
+        self.assert_coupon_data(coupon_data)
+        
         # Finally, add a new coupon that has that UUID, and test that we only get that back when filtering.
         self.data['enterprise_customer'] = {'id': ec_uuid}
+        # Create the coupon...
         self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        # Check that when we filter by EnterpriseCustomer, we get that single coupon
         response = self.client.get(COUPONS_LINK, {'enterprise_customer': ec_uuid})
         self.assertEqual(len(json.loads(response.content)['results']), 1)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+        # Validate the contents of that coupon to match what we created
         coupon_data = json.loads(response.content)['results'][0]
+        self.assert_coupon_data(coupon_data)
+        
+        # And test that without filtering, we get both coupons.
+        response = self.client.get(COUPONS_LINK)
+        self.assertEqual(len(json.loads(response.content)['results']), 2)
+
+    def assert_coupon_data(self, coupon_data):
+        """
+        Verify the coupon data matches what we have stored.
+        """
         self.assertEqual(coupon_data['title'], self.data['title'])
         self.assertEqual(coupon_data['category']['name'], self.data['category']['name'])
         self.assertEqual(coupon_data['client'], self.data['client'])
-        response = self.client.get(COUPONS_LINK)
-        # And test that without filtering, we get both coupons.
-        self.assertEqual(len(json.loads(response.content)['results']), 2)
 
     def test_list_and_details_endpoint_return_custom_code(self):
         """Test that the list and details endpoints return the correct code."""
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -480,6 +480,35 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         self.assertEqual(coupon_data['category']['name'], self.data['category']['name'])
         self.assertEqual(coupon_data['client'], self.data['client'])
 
+    def test_list_coupons_by_enterprise_customer(self):
+        """
+        Test the behavior of the coupon list view when filtering by EnterpriseCustomer UUID
+        """
+        ec_uuid = 'ec1f642c-3e5f-4e30-bdba-2d683d7bcba9'
+        # First, test that when we filter by a UUID that isn't present, we don't get a result.
+        response = self.client.get(COUPONS_LINK, {'enterprise_customer': ec_uuid})
+        self.assertEqual(len(json.loads(response.content)['results']), 0)
+        # Next, test that when we don't pass a real UUID, we don't attempt to filter.
+        response = self.client.get(COUPONS_LINK, {'enterprise_customer': 'fake_uuid'})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        coupon_data = json.loads(response.content)['results'][0]
+        self.assertEqual(coupon_data['title'], self.data['title'])
+        self.assertEqual(coupon_data['category']['name'], self.data['category']['name'])
+        self.assertEqual(coupon_data['client'], self.data['client'])
+        # Finally, add a new coupon that has that UUID, and test that we only get that back when filtering.
+        self.data['enterprise_customer'] = {'id': ec_uuid}
+        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        response = self.client.get(COUPONS_LINK, {'enterprise_customer': ec_uuid})
+        self.assertEqual(len(json.loads(response.content)['results']), 1)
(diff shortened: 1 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @haikuginger! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1247"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -53,6 +54,28 @@ def get_serializer_class(self):
             return CouponListSerializer
         return CouponSerializer
 
+    def get_queryset(self):
+        """
+        Handle cases where we need to filter by fields that aren't
+        natively supported on the Coupon object.
+        """
+        ec_filter = self.request.GET.get('enterprise_customer', None)
```
vkaracic: You don't need ``None`` here.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -53,6 +54,28 @@ def get_serializer_class(self):
...
+        objs = super(CouponViewSet, self).get_queryset()
+        if ec_filter is not None:
```
vkaracic: Can you just say ``if ec_filter:``?

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -480,6 +480,35 @@ def test_list_coupons(self):
...
+        response = self.client.get(COUPONS_LINK, {'enterprise_customer': ec_uuid})
+        self.assertEqual(len(json.loads(response.content)['results']), 1)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        coupon_data = json.loads(response.content)['results'][0]
+        self.assertEqual(coupon_data['title'], self.data['title'])
+        self.assertEqual(coupon_data['category']['name'], self.data['category']['name'])
+        self.assertEqual(coupon_data['client'], self.data['client'])
+        response = self.client.get(COUPONS_LINK)
+        # And test that without filtering, we get both coupons.
+        self.assertEqual(len(json.loads(response.content)['results']), 2)
```
vkaracic: You should add a few blank lines to make the code more legible.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -480,6 +480,35 @@ def test_list_coupons(self):
...
+        # Finally, add a new coupon that has that UUID, and test that we only get that back when filtering.
+        self.data['enterprise_customer'] = {'id': ec_uuid}
+        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+        response = self.client.get(COUPONS_LINK, {'enterprise_customer': ec_uuid})
+        self.assertEqual(len(json.loads(response.content)['results']), 1)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        coupon_data = json.loads(response.content)['results'][0]
+        self.assertEqual(coupon_data['title'], self.data['title'])
+        self.assertEqual(coupon_data['category']['name'], self.data['category']['name'])
+        self.assertEqual(coupon_data['client'], self.data['client'])
```
vkaracic: If I understand this test correctly you get the same thing with a fake UUID as you do with a valid one?

marjev (approved):
I would agree with Vedran on adding a few additional blank lines just to increase the readability of the test.

I would suggest separating different cases with a single new line. Otherwise LGTM.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -480,6 +480,35 @@ def test_list_coupons(self):
...
+        response = self.client.get(COUPONS_LINK)
+        # And test that without filtering, we get both coupons.
```
itsjeyd: @haikuginger I think we should move this comment above the `GET` in the previous line.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -480,6 +480,35 @@ def test_list_coupons(self):
...
+        # First, test that when we filter by a UUID that isn't present, we don't get a result.
+        response = self.client.get(COUPONS_LINK, {'enterprise_customer': ec_uuid})
+        self.assertEqual(len(json.loads(response.content)['results']), 0)
+        # Next, test that when we don't pass a real UUID, we don't attempt to filter.
+        response = self.client.get(COUPONS_LINK, {'enterprise_customer': 'fake_uuid'})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        coupon_data = json.loads(response.content)['results'][0]
+        self.assertEqual(coupon_data['title'], self.data['title'])
+        self.assertEqual(coupon_data['category']['name'], self.data['category']['name'])
+        self.assertEqual(coupon_data['client'], self.data['client'])
```
itsjeyd: @haikuginger Since these assertions are duplicated below it might make sense to move them to a separate method, something like `assert_coupon_data`.

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
```diff
@@ -201,16 +201,6 @@ define([
...
-                it('enterprise customer dropdown should be hidden when a catalog is selected', function () {
-                    view.$('#single-course').prop('checked', true).trigger('change');
-                    expect(SpecUtils.formGroup(view, '[name=enterprise_customer]')).toBeVisible();
-
-                    view.$('#catalog').prop('checked', true).trigger('change');
-                    expect(SpecUtils.formGroup(view, '[name=enterprise_customer]')).not.toBeVisible();
-
-                    view.$('#multiple-courses').prop('checked', true).trigger('change');
-                    expect(SpecUtils.formGroup(view, '[name=enterprise_customer]')).toBeVisible();
-                });
```
itsjeyd: @haikuginger IIUC, we now want the menu for selecting an enterprise customer to be visible at all times. Would it be worth keeping this test and just changing `.not.toBeVisible()` in L209 to `.toBeVisible()`, to make sure any changes to the current behavior wouldn't go unnoticed?

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -53,6 +54,28 @@ def get_serializer_class(self):
...
+                # If we didn't get a valid UUID for enterprise_customer, then skip the filtration
+                pass
+            else:
+                # We got a valid, UUID-like argument for enterprise_customer, for which we
+                # have to do an excessively complicated query to filter on.
+                objs = objs.filter(
+                    coupon_vouchers__vouchers__offers__condition__range__enterprise_customer=ec_filter
+                ).distinct()
+        return objs
+
```
itsjeyd, line 78: @haikuginger Thanks for adding detailed comments about what's happening to this method :)

itsjeyd (changes requested):
@haikuginger Made a first pass over the code. Just needs a little bit of polish; the approach looks good to me.

clintonb (changes requested):
Your team needs to finish https://github.com/edx/ecommerce/pull/1218 before you do any additional coupon work. You're building atop an incorrect model.

itsjeyd:
@haikuginger I tested the changes according to the instructions in the PR description and they are working fine. Couldn't associate a coupon with a catalog because ecommerce is currently failing to retrieve course catalog data from the Course Discovery API, but that's not related to the changes that this PR introduces. The "Enterprise Customer" option is available for any of "Single course", "Multiple courses", "Catalog".

ecommerce/extensions/api/v2/views/coupons.py
haikuginger: No; that'd match blank strings, which aren't valid filters. To be fair, that'd get caught by checking if it's a valid UUID, but I prefer to be explicit.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
haikuginger: Agreed! Legible code is always good.

ecommerce/extensions/api/v2/views/coupons.py
haikuginger: Right you are!

ecommerce/extensions/api/v2/tests/views/test_coupons.py
haikuginger: No. Initially, we have one coupon, and so calling the list API will return that single coupon. When calling with an invalid UUID, we get that single coupon back, since an invalid UUID can't filter anything. We then add an _additional_ coupon that has a UUID attached, and attempt to filter by that valid UUID, resulting in a single (different) result.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
haikuginger: Sure!

ecommerce/extensions/api/v2/tests/views/test_coupons.py
haikuginger: Sure; that makes sense!

ecommerce/static/js/test/specs/views/coupon_form_view_spec.js
haikuginger: We definitely could; that's a good idea.

ecommerce/extensions/api/v2/views/coupons.py
clintonb: Filtering should be done in `filters.py` with `django_filters. See https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/filters.py.

haikuginger:
@clintonb, I think you're probably right that we should prioritize that work. However, we'd still like to get this across the line. Would this change be acceptable to you if we moved the actual query to filter by EnterpriseCustomer UUID into the `ecommerce.enterprise` module?

Doing so would isolate where we need to make changes during the refactor to just that module. Otherwise, the needed behavior should remain the same - we need to get a list of the coupons that are associated with a particular EnterpriseCustomer, whether we do that by doing a database lookup to the related Range, or if we locate an Enterprise-linked Offer and get the linked coupons that way.

haikuginger:
@marjev @vkaracic @itsjeyd, I've addressed your comments. @clintonb, I haven't made any changes to the filtration logic yet, pending a discussion on if and how we should go ahead with implementing this right now, or wait on #1218.

clintonb:
This cannot be merged until #1218 is addressed. Building atop the incorrect model will only make fixing that model more difficult.

haikuginger:
@clintonb, okay! I actually just got off a call with Griff, and we've agreed to defer this until later. I'm going to go ahead and close this PR, as it's no longer necessary.

@itsjeyd, @vkaracic, @marjev, thank you for your reviews!

PR #1248, opened 2017-04-12
[LEARNER-508] Add condition offer API endpoint.

vkaracic:
LEARNER-508

@edx/helio FYI.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1248&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/api/v2/views/offers.py
```diff
@@ -0,0 +1,91 @@
...
+            api = self.request.site.siteconfiguration.course_catalog_api_client
+            program = api.programs(program_uuid).get()
+            cache.set(cache_key, program, settings.PROGRAM_CACHE_TIMEOUT)
+        return program
+
+    def create_range(self, program):
+        """ Create an Oscar Range for the program. """
+        stock_records = []
+        seat_types = program['applicable_seat_types']
+        for course in program['courses']:
```
mjfrey: We would also need to remove and excluded course_runs.

ecommerce/extensions/api/v2/views/offers.py
```diff
@@ -0,0 +1,91 @@
...
+Range = get_model('offer', 'Range')
+StockRecord = get_model('partner', 'StockRecord')
+
+
+class OfferViewSet(viewsets.ModelViewSet):
+    permission_classes = (IsAuthenticated, IsAdminUser)
+    queryset = ConditionalOffer.objects.filter(offer_type=ConditionalOffer.SITE)
+    serializer_class = ConditionalOfferSerialitzer
+
+    def create(self, request, *args, **kwargs):
```
mjfrey: Given a conditional offer does not have an expiration date like Vouchers we need a way to delete / disable these.

ecommerce/extensions/api/v2/views/offers.py
```diff
@@ -0,0 +1,91 @@
...
+            return Response(self.serializer_class(offer).data, status=status.HTTP_201)
+
+    def get_program(self, program_uuid):
+        cache_key = hashlib.md5(
+            'program_{uuid}'.format(uuid=program_uuid)
+        ).hexdigest()
+        program = cache.get(cache_key)
+        if not program:
+            api = self.request.site.siteconfiguration.course_catalog_api_client
+            program = api.programs(program_uuid).get()
```
mjfrey: need exception handling

mjfrey (changes requested):

ecommerce/extensions/api/v2/views/offers.py
vkaracic: Condition offers have expiration dates: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/abstract_models.py#L91

ecommerce/extensions/api/v2/views/offers.py
vkaracic: Those are automatically excluded on the programs API endpoint.

ecommerce/extensions/api/v2/views/offers.py
mjfrey: perfect.  We would need to just add those to the endpoint like coupons.

ecommerce/extensions/api/serializers.py
```diff
@@ -717,3 +718,8 @@ class ProviderSerializer(serializers.Serializer):  # pylint: disable=abstract-me
     status_url = serializers.CharField()
     thumbnail_url = serializers.CharField()
     url = serializers.CharField()
+
+
+class ConditionalOfferSerialitzer(serializers.ModelSerializer):
```
clintonb: `ConditionalOfferSerializer`

ecommerce/extensions/api/v2/views/offers.py
vkaracic: I'm expecting that they are sent to this endpoint (lines 51 and 52 in this file).

vkaracic:
@edx/helio ready for reviews!

mjfrey (approved):
Once tests are fixed.

vkaracic:
@edx/helio this is ready for a second review.

ecommerce/extensions/api/v2/urls.py
```diff
@@ -98,6 +99,7 @@
     .register(r'products', product_views.ProductViewSet,
               base_name='course-product', parents_query_lookups=['course_id'])
 router.register(r'orders', order_views.OrderViewSet)
+router.register(r'offers', offer_views.OfferViewSet, base_name='offers')
```
clintonb, line 102: You should not need the `base_name` kwarg.

ecommerce/extensions/api/v2/views/offers.py
```diff
@@ -0,0 +1,140 @@
...
+        stock_records = []
+        seat_types = program['applicable_seat_types']
+        for course in program['courses']:
+            for course_run in course['course_runs']:
+                _course = Course.objects.get(id=course_run['key'])
+                seat = _course.seat_products.get(
+                    attributes__name='certificate_type',
+                    attribute_values__value_text__in=seat_types
+                )
+                stock_records.append(StockRecord.objects.get(product=seat))
```
clintonb: This shuould be `.filter()`, since the relationship is 1:m.

ecommerce/extensions/api/v2/views/offers.py
```diff
@@ -0,0 +1,140 @@
...
+            api = request.site.siteconfiguration.course_catalog_api_client
+            program = json.loads(api.programs(program_uuid).get())
+            cache.set(cache_key, program, settings.PROGRAM_CACHE_TIMEOUT)
+        return program
+
+    def create_range(self, program):
+        """Create an Oscar Range for the program."""
+        stock_records = []
+        seat_types = program['applicable_seat_types']
+        for course in program['courses']:
```
clintonb: This can be made more efficient.

```python
partner = self.site.siteconfiguration.partner
currency = settings.OSCAR_DEFAULT_CURRENCY

course_runs = []
  course_run += course_runs

product_query = reduce(
  lambda x, y: x | y,
  [Q(course__id=course_run['key'], attributes__name='certificate_type', attribute_values__value_text__in=seat_types) for course_run in course_runs]
)

stock_records = StockRecord.objects.filter(Q(partner=partner, price_currency=currency) & product_query)
```

ecommerce/extensions/api/v2/views/offers.py
```diff
@@ -0,0 +1,140 @@
...
+            for course_run in course['course_runs']:
+                _course = Course.objects.get(id=course_run['key'])
+                seat = _course.seat_products.get(
+                    attributes__name='certificate_type',
+                    attribute_values__value_text__in=seat_types
+                )
+                stock_records.append(StockRecord.objects.get(product=seat))
+
+        name = 'Program: {}'.format(program['title'])
+        catalog = Catalog.objects.create(
```
clintonb: Why are we copying program information to Otto? What happens if the program courses/course runs change? I would prefer that we read the info from the Catalog Service, and cache it for some period of time.

ecommerce/extensions/api/v2/views/offers.py
clintonb: Actually, the SKUs are in the API response. Use them.

clintonb (changes requested):

clintonb:
We should avoid storing program details in the E-Commerce Service to avoid issues with data getting out-of-sync. Pull the info, and cache it, rather than storing it.

ecommerce/extensions/api/v2/urls.py
vkaracic: Without it ``reverse`` can't find the pattern.

vkaracic:
@clintonb I think I've managed to address your comment with the last commit, so please take a look and let me know if that's the way to go. I'll adjust/add tests once I know I'm on the right track.

ecommerce/extensions/offer/models.py
```diff
@@ -282,6 +291,36 @@ def catalog_contains_product(self, product):
...
+        Raises:
+            Exception: if it fails to establish a connection with the Catalog Service.
+        """
+        cache_key = hashlib.md5(
+            'program_{uuid}_seats'.format(uuid=self.program_uuid)
+        ).hexdigest()
+        program_seat_skus = cache.get(cache_key)
+
+        if not program_seat_skus:
+            request = get_current_request()
```
mjfrey, line 306: Do not use this anymore, pass the request in and use it.

ecommerce/extensions/offer/models.py
```diff
@@ -312,12 +354,14 @@ def contains_product(self, product):
     contains = contains_product
 
     def num_products(self):
-        return len(self.all_products())
+        return self.all_products()
```
mjfrey, line 357: why this change?

mjfrey (changes requested):

PR #1250, opened 2017-04-24
Enable all products to be displayed on receipt page

iivic:
[LEARNER-605](https://openedx.atlassian.net/browse/LEARNER-605)
This is the only line that was causing problems. I didn't find more code that could potentially break the receipt page. Feel free to double check.

### DIFF
--- a/ecommerce/templates/edx/checkout/receipt.html
+++ b/ecommerce/templates/edx/checkout/receipt.html
@@ -80,7 +80,9 @@
                 <dt class="course-description sr">{% trans "Description:" %}</dt>
                 <dd class="course-description">
                   <span>{{ line.description }}</span>
-                  <p>{{ line.product.course.id|course_organization }}</p>
+                  {% if line.product.course %}
+                    <p>{{ line.product.course.id|course_organization }}</p>
+                  {% endif %}
                 </dd>
                 <dt class="line-price sr">{% trans "Item Price:" %}</dt>
                 <dd class="line-price price">{{ line.line_price_before_discounts_incl_tax|currency:order.currency }}</dd>


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1250&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb (approved):

ecommerce/templates/edx/checkout/receipt.html
```diff
@@ -80,7 +80,9 @@ <h2 class="thank-you">{% trans "Thank you for your order!" %}</h2>
                 <dt class="course-description sr">{% trans "Description:" %}</dt>
                 <dd class="course-description">
                   <span>{{ line.description }}</span>
-                  <p>{{ line.product.course.id|course_organization }}</p>
+                  {% if line.product.course %}
+                    <p>{{ line.product.course.id|course_organization }}</p>
+                  {% endif %}
```
vkaracic, line 85: If you want to be inclusive, you can also add an ``else`` block and display ``line.product.description``.

vkaracic (approved):

ecommerce/templates/edx/checkout/receipt.html
clintonb: `line.description` is already included above. This just adds the org (e.g. edX, MITx) for course-related products. Frankly, I'd rather this if-statement be removed altogether. It's value is unclear to me.

vkaracic: Didn't see it, my bad.

iivic: I would rather leave it. Maybe someone finds it useful.

PR #1252, opened 2017-04-26
Check for previous purchase before adding to basket

Ayub-Khan:
**ECOM-7646**

**Description:**
As the Enrollment check was reverted in   ECOM-7349 MERGED  . Which use to prevent audit multiple order and audit multiple payment issues. We started facing those issues again. We need to get some more appropriate solution for those two issues.

**Steps to Reproduce:**
1. Go to stage
2. Enroll in a verified course as an audit student
3. Go to Dashboard
4. Open checkout page (by clicking on the upgrade to verified button) in new tab and pay for course
5. Go back to your dashboard tab (which should have been open this whole time)
6. Open checkout page in new tab (again by clicking the upgrade button) and pay for course again
7. You'll see you've created two orders for the same course

**Work:**
1. Add new order check to prevent multiple order issue
2. Remove Existing Enrollment Check
3. Update basket and coupon views to use Order check and write tests

**Expected behaviour:**
Students should not be able to pay for single course seat twice.

**Actual behaviour:** 
Students can pay for single course seat twice.

**Testing:**
[Sandbox](https://order-check.sandbox.edx.org)
screen shot 2017-05-22 at 1 39 16 pm

**Reviewers:**
Style and readability review: @noraiz-anwar 
Code Review: @schenedx 
Code Review: @bderusha 

**Post-review:**
Rebase and squash commits.

### DIFF
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -22,6 +22,7 @@ from ecommerce.enterprise.utils import (
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
+from ecommerce.extensions.order.utils import UserAlreadyPlacedOrder
 from ecommerce.extensions.test.factories import prepare_voucher
 from ecommerce.tests.mixins import ApiMockMixin, LmsApiMockMixin
 from ecommerce.tests.testcases import TestCase
@@ -346,7 +347,6 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin
 
     def redeem_coupon(self, code=COUPON_CODE, consent_token=None):
         self.request.user = self.user
-        self.mock_enrollment_api(self.request, self.user, self.course.id, is_active=False, mode=self.course_mode)
         return self.client.get(self.redeem_url_with_params(code=code, consent_token=consent_token))
 
     def assert_redirects_to_receipt_page(self, code=COUPON_CODE, consent_token=None):
@@ -459,7 +459,6 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin
             enterprise_customer=ENTERPRISE_CUSTOMER
         )
         self.request.user = self.user
-        self.mock_enrollment_api(self.request, self.user, self.course.id, is_active=False, mode=self.course_mode)
         self.mock_enterprise_learner_api(consent_provided=False)
         self.mock_enterprise_course_enrollment_api(results_present=False)
         self.mock_account_api(self.request, self.user.username, data={'is_active': True})
@@ -494,7 +493,6 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin
         """ Verify that the view renders an error when the consent token doesn't match. """
         code = self.prepare_enterprise_data()
         self.request.user = self.user
-        self.mock_enrollment_api(self.request, self.user, self.course.id, is_active=False, mode=self.course_mode)
         self.mock_account_api(self.request, self.user.username, data={'is_active': True})
         self.mock_access_token_response()
         self.mock_specific_enterprise_customer_api(ENTERPRISE_CUSTOMER)
@@ -553,20 +551,18 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin
 
     @httpretty.activate
     def test_already_enrolled_rejection(self):
-        """ Verify a user is rejected from redeeming a coupon for a course she's already enrolled in."""
-        self.mock_enrollment_api(self.request, self.user, self.course.id, is_active=True, mode=self.course_mode)
+        """ Verify a user is rejected from redeeming a coupon for a course he's already enrolled in."""
         self.mock_account_api(self.request, self.user.username, data={'is_active': True})
         self.mock_access_token_response()
-
         self.create_coupon_and_get_code()
-        response = self.client.get(self.redeem_url_with_params())
-        msg = 'You are already enrolled in the course.'
-        self.assertEqual(response.context['error'], msg)
+        with mock.patch.object(UserAlreadyPlacedOrder, 'user_already_placed_order', return_value=True):
+            response = self.client.get(self.redeem_url_with_params())
+            msg = 'You have already purchased {course} seat.'.format(course=self.course.name)
+            self.assertEqual(response.context['error'], msg)
 
     @httpretty.activate
     def test_invalid_email_domain_rejection(self):
         """ Verify a user with invalid email domain is rejected. """
-        self.mock_enrollment_api(self.request, self.user, self.course.id, is_active=True, mode=self.course_mode)
         self.create_coupon_and_get_code(email_domains='example.com')
         response = self.client.get(self.redeem_url_with_params())
         msg = 'You are not eligible to use this coupon.'
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -30,6 +30,7 @@ from ecommerce.extensions.api import exceptions
 from ecommerce.extensions.basket.utils import prepare_basket
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
+from ecommerce.extensions.order.exceptions import AlreadyPlacedOrderException
 from ecommerce.extensions.voucher.utils import get_voucher_and_products_from_code
 
 Applicator = get_class('offer.utils', 'Applicator')
@@ -182,9 +183,6 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
                 }
             )
 
-        if request.user.is_user_already_enrolled(request, product):
-            return render(request, template_name, {'error': _('You are already enrolled in the course.')})
-
         try:
             enterprise_customer = get_enterprise_customer_from_voucher(request.site, voucher)
         except EnterpriseDoesNotExist as e:
@@ -232,7 +230,12 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
                 )
                 return HttpResponseRedirect(redirect_url)
 
-        basket = prepare_basket(request, [product], voucher)
+        try:
+            basket = prepare_basket(request, [product], voucher)
+        except AlreadyPlacedOrderException:
+            msg = _('You have already purchased {course} seat.').format(course=product.course.name)
+            return render(request, template_name, {'error': msg})
+
         if basket.total_excl_tax == 0:
             try:
                 order = self.place_free_order(basket)
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -14,6 +14,8 @@ from ecommerce.core.tests import toggle_switch
 from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.extensions.basket.utils import attribute_cookie_data, prepare_basket
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.extensions.order.exceptions import AlreadyPlacedOrderException
+from ecommerce.extensions.order.utils import UserAlreadyPlacedOrder
 from ecommerce.extensions.partner.models import StockRecord
 from ecommerce.extensions.test.factories import prepare_voucher
 from ecommerce.referrals.models import Referral
@@ -281,6 +283,27 @@ class BasketUtilsTests(CourseCatalogTestMixin, TestCase):
         # test referral record is deleted when no cookies are set
         with self.assertRaises(Referral.DoesNotExist):
             Referral.objects.get(basket_id=basket.id)
+
+    def test_prepare_basket_raises_exception_for_purchased_product(self):
+        """
+        Test prepare_basket raises AlreadyPlacedOrderException if the product is already purchased by user
+        """
+        product = ProductFactory()
+        with mock.patch.object(UserAlreadyPlacedOrder, 'user_already_placed_order', return_value=True):
+            with self.assertRaises(AlreadyPlacedOrderException):
+                prepare_basket(self.request, [product])
+
+    def test_prepare_basket_for_purchased_enrollment_code(self):
+        """
+        Test prepare_basket returns basket with product even if its already been purchased by user
+        """
+        course = CourseFactory()
+        toggle_switch(ENROLLMENT_CODE_SWITCH, True)
+        course.create_or_update_seat('verified', False, 10, self.partner, create_enrollment_code=True)
+        enrollment_code = Product.objects.get(product_class__name=ENROLLMENT_CODE_PRODUCT_CLASS_NAME)
+        with mock.patch.object(UserAlreadyPlacedOrder, 'user_already_placed_order', return_value=True):
+            basket = prepare_basket(self.request, [enrollment_code])
+            self.assertIsNotNone(basket)
 
 
 class BasketUtilsTransactionTests(UserMixin, TransactionTestCase):
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -8,11 +8,16 @@ from django.db import transaction
 from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_class, get_model
 
+from ecommerce.courses.utils import mode_for_seat
+from ecommerce.extensions.order.exceptions import AlreadyPlacedOrderException
+from ecommerce.extensions.order.utils import UserAlreadyPlacedOrder
 from ecommerce.referrals.models import Referral
 
 Applicator = get_class('offer.utils', 'Applicator')
 Basket = get_model('basket', 'Basket')
 StockRecord = get_model('partner', 'StockRecord')
+OrderLine = get_model('order', 'Line')
+Refund = get_model('refund', 'Refund')
 
 logger = logging.getLogger(__name__)
 
@@ -28,7 +33,7 @@ def prepare_basket(request, products, voucher=None):
 
     Arguments:
         request (Request): The request object made to the view.
-        product (List): List of products to be added to the basket.
+        products (List): List of products to be added to the basket.
         voucher (Voucher): Voucher to apply to the basket.
 
     Returns:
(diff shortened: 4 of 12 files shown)


ecommerce/extensions/order/utils.py
```diff
@@ -113,3 +115,30 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
...
+        try:
+            orders = OrderLine.objects.filter(
+                quantity=1,
+                product=product,
+                line_price_incl_tax__gt=0,
+                order__user=user,
+            )
+        except OrderLine.DoesNotExist:
+            return False
+        for order in orders:
```
clintonb: You want to filter by RefundLine not Refund. Orders can have multiple items.

ecommerce/extensions/order/utils.py
```diff
@@ -113,3 +115,30 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
...
+        checks if the user already have a non refunded order for the product.
+        Args:
+            user: (User)
+            product: (Product)
+
+        Returns:
+                Returns True if user have a non refunded order else False
+        """
+        try:
+            orders = OrderLine.objects.filter(
```
clintonb: These are OrderLine objects not orders. Name your variable accordingly.

ecommerce/extensions/order/utils.py
```diff
@@ -113,3 +115,30 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
...
+                quantity=1,
```
clintonb: Why does quantity matter?

ecommerce/extensions/order/utils.py
```diff
@@ -113,3 +115,30 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
...
+                product=product,
+                line_price_incl_tax__gt=0,
```
clintonb: Why does price matter?

ecommerce/extensions/order/utils.py
```diff
@@ -113,3 +115,30 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
...
+                order__user=user,
+            )
+        except OrderLine.DoesNotExist:
+            return False
+        for order in orders:
+            if not Refund.objects.exists(order=order, status='Complete'):
```
clintonb: This can probably be combined with the query above. Remember to account for the other statuses.

ecommerce/extensions/order/utils.py
```diff
@@ -113,3 +115,30 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
             logger.exception('Referral for Order [%d] failed to save.', order.id)
 
         return order
+
+
+class UserAlreadyPlacedOrder(object):
+    @staticmethod
+    def user_already_placed_order(user, product):
```
clintonb: This should accept an array or products.

ecommerce/extensions/basket/views.py
```diff
@@ -113,6 +113,7 @@ def get(self, request):
         if not product.is_enrollment_code_product and code:
             try:
                 if request.user.is_user_already_enrolled(request, product):
+                # if have_already_placed_order(request.user, product):
```
clintonb: This needs to work for multi-item view.

clintonb (dismissed):

ecommerce/extensions/order/utils.py
Ayub-Khan: to avoid orders for audit enrollment

ecommerce/extensions/order/utils.py
Ayub-Khan: to exclude enrollment code orders.

clintonb: Enrollment codes are a different product class.

ecommerce/extensions/basket/views.py
Ayub-Khan: edx allow multi-item purchase against single basket?

ecommerce/extensions/order/utils.py
clintonb: The odds of an audit enrollment coming through here are low. If they did, we really don't care if you buy it twice. If you really want to check for audit, check the product attributes.

ecommerce/extensions/basket/views.py
clintonb: Yes. Even if we didn't, we should never artificially limit our system.

Ayub-Khan: ok

Ayub-Khan: I have divided the work into steps.

- 1. Write code for Order check.
- 2. Write Tests for Order check.
- 3. Remove already existing Enrollment Check and its related code.
- 4. Update single-item view to use the Order Check and write tests.
- 5. Update Order check to work for multiple-item view.
- 6. Update multi-item view to use Order check and write tests.

let me know if you have any suggestions.

Ayub-Khan: @adampalay @clintonb ^^

mjfrey: @Ayub-Khan I would move this check to prepare_basket.  Have that throw an exception and check for it in both single-item as well as multi-item views.  This way you have one place where you check for previous order.

mjfrey (changes requested):

Ayub-Khan:
Sorry just playing around with new git features

ecommerce/extensions/basket/utils.py
```diff
@@ -39,9 +43,21 @@ def prepare_basket(request, products, voucher=None):
...
-        # Call signal handler to notify listeners that something has been added to the basket
-        basket_addition.send(sender=basket_addition, product=product, user=request.user, request=request, basket=basket)
+        if not UserAlreadyPlacedOrder.user_already_placed_order(request.user, product) or \
+                product.is_enrollment_code_product:
+            basket.add_product(product, 1)
+            # Call signal handler to notify listeners that something has been added to the basket
+            basket_addition.send(sender=basket_addition, product=product, user=request.user, request=request,
+                                 basket=basket)
+        else:
+            logger.warning(
```
mjfrey: Please use the messaging framework to tell the user that products they already purchased will not be added to the basket. So it is not confusing to them.

ecommerce/coupons/views.py
```diff
@@ -232,6 +229,9 @@ def get(self, request):
                 return HttpResponseRedirect(redirect_url)
 
         basket = prepare_basket(request, [product], voucher)
+        if not basket:
+            return render(request, template_name, {'error': _('You are already enrolled in the course.')})
```
mjfrey: The message should be something like:
"You have already purchased these products"

ecommerce/extensions/basket/views.py
```diff
@@ -151,7 +154,9 @@ def get(self, request):
             return HttpResponseBadRequest(_('Products with SKU(s) [{skus}] do not exist.').format(skus=', '.join(skus)))
 
         voucher = Voucher.objects.get(code=code) if code else None
-        prepare_basket(request, products, voucher)
+        basket = prepare_basket(request, products, voucher)
+        if not basket:
+            return HttpResponseBadRequest(_('You are already enrolled in courses.'))
```
mjfrey: Same here.  It is not the enrollment but the purchase that has already happened.

mjfrey (changes requested):

ecommerce/extensions/basket/utils.py
Ayub-Khan: done

ecommerce/extensions/order/utils.py
Ayub-Khan: @clintonb as per @mjfrey advice it was implemented in prepare_basket so keeping this to take single input . Because we also log when user tries to purchase course seat again. So i think its batter if we keep logging in prepare_basket. because we might need tp use this in future on checkout view to prevent multiple_payment issue.

Ayub-Khan:
@clintonb  @mjfrey @noraiz-anwar I have added sandbox link. Kindly review.

Ayub-Khan:
As @clintonb @mjfrey have not reviewed this for a while adding new code reviewers.

Ayub-Khan:
@schenedx and @bderusha kindly review.

mjfrey:
The sandbox is not working for me.  It is not setup to use our e-commerce hosted basket page nor any payment processors.

ecommerce/extensions/basket/utils.py
```diff
@@ -39,9 +43,21 @@ def prepare_basket(request, products, voucher=None):
     basket.save()
     basket_addition = get_class('basket.signals', 'basket_addition')
     for product in products:
-        basket.add_product(product, 1)
-        # Call signal handler to notify listeners that something has been added to the basket
-        basket_addition.send(sender=basket_addition, product=product, user=request.user, request=request, basket=basket)
+        if not UserAlreadyPlacedOrder.user_already_placed_order(request.user, product) or \
+                product.is_enrollment_code_product:
```
schenedx: Can you put the condition `product.is_enrollment_code_product` before the `UserAlreadyPlacedOrder.user_already_placed_order()` call? We want the shortcutting to happen

ecommerce/extensions/basket/views.py
```diff
@@ -107,27 +108,13 @@ def get(self, request):
...
-                msg = _('An error occurred while retrieving enrollment details. Please try again.')
-                return HttpResponseBadRequest(msg)
-
         # At this point we're either adding an Enrollment Code product to the basket,
         # or the user is adding a Seat product for which they are not already enrolled
-        prepare_basket(request, [product], voucher)
+        basket = prepare_basket(request, [product], voucher)
+        # If user has already purchased the product basket will be None
+        if not basket:
+            msg = _('You have already purchased this {course} seat.').format(course=product.course.name)
```
schenedx: @Ayub-Khan What if the basket is `None` because of another error or condition that is different from `You already purchased`? Seems to me you are making the assumption that `None` basket means `already ordered`.
I am afraid this assumption is too implicit to understand.
Can we throw an exception in the `prepare_basket` function when the order is already placed, and then handle the exception here? I feel like Exception throwing and catching is the more explicit way of handling

ecommerce/extensions/basket/views.py
```diff
@@ -151,7 +138,10 @@ def get(self, request):
             return HttpResponseBadRequest(_('Products with SKU(s) [{skus}] do not exist.').format(skus=', '.join(skus)))
 
         voucher = Voucher.objects.get(code=code) if code else None
-        prepare_basket(request, products, voucher)
+        basket = prepare_basket(request, products, voucher)
+        if not basket:
```
schenedx: Same comment here about using an exception to signal `already ordered` instead.

ecommerce/extensions/order/utils.py
```diff
@@ -113,3 +115,23 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
...
+            user: (User)
+            product: (Product)
+
+        Returns:
+                Returns True if user have a non refunded order else False
+        """
+        orders_lines = OrderLine.objects.filter(product=product, order__user=user)
+        if orders_lines:
+            for order_line in orders_lines:
+                if not RefundLine.objects.filter(order_line=order_line).exists():
```
schenedx: Can you move this statement into a new utility function? I can think of the function name to be `is_order_line_refunded`

ecommerce/extensions/order/utils.py
```diff
@@ -113,3 +115,23 @@ def create_order_model(self, user, basket, shipping_address, shipping_method, sh
...
+                    return True
```
schenedx: I don't think we need this line. You can just do `return not is_order_line_refunded(order_line)`

schenedx (changes requested):
Commented. Please see if my comments make sense

ecommerce/extensions/order/utils.py
Ayub-Khan: @schenedx i did this to counter a edge case where order is multiple product order and there are multiple refund lines so if the first product is refunded in that case this method will return false even if user is trying to buy second product again.

ecommerce/extensions/basket/utils.py
```diff
@@ -57,7 +58,7 @@ def prepare_basket(request, products, voucher=None):
                 product.course_id
             )
```
schenedx: Shouldn't we raise the `AlreadyPlacedOrderException` here instead of where it is now, out side of the loop?

ecommerce/extensions/order/utils.py
schenedx: You are right. My bad. Thanks for the explanation

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -282,6 +283,27 @@ def test_attribute_cookie_data_multiple_cookies(self):
...
             Referral.objects.get(basket_id=basket.id)
 
+    def test_prepare_basket_returns_none_for_purchased_product(self):
+        """
+        Test prepare_basket returns none if the product is already purchased by user
+        """
+        product = ProductFactory()
+        with mock.patch.object(UserAlreadyPlacedOrder, 'user_already_placed_order', return_value=True):
+            basket = prepare_basket(self.request, [product])
+            self.assertIsNone(basket)
```
schenedx: Can you also assert that exception `AlreadyPlacedOrderException` is raised?

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -282,6 +283,27 @@ def test_attribute_cookie_data_multiple_cookies(self):
...
+        """
+        Test prepare_basket returns basket with product even if its already been purchased by user
+        """
+        course = CourseFactory()
+        toggle_switch(ENROLLMENT_CODE_SWITCH, True)
+        course.create_or_update_seat('verified', False, 10, self.partner, create_enrollment_code=True)
+        enrollment_code = Product.objects.get(product_class__name=ENROLLMENT_CODE_PRODUCT_CLASS_NAME)
+        with mock.patch.object(UserAlreadyPlacedOrder, 'user_already_placed_order', return_value=True):
+            basket = prepare_basket(self.request, [enrollment_code])
+            self.assertIsNotNone(basket)
```
schenedx, line 306: Can you also assert that exception `AlreadyPlacedOrderException` is also raised?

schenedx (changes requested):
I added more comments

ecommerce/extensions/basket/utils.py
Ayub-Khan: In case of multiple-item basket we let user move forward for all products which are not paid. But if we place raise here system will break even if first product is already paid and none other.

ecommerce/extensions/basket/tests/test_utils.py
Ayub-Khan: We let user purchase enrollment code products as many times as he wants so exception wont be raised.

Ayub-Khan:
@schenedx should i write extra tests for is_order_line_refunded?

ecommerce/extensions/basket/tests/test_utils.py
schenedx: I see. If that's the test case, bad suggestion from me.

schenedx:
@Ayub-Khan If I were you, I would. You never know if a future iteration of the changes will use the function `is_order_line_refunded` else where.

ecommerce/extensions/basket/utils.py
```diff
@@ -39,9 +44,21 @@ def prepare_basket(request, products, voucher=None):
...
+                                 basket=basket)
+        else:
+            logger.warning(
+                'User [%s] attempted to repurchase the [%s] seat of course [%s]',
+                request.user.username,
+                mode_for_seat(product),
+                product.course_id
+            )
+    if basket.is_empty:
+        raise AlreadyPlacedOrderException
```
schenedx: @Ayub-Khan I still have trouble on this logic.
The logic here tells me that, if we get an empty basket, no matter what the reason it is, we always show learner that `the order has been placed before` message. I feel like that's a dangerous assumption.
Why can't we do this:

1. Use a dictionary to store any product we encountered in the basket that was already ordered. After the loop, if we have dictionary to have members, and basket is empty, raise the `AlreadyPlacedOrderException` exception
2. If the dictionary have members, but we still have product in the basket, we can let user proceed with a different warning in the basket page telling user which products he has in the basket has been ordered before.
3. If the dictionary is empty, but basket is still empty, return `None`, and let caller handle an empty basket.
4. If the dictionary is empty, but basket has product, proceed as normal.

For point 2 above, I don't know if that would be a realistic case. @mjfrey @clintonb Can we make the assumption that we only have always 1 seat product in basket in all times?

Ayub-Khan: @schenedx i have updated the prepare basket.

Ayub-Khan:
@mjfrey @clintonb @schenedx  I have updated the sandbox. You can confirm the changes.
fyi @adampalay

mjfrey:
The sandbox is still not working for me. I get 500 errors.  The sandbox is also not setup for client-side checkout as well as the otto hosted receipt page.

Ayub-Khan:
@mjfrey Kindly explain the steps you are taking to create an order and on which steps you get 500 errors, so that I can better reproduce and debug this issue. I'm not sure how to setup a sandbox for client-side checkout and otto hosted receipt page. Is there documentation for how to set this up? I haven't been able to find any. In the past, we've made PRs without checking these things on sandboxes—if you're confident with the code workflow, can we instead wait to test on stage?

schenedx (approved):
👍  LGTM

schenedx:
@Ayub-Khan Can you setup your sandbox with the following config values? https://gist.github.com/clintonb/c45f4136c26c03d1f4ed#file-ecom-private-py-L39-L66

Ayub-Khan:
@mjfrey I have updated the sandbox to use client-side checkout as well as otto hosted receipt page. But sadly i have not been able to configure cybersource test account on my sandbox. Its still not working properly. But you can use the paypal to make a payment. 
Kindly use the following steps to confirm the changes on sandbox
1. login to sandbox using email: staff@example.com pass : edx
2. open "Upgrade to Verified" in new tab for course "Test Order Check 2"
3. click paypal for payment 
4. use any paypal sanbox test account to make payment for paypal
5. Notice the receipt page
6. Now from the other tab try to open "Upgrade to Verified" in new tab 
7. Notice the Error Page.
Thanks. 
fyi @schenedx

ecommerce/coupons/tests/test_views.py
```diff
@@ -553,20 +551,18 @@ def test_multiple_vouchers(self):
...
         self.mock_account_api(self.request, self.user.username, data={'is_active': True})
         self.mock_access_token_response()
-
         self.create_coupon_and_get_code()
-        response = self.client.get(self.redeem_url_with_params())
-        msg = 'You are already enrolled in the course.'
-        self.assertEqual(response.context['error'], msg)
+        with mock.patch.object(UserAlreadyPlacedOrder, 'user_already_placed_order', return_value=True):
+            response = self.client.get(self.redeem_url_with_params())
+            msg = 'You have already purchased this {course} seat.'.format(course=self.course.name)
```
mjfrey: "You have already purchased", please remove the "this"

mjfrey (approved):
Just a minor comment.

PR #1254, opened 2017-04-26
[WIP] Create new program admin page in ecommerce

iivic:
@edx/helio  Take a look and comment if this is the right direction or not.


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1254"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb (changes requested):
1. We should avoid copying program data. It can be cached, but it should not be permanently stored in the database.

2. We should start using React. Backbone is deprecated for edX. See http://open-edx-proposals.readthedocs.io/en/latest/oep-0011.html.

iivic:
1. I'm only saving the program uuid so we know to what program to apply the discount.

2. Learning React will take some time.... but I'm looking forward to it.

PR #1256, opened 2017-04-27
Add basket calculate endpoint.

mjfrey:
api/v2/basket/calculate
Creates a temporary basket, adds products and applies an optional voucher.
JSON is returned with the basket totals, including any discounts.

LEARNER-512

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_baskets.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_baskets.py
@@ -1,5 +1,9 @@
 # -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import datetime
 import json
+import urllib
 from collections import namedtuple
 from decimal import Decimal
 
@@ -18,10 +22,15 @@ from ecommerce.extensions.api.v2.tests.views import JSON_CONTENT_TYPE, OrderDeta
 from ecommerce.extensions.api.v2.views.baskets import BasketCreateView
 from ecommerce.extensions.payment import exceptions as payment_exceptions
 from ecommerce.extensions.payment.processors.cybersource import Cybersource
+from ecommerce.extensions.test.factories import prepare_voucher
+from ecommerce.tests.factories import ProductFactory
 from ecommerce.tests.mixins import BasketCreationMixin, ThrottlingMixin
 from ecommerce.tests.testcases import TestCase, TransactionTestCase
 
 Basket = get_model('basket', 'Basket')
+Benefit = get_model('offer', 'Benefit')
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
+Condition = get_model('offer', 'Condition')
 Order = get_model('order', 'Order')
 ShippingEventType = get_model('order', 'ShippingEventType')
 Refund = get_model('refund', 'Refund')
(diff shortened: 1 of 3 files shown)


mjfrey:
@edx/helio please review

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -293,3 +302,77 @@ def test_deletion(self):
...
+
+        self.user = self.create_user(is_staff=True)
+        self.client.login(username=self.user.username, password=self.password)
+
+    def test_no_sku(self):
+        """ Verify bad response when not providing sku(s) """
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 400)
+
+    def test_no_authorization(self):
```
clintonb: `test_no_authentication`

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -293,3 +302,77 @@ def test_deletion(self):
...
+        """ Verify that un-authorized users are rejected """
```
clintonb: un-authenticated

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -293,3 +302,77 @@ def test_deletion(self):
...
+        self.client.logout()
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 401)
+
+    def test_basket_calculate_no_offers(self):
+        """ Verify a successfull basket calculation with no offers"""
+        expected = {
+            'total_excl_discount': Decimal('29.97'),
```
clintonb: Where is 29.97 coming from? This should be an explicit variable/constant rather than assumed.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -293,3 +302,77 @@ def test_deletion(self):
...
+        factories.ConditionalOfferFactory(name='Test Offer', benefit=benefit, condition=condition,
+                                          offer_type=ConditionalOffer.SITE, start_datetime=datetime.datetime.now(),
+                                          end_datetime=datetime.datetime.now() + datetime.timedelta(days=1))
+
+        response = self.client.get(self.url)
+
+        expected = {
+            'total_excl_discount': Decimal('29.97'),
+            'discount_type': 'Percentage',
+            'total_discount': Decimal('10.00'),
```
clintonb: `10` should be a variable.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -293,3 +302,77 @@ def test_deletion(self):
...
+            'total_incl_discount': Decimal('27.00')
+        }
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data, expected)
+
+    def test_basket_calculate_coupon(self):
+        """ Verify sucessfull basket calculation when passing a voucher """
```
clintonb: successful

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -318,3 +326,84 @@ class BasketDestroyView(generics.DestroyAPIView):
     lookup_url_kwarg = 'basket_id'
     permission_classes = (IsAuthenticated, IsSuperuser,)
     queryset = Basket.objects.all()
+
+
+class BasketCalculateView(generics.GenericAPIView):
+    """ Calculate basket totals given a list of sku's
```
clintonb: What does this look like in Swagger? I suspect this needs to be on `get()`.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -293,3 +302,77 @@ def test_deletion(self):
...
+
+        expected = {
+            'total_excl_discount': Decimal('29.97'),
+            'discount_type': 'Percentage',
+            'total_discount': Decimal('100.00'),
+            'total_incl_discount': Decimal('0.00')
+        }
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data, expected)
```
clintonb: You should add a test for flat-rate discounts.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -318,3 +326,84 @@ class BasketDestroyView(generics.DestroyAPIView):
...
+    Returns:
+        JSON: {
+                'total_excl_discount': basket.total_incl_tax_excl_discounts,
+                'total_incl_discount': basket.total_incl_tax,
+                'discounts': discounts,
+                'discount_type': discount_type
+              }
+    """
+    permission_classes = (IsAuthenticated,)
+    renderer_classes = (JSONRenderer, )
```
clintonb: Remove. This is the default renderer. There is no need to specify this.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -318,3 +326,84 @@ class BasketDestroyView(generics.DestroyAPIView):
...
+
+    def get(self, request):
+
```
clintonb: Remove this unnecessary newline.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -318,3 +326,84 @@ class BasketDestroyView(generics.DestroyAPIView):
...
+                if voucher:
+                    basket.vouchers.add(voucher)
+
+                # Calculate any discounts on the basket.
+                Applicator().apply(basket, request)
+
+                total_discount = 0
+
+                if basket.offer_discounts:
```
clintonb: Why not both types of discounts? Today, we don't support applying both; however, we should avoid limiting this endpoint to one type.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -318,3 +326,84 @@ class BasketDestroyView(generics.DestroyAPIView):
...
+                elif basket.voucher_discounts:
+                    discounts = basket.voucher_discounts
+                else:
+                    discounts = None
+
+                if discounts:
+                    for discount in discounts:
+                        benefit = discount.get('offer').benefit
+                        total_discount += benefit.value
+                        discount_type = benefit.type
```
clintonb: This assumes that all discounts are the same type, which isn't necessarily the case. Perhaps we should leave out the discount type and amount for now. We are returning the total and the total with discounts. The client can figure out how it wants to display the info.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -318,3 +326,84 @@ class BasketDestroyView(generics.DestroyAPIView):
...
+                else:
+                    discount_type = None
+
+                response = {
+                    'total_excl_discount': basket.total_incl_tax_excl_discounts,
```
clintonb: total_incl_tax_excl_discounts

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -318,3 +326,84 @@ class BasketDestroyView(generics.DestroyAPIView):
...
+                    'total_incl_discount': basket.total_incl_tax,
```
clintonb: total_incl_tax

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -318,3 +325,76 @@ class BasketDestroyView(generics.DestroyAPIView):
...
+
+        code = request.GET.get('code', None)
+        voucher = Voucher.objects.get(code=code) if code else None
+
+        products = Product.objects.filter(stockrecords__partner=partner, stockrecords__partner_sku__in=skus)
+        if not products:
+            return HttpResponseBadRequest(_('Products with SKU(s) [{skus}] do not exist.').format(skus=', '.join(skus)))
+
+        # We wrap this in an atomic operation so we never commit this to the db.
+        # This is to avaoid merging this temporary basket with a real user bakset.
```
marjev: `avoid`
`basket`

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -293,3 +302,95 @@ def test_deletion(self):
...
+    def setUp(self):
+        super(BasketCalculateViewTests, self).setUp()
+        self.products = ProductFactory.create_batch(3, stockrecords__partner=self.partner)
+        qs = urllib.urlencode({'sku': [product.stockrecords.first().partner_sku for product in self.products]}, True)
+        self.path = reverse('api:v2:baskets:calculate')
+        self.url = '{root}?{qs}'.format(root=self.path, qs=qs)
+        self.range = factories.RangeFactory(includes_all_products=True)
+        self.product_total = 0
+        for product in self.products:
+            self.product_total += product.stockrecords.first().price_excl_tax
```
marjev: `self.product_total = sum(product.stockrecords.first().price_excl_tax for product in self.products)`

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -293,3 +302,95 @@ def test_deletion(self):
...
+        self.assertEqual(response.status_code, 400)
+
+    def test_no_authentication(self):
+        """ Verify that un-authenticated users are rejected """
+        self.client.logout()
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 401)
+
+    def test_basket_calculate_no_offers(self):
+        """ Verify a successfull basket calculation with no offers"""
```
marjev: `successful`

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -293,3 +302,95 @@ def test_deletion(self):
...
+        self.assertEqual(response.data, expected)
+
+    def test_basket_calculate_coupon(self):
+        """ Verify successful basket calculation when passing a voucher """
+        voucher, _ = prepare_voucher(_range=self.range)
+        response = self.client.get(self.url + '&code={code}'.format(code=voucher.code))
+
+        expected = {
+            'total_incl_tax_excl_discounts': self.product_total,
+            'total_discount': Decimal('100.00'),
```
marjev: `self.product_total` instead of `Decimal('100.00')`?

marjev (changes requested):

ecommerce/extensions/api/v2/tests/views/test_baskets.py
mjfrey: No

mjfrey:
@edx/helio another look please

marjev (approved):

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -318,3 +325,75 @@ class BasketDestroyView(generics.DestroyAPIView):
...
+                    discounts.extend(basket.voucher_discounts)
+
+                for discount in discounts:
+                    benefit = discount.get('offer').benefit
+                    total_discount += benefit.value
+
+                response = {
+                    'total_incl_tax_excl_discounts': basket.total_incl_tax_excl_discounts,
+                    'total_incl_tax': basket.total_incl_tax,
+                    'total_discount': total_discount,
```
clintonb: Without the discount type, I don't see the value of including this field. We can add it later, if needed. For now, we can use a simple cross-out display of pricing.

Also, please include the currency field.

clintonb (approved):
Please address my last comment before squashing and merging.

ecommerce/extensions/api/v2/views/baskets.py
mjfrey: @clintonb currency field?  Say more.

clintonb: Price has a currency (unit) and a value. We should never assume that we solely operate in USD.

Add: `'currency': basket.currency`

PR #1258, opened 2017-04-28
ENT-354: Change Enterprise Customer coupon implementation from 'Range' to 'Custom Benefit'

saleem-latif:
Hi @mattdrayer , @douglashall , @clintonb 

Please take a look at this PR, this PR adds required models and functions for moving to Custom Benefits for handling enterprise entitlements.

__Jira Ticket:__ [ENT-354](https://openedx.atlassian.net/browse/ENT-354)

@brittneyexline , @asadiqbal08 , @zubair-arbi -- FYI

__Note:__ I am adding more tests for the changes, I will make some updates to this PR in the upcoming days.


ecommerce/extensions/offer/models.py
```diff
@@ -39,6 +41,90 @@ def clean_value(self):
...
+    # http://django-oscar.readthedocs.io/en/releases-1.4/howto/how_to_create_a_custom_benefit.html
+
+    # Note that we are adding a new UUID field above -- I am assuming that this cannot be a proxy
+    # model because of this, since the intent of a proxy model is to include additional behavior on
+    # top of an existing model class and in this case we also need to include additional state
+    _description = _("%(value)s%% enterprise entitlement discount on %(range)s")
+
+    @property
+    def name(self):
+        return self._description % {
```
clintonb: We use `str.format()`.

ecommerce/extensions/offer/models.py
```diff
@@ -39,6 +41,90 @@ def clean_value(self):
...
+
+    This custom benefit covers use cases having to do with establishing relationships between an
+    Open edX user/learner and an enterprise customer (ref: http://github.com/edx/edx-enterprise).
+    """
+    enterprise_customer_uuid = models.UUIDField(
+        help_text='UUID for an EnterpriseCustomer from the Enterprise Service.'
+    )
+
+    # The below code was bootstrapped from the example found at
+    # http://django-oscar.readthedocs.io/en/releases-1.4/howto/how_to_create_a_custom_benefit.html
```
clintonb: Not really. The bulk of the class, `apply()`, is copied from `PercentageDiscountBenefit`. Why not simply inherit it?

ecommerce/extensions/offer/models.py
```diff
@@ -39,6 +41,90 @@ def clean_value(self):
             )
 
 
+class EnterpriseCustomerUserPercentageBenefit(Benefit):
```
clintonb: Similar to Oscar's benefit implementations, this class belongs in `benefits.py`. This will probably resolve your circular reference.

ecommerce/extensions/offer/models.py
```diff
@@ -39,6 +41,90 @@ def clean_value(self):
...
+    @property
+    def description(self):
+        return self._description % {
+            'value': self.value,
+            'range': self.range.name
+        }
+
+    class Meta:
+        app_label = 'offer'
+        verbose_name = _("Percentage enterprise entitlement discount benefit")
```
clintonb: Single quotes. `'Percentage discount benefit for enterprise customers'`

mattdrayer:
@saleem-latif this implementation looks a little different from the proto-work done in #1218 -- I'm not seeing a new ConditionalOffer class, which is something that was suggested by @clintonb and @mjfrey.  I'm not sure if we need that versus handling the conditional work inside of the Benefit's `apply()` method -- thoughts?

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,82 @@
...
+        proxy = True
+        verbose_name = _("Data sharing consent condition")
+        verbose_name_plural = _("Data sharing consent conditions")
+
+    def is_satisfied(self, offer, basket):
+        """
+        Determines whether enterprise learner is eligible for entitlements.
+        """
+        request = get_current_request()
+        enterprise_learner = enterprise_utils.get_enterprise_learner(request.site, basket.owner.username)
```
clintonb: basket.site. Please avoid get_current_request(). I'm not a fan of thread locals.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,82 @@
...
+    @property
+    def name(self):
+        return self._description
+
+    @property
+    def description(self):
+        return self._description
+
+    class Meta:
+        app_label = 'offer'
```
clintonb: All you need is proxy. The rest of these will not be used since this isn't a real model.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,82 @@
...
+from threadlocals.threadlocals import get_current_request
+
+from oscar.core.loading import get_model
+
+from ecommerce.enterprise import utils as enterprise_utils
+
+
+Condition = get_model('offer', 'Condition')
+
+__all__ = [
```
clintonb: This should not be necessary.

ecommerce/extensions/offer/models.py
```diff
@@ -325,3 +325,5 @@ def all_products(self):
 
 
 from oscar.apps.offer.models import *  # noqa isort:skip pylint: disable=wildcard-import,unused-wildcard-import,wrong-import-position,wrong-import-order,ungrouped-imports
+from ecommerce.extensions.offer.benefits import *  # noqa isort:skip
+from ecommerce.extensions.offer.conditions import *  # noqa isort:skip
```
clintonb: We tend to avoid these. Import only what you need.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,82 @@
...
+    def is_partially_satisfied(self, offer, basket):
+        """
+        Return True if enterprise learner is just one step away from getting entitlements.
+
+        If enterprise learner is already associated with the enterprise but has not
+        consented to data sharing then return False to let the learner know what he/she
+        needs to do in order to avail entitlements.
+        """
+        request = get_current_request()
+        enterprise_learner = enterprise_utils.get_enterprise_learner(request.site, basket.owner.username)
```
clintonb: basket.site

ecommerce/enterprise/utils.py
```diff
@@ -364,3 +393,32 @@ def get_enterprise_customer_data_sharing_consent_token(access_token, course_id,
...
+        >>> is_user_linked_to_enterprise_customer("228921dc-b64d-4006-8ac6-g7sl19zz26pl07e", user)
+        False
+
+    Returns:
+        (bool): `True` if given user belongs to enterprise customer `False` of it does not.
+    """
+    if not is_enterprise_feature_enabled():
+        return False
+
+    # Get HttpRequest object for the ongoing request
```
clintonb: This comment adds not value.

ecommerce/enterprise/utils.py
```diff
@@ -364,3 +393,32 @@ def get_enterprise_customer_data_sharing_consent_token(access_token, course_id,
...
+    request = get_current_request()
```
clintonb: Pass the site as an argument to this function.

ecommerce/enterprise/utils.py
```diff
@@ -364,3 +393,32 @@ def get_enterprise_customer_data_sharing_consent_token(access_token, course_id,
...
+
+def is_user_linked_to_enterprise_customer(enterprise_customer_uuid, user):
+    """
+    Does given user belongs to the given enterprise customer ?
+
+    Arguments:
+     enterprise_customer_uuid (string): UUID for the enterprise customer
+     user (User): django user object
+
+    Examples:
```
clintonb: This example doesn't add much value. It's not needed given how simple the function is.

saleem-latif:
@clintonb In order to resolve pylint error `Old-style class defined. (old-style-class)`. I made [these](https://github.com/edx/ecommerce/pull/1258/files#diff-116bdef327d1558d6d967bafad78ca50R27) changes, Please let me know if I should use `pylint: disable=old-style-class` instead.

Previous: 
```
class Meta:
    proxy = True           
```

New:
```
class Meta(object):
    proxy = True           
```

clintonb:
@saleem-latif Inheriting from `object` is the proper thing to do in Python 2. What you have is correct.

saleem-latif:
@mattdrayer I have added offer conditions, please take a look and share your thoughts

saleem-latif:
@brittneyexline , @douglashall , @clintonb PR is ready for another review, kindly take a llok and share your thoughts.

saleem-latif:
@mjfrey Can you also take a look at the changes ?

ecommerce/extensions/offer/benefits.py
```diff
@@ -0,0 +1,82 @@
...
+    This custom benefit covers use cases having to do with establishing relationships between an
+    Open edX user/learner and an enterprise customer (ref: http://github.com/edx/edx-enterprise).
+    """
+    enterprise_customer_uuid = models.UUIDField(
+        help_text='UUID for an EnterpriseCustomer from the Enterprise Service.'
+    )
+
+    _description = _('%(value)s%% enterprise entitlement discount on %(range)s')
+
+    # Note that we are adding a new UUID field above -- I am assuming that this cannot be a proxy
```
clintonb: That's correct. This comment can go away.

ecommerce/extensions/offer/benefits.py
```diff
@@ -0,0 +1,82 @@
...
+
+class EnterpriseCustomerUserPercentageBenefit(benefits.PercentageDiscountBenefit):
+    """
+    An offer benefit that gives a percentage discount for enterprise learners.
+
+    This custom benefit covers use cases having to do with establishing relationships between an
+    Open edX user/learner and an enterprise customer (ref: http://github.com/edx/edx-enterprise).
+    """
+    enterprise_customer_uuid = models.UUIDField(
+        help_text='UUID for an EnterpriseCustomer from the Enterprise Service.'
```
clintonb: This help text doesn't really say anything different from the field's name. It should go away. In the future, remember to mark the help text for translation.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,81 @@
...
+
+Condition = get_model('offer', 'Condition')
+
+
+class DataSharingConsentCondition(Condition):
+    """
+    An offer condition dependent on the data sharing consent of
+    enterprise learner.
+    """
+    _description = _("Enterprise learner has consented to data sharing.")
```
clintonb: 1. Use single quotes.
2. Move this to the `name` property.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @property
+    def name(self):
+        return self._description
+
+    @property
+    def description(self):
+        return self._description
```
clintonb: `return self.name`

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,81 @@
+
```
clintonb: This newline is unnecessary.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,81 @@
...
+from django.utils.translation import ugettext_lazy as _
+
```
clintonb: This newline is unnecessary.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,81 @@
...
+        # If enterprise customer requires data sharing consent, but enterprise learner has not
+        # consented then return False.
+        if enterprise_utils.enterprise_customer_needs_consent(enterprise_learner['enterprise_customer']) and \
+                not enterprise_utils.enterprise_customer_user_consent_provided(enterprise_learner):
+            return False
+
+        return True
+
+    def is_partially_satisfied(self, offer, basket):  # pylint: disable=unused-argument
```
clintonb: We aren't displaying any upsell messages. Neither this method nor `get_upsell_message` is necessary. These should be removed to avoid the unnecessary overhead of calling the Enterprise API for almost every page load.

ecommerce/extensions/offer/models.py
```diff
@@ -316,3 +316,8 @@ def all_products(self):
 
 
 from oscar.apps.offer.models import *  # noqa isort:skip pylint: disable=wildcard-import,unused-wildcard-import,wrong-import-position,wrong-import-order,ungrouped-imports
+from ecommerce.extensions.offer.benefits import (  # noqa isort:skip pylint: disable=wrong-import-position,unused-import,ungrouped-imports
```
clintonb, line 325: None of this is necessary.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,81 @@
...
+
+        If enterprise learner is already associated with the enterprise but has not
+        consented to data sharing then return False to let the learner know what he/she
+        needs to do in order to avail entitlements.
+        """
+        enterprise_learner = enterprise_utils.get_enterprise_learner(basket.site, basket.owner.username)
+
+        # If enterprise customer requires data sharing consent, but enterprise learner has not
+        # consented then return True.
+        # This means then learner just need to consent to get dispounts. so we show that message to
```
clintonb: Typos. Also, we aren't currently displaying upsell messages.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,81 @@
...
+        # We apply entitlements only if there is some item in the basket and all items
+        # in the basket are discountable.
+        if basket.is_empty or not all(self.can_apply_condition(line) for line in basket.all_lines()):
+            return False
+
+        # There must be a basket owner for enterprise entitlements
+        if not basket.owner:
+            return False
+
+        enterprise_learner = enterprise_utils.get_enterprise_learner(basket.site, basket.owner.username)
```
mjfrey: you need to catch any exceptions that the API call to enterprise app may throw.  If so return False.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def is_partially_satisfied(self, offer, basket):  # pylint: disable=unused-argument
+        """
+        Return True if enterprise learner is just one step away from getting entitlements.
+
+        If enterprise learner is already associated with the enterprise but has not
+        consented to data sharing then return False to let the learner know what he/she
+        needs to do in order to avail entitlements.
+        """
+        enterprise_learner = enterprise_utils.get_enterprise_learner(basket.site, basket.owner.username)
```
mjfrey: same here.

ecommerce/extensions/offer/benefits.py
```diff
@@ -0,0 +1,82 @@
...
+        return super(EnterpriseCustomerUserPercentageBenefit, self).apply(
+            basket=basket,
+            condition=condition,
+            offer=offer,
+            discount_percent=discount_percent,
+            max_total_discount=max_total_discount,
+        )
+
+
+class EnterpriseCustomerUserAbsoluteDiscountBenefit(benefits.AbsoluteDiscountBenefit):
```
mjfrey, line 42: Where are these new benefits used?

mjfrey (changes requested):

ecommerce/enterprise/utils.py
```diff
@@ -70,6 +74,35 @@ def get_enterprise_customer(site, uuid):
...
+    # Handle the case when username is empty. We need this condition as
+    # filtering enterprise learners with empty username will return all learners.
+    if not username:
+        return None
+
+    api = site.siteconfiguration.enterprise_api_client
+    api_resource = 'enterprise-learner'
+    endpoint = getattr(api, api_resource)
+    response = endpoint.get(
+        username=str(username),
```
clintonb: Nit: one line.
`username` is already a string.

ecommerce/enterprise/utils.py
```diff
@@ -70,6 +74,35 @@ def get_enterprise_customer(site, uuid):
...
+    Arguments:
+        site (Site): The site which is handling the current request
+        username (str): The username of the user in the LMS
+
+    Returns:
+        (dict): The single EnterpriseCustomerUser structure provided by the API
+            or returns None if EnterpriseCustomerUser with given username is not found.
+    """
+    # Handle the case when username is empty. We need this condition as
+    # filtering enterprise learners with empty username will return all learners.
```
clintonb: It seems you are using the wrong API endpoint. Why not use `/enterprise-learners/<username>/`?

ecommerce/enterprise/utils.py
```diff
@@ -22,6 +23,9 @@
 CONSENT_FAILED_PARAM = 'consent_failed'
 
 
+logger = logging.getLogger(__name__)
```
clintonb: This is never used. However, you probably should have more logging around your API calls.

ecommerce/enterprise/utils.py
```diff
@@ -70,6 +74,35 @@ def get_enterprise_customer(site, uuid):
...
+    """
+    # Handle the case when username is empty. We need this condition as
+    # filtering enterprise learners with empty username will return all learners.
+    if not username:
+        return None
+
+    api = site.siteconfiguration.enterprise_api_client
+    api_resource = 'enterprise-learner'
+    endpoint = getattr(api, api_resource)
+    response = endpoint.get(
```
clintonb: What happens if this call raises an exception?

clintonb: The response should be cached for some period of time to avoid performance issues.

ecommerce/extensions/offer/benefits.py
```diff
@@ -0,0 +1,82 @@
...
+from decimal import Decimal
+
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+from oscar.apps.offer import benefits, results
+
+from ecommerce.enterprise.utils import is_user_linked_to_enterprise_customer
+
+
+class EnterpriseCustomerUserPercentageBenefit(benefits.PercentageDiscountBenefit):
```
clintonb, line 10: Custom models should almost always be exposed in Django admin.

ecommerce/extensions/offer/benefits.py
```diff
@@ -0,0 +1,82 @@
+
```
clintonb: This is not necessary.

ecommerce/extensions/offer/benefits.py
```diff
@@ -0,0 +1,82 @@
...
+
+class EnterpriseCustomerUserAbsoluteDiscountBenefit(benefits.AbsoluteDiscountBenefit):
+    """
+    An offer benefit that gives a absolute discount for enterprise learners.
+
+    This custom benefit covers use cases having to do with establishing relationships between an
+    Open edX user/learner and an enterprise customer (ref: http://github.com/edx/edx-enterprise).
+    """
+    enterprise_customer_uuid = models.UUIDField(
```
clintonb: Extract this, and your `apply` method to a mixin shared by both classes.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+@ddt.ddt
+@httpretty.activate
+class TestDataSharingConsentCondition(EnterpriseServiceMockMixin):
+    """
+    Tests for data sharing consent offer condition.
+    """
+
+    def setUp(self):
+        """
+        Common setup operations for each test case.
```
clintonb: Docstrings on `setUp` and `tearDown` are unnecessary.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+from oscar.test import factories
+from oscar.test.basket import add_product
+
+from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
+from ecommerce.extensions.offer import models
+
+
+@ddt.ddt
+@httpretty.activate
+class TestDataSharingConsentCondition(EnterpriseServiceMockMixin):
```
clintonb: `DataSharingConsentConditionTests`

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    Tests for data sharing consent offer condition.
```
clintonb: This docstring doesn't tell us anything we don't already know from the class name. It can go away.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
+from decimal import Decimal as D
```
clintonb: Nit: I've never understood why folks shorten `Decimal` to `D`. Just use `Decimal`.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+
+    def setUp(self):
+        """
+        Common setup operations for each test case.
+        """
+        super(TestDataSharingConsentCondition, self).setUp()
+
+        self.learner = self.create_user(is_staff=True)
+        self.range = models.Range.objects.create(
```
clintonb: Use factories: `self.range = RangeFactory(includes_all_products=True)`

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+            name="All products range", includes_all_products=True,
+        )
+
+        self.basket = factories.create_basket(empty=True)
```
clintonb: `self.basket = BasketFactory(owner=self.learner, site=self.site)`

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+    Tests for data sharing consent offer condition.
+    """
+
+    def setUp(self):
+        """
+        Common setup operations for each test case.
+        """
+        super(TestDataSharingConsentCondition, self).setUp()
+
+        self.learner = self.create_user(is_staff=True)
```
clintonb: Nit: we generally use `self.user`.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+        Update basket so that it has a user and a site.
+        """
+        self.basket.owner = self.learner
+        self.basket.site = self.site
+        self.basket.save()
+
+    def test_name_and_description(self):
+        """
+        Test name and description properties od condition.
+        @return:
```
clintonb: Remove.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def _update_basket(self):
+        """
+        Update basket so that it has a user and a site.
+        """
+        self.basket.owner = self.learner
+        self.basket.site = self.site
+        self.basket.save()
+
+    def test_name_and_description(self):
```
clintonb: Split this into `test_name` and `test_description`.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+        self.assertEqual(
+            'Enterprise learner has consented to data sharing.',
+            self.condition.description,
+        )
+
+    def test_empty_basket_fails_condition(self):
+        """
+        Test that condition fails for empty basket.
+        """
+        self.assertFalse(self.condition.is_satisfied(self.offer, self.basket))
```
clintonb: 1. If your condition doesn't rely on the offer parameter, either pass `None` or use `ConditionalOfferFactory`. Don't mock things unnecessarily.
2. Avoid using instance variables unnecessarily. Create an empty basket for this test case with `BasketFactory`.

`self.assertFalse(self.condition.is_satisfied(None, BasketFactory()))`

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def test_not_discountable_item_fails_condition(self):
```
clintonb: I'm not sure of the value of this test. Your condition should not fail because the basket contains an item that is not discountable. It should only fail because data consent has not been granted. The items in the basket don't matter at all.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+            "ecommerce.extensions.offer.conditions.enterprise_utils.get_enterprise_learner", return_value=None,
+        ):
+            self.assertFalse(self.condition.is_satisfied(self.offer, self.basket))
+
+    def test_condition_fails_for_no_basket_owner(self):
+        """
+        Test that condition fails if basket does not have any owner.
+        """
+        # Remove basket owner
+        self.basket.owner = None
```
clintonb: You don't need `self.basket`. Just use `BasketFactory()`.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+        self.basket.save()
+
+        add_product(self.basket, D('26'), product=self.item)
```
clintonb: Products don't matter.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,81 @@
...
+    class Meta(object):
+        proxy = True
+
+    def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
+        """
+        Determines whether enterprise learner is eligible for entitlements.
+        """
+        # We apply entitlements only if there is some item in the basket and all items
+        # in the basket are discountable.
+        if basket.is_empty or not all(self.can_apply_condition(line) for line in basket.all_lines()):
```
clintonb: Why does this condition care about the basket's contents?

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        product = factories.create_product(is_discountable=False)
+        add_product(self.basket, D('26'), product=product)
+        self.assertFalse(self.condition.is_satisfied(self.offer, self.basket))
+
+    def test_condition_fails_for_none_enterprise_learner(self):
+        """
+        Test that condition fails if learner does not belong to any enterprise.
+        """
+        add_product(self.basket, D('26'), product=self.item)
```
clintonb: `self.basket.add_product(self.item)`

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+            name="All products range", includes_all_products=True,
+        )
+
+        self.basket = factories.create_basket(empty=True)
+        self._update_basket()
+
+        self.condition = models.DataSharingConsentCondition(range=self.range)
+        self.offer = mock.Mock()
+        self.item = factories.create_product(price=D('5.00'))
+        self.expensive_item = factories.create_product(price=D('15.00'))
```
clintonb: This is never used, and should be removed.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+        self.range = models.Range.objects.create(
+            name="All products range", includes_all_products=True,
+        )
+
+        self.basket = factories.create_basket(empty=True)
+        self._update_basket()
+
+        self.condition = models.DataSharingConsentCondition(range=self.range)
+        self.offer = mock.Mock()
+        self.item = factories.create_product(price=D('5.00'))
```
clintonb: 1. The price doesn't matter.
2. Use a name that helps describe what the variable represents.

`self.product = factories.create_product()`

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
+from decimal import Decimal as D
+
+import ddt
+import httpretty
+import mock
+from oscar.test import factories
+from oscar.test.basket import add_product
```
clintonb: This function isn't necessary. Use `Basket.add_product()`.

ecommerce/extensions/offer/tests/test_conditions.py
```diff
@@ -0,0 +1,158 @@
...
+
+        self.basket = factories.create_basket(empty=True)
+        self._update_basket()
+
+        self.condition = models.DataSharingConsentCondition(range=self.range)
+        self.offer = mock.Mock()
+        self.item = factories.create_product(price=D('5.00'))
+        self.expensive_item = factories.create_product(price=D('15.00'))
+
+    def _update_basket(self):
```
clintonb: This method is not needed.

ecommerce/enterprise/tests/test_utils.py
```diff
@@ -2,11 +2,16 @@
...
 import ddt
 import httpretty
+from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
+from oscar.test import factories
+from waffle.models import Switch
 
 from ecommerce.core.tests.decorators import mock_enterprise_api_client
 from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
 from ecommerce.enterprise.utils import (enterprise_customer_user_needs_consent, get_enterprise_customer,
```
clintonb: Drop the contained items to a new line.

ecommerce/enterprise/tests/test_utils.py
```diff
@@ -19,6 +24,15 @@ def setUp(self):
         self.learner = self.create_user(is_staff=True)
         self.client.login(username=self.learner.username, password=self.password)
 
+    def _enable_enterprise(self, is_active=True):
+        """
+        Enable Enterprise functionality via waffle switch.
+        """
+        Switch.objects.get_or_create(
```
clintonb: Use `toggle_switch`

ecommerce/enterprise/tests/test_utils.py
```diff
@@ -134,3 +148,105 @@ def test_ecu_needs_consent_link_exists(
...
+        Verify that "get_enterprise_learner" returns an appropriate response from the
+        "enterprise-learner" Enterprise service API endpoint.
+        """
+        self.mock_access_token_response()
+        self.mock_enterprise_learner_api(learner_id=self.learner.id)
+
+        enterprise_learner = get_enterprise_learner(self.site, self.learner.username)
+
+        # assert non-None result is returned when enterprise learner with given username exist
+        self.assertIsNotNone(enterprise_learner)
```
clintonb: You should check the returned value. `assertIsNotNone` is not good enough if the non-null value is still invalid.

ecommerce/enterprise/tests/test_utils.py
```diff
@@ -134,3 +148,105 @@ def test_ecu_needs_consent_link_exists(
...
+
+        enterprise_learner = get_enterprise_learner(self.site, '')
+        # assert None is returned when get_enterprise_learner is called with empty username
+        self.assertIsNone(enterprise_learner)
+
+    @mock_enterprise_api_client
+    def test_get_enterprise_learner_with_no_learner(self):
```
clintonb: This isn't right.

clintonb (changes requested):
I am concerned that this may not work as expected. Using non-proxy models may cause issues when it comes time to actually apply the offer. I would like to see a test that creates a conditional offer with the custom condition and benefits, and actually applies the offer to a basket.

ecommerce/extensions/offer/benefits.py
saleem-latif: Function signature for `apply` of `benefits.PercentageDiscountBenefit` and `benefits.AbsoluteDiscountBenefit` is different. If I try to abstract away conflicting arguments using `*args` and `**kwargs` it would reduce readability. I think it would be better to leave apply function in both classes.

ecommerce/enterprise/utils.py
saleem-latif: Currently, we have only `/enterprise-learners/<id>/` api endpoint and not `/enterprise-learners/<username>/`, There is a ticket open https://openedx.atlassian.net/browse/ENT-374 which might handle this case going forward. That is why currently we are using filtering to get enterprise learner.

ecommerce/extensions/offer/tests/test_conditions.py
saleem-latif: We need `add_product` for `test_not_discountable_item_fails_condition` otherwise it raises `ValueError: Basket lines must all have stock records.`.
I have removed other occurrences of this

ecommerce/extensions/offer/benefits.py
saleem-latif: Done

ecommerce/extensions/offer/benefits.py
saleem-latif: Done

ecommerce/extensions/offer/models.py
saleem-latif: removing these causes isort check failure in quality

saleem-latif:
@clintonb I agree that we need to add an integration test for the newly added benefits and offer conditions, but we have that planned for a later story (https://openedx.atlassian.net/browse/ENT-380).
I tried adding a test but doing that requires some discovery work around conditional offers and their use. So, I am thinking that it would be better to add those tests as part of [ENT-380](https://openedx.atlassian.net/browse/ENT-380).

@mjfrey I hope this also answers your question about where to use these benefits.

ecommerce/extensions/offer/models.py
clintonb: I'm referring to the new lines of code. This file does not need to be modified.

clintonb:
@saleem-latif this can't be merged until there is an integration test. You've created a bunch of code that may not work as you expect. Discovery around conditional offers should have been done before you started work on conditional offers. See https://github.com/edx/ecommerce/commit/e9d66122721424e2fea982389cf4fff0291057cd#diff-3fc00a5e5e9842eaee3b3266d1f0d597 for an example test.

ecommerce/extensions/offer/models.py
saleem-latif: Okay, I will remove these

saleem-latif: I looked around removing these but we need benefits imports here for migrations to work properly. Also this style conforms to [django-oscar convention](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/models.py#L63-L64). So, I think we should leave these as is.

saleem-latif:
@mjfrey @clintonb I have added integration tests, please take a look.

ecommerce/extensions/offer/migrations/0013_auto_20170511_0838.py
```diff
@@ -0,0 +1,54 @@
...
+        migrations.CreateModel(
+            name='DataSharingConsentCondition',
+            fields=[
+            ],
+            options={
+                'proxy': True,
+            },
+            bases=('offer.condition',),
+        ),
+        migrations.AlterField(
```
clintonb: Why is this field being altered?

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,52 @@
...
+Condition = get_model('offer', 'Condition')
+logger = logging.getLogger(__name__)
+
+
+class DataSharingConsentCondition(Condition):
+    """
+    An offer condition dependent on the data sharing consent of
+    enterprise learner.
+    """
+    name = unicode(_('Enterprise learner has consented to data sharing.'))
```
clintonb: Remove the usage of `unicode()`. The translations will never be used since the string will be cast to Unicode (which is the output of `ugettext_lazy`) when the class is loaded.

Also, remove the period.

ecommerce/extensions/offer/models.py
clintonb: Please explain how removing these affects migrations. That is a sign of something wrong. I am concerned that this will lead to circular import issues down the road. Given that these models are not used in this code, the imports should not be here.

We don't follow that particular convention of Oscar because we generally avoid `import *`. You're not even following the convention here because you haven't extended `__all__`.

ecommerce/extensions/offer/tests/test_benefits.py
```diff
@@ -1,21 +1,33 @@
...
+from django.core.urlresolvers import reverse
+from oscar.core.loading import get_model
 from oscar.test import factories
 from oscar.test.basket import add_product, add_products
 from oscar.test.factories import BasketFactory, RangeFactory
 
 from ecommerce.core.tests import toggle_switch
 from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
 from ecommerce.extensions.offer import models
+from ecommerce.tests.factories import (EnterpriseCustomerUserAbsoluteDiscountBenefittFactory,
```
clintonb: Drop the enclosed items to a new line to avoid long indentations.

ecommerce/tests/factories.py
```diff
@@ -39,3 +39,23 @@ class Meta(object):
 class StockRecordFactory(OscarStockRecordFactory):
     product = factory.SubFactory(ProductFactory)
     price_currency = 'USD'
+
```
clintonb: Move all of this to `ecommerce/extensts/test/factories.py`.

ecommerce/tests/factories.py
```diff
@@ -39,3 +39,23 @@ class Meta(object):
...
+
+class EnterpriseCustomerUserPercentageBenefitFactory(factory.DjangoModelFactory):
+    enterprise_customer_uuid = 'b0990312-c5a3-48c4-b0fe-b8d844249623'
```
clintonb: Don't hard-code this. Use `factory.LazyFunction(uuid.uuid4)`

ecommerce/tests/factories.py
```diff
@@ -39,3 +39,23 @@ class Meta(object):
...
+    type = get_model('offer', 'Benefit').PERCENTAGE
```
clintonb: Set `Benefit = get_model('offer', 'Benefit')` at the top of the file.

ecommerce/tests/factories.py
```diff
@@ -39,3 +39,23 @@ class Meta(object):
...
+    value = 10
+    range = factory.SubFactory(RangeFactory)
+
+    class Meta(object):
+        model = get_model('offer', 'EnterpriseCustomerUserPercentageBenefit')
```
clintonb: Your custom models are not dynamic. Import them as you would any non-Oscar class.

ecommerce/extensions/offer/tests/test_benefits.py
```diff
@@ -27,8 +39,10 @@ def setUp(self):
 
         self.range = RangeFactory(includes_all_products=True)
         self.user = self.create_user(is_staff=True)
+        self.client.login(username=self.user.username, password=self.password)
+
         self.condition = models.DataSharingConsentCondition(range=self.range)
-        self.benefit = models.EnterpriseCustomerUserPercentageBenefit(
+        self.benefit = EnterpriseCustomerUserPercentageBenefitFactory(
```
clintonb: Let your factory do the work: `self.benefit = EnterpriseCustomerUserPercentageBenefitFactory()`

ecommerce/extensions/offer/tests/test_benefits.py
```diff
@@ -107,7 +121,61 @@ def test_applies_correctly_to_basket_which_matches_condition(self):
...
+        Verify successful basket calculation with enterprise benefits applied according to the conditions.
+        """
+        self.mock_access_token_response()
+        self.mock_enterprise_learner_api(
+            learner_id=self.user.id,
+            enterprise_customer_uuid=TEST_ENTERPRISE_CUSTOMER_UUID,
+            consent_enabled=is_consent_enabled,
+            consent_provided=is_consent_provided,
+        )
+        benefit = EnterpriseCustomerUserPercentageBenefitFactory(
```
clintonb: As above, use the values from the factory. Only override what is absolutely necessary.

ecommerce/extensions/offer/tests/test_benefits.py
```diff
@@ -107,7 +121,61 @@ def test_applies_correctly_to_basket_which_matches_condition(self):
...
+        products = ProductFactory.create_batch(items, stockrecords__partner=self.partner)
+        product_total = sum(product.stockrecords.first().price_excl_tax for product in products)
+
+        qs = urllib.urlencode({'sku': [product.stockrecords.first().partner_sku for product in products]}, True)
+        path = reverse('api:v2:baskets:calculate')
+        url = '{root}?{qs}'.format(root=path, qs=qs)
+
+        factories.ConditionalOfferFactory(
+            name='Test Offer', benefit=benefit, condition=condition, offer_type=ConditionalOffer.SITE,
+            start_datetime=datetime.datetime.now(), end_datetime=datetime.datetime.now() + datetime.timedelta(days=1),
```
clintonb: Neither the name nor dates need to be overwritten.

ecommerce/extensions/offer/tests/test_benefits.py
```diff
@@ -107,7 +121,61 @@ def test_applies_correctly_to_basket_which_matches_condition(self):
...
+        )
+        condition = factories.ConditionFactory(
+            range=self.range,
+            proxy_class="ecommerce.extensions.offer.models.DataSharingConsentCondition",
+        )
+        products = ProductFactory.create_batch(items, stockrecords__partner=self.partner)
+        product_total = sum(product.stockrecords.first().price_excl_tax for product in products)
+
+        qs = urllib.urlencode({'sku': [product.stockrecords.first().partner_sku for product in products]}, True)
+        path = reverse('api:v2:baskets:calculate')
```
clintonb: Why are you using this endpoint? I provided an example does not involve an API call, and properly validates each line item in the basket. It would be ideal if you followed that model so that we actually test what we want to test: the basket discounts.

ecommerce/extensions/offer/tests/test_benefits.py
```diff
@@ -107,7 +121,61 @@ def test_applies_correctly_to_basket_which_matches_condition(self):
         self.assertEqual(2, self.basket.num_items_with_discount)
         self.assertEqual(0, self.basket.num_items_without_discount)
 
+    @ddt.data(
+        (True, False, 3, Decimal('20.00'), Decimal('29.97')),
+        (True, True, 3, Decimal('20.00'), Decimal('24.00')),
+        (False, True, 3, Decimal('20.00'), Decimal('24.00')),
+        (False, False, 3, Decimal('20.00'), Decimal('24.00')),
```
clintonb: Break this into three tests:

1. Consent disabled.
2. Consent enabled, but not provided.
3. Consent enabled, and provided.

Use helper methods to avoid duplication. As written, it is not totally clear what you are testing in each of these cases. Three months from now it won't make any sense.

Do the same for the test below.

Also, since you are testing more than benefits, move this to a new file, `test_offer.py`, similar to what we did for `programs`.

clintonb (changes requested):

ecommerce/extensions/offer/models.py
saleem-latif: If we remove these imports then migrations are not created, and if we remove these at this point then future changes to the benefit models will not be registered, my understanding is that django searches for models in models.py

clintonb: Go it. Thanks for clarifying. Please add that comment in this file to avoid future confusion.

ecommerce/extensions/offer/migrations/0013_auto_20170511_0838.py
saleem-latif: Field was being altered because migration for `offer.Condition` contained `verbose_name='Program UUID'` in the migration ([added here](https://github.com/edx/ecommerce/pull/1262/files#diff-3b5e7b36f57ff6196a43ba289ce6ef12R18)) but the same verbose name was not present in the model definition [added here](https://github.com/edx/ecommerce/pull/1262/files#diff-49b6e1e79e2754be1784cc6dc9a1fe84R321). Not sure what caused this, but my guess would be that migration for `offer.Condition` was created when verbose_name was present in models definition and later removed (but migrations were not updated).
There are two approaches forward here
1. Do not update `offer.Condition`. This way `AlterField` migration will pop up (as is happening currently)
2. Add `verbose_name='Program UUID'` in the `program_uuid` field of `offer.Condition`

Which one do you prefer ?

saleem-latif:
@clintonb please take a look, I have incorporated your feedback.

ecommerce/extensions/offer/migrations/0013_auto_20170512_0558.py
```diff
@@ -0,0 +1,54 @@
...
+        migrations.CreateModel(
+            name='DataSharingConsentCondition',
+            fields=[
+            ],
+            options={
+                'proxy': True,
+            },
+            bases=('offer.condition',),
+        ),
+        migrations.AlterField(
```
clintonb: This will be fixed by https://github.com/edx/ecommerce/pull/1272.

ecommerce/extensions/offer/benefits.py
clintonb: Digging deeper (both in the code and my understanding of Oscar): these should be conditions, not benefits. By the time we get to applying a benefit, we should know if it's going to be applied. A custom condition is what you want to determine if the basket/user meets the conditions for applying an offer.

ecommerce/extensions/offer/migrations/0013_auto_20170512_0558.py
saleem-latif: I have rebased it, migration is fixed

saleem-latif:
@clintonb can you please take another look.

ecommerce/enterprise/tests/test_utils.py
```diff
@@ -2,11 +2,16 @@
...
 import ddt
 import httpretty
+from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
+from oscar.test import factories
 
+from ecommerce.core.tests import toggle_switch
 from ecommerce.core.tests.decorators import mock_enterprise_api_client
 from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
 from ecommerce.enterprise.utils import (enterprise_customer_user_needs_consent, get_enterprise_customer,
```
clintonb, line 12: Drop the contained items to a new line, and de-indent.

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        Determines whether enterprise learner is eligible for entitlements.
+        """
+        # There must be a basket owner for enterprise entitlements
+        if not basket.owner:
+            return False
+
+        try:
+            enterprise_learner = enterprise_utils.get_enterprise_learner(basket.site, basket.owner)
+        except (ConnectionError, SlumberBaseException, Timeout):
```
clintonb, line 37: What about other exceptions?

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,52 @@
...
+            logger.exception('Failed to retrieve learner data for "%s"', basket.owner.username)
```
clintonb, line 38: We tend to use square brackets rather than double-quotes. Also, this needs to be a bit more informative:
`Failed to retrieve learner [username] from Enterprise API for site [site.name].`

ecommerce/extensions/offer/conditions.py
```diff
@@ -0,0 +1,52 @@
+import logging
+from django.utils.translation import ugettext as _
+from oscar.core.loading import get_model
+from requests.exceptions import ConnectionError, Timeout
+from slumber.exceptions import SlumberBaseException
+from ecommerce.enterprise import utils as enterprise_utils
```
clintonb, line 6: Add a newline above this one.

ecommerce/enterprise/utils.py
```diff
@@ -70,6 +71,30 @@ def get_enterprise_customer(site, uuid):
...
+    """
+    # Handle the case when user is anonymous. We need this condition as
+    # filtering enterprise learners for anonymous user will return all learners.
+    if user.is_anonymous():
+        return None
+
+    response = enterprise_api.fetch_enterprise_learner_data(site, user)
+    results = response.get('results')
+
+    return results[0] if results else None
```
clintonb, line 95: This is a code smell. Rather than return a special value, you should raise an exception (e.g. `EnterpriseUserNotFound`). Otherwise, you run the risk that someone uses this method and misses the fact that a special value is returned. An exception is hard to miss.

ecommerce/enterprise/utils.py
```diff
@@ -364,3 +389,24 @@ def get_enterprise_customer_data_sharing_consent_token(access_token, course_id,
...
+
+    Arguments:
+     site (Site): The site which is handling the current request
+     enterprise_customer_uuid (string): UUID for the enterprise customer
+     user (User): django user object
+
+    Returns:
+        (bool): `True` if given user belongs to enterprise customer `False` of it does not.
+    """
+    if not is_enterprise_feature_enabled():
```
clintonb, line 406: Move this upstream. Keep your utility methods as simple and specific as possible. Utilities should not care about features being enabled. That responsibility belongs at a higher level.

ecommerce/extensions/offer/admin.py
```diff
@@ -14,3 +18,13 @@ class RangeAdminExtended(admin.ModelAdmin):
 @admin.register(Condition)
 class ConditionAdminExtended(ConditionAdmin):
     list_display = ('type', 'value', 'range', 'program_uuid')
+
+
+@admin.register(EnterpriseCustomerUserPercentageBenefit)
+class EnterpriseCustomerUserPercentageBenefitAdmin(admin.ModelAdmin):
+    list_display = ('type', 'value', 'range', 'enterprise_customer_uuid')
```
clintonb, line 25: `'type'` will always be the same for these instances. It can go away. The same applies to the model below.

clintonb (changes requested):
@saleem-latif the suggestion to re-implement this as a custom benefit was not fully-informed. Given that you want to determine if a user is eligible for an offer, you want to create custom `Condition` models, not `Benefit`. You can use the existing `Benefit` models. If you continue down this route, using data consent as your sole condition, you run the risk of breaking other site-wide offers.

@mattdrayer @douglashall it might help if we get a better understanding of how you intend to use offers now and in the future.

FYI @mjfrey

saleem-latif:
@clintonb If we use existing benefits than I do not think we would be able associated a benefit to enterprise customer. For enterprise learners we also want custom analytics events and reports like
1. Total of the enterprise entitlements applied 
2. Time stamps for entitlement application for enterprise learners (This data will be dependent on data sharing consent states of customer and learner)
3. (potential use case) Cost limits applied by enterprise customer (unique for every customer), entitlements should not work after the limit has been exhausted and enterprise customer should be notified accordingly.

There might be other use cases that I am not aware of at this point or that might come up as we move forward with enterprise functionality.

clintonb:
What's the deal with this? Why isn't it done?

douglashall:
@clintonb It has been deprioritized for other work at the moment. We will pick this back up soon.

douglashall:
Closing in favor of https://github.com/edx/ecommerce/pull/1475.

PR #1260, opened 2017-05-01
Updated edx-auth-backends to 1.0.4

clintonb:
This update should resolve the issue with UUIDs being appended to usernames.


clintonb:
See https://github.com/edx/auth-backends/pull/39 for diff of edx-auth-backends.

PR #1261, opened 2017-05-01
Fixed Django Debug Toolbar configuration for devstack

clintonb:
Devstack users can now enable the toolbar by setting the ENABLE_DJANGO_TOOLBAR environment variable.


clintonb:
FYI @edx/learner

PR #1262, opened 2017-05-03
Added program offer administrative views

clintonb:
These views allow staff to create, read, and update conditional offers for programs.

LEARNER-435


clintonb:
This is intended to replace https://github.com/edx/ecommerce/pull/1254 and https://github.com/edx/ecommerce/pull/1248.

clintonb:
screen shot 2017-05-03 at 12 12 54 am
screencapture-localhost-18130-program-offers-new-1493784609309
screencapture-localhost-18130-program-offers-new-1493784659328
screencapture-localhost-18130-program-offers-6-edit-1493784642490

clintonb:
FYI @edx/helio

vkaracic:
Perhaps add a human readable title/name for a program so that someone can more easily search and distinguish programs. Something similar like course name field.

ecommerce/program_offers/views.py
```diff
@@ -54,6 +54,10 @@ def get_context_data(self, **kwargs):
         context = super(ProgramOfferUpdateView, self).get_context_data(**kwargs)
         context.update({
             'editing': True,
+            # TODO: Include program name
+            'program': {
+                'uuid': self.object.condition.program_uuid,
```
iivic: The uuid is already in the context. Why do we need this?

clintonb: See the TODO above. The goal is to provide more program details. This is merely a placeholder.

ecommerce/program_offers/conditions.py
```diff
@@ -15,15 +17,24 @@ def name(self):
...
+             'program_{uuid}'.format(uuid=self.program_uuid)
+        ).hexdigest()
+        program = cache.get(cache_key)
+
+        if not program:
+            try:
+                program = client.programs(str(self.program_uuid)).get()
+                cache.set(cache_key, program, settings.PROGRAM_CACHE_TIMEOUT)
+            except (ConnectionError, SlumberBaseException, Timeout):
+                raise Exception('Unable to connect to Course Catalog service.')
```
clintonb: Don't change the exception type unnecessarily. Instead, log the error and re-raise the exception:

```python
logger.error(...)
raise
```

ecommerce/program_offers/conditions.py
```diff
@@ -15,15 +17,24 @@ def name(self):
 
     def _program(self, site_configuration):
         client = site_configuration.course_catalog_api_client
-        # TODO Cache the data
-        # TODO Handle errors
-        return client.programs(str(self.program_uuid)).get()
+        cache_key = hashlib.md5(
```
clintonb: There is no need to hash the key given that we know the characters are all ASCII.

clintonb:
@edx/helio please review

ecommerce/program_offers/conditions.py
```diff
@@ -0,0 +1,62 @@
...
+
+class ProgramCourseRunSeatsCondition(Condition):
+    class Meta(object):
+        proxy = True
+
+    @property
+    def name(self):
+        return 'Basket contains a seat for every course in program {}'.format(self.program_uuid)
+
+    def _program(self, site_configuration):
```
iivic: We will probably need programs in different places as well. Maybe moving this to an utils module could be helpful in the future?

ecommerce/program_offers/conditions.py
```diff
@@ -0,0 +1,62 @@
...
+            for course_run in course['course_runs']:
+                skus.update(set([seat['sku'] for seat in course_run['seats'] if seat['type'] in applicable_seat_types]))
+
+            # If there is no intersection of the SKUs in the basket and the SKUs
+            # of this course, the basket cannot meet the condition.
+            diff = basket_skus.difference(skus)
+            if diff == basket_skus:
+                return False
+
+            # Remove this course's SKUs the set of basket SKUs under consideration.
```
iivic: This comment is confusing. Please rewrite it or even remove it.

ecommerce/program_offers/views.py
```diff
@@ -0,0 +1,66 @@
+from django.contrib import messages
+from django.core.urlresolvers import reverse
+from django.utils.translation import ugettext_lazy as _
```
iivic: Use ugettext instead of ugettext_lazy.

iivic (approved):
A few comments left but it's looking good. Good job 👍

iivic (changes requested):
A migration is still missing.

ecommerce/program_offers/views.py
clintonb: Both functions produce similar output. `ugettext_lazy` is used here because we don't want the value of `ProgramOfferCreateView.success_message` to be evaluated too soon. I have seen issues in the past with `ugettext` breaking on class fields, and I'd like to avoid recreating those issues.

ecommerce/program_offers/conditions.py
clintonb: Updated.

ecommerce/program_offers/conditions.py
clintonb: I prefer to optimize for that situation when it arises. At the moment I don't fully understand the future use case(s).

ecommerce/core/models.py
```diff
@@ -359,6 +359,7 @@ def course_catalog_api_client(self):
             EdxRestApiClient: The client to access the Course Catalog service.
         """
 
+        # TODO Use URL from SiteConfiguration model.
```
vkaracic, line 362: Are you going to do this in this PR?

ecommerce/extensions/test/factories.py
```diff
@@ -104,3 +109,33 @@ class VoucherFactory(BaseVoucherFactory):  # pylint: disable=function-redefined
 
 class ConditionalOfferFactory(BaseConditionalOfferFactory):  # pylint: disable=function-redefined
     name = factory.Faker('word')
+
+
+class AbsoluteDiscountBenefitWithoutRangeFactory(BenefitFactory):
+    range = None
+    type = ''
```
vkaracic, line 116: If this is an absolute discount benefit shouldn't the type be ``Benefit.ABSOLUTE`` here and ``Benefit.PERCENTAGE`` down below?

ecommerce/programs/api.py
```diff
@@ -0,0 +1,35 @@
...
+        Returns:
+            dict
+        """
+        program_uuid = str(uuid)
+        cache_key = 'program-{uuid}'.format(uuid=program_uuid)
+
+        program = cache.get(cache_key)
+
+        if not program:
+            program = self.client.programs(program_uuid).get()
```
vkaracic: I don't see the usual exceptions (``ConnectionError``, ``Timeout``, ``SlumberException``) getting caught anywhere where this method is used. It might be good to handle those cases.

vkaracic (commented):
Reviewed ~1/4 of the changes just to realize that this PR is way too large to be effectively reviewed in one sitting.

ecommerce/programs/api.py
iivic: We concluded that it's unnecessary. Instead we will let it bubble up since it's not expected to happen often but if it happens the edx staff knows how to report issues.

iivic (approved):

ecommerce/extensions/test/factories.py
clintonb: We are using a custom benefit. No type is necessary. See https://django-oscar.readthedocs.io/en/releases-1.4/howto/how_to_create_a_custom_benefit.html.

ecommerce/core/models.py
clintonb: Probably not. As you pointed out, this PR is already large enough.

ecommerce/programs/api.py
clintonb: If consuming code needs that exception handling, it can add it. For the initial version, at least, we aren't going to bother with exception handling. If the  Programs API is not reachable, we have bigger problems than worrying about the admin UX during such an outage.

mjfrey (approved):

ecommerce/programs/conditions.py
```diff
@@ -0,0 +1,123 @@
...
+    def can_apply_condition(self, line):
+        """ Determines whether the condition can be applied to a given basket line. """
+        if not line.stockrecord_id:
+            return False
+
+        product = line.product
+        return line.stockrecord.partner_sku in self.get_applicable_skus(
+            line.basket.site.siteconfiguration) and product.get_is_discountable()
+
+    def get_applicable_lines(self, offer, basket, most_expensive_first=True):
```
douglashall, line 96: @clintonb Can you remember the reason why you implemented this function here? It looks logically equivalent to the implementation on [AbstractCondition](https://github.com/django-oscar/django-oscar/blob/1.4/src/oscar/apps/offer/abstract_models.py#L749). What am I missing?

PR #1263, opened 2017-05-03
[LEARNER-618] Coupon discount should supersede base Program discount

vkaracic:
Since we don't have official program discounts yet, I'm going with the assumption that they are going to be site-wide offers. I've created a 33% off site-wide offer and a 15% voucher offer, the way they behave after applying this PR is shown below:

out-1

FYI: @edx/helio

### DIFF
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -45,6 +45,8 @@ Applicator = get_class('offer.utils', 'Applicator')
 Basket = get_model('basket', 'Basket')
 Benefit = get_model('offer', 'Benefit')
 Catalog = get_model('catalogue', 'Catalog')
+Condition = get_model('offer', 'Condition')
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
 Product = get_model('catalogue', 'Product')
 ProductAttribute = get_model('catalogue', 'ProductAttribute')
 Selector = get_class('partner.strategy', 'Selector')
@@ -764,6 +766,51 @@ class VoucherAddViewTests(TestCase):
 
         for part in expected_url_parts:
             self.assertIn(part, resp.url)
+
+    def assert_basket_discounts(self, expected_offer_discounts=None, expected_voucher_discounts=None):
+        """Helper to determine if the expected offer is applied to a basket.
+        The basket is retrieved from the response because Oscar uses
+        SimpleLazyObjects to operate with baskets."""
+        expected_offer_discounts = expected_offer_discounts or []
+        expected_voucher_discounts = expected_voucher_discounts or []
+
+        response = self.client.get(reverse('basket:summary'))
+        basket = response.context['basket']
+
+        actual_offer_discounts = [discount['offer'] for discount in basket.offer_discounts]
+        actual_voucher_discounts = [discount['offer'] for discount in basket.voucher_discounts]
+
+        self.assertEqual(actual_offer_discounts, expected_offer_discounts)
+        self.assertEqual(actual_voucher_discounts, expected_voucher_discounts)
+
+    def test_coupon_applied_on_site_offer(self):
+        """Coupon offer supersedes site offer."""
+        product_price = 100
+        site_offer_discount = 20
+        voucher_discount = 10
+
+        voucher, product = prepare_voucher(benefit_value=voucher_discount)
+        stockrecord = product.stockrecords.first()
+        stockrecord.price_excl_tax = product_price
+        stockrecord.save()
+
+        _range = factories.RangeFactory(includes_all_products=True)
+        site_offer = factories.ConditionalOfferFactory(
+            offer_type=ConditionalOffer.SITE,
+            benefit=factories.BenefitFactory(range=_range, value=site_offer_discount),
+            condition=factories.ConditionFactory(type=Condition.COVERAGE, value=1, range=_range)
+        )
+        self.basket.add_product(product)
+        # Only site offer is applied to the basket.
+        self.assert_basket_discounts([site_offer])
+
+        # Only the voucher offer is applied to the basket.
+        self.client.post(reverse('basket:vouchers-add'), data={'code': voucher.code})
+        self.assert_basket_discounts(expected_voucher_discounts=[voucher.offers.first()])
+
+        # Site offer discount is still present after removing voucher.
+        self.client.post(reverse('basket:vouchers-remove', kwargs={'pk': voucher.id}))
+        self.assert_basket_discounts([site_offer])
 
 
 class VoucherRemoveViewTests(TestCase):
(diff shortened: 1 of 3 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1263&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev (approved):

ecommerce/extensions/basket/views.py
```diff
@@ -394,6 +394,7 @@ def get_context_data(self, **kwargs):
 
 class VoucherAddView(BaseVoucherAddView):  # pylint: disable=function-redefined
     def apply_voucher_to_basket(self, voucher):
+        self.request.basket.reset_offer_applications()
```
clintonb: Don't do this until after the voucher has been validated; otherwise, you break site-wide offers when the user enters an invalid code. This line should be right above line 420. It should also have a comment explaining why it is necessary.

ecommerce/extensions/test/factories.py
```diff
@@ -81,13 +82,19 @@ def prepare_voucher(code='COUPONTEST', _range=None, start_datetime=None, end_dat
...
             name='PrepareVoucherOffer',
             benefit=benefit,
             condition=condition,
             max_global_applications=max_usage
         )
     else:
-        offer = ConditionalOfferFactory(name='PrepareVoucherOffer', benefit=benefit, condition=condition)
+        offer = ConditionalOfferFactory(
+            offer_type=ConditionalOffer.VOUCHER,
+            name='PrepareVoucherOffer',
```
clintonb: Drop the name given that the factory will make one for you.

ecommerce/extensions/test/factories.py
```diff
@@ -81,13 +82,19 @@ def prepare_voucher(code='COUPONTEST', _range=None, start_datetime=None, end_dat
     condition = ConditionFactory(value=1, range=_range)
     if max_usage:
         offer = ConditionalOfferFactory(
+            offer_type=ConditionalOffer.VOUCHER,
             name='PrepareVoucherOffer',
```
clintonb: Drop the name given that the factory will make one for you.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -765,6 +767,39 @@ def test_redirects_with_enterprise_customer(self, get_ec):
...
+            condition=factories.ConditionFactory(type=Condition.COVERAGE, value=1, range=_range)
+        )
+        self.basket.add_product(product)
+
+        # Only site offer is applied to the basket.
+        Applicator().apply(self.basket)
+        self.assertEqual(self.basket.total_discount, site_offer_discount)
+
+        # Only the voucher offer is applied to the basket.
+        self.view.apply_voucher_to_basket(voucher)
```
clintonb: Please call the views via URLs (e.g. `self.client.get(...)`). I recall this sort of testing making the Django/Oscar upgrades unnecessarily difficult. Although these are unit tests, and going through URLs might be considered a functional test, going through URLs is how the user will actually experience the product. We should do our best to mimic actual behaviors in our tests (especially given our poor track record with acceptance tests).

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -765,6 +767,39 @@ def test_redirects_with_enterprise_customer(self, get_ec):
...
+        self.assertEqual(self.basket.total_discount, voucher_discount)
```
clintonb: Use the `Basket.offer_discounts` and `Basket.voucher_discounts` to ensure that your offers are actually applied (or not applied). Otherwise, an offer made by another test could fool this assertion. It's a stretch, I know, but we should be explicit as possible about what we are validating.

clintonb (changes requested):

vkaracic:
@clintonb comments addressed!

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -767,6 +767,15 @@ def test_redirects_with_enterprise_customer(self, get_ec):
         for part in expected_url_parts:
             self.assertIn(part, resp.url)
 
+    def assert_basket_discount(self, voucher_discount=0, offer_discount=0):
+        """Helper to determine how many voucher and/or offer discounts are applied
+        to a basket. The basket is retrieved from the response because Oscar uses
+        SimpleLazyObjects to manipulate with baskets."""
+        response = self.client.get(reverse('basket:summary'))
+        basket = response.context['basket']
+        self.assertEqual(len(basket.voucher_discounts), voucher_discount)
```
clintonb: Just as the basket value wasn't really relevant in the first iteration, the number of discounts is not relevant here. We want to assert that the discount we applied is actually applied.

clintonb (commented):
I'm ready to approve once the final comment is addressed.

vkaracic:
@clintonb take another look.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -765,6 +767,44 @@ def test_redirects_with_enterprise_customer(self, get_ec):
         for part in expected_url_parts:
             self.assertIn(part, resp.url)
 
+    def assert_basket_discount(self, expected_offer):
+        """Helper to determine if the expected offer is applied to a basket.
+        The basket is retrieved from the response because Oscar uses
+        SimpleLazyObjects to manipulate with baskets."""
+        response = self.client.get(reverse('basket:summary'))
+        basket = response.context['basket']
+        for offer in basket.applied_offers().values():
```
clintonb: This for-loop seems odd. Consider the following:

```python
def assert_basket_discounts(self, expected_offer_discounts=None, expected_voucher_discounts=None):
  expected_offer_discounts = expected_offer_discounts or []
  expected_voucher_discounts = expected_voucher_discounts or []

  response = self.client.get(reverse('basket:summary'))
  basket = response.context['basket']

  self.assertEqual(basket.offer_discounts, expected_offer_discounts)
  self.assertEqual(basket.voucher_discounts, expected_voucher_discounts)
```

clintonb (approved):
Please update `assert_basket_discount`.

ecommerce/extensions/basket/tests/test_views.py
vkaracic: The problem with that is that `offer_discounts` and `voucher_discounts` don't return a list of offers, but of dictionaries, example:
```
[{'description': None, 'offer': <ConditionalOffer: esse>, 'name': u'esse', 'discount': Decimal('10.00'), 'voucher': <Voucher: distinctio>, 'freq': 1, 'result': BasketDiscount(Decimal('10.00'))}]
```

clintonb: `actual = [discount['offer'] for discount in basket.offer_discounts]`

vkaracic: D'oh! Thanks.

PR #1264, opened 2017-05-03
Installing social-auth-core from Git

clintonb:
The pinned commit contains a fix that resolves the pipeline issues previously seen in production. We will switch to the PyPI version once a new release is available.

LEARNER-480


clintonb:
See https://github.com/python-social-auth/social-core/pull/74 for details on the resolved issue.

I have tested this manually locally, confirming via debugger that our desired pipeline is executed. I'm working on a functional test.

clintonb:
@edx/learner please review. This will allow us to resume deploying this service.

FYI @edx/devops

rlucioni (approved):

PR #1265, opened 2017-05-03
Updated npm dependencies

clintonb:
In addition to updating the dependencies, I have removed the usage of semver modifiers. This is the second time (that I recall) these floating dependencies have been the source of broken tests on Travis, but not locally. In this case karma 1.6.0 is broken due to the issue described at https://github.com/karma-runner/karma/issues/2675.

### DIFF
--- a/ecommerce/static/js/test/specs/models/course_model_spec.js
+++ b/ecommerce/static/js/test/specs/models/course_model_spec.js
@@ -340,7 +340,8 @@ define([
                 });
             });
 
-            describe('getOrCreateSeats', function () {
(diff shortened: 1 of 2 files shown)


clintonb:
FYI @edx/learner 

You may now say, "I told you so," @rlucioni.

bjacobel:
https://yarnpkg.com 🙃

rlucioni (approved):

clintonb:
Merged by https://github.com/edx/ecommerce/pull/1261.

PR #1267, opened 2017-05-05
Removed unnecessary validation from Benefit model

clintonb:
The benefit_type field’s artificial limits should not exist. Limits necessary for coupons should be enforced by the Coupons API endpoint. This change will enable the creation of custom benefits that specify a benefit type that is outside of “percentage” or “fixed”.

LEARNER-435


clintonb:
@edx/helio please review. This is needed for https://github.com/edx/ecommerce/pull/1262.

mjfrey (approved):

PR #1268, opened 2017-05-05
Fix the bug on javascript won't run on receipt page

schenedx:
ECOM-7716

@clintonb @AlasdairSwan @mjfrey Please review

With `define`, I found the javascript snippet here just won't run on the page.
Update it to `require` so it can run. I was able to verify this in the Chrome dev console on stage.

### DIFF
--- a/ecommerce/static/js/apps/order_receipt_app.js
+++ b/ecommerce/static/js/apps/order_receipt_app.js
@@ -1,0 +1,10 @@
+require([
+        'pages/receipt_page'
+    ],
+    function (ReceiptPage) {
+        'use strict';
+
+        new ReceiptPage();
+    }
+);
+
--- a/build.js
+++ b/build.js
@@ -37,7 +37,7 @@
             exclude: ['js/common']
         },
         {
-            name: 'js/pages/receipt_page',
+            name: 'js/apps/order_receipt_app',
             exclude: ['js/common']
         }
     ]
--- a/ecommerce/templates/edx/checkout/receipt.html
+++ b/ecommerce/templates/edx/checkout/receipt.html
@@ -12,7 +12,7 @@
 {% endblock title %}
 
 {% block javascript %}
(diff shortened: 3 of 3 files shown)


mjfrey (approved):

clintonb:
@edx/fedx-team can we get an approval, please?

clintonb (approved):
I'd like to get an approval from a front-end expert.

andy-armstrong:
If the file is defining a module to be loaded by another JavaScript file, then the code was correct as it was. However, if the file is being loaded directly, then you want to use ``require`` to make the code execute. If you load a file with ``define`` then all you've done is define the module for somebody else to invoke.

http://stackoverflow.com/questions/9507606/when-should-i-use-require-and-when-to-use-define

It looks like this file was intended to be loaded and executed directly, so in that case 👍

clintonb:
Thanks @andy-armstrong!

clintonb:
@schenedx I restarted the tests. I sense flakiness; although, not due to this change.

ecommerce/static/js/apps/order_receipt_app.js
```diff
@@ -0,0 +1,12 @@
+require([
+        'jquery',
```
clintonb: jQuery is never used. This can go away.

schenedx:
@clintonb That failed build was not due to flakiness. It was due to the fact that the new `order_receipt_app.js` was not included into the `build.js`
It is now included. The build should pass this time.

clintonb:
Noted. I had not noticed you changed the code beyond the original two lines fix from earlier.

schenedx:
@clintonb With the two line change, the jasmine unit test would fail.
Therefore I have to use a wrapper `app` file to initiate the code on `receipt_page.js`. It's the pattern we use for all other javascript apps. See https://github.com/edx/ecommerce/tree/master/ecommerce/static/js/apps

PR #1270, opened 2017-05-11
Add Coupon creation for Programs.

mjfrey:
Add support to create Coupons given a Program UUID.

LEARNER-938

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -68,6 +68,11 @@ def retrieve_offer(obj):
     return retrieve_voucher(obj).offers.first()
 
 
+def retrieve_range(obj):
+    """Helper method to retrieve the range from coupon."""
+    return retrieve_offer(obj).condition.range
+
+
 def retrieve_quantity(obj):
     """Helper method to retrieve number of vouchers. """
     return obj.attr.coupon_vouchers.vouchers.count()
@@ -563,6 +568,7 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     note = serializers.SerializerMethodField()
     num_uses = serializers.SerializerMethodField()
     payment_information = serializers.SerializerMethodField()
+    program_uuid = serializers.SerializerMethodField()
     quantity = serializers.SerializerMethodField()
     start_date = serializers.SerializerMethodField()
     voucher_type = serializers.SerializerMethodField()
@@ -576,10 +582,12 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
         return retrieve_benefit(obj).value
 
     def get_catalog_query(self, obj):
-        return retrieve_offer(obj).condition.range.catalog_query
+        offer_range = retrieve_range(obj)
+        return offer_range.catalog_query if offer_range else None
 
     def get_course_catalog(self, obj):
-        return retrieve_offer(obj).condition.range.course_catalog
+        offer_range = retrieve_range(obj)
+        return offer_range.course_catalog if offer_range else None
 
     def get_category(self, obj):
         category = ProductCategory.objects.filter(product=obj).first().category
@@ -605,9 +613,14 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
         return _('ACTIVE') if in_time_interval else _('INACTIVE')
 
     def get_course_seat_types(self, obj):
-        offer = retrieve_offer(obj)
-        course_seat_types = offer.condition.range.course_seat_types
-        return course_seat_types.split(',') if course_seat_types else course_seat_types
+        seat_types = []
+        offer_range = retrieve_range(obj)
+
+        if offer_range:
+            course_seat_types = offer_range.course_seat_types or ''
+            seat_types = course_seat_types.split(',')
+
+        return seat_types
 
     def get_email_domains(self, obj):
         offer = retrieve_offer(obj)
@@ -617,10 +630,9 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
         return retrieve_end_date(obj)
 
     def get_enterprise_customer(self, obj):
-        """
-        Get the Enterprise Customer UUID attached to a coupon.
-        """
-        return retrieve_offer(obj).condition.range.enterprise_customer
+        """ Get the Enterprise Customer UUID attached to a coupon. """
+        offer_range = retrieve_range(obj)
+        return offer_range.enterprise_customer if offer_range else None
 
     def get_last_edited(self, obj):
         history = obj.history.latest()
@@ -640,6 +652,10 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
         offer = retrieve_offer(obj)
         return offer.num_applications
 
+    def get_program_uuid(self, obj):
+        """ Get the Program UUID attached to the coupon. """
+        return retrieve_offer(obj).condition.program_uuid
+
     def get_payment_information(self, obj):
         """
         Retrieve the payment information.
@@ -657,16 +673,16 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
         return retrieve_start_date(obj)
 
     def get_seats(self, obj):
-        offer = retrieve_offer(obj)
-        _range = offer.condition.range
+        offer_range = retrieve_range(obj)
         request = self.context['request']
-        if _range.catalog:
-            stockrecords = _range.catalog.stock_records.all()
+
+        if offer_range and offer_range.catalog:
+            stockrecords = offer_range.catalog.stock_records.all()
             seats = Product.objects.filter(id__in=[sr.product.id for sr in stockrecords])
             serializer = ProductSerializer(seats, many=True, context={'request': request})
             return serializer.data
-        else:
-            return None
+
+        return {}
 
     def get_voucher_type(self, obj):
         return retrieve_voucher_usage(obj)
@@ -677,7 +693,7 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
             'benefit_type', 'benefit_value', 'catalog_query', 'course_catalog', 'category',
             'client', 'code', 'code_status', 'coupon_type', 'course_seat_types',
             'email_domains', 'end_date', 'enterprise_customer', 'id', 'last_edited', 'max_uses',
-            'note', 'num_uses', 'payment_information', 'price', 'quantity',
+            'note', 'num_uses', 'payment_information', 'program_uuid', 'price', 'quantity',
             'seats', 'start_date', 'title', 'voucher_type'
         )
 
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -25,6 +25,8 @@ from ecommerce.extensions.api.v2.views.coupons import CouponViewSet
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.extensions.voucher.models import CouponVouchers
 from ecommerce.invoice.models import Invoice
+from ecommerce.programs.constants import BENEFIT_MAP
+from ecommerce.programs.custom import class_path
 from ecommerce.tests.factories import ProductFactory, SiteConfigurationFactory
 from ecommerce.tests.mixins import ThrottlingMixin
 from ecommerce.tests.testcases import TestCase
@@ -34,6 +36,7 @@ Basket = get_model('basket', 'Basket')
 Benefit = get_model('offer', 'Benefit')
 Catalog = get_model('catalogue', 'Catalog')
 Category = get_model('catalogue', 'Category')
+Condition = get_model('offer', 'Condition')
 Course = get_model('courses', 'Course')
 Order = get_model('order', 'Order')
 Product = get_model('catalogue', 'Product')
@@ -76,6 +79,7 @@ class CouponViewSetTest(CouponMixin, CourseCatalogTestMixin, TestCase):
             'catalog_query': None,
             'course_seat_types': None,
             'email_domains': None,
+            'program_uuid': None,
         }
 
     def build_request(self):
@@ -199,7 +203,8 @@ class CouponViewSetTest(CouponMixin, CourseCatalogTestMixin, TestCase):
             'quantity',
             'start_datetime',
             'title',
-            'voucher_type'
+            'voucher_type',
+            'program_uuid'
         ]
         self.assertEqual(sorted(expected_cleaned_voucher_data_keys), sorted(cleaned_voucher_data.keys()))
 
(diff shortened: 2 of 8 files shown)


mjfrey:
@edx/helio please review.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -281,6 +290,28 @@ def test_create_course_catalog_coupon(self):
...
+        program_coupon = self.create_program_coupon(
+            coupon_title=coupon_title,
+            quantity=quantity,
+            program_uuid=program_uuid,
+            course_seat_types='verified',
+        )
+        self.assertEqual(program_coupon.title, coupon_title)
+
+        program_vouchers = program_coupon.attr.coupon_vouchers.vouchers.all()
+        program_voucher_offers = program_vouchers.first().offers
```
vkaracic: You don't seem to be be using ``program_voucher_offers`` anywhere.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -153,6 +154,14 @@ def create_course_catalog_coupon(self, coupon_title, quantity, course_catalog, c
             course_seat_types=course_seat_types,
         )
 
+    def create_program_coupon(self, coupon_title, quantity, program_uuid, course_seat_types):
```
vkaracic: This is unnecessary, you're just renaming ``create_coupon`` to ``create_program_coupon``.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -395,7 +401,7 @@ def create_update_data_dict(self, data, fields):
                     update_dict[field.replace('invoice_', '')] = value
         return update_dict
 
-    def update_coupon_benefit_value(self, benefit_value, coupon, vouchers):
+    def update_coupon_benefit_value(self, benefit_value, coupon, vouchers, program_uuid=None):
```
iivic: Add program_uuid to the docstring.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -334,9 +337,12 @@ def update(self, request, *args, **kwargs):
 
             self.update_range_data(request, vouchers)
 
+            program_uuid = request.data.get('program_uuid')
+
             benefit_value = request.data.get('benefit_value')
             if benefit_value:
```
vkaracic: What happens if only the ``program_uuid`` is sent for an update?

ecommerce/extensions/voucher/utils.py
```diff
@@ -284,28 +287,45 @@ def _get_or_create_offer(
...
-            value=Decimal(benefit_value),
-            max_affected_items=1,
+            type=Condition.COUNT,
+            value=1,
         )
+    try:
+        if program_uuid:
+            proxy_class = class_path(BENEFIT_MAP[benefit_type])
+            offer_benefit = Benefit()
+            offer_benefit.proxy_class = proxy_class
```
iivic: Can you use the [class](https://github.com/edx/ecommerce/blob/master/ecommerce/programs/benefits.py) directly instead of doing this? I mean replace the `Benefit()` with `PercentageDiscountBenefitWithoutRange()` or `AbsoluteDiscountBenefitWithoutRange()?`

vkaracic (changes requested):
You are missing tests for updating a coupon's program UUID. Also you could add an integration tests where you send the required data to the coupon create API endpoint and assert that everything is created as it's suppose to be.

ecommerce/extensions/voucher/utils.py
```diff
@@ -560,7 +587,8 @@ def get_voucher_discount_info(benefit, price):
         }
 
 
-def update_voucher_offer(offer, benefit_value, benefit_type, coupon, max_uses=None, email_domains=None):
+def update_voucher_offer(offer, benefit_value, benefit_type, coupon, max_uses=None,
+                         email_domains=None, program_uuid=None):
```
iivic, line 595: Add `program_uuid` to docstring.

ecommerce/extensions/voucher/tests/test_utils.py
mjfrey: I do on the next line.

ecommerce/extensions/voucher/utils.py
mjfrey: This is better.

ecommerce/extensions/voucher/tests/test_utils.py
vkaracic: What I wanted to say is that you could get away with ``program_voucher_offer = program_vouchers.first().offers.first()``.

ecommerce/extensions/api/v2/views/coupons.py
mjfrey: Right, I think you need both program_uuid and benefit value?  Is that correct @vkaracic

vkaracic: That depends on how you wanted your update logic to work. If you want to able to just adjust the program UUID then you don't need the benefit value (which wouldn't be in the request data anyway). Also be careful if you use ``program_uuid and benefit_value`` updating the non-program coupon benefit values won't work.

ecommerce/extensions/voucher/utils.py
clintonb: @iivic this is the Oscar-recommended method to create a custom `Benefit`.

ecommerce/extensions/voucher/utils.py
```diff
@@ -475,14 +497,18 @@ def create_vouchers(
...
-        product_range, __ = Range.objects.get_or_create(
-            name=range_name,
-            catalog=catalog,
-            catalog_query=catalog_query,
-            course_catalog=course_catalog,
-            course_seat_types=course_seat_types,
-            enterprise_customer=enterprise_customer,
-        )
+        # we do not need a range if this is for a Program
+        if not program_uuid:
```
clintonb: Nit: Invert so the positive check comes first `if program_uuid:`

iivic:
@edx/helio Please take another look.

ecommerce/extensions/api/serializers.py
```diff
@@ -576,10 +582,10 @@ def get_benefit_value(self, obj):
         return retrieve_benefit(obj).value
 
     def get_catalog_query(self, obj):
-        return retrieve_offer(obj).condition.range.catalog_query
+        return retrieve_range(obj).catalog_query if retrieve_range(obj) else None
```
vkaracic: Perhaps this instead of calling the chain of commands twice:
```
range = retrieve_range(obj)
return range.catalog_query if range else None
```

ecommerce/extensions/api/serializers.py
```diff
@@ -576,10 +582,10 @@ def get_benefit_value(self, obj):
...
 
     def get_course_catalog(self, obj):
-        return retrieve_offer(obj).condition.range.course_catalog
+        return retrieve_range(obj).course_catalog if retrieve_range(obj) else None
```
vkaracic: Same here.

ecommerce/extensions/api/serializers.py
```diff
@@ -617,10 +627,8 @@ def get_end_date(self, obj):
         return retrieve_end_date(obj)
 
     def get_enterprise_customer(self, obj):
-        """
-        Get the Enterprise Customer UUID attached to a coupon.
-        """
-        return retrieve_offer(obj).condition.range.enterprise_customer
+        """ Get the Enterprise Customer UUID attached to a coupon. """
+        return retrieve_range(obj).enterprise_customer if retrieve_range(obj) else None
```
vkaracic: Same here.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -1039,6 +1044,55 @@ def test_update_coupon_with_invalid_note(self, invalid_note):
...
+            'stock_record_ids': []
+        })
+        self.assertEqual(Benefit.objects.filter(proxy_class=proxy_class).count(), 0)
+        self.assertEqual(Condition.objects.filter(program_uuid=self.data['program_uuid']).count(), 0)
+
+        details = self._create_and_get_coupon_details()
+        self.assertEqual(details['program_uuid'], program_uuid)
+        self.assertEqual(details['title'], self.data['title'])
+        self.assertEqual(Benefit.objects.filter(proxy_class=proxy_class).count(), 1)
+        self.assertEqual(Condition.objects.filter(program_uuid=self.data['program_uuid']).count(), 1)
```
vkaracic, line 1082: From this line upwards you have almost the same code that you have in the test above. Either DRY it or don't test same things twice.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -395,23 +402,30 @@ def create_update_data_dict(self, data, fields):
...
+            benefit_value (Decimal): Benefit value associated with a new offer
+            program_uuid (str): Program UUID
         """
         voucher_offers = vouchers.first().offers
         voucher_offer = voucher_offers.first()
 
+        if program_uuid:
+            Condition.objects.filter(
+                program_uuid=voucher_offer.condition.program_uuid
+            ).update(program_uuid=program_uuid)
```
vkaracic, line 420: Perhaps you don't need to change this method. Just move the condition update code to the update method directly.
```
program_uuid = request.data.get('program_uuid')

if program_uuid:
  voucher_offer = vouchers.first().offers.first()
  Condition.objects.filter(
      program_uuid=voucher_offer.condition.program_uuid
  ).update(program_uuid=program_uuid)
```

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
+import uuid
```
vkaracic: There should be a space between special imports and standard library ones.

vkaracic (commented):
Mostly nits left.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
iivic: It's part of what I'm testing. I'm just making sure that the Benefit and Condition count is the same after updating the program_uuid. Relying on another test to check that seems silly.
Regarding the DRY... I don't like coupling tests I think it's smarter to keep every test for itself.

vkaracic: Actually in the upper test you are testing that the coupon is created and in this one you're testing that the coupon is created + coupon has been updated. Makes perfect sense to factor our the almost identical piece of code into a custom assert method.

iivic: Readability is also a factor so people that are looking at the code for the first time can see what is going on instead of switching between lines of code. I don't think the trade that we get with DRY is justified.
If you insist I will make the change but I would like to keep it as is.

ecommerce/extensions/api/serializers.py
clintonb: The pattern in the comment above should be followed to avoid calling `retrieve_range` twice. Use `offer_range` instead of `range` to avoid conflicting with the standard library function.

ecommerce/extensions/api/serializers.py
```diff
@@ -605,8 +611,12 @@ def get_code_status(self, obj):
...
 
     def get_course_seat_types(self, obj):
-        offer = retrieve_offer(obj)
-        course_seat_types = offer.condition.range.course_seat_types
+        offer_range = retrieve_range(obj)
+
+        if not offer_range:
+            return None
+
+        course_seat_types = offer_range.course_seat_types
```
clintonb: You run the risk of returning an empty string. This means we have three potential return types: null, empty string, an array. It would be ideal if we returned a single type.

```python
def get_course_seat_types(self, obj):
  seat_types = []
  offer_range = retrieve_range(obj)

  if offer_range:
    course_seat_types = offer_range.course_seat_types or ''
    seat_types = course_seat_types.split(',')

  return seat_types
```

ecommerce/extensions/api/serializers.py
```diff
@@ -657,11 +669,15 @@ def get_start_date(self, obj):
         return retrieve_start_date(obj)
 
     def get_seats(self, obj):
-        offer = retrieve_offer(obj)
-        _range = offer.condition.range
+        offer_range = retrieve_range(obj)
+
+        if not offer_range:
+            return None
```
clintonb: See comment above. We should return one data type.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
vkaracic: Leave it as is, since you're so passionate about it :smile:

iivic:
One more review.

mjfrey:
I can not approve since I started this PR.  But LGTM. 👍

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -334,9 +339,16 @@ def update(self, request, *args, **kwargs):
...
+            if program_uuid:
+                Condition.objects.filter(
+                    program_uuid=vouchers.first().offers.first().condition.program_uuid
+                ).update(program_uuid=program_uuid)
+
             benefit_value = request.data.get('benefit_value')
-            if benefit_value:
-                self.update_coupon_benefit_value(benefit_value=benefit_value, vouchers=vouchers, coupon=coupon)
+            if benefit_value or program_uuid:
+                self.update_coupon_offer(benefit_value=benefit_value, vouchers=vouchers,
```
vkaracic, line 345: Why are you updating this method?

marjev (approved):

ecommerce/extensions/voucher/utils.py
```diff
@@ -284,28 +287,49 @@ def _get_or_create_offer(
...
             range=product_range,
-            type=benefit_type,
-            value=Decimal(benefit_value),
-            max_affected_items=1,
+            type=Condition.COUNT,
+            value=1,
         )
+    try:
+        if program_uuid:
+            proxy_class = class_path(BENEFIT_MAP[benefit_type])
```
vkaracic, line 309: Since you're only using this if there is a benefit, I'd move it into the if-block below.

ecommerce/extensions/voucher/utils.py
```diff
@@ -284,28 +287,49 @@ def _get_or_create_offer(
...
+
+            if not offer_benefit:
+                offer_benefit = Benefit()
+                offer_benefit.proxy_class = proxy_class
+                offer_benefit.value = benefit_value
+                offer_benefit.save()
+
+            offer_name = "Coupon [{}]-{}".format(coupon_id, offer_benefit.name)
+        else:
+            offer_benefit, __ = Benefit.objects.get_or_create(
```
vkaracic, line 320: What happens if you only update the benefit value of a program coupon, would this create a new benefit for the same coupon?

vkaracic (approved):
If you want to treat the update as a PATCH (partial) update, you'll still need to account for scenarios where other than the program UUID fields are updated.
If NOT (since our UI sends all the fields on the update now), I guess you're good to go.

ecommerce/extensions/voucher/utils.py
iivic: It's used in the next line.

iivic:
@vkaracic Nice catch. I will merge it as is so we have something out but we have to take care of it. I will bring it up on standup.

PR #1271, opened 2017-05-11
Only send US formatted order value number to tracking LEARNER-964

schenedx:
Use the international API on javascript to format the order number for tracking consistently.

Please review @mjfrey @clintonb

### DIFF
--- a/ecommerce/extensions/checkout/tests/test_views.py
+++ b/ecommerce/extensions/checkout/tests/test_views.py
@@ -304,3 +304,13 @@ class ReceiptResponseViewTests(CourseCatalogMockMixin, LmsApiMockMixin, RefundTe
         body['course_key'] = seat.attr.course_key
         self.assertEqual(response.status_code, 200)
         self.assertTrue(response.context_data['display_credit_messaging'])
+
+    @httpretty.activate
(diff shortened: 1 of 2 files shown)


ecommerce/static/js/pages/receipt_page.js
```diff
@@ -9,9 +9,10 @@ define([
         'use strict';
 
         function trackPurchase(order_id, total_amount, currency) {
+            var purchaseAmount = new Intl.NumberFormat('en-US').format(parseFloat(total_amount));
```
clintonb: Can we fix this at a higher level: https://github.com/edx/ecommerce/blob/7ee38cd6e1624bb1dadbd2b773be24664dd3bff0/ecommerce/templates/edx/checkout/receipt.html#L27, please?

clintonb (changes requested):

ecommerce/static/js/pages/receipt_page.js
schenedx: @clintonb Why? Any particular concerns that made you think we should fix in the template?

clintonb (approved):
If possible, would you add a test confirming the fix?

clintonb (approved):
Remember to squash your commits before merging.

PR #1272, opened 2017-05-12
Added missing migration

clintonb:
LEARNER-435


clintonb:
It makes more sense to restore the `verbose_name` attribute than add a migration to keep it removed. See https://github.com/edx/ecommerce/pull/1274.

PR #1273, opened 2017-05-12
Log the error message when calculating basket discount.

vkaracic:
Let's see what fails.

### DIFF
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -390,8 +390,14 @@ class BasketCalculateView(generics.GenericAPIView):
                     'total_incl_tax': basket.total_incl_tax,
                     'currency': basket.currency
                 }
-                raise Exception
-        except:  # pylint: disable=bare-except
+                raise api_exceptions.TemporaryBasketException
+        except api_exceptions.TemporaryBasketException:
             pass
+        except:  # pylint: disable=bare-except
+            logger.exception(
+                'Failed to calculate basket discount for SKUs [%s] and voucher [%s].',
+                skus, code
+            )
+            raise
 
         return Response(response)
--- a/ecommerce/extensions/api/exceptions.py
+++ b/ecommerce/extensions/api/exceptions.py
@@ -28,3 +28,9 @@ class ProductNotFoundError(ApiError):
 
 class BadRequestException(APIException):
     status_code = status.HTTP_400_BAD_REQUEST
+
+
+class TemporaryBasketException(Exception):
+    """Exception raised to force a transaction rollback so that the temporary basket
+    is not saved to the database."""
+    pass
--- a/ecommerce/extensions/api/v2/tests/views/test_baskets.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_baskets.py
@@ -411,3 +411,13 @@ class BasketCalculateViewTests(TestCase):
 
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.data, expected)
+
+    @mock.patch('ecommerce.extensions.basket.models.Basket.add_product', mock.Mock(side_effect=Exception))
+    @mock.patch('ecommerce.extensions.api.v2.views.baskets.logger.exception')
+    def test_exception_log(self, mocked_logger):
+        """A log entry is filed when an exception happens."""
+        voucher, _ = prepare_voucher(_range=self.range, benefit_type=Benefit.FIXED, benefit_value=5)
+
+        with self.assertRaises(Exception):
+            self.client.get(self.url + '&code={code}'.format(code=voucher.code))
+            self.assertTrue(mocked_logger.called)


vkaracic:
@mjfrey

edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1273&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mjfrey (approved):
We will revert once we know why this is failing.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -391,7 +391,9 @@ def get(self, request):
                     'currency': basket.currency
                 }
                 raise Exception
-        except:  # pylint: disable=bare-except
-            pass
+        except Exception as e:  # pylint: disable=bare-except
+            logger.error('Cannot calculate basket discount. Error: %s', e.message)
```
clintonb: `logger.exception()`

clintonb (changes requested):

ecommerce/extensions/api/v2/views/baskets.py
vkaracic: Won't that raise the exception? I guess it doesn't matter anyway.

clintonb:
```python
except:
  logger.exception('Failed to calculate basket discount for SKUs [<comma-separated list>] and voucher [<code>].')
```

clintonb: Logging does not raise an exception. See https://docs.python.org/2/library/logging.html#logging.Logger.exception.

vkaracic:
@clintonb re-review please.

clintonb (approved):

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -391,7 +391,12 @@ def get(self, request):
...
-            pass
-
+        except Exception as e:  # pylint: disable=broad-except
+            logger.exception(
+                'Failed to calculate basket discount for SKUs [%s] and voucher [%s].',
+                skus, code
+            )
+
+        if not response:
+            return Response({}, status=404)
```
clintonb: 404 is wrong here.

clintonb (changes requested):

ecommerce/extensions/api/v2/views/baskets.py
vkaracic: 500?

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -391,7 +391,12 @@ def get(self, request):
                     'currency': basket.currency
                 }
                 raise Exception
```
clintonb: Why is this here?

ecommerce/extensions/api/v2/views/baskets.py
clintonb: Yes. Alternatively, just re-raise the exception after logging it. Django will handle the rest.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -391,7 +391,12 @@ def get(self, request):
...
-        except:  # pylint: disable=bare-except
-            pass
-
+        except Exception as e:  # pylint: disable=broad-except
```
clintonb: No need for `as e`.

clintonb (changes requested):

ecommerce/extensions/api/v2/views/baskets.py
mjfrey: That is to force the transaction roll-back.

clintonb: Please add a comment explaining that. We should create a custom exception. That exception can be caught and ignored. The rest can be logged. Otherwise, the logging below will just be noise in the logs.

vkaracic: @mjfrey yeah why is that there? :smile:

vkaracic:
@clintonb please re-review.

clintonb (approved):

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -390,8 +390,8 @@ def get(self, request):
                     'total_incl_tax': basket.total_incl_tax,
                     'currency': basket.currency
                 }
-                raise Exception
-        except:  # pylint: disable=bare-except
+                raise api_exceptions.TemporaryBasketException
+        except api_exceptions.TemporaryBasketException:
             pass
```
clintonb: There is still some value in catching unexpected exceptions, and logging the parameters that resulted in the exception being raised.

vkaracic: Agreed, I've added another except-block for that purpose.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -390,8 +390,13 @@ def get(self, request):
...
+        except:  # pylint: disable=bare-except
+            logger.exception(
+                'Failed to calculate basket discount for SKUs [%s] and voucher [%s].',
+                skus, code
+            )
```
clintonb, line 400: Re-`raise` the exception.

clintonb (approved):

PR #1277, opened 2017-05-12
Create temp basket with site for calculation

mjfrey:
When creating a temp basket used to calculate price / discount, we need to pass the site so things relying on it do not fail.


mjfrey:
@clintonb

clintonb (approved):

clintonb:
Test?

mjfrey:
there is already a test for this, how would I add one to test for the site?

clintonb:
I would write a test that replicates the error on stage/production.

mjfrey:
I see.  Create a program offer and make sure there is no error

clintonb (approved):
Squash your commits.

PR #1279, opened 2017-05-16
Require course seat upgrade deadline

mjfrey:
LEARNER-1021

### DIFF
--- a/ecommerce/static/js/models/course_seats/course_seat.js
+++ b/ecommerce/static/js/models/course_seats/course_seat.js
@@ -34,6 +34,10 @@ define([
 
                     // No validation is needed for empty values or seats not linked to courses.
                     if (_.isEmpty(value) || !course) {
+                        // Do not allow empty upgrade deadline for verified seats.
+                        if (this.getSeatType() === 'verified') {
+                            return gettext('Verified seats must have an upgrade deadline.');
+                        }
                         return;
                     }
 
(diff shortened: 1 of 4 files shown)


clintonb (approved):

ecommerce/static/js/models/course_seats/course_seat.js
```diff
@@ -34,6 +34,10 @@ define([
 
                     // No validation is needed for empty values or seats not linked to courses.
                     if (_.isEmpty(value) || !course) {
+                        // Do noy allow empty value for verified seats.
```
rlucioni: s/noy/not

ecommerce/static/js/models/course_seats/course_seat.js
```diff
@@ -34,6 +34,10 @@ define([
...
+                        if (this.getSeatType() == 'verified') {
+                            return gettext('Upgrade Dealine Required');
```
rlucioni: s/Dealine/Deadline

I'd make this a sentence for clarity, too: "Verified seats must have an upgrade deadline."

rlucioni (approved):
LGTM once test is fixed.

rlucioni:
Please remember to remove the text that appears under the field that refers to a 10-day default.

mjfrey:
@rlucioni Can you take another look?  Especially at the tests.  Want to make sure they are done correctly.

rlucioni (approved):
:shipit:

vkaracic:
Did anyone test this locally? After this change I cannot create a verified seat anymore no matter what value I put for the `Upgrade deadline` field. Doesn't work on stage either. You can try with a new course that I created `course-v1:edX+VK+01`.

Either I'm doing something wrong or there should be a bug ticket with higher priority filed for this bug since it does not allow creating verified seats anymore.

PR #1281, opened 2017-05-18
Add account activation gate toggle to coupon redemption flow.

douglashall:
This adds the ability to turn off the check to see if the user has activated their account before allowing
them to proceed with coupon redemption when using a coupon URL. The activation gate can be toggled for each
site.

ENT-307


ecommerce/core/models.py
```diff
@@ -151,6 +151,11 @@ class SiteConfiguration(models.Model):
         max_length=255,
         blank=True
     )
+    disable_account_activation_requirement = models.BooleanField(
```
clintonb: Nit: Invert, and rename this to `enable_account_activation_requirement`. It will make understanding the functionality a bit simpler.

douglashall:
Thanks @clintonb. I have made the change you requested. Is this looking good now?

clintonb (approved):
Update the translations with `make fake_translations`.

douglashall:
@clintonb Hmm, seems like `make fake_translations` picked up stuff not related to these changes. Is that okay? Or did I do something wrong?

mattdrayer (approved):
Change looks good!

mattdrayer:
re: translations -- I think that it's a "pick up the entire deck" sort of thing?

clintonb:
This is a case of other developers forgetting to update translations, and our failed check not failing the build. Squash commits, and you're good to go.

douglashall:
Cool. Thanks!

PR #1282, opened 2017-05-19
Updated e2e tests to load configuration from a file

clintonb:
LEARNER-509

### DIFF
--- a/e2e/config.py
+++ b/e2e/config.py
@@ -1,6 +1,14 @@
 import os
+from os.path import dirname, join
+
+from dotenv import load_dotenv
 
 from e2e.utils import str2bool
+
+# Load environment variables from an external file. Existing values will take precedence. The variables will be read
+# later to configure the tests. See https://github.com/theskumar/python-dotenv.
+dotenv_path = os.environ.get('DOTENV_PATH', join(dirname(__file__), '.env'))
+load_dotenv(dotenv_path)
 
 OAUTH_ACCESS_TOKEN_URL = os.environ.get('OAUTH_ACCESS_TOKEN_URL')
 OAUTH_CLIENT_ID = os.environ.get('OAUTH_CLIENT_ID')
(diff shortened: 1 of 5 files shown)


clintonb:
@edx/learner @edx/helio please review

e2e/requirements.txt
```diff
@@ -1,7 +1,8 @@
 # Packages required to run e2e tests
-bok-choy==0.6.2
+bok-choy==0.7.0
 ddt==1.1.1
 django-nose==1.4.4
 edx-rest-api-client==1.7.1
 nose-ignore-docstring==0.2
-selenium>=3.0.2
+python-dotenv==0.6.4
+selenium>=3.4.2
```
rlucioni: Please pin this! I thought we'd learned our lesson by now.

e2e/config.py
```diff
@@ -1,7 +1,14 @@
 import os
+from os.path import dirname, join
+
+from dotenv import load_dotenv
 
 from e2e.utils import str2bool
 
+# Load environment variables from an external file
+dotenv_path = os.environ.get('DOTENV_PATH', join(dirname(__file__), '.env'))
```
rlucioni: This is cool, but what's wrong with the more conventional YAML loading approach? Introducing a dependency for this seems like overkill.

clintonb: Why write more code when someone else has already done the work?

e2e/requirements.txt
clintonb: (okay)

e2e/config.py
rlucioni: Well, this dotenv business is now one more thing I need to know about when working in this system. Granted, it's pretty straightforward and I don't expect it to be a problem. I just would have gone with the YAML approach we see elsewhere and been done with it. ¯\\_(ツ)_/¯

rlucioni (approved):

PR #1283, opened 2017-05-19
[LEARNER-936] discount name on receipt page

vkaracic:
https://openedx.atlassian.net/browse/LEARNER-936

@edx/helio please review.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1283&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

vkaracic:
Here are screenshots of the change:
screenshot from 2017-05-19 14-56-33
screenshot from 2017-05-19 14-56-45

mjfrey:
@vkaracic screenshot?

marjev (approved):

ecommerce/extensions/checkout/views.py
```diff
@@ -145,7 +147,23 @@ def get(self, request, *args, **kwargs):
...
+        program_discount_text = None
+
+        # Currently only one type of offer can be applied to the basket.
+        # So the type of the first offer should be the same as (if any) other.
+        if order.basket_discounts:
+            offer = order.basket_discounts[0].offer
+
+            if offer.offer_type == ConditionalOffer.SITE:
+                client = ProgramsApiClient(self.request.site.siteconfiguration.course_catalog_api_client)
+                program = client.get_program(offer.condition.program_uuid)
```
mjfrey: Need to catch an error here and if so the receipt page should just say "Discount for Program"

mjfrey (changes requested):

ecommerce/extensions/checkout/views.py
```diff
@@ -145,7 +147,23 @@ def get(self, request, *args, **kwargs):
...
+                program_discount_text = 'Discount for the {program_title} {program_type} Program'.format(
```
clintonb: We should avoid special use cases for rendering whenever possible. Every if statement poses two new branches that need to be tested. Set the name on the conditional offer instance at https://github.com/edx/ecommerce/blob/master/ecommerce/programs/forms.py#L75. If the program name/type changes in the future, folks can simply re-save the model.

clintonb (changes requested):
Simplify.

vkaracic:
@clintonb take a look at the last commit, let me know if this is more appropriate.

mjfrey (approved):

ecommerce/extensions/checkout/views.py
```diff
@@ -18,6 +18,7 @@
 
 Applicator = get_class('offer.utils', 'Applicator')
 Basket = get_model('basket', 'Basket')
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
```
clintonb: No changes need to be made to this file.

ecommerce/extensions/checkout/tests/test_views.py
```diff
@@ -13,6 +13,7 @@
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.checkout.views import ReceiptResponseView
 from ecommerce.extensions.refund.tests.mixins import RefundTestMixin
+from ecommerce.programs.tests.mixins import ProgramTestMixin
```
clintonb: These changes are not necessary.

ecommerce/programs/forms.py
```diff
@@ -72,7 +77,18 @@ def clean(self):
     def save(self, commit=True):
         program_uuid = self.cleaned_data['program_uuid']
 
-        self.instance.name = 'Discount for program {}'.format(program_uuid)
+        try:
+            client = ProgramsApiClient(self.request.site.siteconfiguration.course_catalog_api_client)
+            program = client.get_program(program_uuid)
+            offer_name = 'Discount for the {program_title} {program_type} Program'.format(
```
clintonb: Mark for translation.

ecommerce/programs/forms.py
```diff
@@ -72,7 +77,18 @@ def clean(self):
...
+                program_title=program['title'],
+                program_type=program['type']
+            )
+        except:  # pylint: disable=bare-except
```
clintonb: Don't catch anything here. We don't want an offer with an invalid name in our system. Let the failure escalate to the user.

ecommerce/static/sass/partials/views/_receipt.scss
```diff
@@ -325,6 +325,10 @@
                     font-size: 15px;
                     font-weight: normal;
                 }
+
```
clintonb, line 328: Is this needed?

ecommerce/programs/tests/test_forms.py
```diff
@@ -95,3 +96,19 @@ def test_save_without_commit(self):
...
+        program_title = 'Test Program'
+        program_type = 'MicroMockers'
+
+        self.mock_program_detail_endpoint(data['program_uuid'], program_title=program_title, program_type=program_type)
+        form = ProgramOfferForm(request=self.request, data=data)
+        form.is_valid()
+        offer = form.save()
+
+        self.assert_program_offer_conditions(offer, data['program_uuid'], data['benefit_value'], data['benefit_type'])
+        self.assertIn(program_title, offer.name)
```
clintonb: Check the full name. New challenge for you: only use `assertIn` when checking the contents of a dict, or list.

ecommerce/programs/tests/mixins.py
```diff
@@ -40,6 +42,8 @@ def mock_program_detail_endpoint(self, program_uuid):
         program_uuid = str(program_uuid)
         data = {
             'uuid': program_uuid,
+            'title': program_title,
```
clintonb: Nit: either hardcode this or use `Faker`. There isn't a good reason to pass these values as `kwargs`.

ecommerce/programs/tests/test_forms.py
```diff
@@ -24,7 +26,6 @@ def generate_data(self, **kwargs):
 
     def assert_program_offer_conditions(self, offer, program_uuid, expected_benefit_value, expected_benefit_type):
         """ Assert the given offer's parameters match the expected values. """
-        self.assertEqual(offer.name, 'Discount for program {}'.format(program_uuid))
```
clintonb: Restore this, and pass in the program title and type. Alternatively, you can make a new method, `assertOfferName`, that is called separately from this one.

clintonb (changes requested):
Please post screenshots for the mobile, tablet, and desktop views of the receipt page.

clintonb:
Never mind. I see the screenshots.

ecommerce/programs/forms.py
vkaracic: Should I at least log the exception?

ecommerce/static/sass/partials/views/_receipt.scss
vkaracic: You mean the blank line? It is, before each new selector.
If you mean the `line-height` change, then yes, otherwise the characters overlap.

ecommerce/programs/tests/mixins.py
vkaracic: I'll hardcode it, no need to introduce another dependency.

ecommerce/programs/forms.py
clintonb: Nope. That will happen when the failure bubbles up to the user level.

vkaracic:
@clintonb comments addressed!

ecommerce/programs/forms.py
```diff
@@ -1,15 +1,19 @@
...
 
+from ecommerce.programs.api import ProgramsApiClient
 from ecommerce.programs.conditions import ProgramCourseRunSeatsCondition
 from ecommerce.programs.constants import BENEFIT_MAP, BENEFIT_PROXY_CLASS_MAP, BENEFIT_TYPE_CHOICES
 from ecommerce.programs.custom import class_path, create_condition
 
 Benefit = get_model('offer', 'Benefit')
 Condition = get_model('offer', 'Condition')
 ConditionalOffer = get_model('offer', 'ConditionalOffer')
+logger = logging.getLogger(__name__)
```
clintonb: Remove.

ecommerce/programs/tests/test_forms.py
```diff
@@ -24,7 +26,7 @@ def generate_data(self, **kwargs):
...
+        self.assertEqual(str(offer.name), 'Discount for the Test Program MicroMockers Program')
```
clintonb, line 29: No need for `str()`.

ecommerce/programs/tests/test_forms.py
```diff
@@ -38,6 +40,10 @@ def assert_form_errors(self, data, expected_errors):
         self.assertFalse(form.is_valid())
         self.assertEqual(form.errors, expected_errors)
 
+    def mock_program_api_endpoint(self, program_uuid):
+        httpretty.enable()
```
clintonb: Use the decorator on the appropriate test methods to ensure `httpretty` is disabled when tests finish.

clintonb (approved):
Address the latest comments, and squash your commits.

ecommerce/programs/tests/test_forms.py
vkaracic: Actually yes there is. Without it the string gets compared with the gettext function representation.

PR #1285, opened 2017-05-22
Program offers list page is displaying program vouchers info.

iivic:
LEARNER-1076

The program offers list page should only display program offers not program coupon data.
@edx/helio Please review.

### DIFF
--- a/ecommerce/programs/tests/test_views.py
+++ b/ecommerce/programs/tests/test_views.py
@@ -10,6 +10,7 @@ from ecommerce.programs.benefits import PercentageDiscountBenefitWithoutRange
 from ecommerce.programs.constants import BENEFIT_PROXY_CLASS_MAP
 from ecommerce.programs.custom import class_path
 from ecommerce.programs.tests.mixins import ProgramTestMixin
+from ecommerce.programs.views import ProgramOfferListView
 from ecommerce.tests.testcases import CacheMixin, TestCase
 
 Benefit = get_model('offer', 'Benefit')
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1285&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev (approved):

vkaracic (approved):

ecommerce/programs/tests/test_views.py
```diff
@@ -82,6 +83,24 @@ def test_get(self):
         response = self.assert_get_response_status(200)
         self.assertEqual(list(response.context['object_list']), program_offers)
 
+    def test_get_queryset(self):
+        """ Should return only Conditional Offers with Site offer type. """
+
+        # Conditional Offer should contain a condition with program uuid set in order to be returned
+        program_condition = factories.ProgramCourseRunSeatsConditionFactory()
```
clintonb, line 90: Use `ProgramOfferFactory`

ecommerce/programs/tests/test_views.py
```diff
@@ -82,6 +83,24 @@ def test_get(self):
...
+            offer_type=ConditionalOffer.SITE
+        )
+        factories.ConditionalOfferFactory(
+            condition=program_condition,
+            offer_type=ConditionalOffer.VOUCHER
+        )
+        factories.ConditionalOfferFactory(
+            offer_type=ConditionalOffer.SITE
+        )
+        self.assertEqual(list(ProgramOfferListView().get_queryset()), [site_conditional_offer])
```
clintonb, line 102: Please don't call views in this manner for testing. Follow the example of the other tests in this file and use `self.client.get()`. When you bypass the framework's view loader you bypass the setting of a request and other values. This may work now, but it will break if/when we make more significant changes to this view.

clintonb (commented):
Please update your tests to align with existing tests, and avoid pain for future development.

PR #1286, opened 2017-05-22
Upgrade ecommerce service to Django 1.10.

adeelkhan:
LEARNER-944
Subtask for LEARNER-480

WIP - Upgrades ecommerce service from 1.9.13 to 1.10.7


clintonb:
Make sure you review all of the release notes, and run `makemigrations`.

schenedx (approved):
👍

schenedx:
@clintonb @mjfrey I think both of you should review this PR

ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -31,6 +31,13 @@ class Command(BaseCommand):
 
     PAYPAL_CONFIG_KEY = "paypal"
 
+    def add_arguments(self, parser):
+
```
clintonb: Remove this unnecessary newline.

ecommerce/extensions/catalogue/management/commands/migrate_course.py
```diff
@@ -153,27 +152,29 @@ def _get_products(self, modes):
...
-                    help='Save the migrated data to the database. If this is not set, '
-                         'migrated data will NOT be saved to the database.'),
-        make_option('--site',
-                    action='store',
-                    dest='site_domain',
-                    default=None,
-                    help='Domain for the ecommerce site providing the course.'),
-    )
+    def add_arguments(self, parser):
+
```
clintonb: Remove this unnecessary newline.

ecommerce/extensions/catalogue/management/commands/convert_course.py
```diff
@@ -27,43 +26,37 @@ class Command(BaseCommand):
...
-            action='store',
-            dest='direction',
-            type='choice',
-            choices=(HONOR_TO_AUDIT, AUDIT_TO_HONOR),
-            default=HONOR_TO_AUDIT,
-            help='Which direction to convert the courses. Options are honor_to_audit, or audit_to_honor.'
-        )
-    )
+    def add_arguments(self, parser):
+
```
clintonb: Remove this unnecessary newline.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -609,7 +609,7 @@ def test_get_voucher_and_products_from_code(self):
 
     def test_no_product(self):
         """ Verify that an exception is raised if there is no product. """
-        code = FuzzyText().fuzz()
+        code = FuzzyText().fuzz().upper()
```
clintonb: Why does case matter?

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -609,7 +609,7 @@ def test_get_voucher_and_products_from_code(self):
...
         voucher = VoucherFactory(code=code)
```
clintonb: `VoucherFactory` already generates a code. Use it instead of unnecessary passing another one. (Yes, I know you didn't write this code; however, since you're touching it, you should take a small step to improve it.)

ecommerce/extensions/payment/management/tests/test_paypal_profile.py
```diff
@@ -50,21 +50,21 @@ def call_command_action(self, action, *args, **options):
 
     def test_list(self, mock_profile):
         mock_profile.all.return_value = [self.mock_profile_instance]
-        self.call_command_action("list")
+        self.call_command_action("list", self.TEST_ID, self. TEST_JSON)
```
clintonb: Why are these changes necessary? You have changed how the tests invoke the command in a way that is backwards-incompatible.

adeelkhan:
@clintonb Made changes, please review.

ecommerce/extensions/payment/management/commands/paypal_profile.py
```diff
@@ -31,6 +31,9 @@ class Command(BaseCommand):
 
     PAYPAL_CONFIG_KEY = "paypal"
 
+    def add_arguments(self, parser):
+        parser.add_argument('vargs', nargs='+', type=str)
```
clintonb: I am fine with all of the changes except for those to this file. You've added a new argument, and I'm not sure why. Let's pull these changes, and the updated requirements into a separate PR so that we can discuss them further without blocking your other updates.

clintonb (changes requested):

ecommerce/extensions/payment/management/commands/paypal_profile.py
clintonb: Why are any of these changes necessary? The [Django 1.10 release notes](https://docs.djangoproject.com/en/1.11/releases/1.10/) make no mention of changes to management command arguments/options.

adeelkhan: @clintonb I was finding the  [error](https://github.com/miracle2k/django-assets/issues/72) while upgrading. [Here](https://docs.djangoproject.com/en/1.9/releases/1.8/#extending-management-command-arguments-through-command-option-list) its mentioned its deprecated in 1.8, so have to change it to pass command arguments the new way. Now as the tests are executing command with variable no of arguments, so collecting them this way.

adeelkhan: @clintonb It seems for 1.9.13, command arguments are successfully passed in args variable, but changing to 1.10.7, no positional arguments are passing to args in handle() function. I have to use vargs to collect positional args and 1.10.7 is populating it into options inside handle. Please guide.

clintonb: See https://docs.python.org/2/howto/argparse.html for more info on argparse. At a minimum give these arguments relevant names.

adeelkhan: @clintonb Would variable name paypal_profile_arguments is fine?

clintonb: Take a look at the `args` class variable. This is not one variable, but two to four variables. I would rather see these listed out as explicit arguments/options, similar to our other commands, than continue doing whatever weirdness is going on here.

Again, let's separate this into a separate sub-task and pull-request. Modifying this management command is worthy of a separate review given the risk and scope of the changes.

adeelkhan:
@clintonb Please review

schenedx (approved):
👍

clintonb (approved):

PR #1287, opened 2017-05-23
Update program coupon: update only benefit value

iivic:
LEARNER-1077

I also included the comments from https://github.com/edx/ecommerce/pull/1285#discussion_r117739230. FYI @clintonb 

@edx/helio Please review.

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -419,6 +419,9 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
                 program_uuid=voucher_offer.condition.program_uuid
             ).update(program_uuid=program_uuid)
 
+        # The program uuid (if program coupon) is required for the benefit and condition update logic
+        program_uuid = program_uuid or voucher_offer.condition.program_uuid
+
         new_offer = update_voucher_offer(
             offer=voucher_offer,
             benefit_value=benefit_value or voucher_offer.benefit.value,
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -1093,6 +1093,25 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         self.assertEqual(Benefit.objects.filter(proxy_class=proxy_class).count(), 1)
         self.assertEqual(Condition.objects.filter(program_uuid=edited_program_uuid).count(), 1)
 
+    def test_update_program_coupon_benefit_value(self):
+        """Verify update benefit value for program coupon."""
+        self.data.update({
+            'program_uuid': str(uuid4()),
+            'title': 'Program Coupon',
+            'enterprise_customer': None,
+            'stock_record_ids': []
+        })
+        self.client.post(COUPONS_LINK, json.dumps(self.data), 'application/json')
+
+        edited_benefit_value = 92
+        coupon = Product.objects.get(title=self.data['title'])
+        response_data = self.get_response_json(
+            'PUT',
+            reverse('api:v2:coupons-detail', kwargs={'pk': coupon.pk}),
+            data={'benefit_value': edited_benefit_value}
+        )
+        self.assertEqual(response_data['benefit_value'], edited_benefit_value)
+
 
 class CouponCategoriesListViewTests(TestCase):
     """ Tests for the coupon category list view. """


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1287&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev (approved):

ecommerce/programs/tests/test_views.py
```diff
@@ -87,19 +86,18 @@ def test_get_queryset(self):
         """ Should return only Conditional Offers with Site offer type. """
 
         # Conditional Offer should contain a condition with program uuid set in order to be returned
-        program_condition = factories.ProgramCourseRunSeatsConditionFactory()
```
clintonb: Move this to a separate commit. It is not related to the other change.

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -2,10 +2,12 @@
 from django.core.exceptions import ValidationError
 from django.utils.timezone import now, timedelta
 from oscar.core.loading import get_model
-from oscar.test import factories
+from oscar.test import factories as oscar_factories
```
clintonb: I believe this module is imported by the factory module below. There should be no need to import this one again.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -419,6 +419,9 @@ def update_coupon_offer(self, coupon, vouchers, benefit_value=None, program_uuid
                 program_uuid=voucher_offer.condition.program_uuid
             ).update(program_uuid=program_uuid)
 
+        # The program uuid (if program coupon) is required for the benefit and condition update logic
+        program_uuid = coupon.program_uuid
```
vkaracic: `program_uuid` is sent as an argument to this method can you use that one?

iivic: When only the benefit value gets updated the program uuid will not be sent and thus it will not be possible to use it.

iivic:
@clintonb comments addressed.

ecommerce/extensions/api/v2/views/coupons.py
clintonb: Why not always send the program UUID? Branching of this sort—arguments only sent under overly-specific conditions—is a code smell to me.

ecommerce/extensions/catalogue/models.py
```diff
@@ -32,6 +32,12 @@ def is_enrollment_code_product(self):
     def is_coupon_product(self):
         return self.get_product_class().name == COUPON_PRODUCT_CLASS_NAME
 
+    @property
+    def program_uuid(self):
```
clintonb: This seems to have only been setup for the call in `ecommerce/extensions/api/v2/views/coupons.py`. That doesn't seem right to me.

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -37,6 +38,20 @@ def update_product_expires(self, product):
         product.save()
         return expiration_datetime
 
+    def _create_and_assert_product(self, expected_result, offer, product_class_name, short_code):
+        """Helper method that creates a product."""
```
clintonb: Creates a product for what purpose. Give some more detail here, please.

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -37,6 +38,20 @@ def update_product_expires(self, product):
         product.save()
         return expiration_datetime
 
+    def _create_and_assert_product(self, expected_result, offer, product_class_name, short_code):
```
clintonb: s/expected_result/expected_program_uuid

clintonb (changes requested):
The ticket, LEARNER-1077, is quite vague. I have no idea what the bug is. Can you provide more details, please? What are you trying to accomplish?

iivic:
@clintonb I don't know what is unclear to you but the ticket goal is simple.
Currently we are not able to update the discount value for a program coupon and this ticket is created in order to enable that.

ecommerce/extensions/api/v2/views/coupons.py
iivic: A PATCH request will be sent if any argument is updated and sending program_uuid every time seems even worse then this solution. Program uuid must be present if the benefit value is updated because that's the current logic of creating/updating program benefits.
I'm open for suggestions.

ecommerce/extensions/catalogue/models.py
iivic: For now that is the only place where it's used but hiding the whole chain of models seems like the right way to expose this data for a coupon. This will be useful in the future as well.

clintonb: This doesn't seem right. The `Product` class should not care about program UUIDs. For the time being, remove this.

clintonb (approved):
Please address the remaining comments, update your commit message, and squash your commits.

PR #1290, opened 2017-05-25
Updated models for multi-tenancy

clintonb:
LEARNER-1129


clintonb:
@edx/helio please review

ecommerce/sailthru/tests/test_signals.py
```diff
@@ -55,16 +52,13 @@ def test_signals_disabled(self, mock_log_error):
     @patch('ecommerce_worker.sailthru.v1.tasks.update_course_enrollment.delay')
     @patch('ecommerce.sailthru.signals.logger.error')
     def test_partner_not_supported(self, mock_log_error, mock_update_course_enrollment):
-        """ Verify Sailthru is not contacted if the Partner does not support Sailthru. """
-        site_configuration = SiteConfigurationFactory(partner__name='TestX')
-        site_configuration.partner.enable_sailthru = False
-        self.request.site.siteconfiguration = site_configuration
+        """ Verify Sailthru is not contacted if the feature is disabled for the site. """
+        self.site_configuration.enable_sailthru = False
```
vkaracic, line 56: It's not required to call ``save()`` here?

ecommerce/core/models.py
```diff
@@ -45,10 +44,10 @@ class SiteConfiguration(models.Model):
     )
     theme_scss_path = models.CharField(
         verbose_name=_('Path to custom site theme'),
-        help_text=_('Path to scss files of the custom site theme'),
+        help_text='DEPRECATED: THIS FIELD WILL BE REMOVED!',
```
vkaracic, line 47: When will this field be removed? Should we log a ticket to keep track of this?

ecommerce/extensions/partner/models.py
```diff
@@ -12,8 +12,7 @@ class Partner(AbstractPartner):
     # short_code is the unique identifier for the 'Partner'
     short_code = models.CharField(max_length=8, unique=True, null=False, blank=False)
     enable_sailthru = models.BooleanField(default=True, verbose_name=_('Enable Sailthru Reporting'),
-                                          help_text=_(
-                                              'Determines if purchases/enrolls should be reported to Sailthru.'))
+                                          help_text='DEPRECATED: Use SiteConfiguration!')
```
vkaracic, line 15: When do we plan to remove this field?

vkaracic (approved):
Just a few questions.

ecommerce/core/models.py
clintonb: These will be removed shortly after this code is released.

ecommerce/sailthru/tests/test_signals.py
clintonb: Not really, since the underlying code uses this exact instance. If we were, for example, calling a view we would need to save the instance to ensure the view loaded the updated instance.

ecommerce/extensions/partner/models.py
clintonb: This field will be removed at some point after this PR is merged/released.

PR #1291, opened 2017-05-25
Added site field to Course model

clintonb:
This commit makes the Course model multi-tenant, ensuring the CAT and API only provide data for a single tenant.

LEARNER-1129


vkaracic:
@clintonb I have a question about a TODO note that you left.
Does [ ``TODO Update API to set this`` ](https://github.com/edx/ecommerce/pull/1291/files#diff-b8728e7ddc2a7b07a4e6e19255ef1f77R33) mean that I should override the ``create`` method on the courses API endpoint class and set it to manually create a course?

clintonb:
There is no manual anything. The API should pull the site from the `request`.

ecommerce/core/migrations/0030_auto_20170526_1259.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
clintonb: This won't be necessary once https://github.com/edx/ecommerce/pull/1290 is merged. This commit should be removed.

vkaracic:
@clintonb is that done automatically? Because there is a test case that sends the name and ID to the API endpoint and it fails because it does not get the site.

clintonb:
Update the serializer.
-- 
Clinton Blackburn

ecommerce/courses/models.py
```diff
@@ -118,6 +119,13 @@ def enrollment_code_product(self):
                 return enrollment_code
         return None
 
+    @property
+    def partner(self):
+        if self.site:
+            return self.site.siteconfiguration.partner
+
+        return None
```
vkaracic: @clintonb once this is ran https://github.com/edx/ecommerce/pull/1291/files#diff-f137d26c66e1636b8e3cc6a127697be8R9 there would be no need for the if-block here, right?

vkaracic:
Ready for review @edx/helio

marjev (approved):

ecommerce/courses/tests/test_models.py
```diff
@@ -159,7 +159,7 @@ def test_create_seat_with_enrollment_code(self):
 
     def test_create_credit_seats(self):
         """Verify that the model's seat creation method allows the creation of multiple credit seats."""
-        course = Course.objects.create(id='a/b/c', name='Test Course')
+        course = Course.objects.create(id='a/b/c', name='Test Course', site=self.site)
```
iivic: Maybe use that CourseFactory you updated?

ecommerce/courses/tests/test_models.py
```diff
@@ -25,7 +25,7 @@ class CourseTests(CourseCatalogTestMixin, TestCase):
     def test_unicode(self):
         """Verify the __unicode__ method returns the Course ID."""
         course_id = u'edx/Demo_Course/DemoX'
-        course = Course.objects.create(id=course_id)
+        course = Course.objects.create(id=course_id, site=self.site)
```
iivic: Maybe use that CourseFactory you updated?

ecommerce/courses/tests/test_models.py
```diff
@@ -85,7 +85,7 @@ def test_publish_to_lms(self):
 
     def test_save_creates_parent_seat(self):
         """ Verify the save method creates a parent seat if one does not exist. """
-        course = Course.objects.create(id='a/b/c', name='Test Course')
+        course = Course.objects.create(id='a/b/c', name='Test Course', site=self.site)
```
iivic: Maybe use that CourseFactory you updated?

ecommerce/courses/tests/test_models.py
```diff
@@ -114,7 +114,7 @@ def assert_course_seat_valid(self, seat, course, certificate_type, id_verificati
 
     def test_create_or_update_seat(self):
         """ Verify the method creates or updates a seat Product. """
-        course = Course.objects.create(id='a/b/c', name='Test Course')
+        course = Course.objects.create(id='a/b/c', name='Test Course', site=self.site)
```
iivic: Maybe use that CourseFactory you updated?

ecommerce/courses/tests/test_models.py
```diff
@@ -194,8 +194,8 @@ def test_collision_avoidance(self):
         Sanity check verifying that course IDs which produced collisions due to a
         lossy slug generation process no longer do so.
         """
-        dotted_course = Course.objects.create(id='a/...course.../id')
-        regular_course = Course.objects.create(id='a/course/id')
+        dotted_course = Course.objects.create(id='a/...course.../id', site=self.site)
```
iivic: Use CourseFactory throughout the file. I will stop writing this comment.

iivic (changes requested):
Instead of using Course.objects to create courses in test files we should be using CourseFactory. The rest is looking good.

iivic (approved):

vkaracic:
@iivic comments addressed, please take another look.

ecommerce/extensions/api/serializers.py
```diff
@@ -280,8 +280,14 @@ class Meta(object):
...
 
 
+class SiteSerializer(serializers.ModelSerializer):
+    class Meta(object):
+        model = Site
+
+
 class CourseSerializer(serializers.HyperlinkedModelSerializer):
     id = serializers.RegexField(COURSE_ID_REGEX, max_length=255)
+    site = SiteSerializer(read_only=True)
```
clintonb, line 290: This is not necessary. Why was it added? The site is implied by the hostname through which the API is accessed.

vkaracic: So we don't care about this attribute being displayed on the API?

clintonb: We don't *need* this attribute to be returned by the API. Why do you think it is needed?

vkaracic: Ah, correct. I'll remove it.

PR #1292, opened 2017-05-25
Update jshint "predef" options

iivic:
This update will allow us to use beforeAll and afterAll in
jasmine tests.

@clintonb since you're the only contributor to this file please review.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -126,6 +126,8 @@
         "expect",
         "waitsFor",
         "runs",
+        "beforeAll",
+        "afterAll",
 
         // jQuery-Jasmine library.
         "loadFixtures",


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1292&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb:
This is fine, but it would be ideal if we dropped JSHint solely for ESLint, and used the edX confit at https://github.com/edx/eslint-config-edx.

iivic:
Can you create a separate ticket for that and assign it to me? For now I just want to fix the bugs I reported once I do that I can take a look how to switch to ESLint.

clintonb:
You have the power to make your own tickets.

clintonb (approved):

PR #1294, opened 2017-05-26
There is no benefit type for program coupons

iivic:
There is no discount/benefit type for program coupons since we use
proxy_class to determine the type. This commit should take care of
displaying the proper benefit type on coupon details/edit page and
basket page.

LEARNER-1087

### DIFF
--- a/ecommerce/extensions/offer/utils.py
+++ b/ecommerce/extensions/offer/utils.py
@@ -56,8 +56,16 @@ def format_benefit_value(benefit):
     Returns:
         benefit_value (str): String value containing formatted benefit value and type.
     """
+
+    # TODO: Find a better way to format benefit value so we can remove this import.
+    # Techdebt ticket: LEARNER-1317
+    # Import is here because of a circular dependency.
+    from ecommerce.programs.constants import BENEFIT_PROXY_CLASS_MAP
+
     benefit_value = _remove_exponent_and_trailing_zeros(Decimal(str(benefit.value)))
-    if benefit.type == Benefit.PERCENTAGE:
+    benefit_type = benefit.type or BENEFIT_PROXY_CLASS_MAP[benefit.proxy_class]
+
+    if benefit_type == Benefit.PERCENTAGE:
         benefit_value = _('{benefit_value}%'.format(benefit_value=benefit_value))
     else:
         converted_benefit = add_currency(Decimal(benefit.value))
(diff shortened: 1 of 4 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1294&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

iivic:
@edx/helio please review.

ecommerce/extensions/offer/utils.py
```diff
@@ -56,8 +56,14 @@ def format_benefit_value(benefit):
     Returns:
         benefit_value (str): String value containing formatted benefit value and type.
     """
+
+    # Circular dependency
+    from ecommerce.programs.constants import BENEFIT_PROXY_CLASS_MAP
```
marjev: For the record: I don't like that we have to resort to this.
https://s-media-cache-ak0.pinimg.com/736x/79/28/2b/79282b550b4bc6d13dfd6ea87e12eb79.jpg

marjev (approved):

ecommerce/extensions/offer/utils.py
clintonb: Where/how is `format_benefit_value` used? We already have methods of displaying conditional offer details. Why do we need to display the benefit details?

clintonb (changes requested):

ecommerce/extensions/offer/utils.py
iivic: In this case it is used to display the discount value for basket page. I'm not sure what you are trying to say. Can you elaborate?

clintonb: Oscar already has a mechanism of rendering this value. We use it on the receipt page: https://github.com/edx/ecommerce/blob/39151c18cd187f59fa73efcf8b6ebfd62c10228f/ecommerce/templates/edx/checkout/receipt.html#L100-L119. Replicate that behavior rather than add further customization.

iivic: I can't find that mechanism on oscar and the receipt page filter uses this method that I'm altering. Did I miss something here @clintonb ?

clintonb: Please file a techdebt ticket, and add a comment in the code. I don't have a good solution to this issue at the moment.

iivic:
@clintonb I created the techdebt ticket and updated the comment. First time doing it so let me know if the info is insufficient.

clintonb (approved):
Squash your commits, and update your commit title/message to tell me what you did rather than describe the bug.

"There is no benefit type for program coupons" isn't helpful when scrolling through commit messages. The actual fix here is that you, "Fixed basket page rendering of program coupons".

PR #1296, opened 2017-05-29
Add name attribute as a sequence to ProgramOfferFactory.

vkaracic:
Flaky errors like [this one](https://gist.github.com/vkaracic/1e7cfae4e1d65dda8a3d15c4f042d012) have become way too common on Travis. Using a sequence instead of relying on `Faker` to always pick a new word from Lorem Ipsum is a better solution.

FYI @edx/helio


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1296&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb (approved):

PR #1297, opened 2017-05-29
Remove site configuration API endpoint.

vkaracic:
The site configuration API endpoint has no purpose anymore. It was introduced here: https://github.com/edx/ecommerce/commit/78607311da1f97f64a65fc355b3721d3c43b6928 just so that the frontend could determine if the enrollment codes checkbox should be displayed or not (https://github.com/edx/ecommerce/commit/78607311da1f97f64a65fc355b3721d3c43b6928#diff-5f987f41c9f03e408b8f898a2d88df01R333). In the meanwhile the logic to determine that has changed which rendered this API endpoint obsolete.

Please review @edx/helio


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1297&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev (approved):

iivic (approved):

PR #1299, opened 2017-05-30
Fix creating program offer when coupon with the same uuid already exists

iivic:
LEARNER-1180

The goal is to enable creating program offers when a program coupon with the same uuid already exists.

@edx/helio please review.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1299&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

marjev (approved):

vkaracic (approved):

PR #1302, opened 2017-05-30
Change voucher factory code attribute to a sequence.

vkaracic:
Same issue as here: https://github.com/edx/ecommerce/pull/1296 happens for the `VoucherFactory.code` attribute. I couldn't find anymore unique attributes in our test factories so we should be fine after this change.

FYI @edx/helio


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1302&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb (approved):

PR #1303, opened 2017-05-30
Drop JSHint and use ESLint

iivic:
LEARNER-1167

We are replacing JSHint with ESLint entirely.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1303&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

iivic:
@clintonb please take a look at the change.
After running `make validate_js` we will get `2970 problems (2961 errors, 9 warnings)`.
 I tried https://github.com/TheSavior/ESLint-Formatter to format automatically in sublime but it refuses to format the code.
Can you suggest a solution or at least a different approach?

clintonb:
http://eslint.org/docs/user-guide/command-line-interface

Use `--fix`.

iivic:
@clintonb please review and maybe propose another reviewer?

package.json
```diff
@@ -10,13 +10,14 @@
     "requirejs": "2.3.3"
   },
   "devDependencies": {
+    "eslint": "^3.19.0",
```
clintonb: Pin all requirements to a specific release, rather than a range, to avoid issues with dependencies changing unexpetedly.

clintonb (approved):

AlasdairSwan (approved):
👍

PR #1307, opened 2017-06-02
[ENT-357] Display course run price with entitlement on Landing page & course description modal

replaceafill:
This PR modifies the `/api/v2/baskets/calculate` endpoint to look for vouchers in the Enterprise entitlements available to the user when a vocher code is not provided.

Related edx-enterprise PR: https://github.com/edx/edx-enterprise/pull/118

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_baskets.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_baskets.py
@@ -418,7 +418,7 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
         self.assertEqual(response.data, expected)
 
     def test_basket_calculate_fixed_coupon(self):
-        """ Verufy successful basket calculation for a fixed price voucher """
+        """ Verify successful basket calculation for a fixed price voucher """
         discount = 5
         voucher, _ = prepare_voucher(_range=self.range, benefit_type=Benefit.FIXED, benefit_value=discount)
 
@@ -442,3 +442,40 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
         with self.assertRaises(Exception):
             self.client.get(self.url + '&code={code}'.format(code=voucher.code))
             self.assertTrue(mocked_logger.called)
+
+    @mock.patch('ecommerce.extensions.api.v2.views.baskets.get_entitlement_voucher')
+    def test_basket_calculate_entitlement_voucher(self, mock_get_entitlement_voucher):
+        """ Verify successful basket calculation considering Enterprise entitlement vouchers """
+
+        discount = 5
+        voucher, _ = prepare_voucher(_range=self.range, benefit_type=Benefit.FIXED, benefit_value=discount)
+        mock_get_entitlement_voucher.return_value = voucher
+
+        # If the list of sku's contains more than one product no entitlement voucher is applied
+        response = self.client.get(self.url)
+
+        expected = {
+            'total_incl_tax_excl_discounts': self.product_total,
+            'total_incl_tax': self.product_total,
+            'currency': 'GBP'
+        }
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data, expected)
+
+        # If it's only one product, the entitlement voucher is applied
+        product = self.products[0]
+        qs = urllib.urlencode({'sku': [product.stockrecords.first().partner_sku]}, True)
+        url = '{root}?{qs}'.format(root=self.path, qs=qs)
+        product_total = product.stockrecords.first().price_excl_tax
+
+        response = self.client.get(url)
+
+        expected = {
+            'total_incl_tax_excl_discounts': product_total,
+            'total_incl_tax': product_total - discount,
+            'currency': 'GBP'
+        }
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data, expected)
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -14,6 +14,7 @@ from rest_framework import generics, status
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 
+from ecommerce.enterprise.entitlements import get_entitlement_voucher
 from ecommerce.extensions.analytics.utils import audit_log
 from ecommerce.extensions.api import data as data_api
 from ecommerce.extensions.api import exceptions as api_exceptions
@@ -334,7 +335,9 @@ class BasketCalculateView(generics.GenericAPIView):
         """ Calculate basket totals given a list of sku's
 
         Create a temporary basket add the sku's and apply an optional voucher code.
-        Then calculate the total proce less discounts.
+        Then calculate the total price less discounts. If a voucher code is not
+        provided apply a voucher in the Enterprise entitlements available
+        to the user.
 
         Arguments:
             sku (string): A list of sku(s) to calculate
@@ -361,6 +364,10 @@ class BasketCalculateView(generics.GenericAPIView):
         products = Product.objects.filter(stockrecords__partner=partner, stockrecords__partner_sku__in=skus)
         if not products:
             return HttpResponseBadRequest(_('Products with SKU(s) [{skus}] do not exist.').format(skus=', '.join(skus)))
+
+        # If there is only one product apply an Enterprise entitlement voucher
+        if not voucher and len(products) == 1:
+            voucher = get_entitlement_voucher(request, products[0])
 
         # We wrap this in an atomic operation so we never commit this to the db.
         # This is to avoid merging this temporary basket with a real user basket.


openedx-webhooks:
Thanks for the pull request, @replaceafill! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1307"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/extensions/api/v2/urls.py
```diff
@@ -74,7 +76,14 @@
 ]
 
 ENTERPRISE_URLS = [
-    url(r'^customers$', enterprise_views.EnterpriseCustomerViewSet.as_view(), name='enterprise_customers')
+    url(r'^customers/$', enterprise_views.EnterpriseCustomerViewSet.as_view(), name='enterprise_customers'),
+    url(
+        r'^sku_price_details_for_user/{sku},{username}/$'.format(
```
itsjeyd: @replaceafill Any specific reason for using a comma to separate `{sku}` and `{username}` (instead of a slash)?

In general, I'm wondering if we should make the URL pattern more generic, to make extending it more straightforward. For example, we might want to introduce additional endpoints for retrieving information related to a specific user in the future. So how about something like:

```python
r'^users/{username}/products/{sku}/price/$'.format(...)
```

ecommerce/enterprise/entitlements.py
```diff
@@ -204,3 +205,54 @@ def get_available_voucher_for_product(request, product, vouchers):
...
+    else:
+        result = price - benefit.value
+        if result < 0:
+            result = 0
+    return decimal.Decimal(result)
+
+
+def get_product_price_details_for_enterprise_customer_user(request, product, user):
+    """
+    Return price information for a product for enterprise customer user.
```
itsjeyd: @replaceafill How about:

```
Return price information for a product for a user who is associated with an enterprise customer.
```

(The current wording sounds like we're dealing with a user who is an enterprise customer.)

If you make that change, I think you can shorten the name of the function to `get_product_price_details_for_user`.

ecommerce/enterprise/entitlements.py
```diff
@@ -204,3 +205,54 @@ def get_available_voucher_for_product(request, product, vouchers):
...
     return None
+
+
+def get_discounted_price(price, voucher):
+    """
+    Calculate discounted price based on voucher benefit.
+
+    Arguments:
+        price (Decimal): original price
+        voucher (Voucher): voucher for getting the benift's type and value
```
itsjeyd: @replaceafill Typo: `benift`

ecommerce/enterprise/tests/test_entitlements.py
```diff
@@ -420,3 +422,36 @@ def test_is_course_in_enterprise_catalog_for_error_in_get_catalog_course_runs(se
         expected_number_of_requests = 1
         log_message = 'Unable to connect to Course Catalog service for catalog contains endpoint.'
         self._assert_is_course_in_enterprise_catalog_for_failure(expected_number_of_requests, log_message)
+
+    def test_get_discounted_price_fixed_benefit(self):
+        """
+        Verify that method "get_discounted_price" returns the substraction of
```
itsjeyd: @replaceafill Typo: `substraction`

Nit: The following wording would be more precise:

> Verify that for a fixed benefit, "get_discounted_price" returns the result of subtracting the benefit value from the original price.

ecommerce/enterprise/tests/test_entitlements.py
```diff
@@ -420,3 +422,36 @@ def test_is_course_in_enterprise_catalog_for_error_in_get_catalog_course_runs(se
...
+        coupon = self.create_coupon(benefit_type=Benefit.FIXED, benefit_value=500)
+        voucher = coupon.attr.coupon_vouchers.vouchers.first()
+        price = decimal.Decimal(200)
+        discounted_price = get_discounted_price(price, voucher)
+        self.assertEquals(decimal.Decimal(0), discounted_price)
+
+    def test_get_discounted_price_percentage_benefit(self):
+        """
+        Verify that method "get_discounted_price" returns a fraction equivalent
+        to the benefit value for a percentage benefit type.
```
itsjeyd: @replaceafill The method should return the result of subtracting a fraction equivalent to the benefit value from the original price (not the fraction itself):

> Verify that for a percentage benefit, "get_discounted_price" returns the result of subtracting a fraction equivalent to the benefit value from the original price.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -63,3 +73,46 @@ def test_get_customers(self, mock_client):
...
             }
         )
+
+
+class TestEnterpriseSkuPriceDetailsForUserView(EnterpriseServiceMockMixin, CouponMixin, TestCase):
+
+    @mock.patch('ecommerce.enterprise.entitlements.get_course_vouchers_for_learner')
+    @mock.patch('ecommerce.enterprise.entitlements.get_available_voucher_for_product')
+    @httpretty.activate
+    def test_get_price_details(self, voucher_for_product_mock, vouchers_for_learner_mock):
```
itsjeyd: @replaceafill Nit: Add docstring

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -63,3 +73,46 @@ def test_get_customers(self, mock_client):
...
+        toggle_switch('enable_enterprise_on_runtime', False)
+        self.assertEqual(result.status_code, 200)
+        self.assertJSONEqual(
+            result.content.decode('utf-8'),
+            {
+                'price': 9.99,
+                'code': coupon_code,
+                'final_price': 4.995,
+            }
+        )
```
itsjeyd: @replaceafill This test looks good, but it only covers one of two happy paths (the other happy path is when we're dealing with a non-staff user who is requesting price details for themselves), and none of the sad paths (non-staff user requesting price details for another user; non-existing user). Could you add tests for these paths as well?

itsjeyd (changes requested):
@replaceafill This is looking pretty good already. Most of my suggestions are about making cosmetic changes. I also had a question about URL patterns, and about extending the tests a bit.

clintonb (changes requested):
Let's have a chat about APIs. We should avoid making Enterprise-specific endpoints when they may benefit all of OpenEdX. The ability to get a price for a user opens up many possibilities on marketing sites. Also @mjfrey has already written a similar endpoint: /baskets/calculate/.

itsjeyd:
@brittneyexline @douglashall Pinging you on this as discussed at standup.

ecommerce/extensions/api/v2/urls.py
replaceafill: @itsjeyd No specific reason. I've just been working with the enrollment API lately and it uses that approach. I changed this to use your suggestion.

clintonb: Please don't follow any examples from the Enrollment API. It's an example of what not to do.

clintonb:
Are you progressing with an Enterprise-specific API? If so, why? What about the existing API is lacking?

replaceafill:
@clintonb I wanted to ask you a couple of questions today, but I decided to address @itsjeyd comments first.

Thank you for pointing me to the calculate endpoint. It seems to do part of what we need. I even tried extracting and reusing [this part](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/v2/views/baskets.py#L369-L392), but couldn't get the discount to work in my current tests. I decided to extract that part because calling the "whole" endpoint from a new endpoint didn't seem right.

My plan was to [get the entitlement voucher](https://github.com/edx/ecommerce/blob/master/ecommerce/enterprise/entitlements.py#L30-L49) and use it with the calculate endpoint code. But the way I've implemented the current endpoint [allows a user to query other user's data](https://github.com/replaceafill/ecommerce/blob/replaceafill/ENT-357/ecommerce/extensions/api/v2/views/enterprise.py#L30-L34) (I recently asked @haikuginger if this is even needed) and I'm not sure if the Applicator in Oscar's offers relies on request.user instead.

This is my first interaction with Ecommerce so any recommendation is highly appreciated.

clintonb:
@replaceafill, would you please describe the problem you are trying to solve?

replaceafill:
@clintonb Sure. The JIRA ticket is [ENT-357](https://openedx.atlassian.net/browse/ENT-357). The goal of the task is that when an Enterprise learner enrolls in a course of an Enterprise, they see a discounted price on the premium (paid) track based on the entitlements (coupons) associated with the Enterprise.

The enterprise code in Ecommerce is already able to [get the voucher](https://github.com/edx/ecommerce/blob/master/ecommerce/enterprise/entitlements.py#L30-L49) based on a product, so all I need to do is to get the Product from the SKU of the paid mode. Then I was thinking of (somehow) sending `voucher code + [product]` to the calculate  endpoint you mentioned.

clintonb:
1. (Not necessarily for you to answer) Why are entitlements implemented as vouchers instead of site-wide conditional offers?
2. Why not update the existing endpoint, `/api/v2/baskets/calculate/`, to account for *all* context, including entitlements?

replaceafill:
@clintonb Correct. The decision on 1. predates my involvement on Enterprise coding :)

Regarding 2. do you think we could make the `/api/v2/baskets/calculate` endpoint to accept an optional username and try to get the voucher code from it if an explicit code is not supplied?

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -63,3 +74,149 @@ def test_get_customers(self, mock_client):
...
+    )
+    @ddt.unpack
+    def test_get_price_details_user_checks_own_information(
+            self,
+            username,
+            is_staff,
+            voucher_for_product_mock,
+            _unused_vouchers_for_learner_mock
+    ):
+        # Create staff user
```
itsjeyd: @replaceafill After the recent updates, this should be just "Create user", or "Create (non-)staff user".

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -63,3 +74,149 @@ def test_get_customers(self, mock_client):
...
+        toggle_switch('enable_enterprise_on_runtime', False)
+
+        # Verify a forbidden response
+        self.assertEqual(result.status_code, 403)
+        self.assertJSONEqual(
+            result.content.decode('utf-8'),
+            {
+                'detail': 'You do not have permission to perform this action.'
+            }
+        )
```
itsjeyd: @replaceafill Thanks for adding these tests! They look good, but contain a good amount of duplication.

Please merge `test_get_price_details_staff_checks_other_user` and `test_get_price_details_non_staff_checks_other_user` using `ddt`. You could name the resulting test `test_get_price_details_other_user`. For the parts that verify the response you can vary behavior based on whether the current user `is_staff`.

It would be fine to have separate tests for "user checks own information" vs. "user checks information for other user", I think, but maybe you can extract some of the common parts from `test_get_price_details_user_checks_own_information` and the new `test_get_price_details_other_user` into (private) helper methods on the `TestEnterpriseSkuPriceDetailsForUserView` class.

Finally, it would be great if you could add docstrings to the tests :)

itsjeyd (changes requested):
@replaceafill Thanks for the updates. This is getting close, but the tests are not quite there yet (see inline comments). Since a final decision on the overall approach is still pending it might make sense to hold off on making further adjustments to the tests (although the feedback that I had should be fairly quick to incorporate).

This also needs a rebase, and (note to self) testing is blocked on the issue that I reported on [OC-2619](https://tasks.opencraft.com/browse/OC-2619).

clintonb:
@replaceafill what is the use case for passing a username as opposed to using the authenticated user? Yes, looking up the entitlements is exactly what I am proposing.

haikuginger:
@clintonb, re: your question about conditional offers, I believe we actually have some stuff in flight to switch over to those.

replaceafill:
@clintonb I'm going to modify the `/api/v2/baskets/calculate` endpoint to also be aware of Enterprise entitlements.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -442,3 +442,22 @@ def test_exception_log(self, mocked_logger):
         with self.assertRaises(Exception):
             self.client.get(self.url + '&code={code}'.format(code=voucher.code))
             self.assertTrue(mocked_logger.called)
+
+    @mock.patch('ecommerce.extensions.api.v2.views.baskets.get_entitlement_voucher')
+    def test_basket_calculate_entitlement_voucher(self, mock_get_entitlement_voucher):
+        """ Verufy successful basket calculation for a fixed price voucher """
```
itsjeyd: @replaceafill Typo: Verufy (please fix for docstring of `test_basket_calculate_fixed_coupon` as well).

Also, can you please update/extend this docstring to explain how the test differs from `test_basket_calculate_fixed_coupon`?

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -362,6 +363,15 @@ def get(self, request):
         if not products:
             return HttpResponseBadRequest(_('Products with SKU(s) [{skus}] do not exist.').format(skus=', '.join(skus)))
 
+        vouchers = []
```
itsjeyd: @replaceafill Please update the following part of the docstring of this method to take into account Enterprise entitlements:

> Create a temporary basket add the sku's and apply an optional voucher code. Then calculate the total proce less discounts.

There's also a typo in there: "proce"

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -362,6 +363,15 @@ def get(self, request):
...
+        if voucher is not None:
+            vouchers.append(voucher)
+        else:
```
itsjeyd: @replaceafill IIUC, passing a `code` to this view will cause this code to skip applying any Enterprise entitlements. Is this what we want? Or is there are use case where we'd pass a code *and* would want the view to apply Enterprise entitlements? Genuine question, CC @clintonb @brittneyexline @douglashall @haikuginger

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -4,10 +4,15 @@
...
 from django.core.urlresolvers import reverse
+from oscar.core.loading import get_model
 
 from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
 
 
+Catalog = get_model('catalogue', 'Catalog')
+ProductClass = get_model('catalogue', 'ProductClass')
+
+
```
itsjeyd: @replaceafill I think you can revert the changes in this file.

ecommerce/enterprise/tests/test_entitlements.py
```diff
@@ -22,6 +22,7 @@
 from ecommerce.tests.testcases import TestCase
 
 COURSE_CATALOG_API_URL = 'https://catalog.example.com/api/v1/'
+Benefit = get_model('offer', 'Benefit')
```
itsjeyd: @replaceafill I think you can revert the changes in this file.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -1,9 +1,13 @@
+from django.contrib.auth import get_user_model
+
 from rest_framework import generics
 from rest_framework.permissions import IsAdminUser, IsAuthenticated
 from rest_framework.response import Response
 
 from ecommerce.enterprise.utils import get_enterprise_customers
 
+User = get_user_model()
+
```
itsjeyd: @replaceafill I think you can revert the changes in this file.

itsjeyd (changes requested):
@replaceafill Thanks for changing the approach! This looks good aside from a few small things that you'll need to clean up. I also had a [question](https://github.com/edx/ecommerce/pull/1307#discussion_r120832216) about applying voucher codes vs. Enterprise entitlements, but we can resolve that during upstream review.

itsjeyd:
@replaceafill One more thing I forgot to mention earlier: Now that the approach has changed it would be worth updating the PR description to reflect that.

ecommerce/extensions/api/v2/views/baskets.py
replaceafill: @itsjeyd I have the same question.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -335,7 +335,9 @@ def get(self, request):
         """ Calculate basket totals given a list of sku's
 
         Create a temporary basket add the sku's and apply an optional voucher code.
-        Then calculate the total proce less discounts.
+        Then calculate the total price less discounts. If a voucher code is not
+        provided look for vouchers in the Enterprise entitlements available
```
itsjeyd: @replaceafill Nit: Replace "look for" (which is pretty generic) with "apply":

> If a voucher code is not provided, apply vouchers from the Enterprise entitlements for the current user instead.

itsjeyd (commented):
@replaceafill One more nit and another rebase, then this should be good to go (manual testing pending).

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -362,6 +365,15 @@ def get(self, request):
         if not products:
             return HttpResponseBadRequest(_('Products with SKU(s) [{skus}] do not exist.').format(skus=', '.join(skus)))
 
+        vouchers = []
+        if voucher is not None:
```
clintonb: `if voucher`

ecommerce/extensions/api/v2/views/baskets.py
clintonb: Look at how the basket page determines priority, and replicate that/extract it to a method you can use here. Generally, user-specified vouchers take precedence over other offers. We will let you shoot yourself in the foot if you choose to use an offer that gives you a worse discount than the one we've selected for you.

replaceafill:
@itsjeyd Rebased/squashed from 3fc7270

ecommerce/extensions/api/v2/views/baskets.py
replaceafill: @clintonb I reviewed how the basket page determines voucher priority and I found that only the [BasketSingleItemView](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/views.py#L43) uses Enterprise entitlements after failing to get a voucher from the request.

What I'm wondering now is how much sense does it make to apply multiple entitlement vouchers to the calculate endpoint (one for each sku). Maybe just adding this in [this part](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/v2/views/baskets.py#L364) of the endpoint would be enough:

```python
        # If it's a single product try to apply an Enterprise entitlement voucher                                               
        if not voucher and len(products) ==  1:
            voucher = get_entitlement_voucher(request, products[0])
```

What do you think?

brittneyexline: hey @clintonb, when you get a chance can you reply to the above implementation question in this thread? thanks!

clintonb: > What I'm wondering now is how much sense does it make to apply multiple entitlement vouchers to the calculate endpoint (one for each sku).

That's a question for the product owner. However, if we only use one entitlement voucher today, I see no reason to change that.

replaceafill:
@itsjeyd I just rebased this from c557da4

ecommerce/extensions/api/v2/views/baskets.py
replaceafill: Thanks @clintonb. I changed the implementation to use this approach. Could you please take a look and let me know if it's good to be merged?

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -362,6 +365,11 @@ def get(self, request):
         if not products:
             return HttpResponseBadRequest(_('Products with SKU(s) [{skus}] do not exist.').format(skus=', '.join(skus)))
 
+        # If there is only one product apply an Enterprise entitlement voucher
+        # This is to match the logic of the '/basket/single-item' view
+        if not voucher and len(products) == 1:
```
clintonb: I clearly misunderstood your initial question. Who is the product owner? Has he/she been notified of this limitation? What you are doing here is probably okay for now, but will be limiting in the future as Enterprise embraces multi-item baskets (which may be sooner, rather than later, for programs). The alternative to this approach is to retrieve multiple vouchers. Ultimately, the Enterprise Team will need to update entitlements to be conditional offers rather than vouchers.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -362,6 +365,11 @@ def get(self, request):
         if not products:
             return HttpResponseBadRequest(_('Products with SKU(s) [{skus}] do not exist.').format(skus=', '.join(skus)))
 
+        # If there is only one product apply an Enterprise entitlement voucher
+        # This is to match the logic of the '/basket/single-item' view
```
clintonb: I would not mention the old view since it will eventually be deleted. The justification for this logic should stand on its own merit.

clintonb (approved):
Squash your commits, and update your commit message to follow the guidelines at https://openedx.atlassian.net/wiki/display/ECOM/Developer+Onboarding.

replaceafill:
Thanks for your help @clintonb. I just rebased/squashed this from 3f91f6d

douglashall (approved):
Can you rebase this again?

replaceafill:
@douglashall Rebased from 02bb022

PR #1308, opened 2017-06-02
Can't edit course two times

iivic:
LEARNER-1293

@edx/helio This should fix the issue of editing a course two times in a row. But a few flaky tests appeared after this fix (three to be precise). I don't want to merge until this is resolved but I didn't find a solution yet. I will fix other issues at hand and then return to this one if someone doesn't find a solution in the mean time.

### DIFF
--- a/ecommerce/static/js/models/course_seats/course_seat.js
+++ b/ecommerce/static/js/models/course_seats/course_seat.js
@@ -30,7 +30,7 @@ define([
                 },
                 expires: function (value) {
                     var verificationDeadline,
-                        course = this.get('course');
+                        course = this.course || this.get('course');
 
                     // No validation is needed when the Seat is not linked to the course
                     if (!course || !course.get('id')) {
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1308&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb:
This issue was introduced by https://github.com/edx/ecommerce/pull/1288.

mjfrey (approved):

iivic:
@clintonb I would like to merge this if you're fine with it.

clintonb (approved):
Please provide some context in your commit message. The commit message tells a story—what changed and why. That story is incomplete in this commit message. 

Based on your comments it HipChat, you are not really sure why this is broken. At the very least your commit title should be something along the lines of: Fixed bug in CAT that prevented courses from being updated multiple times. This at least tells me why this commit is necessary.

PR #1309, opened 2017-06-06
switches added for disable/enable order and user table change/list view

tasawernawaz:
[LEARNER-864](https://openedx.atlassian.net/browse/LEARNER-864)


tasawernawaz:
@clintonb Please review this.

ecommerce/core/admin.py
```diff
@@ -23,6 +25,12 @@ class EcommerceUserAdmin(UserAdmin):
     )
     show_full_result_count = False
 
+    def has_change_permission(self, request, obj=None):
+        if waffle.switch_is_active(USER_LIST_VIEW_SWITCH):
```
clintonb: Add a comment explaining why this is necessary.

ecommerce/core/admin.py
```diff
@@ -1,6 +1,8 @@
+import waffle
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
 from django.utils.translation import ugettext_lazy as _
+from ecommerce.core.constants import USER_LIST_VIEW_SWITCH
```
clintonb: Fix your import groupings.

https://github.com/timothycrosley/isort

ecommerce/core/constants.py
```diff
@@ -10,6 +10,9 @@
 # Seat constants
 SEAT_PRODUCT_CLASS_NAME = 'Seat'
 
+# User constants
```
clintonb: This comment is not helpful. Actually explain what the switch is for.

ecommerce/extensions/order/admin.py
```diff
@@ -1,4 +1,6 @@
+import waffle
 from oscar.apps.order.admin import *  # noqa pylint: disable=wildcard-import,unused-wildcard-import
+from ecommerce.extensions.order.constants import ORDER_LIST_VIEW_SWITCH
```
clintonb: Fix.

ecommerce/extensions/order/constants.py
```diff
@@ -1,3 +1,6 @@
 class PaymentEventTypeName(object):
     PAID = 'Paid'
     REFUNDED = 'Refunded'
+
+# Order Constants
```
clintonb: Again, this comment is not helpful.

ecommerce/extensions/order/admin.py
```diff
@@ -23,6 +25,12 @@ def get_queryset(self, request):
         queryset = queryset.select_related('site', 'user', 'basket', )
         return queryset
 
+    def has_change_permission(self, request, obj=None):
+        if waffle.switch_is_active(ORDER_LIST_VIEW_SWITCH):
```
clintonb: Add a comment explaining what is going on here.

ecommerce/extensions/order/migrations/0014_auto_20170606_0535.py
```diff
@@ -0,0 +1,29 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+from ecommerce.extensions.order.constants import ORDER_LIST_VIEW_SWITCH
+
+
+def create_switch(apps, schema_editor):
+    """Create the `enable_order_list_view` switch if it does not already exist."""
+    Switch = apps.get_model('waffle', 'Switch')
+    Switch.objects.get_or_create(name=ORDER_LIST_VIEW_SWITCH, defaults={'active': True})
```
clintonb: `update_or_create`

ecommerce/core/migrations/0033_auto_20170606_0539.py
```diff
@@ -0,0 +1,29 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+from ecommerce.core.constants import USER_LIST_VIEW_SWITCH
+
+
+def create_switch(apps, schema_editor):
+    """Create the `enable_user_list_view` switch if it does not already exist."""
+    Switch = apps.get_model('waffle', 'Switch')
+    Switch.objects.get_or_create(name=USER_LIST_VIEW_SWITCH, defaults={'active': True})
```
clintonb: `update_or_create`

ecommerce/extensions/order/tests/test_admin.py
```diff
@@ -0,0 +1,29 @@
+from oscar.test.factories import UserFactory
+
+from ecommerce.tests.testcases import TestCase
+from ecommerce.core.tests import toggle_switch
```
clintonb: Alphabetize.

ecommerce/extensions/order/tests/test_admin.py
```diff
@@ -0,0 +1,29 @@
...
+class OrderAdminTests(TestCase):
+    """ Tests Admin page."""
+
+    password = 'test'
+
+    def setUp(self):
+        super(OrderAdminTests, self).setUp()
+        self.user = UserFactory(is_staff=True, is_superuser=True, password=self.password)
+        self.client.login(username=self.user.username, password=self.password)
+        self.order_page_url = '/admin/order/order/'
```
clintonb: Use `reverse()`: https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#reversing-admin-urls

ecommerce/extensions/order/tests/test_admin.py
```diff
@@ -0,0 +1,29 @@
+from oscar.test.factories import UserFactory
+
+from ecommerce.tests.testcases import TestCase
+from ecommerce.core.tests import toggle_switch
+from ecommerce.extensions.order.constants import ORDER_LIST_VIEW_SWITCH
+
+
+class OrderAdminTests(TestCase):
+    """ Tests Admin page."""
```
clintonb: Remove this docstring. It adds no value.

ecommerce/extensions/order/tests/test_admin.py
clintonb: Move this outside of setup. It should be fine to put this at the same level as `password`.

ecommerce/core/admin.py
clintonb: Also, where are the tests for this one?

clintonb (changes requested):

tasawernawaz:
@clintonb Please take another look.

ecommerce/extensions/order/tests/test_admin.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class OrderAdminTests(TestCase):
+
+    def setUp(self):
+        super(OrderAdminTests, self).setUp()
+        self.request.user = UserFactory(is_staff=True, is_superuser=True)
+        self.order_admin = OrderAdminExtended(create_order(), AdminSite())
+
+    def test_with_enable_order_list_view(self):
```
clintonb: Merge these into a single test with ddt.

ecommerce/core/tests/test_admin.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class UserAdminTests(TestCase):
+
+    def setUp(self):
+        super(UserAdminTests, self).setUp()
+        self.request.user = UserFactory(is_staff=True, is_superuser=True)
+        self.user_admin = EcommerceUserAdmin(UserFactory(), AdminSite())
+
+    def test_with_enable_order_list_view(self):
```
clintonb: Merge these into a single test with ddt.

ecommerce/extensions/order/tests/test_admin.py
tasawernawaz: Merging both tests needs a conditional statement in test before assertion, which I believe is not a better solution.

ecommerce/core/tests/test_admin.py
tasawernawaz: same here.

ecommerce/extensions/order/tests/test_admin.py
clintonb:
```python
@ddt.data(True, False)
def test_has_change_permission(self, switch_value):
  """ The method should only return True if the switch is enabled. """
  toggle_switch(ORDER_LIST_VIEW_SWITCH, switch_value)
  self.assertEqual(self.order_admin.has_change_permission(self.request), switch_value)
```

tasawernawaz:
@clintonb Made suggested changes.

ecommerce/core/admin.py
```diff
@@ -23,6 +25,10 @@ class EcommerceUserAdmin(UserAdmin):
     )
     show_full_result_count = False
 
+    def has_change_permission(self, request, obj=None):
+        # prevent users to access USER table (with large dataset) from django admin which was causing site break down
+        return True if waffle.switch_is_active(USER_LIST_VIEW_SWITCH) else False
```
clintonb: `return waffle.switch_is_active(USER_LIST_VIEW_SWITCH)`

ecommerce/extensions/order/admin.py
```diff
@@ -23,6 +26,10 @@ def get_queryset(self, request):
         queryset = queryset.select_related('site', 'user', 'basket', )
         return queryset
 
+    def has_change_permission(self, request, obj=None):
+        # prevent users to access ORDER table (with large dataset) from django admin which was causing site break down
+        return True if waffle.switch_is_active(ORDER_LIST_VIEW_SWITCH) else False
```
clintonb: `waffle.switch_is_active(ORDER_LIST_VIEW_SWITCH)`

clintonb (commented):
I just realized that this disables the change view. We only want to disable the list view.

tasawernawaz:
@clintonb Good catch. There isn't any permission related to list view. We can override `get_queryset`  and return empty set or we can override `changelist_view` and raise `PermissionDenied` exception. Your thoughts will be appreciated here.

ecommerce/core/tests/test_admin.py
```diff
@@ -0,0 +1,23 @@
...
+@ddt.ddt
+class UserAdminTests(TestCase):
+
+    def setUp(self):
+        super(UserAdminTests, self).setUp()
+        self.request.user = UserFactory(is_staff=True, is_superuser=True)
+        self.user_admin = EcommerceUserAdmin(UserFactory(), AdminSite())
+
+    @ddt.data(True, False)
+    def test_has_change_permission(self, switch_value):
```
schenedx: We want to restrict view permission right?. So this should be renamed to`test_has_view_permission`?

ecommerce/extensions/order/tests/test_admin.py
```diff
@@ -0,0 +1,23 @@
...
+@ddt.ddt
+class OrderAdminTests(TestCase):
+
+    def setUp(self):
+        super(OrderAdminTests, self).setUp()
+        self.request.user = UserFactory(is_staff=True, is_superuser=True)
+        self.order_admin = OrderAdminExtended(create_order(), AdminSite())
+
+    @ddt.data(True, False)
+    def test_has_change_permission(self, switch_value):
```
schenedx: Same comment as above.  Should this be renamed to `test_has_view_permission`?

schenedx (changes requested):
I like the idea to override `changelist_view`. I don't know if `permission_denied` is the right exception to raise. Do we need to raise exception? Shouldn't we override the view template in the view? We should tell our users who try to access this view that they are not going to get the list of orders? The message should also include that list of orders are too many and to expansive to render.

tasawernawaz:
@schenedx Implemented your suggestion. Please review.

schenedx (approved):
👍 Nice work

tasawernawaz:
@clintonb Needs your review.

clintonb (approved):

schenedx:
@tasawernawaz Can you please verify this change on stage now?
Looks like the ecom stage deployment will be done 10min or less from now.

tasawernawaz:
@schenedx Verified.

PR #1313, opened 2017-06-08
Added Optimizely snippet to edX base template

clintonb:
LEARNER-1171


clintonb:
FYI @edx/learner

mulby:
:+1:

PR #1314, opened 2017-06-08
Updated js tests

iivic:
LEARNER-1359

@clintonb finally found the source of the flakiness! There where two course seats with the same id in different files. Depending on which one was created first it would be used in the validation.
Please review.

### DIFF
--- a/ecommerce/static/js/models/course_seats/course_seat.js
+++ b/ecommerce/static/js/models/course_seats/course_seat.js
@@ -53,24 +53,26 @@ define([
                         return gettext('The upgrade deadline must occur BEFORE the verification deadline.');
                     }
                     return undefined;
+                },
+                title: {
+                    required: false,
+                    pattern: 'productName'
                 }
             },
 
             // TODO Determine how to use the extended seatType attribute of child classes with Backbone.Relational
             // http://backbonerelational.org/#RelationalModel-subModelTypes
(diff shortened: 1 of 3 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1314&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/static/js/models/course_seats/course_seat.js
```diff
@@ -53,24 +53,26 @@ define([
...
+                title: {
+                    required: false,
+                    pattern: 'productName'
                 }
             },
 
             // TODO Determine how to use the extended seatType attribute of child classes with Backbone.Relational
             // http://backbonerelational.org/#RelationalModel-subModelTypes
             getSeatType: function() {
                 switch (this.get('certificate_type')) {
```
clintonb:
```javascript
var certificateType = this.get('certificate_type');
switch (certificateType) {
  case 'credit':
  case 'honor':
  case 'verified':
    return certificateType;
  case 'professional':
  case 'no-id-professional':
    return 'professional';
  default:
    return 'audit';
}
```

clintonb (approved):
(nice)

PR #1316, opened 2017-06-09
Added management command to create demo data

clintonb:
LEARNER-1375

Depends on https://github.com/edx/edx-platform/pull/15287

### DIFF
--- a/ecommerce/core/management/commands/create_demo_data.py
+++ b/ecommerce/core/management/commands/create_demo_data.py
@@ -1,0 +1,83 @@
+import datetime
+
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+from oscar.core.loading import get_model
+from waffle.models import Flag
+
+from ecommerce.courses.models import Course
+
+Partner = get_model('partner', 'Partner')
+
+
+class Command(BaseCommand):
+    help = 'Create demo data. This should only be run in development environments!'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--partner',
+            action='store',
+            dest='partner_code',
+            type=str,
+            required=True,
+            help='Code for the partner with which the course should be associated'
+        )
+        parser.add_argument(
+            '--course-id',
+            action='store',
+            dest='course_id',
+            type=str,
+            default='course-v1:edX+DemoX+Demo_Course',
+            help='ID of the course to be created/updated. Defaults to course-v1:edX+DemoX+Demo_Course.'
+        )
+        parser.add_argument(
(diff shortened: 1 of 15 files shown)


ecommerce/core/management/commands/create_demo_data.py
```diff
@@ -0,0 +1,56 @@
...
+
+from ecommerce.courses.models import Course
+
+Partner = get_model('partner', 'Partner')
+
+
+class Command(BaseCommand):
+    help = 'Create demo data. This should only be run in development environments!'
+    course_id = 'course-v1:edX+DemoX+Demo_Course'
+    course_title = 'edX Demonstration Course'
```
mjfrey: Can these be arguments?  So we can use this to batch create demo course runs.

mjfrey (changes requested):

ecommerce/core/management/commands/create_demo_data.py
clintonb: Share more, please. I thought about making these arguments, but wanted to keep it simple (since it was 2AM). What's the use case for making these command line arguments?

mjfrey: We have always wanted to have a way to batch create demo data for development.  We have an sensible role that creates at least 3 demo courses in LMS, would be nice if that could be extended to also create the corresponding seats.

clintonb: @mjfrey updated.

ecommerce/core/management/commands/create_demo_data.py
```diff
@@ -0,0 +1,81 @@
...
+            help='Title of the course to be created/updated. Defaults to edX Demonstration Course.'
+        )
+        parser.add_argument(
+            '--price',
+            action='store',
+            dest='price',
+            type=int,
+            default=149,
+            help='Cost of the verified seat. Defaults to 149'
+        )
```
mjfrey, line 48: Can we also add seat_type?  Would be nice to be able to vary between verified and professional.

clintonb: Eh...now we are getting into an area of customization that I'd rather not deal with at this time. ID verification, or no? Credit, or no? Which credit provider?

As is, this command works for at least 90% of usecases. If others really need to automate other seat types, I'd rather they contribute the change.

mjfrey: Alright fine. :)

mjfrey (approved):

PR #1318, opened 2017-06-09
Passing SKU to ecom_worker for sailthru calls.

adeelkhan:
LEARNER-1378 - Subtask for LEARNER-1193

This explicit passing would simplify ecom-worker implementation. Ecom-worker instead of using ecom-api for fetching sku, it would get this from ecommerce as part of signal handling code.

### DIFF
--- a/ecommerce/sailthru/signals.py
+++ b/ecommerce/sailthru/signals.py
@@ -55,6 +55,7 @@ def process_checkout_complete(sender, order=None, user=None, request=None,  # py
 
         # get product
         product = line.product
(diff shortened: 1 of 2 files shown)


schenedx:
@adeelkhan Is there a unit test we can write for this change?

schenedx:
@adeelkhan Also, you may need to merge the update on ecom-worker first before your change here can merge.

schenedx (approved):
👍

ahsan-ul-haq (approved):

PR #1319, opened 2017-06-09
Added payment processor configuration for devstack

clintonb:
This change enables developers to quickly test payments without manual configuration.

LEARNER-1375


clintonb:
@edx/learner please review

schenedx (approved):
👍 This is needed for a while. Thank you for getting this configured!

PR #1321, opened 2017-06-12
LEARNER 1270: Allow localization of Paypal checkout experience

stevenzheng3791:
Allow locale setting for Paypal checkout. Currently does not set the locale (default to US). But when someone implements locale detection in the future, then localization capability is already supported.

LEARNER-1270

### DIFF
--- a/ecommerce/extensions/payment/processors/paypal.py
+++ b/ecommerce/extensions/payment/processors/paypal.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 import logging
+import uuid
 from decimal import Decimal
 from urlparse import urljoin
 
@@ -28,6 +29,7 @@ class Paypal(BasePaymentProcessor):
 
     NAME = 'paypal'
     DEFAULT_PROFILE_NAME = 'default'
+    DEFAULT_LOCALE_CODE = 'default'
 
     def __init__(self, site):
         """
@@ -60,6 +62,40 @@ class Paypal(BasePaymentProcessor):
     @property
     def error_url(self):
         return get_ecommerce_url(self.configuration['error_path'])
+
+    def resolve_paypal_locale(self):
+        """ TODO (LEARNER 1278): Generate appropriate Paypal locale code based on region """
+        return self.DEFAULT_LOCALE_CODE
+
+    def create_temporary_web_profile(self, locale_code):
+        """
+        Generates a temporary Paypal WebProfile that carries the locale setting for a Paypal Payment
+        and returns the id of the WebProfile
+        """
+        try:
+            web_profile = paypalrestsdk.WebProfile({
+                "name": str(uuid.uuid1()),  # Generate a unique identifier
+                "presentation": {
+                    "locale_code": locale_code
+                },
+                "temporary": True  # Persists for 3 hours
+            }, api=self.paypal_api)
+
+            if web_profile.create():
+                msg = "Web Profile[%s] for locale %s created successfully" % (
+                    web_profile.id,
+                    web_profile.presentation.locale_code
+                )
+                logger.info(msg)
+                return web_profile.id
+            else:
+                msg = "Web profile creation encountered error [%s]. Will continue without one" % (
+                    web_profile.error
+                )
+                logger.warning(msg)
+
+        except Exception:  # pylint: disable=broad-except
+            logger.warning("Creating PayPal WebProfile resulted in exception. Will continue without one.")
 
     def get_transaction_parameters(self, basket, request=None, use_client_side_checkout=False, **kwargs):
         """
@@ -112,11 +148,23 @@ class Paypal(BasePaymentProcessor):
             }],
         }
 
-        try:
-            web_profile = PaypalWebProfile.objects.get(name=self.DEFAULT_PROFILE_NAME)
-            data['experience_profile_id'] = web_profile.id
-        except PaypalWebProfile.DoesNotExist:
-            pass
+        if waffle.switch_is_active('create_and_set_webprofile'):
+            # TODO (LEARNER 1278) - set locale code based on user language
+            locale_code = self.resolve_paypal_locale()
+            if locale_code != self.DEFAULT_LOCALE_CODE:
+                # Create a temporary web profile with the correct locale code
+                web_profile_id = self.create_temporary_web_profile(locale_code)
+                if web_profile_id is not None:
+                    data['experience_profile_id'] = web_profile_id
+            else:
+                # Proceed without web profile (Paypal defaults to english)
+                logger.info("Using default locale. Will not create and set web profile")
+        else:
+            try:
+                web_profile = PaypalWebProfile.objects.get(name=self.DEFAULT_PROFILE_NAME)
+                data['experience_profile_id'] = web_profile.id
+            except PaypalWebProfile.DoesNotExist:
+                pass
 
         available_attempts = 1
         if waffle.switch_is_active('PAYPAL_RETRY_ATTEMPTS'):
--- a/ecommerce/extensions/payment/tests/processors/test_paypal.py
+++ b/ecommerce/extensions/payment/tests/processors/test_paypal.py
@@ -192,10 +192,12 @@ class PaypalTests(PaypalMixin, PaymentProcessorTestCaseMixin, TestCase):
 
     @mock.patch('ecommerce.extensions.payment.processors.paypal.paypalrestsdk.Payment')
     @ddt.data(None, Paypal.DEFAULT_PROFILE_NAME, 'some-other-name')
-    def test_web_profiles(self, enabled_profile_name, mock_payment):
+    def test_dummy_web_profiles(self, enabled_profile_name, mock_payment):
         """
         Verify that the payment creation payload references a web profile when one is enabled with the expected name.
+        This should occur when the create_and_set_webprofile waffle is disabled.
         """
+        toggle_switch('create_and_set_webprofile', False)
         mock_payment_instance = mock.Mock()
         # NOTE: This is necessary to avoid the issue in https://code.djangoproject.com/ticket/25493.
         mock_payment_instance.id = FuzzyInteger(low=1).fuzz()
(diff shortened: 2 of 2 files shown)


ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -4,8 +4,12 @@
 import logging
 from decimal import Decimal
 from urlparse import urljoin
+import random
+import string
 
 import paypalrestsdk
+from paypalrestsdk import WebProfile, Payment
```
schenedx: Please make a decision whether you want to import `paypalrestsdk` or the sub modules individually here.
Basically, either delete the line before this line, or remove this line

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,16 +116,36 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
+        # ES - spanish
+        locale_code = "US" # TODO, set the locale code according to user settings, use "US" as default
+        
+        available_attempts = 2
+        for i in range(1, available_attempts + 1):
+            name = ''.join(random.choice(string.ascii_uppercase) for i in range(12)) # Name needs to be unique so just generating a random one
+            web_profile = WebProfile({
+                "name": name,
+                "presentation": {
+                    "locale_code": locale_code
```
schenedx: How is this value transitioned to `ES` when needed?

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,16 +116,36 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
-            web_profile = PaypalWebProfile.objects.get(name=self.DEFAULT_PROFILE_NAME)
-            data['experience_profile_id'] = web_profile.id
-        except PaypalWebProfile.DoesNotExist:
-            pass
+        # for locale code, use country codes found here: https://developer.paypal.com/docs/classic/api/locale_codes
+        # US - english
+        # ES - spanish
+        locale_code = "US" # TODO, set the locale code according to user settings, use "US" as default
+        
+        available_attempts = 2
```
schenedx: Why do we need to make multiple attempts? Are you trying to address a bug? Is 2 attempts going to be enough?

schenedx (changes requested):
Have both questions and comments. Is this PR ready for review?

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,16 +116,36 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
 
-        try:
-            web_profile = PaypalWebProfile.objects.get(name=self.DEFAULT_PROFILE_NAME)
-            data['experience_profile_id'] = web_profile.id
-        except PaypalWebProfile.DoesNotExist:
-            pass
+        # for locale code, use country codes found here: https://developer.paypal.com/docs/classic/api/locale_codes
+        # US - english
+        # ES - spanish
+        locale_code = "US" # TODO, set the locale code according to user settings, use "US" as default
```
BenjiLee: Let's include the ticket number for this comment https://openedx.atlassian.net/browse/LEARNER-1278

stevenzheng3791:
It is possible to create permanent payment experience profiles beforehand, perhaps one for each language, and grab the correct one when the user presses checkout with paypal. I think that would be better, but it would require unknown changes outside of this one file.

ecommerce/extensions/payment/processors/paypal.py
stevenzheng3791: Right now it can't, until language detection is supported (LEARNER 1278). Transitioning to "ES" will be a combination of language detection and then setting that language.

ecommerce/extensions/payment/processors/paypal.py
stevenzheng3791: Removed the multiple attempts logic, it was unnecessary. 
Right now, we create a temporary paypal web profile that lasts 3 hours each time a user clicks checkout with paypal. To create a profile, we need to generate a unique name. In the (highly unlikely I think) event that the generated name is already taken by an existing temporary profile, then it would fail and default to english. I am exploring a solution to avoid temporary profiles altogether.

ecommerce/extensions/payment/processors/paypal.py
stevenzheng3791: Done, thanks

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,16 +115,29 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
             }],
         }
 
-        try:
-            web_profile = PaypalWebProfile.objects.get(name=self.DEFAULT_PROFILE_NAME)
+        # for locale code, use country codes found here: https://developer.paypal.com/docs/classic/api/locale_codes
+        # US - english
```
BenjiLee: The US and ES comments could be removed. The link is sufficient

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,16 +115,29 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
+        # ES - spanish
+        locale_code = "US" # TODO (LEARNER 1278) - set the locale code according to user settings, use "US" as default
+        name = ''.join(random.choice(string.ascii_uppercase) for i in range(12)) # Name needs to be unique so just generating a random one
```
BenjiLee: nit: If `name` is not being used anywhere else you could move it right into the WebProfile

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,16 +115,29 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
             }],
         }
 
-        try:
-            web_profile = PaypalWebProfile.objects.get(name=self.DEFAULT_PROFILE_NAME)
```
bderusha, line 116: @stevenzheng3791 @BenjiLee I missed this before.  We already have a database table for these profiles.

What if we did something where we look up the profile like `PaypalWebProfile.objects.get(name=locale)` and if it exists then we use it, otherwise we create a permanent one on-the-fly and then store it.  The benefits of not making that extra call each time, but less of the cumbersome configuration management.

There might be a tricky bit to map our locales to paypal's... anyway just a thought.

stevenzheng3791: Maintaining the mapping turned out to be rather complicated after accounting for potential errors, as opposed to creating one on the fly, which is logically self-contained whether the API call succeeds or not. If the extra API call does turn out to be too slow, then we can look at mapping as a faster solution.

ecommerce/extensions/payment/tests/processors/test_paypal.py
```diff
@@ -189,12 +188,103 @@ def test_switch_disabled_lms_url(self):
...
+        toggle_switch('create_and_set_webprofile', True)
+        Paypal._resolve_paypal_locale = mock.Mock(return_value="valid_locale")
+        mock_web_profile_init.side_effect = Exception('MissingConfig Exception')
+
+        self.processor.get_transaction_parameters(self.basket, request=self.request)
+        payment_creation_payload = mock_payment.call_args[0][0]
+        self.assertNotIn('experience_profile_id', payment_creation_payload)
 
     @mock.patch('ecommerce.extensions.payment.processors.paypal.paypalrestsdk.Payment')
-    @ddt.data(None, Paypal.DEFAULT_PROFILE_NAME, 'some-other-name')
```
BenjiLee: I think you forgot to revert the changes made to this test.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,11 +131,33 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
+        if waffle.switch_is_active('create_and_set_webprofile'):
+
+            # TODO (LEARNER 1278) - set locale code based on user language
+            locale_code = self._resolve_paypal_locale(self.DEFAULT_REGION_CODE)
+
+            if locale_code != self.DEFAULT_LOCALE_CODE:
+                # Create a temporary web profile with the correct locale code
+                try:
+                    web_profile = paypalrestsdk.WebProfile({
+                        "name": str(uuid.uuid1()),
```
BenjiLee: The comment about paypal requiring a 12 char unique name would be helpful here.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,11 +131,33 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
+                        "presentation": {
+                            "locale_code": locale_code
+                        },
+                        "temporary": True
```
BenjiLee: A comment that says temporary == 3 hours would be helpful here.

ecommerce/extensions/payment/processors/paypal.py
stevenzheng3791: It actually doesn't, 12 char was just an example. I figured it'd be more straightforward to use the unique identifier library for the unique name.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,11 +118,40 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
+                    web_profile = paypalrestsdk.WebProfile({
+                        "name": str(uuid.uuid1()),  # Generate a unique identifier
+                        "presentation": {
+                            "locale_code": locale_code
+                        },
+                        "temporary": True  # Persists for 3 hours
+                    }, api=self.paypal_api)
+
+                    if web_profile.create():
+                        logger.info("Web Profile[%s] created successfully", (web_profile.id))
```
schenedx: We should log down the locale/language that prompted for this profile creation

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,11 +118,40 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
+                        data['experience_profile_id'] = web_profile.id
+                    else:
+                        logger.warning(
+                            "Error [%s]. Web Profile could not be created. Will continue without one.",
```
schenedx: I think it's better to have the log message as `Web profile creation encountered error [%s]. Will continue without one.`

ecommerce/extensions/payment/tests/processors/test_paypal.py
```diff
@@ -213,6 +213,116 @@ def test_web_profiles(self, enabled_profile_name, mock_payment):
...
+        mock_payment_instance.to_dict.return_value = {}
+        mock_payment_instance.links = [mock.Mock(rel='approval_url', href='dummy')]
+        mock_payment.return_value = mock_payment_instance
+
+        toggle_switch('create_and_set_webprofile', True)
+        Paypal.resolve_paypal_locale = mock.Mock(return_value=Paypal.DEFAULT_PROFILE_NAME)
+
+        self.processor.get_transaction_parameters(self.basket, request=self.request)
+        payment_creation_payload = mock_payment.call_args[0][0]
+        self.assertNotIn('experience_profile_id', payment_creation_payload)
```
schenedx: You should also make sure logging messages are done as part of this test

ecommerce/extensions/payment/tests/processors/test_paypal.py
```diff
@@ -213,6 +213,116 @@ def test_web_profiles(self, enabled_profile_name, mock_payment):
...
+        mock_payment_instance.links = [mock.Mock(rel='approval_url', href='dummy')]
+        mock_payment.return_value = mock_payment_instance
+
+        toggle_switch('create_and_set_webprofile', True)
+        Paypal.resolve_paypal_locale = mock.Mock(return_value="valid_locale")
+        mock_web_profile_init.side_effect = Exception('MissingConfig Exception')
+
+        self.processor.get_transaction_parameters(self.basket, request=self.request)
+        payment_creation_payload = mock_payment.call_args[0][0]
+        self.assertNotIn('experience_profile_id', payment_creation_payload)
```
schenedx, line 326: You should test that exception is raised and an error message is logged

ecommerce/extensions/payment/tests/processors/test_paypal.py
```diff
@@ -213,6 +213,116 @@ def test_web_profiles(self, enabled_profile_name, mock_payment):
...
+        mock_payment_instance.id = FuzzyInteger(low=1).fuzz()
+        mock_payment_instance.to_dict.return_value = {}
+        mock_payment_instance.links = [mock.Mock(rel='approval_url', href='dummy')]
+        mock_payment.return_value = mock_payment_instance
+
+        toggle_switch('create_and_set_webprofile', False)
+
+        self.processor.get_transaction_parameters(self.basket, request=self.request)
+        payment_creation_payload = mock_payment.call_args[0][0]
+        self.assertNotIn('experience_profile_id', payment_creation_payload)
```
schenedx: Maybe you should also assert that the `paypalrestsdk.WebProfile` is not called with the flag toggled false?

schenedx (commented):
This makes more sense. Please see if my comments make sense to you

ecommerce/extensions/payment/tests/processors/test_paypal.py
stevenzheng3791: I've gotten rid of this test since it was testing the original functionality, and the existing test for it was already more thorough. I modified the original test to set the waffle flag to false.

ecommerce/extensions/payment/tests/processors/test_paypal.py
stevenzheng3791: I've updated all logger messages, and now verified them and/or exceptions in all my tests

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -112,11 +148,23 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
-            web_profile = PaypalWebProfile.objects.get(name=self.DEFAULT_PROFILE_NAME)
-            data['experience_profile_id'] = web_profile.id
-        except PaypalWebProfile.DoesNotExist:
-            pass
+        if waffle.switch_is_active('create_and_set_webprofile'):
+            # TODO (LEARNER 1278) - set locale code based on user language
+            locale_code = self.resolve_paypal_locale()
+            if locale_code != self.DEFAULT_LOCALE_CODE:
+                # Create a temporary web profile with the correct locale code
+                web_profile_id = self.create_web_profile(locale_code)
```
schenedx: #Minor I think the function name is better as `create_temporary_web_profile`

schenedx (approved):
Just a minor comment, rest LGTM

BenjiLee (approved):

PR #1322, opened 2017-06-12
Always require account activation for email domain restricted coupons

douglashall:
We must always require account activation before a learner can redeem
an email domain restricted coupon, otherwise anyone could create an
account using an email address with a privileged domain and redeem
the coupon.

ENT-418

### DIFF
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -594,6 +594,22 @@ class CouponRedeemViewTests(CouponMixin, CourseCatalogTestMixin, LmsApiMockMixin
         expected_url = self.get_full_url(path=reverse('basket:summary'))
         self.assert_redemption_page_redirects(expected_url)
 
+    @httpretty.activate
+    def test_active_user_email_domain_restricted_coupon_redemption(self):
+        """
+        Verify that a user must activate their account before being allowed
+        to redeem an email domain-restricted coupon.
+        """
+        self.mock_account_api(self.request, self.user.username, data={'is_active': False})
+        self.mock_access_token_response()
+        email_domain = self.user.email.split('@')[1]
+        self.create_coupon(catalog=self.catalog, code=COUPON_CODE, benefit_value=5, email_domains=email_domain)
+
+        response = self.client.get(self.redeem_url_with_params())
+        self.assertIn('edx/email_confirmation_required.html', [t.name for t in response.templates])
+        self.assertEqual(response.context['course_name'], self.course.name)
+        self.assertEqual(response.context['user_email'], self.user.email)
+
 
 class EnrollmentCodeCsvViewTests(TestCase):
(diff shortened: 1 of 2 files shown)


ecommerce/coupons/tests/test_views.py
```diff
@@ -594,6 +594,22 @@ def test_active_user_requirement_disabled(self):
         expected_url = self.get_full_url(path=reverse('basket:summary'))
         self.assert_redemption_page_redirects(expected_url)
 
+    @httpretty.activate
+    def test_active_user_email_domain_restricted_coupon_redemption(self):
+        """
+        Verify that a user must activate their account before being allowed
+        to redeem an email domain restricted coupon.
+        """
+        self.user.email = 'foo@example.com'
```
clintonb: The user already has an email address. Why change it? If you need the domain...
```python
email_domain = self.user.email.split('@')[1]
```

ecommerce/coupons/tests/test_views.py
```diff
@@ -594,6 +594,22 @@ def test_active_user_requirement_disabled(self):
...
+        self.user.save()
+        self.mock_account_api(self.request, self.user.username, data={'is_active': False})
+        self.mock_access_token_response()
+        self.create_coupon(catalog=self.catalog, code=COUPON_CODE, benefit_value=5, email_domains='example.com')
+
+        response = self.client.get(self.redeem_url_with_params())
+        self.assertEqual(response.context['course_name'], self.course.name)
```
clintonb: You should ensure the correct template is rendered as well.

ecommerce/coupons/tests/test_views.py
```diff
@@ -594,6 +594,22 @@ def test_active_user_requirement_disabled(self):
         expected_url = self.get_full_url(path=reverse('basket:summary'))
         self.assert_redemption_page_redirects(expected_url)
 
+    @httpretty.activate
+    def test_active_user_email_domain_restricted_coupon_redemption(self):
+        """
+        Verify that a user must activate their account before being allowed
+        to redeem an email domain restricted coupon.
```
clintonb: domain-restricted

douglashall:
@clintonb Anything else you think we should clean up here?

ecommerce/coupons/views.py
```diff
@@ -169,10 +170,16 @@ def get(self, request):
         if not valid_voucher:
             return render(request, template_name, {'error': msg})
 
-        if not voucher.offers.first().is_email_valid(request.user.email):
```
clintonb, line 172: Why was this changed? The new variable isn't used anywhere else.

clintonb (approved):

ecommerce/coupons/views.py
douglashall: I'm using `offer` on line 174 and line 182.

PR #1323, opened 2017-06-13
Updated SOAP library

clintonb:
suds has been replaced zeep. This resolves compatibility issues with TLS 1.2 and Python 3.

ECOM-7286


schenedx:
@clintonb This looks like a risky change. 
Does it make sense to slowly roll out this change in production to make sure we are not breaking for some specially configured users?
What is your plan to manually verify this before stage? Should we have a sandbox setup to test this out?

clintonb:
@schenedx why do you think this looks risky? SOAP is solely used to process refunds. If they work locally, they should work on stage. If they work on stage, they should work in production.

schenedx:
@clintonb Thank you for the explanation. I am fine to verify the change in stage.

clintonb:
@edx/learner please review

mjfrey (approved):

rlucioni (approved):
@clintonb just to help me understand, are you making this change to help pave the way for Python 3 here? Looks fine to me once the outstanding quality issues are resolved.

clintonb:
@rlucioni this change is primarily prioritized to ensure we have no issues with CyberSource's TLS 1.2 upgrade. The Python 3 preparedness is a bonus.

PR #1324, opened 2017-06-14
SiteConfiguration updates for receipt page

clintonb:
- Receipt page now enabled by default
- Receitp page can be disabled via management command

LEARNER-1375


clintonb:
@edx/learner please review

mjfrey (approved):

PR #1325, opened 2017-06-14
Fix i18n config to enable auto-translation updates

amangano-edx:
Fixes configuration issues that were preventing us from automatically updating translations for the ecommerce repo.

Summary:
- Removes *.po and *.mo from .gitignore. Their presence in the .gitignore file prevented the jenkins transifex/pull job from correctly detecting that translations had changed, and therefore it never attempted to open pull requests to update them.
- Fixes .tx/config so that the translation files are mapped with the correct names.
- adds `--mode reviewed` to the pull_translations command to ensure that only reviewed translations are merged to master.

@rlucioni

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ compile_translations:
 fake_translations: extract_translations dummy_translations compile_translations
 
 pull_translations:
-	cd ecommerce && tx pull -af
+	cd ecommerce && tx pull -af --mode reviewed
 
 push_translations:
 	cd ecommerce && tx push -s
--- a/ecommerce/.tx/config
+++ b/ecommerce/.tx/config
@@ -2,13 +2,15 @@
 host = https://www.transifex.com
 
 [edx-platform.ecommerce]
-file_filter = conf/locale/<lang>/LC_MESSAGES/django-partial.po
-source_file = conf/locale/en/LC_MESSAGES/django-partial.po
-source_lang = en
+file_filter = conf/locale/<lang>/LC_MESSAGES/django.po
+source_file = conf/locale/en/LC_MESSAGES/django.po
+source_lang = en_US
 type = PO
+lang_map = en_US:en
 
 [edx-platform.ecommerce-js]
-file_filter = conf/locale/<lang>/LC_MESSAGES/djangojs-partial.po
-source_file = conf/locale/en/LC_MESSAGES/djangojs-partial.po
-source_lang = en
+file_filter = conf/locale/<lang>/LC_MESSAGES/djangojs.po
+source_file = conf/locale/en/LC_MESSAGES/djangojs.po
+source_lang = en_US
 type = PO
+lang_map = en_US:en


.gitignore
```diff
@@ -43,8 +43,6 @@ unittests.xml
 e2e/xunit.xml
 
 ### Internationalization artifacts
-*.mo
```
clintonb: This shouldn't be necessary given that we are permitting django(js).(mo|po) below.

ecommerce/.tx/config
```diff
@@ -2,13 +2,15 @@
 host = https://www.transifex.com
 
 [edx-platform.ecommerce]
-file_filter = conf/locale/<lang>/LC_MESSAGES/django-partial.po
-source_file = conf/locale/en/LC_MESSAGES/django-partial.po
-source_lang = en
+file_filter = conf/locale/<lang>/LC_MESSAGES/django.po
+source_file = conf/locale/en/LC_MESSAGES/django.po
+source_lang = en_US
```
clintonb, line 7: Will this change need to be made in other repos? edx/course-discovery, for example, uses en.

ecommerce/.tx/config
```diff
@@ -2,13 +2,15 @@
 host = https://www.transifex.com
 
 [edx-platform.ecommerce]
-file_filter = conf/locale/<lang>/LC_MESSAGES/django-partial.po
-source_file = conf/locale/en/LC_MESSAGES/django-partial.po
-source_lang = en
+file_filter = conf/locale/<lang>/LC_MESSAGES/django.po
+source_file = conf/locale/en/LC_MESSAGES/django.po
```
clintonb, line 6: Based on your change below, should Django be updated to actually write to en_US?

.gitignore
amangano-edx: Maybe you're right, since before it was writing files to django-partial.po instead of django.po. That may be why git wasn't picking them up.

ecommerce/.tx/config
amangano-edx: Probably, since it looks like in most cases we've configured the source file in transifex to be en_US, not en.

ecommerce/.tx/config
amangano-edx: I don't think it matters, since transifex can tell that the /en/ file is the source thanks to the lang_map parameter.

Makefile
```diff
@@ -113,7 +113,7 @@ compile_translations:
 fake_translations: extract_translations dummy_translations compile_translations
 
 pull_translations:
-	cd ecommerce && tx pull -af
+	cd ecommerce && tx pull -af --mode reviewed
```
rlucioni, line 116: Is it really necessary to change directories before pulling? We don't do this in other repos.

rlucioni (approved):

Makefile
amangano-edx: I think it is in this case since the `.tx/` directory is a child of `ecommerce`. I do not know why we placed it there as opposed to at the root of the repository, as we've done in others.

PR #1327, opened 2017-06-14
ENT-418 Check for account activation in VoucherAddView.

douglashall:
We need to also check to see if account activation is required
when a user manually enters a coupon code on the basket page.

ENT-418

### DIFF
--- a/ecommerce/extensions/offer/utils.py
+++ b/ecommerce/extensions/offer/utils.py
@@ -1,6 +1,7 @@
 """Offer Utility Methods. """
 from decimal import Decimal
 
+from django.shortcuts import render
 from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_model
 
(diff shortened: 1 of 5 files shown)


ecommerce/extensions/offer/utils.py
```diff
@@ -71,3 +72,36 @@ def format_benefit_value(benefit):
...
+
+    Arguments:
+        request (HttpRequest): The current HttpRequest.
+        offer (ConditionalOffer): The offer to be redeemed.
+        product (Product): The
+
+    Returns:
+        HttpResponse or None: An HttpResponse which renders the email confirmation template if required.
+    """
+    require_account_activation = request.site.siteconfiguration.require_account_activation or offer.email_domains
```
mattdrayer, line 98: Semantically speaking, it doesn't seem like the determination we're trying to make here is whether or not account activation is required -- it's whether or not email confirmation is required, so I might change the name of the variable.

ecommerce/extensions/offer/utils.py
```diff
@@ -71,3 +72,36 @@ def format_benefit_value(benefit):
...
+    if require_account_activation and not request.user.account_details(request).get('is_active'):
```
mattdrayer, line 99: ...or maybe not...I guess I could take it or leave it...

mattdrayer:
One nit from me (take or leave), otherwise 👍

mattdrayer (approved):

douglashall:
@mattdrayer Yes, I went back and forth between the two names when I originally created the flag on the SiteConfiguration. I'm going to leave this as account activation related for now. Thanks for the review!

PR #1329, opened 2017-06-16
WIP: LEARNER-1278: Add locale to User

BenjiLee:
Gets's locale from LMS in this change.  https://github.com/edx/edx-platform/pull/15366

Tests might be failed because of a migration issue. Still trying to figure how to test on this repo.

TODO:
- [ ] Verify that the LMS change points to the correct ecommerce version.


ecommerce/core/admin.py
```diff
@@ -15,10 +15,10 @@ class SiteConfigurationAdmin(admin.ModelAdmin):
 
 @admin.register(User)
 class EcommerceUserAdmin(UserAdmin):
-    list_display = ('username', 'email', 'full_name', 'first_name', 'last_name', 'is_staff')
+    list_display = ('username', 'email', 'full_name', 'first_name', 'last_name', 'locale', 'is_staff')
```
clintonb: Don't put this on the list display. That page loads enough information already.

ecommerce/core/admin.py
```diff
@@ -15,10 +15,10 @@ class SiteConfigurationAdmin(admin.ModelAdmin):
...
     fieldsets = (
         (None, {'fields': ('username', 'password')}),
-        (_('Personal info'), {'fields': ('full_name', 'first_name', 'last_name', 'email')}),
+        (_('Personal info'), {'fields': ('full_name', 'first_name', 'last_name', 'locale', 'email')}),
```
clintonb: Nit: move this after email.

ecommerce/core/models.py
```diff
@@ -405,6 +405,7 @@ class User(AbstractUser):
     """Custom user model for use with OIDC."""
 
     full_name = models.CharField(_('Full Name'), max_length=255, blank=True, null=True)
+    locale = models.CharField(_('User locale'), max_length=255, blank=True, null=True)
```
clintonb: s/User locale/Locale 
Every field is a User field.

ecommerce/settings/devstack.py
```diff
@@ -54,6 +54,16 @@
 }
 # END PAYMENT PROCESSING
 
+EDX_DRF_EXTENSIONS = {
```
clintonb: Remove. This should be in the base settings.

clintonb (commented):
Add tests.

ecommerce/settings/devstack.py
BenjiLee: These settings do not get set locally unless they are declared in this file.

clintonb: As mentioned in chat, update https://github.com/edx/ecommerce/blob/57bcc22b5337a86e4cff3baa75cf6aeb18bcd331/ecommerce/settings/production.py#L46 so that `EDX_DRF_EXTENSIONS` is merged instead of overwritten by external configuration.

BenjiLee:
Closed in favor of https://github.com/edx/ecommerce/pull/1345

PR #1331, opened 2017-06-21
Removed usage of SiteConfiguration.enable_otto_receipt_page

clintonb:
Otto's receipt page is now the only receipt page we support. This field will be removed in a subsequent commit/release. Additionally, the `CybersourceNotifyView` (which relied on this field) has been removed. CyberSource payers should be redirected to the interstitial.

LEARNER-1528

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -117,6 +117,7 @@ class SiteConfiguration(models.Model):
         blank=True,
         default="",
     )
+    # TODO Remove this field after all usages have been removed/released to production.
     enable_otto_receipt_page = models.BooleanField(
         verbose_name=_('Enable Otto receipt page'),
         help_text=_('Enable the usage of Otto receipt page.'),
--- a/ecommerce/core/management/commands/create_or_update_site.py
+++ b/ecommerce/core/management/commands/create_or_update_site.py
@@ -137,7 +137,6 @@ class Command(BaseCommand):
         segment_key = options.get('segment_key')
         from_email = options.get('from_email')
         enable_enrollment_codes = True if options.get('enable_enrollment_codes') else False
-        enable_otto_receipt_page = not options.get('disable_otto_receipt_page')
         payment_support_email = options.get('payment_support_email', '')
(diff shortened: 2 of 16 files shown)


clintonb:
@edx/learner please review

mjfrey (approved):
yes, remove code.

rlucioni (approved):
Thanks for removing this @clintonb. The story to add the interstitial shouldn't have been closed until this was done.

ecommerce/core/models.py
```diff
@@ -117,6 +117,7 @@ class SiteConfiguration(models.Model):
         blank=True,
         default="",
     )
+    # TODO Remove this field after all usages have been removed/released to production.
```
bderusha, line 120: Is there a ticket tracking this?

bderusha (approved):

ecommerce/core/models.py
clintonb: There is now. See the sub-tasks of https://openedx.atlassian.net/browse/LEARNER-1528.

robrap: FYI: Hopefully there will be an OEP soon on this, but I'd recommend the following format for TODO comments:
```
# TODO: TICKET: DUE DATE: here's my comment
```
For example:
```
# TODO: LEARNER-1528: 07/2017: Remove this field after all usages have been removed/released to production.
```

Note, the point of the date is to is so that if you see a TODO comment that is 6 months or 2 years past due, you might be prompted to review the ticket and make a decision about how to proceed (e.g. do it, or close ticket and remove to dos, or whatever makes the most sense).

clintonb: I appreciate the feedback. I personally don't let these remain in place for longer than a day. I intend to remove the field within hours of this change going to production.

robrap: No worries.  You don't have to change this one, but I still think it's a good practice because it's a shared codebase, and it's useful info for others (no matter how temporary), and sometimes things fall through the cracks.

PR #1332, opened 2017-06-21
Added switch for repeat purchase check

clintonb:
This check to avoid allowing learners purchasing the same item multiple times is a nuisance in development and test environments. If refund deadlines pass, there is no simple way for developers to get into a state where they can view the checkout page. Disabling this check allows for the addition of items regardless of whether they have been previously purchased.


clintonb:
@edx/learner please review

mjfrey (approved):
I still am not a fan of the check in the first place.

schenedx (approved):
👍

schenedx:
@Ayub-Khan Please be aware of this change

Ayub-Khan:
@clintonb  @schenedx @mjfrey I Just had a look at this pull request . This pull request should not have been merged. Current implementation of 'allow_repeat_purchase' is doing the exact opposite job as compared to its name. If this flag is disabled It will work as a kill switch for all the basket views.

clintonb:
The switch is misnamed. It should be called disallow_repeat_purchase. The logic itself is correct. I'll update the name later today.

ecommerce/extensions/basket/utils.py
```diff
@@ -43,24 +45,26 @@ def prepare_basket(request, products, voucher=None):
...
-            )
-    if already_purchased_products and basket.is_empty:
-        raise AlreadyPlacedOrderException
+
+    if waffle.switch_is_active(REPEAT_PURCHASE_SWITCH_NAME):
+        already_purchased_products = []
+        for product in products:
+            if product.is_enrollment_code_product or \
+                    not UserAlreadyPlacedOrder.user_already_placed_order(request.user, product):
+                basket.add_product(product, 1)
```
schenedx, line 54: @clintonb If the switch is not active, according to the change here, where do we add the products to the basket?
Doesn't this mean if the switch is not active, we are not adding ANY product to the basket?

clintonb: Hmm...I don't have time to properly debug this. I will revert the commit.

clintonb: Correction: my initial statement stands. We should not deactivate this switch. I'll fix it later today.

Ayub-Khan: @clintonb no need to revert this. its working fine unless the switch is deactivated. another quick question was this switch in off state yesterday for some time ?.

ecommerce/extensions/basket/utils.py
```diff
@@ -43,24 +45,26 @@ def prepare_basket(request, products, voucher=None):
...
-                mode_for_seat(product),
-                product.course_id
-            )
-    if already_purchased_products and basket.is_empty:
-        raise AlreadyPlacedOrderException
+
+    if waffle.switch_is_active(REPEAT_PURCHASE_SWITCH_NAME):
+        already_purchased_products = []
+        for product in products:
+            if product.is_enrollment_code_product or \
```
Ayub-Khan, line 52: I think adding switch here as 'or' will fix the pr.

clintonb:
I created https://github.com/edx/ecommerce/pull/1339 to fix the identified issues.

PR #1333, opened 2017-06-21
ENT-242 Updated the payment page UI with Enterprise msg.

asadiqbal08:
JIRA - [ENT-242](https://openedx.atlassian.net/browse/ENT-242)

screen shot 2017-06-21 at 5 10 21 pm

@zubair-arbi @saleem-latif your initial thoughts on it ?


zubair-arbi (approved):
Once build is green 👍

ecommerce/coupons/views.py
```diff
@@ -241,6 +242,18 @@ def get(self, request):
             except:  # pylint: disable=bare-except
                 logger.exception('Failed to create a free order for basket [%d]', basket.id)
                 return HttpResponseRedirect(reverse('checkout:error'))
+
+        if enterprise_customer:
+            messages.info(
+                self.request,
+                _('{html_start} A discount has been applied, courtesy of {enterprise_customer_name}. {html_end}').
```
clintonb: There is no need to have the HTML in your translation string.

ecommerce/coupons/views.py
```diff
@@ -14,6 +14,7 @@
 from django.utils.translation import ugettext as _
 from django.views.generic import TemplateView, View
 from oscar.core.loading import get_class, get_model
+from django.contrib import messages
```
clintonb: Sort.

clintonb (commented):
Please post a screenshot.

ecommerce/coupons/views.py
asadiqbal08: so no need for `fa-info-circle` ?

ecommerce/coupons/views.py
```diff
@@ -241,6 +242,19 @@ def get(self, request):
             except:  # pylint: disable=bare-except
                 logger.exception('Failed to create a free order for basket [%d]', basket.id)
                 return HttpResponseRedirect(reverse('checkout:error'))
+
+        if enterprise_customer:
+            message = '{html_start}'.format(html_start='<i class="fa fa-info-circle"> ',) + \
```
clintonb: Why is the icon necessary? Do we display an icon for any other messages? Will this conflict with that? Should we display the same icon for all info messages? If so, that should be a template, not here.

If you do need the icon here, stop using format unnecessarily and stop italicizing the entire string:
```python
message = _('A discount has been applied, courtesy of {enterprise_customer_name}.').format(enterprise_customer_name=enterprise_customer.get('name'))
message = '<i class="fa fa-info-circle"></i> {}'.format(message)
```

brittneyexline (commented):
Code looks good! I will also approve once the build is green.
Is there a sandbox available for this? If it is too burdensome to spin that up yourself, we should try to get this onto the business sandbox for testing.

ecommerce/coupons/views.py
```diff
@@ -241,6 +242,18 @@ def get(self, request):
...
                 logger.exception('Failed to create a free order for basket [%d]', basket.id)
                 return HttpResponseRedirect(reverse('checkout:error'))
+
+        if enterprise_customer:
+            message = _('A discount has been applied, courtesy of {enterprise_customer_name}.').format(
+                enterprise_customer_name=enterprise_customer.get('name')
+            )
+            message = '<i class="fa fa-info-circle"></i> {}'.format(message)
+            messages.info(
+                self.request,
```
clintonb: This can go on one line.

clintonb (changes requested):
You need tests. Also, your commit message has a typo in the ticket number.

asadiqbal08:
added test

ecommerce/coupons/tests/test_views.py
```diff
@@ -537,6 +538,26 @@ def test_enterprise_customer_successful_redemption(self):
...
+    def test_enterprise_customer_successful_redemption_message(self):
+        """ Verify the info message appears on successful redemption. """
+        expected_message = '<i class="fa fa-info-circle"></i> A discount has been applied, courtesy of TestShib.'
+        self.toggle_ecommerce_receipt_page(True)
+        code = self.prepare_enterprise_data(benefit_value=5, consent_enabled=False)
+        self.mock_enterprise_learner_api_for_learner_with_no_enterprise()
+        self.mock_enterprise_learner_post_api()
+
+        expected_url = self.get_full_url(path=reverse('basket:summary'))
+        self.request.user = self.user
```
clintonb: This line does nothing because self.request will never be used in this test.

ecommerce/coupons/tests/test_views.py
```diff
@@ -537,6 +538,26 @@ def test_enterprise_customer_successful_redemption(self):
...
     @httpretty.activate
+    def test_enterprise_customer_successful_redemption_message(self):
+        """ Verify the info message appears on successful redemption. """
+        expected_message = '<i class="fa fa-info-circle"></i> A discount has been applied, courtesy of TestShib.'
+        self.toggle_ecommerce_receipt_page(True)
+        code = self.prepare_enterprise_data(benefit_value=5, consent_enabled=False)
+        self.mock_enterprise_learner_api_for_learner_with_no_enterprise()
+        self.mock_enterprise_learner_post_api()
+
+        expected_url = self.get_full_url(path=reverse('basket:summary'))
```
clintonb: There is no need to use get_full_url because the test client knows the domain. Also, the variable name doesn't correspond to its usage. 

`path = reverse('basket:summary')`

ecommerce/coupons/tests/test_views.py
```diff
@@ -537,6 +538,26 @@ def test_enterprise_customer_successful_redemption(self):
         self.assertEqual(last_request.method, 'POST')
 
     @httpretty.activate
+    def test_enterprise_customer_successful_redemption_message(self):
+        """ Verify the info message appears on successful redemption. """
+        expected_message = '<i class="fa fa-info-circle"></i> A discount has been applied, courtesy of TestShib.'
+        self.toggle_ecommerce_receipt_page(True)
+        code = self.prepare_enterprise_data(benefit_value=5, consent_enabled=False)
```
clintonb: The benefit value doesn't matter, so don't override the default. Overriding a default signals to other developers that your test cares about the value. This test doesn't, so why bother overriding?

ecommerce/coupons/tests/test_views.py
```diff
@@ -537,6 +538,26 @@ def test_enterprise_customer_successful_redemption(self):
...
+
+        expected_url = self.get_full_url(path=reverse('basket:summary'))
+        self.request.user = self.user
+        response = self.client.get(self.redeem_url_with_params(code=code))
+        # pylint: disable=protected-access, no-member
+        response_messages = messages.storage.cookie.CookieStorage(response)._decode(response.cookies['messages'].value)
+        self.assertEqual(response_messages[0].tags, 'safe info')
+        self.assertEqual(response_messages[0].message, expected_message)
+        self.assertRedirects(
+            response, expected_url, status_code=302, target_status_code=200, fetch_redirect_response=False
```
clintonb: target_status_code is not needed given that you aren't fetching a response.

ecommerce/coupons/tests/test_views.py
```diff
@@ -537,6 +538,26 @@ def test_enterprise_customer_successful_redemption(self):
...
+        self.toggle_ecommerce_receipt_page(True)
+        code = self.prepare_enterprise_data(benefit_value=5, consent_enabled=False)
+        self.mock_enterprise_learner_api_for_learner_with_no_enterprise()
+        self.mock_enterprise_learner_post_api()
+
+        expected_url = self.get_full_url(path=reverse('basket:summary'))
+        self.request.user = self.user
+        response = self.client.get(self.redeem_url_with_params(code=code))
+        # pylint: disable=protected-access, no-member
+        response_messages = messages.storage.cookie.CookieStorage(response)._decode(response.cookies['messages'].value)
```
clintonb: There's no need to decode cookies. Pull messages from context. See https://github.com/edx/ecommerce/blob/8fde125db17565c28cdf182fd35416c5c05b9dee/ecommerce/extensions/dashboard/tests.py#L8.

clintonb (changes requested):

ecommerce/coupons/tests/test_views.py
asadiqbal08: @zubair-arbi fyi

ecommerce/coupons/tests/test_views.py
asadiqbal08: I did not change the default value, I am just overriding in my own test and with `benefit_value=100`, I am getting [redirection](https://github.com/edx/ecommerce/blob/master/ecommerce/coupons/views.py#L240)

clintonb: Add a comment with that explanation. Try not leave others guessing about your changes. That's easy to do with comments. 

Setting benefit value to a low amount to ensure the basket is not free, and calls to the checkout page do not redirect away from the checkout page.

asadiqbal08:
@brittneyexline build is green.

ecommerce/coupons/tests/test_views.py
```diff
@@ -537,6 +537,27 @@ def test_enterprise_customer_successful_redemption(self):
...
+        # and calls to the checkout page do not redirect away from the checkout page.
+        code = self.prepare_enterprise_data(benefit_value=5, consent_enabled=False)
+
+        self.toggle_ecommerce_receipt_page(True)
+        self.mock_enterprise_learner_api_for_learner_with_no_enterprise()
+        self.mock_enterprise_learner_post_api()
+
+        response = self.client.get(self.redeem_url_with_params(code=code), follow=True)
+        messages = []
+        for context in response.context:
```
clintonb: Why is this a loop? There is only one context.

ecommerce/coupons/tests/test_views.py
```diff
@@ -537,6 +537,25 @@ def test_enterprise_customer_successful_redemption(self):
...
+        # and calls to the checkout page do not redirect away from the checkout page.
+        code = self.prepare_enterprise_data(benefit_value=5, consent_enabled=False)
+
+        self.toggle_ecommerce_receipt_page(True)
+        self.mock_enterprise_learner_api_for_learner_with_no_enterprise()
+        self.mock_enterprise_learner_post_api()
+
+        response = self.client.get(self.redeem_url_with_params(code=code), follow=True)
+        messages = []
+        messages += response.context[0].get('messages', [])
```
clintonb: You still haven't answered my question: Is there more than one context? 

The answer is no, there is only one context. `response.context` is a `dict`, not a `list`. See https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.Response.context for additional details.

Here is the corrected code: 
```python
messages = response.context['messages']
self.assertEqual(len(messages), 1)
```

clintonb (changes requested):

ecommerce/coupons/tests/test_views.py
asadiqbal08: @clintonb there are more than one context in response that's why I used the loop. However I have eliminated the loop logic. Kindly approve the PR. we are just verifying the message here.

ecommerce/coupons/tests/test_views.py
```diff
@@ -537,6 +537,25 @@ def test_enterprise_customer_successful_redemption(self):
...
+        # Setting benefit value to a low amount to ensure the basket is not free,
+        # and calls to the checkout page do not redirect away from the checkout page.
+        code = self.prepare_enterprise_data(benefit_value=5, consent_enabled=False)
+
+        self.toggle_ecommerce_receipt_page(True)
+        self.mock_enterprise_learner_api_for_learner_with_no_enterprise()
+        self.mock_enterprise_learner_post_api()
+
+        response = self.client.get(self.redeem_url_with_params(code=code), follow=True)
+        messages = []
```
clintonb, line 550: `messages = response.context['messages']`

clintonb (approved):
Please address the last comment.

Note that I am not intentionally trying to be a roadblock. Out jobs as engineers is to not simply solve the problem, but to effectively communicate anything that might trip someone up later (including our future selves).

While the test may have seemed simple, it differed significantly from every test in our codebase. Differences are okay *if* they are properly explained via comments. Otherwise, seemingly simple solutions end up with a lot of questions.

ecommerce/coupons/tests/test_views.py
asadiqbal08: `response.context['messages']` gives me a `FallbackStorage` object but in list it is a `storage.base.Message` object that is what I get from [storage beckends](https://hdknr.github.io/docs/django/django.contrib.messages.storage.html)

asadiqbal08:
@clintonb Thanks for the feedback on the PR. I agree with you but if I look around the other existing tests in same file then sure there are many places that need to be documented and named properly. 
As an example, If you look around the method that is basically returning a `code` but its name is  something confusing --- [prepare_enterprise_data](https://github.com/edx/ecommerce/blob/master/ecommerce/coupons/tests/test_views.py#L454)  and it is used many other places.
As an other example, I could not understand the reason why this [value is coded `100`](https://github.com/edx/ecommerce/blob/master/ecommerce/coupons/tests/test_views.py#L457) here, while I was writing the test. 

So I agree, we should effectively communicate anything that might trip someone up later.

clintonb:
If you see a problem with the code, fix it! No one else will maintain our codebase for us.

douglashall:
@asadiqbal08  I have cherry-picked this commit onto a sprint branch (edx/ecommerce:mighty_mouse) and deployed that branch to https://business.sandbox.edx.org.

brittneyexline:
rebased from https://github.com/edx/ecommerce/pull/1333/commits/1a7be4707ebcc3bd744aab97d5869eaee12d2927

PR #1334, opened 2017-06-21
Logging payments made against non-frozen baskets

clintonb:
We should only accept payments for frozen baskets. Accepting payments for baskets in other states leads to issues that require manual intervention. This change logs an error for these payments, and displays an error message to the user.

LEARNER-867

### DIFF
--- a/ecommerce/extensions/payment/views/cybersource.py
+++ b/ecommerce/extensions/payment/views/cybersource.py
@@ -90,7 +90,7 @@ class CybersourceSubmitView(FormView):
         # Ensure we aren't attempting to purchase a basket that has already been purchased, frozen,
         # or merged with another basket.
         if basket.status != Basket.OPEN:
-            logger.debug('Basket %d must be in the "Open" state. It is currently in the "%s" state.',
+            logger.error('Basket %d must be in the "Open" state. It is currently in the "%s" state.',
                          basket.id, basket.status)
             error_msg = _('Your basket may have been modified or already purchased. Refresh the page to try again.')
             return self._basket_error_response(error_msg)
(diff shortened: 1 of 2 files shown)


clintonb:
FYI @edx/learner 

I want to wait a bit before merging this to dig more into this issue. If this gets merged, we ultimately leave the learner in horrible state in which they can make a payment, but not actually get the product (because we show an error!). Until someone, or the code, updates their basket status, they are screwed!

clintonb:
Ultimately we should figure out how baskets end up paid but in a non-frozen/non-submitted state.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -196,6 +196,13 @@ def validate_notification(self, notification):
             if not basket:
                 logger.error('Received payment for non-existent basket [%s].', basket_id)
                 raise InvalidBasketError
+
+            if basket.status != Basket.FROZEN:
+                logger.error(
+                    'CyberSource payment received for basket [%d] which is in a non-frozen state, [%s]',
+                    basket.id, basket.status
+                )
+                raise InvalidBasketError
```
rlucioni: @clintonb could you merge this if this line was removed? That way we can collect some additional information without changing the system's current behavior.

bderusha:
You basically brought up my main question here, which is what is the recourse for a learner if this error occurs?

ecommerce/extensions/payment/views/cybersource.py
clintonb: Yeah. (okay)

clintonb:
@rlucioni @bderusha I have removed the `raise` call. We are now just logging. Thanks for the feedback! Unless you have additional feedback, I will merge this after the build passes.

PR #1338, opened 2017-06-22
Stop thawing baskets

rlucioni:
Thawing the user's basket when there's a problem handling their CyberSource payment causes that basket to be re-used when making subsequent payments. This thaw can be triggered when CyberSource, seemingly unprompted, sends the service an error response soon after a successful payment.

The response contains error code 104, indicating that payment processing failed because the "merchant reference code" (i.e., order number) matches a reference code sent in the last 15 minutes. After this happens, the user's basket is left open. The next time they try to check out, the basket middleware retrieves this open basket. Although we do re-freeze the basket at this point, we will silently fail to place a new order because an order with the number corresponding to the now-frozen basket already exists. If the user is attempting to pay for something at this point, we'll charge them, but they won't receive the product they paid for and we'll show them a receipt page corresponding to their most recently placed order. Since we're re-freezing the basket, the user will be able to place a new order the next time they try to do so (the middleware will create a new basket for them).

This change prevents us from thawing the user's basket when errors occur while handling CyberSource payments. As a result, a new basket should be used for every attempted order, preventing order number collisions. This change also prevents order number collisions from causing order placement to fail silently.

LEARNER-867

@edx/learner FYI.

### DIFF
--- a/ecommerce/extensions/checkout/mixins.py
+++ b/ecommerce/extensions/checkout/mixins.py
@@ -101,25 +101,22 @@ class EdxOrderPlacementMixin(OrderPlacementMixin):
         and basket submission in a transaction. Should be used only in
         the context of an exception handler.
         """
-        try:
-            return Order.objects.get(number=order_number)
-        except Order.DoesNotExist:
-            with transaction.atomic():
-                order = self.place_order(
-                    order_number=order_number,
-                    user=user,
-                    basket=basket,
-                    shipping_address=shipping_address,
-                    shipping_method=shipping_method,
-                    shipping_charge=shipping_charge,
-                    order_total=order_total,
-                    billing_address=billing_address,
-                    **kwargs
-                )
+        with transaction.atomic():
+            order = self.place_order(
+                order_number=order_number,
+                user=user,
+                basket=basket,
+                shipping_address=shipping_address,
+                shipping_method=shipping_method,
+                shipping_charge=shipping_charge,
+                order_total=order_total,
+                billing_address=billing_address,
+                **kwargs
+            )
 
-                basket.submit()
+            basket.submit()
 
-            return self.handle_successful_order(order, request)
+        return self.handle_successful_order(order, request)
 
     def handle_successful_order(self, order, request=None):  # pylint: disable=arguments-differ
         """Send a signal so that receivers can perform relevant tasks (e.g., fulfill the order)."""
--- a/ecommerce/extensions/payment/views/cybersource.py
+++ b/ecommerce/extensions/payment/views/cybersource.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 import logging
 
 import six
-from django.contrib import messages
 from django.contrib.auth.decorators import login_required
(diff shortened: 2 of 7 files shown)


ecommerce/extensions/checkout/mixins.py
```diff
@@ -101,25 +101,22 @@ def handle_order_placement(self,
...
-                    shipping_charge=shipping_charge,
-                    order_total=order_total,
-                    billing_address=billing_address,
-                    **kwargs
-                )
-
-                basket.submit()
-
-            return self.handle_successful_order(order, request)
+        with transaction.atomic():
```
rlucioni, line 104: This reverts a change that should never have been made in https://github.com/edx/ecommerce/commit/13c0891c3ce14a7e1cff479a95ea02fbfb8a4ebd.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -101,25 +101,22 @@ def handle_order_placement(self,
         and basket submission in a transaction. Should be used only in
         the context of an exception handler.
         """
-        try:
-            return Order.objects.get(number=order_number)
```
rlucioni, line 105: This is what made debugging this so difficult. This circumvents the logging infrastructure we have in place to track order placement failures, disguising order number collisions.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -307,17 +307,6 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
         try:
             notification = request.POST.dict()
             basket = self.validate_notification(notification)
-        except (InvalidBasketError, InvalidSignatureError):
-            return redirect(reverse('payment_error'))
-        except (UserCancelled, TransactionDeclined, PaymentError):
```
rlucioni, line 312: User cancellation is a holdover from when we used the CyberSource hosted payment page. Users are not able to cancel CyberSource payments today.

ecommerce/templates/oscar/checkout/payment_error.html
```diff
@@ -29,7 +29,7 @@
     <p>
       {% with "<a class='nav-link' href='"|add:basket_url|add:"'>"|safe as start_link %}
         {% blocktrans with end_link="</a>"|safe %}
-          To try again, visit the {{ start_link }}basket page{{ end_link }}.
```
rlucioni: Referring to it as the "basket page" only means something to us. Something more conventional like "checkout page" is more likely to be understood by our users.

ecommerce/templates/oscar/checkout/payment_error.html
```diff
@@ -15,7 +15,7 @@
     <h1>{% trans "Payment Failed" %}</h1>
     <p>
       {% blocktrans %}
```
clintonb: Add `trimmed` here and below. See https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#blocktrans-template-tag for details.

rlucioni: @clintonb updated.

PR #1339, opened 2017-06-22
Fixed switch to disable repeat order check

clintonb:
The switch has been moved to the UserAlreadyPlacedOrder class to avoid breaking the ability to add items to baskets.


clintonb:
@edx/learner please review

Ayub-Khan (approved):
Looks Good.

PR #1340, opened 2017-06-23
WIP Move edx.org theme into ecommerce repo

amangano-edx:
Moves the edx.org theme from https://github.com/edx/edx-themes into this repo so that we can translate the footer more easily.

Corresponding edx-themes PR: https://github.com/edx/edx-themes/pull/521

JIRA: https://openedx.atlassian.net/browse/LEARNER-1576 (subtask of https://openedx.atlassian.net/browse/LEARNER-1271)

WIP because this hasn't been fully tested, and I'm not sure if any configuration changes are necessary.


clintonb (changes requested):
We should setup some time to chat about theming. Moving any theme back into this repository is not ideal. I would much rather move toward adopting a model similar to what we do with credentials, where the theme is its own app, and setup those repos to translations.

amangano-edx:
@clintonb Our team is aware that this is not an ideal solution, but we decided to go with this approach for now so that we can meet our deadline by the end of this week. I talked with @douglashall yesterday about a better approach that we can apply when we have more time. We will definitely be doing that as we will need it for credentials, which also uses a separate themes repo. Please see the comments in https://openedx.atlassian.net/browse/LEARNER-1271 for more context on what was discussed and our decision making process.

ecommerce/themes/edx.org/static/sass/base/overrides.scss
```diff
@@ -0,0 +1,55 @@
...
+  .site-details, .external-links {
+    margin-left: 0px;
+  }
+}
+
+@media only screen and (max-width: 768px) {
+  .footer-main .grid-container .external-links {
+    float: left;
+  }
+}
```
douglashall, line 55: Add an extra line.

douglashall (approved):
Do you want to also run makemessages or will you do that in a separate PR?

BenjiLee:
closing in favor of https://github.com/edx/ecommerce/pull/1350

PR #1342, opened 2017-06-23
Removed SiteConfiguration.enable_otto_receipt_page

clintonb:
This field is no longer needed. All usages were removed in a previous commit.

LEARNER-1528


clintonb:
@edx/learner please review

schenedx (approved):
👍

PR #1343, opened 2017-06-23
Link to the dashboard from the payment error page

rlucioni:
Users seeing the payment error page will be given a new, empty basket. Returning them to the basket page with an empty basket isn't helpful and forces them to make an extra click to get back to their dashboard.

LEARNER-867

This a follow-up to https://github.com/edx/ecommerce/pull/1338.

### DIFF
--- a/ecommerce/extensions/payment/tests/views/test_cybersource.py
+++ b/ecommerce/extensions/payment/tests/views/test_cybersource.py
@@ -12,6 +12,7 @@ from oscar.apps.payment.exceptions import TransactionDeclined
 from oscar.core.loading import get_class, get_model
 from oscar.test import factories
 
+from ecommerce.core.url_utils import get_lms_url
 from ecommerce.extensions.payment.exceptions import InvalidBasketError, InvalidSignatureError
 from ecommerce.extensions.payment.tests.mixins import CybersourceMixin, CybersourceNotificationTestsMixin
 from ecommerce.extensions.payment.views.cybersource import CybersourceInterstitialView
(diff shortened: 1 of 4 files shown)


clintonb (commented):
This assumes the learner arrived at the basket from the dashboard. Not all learners arrive from the dashboard. See https://analytics.google.com/analytics/web/?authuser=1#report/content-pages/a35248639w83315850p86300562/%3Fexplorer-table.plotKeys%3D%5B%5D%26explorer-table.rowCount%3D500%26explorer-table.secSegmentId%3Danalytics.referralPath%26explorer-table.advFilter%3D%5B%5B0%2C%22analytics.pageTitle%22%2C%22EQ%22%2C%22Basket%22%2C0%5D%5D%26explorer-segmentExplorer.segmentId%3Danalytics.pageTitle/.

rlucioni:
@clintonb that's fair. Where would you suggest we send them, then? In the short-term, the dashboard seems like an improvement over an empty basket page.

clintonb:
Ideally, we should recreate the basket and put the user back on the checkout page. Is that feasible?

rlucioni:
@clintonb, @mjfrey pointed me to an example of [manual basket creation](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/v2/views/baskets.py#L376-L386). If what's done there is enough to create a valid basket, then we may be able to manually create a fresh basket, merge the frozen one into it, and send the user back to the basket page.

At this point, I'd prefer to merge this and open a ticket to improve on it incrementally. What's presented here is objectively better than what's currently in prod. Anything more involved is beyond the scope of [LEARNER-867](https://openedx.atlassian.net/browse/LEARNER-867).

clintonb:
What's the time frame for getting this done? We should fix this flow ASAP. At this point the user has clearly indicated a desire to give us money. We should do our best to accept payment and fulfill the order. Leaving the user in a lurch is likely to push the user away.

rlucioni:
@clintonb agreed, we're on the same page. I can jump on that after getting this out. Like I said, this is a smaller, faster, and safer fix that improves on what's out there right now. I might be able to get this out before the weekend, but I'm less confident about the basket merging approach.

clintonb (approved):

mjfrey (approved):

PR #1344, opened 2017-06-23
Consolidated Segment tracking code in analytics.utils

clintonb:
The track_segment_event function can now be used fire events. It will handle creating the context and verifying that events can be fired for the site.

LEARNER-1563

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -2,14 +2,7 @@ import json
 import logging
 from functools import wraps
 
-from threadlocals.threadlocals import get_current_request
-
 logger = logging.getLogger(__name__)
-
-
-def is_segment_configured():
-    """Returns a Boolean indicating if Segment has been configured for use."""
-    return bool(get_current_request().site.siteconfiguration.segment_key)
 
 
 def parse_tracking_context(user):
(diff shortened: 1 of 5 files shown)


ecommerce/extensions/analytics/utils.py
```diff
@@ -118,3 +114,33 @@ def prepare_analytics_data(user, segment_key):
...
+
+    Returns:
+        (success, msg): Tuple indicating the success of enqueuing the event on the message queue.
+            This can be safely ignored unless needed for debugging purposes.
+    """
+    site_configuration = site.siteconfiguration
+    if not site_configuration.segment_key:
+        msg = 'Event [{event}] was NOT fired because no Segment key is set for site configuration [{site_id}]'
+        msg = msg.format(event=event, site_id=site_configuration.pk)
+        logger.debug(msg)
```
rlucioni, line 136: This message seems useful. Should it be logged at the INFO or WARNING levels?

ecommerce/extensions/analytics/utils.py
```diff
@@ -118,3 +114,33 @@ def prepare_analytics_data(user, segment_key):
...
+        return False, msg
```
rlucioni, line 137: Why return the message? Logging it here and moving on seems sufficient. The message isn't being used anywhere else.

rlucioni (approved):

ecommerce/extensions/analytics/utils.py
clintonb: We don't want this in Splunk for edx.org. None of the white label sites use Segment, so this will just add spam from those sites.

ecommerce/extensions/analytics/utils.py
clintonb: I am maintaining the return value structure of Segment's `track()` method.

PR #1345, opened 2017-06-23
Setup internationalize on checkout and paypal page

BenjiLee:
https://openedx.atlassian.net/browse/LEARNER-1278
https://openedx.atlassian.net/browse/LEARNER-914

Goals:
* Enable django language cookie
* Verify i18n on checkout page'
* Send user's language to paypal

@edx/learner Please review

https://github.com/edx/configuration/pull/3950
https://github.com/edx/ecommerce/pull/1345/
https://github.com/edx/edx-internal/pull/115

### DIFF
--- a/ecommerce/extensions/payment/constants.py
+++ b/ecommerce/extensions/payment/constants.py
@@ -29,3 +29,12 @@ CYBERSOURCE_CARD_TYPE_MAP = {
 }
 
 CLIENT_SIDE_CHECKOUT_FLAG_NAME = 'enable_client_side_checkout'
+
+# Paypal only supports 4 languages, which are prioritized by country.
+# https://developer.paypal.com/docs/classic/api/locale_codes/
+PAYPAL_LOCALES = {
+    'zh': 'CN',
+    'fr': 'FR',
+    'en': 'US',
+    'es': 'MX',
+}
--- a/ecommerce/extensions/payment/processors/paypal.py
+++ b/ecommerce/extensions/payment/processors/paypal.py
@@ -2,17 +2,21 @@
 from __future__ import unicode_literals
 
 import logging
+import re
 import uuid
 from decimal import Decimal
 from urlparse import urljoin
 
 import paypalrestsdk
 import waffle
+from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.utils.functional import cached_property
+from django.utils.translation import get_language
 from oscar.apps.payment.exceptions import GatewayError
 
 from ecommerce.core.url_utils import get_ecommerce_url
+from ecommerce.extensions.payment.constants import PAYPAL_LOCALES
 from ecommerce.extensions.payment.models import PaypalProcessorConfiguration, PaypalWebProfile
 from ecommerce.extensions.payment.processors import BasePaymentProcessor, HandledProcessorResponse
 from ecommerce.extensions.payment.utils import middle_truncate
@@ -29,7 +33,6 @@ class Paypal(BasePaymentProcessor):
 
     NAME = 'paypal'
     DEFAULT_PROFILE_NAME = 'default'
-    DEFAULT_LOCALE_CODE = 'default'
 
     def __init__(self, site):
         """
@@ -63,9 +66,12 @@ class Paypal(BasePaymentProcessor):
     def error_url(self):
         return get_ecommerce_url(self.configuration['error_path'])
 
-    def resolve_paypal_locale(self):
-        """ TODO (LEARNER 1278): Generate appropriate Paypal locale code based on region """
-        return self.DEFAULT_LOCALE_CODE
+    def resolve_paypal_locale(self, language_code):
+        default_paypal_locale = PAYPAL_LOCALES.get(re.split(r'[_-]', get_language())[0].lower())
+        if not language_code:
+            return default_paypal_locale
+        else:
+            return PAYPAL_LOCALES.get(re.split(r'[_-]', language_code)[0].lower(), default_paypal_locale)
 
     def create_temporary_web_profile(self, locale_code):
         """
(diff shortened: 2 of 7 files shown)


ecommerce/settings/base.py
```diff
@@ -478,7 +486,7 @@
 # Detailed information at: https://docs.djangoproject.com/en/dev/ref/settings/
 SESSION_COOKIE_NAME = 'ecommerce_sessionid'
 CSRF_COOKIE_NAME = 'ecommerce_csrftoken'
-LANGUAGE_COOKIE_NAME = 'ecommerce_language'
+LANGUAGE_COOKIE_NAME = 'openedx-language-preference'
```
BenjiLee: @cpennington I was thinking about setting this cookie to the language cookie that we have set in lms and mktg to resolve user language in ecommerce. Are there any downsides to this approach?

cpennington: The only concern I would have is if the ecommerce app is setting the language to something like `en-us` while the other apps are setting `en_us` or just `en`. I don't know if that's actually the case or not, but that's the main thing I'd worry about.

BenjiLee: It can always be changed https://github.com/edx/ecommerce/blob/master/ecommerce/settings/base.py#L69

If I changed it to 'en' or whatever the LMS was using, would that be sufficient?

BenjiLee: Looks like in LMS they are following the recommendation from http://www.i18nguy.com/unicode/language-identifiers.html 

> Note that depending on the purpose or intended audience of the content, a very precise label may not be needed and a one-level language tag may be adequate. For example, using "en" for this page may be sufficient instead of en-US.

ecommerce/extensions/payment/constants.py
```diff
@@ -29,3 +29,11 @@
 }
 
 CLIENT_SIDE_CHECKOUT_FLAG_NAME = 'enable_client_side_checkout'
+
+# Paypal only supports 5 languages, which are prioritized by country.
+# https://developer.paypal.com/docs/classic/api/locale_codes/
+PAYPAL_LOCALES = {
+    'en': 'US',
+    'es': 'AR',
+    'es-419': 'AR'
```
BenjiLee: TODO change to mexico

ecommerce/settings/production.py
```diff
@@ -82,4 +82,7 @@ def get_env_setting(setting):
         })
 # END PAYMENT PROCESSOR OVERRIDES
 
+# Language cookie
+LANGUAGE_COOKIE_NAME = 'prod-edx-language-preference'
```
BenjiLee: TODO remve this

ecommerce/settings/base.py
BenjiLee: TODO undo this specific line since it get's overriden

ecommerce/extensions/payment/tests/processors/test_paypal.py
```diff
@@ -264,29 +264,6 @@ def test_web_profile_with_invalid_locale(self, mock_web_profile, mock_payment, m
 
     @mock.patch('ecommerce.extensions.payment.processors.paypal.logger')
     @mock.patch('ecommerce.extensions.payment.processors.paypal.paypalrestsdk.Payment')
-    def test_web_profile_with_default_locale(self, mock_payment, mock_logger):
```
BenjiLee, line 267: This condition no longer exists. web profile is always created since Paypal also stores a user's language in a cookie.

ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -149,16 +151,10 @@ def get_transaction_parameters(self, basket, request=None, use_client_side_check
...
-            # TODO (LEARNER 1278) - set locale code based on user language
-            locale_code = self.resolve_paypal_locale()
-            if locale_code != self.DEFAULT_LOCALE_CODE:
-                # Create a temporary web profile with the correct locale code
-                web_profile_id = self.create_temporary_web_profile(locale_code)
-                if web_profile_id is not None:
-                    data['experience_profile_id'] = web_profile_id
-            else:
-                # Proceed without web profile (Paypal defaults to english)
-                logger.info("Using default locale. Will not create and set web profile")
```
bderusha, line 161: Should we capture a log message like this somewhere else? 

Might be helpful to have a message indicating the selected language by writing out own `.get` method logic

```
try
    PAYPAL_LOCALES[key]
    log!
except KeyError?
    default_locale
    log!
```

bderusha (approved):

bderusha:
@BenjiLee can you link to the configuration PR when you have one so we can keep track of the linked PRs

ecommerce/extensions/payment/processors/paypal.py
BenjiLee: This error no longer exists since we are always fetching a webprofile. Before we decided to not fetch a webprofile if the user's language == default to save an API call but it turns out paypal has their own language cookie and we need to send it everytime. Otherwise if a user requests the paypal page in ES, and then at a later time requests it in EN, since paypal stored ES in a cookie, the ES page would should up if we do not send a webprofile.

BenjiLee:
@bderusha Please review.

Configurations are WIP.

ecommerce/settings/base.py
```diff
@@ -78,6 +81,12 @@
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#use-i18n
 USE_I18N = True
 
+LANGUAGES = (
```
BenjiLee: @bderusha Does this also have to go in the config?

ecommerce/extensions/payment/tests/processors/test_paypal.py
```diff
@@ -312,6 +290,32 @@ def test_web_profile_with_exception(self, mock_web_profile_init, mock_payment, m
...
         mock_logger.warning.assert_any_call(msg)
 
+    @ddt.unpack
+    @ddt.data(
+        ['zh', 'en', 'US'],
+        ['zh', 'es', 'MX'],
+        ['zh', 'es-419', 'MX'],
+        ['zh', 'invalid', 'CN'],
+        ['zh-zh', '', 'CN'],
+        ['invalid default', 'invalid cookie', None]
```
bderusha, line 300: What happens when we call `create_temporary_web_profile` with `None` ?  Just want to confirm that that is a valid thing for PayPal

ecommerce/settings/base.py
bderusha, line 84: I don't think so.  This is a sane default that we do not need to override int he config.  I think just here is fine

bderusha (approved):

ecommerce/extensions/payment/tests/processors/test_paypal.py
BenjiLee: Paypal accepts the None value and does nothing. Hopefully we never send none, but at least we know what would cause None to be sent.

PR #1346, opened 2017-06-23
Sending Cart Viewed event to Segment when user views basket page

clintonb:
LEARNER-1575


clintonb:
Built atop #1344 
@edx/learner please review

mjfrey (approved):

schenedx (approved):
👍

PR #1349, opened 2017-06-26
Redirect to checkout when errors occur during payment

rlucioni:
It may be easy to correct some payment errors. For example, the user may have entered an incorrect CVV. When these errors occur, we redirect the learner back to the basket summary page to reduce friction in the checkout process.

LEARNER-867

Follow-up to https://github.com/edx/ecommerce/pull/1338.

### DIFF
--- a/ecommerce/extensions/payment/tests/views/test_cybersource.py
+++ b/ecommerce/extensions/payment/tests/views/test_cybersource.py
@@ -165,7 +165,37 @@ class CybersourceInterstitialViewTests(CybersourceNotificationTestsMixin, TestCa
     path = reverse('cybersource:redirect')
     view = CybersourceInterstitialView
 
-    @ddt.data(InvalidSignatureError, InvalidBasketError, TransactionDeclined, Exception)
+    def test_payment_declined(self):
+        """
+        Verify that the user is redirected to the basket summary page when their
+        payment is declined.
+        """
+        # Basket merging clears lines on the old basket. We need to take a snapshot
+        # of lines currently on this basket before it gets merged with a new basket.
+        old_lines = list(self.basket.lines.all())
+
+        notification = self.generate_notification(
+            self.basket,
+            billing_address=self.billing_address,
+        )
+
+        with mock.patch.object(self.view, 'validate_notification', side_effect=TransactionDeclined):
+            response = self.client.post(self.path, notification)
+
+            self.assertRedirects(
+                response,
+                self.get_full_url(path=reverse('basket:summary')),
+                status_code=302,
+                fetch_redirect_response=False
+            )
+
+            new_basket = Basket.objects.get(status='Open')
+            merged_basket_count = Basket.objects.filter(status='Merged').count()
+
+            self.assertEqual(list(new_basket.lines.all()), old_lines)
+            self.assertEqual(merged_basket_count, 1)
+
+    @ddt.data(InvalidSignatureError, InvalidBasketError, Exception)
     def test_invalid_payment_error(self, error_class):
         """
         Verify that the view redirects to the payment error page when an error
--- a/ecommerce/extensions/payment/views/cybersource.py
+++ b/ecommerce/extensions/payment/views/cybersource.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import logging
 
 import six
+from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.urlresolvers import reverse
@@ -10,6 +11,7 @@ from django.db import transaction
 from django.http import JsonResponse
 from django.shortcuts import redirect
 from django.utils.decorators import method_decorator
+from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext as _
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import FormView, View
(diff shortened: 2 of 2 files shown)


ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -293,6 +296,32 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
             # for an existing order. If this happens, we can redirect the browser
             # to the receipt page for the existing order.
             return self.redirect_to_receipt_page(notification)
+        except TransactionDeclined:
```
rlucioni, line 298: @clintonb @mjfrey my thinking here is that this error tends to indicate an error that is easy to correct (e.g., an incorrect CVV). Declined transactions are the most common cause of errors during payment processing. The recovery path is not as clear for other exceptions, so I'm continuing to let those drop through to the payment error page.

rlucioni: I'll add a comment to this effect. Let me know if you disagree.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -293,6 +296,32 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            order_number = request.POST.get('req_reference_number')
+            old_basket_id = OrderNumberGenerator().basket_id(order_number)
+            old_basket = Basket.objects.get(id=basket_id)
+
+            new_basket = Basket(owner=request.user, site=request.site)
+            new_basket.merge(old_basket, add_quantities=False)
+            new_basket.strategy = old_basket.strategy
+
```
mjfrey: Do you need to save() the basket here?

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -165,7 +165,34 @@ class CybersourceInterstitialViewTests(CybersourceNotificationTestsMixin, TestCa
     path = reverse('cybersource:redirect')
     view = CybersourceInterstitialView
 
-    @ddt.data(InvalidSignatureError, InvalidBasketError, TransactionDeclined, Exception)
+    @ddt.data(TransactionDeclined)
```
schenedx: Do we need `ddt` if you are covering only one exception case?
Also, shouldn't we change the name of the test function?

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -293,6 +296,32 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
             # for an existing order. If this happens, we can redirect the browser
             # to the receipt page for the existing order.
             return self.redirect_to_receipt_page(notification)
+        except TransactionDeclined:
+            order_number = request.POST.get('req_reference_number')
+            old_basket_id = OrderNumberGenerator().basket_id(order_number)
+            old_basket = Basket.objects.get(id=basket_id)
+
+            new_basket = Basket(owner=request.user, site=request.site)
+            new_basket.merge(old_basket, add_quantities=False)
```
schenedx: Can you add some comments here on why we want to merge into new baskets instead of opening the old basket?

schenedx (commented):
Few questions inline

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -293,6 +296,36 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            new_basket.strategy = old_basket.strategy
+
+            # Voucher usage is not recorded until an order is placed, so it's
+            # safe to reapply any vouchers here. For more, see
+            # https://github.com/django-oscar/django-oscar/blob/releases/1.4/src/oscar/apps/order/utils.py#L90.
+            Applicator().apply(new_basket, user=new_basket.owner, request=request)
+
```
mjfrey: What happens to the old_basket?  We should delete it.

ecommerce/extensions/payment/tests/views/test_cybersource.py
clintonb: This looks like Git/Github messing up the diff. See the last added line in this batch of lines.

ecommerce/extensions/payment/views/cybersource.py
clintonb: `merge()` calls `save()`.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -293,6 +296,36 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            message = _(
+                'An error occurred while processing your payment. You have not been charged. '
+                'Please double-check the information you provided and try again. '
+                'For help, contact {link_start}support{link_end}.'
```
clintonb: Move "contact" inside the link (e.g. <a>contact support</a>) to provide a slightly larger click target.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -293,6 +296,36 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            # Declined transactions are the most common cause of errors during payment
+            # processing and tend to be easy to correct (e.g., an incorrect CVV may have
+            # been provided). The recovery path is not as clear for other exceptions,
+            # so we let those drop through to the payment error page.
+            order_number = request.POST.get('req_reference_number')
+            old_basket_id = OrderNumberGenerator().basket_id(order_number)
+            old_basket = Basket.objects.get(id=old_basket_id)
+
+            new_basket = Basket(owner=request.user, site=request.site)
+            new_basket.merge(old_basket, add_quantities=False)
```
clintonb: Per @schenedx's request, explain why we are merging instead of reusing the old basket.

ecommerce/extensions/payment/views/cybersource.py
rlucioni: @mjfrey I agree. However, I think that doing that now is starting down a path I don't have the time to be on right now. Deleting the old basket breaks other tests that expect to be able to assert on it. I'd like to leave this change for when we modify the view to delete all baskets after order placement.

rlucioni:
@clintonb @mjfrey @schenedx I've addressed your comments.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -293,6 +295,33 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
             # to the receipt page for the existing order.
             return self.redirect_to_receipt_page(notification)
+        except TransactionDeclined:
+            # Declined transactions are the most common cause of errors during payment
+            # processing and tend to be easy to correct (e.g., an incorrect CVV may have
+            # been provided). The recovery path is not as clear for other exceptions,
+            # so we let those drop through to the payment error page.
+            order_number = request.POST.get('req_reference_number')
+            old_basket_id = OrderNumberGenerator().basket_id(order_number)
+            old_basket = Basket.objects.get(id=old_basket_id)
```
schenedx, line 305: @rlucioni The case we went through together today, reminded me that, the old basket could already be in "submitted" status at this point.
Should we filter out those cases?

schenedx (commented):
I think we should be careful not to merge already "submitted" baskets with brand new ones

ecommerce/extensions/payment/views/cybersource.py
rlucioni: @schenedx submitted baskets will never make it to this point. Baskets selected by the middleware must be open. The basket we looked at earlier today was open. It had been submitted, thawed, and later flushed before being re-used.

mjfrey (approved):

PR #1350, opened 2017-06-26
Do not include the footer on the checkout page

BenjiLee:
https://openedx.atlassian.net/browse/LEARNER-1271

**How to test:**
The footer on devstack is included from https://github.com/edx/ecommerce/blob/master/ecommerce/templates/edx/footer.html

Type some thing in there and it should show up in the footer. With this PR, nothing should show up on the checkout page.

Before: 
screen shot 2017-06-26 at 2 43 04 pm
After: 
screen shot 2017-06-26 at 2 42 25 pm

### DIFF
--- a/ecommerce/templates/oscar/basket/basket.html
+++ b/ecommerce/templates/oscar/basket/basket.html
@@ -68,3 +68,7 @@
     {% include client_side_payment_processor.get_template_name %}
   {% endif %}
 {% endblock %}
+
+{% block footer %}
+<!--Do not include the footer.-->
+{% endblock footer %}


clintonb:
Screenshot? For our open source users, can you put this behind a setting or make it part of `SiteConfiguration`?

jasperho:
@clintonb - can you elaborate a little more on wanting for this to be configurable? I'm having a hard time coming up scenarios where our open source users would want a footer in the checkout.  

I think that having the footer is a distraction to the goal of this page: to complete checkout. That, and it's also difficult to navigate back to the checkout page once you've left it.

clintonb:
I don't know what customizations, if any, others have made to the footer. Given that uncertainty, we should at least inform the community that the footer is going away. The happy-medium is to allow them to keep it via setting.

bderusha (commented):
I have questions about how this impacts whitelabel sites.

BenjiLee:
> I have questions about how this impacts whitelabel sites.

We discussed offline and one thought was to remove the footer entirely in the edX theme. @jasperho The footer is used on the receipts page. Is this a place where we can remove the footer as well?

BenjiLee:
> We discussed offline and one thought was to remove the footer entirely in the edX theme. @jasperho The footer is used on the receipts page. Is this a place where we can remove the footer as well?

Had an offline conversation with @jasperho . We are getting rid of the footer in the edX theme for ecommerce.

PR #1352, opened 2017-06-28
Check for embargo course access.

mjfrey:
LEARNER-1525

Check for course access via the LMS embargo API.


mjfrey:
@edx/helio Please review.

ecommerce/core/models.py
```diff
@@ -370,6 +376,11 @@ def course_catalog_api_client(self):
         # TODO Use URL from SiteConfiguration model.
         return EdxRestApiClient(settings.COURSE_CATALOG_API_URL, jwt=self.access_token)
 
+    @property
```
clintonb: `@cached_property`

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -36,6 +43,16 @@ def setUp(self):
...
 
+    def mock_embargo_api(self, body=None, status=200):
+        httpretty.register_uri(
+            httpretty.GET,
+            re.compile(self.site_configuration.build_lms_url('/api/embargo/v1')),
+            status=status,
+            body=body,
+            content_type='application/json'
+        )
+        self.site_configuration.enable_embargo_check = True
```
clintonb: You can avoid future confusion by putting this line in your tests rather than the mock method.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -36,6 +43,16 @@ def setUp(self):
         self.site_configuration.utm_cookie_name = 'test.edx.utm'
         toggle_switch(DISABLE_REPEAT_ORDER_CHECK_SWITCH_NAME, False)
 
+    def mock_embargo_api(self, body=None, status=200):
+        httpretty.register_uri(
+            httpretty.GET,
+            re.compile(self.site_configuration.build_lms_url('/api/embargo/v1')),
+            status=status,
+            body=body,
```
clintonb: Nit: Do your JSON conversion here.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -118,6 +135,26 @@ def test_prepare_basket_calls_attribution_method(self):
...
+        self.mock_access_token_response()
+        self.mock_embargo_api(body=json.dumps({'access': False}))
+        course = CourseFactory()
+        product = course.create_or_update_seat('verified', False, 10, self.partner)
+        basket = prepare_basket(self.request, [product])
+        self.assertEqual(basket.lines.count(), 0)
+
+    @httpretty.activate
+    def test_prepare_basket_embargo_check_exception(self):
+        """ Verify embargo check passes when API call thorws an exception. """
```
clintonb: throws

ecommerce/core/models.py
iivic: I noticed that we use @cached_property when jwt is set.

ecommerce/extensions/basket/tests/test_utils.py
iivic: Nit: thorws

ecommerce/extensions/payment/utils.py
```diff
@@ -66,6 +67,39 @@ def clean_field_value(value):
     return re.sub(r'[\^:"\']', '', value)
 
 
+def embargo_check(request, products):
+    """ Checks if the user has access to purchase products by calling the LMS embargo API.
+
+    Args:
+        request (object): The current request
+        courses (list): A list of courses to check access against
```
iivic: Nit: your parameter is called `products` while here you state `courses`

ecommerce/extensions/payment/utils.py
```diff
@@ -66,6 +67,39 @@ def clean_field_value(value):
...
+        'ip_address': ip,
+        'course_ids': courses
+    }
+
+    try:
+        response = request.site.siteconfiguration.embargo_api_client.course_access.get(**params)
+        if not response.get('access'):
+            return False
+    except:  # pylint: disable=bare-except
+        # We are going to allow purchase if the API is un-reachable.
```
iivic: We usually catch three types of exceptions ConnectionError, SlumberBaseException and Timeout. Do you want to allow any kind of error or just those for connection issues?

ecommerce/extensions/payment/utils.py
```diff
@@ -66,6 +67,39 @@ def clean_field_value(value):
...
+
+    params = {
+        'user': request.user,
+        'ip_address': ip,
+        'course_ids': courses
+    }
+
+    try:
+        response = request.site.siteconfiguration.embargo_api_client.course_access.get(**params)
+        if not response.get('access'):
```
iivic: Why not just return response.get('access')?

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -36,6 +43,16 @@ def setUp(self):
         self.site_configuration.utm_cookie_name = 'test.edx.utm'
         toggle_switch(DISABLE_REPEAT_ORDER_CHECK_SWITCH_NAME, False)
 
+    def mock_embargo_api(self, body=None, status=200):
+        httpretty.register_uri(
+            httpretty.GET,
+            re.compile(self.site_configuration.build_lms_url('/api/embargo/v1')),
```
vkaracic: Why are you using regex here for the URL?

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -118,6 +135,26 @@ def test_prepare_basket_calls_attribution_method(self):
...
 
+    @httpretty.activate
+    def test_prepare_basket_embargo_check_fail(self):
+        """ Verify an empty basket is returned and embargo_check method called. """
+        self.mock_access_token_response()
+        self.mock_embargo_api(body=json.dumps({'access': False}))
+        course = CourseFactory()
+        product = course.create_or_update_seat('verified', False, 10, self.partner)
+        basket = prepare_basket(self.request, [product])
+        self.assertEqual(basket.lines.count(), 0)
```
vkaracic, line 145: The docstring states that you verify if the `embargo_check` is called yet I don't see that assertion here.

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -152,3 +153,37 @@ def test_deactivate_user(self):
...
+            'ip_address': '0.0.0.0',
+            'course_ids': ['foo-course']
+        }
+
+    def mock_embargo_response(self, response, status_code=200):
+        """ Mock the embargo check API endpoint response. """
+
+        httpretty.register_uri(
+            httpretty.GET,
+            re.compile(self.site_configuration.build_lms_url('/api/embargo/v1')),
```
vkaracic: Same question for regex here.

ecommerce/extensions/payment/utils.py
mjfrey: EdxRestAPI just returns a response

ecommerce/extensions/payment/utils.py
mjfrey: I really want to catch all exceptions.

ecommerce/extensions/payment/utils.py
```diff
@@ -66,6 +67,39 @@ def clean_field_value(value):
...
+        courses (list): A list of courses to check access against
+
+    Returns:
+        True or False
+    """
+    courses = []
+    _, _, ip = parse_tracking_context(request.user)
+
+    for product in products:
+        courses.append(product.course.id)
```
vkaracic: What happens if you try and purchase enrollment codes?

ecommerce/extensions/payment/tests/test_utils.py
```diff
@@ -152,3 +153,37 @@ def test_deactivate_user(self):
...
+        )
+
+    @httpretty.activate
+    def test_embargo_check_match(self):
+        """ Verify the embargo check returns False. """
+        embargo_response = {'access': False}
+        self.mock_access_token_response()
+        self.mock_embargo_response(json.dumps(embargo_response))
+        response = self.site.siteconfiguration.embargo_api_client.course_access.get(**self.params)
+        self.assertEqual(response, embargo_response)
```
vkaracic, line 188: I think you should be testing how `embargo_check()` works here. I think you're just testing here if you'll get the response from going directly to the endpoint that you previously mocked.

vkaracic (changes requested):

ecommerce/extensions/payment/tests/test_utils.py
mjfrey: This is the way to match on the API with kwargs

ecommerce/extensions/basket/tests/test_utils.py
mjfrey: The only way to match the API with kwargs

ecommerce/extensions/payment/utils.py
mjfrey: Good catch, we only are checking against seats.  I will add the check.

ecommerce/extensions/basket/tests/test_utils.py
mjfrey: comment fixed.

ecommerce/extensions/payment/tests/test_utils.py
mjfrey: I am testing this also in prepare_basket.

ecommerce/extensions/payment/tests/test_utils.py
vkaracic: Can you give me an example of what URL is not working for you? Because a URL with any kwargs/string-queries should work just fine.
```
In [1]: import requests

In [2]: import httpretty

In [3]: httpretty.enable()

In [4]: httpretty.register_uri(httpretty.GET, 'http://test-me.com', body="HIT!!")

In [5]: print requests.get('http://test-me.com')
<Response [200]>

In [6]: print requests.get('http://test-me.com').content
HIT!!

In [7]: print requests.get('http://test-me.com/').content
HIT!!

In [8]: print requests.get('http://test-me.com/?a=b').content
HIT!!
```

mjfrey: I am not sure what your problem is with the regex?  Why does it bother you?

ecommerce/extensions/payment/tests/test_utils.py
vkaracic: I was referring that this test doesn't make much sense. You're basically testing if mocking works here.

mjfrey: you are right, it is a bit extraneous, but the more tests the better :)

ecommerce/extensions/payment/tests/test_utils.py
vkaracic: I don't have problems with regex. I think it's not necessary to use it, since we never needed it before to register a URL.

ecommerce/extensions/basket/tests/test_utils.py
schenedx: @mjfrey can you add a comment here for using the regex?

ecommerce/extensions/basket/utils.py
```diff
@@ -45,6 +47,18 @@ def prepare_basket(request, products, voucher=None):
     basket_addition = get_class('basket.signals', 'basket_addition')
     already_purchased_products = []
 
+    if request.site.siteconfiguration.enable_embargo_check:
+        if not embargo_check(request, products):
+            messages.error(
+                request,
+                _('Due to export controls, we cannot allow you to access this course at this time.')
```
schenedx: @mjfrey Is the word `access` what you wrote? Or it's part of the message agreed with copy writers and legal?

mjfrey: @schenedx This is virtually  the copy we have in the LMS today.

schenedx (approved):
👍

mjfrey:
@clintonb can you take a look

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -118,6 +136,38 @@ def test_prepare_basket_calls_attribution_method(self):
...
+
+    @httpretty.activate
+    def test_prepare_basket_embargo_with_enrollment_code(self):
+        """ Verify an basket is returned after adding enrollment code. """
```
clintonb: a basket

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -118,6 +136,38 @@ def test_prepare_basket_calls_attribution_method(self):
...
+        self.mock_access_token_response()
+        self.mock_embargo_api(body=json.dumps({'access': True}))
+        toggle_switch(ENROLLMENT_CODE_SWITCH, True)
+        course = CourseFactory()
+        course.create_or_update_seat('verified', False, 10, self.partner, create_enrollment_code=True)
+        product = Product.objects.get(product_class__name=ENROLLMENT_CODE_PRODUCT_CLASS_NAME)
+        basket = prepare_basket(self.request, [product])
+        self.assertEqual(basket.lines.count(), 1)
```
clintonb, line 158: Nit: You should ensure the correct product is in the basket.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -118,6 +136,38 @@ def test_prepare_basket_calls_attribution_method(self):
...
+
+    @httpretty.activate
+    def test_prepare_basket_embargo_check_exception(self):
+        """ Verify embargo check passes when API call throws an exception. """
+        self.mock_access_token_response()
+        self.mock_embargo_api(body=timeoutException, status=400)
```
clintonb: Nit: replace `timeoutException` with `lambda: raise requests.Timeout`.
Also, there is no need for the status kwarg since it won't be used.

ecommerce/extensions/payment/utils.py
```diff
@@ -66,6 +68,42 @@ def clean_field_value(value):
     return re.sub(r'[\^:"\']', '', value)
 
 
+def embargo_check(request, products):
```
clintonb: Avoid passing request objects as this constricts this method to usage solely within the request cycle. Instead pass the user and site separately since they are the only parts of the request you need.

ecommerce/extensions/payment/utils.py
```diff
@@ -66,6 +68,42 @@ def clean_field_value(value):
...
+    if courses:
+        params = {
+            'user': request.user,
+            'ip_address': ip,
+            'course_ids': courses
+        }
+
+        try:
+            response = request.site.siteconfiguration.embargo_api_client.course_access.get(**params)
+            if not response.get('access'):
```
clintonb: `return response.get('access', True)`

ecommerce/extensions/payment/utils.py
```diff
@@ -66,6 +68,42 @@ def clean_field_value(value):
...
 
+def embargo_check(request, products):
+    """ Checks if the user has access to purchase products by calling the LMS embargo API.
+
+    Args:
+        request (object): The current request
+        products (list): A list of products to check access against
+
+    Returns:
+        True or False
```
clintonb: bool

clintonb (approved):

ecommerce/extensions/basket/tests/test_utils.py
mjfrey: That syntax does not seem to work.

vkaracic: Try with `lambda: requests.Timeout`.

vkaracic (approved):

PR #1354, opened 2017-06-28
Removed unnecessary message on basket view

clintonb:
This message is confusing to buyers, and makes no sense given that you can never view the page if you have previously-purchased items in your basket.

LEARNER-1650


schenedx (approved):
👍

schenedx:
@clintonb This definitely addresses the bug.
Can you help me understand why you say learner cannot view basket page if they have an already purchased item in their basket?

clintonb:
@schenedx if the item has already been purchased, the learner will be show the error page. See https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/views.py#L117-L119

schenedx:
@clintonb I see. I just never seen the error message. Make sense.

PR #1355, opened 2017-06-28
Add logging to CyberSource submit view

rlucioni:
These log messages are meant to help verify that this view is being hit consistently by users' browsers.

LEARNER-1671

### DIFF
--- a/ecommerce/extensions/payment/views/cybersource.py
+++ b/ecommerce/extensions/payment/views/cybersource.py
@@ -59,6 +59,12 @@ class CybersourceSubmitView(FormView):
 
     @method_decorator(login_required)
     def dispatch(self, request, *args, **kwargs):
+        logger.info(
+            'CyberSource submit view called for basket [%d]. It is in the [%s] state.',
+            request.basket.id,
+            request.basket.status
+        )
+
         return super(CybersourceSubmitView, self).dispatch(request, *args, **kwargs)
 
     def get_form_kwargs(self):
@@ -73,6 +79,11 @@ class CybersourceSubmitView(FormView):
         return JsonResponse(data, status=400)
 
     def form_invalid(self, form):
+        logger.info(
+            'Invalid payment form submitted for basket [%d].',
+            self.request.basket.id
+        )
+
         errors = {field: error[0] for field, error in form.errors.iteritems()}
         logger.debug(errors)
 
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -83,6 +94,11 @@ def form_invalid(self, form):
         return JsonResponse({'field_errors': errors}, status=400)
 
     def form_valid(self, form):
+        logger.info(
+            'Valid payment form submitted for basket [%d].',
+            basket.id
```
mjfrey: Maybe  move this down to after generating the params and log the transaction ID?

rlucioni: @mjfrey moving, I also missed that the basket variable isn't yet available here.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -73,6 +79,11 @@ def _basket_error_response(self, error_msg):
         return JsonResponse(data, status=400)
 
     def form_invalid(self, form):
+        logger.info(
+            'Invalid payment form submitted for basket [%d].',
+            basket.id
```
schenedx: Where does this `basket` variable come from?

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -59,6 +59,12 @@ class CybersourceSubmitView(FormView):
 
     @method_decorator(login_required)
     def dispatch(self, request, *args, **kwargs):
+        logger.info(
+            'CyberSource submit view called for basket [%d]. It is in the [%s] state.',
+            basket.id,
```
schenedx: Where does the variable `basket` come from?

schenedx (commented):
I got a few questions

ecommerce/extensions/payment/views/cybersource.py
rlucioni: @schenedx it doesn't, see above. This is what happens when you try to do too many things at once 😟 . Fixing.

rlucioni:
@mjfrey @schenedx I've addressed your comments.

mjfrey (approved):

schenedx (approved):
👍

clintonb (approved):

rlucioni:
I added one additional message here to the checkout view, just to help rule it out.

rlucioni:
Looks like Codecov is having problems: https://status.codecov.io/.

PR #1356, opened 2017-06-28
Add Enterprise entitlement application to BasketMultipleItemsView.

douglashall:
We need to apply available Enterprise entitlements to multi-item baskets. Since we have now switched LMS over to using the new `BasketMultipleItemsView` for creating new baskets, we need to make sure that this view is capable of applying Enterprise entitlements just as the `BasketSingleItemView` was capable of doing.

Sandbox link: https://business.sandbox.edx.org/enterprise/2b5a2956-7746-4fc9-9587-bc887ba45973/course/course-v1:edX+DemoX+Demo_Course/enroll/


brittneyexline (approved):
I haven't tested on the sandbox yet but this looks good.

PR #1357, opened 2017-06-28
Add logging to CyberSource signature generation

rlucioni:
This is meant to help us track where CyberSource data is being signed. The signature itself is not logged.

LEARNER-1671

This could help paint a more complete picture of the behavior we're seeing on the off-chance that Nginx access logs don't tell the whole story.

### DIFF
--- a/ecommerce/extensions/payment/processors/cybersource.py
+++ b/ecommerce/extensions/payment/processors/cybersource.py
@@ -9,6 +9,7 @@ from decimal import Decimal
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from oscar.apps.payment.exceptions import GatewayError, TransactionDeclined, UserCancelled
+from oscar.core.loading import get_class
 from zeep import Client
 from zeep.helpers import serialize_object
 from zeep.wsse import UsernameToken
@@ -26,6 +27,8 @@ from ecommerce.extensions.payment.processors import BaseClientSidePaymentProcess
 from ecommerce.extensions.payment.utils import clean_field_value
 
 logger = logging.getLogger(__name__)
+
+OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
 
 
 class Cybersource(BaseClientSidePaymentProcessor):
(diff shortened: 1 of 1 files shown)


clintonb (approved):

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -273,6 +276,23 @@ def _generate_signature(self, parameters, use_sop_profile):
         Returns:
             unicode: the signature for the given parameters
         """
+        order_number = None
+        basket_id = None
+
+        if 'reference_number' in parameters:
```
rlucioni, line 282: @clintonb this caught me off guard. The parameters we generate, sign, and send to the browser don't contain the `req_` prefix, but the parameters that come back from CyberSource do. I don't see us adding or removing that prefix anywhere in our code. How can the signature we send initially match the one we generate when handling CyberSource's response?

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -273,6 +276,23 @@ def _generate_signature(self, parameters, use_sop_profile):
...
+            order_number = parameters['reference_number']
+        elif 'req_reference_number' in parameters:
+            order_number = parameters['req_reference_number']
+
+        if order_number:
+            basket_id = str(OrderNumberGenerator().basket_id(order_number))
+
+        logger.info(
+            'Signing CyberSource payment data for basket [%s], to become order [%s].',
+            basket_id,
```
schenedx, line 292: Does `logger.info` handle null basket_id and/or order_number?

schenedx (approved):
👍

ecommerce/extensions/payment/processors/cybersource.py
clintonb: See the signed_parameters field. The signature is generated based on that field's value. The keys used for the signature are not hard coded.

rlucioni: @clintonb I see that, but we include the keys in the signed message. If the keys change after we sign the initial set of parameters, shouldn't that be a problem?

ecommerce/extensions/payment/processors/cybersource.py
rlucioni, line 292: @schenedx yes. The string representation is used: 'None'.

ecommerce/extensions/payment/processors/cybersource.py
clintonb: The keys may change, but the values should be the same. If they differ, the signatures should not match.

PR #1358, opened 2017-06-29
Fix the BasketCalculateView so it does not throw an error when used.

douglashall:
The method [signature](https://github.com/django-oscar/django-oscar/blame/1.4/src/oscar/apps/offer/applicator.py#L19) of the Applicator apply method is Applicator.apply(self, basket, user=None, request=None).
We were calling this method from the [BasketCalculateView](https://github.com/edx/ecommerce/blame/master/ecommerce/extensions/api/v2/views/baskets.py#L386) with Applicator().apply(basket, request) causing
a Request object to be used when a user object was expected.

This bug is only surfaced when this view is working with a basket that has a voucher applied to it. If a voucher is not applied to the basket the Oscar code never uses the second parameter passed to Applicator.apply method. The enterprise entitlement feature applies a voucher within the BasketCalculateView if the basket qualifies for an enterprise entitlement so this is the only current use case which would have surfaced this bug.

### DIFF
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -383,7 +383,7 @@ class BasketCalculateView(generics.GenericAPIView):
                     basket.vouchers.add(voucher)
 
                 # Calculate any discounts on the basket.
-                Applicator().apply(basket, request)
+                Applicator().apply(basket, user=request.user, request=request)
 
                 discounts = []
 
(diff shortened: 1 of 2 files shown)


mjfrey (approved):

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -383,7 +383,7 @@ def get(self, request):
                     basket.vouchers.add(voucher)
 
                 # Calculate any discounts on the basket.
-                Applicator().apply(basket, request)
+                Applicator().apply(basket, request.user, request)
```
clintonb: Use the kwarg names.

PR #1360, opened 2017-06-29
LEARNER-1369 Update the order receipt frontend event for 'Completed Purchase'

schenedx:
According to the Google Tag at https://tagmanager.google.com/?hl=en#/container/accounts/12009721/containers/652832/workspaces/80/triggers
we can update the front end trigger event name so we don't double count the order in GA

@mulby @clintonb Please help review

Once this gets deployed, I will make "API name" change in OptimizelyX "Order Completed" event

### DIFF
--- a/ecommerce/static/js/pages/receipt_page.js
+++ b/ecommerce/static/js/pages/receipt_page.js
@@ -9,7 +9,7 @@ define([
         'use strict';
 
         function trackPurchase(orderId, totalAmount, currency) {
-            window.analytics.track('Order Completed', {
+            window.analytics.track('Completed Purchase', {
                 orderId: orderId,
                 total: totalAmount,
                 currency: currency


mulby:
This looks fine, did you run the tests we talked about in Optimizely?

schenedx:
@mulby I haven't had a chance to test. If we want to test this in stage, I'd have to merge this change into master first.

mulby (approved):

clintonb (approved):

schenedx:
@mulby I tested this change on stage. Once I updated the Optimizely "Order Completed" event api-name to "Completed Purchase", the results did not clear out. The result stays and the new event was captured successfully. I think we are good to move this change into production. I plan to do this on Wednesday next week. Once we move this into production, I will update the event api-name in Optimizely for production

PR #1361, opened 2017-06-29
Escaped string for translation

BenjiLee:
@edx/learner Please review.


ecommerce/extensions/payment/forms.py
```diff
@@ -22,7 +22,7 @@ def country_choices():
     )
     # Inserting a placeholder here so that the first option
     # when rendering the dropdown isn't a valid country.
-    countries.insert(0, ('', _('<Choose country>')))
+    countries.insert(0, ('', _('&lt;Choose country&gt;')))
```
clintonb: The symbols do not need to be translated. Move them outside of the call to `gettext`. Also, why are they HTML instead of the existing characters?

clintonb (changes requested):

ecommerce/extensions/payment/forms.py
BenjiLee: Transifex was trying to treat the <> as an HTML tag so I replaced them. Your suggestion to not wrap the tags is a better idea.

BenjiLee:
@clintonb Updated PR. Ready for another pass.

clintonb (approved):

clintonb:
Generate updated translation files (skip djangojs.*).

PR #1362, opened 2017-06-29
Prevent accidental resubmission of payment form

rlucioni:
Adds JavaScript to the receipt page that manipulates the contents of the history stack using the URL fragment to prevent users from accidentally resubmitting the payment form by discouraging them from using their browser's back button.

LEARNER-1671

@edx/learner FYI


ecommerce/extensions/checkout/views.py
```diff
@@ -146,6 +147,9 @@ def get_context_data(self, **kwargs):
         context = super(ReceiptResponseView, self).get_context_data(**kwargs)
         order = context[self.context_object_name]
         context.update({
+            'back_button_warning': _(
+                'Caution! Using the back button on this page may cause you to be charged again.'
```
rlucioni: I'm debating whether this message is the best thing to display. I want to let users know why their back button doesn't work on this page, but I also don't want to encourage more curious users from trying to circumvent this just to see what happens.

rlucioni:
Looks like there's a problem finding the `xvfb-run` command on Travis. This is what's failing the build.

ecommerce/templates/edx/checkout/receipt.html
```diff
@@ -23,6 +23,7 @@
 {% block content %}
   <div id="receipt-container"
        class="receipt container content-container"
+       data-back-button-warning="{{ back_button_warning }}"
```
clintonb: There is no need to pass this from Django to JS. Just set the string in your `alert()` call and mark it with `gettext()`.

clintonb (approved):
Move the text to JS, and generate translation files.

rlucioni:
It took a look at the failing Travis builds. [Build #15413](https://travis-ci.org/edx/ecommerce/builds/248522910) is one of the most recent clean builds I can find. [Build #15451](https://travis-ci.org/edx/ecommerce/builds/248608463) is an example of a failing build. Both ran on Travis' new Trusty `sudo: required` environment, but we should't be affected because we're using Docker.

I noticed that earlier, successful builds pulled an edxops/ecommerce image with SHA `8b2b438`. This and other failing builds are pulling edxops/ecommerce image with SHA `67ec7b3`. I suspect that a difference between the two images is causing newer builds to fail because they can't find `xvfb-run`. [Docker Hub](https://hub.docker.com/r/edxops/ecommerce/tags/) confirms that a new edxops/ecommerce image was pushed roughly 5 hours from this writing:



The [Dockerfile](https://github.com/edx/configuration/tree/master/docker/build/ecommerce) in edx/configuration hasn't been touched in 5 months. @clintonb is there any chance you're familiar with the changes in the new image or can point me in the right direction? At the very least, this goes to show that we should be pinning to versioned images instead of using the floating `latest` tag.

@BenjiLee FYI, this same issue is breaking your build over on https://github.com/edx/ecommerce/pull/1361.

clintonb:
I'm not familiar with any changes in that image. I am building a new one. It should be pushed in about 25 minutes.

Pinning images does not solve this problem. In general the image with the `latest` tag should contain the code and requirements from the tip of the `master` branch. @edx/devops is finishing up a Jenkins job to handle creating Ansible-based images. I still prefer the approach used in edx/credentials where we build the image for every Travis build.

rlucioni:
@clintonb thank you, I'll try rerunning the build here once that image is available. I have some more questions about pinning but will save them for later. I also prefer the Travis (or Docker Cloud) approach where pushes to master trigger image builds and pushes.

clintonb:
Image pushed. Builds restarted.

ecommerce/templates/edx/checkout/receipt.html
rlucioni: @clintonb thanks for the reminder. This is updated.

rlucioni:
@clintonb I've addressed your comments. We have a job for automatically extracting strings marked for translation. It currently only runs against edx/edx-mktg. I'm wondering if we should apply it to other repos to avoid relying on developers for this extraction.

I've also made a minor revision here after playing around with this locally. The replacement hash is now derived from a timestamp to help keep this logic working as expected if a user uses the back button to return to this page after the replacement hash has been pushed onto the history stack (e.g., receipt page -> verification pages -> back to receipt page).

schenedx (approved):
👍

rlucioni:
The image being used for testing changed again, and `xvfb-run` is missing again.

rlucioni:
We pushed a previous, working image up to get a clean build here. We're speaking with Devops about the job that's pushing broken images.

PR #1363, opened 2017-06-30
Corrected refund event properties

clintonb:
LEARNER-1565


clintonb:
FYI @edx/learner this explains whey we see no refund data in Google Analytics.

clintonb:
Never mind: https://community.segment.com/t/80t5rt/whats-with-the-schema-inconsistencies. Segment supports both.

PR #1365, opened 2017-06-30
Updated analytics event code to fallback to product name

clintonb:
Analytics events no longer assume products are associated with courses. If no course is associated, product.title is used.

LEARNER-1566


ecommerce/extensions/analytics/tests/test_utils.py
```diff
@@ -102,5 +104,17 @@ def test_translate_basket_line_for_segment(self):
...
         }
+        self.assertEqual(translate_basket_line_for_segment(line), expected)
+
+        # Products not associated with a Course should still be reported with the product's title instead of
+        # the course ID.
+        seat.course = None
+        seat.save()
+
+        # Refresh the basket
+        basket.lines.all().delete()
```
mjfrey: basket.flush()

mjfrey (approved):
small nit.

PR #1369, opened 2017-07-03
Fire Product Added event when item added to basket

McKenzieW:
Updated Basket model add_product method to include a call to the Google Analytics "Product Added" event.

LEARNER-1566

### DIFF
--- a/ecommerce.yml
+++ b/ecommerce.yml
@@ -1,0 +1,83 @@
+# Ansible managed
+
+AFFILIATE_COOKIE_KEY: dev_affiliate_id
+BROKER_URL: amqp://celery:celery@172.17.0.2:5672
+CACHES:
+    default:
+        BACKEND: django.core.cache.backends.memcached.MemcachedCache
+        KEY_PREFIX: ecommerce
+        LOCATION:
+        - edx.devstack.memcached:11211
+COMMERCE_API_URL: http://127.0.0.1:18000/api/commerce/v1/
+COMPREHENSIVE_THEME_DIRS: null
+COURSE_CATALOG_API_URL: http://edx.devstack.discovery:18381/api/v1/
+DATABASES:
+    default:
+        ATOMIC_REQUESTS: true
+        CONN_MAX_AGE: 60
+        ENGINE: django.db.backends.mysql
+        HOST: edx.devstack.mysql
+        NAME: ecommerce
+        OPTIONS:
+            connect_timeout: 10
+        PASSWORD: password
+        PORT: '3306'
+        USER: ecomm001
+DEFAULT_SITE_THEME: null
+ECOMMERCE_URL_ROOT: http://localhost:18130
+EDX_API_KEY: PUT_YOUR_API_KEY_HERE
+EDX_DRF_EXTENSIONS:
+    OAUTH2_USER_INFO_URL: http://127.0.0.1:18000/oauth2/user_info
+ENABLE_COMPREHENSIVE_THEMING: false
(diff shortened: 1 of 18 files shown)


ecommerce.yml
```diff
@@ -0,0 +1,84 @@
+---
```
clintonb: This shouldn't b here.

ecommerce/extensions/analytics/tests/test_utils.py
```diff
@@ -1,19 +1,19 @@
 import json
-
 import mock
+
```
clintonb: Remove. mock is a third-party lib.

ecommerce/extensions/basket/models.py
```diff
@@ -82,7 +91,3 @@ class BasketAttribute(models.Model):
 
     class Meta(object):
         unique_together = ('basket', 'attribute_type')
-
-
-# noinspection PyUnresolvedReferences
-from oscar.apps.basket.models import *  # noqa isort:skip pylint: disable=wildcard-import,unused-wildcard-import,wrong-import-position
```
clintonb: Restore this block.

ecommerce/extensions/basket/utils.py
```diff
@@ -14,7 +14,6 @@
 from ecommerce.extensions.order.utils import UserAlreadyPlacedOrder
 from ecommerce.extensions.payment.utils import embargo_check
 from ecommerce.referrals.models import Referral
-
```
clintonb: Revert.

clintonb (changes requested):
I think your IDE may be doing more than it should.

Git pro-tips:

1. Squash your commits before requesting a review. As you gain experience you will only need one round of reviews. Get everything in order so you can simply press the merge button without having to debase and wait for the build to pass again. 
2. Rebate on master instead of merging it into your branch to avoid merge bubbles.

ecommerce/extensions/basket/models.py
```diff
@@ -45,6 +47,13 @@ def get_basket(cls, user, site):
 
         return basket
 
+    def add_product(self, product, quantity=1, options=None):
```
marjev, line 49: Please add a docstring for this method. We're trying to write a Google style Python docstring for each method.
See the example here: https://gist.github.com/dsaiztc/9b4c15c6dfef08a957c1
Basic idea is to have a short description, a list of parameters and return values available in a docstring.

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -113,3 +111,15 @@ def test_create_basket(self):
         basket = Basket.create_basket(self.site1, user)
         self.assertEqual(basket.site, self.site1)
         self.assertEqual(basket.owner, user)
+
+    def test_add_product(self):
+        """ Verify the method fires Product Added analytic event when a product
+            is added to the basket """
```
marjev: I think this docstring can fit in a single line. I think we allow 120 characters per line.

marjev (commented):
I see Clinton already left valuable comments. I only noticed two smaller things that might be changed.

ecommerce/extensions/basket/models.py
McKenzieW: This causes issues with the quality check. Should it not be at the top of of the file?

clintonb: Nope. It belongs here. We want our definition of the Basket model to take precedence. Quality checks haven't failed in the months this has been here. It's fine.

McKenzieW:
https://github.com/edx/ecommerce/pull/1382

PR #1371, opened 2017-07-03
Added support for Apple Pay

clintonb:
LEARNER-1668

- [x] Finish unit tests
- [x] Decide on whether to add e2e test (with mocked Apple Pay endpoints)
- [x] Site configuration flag to disable
- [x] Finish JS tests
- [x] Increase overall test coverage


clintonb:
@edx/learner please review

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -348,3 +357,99 @@ def issue_credit(self, order, reference_number, amount, currency):
                 'Failed to issue CyberSource credit for order [{order_number}]. '
                 'Complete response has been recorded in entry [{response_id}]'.format(
                     order_number=order.number, response_id=ppr.id))
+
+    def request_apple_pay_authorization(self, basket, billing_address, payment_token):
```
mjfrey, line 362: Why attach Apple Pay to Cybersource payment processor?  What if we move or add a new processor?

clintonb: Making a generic Apple Pay setup for the first pass is overkill. I thought about it, but the implementations are somewhat processor-specific. Stripe, for example does much of the work for merchants when starting the session. What's written is specific to CyberSource for the simple fact that making it generic would be a bit of a hack.

clintonb:
Related: https://github.com/edx/edx-themes/pull/528

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -348,3 +357,99 @@ def issue_credit(self, order, reference_number, amount, currency):
...
+                'postalCode': billing_address.postcode,
+                'country': billing_address.country.iso_3166_1_a2,
+                'email': basket.owner.email,
+            }
+            purchase_totals = {
+                'currency': basket.currency,
+                'grandTotalAmount': str(basket.total_incl_tax),
+            }
+            encrypted_payment = {
+                'descriptor': 'RklEPUNPTU1PTi5BUFBMRS5JTkFQUC5QQVlNRU5U',
```
rlucioni, line 399: What's this? It looks like hardcoded configuration.

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -348,3 +357,99 @@ def issue_credit(self, order, reference_number, amount, currency):
...
+            response = client.service.runTransaction(
+                merchantID=self.merchant_id,
+                merchantReferenceCode=basket.order_number,
+                billTo=bill_to,
+                purchaseTotals=purchase_totals,
+                encryptedPayment=encrypted_payment,
+                card=card,
+                ccAuthService=auth_service,
+                ccCaptureService=capture_service,
+                paymentSolution='001',
```
rlucioni, line 431: Does this belong in settings?

ecommerce/settings/_oscar.py
```diff
@@ -131,6 +132,13 @@
 }
 
 PAYMENT_PROCESSOR_SWITCH_PREFIX = 'payment_processor_active_'
+
+# TODO Move to CyberSource settings
```
rlucioni: Looks like you took care of this TODO? I don't see these settings being used anywhere.

ecommerce/static/js/payment_processors/cybersource.js
```diff
@@ -123,6 +127,98 @@ require([
...
+                    },
+                    data: JSON.stringify({url: evt.validationURL}),
+                    contentType: 'application/json',
+                    success: function(data) {
+                        console.log('Merchant validation succeeded.');
+                        console.log(data);
+                        session.completeMerchantValidation(data);
+                    },
+                    error: function(jqXHR, textStatus, errorThrown) {
+                        // TODO Display an error message to the learner
```
rlucioni: Should these console logs be removed?

ecommerce/urls.py
```diff
@@ -40,7 +41,13 @@ def handler403(_):
 # NOTE 2: These same patterns are used for rest_framework's browseable API authentication links.
 AUTH_URLS = [url(r'^logout/$', LogoutView.as_view(), name='logout'), ] + auth_urlpatterns
 
-urlpatterns = AUTH_URLS + [
+WELL_KNOWN_URLS = [
+    # TODO Add view for /.well-known/apple-developer-merchantid-domain-association
```
rlucioni: Looks like you took care of this, too.

rlucioni (commented):
@clintonb would it make sense to add a section to the docs for this service that gives a high level overview of how Apple Pay works and how to enable it?

ecommerce/extensions/payment/processors/cybersource.py
clintonb: That is intentional. It's required by CyberSource.

ecommerce/extensions/payment/processors/cybersource.py
clintonb: That is intentional. It's required by CyberSource.

ecommerce/extensions/payment/processors/cybersource.py
rlucioni: @clintonb a comment explaining what this is would be helpful to myself and other maintainers who aren't as familiar with this.

clintonb: Hmm...how about a link to the CyberSource docs in the method's docstring? Candidly, if folks don't know what they are doing—haven't read the docs—they shouldn't be messing around in payment code.

rlucioni: @clintonb sure, that link would be good.

clintonb: Docs and docstring updated.

ecommerce/static/js/payment_processors/cybersource.js
clintonb: I find them useful for development. They don't affect the end user, so I see no great need to remove them.

clintonb:
@rlucioni @mjfrey please take another look when you can. I have higher priority issues, so it will take some time for me to fix the coverage issues and the last test, however, the server side components are ready to go.

.gitignore
```diff
@@ -92,3 +92,7 @@ docs/_build
 
 # Visual Studio Code
 .vscode
+
+# Certificates
+*.cer
+*.pem
```
rlucioni, line 98: @clintonb are there Ansible changes corresponding to this PR that are necessary to manage this certificate?

rlucioni (approved):

.gitignore
clintonb: Yes.

clintonb:
Related PR for certificates: https://github.com/edx/configuration/pull/3990

PR #1372, opened 2017-07-04
Multi-Tenant Support for E-Commerce Service API endpoints

vkaracic:
LEARNER-1130

@edx/helio please review.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_catalog.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_catalog.py
@@ -4,7 +4,6 @@ import ddt
 import httpretty
 import mock
 from django.core.urlresolvers import reverse
-from django.test import RequestFactory
 from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import SlumberBaseException
@@ -14,6 +13,7 @@ from ecommerce.coupons.tests.mixins import CourseCatalogMockMixin
 from ecommerce.courses.tests.mixins import CourseCatalogServiceMockMixin
 from ecommerce.extensions.api.serializers import ProductSerializer
 from ecommerce.extensions.api.v2.tests.views.mixins import CatalogMixin
+from ecommerce.tests.factories import PartnerFactory
 from ecommerce.tests.mixins import ApiMockMixin
 from ecommerce.tests.testcases import TestCase
 
@@ -31,12 +31,6 @@ class CatalogViewSetTest(CatalogMixin, CourseCatalogMockMixin, CourseCatalogServ
     def setUp(self):
         super(CatalogViewSetTest, self).setUp()
         self.client.login(username=self.user.username, password=self.password)
-
-    def prepare_request(self, url):
-        factory = RequestFactory()
-        request = factory.get(url)
-        request.site = self.site
-        return request
 
     def test_staff_authorization_required(self):
         """Verify that only users with staff permissions can access the API. """
@@ -57,7 +51,9 @@ class CatalogViewSetTest(CatalogMixin, CourseCatalogMockMixin, CourseCatalogServ
         self.assertEqual(response.status_code, 403)
 
     def test_catalog_list(self):
-        """Verify the endpoint returns all catalogs."""
+        """The list endpoint should return only catalogs with current site's partner."""
+        Catalog.objects.create(name='New catalog', partner=PartnerFactory())
+        self.assertEqual(Catalog.objects.count(), 2)
         response = self.client.get(self.catalog_list_path)
         expected_data = self.serialize_catalog(self.catalog)
         response_data = json.loads(response.content)
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -27,7 +27,7 @@ from ecommerce.extensions.voucher.models import CouponVouchers
 from ecommerce.invoice.models import Invoice
 from ecommerce.programs.constants import BENEFIT_MAP
 from ecommerce.programs.custom import class_path
-from ecommerce.tests.factories import ProductFactory, SiteConfigurationFactory
+from ecommerce.tests.factories import PartnerFactory, ProductFactory, SiteConfigurationFactory
 from ecommerce.tests.mixins import ThrottlingMixin
 from ecommerce.tests.testcases import TestCase
 
@@ -480,7 +480,9 @@ class CouponViewSetFunctionalTest(CouponMixin, CourseCatalogTestMixin, CourseCat
         self.assert_post_response_status(self.data, status.HTTP_403_FORBIDDEN)
 
     def test_list_coupons(self):
-        """Test that the endpoint returns information needed for the details page."""
+        """The list endpoint should return only coupons with current site's partner."""
+        self.create_coupon(partner=PartnerFactory())
+        self.assertEqual(Product.objects.filter(product_class__name='Coupon').count(), 2)
         response = self.client.get(COUPONS_LINK)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         coupon_data = json.loads(response.content)['results'][0]
--- a/ecommerce/extensions/api/v2/tests/views/test_courses.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_courses.py
@@ -25,7 +25,6 @@ User = get_user_model()
 
 
 class CourseViewSetTests(ProductSerializerMixin, CourseCatalogTestMixin, TestCase):
-    maxDiff = None
     list_path = reverse('api:v2:course-list')
 
     def setUp(self):
@@ -98,6 +97,9 @@ class CourseViewSetTests(ProductSerializerMixin, CourseCatalogTestMixin, TestCas
 
     def test_list(self):
         """ Verify the view returns a list of Courses. """
+        CourseFactory()
+        self.assertEqual(Course.objects.count(), 2)
+
         response = self.client.get(self.list_path)
         self.assertEqual(response.status_code, 200)
         self.assertListEqual(json.loads(response.content)['results'], [self.serialize_course(self.course)])
--- a/ecommerce/extensions/api/v2/tests/views/test_products.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_products.py
@@ -14,6 +14,7 @@ from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.extensions.api.serializers import ProductSerializer
 from ecommerce.extensions.api.v2.tests.views import JSON_CONTENT_TYPE, ProductSerializerMixin
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
+from ecommerce.tests.factories import PartnerFactory, ProductFactory
 from ecommerce.tests.testcases import TestCase
 
 Benefit = get_model('offer', 'Benefit')
@@ -21,6 +22,8 @@ Catalog = get_model('catalogue', 'Catalog')
 Product = get_model('catalogue', 'Product')
 ProductClass = get_model('catalogue', 'ProductClass')
 Voucher = get_model('voucher', 'Voucher')
+
+PRODUCT_LIST_PATH = reverse('api:v2:product-list')
 
 
 class ProductViewSetBase(ProductSerializerMixin, CourseCatalogTestMixin, TestCase):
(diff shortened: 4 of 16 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1372"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

vkaracic:
@clintonb please review.

ecommerce/extensions/api/v2/views/catalog.py
```diff
@@ -26,6 +26,11 @@ class CatalogViewSet(NestedViewSetMixin, ReadOnlyModelViewSet):
     serializer_class = serializers.CatalogSerializer
     permission_classes = (IsAuthenticated, IsAdminUser,)
 
+    def get_queryset(self):
+        return super(CatalogViewSet, self).get_queryset().filter(
```
clintonb: Nit: Ditch `super` and the `queryset` field above.

`return Catalog.objects.filter(partner=self.request.site.siteconfiguration.partner)`

ecommerce/extensions/api/v2/views/courses.py
```diff
@@ -21,18 +21,19 @@ class CourseViewSet(NonDestroyableModelViewSet):
...
-    queryset = Course.objects.all().prefetch_related(
-        products_prefetch, product_attribute_value_prefetch, 'products__stockrecords'
-    ).all()
     serializer_class = serializers.CourseSerializer
     permission_classes = (IsAuthenticated, IsAdminUser,)
 
+    def get_queryset(self):
+        return Course.objects.filter(site=self.request.site).prefetch_related(
+            self.products_prefetch, self.product_attribute_value_prefetch, 'products__stockrecords'
+        ).all()
```
clintonb: Drop `.all()`

ecommerce/extensions/api/v2/views/products.py
```diff
@@ -12,8 +13,14 @@
 
 
 class ProductViewSet(NestedViewSetMixin, NonDestroyableModelViewSet):
-    queryset = Product.objects.all()
     serializer_class = serializers.ProductSerializer
     filter_backends = (filters.DjangoFilterBackend,)
     filter_class = ProductFilter
     permission_classes = (IsAuthenticated, IsAdminUser,)
+    queryset = Product.objects.all()
```
clintonb: Move this into the method below.

ecommerce/extensions/api/v2/views/products.py
```diff
@@ -12,8 +13,14 @@
...
+
+    def get_queryset(self):
+        return super(ProductViewSet, self).get_queryset().filter(
+            Q(stockrecords__partner=self.request.site.siteconfiguration.partner) |
+            Q(course__site=self.request.site)
```
clintonb: Don't assume products are associated with courses. Remove this. The stock record filtering is sufficient.

clintonb (changes requested):
Every viewset needs at least one test on the list endpoint validating these changes.

ecommerce/extensions/api/v2/views/products.py
vkaracic: Parent course products don't have stock records associated with them.

ecommerce/extensions/api/v2/views/catalog.py
vkaracic: This won't work with nested catalogs. `NestedViewSetMixin` should first do the filtering based on the parent ID and then filter those results by partner. Filtering directly here will render the nesting useless since it would return all the catalogs (filtered by partner) regardless of the parent ID.

clintonb: Add a comment explaining this.

ecommerce/extensions/api/v2/views/products.py
clintonb: Add a comment explaining this.

clintonb (approved):

clintonb:
Remember to add comments for the last two items, and squash your commits.

vkaracic:
@clintonb I've added tests for each viewset. Please take another look.

clintonb (changes requested):
You did a bit too much work for these tests. Simplify:

```python
def test_list(self):
  # TODO Create data for multiple partners
  # TODO Verify data is returned for a single partner
```
For the existing tests, you should only need to create some data for another partner. The existing tests are sufficient.

vkaracic:
@clintonb take another look please.

vkaracic:
@clintonb ping.

ecommerce/extensions/api/v2/tests/views/test_catalog.py
```diff
@@ -56,14 +51,15 @@ def test_authentication_required(self):
         response = self.client.get(self.catalog_list_path)
         self.assertEqual(response.status_code, 403)
 
+    def assert_catalog_list_response(self, catalog):
+        """Assert that the catalog list endpoint response content is as expected."""
```
marjev: I think you're missing the body of the method.

ecommerce/extensions/api/v2/tests/views/test_catalog.py
```diff
@@ -56,14 +51,15 @@ def test_authentication_required(self):
...
-        """Verify the endpoint returns all catalogs."""
-        response = self.client.get(self.catalog_list_path)
-        expected_data = self.serialize_catalog(self.catalog)
-        response_data = json.loads(response.content)
-        self.assertEqual(response.status_code, 200)
-        self.assertEqual(response_data['count'], 1)
-        self.assertListEqual(response_data['results'], [expected_data])
+        """The list endpoint should return only catalogs with current site's partner."""
+        new_partner = PartnerFactory()
+        Catalog.objects.create(name='New catalog', partner=new_partner)
```
marjev: Can you send `PartnerFactory()` directly? I don't see `new_partner` being used again.

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -480,13 +482,11 @@ def test_authorization_required(self):
...
-        """Test that the endpoint returns information needed for the details page."""
-        response = self.client.get(COUPONS_LINK)
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-        coupon_data = json.loads(response.content)['results'][0]
-        self.assertEqual(coupon_data['title'], self.data['title'])
-        self.assertEqual(coupon_data['category']['name'], self.data['category']['name'])
-        self.assertEqual(coupon_data['client'], self.data['client'])
+        """The list endpoint should return only coupons with current site's partner."""
+        new_partner = PartnerFactory()
+        self.create_coupon(partner=new_partner)
```
marjev: Can these two lines be one? Is `new_partner` ever reused?

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -37,21 +41,38 @@ def setUp(self):
 
 class ProductViewSetTests(ProductViewSetBase):
     def test_list(self):
-        """ Verify a list of products is returned. """
-        path = reverse('api:v2:product-list')
-        response = self.client.get(path)
+        """The list endpoint should return only products with current site's partner."""
+        new_partner = PartnerFactory()
+        ProductFactory(stockrecords__partner=new_partner)
```
marjev: Same comment as before.

ecommerce/extensions/api/v2/tests/views/test_stockrecords.py
```diff
@@ -30,10 +31,15 @@ def setUp(self):
 
     def test_list(self):
         """ Verify a list of stock records is returned. """
+        new_partner = PartnerFactory()
+        StockRecordFactory(partner=new_partner)
```
marjev: Same comment as before.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -43,15 +47,25 @@ def setUp(self):
...
         for voucher in vouchers:
             voucher.offers.add(ConditionalOfferFactory())
+            coupon_vouchers.vouchers.add(voucher)
+        return vouchers
 
+    def test_list(self):
+        """ Verify the endpoint lists all vouchers. """
+        vouchers = self.create_vouchers(count=3)
+        new_partner = PartnerFactory()
+        self.create_vouchers(partner=new_partner)
```
marjev: Same comment as before.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -827,7 +827,7 @@ def test_redirects_with_enterprise_customer(self, get_ec):
             reverse('coupons:redeem'),
             'sku={sku}'.format(sku=stock_record.partner_sku),
             'code={code}'.format(code=COUPON_CODE),
-            'failure_url=http%3A%2F%2Ftestserver%2Fbasket%2F%3Fconsent_failed%3D{code}'.format(code=COUPON_CODE)
+            'failure_url=http%3A%2F%2Ftestserver.fake%2Fbasket%2F%3Fconsent_failed%3D{code}'.format(code=COUPON_CODE)
```
marjev: Shouldn't this be part of the format parameters as well? I think `testserver.fake` comes from either a Factory instance or settings.

marjev (changes requested):

vkaracic:
@marjev comments addressed.

ecommerce/extensions/api/v2/tests/views/mixins.py
```diff
@@ -34,3 +36,18 @@ def serialize_catalog(self, catalog):
...
+        """Assert that the list endpoint response content is as expected."""
+        response = self.client.get(path)
+        # We're doing JSON dumping and loading here to convert any tuples to lists
+        # the same way they are converted by the endpoint.
+        expected_data = json.dumps(serializer(obj, context={'request': self.request}).data)
+        expected_data = json.loads(expected_data)
+        response_data = json.loads(response.content)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response_data['count'], count)
```
clintonb: Nit: If the results match, checking the count doesn't really matter.

ecommerce/extensions/api/v2/tests/views/mixins.py
```diff
@@ -34,3 +36,18 @@ def serialize_catalog(self, catalog):
...
+
+class ApiTestMixin(object):
+    def assert_endpoint_list_response(self, path, serializer, obj, count=1):
+        """Assert that the list endpoint response content is as expected."""
+        response = self.client.get(path)
+        # We're doing JSON dumping and loading here to convert any tuples to lists
+        # the same way they are converted by the endpoint.
+        expected_data = json.dumps(serializer(obj, context={'request': self.request}).data)
+        expected_data = json.loads(expected_data)
+        response_data = json.loads(response.content)
```
clintonb: Use `response.data` to get Python instead of JSON, and you can ditch all of the conversion.

ecommerce/extensions/api/v2/tests/views/mixins.py
```diff
@@ -34,3 +36,18 @@ def serialize_catalog(self, catalog):
...
         return data
+
+
+class ApiTestMixin(object):
+    def assert_endpoint_list_response(self, path, serializer, obj, count=1):
+        """Assert that the list endpoint response content is as expected."""
+        response = self.client.get(path)
+        # We're doing JSON dumping and loading here to convert any tuples to lists
+        # the same way they are converted by the endpoint.
+        expected_data = json.dumps(serializer(obj, context={'request': self.request}).data)
```
clintonb: Nit: Use the `many` kwarg: `expected = serializer([obj], context={'request': self.request}, many=True).data`

clintonb (commented):
You're still doing way too much with these tests. I'm not a fan of the mixin because it's a bit too generic.

Literally all you need to do for the existing tests is create some for the non-default partner and add a comment informing future developers the data should not be returned. There is no need for any of these other changes to test code

```python
def test_list(self):
  # This data won't be returned because it's associated with the wrong site/partner.
  ProductFactory.create_batch(3, partner=PartnerFactory())

  # <Existing Test>
```

vkaracic:
@clintonb comments addressed!

ecommerce/extensions/api/v2/tests/views/test_courses.py
```diff
@@ -98,6 +98,10 @@ def test_jwt_authentication(self):
 
     def test_list(self):
         """ Verify the view returns a list of Courses. """
+        new_site = SiteFactory()
+        CourseFactory(id='new/test/course', site=new_site)
```
clintonb: All you need is `CourseFactory()`. It will do all the work for you.

ecommerce/extensions/api/v2/tests/views/test_stockrecords.py
```diff
@@ -30,10 +30,14 @@ def setUp(self):
 
     def test_list(self):
         """ Verify a list of stock records is returned. """
+        StockRecordFactory(partner=PartnerFactory())
```
clintonb: Just use `StockRecordFactory()`.

clintonb (approved):

vkaracic:
@marjev please re-review.

marjev (approved):

PR #1373, opened 2017-07-05
Fire Payment Info Entered event after successful payment entry

McKenzieW:
Updated handle_payment method in EdxOrderPlacementMixin to fire the segment analytic event 'Payment Info Entered' when payment information has been validated.

### DIFF
--- a/ecommerce/extensions/checkout/mixins.py
+++ b/ecommerce/extensions/checkout/mixins.py
@@ -10,7 +10,7 @@ from ecommerce_worker.fulfillment.v1.tasks import fulfill_order
 from oscar.apps.checkout.mixins import OrderPlacementMixin
 from oscar.core.loading import get_class, get_model
 
-from ecommerce.extensions.analytics.utils import audit_log
+from ecommerce.extensions.analytics.utils import audit_log, track_segment_event
 from ecommerce.extensions.api import data as data_api
 from ecommerce.extensions.checkout.exceptions import BasketNotFreeError
 from ecommerce.extensions.customer.utils import Dispatcher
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/checkout/mixins.py
```diff
@@ -52,6 +52,8 @@ def handle_payment(self, response, basket):
         linked to the order when it is saved later on.
         """
         handled_processor_response = self.payment_processor.handle_processor_response(response, basket=basket)
+        track_segment_event(basket.site, basket.owner, 'Payment Info Entered',
+                            {'checkout_id': handled_processor_response.transaction_id})
```
clintonb: This should be `basket.id`.

ecommerce/extensions/checkout/tests/test_mixins.py
```diff
@@ -50,6 +51,7 @@ def test_handle_payment_logging(self, __):
         user = factories.UserFactory()
```
mjfrey: no need to create another user.  Just use self.user this is created in setUp().

mjfrey (changes requested):

clintonb (changes requested):

ecommerce/extensions/checkout/mixins.py
clintonb: Actually, use `basket.order_number`.

mjfrey (approved):
looks good.

ecommerce/extensions/checkout/mixins.py
```diff
@@ -52,6 +52,8 @@ def handle_payment(self, response, basket):
         linked to the order when it is saved later on.
         """
         handled_processor_response = self.payment_processor.handle_processor_response(response, basket=basket)
+        track_segment_event(basket.site, basket.owner, 'Payment Info Entered',
+                            {'checkout_id': basket.order_number})
```
clintonb: Nit: single line, if it fits.

clintonb (approved):

PR #1374, opened 2017-07-06
Disabled client-side tracking if Segment ID is empty

clintonb:
Tracking is now disabled if the Segment application ID is set to an empty string (e.g. the default for local development).

LEARNER-1668


clintonb:
FYI @edx/learner

PR #1375, opened 2017-07-06
Removed JSCS

clintonb:
ESLint is sufficient as JSCS is no longer maintained.

LEARNER-1668


clintonb:
FYI @edx/learner

clintonb:
http://eslint.org/docs/user-guide/migrating-from-jscs

andy-armstrong (approved):
👍  nice clean up! As you say, ESLint covers all the same bases, so JSCS is unnecessary.

PR #1383, opened 2017-07-11
Running Pylint with parallel execution

clintonb:
Pylint now uses all cores/CPUs on the machine.


clintonb:
FYI @edx/learner

rlucioni:
Thanks @clintonb!

PR #1384, opened 2017-07-11
Reporting coverage for all JS files

clintonb:
Our current coverage reports are invalid because they exclude JS files that are not loaded by the tests. These files have 0% coverage, thus our reports over-estimate test coverage.


clintonb:
Related: https://github.com/karma-runner/karma-coverage/issues/125

clintonb:
FYI @edx/learner

PR #1368, opened 2017-07-03
Updated E-Commerce Service to use a site-specific URL for Catalog Service

iivic:
LEARNER-1135

TODO:

- [x] Check for program caching and enable multitenancy support for caching.

- [ ] Merge https://github.com/edx/devstack/pull/178 after this PR merges.

PR #1385, opened 2017-07-11
Updated E-Commerce Service to use a site-specific URL for Discovery Service

iivic:
LEARNER-1135

Splitting https://github.com/edx/ecommerce/pull/1368/files into three pieces this one should be merged first.

@clintonb FYI.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1385"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/tests/factories.py
```diff
@@ -35,6 +35,7 @@ class Meta(object):
     send_refund_notifications = False
     enable_sdn_check = False
     enable_embargo_check = False
+    discovery_api_url = factory.LazyAttribute(lambda obj: 'http://{domain}/api/v1/'.format(domain=obj.site.domain))
```
clintonb: Use faker: http://faker.readthedocs.io/en/master/providers/faker.providers.internet.html?highlight=url#faker-providers-internet

ecommerce/programs/tests/test_forms.py
```diff
@@ -73,6 +74,7 @@ def test_clean_with_conflicting_program_uuid(self):
     @httpretty.activate
     def test_save_create(self):
         """ A new ConditionalOffer, Benefit, and Condition should be created. """
+        toggle_switch('use_multi_tenant_discovery_api_urls', True)
```
clintonb: Why duplicate? Move to `setUp()`.

ecommerce/programs/tests/test_conditions.py
```diff
@@ -23,6 +24,7 @@ def test_name(self):
     @httpretty.activate
     def test_get_program(self):
         """ The method should return data from the Catalog Service API. Data should be cached for subsequent calls. """
+        toggle_switch('use_multi_tenant_discovery_api_urls', True)
```
clintonb: Move to `setUp()`.

ecommerce/programs/tests/mixins.py
```diff
@@ -2,14 +2,15 @@
...
 import httpretty
-from django.conf import settings
 
 from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.courses.utils import mode_for_seat
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 
 
 class ProgramTestMixin(CourseCatalogTestMixin):
+    DEFAULT_DISCOVERY_API_URL = 'http://testserver.fake/api/v1'
```
clintonb: No more hardcoding. This should come from `self.site_configuration`.

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -139,6 +140,7 @@ def create_catalog_coupon(
             catalog_query='*:*',
             course_seat_types='verified'
     ):
+        toggle_switch("use_multi_tenant_discovery_api_urls", True)
```
clintonb: Move to `setUp()`. Also, use single quotes.

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -107,6 +108,7 @@ def test_run_catalog_query(self):
         """
         run_course_query() should return True for included course run ID's.
         """
+        toggle_switch('use_multi_tenant_discovery_api_urls', True)
```
clintonb: Move to `setUp()`.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -130,6 +130,7 @@ def test_with_entitlement_voucher_consent_failed(self, mock_get_entitlement_vouc
         """
         Since consent has already failed, we ought to follow the standard flow, rather than looping forever.
         """
+        toggle_switch("use_multi_tenant_discovery_api_urls", True)
```
clintonb: Move to `setUp()`.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -78,6 +79,7 @@ def prepare_get_offers_response(self, quantity=1, seat_type='verified', seats=No
         Returns:
             The products, request and vouchers created.
         """
+        toggle_switch('use_multi_tenant_discovery_api_urls', True)
```
clintonb: Move to `setUp()`.

ecommerce/extensions/api/v2/tests/views/test_catalog.py
```diff
@@ -102,6 +103,7 @@ def test_catalog_products(self):
     @mock_course_catalog_api_client
     def test_preview_success(self):
         """ Verify the endpoint returns a list of catalogs from the Catalog API. """
+        toggle_switch("use_multi_tenant_discovery_api_urls", True)
```
clintonb: Move to `setUp()`.

ecommerce/enterprise/tests/test_entitlements.py
```diff
@@ -138,6 +138,7 @@ def test_get_entitlement_voucher_with_enterprise_feature_enabled(self):
         Verify that method "get_entitlement_voucher" returns a voucher if
         the enterprise feature is enabled.
         """
+        toggle_switch('use_multi_tenant_discovery_api_urls', True)
```
clintonb: Move to `setUp()`.

ecommerce/coupons/tests/mixins.py
```diff
@@ -19,12 +18,15 @@
 
 class CourseCatalogMockMixin(object):
     """ Mocks for the Course Catalog responses. """
+    DEFAULT_DISCOVERY_API_URL = 'http://testserver.fake/api/v1/'
```
clintonb: Create from `self.site_configuration`.

clintonb (changes requested):
There is a lot of unnecessary duplication of the `toggle_switch()` call. I suggest moving it to `setUp()` of the base `TestCase` class. That way there is very little code to remove to roll this out permanently.

ecommerce/tests/factories.py
iivic: The idea was to activate the switch for tests that really need it instead of running it for every test. Since it will get removed anyway your suggestion makes more sense.

ecommerce/programs/tests/test_forms.py
iivic: The idea was to activate the switch for tests that really need it. Since it will get removed anyway your suggestion makes more sense.

ecommerce/tests/factories.py
iivic: In order to add the fake url the `@mock_course_catalog_api_client` decorator must be removed. I was planing to do so in the refactor PR. I will include this comment there.

clintonb: The factory is not dependent on decorators. You can modify the factory with faker instead of using the lambda with no issue.

iivic: Decorator mocks the SiteConfiguration.discovery_api_client while mocks like mock_dynamic_catalog_single_course_runs_api use the factory discovery_api_url to create the discovery client and if we add the faker without removing the decorator the urls will not match.

clintonb: Perhaps the decorator should have been removed first. Add the comment, and remember to switch to `Faker` after the decorator is removed.

iivic:
@clintonb please verify everything is in place. I would like to squash and merge.

ecommerce/tests/testcases.py
```diff
@@ -22,7 +23,9 @@ class TestCase(TestServerUrlMixin, UserMixin, SiteMixin, CacheMixin, DjangoTestC
 
     This class guarantees that tests have a Site and Partner available.
     """
-    pass
+    def setUp(self):
+        super(TestCase, self).setUp()
+        toggle_switch('use_multi_tenant_discovery_api_urls', True)
```
clintonb: Add a comment explaining that this is temporary, and linking to the ticket to remove this.

clintonb (approved):

iivic:
@clintonb strange I covered my changes but codecov is complaining. What is the procedure in this case?

clintonb:
You touched files that previously had less than 100% coverage. Our overall coverage dropped after https://github.com/edx/ecommerce/pull/1384 was merged because not all JS files are covered. I'll merge and update Codecov's threshold.

PR #1386, opened 2017-07-11
Fire Product Removed event when item removed from basket

McKenzieW:
Before flushing a basket, fire 'Product Removed' Segment event for each product that was in the basket.

[LEARNER-1567]

### DIFF
--- a/ecommerce/extensions/basket/models.py
+++ b/ecommerce/extensions/basket/models.py
@@ -3,6 +3,9 @@ from django.utils.translation import ugettext_lazy as _
 from oscar.apps.basket.abstract_models import AbstractBasket
 from oscar.core.loading import get_class
 
+from ecommerce.extensions.analytics.utils import track_segment_event, translate_basket_line_for_segment
+
+
 OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
 Selector = get_class('partner.strategy', 'Selector')
 
@@ -45,6 +48,13 @@ class Basket(AbstractBasket):
 
         return basket
 
+    def flush(self):
+        """Remove all products in basket and fire Segment 'Product Removed' Analytic event for each"""
+        for line in self.all_lines():
+            properties = translate_basket_line_for_segment(line)
+            track_segment_event(self.site, self.owner, 'Product Removed', properties)
+        super(Basket, self).flush()  # pylint: disable=bad-super-call
+
     def clear_vouchers(self):
         """Remove all vouchers applied to the basket."""
         for v in self.vouchers.all():
@@ -85,4 +95,4 @@ class BasketAttribute(models.Model):
 
 
 # noinspection PyUnresolvedReferences
-from oscar.apps.basket.models import *  # noqa isort:skip pylint: disable=wildcard-import,unused-wildcard-import,wrong-import-position
+from oscar.apps.basket.models import *  # noqa isort:skip pylint: disable=wildcard-import,unused-wildcard-import,wrong-import-position,wrong-import-order,ungrouped-imports
--- a/ecommerce/extensions/basket/tests/test_models.py
+++ b/ecommerce/extensions/basket/tests/test_models.py
@@ -1,9 +1,15 @@
 import itertools
+import mock
 
+from analytics import Client
 from django.contrib.sites.models import Site
 from oscar.core.loading import get_class, get_model
 from oscar.test import factories
 
+from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.extensions.analytics.utils import parse_tracking_context, translate_basket_line_for_segment
+from ecommerce.extensions.test.factories import create_basket
+from ecommerce.tests.factories import SiteConfigurationFactory
 from ecommerce.tests.testcases import TestCase
 
 Basket = get_model('basket', 'Basket')
@@ -113,3 +119,44 @@ class BasketTests(TestCase):
         basket = Basket.create_basket(self.site1, user)
         self.assertEqual(basket.site, self.site1)
         self.assertEqual(basket.owner, user)
+
+    def test_flush_with_product(self):
+        """
+        Verify the method fires 'Product Removed' Segment event with the correct information when basket is not empty
+        """
+        self.site1.siteconfiguration = SiteConfigurationFactory()
+        self.site1.siteconfiguration.segment_key = 'fake_key'
+        basket = create_basket(empty=True)
+        basket.owner = factories.UserFactory()
+        basket.site = self.site1
+        basket.save()
+
+        course = CourseFactory()
+        seat = course.create_or_update_seat('verified', True, 100, self.partner)
+        basket.add_product(seat)
+
+        properties = translate_basket_line_for_segment(basket.lines.first())
+        user_tracking_id, lms_client_id, lms_ip = parse_tracking_context(basket.owner)
+        context = {
+            'ip': lms_ip,
+            'Google Analytics': {
+                'clientId': lms_client_id
+            }
+        }
+
+        with mock.patch.object(Client, 'track') as mock_track:
+            basket.flush()
+            mock_track.assert_called_once_with(user_tracking_id, 'Product Removed', properties, context=context)
+
+    def test_flush_without_product(self):
+        """ Verify the method does not fireSegment event when basket is empty """
+        self.site1.siteconfiguration = SiteConfigurationFactory()
+        self.site1.siteconfiguration.segment_key = 'fake_key'
+
+        basket = create_basket(empty=True)
+        basket.owner = factories.UserFactory()
+        basket.site = self.site1
+
+        with mock.patch.object(Client, 'track') as mock_track:
+            basket.flush()
+            self.assertEqual(mock_track.call_count, 0)
--- a/ecommerce/programs/tests/test_conditions.py
+++ b/ecommerce/programs/tests/test_conditions.py
@@ -35,7 +35,7 @@ class ProgramCourseRunSeatsConditionTests(ProgramTestMixin, TestCase):
         """ The method should return True if the basket contains one course run seat corresponding to each
         course in the program. """
         offer = factories.ProgramOfferFactory(condition=self.condition)
-        basket = factories.BasketFactory(site=self.site)
+        basket = factories.BasketFactory(site=self.site, owner=factories.UserFactory())
         program = self.mock_program_detail_endpoint(
             self.condition.program_uuid, self.site_configuration.discovery_api_url
         )


ecommerce/extensions/basket/utils.py
```diff
@@ -42,6 +43,9 @@ def prepare_basket(request, products, voucher=None):
         basket (Basket): Contains the product to be redeemed and the Voucher applied.
     """
     basket = Basket.get_basket(request.user, request.site)
+    for line in basket.all_lines():
```
mjfrey: Rather than doing this here.  Override the flush method in the Basket model.  Then call Super.

mjfrey (changes requested):

ecommerce/extensions/basket/models.py
```diff
@@ -45,6 +48,12 @@ def get_basket(cls, user, site):
 
         return basket
 
+    def flush(self):
```
mjfrey, line 51: you need to call the Super method so the default behavior still happens.

McKenzieW: yes, so silly of me to forget. Thank you!

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -355,6 +357,47 @@ def test_prepare_basket_for_purchased_enrollment_code(self):
             basket = prepare_basket(self.request, [enrollment_code])
             self.assertIsNotNone(basket)
 
+    def _prepare_basket_segment_event(self, isEmpty):
```
clintonb: Use snake_casing for Python (e.g. `is_empty`). That said, `empty` is sufficient.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -355,6 +357,47 @@ def test_prepare_basket_for_purchased_enrollment_code(self):
...
+        """
+        Helper function for Segment event firing functionality in prepare_basket
+        """
+        basket = BasketFactory(owner=self.request.user, site=self.site_configuration.site)
```
clintonb: Avoid relying on `self.request`. We should not depend on requests outside of views. Use `self.create_user()`.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -355,6 +357,47 @@ def test_prepare_basket_for_purchased_enrollment_code(self):
...
+            context = {
+                'ip': lms_ip,
+                'Google Analytics': {
+                    'clientId': lms_client_id
+                }
+            }
+
+        with mock.patch.object(Client, 'track') as mock_track:
+            prepare_basket(self.request, [product])
+            if isEmpty:
```
clintonb: If statements in tests are a code smell. Your test methods should do one thing well. Move these assertions to the test methods. This method should only be concerned with creating the basket.

clintonb (changes requested):

ecommerce/extensions/basket/models.py
```diff
@@ -45,6 +48,13 @@ def get_basket(cls, user, site):
...
+        """Remove all products in basket and fire Segment 'Product Removed' Analytic event for each"""
+        super(Basket, self).flush()  # pylint: disable=bad-super-call
```
mjfrey: This will need to go after the track calls otherwise there will be no lines.

McKenzieW:
@mjfrey @clintonb can I get a confirmation from y'all on the changes you requested? Otherwise, ready to merge.

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -113,3 +118,39 @@ def test_create_basket(self):
...
         self.assertEqual(basket.owner, user)
+
+    def _flush_segment_event(self, empty):
+        """
+        Helper function for testing Segment event firing in basket.flush()
+        """
+        site = SiteConfigurationFactory().site
+        site.siteconfiguration.segment_key = 'fake_key'
+        user = factories.UserFactory()
+        basket = self._create_basket(user, site)
```
clintonb: Can you pass the empty flag here? It seems strange to flush the basket on the line below only to create another product and add it to the basket.

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -113,3 +118,39 @@ def test_create_basket(self):
...
+            context = {
+                'ip': lms_ip,
+                'Google Analytics': {
+                    'clientId': lms_client_id
+                }
+            }
+
+        with mock.patch.object(Client, 'track') as mock_track:
+            basket.flush()
+            if empty:
```
clintonb: An if statement in the test means you're doing too much for one test, or for one helper method in this case. This should be split appropriately. De-duplication is nice, but not the extent that it makes your tests unreadable.

Compare what you have to this below.

```python
    def test_flush_events(self):
        """ Flushing the basket should fire the "Product Removed" event for all lines in the basket. """
        user = factories.UserFactory()
        basket = Basket.create_basket(self.site, user)

        properties = translate_basket_line_for_segment(basket.lines.first())
        user_tracking_id, lms_client_id, lms_ip = parse_tracking_context(basket.owner)
        context = {
            'ip': lms_ip,
            'Google Analytics': {
                'clientId': lms_client_id
            }
        }

        with mock.patch.object(Client, 'track') as mock_track:
            basket.flush()
            mock_track.assert_called_once_with(user_tracking_id, 'Product Removed', properties, context=context)

    def test_flush_events_for_empty_basket(self):
        """ Flush an empty basket should NOT fire any events. """
        user = factories.UserFactory()
        basket = Basket.create_basket(self.site, user)

        with mock.patch.object(Client, 'track') as mock_track:
            basket.flush()
            self.assertEqual(mock_track.call_count, 0)
```

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -113,3 +118,39 @@ def test_create_basket(self):
...
+                self.assertEqual(mock_track.call_count, 0)
+            else:
+                mock_track.assert_called_once_with(user_tracking_id, 'Product Removed', properties, context=context)
+
+    def test_flush_with_product(self):
```
clintonb: Add docstrings explaining what these tests do.

clintonb (changes requested):
Ditch `_flush_segment_event`. Convenience methods that are too convenient end up doing us a disservice because the branching logic hides potential issues and/or makes them difficult to process mentally. Since there is little commonality amongst your two test cases, there is little need for a convenience method.

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -113,3 +118,39 @@ def test_create_basket(self):
...
+
+    def _flush_segment_event(self, empty):
+        """
+        Helper function for testing Segment event firing in basket.flush()
+        """
+        site = SiteConfigurationFactory().site
+        site.siteconfiguration.segment_key = 'fake_key'
+        user = factories.UserFactory()
+        basket = self._create_basket(user, site)
+        basket.flush()
```
mjfrey: I don't think you need this as new baskets are empty.

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -113,3 +118,39 @@ def test_create_basket(self):
         basket = Basket.create_basket(self.site1, user)
         self.assertEqual(basket.site, self.site1)
         self.assertEqual(basket.owner, user)
+
+    def _flush_segment_event(self, empty):
+        """
+        Helper function for testing Segment event firing in basket.flush()
+        """
+        site = SiteConfigurationFactory().site
```
mjfrey: you could use self.site1 rather than creating a new one.

McKenzieW:
@clintonb @mjfrey your requests have been addressed

mjfrey (approved):

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -113,3 +119,46 @@ def test_create_basket(self):
         basket = Basket.create_basket(self.site1, user)
         self.assertEqual(basket.site, self.site1)
         self.assertEqual(basket.owner, user)
+
+    def test_flush_with_product(self):
+        """
+        Verify the method fires 'Product Removed' Segment event with the correct information when basket is not empty
+        """
+        self.site1.siteconfiguration = SiteConfigurationFactory()
```
clintonb, line 127: Do you need to make a new instance? `self.site` already exists. If you aren't doing anything special with Site or SiteConfiguration, there's no need to make more of them.

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -113,3 +119,46 @@ def test_create_basket(self):
...
+        self.site1.siteconfiguration.segment_key = 'fake_key'
+        basket = create_basket(empty=True)
+        basket.owner = factories.UserFactory()
+        basket.site = self.site1
+        basket.save()
+
+        course = CourseFactory()
+        seat = course.create_or_update_seat('verified', True, 100, self.partner)
+        basket.add_product(seat)
+        basket.save()
```
clintonb: This call should not be needed.

ecommerce/extensions/basket/tests/test_models.py
```diff
@@ -113,3 +119,46 @@ def test_create_basket(self):
...
+            }
+        }
+
+        with mock.patch.object(Client, 'track') as mock_track:
+            basket.flush()
+            mock_track.assert_called_once_with(user_tracking_id, 'Product Removed', properties, context=context)
+
+    def test_flush_without_product(self):
+        """ Verify the method does not fireSegment event when basket is empty """
+        self.site1.siteconfiguration = SiteConfigurationFactory()
```
clintonb, line 153: Same comment as above.

clintonb (approved):
* Address the remaining comments
* Add the ticket number to your commit message. See the guidelines at https://openedx.atlassian.net/wiki/display/ECOM/Developer+Onboarding.

ecommerce/extensions/basket/tests/test_models.py
McKenzieW: sorry, which part of this should not be needed?

clintonb: This line to save the basket. The call to `add_product()` updates the m2m relations. There is no need to save after this.

McKenzieW: Ah ok, I didn't realize. Thank you

clintonb: You're welcome.

ecommerce/extensions/basket/tests/test_models.py
McKenzieW: There needs to be a site configuration for the segment event to work, and there doesn't already exist one for any of the site objects, so I need to add it.

clintonb: `self.site` is created by `SiteMixin`. It is already configured appropriately. See https://github.com/edx/ecommerce/blob/ac971e8b72c3f4fd4aca1a438d268286c85e507b/ecommerce/tests/mixins.py#L253.

McKenzieW: That is crazy because when I tried to use self.site I got an error saying that it had no siteconfiguration attribute

PR #1387, opened 2017-07-12
Updated custom Jasmine matchers

clintonb:
- Removed toHaveClass() since it is already provided by jasmine-jquery
- Renamed toBeVisible() to toNotHaveHiddenClass() since toBeVisible() is already provided by jasmine-jquery

LEARNER-1811

### DIFF
--- a/ecommerce/static/js/test/custom-matchers.js
+++ b/ecommerce/static/js/test/custom-matchers.js
@@ -3,25 +3,10 @@ define(['jquery'], function($) {
 
     beforeEach(function() {
         jasmine.addMatchers({
-            /**
-             * Returns a Boolean value that indicates whether an element is visible,
-             * which it can't be if it's not in the DOM or is hidden.
-             */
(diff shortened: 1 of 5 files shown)


ecommerce/static/js/test/custom-matchers.js
```diff
@@ -4,26 +4,18 @@ define(['jquery'], function($) {
     beforeEach(function() {
         jasmine.addMatchers({
             /**
-             * Returns a Boolean value that indicates whether an element is visible,
-             * which it can't be if it's not in the DOM or is hidden.
+             * Returns a Boolean value that indicates whether an element does NOT the hidden class.
```
vkaracic: `does not have the hidden class.`

vkaracic: Also, IIRC, this part `$(actual).length > 0` asserts whether the element is in the DOM or not.

PR #1388, opened 2017-07-12
Decreased Codecov threshold to 94%

clintonb:
The recent change reporting all JS files lead to an overall decrease in coverage. Dropping Codecov's threshold so that open PRs are not penalized for code that is not relevant to those PRs.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,4 +6,4 @@ coverage:
         target: 95
     project:
       default:
-        target: 95
+        target: 94


clintonb:
FYI @edx/learner

codecov.yml
```diff
@@ -6,4 +6,4 @@ coverage:
         target: 95
     project:
       default:
-        target: 95
+        target: 94
```
BenjiLee, line 9: Does patch on line 6 also have to be updated?

clintonb: Nope. New patches should still maintain 95% coverage or higher. It's only the overall project that has decreased coverage.

clintonb:
This is caused by https://github.com/edx/ecommerce/pull/1384.

rlucioni (approved):
😭

clintonb:
Don't cry. Update the JS tests!

PR #1390, opened 2017-07-13
Refactored e2e tests

clintonb:
The previous tests were broken and somewhat confusing to decipher. The tests are now run by pytest, rather than nose (which is no longer maintained). We are now interfacing directly with Selenium rather than going through bok-choy.

The broken tests have been removed. What remains are tests of the basic SSO integration and checkout. Usage of the course detail pages on the marketing site has been intentionally removed to (a) avoid false-negatives that result from issues with the marketing site and (b) keep the tests focused solely on the E-Commerce Service.

LEARNER-509


clintonb:
FYI @edx/learner @edx/testeng 

Our e2e tests atrophied due to lack of running/updating them. I decided to start over and focus solely on the most-important operation—payment—and solely on this service. I don't want to repeat Wednesday's frustrations that resulted from the reliance on the marketing site.

If we value coupons or bulk purchasing we should rewrite those tests in this repo and maintain them. If we value enrollment+logistration we should write tests for it in edx/edx-platform. If we value the interaction on the course details page of the marketing site, we should write tests for it in that repo.

e2e/test_auth.py
```diff
@@ -1,51 +1,52 @@
...
+    # Visit the Otto dashboard to trigger an OpenID Connect login
+    EcommerceHelpers.visit_dashboard(selenium)
+
+    LmsHelpers.logout(selenium)
+
+    # Now that the user has been logged out, navigating to the dashboard should result in the user being
+    # redirected to the identity provider's login page. This indicates the user has been logged out of both systems.
+    try:
+        # Decrease the wait time since we don't expect to find the element.
+        selenium.implicitly_wait(3)
```
rlucioni: Is this necessary? If you leave this, does it affect other `selenium` instances? My guess is no, just wondering.

e2e/conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+
+@pytest.fixture
+def selenium(selenium):
```
rlucioni: Nice use of a fixture.

rlucioni (approved):
@clintonb nice, thank you for taking this cleanup on.

e2e/test_auth.py
clintonb: Not necessary, and there's a valid argument that this is premature optimization. It shouldn't affect other `selenium` instances since they are recreated for each test run.

rlucioni: @clintonb cool. I do think it's premature. Your call.

jzoldak:
Thx for the heads up @clintonb

PR #1391, opened 2017-07-13
Refactoring code

iivic:
Separated this change from the initial PR for a cleaner view of the work.

LEARNER-1135


iivic:
@clintonb please review.

ecommerce/coupons/tests/mixins.py
```diff
@@ -22,6 +22,10 @@ def setUp(self):
         super(CourseCatalogMockMixin, self).setUp()
         cache.clear()
 
+    @staticmethod
+    def build_discovery_catalogs_url(discovery_api_url):
```
vkaracic: I would add a `catalog_id=''` kwarg here since you use that in almost all instances.

vkaracic (approved):

ecommerce/tests/factories.py
```diff
@@ -35,9 +36,7 @@ class Meta(object):
     send_refund_notifications = False
     enable_sdn_check = False
     enable_embargo_check = False
-    # TODO: Modify discovery_api_url to use Faker().url() and use
-    # mock_access_token_response instead of @mock_course_catalog_api_client decorator
-    discovery_api_url = factory.LazyAttribute(lambda obj: 'http://{domain}/api/v1/'.format(domain=obj.site.domain))
+    discovery_api_url = 'http://{}.fake/'.format(Faker().domain_name())
```
clintonb, line 40: Just fake the entire URL.

http://faker.readthedocs.io/en/master/providers/faker.providers.internet.html

iivic:
@clintonb just reminding you to review.

clintonb (approved):

clintonb:
Remember to squash your commits. Also, check out the guidelines for create an informative commit message: https://openedx.atlassian.net/wiki/display/OpenDev/Git+Best+Practices#GitBestPractices-CommitMessages.

ecommerce/tests/factories.py
iivic: I tried the first time and `fake.url()` sometimes creates a url with `https` which will cause the handshake to break in tests. Can't remember the exact issue but I decided that it is less time consuming to use this approach with the same result.

PR #1392, opened 2017-07-14
[WIP] Add management command to create a set of courses from json

stevenzheng3791:
LEARNER-1817
https://openedx.atlassian.net/browse/LEARNER-1817

This management command allows you to create a set of courses (e.g. course set containing all course types and valid course seat configurations) from properly-formatted json. Eventually, a provisioning script from docker devstack should be used to call this command, as opposed to using this command directly (It would be cumbersome to specify json directly on the command line)

### DIFF
--- a/ecommerce/core/management/commands/generate_courses.py
+++ b/ecommerce/core/management/commands/generate_courses.py
@@ -1,0 +1,136 @@
+"""
+Django management command to generate a test course for a given course id on LMS
+"""
+import datetime
+import json
+import logging
+
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+from django.utils import timezone
+from oscar.core.loading import get_model
+from waffle.models import Flag
+
+from ecommerce.courses.models import Course
+
+Partner = get_model('partner', 'Partner')
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+
+    help = 'Generate courses on ecommerce from a json list of courses. Should only be run in development environments!'
+
+    valid_seat_types = ["audit", "verified", "honor", "professional"]
+    default_verified_price = 100
+    default_professional_price = 1000
+    default_upgrade_deadline = timezone.now() + datetime.timedelta(days=365)
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'courses',
+            help='courses to create in JSON format'  # TODO - link test-course JSON format
+        )
+
+    def handle(self, *args, **options):
+        if not settings.DEBUG:
+            # DEBUG is turned on in development settings and off in production settings
+            raise CommandError("Command should only be run in development environments")
+        try:
+            logger.info("Attempting to create course from the following json object: %s", options["courses"])
+            arg = json.loads(options["courses"])
+        except ValueError:
+            raise CommandError("Invalid JSON object")
+        try:
+            courses = arg["courses"]
+        except KeyError:
+            raise CommandError("JSON object is missing courses field")
+
+        partner = Partner.objects.get(short_code='edx')
+        site = partner.siteconfiguration.site
+        Flag.objects.update_or_create(name='enable_client_side_checkout', defaults={'everyone': True})
+
+        for course_settings in courses:
+            if not self._course_is_valid(course_settings):
+                logger.warning("Can't create course, proceeding to next course")
+                continue
+
+            # Create the course
+            org = course_settings["organization"]
+            num = course_settings["number"]
+            run = course_settings["run"]
+            course_name = course_settings["fields"]["display_name"]
+            course_id = "course-v1:{org}+{num}+{run}".format(org=org, num=num, run=run)
+            defaults = {'name': course_name}
+            course, __ = Course.objects.update_or_create(id=course_id, site=site, defaults=defaults)
+            logger.info("Created course with id %s", course.id)
+
+            # Create seats
+            for seat in course_settings["seats"]:
+                self._create_seat(course, seat, partner)
+
+            # Publish the data to the LMS
+            course.publish_to_lms()
+
+    def _course_is_valid(self, course):
+        """ Returns true if the course is properly formatted and contains acceptable values """
+        is_valid = True
+
+        # Check course settings
+        missing_settings = []
+        if "organization" not in course:
+            missing_settings.append("organization")
+        if "number" not in course:
+            missing_settings.append("number")
+        if "run" not in course:
+            missing_settings.append("run")
+        if "fields" not in course:
+            missing_settings.append("fields")
+        if "seats" not in course:
+            missing_settings.append("seats")
+        if missing_settings:
+            logger.warning("Course json is missing the following fields: " + str(missing_settings))
+            is_valid = False
+
+        # Check fields settings
+        if ("fields" in course) and ("display_name" not in course["fields"]):
+            logger.warning("Fields json is missing display_name")
+            is_valid = False
+
+        # Check seats types
+        if "seats" in course:
+            for seat in course["seats"]:
+                if ("seat_type" not in seat) or (seat["seat_type"] not in self.valid_seat_types):
+                    logger.warning("Seat type must be one of " + str(self.valid_seat_types))
+                    is_valid = False
+                    break
+
+        return is_valid
+
+    def _create_seat(self, course, seat, partner):
+        """ Add the specified seat to the course """
+        seat_type = seat["seat_type"]
+        if seat_type == "audit":
+            course.create_or_update_seat("", False, 0, partner)
+        elif seat_type == "verified":
+            course.create_or_update_seat(
+                "verified",
+                True,
+                self.default_verified_price,
+                partner,
+                expires=self.default_upgrade_deadline
+            )
+        elif seat_type == "honor":
+            course.create_or_update_seat("honor", False, 0, partner)
+        elif seat_type == "professional":
+            if "id_verification_required" in seat:
+                id_verification_required = seat["id_verification_required"]
+            else:
+                id_verification_required = True
+            course.create_or_update_seat(
+                "professional",
+                id_verification_required,
+                self.default_professional_price,
+                partner
+            )
+        logger.info("Created %s seat for course %s", seat_type, course.id)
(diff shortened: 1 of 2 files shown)


ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,130 @@
...
+
+        try:
+            courses = arg["courses"]
+        except KeyError:
+            raise CommandError("JSON object is missing courses field")
+
+        partner = Partner.objects.get(short_code='edx')
+        site = partner.siteconfiguration.site
+        # Upgrade deadline must occur before verification deadline
+        # verification_deadline = timezone.now() + datetime.timedelta(days=366)
```
schenedx: #nitpick This commented out line should be removed?

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,130 @@
...
+        upgrade_deadline = timezone.now() + datetime.timedelta(days=365)
+        # Enable client-side checkout
+        # TODO Use the management command in edx/devstack once https://github.com/jsocol/django-waffle/issues/199
+        # is resolved.
+        Flag.objects.update_or_create(name='enable_client_side_checkout', defaults={'everyone': True})
+
+        for course_settings in courses:
+            # Create the course from course settings
+            try:
+                store = course_settings["store"]
```
schenedx: Can we be explicit for the variable? Just say `module_store` instead of `store`?

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,130 @@
...
+                        org=org, num=num, run=run
+                    )
+                elif store == "mongo":
+                    course_id = "course-v1:{org}/{num}/{run}".format(
+                        org=org, num=num, run=run
+                    )
+                else:
+                    raise CommandError('Modulestore {store} is invalid'.format(store=store))
+            except KeyError:
+                raise CommandError("Course JSON object is missing required id fields")
```
schenedx: Wouldn't this stop the command? Should we do that? Should we continue the command so we can consume other courses specified in the json?

ecommerce/core/tests/test_generate_courses.py
```diff
@@ -0,0 +1,207 @@
...
+                '"store":"split",' +
+                '"organization":"test-course-generator"}]}'
+            )
+            call_command("generate_courses", arg)
+
+    def test_invalid_store(self):
+        """
+        Test that providing an invalid store option will result in the appropriate command error
+        """
+        error_msg = "Modulestore invalid_store is invalid"
```
schenedx: This is a very confusing error message. Can we make this message better?

ecommerce/core/tests/test_generate_courses.py
```diff
@@ -0,0 +1,207 @@
...
+                '"organization":"test-course-generator",' +
+                '"number":"1",' +
+                '"run":"1",' +
+                '"seats":[{"seat_type":"invalid_seat_type"}],' +
+                '"fields":{"display_name":"test-course"}}]}'
+            )
+            call_command("generate_courses", arg)
+
+    @httpretty.activate
+    def test_create_audit_seat(self):
```
schenedx: From this point in the file, the test cases pattern is quite repetitive. Can we refactor so we can use ddt to make this a single test with different test inputs?

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,130 @@
...
+from waffle.models import Flag
+
+from ecommerce.courses.models import Course
+
+Partner = get_model('partner', 'Partner')
+
+
+class Command(BaseCommand):
+
+    help = 'Generate courses on ecommerce from a json list of courses. Should only be run in development environments!'
```
schenedx, line 22: Should we implement this command so if this is run in the production environment, throws an exception? You should be able to determine whether this is in production environment through settings. I don't know exactly which setting tho. Can you find out?

schenedx (changes requested):
Please see my feedback inline.

Also, how would this management command be used? I see all the courses are passed in as a large string of json format. How is that being passed in?

stevenzheng3791:
Eventually, there will be a test course provisioning tool on docker devstack for developers to use to that will have different ways to auto-generate course setting permutations (in json) and pass that to this command in ecommerce, as well as an equivalent command in studio. So, it will not be ideal to use this command directly.

ecommerce/core/management/commands/generate_courses.py
stevenzheng3791: Yes, thank you for point this out. I've updated it to only raise exception if the preliminary checks fail.

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,139 @@
...
+            try:
+                module_store = course_settings["store"]
+                org = course_settings["organization"]
+                num = course_settings["number"]
+                run = course_settings["run"]
+                course_name = course_settings["fields"]["display_name"]
+                if module_store == "split":
+                    course_id = "course-v1:{org}+{num}+{run}".format(org=org, num=num, run=run)
+                elif module_store == "mongo":
+                    course_id = "course-v1:{org}/{num}/{run}".format(org=org, num=num, run=run)
```
BenjiLee: Mongo course_id's do not have `course-v1:` prepended. They should look like `{org}/{num}/{run}`

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,139 @@
...
+                    continue
+            except KeyError:
+                self._abort_course_generation(None, "Course JSON object is missing required id fields")
+                continue
+
+            # Create the course
+            defaults = {'name': course_name}
+            course, __ = Course.objects.update_or_create(id=course_id, site=site, defaults=defaults)
+            msg = "Created course with id %s" % (course.id)
+            logger.info(msg)
```
BenjiLee: nit: You could probably fit this all in one line.

`logger.info("Created course with id {}".format(course.id))`

More about python string formatting: https://pyformat.info/

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,139 @@
...
+
+            # Create seats
+            seats = course_settings["seats"]
+            seat_types = [seat["seat_type"] for seat in seats]
+            if not set(seat_types).issubset(self.valid_seat_types):
```
BenjiLee: It might be better to do this verification before even attempting to create a course. That way, we won't have to worry about the clean up.

If there is any validation  that can be done before the actual creation of the course, that would be preferred.

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,139 @@
...
+        elif seat_type == "professional":
+            # TODO - Enable verification deadline if developers want it
+            course.create_or_update_seat(
+                "professional",
+                seat["id_verification_required"],
+                self.default_professional_price,
+                partner
+            )
+        elif seat_type == "credit":
+            # Requires manually creating TestCreditProvider in the LMS django admin
```
BenjiLee: Should this raise an exception if `TestCreditProvider` is not set?

schenedx (commented):
Looks better. I agree with comments made by @BenjiLee

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,139 @@
...
+        try:
+            arg = json.loads(options["courses"])
+        except ValueError:
+            raise CommandError("Invalid JSON object")
+        try:
+            courses = arg["courses"]
+        except KeyError:
+            raise CommandError("JSON object is missing courses field")
+
+        partner = Partner.objects.get(short_code='edx')
```
BenjiLee, line 49: Do partner have to be hardcoded?

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'courses',
+            help='courses to create in JSON format'  # TODO - link test-course JSON format
+        )
+
+    def handle(self, *args, **options):
+        # DEBUG is true in development environments and false in production
+        if not settings.DEBUG:
```
BenjiLee: Could you describe whats `settings.DEBUG` is in a github comment?

BenjiLee:
@stevenzheng3791 Done with first pass

ecommerce/core/management/commands/generate_courses.py
stevenzheng3791: I've decided to get rid of this case for now. It isn't very functional right now without provisioning a credit provider before hand. In the future, if we find out we need credit courses, then we can figure out the best way to go about handling it.

ecommerce/core/management/commands/generate_courses.py
stevenzheng3791: Right now, it is because partner and site for 'edx' are already provisioned in docker devstack. Unless it turns out this is one of the settings developers want to be able to manipulate (would require manually provisioning a partner/site right now), I think this is the simplest way to allow course creation

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,133 @@
...
+            logger.warning("Missing setting in course json")
+            return False
+        for seat in course["seats"]:
+            if seat["seat_type"] not in self.valid_seat_types:
+                logger.warning("Seat type must be one of " + str(self.valid_seat_types))
+                return False
+            elif (seat["seat_type"] == "professional") and ("id_verification_required" not in seat):
+                logger.warning("Missing setting in seat json")
+                return False
+        return True
```
BenjiLee: We should log all the missing fields in one run. If a user is missing all of the fields, they will only get the error message for one.

BenjiLee (approved):

ecommerce/core/management/commands/generate_courses.py
```diff
@@ -0,0 +1,136 @@
...
+            'courses',
+            help='courses to create in JSON format'  # TODO - link test-course JSON format
+        )
+
+    def handle(self, *args, **options):
+        if not settings.DEBUG:
+            # DEBUG is turned on in development settings and off in production settings
+            raise CommandError("Command should only be run in development environments")
+        try:
+            print options["courses"]
```
schenedx: Should we use `print` or use `logger` here?

schenedx (approved):
Just one place I believe we should update.
rest LGTM

PR #1396, opened 2017-07-17
ENT-367 Set max_age on Enterprise cookie.

douglashall:
We set a cookie which identifies a user as being associated with
a given enterprise based on how they navigated to the Open edX site.
This cookie gets shared across edX services, so that the enterprise
context can be maintained when navigating between UIs across these
services. This cookie needs to be short-lived so that the Enterprise
context does not linger longer than it should. This change sets a
60 second max age on the Enterprise cookie.

ENT-367

### DIFF
--- a/ecommerce/enterprise/decorators.py
+++ b/ecommerce/enterprise/decorators.py
@@ -16,15 +16,23 @@ def set_enterprise_cookie(func):
     """
     @wraps(func)
     def _decorated(request, *args, **kwargs):
-        code, enterprise_customer_uuid = request.GET.get('code'), None
-        if code:
-            enterprise_customer_uuid = utils.get_enterprise_customer_uuid(code)
-
         response = func(request, *args, **kwargs)
 
         # Set enterprise customer cookie if enterprise customer uuid is available.
-        if enterprise_customer_uuid:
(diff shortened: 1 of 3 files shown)


saleem-latif (approved):
LGTM 👍 , I have also verified it on my devstack and it works as expected.

ecommerce/enterprise/decorators.py
```diff
@@ -16,15 +16,23 @@ def set_enterprise_cookie(func):
     """
     @wraps(func)
     def _decorated(request, *args, **kwargs):
-        code, enterprise_customer_uuid = request.GET.get('code'), None
-        if code:
-            enterprise_customer_uuid = utils.get_enterprise_customer_uuid(code)
-
```
itsjeyd, line 22: @douglashall Moving this code below the call to `func` is just a clean up, right?

itsjeyd (approved):
@douglashall Thanks for following up on ENT-367! I had one question, but this LGTM 👍

ecommerce/enterprise/decorators.py
douglashall: @itsjeyd Correct. No need to look for the enterprise customer until after the wrapped function is called.

PR #1399, opened 2017-07-19
Enabling CyberSource Export Compliance for Apple Pay

clintonb:
Export Compliance checks the SDN and other lists to ensure we reject payments from certain individuals and locations.

LEARNER-1898


clintonb:
FYI @mjfrey

PR #1400, opened 2017-07-19
Updated Order Completed segment event to include Coupon data

Mjloturco:
Segment events were not tracking coupon or discount values, events fired now have include a coupon field containing the name of the coupon and a discount field containing the value of the discount they have received.

 [LEARNER-1918] cloned from [LEARNER-1568]

### DIFF
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -23,10 +23,15 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
     if order.total_excl_tax <= 0:
         return
 
+    voucher = order.basket_discounts.filter(voucher_id__isnull=False).first()
+    coupon = voucher.voucher_code if voucher else None
+
     properties = {
         'orderId': order.number,
         'total': str(order.total_excl_tax),
         'currency': order.currency,
+        'coupon': coupon,
+        'discount': str(order.total_discount_incl_tax),
         'products': [
             {
                 # For backwards-compatibility with older events the `sku` field is (ab)used to
--- a/ecommerce/extensions/checkout/tests/test_signals.py
+++ b/ecommerce/extensions/checkout/tests/test_signals.py
@@ -3,6 +3,8 @@ import json
 import httpretty
 import mock
 from django.core import mail
+from oscar.core.loading import get_class, get_model
+from oscar.test import factories
 from oscar.test.newfactories import BasketFactory
 from testfixtures import LogCapture
 
@@ -12,8 +14,14 @@ from ecommerce.courses.utils import mode_for_seat
 from ecommerce.extensions.catalogue.tests.mixins import CourseCatalogTestMixin
 from ecommerce.extensions.checkout.signals import send_course_purchase_email, track_completed_order
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
-from ecommerce.extensions.test.factories import create_order
+from ecommerce.extensions.test.factories import create_order, prepare_voucher
+from ecommerce.tests.factories import ProductFactory
 from ecommerce.tests.testcases import TestCase
+
+Applicator = get_class('offer.utils', 'Applicator')
+Benefit = get_model('offer', 'Benefit')
+Condition = get_model('offer', 'Condition')
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
 
 LOGGER_NAME = 'ecommerce.extensions.checkout.signals'
 
@@ -111,10 +119,14 @@ class SignalTests(CourseCatalogTestMixin, TestCase):
             )
 
     def _generate_event_properties(self, order):
+        voucher = order.basket_discounts.filter(voucher_id__isnull=False).first()
+        coupon = voucher.voucher_code if voucher else None
         return {
             'orderId': order.number,
             'total': str(order.total_excl_tax),
             'currency': order.currency,
+            'coupon': coupon,
+            'discount': str(order.total_discount_incl_tax),
             'products': [
                 {
                     'id': line.partner_sku,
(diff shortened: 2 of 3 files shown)


ecommerce/extensions/checkout/signals.py
```diff
@@ -23,10 +23,18 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
     if order.total_excl_tax <= 0:
         return
 
+    coupon_name = None
+    discount_value = 0
+    for discount in order.discounts.all():
+        coupon_name = discount.voucher.name
```
mjfrey: A few things here.  There may be a discount that does not have a voucher.
For example, we have program offers which are discounts that the learner does not have to enter a code at checkout time.  

You can get those via:
discount.offer

I would suggest we send a list of vouchers, offers and total all the discounts together.

e2e/test_auth.py
```diff
@@ -1,4 +1,4 @@
-import pytest
+import pytest  # isort:skip
```
clintonb: Revert. General rule of thumb: if merged code has a quality issue, the issue probably lies with your machine, not the merged code. In this case, you probably haven't installed Pytest. Run `make requirements` to install the e2e test requirements.

ecommerce/extensions/checkout/signals.py
```diff
@@ -23,10 +23,27 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
     if order.total_excl_tax <= 0:
         return
 
+    # coupon_name = None
```
clintonb: No dead code in the repo.

ecommerce/extensions/checkout/signals.py
```diff
@@ -23,10 +23,27 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
...
+            discount_names.append(discount.voucher.name)
+        elif discount.offer:
+            discount_names.append(discount.offer.name)
+        discount_total += discount.amount
+
     properties = {
         'orderId': order.number,
         'total': str(order.total_excl_tax),
         'currency': order.currency,
+        'discount_names': discount_names,
```
clintonb: Where did these field names come from? They are not listed at https://segment.com/docs/spec/ecommerce/v2/#order-completed.

clintonb (changes requested):
These are not valid properties. See the documentation at https://segment.com/docs/spec/ecommerce/v2/#order-completed.

e2e/test_auth.py
Mjloturco: Got it, will revert, thanks

clintonb:
@Mjloturco I'm going to let you do the digging, instead of telling you exactly what to change, to help you learn more about Otto and Oscar. Here are a few hints, however.

1. We only allow one coupon/voucher per order.
2. We may have multiple offers per order (but only one will be voucher-based).
3. The `Order` model has everything you need.

Mjloturco:
@clintonb thanks for the hints, I'll make sure to stick with actual segment event properties. Are we interested in tracking the names of applied offers in some way as well?  The docs specify the "discount" property is a total discount, are we worried this may appear as if the entire discount is due to the one voucher?

Example case: order pre-offer total is $100, a voucher "examplevoucher" discounts $10, and a separate offer discounts $10 as well. Discount would be 20, despite the coupon only accounting for half of that.

Mjloturco:
@clintonb I made the changes you requested.

ecommerce/extensions/checkout/signals.py
```diff
@@ -23,10 +23,21 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
...
+        discount_value += discount.amount
+    # Note discount_value is the sum of all offer discounts, and so may
+    # be greater than the discount associated with only the voucher.
+
     properties = {
         'orderId': order.number,
         'total': str(order.total_excl_tax),
         'currency': order.currency,
+        'coupon': coupon,
+        'discount': str(discount_value),
```
clintonb: See https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/order/abstract_models.py#L201.

```python
str(order. total_discount_incl_tax)
```

ecommerce/extensions/checkout/signals.py
```diff
@@ -23,10 +23,21 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
     if order.total_excl_tax <= 0:
         return
 
+    coupon = None
+    discount_value = 0
+    for discount in order.discounts.all():
```
clintonb: There is no need to iterate over all discounts. You can filter in the database.

```python
voucher = order.basket_discounts.filter(voucher_id__isnull=False).first()
coupon = voucher.voucher_code if voucher else None
```

clintonb (changes requested):

ecommerce/extensions/checkout/signals.py
Mjloturco: Fixed

ecommerce/extensions/checkout/signals.py
Mjloturco: Fixed

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -111,10 +119,14 @@ def test_post_checkout_callback_no_credit_provider(self):
             )
 
     def _generate_event_properties(self, order):
+        voucher = order.basket_discounts.filter(voucher_id__isnull=False).first()
```
clintonb, line 122: Pass the voucher in as an argument. You want to avoid using the code being tested in your test in case that code is wrong.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -142,3 +154,62 @@ def test_track_completed_order(self):
...
+            basket.add_product(product)
+            basket.vouchers.add(voucher)
+            Applicator().apply(basket, user=basket.owner, request=self.request)
+
+            order = factories.create_order(basket=basket, user=self.user)
+            track_completed_order(None, order)
+            properties = self._generate_event_properties(order)
+            mock_track.assert_called_once_with(order.site, order.user, 'Order Completed', properties)
+
+    def test_track_completed_fixed_discounted_order(self):
```
clintonb, line 177: The voucher type shouldn't matter. I see no need for a separate test case.

clintonb: I would convert this to a test with both a voucher and another offer (without a voucher).

clintonb (approved):
Fix the tests, and you're good to go.

Mjloturco:
Just realized it was approved only with changes, making a new pr right now to fix.

PR #1401, opened 2017-07-19
Updated Program Conditional Offer logic to allow for partial programs

McKenzieW:
ProgramCourseRunSeatsCondition model's is_satisfied method checks enrollments and items in basket  for condition satisfaction, allowing for discount application on remaining courses in a program when a user is already enrolled in some of the program courses.

**Test**
Add two course products in ecommerce.
Create corresponding courses with verified seats in CAT.
Create a program offer with these two courses.
Enroll a user in a verified seat of one course.
Add the verified seat for the other course to the basket.
Discount should appear for the one course.

LEARNER-1901

### DIFF
--- a/ecommerce/courses/utils.py
+++ b/ecommerce/courses/utils.py
@@ -1,10 +1,13 @@
 import hashlib
+import logging
 
 from django.conf import settings
 from django.core.cache import cache
 from django.utils.translation import ugettext_lazy as _
 
 from ecommerce.core.utils import traverse_pagination
+
+logger = logging.getLogger(__name__)
 
 
 def mode_for_seat(product):
--- a/ecommerce/programs/conditions.py
+++ b/ecommerce/programs/conditions.py
@@ -1,10 +1,11 @@
+from __future__ import unicode_literals
+
 import logging
 import operator
 
-from oscar.apps.offer import utils
+import requests
+from oscar.apps.offer import utils as oscar_utils
 from oscar.core.loading import get_model
-
-from requests import Timeout
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
 
 from ecommerce.programs.api import ProgramsApiClient
@@ -53,21 +54,37 @@ class ProgramCourseRunSeatsCondition(Condition):
         return program_course_run_skus
 
     def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
+        """
+        Determines if a user is eligible for a program offer based on products in their basket
+        and their existing course enrollments.
+
+        Args:
+            basket : contains information on line items for order, associated siteconfiguration
+                        for retrieving program details, and associated user for retrieving enrollments
+        Returns:
+            bool
+        """
+
         if basket.is_empty:
             return False
 
         basket_skus = set([line.stockrecord.partner_sku for line in basket.all_lines()])
-
         try:
             program = self.get_program(basket.site.siteconfiguration)
-        except (HttpNotFoundError, SlumberBaseException, Timeout):
+        except (HttpNotFoundError, SlumberBaseException, requests.Timeout):
             return False
 
         applicable_seat_types = program['applicable_seat_types']
+        enrollments = basket.site.siteconfiguration.enrollment_api_client.enrollment.get(user=basket.owner.username)
 
         for course in program['courses']:
-            # If the basket has no SKUs, but we still have courses over which to iterate,
-            # the basket cannot meet the condition that all courses be represented.
+            # If the user is already enrolled in a course, we do not need to check their basket for it
+            if any(course['key'] in enrollment['course_details']['course_id'] and
+                   enrollment['mode'] in applicable_seat_types for enrollment in enrollments):
+                continue
+
+            # If the  basket has no SKUs left, but we still have courses over which
+            # to iterate, the user cannot meet the condition that all courses be represented.
             if not basket_skus:
                 return False
 
@@ -76,9 +93,9 @@ class ProgramCourseRunSeatsCondition(Condition):
             for course_run in course['course_runs']:
                 skus.update(set([seat['sku'] for seat in course_run['seats'] if seat['type'] in applicable_seat_types]))
 
-            # The lack of a difference in the set of SKUs in the basket and the course indicates
-            # that there is no intersection. Therefore, the basket contains no SKUs for the current
-            # course. It follows that the program condition is not met.
+            # The lack of a difference in the set of SKUs in the basket and the course indicates that
+            # that there is no intersection. Therefore, the basket contains no SKUs for the current course.
+            # Because the user is also not enrolled in the course, it follows that the program condition is not met.
             diff = basket_skus.difference(skus)
             if diff == basket_skus:
                 return False
@@ -107,7 +124,7 @@ class ProgramCourseRunSeatsCondition(Condition):
             if not self.can_apply_condition(line):
                 continue
 
-            price = utils.unit_price(offer, line)
+            price = oscar_utils.unit_price(offer, line)
             if not price:
                 continue
             line_tuples.append((price, line))
--- a/ecommerce/programs/tests/mixins.py
+++ b/ecommerce/programs/tests/mixins.py
@@ -3,6 +3,7 @@ from decimal import Decimal
 
 import httpretty
 
+from ecommerce.core.url_utils import get_lms_enrollment_api_url
 from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.courses.utils import mode_for_seat
 from ecommerce.extensions.catalogue.tests.mixins import DiscoveryTestMixin
@@ -18,9 +19,9 @@ class ProgramTestMixin(DiscoveryTestMixin):
             dict: Mocked program data.
         """
         courses = []
-        for __ in range(1, 5):
+        for i in range(1, 5):
+            key = 'course-v1:test-org+course+' + str(i)
             course_runs = []
-
             for __ in range(1, 4):
                 course_run = CourseFactory()
                 course_run.create_or_update_seat('audit', False, Decimal(0), self.partner)
@@ -34,7 +35,7 @@ class ProgramTestMixin(DiscoveryTestMixin):
                     } for seat in course_run.seat_products]
                 })
 
-            courses.append({'course_runs': course_runs, })
+            courses.append({'key': key, 'course_runs': course_runs, })
 
         program_uuid = str(program_uuid)
         data = {
@@ -56,3 +57,17 @@ class ProgramTestMixin(DiscoveryTestMixin):
             content_type='application/json'
         )
         return data
+
+    def mock_enrollment_api(self, username, enrollments=[]):  # pylint: disable=dangerous-default-value
+        """ Mocks enrollment retrieval from LMS
+        Returns:
+            list: Mocked enrollment data
+        """
+        self.mock_access_token_response()
+        httpretty.register_uri(
+            method=httpretty.GET,
+            uri='{}?user={}'.format(get_lms_enrollment_api_url(), username),
+            body=json.dumps(enrollments),
+            content_type='application/json'
+        )
+        return enrollments
--- a/ecommerce/programs/tests/test_conditions.py
+++ b/ecommerce/programs/tests/test_conditions.py
@@ -13,6 +13,7 @@ from ecommerce.tests.factories import ProductFactory
 from ecommerce.tests.testcases import TestCase
 
 Product = get_model('catalogue', 'Product')
+LOGGER_NAME = 'ecommerce.programs.conditions'
 
 
 @ddt.ddt
@@ -42,7 +43,7 @@ class ProgramCourseRunSeatsConditionTests(ProgramTestMixin, TestCase):
         self.assertEqual(self.condition.get_program(self.site.siteconfiguration), data)
 
     @httpretty.activate
-    def test_is_satisfied(self):
+    def test_is_satisfied_no_enrollments(self):
         """ The method should return True if the basket contains one course run seat corresponding to each
         course in the program. """
         offer = factories.ProgramOfferFactory(condition=self.condition)
@@ -63,6 +64,7 @@ class ProgramCourseRunSeatsConditionTests(ProgramTestMixin, TestCase):
                 else:
                     audit_seats.append(seat)
 
+        self.mock_enrollment_api(basket.owner.username)
         # Empty baskets should never be satisfied
         basket.flush()
         self.assertTrue(basket.is_empty)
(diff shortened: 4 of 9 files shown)


ecommerce/programs/conditions.py
```diff
@@ -65,7 +66,18 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
 
         applicable_seat_types = program['applicable_seat_types']
 
+        enrollments = get_enrollments(basket.owner)
+        verified_enrollments = set()
```
mjfrey: Take a look below.  You will need to check if any of the modes are in the applicable_seat_types.  Not just verified.

McKenzieW:
@edx/helio please review

ecommerce/programs/conditions.py
```diff
@@ -52,22 +54,56 @@ def get_applicable_skus(self, site_configuration):
 
         return program_course_run_skus
 
+    def get_enrollments(self, username):
+        """ Retrieve the enrollments for the specified User """
```
marjev: Can you add args and return value to the docstring the same way it's done in e. g. `get_program` method found above?

ecommerce/programs/conditions.py
```diff
@@ -52,22 +54,56 @@ def get_applicable_skus(self, site_configuration):
...
+
+            response = requests.get(url, headers=headers, timeout=timeout)
+            status_code = response.status_code
+            if status_code == 200:
+                return json.loads(response.json())
+            else:
+                logger.warning(u'Failed to retrieve enrollments for [%s]. Enrollment API returned status code [%d].',
+                               username, status_code)
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(u'An unexpected error occurred while retrieving enrollments for [%s].', username)
```
marjev: You can import `from __future__ import unicode_literals` at the beginning of the file to avoid having explicit unicode string literals.

ecommerce/programs/conditions.py
```diff
@@ -52,22 +54,56 @@ def get_applicable_skus(self, site_configuration):
...
+        return []
+
+    def enrolled(self, enrollments, course, applicable_seat_types):
+        """ Determine whether or not the user is enrolled in a given course """
```
marjev: Please extend the docstring with descriptions of args and return value.

ecommerce/programs/conditions.py
```diff
@@ -52,22 +54,56 @@ def get_applicable_skus(self, site_configuration):
...
+            if status_code == 200:
+                return json.loads(response.json())
+            else:
+                logger.warning(u'Failed to retrieve enrollments for [%s]. Enrollment API returned status code [%d].',
+                               username, status_code)
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(u'An unexpected error occurred while retrieving enrollments for [%s].', username)
+        return []
+
+    def enrolled(self, enrollments, course, applicable_seat_types):
```
marjev: Can you please rename this and give it a more descriptive name? I suggest `determine_user_enrollment` since it's a method and I like to name methods as actions. It's only a suggestion and perhaps you can think of a more descriptive name.

ecommerce/programs/conditions.py
```diff
@@ -52,22 +54,56 @@ def get_applicable_skus(self, site_configuration):
...
+        """ Determine whether or not the user is enrolled in a given course """
+        for enrollment in enrollments:
+            if course['uuid'] == enrollment['course_details']['course_id']:
+                if enrollment['mode'] in applicable_seat_types:
+                    return True
+        return False
```
marjev: If you want, you can use Python's any built-in function like this:
```
return any(
    course['uuid'] == enrollment['course_details']['course_id'] and enrollment['mode'] in applicable_seat_types 
    for enrollment in enrollments
)
```

ecommerce/programs/tests/mixins.py
```diff
@@ -18,23 +18,33 @@ def mock_program_detail_endpoint(self, program_uuid, discovery_api_url):
             dict: Mocked program data.
         """
         courses = []
-        for __ in range(1, 5):
+        for i in range(1, 5):
+            course_id = i
```
marjev: I'm not sure I understand why this is needed. Why's the counter value assigned as the seat ID?

marjev (changes requested):
Mostly suggestions. Let me know if you need any help or clarification of some of my comments.

ecommerce/extensions/offer/models.py
```diff
@@ -320,5 +320,4 @@ def all_products(self):
 class Condition(AbstractCondition):
     program_uuid = models.UUIDField(null=True, blank=True, verbose_name=_('Program UUID'))
 
-
```
clintonb: Revert.

ecommerce/programs/conditions.py
```diff
@@ -52,22 +54,56 @@ def get_applicable_skus(self, site_configuration):
 
         return program_course_run_skus
 
+    def get_enrollments(self, username):
+        """ Retrieve the enrollments for the specified User """
+        try:
+            url = '{}?user={}'.format(get_lms_enrollment_api_url(), username)
```
clintonb: Create an `enrollment_api_client` property on the `SiteConfiguration` model and use it, similar to how we call other APIs. Also, use JWT authentication.

ecommerce/programs/conditions.py
```diff
@@ -52,22 +54,56 @@ def get_applicable_skus(self, site_configuration):
...
+
+            headers = {
+                'Content-Type': 'application/json',
+                'X-Edx-Api-Key': settings.EDX_API_KEY
+            }
+
+            response = requests.get(url, headers=headers, timeout=timeout)
+            status_code = response.status_code
+            if status_code == 200:
+                return json.loads(response.json())
```
clintonb: `response.json()` will give you a Python object. There is no need to call `json.loads()`. In fact, this line will fail. (Make sure you've manually tested this locally.)

ecommerce/programs/conditions.py
```diff
@@ -52,22 +54,56 @@ def get_applicable_skus(self, site_configuration):
...
+                logger.warning(u'Failed to retrieve enrollments for [%s]. Enrollment API returned status code [%d].',
+                               username, status_code)
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(u'An unexpected error occurred while retrieving enrollments for [%s].', username)
+        return []
+
+    def enrolled(self, enrollments, course, applicable_seat_types):
+        """ Determine whether or not the user is enrolled in a given course """
+        for enrollment in enrollments:
+            if course['uuid'] == enrollment['course_details']['course_id']:
```
clintonb: The course's UUID will never be equivalent to a course ID. The simplest check is to determine if `enrollment['course_details']['course_id'].startswith(course['key'])`.

ecommerce/programs/conditions.py
```diff
@@ -52,22 +54,56 @@ def get_applicable_skus(self, site_configuration):
...
+                if enrollment['mode'] in applicable_seat_types:
```
clintonb: Combine with the above if statement.

ecommerce/programs/conditions.py
clintonb: Is this endpoint paginated? You may need to account for that.

ecommerce/programs/tests/mixins.py
```diff
@@ -18,23 +18,33 @@ def mock_program_detail_endpoint(self, program_uuid, discovery_api_url):
...
             course_runs = []
 
             for __ in range(1, 4):
                 course_run = CourseFactory()
-                course_run.create_or_update_seat('audit', False, Decimal(0), self.partner)
                 course_run.create_or_update_seat('verified', True, Decimal(100), self.partner)
+                course_run.create_or_update_seat('audit', False, Decimal(0), self.partner)
```
clintonb: This change is unnecessary. Revert.

ecommerce/programs/tests/mixins.py
```diff
@@ -18,23 +18,33 @@ def mock_program_detail_endpoint(self, program_uuid, discovery_api_url):
...
-                    'key': course_run.id,
-                    'seats': [{
+                seats = []
+                for seat in course_run.seat_products:
+                    if mode_for_seat(seat) == 'verified':
+                        seat_id = course_id
+                    else:
+                        seat_id = 2
+                    seats.append({
+                        'id': seat_id,
```
clintonb: Our code doesn't care about seat IDs. The API doesn't even expose them. This needs to go away.

ecommerce/programs/tests/mixins.py
```diff
@@ -18,23 +18,33 @@ def mock_program_detail_endpoint(self, program_uuid, discovery_api_url):
...
-                    } for seat in course_run.seat_products]
+                    })
+
+                course_runs.append({
+                    'key': course_run.id,
+                    'seats': seats
                 })
 
-            courses.append({'course_runs': course_runs, })
+            courses.append({'uuid': course_id, 'course_runs': course_runs, })
```
clintonb: I'm not sure what's going on here in this file. The only change needed to the data is the addition of a `'key'` field on this line. The rest is unnecessary.

clintonb: See the API for the exposed fields: https://prod-edx-discovery.edx.org/api/v1/programs/3fa369a7-4055-4f6d-8e31-b5b50c19ac73/.

ecommerce/programs/tests/test_conditions.py
```diff
@@ -20,7 +21,7 @@ class ProgramCourseRunSeatsConditionTests(ProgramTestMixin, TestCase):
     def setUp(self):
         super(ProgramCourseRunSeatsConditionTests, self).setUp()
         self.condition = factories.ProgramCourseRunSeatsConditionFactory()
-        self.test_product = ProductFactory()
+        self.test_product = ProductFactory(categories=[])
```
clintonb: This is not necessary. If your tests are dependent on product categories something is wrong.

ecommerce/programs/tests/test_conditions.py
```diff
@@ -39,7 +40,18 @@ def test_get_program(self):
         self.assertEqual(self.condition.get_program(self.site.siteconfiguration), data)
 
     @httpretty.activate
-    def test_is_satisfied(self):
+    def test_get_enrollments(self):
+        """ The method should return an empty list if their is an error when trying to retrieve enrollments """
```
clintonb: s/their/there

ecommerce/programs/tests/test_conditions.py
```diff
@@ -39,7 +40,18 @@ def test_get_program(self):
...
+        user = factories.UserFactory()
+        httpretty.register_uri(
+            method=httpretty.GET,
```
clintonb: Nit: these are the first two arguments. There is no need for the `method=` and `uri=` components.

ecommerce/programs/tests/test_conditions.py
```diff
@@ -60,27 +72,63 @@ def test_is_satisfied(self):
                 else:
                     audit_seats.append(seat)
 
-        # Empty baskets should never be satisfied
-        basket.flush()
-        self.assertTrue(basket.is_empty)
-        self.assertFalse(self.condition.is_satisfied(offer, basket))
+        with mock.patch('ecommerce.programs.conditions.ProgramCourseRunSeatsCondition.get_enrollments',
```
clintonb: Nit: Make this a decorator over the entire method.

ecommerce/programs/tests/test_conditions.py
```diff
@@ -60,27 +72,63 @@ def test_is_satisfied(self):
...
+            # needed for the program, the condition should NOT be satisfied
+            basket.flush()
+            for verified_seat in verified_seats[1:len(verified_seats)]:
+                basket.add_product(verified_seat)
+            self.assertFalse(self.condition.is_satisfied(offer, basket))
+
+    @httpretty.activate
+    def test_is_satisfied_with_enrollments(self):
+        """ The condition should be satisfied if one valid course run from each course is in either the
+        basket or the user's enrolled courses """
```
clintonb: Add a period at the end of the sentence.

ecommerce/programs/tests/test_conditions.py
```diff
@@ -60,27 +72,63 @@ def test_is_satisfied(self):
...
+        offer = factories.ProgramOfferFactory(condition=self.condition)
+        basket = factories.BasketFactory(site=self.site, owner=factories.UserFactory())
+        program = self.mock_program_detail_endpoint(
+            self.condition.program_uuid, self.site_configuration.discovery_api_url
+        )
+        enrollments = [{'mode': 'verified', 'course_details': {'course_id': 1}},
```
clintonb: These are not valid course keys. See `CourseFactory` for an example generator.

clintonb (changes requested):

ecommerce/programs/tests/test_conditions.py
McKenzieW: Sorry, I'm not sure what you mean by this.

clintonb:
```python
uri = '{}?user={}'.format(get_lms_enrollment_api_url(), user.username)
httpretty.register_uri(httpretty.GET, uri, status=404)
```

You don't need to specify `method=` or `uri=` because they are positional arguments. See https://docs.python.org/3/glossary.html#term-argument for an example of the difference between positional and keyword arguments.

McKenzieW: Oh ok I see. I didn't realize that you were mentioning that they were positional arguments.

McKenzieW:
@edx/helio please review

ecommerce/courses/utils.py
```diff
@@ -87,3 +93,37 @@ def get_certificate_type_display_value(certificate_type):
...
+    Returns:
+        list of enrollments
+    """
+    try:
+        url = '{}?user={}'.format(get_lms_enrollment_api_url(), username)
+        timeout = settings.ENROLLMENT_FULFILLMENT_TIMEOUT
+
+        headers = {
+            'Content-Type': 'application/json',
+            'X-Edx-Api-Key': settings.EDX_API_KEY
```
clintonb: This endpoint now supports using JWT for authentication, and staff users can now read all enrollments. Please update this code to use the JWT. In general, if you find yourself needing the API key header, you should update the API to use proper authentication and authorization schemes.

staubina: @clintonb Can you please provide a link to documentation or an example of how to upgrade the authentication method.  Thank you.

ecommerce/courses/utils.py
```diff
@@ -87,3 +93,37 @@ def get_certificate_type_display_value(certificate_type):
...
+        }
+
+        response = requests.get(url, headers=headers, timeout=timeout)
+        status_code = response.status_code
+        if status_code == 200:
+            return response.json()
+        else:
+            logger.warning('Failed to retrieve enrollments for [%s]. Enrollment API returned status code [%d].',
+                           username, status_code)
+    except Exception:  # pylint: disable=broad-except
```
staubina: nit: You do not need to changes this I know it was a code move. Just as a general note I am not a fan of using general Exception catching.  It can be risky and hide/mask real errors.  There is logging in this case so we should be ok, but it should be avoided.

ecommerce/extensions/dashboard/users/tests/test_views.py
```diff
@@ -47,7 +46,7 @@ def test_enrollments_switch_inactive(self):
         self.switch.save()
 
         mock_get_enrollments = mock.Mock()
-        with mock.patch.object(UserDetailView, '_get_enrollments', mock_get_enrollments):
+        with mock.patch('ecommerce.courses.utils.get_enrollments', mock_get_enrollments):
```
staubina: I am glad we were able to work out this Mocking issue 👍

ecommerce/programs/conditions.py
```diff
@@ -57,17 +59,22 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
...
+        except (HttpNotFoundError, SlumberBaseException, requests.Timeout):
             return False
 
         applicable_seat_types = program['applicable_seat_types']
+        enrollments = utils.get_enrollments(basket.owner.username)
 
         for course in program['courses']:
-            # If the basket has no SKUs, but we still have courses over which to iterate,
-            # the basket cannot meet the condition that all courses be represented.
+            # If the user is alread enrolled in a course, we do not need to check their basket for it
```
staubina: Small issue, Typo in the comment

ecommerce/courses/utils.py
McKenzieW: I think I've gotten it mostly sorted, just can't find information on a JWT key that would allow us to access returned data (all the examples I found in the code base were for payment processing)

staubina (changes requested):
Some minor changes requested. Otherwise looks good.

ecommerce/courses/utils.py
clintonb: See https://github.com/edx/ecommerce/blob/06dcd6566c07517088ce9a71ce22fe335f2a156c/ecommerce/core/models.py#L422-L424, and mimic that behavior for the Enrollment API.

marjev (approved):

marjev:
You should fix quality, otherwise looks good.

staubina (approved):
Good work.  Changes are working in the basket as expected.  Remember to wait on merging this.

👍

McKenzieW:
Changes from this PR are included in https://github.com/edx/ecommerce/pull/1434

PR #1402, opened 2017-07-20
Added a timeout to e2e tests

clintonb:
Each test now has a 45s timeout. This is one fix to help prevent
Jenkins jobs from running for hours.

LEARNER-1998


clintonb:
FYI @edx/learner

PR #1403, opened 2017-07-21
Updated rendering of Apple Pay button on mobile clients

clintonb:
All payment methods are now grouped together to avoid alignment issues on narrow windows. Additionally, we are now using the Apple Pay button without text. This style is slightly bolder and allows us to maintain a 120px width.

LEARNER-2002

### DIFF
--- a/ecommerce/static/sass/partials/views/_basket.scss
+++ b/ecommerce/static/sass/partials/views/_basket.scss
@@ -238,32 +238,27 @@
             }
 
             @supports (-webkit-appearance: -apple-pay-button) {
-                .apple-pay-set-up-button {
+                .apple-pay-button {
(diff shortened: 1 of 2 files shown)


clintonb:
screen shot 2017-07-21 at 1 20 27 am
screen shot 2017-07-21 at 1 20 49 am
screen shot 2017-07-21 at 1 21 03 am

clintonb:
FYI @mjfrey, this isn't perfect but is better than what's currently on master. Ultimately, getting this right will require adjusting the credit card button. The resolution of that image is too small, and its width doesn't match the others—150px instead of 120px. Once that's fixed, we can look at options for placing each button on its own row for small screens and adding a bit of margin between them.

PR #1404, opened 2017-07-21
Firing checkout step events

clintonb:
The firing of these events will allow us to measure our checkout funnel using Google Analytics.

LEARNER-1563 and LEARNER-1668


clintonb:
@edx/helio please review. My goal is to ship this by 11AM.

mjfrey (approved):

ecommerce/extensions/checkout/mixins.py
```diff
@@ -116,6 +131,7 @@ def handle_order_placement(self,
 
             basket.submit()
 
+        # TODO Fire "Checkout Step Completed" -- Step 2 -- Payment Info Entered
```
vkaracic: You left this TODO as a tech debt?

clintonb: Nope. That's a screw-up that will be fixed shortly.

PR #1412, opened 2017-07-25
Updated production requirements

clintonb:
This is an attempt to resolve production issues we are seeing. Also,
some of these are wildly outdated (circa 2015).


schenedx (approved):
👍

PR #1413, opened 2017-07-25
Upgraded Segment client

clintonb:


clintonb:
https://openedx.atlassian.net/browse/LEARNER-2076

PR #1418, opened 2017-07-27
Allow site-specific Segment event filtering

clintonb:
Each site can now control which Segment events get fired for the site. This will allow us to experiment with combinations to determine the source of our CPU issues in production.

LEARNER-2076


clintonb:
FYI @schenedx @edx/learner @edx/devops 

My current theory is that the Product Added/Removed events, which are fired more than any other event, are the culprits. I'm going to activate all events except those once this is deployed.

schenedx (commented):
👍

PR #1420, opened 2017-07-27
Disabled Product Added/Removed events for free items

clintonb:
The volume we see for edX.org leads to a dramatic increase in CPU usage. Given that orders for free items are ignored, there is no need for these events.

LEARNER-2076


clintonb:
FYI @schenedx @edx/devops @edx/learner

PR #1421, opened 2017-07-28
Removed SiteConfiguration._allowed_segment_events

clintonb:
Crisis resolved. This field is no longer needed.

LEARNER-2076


clintonb:
FYI @edx/learner @edx/devops this the cleanup from this week's CPU crisis.

feanil:
Will removing the column break the existing running app?

clintonb:
@feanil thanks for asking! Usage of the field was removed by the first commit of https://github.com/edx/ecommerce/pull/1420.

feanil (approved):
We should make a note on the (Hawthorne)[https://openedx.atlassian.net/wiki/display/COMM/Hawthorn) page.  If other people in the community are using this it will break them while they are deploying.

@nedbat FYI

clintonb:
@feanil this field was created yesterday for debugging purposes with help text stating the field was temporary. The likelihood of someone coming to rely on this field is quite low.

feanil:
Gotcha. Thanks!

PR #1422, opened 2017-07-28
Upgraded Segment client from 1.2.3 to 1.2.7

clintonb:
In addition to the upgrade we are no longer queueing events for tests. This should prevent the tests from spinning up unnecessary threads to fire fake events to a fake Segment property.

LEARNER-2076

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -282,7 +282,7 @@ class SiteConfiguration(models.Model):
 
     @cached_property
     def segment_client(self):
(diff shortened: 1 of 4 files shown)


clintonb:
FYI @edx/learner @edx/devops I suspect the tests were failing for previous upgrades due to threads not being able to exit. The threads weren't able to exit because they were never able to send the fake events to Segment. Disabling sending seems to do the trick. Note that we still call Segment in the tests, but the events are simply dropped.

PR #1425, opened 2017-07-31
Renaming classes and methods

iivic:
We are using the term discovery service instead of course catalog.
Other methods are renamed to better reflect their functionality.
This commit is separated from the PR containing work described in ticket

LEARNER-1135


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1425&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/coupons/tests/mixins.py
```diff
@@ -178,7 +180,7 @@ def callback(request, uri, headers):  # pylint: disable=unused-argument
 
     def mock_get_catalog_course_runs_for_failure(self, partner_code, query, error, discovery_api_url):
```
vkaracic: You should probably rename this too.

vkaracic (commented):
Should the rename include .js files as well? Since `Catalog Service` is used there also. I'm for that all instances of that name be renamed, not just class and method names, in which case you [missed](https://github.com/edx/ecommerce/blob/e18ee24c85a739b52866d2fea347761bc650dc73/ecommerce/courses/utils.py#L40) [a](https://github.com/edx/ecommerce/blob/e18ee24c85a739b52866d2fea347761bc650dc73/ecommerce/enterprise/entitlements.py#L180) [few](https://github.com/edx/ecommerce/blob/e18ee24c85a739b52866d2fea347761bc650dc73/ecommerce/enterprise/tests/test_entitlements.py#L403) [spots](https://github.com/edx/ecommerce/blob/e18ee24c85a739b52866d2fea347761bc650dc73/ecommerce/extensions/api/v2/views/vouchers.py#L267).

ecommerce/coupons/tests/mixins.py
iivic: I will remove the method altogether since I can't find where it's being used.

ecommerce/extensions/offer/migrations/0013_auto_20170731_1551.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('offer', '0012_condition_program_uuid'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='range',
+            name='course_catalog',
+            field=models.PositiveIntegerField(blank=True, help_text='Course catalog id from the Discovery Service.', null=True),
```
clintonb: Catalog ID...

iivic:
Please take another look.

ecommerce/extensions/api/v2/tests/views/test_vouchers.py
```diff
@@ -598,7 +598,7 @@ def test_offers_api_endpoint_for_course_catalog_voucher(self):
 
     def test_get_offers_for_course_catalog_voucher_api_error(self):
         """
-        Verify that offers api endpoint returns proper message if course catalog api returns error.
+        Verify that offers api endpoint returns proper message if Discovery api returns error.
```
vkaracic: `API`.

vkaracic (approved):

iivic:
@clintonb is this good to go?

clintonb:
Squash your commits.

PR #1426, opened 2017-08-01
[ENT-493] New data sharing consent API

haikuginger:
This pull request updates ecommerce to use the new generic data sharing consent API, as well as adopt a different format for communicating the linked enterprise to the LMS for use in consent creation.

Depends on edx/edx-platform#15706.

Testing instructions:

1. Create a premium course
2. Create a coupon attached to an EnterpriseCustomer
3. Ensure that the premium course is included in the catalog linked to the EnterpriseCustomer
4. Verify that the EnterpriseCustomer is configured to require data sharing consent 
5. Create a new user and go to the coupon redemption link
6. Verify that you're prompted for data sharing consent as expected.

### DIFF
--- a/ecommerce/enterprise/utils.py
+++ b/ecommerce/enterprise/utils.py
@@ -158,60 +158,6 @@ def get_or_create_enterprise_customer_user(site, enterprise_customer_uuid, usern
     return response
 
 
-def enterprise_customer_needs_consent(enterprise_customer_data):
-    """
-    Determine if consent should be prompted for on this enterprise customer.
-
-    Args:
-        enterprise_customer_data: A dictionary isomorphic with the EnterpriseCustomer
-            object returned by various endpoints of the Enterprise API.
-
-    Returns:
-        bool: Whether, in general, a user must provide consent to use offers provided by this EnterpriseCustomer.
-    """
-    if not enterprise_customer_data['enable_data_sharing_consent']:
-        return False
-
-    return enterprise_customer_data['enforce_data_sharing_consent'] in ('at_login', 'at_enrollment')
-
-
-def enterprise_customer_user_consent_provided(ec_user_data):
-    """
-    Determine if the EnterpriseCustomerUser has provided consent at an account level.
-
-    Args:
-        ec_user_data: A dictionary isomorphic with the EnterpriseCustomerUser
-            object returned by various endpoints of the Enterprise API.
-    """
-    return ec_user_data['data_sharing_consent'] and ec_user_data['data_sharing_consent'][0]['enabled']
-
-
-def get_enterprise_customer_user(site, username, enterprise_customer_uuid):
-    """
-    Get the EnterpriseCustomerUser with a particular username and linked to a particular
-    EnterpriseCustomer if it exists; otherwise, return None.
-
-    Args:
-        site (Site): The site which is handling the current request
-        username (str): The username of the user in the LMS
-        enterprise_customer_uuid (str): The UUID of the EnterpriseCustomer in the LMS
-
-    Returns:
-        dict: The single EnterpriseCustomerUser structure provided by the API
-        NoneType: Returns None if no EnterpriseCustomerUser is found
-    """
-    api = site.siteconfiguration.enterprise_api_client
-    api_resource = 'enterprise-learner'
-    endpoint = getattr(api, api_resource)
-    response = endpoint.get(
-        enterprise_customer=str(enterprise_customer_uuid),
-        username=str(username),
-    )
-    results = response.get('results')
-
-    return results[0] if results else None
-
-
 def get_enterprise_course_enrollment(site, enterprise_customer_user, course_id):
     """
     Get the EnterpriseCourseEnrollment between a particular EnterpriseCustomerUser and
@@ -254,29 +200,13 @@ def enterprise_customer_user_needs_consent(site, enterprise_customer_uuid, cours
             that the EnterpriseCustomer specified by the enterprise_customer_uuid
             argument provides for the course specified by the course_id argument.
     """
-    account_consent_provided = False
-    course_consent_provided = False
-
-    ec_user = get_enterprise_customer_user(site, username, enterprise_customer_uuid)
-
-    if ec_user:
-        account_consent_provided = enterprise_customer_user_consent_provided(ec_user)
-        enterprise_customer = ec_user['enterprise_customer']
-    else:
-        enterprise_customer = get_enterprise_customer(site, enterprise_customer_uuid)
-
-    consent_needed = enterprise_customer_needs_consent(enterprise_customer)
-
-    if consent_needed and not account_consent_provided and ec_user:
-        existing_course_enrollment = get_enterprise_course_enrollment(
-            site,
-            course_id=str(course_id),
-            enterprise_customer_user=ec_user['id'],
-        )
-        if existing_course_enrollment:
-            course_consent_provided = existing_course_enrollment.get('consent_granted', False)
-
-    return consent_needed and not (account_consent_provided or course_consent_provided)
+    consent_client = site.siteconfiguration.consent_api_client
+    endpoint = consent_client.data_sharing_consent
+    return endpoint.get(
+        username=username,
+        enterprise_customer_uuid=enterprise_customer_uuid,
+        course_id=course_id
+    )['consent_required']
 
 
 def get_enterprise_customer_from_voucher(site, voucher):
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -147,7 +147,7 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
                 pass
 
             if enterprise_customer_uuid is not None:
-                data['enterprise_course_consent'] = True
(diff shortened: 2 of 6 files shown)


openedx-webhooks:
Thanks for the pull request, @haikuginger! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1426"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/enterprise/utils.py
```diff
@@ -254,30 +254,17 @@ def enterprise_customer_user_needs_consent(site, enterprise_customer_uuid, cours
...
-            site,
-            course_id=str(course_id),
-            enterprise_customer_user=ec_user['id'],
-        )
-        if existing_course_enrollment:
-            course_consent_provided = existing_course_enrollment.get('consent_granted', False)
-
-    return consent_needed and not (account_consent_provided or course_consent_provided)
-
+    from slumber.exceptions import HttpNotFoundError
```
brittneyexline: is there a reason this line is here if it's getting imported above?

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -147,7 +147,7 @@ def _add_enterprise_data_to_enrollment_api_post(self, data, order):
                 pass
 
             if enterprise_customer_uuid is not None:
-                data['enterprise_course_consent'] = True
+                data['linked_enterprise_customer'] = str(enterprise_customer_uuid)
```
brittneyexline, line 150: i don't understand the purpose or effect of this change, can you explain?

ecommerce/enterprise/utils.py
```diff
@@ -254,30 +254,17 @@ def enterprise_customer_user_needs_consent(site, enterprise_customer_uuid, cours
             that the EnterpriseCustomer specified by the enterprise_customer_uuid
             argument provides for the course specified by the course_id argument.
     """
-    account_consent_provided = False
-    course_consent_provided = False
-
-    ec_user = get_enterprise_customer_user(site, username, enterprise_customer_uuid)
```
itsjeyd, line 260: @haikuginger `get_enterprise_customer_user` seems to be unused now.

ecommerce/enterprise/utils.py
```diff
@@ -254,30 +254,17 @@ def enterprise_customer_user_needs_consent(site, enterprise_customer_uuid, cours
...
-
-    if ec_user:
-        account_consent_provided = enterprise_customer_user_consent_provided(ec_user)
```
itsjeyd, line 263: @haikuginger `enterprise_customer_user_consent_provided` seems to be unused now.

ecommerce/enterprise/utils.py
```diff
@@ -254,30 +254,17 @@ def enterprise_customer_user_needs_consent(site, enterprise_customer_uuid, cours
...
-        enterprise_customer = ec_user['enterprise_customer']
-    else:
-        enterprise_customer = get_enterprise_customer(site, enterprise_customer_uuid)
-
-    consent_needed = enterprise_customer_needs_consent(enterprise_customer)
```
itsjeyd, line 268: @haikuginger `enterprise_customer_needs_consent` seems to be unused now.

itsjeyd (changes requested):
@haikuginger This looks good to me in general. I had a few comments about functions that no longer seem to be referenced anywhere, though, and would also like to know the answers to the questions that @brittneyexline raised.

> There is a a concern that we are not able to use the new endpoint to determine whether or not consent is required in ecommerce at the time of enrollment

Does the implementation address this?

ecommerce/enterprise/utils.py
haikuginger: Ah, no. Code cruft that I need to remove.

ecommerce/extensions/fulfillment/modules.py
haikuginger: Basically, previously, when this POST hit the enrollment API, we were implicitly getting the linked EnterpriseCustomer by looking for an EnterpriseCustomerUser, and then adding a consent record for the enrolled course with that user. The new consent API expects us to be more explicit, so now, the actual EnterpriseCustomer that was linked to the coupon is being transmitted across so that we can use it in an explicit call to the consent API.

ecommerce/enterprise/utils.py
haikuginger: 👍

ecommerce/enterprise/utils.py
haikuginger: 👍

ecommerce/enterprise/utils.py
haikuginger: 👍

haikuginger:
Rebased from fa852e4f.

haikuginger:
@itsjeyd @brittneyexline, tests are set to go and comments are addressed.

haikuginger:
Squashed from 9ab32b25.

itsjeyd (commented):
@haikuginger Recent updates to the code look good to me. I tried testing this on the sandbox, and found that I wasn't able to create a coupon (for Pied Piper and the VNC course); got

> Error! An error occurred while saving the data.

when trying to submit coupon data.

haikuginger:
@itsjeyd I'm guessing there's a migration in the Pink Panther branch that I'm missing here. Stand by.

haikuginger:
@itsjeyd Yeah, just needed to run migrations. This should be set for you to run through on the sandbox now.

itsjeyd (approved):
@haikuginger Yup, was able to complete the steps from the test instructions now :+1:

- [x] I tested this: Works as advertised in PR description
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~ N/A (no new flows)
- [x] ~~Includes documentation~~ N/A

ecommerce/enterprise/utils.py
```diff
@@ -254,29 +200,16 @@ def enterprise_customer_user_needs_consent(site, enterprise_customer_uuid, cours
...
+    consent_client = site.siteconfiguration.consent_api_client
+    endpoint = consent_client.data_sharing_consent
+    try:
+        return endpoint.get(
+            username=username,
+            enterprise_customer_uuid=enterprise_customer_uuid,
+            course_id=course_id
+        )['consent_required']
+    except HttpNotFoundError as e:
+        return endpoint._try_to_serialize_response(e.response)['consent_required']  # pylint: disable=protected-access
```
douglashall: What is the purpose of this code? Why are we calling this protected function?

haikuginger: Slumber doesn't have a native way for us to get a deserialized response body when the server returned an error code. So we can either access Slumber's "private" functions to do it for us in a way that matches the way it behaves when it receives a non-error response, or we can just naively use the `json` module to do it ourselves - which is more Pylint-compliant, but also more brittle.

douglashall: Why are we returning a 404 status from the consent API if we are actually returning consent data?

douglashall (changes requested):
@haikuginger One last question to address here.

ecommerce/enterprise/utils.py
haikuginger: We return a 404 as an early distinguisher about whether a relevant record exists. There's no reason we can't also return a response body on a 404 with additional details, and the 404 status code conceptually maps to "entity does not exist" in the HTTP lexicon.

douglashall: I think 200 is the more appropriate status code here if you are returning a non-error response body from the API endpoint. If we need a way to identify that the data sharing consent record does not exist then let's add that to the data returned from the API endpoint.

haikuginger: Okay, making that change!

haikuginger:
@douglashall, addressed.

haikuginger:
Looks like we've got unrelated ESLint tests that are failing on multiple branches. I'll try rebasing on master to see if that resolves it.

EDIT: Already on master, unfortunately... not sure what changed.

douglashall:
@haikuginger Looks like the build failure here is not related to ESLint, but instead:

```
ecommerce/enterprise/utils.py:269:1: E302 expected 2 blank lines, found 1
Makefile:60: recipe for target 'quality' failed
make: *** [quality] Error 1
```

haikuginger:
@douglashall I don't see those lines in the Travis log at all. Here's what I see:

```
Message:

    Failed with 2 errors

Makefile:65: recipe for target 'validate_js' failed

make: *** [validate_js] Error 1
```

haikuginger:
Rebased from c69bc7d0.

haikuginger:
@douglashall, we've got a green check here, so once you sign off, we're set to merge. No rush, though, since we want to delay the rollout on this somewhat.

douglashall (approved):

haikuginger:
Squashed from 8ada68da.

PR #1427, opened 2017-08-01
Remove COURSE_CATALOG_API_URL

vkaracic:
Follow-up to the TODO left in `ecommerce/core/models.py`.

@edx/helio please review.

---
TODO:
- [ ] Remove the `use_multi_tenant_discovery_api_urls` switch
- [x] Update docker devstack to include `discovery-api-url` flag value when running create_or_update_site (It's already [set](https://github.com/edx/devstack/blob/master/provision-ecommerce.sh#L7))


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1427"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/core/models.py
```diff
@@ -381,10 +380,7 @@ def discovery_api_client(self):
             EdxRestApiClient: The client to access the Discovery service.
         """
 
-        # TODO Once the change is verified remove the switch and COURSE_CATALOG_API_URL from settings.
-        if waffle.switch_is_active('use_multi_tenant_discovery_api_urls') and self.discovery_api_url:
-            return EdxRestApiClient(self.discovery_api_url, jwt=self.access_token)
-        return EdxRestApiClient(settings.COURSE_CATALOG_API_URL, jwt=self.access_token)
+        return EdxRestApiClient(self.discovery_api_url, jwt=self.access_token)
```
iivic, line 390: What happens when the discovery_api_url is not set?

iivic (commented):
You should verify with someone if this feature is approved just to make sure nothing breaks.

ecommerce/core/models.py
vkaracic: Now it will be a required field, as it should be since I the ecommerce service is dependent on the discovery service.

vkaracic:
@edx/helio please review.

iivic (approved):

ecommerce/core/management/commands/create_or_update_site.py
```diff
@@ -128,7 +128,7 @@ def add_arguments(self, parser):
                             action='store',
                             dest='discovery_api_url',
                             type=str,
-                            required=False,
+                            required=True,
                             default='',
```
clintonb: Remove this default. Also, set blank to False.

clintonb: Ignore the bit about blank.

ecommerce/core/tests/test_commands.py
```diff
@@ -184,7 +184,25 @@ def test_update_site(self):
          '--theme-scss-path=site/sass/css/', '--payment-processors=cybersource'],
         ['--site-id=1', '--site-domain=fake.server', '--partner-name=fake_partner',
          '--theme-scss-path=site/sass/css/', '--payment-processors=cybersource',
-         '--segment-key=abc']
+         '--segment-key=abc'],
+        ['--site-id=1', '--site-domain=fake.server', '--partner-name=fake_partner',
```
clintonb, line 188: Why did you add so many cases?

vkaracic: Because when new required arguments were added, this test was not updated to include them. I'm including the `discovery_api_url` argument together with all the other omitted ones.

clintonb (approved):

clintonb:
Squash your commits and link to an appropriate ticket.

PR #1433, opened 2017-08-02
Standardize product expiration date value.

vkaracic:
There was a discrepancy between the product expiration date outputs. The tests format the expiration date according to the ISO 8601 standard while the `ProductSerializer` would print out the `datetime` object (which included milliseconds which aren't part of the standard).
Instead of changing the way the tests' serializer outputs the value, I've changed the way the `ProductSerializer` outputs the expiration date value.

@edx/helio please review.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1433"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

clintonb:
See https://github.com/encode/django-rest-framework/issues/4255. This problem will go away when we upgrade DRF to 3.4+.

vkaracic:
Very well.

PR #1435, opened 2017-08-04
Add Site fk to ConditionalOffer model

bill-filler:
This will make Program Offers multi-site aware by adding a foreign key
to the Sites table. Null values are accepted for this phase, and
django-admin has been modified to expose Site to manaully change the
data.

WL-1152


ecommerce/extensions/offer/models.py
```diff
@@ -35,6 +35,9 @@ def clean_value(self):
 class ConditionalOffer(AbstractConditionalOffer):
     UPDATABLE_OFFER_FIELDS = ['email_domains', 'max_uses']
     email_domains = models.CharField(max_length=255, blank=True, null=True)
+    site = models.ForeignKey(
+        'sites.Site', verbose_name=_("Site"), null=True, blank=True, default=None, on_delete=models.SET_NULL
```
clintonb: Ditch the on_delete. If a site goes away, we don't want to keep a bunch of unassociatrd data.

clintonb (changes requested):
Tests?

clintonb:
Drop the merge bubble. Rebase on master instead of merging it into your branch.

ecommerce/extensions/offer/models.py
bill-filler: ok

bill-filler:
@clintonb addressed issues. tests written, removed on_delete and rebased my branch on master. Let me know if anything else.

ecommerce/extensions/offer/models.py
```diff
@@ -35,6 +35,9 @@ def clean_value(self):
 class ConditionalOffer(AbstractConditionalOffer):
     UPDATABLE_OFFER_FIELDS = ['email_domains', 'max_uses']
     email_domains = models.CharField(max_length=255, blank=True, null=True)
+    site = models.ForeignKey(
+        'sites.Site', verbose_name=_("Site"), null=True, blank=True, default=None
```
clintonb: nit: single quotes

ecommerce/extensions/offer/tests/test_admin.py
```diff
@@ -0,0 +1,19 @@
...
+class OfferAdminTests(TestCase):
+
+    offer_add_page_url = reverse('admin:offer_conditionaloffer_add')
+
+    def setUp(self):
+        super(OfferAdminTests, self).setUp()
+        self.user = UserFactory(is_staff=True, is_superuser=True, password=self.password)
+        self.client.login(username=self.user.username, password=self.password)
+
+    def test_site_displayed(self):
```
clintonb: This test isn't useful. Unless you actually check the HTML to determine if the site field is present, you're just testing the frameworks—Django and Oscar. This file can be removed.

ecommerce/programs/forms.py
```diff
@@ -91,6 +91,7 @@ def save(self, commit=True):
         self.instance.status = ConditionalOffer.OPEN
         self.instance.offer_type = ConditionalOffer.SITE
         self.instance.max_basket_applications = 1
+        self.instance.site = site
```
clintonb, line 94: This is really what I was looking for tests around. We need to ensure the form associates a `Site` with the offer. See https://github.com/edx/ecommerce/blob/master/ecommerce/programs/tests/test_forms.py. You should be able to get away with adding a line to `test_save_create` and `test_save_edit`.

bill-filler: fixed. actually added the check to assert_program_offer_conditions as that gets called from both functions as well as test_save_offer_name

clintonb (approved):

clintonb:
Remember to update the [Hawthorn release notes](https://openedx.atlassian.net/wiki/display/COMM/Hawthorn).

bill-filler:
Hawthorn release notes updated

PR #1437, opened 2017-08-07
ENT-530: Coupon code gets valid for 'all the courses' if an enterprise customer is selected during coupon creation.

saleem-latif:
Hi @brittneyexline , @douglashall , @zubair-arbi , @asadiqbal08 

Please take a look at the changes, This PR updates error message when coupons is not applicable to the selected course.

__Steps To Reproduce:__
1. Open coupon creation form in ecommerce:
https://ecommerce-business.sandbox.edx.org/coupons/new/
2. Select "Single Course" radio button and enter course ID in "Course ID" field.
3. Select any enterprise customer from "Enterprise Customer" drop down.
4. Input all other required data and press "Create Coupon" button.
5. Download coupon code.
6. Login/Register to LMS with learner.
7. Select a course that is not linked with coupon. 
8. Click "Enroll in Course" button from course about page.
9. Copy coupon code from downloaded coupon file.
10. Paste coupon code in basket page.
11. Click apply button and view.

__Expected Result:__
Error message should appear as it is appearing for coupons that are created without enterprise customer "This coupon code is not valid for this course. Try a different course."

### DIFF
--- a/ecommerce/coupons/utils.py
+++ b/ecommerce/coupons/utils.py
@@ -163,3 +163,21 @@ def fetch_course_catalog(site, catalog_id):
 
     cache.set(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
     return response
+
+
+def is_voucher_applied(basket, voucher):
+    """
+    Check if given voucher is applied to the given basket.
+
+    Arguments:
+        basket (Basket): oscar basket object to checked for discount voucher.
+        voucher (Voucher): Discount voucher.
+
+    Returns:
+         (bool): True if given voucher is applied to the basket, False otherwise
+    """
+    # Look for discounts from this new voucher
+    for discount in basket.offer_applications:
+        if discount['voucher'] and discount['voucher'] == voucher:
+            return True
+    return False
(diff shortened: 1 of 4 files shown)


ecommerce/coupons/utils.py
```diff
@@ -163,3 +163,21 @@ def fetch_course_catalog(site, catalog_id):
 
     cache.set(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
     return response
+
+
+def is_voucher_applied(basket, voucher):
```
brittneyexline, line 168: maybe this should be renamed to is_enterprise_voucher_applied, though i'm conflicted because this doesn't actually have enterprise specific logic, but is called only from an enterprise context.

brittneyexline (commented):
Looks good in general, still need to test on the sandbox.

ecommerce/coupons/utils.py
saleem-latif: @brittneyexline although it is used only in enterprise related code at the moment, `is_voucher_applied ` is general and can be used in other places, that is why I named it `is_voucher_applied`

brittneyexline (approved):
I've verified this on the sandbox.

zubair-arbi:
Changes looks good 👍

zubair-arbi (approved):

PR #1439, opened 2017-08-08
LEARNER-1184 Disable enrollment code product revocation in refund process

Ayub-Khan:
[Learner-1184](https://openedx.atlassian.net/browse/LEARNER-1184)

Description:
Refund revocataion process was breaking for enrollment codes because of no enrollment for user. Currently do not have proper rules for implementation so bypassing it. People were able to get multiple refunds for single orders. This pr will disable that functionality.

### DIFF
--- a/ecommerce/extensions/refund/models.py
+++ b/ecommerce/extensions/refund/models.py
@@ -128,13 +128,16 @@ class Refund(StatusMixin, TimeStampedModel):
 
             status = getattr(settings, 'OSCAR_INITIAL_REFUND_LINE_STATUS', REFUND_LINE.OPEN)
             for line in unrefunded_lines:
-                RefundLine.objects.create(
-                    refund=refund,
-                    order_line=line,
-                    line_credit_excl_tax=line.line_price_excl_tax,
-                    quantity=line.quantity,
-                    status=status
-                )
+                # LEARNER-1184 Disabling refund line creation for enrollment code, Because
+                # it causes Multiple Refunds errors.
+                if not line.product.is_enrollment_code_product:
+                    RefundLine.objects.create(
+                        refund=refund,
+                        order_line=line,
+                        line_credit_excl_tax=line.line_price_excl_tax,
+                        quantity=line.quantity,
+                        status=status
+                    )
 
             if total_credit_excl_tax == 0:
                 refund.approve(notify_purchaser=False)
--- a/ecommerce/extensions/refund/tests/test_models.py
+++ b/ecommerce/extensions/refund/tests/test_models.py
@@ -8,13 +8,19 @@ import mock
 from django.conf import settings
 from oscar.apps.payment.exceptions import PaymentError
 from oscar.core.loading import get_class, get_model
-from oscar.test.newfactories import UserFactory
+from oscar.test.newfactories import BasketFactory, UserFactory
 from testfixtures import LogCapture
 
-from ecommerce.core.constants import SEAT_PRODUCT_CLASS_NAME
+from ecommerce.core.constants import (
+    ENROLLMENT_CODE_PRODUCT_CLASS_NAME,
+    ENROLLMENT_CODE_SWITCH,
+    SEAT_PRODUCT_CLASS_NAME
+)
+from ecommerce.core.tests import toggle_switch
 from ecommerce.core.url_utils import get_lms_enrollment_api_url
 from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.extensions.checkout.utils import format_currency, get_receipt_page_url
+from ecommerce.extensions.fulfillment.status import ORDER
 from ecommerce.extensions.payment.tests.processors import DummyProcessor
 from ecommerce.extensions.refund import models
 from ecommerce.extensions.refund.exceptions import InvalidStatus
@@ -26,6 +32,7 @@ from ecommerce.tests.testcases import TestCase
 
 PaymentEventType = get_model('order', 'PaymentEventType')
 post_refund = get_class('refund.signals', 'post_refund')
+Product = get_model('catalogue', 'Product')
 Refund = get_model('refund', 'Refund')
 Source = get_model('payment', 'Source')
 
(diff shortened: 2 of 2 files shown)


Ayub-Khan:
@mjfrey @clintonb  Thoughts?

ecommerce/extensions/fulfillment/api.py
```diff
@@ -134,13 +134,24 @@ def revoke_fulfillment_for_refund(refund):
...
         for refund_line in refund.lines.all():
             order_line = refund_line.order_line
-            modules = get_fulfillment_modules_for_line(order_line)
-
-            for module in modules:
-                if module().revoke_line(order_line):
-                    refund_line.set_status(REFUND_LINE.COMPLETE)
-                else:
+            # TODO: Disabling revocation for enrollment code products until we have proper rules for that
+            if not order_line.product.is_enrollment_code_product:
```
clintonb: There should be no product-specific code here. Take a look at https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/fulfillment/modules.py#L482. There is no revocation function for enrollment codes.

The simpler solution is to update the code that creates refunds to never create refund lines for enrollment code lines.

clintonb (changes requested):

ecommerce/extensions/refund/models.py
```diff
@@ -128,13 +128,14 @@ def create_with_lines(cls, order, lines):
...
             status = getattr(settings, 'OSCAR_INITIAL_REFUND_LINE_STATUS', REFUND_LINE.OPEN)
             for line in unrefunded_lines:
-                RefundLine.objects.create(
-                    refund=refund,
-                    order_line=line,
-                    line_credit_excl_tax=line.line_price_excl_tax,
-                    quantity=line.quantity,
-                    status=status
-                )
+                if not line.product.is_enrollment_code_product:
```
clintonb: Add a comment, and ticket number, explaining why we do not allow enrollment code products to be refunded.

ecommerce/extensions/refund/tests/test_models.py
```diff
@@ -150,6 +150,21 @@ def test_create_with_lines_with_existing_refund(self, refund_status, refund_crea
             else:
                 l.check()
 
+    def test_create_with_lines_for_enrollment_code_order_line(self):
+        """
+        Test Refund.create_with_lines should not create RefundLines for enrollment
+        coe order line
```
clintonb: code

ecommerce/extensions/refund/tests/test_models.py
```diff
@@ -150,6 +150,21 @@ def test_create_with_lines_with_existing_refund(self, refund_status, refund_crea
             else:
                 l.check()
 
+    def test_create_with_lines_for_enrollment_code_order_line(self):
+        """
+        Test Refund.create_with_lines should not create RefundLines for enrollment
```
clintonb: Nit: No need for the word "Test" here.

ecommerce/extensions/refund/tests/test_models.py
```diff
@@ -150,6 +150,21 @@ def test_create_with_lines_with_existing_refund(self, refund_status, refund_crea
...
+        coe order line
+        """
+        order = self.create_order(user=UserFactory(), enrollment_code_order=True)
+        line = order.lines.first()
+
+        with LogCapture(LOGGER_NAME) as l:
+            refund = Refund.create_with_lines(order, [line])
+            self.assertEqual(isinstance(refund, Refund), True)
+            self.assert_refund_creation_logged(l, refund, order)
```
clintonb, line 182: Checking logs is nice, but you really need to verify that no `RefundLine` is associated with `line`.

ecommerce/extensions/refund/tests/mixins.py
```diff
@@ -24,20 +27,31 @@
 class RefundTestMixin(DiscoveryTestMixin):
     def setUp(self):
         super(RefundTestMixin, self).setUp()
+        toggle_switch(ENROLLMENT_CODE_SWITCH, True)
```
clintonb: Move all of this to the test case that needs it. We don't need to do all of this work for every single test case.

clintonb (changes requested):

schenedx:
@Ayub-Khan I agree this is the better approach than your first attempt.

ecommerce/extensions/refund/models.py
```diff
@@ -128,13 +128,16 @@ def create_with_lines(cls, order, lines):
...
-                RefundLine.objects.create(
-                    refund=refund,
-                    order_line=line,
-                    line_credit_excl_tax=line.line_price_excl_tax,
-                    quantity=line.quantity,
-                    status=status
-                )
+                # LEARNER-1184 Disabling refund line creation for enrollment code, Because
+                # it causes Multiple Refunds errors.
+                if not line.product.is_enrollment_code_product:
```
schenedx, line 133: Question: If a refund on the order is created without any refund lines, what happens? Is that a realistic case?

clintonb: Good question. This check should be moved up to the declaration of `unrefunded_lines`. If that array is empty no `Refund` will be created.

Ayub-Khan: I think if no refund is generated, we wont be even able to refund the money. we will totally disable the refund option for enrollment codes.

clintonb: That's the point. No one should be buying enrollment codes with real money.

mjfrey: Actually on WL sites, enrollment codes are purchased with money.  When a quantity greater than 1 for the seat is entered, enrollment codes are generated and sent to the purchaser.

clintonb: Bah! Who said you could go on vacation! @Ayub-Khan please work with Product (Deen and Mike Martin) to determine the correct behavior. Once that is determined, please document that behavior in `/docs` and implement it.

Ayub-Khan: @clintonb I will create a hipchat room including all the related people to find a resolution fast.

schenedx:
@Ayub-Khan Close this PR? Let the user story do this work.

PR #1442, opened 2017-08-08
[WIP] BasketCalculateView supports username parameter.

Mjloturco:
In order for basket calculate view to correctly return discounts to marketing site, an optional parameter of username is included in the BasketCalculateView instead of always relying on request.user.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_baskets.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_baskets.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import datetime
 import json
 import urllib
+import logging
 from collections import namedtuple
 from decimal import Decimal
 
@@ -18,6 +19,8 @@ from oscar.test import factories
 from oscar.test.factories import BasketFactory
 from rest_framework.throttling import UserRateThrottle
 
+from ecommerce.courses.models import Course
+from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.extensions.api import exceptions as api_exceptions
 from ecommerce.extensions.api.v2.tests.views import JSON_CONTENT_TYPE, OrderDetailViewTestMixin
 from ecommerce.extensions.api.v2.views.baskets import BasketCreateView
@@ -42,6 +45,8 @@ Voucher = get_model('voucher', 'Voucher')
 
 LOGGER_NAME = 'ecommerce.extensions.api.v2.views.baskets'
 
+
+logger = logging.getLogger(__name__)
 
 @ddt.ddt
 @override_settings(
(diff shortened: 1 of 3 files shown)


Mjloturco:
@edx/helio Hey all, feel free to give this a look over

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -369,12 +372,18 @@ def get(self, request):
         if not voucher and len(products) == 1:
             voucher = get_entitlement_voucher(request, products[0])
 
+        username = request.GET.get('username')
+        try:
+            user = User.objects.get(username=username) if username else request.user
+        except User.DoesNotExist:
+            raise api_exceptions.BadRequestException('User "{}" does not exist.'.format(username))
```
clintonb: We generally use square brackets (e.g. `[username]`) rather than quotes.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -369,12 +372,18 @@ def get(self, request):
         if not voucher and len(products) == 1:
             voucher = get_entitlement_voucher(request, products[0])
 
+        username = request.GET.get('username')
+        try:
+            user = User.objects.get(username=username) if username else request.user
```
clintonb: What happens if I pass in the username of someone else? There should be permissions on this endpoint to avoid exposing details about one user to another.

clintonb: Drop the `else` and add `user = request.user` above the `try` block.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -369,12 +372,19 @@ def get(self, request):
...
+        except User.DoesNotExist:
+            user = request.user
+            #raise api_exceptions.BadRequestException('User "{}" does not exist.'.format(username))
```
clintonb: Dead code should be removed.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -369,12 +372,19 @@ def get(self, request):
         if not voucher and len(products) == 1:
             voucher = get_entitlement_voucher(request, products[0])
 
+        username = request.GET.get('username')
+        try:
+            user = User.objects.get(username=username) if username else request.user
+        except User.DoesNotExist:
+            user = request.user
```
clintonb: Add a debug logging statement to aid local development but avoid spamming production logs.

clintonb (changes requested):
Follow the [commit message guidelines](https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/161880915/Git+Best+Practices#GitBestPractices-CommitMessages).

ecommerce/extensions/api/v2/views/baskets.py
Mjloturco: Looping in @mjfrey on this, he'd mentioned that since only pricing info is being returned no user details are really exposed.

clintonb: You are exposing the fact that a learner may, or may not, be enrolled in certain course runs. It may seem minor, but it's a potential breach of learner trust.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -436,6 +436,18 @@ def test_basket_calculate_fixed_coupon(self):
...
+    def test_basket_calculate_valid_username(self):
+        """Verify that a valid username passed calculates a basket for that specific user"""
+        response = self.client.get(self.url + '&username={username}'.format(username=self.user.username))
+        self.assertEqual(response.status_code, 200)
+
+    @httpretty.activate
+    def test_basket_calculate_invalid_username_passed(self):
+        """Verify that an invalid username still gets a response"""
+        response = self.client.get(self.url + '&username=userdoesntexist')
+        self.assertEqual(response.status_code, 200)
```
vkaracic: How can you be sure that the response is not the same in these two tests? I suggest asserting that the response data is the one that you expect for a particular case in addition to the status code assertion.

ecommerce/extensions/api/v2/views/baskets.py
mjfrey: I understand that a user could after some real digging determine that another user has already purchased a verified / professional seat based on a list of skus.  All this endpoint returns is price.  If we are really concerned with this, how would we ensure that only a privileged user or the actual learner can call this?

clintonb: If `request.user` is a staff member, allow all calls. Otherwise, if `request.user.username` does not match the passed in username (remember to account for cAsE), return HTTP 403 (Forbidden) with an error message stating that users can only view their own data.

PR #1445, opened 2017-08-10
Dropped Promise return value for unsupported browsers

clintonb:
If the browser does not support Promises, an empty promise will NOT be
returned if Apple Pay is not supported. Users with browsers that support
the Promise feature will not be affected.

LEARNER-2218


clintonb:
FYI @edx/learner

PR #1448, opened 2017-08-11
Track programs or other bundles associated with a completed order

McKenzieW:
Track programs or other bundles associated with a completed order

prepare_basket creates a 'bundle' basket attribute if the request includes a bundle, and this is tracked along with the basket's products in the 'Order Completed' analytic event

[LEARNER-1721]
https://openedx.atlassian.net/browse/LEARNER-1721

### DIFF
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -25,6 +25,9 @@ from ecommerce.tests.testcases import TestCase, TransactionTestCase
 
 Benefit = get_model('offer', 'Benefit')
 Basket = get_model('basket', 'Basket')
+BasketAttribute = get_model('basket', 'BasketAttribute')
+BasketAttributeType = get_model('basket', 'BasketAttributeType')
+BUNDLE = 'bundle_identifier'
 Product = get_model('catalogue', 'Product')
 
 
@@ -358,6 +361,20 @@ class BasketUtilsTests(DiscoveryTestMixin, TestCase):
         with mock.patch.object(UserAlreadyPlacedOrder, 'user_already_placed_order', return_value=True):
             basket = prepare_basket(self.request, [enrollment_code])
             self.assertIsNotNone(basket)
+
+    def test_prepare_basket_with_bundle(self):
+        """
+        Test prepare_basket updates or creates a basket attribute for the associated bundle
+        """
+        product = ProductFactory()
+        request = self.request
+        basket = prepare_basket(request, [product])
+        with self.assertRaises(BasketAttribute.DoesNotExist):
+            BasketAttribute.objects.get(basket=basket, attribute_type__name=BUNDLE)
+        request.GET = {'bundle': 'test_bundle'}
+        basket = prepare_basket(request, [product])
+        bundle_id = BasketAttribute.objects.get(basket=basket, attribute_type__name=BUNDLE).value_text
+        self.assertEqual(bundle_id, 'test_bundle')
 
 
 class BasketUtilsTransactionTests(TransactionTestCase):
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -2,13 +2,17 @@ import logging
 
 import waffle
 from django.dispatch import receiver
-from oscar.core.loading import get_class
+from oscar.core.loading import get_class, get_model
 
 from ecommerce.courses.utils import mode_for_seat
 from ecommerce.extensions.analytics.utils import silence_exceptions, track_segment_event
 from ecommerce.extensions.checkout.utils import get_credit_provider_details, get_receipt_page_url
 from ecommerce.notifications.notifications import send_notification
+from ecommerce.programs.utils import get_program
 
+BasketAttribute = get_model('basket', 'BasketAttribute')
+BasketAttributeType = get_model('basket', 'BasketAttributeType')
+BUNDLE = 'bundle_identifier'
 logger = logging.getLogger(__name__)
 post_checkout = get_class('checkout.signals', 'post_checkout')
 
@@ -29,6 +33,10 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
 
     voucher = order.basket_discounts.filter(voucher_id__isnull=False).first()
     coupon = voucher.voucher_code if voucher else None
+    try:
+        bundle_id = BasketAttribute.objects.get(basket=order.basket, attribute_type__name=BUNDLE).value_text
+    except BasketAttribute.DoesNotExist:
+        bundle_id = None
 
     properties = {
         'orderId': order.number,
@@ -51,6 +59,26 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
             } for line in order.lines.all()
         ],
     }
+
+    try:
+        bundle_id = BasketAttribute.objects.get(basket=order.basket, attribute_type__name=BUNDLE).value_text
+        program = get_program(bundle_id, order.basket.site.siteconfiguration)
+        if len(order.lines.all()) < len(program['courses']):
+            variant = 'partial'
+        else:
+            variant = 'full'
+        bundle_product = {
+            'id': bundle_id,
+            'price': '0',
+            'quantity': str(len(order.lines.all())),
+            'category': 'bundle',
+            'variant': variant,
+            'name': program['name']
+        }
+        properties['products'].append(bundle_product)
+    except BasketAttribute.DoesNotExist:
+        logger.info('There is no program or bundle associated with order number %s', order.number)
+
     track_segment_event(order.site, order.user, 'Order Completed', properties)
 
 
--- a/ecommerce/extensions/checkout/tests/test_signals.py
+++ b/ecommerce/extensions/checkout/tests/test_signals.py
@@ -13,24 +13,26 @@ from ecommerce.core.tests import toggle_switch
 from ecommerce.coupons.tests.mixins import CouponMixin
 from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.courses.utils import mode_for_seat
-from ecommerce.extensions.catalogue.tests.mixins import DiscoveryTestMixin
 from ecommerce.extensions.checkout.signals import send_course_purchase_email, track_completed_order
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.test.factories import create_order, prepare_voucher
+from ecommerce.programs.tests.mixins import ProgramTestMixin
 from ecommerce.tests.factories import ProductFactory
 from ecommerce.tests.testcases import TestCase
 
 Applicator = get_class('offer.utils', 'Applicator')
+BasketAttribute = get_model('basket', 'BasketAttribute')
+BasketAttributeType = get_model('basket', 'BasketAttributeType')
 Benefit = get_model('offer', 'Benefit')
 Condition = get_model('offer', 'Condition')
 ConditionalOffer = get_model('offer', 'ConditionalOffer')
-
-LOGGER_NAME = 'ecommerce.extensions.checkout.signals'
-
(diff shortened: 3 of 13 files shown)


McKenzieW:
@edx/helio please review

ecommerce/extensions/basket/models.py
```diff
@@ -15,6 +15,10 @@ class Basket(AbstractBasket):
         'sites.Site', verbose_name=_("Site"), null=True, blank=True, default=None, on_delete=models.SET_NULL
     )
 
+    bundle = models.CharField(
```
clintonb: This table has 15,785,356 rows. Adding a field will lock the table, and break enrollments for the duration of the lock. We have a management command, `delete_ordered_baskets`, that should remove many of the baskets, but that might not remove enough of them.

clintonb: Actually...where is this field used?

ecommerce/sailthru/migrations/0002_add_basket_attribute_type.py
```diff
@@ -3,13 +3,15 @@
...
 
+from ecommerce.extensions.checkout.signals import BUNDLE
 from ecommerce.sailthru.signals import SAILTHRU_CAMPAIGN
 
 
 def create_attribute(apps, schema_editor):
     BasketAttributeType = apps.get_model('basket', 'BasketAttributeType')
 
     BasketAttributeType.objects.create(name=SAILTHRU_CAMPAIGN)
+    BasketAttributeType.objects.create(name=BUNDLE)
```
clintonb: This migration will not be run again in production. If you want this created, make a new migration. Even if this migration had not been run, you should avoid cross-pollinating app migrations in this manner. Your changes have nothing to do with Sailthru, thus your changes should not rely on Sailthru's migrations.

clintonb (changes requested):
Grenade coming...you can't rely on baskets for reporting. Baskets aren't intended for longterm storage. If you want to store this info, it should be associated with an order since we will never delete orders.

ecommerce/extensions/basket/models.py
McKenzieW: Wow that was a dumb mistake on my part. This is from an old approach and I must've just forgotten to delete that code.

clintonb:
@McKenzieW this PR cannot be merged. See my note about baskets.

mjfrey:
Why can't the basket be used here for firing events?  There is no order when items are added to the basket.  Do we need persistence here or is it just fine to fire the event?  A basket would not be gone before the order complete event happens?

clintonb:
I see. I missed the fact that you are firing an event. As long as everyone understands that the basket may go away after the order is placed, all is well.

McKenzieW:
Yupp the reason we're associating the attribute with the basket as opposed to the offer is that we only need it for the event

ecommerce/extensions/checkout/signals.py
```diff
@@ -29,6 +32,11 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
 
     voucher = order.basket_discounts.filter(voucher_id__isnull=False).first()
     coupon = voucher.voucher_code if voucher else None
+    try:
+        bundle_id = BasketAttribute.objects.get(basket=order.basket,
```
clintonb: You don't need `BasketAttributeType` here:
```python
BasketAttribute.objects.get(basket=order.basket, attribute_type__name=BUNDLE).value_text
```

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -26,7 +26,9 @@
 ConditionalOffer = get_model('offer', 'ConditionalOffer')
 
 LOGGER_NAME = 'ecommerce.extensions.checkout.signals'
-
+BasketAttribute = get_model('basket', 'BasketAttribute')
+BasketAttributeType = get_model('basket', 'BasketAttributeType')
+BUNDLE = 'bundle_identifier'
```
clintonb: Nit: separate your constants from the class names.

ecommerce/programs/migrations/0002_add_basket_attribute_type.py
```diff
@@ -0,0 +1,27 @@
...
+
+def create_attribute(apps, schema_editor):
+    BasketAttributeType = apps.get_model('basket', 'BasketAttributeType')
+
+    BasketAttributeType.objects.create(name=BUNDLE)
+
+
+def delete_attribute(apps, schema_editor):
+    BasketAttributeType = apps.get_model('basket', 'BasketAttributeType')
+    BasketAttributeType.objects.get(name=BUNDLE).delete()
```
clintonb: Catch the `DoesNotExist` exception, and do nothing, just in case something goes awry.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -151,13 +163,20 @@ def test_track_completed_order(self):
...
 
-            # We should be able to fire events even if the product is not releated to a course.
+            # We should be able to fire events even if the product is not related to a course.
             mock_track.reset_mock()
             order = create_order()
             track_completed_order(None, order)
             properties = self._generate_event_properties(order)
             mock_track.assert_called_once_with(order.site, order.user, 'Order Completed', properties)
 
+            # If the order is a bundle purchase, we should track the associated bundle in the properties
```
clintonb: Nit: make this a separate test case.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -37,7 +39,7 @@ def setUp(self):
         self.request.user = self.user
         toggle_switch('ENABLE_NOTIFICATIONS', True)
 
-    def prepare_order(self, seat_type, credit_provider_id=None):
+    def prepare_order(self, seat_type, credit_provider_id=None, bundle_id=None):
```
clintonb: You're relying on the bundle in exactly one test case. Move this code to that test case.

ecommerce/extensions/checkout/signals.py
```diff
@@ -51,6 +59,10 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
             } for line in order.lines.all()
         ],
     }
+
+    if bundle_id:
+        properties['products'].append({'bundle': bundle_id})
```
clintonb: Why is this nested under `'products'`? That array follows semantics set by Google and Segment. I fear adding a product that does not follow the semantics will cause issues.

ecommerce/extensions/checkout/signals.py
```diff
@@ -51,6 +59,10 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
             } for line in order.lines.all()
         ],
     }
+
+    if bundle_id:
```
clintonb: It might be cleaner to always include the bundle ID, even if it is `null`.

clintonb (changes requested):
Biggest cause of concern are the event properties. We shouldn't be putting non-products in the products array.

ecommerce/extensions/checkout/signals.py
mjfrey: Speaking to Adley he suggested sending the UUID as a label

properties['bundle_id'] = bundle_id

clintonb: That is cleaner and shouldn't conflict with existing properties.

McKenzieW: Do you know if it is possible to add another property to the event that is not listed in its spec? If not, would 'affiliation' be a better property to set the the program identifier?

ecommerce/extensions/checkout/signals.py
McKenzieW: but isn't it possible that the query would return the sailthru attribute otherwise?

clintonb: If the Sailthru attributes have the name set to the value of `BUNDLE`, that's going to be returned regardless of implementation.

McKenzieW: Oh sorry right, I didn't read carefully. I thought at first you were suggesting that I leave out the parameter entirely.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -359,6 +362,21 @@ def test_prepare_basket_for_purchased_enrollment_code(self):
...
 
+    def test_prepare_basket_with_bundle(self):
+        """
+        Test prepare_basket updates or creates a basket attribute for the associated bundle
+        """
+        product = ProductFactory()
+        request = self.request
+        basket = prepare_basket(request, [product])
+        with self.assertRaises(BasketAttribute.DoesNotExist):
+            BasketAttribute.objects.get(basket=basket, attribute_type=BasketAttributeType.objects.get(name=BUNDLE))
```
clintonb: `BasketAttribute.objects.get(basket=basket, attribute_type__name=BUNDLE)`

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -359,6 +362,21 @@ def test_prepare_basket_for_purchased_enrollment_code(self):
...
+        request.GET = {'bundle': 'test_bundle'}
+        basket = prepare_basket(request, [product])
+        bundle_id = BasketAttribute.objects.get(basket=basket,
```
clintonb: `BasketAttribute.objects.get(basket=basket, attribute_type__name=BUNDLE)`

ecommerce/extensions/checkout/signals.py
```diff
@@ -2,20 +2,43 @@
...
+BasketAttributeType = get_model('basket', 'BasketAttributeType')
+BUNDLE = 'bundle_identifier'
 logger = logging.getLogger(__name__)
 post_checkout = get_class('checkout.signals', 'post_checkout')
 
 # Number of orders currently supported for the email notifications
 ORDER_LINE_COUNT = 1
 
 
+def get_program_data(program_uuid, site_configuration):
```
mjfrey: I feel like we have this implemented somewhere else.  We should make this a utility and not repeat.

ecommerce/extensions/checkout/tests/test_signals.py
```diff
@@ -55,6 +57,27 @@ def prepare_order(self, seat_type, credit_provider_id=None):
...
         return order
 
+    def mock_get_program_data(self, isFull):
+        data = {'name': 'test_program', 'courses': [{}]}
+        if isFull:
+            data['courses'].append({})
+        return data
+
+    @httpretty.activate
+    def test_get_program_data(self):
```
mjfrey: Same for this test.  I think we already have a test exactly like this.

mjfrey (changes requested):

ecommerce/extensions/checkout/signals.py
McKenzieW: So there are two other methods:
https://github.com/edx/ecommerce/blob/master/ecommerce/programs/conditions.py#L29-L44
https://github.com/edx/ecommerce/blob/master/ecommerce/programs/views.py#L26-L38

that also essentially just call the api method:
https://github.com/edx/ecommerce/blob/master/ecommerce/programs/api.py


Do you want me to remove those methods as well?

mjfrey: Yes, I would consolidate all of those methods into one utility.

ecommerce/extensions/checkout/utils.py
```diff
@@ -77,3 +79,22 @@ def add_currency(amount):
         str: Formatted price with currency.
     """
     return format_currency(settings.OSCAR_DEFAULT_CURRENCY, amount, u'#,##0.00')
+
+
+def get_program(program_uuid, siteconfiguration):
```
mjfrey: Good, but please move this to:
ecommerce/programs/utils.py

McKenzieW:
@mjfrey good to go?

mjfrey (approved):
looks good.  Please squash commits.

PR #1450, opened 2017-08-16
Fix eslint errors.

douglashall:
ESLint was complaining about missing whitespace in comments at the
beginning of a couple of files. I am removing those comments because
they do not offer any value.


schenedx (approved):
👍

McKenzieW (approved):

mjfrey (approved):

PR #1454, opened 2017-08-21
Fix broken docs links in ecommerce docs

schenedx:
@mjfrey @clintonb Please review
Just trying to clean up the errors when compiling docs on ecommerce

### DIFF
--- a/docs/create_products/create_products_overview.rst
+++ b/docs/create_products/create_products_overview.rst
@@ -4,7 +4,19 @@
 Creating Products Overview
 ###########################
 
-.. include:: ../products_desc.txt
+The edX platform offers several types of products. You create these products in
+E-Commerce web pages.
+
+* Course seats represent an enrollment track. Each course seat has an
+  associated set of attributes, such as price and certificate availability. The
+  edX code uses course seats to determine how a given enrollment should be
+  handled. For more information, see :ref:`Create Course Seats`.
+
+* Coupons allow users to offer learners a discount, either percentage or fixed
+  amount, off a course enrollment. For more information, see :ref:`Create and
+  Manage Coupons`.
+
+* Enrollment codes allow users to purchase bulk enrollments for a course.
 
 ******************************
 Prepare to Create a Product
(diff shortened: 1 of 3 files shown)


mjfrey (approved):

docs/index.rst
```diff
@@ -23,6 +21,7 @@ use the E-Commerce service with the Open edX platform.
    test_features
    test_ecommerce
    additional_features/index
+   theming
```
bderusha: `theming` already exists on line 17

docs/create_products/create_products_overview.rst
```diff
@@ -4,8 +4,6 @@
 Creating Products Overview
 ###########################
 
-.. include:: ../products_desc.txt
```
bderusha, line 7: Do we know why this is broken?  Isn't this information still relevant to include here?

docs/index.rst
```diff
@@ -13,8 +13,6 @@ use the E-Commerce service with the Open edX platform.
 .. toctree::
    :maxdepth: 2
 
-   about_ecommerce
-   flow
```
bderusha, line 17: Same question as above:
 Do we know why this is broken?  Isn't this information still relevant to include here?

bderusha (changes requested):

docs/index.rst
schenedx: I'll remove that

docs/index.rst
schenedx: @bderusha Looks like these broken links are added as part of https://github.com/edx/ecommerce/commit/e85bad23373c314c28b39ef91d7aa177d97af65e
The commit does not have any PR associated. I remember David had trouble revert and re-introduce commits when he is working on this task.
I think the thing to do that make most sense is to revert commit e85bad23373c314c28b39ef91d7aa177d97af65e

bderusha: Agreed.  Let's revert the commit and make sure `theming` gets added back if that is working.

bderusha (approved):
:shipit:

PR #1456, opened 2017-08-22
Filter Program Offers to only show offers for the current site

bill-filler:
Program Offers are now site-aware and the view is filtered to only show
offers for the current site. Site field also added to dashboard/offers
UI when creating/displaying offers from the dashboard.

WL-1152

### DIFF
--- a/ecommerce/extensions/dashboard/offers/forms.py
+++ b/ecommerce/extensions/dashboard/offers/forms.py
@@ -1,0 +1,12 @@
+from django.contrib.sites.models import Site
+from django.forms import ModelChoiceField
+from oscar.apps.dashboard.offers.forms import MetaDataForm as CoreMetaDataForm
+from oscar.core.loading import get_model
+
+
+class MetaDataForm(CoreMetaDataForm):
+    site = ModelChoiceField(queryset=Site.objects.all(), required=True)
+
+    class Meta(object):
+        model = get_model('offer', 'ConditionalOffer')
+        fields = ('name', 'description', 'site')
--- a/ecommerce/programs/views.py
+++ b/ecommerce/programs/views.py
@@ -25,6 +25,7 @@ class ProgramOfferViewMixin(StaffOnlyMixin):
 
     def get_queryset(self):
         return super(ProgramOfferViewMixin, self).get_queryset().filter(
+            site=self.request.site.id,
             condition__program_uuid__isnull=False,
             offer_type=ConditionalOffer.SITE
         )
--- a/ecommerce/templates/oscar/dashboard/offers/offer_detail.html
+++ b/ecommerce/templates/oscar/dashboard/offers/offer_detail.html
@@ -1,0 +1,117 @@
+{% extends 'dashboard/layout.html' %}
+{% load currency_filters %}
+{% load i18n %}
+
+{% block title %}
+    {% blocktrans with name=offer.name %}
+        {{ name }} | Offers
+    {% endblocktrans %} | {{ block.super }}
+{% endblock %}
+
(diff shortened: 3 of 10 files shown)


bill-filler:
I think the codecov is failing because of the oscar templates (which are large html files) I had to include to add Site to the summary and details page:
https://github.com/edx/ecommerce/pull/1456/files#diff-bd70d183ff3ef2312b933085571eb273
https://github.com/edx/ecommerce/pull/1456/files#diff-89356728367e81c60819b61474c73c72
Not sure what to do about that.

mjfrey:
I would add a view test that submits a form so the new code that stores the site in the kwargs gets tested.

ecommerce/extensions/dashboard/offers/forms.py
```diff
@@ -0,0 +1,8 @@
+from oscar.apps.dashboard.offers.forms import MetaDataForm as CoreMetaDataForm
+from oscar.core.loading import get_model
+
+
+class MetaDataForm(CoreMetaDataForm):
+    class Meta:  # pylint: disable=C1001
```
clintonb: Don't disable this. Fix it: `class Meta(object)`.

ecommerce/programs/views.py
```diff
@@ -24,7 +24,9 @@ def get_context_data(self, **kwargs):
         return context
 
     def get_queryset(self):
+        site = self.request.site
```
clintonb: Nit: Merge this with the line below. The `site` variable isn't necessary.

ecommerce/templates/oscar/dashboard/offers/offer_detail.html
```diff
@@ -0,0 +1,102 @@
+{% extends 'dashboard/layout.html' %} {% load currency_filters %} {% load i18n %} {% block title %} {% blocktrans with name=offer.name %} {{ name }} | Offers {% endblocktrans %} | {{ block.super }} {% endblock %} {% block breadcrumbs %}
```
clintonb: Break this up into multiple lines. Each `extends` and `load` call is on its own line. Each `block` is its own block.

ecommerce/templates/oscar/dashboard/offers/summary.html
```diff
@@ -0,0 +1,41 @@
...
+
+        <ul>
+            {% for restriction in offer.availability_restrictions %}
+            <li>{{ restriction.description }}</li>
+            {% endfor %}
+        </ul>
+        {% else %} - {% endif %}
+    </div>
+    {% endif %}
+</div>
```
clintonb: Add a new line.

ecommerce/templates/oscar/dashboard/offers/offer_detail.html
```diff
@@ -0,0 +1,102 @@
...
+        <tr>
+            <td><a href="{% url 'dashboard:order-detail' number=order.number %}">{{ order.number }}</a></td>
+            <td>{{ order.date_placed }}</td>
+            <td>{{ order.total_incl_tax|currency }}</td>
+            <td>{{ discount.amount|currency }}</td>
+        </tr>
+        {% endwith %} {% endfor %}
+    </tbody>
+</table>
+{% include 'dashboard/partials/pagination.html' %} {% endif %} {% endblock dashboard_content %}
```
clintonb: Add a new line.

clintonb (changes requested):
You need tests. Codecov doesn't look at HTML templates. You've added Python code that is not tested.

clintonb (approved):
Remember to squash commits.

PR #1457, opened 2017-08-22
Do not tell sailthru about multi-product baskets 

MatthewPiatetsky:
It seems we already do this for some things like coupons.
Sailthru is not currently used for bundles.
Product has determined that we want to implement the abandoned cart email for now only for courses and have a separate effort to implement it for bundles (programs).

### DIFF
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -149,6 +149,7 @@ class BasketCreateView(EdxOrderPlacementMixin, generics.CreateAPIView):
 
             requested_products = request.data.get('products')
             if requested_products:
+                is_multi_product_basket = True if len(requested_products) > 1 else False
                 for requested_product in requested_products:
                     # Ensure the requested products exist
                     sku = requested_product.get('sku')
@@ -182,8 +183,8 @@ class BasketCreateView(EdxOrderPlacementMixin, generics.CreateAPIView):
 
                     # Call signal handler to notify listeners that something has been added to the basket
                     basket_addition = get_class('basket.signals', 'basket_addition')
-                    basket_addition.send(sender=basket_addition, product=product, user=request.user,
-                                         request=request, basket=basket)
+                    basket_addition.send(sender=basket_addition, product=product, user=request.user, request=request,
+                                         basket=basket, is_multi_product_basket=is_multi_product_basket)
             else:
                 # If no products were included in the request, we cannot checkout.
                 return self._report_bad_request(
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -85,13 +85,14 @@ def prepare_basket(request, products, voucher=None):
             )
             return basket
 
+    is_multi_product_basket = True if len(products) > 1 else False
     for product in products:
         if product.is_enrollment_code_product or \
                 not UserAlreadyPlacedOrder.user_already_placed_order(request.user, product):
             basket.add_product(product, 1)
             # Call signal handler to notify listeners that something has been added to the basket
             basket_addition.send(sender=basket_addition, product=product, user=request.user, request=request,
-                                 basket=basket)
+                                 basket=basket, is_multi_product_basket=is_multi_product_basket)
         else:
             already_purchased_products.append(product)
             logger.warning(
--- a/ecommerce/sailthru/signals.py
+++ b/ecommerce/sailthru/signals.py
@@ -51,8 +51,12 @@ def process_checkout_complete(sender, order=None, user=None, request=None,  # py
     #  If multi product orders become common it may be worthwhile to pass an array of
     #  orders to the worker in one call to save overhead, however, that would be difficult
     #  because of the fact that there are different templates for free enroll versus paid enroll
-    for line in order.lines.all():
-
+    lines = order.lines.all()
+    # We are not sending multi product orders to sailthru for now, because
+    # the abandoned cart email does not yet support baskets with multiple products
+    if len(lines) > 1:
+        return
+    for line in lines:
         # get product
         product = line.product
         sku = line.partner_sku
@@ -72,7 +76,7 @@ def process_checkout_complete(sender, order=None, user=None, request=None,  # py
 
 @receiver(basket_addition)
 @silence_exceptions("Failed to call Sailthru upon basket addition.")
-def process_basket_addition(sender, product=None, user=None, request=None, basket=None,
+def process_basket_addition(sender, product=None, user=None, request=None, basket=None, is_multi_product_basket=None,
                             **kwargs):  # pylint: disable=unused-argument
     """Tell Sailthru when payment started.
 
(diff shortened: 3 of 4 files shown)


mjfrey:
You can probably do something like this:
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/checkout/signals.py#L37

schenedx:
@MatthewPiatetsky I agree with @mjfrey.
I think the assumption that multiple products in basket is always a bundle is too risky.
I have seen bundles sold with just 1 item. Or orders with multiple items in it that's not a bundle.

schenedx (changes requested):
Let's use to Bundle ID to determine whether the basket is a bundled basket or not.

MatthewPiatetsky:
@schenedx what is an example of orders that are not a bundle but have multiple items?
I can check with product, but I think the requirement at the moment might just be that we wanted to do a single item email first and a multi item email later?
If the only multi item carts we have are bundles then using this bundle check makes sense, although if not, perhaps not

also @mjfrey you said this was using the wrong api?

mjfrey:
This is the API that gets called when purchasing bundles.
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/views.py#L95

MatthewPiatetsky:
@mjfrey I think the function you linked to eventually calls the code I modified.
BasketMultipleItemsView calls prepare_basket
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/views.py#L130
prepare_basket calls basket_addition
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/utils.py#L91

schenedx:
@MatthewPiatetsky Yes. I have an example. See 19858657

MatthewPiatetsky:
@schenedx @mjfrey if we have orders that are not bundles with multiple lines, i think the product requirement here would be to check the number of lines (not whether something is a bundle) because the emails would not be correct for the non-bundle multi line orders either

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -149,6 +149,7 @@ def create(self, request, *args, **kwargs):
 
             requested_products = request.data.get('products')
             if requested_products:
+                is_bundle = True if len(requested_products) > 1 else False
```
schenedx: I think calling this variable `is_bundle` is going to confuse others. 
What about calling this `is_multi_product_basket`?

ecommerce/sailthru/signals.py
```diff
@@ -51,8 +51,11 @@ def process_checkout_complete(sender, order=None, user=None, request=None,  # py
     #  If multi product orders become common it may be worthwhile to pass an array of
     #  orders to the worker in one call to save overhead, however, that would be difficult
     #  because of the fact that there are different templates for free enroll versus paid enroll
-    for line in order.lines.all():
-
+    lines = order.lines.all()
+    is_bundle = True if len(lines) > 1 else False
+    if is_bundle:
```
schenedx: Why do we need 3 lines while just 2 lines can achieve the same?
```
if len(lines) > 1:
    return
```

ecommerce/sailthru/signals.py
```diff
@@ -108,7 +111,7 @@ def process_basket_addition(sender, product=None, user=None, request=None, baske
         # an item has been added to the shopping cart so that an abandoned cart message can be sent
         # later if the purchase is not completed.  Abandoned cart support is only for purchases, not
         # for free enrolls
-        if price:
+        if price and not is_bundle:
```
schenedx: Again, the variable name is not accurate in my opinion

ecommerce/sailthru/tests/test_signals.py
```diff
@@ -82,6 +82,29 @@ def test_unsupported_product_class(self, mock_log_error, mock_update_course_enro
         self.assertFalse(mock_log_error.called)
 
     @patch('ecommerce_worker.sailthru.v1.tasks.update_course_enrollment.delay')
+    @patch('ecommerce.sailthru.signals.logger.error')
+    def test_unsupported_bundles(self, mock_log_error, mock_update_course_enrollment):
```
schenedx: I think a more accurate name for this test case is `test_stop_sailthru_update_on_multi_product_baskets`

schenedx (changes requested):
@MatthewPiatetsky If you are not specifically checking for bundles, and instead want to stop Sailthru basket update for all multi-product baskets, we should update the variables and comments accordingly to this requirement

MatthewPiatetsky:
@schenedx  did i address your comments?

ecommerce/extensions/basket/utils.py
```diff
@@ -85,13 +85,14 @@ def prepare_basket(request, products, voucher=None):
             )
             return basket
 
+    is_bundle = True if len(products) > 1 else False
```
schenedx: `is_multi_product_basket`

ecommerce/extensions/basket/utils.py
```diff
@@ -85,13 +85,14 @@ def prepare_basket(request, products, voucher=None):
...
     for product in products:
         if product.is_enrollment_code_product or \
                 not UserAlreadyPlacedOrder.user_already_placed_order(request.user, product):
             basket.add_product(product, 1)
             # Call signal handler to notify listeners that something has been added to the basket
             basket_addition.send(sender=basket_addition, product=product, user=request.user, request=request,
-                                 basket=basket)
+                                 basket=basket, is_bundle=is_bundle)
```
schenedx: You'll want to update this variable as well to `is_multi_product_basket`

ecommerce/sailthru/signals.py
```diff
@@ -51,8 +51,10 @@ def process_checkout_complete(sender, order=None, user=None, request=None,  # py
     #  If multi product orders become common it may be worthwhile to pass an array of
     #  orders to the worker in one call to save overhead, however, that would be difficult
     #  because of the fact that there are different templates for free enroll versus paid enroll
-    for line in order.lines.all():
-
+    lines = order.lines.all()
+    if len(lines) > 1:
+        return
```
schenedx: Add some comments about multiple product basket case, and why we choose to skip sending to Sailthru

ecommerce/sailthru/tests/test_signals.py
```diff
@@ -82,6 +82,29 @@ def test_unsupported_product_class(self, mock_log_error, mock_update_course_enro
...
+        seat = self.course.create_or_update_seat('verified', False, 100, self.partner, None)
+        other_course = CourseFactory(site=self.site)
+        other_seat = other_course.create_or_update_seat('verified', False, 100, self.partner, None)
+        basket = BasketFactory(owner=self.user, site=self.site)
+        basket.add_product(seat)
+        basket.add_product(other_seat)
+        bundle_order = create_order(number=2, basket=basket, user=self.user, site=self.site)
+
+        # This method takes an argument to determine whether that product is part of a multi-product basket
+        process_basket_addition(None, request=self.request, user=self.user, product=seat, is_bundle=True)
```
schenedx: Should we update the variable here from `is_bundle` to `is_multi_product_basket`?

schenedx (commented):
See my inline comments

mjfrey:
It really seems to me that we are doing to much work to not send notifications to bundle purchase users.  Why do we not just teach sailthru about multi-line item basket/orders?

MatthewPiatetsky:
@mjfrey sounds like we decided to go ahead with this after talking
@schenedx further comments?

schenedx (approved):
👍

PR #1461, opened 2017-08-25
Delete invalid basket attributes

McKenzieW:
Updated prepare_basket method to remove old entries from the Basket Attribute table such that baskets are not mistakenly associated with bundle purchases in orders after a bundle purchase by the same user for single courses.

### DIFF
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -70,6 +70,8 @@ def prepare_basket(request, products, voucher=None):
             attribute_type=BasketAttributeType.objects.get(name=BUNDLE),
             defaults={'value_text': bundle}
         )
+    else:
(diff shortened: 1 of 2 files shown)


McKenzieW:
@edx/learner-helio review please :)

ecommerce/extensions/basket/utils.py
```diff
@@ -71,10 +71,7 @@ def prepare_basket(request, products, voucher=None):
             defaults={'value_text': bundle}
         )
     else:
-        try:
```
McKenzieW: The filter().delete() call does not raise the DoesNotExist exception when the object does not exist, so the try and except are unnecessary

staubina (approved):
:shipit:

mjfrey (approved):

PR #1465, opened 2017-08-30
Deprecated optional.txt

clintonb:
The New Relic package has been upgraded and moved to production.txt. As we move toward consolidating Ansible configuration, we are adopting best practices from more recent IDAs here in this repo. We no longer use optional.txt. The cost of installing the optional package is minimal relative to the administrative overhead of optionally installing it, thus it will always be installed. However, the code will not be active on servers that are not running the New Relic agent.

LEARNER-818

### DIFF
--- a/requirements/monitoring/requirements.txt
+++ b/requirements/monitoring/requirements.txt
@@ -7,7 +7,6 @@
 -r ../base.txt
 -r ../docs.txt
 -r ../local.txt
--r ../optional.txt
 -r ../production.txt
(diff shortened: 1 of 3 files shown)


clintonb:
@edx/devops please review
@edx/learner FYI

The `optional.txt` file will be deleted after https://github.com/edx/configuration/pull/3941 is deployed and deemed stable post-release.

jibsheet (approved):
Seems reasonable, this is what we do in platform

PR #1466, opened 2017-08-30
Added .env file to run e2e tests on devstack

clintonb:


clintonb:
FYI @edx/learner @edx/docker-devstack-working-group

PR #1468, opened 2017-09-05
[ENT-587] More gracefully handle errors when hitting APIs

UmanShahzad:
This PR is meant to handle errors more gracefully in the codebase -- specifically those places where `edx/edx-enterprise` requires it.

**JIRA tickets**: [ENT-587](https://openedx.atlassian.net/browse/ENT-587)

**Merge deadline**: By the end of the week.

**Testing instructions**:

It's hard to test this because you'd need to force an error somehow. In many cases you can force the errors through over-throttling. But other than that, the tests should build the confidence you need.

But since this is meant primarily for the edx-enterprise's calls to ecommerce, know that edx-enterprise only makes calls to the basket calculation view. In that case, you can over-throttle /api/v2/baskets/calculate and see the result.

**Reviewers**
- [ ] @haikuginger 
- [ ] @douglashall @brittneyexline

### DIFF
--- a/ecommerce/programs/conditions.py
+++ b/ecommerce/programs/conditions.py
@@ -3,11 +3,11 @@ from __future__ import unicode_literals
 import logging
 import operator
 
-import requests
 from django.conf import settings
 from django.core.cache import cache
 from oscar.apps.offer import utils as oscar_utils
 from oscar.core.loading import get_model
+from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
 
 from ecommerce.core.utils import get_cache_key
@@ -62,7 +62,7 @@ class ProgramCourseRunSeatsCondition(Condition):
         basket_skus = set([line.stockrecord.partner_sku for line in basket.all_lines()])
         try:
             program = get_program(self.program_uuid, basket.site.siteconfiguration)
-        except (HttpNotFoundError, SlumberBaseException, requests.Timeout):
+        except (HttpNotFoundError, SlumberBaseException, Timeout):
             return False
 
         if program:
@@ -70,6 +70,7 @@ class ProgramCourseRunSeatsCondition(Condition):
         else:
             return False
 
+        enrollments = []
         if basket.site.siteconfiguration.enable_partial_program:
             api_resource = 'enrollments'
             cache_key = get_cache_key(
@@ -77,14 +78,16 @@ class ProgramCourseRunSeatsCondition(Condition):
                 resource=api_resource,
                 username=basket.owner.username,
             )
-            enrollments = cache.get(cache_key)
+            enrollments = cache.get(cache_key, [])
             if not enrollments:
                 api = basket.site.siteconfiguration.enrollment_api_client
                 user = basket.owner.username
-                enrollments = api.enrollment.get(user=user)
-                cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
-        else:
-            enrollments = []
+                try:
+                    enrollments = api.enrollment.get(user=user)
+                except (ConnectionError, SlumberBaseException, Timeout) as exc:
+                    logger.error('Failed to retrieve enrollments: %s', str(exc))
+                else:
+                    cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
 
         for course in program['courses']:
             # If the user is already enrolled in a course, we do not need to check their basket for it
--- a/ecommerce/programs/tests/mixins.py
+++ b/ecommerce/programs/tests/mixins.py
@@ -63,7 +63,7 @@ class ProgramTestMixin(DiscoveryTestMixin):
         )
         return data
 
-    def mock_enrollment_api(self, username, enrollments=[]):  # pylint: disable=dangerous-default-value
+    def mock_enrollment_api(self, username, enrollments=None, response_code=200):
         """ Mocks enrollment retrieval from LMS
         Returns:
             list: Mocked enrollment data
@@ -72,7 +72,8 @@ class ProgramTestMixin(DiscoveryTestMixin):
         httpretty.register_uri(
             method=httpretty.GET,
             uri='{}?user={}'.format(get_lms_enrollment_api_url(), username),
-            body=json.dumps(enrollments),
+            body=json.dumps([] if enrollments is None else enrollments),
+            status=response_code,
             content_type='application/json'
         )
         return enrollments
--- a/ecommerce/programs/tests/test_conditions.py
+++ b/ecommerce/programs/tests/test_conditions.py
@@ -127,7 +127,7 @@ class ProgramCourseRunSeatsConditionTests(ProgramTestMixin, TestCase):
             self.assertTrue(self.condition.is_satisfied(offer, basket))
 
     @ddt.data(HttpNotFoundError, SlumberBaseException, Timeout)
-    def test_is_satisfied_with_exception(self, value):
+    def test_is_satisfied_with_exception_for_programs(self, value):
         """ The method should return False if there is an exception when trying to get program details. """
         offer = factories.ProgramOfferFactory(condition=self.condition)
         basket = factories.BasketFactory(site=self.site, owner=factories.UserFactory())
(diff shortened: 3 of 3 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1468"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

UmanShahzad:
I've covered the case that was written in the ticket, but there may be more places that the basket calculation logic takes us where API calls are being made without an outer try-catch block. I'll have to investigate more to be sure.

UmanShahzad:
@haikuginger @brittneyexline @douglashall I did more testing to see if there's any more weak spots that aren't handling errors gracefully, but I can't find them. I think at this point instead of investing more effort into trying to magically run into a new pain point, we can review / merge this change and continue to do something similar for more ungracefully-handled errors in the future as they pop up in logs.

ecommerce/programs/conditions.py
```diff
@@ -81,8 +81,13 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
...
-                cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
+                try:
+                    enrollments = api.enrollment.get(user=user)
+                except (ConnectionError, SlumberBaseException, Timeout) as exc:
+                    logger.error('Failed to retrieve enrollments: %s', str(exc))
+                    enrollments = []
+                else:
+                    cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
         else:
             enrollments = []
```
douglashall: Could we do this instead to simplify a bit:
```
enrollments = []
if basket.site.siteconfiguration.enable_partial_program:
    api_resource = 'enrollments'
    cache_key = get_cache_key(
        site_domain=basket.site.domain,
        resource=api_resource,
        username=basket.owner.username,
    )
    enrollments = cache.get(cache_key)
    if not enrollments:
        api = basket.site.siteconfiguration.enrollment_api_client
        user = basket.owner.username
        enrollments = api.enrollment.get(user=user)
        cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
        try:
            enrollments = api.enrollment.get(user=user)
            logger.exception('Failed to retrieve enrollments from enrollment API:')
        else:
            cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
```

UmanShahzad: That's much better, yep! Thanks.

douglashall (approved):

douglashall (approved):

douglashall:
@UmanShahzad Reminder to rebase/squash before merging.

UmanShahzad:
@douglashall Will do :+1:

@haikuginger Can you take a look when you have time? I'll then rebase and squash.

ecommerce/programs/tests/mixins.py
```diff
@@ -72,7 +72,7 @@ def mock_enrollment_api(self, username, enrollments=[]):  # pylint: disable=dang
         httpretty.register_uri(
             method=httpretty.GET,
             uri='{}?user={}'.format(get_lms_enrollment_api_url(), username),
-            body=json.dumps(enrollments),
+            body=json.dumps(enrollments) if enrollments else self.throw_exception,
```
haikuginger: I don't think this'll do what you want it to do. What you want to do is set the status code to an error code; HTTP requests to the given URL will then receive a response with that code, and Slumber (the HTTP client itself) will interpret that as an error and raise an exception.

I also think you probably want to do `if enrollments is None` rather than just checking for truthiness so that an empty list can be explicitly passed if desired.

haikuginger (changes requested):
One change request to the testing harness; I'm going to test this manually now.

ecommerce/settings/base.py
```diff
@@ -457,7 +457,7 @@
         'rest_framework.throttling.UserRateThrottle',
     ),
     'DEFAULT_THROTTLE_RATES': {
-        'user': '50/minute',
+        'user': '500/minute',
```
haikuginger: I'm unclear on how this is relevant to the other changes.

UmanShahzad: Oh my, sorry this slipped through, that was for testing.

ecommerce/programs/tests/mixins.py
UmanShahzad: I think you're right, although I do think it throws the error as I hoped because (apparently) I have test coverage? Either way, that's a way simpler solution that I just didn't know about.

ecommerce/programs/tests/mixins.py
```diff
@@ -63,7 +63,7 @@ def mock_program_detail_endpoint(self, program_uuid, discovery_api_url, empty=Fa
         )
         return data
 
-    def mock_enrollment_api(self, username, enrollments=[]):  # pylint: disable=dangerous-default-value
+    def mock_enrollment_api(self, username, enrollments=None):
```
haikuginger: Looks like we've got several tests throughout the testing suite that use the default parameter here, so now they're testing something else - instead of testing an explicitly returned empty set of enrollments, they're testing the exception behavior.

Can you modify this method to have an extra optional parameter `response_code` that accepts a numeric response code, defaulting to 200, and then add explicit tests of the error case? Something like this:

```python
def mock_enrollment_api(self, username, enrollments=None, response_code=200):
    enrollments = enrollments or []
    httpretty.register_uri(
        method=httpretty.GET,
        uri='{}?user={}'.format(get_lms_enrollment_api_url(), username),
        body=json.dumps(enrollments),
        status=response_code,
    )
    return enrollments
```

EDIT: Suggestion: mock an error status _with_ a body so that we actually test that the response body isn't being used in the error case.

haikuginger (changes requested):

ecommerce/programs/conditions.py
```diff
@@ -81,10 +82,13 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
...
                 api = basket.site.siteconfiguration.enrollment_api_client
                 user = basket.owner.username
-                enrollments = api.enrollment.get(user=user)
-                cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
-        else:
-            enrollments = []
+                try:
+                    enrollments = api.enrollment.get(user=user)
+                except (ConnectionError, SlumberBaseException, Timeout) as exc:
+                    enrollments = []
```
UmanShahzad: @douglashall Turns out I still had to have this, because it's being gotten from the cache above, which may give `None`, and then it gets used below as a list, causing errors.

douglashall: You could do `cache.get(cache_key, [])`.

UmanShahzad: /facepalm. Yeah, I need a rest. :P Thanks.

ecommerce/programs/tests/mixins.py
```diff
@@ -72,7 +72,8 @@ def mock_enrollment_api(self, username, enrollments=[]):  # pylint: disable=dang
         httpretty.register_uri(
             method=httpretty.GET,
             uri='{}?user={}'.format(get_lms_enrollment_api_url(), username),
-            body=json.dumps(enrollments),
+            body=json.dumps([] if enrollments is None else enrollments),
```
UmanShahzad, line 75: I do this to avoid the pylint suppression above.

ecommerce/programs/tests/test_conditions.py
```diff
@@ -2,7 +2,7 @@
 import httpretty
 import mock
 from oscar.core.loading import get_model
-from requests import Timeout
+from requests import ConnectionError, Timeout
```
UmanShahzad: Sorry, this is unused.

UmanShahzad:
@haikuginger Ok, attempted to apply the test suggestion.

douglashall:
@haikuginger Is this ready for merge now?

haikuginger (approved):
:+1:

- [x] I tested this: Verified the cleaner behavior when receiving an error code from the enrollment API
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~ N/A
- [x] Includes documentation

UmanShahzad:
@douglashall Am I good to merge? Do we need any other reviews for ecommerce PRs?

douglashall:
@UmanShahzad I think you are good to merge once tests pass. Thanks.

ecommerce/programs/conditions.py
```diff
@@ -62,29 +62,32 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
...
-                enrollments = api.enrollment.get(user=user)
-                cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
-        else:
-            enrollments = []
+                try:
+                    enrollments = api.enrollment.get(user=user)
+                except (ConnectionError, SlumberBaseException, Timeout) as exc:
+                    logger.error('Failed to retrieve enrollments: %s', str(exc))
+                else:
+                    cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
```
clintonb, line 90: You're potentially caching invalid data for an hour. This is not good.

UmanShahzad: Sorry -- we can change this to skip the caching step if it's invalid.

PR #1469, opened 2017-09-05
Call isort on codebase

UmanShahzad:
We seem to do an isort check on the codebase as part of quality, but there are files that haven't actually been sorted.

This just calls it in a separate PR to make the diff easier to read.

P.S. This also adds an `isort` make target for future ease.

### DIFF
--- a/ecommerce/core/migrations/0033_auto_20170606_0539.py
+++ b/ecommerce/core/migrations/0033_auto_20170606_0539.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 from django.db import migrations
+
 from ecommerce.core.constants import USER_LIST_VIEW_SWITCH
 
 
--- a/ecommerce/core/tests/test_generate_courses.py
+++ b/ecommerce/core/tests/test_generate_courses.py
@@ -1,4 +1,5 @@
 import json
+
 import ddt
 import httpretty
 import mock
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -17,7 +17,8 @@ from ecommerce.coupons.tests.mixins import CouponMixin, DiscoveryMockMixin
 from ecommerce.coupons.views import voucher_is_valid
 from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
 from ecommerce.enterprise.utils import (
-    get_enterprise_course_consent_url, get_enterprise_customer_data_sharing_consent_token
+    get_enterprise_course_consent_url,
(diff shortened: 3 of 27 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1469"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

UmanShahzad:
@clintonb This is not a significant change -- can you give a :+1: if it looks fine?

Makefile
```diff
@@ -61,6 +61,9 @@ quality:
 	pep8 --config=.pep8 ecommerce e2e
 	pylint -j 0 --rcfile=pylintrc ecommerce e2e
 
+isort:
```
clintonb: The quality target already exists. Remove this

clintonb (changes requested):
The make target is not necessary. I am generally in favor of better documentation rather than more make targets to obfuscate the commands. 

Update your commit message better describe what actually occurred: Ran isort on entire codebase

Makefile
UmanShahzad: The quality target applies `--check-only`. This doesn't. It's a fast way to run isort on the entire codebase, instead of having to type out everything.

clintonb: I understand. Again, documentation over convenience. This is not an issue for other repos. The only reason you found isort issues here is due to the manner in which the tests are run.

clintonb: Correction: I fixed the issues with test runs a while ago. All of the changes you've made are purely cosmetic; `isort` isn't failing.

I stand by my statement that devs should understand exactly what they are running rather than rely on Make targets for everything. I'll approve the cosmetic changes. I have yet to hear a convincing argument to add another Make target.

UmanShahzad: As part of #1468, I added/removed some imports and I knew that caused `isort` to fail in one of the test runs. I then called `isort --recursive e2e ecommerce` assuming it'd solve my problem, but it instead generated this much larger changeset. So I figured it'd be good to not slow developers down by `isort` issues by allowing a convenience command to solve it automatically, and opened a separate PR for that purpose, that is, this one.

I'm OK with removing this as a make target, but we should consider a weekly/monthly bot that runs `isort` on the codebase. It's not important, but it allows individual developers to fix their PR-local `isort` quality issues without much effort and without acquiring additional changeset overhead, which could potentially confuse reviewers. Just a recommendation.

UmanShahzad:
@clintonb The make target is removed & commits squashed to your recommended commit message.

clintonb (approved):

PR #1470, opened 2017-09-06
Pre Django 1.11 updates

ahsan-ul-haq:
This includes the work involve in upgrading the libraries to support django 1.11
[LEARNER-1306](https://openedx.atlassian.net/browse/LEARNER-1306)


requirements/base.txt
```diff
@@ -1,23 +1,23 @@
...
+django-compressor==2.2
 django-crispy-forms==1.6.1
-django_extensions==1.5.5
-django-filter==1.0.1
+django_extensions==1.9.0
+django-filter==1.0.4
 django-libsass==0.5
-django-oscar==1.4
+django-oscar==1.5
 django-rest-swagger[reST]==0.3.10
```
clintonb: This needs to be updated. Breaking Swagger for edx/course-discovery affected some users. Let's make sure we don't repeat that .

schenedx (approved):
👍  LGTM
After you merge, could you do some manual testing on stage?
Remember to squash your commits!

ahsan-ul-haq:
@clintonb please review

requirements/test.txt
```diff
@@ -5,6 +5,7 @@ coverage==4.3.4
 ddt==1.1.1
 diff-cover==0.9.6
 django-nose==1.4.4
+django-webtest==1.9.2
```
clintonb, line 8: What is this, and why is it necessary?

ecommerce/settings/base.py
```diff
@@ -232,7 +233,6 @@
     'ecommerce.theming.middleware.CurrentSiteThemeMiddleware',
     'ecommerce.theming.middleware.ThemePreviewMiddleware',
 )
-# END MIDDLEWARE CONFIGURATION
```
clintonb, line 235: Restore this.

ecommerce/settings/base.py
```diff
@@ -211,6 +211,7 @@
 
 # MIDDLEWARE CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#middleware-classes
+# END MIDDLEWARE CONFIGURATION
```
clintonb, line 214: Remove this.

ecommerce/extensions/order/migrations/0016_auto_20170907_1306.py
```diff
@@ -0,0 +1,25 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='historicalline',
+            name='history_change_reason',
+            field=models.CharField(max_length=100, null=True),
+        ),
+        migrations.AddField(
+            model_name='historicalorder',
```
clintonb, line 21: How big is this table?

ecommerce/extensions/basket/views.py
```diff
@@ -451,12 +451,7 @@ def apply_voucher_to_basket(self, voucher):
             if discount['voucher'] and discount['voucher'] == voucher:
                 found_discount = True
                 break
-        if not found_discount:
-            messages.warning(
```
clintonb, line 455: Why was this removed?

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -800,9 +801,8 @@ def test_voucher_has_no_discount_error_msg(self):
         """ Verify correct error message is returned when voucher has no discount. """
         self.mock_access_token_response()
         self.mock_account_api(self.request, self.user.username, data={'is_active': True})
-        __, product = prepare_voucher(code=COUPON_CODE, benefit_value=0)
-        self.basket.add_product(product)
-        self.assert_form_valid_message("Your basket does not qualify for a coupon code discount.")
+        with self.assertRaises(ValidationError):
```
clintonb, line 804: Revert. The desired functionality is to display an error message, not raise an exception message that is not helpful to the user.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -978,5 +978,5 @@ def test_post_with_missing_voucher(self):
         self.assertEqual(response.status_code, 302)
 
         actual = list(get_messages(request))[-1].message
-        expected = "No coupon found with id '{}'".format(pk)
+        expected = "No voucher found with id '{}'".format(pk)
```
clintonb, line 981: Revert. This is probably an issue with translations.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -835,7 +835,7 @@ def test_voucher_not_valid_for_bundle(self):
         """ Verify correct error message is returned when voucher is used against a bundle. """
         self.mock_access_token_response()
         self.mock_account_api(self.request, self.user.username, data={'is_active': True})
-        voucher, product = prepare_voucher(code=COUPON_CODE, benefit_value=0)
+        voucher, product = prepare_voucher(code=COUPON_CODE, benefit_value=10)
```
clintonb, line 838: Why was this changed?

ecommerce/extensions/api/tests/test_views.py
```diff
@@ -0,0 +1,31 @@
...
+from django.test import RequestFactory, TestCase
+from django.urls import reverse
+from oscar.test.factories import UserFactory
+
+from ecommerce.extensions.api.views import api_docs_permission_denied_handler
+
+
+@ddt.ddt
+class ApiDocsPermissionDeniedHandlerTests(TestCase):
+    def setUp(self):
```
clintonb, line 13: Ditch the `setUp` method. There is nothing special about any of the variables you're making here. Make them in the test cases. We tend to do too much in `setUp`, making tests difficult to debug later.

ecommerce/extensions/api/tests/test_views.py
```diff
@@ -0,0 +1,31 @@
...
+        super(ApiDocsPermissionDeniedHandlerTests, self).setUp()
+        self.request_path = '/'
+        self.request = RequestFactory().get(self.request_path)
+
+    def test_authenticated(self):
+        """ Verify the view raises `PermissionDenied` if the request is authenticated. """
+        user = UserFactory()
```
clintonb, line 20: This doesn't need to be a separate variable.

ecommerce/extensions/api/tests/test_views.py
```diff
@@ -0,0 +1,31 @@
...
+
+@ddt.ddt
+class ApiDocsPermissionDeniedHandlerTests(TestCase):
+    def setUp(self):
+        super(ApiDocsPermissionDeniedHandlerTests, self).setUp()
+        self.request_path = '/'
+        self.request = RequestFactory().get(self.request_path)
+
+    def test_authenticated(self):
+        """ Verify the view raises `PermissionDenied` if the request is authenticated. """
```
clintonb, line 19: Permission denied errors should be raised if the user lacks the appropriate permissions, not if the user is authenticated. This docstring is more confusing than helpful. Please update it to reflect what is actually going on.

ecommerce/extensions/api/views.py
```diff
@@ -0,0 +1,55 @@
...
+    renderer_classes = [
+        CoreJSONRenderer,
+        OpenAPIRenderer,
+        SwaggerUIRenderer,
+    ]
+
+    exclude_from_schema = True
+
+    def get(self, request):
+        generator = SchemaGenerator(title='Discovery API')
```
clintonb, line 45: This is E-Commerce.

clintonb (changes requested):
You've changed some of the functionality that should have been changed. Also, some of your migrations may touch large tables and result in downtime. Please work with @edx/devops to test the effects of these migrations.

For the sake of time and simplicity, it may be better to split this PR into two parts:

1. Upgrade everything except Oscar
2. Upgrade Oscar

ecommerce/extensions/basket/tests/test_views.py
ahsan-ul-haq: Because having this value as 0 will cause validation error in Oscar 1.5

ecommerce/extensions/basket/views.py
ahsan-ul-haq: this is now being handled in the oscar that will generate validation error. If we want to have this I can override that behavior to not raise the error.

ecommerce/extensions/order/migrations/0016_auto_20170907_1306.py
ahsan-ul-haq: Are you refering to number of record in this table?

clintonb: Yes.

requirements/test.txt
ahsan-ul-haq: Its a required package for testing in Oscar 1.5

PR #1471, opened 2017-09-08
LEARNER-2456 Added loggings to track Missing Orders

Ayub-Khan:
**[LEARNER-2456](https://openedx.atlassian.net/browse/LEARNER-2456)**

Added loggings so that if there is missing order issue we can track it using the splunk alerts.

### DIFF
--- a/ecommerce/extensions/payment/views/cybersource.py
+++ b/ecommerce/extensions/payment/views/cybersource.py
@@ -43,6 +43,7 @@ NoShippingRequired = get_class('shipping.methods', 'NoShippingRequired')
 Order = get_model('order', 'Order')
 OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
 OrderTotalCalculator = get_class('checkout.calculators', 'OrderTotalCalculator')
+PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
 
 
 class CyberSourceProcessorMixin(object):
@@ -294,12 +295,21 @@ class CybersourceNotificationMixin(CyberSourceProcessorMixin, OrderCreationMixin
                 )
                 raise
             except DuplicateReferenceNumber:
-                logger.info(
-                    'Received CyberSource payment notification for basket [%d] which is associated '
-                    'with existing order [%s]. No payment was collected, and no new order will be created.',
-                    basket.id,
-                    order_number
-                )
+                if Order.objects.filter(number=order_number).exists() or PaymentProcessorResponse.objects.filter(
+                        basket=basket).exclude(transaction_id__isnull=True).exclude(transaction_id='').exists():
+                    logger.info(
+                        'Received CyberSource payment notification for basket [%d] which is associated '
+                        'with existing order [%s] or had an existing valid payment notification. '
+                        'No payment was collected, and no new order will be created.',
+                        basket.id,
+                        order_number
+                    )
+                else:
+                    logger.info(
+                        'Received duplicate CyberSource payment notification for basket [%d] which is not associated '
+                        'with any existing order (Missing Order Issue)',
+                        basket.id,
+                    )
                 raise
             except PaymentError:
                 logger.exception(
--- a/ecommerce/extensions/payment/tests/mixins.py
+++ b/ecommerce/extensions/payment/tests/mixins.py
@@ -630,6 +630,55 @@ class CybersourceNotificationTestsMixin(CybersourceMixin):
 
         self.assertRedirects(response, expected_redirect, fetch_redirect_response=False)
 
+    @mock.patch('ecommerce.extensions.payment.models.PaymentProcessorResponse.objects')
+    @mock.patch('ecommerce.extensions.order.models.Order.objects')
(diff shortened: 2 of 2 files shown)


ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -294,12 +294,20 @@ def validate_notification(self, notification):
                 )
                 raise
             except DuplicateReferenceNumber:
-                logger.info(
-                    'Received CyberSource payment notification for basket [%d] which is associated '
-                    'with existing order [%s]. No payment was collected, and no new order will be created.',
-                    basket.id,
-                    order_number
-                )
+                order = Order.objects.get(number=order_number)
```
mjfrey: This will throw an exception if no order exists.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -294,12 +294,20 @@ def validate_notification(self, notification):
...
                 raise
             except DuplicateReferenceNumber:
-                logger.info(
-                    'Received CyberSource payment notification for basket [%d] which is associated '
-                    'with existing order [%s]. No payment was collected, and no new order will be created.',
-                    basket.id,
-                    order_number
-                )
+                try:
+                    Order.objects.get(number=order_number)
```
schenedx: @Ayub-Khan Why not do
`if Order.objects.filter(number=order_number).exists():`? This way, you don't need to do try catch

schenedx (changes requested):
You would also need to exclude the case where there already exists of a paymentprocessorresponse row with a valid transaction_id

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -294,12 +295,20 @@ def validate_notification(self, notification):
                 )
                 raise
             except DuplicateReferenceNumber:
-                logger.info(
-                    'Received CyberSource payment notification for basket [%d] which is associated '
-                    'with existing order [%s]. No payment was collected, and no new order will be created.',
-                    basket.id,
-                    order_number
-                )
+                if Order.objects.filter(number=order_number).exists() and \
```
schenedx: Should be `or` in the if condition?

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -294,12 +295,20 @@ def validate_notification(self, notification):
...
+                        PaymentProcessorResponse.objects.filter(basket=basket).exclude(transaction_id__isnull=True):
+                    logger.info(
+                        'Received CyberSource payment notification for basket [%d] which is associated '
+                        'with existing order [%s]. No payment was collected, and no new order will be created.',
+                        basket.id,
+                        order_number
+                    )
+                else:
+                    logger.info(
+                        'Received CyberSource payment notification for basket [%d] which is not associated '
```
schenedx: `Received duplicate Cybersource payment notification for basket...`

schenedx (changes requested):
Question on the logic

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -294,12 +295,22 @@ def validate_notification(self, notification):
...
-                logger.info(
-                    'Received CyberSource payment notification for basket [%d] which is associated '
-                    'with existing order [%s]. No payment was collected, and no new order will be created.',
-                    basket.id,
-                    order_number
-                )
+                if Order.objects.filter(number=order_number).exists() or \
+                        PaymentProcessorResponse.objects.filter(
+                            basket=basket
+                        ).exclude(transaction_id__isnull=True).exclude(transaction_id=""):
```
schenedx: #Nitpick please use single quote to keep the file consistent.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -294,12 +295,22 @@ def validate_notification(self, notification):
...
+                    logger.info(
+                        'Received CyberSource payment notification for basket [%d] which is associated '
```
schenedx: This log message is no longer accurate with the `OR` condition above. Can you update the message?

schenedx (approved):
👍

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -294,12 +295,23 @@ def validate_notification(self, notification):
...
-                    'with existing order [%s]. No payment was collected, and no new order will be created.',
-                    basket.id,
-                    order_number
-                )
+                if Order.objects.filter(number=order_number).exists() or \
+                        PaymentProcessorResponse.objects.filter(
+                            basket=basket
+                        ).exclude(transaction_id__isnull=True).exclude(transaction_id=''):
+                    logger.info(
+                        'Received CyberSource payment notification for basket [%d] which is associated '
```
schenedx: `Received CyberSource duplicate payment notification for basket....or had an existing valid payment notification.`

mjfrey (approved):

ahsan-ul-haq (approved):
:+1:

PR #1474, opened 2017-09-12
[ENT-587] Don't cache invalid enrollment data on API error.

UmanShahzad:
This is a quick fix of https://github.com/edx/ecommerce/pull/1468 to avoid caching invalid enrollment data if we get an API error.

See https://github.com/edx/ecommerce/pull/1468/files#r138332927 for original discussion.

@haikuginger @clintonb -- please double-check that this logic is fine.

### DIFF
--- a/ecommerce/programs/conditions.py
+++ b/ecommerce/programs/conditions.py
@@ -84,10 +84,9 @@ class ProgramCourseRunSeatsCondition(Condition):
                 user = basket.owner.username
                 try:
                     enrollments = api.enrollment.get(user=user)
+                    cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
                 except (ConnectionError, SlumberBaseException, Timeout) as exc:
                     logger.error('Failed to retrieve enrollments: %s', str(exc))
-                else:
-                    cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
 
         for course in program['courses']:
             # If the user is already enrolled in a course, we do not need to check their basket for it


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fecommerce&number=1474"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

ecommerce/programs/conditions.py
```diff
@@ -86,7 +86,7 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
                     enrollments = api.enrollment.get(user=user)
                 except (ConnectionError, SlumberBaseException, Timeout) as exc:
                     logger.error('Failed to retrieve enrollments: %s', str(exc))
-                else:
+                if enrollments:
                     cache.set(cache_key, enrollments, settings.ENROLLMENT_API_CACHE_TIMEOUT)
```
clintonb: Move this line inside the `try` block.

clintonb (commented):
Remember to update the tests.

ecommerce/programs/conditions.py
UmanShahzad: Ah, right.

UmanShahzad:
@clintonb No need to actually update the tests I believe; what we have already covers the scenarios.

clintonb:
Okay.

clintonb (approved):

haikuginger:
I don't understand how this changes the behavior at all. The try/except/else block you had initially will only evaluate the `else` block if an exception isn't raised; essentially, the same behavior as putting the cache in the try block after the statement that might raise. @clintonb, was there some specific condition (indicating invalid data) that you wanted to be handled other than the possibility of an HTTP error code?

UmanShahzad:
@haikuginger Yeah, I just realized that myself. ._.

clintonb:
(facepalm) I misread the `else` as `finally`. For clarity's sake, let's move forward with this change, please.

UmanShahzad:
@haikuginger @clintonb Just going to amend the commit, push, and merge. Thanks!

PR #1475, opened 2017-09-13
ENT-638 Enterprise Offers

douglashall:
This PR adds the Enterprise offers feature. An Enterprise offer consists of a custom Condition model which checks that the user is linked to the specified EnterpriseCustomer, the course run(s) exist in that EnterpriseCustomer's catalogs, and the user has granted consent to share data for the course run(s).

Enterprise offers can be set up in the ecommerce administration tool.

ENT-638

**Testing Instructions:**
1. Visit [Enterprise Offers Admin](https://ecommerce-business.sandbox.edx.org/enterprise/offers) and verify that you can successfully create an Enterprise Offer using the existing [EnterpriseCustomerCatalogs](https://business.sandbox.edx.org/admin/enterprise/enterprisecustomercatalog/).
2. Visit https://business.sandbox.edx.org/enterprise/76afe510-a9fd-4b59-9ba1-49b537864205/course/course-v1:Microsoft+DAT206x+2017_T4/enroll/ as an unauthenticated user.
3. Log in through TestShib.
4. Verify that the course enrollment landing page displays the 50% Enterprise Offer discount.
5. Select Verified and click continue.
6. Verify the basket page displays the 50% Enterprise Offer discount.


ecommerce/enterprise/tests/mixins.py
```diff
@@ -174,9 +174,12 @@ def mock_enterprise_learner_api(
...
                         {
-                            'user': 1,
-                            'state': 'enabled' if consent_provided else 'disabled',
-                            'enabled': consent_provided
+                            'username': 'verified',
+                            'enterprise_customer_uuid': enterprise_customer_uuid,
+                            'exists': consent_provided,
+                            'course_id': 'course-v1:FooX+FooX+Demo_Course',
+                            'consent_provided': consent_provided,
+                            'consent_required': consent_enabled
```
UmanShahzad: Ah, looks like I'm also doing this change over at https://github.com/edx/ecommerce/pull/1494/files, although I have some required differences -- up to you if you want to use what I have and put it in this PR or let that PR handle it, I'm neutral. Unless there's something I got wrong with my changes?

ecommerce/enterprise/offer/benefits.py
```diff
@@ -0,0 +1,28 @@
...
+class EnterprisePercentageDiscountBenefit(BenefitWithoutRangeMixin, PercentageDiscountBenefit):
+    """ Enterprise-related PercentageDiscountBenefit without an attached range. """
+
+    class Meta(object):
+        app_label = 'enterprise'
+        proxy = True
+
+    @property
+    def name(self):
+        return _('{value}% enterprise discount').format(value=self.value)
```
UmanShahzad: I believe this can be safely removed because `BenefitWithoutRangeMixin`'s version would output the same thing. (Of course not true for the others).

ecommerce/enterprise/tests/mixins.py
UmanShahzad: Additionally, mine also changes a doc in `ecommerce/enterprise/api.py`.

ecommerce/enterprise/offer/conditions.py
```diff
@@ -0,0 +1,98 @@
...
+
+        enterprise_customer = learner_data['enterprise_customer']
+        if self.enterprise_customer_uuid != enterprise_customer['uuid']:
+            # Learner is not linked to the EnterpriseCustomer associated with this condition.
+            return False
+
+        course_runs_with_consent = self._get_course_runs_with_consent(learner_data['data_sharing_consent_records'])
+        course_run_ids = []
+        for line in basket.all_lines():
+            product = line.product
```
UmanShahzad: I see that `product` isn't used elsewhere below, so maybe we can squish things into `course = line.product.course` to be succinct.

UmanShahzad:
I like where this is headed and I think it should be enough back-end logic for the front-end work to come soon.

ecommerce/programs/conditions.py
```diff
@@ -141,18 +142,3 @@ def get_applicable_lines(self, offer, basket, most_expensive_first=True):
...
-
-        Args:
-            offer (AbstractConditionalOffer)
-            basket (AbstractBasket)
-            affected_lines (tuple[]): The lines that have been affected by the discount.
-                This should be list of tuples (line, discount, qty)
-        """
-        for line, __, __ in affected_lines:
-            quantity_to_consume = min(line.quantity_without_discount, 1)
-            line.consume(quantity_to_consume)
```
UmanShahzad: Just a nitty organizational suggestion: would it be worth moving `programs/condition.py` and `programs/benefits.py` into an `program/offer` folder, the way you've started for `enterprise/offer`? Or maybe since this existed beforehand, remove `enterprise/offer`? Following one way or the other across these apps helps a lot for seeing the landscape and similarities among them at a glance.

ecommerce/enterprise/offer/conditions.py
```diff
@@ -0,0 +1,98 @@
...
+                site.domain,
+                user.username,
+            )
+            return False
+        except IndexError:
+            # Learner is not linked to any EnterpriseCustomer.
+            return False
+
+        enterprise_customer = learner_data['enterprise_customer']
+        if self.enterprise_customer_uuid != enterprise_customer['uuid']:
```
UmanShahzad: We should change the left-hand side to `self.enterprise_customer['id']` since that's what our `ecommerce.enterprise.utils.get_enterprise_customer` function returns in the dictionary, and we presumably have access to `self.enterprise_customer` from the above `name` property.

ecommerce/enterprise/offer/conditions.py
```diff
@@ -0,0 +1,98 @@
...
+                # Basket contains products not related to a course_run.
+                return False
+
+            if enterprise_customer['enable_data_sharing_consent'] and course.id not in course_runs_with_consent:
+                # Basket contains course_runs for which the learner has not given consent to share data.
+                return False
+
+            course_run_ids.append(course.id)
+
+        if not catalog_contains_course_runs(basket.site, course_run_ids, self.enterprise_customer_uuid,
```
UmanShahzad: Same here as above with the enterprise customer UUID. But I think we can leave the catalog's UUID alone.

ecommerce/enterprise/offer/benefits.py
douglashall: Good call.

ecommerce/enterprise/offer/conditions.py
douglashall: Yup.

ecommerce/enterprise/tests/mixins.py
douglashall: Yeah, we can work it out as we get closer to merging either this or that.

ecommerce/programs/conditions.py
douglashall: Yes, consistency is good. I kind of like the enterprise/offer directory structure, but I think it would be better to match what we already have for expediency's sake.

ecommerce/enterprise/offer/conditions.py
douglashall: Actually, we don't have access to the enterprise_customer from the Condition object. I am going to denormalize and add both `enterprise_customer_uuid` and `enterprise_customer_name` fields to the Condition model since we won't be able to get the enterprise_customer_name using the enterprise_customer_uuid without the Site (in order to construct an API client) and we don't have access to the Site here either.

UmanShahzad: Yeah, you're right. That's why tests are failing on #1497 -- I tried putting `enterprise_customer` on the condition factory :)

ecommerce/enterprise/offer/conditions.py
UmanShahzad: Nevermind.

ecommerce/enterprise/offer/conditions.py
UmanShahzad: With that sorted out we should be well on our way, I believe, minus testing / any outstanding review for this overall work.

ecommerce/enterprise/forms.py
```diff
@@ -0,0 +1,124 @@
...
+            if hasattr(self.instance, 'condition'):
+                self.instance.condition.enterprise_customer_uuid = enterprise_customer_uuid
+                self.instance.condition.enterprise_customer_catalog_uuid = enterprise_customer_catalog_uuid
+                self.instance.condition.save()
+            else:
+                self.instance.condition = create_condition(
+                    EnterpriseCustomerCondition,
+                    enterprise_customer=enterprise_customer,
+                    enterprise_customer_catalog_uuid=enterprise_customer_catalog_uuid,
+                )
```
UmanShahzad: @douglashall This is where things crash, AFAIK. The condition expects `enterprise_customer_name`, `enterprise_customer_uuid`, and `enterprise_customer_catalog_uuid`, so we need to pass those in correctly. This reflects the old state of things before we decided to use `enterprise_customer_name` directly. The same of course goes for the `if` part of the branch -- there only `enterprise_customer_name` remains.

douglashall: Oh wait, I have fixed that already. Let me push.

ecommerce/extensions/offer/models.py
```diff
@@ -324,6 +324,22 @@ def all_products(self):
 
 
 class Condition(AbstractCondition):
+    enterprise_customer_uuid = models.UUIDField(
+        null=True,
+        blank=True,
+        verbose_name=_('EnterpriseCustomer UUID')
+    )
+    # De-normalizing the EnterpriseCustomer name for optimization purposes.
+    enterprise_customer_name = models.UUIDField(
```
UmanShahzad: ^^^ @douglashall It's being saved as a UUID

douglashall:
@mjfrey Any chance you could review this? I appreciate that the diff is large and I have attempted to separate the work into logical commits. Let me know if there is anything I can do to make the review go smoother for you.

douglashall:
@mjfrey I have opened a new PR containing just the first commit from this branch as you requested. The second two commits on this branch depend on that first commit.

ecommerce/invoice/migrations/0005_auto_20171003_1029.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.7 on 2017-10-03 14:29
+from __future__ import unicode_literals
+
+import django_extensions.db.fields
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
mjfrey: What is the reason for this change?  Have we also tested how long this might take in production to run?

ecommerce/extensions/refund/migrations/0003_auto_20171003_1029.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.7 on 2017-10-03 14:29
+from __future__ import unicode_literals
+
+import django_extensions.db.fields
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
mjfrey: Same here.  Why this change.  This table is quite large in production.

ecommerce/invoice/migrations/0005_auto_20171003_1029.py
douglashall: This must be a previously missed migration that got picked up by running `makemigrations`. Have we been ignoring the missing migrations warning on purpose?

ecommerce/extensions/refund/migrations/0003_auto_20171003_1029.py
UmanShahzad: I think they just came along when running `makemigrations`.

ecommerce/invoice/migrations/0005_auto_20171003_1029.py
mjfrey: I do not see missing migrations.  We should remove these.

douglashall: @mjfrey Sounds good to me. I will do that now.

douglashall:
@mjfrey How are we looking here? Do you think we'll be able to merge this today?

mjfrey (approved):
LGTM.  What is the plan to phase out the Coupon based offers and use this?  Are you planning on removing the old code?

douglashall:
@mjfrey Thanks!

I think we still need to support the ability to associate an EnterpriseCustomer to a coupon, but there is followup work here to modify the coupon admin tool to create an Enterprise Offer. I will write up that story now and get it prioritized in an upcoming sprint.

PR #1477, opened 2017-09-15
LEARNER-2355 Make address, suite, state and zip optional

MatthewPiatetsky:
**Background**
We have consulted with cybersource and internally with enterprise/finance/legal and a number of other teams about those 4 fields on the checkout form.
We determined there is some use for them, especially reducing risk of fraud and credit card declines, but we also believe there is a potentially greater benefit of increased revenue. We know how to measure the revenue and declined cards and want to a/b test a new checkout page that would all fit on one desktop screen with fewer fields against our current page.

**This PR**
This PR would enable the variation that will make these fields optional to not pass them, while preserving current behavior for the other variation.

An optimizely flag would result in the same behavior for half the users by preserving the client side validation.
(Only question is zip code. Currently it is required server side if US/CA is selected, but is not marked as required client side and there is no client side validation. If necessary, I can add a client side requirement for it when US/CA are selected like is currently done for state)

**Optimizely**
The optimizely test would look roughly like [this.](https://app.optimizely.com/v2/projects/1743970571/experiments/8728653307)
Of course we would test further first before turning on

### DIFF
--- a/ecommerce/extensions/payment/forms.py
+++ b/ecommerce/extensions/payment/forms.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import logging
 
 import pycountry
+import waffle
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import HTML, Div, Layout
 from django import forms
@@ -93,7 +94,7 @@ class PaymentForm(forms.Form):
     basket = forms.ModelChoiceField(queryset=Basket.objects.all(), widget=forms.HiddenInput(), required=False)
     first_name = forms.CharField(max_length=60, label=_('First Name'))
     last_name = forms.CharField(max_length=60, label=_('Last Name'))
-    address_line1 = forms.CharField(max_length=60, label=_('Address'))
+    address_line1 = forms.CharField(max_length=60, label=_('Address'), required=False)
     address_line2 = forms.CharField(max_length=29, required=False, label=_('Suite/Apartment Number'))
     city = forms.CharField(max_length=32, label=_('City'))
     # max_length for state field is set to default 60, if it needs to be changed,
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/payment/forms.py
```diff
@@ -110,28 +110,37 @@ def clean(self):
...
-            # Ensure that a valid 2-character state/province code is specified.
-            if not state:
-                raise ValidationError({'state': _('This field is required.')})
-
-            code = '{country}-{state}'.format(country=country, state=state)
+            # Temporarily comment out below code that requires state while running this test
+            # https://openedx.atlassian.net/browse/LEARNER-2355
+            # State will still be required client side for users not in the hide location fields variation
 
-            try:
```
MatthewPiatetsky: this is the same below just inside the if statement

ecommerce/extensions/payment/forms.py
```diff
@@ -110,28 +110,37 @@ def clean(self):
...
+                    logger.debug(msg)
+                    raise ValidationError({'state': msg})
 
             # Ensure the postal code is present, and limited to 9 characters
             postal_code = cleaned_data.get('postal_code')
-            if not postal_code:
-                raise ValidationError({'postal_code': _('This field is required.')})
+            # Temporarily comment out below code that requires postal code while running this test
+            # https://openedx.atlassian.net/browse/LEARNER-2355
+            # I could require it client side, but am not because it is not currently marked as required for the user
```
mikedikan: Do I understand this to mean that we don't CURRENTLY have code to validate zip client-side?

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -210,7 +210,7 @@ def _get_billing_address(self, cybersource_response):
         return BillingAddress(
             first_name=cybersource_response['req_bill_to_forename'],
             last_name=cybersource_response['req_bill_to_surname'],
-            line1=cybersource_response['req_bill_to_address_line1'],
+            line1=cybersource_response.get('req_bill_to_address_line1', ''),
```
mikedikan: Worth leaving a comment here about the test?

mikedikan (commented):
Just a cup'l questions

ecommerce/extensions/payment/forms.py
MatthewPiatetsky: correct

ecommerce/extensions/payment/views/cybersource.py
MatthewPiatetsky: added comment

ecommerce/extensions/payment/forms.py
```diff
@@ -110,28 +110,37 @@ def clean(self):
         if country in ('US', 'CA'):
             state = cleaned_data.get('state')
 
-            # Ensure that a valid 2-character state/province code is specified.
-            if not state:
-                raise ValidationError({'state': _('This field is required.')})
-
-            code = '{country}-{state}'.format(country=country, state=state)
+            # Temporarily comment out below code that requires state while running this test
```
mjfrey: Rather than commenting this out, can we add a siteconfig option like "enable minimal payment fields: bool"?  This way we can turn it on / off for all the different sites. 

Once we feel it is good globally we can remove the siteconfig flag.

mjfrey (changes requested):

ecommerce/extensions/payment/forms.py
mikedikan: I think this makes sense once this goes to implementation, but right now it's just an A/B test and is more likely to be reverted completely than anything else.  @mjfrey  Can we save this feedback for when we have a successful A/B test and are moving towards productizing?

mjfrey: But doing it this way will force the test and behavior on all sites including the WL ones.  This could be an issue. Not that much extra to just add a siteconfig flag.

MatthewPiatetsky: i don't mind adding siteconfig flag but wondering if that's the best fix here.
optimizely test won't trigger on white label sites so if we wanted to preserve identical behavior from a user perspective we could add zip code to the list of required fields client side, and all the current required fields would remain required and behave the same.
this may make more sense anyway since the current state is either a product bug or implementation bug.

If the actual reason we wanted to add a flag is because we were concerned about this change breaking something and wanted to turn off the server side changes to make the fields optional though they are required client side, i'm wondering if a feature flag makes more sense.
in both cases we'd have to additional tests to pass codecov, but then we would not need to write a migration to add the field, remove the field, and squash migrations

mjfrey: Only issue with a feature flag, it is global.  A siteconfig flag is per site.

MatthewPiatetsky: but what case wouldn't be covered by globally identical client-side behavior and a global feature flag?

mjfrey: I guess I do not understand the proposal.  I am suggesting that we support the current behavior on some sites and the test on others.  Can the be done with a feature flag being global?

MatthewPiatetsky: my proposal would be to keep all the fields required client side on all sites other than edx.org to preserve their current behavior.
If we were concerned that someone would commit something breaking the client side code during the test, we could use a feature flag as a fallback safety measure to turn off the test changes until we can fix the bug.
then all the code would be reverted after the test is over

mjfrey: so you would modify the client side JS?  Isn't adding a siteconfig flag easier?  I did not mean for this to be a big deal, just that is needs to be configurable per site.

MatthewPiatetsky: this pr already modifies the client side js to preserve current behavior unless a flag an optimizely flag is triggered, which would only happen on edx.org
the one exception is that unlike the other fields which are required both client side and server side, zip code is only required server side.
so as part of my proposal, also requiring it client side would mean that users would mean that users won't be able to submit without any of the fields now optional server side
i think this would work unless white label sites use different javascript, in which case a siteconfiguration would be necessary
do white label sites use different javascript client side?

mjfrey: okay I understand.  I am not sure about WL sites modifying the JS.  I do not think they do but you should check the themes to see.

MatthewPiatetsky: i looked through themes and some of the actual sites, and the payment pages look the same to me. i've gone ahead and ensured that all fields including zip continue to be required client side. just in case, the server side changes are now also controlled by a waffle switch

mjfrey (approved):

PR #1479, opened 2017-09-18
Use GA client id on order completed event instead from DB.

waheedahmed:
Our client ID is stored in ecommerce database, and that won't change anymore even if the user is moved to a different browse. We should not use the client ID from the database instead we should get the client ID from the cookie of the request.

**Testing:**
    Tested locally and working as expected.

LEARNER-2232

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -118,7 +118,7 @@ def prepare_analytics_data(user, segment_key):
     return json.dumps(data)
 
 
-def track_segment_event(site, user, event, properties):
+def track_segment_event(site, user, event, properties, ga_client_id=None):
     """ Fire a tracking event via Segment.
 
     Args:
@@ -126,12 +126,12 @@ def track_segment_event(site, user, event, properties):
         user (User): User to which the event should be associated.
         event (str): Event name.
         properties (dict): Event properties.
+        ga_client_id (str): Google Analytics clientId.
 
     Returns:
         (success, msg): Tuple indicating the success of enqueuing the event on the message queue.
             This can be safely ignored unless needed for debugging purposes.
     """
-
     site_configuration = site.siteconfiguration
     if not site_configuration.segment_key:
         msg = 'Event [{event}] was NOT fired because no Segment key is set for site configuration [{site_id}]'
@@ -143,7 +143,7 @@ def track_segment_event(site, user, event, properties):
     context = {
         'ip': lms_ip,
         'Google Analytics': {
-            'clientId': lms_client_id
+            'clientId': ga_client_id if ga_client_id else lms_client_id
         }
     }
     return site.siteconfiguration.segment_client.track(user_tracking_id, event, properties, context=context)
@@ -171,3 +171,17 @@ def translate_basket_line_for_segment(line):
         'quantity': line.quantity,
         'category': line.product.get_product_class().name,
     }
+
+
+def get_google_analytics_client_id(request):
+    """Get google analytics client ID from request cookies."""
+    if not request:
+        return None
+
+    # Google Analytics uses the clientId to keep track of unique visitors. A GA cookie looks like
+    # this: _ga=GA1.2.1033501218.1368477899 and the clientId is this part: 1033501218.1368477899
+    google_analytics_cookie = request.COOKIES.get('_ga')
+    if google_analytics_cookie:
+        return '.'.join(google_analytics_cookie.split('.')[2:])
+
+    return None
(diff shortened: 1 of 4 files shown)


waheedahmed:
@edx/learner-spartans, @mulby Please review.

ecommerce/extensions/analytics/utils.py
```diff
@@ -171,3 +171,17 @@ def translate_basket_line_for_segment(line):
...
+
+def get_google_analytics_client_id(request):
+    """Get google analytics client ID from request cookies."""
+    if not request:
+        return None
+
+    # Google Analytics uses the clientId to keep track of unique visitors. A GA cookie looks like
+    # this: _ga=GA1.2.1033501218.1368477899 and the clientId is this part: 1033501218.1368477899
+    google_analytics_cookie = request.COOKIES.get('_ga')
+    if google_analytics_cookie:
```
mulby, line 184: The LMS falls back on a header, not sure why... but I wonder if we shouldn't implement the same logic here?

https://github.com/edx/edx-platform/blob/e8a36957b1f732974260e7b9b42b9c25148b492c/common/djangoapps/track/middleware.py#L148

ecommerce/extensions/checkout/signals.py
```diff
@@ -79,7 +83,9 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
     except BasketAttribute.DoesNotExist:
         logger.info('There is no program or bundle associated with order number %s', order.number)
 
-    track_segment_event(order.site, order.user, 'Order Completed', properties)
+    ga_client_id = get_google_analytics_client_id(kwargs.get('request'))
+
+    track_segment_event(order.site, order.user, 'Order Completed', properties, ga_client_id=ga_client_id)
```
mulby, line 88: I think we want to pass in "ga_client_id" on *every* call to `track_segment_event`

mulby (commented):
Approach looks good to me, I just think we need to roll it out more broadly in the code base.

ecommerce/extensions/analytics/utils.py
waheedahmed: @mulby We can't, because it's same client ID saved in ecommerce DB sent on enrollment API call from ecommerce. https://github.com/edx/ecommerce/blob/78ba7a8a960fb0eed4809da57e0d9d041995c542/ecommerce/extensions/fulfillment/modules.py#L118-L126

ecommerce/extensions/checkout/signals.py
waheedahmed: @mulby Yes we are, already discussed it with @schenedx and it was going out of scope of this ticket and decided to create a new ticket for this [LEARNER-2596](https://openedx.atlassian.net/browse/LEARNER-2596)

ecommerce/extensions/analytics/utils.py
mulby: Oh, I see.

Shouldn't this also read the client ID from the cookie and not the database and send *that* to the LMS?

ecommerce/extensions/checkout/signals.py
mulby: Ah - OK

mulby (approved):
LGTM - assuming other issues are addressed in a follow-on ticket.

waheedahmed:
@mulby Can we use [event-tracking](https://github.com/edx/event-tracking) in ecommerce to set client id in a tracker context from a middleware like you did in LMS instead of passing request to every call?

waheedahmed:
@mulby I also have another solution in which we can save `ga_client_id` in `user.tracking_context` and then get it from there like we get `lms_course_id` but it will increase DB calls.

ecommerce/extensions/analytics/utils.py
waheedahmed: Yes, it should.

mulby:
@waheedahmed - I wouldn't recommend using event-tracking just for this purpose. I do like the idea, however, of implementing middleware that stores the client ID in such a way that the tracking calls can pull it without having a reference to the request object. event-tracking just uses thread local storage for this purpose.

It looks like this project already uses threadlocals. You could use that? Or maybe just `request.session`? Or your current tracking_context in the DB?

Regardless of the storage mechanism I like the idea of:

* Create a middleware class that parses the cookie and stores the client ID somewhere that can be accessed without having a reference to the request object
* When emitting events and making requests to the LMS, pull from that storage location and use the client ID

schenedx (approved):
👍

waheedahmed:
@mulby You're right, we are already using threadlocals in ecommerce and we can get current request object directly from it with `get_current_request()` in `track_segment_event()`, I will check if it works then I'm going to use it.

PR #1480, opened 2017-09-18
change django admin title

tasawernawaz:
[LEARNER-2555](https://openedx.atlassian.net/browse/LEARNER-2555)

Please review @edx/learner-spartans


ecommerce/urls.py
```diff
@@ -36,6 +37,8 @@ def handler403(_):
 
 
 admin.autodiscover()
+admin.site.site_header = _('Ecommerce Administration')
```
clintonb: E-Commerce Service Administration

clintonb (approved):

PR #1482, opened 2017-09-19
Pre Django Oscar upgrades

ahsan-ul-haq:
LEARNER-1306


ahsan-ul-haq:
@clintonb I went through historical order table and apparently there are more than 56 million record in it. So what do you propose about running the migration on prod?

clintonb:
We will need to determine if those historical orders are needed, and clean out as much as we can. You should do similar data gathering for the other tables that are being affected. 

Also, again, split this PR up. You've done some great work that we can capture value from immediately. Let's do that, and worry about Oscar 1.5 separately.

ahsan-ul-haq:
@clintonb there no oscar updates in this PR.

clintonb:
Got it. I was under the impression the order changes were due to Oscar, not Simple History.

requirements/base.txt
```diff
@@ -1,23 +1,23 @@
...
 django-crispy-forms==1.6.1
-django_extensions==1.5.5
-django-filter==1.0.1
+django_extensions==1.9.0
+django-filter==1.0.4
 django-libsass==0.5
 django-oscar==1.4
-django-rest-swagger[reST]==0.3.10
-django-simple-history==1.8.2
+django-rest-swagger==2.1.2
```
clintonb: We can get rid of Swagger and use DRF's built-in documentation system. See https://github.com/edx/credentials/commit/acb0f2dd3265ceb6a0370c3e897dda32269937cf.

clintonb (commented):
We need to further split this PR. There is still too much risk due to the Simple History upgrade/migrations.

Proposed split
1. All of the simple package upgrades.
2. DRF + removal of Swagger
3. Simple History

I don't mean to make this process suck, and appreciate your taking on this effort, but we want to minimize the damage should one of these upgrades go wrong. The plan above allows us to change our parts in increments, rather than do a giant change and have to roll it all back.

clintonb:
This will help with the eventual removal of history data: https://github.com/treyhunner/django-simple-history/issues/198#issuecomment-208147874

clintonb:
Table | Row Count
-- | --
order_historicalorder | 56,686,139
order_historicalline | 19,416,406
refund_historicalrefund | 5,713,141
refund_historicalrefundline | 3,808,800
catalogue_historicalproductattributevalue | 62,940
catalogue_historicalproduct | 40,295
partner_historicalstockrecord | 19,866
courses_historicalcourse | 13,560
invoice_historicalinvoice | 2,713

```sql
SELECT
  (SELECT COUNT(*) FROM catalogue_historicalproduct) as catalogue_historicalproduct,
  (SELECT COUNT(*) FROM catalogue_historicalproductattributevalue) as catalogue_historicalproductattributevalue,
  (SELECT COUNT(*) FROM courses_historicalcourse) as courses_historicalcourse,
  (SELECT COUNT(*) FROM invoice_historicalinvoice) as invoice_historicalinvoice,
  (SELECT COUNT(*) FROM order_historicalline) as order_historicalline,
  (SELECT COUNT(*) FROM order_historicalorder) as order_historicalorder,
  (SELECT COUNT(*) FROM partner_historicalstockrecord) as partner_historicalstockrecord,
  (SELECT COUNT(*) FROM refund_historicalrefund) as refund_historicalrefund,
  (SELECT COUNT(*) FROM refund_historicalrefundline) as refund_historicalrefundline
```

We will need to clean history for anything with over 1 million rows.

ahsan-ul-haq:
@clintonb made the change to only have simple package upgrades

clintonb (approved):

PR #1484, opened 2017-09-20
Use GA clientId from request cookie on all analytics events instead of DB.

waheedahmed:
Our client ID is stored in ecommerce database, and that won't change anymore even if the user is moved to a different browser. We should not use the client ID from the database instead we should get the client ID from the cookie of the request.

LEARNER-2596

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -14,7 +14,7 @@ def parse_tracking_context(user):
         user (User): An instance of the User model.
 
     Returns:
-        Tuple of strings: user_tracking_id, lms_client_id, lms_ip
+        Tuple of strings: user_tracking_id, ga_client_id, lms_ip
     """
     tracking_context = user.tracking_context or {}
 
@@ -26,10 +26,10 @@ def parse_tracking_context(user):
         # at some point.
         user_tracking_id = 'ecommerce-{}'.format(user.id)
 
-    lms_client_id = tracking_context.get('lms_client_id')
     lms_ip = tracking_context.get('lms_ip')
+    ga_client_id = tracking_context.get('ga_client_id')
 
-    return user_tracking_id, lms_client_id, lms_ip
+    return user_tracking_id, ga_client_id, lms_ip
 
 
 def silence_exceptions(msg):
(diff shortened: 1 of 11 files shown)


waheedahmed:
@mulby @edx/learner-spartans Please review.

ecommerce/extensions/analytics/utils.py
```diff
@@ -26,10 +28,10 @@ def parse_tracking_context(user):
         # at some point.
         user_tracking_id = 'ecommerce-{}'.format(user.id)
 
-    lms_client_id = tracking_context.get('lms_client_id')
     lms_ip = tracking_context.get('lms_ip')
+    ga_client_id = get_google_analytics_client_id(get_current_request())
```
mulby: It just occurred to me while looking at this, but is this call ever made from a celery worker? If so, the request probably won't be available...

Actually, we might have a similar problem with the previous PR as well. If the request coming into the ecommerce service is coming from cybersource instead of the user directly, then the GA cookie won't be set properly on that request.

waheedahmed: @mulby In that case I think we should save it in user tracking context in DB from a middleware, what do you think?

waheedahmed: But if this call is coming from celery worker then clientId should be None, isn't it?

mulby: Using the DB makes sense.

RE: the celery worker. I think it depends. If the worker is doing something on behalf of the user than it makes sense to use the client ID for that user. You could imagine a scenario where we did all of the work while processing the HTTP request. Maybe that wasn't performant enough so we moved it out into a celery worker. In that case we should maintain the user's client ID. Does that make sense?

waheedahmed: @mulby Yes, completely make sense, thank you!

mulby:
@waheedahmed - I love this!

mulby (approved):

waheedahmed:
@mulby thank you! @edx/learner-spartans Can someone please take a second look on this PR?

schenedx (approved):
👍 Very nice
Please remember to squash your commits

PR #1486, opened 2017-09-22
(WIP) added the prevention of multiple-refunds for one day

Ayub-Khan:
Preventing Multiple Refunds for one day .


ecommerce/extensions/refund/models.py
```diff
@@ -287,6 +291,14 @@ def deny(self):
 
         return result
 
+    def _can_issue_refund(self):
+        """ Disable the multiple refund for one day if already exists a payment event"""
+        refund_event = PaymentEventType.objects.get(name=PaymentEventTypeName.REFUNDED)
```
schenedx: You don't need to make two database call here.
all you need is to do `event_type__name=PaymentEventTypeName.REFUNDED`

ecommerce/extensions/refund/models.py
```diff
@@ -287,6 +291,14 @@ def deny(self):
...
+        return not (PaymentEvent.objects.filter(
+            order=self.order,
+            date_created__range=[datetime.timedelta(days=1), datetime.datetime.now()],
```
schenedx: Please make this value configurable using a value in configuration model

ecommerce/extensions/refund/models.py
```diff
@@ -233,6 +234,9 @@ def _revoke_lines(self):
             self.set_status(REFUND.REVOCATION_ERROR)
 
     def approve(self, revoke_fulfillment=True, notify_purchaser=True):
+        if not self._can_issue_refund():
+            logger.info('Refund [%d] have a existing refund Payment Event in last day.', self.id)
```
schenedx: Update the comment with the time range from configuration model

schenedx (commented):
The general direction looks good to me.

Ayub-Khan:
moved to https://github.com/edx/ecommerce/pull/1490

PR #1490, opened 2017-09-26
Added Disable Multiple Refund Flag Config Model

Ayub-Khan:
New Pull Request for LEARNER-2161


ecommerce/extensions/refund/admin.py
```diff
@@ -21,3 +24,6 @@ class RefundAdmin(admin.ModelAdmin):
     fields = ('order', 'user', 'status', 'total_credit_excl_tax', 'currency', 'created', 'modified',)
     readonly_fields = ('order', 'user', 'total_credit_excl_tax', 'currency', 'created', 'modified',)
     inlines = (RefundLineInline,)
+
+
+admin.site.register(DisableMultipleRefund, ConfigurationModelAdmin)
```
clintonb: Add a newline.

ecommerce/extensions/refund/models.py
```diff
@@ -1,17 +1,14 @@
 from __future__ import unicode_literals
 
 import logging
+from datetime import datetime
 
 from django.conf import settings
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
-from django_extensions.db.models import TimeStampedModel
```
clintonb: Undo. See https://www.python.org/dev/peps/pep-0008/#imports for import ordering.

ecommerce/extensions/refund/models.py
```diff
@@ -44,7 +46,6 @@ def available_statuses(self):
     # pylint: disable=access-member-before-definition,attribute-defined-outside-init
     def set_status(self, new_status):
         """Set a new status for this object.
-
```
clintonb, line 47: Revert your changes to these docstrings. They are unnecessary and make the docstrings less legible.

ecommerce/extensions/refund/models.py
```diff
@@ -315,3 +328,17 @@ class RefundLine(StatusMixin, TimeStampedModel):
...
         return True
+
+
+class DisableMultipleRefund(ConfigurationModel):
+    """ When the flag is active disable Multiple Refund for specified duration of time"""
+
+    class Meta(ConfigurationModel.Meta):
+        app_label = "refund"
+
+    duration = models.PositiveIntegerField(
```
clintonb, line 339: Why is this necessary? Why not prevent multiple refunds always? I can't imagine anyone waiting a day to issue the refund again.

ecommerce/extensions/refund/models.py
```diff
@@ -315,3 +328,17 @@ class RefundLine(StatusMixin, TimeStampedModel):
     def deny(self):
         self.set_status(REFUND_LINE.DENIED)
         return True
+
+
+class DisableMultipleRefund(ConfigurationModel):
+    """ When the flag is active disable Multiple Refund for specified duration of time"""
+
+    class Meta(ConfigurationModel.Meta):
+        app_label = "refund"
```
clintonb, line 337: Single quotes

ecommerce/extensions/refund/models.py
```diff
@@ -315,3 +328,17 @@ class RefundLine(StatusMixin, TimeStampedModel):
...
+
+    duration = models.PositiveIntegerField(
+        default=1,
+        help_text="Number of Days you want to disable Multiple Refund.",
```
clintonb, line 341: Single quotes

requirements/base.txt
```diff
@@ -1,6 +1,7 @@
 analytics-python==1.2.7
 django==1.10.7
 django-compressor==2.2
+django-config-models==0.1.8
```
clintonb, line 4: Why config models rather than Waffle (to start) followed by a Django setting. This doesn't seem like functionality that will ever need to be disabled after the initial rollout.

clintonb (changes requested):

ecommerce/extensions/refund/models.py
Ayub-Khan: @schenedx ^^

requirements/base.txt
Ayub-Khan: This gives us more flexibility around duration we want to disable. in case of Django setting we will have to create a pull request if we want to change it.

clintonb:
I question the need for this work. This issue occurred one time because I manually modified a refund line. That was an exceptional case. The majority of our refunds are automated. We don't need more business logic here. We also don't need arbitrary waiting periods. If a refund failed, waiting a day won't magically make it work again.

Ayub-Khan:
@clintonb Yes i agree. But we are testing our condition against payment event so if there was not any payment event it will never trigger and allow refund again. But in case if payment event was there and still refund fail we have seen multiple cases in past where support refunded orders multiple times. so if we disable it for one day we will be able to counter all those loses and step in, in that one day time.

clintonb:
How much money has this issue cost us? How many refunds have been affected? I suspect both numbers are low.

The issue is not that we attempt to approve refunds multiple times. The issue is with the state model and how it responds to API errors/manual changes. Again, the JOC order mentioned in the ticket was an exceptional case hence the weird manual changes. I can vaguely think of only two other times where I have had to make such changes over the past two years.

Bluntly stated: Unless you can show me data demonstrating otherwise, I think we are wasting time working on LEARNER-2161. The amount of time you've worked on this PR and I've reviewed it have most likely already surpassed what we lost from duplicate refunds over the last two years.

schenedx:
@clintonb Good call for getting data to inform us.
I did a data pull in our ecommerce production database. 
The result is that, we had 5 multiple refund incidents
The total loss of money is $1941.15. That indeed is a very small loss amount. See my query below

@clintonb I thought this is a good change to make to prevent us from making mistakes above in the future. The historical small number won't prevent us getting bitten by this hole in the system in the future. Just like we joked, someone can certainly explore this system behavior to get themselves multiple refunds in a very short amount of time. 

Here is the query I used.
```
select sum(pe.amount), count(pe.id), oo.total_excl_tax from order_paymentevent pe
join order_order oo on oo.id = pe.order_id
where pe.event_type_id = 2
GROUP BY pe.order_id
having count(pe.id) > 0
and sum(pe.amount) > oo.total_excl_tax;
```

PR #1491, opened 2017-09-26
Make sure we clear any vouchers from basket when adding a bundle.

mjfrey:
Any vouchers that may be applied to the basket must be cleared when adding a bundle.

LEARNER-2683


mjfrey:
@edx/helio please take a look.

McKenzieW (approved):
LGTM 👍

staubina (approved):

PR #1493, opened 2017-09-28
LEARNER-2594 course names discrepancy for coupon redemption

Ayub-Khan:
[**LEARNER-2594**](https://openedx.atlassian.net/browse/LEARNER-2594)
Updated code to use the course title from the discovery. If unable to find it than use the old one.

### DIFF
--- a/ecommerce/extensions/api/v2/views/vouchers.py
+++ b/ecommerce/extensions/api/v2/views/vouchers.py
@@ -286,6 +286,6 @@ class VoucherViewSet(NonDestroyableModelViewSet):
             'credit_provider_price': credit_provider_price,
             'seat_type': product.attr.certificate_type,
             'stockrecords': serializers.StockRecordSerializer(stock_record).data,
-            'title': course.name,
+            'title': course_info.get('title', course.name),
             'voucher_end_date': voucher.end_datetime
         }


schenedx:
@mjfrey Please review

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -286,6 +286,6 @@ def get_course_offer_data(
             'credit_provider_price': credit_provider_price,
             'seat_type': product.attr.certificate_type,
             'stockrecords': serializers.StockRecordSerializer(stock_record).data,
-            'title': course.name,
+            'title': course_info.get('title', course.name),
```
schenedx, line 289: Should we default back to `course.name` if the title property has no value in `course_info`?

schenedx (commented):
Got a question here

ecommerce/extensions/api/v2/views/vouchers.py
Ayub-Khan: I think if we do not have any title its better to show course.name instead of empty string .

schenedx (approved):
👍

mjfrey (approved):

PR #1498, opened 2017-10-04
jbradley/learner 2413 Add Geolocation-Based Local Currency To Basket

jaebradley:
## Purpose

[JIRA](https://openedx.atlassian.net/browse/LEARNER-2413)

Translate basket prices into local currency. Local currency is defined by information stored in the `edx-price-l10n` cookie, which should be set beforehand.

cc: @MatthewPiatetsky @AlasdairSwan @mikedikan 

### Before
image

### No Cookie
image

### With Cookie (But `USA` Country Code)
image

### Cookie With Chinese Yen Values
image

## To Do
- [x] Check what happens when coupon is added
- [x] Update spec
- [x] Check about warning text on conversion accuracy
- [x] Squash commits

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -16,452 +16,492 @@ define([
               Cookies) {
         'use strict';
 
-        function hideVoucherForm() {
-            $('#voucher_form_container').hide();
-            $('#voucher_form_link').show();
-        }
+        var BasketPage = {
+            hideVoucherForm: function() {
+                $('#voucher_form_container').hide();
+                $('#voucher_form_link').show();
+            },
 
-        function onFail() {
-            var message = gettext('Problem occurred during checkout. Please contact support');
-            $('#messages').html(_s.sprintf('<div class="alert alert-error">%s</div>', message));
-        }
+            onFail: function() {
+                var message = gettext('Problem occurred during checkout. Please contact support.');
+                $('#messages').html(_s.sprintf('<div class="alert alert-error">%s</div>', message));
+            },
 
-        function onSuccess(data) {
-            var $form = $('<form>', {
-                class: 'hidden',
-                action: data.payment_page_url,
-                method: 'POST',
-                'accept-method': 'UTF-8'
-            });
+            onSuccess: function(data) {
+                var $form = $('<form>', {
+                    class: 'hidden',
+                    action: data.payment_page_url,
+                    method: 'POST',
+                    'accept-method': 'UTF-8'
+                });
 
-            _.each(data.payment_form_data, function(value, key) {
-                $('<input>').attr({
-                    type: 'hidden',
-                    name: key,
-                    value: value
-                }).appendTo($form);
-            });
+                _.each(data.payment_form_data, function(value, key) {
+                    $('<input>').attr({
+                        type: 'hidden',
+                        name: key,
+                        value: value
+                    }).appendTo($form);
+                });
 
-            $form.appendTo('body').submit();
-        }
+                $form.appendTo('body').submit();
+            },
 
-        function checkoutPayment(data) {
-            $.ajax({
-                url: '/api/v2/checkout/',
-                method: 'POST',
-                contentType: 'application/json; charset=utf-8',
-                dataType: 'json',
-                headers: {
-                    'X-CSRFToken': Cookies.get('ecommerce_csrftoken')
-                },
-                data: JSON.stringify(data),
-                success: onSuccess,
-                error: onFail
-            });
-        }
+            checkoutPayment: function(data) {
+                $.ajax({
+                    url: '/api/v2/checkout/',
+                    method: 'POST',
+                    contentType: 'application/json; charset=utf-8',
+                    dataType: 'json',
+                    headers: {
+                        'X-CSRFToken': Cookies.get('ecommerce_csrftoken')
+                    },
+                    data: JSON.stringify(data),
+                    success: BasketPage.onSuccess,
+                    error: BasketPage.onFail
+                });
+            },
 
-        function appendCardValidationErrorMsg(event, field, msg) {
-            field.find('~.help-block').append('<span>' + msg + '</span>');
-            field.focus();
-            event.preventDefault();
-            $('.payment-form').attr('data-has-error', true);
-        }
+            appendCardValidationErrorMsg: function(event, field, msg) {
+                event.preventDefault();
+                field.find('~.help-block').append('<span>' + msg + '</span>');
+                field.focus();
+                $('.payment-form').attr('data-has-error', true);
+            },
 
-        function appendCardHolderValidationErrorMsg(field, msg) {
-            field.parentsUntil('form-item').find('~.help-block').append(
-                '<span>' + msg + '</span>'
-            );
-        }
+            appendCardHolderValidationErrorMsg: function(field, msg) {
+                field.parentsUntil('form-item').find('~.help-block').append(
+                    '<span>' + msg + '</span>'
+                );
+            },
 
-        function cardHolderInfoValidation(event) {
-            var requiredFields = [
-                    'input[name=first_name]',
-                    'input[name=last_name]',
-                    'input[name=city]',
-                    'select[name=country]'
-                ],
-                experiments = window.experimentVariables || {};
-            // Only require address and state if we are not in the hide location fields variation of this experiment
-            // https://openedx.atlassian.net/browse/LEARNER-2355
-            if (!(experiments && experiments.hide_location_fields)) {
-                requiredFields.push('input[name=address_line1]');
-                if (['US', 'CA'].indexOf($('select[name=country]').val()) > -1) {
-                    requiredFields.push('select[name=state]');
-                    requiredFields.push('input[name=postal_code]');
+            cardHolderInfoValidation: function(event) {
+                var requiredFields = [
+                        'input[name=first_name]',
+                        'input[name=last_name]',
+                        'input[name=city]',
+                        'select[name=country]'
+                    ],
+                    countriesWithRequiredStateAndPostalCodeValues = ['US', 'CA'],
+                    experiments = window.experimentVariables || {};
+                // Only require address and state if we are not in the hide location fields variation of this experiment
+                // https://openedx.atlassian.net/browse/LEARNER-2355
+                if (!(experiments && experiments.hide_location_fields)) {
+                    requiredFields.push('input[name=address_line1]');
+                    if (countriesWithRequiredStateAndPostalCodeValues.indexOf($('select[name=country]').val()) > -1) {
+                        requiredFields.push('select[name=state]');
+                        requiredFields.push('input[name=postal_code]');
+                    }
                 }
-            }
 
-            _.each(requiredFields, function(field) {
-                if ($(field).val() === '') {
-                    event.preventDefault();
-                    appendCardHolderValidationErrorMsg($(field), gettext('This field is required'));
-                    $('.payment-form').attr('data-has-error', true);
-                }
-            });
+                _.each(requiredFields, function(field) {
+                    if ($(field).val() === '') {
+                        event.preventDefault();
+                        BasketPage.appendCardHolderValidationErrorMsg($(field), gettext('This field is required'));
+                        $('.payment-form').attr('data-has-error', true);
+                    }
+                });
 
-            // Focus the first element that has an error message.
-            $('.help-block > span')
-            .first().parentsUntil('fieldset').last()
-            .find('input')
-            .focus();
-        }
+                // Focus the first element that has an error message.
+                $('.help-block > span')
+                .first().parentsUntil('fieldset').last()
+                .find('input')
+                .focus();
+            },
 
-        function isCardTypeSupported(cardType) {
-            return $.inArray(cardType, ['amex', 'discover', 'mastercard', 'visa']) > -1;
-        }
+            isCardTypeSupported: function(cardType) {
+                return $.inArray(cardType, ['amex', 'discover', 'mastercard', 'visa']) > -1;
+            },
 
-        function cardInfoValidation(event) {
-            var cardType,
-                // We are adding 1 here because month in js style date-time starts with 0
-                // i.e. 0 for JAN, 1 for FEB etc. However, credit card expiry months start with 1
-                // i.e 1 for JAN, 2 for FEB etc.
-                currentMonth = new Date().getMonth() + 1,
-                currentYear = new Date().getFullYear(),
-                $number = $('#card-number'),
-                $cvn = $('#card-cvn'),
-                $expMonth = $('#card-expiry-month'),
-                $expYear = $('#card-expiry-year'),
-                cardNumber = $number.val(),
-                cvnNumber = $cvn.val(),
-                cardExpiryMonth = $expMonth.val(),
-                cardExpiryYear = $expYear.val();
+            cardInfoValidation: function(event) {
+                var cardType,
+                    // We are adding 1 here because month in js style date-time starts with 0
+                    // i.e. 0 for JAN, 1 for FEB etc. However, credit card expiry months start with 1
+                    // i.e 1 for JAN, 2 for FEB etc.
+                    currentMonth = new Date().getMonth() + 1,
+                    currentYear = new Date().getFullYear(),
+                    $number = $('#card-number'),
+                    $cvn = $('#card-cvn'),
+                    $expMonth = $('#card-expiry-month'),
+                    $expYear = $('#card-expiry-year'),
+                    cardNumber = $number.val(),
+                    cvnNumber = $cvn.val(),
+                    cardExpiryMonth = $expMonth.val(),
+                    cardExpiryYear = $expYear.val();
 
-            cardType = CreditCardUtils.getCreditCardType(cardNumber);
-
-            if (!CreditCardUtils.isValidCardNumber(cardNumber)) {
-                appendCardValidationErrorMsg(event, $number, gettext('Invalid card number'));
-            } else if (_.isUndefined(cardType) || !isCardTypeSupported(cardType.name)) {
-                appendCardValidationErrorMsg(event, $number, gettext('Unsupported card type'));
-            } else if (cvnNumber.length !== cardType.cvnLength || !Number.isInteger(Number(cvnNumber))) {
-                appendCardValidationErrorMsg(event, $cvn, gettext('Invalid security number'));
-            }
-
-            if (!Number.isInteger(Number(cardExpiryMonth)) ||
-                Number(cardExpiryMonth) > 12 || Number(cardExpiryMonth) < 1) {
-                appendCardValidationErrorMsg(event, $expMonth, gettext('Invalid month'));
-            } else if (!Number.isInteger(Number(cardExpiryYear)) || Number(cardExpiryYear) < currentYear) {
-                appendCardValidationErrorMsg(event, $expYear, gettext('Invalid year'));
-            } else if (Number(cardExpiryMonth) < currentMonth && Number(cardExpiryYear) === currentYear) {
-                appendCardValidationErrorMsg(event, $expMonth, gettext('Card expired'));
-            }
-        }
-
-        function detectCreditCard() {
-            var card,
-                $input = $('#card-number'),
-                cardNumber = $input.val().replace(/\s+/g, ''),
-                iconPath = '/static/images/credit_cards/';
-
-            if (cardNumber.length > 12) {
-                card = CreditCardUtils.getCreditCardType(cardNumber);
+                cardType = CreditCardUtils.getCreditCardType(cardNumber);
 
                 if (!CreditCardUtils.isValidCardNumber(cardNumber)) {
-                    $('.card-type-icon').attr('src', '').addClass('hidden');
-                    return;
+                    BasketPage.appendCardValidationErrorMsg(event, $number, gettext('Invalid card number'));
+                } else if (_.isUndefined(cardType) || !BasketPage.isCardTypeSupported(cardType.name)) {
+                    BasketPage.appendCardValidationErrorMsg(event, $number, gettext('Unsupported card type'));
+                } else if (cvnNumber.length !== cardType.cvnLength || !Number.isInteger(Number(cvnNumber))) {
+                    BasketPage.appendCardValidationErrorMsg(event, $cvn, gettext('Invalid security number'));
                 }
 
-                if (typeof card !== 'undefined') {
-                    $('.card-type-icon').attr(
-                        'src',
-                        iconPath + card.name + '.png'
-                    ).removeClass('hidden');
-                    $input.trigger('cardType:detected', {type: card.name});
+                if (!Number.isInteger(Number(cardExpiryMonth)) ||
+                    Number(cardExpiryMonth) > 12 || Number(cardExpiryMonth) < 1) {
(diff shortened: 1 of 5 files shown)


ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,38 @@ define([
...
 
+        function calculateLocalizedPrice(priceInUSD) {
+            var countryData = Cookies.getJSON('edx-price-l10n');
+            // Default to USD when
+            // The exchange rate cookie doesn't exist
+            // The country code is USA
+            // The symbol, rate, or code values are missing from the cookie
+            if (!countryData || countryData.countryCode === 'USA'
+                || (!countryData.symbol || !countryData.rate || !countryData.code)) {
+                return '$' + priceInUSD;
```
AlasdairSwan: It may be cleaner to just set countryData.symbol to '$' and countryData.rate = 1 so the formatting of the number is consistent.

MatthewPiatetsky:
good to dos, also we put the ticket number on the third line of the squashed commit message
you can create a coupon here http://localhost:18130/coupons/new/ and apply that

you can check with about how the disclaimer text/usd price should be displayed, although i think we also wanted to display the currency code after the number in addition to the symbol before the number

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,38 @@ define([
...
+        }
+
+        function localizedPriceTranslation() {
+            $('.price').each(function() {
+                // Stringified price value with a USD symbol prefix like $100.00
+                var rawPriceValue = $(this).text();
+                var priceInUSD = parseFloat(
+                    rawPriceValue.substring(
+                        rawPriceValue.indexOf('$') + 1,
+                        rawPriceValue.length
```
AlasdairSwan: You don't need to include this, if omitted it defaults to the length of the string.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,38 @@ define([
...
+                    )
+                ).toFixed(2);
+                $(this).text(calculateLocalizedPrice(priceInUSD));
```
AlasdairSwan: I think it would be worth doing a check to ensure the new value is different than the current value before updating the DOM.

AlasdairSwan (changes requested):

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,63 @@ define([
...
+            $('.price').each(function() {
+                var priceText = $(this).text();
+                var localPriceText = generateLocalPriceText($(this).text());
+                if (priceText !== localPriceText) {
+                    $(this).text(localPriceText);
+                }
+            });
+
+            $('.voucher').each(function() {
+                var priceText = $(this).text();
```
jaebradley: This is repeated code from above - I can move to it's own function if necessary.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,63 @@ define([
...
+        }
+
+        function formatToLocalPrice(priceInUsd) {
+            var currencyValue;
+
+            // Default to USD when the exchange rate cookie doesn't exist
+            var countryData = Cookies.getJSON('edx-price-l10n');
+            if (countryData && countryData.countryCode !== 'USA') {
+                currencyValue = Math.round(priceInUsd * countryData.rate)
+                    .toLocaleString(undefined, {maximumFractionDigits: 0});
```
jaebradley: This is slightly different than the implementation [here](https://github.com/edx/edx-mktg/pull/1612/files#diff-873ba66872fcc5d2e711a09db7daddc2R194) and provides additional numerical formatting (`1000` => `1,000` - you can see the Yen screenshot in the PR description for a live example).

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,63 @@ define([
             );
         }
 
+        function addPriceDisclaimer() {
+            var price = $('#basket-total').find('> .price').text();
+            var countryData = Cookies.getJSON('edx-price-l10n');
+            if (countryData && countryData.countryCode !== 'USA') {
+                $('<span>').attr('class', 'price-disclaimer')
```
jaebradley: There's got to be a better way to do this...open to suggestions!

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,63 @@ define([
             );
         }
 
+        function addPriceDisclaimer() {
+            var price = $('#basket-total').find('> .price').text();
+            var countryData = Cookies.getJSON('edx-price-l10n');
```
MatthewPiatetsky: define all var variables on first line like this

```
var x=1,
    y=2;
```

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,63 @@ define([
...
+            var currencyValue;
+
+            // Default to USD when the exchange rate cookie doesn't exist
+            var countryData = Cookies.getJSON('edx-price-l10n');
+            if (countryData && countryData.countryCode !== 'USA') {
+                currencyValue = Math.round(priceInUsd * countryData.rate)
+                    .toLocaleString(undefined, {maximumFractionDigits: 0});
+                return countryData.symbol + currencyValue + ' ' + countryData.code + ' *';
+            } else {
+                return '$' + priceInUsd;
```
MatthewPiatetsky: we can double check if we need USD after, if we don't want to change the current behavior, perhaps we can just do nothing in that case?

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,63 @@ define([
...
+            }
+        }
+
+        function formatToLocalPrice(priceInUsd) {
+            var currencyValue;
+
+            // Default to USD when the exchange rate cookie doesn't exist
+            var countryData = Cookies.getJSON('edx-price-l10n');
+            if (countryData && countryData.countryCode !== 'USA') {
+                currencyValue = Math.round(priceInUsd * countryData.rate)
```
MatthewPiatetsky: maybe lets do ceiling here?

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,63 @@ define([
             );
         }
 
+        function addPriceDisclaimer() {
+            var price = $('#basket-total').find('> .price').text();
```
MatthewPiatetsky: what is the > in the selector?

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,63 @@ define([
...
+            var countryData = Cookies.getJSON('edx-price-l10n');
+            if (countryData && countryData.countryCode !== 'USA') {
+                $('<span>').attr('class', 'price-disclaimer')
+                    .text('* This total contains an approximate conversion. You will be charged ' + price + ' USD.')
+                    .css('font-style', 'italic')
+                    .appendTo('div[aria-labelledby="order-details-region"]');
```
MatthewPiatetsky: defer to @AlasdairSwan on whether this is accessible

package-lock.json
```diff
@@ -0,0 +1,6907 @@
+{
```
MatthewPiatetsky: i think you didn't mean to commit this file?

ecommerce/static/js/pages/basket_page.js
```diff
@@ -234,10 +234,63 @@ define([
...
+                    .toLocaleString(undefined, {maximumFractionDigits: 0});
+                return countryData.symbol + currencyValue + ' ' + countryData.code + ' *';
+            } else {
+                return '$' + priceInUsd;
+            }
+        }
+
+        function generateLocalPriceText(usdPriceText) {
+            var localPriceText = usdPriceText;
+            usdPriceText.match(/\$[0-9]+\.[0-9]+/g).forEach(function(value) {
```
MatthewPiatetsky: perhaps add a comment showing what this regex does with input/output?

MatthewPiatetsky:
We might want to wrap this (and the platform work) in a feature gate
we can decide during standup tomorrow

MatthewPiatetsky (commented):
We might want to wrap this (and the platform work) in a feature gate
we can decide during standup tomorrow

left some comments

ecommerce/static/js/pages/basket_page.js
```diff
@@ -16,17 +16,19 @@ define([
               Cookies) {
         'use strict';
 
-        function hideVoucherForm() {
+        var BasketPage = {};
```
jaebradley: The reason I'm creating (and returning) this object is because the current structure is not conducive to Jasmine tests (particularly spies).

## Example Time

Consider this `foobar.js`

```javascript
// foobar.js

define([],
    function() {
        'use strict';

        function bar() {
            console.log('bar has been called');
        }

        function foo() {
            console.log('foo has been called');
            bar();
        }

        return {
            foo: foo,
            bar: bar
        };
    }
);
```

The previous implementation of `foobar.js` fails this spec

```javascript
// foobar_spec.js

define([
    'foobar'
],
    function(Foobar) {
        'use strict';

        describe('Testing foo', function() {
            it('should call bar', function() {
                console.log(Foobar);
                spyOn(Foobar, 'bar');
                Foobar.foo();
                expect(Foobar.bar).toHaveBeenCalled();
            });
        });
    }
);
```

while this `foobar.js` passes

```javascript
define([],
    function() {
        'use strict';

        var obj = {};

        obj.bar = function() {
            console.log('bar has been called');
        };

        obj.foo = function() {
            console.log('foo has been called');
            obj.bar();
        };

        return obj;
    }
);
```

The first `foobar.js` fails (I think) because the `Foobar.bar` property (which Jasmine puts a spy on) is different than the `bar` method that the method `foo` (which `Foobar.foo` has a reference to) calls. Thus, while the method `bar` is called by `Foobar.foo`, the property `Foobar.bar` is never called, and thus, `expect(Foobar.bar).toHaveBeenCalled()` will fail.

The second `foobar.js` succeeds for the same reason that the first `foobar.js` fails.

The implication is that in order to write sane Jasmine tests for `BasketPage` the returned object needs to follow the pattern established in the second `foobar.js` instead of the current pattern, which is identical to the first `foobar.js`.

package-lock.json
jaebradley: From [the `package-lock.json` documentation](https://docs.npmjs.com/files/package-lock.json)
> This file is intended to be committed into source repositories

I've deleted it for now and added `package-lock.json` to the `.gitignore`.

ecommerce/static/js/pages/basket_page.js
jaebradley: A [child combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_selectors)!
> It matches only those elements matched by the second selector that are the children of elements matched by the first.

I initially had
image

ecommerce/static/js/pages/basket_page.js
jaebradley: I tested the updated Basket Page using `axe` and I didn't see any warnings over the added `span`.

ecommerce/static/js/pages/basket_page.js
jaebradley: I was just copying from [here](https://github.com/edx/edx-mktg/pull/1612/files#diff-873ba66872fcc5d2e711a09db7daddc2R200)

image

MatthewPiatetsky: ok i guess i'll just my code to round :(

ecommerce/static/js/pages/basket_page.js
jaebradley: I see what you mean.

I do like the modularity of this method and how it encapsulates the implementation details around the `edx-price-l10n` cookie.

I did include [checks](https://github.com/edx/ecommerce/pull/1498/files#diff-a77e90eb28521ce0645b1b9831b34192R277) that don't update the DOM element if the text value is the same as the formatted local price value.

MatthewPiatetsky (approved):
once Alasdair approves and you've tested locally that you can still successfully checkout, good from my end.
this can also be released before the platform work, since no user should currently have the cookie (so this should do nothing when it is released)

jaebradley:
@AlasdairSwan mind taking another look? Thanks!

ecommerce/static/js/pages/basket_page.js
AlasdairSwan: Why not just do this;
```
var BasketPage = {
    onFail: function() {....}
};
```

ecommerce/static/js/pages/basket_page.js
AlasdairSwan: @jaebradley where is the element with this id?

ecommerce/static/js/pages/basket_page.js
```diff
@@ -232,22 +234,76 @@ define([
...
+        };
 
-        function onReady() {
+        BasketPage.addPriceDisclaimer = function() {
+            var price = $('#basket-total').find('> .price').text(),
+                countryData = Cookies.getJSON('edx-price-l10n');
+            if (countryData && countryData.countryCode !== 'USA') {
+                $('<span>').attr('class', 'price-disclaimer')
+                    .text('* This total contains an approximate conversion. You will be charged ' + price + ' USD.')
+                    .css('font-style', 'italic')
```
AlasdairSwan: Please update to use a CSS class for styling, we shouldn't be styling with JS.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -232,22 +234,76 @@ define([
...
+            usdPriceText.match(/\$[0-9]+\.[0-9]+/g).forEach(function(value) {
+                localPriceText = localPriceText.replace(value, BasketPage.formatToLocalPrice(value.substring(1)));
+            });
+            return localPriceText;
+        };
+
+        BasketPage.translateToLocalPrices = function() {
+            $('.price').each(function() {
+                var priceText = $(this).text();
+                var localPriceText = BasketPage.generateLocalPriceText($(this).text());
```
AlasdairSwan: There should only be 1 var declaration at the top of the function, comma separating the variable names.

AlasdairSwan: Also, you should just pass ```priceText``` to the function on line 276 instead of getting it again.

ecommerce/static/js/pages/basket_page.js
AlasdairSwan: Yeah, please move the function called in the each loop to its own named function so it can be called for each loop instead of duplicating it.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -257,20 +313,20 @@ define([
...
             $('#card-cvn-help').on('click touchstart', function(event) {
                 event.preventDefault();
-                toggleCvvTooltip();
+                BasketPage.toggleCvvTooltip();
             });
 
-            $('#card-cvn-help').blur(hideCvvTooltip);
-            $('#card-cvn-help').hover(showCvvTooltip, hideCvvTooltip);
+            $('#card-cvn-help').blur(BasketPage.hideCvvTooltip);
+            $('#card-cvn-help').hover(BasketPage.showCvvTooltip, BasketPage.hideCvvTooltip);
```
AlasdairSwan: The jQuery object ```$('#card-cvn-help')``` is being created 3 times, please assign it to a variable so it only needs to be created once. Also, remember you can chain jQuery methods.

AlasdairSwan (changes requested):
Just some minor changes and a few comments.

ecommerce/static/js/pages/basket_page.js
jaebradley: image

ecommerce/static/js/pages/basket_page.js
jaebradley: This was existing code that I edited to deal with the Jasmine test issues. Would you still like me to make these changes (I know they're quite small) in this PR?

ecommerce/static/js/pages/basket_page.js
jaebradley: @AlasdairSwan ha - good point!

ecommerce/static/js/pages/basket_page.js
AlasdairSwan: Yes please, if you're in there already I think any code clean up that can be done should be.

jaebradley:
@AlasdairSwan mind taking another pass here?

ecommerce/static/js/pages/basket_page.js
```diff
@@ -16,452 +16,493 @@ define([
...
-                }
-            }
+        var BasketPage = {
+            hideVoucherForm: function() {
+                $('#voucher_form_container').hide();
+                $('#voucher_form_link').show();
+            },
+
+            onFail: function() {
+                var message = gettext('Problem occurred during checkout. Please contact support');
```
AlasdairSwan: Missing ```.```

ecommerce/static/js/pages/basket_page.js
```diff
@@ -16,452 +16,493 @@ define([
...
+                        'input[name=last_name]',
+                        'input[name=city]',
+                        'select[name=country]'
+                    ],
+                    experiments = window.experimentVariables || {};
+                // Only require address and state if we are not in the hide location fields variation of this experiment
+                // https://openedx.atlassian.net/browse/LEARNER-2355
+                if (!(experiments && experiments.hide_location_fields)) {
+                    requiredFields.push('input[name=address_line1]');
+                    if (['US', 'CA'].indexOf($('select[name=country]').val()) > -1) {
```
AlasdairSwan: It would be clearer what is going on if you moved the array to a variable.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -16,452 +16,493 @@ define([
...
 
-        function detectCreditCard() {
-            var card,
-                $input = $('#card-number'),
-                cardNumber = $input.val().replace(/\s+/g, ''),
-                iconPath = '/static/images/credit_cards/';
+            appendCardValidationErrorMsg: function(event, field, msg) {
+                field.find('~.help-block').append('<span>' + msg + '</span>');
+                field.focus();
+                event.preventDefault();
```
AlasdairSwan: This should be called at the start of the function.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -16,452 +16,493 @@ define([
...
+            showVoucherForm: function() {
+                $('#voucher_form_container').show();
+                $('#voucher_form_link').hide();
+                $('#id_code').focus();
+            },
+
+            showCvvTooltip: function() {
+                $('#cvvtooltip').show();
+                $('#card-cvn-help').attr('aria-haspopup', 'false');
+                $('#card-cvn-help').attr('aria-expanded', 'true');
```
AlasdairSwan: You can chain jQuery methods instead of instantiating the same jQuery object twice, but here it would make sense to just set both attributes at the same time http://api.jquery.com/attr/#setting-several-attrs.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -16,452 +16,493 @@ define([
...
+            },
+
+            hideCvvTooltip: function() {
+                $('#cvvtooltip').hide();
+                $('#card-cvn-help').attr('aria-haspopup', 'true');
+                $('#card-cvn-help').attr('aria-expanded', 'false');
```
AlasdairSwan: You can chain jQuery methods instead of instantiating the same jQuery object twice, but here it would make sense to just set both attributes at the same time http://api.jquery.com/attr/#setting-several-attrs.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -16,452 +16,493 @@ define([
...
+                var $cvnHelpButton = $('#card-cvn-help');
+                $('#cvvtooltip').toggle();
+                $cvnHelpButton.attr(
+                    'aria-haspopup',
+                    $cvnHelpButton.attr('aria-haspopup') === 'true' ? 'false' : 'true'
+                );
+                $cvnHelpButton.attr(
+                    'aria-expanded',
+                    $cvnHelpButton.attr('aria-expanded') === 'true' ? 'false' : 'true'
+                );
```
AlasdairSwan: You can chain jQuery methods, but here it would make sense to just set both attributes at the same time http://api.jquery.com/attr/#setting-several-attrs.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -16,452 +16,493 @@ define([
...
+            },
+
+            formatToLocalPrice: function(priceInUsd) {
+                var currencyValue,
+                    countryData = Cookies.getJSON('edx-price-l10n');
+
+                // Default to USD when the exchange rate cookie doesn't exist
+                if (countryData && countryData.countryCode !== 'USA') {
+                    currencyValue = Math.round(priceInUsd * countryData.rate)
+                        .toLocaleString(undefined, {maximumFractionDigits: 0});
```
AlasdairSwan: Why no decimals? In edx-mktg it rounds to 2 decimal places following a standard already set by the program bundling work.

ecommerce/static/js/pages/basket_page.js
```diff
@@ -16,452 +16,493 @@ define([
...
-                            'West Virginia': 'WV',
-                            Wisconsin: 'WI',
-                            Wyoming: 'WY'
+                $('#card-cvn-help').on('click touchstart', function(event) {
+                    event.preventDefault();
+                    BasketPage.toggleCvvTooltip();
+                });
+
+                $('#card-cvn-help').blur(BasketPage.hideCvvTooltip);
+                $('#card-cvn-help').hover(BasketPage.showCvvTooltip, BasketPage.hideCvvTooltip);
```
AlasdairSwan: You can chain jQuery methods instead of instantiating the same jQuery object twice.

ecommerce/static/js/test/specs/pages/basket_page_spec.js
```diff
@@ -554,6 +554,124 @@ define([
                     expect(window.analytics.page).toHaveBeenCalled();
                 });
             });
+
+            describe('formatToLocalPrice', function() {
+                var EDX_PRICE_LOCATION_COOKIE_NAME = 'edx-price-l10n';
+                var USD_VALUE = 100.25;
+                var EXPECTED_USD_PRICE = '$100.25';
+                var COOKIE_VALUES = {countryCode: 'FOO', rate: 2, code: 'BAR', symbol: 'BAZ'};
```
AlasdairSwan: There should only be 1 var declaration at the top of the function, comma separating the variable names.

AlasdairSwan (changes requested):

ecommerce/static/js/pages/basket_page.js
jaebradley: [Learner-2012](https://openedx.atlassian.net/browse/LEARNER-2012) seemed to suggest that when currency converting we should not have decimals
> Verify that format for prices without a label is $213 USD

AlasdairSwan: @jaebradley can you please confirm with @stroilova the desired format?

jaebradley: Based on the conversation in the Learner - Growth HipChat, it seems like no decimals is the desired format for converted values?

AlasdairSwan: 👍

AlasdairSwan (approved):
:shipit:

PR #1499, opened 2017-10-06
WL-1270 | Included an empty file on receipt page to include extra js …

hasnain-naveed:
…code in WL-themes.


afzaledx:
@mjfrey, we need to include GA code for a WL site on the receipt page. We'll override the file we've pulled out in themes. Can you review this?

afzaledx:
@bill-filler

PR #1502, opened 2017-10-10
LEARNER-2714 Fix for cat not saving credit hours.

Ayub-Khan:
[LEARNER-2714](https://openedx.atlassian.net/browse/LEARNER-2714):

**Current behavior**
When I attempt to edit the 'credit hours' field in the CAT/Ecomm instance for course run course run ASUx+CHM114x+2184C, the page won't save the change.

**Steps to reproduce**
Navigate to the CAT/Ecomm instance for course run ASUx+CHM114x+2184C (https://ecommerce.edx.org/courses/course-v1:ASUx+CHM114x+2184C).
Edit the page
Set the 'credit hours' field for course as '4' (currently has a value of '3')
Save the page

**Scope & Impact**
The current value of '3' misleads learners who are paying to convert this course to university credit.
Screenshots attached.

### DIFF
--- a/ecommerce/courses/models.py
+++ b/ecommerce/courses/models.py
@@ -225,6 +225,8 @@ class Course(models.Model):
         seat.title = self.get_course_seat_name(certificate_type, id_verification_required)
         seat.expires = expires
 
+        seat.save()
+
         # If a ProductAttribute is saved with a value of None or the empty string, the ProductAttribute is deleted.
         # As a consequence, Seats derived from a migrated "audit" mode do not have a certificate_type attribute.
         seat.attr.certificate_type = certificate_type
@@ -241,7 +243,7 @@ class Course(models.Model):
         if credit_hours:
             seat.attr.credit_hours = credit_hours
 
-        seat.save()
+        seat.attr.save()
 
         try:
             stock_record = StockRecord.objects.get(product=seat, partner=partner)
--- a/ecommerce/courses/tests/test_models.py
+++ b/ecommerce/courses/tests/test_models.py
@@ -189,6 +189,44 @@ class CourseTests(DiscoveryTestMixin, TestCase):
         # Expected seat total, with one being the parent seat product.
         self.assertEqual(course.products.count(), len(credit_data) + 1)
 
+    def test_update_credit_seat(self):
+        """
+        Tests that model's seat update method updates the seat and attribute values
+        """
+        course = CourseFactory()
+        credit_provider = 'MIT'
+        credit_hours = 2
+        certificate_type = 'credit'
+        id_verification_required = True
+        price = 10
+        course.create_or_update_seat(
+            certificate_type,
+            id_verification_required,
+            price,
+            self.partner,
+            credit_provider=credit_provider,
+            credit_hours=credit_hours
+        )
+        credit_hours = 4
+        price = 100
+        credit_seat = course.create_or_update_seat(
+            certificate_type,
+            id_verification_required,
+            price,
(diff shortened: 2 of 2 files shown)


clintonb (changes requested):
Write a test proving you've fixed the problem.

ecommerce/courses/models.py
```diff
@@ -224,7 +224,7 @@ def create_or_update_seat(
         seat.is_discountable = True
         seat.title = self.get_course_seat_name(certificate_type, id_verification_required)
         seat.expires = expires
-
+        seat.save()
```
clintonb: Nit: add a newline after this line for legibility.

ecommerce/courses/tests/test_models.py
```diff
@@ -189,6 +189,43 @@ def test_create_credit_seats(self):
         # Expected seat total, with one being the parent seat product.
         self.assertEqual(course.products.count(), len(credit_data) + 1)
 
+    def test_update_credit_seat(self):
+        """
+        Verify that model's seat updation method updates the seat
+        """
+        course = CourseFactory(id='a/test/c', name='Test Course', site=self.site)
```
clintonb: None of these arguments are relevant to your test. Get rid of them. Don't waste time, or add confusion to later readers, by adding unnecessary factory arguments.

ecommerce/courses/tests/test_models.py
```diff
@@ -189,6 +189,43 @@ def test_create_credit_seats(self):
         # Expected seat total, with one being the parent seat product.
         self.assertEqual(course.products.count(), len(credit_data) + 1)
 
+    def test_update_credit_seat(self):
+        """
+        Verify that model's seat updation method updates the seat
```
clintonb: This docstring doesn't tell the reader anything that isn't obvious from the name of the test. It can go. Docstrings should go further than tell us what the test does. They should explain _why_ it's being done. There isn't much explanation needed, so this can go.

ecommerce/courses/tests/test_models.py
```diff
@@ -189,6 +189,43 @@ def test_create_credit_seats(self):
...
+        """
+        course = CourseFactory(id='a/test/c', name='Test Course', site=self.site)
+        credit_data = ['MIT', 2]
```
clintonb: There is no reason for this to be an array. Break this into two variables: `credit_provider` and `credit_hours`.

clintonb (changes requested):
1. Why haven't we seen this problem until now?
2. Update your commit message to follow the best practices at https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/161880915/Git+Best+Practices.

Ayub-Khan:
@clintonb Im not sure about the question why we discovered this so late or why it did not occur before. Could be that we do not update credit seats once they have been created.

ecommerce/courses/tests/test_models.py
```diff
@@ -189,6 +189,44 @@ def test_create_credit_seats(self):
...
+            credit_provider=credit_provider,
+            credit_hours=credit_hours
+        )
+        self.assert_course_seat_valid(
+            credit_seat,
+            course,
+            certificate_type,
+            id_verification_required,
+            price,
+            credit_provider=credit_data[0],
```
clintonb: Make sure you run your tests locally _before_ pushing. Doing so helps decreases the latency between push and merge.

clintonb (approved):

schenedx (approved):
👍

PR #1504, opened 2017-10-11
Add new Course Entitlement product class.

mjfrey:
This PR adds a new product class called "Course Entitlement".  It will allow learners to purchase a "course" rather than an individual course-run.

LEARNER-1728

### DIFF
--- a/ecommerce/core/constants.py
+++ b/ecommerce/core/constants.py
@@ -21,6 +21,9 @@ ENROLLMENT_CODE_PRODUCT_CLASS_NAME = 'Enrollment Code'
 ENROLLMENT_CODE_SWITCH = 'create_enrollment_codes'
 ENROLLMENT_CODE_SEAT_TYPES = ['verified', 'professional', 'no-id-professional']
 
+# Course Entitlement constant
+COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME = 'Course Entitlement'
+
 # Discovery Service constants
 DEFAULT_CATALOG_PAGE_SIZE = 100
 
--- a/ecommerce/extensions/catalogue/tests/test_course_entitlement.py
+++ b/ecommerce/extensions/catalogue/tests/test_course_entitlement.py
@@ -1,0 +1,25 @@
+from oscar.core.loading import get_model
+from oscar.test.factories import ProductFactory
+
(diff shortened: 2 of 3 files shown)


ecommerce/core/constants.py
```diff
@@ -21,6 +21,9 @@
 ENROLLMENT_CODE_SWITCH = 'create_enrollment_codes'
 ENROLLMENT_CODE_SEAT_TYPES = ['verified', 'professional', 'no-id-professional']
 
+# Course Entitlement contrant
+COURSE_ENTITLEMENT_CLASS_NAME = 'Course Entitlement'
```
clintonb: `COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME`

ecommerce/extensions/catalogue/tests/test_course_entitlement.py
```diff
@@ -0,0 +1,28 @@
...
+ProductClass = get_model('catalogue', 'ProductClass')
+
+
+class CourseEntitlementTest(TestCase):
+    """ Test course entitlement products."""
+
+    def test_course_entitlement_product(self):
+        """Test if a course entitlement product is properly created."""
+        course_entitlement_product_class, _ = ProductClass.objects.get_or_create(name=COURSE_ENTITLEMENT_CLASS_NAME)
+        course_entitlement_product = ProductFactory()
```
clintonb: There isn't much value in validating anything besides the custom attributes. The model attributes are part of Django.

```python
product = ProductFactory(product_class=course_entitlement_product_class)
product.attr.course_key = 'foo-bar'
product.attr.certificate_type = 'verified'
product.attr.save()

product.refresh_from_db()
self.assertEqual(product.attr.course_key, 'foo-bar')
self.assertEqual(product.attr.certificate_type, 'verified')
```

clintonb (approved):

ecommerce/core/constants.py
```diff
@@ -21,6 +21,9 @@
 ENROLLMENT_CODE_SWITCH = 'create_enrollment_codes'
 ENROLLMENT_CODE_SEAT_TYPES = ['verified', 'professional', 'no-id-professional']
 
+# Course Entitlement contrant
```
catong: nit: typo "contrant"

staubina (approved):
:+1:

PR #1506, opened 2017-10-11
add logs in clean history command

ahsan-ul-haq:


ecommerce/core/management/commands/clean_history.py
```diff
@@ -71,8 +71,12 @@ def handle(self, *args, **options):
...
                 history_batch = qs.values_list('id', flat=True)
+                batch_list = list(history_batch)
+
                 with transaction.atomic():
-                    model.history.filter(pk__in=list(history_batch)).delete()
-                    time.sleep(sleep_time)
+                    model.history.filter(pk__in=batch_list).delete()
 
+                time.sleep(sleep_time)
+                message = 'Deleted {}s from pk: {} to pk: {}'.format(model.__name__, batch_list[-1], batch_list[0])
```
clintonb: Ditch `str.format()` and use the logger's built-in interpolation.

```python
logger.info('Deleted instances of %s with PKs %d and %d', model.__name__, batch_list[0], batch_list[-1])
```

ecommerce/core/management/commands/clean_history.py
```diff
@@ -71,8 +71,12 @@ def handle(self, *args, **options):
             qs = qs[:batch_size]
             while qs.exists():
                 history_batch = qs.values_list('id', flat=True)
+                batch_list = list(history_batch)
```
clintonb: There is no need for a new variable: `history_batch = list(qs.values_list('id', flat=True))`

feanil (approved):
This is an improvement, not sure if this will fix the issue that caused it to hang but will hopefully give us some more info and reduce the number of things inside the atomic block.

ecommerce/core/management/commands/clean_history.py
clintonb: Move the log message into the transaction block.

clintonb (changes requested):

ecommerce/core/management/commands/clean_history.py
```diff
@@ -70,9 +70,14 @@ def handle(self, *args, **options):
...
             while qs.exists():
-                history_batch = qs.values_list('id', flat=True)
+                history_batch = list(qs.values_list('id', flat=True))
+
                 with transaction.atomic():
-                    model.history.filter(pk__in=list(history_batch)).delete()
-                    time.sleep(sleep_time)
+                    model.history.filter(pk__in=history_batch).delete()
+                    logger.info(
+                        'Deleted instances of %s with PKs %d and %d',
```
clintonb: with PKs between

clintonb (approved):
Update the commit message to follow our best practices: https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/161880915/Git+Best+Practices. Use the previous ticket number.

ahsan-ul-haq:

clintonb (approved):

PR #1507, opened 2017-10-12
Fulfillment module for Course Entitlement Products.

mjfrey:
Fulfilment module to fulfill and revoke course entitlement products.

Utils to create or update course entitlement products.

LEARNER-1741

### DIFF
--- a/ecommerce/courses/utils.py
+++ b/ecommerce/courses/utils.py
@@ -7,7 +7,7 @@ from django.utils.translation import ugettext_lazy as _
 from ecommerce.core.utils import traverse_pagination
 
 
-def mode_for_seat(product):
+def mode_for_product(product):
     """
     Returns the enrollment mode (aka course mode) for the specified product.
     If the specified product does not include a 'certificate_type' attribute it is likely the
--- a/ecommerce/entitlements/tests/test_utils.py
+++ b/ecommerce/entitlements/tests/test_utils.py
@@ -1,0 +1,40 @@
+from oscar.core.loading import get_model
+
+from ecommerce.entitlements.utils import create_or_update_course_entitlement
+from ecommerce.tests.testcases import TestCase
+
+StockRecord = get_model('partner', 'StockRecord')
+
+
+class TestCourseEntitlementProductCreation(TestCase):
+
+    def setUp(self):
+        super(TestCourseEntitlementProductCreation, self).setUp()
+
+        self.partner = self.site.siteconfiguration.partner
+
+    def test_course_entitlement_creation(self):
+        """ Test course entitlement product creation """
+
+        product = create_or_update_course_entitlement(
+            'verified', 100, self.partner, 'foo-bar', 'Foo Bar Entitlement')
+        self.assertEqual(product.title, 'Course Entitlement for Foo Bar Entitlement')
+        self.assertEqual(product.attr.UUID, 'foo-bar')
+
+        stock_record = StockRecord.objects.get(product=product, partner=self.partner)
+        self.assertEqual(stock_record.price_excl_tax, 100)
+
+    def test_course_entitlement_update(self):
+        """ Test course entitlement product update """
+
+        partner = self.site.siteconfiguration.partner
+        product = create_or_update_course_entitlement('verified', 100, self.partner, 'foo-bar', 'Foo Bar Entitlement')
+        stock_record = StockRecord.objects.get(product=product, partner=self.partner)
+
+        self.assertEqual(stock_record.price_excl_tax, 100)
+        self.assertEqual(product.title, 'Course Entitlement for Foo Bar Entitlement')
+
+        product = create_or_update_course_entitlement('verified', 200, partner, 'foo-bar', 'Foo Bar Entitlement')
+
+        stock_record = StockRecord.objects.get(product=product, partner=self.partner)
+        self.assertEqual(stock_record.price_excl_tax, 200)
--- a/ecommerce/entitlements/utils.py
+++ b/ecommerce/entitlements/utils.py
@@ -1,0 +1,88 @@
+from __future__ import unicode_literals
+
+import logging
+
+from django.conf import settings
+
+from django.db.models import Q
+from oscar.core.loading import get_model
+
+
+from ecommerce.core.constants import COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME
+from ecommerce.extensions.catalogue.utils import generate_sku
+
+logger = logging.getLogger(__name__)
+Category = get_model('catalogue', 'Category')
+Product = get_model('catalogue', 'Product')
+ProductCategory = get_model('catalogue', 'ProductCategory')
+ProductClass = get_model('catalogue', 'ProductClass')
+StockRecord = get_model('partner', 'StockRecord')
+
+
+def create_parent_course_entitlement(name, UUID):
+    """ Create the parent course entitlement product if it does not already exist. """
+    parent, created = Product.objects.get_or_create(
+        structure=Product.PARENT,
+        product_class=ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME),
+    )
+
+    if created:
+        logger.debug('Created new parent course_entitlement [%d] for [%s].', parent.id, UUID)
+    else:
+        logger.debug('Parent course_entitlement [%d] already exists for [%s].', parent.id, UUID)
+
+    ProductCategory.objects.get_or_create(category=Category.objects.get(name='Course Entitlements'), product=parent)
+    parent.title = 'Parent Course Entitlement for {}'.format(name)
+    parent.is_discountable = True
+    parent.attr.UUID = UUID
+    parent.save()
+
+    return parent, created
+
+
+def create_or_update_course_entitlement(certificate_type, price, partner, UUID, name):
+    """ Create or Update Course Entitlement Products """
+
+    certificate_type = certificate_type.lower()
+    UUID = unicode(UUID)
+
+    certificate_type_query = Q(
+        title='Course Entitlement for {}'.format(name),
+        attributes__name='certificate_type',
+        attribute_values__value_text=certificate_type
+    )
+
+    try:
+        parent_entitlement, __ = create_parent_course_entitlement(name, UUID)
+        all_products = parent_entitlement.children.all().prefetch_related('stockrecords')
+        course_entitlement = all_products.get(certificate_type_query)
+    except Product.DoesNotExist:
+        course_entitlement = Product()
+
+    course_entitlement.structure = Product.CHILD
+    course_entitlement.is_discountable = True
+    course_entitlement.title = 'Course Entitlement for {}'.format(name)
+    course_entitlement.attr.certificate_type = certificate_type
+    course_entitlement.attr.UUID = UUID
+    course_entitlement.parent = parent_entitlement
+    course_entitlement.save()
+
+    StockRecord.objects.update_or_create(
+        product=course_entitlement, partner=partner,
+        defaults={
+            'product': course_entitlement,
+            'partner': partner,
+            'partner_sku': generate_sku(course_entitlement, partner),
+            'price_excl_tax': price,
+            'price_currency': settings.OSCAR_DEFAULT_CURRENCY,
+        }
+    )
+
+    logger.info(
+        'Course course entitlement product stock record with certificate type [%s] for [%s] does not exist. '
+        'Instantiated a new instance.',
+        certificate_type,
+        UUID
+    )
+
+    return course_entitlement
--- a/ecommerce/extensions/catalogue/migrations/0026_course_entitlement_attr_change.py
+++ b/ecommerce/extensions/catalogue/migrations/0026_course_entitlement_attr_change.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from oscar.core.loading import get_model
+from oscar.core.utils import slugify
+
+from ecommerce.core.constants import COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME
+
+Category = get_model("catalogue", "Category")
+Product = get_model('catalogue', 'Product')
+ProductAttribute = get_model("catalogue", "ProductAttribute")
+ProductClass = get_model("catalogue", "ProductClass")
+
+
+def rename_product_attr(apps, schema_editor):
+    """ Rename course_entitlement product attr. """
+    course_entitlement_class = ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME)
+    product_attr = ProductAttribute.objects.get(product_class=course_entitlement_class, name="course_key")
+
+    product_attr.name = 'UUID'
+    product_attr.code = 'UUID'
+    product_attr.save()
+
+
+def remove_product_attr(apps, schema_editor):
+    """ Remove product attr """
+    course_entitlement_class = ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME)
+    product_attr = ProductAttribute.objects.get(product_class=course_entitlement_class, name="UUID")
+
+    product_attr.name = 'course_key'
+    product_attr.code = 'course_key'
+    product_attr.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0001_initial'),
+        ('catalogue', '0025_course_entitlement')
+    ]
+
+    operations = [
+        migrations.RunPython(rename_product_attr, remove_product_attr),
+    ]
--- a/ecommerce/extensions/catalogue/migrations/0027_catalogue_entitlement_option.py
+++ b/ecommerce/extensions/catalogue/migrations/0027_catalogue_entitlement_option.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from oscar.core.loading import get_model
+
+
+Option = get_model('catalogue', 'Option')
+
+
+def create_entitlement_option(apps, schema_editor):
+    """ Create catalogue entitlement option. """
+    course_entitlement_option = Option()
+    course_entitlement_option.name = 'Course Entitlement'
+    course_entitlement_option.code = 'course_entitlement'
+    course_entitlement_option.type = Option.OPTIONAL
+    course_entitlement_option.save()
+
+
+def remove_entitlement_option(apps, schema_editor):
+    """ Remove course entitlement option """
+    course_entitlement_option = Option.objects.get(code='course_entitlement')
+    course_entitlement_option.delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0001_initial'),
+        ('catalogue', '0026_course_entitlement_attr_change')
+    ]
+
+    operations = [
+        migrations.RunPython(create_entitlement_option, remove_entitlement_option),
+    ]
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -11,14 +11,15 @@ import logging
 import requests
 from django.conf import settings
 from django.core.urlresolvers import reverse
+from edx_rest_api_client.client import EdxRestApiClient
 from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout  # pylint: disable=ungrouped-imports
 from rest_framework import status
 
 from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME
-from ecommerce.core.url_utils import get_lms_enrollment_api_url
+from ecommerce.core.url_utils import get_lms_enrollment_api_url, get_lms_entitlement_api_url
 from ecommerce.courses.models import Course
-from ecommerce.courses.utils import mode_for_seat
+from ecommerce.courses.utils import mode_for_product
 from ecommerce.enterprise.utils import get_or_create_enterprise_customer_user
 from ecommerce.extensions.analytics.utils import audit_log, parse_tracking_context
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
@@ -28,6 +29,7 @@ from ecommerce.extensions.voucher.utils import create_vouchers
 from ecommerce.notifications.notifications import send_notification
 
 Benefit = get_model('offer', 'Benefit')
+Option = get_model('catalogue', 'Option')
 Product = get_model('catalogue', 'Product')
 Range = get_model('offer', 'Range')
 Voucher = get_model('voucher', 'Voucher')
@@ -208,7 +210,7 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
 
         for line in lines:
             try:
-                mode = mode_for_seat(line.product)
+                mode = mode_for_product(line.product)
                 course_key = line.product.attr.course_key
             except AttributeError:
                 logger.error("Supported Seat Product does not have required attributes, [certificate_type, course_key]")
@@ -292,7 +294,7 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
         try:
             logger.info('Attempting to revoke fulfillment of Line [%d]...', line.id)
 
-            mode = mode_for_seat(line.product)
+            mode = mode_for_product(line.product)
             course_key = line.product.attr.course_key
             data = {
                 'user': line.order.user.username,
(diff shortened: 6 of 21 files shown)


ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+StockRecord = get_model('partner', 'StockRecord')
+
+
+def create_parent_course_entitlement(name, course_id):
+    """ Create the parent course entitlement product if it does not already exist. """
+    parent, created = Product.objects.get_or_create(
+        structure=Product.PARENT,
+        product_class=ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME),
+    )
+    ProductCategory.objects.get_or_create(category=Category.objects.get(name='Course Entitlements'), product=parent)
```
staubina: Do we care whether or not this was created?  Should we include the data below in the context of the get_or_create using the defaults parameter? If we get an entitlement to we want to update it? 

Would update_or_create work better?

staubina (commented):
Early review comments I will take a deeper look on Monday.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,129 @@ def send_email(self, order):
...
+                order_line_id=line.id,
+                order_number=line.order.number,
+                product_class=line.product.get_product_class().name,
+                course_id=course_key,
+                certificate_type=getattr(line.product.attr, 'certificate_type', ''),
+                user_id=line.order.user.id
+            )
+
+            return True
+        except Exception:  # pylint: disable=broad-except
```
McKenzieW, line 628: What kind of exception(s) can we expect to be raised here? Should they possible be handled / logged differently?

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    try:
+        parent_entitlement = Product.objects.get(
+            product_class__name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME, structure=Product.PARENT)
+        all_products = parent_entitlement.children.all().prefetch_related('stockrecords')
+        course_entitlement = all_products.get(certificate_type_query)
+    except Product.DoesNotExist:
+        course_entitlement = Product()
+        parent_entitlement = create_parent_course_entitlement(name, course_id)
+
+    course_entitlement.structure = Product.CHILD
```
staubina: Could you also use update_or_create here for course_entitlement?

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    course_entitlement.save()
+
+    try:
+        stock_record = StockRecord.objects.get(product=course_entitlement, partner=partner)
+        logger.info(
+            'Retrieved course_entitlement product stock record with certificate type [%s] for [%s] from database.',
+            certificate_type,
+            course_id
+        )
+    except StockRecord.DoesNotExist:
```
staubina: Should we use get_or_create here instead of relying on the exception for the stock_record?

Generate Sku could be used in the data for the create portion of the get_or_create.

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+        logger.info(
+            'Course course entitlement product stock record with certificate type [%s] for [%s] does not exist. '
+            'Instantiated a new instance.',
+            certificate_type,
+            course_id
+        )
+
+    stock_record.price_excl_tax = price
+    stock_record.price_currency = settings.OSCAR_DEFAULT_CURRENCY
+    stock_record.save()
```
staubina: Looking at this area, we might want to consider an update_or_create.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,129 @@ def send_email(self, order):
...
+                be "Course Entitlement" products.
+        Returns:
+            The original set of lines, with new statuses set based on the success or failure of fulfillment.
+        """
+        logger.info('Attempting to fulfill "Course Entitlement" product types for order [%s]', order.number)
+
+        for line in lines:
+            try:
+                mode = mode_for_seat(line.product)
+                course_key = line.product.attr.course_key
```
staubina: Note from yesterday's conversation.  Will we get UUID here as well?

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,129 @@ def send_email(self, order):
...
+                line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
+                continue
+
+            data = {
+                'user': order.user.username,
+                'is_active': True,
+                'mode': mode,
+                'expiration_date': '{}'.format(now() + timedelta(days=365)),
+                'course_entitlement_details': {
+                    'course_id': course_key
```
staubina: Note from yesterday's conversation: Changing to UUID

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,129 @@ def send_email(self, order):
...
+            logger.info('Attempting to revoke fulfillment of Line [%d]...', line.id)
+
+            mode = mode_for_seat(line.product)
+            course_key = line.product.attr.course_key
+            data = {
+                'user': line.order.user.username,
+                'is_active': False,
+                'mode': mode,
+                'course_entitlement_details': {
+                    'course_id': course_key,
```
staubina: Note from yesterday's conversation: UUID here also I assume

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -517,3 +520,140 @@ def test_revoke_line(self):
         line = self.order.lines.first()
         with self.assertRaises(NotImplementedError):
             EnrollmentCodeFulfillmentModule().revoke_line(line)
+
+
+class EntitlementFulfillmentModuleTests(FulfillmentTestMixin, TestCase):
```
staubina, line 526: Note from yesterday's conversation: Add tests for the UUID as well please.

staubina (changes requested):
Some requested changes, no major logic issues from what I can see. Let me know if you have any questions.

ecommerce/entitlements/tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    def setUp(self):
+        super(TestCourseEntitlementProductCreation, self).setUp()
+
+        self.partner = self.site.siteconfiguration.partner
+
+    def test_course_entitlement_creation(self):
+        """ Test course entitlement product creation """
+
+        product = create_or_update_course_entitlement(
+            'verfied', 100, self.partner, 'foo-bar', 'Foo Bar Entitlement')
```
bderusha: `verfied` spelling

ecommerce/entitlements/tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
...
+        self.assertEqual(product.title, 'Seat in Foo Bar Entitlement')
```
bderusha: Why does the title of this use the word "seat".  Seems confusing to me since these are outside of seats, yes?

ecommerce/entitlements/tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
...
+        self.assertEqual(product.attr.course_key, 'foo-bar')
+
+        stock_record = StockRecord.objects.get(product=product, partner=self.partner)
+        self.assertEqual(stock_record.price_excl_tax, 100)
+
+    def test_course_entitlement_update(self):
+        """ Test course entitlement product update """
+
+        partner = self.site.siteconfiguration.partner
+        product = create_or_update_course_entitlement('verfied', 100, self.partner, 'foo-bar', 'Foo Bar Entitlement')
```
bderusha: `verfied`

ecommerce/entitlements/tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
...
+        stock_record = StockRecord.objects.get(product=product, partner=self.partner)
+
+        self.assertEqual(stock_record.price_excl_tax, 100)
+        self.assertEqual(product.title, 'Seat in Foo Bar Entitlement')
+
+        product = create_or_update_course_entitlement('verfied', 200, partner, 'foo-bar', 'Foo Bar Entitlement')
```
bderusha: `verfied`

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+
+logger = logging.getLogger(__name__)
+Category = get_model('catalogue', 'Category')
+Product = get_model('catalogue', 'Product')
+ProductCategory = get_model('catalogue', 'ProductCategory')
+ProductClass = get_model('catalogue', 'ProductClass')
+StockRecord = get_model('partner', 'StockRecord')
+
+
+def create_parent_course_entitlement(name, course_id):
```
bderusha: Consistent use of `course_entitlement` VS `parent_course_entitlement` would be nice.  My preference is to drop the word parent, but my only request is consistency.

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    """ Create the parent course entitlement product if it does not already exist. """
+    parent, created = Product.objects.get_or_create(
+        structure=Product.PARENT,
+        product_class=ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME),
+    )
+    ProductCategory.objects.get_or_create(category=Category.objects.get(name='Course Entitlements'), product=parent)
+    parent.title = 'Seat in {}'.format(name)
```
bderusha: Reiterating comment in tests. "Seat in" seems like the wrong way to describe these products.

ecommerce/entitlements/utils.py
bderusha: Can we move 'Course Entitlements' to a constant like `COURSE_ENTITLEMENT_PRODUCT_CATEGORY_NAME`?

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    parent.is_discountable = True
+    parent.attr.course_key = course_id
```
bderusha: We discussed this offline, but wanted to come back to it.  Can we refactor this to use `uuid` instead?

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    parent.save()
+
+    if created:
+        logger.debug('Created new parent course_entitlement [%d] for [%s].', parent.id, course_id)
+    else:
+        logger.debug('Parent course_entitlement [%d] already exists for [%s].', parent.id, course_id)
```
bderusha: NIT: move these lines to right after the `get_or_create` so they still log even if something goes wrong re-saving the parent

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+            'Instantiated a new instance.',
+            certificate_type,
+            course_id
+        )
+
+    stock_record.price_excl_tax = price
+    stock_record.price_currency = settings.OSCAR_DEFAULT_CURRENCY
+    stock_record.save()
+
+    return course_entitlement
```
bderusha: Would be nice if this had the same API contract as django `update_or_create`, a tuple of `(course_entitlement_obj, boolean_created)`

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def create_or_update_course_entitlement(certificate_type, price, partner, course_id, name):
+    """ Create or Update Course Entitlement Products """
+
+    certificate_type = certificate_type.lower()
+    course_id = unicode(course_id)
+
+    certificate_type_query = Q(
+        title='Seat in {}'.format(name),
```
bderusha: Noting the use of `Seat in`  here as well for cleanup

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,92 @@
...
+
+    if created:
+        logger.debug('Created new parent course_entitlement [%d] for [%s].', parent.id, course_id)
+    else:
+        logger.debug('Parent course_entitlement [%d] already exists for [%s].', parent.id, course_id)
+
+    return parent
+
+
+def create_or_update_course_entitlement(certificate_type, price, partner, course_id, name):
```
bderusha: Is this only used by tests?  If so maybe we can move it to a test_utils file instead of the main application code

ecommerce/entitlements/utils.py
bderusha: Agreed.  `get_or_create` is a little cleaner here.

ecommerce/entitlements/utils.py
bderusha: Also the use of `course_id` where it has been called `course_key` should be updated for consistency.  Obviously if we refactor to use UUID this would also change.

ecommerce/entitlements/utils.py
bderusha: inconsistent use of `course_id` and `course_key`.  Let's pick one and use it everywhere.

bderusha (changes requested):
This is 1/2 a review.  I wanted you to get the comments now.  I will review the 2nd half of this PR after lunch 🍔 🍟

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,129 @@ def send_email(self, order):
...
+            lines (List of Lines): Order Lines, associated with purchased products in an Order. These should only
+                be "Course Entitlement" products.
+        Returns:
+            The original set of lines, with new statuses set based on the success or failure of fulfillment.
+        """
+        logger.info('Attempting to fulfill "Course Entitlement" product types for order [%s]', order.number)
+
+        for line in lines:
+            try:
+                mode = mode_for_seat(line.product)
```
bderusha: Mode for seat is not really the correct name for this method.  Can we update the method to be more accurately named `mode_for_product`?

ecommerce/entitlements/utils.py
mjfrey: This will eventually be used by the product API endpoint to create new products.  That is why it is in a util and not just the test.

ecommerce/entitlements/tests/test_utils.py
mjfrey: I am looking for a better word.  Any suggestions?

ecommerce/entitlements/utils.py
mjfrey: Yes, I will add the migration change to this PR and replace the use of the course_key with UUID.

ecommerce/entitlements/utils.py
mjfrey: What specifically is not consistent with parent vs not?  I think the use is correct.

ecommerce/entitlements/utils.py
mjfrey: I originally had 'Course Entitlement in', what do you think about that?

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,129 @@ def send_email(self, order):
...
+                UUID = line.product.attr.UUID
+            except AttributeError:
```
staubina: If for some reason you get all Attribute errors it looks like you will never create a data object.  Then on line 584 when you post_to_entitlement_api you will send a non existent data object.  Should we initialize data as an empty object just in case?

ecommerce/extensions/fulfillment/modules.py
staubina: Did you get a chance to look at this one @mjfrey ? Can we avoid a Broad Except?

staubina (commented):
One open question from @McKenzieW about an open exception, otherwise 👍

ecommerce/extensions/fulfillment/modules.py
mjfrey: We generally catch all Exceptions here rather than enumerating all possible exceptions.  Take a look at the Seat revoke.  If anything goes wrong, we just do  not revoke.

McKenzieW: 👍 works for me

staubina (approved):

ecommerce/entitlements/utils.py
```diff
@@ -0,0 +1,88 @@
...
+        product_class=ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME),
+    )
+
+    if created:
+        logger.debug('Created new parent course_entitlement [%d] for [%s].', parent.id, UUID)
+    else:
+        logger.debug('Parent course_entitlement [%d] already exists for [%s].', parent.id, UUID)
+
+    ProductCategory.objects.get_or_create(category=Category.objects.get(name='Course Entitlements'), product=parent)
+    parent.title = 'Course Entitlement for {}'.format(name)
```
bderusha: Should we have different names for the parent/child products?

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,129 @@ def send_email(self, order):
...
+                logger.error('Entitlement Product does not have required attributes, [certificate_type, UUID]')
+                line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
+                continue
+
+            data = {
+                'user': order.user.username,
+                'is_active': True,
+                'mode': mode,
+                'expiration_date': '{}'.format(now() + timedelta(days=365)),
```
bderusha: Talking with Deen and finance, etc.  It seems like `expiration_date` may be too naive for the use cases we need to support.  Let's talk about this.

bderusha (changes requested):

bderusha:
Looks good code-wise.  Biggest question is actually more LMS side model design.

ecommerce/entitlements/utils.py
mjfrey: We currently do not for Seats.  We could, but the parent is never actually purchased.

ecommerce/extensions/fulfillment/modules.py
mjfrey: I can rename

ecommerce/extensions/fulfillment/modules.py
mjfrey: say more.  Make this configurable?

mjfrey:
@bderusha @staubina Removed expiration.  Should be good to go. Take another look.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +508,128 @@ def send_email(self, order):
...
+                mode = mode_for_product(line.product)
+                UUID = line.product.attr.UUID
+            except AttributeError:
+                logger.error('Entitlement Product does not have required attributes, [certificate_type, UUID]')
+                line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
+                continue
+
+            data = {
+                'user': order.user.username,
+                'is_active': True,
```
staubina: Should we remove is_active from the data as well?

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +508,128 @@ def send_email(self, order):
...
+
+    def revoke_line(self, line):
+        try:
+            logger.info('Attempting to revoke fulfillment of Line [%d]...', line.id)
+
+            mode = mode_for_product(line.product)
+            UUID = line.product.attr.UUID
+            data = {
+                'user': line.order.user.username,
+                'is_active': False,
```
staubina: As mentioned above.

ecommerce/extensions/fulfillment/modules.py
mjfrey: oops
Yes, forgot.

staubina (approved):

ecommerce/courses/utils.py
```diff
@@ -7,7 +7,7 @@
 from ecommerce.core.utils import traverse_pagination
 
 
-def mode_for_seat(product):
+def mode_for_product(product):
```
bderusha, line 10: 🎸 🔥 ✊

ecommerce/entitlements/utils.py
bderusha: More for administration,  if we don't really use the parent product it would be great to have an easy way to distinguish between them when looking at names in django admin

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +508,128 @@ def send_email(self, order):
...
+        try:
+            logger.info('Attempting to revoke fulfillment of Line [%d]...', line.id)
+
+            mode = mode_for_product(line.product)
+            UUID = line.product.attr.UUID
+            data = {
+                'user': line.order.user.username,
+                'mode': mode,
+                'expired_at': str(datetime.datetime.now()),
+                'revoke': True,
```
bderusha: What is this? How is it used?  I have questions

bderusha (commented):
My only question is around the revocation API.  1 question about the revoke attribute.  Another not embedded in the code is whether or not it would make sense for the fulfillment API to return a primary key we could look up rather than looking up by username, uuid, mode, etc.

ecommerce/extensions/fulfillment/modules.py
mjfrey: @staubina  and I where thinking that an explicit revoke would be sent to LMS.  This would tell LMS that in fact it is final. 

Not sure I understand the above question.  Can you explain?

bderusha: If revoke isn't actually part of the entitlement model this breaks a lot of standards and expectations.  I don't understand what telling the LMS that this is final is meant to prevent or ensure.  Can you explain more?

If we do need some kind of assurance I would prefer to have a separate endpoint or argument  and not pass this as part of the payload.  But ideally I want to know why we need it at all

mjfrey: I was trying to mimic the entitlement API, there is the explicit active attribute.  The other thing is that revoke is just calling the same entitlement API put endpoint.  
Do you suggest we just fill out the expired_at data?

mjfrey:
@bderusha @staubina ready for a second round.  Now calling the new Entitlement API.

staubina (approved):
:shipit: 👍

ecommerce/extensions/catalogue/migrations/0027_catalogue_entitlement_option.py
```diff
@@ -0,0 +1,33 @@
...
+def create_entitlement_option(apps, schema_editor):
+    """ Create catalogue entitlement option. """
+    course_entitlement_option = Option()
+    course_entitlement_option.name = 'Course Entitlement'
+    course_entitlement_option.code = 'Course Entitlement'
+    course_entitlement_option.type = Option.OPTIONAL
+    course_entitlement_option.save()
+
+
+def remove_catalog_option(apps, schema_editor):
```
bderusha: rename to `remove_entitlement_option`

ecommerce/extensions/catalogue/migrations/0027_catalogue_entitlement_option.py
```diff
@@ -0,0 +1,33 @@
...
+    """ Remove course entitlement option """
+
```
bderusha: Can we add the rollback functionality

ecommerce/extensions/catalogue/migrations/0026_course_entitlement_attr_change.py
```diff
@@ -0,0 +1,39 @@
...
+    course_entitlement_class = ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME)
+    product_attr = ProductAttribute.objects.get(product_class=course_entitlement_class, name="course_key")
+
+    product_attr.name = 'UUID'
+    product_attr.code = 'UUID'
+    product_attr.save()
+
+
+def remove_product_attr(apps, schema_editor):
+    """ Remove product attr """
```
bderusha, line 27: Can we add the rollback functionality

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,118 @@ def send_email(self, order):
...
+
+            data = {
+                'username': order.user.username,
+                'course_uuid': UUID,
+                'mode': mode,
+                'order_number': order.number,
+            }
+
+            try:
+                entitlement_option = Option.objects.get(name='Course Entitlement')
```
bderusha: we should get by `code` as a practice since that should never change.

Might also be worth updating the migration to differentiate the two values like `course_entitlement` or something

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,118 @@ def send_email(self, order):
...
+
+                entitlement_api_client = EdxRestApiClient(
+                    get_lms_entitlement_api_url(),
+                    jwt=order.site.siteconfiguration.access_token
+                )
+
+                # POST to the Entitlement API.
+                response = entitlement_api_client.entitlements.post(data)
+                results = response['results']
```
bderusha, line 574: I'm assuming we don't need to do any response code checks at this level, but wanted to double check that it's being handled by the client

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,118 @@ def send_email(self, order):
...
+                    'line_fulfilled',
+                    order_line_id=line.id,
+                    order_number=order.number,
+                    product_class=line.product.get_product_class().name,
+                    UUID=UUID,
+                    mode=mode,
+                    user_id=order.user.id,
+                )
+            except Exception:  # pylint: disable=broad-except
+                logger.error(
```
bderusha: `logger.exception` so we also get the stack trace in case it's something other than a network error

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,118 @@ def send_email(self, order):
...
+                    'Unable to fulfill line [%d] of order [%s] due to a network problem', line.id, order.number
```
bderusha: with the generic exception we do not know that it was a network error.

Might be worth keeping the original exceptions with this error and then add the broad exception with a different message

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,118 @@ def send_email(self, order):
...
+
+        logger.info('Finished fulfilling "Course Entitlement" product types for order [%s]', order.number)
+        return order, lines
+
+    def revoke_line(self, line):
+        try:
+            logger.info('Attempting to revoke fulfillment of Line [%d]...', line.id)
+
+            UUID = line.product.attr.UUID
+            entitlement_option = Option.objects.get(name='Course Entitlement')
```
bderusha: same comment.  get by `code`

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -507,3 +509,118 @@ def send_email(self, order):
...
+            course_entitlement_uuid = line.attributes.get(option=entitlement_option).value
+
+            entitlement_api_client = EdxRestApiClient(
+                get_lms_entitlement_api_url(),
+                jwt=line.order.site.siteconfiguration.access_token
+            )
+
+            # DELETE to the Entitlement API.
+            entitlement_api_client.entitlements(course_entitlement_uuid).delete()
```
bderusha, line 615: do we need to check the response code here to ensure that it was revoked?

bderusha (changes requested):

ecommerce/extensions/catalogue/migrations/0026_course_entitlement_attr_change.py
mjfrey: do we need it?  There will be FK to things we do not want to delete.

ecommerce/extensions/catalogue/migrations/0027_catalogue_entitlement_option.py
mjfrey: Same comment as above.

ecommerce/extensions/fulfillment/modules.py
mjfrey: Yes, we rely on the client.  If it is a bad response we will get an exception.

ecommerce/extensions/fulfillment/modules.py
mjfrey: No.  If it did not work we would get an exception.

ecommerce/extensions/catalogue/migrations/0026_course_entitlement_attr_change.py
bderusha: If something goes wrong for some reason and we do a rollback to older code the code will break because it expects `course_key` but the database only has `UUID` since we didnt actually reverse the change

ecommerce/extensions/catalogue/migrations/0027_catalogue_entitlement_option.py
```diff
@@ -0,0 +1,33 @@
...
+    """ Create catalogue entitlement option. """
+    course_entitlement_option = Option()
+    course_entitlement_option.name = 'Course Entitlement'
+    course_entitlement_option.code = 'course_entitlement'
+    course_entitlement_option.type = Option.OPTIONAL
+    course_entitlement_option.save()
+
+
+def remove_entitlement_option(apps, schema_editor):
+    """ Remove course entitlement option """
```
bderusha, line 21: This code is better since there is no older code referencing anything.  I don't see why we can't just call `course_entitlement_option.delete()` for completeness though.

bderusha (approved):

bderusha:
2 comments on the migration rollback.  I don't feel like I need to review the changes since they are pretty straightforward so I gave it a preemptive thumb

PR #1508, opened 2017-10-12
 Fix order and user admin list page 

ahsan-ul-haq:
Order and user admin page broken becacuse of large data.
Added switch to get_queryset to not query database

LEARNER-2806


ecommerce/core/admin.py
```diff
@@ -25,6 +25,12 @@ class EcommerceUserAdmin(UserAdmin):
     )
     show_full_result_count = False
 
+    def get_queryset(self, request):
+        if not waffle.switch_is_active(ORDER_LIST_VIEW_SWITCH):
+            return
```
clintonb: You should always return a `QuerySet`. See [`none()`](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#none).

ecommerce/core/admin.py
```diff
@@ -25,6 +25,12 @@ class EcommerceUserAdmin(UserAdmin):
     )
     show_full_result_count = False
 
+    def get_queryset(self, request):
+        if not waffle.switch_is_active(USER_LIST_VIEW_SWITCH):
+            return User.objects.none()
```
clintonb: Add a message, informing the user what's going on. Do the same for the one below.

```python
# Translators: "Waffle" is the name of a third-party library. It should not be translated
msg = _('User administration has been disabled due to the load on the database. This functionality can be restored by activating the {switch_name} Waffle switch. Be careful when re-activating this switch!').format(USER_LIST_VIEW_SWITCH)

self.message_user(request, msg, level=messages.WARNING)
```

ecommerce/core/admin.py
```diff
@@ -25,6 +25,12 @@ class EcommerceUserAdmin(UserAdmin):
...
+
+        return super(EcommerceUserAdmin, self).get_queryset(request)
+
     def changelist_view(self, request, extra_context=None):
         if not waffle.switch_is_active(USER_LIST_VIEW_SWITCH):
```
clintonb: This is no longer needed. The template can be removed, too.

clintonb (changes requested):

ecommerce/core/admin.py
schenedx: If we are using a waffle flag, the template is still needed, right?

schenedx (commented):
LGTM after the @clintonb's suggestion is addressed

ahsan-ul-haq:
@clintonb made suggested changes

schenedx (approved):
👍

ecommerce/extensions/order/tests/test_admin.py
```diff
@@ -27,4 +27,9 @@ def test_changelist_view_disable_switch(self):
         toggle_switch(ORDER_LIST_VIEW_SWITCH, False)
         response = self.client.get(self.order_page_url)
         self.assertEqual(response.status_code, 200)
-        self.assertContains(response, 'List view is temporarily disabled due to large number of records.')
+        self.assertContains(
```
clintonb: There is a proper way to validate messages: https://stackoverflow.com/a/16143864/592820.

ecommerce/extensions/order/tests/test_admin.py
```diff
@@ -26,5 +26,10 @@ def test_changelist_view_disable_switch(self):
         """ Overridden template will load on the list page, if the switch is disabled. """
         toggle_switch(ORDER_LIST_VIEW_SWITCH, False)
         response = self.client.get(self.order_page_url)
+        msg = 'User administration has been disabled due to the load on the database. ' \
+              'This functionality can be restored by activating the {switch_name} Waffle switch. ' \
+              'Be careful when re-activating this switch!'.format(switch_name=ORDER_LIST_VIEW_SWITCH)
         self.assertEqual(response.status_code, 200)
-        self.assertContains(response, 'List view is temporarily disabled due to large number of records.')
+        message = list(response.context.get('messages'))[0]
+        self.assertEqual(message.tags, "warning")
```
clintonb: `self.assertEqual(message.level, messages.WARNING)`

ecommerce/extensions/order/tests/test_admin.py
```diff
@@ -26,5 +26,10 @@ def test_changelist_view_disable_switch(self):
...
+        self.assertTrue(msg in message.message)
```
clintonb: `self.assertEqual(message.message, msg)`

clintonb (commented):
Update translations.

clintonb (commented):
Include the the .mo and .po files for both the en and es language codes.

ahsan-ul-haq:
@clintonb made suggested changes

ecommerce/extensions/order/admin.py
```diff
@@ -22,18 +24,19 @@ class OrderAdminExtended(OrderAdmin):
     show_full_result_count = False
 
     def get_queryset(self, request):
+        if not waffle.switch_is_active(ORDER_LIST_VIEW_SWITCH):
+            # Translators: "Waffle" is the name of a third-party library. It should not be translated
+            msg = _('User administration has been disabled due to the load on the database. '
```
clintonb: s/User/Order

clintonb (approved):

clintonb:
You forgot to update translations after modifying the string for the order admin. No big deal, but please submit an update at your convenience.

PR #1510, opened 2017-10-12
Updated translation for order admin page

ahsan-ul-haq:
updated the missing translations for order admin page
LEARNER-2806


clintonb:
You don't need my approval for a translation update.

PR #1511, opened 2017-10-13
Clean history bug resolved

ahsan-ul-haq:
For a large data using batch size was not filtering proberly.
So updated the batch logic
LEARNER-2697

### DIFF
--- a/ecommerce/core/management/commands/clean_history.py
+++ b/ecommerce/core/management/commands/clean_history.py
@@ -63,21 +63,34 @@ class Command(BaseCommand):
         )
 
         for model in models:
-            qs = model.history.filter(history_date__lte=cutoff_date)
+            qs = model.history.filter(history_date__lte=cutoff_date).order_by('-pk')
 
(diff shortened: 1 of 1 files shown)


ecommerce/core/management/commands/clean_history.py
```diff
@@ -63,21 +63,32 @@ def handle(self, *args, **options):
...
+            qs = model.history.filter(history_date__lte=cutoff_date).order_by('-pk')
 
             message = 'Cleaning {} rows from {} table'.format(qs.count(), model.__name__)
             logger.info(message)
 
-            qs = qs[:batch_size]
-            while qs.exists():
-                history_batch = list(qs.values_list('id', flat=True))
+            try:
+                max_pk = qs[0].pk
```
schenedx: can you explain why we need to figure out the batch index based on the `pk`?

schenedx: It would be nice to see the explanation in a comment in the code

schenedx (approved):
Just a #NIT 
👍

PR #1515, opened 2017-10-18
Refactor offer code to keep it DRY in preparation for Enterprise Offers feature.

douglashall:
Separating first commit from https://github.com/edx/ecommerce/pull/1475 into separate PR at the request of @mjfrey.


ecommerce/programs/constants.py
```diff
@@ -1,19 +1,14 @@
...
 from ecommerce.programs.benefits import AbsoluteDiscountBenefitWithoutRange, PercentageDiscountBenefitWithoutRange
-from ecommerce.programs.custom import class_path
 
 Benefit = get_model('offer', 'Benefit')
 
 BENEFIT_MAP = {
     Benefit.FIXED: AbsoluteDiscountBenefitWithoutRange,
     Benefit.PERCENTAGE: PercentageDiscountBenefitWithoutRange,
 }
-BENEFIT_PROXY_CLASS_MAP = dict(
```
mjfrey, line 14: Why the removal of this?

douglashall: It allows us to remove [this](https://github.com/edx/ecommerce/pull/1515/files#diff-9ca3ee6b16cb77b48ebe60de5adbbae5L61) TODO and is replaced with a better way of getting the benefit type string from the proxy class. That better way is by pulling it off of the proxy class itself instead of relying on this map. It wouldn't make any sense to add the two new Enterprise-related Benefit proxy class types the this dict in this program-related module.

douglashall: You can see the new benefits being added [here](https://github.com/edx/ecommerce/pull/1475/files#diff-afd5a30a03163c7ceab3a5a481e98475R10).

douglashall:
@mjfrey Can you give this an explicit approval when you are ready? Thanks.

mjfrey (approved):

clintonb:
This has broken deployments. I can't even checkout locally.

https://gocd.tools.edx.org/go/tab/build/detail/stage-ecommerce/259/build_ami/2/stage_edx

PR #1516, opened 2017-10-18
Fix Currency Translation Issue for Percentage Vouchers

jaebradley:
# Introduction

Related to PR #1498 

When the coupon text field does not contain a USD value, the `formatToLocalPrice` method throws an exception. This exception causes the `onReady` method to exit without setting callbacks that are necessary for successful payment completion.

To fix this, I've moved the `addPriceDisclaimer` and `translateToLocalPrices` method calls to the end of the `onReady` method after the callbacks have been set. Additionally, I've fixed the issue with `generateLocalPriceText` when it does not match a USD value.

## % Coupon Added And Order Complete
image
image

## USD Coupon And Order Complete
image
image

cc: @mikedikan @mjfrey @AlasdairSwan 

Test checklist
- [ ] Without coupon
- [ ] With discreet USD value coupon
- [ ] With % coupon
- [ ] Multiple coupons

- [ ] Repeat for with cookie / without cookie
- [ ] Repeat for single course / bundles

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -259,38 +259,59 @@ define([
             },
 
             generateLocalPriceText: function(usdPriceText) {
-                var localPriceText = usdPriceText;
-                // Matches on all $ab.cd strings
-                usdPriceText.match(/\$[0-9]+\.[0-9]+/g).forEach(function(value) {
-                    localPriceText = localPriceText.replace(value, BasketPage.formatToLocalPrice(value.substring(1)));
-                });
+                var localPriceText = usdPriceText,
+                    // Matches on all $ab.cd strings
+                    usdPriceMatches = usdPriceText.match(/\$[0-9]+\.[0-9]+/g);
+
+                // If matches exist translate
+                if (usdPriceMatches) {
+                    usdPriceMatches.forEach(function(value) {
+                        localPriceText = localPriceText
+                            .replace(value, BasketPage.formatToLocalPrice(value.substring(1)));
+                    });
+                }
+
                 return localPriceText;
             },
 
-            translateElementToLocalPrices: function(element) {
-                var priceText = element.text(),
-                    localPriceText = BasketPage.generateLocalPriceText(priceText);
-                if (priceText !== localPriceText) {
-                    element.text(localPriceText);
+            replaceElementText: function(element, text) {
+                if (element.text() !== text) {
+                    element.text(text);
                 }
             },
 
             translateToLocalPrices: function() {
-                $('.price').each(function() {
-                    BasketPage.translateElementToLocalPrices($(this));
+                var translatedItemPrices = [],
+                    translatedVoucherPrices = [],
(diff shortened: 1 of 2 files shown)


MatthewPiatetsky:
@jaebradley can you tell what code not executing caused the symptom of the card_type not being populated? i'm wondering if we can write a test that just causes the local currency code to throw a random error and verify that the card_type is still populated

jaebradley:
@MatthewPiatetsky 

In [`cybersource.js` there's a `cardType:detected` event handler](https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/payment_processors/cybersource.js#L41-L43).

This event gets triggered by the [`detectCreditCard` method in `basket_page.js`](https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/pages/basket_page.js#L170) which is set as an event callback in the `onReady` method of `basket_page` ([for example](https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/pages/basket_page.js#L439)).

If the callback doesn't get set, then Cybersource is not sent the card type.

MatthewPiatetsky:
👍 this looks good to me, i can help double check when we deploy this that it does not trigger the same issue.
lets wait to have @mjfrey review this as well

ecommerce/static/js/pages/basket_page.js
```diff
@@ -499,6 +504,14 @@ define([
                         $btn.prev('disabled', true);
                     }
                 });
+
+                try {
+                    // local currency translation
+                    BasketPage.addPriceDisclaimer();
+                    BasketPage.translateToLocalPrices();
+                } catch (e) {
+                    console.error(e); // eslint-disable-line no-console
```
AlasdairSwan: Do we want to show the user a console error? Is there a better way to track these errors?

ecommerce/static/js/pages/basket_page.js
```diff
@@ -499,6 +504,12 @@ define([
                         $btn.prev('disabled', true);
                     }
                 });
+
+                try {
+                    // local currency translation
+                    BasketPage.addPriceDisclaimer();
+                    BasketPage.translateToLocalPrices();
+                } catch (e) {}
```
mjfrey: What is the effect here?  What happens when there is an exception in either one of these routines?

jaebradley: @mjfrey and I talked about this offline. There was a slight problem with the current implementation. If `generateLocalPriceText` ever threw an exception, some of the price and voucher elements could theoretically be currency converted but others would not because I was replacing the price and voucher element text in-place. Instead, I get the full array of currency converted values and then apply these values after iterating through all the price and voucher elements.

mjfrey (approved):
better.  Let's test the heck out of this on stage.

AlasdairSwan (approved):
👍

jaebradley:
Test results on Stage

[stage-test.zip](https://github.com/edx/ecommerce/files/1402668/stage-test.zip)

PR #1518, opened 2017-10-19
CyberSource Secure Acceptance config is now optional

clintonb:
The configuration for Secure Acceptance is now optional. However,
settings for one of Secure Acceptance or Silent Order POST are required.

Additionally, the unused receipt_path setting has been removed from both
the CyberSource and PayPal settings.

LEARNER-818


clintonb:
@edx/learner please review

PR #1519, opened 2017-10-19
Refactor offer code to keep it DRY in preparation for Enterprise Offers feature.

douglashall:


douglashall:
@mjfrey @clintonb Reintroducing this work now that we have ironed out the deployment issue with https://github.com/edx/configuration/pull/4164.

PR #1520, opened 2017-10-20
put optimizely after jQuery

MatthewPiatetsky:
so we can use jQuery in optimizely on ecommerce without doing a wait


clintonb (approved):
Update your commit message to follow the best practices: https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/161880915/Git+Best+Practices#GitBestPractices-CommitMessages

PR #1521, opened 2017-10-22
[ENT-720] Fix error when trying to apply a voucher.

UmanShahzad:
Without this, there are cases where people try to apply a coupon on the basket page and because there are no lines in the basket, they get an error since some code assumes there's always at least one line. So we just ensure lines exist in the basket before accessing them.

**JIRA tickets**: [ENT-720](https://openedx.atlassian.net/browse/ENT-720)

**Merge deadline**: End of sprint.

**Testing instructions**:

I don't know how to get into a situation where the basket has no lines when trying to apply the coupon -- sorry, just lack of eCommerce knowledge here :\ Tried looking but couldn't figure it out. @douglashall ?

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -464,7 +464,7 @@ class VoucherAddView(BaseVoucherAddView):  # pylint: disable=function-redefined
 
     def form_valid(self, form):
         code = form.cleaned_data['code']
-        if not self.request.basket.id:
+        if self.request.basket.is_empty:
             return redirect_to_referrer(self.request, 'basket:summary')
         if self.request.basket.contains_voucher(code):
             messages.error(
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -769,12 +769,14 @@ class VoucherAddViewTests(LmsApiMockMixin, TestCase):
 
     def test_no_voucher_error_msg(self):
         """ Verify correct error message is returned when voucher can't be found. """
+        self.basket.add_product(ProductFactory())
         self.assert_form_valid_message("Coupon code '{code}' does not exist.".format(code=COUPON_CODE))
 
     def test_voucher_already_in_basket_error_msg(self):
         """ Verify correct error message is returned when voucher already in basket. """
-        voucher = factories.VoucherFactory(code=COUPON_CODE)
+        voucher, product = prepare_voucher(code=COUPON_CODE)
         self.basket.vouchers.add(voucher)
+        self.basket.add_product(product)
         self.assert_form_valid_message(
             "You have already added coupon code '{code}' to your basket.".format(code=COUPON_CODE))
 


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=1521&repo=edx%2Fecommerce"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

douglashall (commented):
Looks like we need a new test case.

UmanShahzad:
@douglashall Yep, will be doing that today or tomorrow. Would maybe shed some light on how this happens in practice.

mjfrey:
I do not know what causes this state.  I think just protecting against it is fine.

UmanShahzad:
@mjfrey Alright, no problem.

ecommerce/extensions/basket/views.py
```diff
@@ -480,46 +480,47 @@ def form_valid(self, form):
...
-                            'code': code,
-                            'sku': stock_record.partner_sku,
-                            'failure_url': self.request.build_absolute_uri(
-                                '{path}?{params}'.format(
-                                    path=reverse('basket:summary'),
-                                    params=urlencode(
-                                        {
-                                            CONSENT_FAILED_PARAM: code
-                                        }
+                if self.request.basket.num_lines:
```
mjfrey: I would just at the top of this block check to see if the basket is empty and if so -- re-direct to the summary page.

mjfrey (changes requested):

ecommerce/extensions/basket/views.py
UmanShahzad: :+1:

UmanShahzad: The basket is empty if `id is None and num_lines == 0`, so I'll just change the check for the ID above to check for emptiness instead. In that case, we won't even need this new check here, making the overall changeset a 1 line diff.

UmanShahzad:
@douglashall @mjfrey One last look? This time I don't think we'll need any new test.

mjfrey (approved):

UmanShahzad:
Had to fix a tiny test.

ecommerce/extensions/basket/views.py
```diff
@@ -464,7 +464,7 @@ def apply_voucher_to_basket(self, voucher):
 
     def form_valid(self, form):
         code = form.cleaned_data['code']
-        if not self.request.basket.id:
+        if self.request.basket.is_empty:
```
douglashall, line 467: We are okay making the assumption here that if a basket is not empty, then it has an id?

mjfrey: if you look at oscar, they check for that in the is_empty method.

douglashall: Do we have a test for this condition?

douglashall: @mjfrey Awesome. Thanks.

douglashall (approved):

ecommerce/extensions/basket/views.py
UmanShahzad: Yep, tests already exist. I'm surprised this wasn't set to `is_empty` in the first place. We should search through the codebase for other similar scenarios. Might make an OSPR for that.

itsjeyd (commented):
@UmanShahzad :+1:

- [x] ~~I tested this:~~ [N/A](https://github.com/edx/ecommerce/pull/1521#issuecomment-338626152)
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~ N/A
- [x] ~~Includes documentation~~ N/A

PR #1523, opened 2017-10-23
Add currency cookie validation

jaebradley:
Add more stringent currency cookie validation before attempting currency conversion in case cookie is malformed.

TODO:
- [x] squash commits


MatthewPiatetsky:
👍 after squashing commits

AlasdairSwan (approved):
Once commits are squashed :shipit:

PR #1525, opened 2017-10-25
Price value prefix checking

jaebradley:
# Introduction

When attempting local currency translation we look at the value of all HTML elements where the class name is `.price` or `.voucher`. However, not all values start with `$` - when the local language selection is Spanish the prefix is `USD`

spanish

# Discussion

The initial idea was to add a data attribute in the [client side checkout basket HTML template](https://github.com/edx/ecommerce/blob/master/ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html#L48), however, it turns out when you do that, the `basket.total_incl_tax_excl_discounts` value is locale formatted.

spanish_locale_format

### Spanish (`es-419`)

spanish

### Arabic (`ar`)

arabic

### French (`fr`)

french

### Simplified Chinese (`zh-cn`)

chinese

Since it seems there are currently only two variants for price formatting (`$` prefixed and `USD` prefixed) for all our language options, modify the existing currency conversion logic to take into account these prefixes, and if either of these prefixes don't exist, fallback to the existing value.

image

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -243,6 +243,10 @@ define([
             addPriceDisclaimer: function() {
                 var price = $('#basket-total').find('> .price').text(),
                     countryData = Cookies.getJSON('edx-price-l10n');
+
+                // when the price value has a USD prefix, replace it with a $
+                price = price.replace('USD', '$');
+
                 if (BasketPage.isValidLocalCurrencyCookie(countryData) && countryData.countryCode !== 'USA') {
                     $('<span>').attr('class', 'price-disclaimer')
                         .text('* This total contains an approximate conversion. You will be charged ' + price + ' USD.')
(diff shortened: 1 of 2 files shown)


ecommerce/static/js/pages/basket_page.js
```diff
@@ -260,15 +260,14 @@ define([
...
-                if (usdPriceMatches) {
-                    usdPriceMatches.forEach(function(value) {
-                        localPriceText = localPriceText
-                            .replace(value, BasketPage.formatToLocalPrice(value.substring(1)));
-                    });
+                    dollarPrefix = '$',
+                    usdPrefix = 'USD';
+
+                // assumes all prices start with $ or USD
+                if (localPriceText.startsWith(dollarPrefix)) {
```
AlasdairSwan: Support for ```startsWith``` is not there for all the browsers edX supports (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith#Browser_compatibility).

AlasdairSwan (changes requested):

ecommerce/static/js/pages/basket_page.js
```diff
@@ -302,6 +303,9 @@ define([
                 // Replace local price text values
                 if ($prices.length === translatedItemPrices.length
                     && $vouchers.length === translatedVoucherPrices.length) {
+                    // Add price disclaimer after all translated values have been calculated
+                    BasketPage.addPriceDisclaimer();
```
MatthewPiatetsky, line 307: can you help me understand why you moved this up and what the code after this line does?

MatthewPiatetsky:
👍 squash commits/restart the travis build in travis

jaebradley:
@AlasdairSwan mind taking another 👀 ?

AlasdairSwan (approved):
👍

PR #1526, opened 2017-10-25
Do not check for data sharing consent when checking for valid Enterprise Offers.

douglashall:
We originally implemented a check to make sure that the learner had
granted consent for the course runs that were in the basket when
deciding if a given Enterprise Offer should be applied to the basket.
This prevents us from accurately displaying the discounted price on
the enterprise enrollment landing pages. This PR removes the data
sharing consent check. We will rely on the Enterprise onboarding
workflow to ensure that the learner has granted consent before they
are able to purchase the course seat at the discounted price.

ENT-638


mjfrey (approved):

UmanShahzad:
Yeah, I think DSC being handled elsewhere is probably preferred -- we seem to have so many places where we do the check!

The travis log failures make no sense to me, though.

PR #1527, opened 2017-10-25
Wrap disclaimer text

jaebradley:
How things look when currency converted, the language is Spanish, and a % discount is applied

image

How things look when currency converted, the language is English, and a % discount is applied

image


MatthewPiatetsky:
👍

PR #1528, opened 2017-10-26
Default to local price prefix instead of $

jaebradley:
When a price has insufficient information to be currency converted (invalid currency cookie, for example) return the price value with a `$` prefix. However, for `es-419` users, the prefix for US dollars is `USD`.


MatthewPiatetsky:
👍

PR #1529, opened 2017-10-26
Updated Travis to build an image for testing

clintonb:
Previously we relied on the firefox and xvfb packages living in our images. This is no longer the case as these packages are not needed for production-like environments. We now create a new image, with the "test" tag, that includes these packages.

LEARNER-818


clintonb:
FYI @edx/learner @edx/devops 

Note that this image will _not_ be pushed to Docker Hub.

PR #1532, opened 2017-10-27
ENT-734 Add ability to apply a voucher to a basket with an Enterprise Offer.

douglashall:
Voucher-based offers should take precedence over any other offers
that may be valid for a given basket. This change will set the
priority on new ConditionalOffers higher than Enterprise offers
or Program offers.

ENT-734

*Testing Instructions*
1. Visit https://business.sandbox.edx.org/enterprise/76afe510-a9fd-4b59-9ba1-49b537864205/course/course-v1:Microsoft+DAT206x+2017_T4/enroll/?catalog=a38ef858-9849-4e76-8624-55915d8d33ff&utm_medium=enterprise&utm_source=microsoft.
2. Sign in via TestShib.
3. Create account.
4. Selected Verified and click continue.
5. Grant consent to share data.
6. Apply code 20OFF to basket.
7. Verify the code discount is applied instead of the enterprise offer discount.

### DIFF
--- a/ecommerce/extensions/test/factories.py
+++ b/ecommerce/extensions/test/factories.py
@@ -9,6 +9,7 @@ from oscar.test.factories import *  # pylint:disable=wildcard-import,unused-wild
 
 from ecommerce.enterprise.benefits import EnterpriseAbsoluteDiscountBenefit, EnterprisePercentageDiscountBenefit
 from ecommerce.enterprise.conditions import EnterpriseCustomerCondition
+from ecommerce.extensions.offer.models import OFFER_PRIORITY_VOUCHER
 from ecommerce.programs.benefits import AbsoluteDiscountBenefitWithoutRange, PercentageDiscountBenefitWithoutRange
 from ecommerce.programs.conditions import ProgramCourseRunSeatsCondition
 from ecommerce.programs.custom import class_path
@@ -104,7 +105,8 @@ def prepare_voucher(code='COUPONTEST', _range=None, start_datetime=None, end_dat
             benefit=benefit,
             condition=condition,
             max_global_applications=max_usage,
-            email_domains=email_domains
+            email_domains=email_domains,
+            priority=OFFER_PRIORITY_VOUCHER
         )
     else:
         offer = ConditionalOfferFactory(
@@ -112,7 +114,8 @@ def prepare_voucher(code='COUPONTEST', _range=None, start_datetime=None, end_dat
             benefit=benefit,
             condition=condition,
             email_domains=email_domains,
-            site=site
+            site=site,
+            priority=OFFER_PRIORITY_VOUCHER
         )
     voucher.offers.add(offer)
     return voucher, product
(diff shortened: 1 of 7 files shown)


ecommerce/extensions/voucher/utils.py
```diff
@@ -377,7 +377,10 @@ def _get_or_create_offer(
         benefit=offer_benefit,
         max_global_applications=max_uses,
         email_domains=email_domains,
-        site=site
+        site=site,
+        # Setting priority here so that Voucher-based offers take precedence
+        # over Program(priority=0) and Enterprise(priority=10) offers.
+        priority=20,
```
mjfrey: Can we create priority constants so they can be more descriptive.
VOUCHER_PRIORITY = 20
ENTERPRISE_OFFER_PRIORITY = 10
...

ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -199,6 +199,7 @@ def test_create_vouchers(self):
         self.assertEqual(voucher_offer.benefit.value, 100.00)
         self.assertEqual(voucher_offer.benefit.range.catalog, self.catalog)
         self.assertEqual(voucher_offer.email_domains, email_domains)
+        self.assertEqual(voucher_offer.priority, 20)
```
mjfrey: Add a couple more tests to prove that a voucher wins over offers.

mjfrey (changes requested):

ecommerce/extensions/voucher/utils.py
douglashall: Done.

ecommerce/extensions/voucher/tests/test_utils.py
douglashall: I'm not sure it's worth adding a test like that. It would only be testing the Otto code which sorts the offers [here](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/applicator.py#L59). The tests we already have ensure that the offer/coupon creation forms will create offers with the appropriate priority.

douglashall:
@mjfrey Can you give this another look when you get a chance? Thanks.

douglashall:
@mjfrey Anything else to address here?

mjfrey:
Looks fine, I just still would love a couple tests to make sure Oscar is acting the way we think it should.

douglashall:
@mjfrey Where would you add such a test?

mjfrey:
In either the ProgramOffers Tests or the Condition tests.  Would just be nice to know when we update oscar that our assumptions are still valid and do not break our coupons / offers.

I would create the scenario that this PR assumes and just test it is true.

douglashall:
@mjfrey How does this look?

ecommerce/extensions/test/factories.py
```diff
@@ -104,15 +104,17 @@ def prepare_voucher(code='COUPONTEST', _range=None, start_datetime=None, end_dat
...
         )
     else:
         offer = ConditionalOfferFactory(
             offer_type=ConditionalOffer.VOUCHER,
             benefit=benefit,
             condition=condition,
             email_domains=email_domains,
-            site=site
+            site=site,
+            priority=20
```
mjfrey: OFFER_PRIORITY_VOUCHER

mjfrey (approved):
Yes thanks.  Just 2 small Nits.

PR #1533, opened 2017-10-27
Trailing text in price regex

jaebradley:


jaebradley:
[localtests.zip](https://github.com/edx/ecommerce/files/1422824/localtests.zip)

PR #1539, opened 2017-10-31
Updated documentation for payment processors

clintonb:
The documentation now explains how to configured the payment processors.


clintonb:
FYI @edx/learner @edx/doc

PR #1540, opened 2017-10-31
Extracted Apple Pay view and processor code out of CyberSource module

clintonb:
The Apple Pay code is not all CyberSource-specific. Extracting it allows it to be used with other payment processors.


clintonb:
FYI @edx/learner

PR #1541, opened 2017-10-31
Fixed bug in translations

clintonb:
ENT-638


clintonb:
FYI @douglashall please ensure you run `make fake_translations` when adding code that contains strings marked for translation.

douglashall (approved):

PR #1550, opened 2017-11-02
Management command for creating bulk orders

ahsan-ul-haq:
Bulk orders are needed to create for testing migrations on
huge date
LEARNER-3075

### DIFF
--- a/ecommerce/extensions/order/management/commands/create_fake_orders.py
+++ b/ecommerce/extensions/order/management/commands/create_fake_orders.py
@@ -1,0 +1,90 @@
+from __future__ import unicode_literals
+
+import logging
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+from oscar.core.loading import get_class, get_model
+from oscar.test.newfactories import UserFactory
+
+from ecommerce.core.models import SiteConfiguration
+
+logger = logging.getLogger(__name__)
+Order = get_model('order', 'Order')
+Basket = get_model('basket', 'Basket')
+Product = get_model('catalogue', 'Product')
+StockRecord = get_model('partner', 'StockRecord')
+Default = get_class('partner.strategy', 'Default')
+Free = get_class('shipping.methods', 'Free')
+OrderTotalCalculator = get_class('checkout.calculators', 'OrderTotalCalculator')
+OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
+OrderCreator = get_class('order.utils', 'OrderCreator')
+
+
+def create_basket(owner, product, site):
+    """Create basket to place order"""
+    basket = Basket.objects.create(site=site, owner=owner)
+    basket.strategy = Default()
+    basket.add_product(product)
+    basket.save()
+    return basket
+
+
+class Command(BaseCommand):
+    help = 'Clean history data'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--count',
+                            action='store',
+                            dest='orders',
+                            type=int,
+                            required=True,
+                            help='Number of orders to create.')
+        parser.add_argument('--sku',
+                            action='store',
+                            dest='sku',
+                            type=str,
+                            required=True,
+                            help='SKU corresponding to the product for which orders will be created.')
+
+    def handle(self, *args, **options):
+        orders = options['orders']
+        sku = options['sku']
+
+        try:
+            stock_record = StockRecord.objects.get(partner_sku=sku)
+            product = stock_record.product
+            partner = stock_record.partner
+        except StockRecord.DoesNotExist:
+            msg = 'No StockRecord for partner_sku {} exists.'.format(sku)
+            logger.exception(msg)
+            raise CommandError(msg)
+
+        try:
+            site = partner.siteconfiguration.site
+        except SiteConfiguration.DoesNotExist:
+            msg = 'No Site Configuration exists for partner {}!'.format(partner.id)
+            logger.exception(msg)
+            raise CommandError(msg)
+
+        user = UserFactory()
+
+        for __ in range(orders):
+            basket = create_basket(user, product, site)
+
+            shipping_method = Free()
+            shipping_charge = shipping_method.calculate(basket)
+            total = OrderTotalCalculator().calculate(basket, shipping_charge)
+            number = OrderNumberGenerator().order_number(basket)
+            with transaction.atomic():
+                OrderCreator().place_order(
+                    order_number=number,
+                    user=user,
+                    basket=basket,
+                    shipping_address=None,
+                    shipping_method=shipping_method,
+                    shipping_charge=shipping_charge,
+                    billing_address=None,
+                    total=total)
+
+                basket.set_as_submitted()
--- a/ecommerce/extensions/order/management/commands/tests/test_create_fake_orders.py
+++ b/ecommerce/extensions/order/management/commands/tests/test_create_fake_orders.py
@@ -1,0 +1,54 @@
+from django.core.management import call_command
+from django.core.management.base import CommandError
+from oscar.core.loading import get_model
+from oscar.test.factories import create_product, create_stockrecord
+from testfixtures import LogCapture
+
+from ecommerce.tests.factories import PartnerFactory, SiteConfigurationFactory
+from ecommerce.tests.testcases import TestCase
+
+
+LOGGER_NAME = 'ecommerce.extensions.order.management.commands.create_fake_orders'
+Order = get_model('order', 'Order')
+Basket = get_model('basket', 'Basket')
+
+
+class FakeOrdersTests(TestCase):
+
+    def test_product_does_not_exist(self):
+        with LogCapture(LOGGER_NAME) as log:
+            with self.assertRaises(CommandError):
+                call_command('create_fake_orders', '--count=5', '--sku=sku_11111')
+                log.check(
+                    (
+                        LOGGER_NAME,
+                        'EXCEPTION',
+                        'No StockRecord for partner_sku sku_11111 exists.'
+                    )
+                )
+
+    def test_site_configuration_does_not_exist(self):
+        partner = PartnerFactory()
+        product = create_product()
+        stockrecord = create_stockrecord(product=product, partner_name=partner.name)
+        with LogCapture(LOGGER_NAME) as log:
+            with self.assertRaises(CommandError):
+                call_command('create_fake_orders', '--count=5', '--sku={}'.format(stockrecord.partner_sku))
+                log.check(
+                    (
+                        LOGGER_NAME,
+                        'EXCEPTION',
+                        'No Site Configuration exists for partner {}!'.format(partner.id)
+                    )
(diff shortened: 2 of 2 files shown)


ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
+from __future__ import unicode_literals
```
clintonb: Move this to ecommerce/extentions/order/management/commands/create_fake_orders.py.

We want to keep order-related code together. Also, we want to make it clear that this data is fake and should not be used in production.

ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
...
+    basket.strategy = Default()
+    basket.add_product(product)
+    return basket
+
+
+class Command(BaseCommand):
+    help = 'Clean history data'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--orders',
```
clintonb: Let's call this count so that its purpose is obvious.

ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
...
+                            action='store',
+                            dest='orders',
+                            type=int,
+                            required=True,
+                            help='Number of orders to be generated as bulk order.')
```
clintonb: The term "bulk order" implies something different for me. I would rephrase this to: Number of orders to create.

ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
...
+        parser.add_argument('--product_id',
```
clintonb: Use SKU as that is how we generally present products to end users/API clients. Also, the corresponding `StockRecord` will keep track of the price and currency being used.

ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if product_id:
+            try:
+                product = Product.objects.get(id=product_id)
+            except Product.DoesNotExist:
+                msg = 'No product for product_id {} exists.'.format(product_id)
+                logger.exception(msg)
+                raise CommandError(msg)
+
+            if not product.is_seat_product:
```
clintonb: Why does the product class matter? I see no reason to limit this to seat products.

ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
...
+                            action='store',
+                            dest='product_id',
+                            type=int,
+                            help='Product Id for which orders are to be generated.')
+
+    def handle(self, *args, **options):
+        orders = options['orders']
+        product_id = options['product_id']
+
+        if product_id:
```
clintonb: Make this required and avoid the extra work.

ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
...
+                logger.exception(msg)
+                raise CommandError(msg)
+
+            if not product.is_seat_product:
+                msg = 'Product does not have seat.'
+                logger.exception(msg)
+                raise CommandError(msg)
+
+        else:
+            products = Product.objects.all()
```
clintonb: This should go away but, in the future, just filter by product class if you only care about a single class of product.

ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
...
+                    break
+
+            if not product:
+                msg = 'No product with seat exists.'
+                logger.exception(msg)
+                raise CommandError(msg)
+
+        user = User.objects.order_by('id').first()
+
+        for _ in range(orders):
```
clintonb: Use double underscore, `__`. A single underscore, `_`, is typically reserved as an alias to `ugettext()`.

ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
...
+            basket = create_basket(user, product)
+            if not basket.id:
```
clintonb: Given that you wrote `create_basket`, you should guarantee the basket has an ID. This should go away.

ecommerce/core/management/commands/generate_bulk_orders.py
```diff
@@ -0,0 +1,100 @@
...
+                if product.is_seat_product:
+                    product = product
+                    break
+
+            if not product:
+                msg = 'No product with seat exists.'
+                logger.exception(msg)
+                raise CommandError(msg)
+
+        user = User.objects.order_by('id').first()
```
clintonb: I would either create a new user, making this a great way to also populate the user table, or allow the user to pass in the username. I'm kinda leaning toward making a new one.

clintonb (changes requested):

ecommerce/core/management/commands/generate_bulk_orders.py
ahsan-ul-haq: Are you referring to partner_sku?

clintonb: Correct.

clintonb: Also, note that you should probably use the partner from that SKU to get the `Site` that should be associated with the basket/order.

ahsan-ul-haq:
@clintonb made the suggested changes.
FYI: @edx/learner-spartans

ecommerce/extensions/order/management/commands/create_fake_orders.py
```diff
@@ -0,0 +1,89 @@
...
+    help = 'Clean history data'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--count',
+                            action='store',
+                            dest='orders',
+                            type=int,
+                            required=True,
+                            help='Number of orders to create.')
+        parser.add_argument('--partner_sku',
```
clintonb: Nit: "sku" is sufficient.

ecommerce/extensions/order/management/commands/create_fake_orders.py
```diff
@@ -0,0 +1,89 @@
...
+                            action='store',
+                            dest='partner_sku',
+                            type=str,
+                            required=True,
+                            help='Partner SKU for which orders are to be generated.')
```
clintonb: SKU corresponding to the product for which orders will be created.

ecommerce/extensions/order/management/commands/create_fake_orders.py
```diff
@@ -0,0 +1,89 @@
...
+            product = Product.objects.get(stockrecords__partner_sku=partner_sku)
+        except Product.DoesNotExist:
+            msg = 'No product for partner_sku {} exists.'.format(partner_sku)
+            logger.exception(msg)
+            raise CommandError(msg)
+
+        partner = StockRecord.objects.get(partner_sku=partner_sku).partner
+
+        try:
+            site = SiteConfiguration.objects.get(partner=partner).site
```
clintonb: `site = partner.siteconfiguration.site`

ecommerce/extensions/order/management/commands/create_fake_orders.py
```diff
@@ -0,0 +1,89 @@
...
+                            type=str,
+                            required=True,
+                            help='Partner SKU for which orders are to be generated.')
+
+    def handle(self, *args, **options):
+        orders = options['orders']
+        partner_sku = options['partner_sku']
+
+        try:
+            product = Product.objects.get(stockrecords__partner_sku=partner_sku)
```
clintonb: Minor, but you can decrease the number of queries:

```python
stock_record = StockRecord.objects.get(partner_sku=partner_sku)
product = stock_record.product
partner = stock_record.partner
```

ecommerce/extensions/order/management/commands/create_fake_orders.py
```diff
@@ -0,0 +1,89 @@
...
+            msg = 'No product for partner_sku {} exists.'.format(partner_sku)
+            logger.exception(msg)
+            raise CommandError(msg)
+
+        partner = StockRecord.objects.get(partner_sku=partner_sku).partner
+
+        try:
+            site = SiteConfiguration.objects.get(partner=partner).site
+        except SiteConfiguration.DoesNotExist:
+            msg = 'Site configuration for partner_sku {} does not exists.'.format(partner_sku)
```
clintonb: Site configuration is linked to partners, not SKUs.

`No Site Configuration exists for partner {}!'.format(partner.id)`

ecommerce/extensions/order/management/commands/create_fake_orders.py
```diff
@@ -0,0 +1,89 @@
...
+            logger.exception(msg)
+            raise CommandError(msg)
+
+        user = User.objects.create(username='fake_user', email='fake_user@example.com')
```
clintonb: This will fail the second time this command is run. Use `user = UserFactory()`

ecommerce/extensions/order/management/commands/tests/test_create_fake_orders.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+LOGGER_NAME = 'ecommerce.extensions.order.management.commands.create_fake_orders'
+Order = get_model('order', 'Order')
+Basket = get_model('basket', 'Basket')
+
+
+class FakeOrdersTests(TestCase):
+
+    def test_product_not_exist(self):
```
clintonb: `test_product_does_not_exist`

ecommerce/extensions/order/management/commands/tests/test_create_fake_orders.py
```diff
@@ -0,0 +1,54 @@
...
+                call_command('create_fake_orders', '--count=5', '--partner_sku=sku_11111')
+                log.check(
+                    (
+                        LOGGER_NAME,
+                        'EXCEPTION',
+                        'No product for partner_sku sku_11111 exists.'
+                    )
+                )
+
+    def test_site_config_not_exist(self):
```
clintonb: `test_site_configuration_does_not_exist`

clintonb (changes requested):

ecommerce/extensions/order/management/commands/create_fake_orders.py
ahsan-ul-haq: May be we can use get_or_create to avoid failure I am bit uncomfortable with using factories here.

clintonb: Write your own, non-test, factory then. The advantage of using the factory is that we get to create a large set of users. This is advantageous for times when we want to run tests on large user tables.

ahsan-ul-haq:

clintonb (approved):

PR #1552, opened 2017-11-03
Adding data migration for enterprise switch.

douglashall:
The `enable_enterprise_on_runtime` was created via Django admin in stage/prod. This PR adds a data migration to create the switch on provisioning of new environments.

ENT-479


douglashall:
@mjfrey Quick thumb on this one?

mjfrey (approved):

PR #1558, opened 2017-11-07
Truncate voucher names when they are longer than model allows

tuchfarber:
Long course names are breaking saving of vouchers. This truncates voucher names so they always fit.

[WL-1326]

### DIFF
--- a/ecommerce/extensions/voucher/tests/test_utils.py
+++ b/ecommerce/extensions/voucher/tests/test_utils.py
@@ -206,6 +206,21 @@ class UtilTests(CouponMixin, DiscoveryMockMixin, DiscoveryTestMixin, LmsApiMockM
         self.assertEqual(voucher.start_datetime, self.data['start_datetime'])
         self.assertEqual(voucher.usage, Voucher.SINGLE_USE)
 
+    def test_create_voucher_with_long_name(self):
+        self.data.update({
+            'name': (
+                'This Is A Really Really Really Really Really Really Long '
+                'Voucher Name That Needs To Be Trimmed To Fit Into The Name Column Of The DB'
+            )
+        })
+        trimmed = (
+            'This Is A Really Really Really Really Really Really Long '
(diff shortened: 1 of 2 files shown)


tuchfarber:
We currently have a 65 character course name that people can't buy enrollment codes for, so that's why I would prefer this fix. 

As far as I could tell, this name column was only for viewing and nothing depends on it's format.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -449,7 +449,7 @@ def fulfill_product(self, order, lines):
                 _range.add_product(seat)
 
             vouchers = create_vouchers(
-                name='Enrollment code voucher [{}]'.format(line.product.title),
+                name='Voucher: {}'.format(line.product.title),
```
douglashall: This seems like a stop gap for the current issue. Why don't we truncate the product title instead?

tuchfarber: (facepalm). You're right. Will fix.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -449,7 +449,7 @@ def fulfill_product(self, order, lines):
                 _range.add_product(seat)
 
             vouchers = create_vouchers(
-                name='Enrollment code voucher [{}]'.format(line.product.title),
+                name='Enrollment code voucher [{}]'.format(line.product.title[:90]), # model max-length 128
```
douglashall: This is good. I think we should do this further [down the stack](https://github.com/edx/ecommerce/blob/bec2e16fe450d9ccddfaa0d28eed74161f1a6e70/ecommerce/extensions/voucher/utils.py#L455) though.

tuchfarber: Just saw this. Updating.

tuchfarber:
@douglashall Truncates at the voucher creation now.

douglashall (approved):
It would be good to have a test for this change.

PR #1560, opened 2017-11-07
Removed JWT decode handler override for devstack

clintonb:
E-Commerce must still use its custom decode handler until we change the configuration/settings to match the format of the other IDAs.

LEARNER-818


clintonb:
FYI @staubina @edx/docker-devstack-working-group @edx/learner

staubina:
@clintonb Does this resolve the issues related to https://github.com/edx/configuration/pull/4192 also?

clintonb:
Yes, this completely resolves those issues. See https://github.com/edx/edx-internal/pull/251#issuecomment-342818056.

PR #1564, opened 2017-11-13
Update program offer logic to account for entitlements

McKenzieW:
Take existing user entitlements into account for partial program offers and calculate offers for programs with course entitlement products.

[LEARNER-2659](https://openedx.atlassian.net/browse/LEARNER-2659)

### DIFF
--- a/ecommerce/entitlements/utils.py
+++ b/ecommerce/entitlements/utils.py
@@ -43,11 +43,13 @@ def create_or_update_course_entitlement(certificate_type, price, partner, UUID, 
 
     certificate_type = certificate_type.lower()
     UUID = unicode(UUID)
+    entitlement_class = ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME)
 
     certificate_type_query = Q(
         title='Course Entitlement for {}'.format(name),
         attributes__name='certificate_type',
-        attribute_values__value_text=certificate_type
+        attribute_values__value_text=certificate_type,
+        product_class=entitlement_class
     )
 
     try:
@@ -63,6 +65,7 @@ def create_or_update_course_entitlement(certificate_type, price, partner, UUID, 
     course_entitlement.attr.certificate_type = certificate_type
     course_entitlement.attr.UUID = UUID
     course_entitlement.parent = parent_entitlement
+    course_entitlement.product_class = entitlement_class
     course_entitlement.save()
 
     StockRecord.objects.update_or_create(
--- a/ecommerce/extensions/api/v2/tests/views/test_baskets.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_baskets.py
@@ -388,7 +388,7 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
         )
         program_uuid = offer.condition.program_uuid
         self.mock_program_detail_endpoint(program_uuid, self.site_configuration.discovery_api_url)
-        self.mock_enrollment_api(self.user.username)
+        self.mock_user_data(self.user.username)
 
         response = self.client.get(self.url)
         expected = {
@@ -473,7 +473,7 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
         products = self._get_program_verified_seats(program)
         url = self._generate_sku_username_url(products, differentuser.username)
         enrollment = [{'mode': 'verified', 'course_details': {'course_id': program['courses'][0]['key']}}]
-        self.mock_enrollment_api(differentuser.username, enrollment)
+        self.mock_user_data(differentuser.username, owned_products=enrollment)
 
         expected = {
             'total_incl_tax_excl_discounts': sum(product.stockrecords.first().price_excl_tax
@@ -505,7 +505,7 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
         products = self._get_program_verified_seats(program)
         url = self._generate_sku_username_url(products, 'invalidusername')
         enrollment = [{'mode': 'verified', 'course_details': {'course_id': program['courses'][0]['key']}}]
-        self.mock_enrollment_api(differentuser.username, enrollment)
+        self.mock_user_data(differentuser.username, enrollment)
 
         expected = {
             'total_incl_tax_excl_discounts': sum(product.stockrecords.first().price_excl_tax
--- a/ecommerce/programs/conditions.py
+++ b/ecommerce/programs/conditions.py
@@ -10,7 +10,7 @@ from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
 
-from ecommerce.core.utils import get_cache_key
+from ecommerce.core.utils import get_cache_key, traverse_pagination
 from ecommerce.extensions.offer.decorators import check_condition_applicability
 from ecommerce.extensions.offer.mixins import SingleItemConsumptionConditionMixin
 from ecommerce.programs.utils import get_program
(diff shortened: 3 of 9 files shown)


McKenzieW:
@edx/helio I'm still having some issues with the updates I made for tests, but here's the PR so you can start getting a look at my code.

ecommerce/programs/conditions.py
```diff
@@ -30,18 +30,55 @@ def name(self):
...
+                    program_skus.update(
                         set([seat['sku'] for seat in course_run['seats'] if seat['type'] in applicable_seat_types])
                     )
+                for entitlement in course['entitlements']:
+                    if entitlement['mode'].lower() in applicable_seat_types:
+                        program_skus.add(entitlement['sku'])
+        return program_skus
 
-        return program_course_run_skus
+    def get_user_ownership_data(self, basket, retrieve_entitlements):
```
mjfrey: default should be False for retrieve_entitlements

ecommerce/programs/conditions.py
```diff
@@ -30,18 +30,55 @@ def name(self):
...
+        """
+        Retrieves existing enrollments and entitlements for a user from LMS
+        """
+        enrollments = []
+        entitlements = []
+        api_resources = ['enrollments']
```
mjfrey: I would create a tuple set to have the api that corresponds to the resource.

ecommerce/programs/conditions.py
```diff
@@ -30,18 +30,55 @@ def name(self):
...
+                cache_key = get_cache_key(
+                    site_domain=basket.site.domain,
+                    resource=api_resource,
+                    username=basket.owner.username,
+                )
+                user_data_list = cache.get(cache_key, [])
+                if not user_data_list:
+                    user = basket.owner.username
+                    try:
+                        if api_resource == 'enrollments':
```
mjfrey: use the tuple above to just write this once and eliminate the if clauses.

ecommerce/programs/conditions.py
```diff
@@ -100,6 +128,9 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
             skus = set()
             for course_run in course['course_runs']:
                 skus.update(set([seat['sku'] for seat in course_run['seats'] if seat['type'] in applicable_seat_types]))
+            for entitlement in course['entitlements']:
```
mjfrey, line 137: Will the element 'entitlements' always exists in the JSON?

mjfrey (changes requested):

ecommerce/programs/conditions.py
McKenzieW: Yes, it will just be an empty list if there aren't any entitlements

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -484,6 +484,10 @@ def test_basket_calculate_by_staff_user_other_username(self):
 
         response = self.client.get(url)
         self.assertEqual(response.status_code, 200)
+        print'Response Data : '
+        print response.data
+        print'Expected Data : '
+        print expected
```
staubina: Are these just here for debugging the tests you are currently working on?

ecommerce/programs/conditions.py
```diff
@@ -30,18 +30,55 @@ def name(self):
...
             applicable_seat_types = program['applicable_seat_types']
 
             for course in program['courses']:
                 for course_run in course['course_runs']:
-                    program_course_run_skus.update(
+                    program_skus.update(
                         set([seat['sku'] for seat in course_run['seats'] if seat['type'] in applicable_seat_types])
                     )
+                for entitlement in course['entitlements']:
+                    if entitlement['mode'].lower() in applicable_seat_types:
```
staubina: Is the lower() needed here?  The expectation elsewhere seems to be that the mode slugs will match the seat type slugs.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
McKenzieW: Yes

ecommerce/programs/conditions.py
```diff
@@ -30,18 +30,55 @@ def name(self):
...
+        """
+        Retrieves existing enrollments and entitlements for a user from LMS
+        """
+        enrollments = []
+        entitlements = []
+        api_resources = ['enrollments']
+        if retrieve_entitlements:
+            api_resources.append('entitlements')
+        if basket.site.siteconfiguration.enable_partial_program:
+            for api_resource in api_resources:
```
staubina: It seems like this should not be a loop and instead should just be two code paths, one for enrollments and one for entitlements. 

The loop seems unnecessary.

ecommerce/programs/conditions.py
```diff
@@ -100,6 +128,9 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
             skus = set()
             for course_run in course['course_runs']:
                 skus.update(set([seat['sku'] for seat in course_run['seats'] if seat['type'] in applicable_seat_types]))
+            for entitlement in course['entitlements']:
+                if entitlement['mode'].lower() in applicable_seat_types:
```
staubina, line 138: Same as earlier comment, is the lower() necessary?

staubina (changes requested):
Some comments on the code per our discussion.

Keep up the great work.

ecommerce/programs/conditions.py
```diff
@@ -30,18 +30,52 @@ def name(self):
...
+                        program_skus.add(entitlement['sku'])
+        return program_skus
+
+    def get_from_cache_or_endpoint(self, basket, resource_name, endpoint):
+        cache_key = get_cache_key(
+            site_domain=basket.site.domain,
+            resource=resource_name,
+            username=basket.owner.username,
+        )
+        data_list = cache.get(cache_key, [])
```
mjfrey: No need for the empty list.

ecommerce/programs/conditions.py
```diff
@@ -30,18 +30,52 @@ def name(self):
...
                         set([seat['sku'] for seat in course_run['seats'] if seat['type'] in applicable_seat_types])
                     )
-
-        return program_course_run_skus
+                for entitlement in course['entitlements']:
+                    if entitlement['mode'].lower() in applicable_seat_types:
+                        program_skus.add(entitlement['sku'])
+        return program_skus
+
+    def get_from_cache_or_endpoint(self, basket, resource_name, endpoint):
```
mjfrey: get_entitlements_and_enrollments

ecommerce/programs/tests/mixins.py
```diff
@@ -18,10 +26,53 @@ def mock_program_detail_endpoint(self, program_uuid, discovery_api_url, empty=Fa
         Returns:
             dict: Mocked program data.
         """
+        partner = PartnerFactory()
         data = None
         if not empty:
             courses = []
             for i in range(1, 5):
+                entitlement_class, created = ProductClass.objects.get_or_create(  # pylint: disable=unused-variable
```
mjfrey: you can just use the existing util class:
create_or_update_course_entitlement
https://github.com/edx/ecommerce/blob/master/ecommerce/entitlements/utils.py#L41

mjfrey (changes requested):

ecommerce/programs/conditions.py
McKenzieW: This method only retrieves one resource at a time, and can in theory be used to retrieve any data from LMS.  How about, `get_lms_resource`?

ecommerce/programs/tests/mixins.py
staubina: Why is the # pylint: disable=unused-variable needed?  Can we remove the unused variable?

We should avoid hiding pylint warnings.

McKenzieW: it's just that the method returns two variables and we only need one of them, but I'm refactoring this code to use a util method anyway.

staubina: If you only need one of the values you can do something like

```
def foo():
   return 'Hello', 'World'

one = foo()[0]
two = foo()[1]
```

McKenzieW: Probably, yes. I will keep that in mind for the future. But in this case I won't need to do that because the util method I'll be using just returns one thing

staubina: Just keep the following as an FYI then.

ecommerce/programs/conditions.py
McKenzieW: I do need the empty list somewhere because otherwise the method will return a NoneType object instead of a list if there aren't any enrollments or entitlements, but I'm adding it to the return statement instead of here

ecommerce/entitlements/utils.py
```diff
@@ -63,6 +65,7 @@ def create_or_update_course_entitlement(certificate_type, price, partner, UUID,
     course_entitlement.attr.certificate_type = certificate_type
     course_entitlement.attr.UUID = UUID
     course_entitlement.parent = parent_entitlement
+    course_entitlement.product_class = entitlement_class
```
mjfrey, line 68: Thanks for catching this.

ecommerce/programs/tests/mixins.py
```diff
@@ -2,26 +2,49 @@
...
 
-from ecommerce.core.url_utils import get_lms_enrollment_api_url
+from ecommerce.core.url_utils import get_lms_enrollment_api_url, get_lms_entitlement_api_url
 from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.courses.utils import mode_for_product
+from ecommerce.entitlements.utils import create_or_update_course_entitlement
 from ecommerce.extensions.catalogue.tests.mixins import DiscoveryTestMixin
+from ecommerce.tests.factories import PartnerFactory
+
+Product = get_model('catalogue', 'Product')
```
mjfrey: Are these used anymore?

ecommerce/programs/tests/mixins.py
```diff
@@ -2,26 +2,49 @@
...
             dict: Mocked program data.
         """
+        partner = PartnerFactory()
         data = None
         if not empty:
             courses = []
             for i in range(1, 5):
+                uuid = '268afbfc-cc1e-415b-a5d8-c58d955bcfc' + str(i)
+                entitlement = create_or_update_course_entitlement('verified', 10, partner, uuid, uuid)
+                sku = entitlement.stockrecords.first().partner_sku
```
mjfrey: no need to assign the variable if only using it once.

ecommerce/programs/tests/mixins.py
McKenzieW: The partner factory is because partner has to be passed to the util for creating course entitlement products, but the Product model is not

ecommerce/programs/tests/mixins.py
McKenzieW: I used uuid twice, as the uuid for the course and in the name of the entitlement to indicate which course it is for.

mjfrey (approved):
Remember to squash.

McKenzieW:
@mjfrey I just want to do a bit more testing locally with the new pagination handling before I merge

McKenzieW:
Tested locally and all the different combinations of basket lines and existing enrollments and entitlements are working as expected.

@staubina do you have any further comments before I merge?

staubina (approved):

PR #1565, opened 2017-11-14
Only cache enrollments for 30 seconds.

mjfrey:
Only cache user enrollments for 30 seconds as apposed to an hour.

This should reduce the size of memcache and make gets faster.


clintonb (approved):
If there is an associated ticket, please include it in the commit message.

PR #1567, opened 2017-11-17
ENT-755: Display 'Undefined' in case coupon history record not found to get last edit date.

saleem-latif:
**JIRA Ticket:** [ENT-755](https://openedx.atlassian.net/browse/ENT-755)

**Description:**
Coupon details page was not being rendered for certain coupons. I was able to find the root cause for the error.

Coupon details page displays "Last Edit" information segment in the header for coupon details page that contains the [username and date associated with the coupon's most recent modification](https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/views/coupon_detail_view.js#L142). This info is not itself saved in the `Product` aka `Coupon` model but is retrieved via the `ProductHistory` table. If there is no history record for a coupon than `last_edit` returned by coupon details API endpoint at `/api/v2/coupons/{coupon_id}/` is `None` which causes [coupon details view](https://github.com/edx/ecommerce/blob/master/ecommerce/static/js/views/coupon_detail_view.js#L43) to break.

I have added a fix to resolve the issue.

**Steps to Reproduce:**
Since, the issue appears randomly there is no simple instructions to reproduce it however with access of mysql console we can create conditions to cause the bug to appear.

1. Create a new coupon at https://ecommerce-business.sandbox.edx.org/coupons/new/
2. Note the coupon's id (it is the number at the end of coupon details page url, e.g. the number 131 in url https://ecommerce-business.sandbox.edx.org/coupons/131)
3. SSH into business sandbox and open mysql console and select ecommerce database and run this query
```
 DELETE FROM `catalogue_historicalproduct` WHERE `catalogue_historicalproduct`.`id` = {coupon_id};
```
4. Now, visit again coupon details page, it will not render and there will be an error in the console.

**Testing Instructions:**

1. Checkout this PR's branch and do 1-3 steps from above "Steps to Reproduce"
2. Visit coupon details page and make sure it is rendered correctly and make sure that `Last edit:` is display on the right below `Edit Coupon` button.

### DIFF
--- a/ecommerce/static/js/views/coupon_detail_view.js
+++ b/ecommerce/static/js/views/coupon_detail_view.js
@@ -120,6 +120,7 @@ define([
                     courseCatalogName = '',
                     invoiceData = this.formatInvoiceData(),
                     emailDomains = this.model.get('email_domains'),
+                    lastEdited = this.model.get('last_edited'),
                     templateData,
                     price = null;
 
@@ -139,7 +140,7 @@ define([
                     courseSeatType: this.formatSeatTypes(),
                     discountValue: this.discountValue(),
                     endDateTime: this.formatDateTime(this.model.get('end_date')),
-                    lastEdited: this.formatLastEditedData(this.model.get('last_edited')),
+                    lastEdited: lastEdited ? this.formatLastEditedData(lastEdited) : '',
                     price: price,
                     startDateTime: this.formatDateTime(this.model.get('start_date')),
                     usage: this.usageLimitation(),


saleem-latif:
Note: Above PR is deployed to https://ecommerce-business.sandbox.edx.org sandbox so bug will not appear there anymore. It can be reproduce on the local machine though by following the instructions.

ecommerce/static/js/views/coupon_detail_view.js
```diff
@@ -139,7 +140,7 @@ define([
                     courseSeatType: this.formatSeatTypes(),
                     discountValue: this.discountValue(),
                     endDateTime: this.formatDateTime(this.model.get('end_date')),
-                    lastEdited: this.formatLastEditedData(this.model.get('last_edited')),
+                    lastEdited: lastEdited ? this.formatLastEditedData(lastEdited) : 'Undefined',
```
douglashall: I think an empty string would be better than `Undefined` here.

saleem-latif: Done

douglashall (approved):
Reminder to squash before merge.

PR #1568, opened 2017-11-17
Don't set product_class of child products

McKenzieW:
Course Entitlement child products inherit their product_class from their parent product, so it does not need to and should not be explicitly set.


Mjloturco (commented):
Looks good to me.

McKenzieW:
@mjfrey any issues with me merging this one?

mjfrey (approved):

PR #1571, opened 2017-11-20
use edx segment fork

bderusha:


mjfrey (approved):

douglashall:
@bderusha @mjfrey Nice work here guys!

PR #1573, opened 2017-11-22
Remove simple history from models

ahsan-ul-haq:
LEARNER-3332


ecommerce/extensions/payment/migrations/0019_auto_20171122_1220.py
```diff
@@ -0,0 +1,26 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('payment', '0018_create_stripe_switch'),
+    ]
+
+    operations = [
```
clintonb, line 15: What's going on here? Why is this model being changed?

clintonb (changes requested):
There are seemingly unrelated migrations. Also, your tests are failing.

clintonb:
FYI @edx/learner @edx/devops

feanil:
Won't these removals try to remove tables that are really large?  Also, won't this break the deploy in multiple ways?  Is it possible to stop using the simplehistory module without generating these migrations.  I'd rather we do this cleanup by hand after the software no longer cares about these tables.  We should also make a note about this on the Hawthorn Page https://openedx.atlassian.net/wiki/spaces/COMM/pages/158045104/Hawthorn

PR #1578, opened 2017-11-29
Add redis to requirements-OPS-2490

nadeemshahzad:
@edx/devops Please review

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -43,4 +43,3 @@ sailthru-client==2.2.3
 six==1.10.0
 stripe==1.70.0
 zeep==2.1.1
-redis==2.10.6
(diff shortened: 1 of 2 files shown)


requirements/base.txt
```diff
@@ -43,3 +43,4 @@ sailthru-client==2.2.3
 six==1.10.0
 stripe==1.70.0
 zeep==2.1.1
+redis==2.10.6
```
feanil: @nadeemshahzad  I think this belongs in https://github.com/edx/ecommerce/blob/master/requirements/production.txt instead.

feanil (changes requested):

feanil (approved):

PR #1579, opened 2017-11-29
LEARNER-2842 Use bundle attribute rather than num_lines to check for bundle and add sku to DOM

MatthewPiatetsky:
JIRA Ticket - https://openedx.atlassian.net/browse/LEARNER-2842
this pr supports this [optimizely test](https://app.optimizely.com/v2/projects/1743970571/experiments/9623001520)
[Preview WIP Optimizely Test Behavior](https://ecommerce.edx.org/basket/add/?sku=F014F1A&optimizely_token=1250c451b083fe2ebf0a1c134163e2e437f3352d086c244f588f3919137db3e5&optimizely_x=9623361107&optimizely_x_audiences=8737273782,8936182252&optimizely_preview_layer_ids=9634150725&optimizely_snippet=s3-1743970571)

new donations product - https://ecommerce.edx.org/dashboard/catalogue/products/18572/

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -275,6 +275,7 @@ class BasketSummaryView(BasketView):
                 benefit_value = None
 
             line_data.update({
+                'sku': line.product.stockrecords.first().partner_sku,
                 'benefit_value': benefit_value,
                 'enrollment_code': line.product.is_enrollment_code_product,
                 'line': line,
@@ -438,7 +439,14 @@ class VoucherAddView(BaseVoucherAddView):  # pylint: disable=function-redefined
             return
 
         # Do not allow single course run coupons used on bundles.
-        if self.request.basket.num_lines > 1 and not voucher.offers.first().condition.program_uuid:
+        BUNDLE = 'bundle_identifier'
+        BasketAttribute = get_model('basket', 'BasketAttribute')
+        BasketAttributeType = get_model('basket', 'BasketAttributeType')
+        bundle_attribute = BasketAttribute.objects.filter(
+            basket=self.request.basket,
+            attribute_type=BasketAttributeType.objects.get(name=BUNDLE)
+        )
+        if len(bundle_attribute) > 0 and not voucher.offers.first().condition.program_uuid:
             messages.error(
                 self.request,
                 _("Coupon code '{code}' is not valid for this basket.").format(code=code))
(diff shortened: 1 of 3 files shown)


ecommerce/extensions/basket/views.py
```diff
@@ -275,6 +275,7 @@ def _process_basket_lines(self, lines):
                 benefit_value = None
 
             line_data.update({
+                'sku': line.product.stockrecords.first().partner_sku,
```
MatthewPiatetsky, line 278: not sure if this is the best way to do this

mjfrey: Why do this? You already have the line in the context

MatthewPiatetsky: i'm not sure if i can have queries into the querysets inside the template.
if i try to do 
`data-sku={{line_data.line.product.stockrecords.first().partner_sku}}`
I get an error that looks like

`Could not parse the remainder: 
'().partner_sku' from 'line_data.line.product.stockrecords.first().partner_sku'`

ecommerce/extensions/basket/views.py
```diff
@@ -438,7 +439,14 @@ def apply_voucher_to_basket(self, voucher):
             return
 
         # Do not allow single course run coupons used on bundles.
-        if self.request.basket.num_lines > 1 and not voucher.offers.first().condition.program_uuid:
+        BUNDLE = 'bundle_identifier'
+        BasketAttribute = get_model('basket', 'BasketAttribute')
+        BasketAttributeType = get_model('basket', 'BasketAttributeType')
+        is_bundle = BasketAttribute.objects.filter(
```
jaebradley: Nit: the naming implies a `boolean` value whereas it's really a list of `BasketAttribute` objects. This makes [line#449](https://github.com/edx/ecommerce/pull/1579/files#diff-398e80deb2588fea52e2551897aac3e2R449) a little hard to read when it's `len(is_bundle) > 0`

jaebradley (approved):
👍

ecommerce/extensions/basket/views.py
MatthewPiatetsky: will update variable name

mjfrey:
I really think this should be implemented in the View to add the donation product so we do not re-create the basket.  But up to you for the test.  
👍

mjfrey (approved):

PR #1580, opened 2017-12-01
Update segment fork version

bderusha:
Found an issue with our initial implementation using a fork of the segment analytics client.  Fixed on the fork.  Need to bump the version here


mjfrey (approved):
nice

bderusha:
FYI @mjfrey

PR #1582, opened 2017-12-03
Add support for refunding course entitlements.

mjfrey:
Add to the refund API the ability to refund orders that have course entitlements.

LEARNER-3414

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_refunds.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_refunds.py
@@ -40,8 +40,14 @@ class RefundCreateViewTests(RefundTestMixin, AccessTokenMixin, JwtMixin, TestCas
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(json.loads(response.content), [])
 
-    def _get_data(self, username=None, course_id=None):
+    def _get_data(self, username=None, course_id=None, order_number=None, entitlement_uuid=None):
         data = {}
+
+        if order_number:
+            data['order_number'] = order_number
+
+        if entitlement_uuid:
+            data['entitlement_uuid'] = entitlement_uuid
 
         if username:
             data['username'] = username
@@ -149,6 +155,48 @@ class RefundCreateViewTests(RefundTestMixin, AccessTokenMixin, JwtMixin, TestCas
         # A second call should result in no additional refunds being created
         response = self.client.post(self.path, data, JSON_CONTENT_TYPE)
         self.assert_ok_response(response)
+
+    def test_valid_entitlement_order(self):
+        """
+        View should create a refund if an entitlement order/line are found eligible for refund.
+        """
+
+        order = self.create_order(entitlement=True)
+        self.assertFalse(Refund.objects.exists())
+        data = self._get_data(username=self.user.username, order_number=order.number, entitlement_uuid='111')
+        response = self.client.post(self.path, data, JSON_CONTENT_TYPE)
+        refund = Refund.objects.latest()
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(json.loads(response.content), [refund.id])
+        self.assert_refund_matches_order(refund, order)
+
+        # A second call should result in no additional refunds being created
+        response = self.client.post(self.path, data, JSON_CONTENT_TYPE)
+        self.assert_bad_request_response(response, 'Order {} does not exist.'.format(order.number))
+
+    def test_invalid_entitlement_order(self):
+        """
+        View should not create a refund if an invalid order number is passed
+        """
+
+        self.create_order(entitlement=True)
+        data = self._get_data(username=self.user.username, order_number='123', entitlement_uuid='111')
+        response = self.client.post(self.path, data, JSON_CONTENT_TYPE)
+
+        self.assert_bad_request_response(response, 'Order 123 does not exist.')
+
+    def test_invalid_entitlement_order_line(self):
+        """
+        View should not create a refund if an entitlement order/line is invalid.
+        """
+
+        order = self.create_order(entitlement=True)
+        self.assertFalse(Refund.objects.exists())
+        data = self._get_data(username=self.user.username, order_number=order.number, entitlement_uuid='11')
+        response = self.client.post(self.path, data, JSON_CONTENT_TYPE)
+
+        self.assert_bad_request_response(response, 'Order {} does not exist.'.format(order.number))
 
     def test_refunded_line(self):
         """
--- a/ecommerce/extensions/api/v2/views/refunds.py
+++ b/ecommerce/extensions/api/v2/views/refunds.py
@@ -9,8 +9,14 @@ from rest_framework.response import Response
 from ecommerce.extensions.api import serializers
 from ecommerce.extensions.api.exceptions import BadRequestException
 from ecommerce.extensions.api.permissions import CanActForUser
-from ecommerce.extensions.refund.api import create_refunds, find_orders_associated_with_course
+from ecommerce.extensions.refund.api import (
+    create_refunds,
+    create_refunds_for_entitlement,
+    find_orders_associated_with_course
+)
 
+Order = get_model('order', 'Order')
+OrderLine = get_model('order', 'Line')
 Refund = get_model('refund', 'Refund')
 User = get_user_model()
 
@@ -18,12 +24,12 @@ User = get_user_model()
 class RefundCreateView(generics.CreateAPIView):
     """Creates refunds.
 
-    Given a username and course ID, this view finds and creates a refund for each order
-    matching the following criteria:
+    Given a username and course ID or an order number and a course entitlement,
+    this view finds and creates a refund for each order matching the following criteria:
 
         * Order was placed by the User linked to username.
         * Order is in the COMPLETE state.
-        * Order has at least one line item associated with the course ID.
+        * Order has at least one line item associated with the course ID or Course Entitlement.
 
     Note that only the line items associated with the course ID will be refunded.
     Items associated with a different course ID, or not associated with any course ID, will NOT be refunded.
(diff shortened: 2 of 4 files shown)


mjfrey:
@amangano-edx FYI

mjfrey:
@edx/helio Please take a look.

ecommerce/extensions/api/v2/views/refunds.py
```diff
@@ -41,27 +45,39 @@ def get_serializer(self):
...
 
         refunds = []
 
-        # We can only create refunds if the user has orders.
-        if user.orders.exists():
-            orders = find_orders_associated_with_course(user, course_id)
-            refunds = create_refunds(orders, course_id)
+        # Try and create a refund for the passed in order
+        if order_number:
+            order = user.orders.filter(number=order_number).first()
```
McKenzieW: since there should only be one order with the given number, could you not just use `get(number=order_number)` rather than `filter(number=order_number).first()`?

McKenzieW: which then I suppose you'd have to wrap in a `get..except` rather than using the `if..else` block below

ecommerce/extensions/refund/api.py
```diff
@@ -36,6 +36,31 @@ def find_orders_associated_with_course(user, course_id):
...
+    order (Order): The order for which to create the refund
+    entitlement_uuid (UUID): The entitlement in the order for which to refund
+
+    Returns:
+        list: refunds created
+    """
+
+    refunds = []
+
+    line = order.lines.filter(refund_lines__id__isnull=True,
```
McKenzieW: I believe the same logic should apply here as above, where there would only be one line in a given order with the given entitlement UUID, so `get` rather than `filter(...).first()` should be appropriate with a `try..except` rather than an if statement below

ecommerce/extensions/api/v2/views/refunds.py
```diff
@@ -41,27 +46,39 @@ def get_serializer(self):
...
-        if not username:  # pragma: no cover
-            raise BadRequestException('No username specified.')
-
-        try:
-            user = User.objects.get(username=username)
-        except User.DoesNotExist:
-            raise BadRequestException('User "{}" does not exist.'.format(username))
+        order_number = request.data.get('order_number')
+        entitlement_uuid = request.data.get('entitlement_uuid')
+        user = request.user
```
staubina: This will move the User from the username in the data to the user making the request, is this expected? It seems to be a change that could cause problems for the existing code.

staubina: Is it possible that the user making the request is say a Staff/Admin user while the refund is for a different username?

staubina: Looking later in the code I might be missing something, but I do not see the reason to set user here to the requesting user.

ecommerce/extensions/refund/api.py
```diff
@@ -36,6 +37,33 @@ def find_orders_associated_with_course(user, course_id):
...
+    refunds = []
+
+    try:
+        line = order.lines.get(refund_lines__id__isnull=True,
+                               product__attribute_values__attribute__code='UUID',
+                               product__attribute_values__value_text=entitlement_uuid)
+
+        refund = Refund.create_with_lines(order, [line])
+        if refund is not None:
+            refunds.append(refund)
```
staubina: Should we log something if there is an attempt to refund that is None here? Or is it handled with Exception logs?

ecommerce/extensions/refund/api.py
```diff
@@ -36,6 +37,33 @@ def find_orders_associated_with_course(user, course_id):
...
+    except OrderLine.DoesNotExist:
+        pass
```
staubina: Do we need to be concerned here?  Should we log before passing with exception?

staubina (changes requested):
Some open questions and comments.

ecommerce/extensions/api/v2/views/refunds.py
mjfrey: You are right.  This is left over.  I will remove and only use the username param.

ecommerce/extensions/refund/api.py
mjfrey: I should log something here.

ecommerce/extensions/api/v2/views/refunds.py
```diff
@@ -41,11 +47,13 @@ def get_serializer(self):
...
         username = request.data.get('username')
+        order_number = request.data.get('order_number')
+        entitlement_uuid = request.data.get('entitlement_uuid')
 
-        if not course_id:
-            raise BadRequestException('No course_id specified.')
+        refunds = []
 
         # We should always have a username value as long as CanActForUser is in place.
         if not username:  # pragma: no cover
```
amangano-edx, line 76: This seems to indicate that username is required for entitlement refunds. We should clarify that in the docstring.

ecommerce/extensions/api/v2/views/refunds.py
```diff
@@ -56,12 +64,21 @@ def create(self, request, *args, **kwargs):
...
             raise BadRequestException('User "{}" does not exist.'.format(username))
 
-        refunds = []
-
-        # We can only create refunds if the user has orders.
-        if user.orders.exists():
-            orders = find_orders_associated_with_course(user, course_id)
-            refunds = create_refunds(orders, course_id)
+        # Try and create a refund for the passed in order
+        if order_number:
```
amangano-edx: I think it would be better to check for `entitlement_uuid` as that's more specific to entitlements.

ecommerce/extensions/api/v2/views/refunds.py
```diff
@@ -56,12 +64,21 @@ def create(self, request, *args, **kwargs):
...
+            except (Order.DoesNotExist, OrderLine.DoesNotExist):
+                raise BadRequestException('Order {} does not exist.'.format(order_number))
+        else:
+            if not course_id:
+                raise BadRequestException('No course_id specified.')
+
+            # We can only create refunds if the user has orders.
+            if user.orders.exists():
+                orders = find_orders_associated_with_course(user, course_id)
+                refunds = create_refunds(orders, course_id)
```
amangano-edx, line 98: Could we move this code to two separate methods, `refund_entitlement` and `refund_courserun`?

ecommerce/extensions/api/v2/views/refunds.py
```diff
@@ -56,12 +64,21 @@ def create(self, request, *args, **kwargs):
...
-
-        # We can only create refunds if the user has orders.
-        if user.orders.exists():
-            orders = find_orders_associated_with_course(user, course_id)
-            refunds = create_refunds(orders, course_id)
+        # Try and create a refund for the passed in order
+        if order_number:
+            try:
+                order = user.orders.get(number=order_number)
+                refunds = create_refund_for_entitlement(order, entitlement_uuid)
```
amangano-edx: We should make sure both `order_number` and `entitlement_uuid` exist before making these calls.

ecommerce/extensions/api/v2/views/refunds.py
```diff
@@ -56,12 +64,21 @@ def create(self, request, *args, **kwargs):
...
+            except (Order.DoesNotExist, OrderLine.DoesNotExist):
+                raise BadRequestException('Order {} does not exist.'.format(order_number))
+        else:
+            if not course_id:
+                raise BadRequestException('No course_id specified.')
+
+            # We can only create refunds if the user has orders.
+            if user.orders.exists():
```
amangano-edx, line 96: Why do we need to do this for the course_run case but not for entitlements?

ecommerce/extensions/refund/api.py
```diff
@@ -36,6 +36,30 @@ def find_orders_associated_with_course(user, course_id):
     return list(orders)
 
 
+def create_refund_for_entitlement(order, entitlement_uuid):
```
amangano-edx: Nit: this returns a list, but the method name is singular. Giving this a plural name would make things clearer.

ecommerce/extensions/api/v2/tests/views/test_refunds.py
```diff
@@ -150,6 +156,48 @@ def test_valid_order(self):
...
         self.assert_ok_response(response)
 
+    def test_valid_entitlement_order(self):
+        """
+        View should create a refund if an entitlement order/line are found eligible for refund.
+        """
+
+        order = self.create_order(entitlement=True)
+        self.assertFalse(Refund.objects.exists())
+        data = self._get_data(username=self.user.username, order_number=order.number, entitlement_uuid='111')
```
amangano-edx, line 166: Nit: it would be good if we could avoid hardcoding the skus.

ecommerce/extensions/api/v2/tests/views/test_refunds.py
```diff
@@ -40,9 +40,15 @@ def assert_ok_response(self, response):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(json.loads(response.content), [])
 
-    def _get_data(self, username=None, course_id=None):
+    def _get_data(self, username=None, course_id=None, order_number=None, entitlement_uuid=None):
```
amangano-edx, line 43: What API endpoint does this hit?

ecommerce/extensions/api/v2/views/refunds.py
mjfrey: not really.  An order_number is the key here.

ecommerce/extensions/api/v2/views/refunds.py
mjfrey: Do not see the need.  They both create refunds.  No need to break out and create new methods.

ecommerce/extensions/api/v2/views/refunds.py
mjfrey: Since no order_number is sent this has been here as a check to make sure the user has orders before attempting.

ecommerce/extensions/api/v2/tests/views/test_refunds.py
mjfrey: no API it is a method

amangano-edx: ohh, oops. the diff made it look like it was making a post.

ecommerce/extensions/api/v2/tests/views/test_refunds.py
```diff
@@ -150,6 +156,48 @@ def test_valid_order(self):
...
+
+        self.create_order(entitlement=True)
+        data = self._get_data(username=self.user.username, order_number='123', entitlement_uuid='111')
+        response = self.client.post(self.path, data, JSON_CONTENT_TYPE)
+
+        self.assert_bad_request_response(response, 'Order 123 does not exist.')
+
+    def test_invalid_entitlement_order_line(self):
+        """
+        View should not create a refund if an entitlement order/lineis invalid.
```
amangano-edx: Nit: 'order/line is'

ecommerce/extensions/api/v2/views/refunds.py
amangano-edx: Methods that handle one isolated piece of functionality are easier to read/maintain than long, monothic methods that do many things and contain many if/else blocks.

staubina (approved):

PR #1583, opened 2017-12-04
Remove usage of simple history

ahsan-ul-haq:
LEARNER-3348


ecommerce/courses/models.py
```diff
@@ -39,7 +38,14 @@ class Course(models.Model):
         blank=True,
         help_text=_('Last date/time on which verification for this product can be submitted.')
     )
-    history = HistoricalRecords()
+    created = models.DateTimeField(
```
clintonb: Don't manually make these fields. Use `from django_extensions.db.models import TimeStampedModel`. Also, make sure these fields are populated with the data from the first and last historical record.

clintonb (changes requested):

ecommerce/courses/models.py
ahsan-ul-haq: I tried using that but it requires a default value as, I believe, it does not allow null=True.

clintonb: Drop the fields.

ahsan-ul-haq: So we are not going to need that as well?

clintonb: Sorry. Keep your customizations, and populate with the historical records.

ahsan-ul-haq:
@clintonb madee the change

ahsan-ul-haq:
@clintonb please take a look.

ecommerce/courses/migrations/0007_auto_20171205_0800.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
```
clintonb: Merge this with the migration above.

ecommerce/courses/migrations/0007_auto_20171205_0800.py
```diff
@@ -0,0 +1,50 @@
...
+# Generated by Django 1.10.7 on 2017-12-05 08:00
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from ecommerce.core.constants import ISO_8601_FORMAT
+from ecommerce.courses.models import Course
```
clintonb: This is not the proper way to access models in migrations. Use `apps.get_models()`. See https://docs.djangoproject.com/en/1.11/topics/migrations/#data-migrations.

ecommerce/courses/migrations/0007_auto_20171205_0800.py
```diff
@@ -0,0 +1,50 @@
...
+        for course in courses:
+            course.modified = None
+            course.save()
+
+    dependencies = [
+        ('courses', '0006_auto_20171204_1036'),
+    ]
+
+    operations = [
+        migrations.RunPython(add_created_date, remove_created_date),
```
clintonb: Merge these into a single method, and avoid iterating over the data twice.

ecommerce/courses/migrations/0007_auto_20171205_0800.py
```diff
@@ -0,0 +1,50 @@
...
+from ecommerce.courses.models import Course
+
+
+class Migration(migrations.Migration):
+
+    def add_created_date(apps, schema_editor):
+        courses = Course.objects.all()
+
+        for course in courses:
+            created_date = course.history.earliest().history_date.strftime(ISO_8601_FORMAT)
```
clintonb: Why are you converting the `datetime` object to a string? That is not necessary.

ecommerce/courses/migrations/0007_auto_20171205_0800.py
clintonb: Also, don't bother with a revert. If we revert the merged migration, the column will be deleted. Use [`RunPython.noop`](https://docs.djangoproject.com/en/1.11/ref/migration-operations/#django.db.migrations.operations.RunPython.noop).

ecommerce/courses/models.py
clintonb: Nit: Both of these declarations can be on one line.

ecommerce/courses/models.py
```diff
@@ -39,8 +39,16 @@ class Course(models.Model):
...
+        null=True,
+        auto_now_add=True
+    )
+    modified = models.DateTimeField(
+        null=True,
+        auto_now=True
+    )
     thumbnail_url = models.URLField(null=True, blank=True)
+    history = HistoricalRecords()
```
clintonb: Nit: put this back where it was.

ecommerce/extensions/api/serializers.py
```diff
@@ -298,10 +297,7 @@ def __init__(self, *args, **kwargs):
             self.fields.pop('products', None)
 
     def get_last_edited(self, obj):
-        try:
-            return obj.history.latest().history_date.strftime(ISO_8601_FORMAT)
-        except ObjectDoesNotExist:
-            return None
+        return obj.modified.strftime(ISO_8601_FORMAT)
```
clintonb: What happens if the value is `None`?

ecommerce/courses/models.py
```diff
@@ -39,8 +39,16 @@ class Course(models.Model):
         blank=True,
         help_text=_('Last date/time on which verification for this product can be submitted.')
     )
-    history = HistoricalRecords()
+    created = models.DateTimeField(
+        null=True,
+        auto_now_add=True
+    )
+    modified = models.DateTimeField(
+        null=True,
```
clintonb: Make these fields non-nullable in a follow-on PR.

clintonb (commented):
Fix the migration, and update the translations.

ahsan-ul-haq:
@clintonb made the suggested changes.

ecommerce/courses/migrations/0006_auto_20171204_1036.py
```diff
@@ -0,0 +1,46 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    def add_created_modified_date(apps, schema_editor):
+        Course = apps.get_model('courses', 'Course')
+        courses = Course.objects.all()
+
+        for course in courses:
+            created_date = course.history.earliest().history_date
```
waheedahmed: No need to create variables, just directly assign the value to model fields.

waheedahmed (approved):
Just one comment otherwise :+1:

ahsan-ul-haq:
@waheedahmed made the change. @clintonb please review

clintonb (approved):
Please ensure the [Hawthorn release notes](https://openedx.atlassian.net/wiki/spaces/COMM/pages/158045104/Hawthorn) are updated to inform maintainers that django-simple-history is being removed from this service.

clintonb:
FYI @edx/learner @edx/devops @edx/open-source-team

ahsan-ul-haq:
Closing this and creating a new PR https://github.com/edx/ecommerce/pull/1587

PR #1584, opened 2017-12-04
Fix fulfillment error for donations

MatthewPiatetsky:
Fulfillment of donation failed which contributed to it not being reported properly in the data warehouse.
https://ecommerce.edx.org/dashboard/orders/EDX-31196785/

For the purposes of the test at least, we do not need to do anything for the fulfillment step.
In theory there could be a fullfillment module with empty methods that contain "pass" but that seems overly complex for a noop.
If we wanted something more "extensible" for the test we could have a constant with a list of SKIP_FULFILLMENT_LINE_TYPES?

### DIFF
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -16,7 +16,10 @@ from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout  # pylint: disable=ungrouped-imports
 from rest_framework import status
 
-from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME
+from ecommerce.core.constants import (
+    DONATIONS_FROM_CHECKOUT_TESTS_PRODUCT_TYPE_NAME,
+    ENROLLMENT_CODE_PRODUCT_CLASS_NAME
+)
 from ecommerce.core.url_utils import get_lms_enrollment_api_url, get_lms_entitlement_api_url
 from ecommerce.courses.models import Course
 from ecommerce.courses.utils import mode_for_product
@@ -101,6 +104,58 @@ class BaseFulfillmentModule(object):  # pragma: no cover
             True, if the product is revoked; otherwise, False.
         """
         raise NotImplementedError("Revoke method not implemented!")
+
+
+class DonationsFromCheckoutTestFulfillmentModule(BaseFulfillmentModule):
+    """
+    Fulfillment module for fulfilling donations as a part of LEARNER-2842 - Test Donations on Checkout.
+    If that test, or any follow up tests around donations at checkout are not implemented, this module will be reverted.
+    Don't use this code for your own purposes, thanks.
+    """
+    def supports_line(self, line):
+        """
+        Returns True if the given Line has a donation product.
+        """
+        return line.product.get_product_class().name == DONATIONS_FROM_CHECKOUT_TESTS_PRODUCT_TYPE_NAME
+
+    def get_supported_lines(self, lines):
+        """ Return a list of supported lines (that contain a donation product)
+
+        Args:
+            lines (List of Lines): Order Lines, associated with purchased products in an Order.
+
+        Returns:
+            A supported list of lines, unmodified.
+        """
+        return [line for line in lines if self.supports_line(line)]
+
+    def fulfill_product(self, order, lines):
+        """ Fulfills the specified lines in the order.
+        Marks the line status as complete. Does not change anything else.
+
+        Args:
+            order (Order): The Order associated with the lines to be fulfilled
+            lines (List of Lines): Order Lines, associated with purchased products in an Order.
+
+        Returns:
+            The original set of lines, with new statuses set based on the success or failure of fulfillment.
+
+        """
+        for line in lines:
+            line.set_status(LINE.COMPLETE)
+        return order, lines
+
+    def revoke_line(self, line):
+        """ Revokes the specified line.
+        (Returning true to avoid unnecessary errors)
+
+        Args:
+            line (Line): Order Line to be revoked.
+
+        Returns:
+            True, if the product is revoked; otherwise, False.
+        """
+        return True
 
 
 class EnrollmentFulfillmentModule(BaseFulfillmentModule):
--- a/ecommerce/extensions/fulfillment/tests/test_modules.py
+++ b/ecommerce/extensions/fulfillment/tests/test_modules.py
@@ -16,6 +16,7 @@ from testfixtures import LogCapture
 from ecommerce.core.constants import (
     COUPON_PRODUCT_CLASS_NAME,
     COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME,
+    DONATIONS_FROM_CHECKOUT_TESTS_PRODUCT_TYPE_NAME,
     ENROLLMENT_CODE_PRODUCT_CLASS_NAME,
     ENROLLMENT_CODE_SWITCH,
     SEAT_PRODUCT_CLASS_NAME
@@ -30,6 +31,7 @@ from ecommerce.extensions.catalogue.tests.mixins import DiscoveryTestMixin
 from ecommerce.extensions.fulfillment.modules import (
     CouponFulfillmentModule,
     CourseEntitlementFulfillmentModule,
+    DonationsFromCheckoutTestFulfillmentModule,
     EnrollmentCodeFulfillmentModule,
     EnrollmentFulfillmentModule
 )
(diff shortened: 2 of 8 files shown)


mjfrey (approved):
Fine for the test. But will need a real fulfillment module to support donation products.

ecommerce/extensions/fulfillment/api.py
```diff
@@ -56,14 +56,17 @@ def fulfill_order(order, lines):
...
                 module.fulfill_product(order, supported_lines)
-
         # Check to see if any line items in the order have not been accounted for by a FulfillmentModule
         # Any product does not line up with a module, we have to mark a fulfillment error.
         for line in line_items:
             product_type = line.product.get_product_class().name
-            logger.error("Product Type [%s] does not have an associated Fulfillment Module. It cannot be fulfilled.",
-                         product_type)
-            line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
+            if product_type == 'Donation':
```
clintonb: Use a fulfillment module.

clintonb (changes requested):

clintonb:
This is _not_ fine for the test. You're taking a completely unnecessary shortcut.

ecommerce/extensions/fulfillment/api.py
MatthewPiatetsky: what is the purpose of using the fulfillment module if I am explicitly not fulfilling anything?
a fulfillment module that does not fulfill anything is not a fulfillment module?
i could see the argument for refactoring earlier code to specifically check for SKIP_FULFILLMENT_PRODUCT_TYPES

clintonb: The purpose of the fulfillment module is to fulfill the line. If fulfilling is simply setting the line status to fulfilled, do it. Make a module called `NoopFulfillmentModule`, if needed. However, don't dirty the codebase with hard-coded product classes like this.

MatthewPiatetsky: i'm not sure i understand why writing a module is cleaner, but let me know if the module i added is what you're looking for.
if so I will remove this line and add the relevant tests.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -103,6 +103,57 @@ def revoke_line(self, line):
...
+        """ Revokes the specified line.
+        (Not implemented since revoking a noop is meaningless)
+
+        Args:
+            line (Line): Order Line to be revoked.
+
+        Returns:
+            True, if the product is revoked; otherwise, False.
+        """
+        raise NotImplementedError("Revoke method not implemented!")
```
clintonb: Revocation may be a no-op, but that doesn't mean you can raise an exception. If someone wants a refund, this line will blow up. Just return `True`.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -103,6 +103,57 @@ def revoke_line(self, line):
...
+class DoNothingDonationFulfillmentModule(BaseFulfillmentModule):
+    """
+    Fulfillment module for fulfilling donations as a part of LEARNER-2842 - Test Donations on Checkout.
+    If that test, or a follow up test around donations at checkout is not implemented, this module will be reverted.
+    """
+    def supports_line(self, line):
+        """
+        Returns True if the given Line has a donation product.
+        """
+        return line.product.get_product_class().name == "Donation"
```
clintonb: This should be a constant in `ecommerce/core/constants.py`.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -103,6 +103,57 @@ def revoke_line(self, line):
         raise NotImplementedError("Revoke method not implemented!")
 
 
+class DoNothingDonationFulfillmentModule(BaseFulfillmentModule):
```
clintonb: Given that this only applies to donations, the proper name is `DonationFulfillmentModule`.

ecommerce/extensions/fulfillment/api.py
clintonb: We use the [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern) for fulfillment. This ensures the code necessary to fulfill each product class is self-contained, and we _don't_ end up with a series of if-else statements.

Running a test doesn't give teams/developers an exemption from maintaining best practices and writing clean code. This is especially true for the E-Commerce Service, where we should be extra diligent about what we release.

ecommerce/extensions/test/factories.py
```diff
@@ -24,27 +25,31 @@
...
     if site is None:
         site = SiteConfigurationFactory().site
     if owner is None:
         owner = UserFactory()
     basket = Basket.objects.create(site=site, owner=owner)
     basket.strategy = Default()
     if not empty:
-        product = create_product()
+        if product_class:
+            product_class_instance = newfactories.ProductClassFactory(name=product_class)
```
clintonb: The product class should be created via migration, as we do with all product classes.

clintonb (changes requested):

ecommerce/extensions/fulfillment/modules.py
MatthewPiatetsky: why do CouponFulfillmentModule/EnrollmentCodeFulfillmentModule raise a not implemented error?

ecommerce/extensions/fulfillment/modules.py
MatthewPiatetsky: why should this be a constant if this module is specific only to donations for this test?

MatthewPiatetsky:
@clintonb I'm confused by your suggestions to name the module DonationFulfilllmentModule and to create the Donation product class with a migration.
That naming and the existence of a migration implies these things are supported for other uses. When our team goes to revert the test changes and delete/squash the migration in 2-3 weeks, if someone has gone and used that code not understanding it would be reverted in a matter of weeks, suddenly we will be breaking new code when cleaning up the test.
By making these changes i'd be signing up our team for a potentially very difficult clean up process.
At the very least if it was done this way, all the modules/constants/migrations would need to contain the ticket number in the name with a DO NOT USE comment

ecommerce/extensions/fulfillment/modules.py
clintonb: Those products are not refundable.

clintonb:
In 2-3 weeks no one will have used the code because few in the OpenEdX community follow the master branch, and we aren't releasing an open source release in that time. I also highly doubt any other internal teams are working on donations.

Again, just because you're running a test doesn't mean you are exempted of following the style/standards of the codebase. Every product class is represented as a migration because that is how we ensure product classes are in sync in every environment. If you choose to remove the product class after testing, delete the body of the migration creating the product class, and add a new migration that deletes the product class if it exists (which should be the same as the revert method of the initial migration).

Fulfillment modules are generally named `<ProductClassName>FulfillmentModule`. The addition of `Noop` to the name doesn't really do anything. Also, if you end up keeping the code, you'll have to rename it.

You can, and should, have `FIXME` comments linked to tickets, and explaining the unstable nature of the code. However, you should still comply with the standards of the codebase. We've established patterns and behaviors based on the last couple years' of success _and_ failures. This new product class does not require us to deviate from those learned patterns and behaviors.

MatthewPiatetsky:
There are multiple internal teams that I am aware of involved with donations at this very moment. I will include the ticket number in all the naming which should make it clear that this is specific to this test to potential consumers.

ecommerce/extensions/fulfillment/modules.py
MatthewPiatetsky: Right, product did not say these should be refundable but I will ask them to change requirements

MatthewPiatetsky:
@clintonb have i addressed your comments

ecommerce/extensions/fulfillment/api.py
```diff
@@ -56,12 +56,12 @@ def fulfill_order(order, lines):
             if supported_lines:
                 line_items = list(set(line_items) - set(supported_lines))
                 module.fulfill_product(order, supported_lines)
-
```
clintonb: No changes are needed for this file.

ecommerce/extensions/fulfillment/tests/test_modules.py
```diff
@@ -476,6 +478,48 @@ def test_revoke_line(self):
             CouponFulfillmentModule().revoke_line(line)
 
 
+class DonationsFromCheckoutTestFulfillmentModuleTest(FulfillmentTestMixin, TestCase):
+    """ Test donation fulfillment. """
+
+    def setUp(self):
+        super(DonationsFromCheckoutTestFulfillmentModuleTest, self).setUp()
+        donation_class, _ = ProductClass.objects.get_or_create(
```
clintonb: Use a double underscore for unused variables. A single underscore is reserved for ugettext.

clintonb: Why is this product class not being created via migration?

clintonb (changes requested):

ecommerce/extensions/fulfillment/api.py
MatthewPiatetsky: removed

ecommerce/extensions/fulfillment/tests/test_modules.py
MatthewPiatetsky: it was nice to not have to run migrations for the test locally but i've removed it

MatthewPiatetsky:
@clintonb have i addressed your comments

ecommerce/extensions/order/tests/0028_donations_from_checkout_tests_product_type.py
```diff
@@ -0,0 +1,61 @@
...
+        track_stock=False,
+        requires_shipping=False,
+        name=DONATIONS_FROM_CHECKOUT_TESTS_PRODUCT_TYPE_NAME,
+        slug=slugify(DONATIONS_FROM_CHECKOUT_TESTS_PRODUCT_TYPE_NAME)
+    )
+
+    # Create product attributes for donation products
+    ProductAttribute.objects.get_or_create(
+        product_class=donation,
+        name="certificate_type",
```
clintonb: Why does a donation have a certificate type?

ecommerce/extensions/order/tests/0028_donations_from_checkout_tests_product_type.py
```diff
@@ -0,0 +1,61 @@
...
+        name="Donations"
+    )
+
+
+def remove_product_class(apps, schema_editor):  # pylint: disable=unused-argument
+    """ Reverse function. """
+    donation_class = ProductClass.objects.get(name=DONATIONS_FROM_CHECKOUT_TESTS_PRODUCT_TYPE_NAME)
+    Product.objects.filter(product_class=donation_class).delete()
+    Category.objects.filter(slug='donations').delete()
+    ProductClass.objects.filter(name=DONATIONS_FROM_CHECKOUT_TESTS_PRODUCT_TYPE_NAME).delete()
```
clintonb: `donation_class.delete()`

ecommerce/extensions/test/factories.py
clintonb: Migrations are run for Travis, thus the product class is created by the migration. This will result in a DB error on the name field, which should be unique.

MatthewPiatetsky: i can change this but i don't understand your comment?
the test using this passed multiple times both locally and on travis

clintonb: Did you have the migration in the previous commits?

MatthewPiatetsky:
@clintonb have i addressed your comments

clintonb (approved):

clintonb:
Remember to squash your commits, and include a ticket number in the commit message.

PR #1588, opened 2017-12-07
Fix migration issue

ahsan-ul-haq:


mjfrey (approved):

clintonb:
Why is this necessary? There are no details in the commit message.

ahsan-ul-haq:
Apparently running migrations without this causes `'Course' object has no attribute 'history'` error

clintonb (approved):
Update the commit message to follow our best practices at https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/161880915/Git+Best+Practices#GitBestPractices-CommitMessages.

PR #1589, opened 2017-12-07
Removed course migration

ahsan-ul-haq:
The  migration is causing stage deployment failure so have to remove it


clintonb (approved):

feanil:
Was this fixed a different way?  Because it's still breaking the Sandbox CI Environment: https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/7255/console

PR #1592, opened 2017-12-08
Added course migration

ahsan-ul-haq:
Added the modified and created date in course model. That was removed as it was causing some issues in the stage which have been dealt with.


ecommerce/extensions/api/v2/tests/views/test_courses.py
```diff
@@ -44,7 +44,7 @@ def serialize_course(self, course, include_products=False):
 
         last_edited = None
         try:
-            last_edited = course.history.latest().history_date.strftime(ISO_8601_FORMAT)
+            last_edited = course.modified.strftime(ISO_8601_FORMAT)
         except ObjectDoesNotExist:
```
waheedahmed: Remove this try except statements.

waheedahmed (approved):
Just one comment, otherwise looks good

ahsan-ul-haq:
@waheedahmed  made the suggested change. @clintonb please take a look.

ecommerce/core/management/commands/clean_history.py
```diff
@@ -0,0 +1,96 @@
+from __future__ import unicode_literals
```
clintonb, line 1: Why is this file here?

ecommerce/extensions/catalogue/models.py
```diff
@@ -80,6 +82,10 @@ def update_enrollment_code(sender, **kwargs):  # pylint: disable=unused-argument
         instance.original_expires = instance.expires
 
 
+class ProductAttributeValue(AbstractProductAttributeValue):
```
clintonb, line 85: Why are these being added back?

clintonb (changes requested):
Why are you adding so much code back to the repo? You only need to modify the `Course` model.

ecommerce/core/management/commands/clean_history.py
ahsan-ul-haq: because as per @feanil suggested history tables need to be emptied before we remove them otherwise, when we drop the table, it will take down the database because it will use up all of our disk IO.

ecommerce/extensions/catalogue/models.py
ahsan-ul-haq: because [here](https://github.com/edx/ecommerce/pull/1592/files#diff-612c9c2724929804eb68ea7d3f14a15bR62) we want to remove ProductAttributeValue  historical data as well.

ahsan-ul-haq:
@clintonb please review.

clintonb (approved):
Squash your commits and update your commit message to follow our guidelines: https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/161880915/Git+Best+Practices#GitBestPractices-CommitMessages.

clintonb:
@ahsan-ul-haq why do you keep opening/closing this PR?

ahsan-ul-haq:
Apparently the codecov is not working on this last time mike suggest closing and then opening it trigger that so I was trying that.

clintonb:
Given that your branch is out of date, rebase it and push. Alternatively, work with IT to ensure the webooks still work.

ahsan-ul-haq:
Closing this as this is being done in https://github.com/edx/ecommerce/pull/1612

PR #1595, opened 2017-12-12
Update the package-lock.json file from node 8.

feanil:
I built a successful container by removing the old lockfile.  This is the new one it generated.


arizzitano (approved):
🙌🙌🙌🙌

PR #1597, opened 2017-12-14
LEARNER-3327 Disable order placement for expired products

Ayub-Khan:
***[LEARNER-3327](https://openedx.atlassian.net/browse/LEARNER-3327)***

***Description:***
There is no existing product expiration check on the basket page. which enable students place orders for expired products using the sku. 

TODO: Will create another pull request for multiple item basket once this is done. 

Sandbox:
In progress.


mjfrey:
We already do this check in the middleware.  There is a strategy that is checked for each product checks the expiration.  

see:
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/partner/strategy.py#L21

Ayub-Khan:
@mjfrey thanks for clearing. I was able to place order because of staff account and assumed it allows all users did not know about this.

PR #1599, opened 2017-12-14
Point to program dashboard in receipt

mikix:
When a user purchases a program (rather than a single course run), we want the receipt to point them at the dashboard of the program, not the learner's normal course dashboard.

https://openedx.atlassian.net/browse/LEARNER-3525


Mjloturco (approved):
Nice. LGTM

mjfrey (approved):
nice

mikix:
@feanil codecov seems stuck -- this is fixable on our end yeah?

PR #1601, opened 2017-12-15
Updated source translation files

clintonb:


clintonb:
FYI @edx/learner please remember to run `make fake_translations`, and commit the source translations, when making changes to user-facing strings.

PR #1602, opened 2017-12-15
Added Alipay support via Stripe

clintonb:


clintonb:
FYI @edx/learner, in case you ever need Alipay.

clintonb:
FYI @nedbat @mjfrey @bderusha @mikedikan @douglashall

PR #1604, opened 2017-12-18
Delete codecov.yml

feanil:


feanil:
@mjfrey  pre our conversations last week, I think something about the codecov config override that is in ecommerce is at the root of the issue with CodeCov, removing it seems to fix it.

FYI @mikix

mikix:
Works for me

mikix (approved):

mjfrey (approved):
okay -- let's try this.

PR #1605, opened 2017-12-18
Check for duplicate course entitlement purchase

mjfrey:
Only allow a learner to purchase a second course entitlement they have already purchased if it has been refunded or has expired.

LEARNER-3572

### DIFF
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -89,7 +89,8 @@ def prepare_basket(request, products, voucher=None):
     is_multi_product_basket = True if len(products) > 1 else False
     for product in products:
         if product.is_enrollment_code_product or \
-                not UserAlreadyPlacedOrder.user_already_placed_order(request.user, product):
+                not UserAlreadyPlacedOrder.user_already_placed_order(user=request.user,
+                                                                     product=product, site=request.site):
             basket.add_product(product, 1)
             # Call signal handler to notify listeners that something has been added to the basket
             basket_addition.send(sender=basket_addition, product=product, user=request.user, request=request,
--- a/ecommerce/extensions/order/tests/test_utils.py
+++ b/ecommerce/extensions/order/tests/test_utils.py
@@ -1,16 +1,21 @@
 """Test Order Utility classes """
+import json
 import logging
 
 import ddt
+import httpretty
 import mock
 from django.test.client import RequestFactory
 from oscar.core.loading import get_class, get_model
 from oscar.test.newfactories import BasketFactory
+from requests import Timeout
 from testfixtures import LogCapture
 
+from ecommerce.core.url_utils import get_lms_entitlement_api_url
 from ecommerce.extensions.fulfillment.status import ORDER
 from ecommerce.extensions.order.utils import UserAlreadyPlacedOrder
 from ecommerce.extensions.refund.tests.factories import RefundFactory
+from ecommerce.extensions.refund.tests.mixins import RefundTestMixin
 from ecommerce.extensions.test.factories import create_basket, create_order
 from ecommerce.referrals.models import Referral
 from ecommerce.tests.factories import PartnerFactory, SiteConfigurationFactory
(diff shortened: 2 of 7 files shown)


ecommerce/extensions/basket/utils.py
```diff
@@ -89,7 +89,8 @@ def prepare_basket(request, products, voucher=None):
     is_multi_product_basket = True if len(products) > 1 else False
     for product in products:
         if product.is_enrollment_code_product or \
-                not UserAlreadyPlacedOrder.user_already_placed_order(request.user, product):
+                not UserAlreadyPlacedOrder.user_already_placed_order(user=request.user,
+                                                                     product=product, site=request.site):
```
staubina, line 93: Small nit here, the parameters here do not match the Python code style guidelines.

Ref: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#breaking-long-lines

staubina: There are a few of these in the PR, they might not pass quality.

ecommerce/extensions/order/tests/test_utils.py
```diff
@@ -201,14 +211,42 @@ def test_already_have_not_refunded_order(self):
...
         Test the case that user have a non refunded order for the product.
         """
-        self.assertTrue(UserAlreadyPlacedOrder.user_already_placed_order(user=self.user, product=self.product))
+        self.assertTrue(UserAlreadyPlacedOrder.user_already_placed_order(user=self.user, product=self.product,
+                                                                         site=self.site))
+
+    @httpretty.activate
+    def test_already_have_not_refunded_entitlement_order(self):
+        """
+        Test the caste that user has a non refunded order for the course entitlement
```
staubina: comment typo

staubina (commented):
Some minor comments I will take another look in the morning.

ecommerce/extensions/order/tests/test_utils.py
```diff
@@ -201,14 +212,56 @@ def test_already_have_not_refunded_order(self):
...
+        }
+        httpretty.register_uri(httpretty.GET, get_lms_entitlement_api_url() +
+                               'entitlements/' + self.course_entitlement_uuid + '/',
+                               status=200, body=json.dumps(body), content_type='application/json')
+        self.assertTrue(UserAlreadyPlacedOrder.user_already_placed_order(user=self.user,
+                                                                         product=self.course_entitlement,
+                                                                         site=self.site))
+
+    @httpretty.activate
+    def test_refunded_entitlement_order_connection_timeout(self):
```
jlajoie, line 242: Just double checking how this works, you're testing the error handling by forcing an "error" via the "Timeout" side_effect?

ecommerce/extensions/order/utils.py
```diff
@@ -126,7 +137,33 @@ class UserAlreadyPlacedOrder(object):
...
+
+        Returns:
+            bool: True if the entitlement is expired
+
+        """
+        entitlement_api_client = EdxRestApiClient(get_lms_entitlement_api_url(),
+                                                  jwt=site.siteconfiguration.access_token)
+        partner_short_code = site.siteconfiguration.partner.short_code
+        key = 'course_entitlement_detail_{}{}'.format(entitlement_uuid, partner_short_code)
+        cache_key = hashlib.md5(key).hexdigest()
```
jlajoie: Might be easier to talk about this in person, but two questions:

Is there a reason to use this md5 hash of a string, rather than just using the `entitlement_uuid` which we know will be unique?

Also, are we okay with the tradeoff of potentially missing setting the `expired_at_datetime` in the time range of whatever this key's TTL is, by hitting the cache and bypassing the REST endpoint?

ecommerce/extensions/order/tests/test_utils.py
mjfrey: Yes exactly.  Forcing the API call to LMS to throw a Timeout error.  Cover that use case.

ecommerce/extensions/order/utils.py
mjfrey: Good point we could just use the uuid.  I can change that.  Yes I am okay that we potentially could have a race condition.  However the TTL is only a minute so would probably be a very small chance.

ecommerce/extensions/basket/utils.py
mjfrey: This is the style of ecommerce.  Just using our pylint settings.

staubina (approved):

ecommerce/extensions/basket/utils.py
staubina: We should consider updating these to match the settings in Platform.

jlajoie (approved):
Lgtm :+1:

PR #1611, opened 2017-12-19
Add support for refund emils for course entitlements.

mjfrey:
Send refund emails for course entitlement products as well as course seats.

### DIFF
--- a/ecommerce/extensions/refund/models.py
+++ b/ecommerce/extensions/refund/models.py
@@ -11,7 +11,7 @@ from oscar.apps.payment.exceptions import PaymentError
 from oscar.core.loading import get_class, get_model
 from oscar.core.utils import get_default_currency
 
-from ecommerce.core.constants import SEAT_PRODUCT_CLASS_NAME
+from ecommerce.core.constants import COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME, SEAT_PRODUCT_CLASS_NAME
 from ecommerce.extensions.analytics.utils import audit_log
 from ecommerce.extensions.checkout.utils import format_currency, get_receipt_page_url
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/refund/models.py
```diff
@@ -206,15 +206,15 @@ def _notify_purchaser(self):
...
+        if product_class != (SEAT_PRODUCT_CLASS_NAME or COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME):
             logger.warning(
                 ('No refund notification will be sent for Refund [%d]. The notification supports product lines '
                  'of type Course, not [%s].'),
                 self.id, product_class
             )
             return
 
-        course_name = self.lines.first().order_line.product.course.name
+        course_name = self.lines.first().order_line.product.title
```
jaebradley: Super nit pick: does the `course_name` variable name need to be changed? not a blocker if this really needs to go out

ecommerce/extensions/refund/models.py
```diff
@@ -206,15 +206,15 @@ def _notify_purchaser(self):
         product = self.lines.first().order_line.product
         product_class = product.get_product_class().name
 
-        if product_class != SEAT_PRODUCT_CLASS_NAME:
+        if product_class != (SEAT_PRODUCT_CLASS_NAME or COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME):
```
jaebradley: Does this need to get broken up into ~~two separate `or`s~~ whoops I meant an `and`?

jaebradley (approved):

ecommerce/extensions/refund/models.py
jaebradley: @mjfrey and I talked about this offline

PR #1612, opened 2017-12-21
Added course migration and Clean history

ahsan-ul-haq:
Added course migration and clean history command
so we can remove history data before we can remove tables

LEARNER-3526


ahsan-ul-haq:
This has already approved in https://github.com/edx/ecommerce/pull/1592

PR #1618, opened 2017-12-26
remove django history from course

ahsan-ul-haq:
LEARNER-3348


ecommerce/courses/management/commands/sync_history_with_course.py
```diff
@@ -0,0 +1,33 @@
...
+                course.save()
+            except ObjectDoesNotExist:
+                logger.warning(
+                    'History object for course with course_id: %s does not exist',
+                    course.id
+                )
+                course.created = None
+                course.modified = None
+                course.save()
+                continue
```
waheedahmed: No need for this `continue`, remove it.

ecommerce/courses/management/commands/sync_history_with_course.py
```diff
@@ -0,0 +1,33 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Sync course history data with course created modified date.'
+
+    def handle(self, *args, **options):
+        courses = Course.objects.all()
+
+        for course in courses:
```
waheedahmed: Do we need to do this in atomic block?

waheedahmed (approved):
Couple of comments otherwise looks good.

ahsan-ul-haq:
@mjfrey please review.

mjfrey:
What is the impact of running these migrations in production?  Will this take a long time?  We need to prepare devops if so and maybe break this up into 2.

ahsan-ul-haq:
It will not take long time. As I don't think number of courses are more 3500.

mjfrey (approved):
Please squash commits.

PR #1619, opened 2017-12-28
WL-1325 | Management command for creating Sites, SiteThemes, SiteConf…

hasnain-naveed:
…igurations and Partners.


bill-filler:
this branch is no longer needed correct?

clintonb:
There are already management commands that do this:
* https://github.com/edx/ecommerce/blob/master/ecommerce/core/management/commands/create_or_update_site.py
* https://github.com/edx/ecommerce/blob/master/ecommerce/theming/management/commands/create_or_update_site_theme.py

PR #1621, opened 2018-01-03
Update internationalization documentation

BenjiLee:
https://openedx.atlassian.net/browse/LEARNER-2986

DOCUMENT in read-the-docs:
- the translation process (make translations, etc) **Taken from https://github.com/edx/credentials/blob/master/docs/internationalization.rst**
- how ecommerce resolves which language to show
- PayPal language

@edx/learner-heart Please review

### DIFF
--- a/docs/internationalization.rst
+++ b/docs/internationalization.rst
@@ -1,0 +1,71 @@
+Internationalization
+====================
+Follow the `internationalization coding guidelines`_ in the edX Developer's Guide when developing new features.
+
+Languages are enabled in
+the settings file, for example in ``ecommerce/settings/base.py``
+
+.. code-block:: python
+
+  LANGUAGES = (
+    ('en', _('English')),
+    ('es', _('Spanish')),
+    ('es-419', _('Spanish (Latin American)')),
+  )
+
+Ecommerce Language Negotiation
+~~~~~~~~~~~~~~~~~~~~~
+Language negotiation for ecommerce is handled by `Django's Locale Middleware`_. Django's Language Negotiation rules, in
+priority order, are as follows.
+
+#. Language prefix is not enabled by default for the ecommerce application.
+#. LANGUAGE_SESSION_KEY is not used.
+#. LANGUAGE_COOKIE_NAME is used to negotiate language. A user's language is set in their account settings in
+edx-platform which sets the language cookie. The language cookie name should be the same cookie name set by
+edx-platform. This can be configured in ``ecommerce/settings/base.py``.
+
+.. code-block:: python
+
+    LANGUAGE_COOKIE_NAME = 'openedx-language-preference'
+#. The Accept-Language HTTP header is used.
+#. LANGUAGE_CODE is used as the last resort. This can be set in 'ecommerce/settings/base.py'
+
+
+PayPal Language Negotiation
+~~~~~~~~~~~~~~~~~~~~~
+To enable localization for PayPal, follow these steps.
+
+#. Sign in to the LMS Django administration console for your base URL. For
+   example, ``http://{your_URL}/admin``.
+
+#. On the **Site Administration** page, locate **Waffle**.
+
+#. In the **Waffle** section, next to **Switches**, select **Add**.
+
+#. On the **Add switch** page, locate the **Name** field, and then add ``create_and_set_webprofile``
+
+#. On the **Add switch** page, locate the **Active** checkbox and check it.
+
+#. Select **Save**.
+
+A language code for ecommerce will be taken from a cookie as described in Ecommerce Language Negotiation. When the
+language code is fetched from the cookie, only the base language is used. For example, ``es-419`` resolve to ``es``.
+PayPal requires a country code. To get the country code, we use the language code to map it to a country. For example,
+the language code ``es`` will map to the country code ``MX`` when it is sent to PayPal. To add your language for PayPal,
+look up `PayPal's country to language mapping`_ and add it to PAYPAL_LOCALES in ``ecommerce/extensions/payment/constants.py``.
+
+.. code-block:: python
+
+    PAYPAL_LOCALES = {
+        'zh': 'CN',
+        'fr': 'FR',
+        'en': 'US',
+        'es': 'MX',
+    }
+
+If a language fetched from the cookie cannot be found in PAYPAL_LOCALES, the LANGUAGE_CODE is used. If the LANGUAGE_CODE
+ does not exist in PAYPAL_LOCALES, PayPal will use its own language negotiation.
+
+.. _internationalization coding guidelines: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/i18n.html
+.. _Django's Locale Middleware: https://docs.djangoproject.com/en/2.0/topics/i18n/translation/#how-django-discovers-language-preference
+.. _PayPal's country to language mapping: https://developer.paypal.com/docs/classic/api/locale_codes/
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -21,6 +21,7 @@ use the E-Commerce service with the Open edX platform.
    test_features
    test_ecommerce
    additional_features/index
+   internationalization
 
 To complete the procedures that this section describes, you use both the Django
 administration site and the Course Administration Tool (CAT). The CAT is a web
--- a/ecommerce/extensions/payment/constants.py
+++ b/ecommerce/extensions/payment/constants.py
@@ -38,8 +38,7 @@ CYBERSOURCE_CARD_TYPE_MAP = {
 
 CLIENT_SIDE_CHECKOUT_FLAG_NAME = 'enable_client_side_checkout'
 
-# Paypal only supports 4 languages, which are prioritized by country.
-# https://developer.paypal.com/docs/classic/api/locale_codes/
+# Paypal uses Country Codes to resolve language. https://developer.paypal.com/docs/classic/api/locale_codes/
 PAYPAL_LOCALES = {
     'zh': 'CN',
     'fr': 'FR',


docs/internationalization.rst
```diff
@@ -0,0 +1,95 @@
+Internationalization
+====================
+All user-facing text content should be marked for translation. Even if this application is only run in English, our open
+source users may choose to use another language. Marking content for translation ensures our users have this choice.
```
mikix: This section reads oddly.  "Even if this application is only run in English" -- "this application" means ecommerce at edx.org I guess?  And by open source users you mean server admins running an instance of openedx rather than users of the site?  It all makes it sound like you're talking to other edX employees rather than the community.

Maybe something like "An installation of ecommerce can choose to support any number of non-English languages" or something nicer.

docs/internationalization.rst
```diff
@@ -0,0 +1,95 @@
...
+    ('es', _('Spanish')),
+    ('es-419', _('Spanish (Latin American)')),
+  )
+
+A user's language is set in their account settings in edx-platform which sets the language cookie. The language cookie
+name should be the same cookie name set by edx-platform. This can be configured in ecommerce/settings/base.py.
+
+.. code-block:: bash
+
+  LANGUAGE_COOKIE_NAME = 'ecommerce_language'
```
mikix: That seems like a misleading "suggested" name.  Since right above you say this name should be shared between platform and ecommerce.  Maybe...  'openedx-language' (dashes or underscores?)

docs/internationalization.rst
```diff
@@ -0,0 +1,95 @@
...
+.. code-block:: bash
+
+    PAYPAL_LOCALES = {
+        'en': 'US',  # default
+        'zh': 'CN',
+        'fr': 'FR',
+        'es': 'MX',
+    }
+
+.. _Django's Locale Middleware: https://docs.djangoproject.com/en/2.0/topics/i18n/translation/#how-django-discovers-language-preference
```
mikix: This link feels like it's in the wrong section (original link was in previous one).

docs/internationalization.rst
```diff
@@ -0,0 +1,95 @@
...
+name should be the same cookie name set by edx-platform. This can be configured in ecommerce/settings/base.py.
+
+.. code-block:: bash
+
+  LANGUAGE_COOKIE_NAME = 'ecommerce_language'
+
+Paypal Language Negotiation
+~~~~~~~~~~~~~~~~~~~~~
+
+To enable localization for Paypal, enable the waffle switch "create_and_set_webprofile" in the admin. Waffle switches
```
mikix: Hah.  Reading this for the first time, without any context, I have no idea why that switch would control this. Not sure it's worth having a parenthetical here or not, though.

docs/internationalization.rst
```diff
@@ -0,0 +1,95 @@
...
+can be enabled at http://YOUR_HOST_NAME/admin/waffle/switch/
+
+Paypal only supports 4 languages via their SOAP API. This language is resolved from a country code.
+`More documentation can be found here`_.
+
+The available languages are English, Spanish, Chinese, and French. The language code for ecommerce will be taken from a
+cookie When the language code is fetched from the cookie (as described in Ecommerce Language Negotiation), only the base
```
mikix: Missing a period after cookie.

docs/internationalization.rst
```diff
@@ -0,0 +1,95 @@
...
+language will be used. For example, es-419 will resolve to es. If no compatible language is found, the first language
+set in PAYPAL_LOCALES in ecommerce/extensions/payment/constants.py will be used.
```
mikix: Is that for realz? We didn't have en first before this PR. Were we defaulting to Chinese?

docs/internationalization.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. code-block:: bash
+
+  LANGUAGE_COOKIE_NAME = 'ecommerce_language'
+
+Paypal Language Negotiation
+~~~~~~~~~~~~~~~~~~~~~
+
+To enable localization for Paypal, enable the waffle switch "create_and_set_webprofile" in the admin. Waffle switches
+can be enabled at http://YOUR_HOST_NAME/admin/waffle/switch/
```
mikix: Is there a way to mark that as not-a-link so that it doesn't get rendered as clickable?

docs/internationalization.rst
```diff
@@ -0,0 +1,95 @@
...
+source users may choose to use another language. Marking content for translation ensures our users have this choice.
+
+Follow the `internationalization coding guidelines`_ in the edX Developer's Guide when developing new features.
+
+.. _internationalization coding guidelines: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/i18n.html
+
+Ecommerce Language Negotiation
+~~~~~~~~~~~~~~~~~~~~~
+
+Language negotiation for the ecommerce site will be handled by `Django's Locale Middleware`_. Languages are enabled in
```
mikix: This left me wanting somehow.  I can read that doc, but like...  which parts of the normal django negotiation do we actually rely on? (do we take advantage of i18n_patterns?, do we set LANGUAGE_SESSION_KEY ourselves?, or do we just jump to the LANGUAGE_COOKIE_NAME part of it?)

mikix (changes requested):

BenjiLee:
@catong

docs/internationalization.rst
BenjiLee: Moved.

docs/internationalization.rst
BenjiLee: Updating code in https://github.com/edx/ecommerce/pull/1630

docs/internationalization.rst
BenjiLee: Perioded.

docs/internationalization.rst
BenjiLee: I heard it's impossible.

docs/internationalization.rst
BenjiLee: Added.

docs/internationalization.rst
BenjiLee: Changed to openedx's actual cookie name.

docs/internationalization.rst
```diff
@@ -0,0 +1,88 @@
...
+~~~~~~~~~~~~~~~~~~~~~
+This project uses `Transifex`_ to translate content. After new features are developed the translation source files
+should be pushed to Transifex. Our translation community will translate the content, after which we can retrieve the
+translations.
+
+.. _Transifex: https://www.transifex.com/
+
+Pushing source translation files to Transifex requires access to the edx-platform. Request access from the Open Source
+Team if you will be pushing translation files. You should also `configure the Transifex client`_ if you have not done so
+already.
```
amangano-edx: Is this relevant for the open source community? They could choose to use a different provider if they wanted right?

BenjiLee:
@mikix Ready for another pass

docs/internationalization.rst
```diff
@@ -0,0 +1,65 @@
...
+.. _internationalization coding guidelines: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/i18n.html
+
+Ecommerce Language Negotiation
+~~~~~~~~~~~~~~~~~~~~~
+Language negotiation for the ecommerce site will be handled by `Django's Locale Middleware`_. Below is the list of
+Django's Language Negotiation rules in priority order.
+
+1. Language prefix is not enabled by default for the ecommerce application.
+2. LANGUAGE_SESSION_KEY is not used.
+3. LANGUAGE_COOKIE_NAME is used to negotiate language. A user's language is set in their account settings in
```
mikix: The indentation of point 3 is wrong...  Not sure why looking at this, but maybe the code block below is messing it up?

docs/internationalization.rst
```diff
@@ -0,0 +1,65 @@
...
+Paypal Language Negotiation
+~~~~~~~~~~~~~~~~~~~~~
+
+To enable localization for Paypal, enable the waffle switch "create_and_set_webprofile" (The language request is sent to
+Paypal in a "WebProfile") in the admin. Waffle switches
+can be enabled at http://YOUR_HOST_NAME/admin/waffle/switch/
+
+The language code for ecommerce will be taken from a cookie as described in Ecommerce Language Negotation. When the
+language code is fetched from the cookie, only the base language will be used. For example, 'es-419' will resolve to
+'es'. Paypal takes a country code and then resolves a language from this code so 'es' will map to 'MX' when it is sent
```
mikix: This sentence is confusing.  You talk about country -> language, but then use an example of language -> country.

Maybe the following is clearer?  "then resolves a language from this code so 'es' on the ecommerce side will map to 'MX' which is sent to Paypal who then resolves 'MX' back to 'es'"

mikix (changes requested):
Looks real good.  Couple comments.

docs/internationalization.rst
```diff
@@ -0,0 +1,65 @@
...
+to Paypal. To add your language for Paypal, look up `Paypal's country to language mapping`_ and add it to PAYPAL_LOCALES
+in 'ecommerce/extensions/payment/constants.py'.
+
+.. code-block:: bash
```
mikix: bash not python? (used two places here)

docs/internationalization.rst
mikix: Or "then resolves a language from this code.  For example, ...."

docs/internationalization.rst
BenjiLee: updated

docs/internationalization.rst
BenjiLee: I was artificially newlining the text. I took all the new lines out. I believe readthedocs and github wiki should handle that for us.

docs/internationalization.rst
```diff
@@ -0,0 +1,52 @@
...
+
+.. code-block:: python
+
+  LANGUAGES = (
+    ('en', _('English')),
+    ('es', _('Spanish')),
+    ('es-419', _('Spanish (Latin American)')),
+  )
+
+.. _internationalization coding guidelines: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/i18n.html
```
catong: Suggest either moving all links down to bottom of the file, or move to a separate links.rst document so that common links used in multiple files don't need to be duplicated and can be maintained in one place. If you use a separate links.rst file, then you will need to add an "include" in the files that refer to links.rst, e.g. ```.. include:: ../../links.rst```

docs/internationalization.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Ecommerce Language Negotiation
+~~~~~~~~~~~~~~~~~~~~~
+Language negotiation for the ecommerce site will be handled by `Django's Locale Middleware`_. Below is the list of Django's Language Negotiation rule in priority order.
```
catong: "will be handled" -> "is handled"

catong: "Django's Language Negotiation rules, in priority order, are as follows."

docs/internationalization.rst
```diff
@@ -0,0 +1,52 @@
...
+
+1. Language prefix is not enabled by default for the ecommerce application.
```
catong: Use ```#.``` for automatic list numbering.

docs/internationalization.rst
```diff
@@ -0,0 +1,52 @@
...
+2. LANGUAGE_SESSION_KEY is not used.
+3. LANGUAGE_COOKIE_NAME is used to negotiate language. A user's language is set in their account settings in edx-platform which sets the language cookie. The language cookie name should be the same cookie name set by edx-platform. This can be configured in 'ecommerce/settings/base.py'.
```
catong: For filename/monospace formatting use double backquotes (``) instead of single straight quote.

docs/internationalization.rst
```diff
@@ -0,0 +1,52 @@
...
+  .. code-block:: python
+
+    LANGUAGE_COOKIE_NAME = 'openedx-language-preference'
+
+4. The Accept-Language HTTP header is used.
+5. LANGUAGE_CODE is used as the last resort. This can be set in 'ecommerce/settings/base.py'
+
+.. _Django's Locale Middleware: https://docs.djangoproject.com/en/2.0/topics/i18n/translation/#how-django-discovers-language-preference
+
+Paypal Language Negotiation
```
catong: Two uppercase "P"s in "PayPal"

docs/internationalization.rst
```diff
@@ -0,0 +1,52 @@
...
+~~~~~~~~~~~~~~~~~~~~~
+
+To enable localization for Paypal, enable the waffle switch "create_and_set_webprofile" (The language request is sent to Paypal in a "WebProfile") in the admin. Waffle switches can be enabled at http://YOUR_HOST_NAME/admin/waffle/switch/
```
catong: "Paypal" -> "PayPal" throughout

docs/internationalization.rst
catong: Throughout, apply line wrapping to make it easier to comment on specific lines. Can be done using Cmd+option+Q with your cursor somewhere in the paragraph.

docs/internationalization.rst
catong: "in the admin" seems a bit unclear - do you mean in Django Admin? Suggest following a consistent pattern for referring to Django Admin, as in: 

Also, http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/ecommerce/enable_receipt_page.html

Remove the parenthetical statement and make it the following sentence, rather than breaking up the first sentence.

docs/internationalization.rst
```diff
@@ -0,0 +1,52 @@
...
+
+A language code for the ecommerce IDA will be taken from a cookie as described in Ecommerce Language Negotation. When the language code is fetched from the cookie, only the base language will be used. For example, 'es-419' will resolve to 'es'. Paypal requires a country code. To get the country code, we use language code to map it to a country. For example, the language code 'es' will map to the country code 'MX' when it is sent to Paypal. To add your language for Paypal, look up `Paypal's country to language mapping`_ and add it to PAYPAL_LOCALES in 'ecommerce/extensions/payment/constants.py'.
+
+.. code-block:: python
```
catong: This code block is not showing up in the generated HTML.

docs/internationalization.rst
```diff
@@ -0,0 +1,52 @@
...
+5. LANGUAGE_CODE is used as the last resort. This can be set in 'ecommerce/settings/base.py'
+
+.. _Django's Locale Middleware: https://docs.djangoproject.com/en/2.0/topics/i18n/translation/#how-django-discovers-language-preference
+
+Paypal Language Negotiation
+~~~~~~~~~~~~~~~~~~~~~
+
+To enable localization for Paypal, enable the waffle switch "create_and_set_webprofile" (The language request is sent to Paypal in a "WebProfile") in the admin. Waffle switches can be enabled at http://YOUR_HOST_NAME/admin/waffle/switch/
+
+A language code for the ecommerce IDA will be taken from a cookie as described in Ecommerce Language Negotation. When the language code is fetched from the cookie, only the base language will be used. For example, 'es-419' will resolve to 'es'. Paypal requires a country code. To get the country code, we use language code to map it to a country. For example, the language code 'es' will map to the country code 'MX' when it is sent to Paypal. To add your language for Paypal, look up `Paypal's country to language mapping`_ and add it to PAYPAL_LOCALES in 'ecommerce/extensions/payment/constants.py'.
```
catong: Missing "i" in "Negotation"
"base language will be used" -> "base language is used"
"will resolve to" -> "resolves to"

Use double backquotes (``) for monospace formatting rather than single straight quote/

docs/internationalization.rst
```diff
@@ -0,0 +1,52 @@
...
+
+.. code-block:: python
+    PAYPAL_LOCALES = {
+        'zh': 'CN',
+        'fr': 'FR',
+        'en': 'US',
+        'es': 'MX',
+    }
+
+If a language fetched from the cookie cannot be found in PAYPAL_LOCALES, the LANGUAGE_CODE is used. If the LANGUAGE_CODE does not exist in PAYPAL_LOCALES, Paypal will use its own language negotiation.
```
catong: "Paypal" -> "PayPal"

catong (changes requested):
@BenjiLee Did a first pass and left comments inline.

docs/internationalization.rst
BenjiLee: I ended up force quitting :(

BenjiLee:
Closed in favor of https://github.com/edx/edx-documentation/pull/1641

PR #1635, opened 2018-01-19
Migrations to remove Django history tables

ahsan-ul-haq:
WE No longer using the django history tables so the purpose of
this PR is to add migrations that will eventually delete history tables
LEARNER-3349


ahsan-ul-haq:
Running this migration on local we get following on bash
```
The following content types are stale and need to be deleted:

    app_name | model_name

Any objects related to these content types by a foreign key will also
be deleted. Are you sure you want to delete these content types?
If you're unsure, answer 'no'.

    Type 'yes' to continue, or 'no' to cancel: 
```
By typing yes it work fine. So after merging these will our goCD pipeline automated stage depoyment would affected? Also how we can run these on production?
cc: @edx/devops

mikedikan:
After discussing with @mjfrey - I think we need to ask devops, we don't believe that we can have prompts on the command line, but they might know better.  I recommend you tag devops on the PR and ask advice.

jibsheet:
You can test locally using the ansible playbook that gocd users and see what it does
https://github.com/edx/configuration/blob/master/playbooks/continuous_delivery/run_migrations.yml
That's run out of edx-gomatic which you can search to better mirror the invocation

ahsan-ul-haq:
@jibsheet thanks for providing the direction to investigate.
@mikedikan after looking into the file provided by Kevin it seems like during goCD deployment migration is run by the management command provided [here](https://github.com/edx/edx-django-release-util/blob/master/release_util/management/commands/run_migrations.py) which is run with [--noinput](https://docs.djangoproject.com/en/2.0/ref/django-admin/#cmdoption-migrate-noinput) argument so prompts will not be generated on goCD. This means running this migration won't effect deployment.

mikedikan:
Great, thanks for digging in @ahsan-ul-haq

mikedikan (approved):

PR #1640, opened 2018-01-23
Fix migrations to avoid prompt in ansible run

nadeemshahzad:
Ansible is failing for sandbox running migration for ecommerce after this PR is merged https://github.com/edx/ecommerce/pull/1635

https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/8523/consoleFull

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ production-requirements: requirements.js
 	pip install -r requirements.txt --exists-action w
 
 migrate:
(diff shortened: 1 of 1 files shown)


coryleeio (approved):

coryleeio:
Approving this bc it's the way we do it in prod, so it should be fine for dev, and it is broken in devstack as well.

PR #1641, opened 2018-01-24
upgrade django oscar to latest version

ahsan-ul-haq:
LEARNER-2634

FYI: @edx/learner @edx/learner-spartans


ahsan-ul-haq:
@mikedikan @mjfrey please review

mikedikan:
@ahsan-ul-haq - I just connected with @mjfrey and we will be able to review, but not today.

ahsan-ul-haq:
@mikedikan thats great thanks for the follow up

ahsan-ul-haq:
Here are number of records in in the tables which will be affected by migrations
```
catalogue_productattributevalue = 38255
catalogue_productattribute = 13
customer_email = 254
offer_conditionaloffer = 18916
voucher_voucher = 174274
voucher_voucherapplication = 63398
```

mjfrey (approved):
Did some local testing and seems fine.  Please test this a lot on stage before releasing. Also give devops a heads up about the migrations.

ecommerce/extensions/offer/models.py
```diff
@@ -37,6 +37,12 @@ def clean_value(self):
                 'Failed to create Benefit. Benefit value may not be a negative number.'
             )
 
+    def clean_percentage(self):
```
bderusha, line 40: How does this get called?

bderusha (approved):
In general when we see `AlterField` in migrations we need to be cautious.  For large tables this can cause an issue with our deployments.  In this case the tables in question are small so it should be fine to have a lock on that table for a few milliseconds while the changes are applied.

ecommerce/extensions/offer/models.py
ahsan-ul-haq: This get called when `Benefit` object is created.

PR #1642, opened 2018-01-29
Support creating entitlements

mikix:
Allow creating entitlement products via the AtomicPublication API. Just pass them in like normal seat products, with an entitlement product class instead of a seat class.

**Notes**
* I refactor a few places to more cleanly separate seat paths and entitlement paths (in both real code and tests). Even making a couple dummy static classes to group them together.
* I make a call out to the catalog to lookup the UUID for the course. That appears to the be source of truth for that info, and we didn't have it locally before. But I do it inside the atomic transaction, so if that fails for any reason, we'll safely roll back.
* I switched some code that looked up if an entitlement already existed based on the title and made it do same thing based on course UUID. Seems like a better choice and was something I ran into when testing. But just calling it out since it might look odd at first blush.
* I removed the old entitlement-creation endpoint (POSTing to our products endpoint). That had been added in anticipation of publisher using it (no one else uses it). But we want to use the newer atomic-publish endpoint model.

https://openedx.atlassian.net/browse/LEARNER-3891

This will need a matching course-discovery branch to push entitlement products. But this is the receiving side of that which can and should land before the discovery side does. https://github.com/edx/course-discovery/pull/1301 for the discovery side.

### DIFF
--- a/ecommerce/entitlements/utils.py
+++ b/ecommerce/entitlements/utils.py
@@ -22,8 +22,15 @@ def create_parent_course_entitlement(name, UUID):
     parent, created = Product.objects.get_or_create(
         structure=Product.PARENT,
         product_class=ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME),
-        title='Parent Course Entitlement for {}'.format(name),
+        attributes__name='UUID',
+        attribute_values__value_text=UUID,
+        defaults={
+            'title': 'Parent Course Entitlement for {}'.format(name),
+            'is_discountable': True,
+        },
     )
+    parent.attr.UUID = UUID
+    parent.attr.save()
 
     if created:
         logger.debug('Created new parent course_entitlement [%d] for [%s].', parent.id, UUID)
@@ -31,10 +38,6 @@ def create_parent_course_entitlement(name, UUID):
         logger.debug('Parent course_entitlement [%d] already exists for [%s].', parent.id, UUID)
 
     ProductCategory.objects.get_or_create(category=Category.objects.get(name='Course Entitlements'), product=parent)
-    parent.title = 'Parent Course Entitlement for {}'.format(name)
-    parent.is_discountable = True
-    parent.attr.UUID = UUID
-    parent.save()
 
     return parent, created
 
@@ -45,8 +48,11 @@ def create_or_update_course_entitlement(certificate_type, price, partner, UUID, 
     certificate_type = certificate_type.lower()
     UUID = unicode(UUID)
 
+    uuid_query = Q(
+        attributes__name='UUID',
+        attribute_values__value_text=UUID,
+    )
     certificate_type_query = Q(
-        title='Course {}'.format(name),
         attributes__name='certificate_type',
         attribute_values__value_text=certificate_type,
     )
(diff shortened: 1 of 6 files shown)


mikix:
@Mjloturco tagging you so you can do a review if you like -- even just a sanity check on the description would be good.

ecommerce/extensions/api/serializers.py
```diff
@@ -317,6 +325,73 @@ class Meta(object):
...
 
+class EntitlementProductHelper(object):
+    @staticmethod
+    def validate(product):
+        attrs = _flatten(product['attribute_values'])
+
+        if not attrs.get('certificate_type'):
+            raise serializers.ValidationError(_(u"Products must have a certificate type."))
+
+        if product.get('price') is None:
```
jlajoie: nit: Should this match the conditional above it checking for existence of that key?

ecommerce/extensions/api/serializers.py
```diff
@@ -317,6 +325,73 @@ class Meta(object):
...
+            raise serializers.ValidationError(_(u"Products must have a price."))
+
+    @staticmethod
+    def save(partner, course, uuid, product):
```
jlajoie: Just double checking, save only runs after validate I'm assuming?

ecommerce/entitlements/utils.py
```diff
@@ -45,22 +48,25 @@ def create_or_update_course_entitlement(certificate_type, price, partner, UUID,
...
         all_products = parent_entitlement.children.all().prefetch_related('stockrecords')
-        course_entitlement = all_products.get(certificate_type_query)
+        course_entitlement = all_products.filter(uuid_query).get(certificate_type_query)
     except Product.DoesNotExist:
         course_entitlement = Product()
 
     course_entitlement.structure = Product.CHILD
     course_entitlement.is_discountable = True
-    course_entitlement.title = 'Course {}'.format(name)
+    course_entitlement.title = 'Course Entitlement for {}'.format(name)
```
mjfrey: We certainly do not want this change.  This title is what is shown to the user in the basket.  The word entitlement should never be user visible.

mikix: Oof OK. That change is not critical to this PR, I can undo.

It's not translated (and neither is the Seat title). I'm guessing that's because of Oscar difficulties with storing the translation somewhere?

ecommerce/extensions/api/serializers.py
mikix: Yeah. This class uses the django-rest-framework flow. save() operates on self.validated_data[] which is filled in by the framework during its validation (which calls stuff like our validate_products() method because we named it validate_XXX where XXX is an expected field).

ecommerce/entitlements/utils.py
mikix: Changed back. Thanks for heads up!

ecommerce/extensions/api/serializers.py
mikix: Fixed. (I actually changed both to use the "key not in list" format.)

jlajoie (approved):
Lgtm 👍

mjfrey (approved):

PR #1643, opened 2018-01-31
Upgrade django to latest version

ahsan-ul-haq:
LEARNER-1306

FYI: @edx/learner @edx/learner-spartans

### DIFF
--- a/ecommerce/core/tests/test_admin.py
+++ b/ecommerce/core/tests/test_admin.py
@@ -1,5 +1,5 @@
 from django.contrib import messages
-from django.core.urlresolvers import reverse
+from django.urls import reverse
 from oscar.test.factories import UserFactory
 
 from ecommerce.core.constants import USER_LIST_VIEW_SWITCH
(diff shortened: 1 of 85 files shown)


ecommerce/settings/base.py
```diff
@@ -155,7 +155,7 @@
 # SITE CONFIGURATION
 # Hosts/domain names that are valid for this site
 # See https://docs.djangoproject.com/en/1.5/ref/settings/#allowed-hosts
-ALLOWED_HOSTS = []
+ALLOWED_HOSTS = ['*']
```
bderusha: do we override this in production?

bderusha (approved):
One question about the ALLOWED_HOSTS setting.  Just want to make sure this doesn't open us up to any security risk.  Otherwise looks good!

mjfrey (approved):
Looks good.

ecommerce/settings/base.py
ahsan-ul-haq: I don't think so. We can add it in the test settings and local settings and keep it `[]` here.

PR #1646, opened 2018-02-08
Allow multi-use coupons for bundles.

mjfrey:
Only allow bundle coupons and multi-use vouchers for baskets that contain bundles.

LEARNER-3466

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -454,7 +454,9 @@ class VoucherAddView(BaseVoucherAddView):  # pylint: disable=function-redefined
             basket=self.request.basket,
             attribute_type=BasketAttributeType.objects.get(name=BUNDLE)
         )
-        if len(bundle_attribute) > 0 and not voucher.offers.first().condition.program_uuid:
+        if len(bundle_attribute) > 0 and not (
+                voucher.offers.first().condition.program_uuid or
+                voucher.usage == Voucher.MULTI_USE):
             messages.error(
                 self.request,
                 _("Coupon code '{code}' is not valid for this basket.").format(code=code))


ecommerce/extensions/basket/views.py
```diff
@@ -454,7 +454,9 @@ def apply_voucher_to_basket(self, voucher):
             basket=self.request.basket,
             attribute_type=BasketAttributeType.objects.get(name=BUNDLE)
         )
-        if len(bundle_attribute) > 0 and not voucher.offers.first().condition.program_uuid:
+        if len(bundle_attribute) > 0 and (
```
Mjloturco: Would it be worth it to just break these each into separate failure conditions? Then our logging could be more Granular as well like 
"Coupon code '{code}' is not valid for this basket, because coupon is Single use").format(code=code)
and 
"Coupon code '{code}' is not valid for this basket, this coupon cannot be used for programs").format(code=code)

Might make the boolean logic less painful too? (This logic takes a bit to parse)

Mjloturco: Looked at this more, I agree this boolean logic hurts. Is our desired behavior is to allow users to use coupons when the coupon is a program coupon, or the coupon is a multiuse coupon?

ecommerce/extensions/basket/views.py
```diff
@@ -454,7 +454,9 @@ def apply_voucher_to_basket(self, voucher):
...
+                not voucher.offers.first().condition.program_uuid or
+                voucher.type == Voucher.MULTI_USE):
```
Mjloturco: I think its 
```
not voucher.offers.first().condition.program_uuid and
                not voucher.type == Voucher.MULTI_USE):
```
Which with a little manipulation is maybe more humanly readable
```
not (voucher.offers.first().condition.program_uuid or voucher.type == Voucher.MULTI_USE)
```
Which is pretty much what you have but order of operations would make the not evaluate before the or, so the parentheses are needed.

staubina (approved):

mikix (approved):
Seems reasonable

Mjloturco (approved):
Nice.

PR #1653, opened 2018-02-15
ENT-803:  Added management command to create enrollment codes for courses.

saleem-latif:
__Description:__
This PR adds a management command to add enrollment codes for all existing courses.

**JIRA Ticket:** [ENT-803](https://openedx.atlassian.net/browse/ENT-803)

### DIFF
--- a/ecommerce/courses/management/commands/create_enrollment_codes.py
+++ b/ecommerce/courses/management/commands/create_enrollment_codes.py
@@ -1,0 +1,230 @@
+"""
+This command generates enrollment codes for courses.
+"""
+from __future__ import unicode_literals
+
+import logging
+import os
+
+from django.core.management import BaseCommand, CommandError
+
+from ecommerce.core.constants import ENROLLMENT_CODE_SEAT_TYPES
+from ecommerce.courses.models import Course
+
+logger = logging.getLogger(__name__)
+
+
+class CourseInfoError(Exception):
+    """
+    Raised when course does not have all the required data.
+    """
+    pass
+
+
+class Command(BaseCommand):
+    """
+    Creates enrollment codes for courses.
+    """
+
+    help = 'Create enrollment codes for courses.'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--course-ids-file',
+            action='store',
+            dest='course_ids_file',
+            default=None,
+            help='Path of the file to read courses id from.',
+            type=str,
+        )
+        parser.add_argument(
+            '--batch-limit',
+            action='store',
+            dest='batch_limit',
+            default=100,
+            help='Number of courses in each batch of enrollment code creation.',
+            type=int,
+        )
+
+    def handle(self, *args, **options):
+        course_ids_file = options['course_ids_file']
+        batch_limit = options['batch_limit']
+
+        if course_ids_file:
+            if not os.path.exists(course_ids_file):
+                raise CommandError('Pass the correct absolute path to course ids file as --course_ids_file argument.')
+
+            total_courses, failed_courses = self._generate_enrollment_codes_from_file(course_ids_file)
+        else:
+            total_courses, failed_courses = self._generate_enrollment_codes_from_db(batch_limit)
+
+        if failed_courses:
+            logger.error('Completed enrollment codes generation. %d of %d failed.', len(failed_courses), total_courses)
+            logger.error('\n'.join(['Failed courses:'] + failed_courses))
+        else:
+            logger.info('Successfully generated enrollment codes for the batch of %s courses.', total_courses)
+
+    def _generate_enrollment_codes_from_db(self, batch_limit):
+        """
+        Generate enrollment codes for the course.
+
+        Arguments:
+            batch_limit (int): How many courses to fetch from db to process in each batch.
+
+        Returns:
+            (total_course, failed_course): a tuple containing count of course processed and a list containing ids of
+                courses whose enrollment codes could not be generated.
+        """
+        failed_courses = []
+        total_courses = 0
+
+        courses = Course.objects.all()[0:batch_limit]
+        while len(courses) > 0:
+            total_courses += len(courses)
+            logger.info('Creating enrollment code for %d courses.', courses.count())
+
+            for course in courses:
+                try:
+                    self._generate_enrollment_code(course)
+                except CourseInfoError as error:
+                    logger.error(
+                        'Enrollment code generation failed for "%s" course. Because %s',
+                        course.id,
+                        error.message,
+                    )
+                    failed_courses.append(course.id)
+
+            courses = Course.objects.all()[total_courses:total_courses + batch_limit]
+        return total_courses, failed_courses
+
+    def _generate_enrollment_codes_from_file(self, course_ids_file):
+        """
+        Generate enrollment codes for the course provided in the course ids file.
+
+        Arguments:
+            course_ids_file (str): path of the file containing course ids.
+
+        Returns:
+            (total_course, failed_course): a tuple containing count of course processed and a list containing ids of
+                courses whose enrollment codes could not be generated.
+        """
+        failed_courses = []
+
+        with open(course_ids_file, 'r') as file_handler:
+            course_ids = file_handler.readlines()
+            total_courses = len(course_ids)
+            logger.info('Creating enrollment code for %d courses.', total_courses)
+            for index, course_id in enumerate(course_ids, start=1):
+                try:
+                    course_id = course_id.strip()
+                    course = Course.objects.get(id=course_id)
+                    self._generate_enrollment_code(course)
+                except Course.DoesNotExist:
+                    failed_courses.append(course_id)
+                    logger.error(
+                        '(%d/%d) Failed to generate enrollment codes for "%s": Course does not exist.',
+                        index,
+                        total_courses,
+                        course_id
+                    )
+                except CourseInfoError as error:
+                    logger.error(
+                        'Enrollment code generation failed for "%s" course. Because %s',
+                        course.id,
+                        error.message,
+                    )
+                    failed_courses.append(course.id)
+        return total_courses, failed_courses
+
+    def _generate_enrollment_code(self, course):
+        """
+        Generate enrollment code for the given course.
+
+        Enrollment code is generated if
+            1. The course has a course mode that supports enrollment codes AND
+            2. The course does not already have an enrollment code for that course mode.
+
+        Arguments:
+            course (Course): E-Commerce course object.
+        """
+        if self.is_course_eligible_for_enrollment_code(course) and not course.get_enrollment_code():
+            seat_type, price, id_verification_required = self.get_course_info(course)
+            course._create_or_update_enrollment_code(  # pylint: disable=protected-access
+                seat_type=seat_type,
+                id_verification_required=id_verification_required,
+                partner=course.partner,
+                price=price,
+                expires=None
+            )
+            logger.info('Enrollment code generated for "%s" course.', course.id)
+        elif self.is_course_eligible_for_enrollment_code(course) and course.get_enrollment_code():
+            logger.info(
+                'Skipping enrollment code generation for "%s" course due to existing enrollment codes.',
+                course.id,
+            )
+        else:
+            logger.info(
+                'Skipping enrollment code generation for "%s" course. '
+                'Because enrollment codes are not allowed for "%s" seat type.',
+                course.id,
+                ', '.join([getattr(seat.attr, 'certificate_type', '').lower() for seat in course.seat_products]),
+            )
+
+    @staticmethod
+    def is_course_eligible_for_enrollment_code(course):
+        """
+        Determine if given course is eligible for an enrollment code.
+
+        Courses that have one of the following seat types are eligible for an enrollment code.
+         1. verified
+         2. professional
+         3. no-id-professional
+
+        Arguments:
+            course (Course): E-Commerce course object.
+
+        Returns:
+            (bool): True if given course is eligible for enrollment code, False otherwise.
+        """
+        seat_types = [getattr(seat.attr, 'certificate_type', '').lower() for seat in course.seat_products]
+        for seat_type in seat_types:
+            if seat_type in ENROLLMENT_CODE_SEAT_TYPES:
+                return True
+        return False
+
+    @staticmethod
+    def get_course_info(course):
+        """
+        Get course info required for the creation of enrollment code.
+
+        Arguments:
+            course (Course): E-Commerce course object.
+
+        Returns:
+            (seat_type, price, id_verification_required): A tuple containing the following info
+                seat_type: Course seat type e.g. verified, professional or no-id-professional
+                price: Price of the course.
+                id_verification_required: boolean indicating whether course requires student identity verification.
+        Raises:
+            (Exception): Raised if given course has either multiple seats eligible for enrollment code or no seat
+                eligible for enrollment code.
+        """
+        seats = [
+            seat for seat in course.seat_products if
+            getattr(seat.attr, 'certificate_type', '').lower() in ENROLLMENT_CODE_SEAT_TYPES
+        ]
+        if len(seats) == 1:
+            seat = seats[0]
+            seat_type = getattr(seat.attr, 'certificate_type', '').lower()
+            price = seat.stockrecords.all()[0].price_excl_tax
+            id_verification_required = getattr(seat.attr, 'id_verification_required', False)
+
+            return seat_type, price, id_verification_required
+        elif len(seats) > 1:
+            raise CourseInfoError(
+                'Course "%s" has multiple seats eligible for enrollment codes.' % course.id
+            )
+        else:
+            raise CourseInfoError(
+                'Course "%s" does not have any seat eligible for enrollment codes.' % course.id
+            )
--- a/ecommerce/courses/tests/test_create_enrollment_codes_command.py
+++ b/ecommerce/courses/tests/test_create_enrollment_codes_command.py
@@ -1,0 +1,340 @@
+# encoding: utf-8
+"""Contains the tests for enrollment code creation command."""
+
+from __future__ import unicode_literals
+
+import logging
+import os
+import tempfile
+from decimal import Decimal
+
+from django.core.management import CommandError, call_command
+from testfixtures import LogCapture
+
+from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.extensions.catalogue.tests.mixins import DiscoveryTestMixin
+from ecommerce.tests.testcases import TransactionTestCase
+
+logger = logging.getLogger(__name__)
+LOGGER_NAME = 'ecommerce.courses.management.commands.create_enrollment_codes'
+
+
+class CreateEnrollmentCodesTests(DiscoveryTestMixin, TransactionTestCase):
+    """
+    Tests the enrollment code creation command.
+    """
+    tmp_file_path = os.path.join(tempfile.gettempdir(), 'tmp-courses.txt')
+
+    def setUp(self):
+        """
+        Create test courses and and temp file containing course ids.
+        """
+        super(CreateEnrollmentCodesTests, self).setUp()
+        self.professional_course_1 = self.create_course(seat_type=str('professional'))
+        self.professional_course_2 = self.create_course(seat_type=str('professional'))
+        self.audit_course = self.create_course(seat_type=str('audit'))
+        self.verified_course = self.create_course(seat_type=str('verified'))
+
+        self.create_course_ids_file(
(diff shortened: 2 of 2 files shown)


ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+from django.core.management import BaseCommand, CommandError
+
+from ecommerce.core.constants import ENROLLMENT_CODE_SEAT_TYPES
+from ecommerce.courses.models import Course
+
+logger = logging.getLogger(__name__)
+
+
+class CourseInfoError(Exception):
+    pass
```
zubair-arbi: Please add some docstring for this exception, e.g. `""" Raised when we fail to generate enrollment code for a course. """`
Also since this is generic exception which we will raise when we fail to generate enrollment codes for a course so maybe we could name it something like `CourseEnrollmentCodeGenerationError`.

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+
+class Command(BaseCommand):
+    """
+    Creates enrollment codes for courses.
+    """
+
+    help = 'Create enrollment codes for courses.'
+
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.DEBUG)
```
zubair-arbi: nit: remove this code for debugging,

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+    logger.addHandler(ch)
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--course_ids_file',
+            action='store',
+            dest='course_ids_file',
+            default=None,
+            help='Path to file to read courses from.',
```
zubair-arbi: `Path of the file to read courses id from.'`

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+        parser.add_argument(
+            '--offset',
+            action='store',
+            dest='offset',
+            default=0,
+            help='Number of courses to skip for enrollment code creation.',
+            type=int,
+        )
+        parser.add_argument(
+            '--limit',
```
zubair-arbi: for consistency with `batch-sleep-time` argument we can rename it to `batch-limit`

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+            '--batch-sleep-time',
+            action='store',
+            dest='batch_sleep_time',
+            default=3,
+            help='Sleep time (in minutes) between the processing batches for courses.',
+            type=int,
+        )
+
+    def handle(self, *args, **options):
+        failed = []
```
zubair-arbi: nit `failed_courses`

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+                    except CourseInfoError as error:
+                        logger.error(
+                            'Enrollment code generation failed for "%s" course. Because %s',
+                            course.id,
+                            error.message,
+                        )
+                        failed.append(course.id)
+        else:
+            courses = Course.objects.all()[offset:offset + limit]
+            while courses.count() > 0:
```
zubair-arbi: Use len(courses)
> Note that if you want the number of items in a QuerySet and are also retrieving model instances from it (for example, by iterating over it), it’s probably more efficient to use len(queryset) which won’t cause an extra database query like count() would.
https://docs.djangoproject.com/en/dev/ref/models/querysets/#count

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+                            course.id,
+                            error.message,
+                        )
+                        failed.append(course.id)
+
+                time.sleep(batch_sleep_time * 60)
+                courses = Course.objects.all()[offset:offset + limit]
+
+        if failed:
+            logger.error('Completed enrollment codes generation. %d of %d failed.', len(failed), total_courses)
```
zubair-arbi: `'Completed processing for {total_courses_count} total courses for enrollment codes generation out of which {failed_courses_count} courses failed.'`

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+            logger.error('\n'.join(['Failed courses:'] + failed))
+        else:
+            logger.info('Enrollment codes created for %d courses.', total_courses)
```
zubair-arbi: `Successfully generated enrollment codes for all {total_courses_count} courses.`

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+        else:
+            logger.info(
+                'Enrollment code generation skipped for "%s" course. '
+                'Because enrollment codes are not allowed for "%s" seat type.',
+                course.id,
+                ', '.join([getattr(seat.attr, 'certificate_type', '').lower() for seat in course.seat_products]),
+            )
+
+    @staticmethod
+    def course_is_eligible_for_enrollment_code(course):
```
zubair-arbi: nit `course_is_eligible_for_enrollment_code` -> `is_course_eligible_for_enrollment_code`

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+            logger.error('\n'.join(['Failed courses:'] + failed))
+        else:
+            logger.info('Enrollment codes created for %d courses.', total_courses)
+
+    def _create_enrollment_code(self, course):
+        """
+        Create enrollment code for the given course.
+
+        Enrollment code is generated if
+            1. Course have a type that supports enrollment codes AND
```
zubair-arbi: Course have course modes that supports enrollment codes

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+                seat_type=seat_type,
+                id_verification_required=id_verification_required,
+                partner=course.partner,
+                price=price,
+                expires=None
+            )
+            logger.info('Enrollment code created for "%s" course.', course.id)
+        elif self.course_is_eligible_for_enrollment_code(course) and course.get_enrollment_code():
+            logger.info(
+                'Enrollment code generation skipped for "%s" course. Because it already has enrollment codes.',
```
zubair-arbi: `'Skipping enrollment code generation for "%s" course due to existing enrollment codes.'`

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+                course.id,
+            )
+        else:
+            logger.info(
+                'Enrollment code generation skipped for "%s" course. '
```
zubair-arbi: `'Skipping enrollment code generation for "%s" course.`

ecommerce/courses/tests/test_create_enrollment_codes_command.py
```diff
@@ -0,0 +1,313 @@
...
+    """
+    tmp_file_path = os.path.join(tempfile.gettempdir(), 'tmp-courses.txt')
+
+    def setUp(self):
+        """
+        Create test courses and and temp file containing course ids.
+        """
+        super(CreateEnrollmentCodesTests, self).setUp()
+        self.professional_course = self.create_course(seat_type=str('professional'))
+        self.professional_course2 = self.create_course(seat_type=str('professional'))
```
zubair-arbi: `self.professional_course_1` and `self.professional_course_2`

ecommerce/courses/tests/test_create_enrollment_codes_command.py
```diff
@@ -0,0 +1,313 @@
...
+        """
+        with open(file_path, 'w') as temp_file:
+            temp_file.write(str("\n".join(course_ids)))
+
+    def test_invalid_file_path(self):
+        """
+        Verify command raises the CommandError for invalid file path.
+        """
+        with self.assertRaises(CommandError):
+            call_command('create_enrollment_codes', course_ids_file="fake/path", batch_sleep_time=0)
```
zubair-arbi: `fake/path` -> `invalid/courses_id/file/path`

ecommerce/courses/tests/test_create_enrollment_codes_command.py
```diff
@@ -0,0 +1,313 @@
...
+                'ERROR',
+                'Completed enrollment codes generation. 1 of 1 failed.'
+            ),
+            (
+                LOGGER_NAME,
+                'ERROR',
+                '\n'.join(['Failed courses:', fake_course_id])
+            ),
+        )
+        with LogCapture(LOGGER_NAME) as lc:
```
zubair-arbi: nit `lc` -> `log_capture`

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+        failed = []
+        total_courses = 0
+
+        course_ids_file = options['course_ids_file']
+        offset = options['offset'] or 0
+        limit = options['limit'] or 100
+        batch_sleep_time = options['batch_sleep_time']
+
+        if course_ids_file:
+            if not os.path.exists(course_ids_file):
```
zubair-arbi: I think we can create 2 sub method for generating enrollment codes from file or from db which can return, counts for total processed courses and total failed courses, like:
`_generate_enrollment_code_from_file`
`_generate_enrollment_code_from_db`

zubair-arbi (changes requested):
Overall functionality looks good.

saleem-latif:
@georgebabey can you please take a look at the PR and share your thoughts?

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+            help='Sleep time (in minutes) between the processing batches for courses.',
+            type=int,
+        )
+
+    def handle(self, *args, **options):
+        failed = []
+        total_courses = 0
+
+        course_ids_file = options['course_ids_file']
+        offset = options['offset'] or 0
```
brittneyexline: is this redundant with the defaults provided from the parser?

ecommerce/courses/management/commands/create_enrollment_codes.py
brittneyexline: i agree with Zubair.

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+                        self._create_enrollment_code(course)
+                    except CourseInfoError as error:
+                        logger.error(
+                            'Enrollment code generation failed for "%s" course. Because %s',
+                            course.id,
+                            error.message,
+                        )
+                        failed.append(course.id)
+
+                time.sleep(batch_sleep_time * 60)
```
brittneyexline: can you explain why we need this sleep statement?

ecommerce/courses/management/commands/create_enrollment_codes.py
brittneyexline: The course has a course mode that supports enrollment codes

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+        else:
+            logger.info('Enrollment codes created for %d courses.', total_courses)
+
+    def _create_enrollment_code(self, course):
+        """
+        Create enrollment code for the given course.
+
+        Enrollment code is generated if
+            1. Course have a type that supports enrollment codes AND
+            2. Course does not already has an enrollment code
```
brittneyexline: The course does not already have an enrollment code for that course mode

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+         3. no-id-professional
+
+        Arguments:
+            course (Course): E-Commerce course object.
+
+        Returns:
+            (bool): True if given course is eligible for enrollment code, False otherwise.
+        """
+        seat_types = [getattr(seat.attr, 'certificate_type', '').lower() for seat in course.seat_products]
+        for seat_typ in seat_types:
```
brittneyexline: seat_type

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,219 @@
...
+        ]
+        if len(seats) == 1:
+            seat = seats[0]
+            seat_type = getattr(seat.attr, 'certificate_type', '').lower()
+            price = seat.stockrecords.all()[0].price_excl_tax
+            id_verification_required = getattr(seat.attr, 'id_verification_required', False)
+
+            return seat_type, price, id_verification_required
+
+        else:
```
brittneyexline: it probably shouldn't happen, but what if len(seats) > 1?

ecommerce/courses/tests/test_create_enrollment_codes_command.py
```diff
@@ -0,0 +1,313 @@
...
+            call_command('create_enrollment_codes', course_ids_file=self.tmp_file_path, batch_sleep_time=0)
+            lc.check(*expected)
+
+    def test_create_enrollment_codes_successfully_with_course_ids_file(self):
+        """
+        Verify enrollment codes are created successfully for all courses given in course ids file.
+        """
+        expected = (
```
brittneyexline, line 116: seems like this is copied in each test, and you have the ddt decorator above, is the intention here to refactor to use ddt?
i also think you should fetch the created enrollment code product from the database and check that instead of just checking log messages.

brittneyexline (changes requested):
great start!

ecommerce/courses/management/commands/create_enrollment_codes.py
saleem-latif: I added this in case we want to add sleep time before processing each batch of courses to reduce strain on database.

ecommerce/courses/tests/test_create_enrollment_codes_command.py
saleem-latif: I tried converting the tests to use ddt but setup for each test case is very different. I have removed the ddt import and class decorator.
I have also updated the tests to verify enrollment codes are actually created in the database.

saleem-latif:
@zubair-arbi , @brittneyexline I have made the changes you suggested, please take another look.

zubair-arbi (approved):
Changes looks good 👍 

@georgebabey @brittneyexline @saleem-latif  We need to decide how we will use this management command, e.g. running it against all courses at a time or manually providing the file with specific courses ids.

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,257 @@
...
+            batch_sleep_time (int): Time in seconds to sleep before processing next batch of courses.
+
+        Returns:
+            (total_course, failed_course): a tuple containing count of course processed and a list containing ids of
+                courses whose enrollment codes could not be generated.
+        """
+        failed_courses = []
+        total_courses = 0
+
+        courses = Course.objects.all()[batch_offset:batch_offset + batch_limit]
```
brittneyexline: I think we should be filtering this at least on verification_deadline not in the past. it also seems weird to me that we're calling for all the Course objects then taking a slice at two separate points - I am not sure we're getting the intended benefits of batching.

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,257 @@
...
+                    self._generate_enrollment_code(course)
+                except CourseInfoError as error:
+                    logger.error(
+                        'Enrollment code generation failed for "%s" course. Because %s',
+                        course.id,
+                        error.message,
+                    )
+                    failed_courses.append(course.id)
+
+            time.sleep(batch_sleep_time * 60)
```
brittneyexline: what evidence do you have that this is necessary, and if it is, that it should be a parameter that is passed in by the caller?

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,257 @@
...
+                batch_sleep_time,
+            )
+
+        if failed_courses:
+            logger.error('Completed enrollment codes generation. %d of %d failed.', len(failed_courses), total_courses)
+            logger.error('\n'.join(['Failed courses:'] + failed_courses))
+        else:
+            logger.info('Successfully generated enrollment codes for the batch of %s courses.', total_courses)
+
+    def _generate_enrollment_codes_from_db(self, batch_limit, batch_offset, batch_sleep_time):
```
brittneyexline: i would prefer we remove batch_offset as an option to be passed into the script. as a caller, i'm not sure why it would be useful, but i see why it is needed to keep track of batches below.

ecommerce/courses/management/commands/create_enrollment_codes.py
```diff
@@ -0,0 +1,257 @@
...
+            (total_course, failed_course): a tuple containing count of course processed and a list containing ids of
+                courses whose enrollment codes could not be generated.
+        """
+        failed_courses = []
+        total_courses = 0
+
+        courses = Course.objects.all()[batch_offset:batch_offset + batch_limit]
+        while len(courses) > 0:
+            total_courses += len(courses)
+            batch_offset += courses.count()
```
brittneyexline: does this mean that total_courses and batch_offset are always the same if we're starting from 0? if we get rid of batch_offset as a parameter then we could simplify this and have one variable.

ecommerce/courses/management/commands/create_enrollment_codes.py
saleem-latif: Verification deadline is not a required field and can be null, so we might miss some courses if we filter on verification_deadline. 
>it also seems weird to me that we're calling for all the Course objects then taking a slice at two separate points - I am not sure we're getting the intended benefits of batching.

https://docs.djangoproject.com/en/2.0/topics/db/queries/#limiting-querysets
Django handles slicing on queryset by creating a corresponding query with limit and offset so all courses will not fetched.

ecommerce/courses/management/commands/create_enrollment_codes.py
saleem-latif: My thinking is that if the command runs for say 5 minutes than it will make a lot of queries to database (i.e. to fetch courses, to create enrollment code products, to check if enrollment code product exists etc.) that may cause strain on database. I am not sure if it will. I wanted add this just in case.
I can update the default `batch_sleep_time` to 0 so that there is no sleep time in normal runs of the management command?

ecommerce/courses/management/commands/create_enrollment_codes.py
saleem-latif: Let's say we run the management command and it processes say 20% of the courses and then we terminate the command or it errors out for some reason. In this case we can pass in the appropriate `batch_offset` and start processing only the courses that need to be updated.

ecommerce/courses/management/commands/create_enrollment_codes.py
brittneyexline: i feel like there has got to be a way to more intelligently filter "all" courses. there are almost 6000 courses in that table on prod, and at least 2/3 of these can't possibly be active courses that you can buy a seat in. it may be that the best answer to the filtering question is we pull the list of courses and use the file option when running this command (in which case this branch isn't necessary), but i think that is something you should consider.

ecommerce/courses/management/commands/create_enrollment_codes.py
brittneyexline: we should be able to avoid "erroring out for some reason" short of a hardware failure. we could also send in the list of unprocessed courses, which at this point seems like the safer option.

i think ultimately, if you feel like you have to code defensively for unknown failures or performance issues for processing all courses, maybe it is better to force the caller to specify courses and remove the logic for running with all courses. i will leave this up to you.

saleem-latif:
@brittneyexline I have removed `batch_offset` and `batch_sleep_time` from the command please take another look.

brittneyexline (approved):
considering we will likely run this with the course id file, this looks good to go. also, we don't have to wait on merging this because of WL-1457, since we decide when to run the command.

PR #1655, opened 2018-02-16
[ENT-860]Update Basket and Receipt Page UI to include Bulk Enrollment Information

asadiqbal08:
##### Updated the screen shots
[ENT-860](https://openedx.atlassian.net/browse/ENT-860)

**Testing Intructions**

- Visit [business sandbox](https://ecommerce-business.sandbox.edx.org/courses/) and create a new verified course in ecommerce with bulk enrollment enabled. 
- Enroll a user through LMS and visit basket page. 
- View the UI changes there.

##### NOTE: Currently, Getting some error while adding a course in ecommerce on sandbox. A ticket is already logged [here](https://openedx.atlassian.net/browse/ENT-890). In that case, Try to [Hit the URL with a new user](https://ecommerce-business.sandbox.edx.org/basket/single-item/?sku=019379F) to view the page directly that I configured manually for a course on sandbox.

**1- Notifications when update the quantity and jump back to single purchase after clicking update button. 
2- Added the organization field, with required field validation, when there is an enrollment code product in basket.** 

**3- Update quantity message in notification bar.** 

quantity

**4- Single purchase message in notification bar along with a link to course about page "Click here to enroll directly"**

screen shot 2018-02-23 at 2 30 43 pm

**5- Receipt page changes.**

receipt page

**6- Organization is Required Field**

screen shot 2018-02-22 at 3 50 48 pm

#### NOTE: We need to update the receipt page message as I posted a comment on [JIRA](https://openedx.atlassian.net/browse/ENT-860)

PR #1657, opened 2018-02-19
ENT-804 Link order with enrollment code to the provided business client

zubair-arbi:
@asadiqbal08 @saleem-latif @georgebabey 
**Description:** Store `Company` Field data from Basket Page in `BusinessClient` model for bulk enrollment code purchases and link the order with the that Business Client for payment processors `cybersource`, `paypal`, `stripe` if business client value is passed from basket page as `organization`.

**JIRA:** [ENT-804](https://openedx.atlassian.net/browse/ENT-804)
**Related JIRA:** [ENT-860](https://openedx.atlassian.net/browse/ENT-860)

**Implemented Tasks:**
* Add `Bulk purchase` type in the `Invoice` model and also added the related migration.
* Added a method `handle_post_order` in common order mixin .`EdxOrderPlacementMixin` which is used by all payment processor (`cybersource`, `paypal`,  `stripe`).
* Called this method `handle_post_order` immediately after an order is created to link the current order with the provided business client for bulk enrollment purchase through `Invoice` model.
* Add unittests.

**Dependencies:** https://github.com/edx/ecommerce/pull/1655

**Merge deadline:** N/A

**Installation instructions:** Checkout ecommerce on this branch `zub/ENT-804-link-enrollment-code-order-with-client` and run order model migrations like `./manage.py migrate order`.

**Testing instructions:**

1. Create a course with in **LMS** and create its related modes in **ECOM** for single seat `SKU` and bulk purchase `Bulk SKU`
screen shot 2018-02-19 at 5 04 37 pm
2. In Ecom add bulk purchase product in the baseket by going to a url like: `http://localhost:8002/basket/single-item/?sku=4040F39`
screen shot 2018-02-19 at 5 06 17 pm
3. Provide the value for organization field (not present currently), e.g. `TEST CLIENT 1`. This field will be available once the related PR https://github.com/edx/ecommerce/pull/1655 is merged.
4. Click on "Place Order" button after providing the payment form values.
5. Now go to Ecom django admin for `core/businessclient` and verify that a new Business Client with the name `TEST CLIENT 1` has been added.
screen shot 2018-02-19 at 5 11 39 pm
6. Now go to Ecom django admin for `order/order` and verify that a new order has been placed with the provided payment values.
screen shot 2018-02-28 at 6 24 23 pm

7. Now go to Ecom django admin for `invoice/invoice` and verify that a new invoice has been generated against the above order.
screen shot 2018-02-28 at 6 24 36 pm

8. Now verify that above invoice has type `Bulk purchase` with state `Paid` and with relation to the new business client `TEST CLIENT 1`.
screen shot 2018-02-28 at 6 25 00 pm

9. Repeat the process for single seat purchase and verify that there is no invoice record against that order (order not linked with business client).


zubair-arbi:
@mjfrey @brittneyexline @georgebabey Please have look on these changes for linking orders with the business client provided from the basket page as field `organization`.

ecommerce/extensions/order/models.py
```diff
@@ -2,10 +2,14 @@
 from django.utils.translation import ugettext_lazy as _
 from oscar.apps.order.abstract_models import AbstractOrder, AbstractPaymentEvent
 
+from ecommerce.core.models import BusinessClient
 from ecommerce.extensions.fulfillment.status import ORDER
 
 
 class Order(AbstractOrder):
+    client = models.ForeignKey(
```
brittneyexline: i'm sorry to change this on you, but i don't  think we can add this to order. there are currently ~28.5 million rows in the order table in prod ecommerce, so altering this table is not really feasible.
unfortunately, i think it would be better to create a new model that foreign keys to both order and business client.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -320,7 +320,8 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
             return redirect(reverse('payment_error'))
 
         try:
-            self.create_order(request, basket, self._get_billing_address(notification))
+            order = self.create_order(request, basket, self._get_billing_address(notification))
+            self.link_order_with_business_client(request.POST, order)
```
brittneyexline: i would rename this method to something like handle_post_order because we're not necessarily linking all orders to a business client, and other post order handling might happen down the line that should probably be in one place.

ecommerce/extensions/order/models.py
zubair-arbi: @brittneyexline According to my understanding there are 2 ways to avoid downtime while running migrations:
1. Adding a (nullable) field 
2. Adding a new table

I have used the management command `create_fake_orders` to create 100000 dummy orders on **business sandbox** for testing these changes _(It took ~40 min to create 100000 fake orders)_.
```bash
./manage.py create_fake_orders --count=5 --sku=319955C
```
* On business sandbox running this migration took only 1 sec to run as we are adding a nullable field.
```
./manage.py migrate order
```
* Reverting this migrations took 3 sec
```
./manage.py migrate order 0016
```
You can verify migration time on the business sandbox.

I believe this is clean solution instead of adding a new m2m table.

brittneyexline (approved):
thanks for doing the research on the migration - let's move forward with your solution.

ecommerce/extensions/order/models.py
```diff
@@ -2,10 +2,14 @@
...
+        BusinessClient, verbose_name=_('Business Client'), null=True, blank=True
```
mjfrey: We certainly do not want to do this.  Do not change the built in Order model from Oscar.  Please add a key to the Invoice model.  That is a cleaner way as to not add a custom column to the generic Order model.

mjfrey (changes requested):
We do not want to modify the generic Order model that comes with Oscar.

zubair-arbi:
@mjfrey @brittneyexline @georgebabey I have updated the PR to use `Invoice` model for linking the bulk purchase orders with business clients. Please have another look.
Currently there are `1926` records in `Invoice` model (prod-replice) so the running the new `Invoice` migration won't have any problems.

zubair-arbi:
@asadiqbal08 @brittneyexline I have add the changes to store `organization` value from basket page as `BasketAttribute` and use it later to link order with the business client on CyberSource Silent Order POST.
Task Remaining: Tests update

zubair-arbi:
@mjfrey @brittneyexline @georgebabey @asadiqbal08 I have verified these changes on business sandbox.

ecommerce/extensions/payment/views/paypal.py
```diff
@@ -66,6 +67,8 @@ def _get_basket(self, payment_id):
             ).basket
             basket.strategy = strategy.Default()
             Applicator().apply(basket, basket.owner, self.request)
+
+            basket_add_organization_attribute(basket, self.request.GET)
```
brittneyexline, line 71: it seems odd to me to put this in each implementation of specific payment processors when this has nothing to do with individual payment processors. is there any way to put this somewhere common? where do other basket attributes get attached to the basket?

zubair-arbi: @brittneyexline I have made this change in order to have a consistent flow for each payment processor, e.g. saving the business client name as organization attribute on the basket immediately after we create a basket and retrieve it later when we have done creating order for linking the order with business client.
Also Cybersource and Stripe triggers order creation on POST request but PayPal triggers order creation on GET request, that's why I am manually providing request data to method `basket_add_organization_attribute`.

I am not sure if there is common place where we can put this maybe someone from @edx/ecommerce can give us suggestion.

There are few other places where we set basket attributes, e.g. `prepare_basket` but that method is called when basket is initialized to render a basket page, before providing the payment details on that basket page by the user.

brittneyexline (approved):

asadiqbal08 (approved):

PR #1658, opened 2018-02-20
Get UUID from discovery

mikix:
Course publishing was failing (I'm 80% sure) because -we- I misspelled a UUID lookup in ecommerce (should have been 'uuid' but was 'UUID'). But in looking at the fix for that, I realized I had gone about this wrong in the first place. I had added a call from ecommerce->discovery to get a course UUID. When discovery should just provide that from the get-go.

To my knowledge, no one else but discovery uses this endpoint, so we can safely change it in lockstep with discovery.

I still can't really test this end-to-end workflow locally. So I'm looking to land these in stage and confirm that it fixes the publishing problem I was seeing.

Related to https://github.com/edx/course-discovery/pull/1325

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_publication.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_publication.py
@@ -37,11 +37,13 @@ class AtomicPublicationTests(DiscoveryTestMixin, TestCase):
         super(AtomicPublicationTests, self).setUp()
 
         self.course_id = 'BadgerX/B101/2015'
+        self.course_uuid = '394a5ce5-6ff4-4b2b-bea1-a273c6920ae1'
         self.course_name = 'Dances with Badgers'
         self.create_path = reverse('api:v2:publication:create')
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/api/v2/tests/views/test_publication.py
```diff
@@ -364,6 +346,22 @@ def test_invalid_course_id(self):
         self.assertEqual(response.status_code, 400)
         self.assert_course_does_not_exist(self.course_id)
 
+    def test_missing_course_uuid(self):
+        """Verify that attempting to save a course without a UUID yields a 400."""
+        self.data.pop('uuid')
```
jlajoie, line 351: TIL there's a pop method for dictionaries.

jlajoie (approved):
:+1:

saleem-latif:
@mikix This change is causing an error on creating a course in the ecommerce https://openedx.atlassian.net/browse/ENT-890 is the related bug. 
The API endpoint updated here is used by ecommerce course admin tool which was not updated along with this change.

@zubair-arbi , @georgebabey -- FYI

PR #1660, opened 2018-02-22
ENT-803: Added functionality to always create enrollment codes with course creation.

saleem-latif:
@brittneyexline , @zubair-arbi , @georgebabey  -- FYI
__Description:__
This PR adds the functionality to create enrollment code be default along with course creation.

**JIRA Ticket:** [ENT-803](https://openedx.atlassian.net/browse/ENT-803)

**Testing Instructions:**

1. Go to studio (https://studio-business.sandbox.edx.org/home/) and create a new course, use this course to perform the following operations.
2. Make sure `Enable enrollment codes` at https://ecommerce-business.sandbox.edx.org/admin/core/siteconfiguration/1/change/ is unchecked.
3. Make sure `create_enrollment_codes` switch at https://ecommerce-business.sandbox.edx.org/admin/waffle/switch/3/change/ is not active.
4. Go to course creation page on ecommerce at https://ecommerce-business.sandbox.edx.org/courses/new/ and create a new course with professional mode.
5. Go to Product admin list view at https://ecommerce-business.sandbox.edx.org/admin/catalogue/product/ and make sure there is a product with the above course id and `Enrollment Code` product class.
6. Repeat 1-4 with an audit course and make sure product with the above course id and `Enrollment Code` product class is not created.


zubair-arbi (approved):
After testing 👍

brittneyexline (changes requested):
this is a good start, but there seem to be some places you haven't removed the ENROLLMENT_CODE_SWITCH, (like https://github.com/edx/ecommerce/blob/master/ecommerce/courses/views.py#L39), so please make sure you have found all references.

saleem-latif:
@brittneyexline do we want to remove all occurrences of `ENROLLMENT_CODE_SWITCH` everywhere from ecommerce? because I thought we just wanted to remove them from course creation flow. 
In my opinion (in case we want to remove it everywhere from ecommerce) it would be better to first enable `ENROLLMENT_CODE_SWITCH` the switch and test if everything works as expected as this would have the same effect as removing its occurrences. Thoughts?

brittneyexline:
@saleem-latif i think it would be good if you outlined what logic this switch gates so we can assess what to do. it wasn't clear to me why other references were excluded from this PR, so it seems like we need that information before we make a decision about the path for removing this flag.

saleem-latif:
@brittneyexline here are the uses of the `ENROLLMENT_CODE_SWITCH` and `SiteConfiguration.enable_enrollment_codes`.

1. https://github.com/edx/ecommerce/blob/master/ecommerce/courses/views.py#L38-L40
Here the switch decides to display 'bulk enrollment checkbox' on course creation page. If `ENROLLMENT_CODE_SWITCH` is active than 'bulk enrollment checkboxes' will be displayed but inactive so user can not select any. If both  `ENROLLMENT_CODE_SWITCH` and `SiteConfiguration.enable_enrollment_codes` and  'bulk enrollment checkboxes' will be displayed and clickable.
2. One occurrence that is removed in this PR is from Publisher serializer that contains the logic to add/updated courses in ecommerce via REST API.
3. Other occurrences are in the test cases [here](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/v2/tests/views/test_publication.py#L460), [here](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/tests/test_views.py#L415), [here](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/catalogue/tests/mixins.py#L152), [here](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/checkout/tests/test_signals.py#L292) and [here](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/tests/test_utils.py#L85),

brittneyexline:
@saleem-latif if the main other use is displaying the option in the UI, then I think it makes sense to remove that code so that it is clear that there is no longer an option to not create the enrollment codes. keep in mind that the feature switch that we'll ultimately need for this work will really be in making these codes available for purchase in the UI, their existence on their own shouldn't impact much.

however, before we merge we should get an answer on WL-1457 to make sure we aren't going to exacerbate an existing bug.

brittneyexline:
@saleem-latif is this deployed to the business sandbox? can you include the testing instructions?

saleem-latif:
@brittneyexline changes are deployed on business sandbox and I have updated PR description to contain testing instructions.

ecommerce/extensions/api/serializers.py
```diff
@@ -466,9 +468,7 @@ def save(self):
                 course.verification_deadline = course_verification_deadline
                 course.save()
 
-                create_enrollment_code = False
-                if waffle.switch_is_active(ENROLLMENT_CODE_SWITCH) and site.siteconfiguration.enable_enrollment_codes:
-                    create_enrollment_code = create_or_activate_enrollment_code
+                create_enrollment_code = create_or_activate_enrollment_code
```
brittneyexline: seems like you could get rid of this line and just use the create_or_activate_enrollment_code variable wherever create_enrollment_code was being used?

brittneyexline (commented):
seems like we should not leave the enable_enrollment_codes field on the site configuration object, if it does nothing. can you deprecate the field in accordance with https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/23003228/Everything+About+Database+Migrations?

maybe we should also add a migration to delete the waffle switch ENROLLMENT_CODE_SWITCH while we're at it?

saleem-latif:
@brittneyexline I have removed `ENROLLMENT_CODE_SWITCH` but I did not remove `enable_enrollment_codes` field from the model. This field is used by [create_or_update_site](https://github.com/edx/ecommerce/blob/master/ecommerce/core/management/commands/create_or_update_site.py#L177), It can be deprecated but I think it would be better to log a new ticket for this where we may also decide on whether to completely remove this or deprecate and if we go with deprecation then how to deprecate it as https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/23003228/Everything+About+Database+Migrations only gives proposals about field depreaction.

brittneyexline (approved):

PR #1663, opened 2018-02-28
Allow UUID to be optional for seats

mikix:
Only make the course UUID required if an entitlement is being created.

https://openedx.atlassian.net/browse/ENT-890


saleem-latif (approved):
👍 I deployed the changes on business sandbox (https://ecommerce-business.sandbox.edx.org) and I was able to add a new course successfully.

PR #1664, opened 2018-02-28
check for entitlement products in addition to course run products

McKenzieW:
Allow courses in addition to course runs to be included in Coupon Ranges' catalog queries

[LEARNER-4166](https://openedx.atlassian.net/browse/LEARNER-4166), [Ecommerce subtask](https://openedx.atlassian.net/browse/LEARNER-4359)

### DIFF
--- a/ecommerce/extensions/catalogue/tests/mixins.py
+++ b/ecommerce/extensions/catalogue/tests/mixins.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import logging
+from uuid import uuid4
 
 from oscar.core.loading import get_model
 from oscar.core.utils import slugify
@@ -73,6 +74,16 @@ class DiscoveryTestMixin(object):
 
         seat = course.create_or_update_seat(seat_type, id_verification, price, partner)
         return course, seat
+
+    def create_entitlement_product(self, course_uuid=None, certificate_type='verified'):
+        entitlement = factories.ProductFactory(
+            product_class=self.entitlement_product_class, stockrecords__partner=PartnerFactory(),
+            stockrecords__price_currency='USD'
+        )
+        entitlement.attr.UUID = course_uuid if course_uuid else uuid4()
+        entitlement.attr.certificate_type = certificate_type
+        entitlement.save()
+        return entitlement
 
     def _create_product_class(self, class_name, slug, attributes):
         """ Helper method for creating product classes.
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import logging
 import re
 
 from django.conf import settings
@@ -20,6 +21,8 @@ from ecommerce.core.utils import get_cache_key, log_message_and_raise_validation
 
 OFFER_PRIORITY_ENTERPRISE = 10
 OFFER_PRIORITY_VOUCHER = 20
+
+logger = logging.getLogger(__name__)
 
 
 class Benefit(AbstractBenefit):
@@ -42,6 +45,77 @@ class Benefit(AbstractBenefit):
             log_message_and_raise_validation_error('Percentage benefits require a product range')
         if self.value > 100:
             log_message_and_raise_validation_error('Percentage discount cannot be greater than 100')
+
+    def _filter_for_paid_course_products(self, lines, applicable_range):
+        """" Filters out products that aren't seats or entitlements or that don't have a paid certificate type. """
+        return [
+            line for line in lines
+            if line.product.is_seat_product or line.product.is_course_entitlement_product and
+            hasattr(line.product.attr, 'certificate_type') and
+            line.product.attr.certificate_type.lower() in applicable_range.course_seat_types
+        ]
+
+    def _identify_uncached_product_identifiers(self, lines, domain, partner_code, query):
+        """
+        Checks the cache to see if each line is in the catalog range specified by the given query
+        and tracks identifiers for which discovery service data is still needed.
+        """
+        course_run_ids = []
+        course_uuids = []
+        applicable_lines = lines
+        for line in applicable_lines:
+            product = line.product
+            cache_key = get_cache_key(
+                site_domain=domain,
+                partner_code=partner_code,
+                resource='catalog_query.contains',
+                course_id=product.course_id if product.is_seat_product else product.attr.UUID,
+                query=query
+            )
+            response = cache.get(cache_key)
+            if response is False:
+                applicable_lines.remove(line)
+            elif response is None:
+                if product.is_seat_product:
+                    course_run_ids.append({'id': product.course.id, 'cache_key': cache_key, 'line': line})
+                else:
+                    course_uuids.append({'id': product.attr.UUID, 'cache_key': cache_key, 'line': line})
+        return course_run_ids, course_uuids, applicable_lines
+
+    def get_applicable_lines(self, offer, basket, range=None):  # pylint: disable=redefined-builtin
+        applicable_range = range if range else self.range
+        if applicable_range and applicable_range.catalog_query is not None:
+            query = applicable_range.catalog_query
+            applicable_lines = self._filter_for_paid_course_products(basket.all_lines(), applicable_range)
+            site = basket.site
+            partner_code = site.siteconfiguration.partner.short_code
+            course_run_ids, course_uuids, applicable_lines = self._identify_uncached_product_identifiers(
+                applicable_lines, site.domain, partner_code, query
+            )
+            if course_run_ids or course_uuids:
+                # Hit Discovery Service to determine if remaining courses and runs are in the range.
+                try:
+                    response = site.siteconfiguration.discovery_api_client.catalog.query_contains.get(
+                        course_run_ids=','.join([metadata['id'] for metadata in course_run_ids]),
+                        course_uuids=','.join([metadata['id'] for metadata in course_uuids]),
+                        query=query,
+                        partner=partner_code
+                    )
+                except Exception as err:  # pylint: disable=bare-except
+                    logger.warning(
+                        '%s raised while attempting to contact Discovery Service for offer catalog_range data.', err
+                    )
+                    raise Exception('Failed to contact Discovery Service to retrieve offer catalog_range data.')
+
+                # Cache range-state individually for each course or run identifier and remove lines not in the range.
+                for metadata in course_run_ids + course_uuids:
+                    in_range = response[str(metadata['id'])]
+                    cache.set(metadata['cache_key'], in_range, settings.COURSES_API_CACHE_TIMEOUT)
+                    if not in_range:
+                        applicable_lines.remove(metadata['line'])
+            return applicable_lines
+        else:
+            return super(Benefit, self).get_applicable_lines(offer, basket, range=range)  # pylint: disable=bad-super-call
 
 
 class ConditionalOffer(AbstractConditionalOffer):
@@ -157,6 +231,10 @@ class ConditionalOffer(AbstractConditionalOffer):
         """
         if not self.is_email_valid(basket.owner.email):
             return False
+        if self.benefit.range and self.benefit.range.catalog_query:
+            # The condition is only satisfied if all basket lines are in the offer range
+            return len(self.benefit.get_applicable_lines(self, basket)) == basket.all_lines().count()
+
         return super(ConditionalOffer, self).is_condition_satisfied(basket)  # pylint: disable=bad-super-call
 
 
@@ -231,33 +309,6 @@ class Range(AbstractRange):
         if self.course_seat_types:
             validate_credit_seat_type(self.course_seat_types)
 
-    def run_catalog_query(self, product):
-        """
-        Retrieve the results from running the query contained in catalog_query field.
-        """
-        request = get_current_request()
-        partner_code = request.site.siteconfiguration.partner.short_code
-        cache_key = get_cache_key(
-            site_domain=request.site.domain,
-            partner_code=partner_code,
-            resource='course_runs.contains',
-            course_id=product.course_id,
-            query=self.catalog_query
-        )
-        response = cache.get(cache_key)
-        if not response:  # pragma: no cover
-            try:
-                response = request.site.siteconfiguration.discovery_api_client.course_runs.contains.get(
-                    query=self.catalog_query,
-                    course_run_ids=product.course_id,
-                    partner=partner_code
-                )
-                cache.set(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
-            except:  # pylint: disable=bare-except
-                raise Exception('Could not contact Discovery Service.')
-
-        return response
-
     def catalog_contains_product(self, product):
         """
         Retrieve the results from using the catalog contains endpoint for
@@ -298,13 +349,6 @@ class Range(AbstractRange):
                 # Range can have a catalog query and 'regular' products in it,
                 # therefor an OR is used to check for both possibilities.
                 return ((response['courses'][product.course_id]) or
-                        super(Range, self).contains_product(product))  # pylint: disable=bad-super-call
-        elif self.catalog_query and self.course_seat_types:
-            if product.attr.certificate_type.lower() in self.course_seat_types:  # pylint: disable=unsupported-membership-test
-                response = self.run_catalog_query(product)
-                # Range can have a catalog query and 'regular' products in it,
-                # therefor an OR is used to check for both possibilities.
-                return ((response['course_runs'][product.course_id]) or
                         super(Range, self).contains_product(product))  # pylint: disable=bad-super-call
         elif self.catalog:
             return (
--- a/ecommerce/extensions/offer/tests/test_models.py
+++ b/ecommerce/extensions/offer/tests/test_models.py
@@ -3,16 +3,12 @@ from __future__ import unicode_literals
 
 import ddt
 import httpretty
-import mock
-from django.core.cache import cache
 from django.core.exceptions import ValidationError
-from django.test import RequestFactory
 from oscar.core.loading import get_model
 from oscar.test import factories
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import SlumberBaseException
 
-from ecommerce.core.utils import get_cache_key
 from ecommerce.coupons.tests.mixins import CouponMixin, DiscoveryMockMixin
 from ecommerce.extensions.catalogue.tests.mixins import DiscoveryTestMixin
 from ecommerce.tests.testcases import TestCase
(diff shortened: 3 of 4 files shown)


McKenzieW:
@edx/helio

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,55 @@ def clean_percentage(self):
         if self.value > 100:
             log_message_and_raise_validation_error('Percentage discount cannot be greater than 100')
 
+    def get_applicable_lines(self, offer, basket, range=None):  # pylint: disable=redefined-builtin
+        query = None
+        if self.range:
```
Mjloturco: is this supposed to be different from the ranger param passed into this function?

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,55 @@ def clean_percentage(self):
...
+            query = self.range.catalog_query
+        if not query:
+            return super(Benefit, self).get_applicable_lines(offer, basket, range=range)  # pylint: disable=bad-super-call
+        else:
+            applicable_lines = [line for line in basket.all_lines()
```
Mjloturco: Is this you filtering for entitlement products? do products have a more explicit way of identifying type? i.e. just like line.product.type == 'Course Entitlement' or something?

ecommerce/extensions/offer/models.py
McKenzieW: Yes you can pass a range to the base function if you want to use that instead of the benefit's range. I suppose if the passed in range is what is being used to determine applicable lines, though, we should be using that for the logic that follows. Good catch

ecommerce/extensions/offer/models.py
```diff
@@ -157,6 +206,10 @@ def is_condition_satisfied(self, basket):
         """
         if not self.is_email_valid(basket.owner.email):
             return False
+        if self.benefit.range:
+            if self.benefit.range.catalog_query:
+                return len(self.benefit.get_applicable_lines(self, basket)) == basket.all_lines().count()
```
Mjloturco: This line is pretty hard to read and make sense of, if it could be wrapped in a helpfully named function that might be useful, otherwise atleast a comment would help.

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,56 @@ def clean_percentage(self):
...
+                    partner_code=partner_code,
+                    resource='catalog_query.contains',
+                    course_id=product.course_id if product.course_id else product.attr.UUID,
+                    query=query
+                )
+                response = cache.get(cache_key)
+                if response is False:
+                    applicable_lines.remove(line)
+                elif response is None:
+                    if product.course:
```
Mjloturco: This also seems like an implicit way of determining product type, may break heavily if we have products in the future that have similar attributes but different types.

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,56 @@ def clean_percentage(self):
...
+                try:
+                    response = site.siteconfiguration.discovery_api_client.catalog_query.contains.get(
+                        course_run_ids=','.join([course_run_id for (course_run_id, cache_key, line) in course_run_ids]),
+                        course_uuids=','.join([course_uuid for (course_uuid, cache_key, line) in course_uuids]),
+                        query=query,
+                        partner=partner_code
+                    )
+                except:  # pylint: disable=bare-except
+                    raise Exception('Could not contact Discovery Service.')
+            for identifier, cache_key, line in course_run_ids + course_uuids:
```
Mjloturco: should this be under the if conditional on line 78? or does this always need to be executed.

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,56 @@ def clean_percentage(self):
...
             log_message_and_raise_validation_error('Percentage discount cannot be greater than 100')
 
+    def get_applicable_lines(self, offer, basket, range=None):  # pylint: disable=redefined-builtin
+        query = None
+        applicable_range = range if range else self.range
+        if applicable_range:
+            query = applicable_range.catalog_query
+        if not query:
+            return super(Benefit, self).get_applicable_lines(offer, basket, range=range)  # pylint: disable=bad-super-call
+        else:
```
Mjloturco: This whole else statement is really big, is there any chance pieces of it could be broken into a seperate function?

Mjloturco (changes requested):
Logic seems good, I think it could be broken into some more readable functions, would be happy to identify possible functionization in person if you'd like (thats a new word).

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,73 @@ def clean_percentage(self):
         if self.value > 100:
             log_message_and_raise_validation_error('Percentage discount cannot be greater than 100')
 
+    def _filter_for_paid_course_products(self, lines, applicable_range):
+        """" Filters out products that aren't seats or entitlements or that don't have a paid certificate type. """
+        return [line for line in lines
+                if line.product.is_seat_product or line.product.is_course_entitlement_product and
+                hasattr(line.product.attr, 'certificate_type') and
+                line.product.attr.certificate_type.lower() in applicable_range.course_seat_types]
```
staubina: Slight code style thing here for this you might want to adjust the spacing to something more like 

```
return [
        line for line in lines 
        if line.product.is_seat_product or 
           line.product.is_course_entitlement_product and 
           hasattr(line.product.attr, 'certificate_type') and
           line.product.attr.certificate_type.lower() in applicable_range.course_seat_types
    ]
```

staubina: More in line with come of the topics here: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#breaking-long-lines

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,73 @@ def clean_percentage(self):
...
+            if response is False:
+                lines.remove(line)
+            elif response is None:
+                if product.is_seat_product:
+                    course_run_ids.append((product.course.id, cache_key, line))
+                else:
+                    course_uuids.append((product.attr.UUID, cache_key, line))
+        return course_run_ids, course_uuids, lines
+
+    def get_applicable_lines(self, offer, basket, range=None):  # pylint: disable=redefined-builtin
```
staubina: Can we avoid adding the pylint disable?  I need to look into what triggers it.

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,73 @@ def clean_percentage(self):
...
+        query = None
+        applicable_range = range if range else self.range
+        if applicable_range:
+            query = applicable_range.catalog_query
+        if not query:
+            return super(Benefit, self).get_applicable_lines(offer, basket, range=range)  # pylint: disable=bad-super-call
```
staubina: Please look at why this disable is required and see if it can be removed.

ecommerce/extensions/offer/models.py
staubina: The else if also probably not required because you are returning in the if, so you will not get to the code that is currently in the else.

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,73 @@ def clean_percentage(self):
...
+            if course_run_ids or course_uuids:
+                # Hit Discovery Service to determine if remaining courses and runs are in the range.
+                try:
+                    response = site.siteconfiguration.discovery_api_client.catalog_query.contains.get(
+                        course_run_ids=','.join([course_run_id for (course_run_id, cache_key, line) in course_run_ids]),
+                        course_uuids=','.join([course_uuid for (course_uuid, cache_key, line) in course_uuids]),
+                        query=query,
+                        partner=partner_code
+                    )
+                except:  # pylint: disable=bare-except
```
staubina: Please avoid bare exceptions if we can.  Are there certain exceptions we can catch here and interpret.

ecommerce/extensions/offer/models.py
```diff
@@ -157,6 +223,11 @@ def is_condition_satisfied(self, basket):
         """
         if not self.is_email_valid(basket.owner.email):
             return False
+        if self.benefit.range:
```
staubina: I think you can combine these two if statements 

```
if self.benefit.range and self.benefit.range.catalog_query:
```

staubina (changes requested):
Please take a look at the comments I posted.  I have not dug into the tests, I need to wrap a few things up before I leave today.  I will take a look first thing when I get back.

ecommerce/extensions/offer/models.py
McKenzieW: unfortunately python doesn't seem to break at the first false in boolean `and` calculations

ecommerce/extensions/offer/models.py
McKenzieW: This code is moved from it's previous location in the Range model

ecommerce/extensions/offer/models.py
Mjloturco: I think python does short circuit evaluate, is there something else going on ? https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not

ecommerce/extensions/catalogue/tests/mixins.py
```diff
@@ -74,6 +75,18 @@ def create_course_and_seat(
         seat = course.create_or_update_seat(seat_type, id_verification, price, partner)
         return course, seat
 
+    def create_entitlement_product(self, course_uuid=None, certificate_type='verified'):
+        course_uuid = course_uuid if course_uuid else uuid4()
```
jaebradley: Nit: unnecessary variable assignment? Looks like this is solely used on [line 85](https://github.com/edx/ecommerce/pull/1664/files#diff-439a8341d45fbef63fb11f0da08d17cbR85)? 

Maybe

```python
entitlement.attr.UUID = course_uuid if course_uuid else uuid4()
```

is more terse?

ecommerce/extensions/catalogue/tests/mixins.py
```diff
@@ -74,6 +75,18 @@ def create_course_and_seat(
...
+        partner = PartnerFactory()
```
jaebradley: Same thing here?

```python
entitlement = factories.ProductFactory(
  product_class=self.entitlement_product_class, 
  stockrecords__partner=PartnerFactory(),
  stockrecords__price_currency='USD'
)
```

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,75 @@ def clean_percentage(self):
...
+                lines.remove(line)
+            elif response is None:
+                if product.is_seat_product:
+                    course_run_ids.append((product.course.id, cache_key, line))
+                else:
+                    course_uuids.append((product.attr.UUID, cache_key, line))
+        return course_run_ids, course_uuids, lines
+
+    def get_applicable_lines(self, offer, basket, range=None):  # pylint: disable=redefined-builtin
+        query = None
```
jaebradley: It looks like query is `None` when `applicable_range` is `None` 

**or** if `applicable_range` is `not None` but `applicable_range.catalog_query` is `None`.

So maybe this could be simplified to

```python
applicable_range = range if range else self.range
# if applicable_range is None, this will return
# if applicable_range.catalog_query is None, this will return
# ...right?
if not applicable_range or not applicable_range.catalog_query
  return super(..etc)
query = applicable_range.catalog_query
applicable_lines = self._filter_for_paid_course_products(basket.all_lines(), applicable_range)
...etc
```

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,75 @@ def clean_percentage(self):
...
+    def _filter_for_paid_course_products(self, lines, applicable_range):
+        """" Filters out products that aren't seats or entitlements or that don't have a paid certificate type. """
+        return [
+            line for line in lines
+            if line.product.is_seat_product or line.product.is_course_entitlement_product and
+            hasattr(line.product.attr, 'certificate_type') and
+            line.product.attr.certificate_type.lower() in applicable_range.course_seat_types
+        ]
+
+    def _try_cache_for_contains_data(self, lines, domain, partner_code, query):
```
jaebradley: `_try_cache_for_contains_data` seems a little vague...maybe `identify_uncached_product_identifiers`?

alt-text

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,75 @@ def clean_percentage(self):
...
+            cache_key = get_cache_key(
+                site_domain=domain,
+                partner_code=partner_code,
+                resource='catalog_query.contains',
+                course_id=product.course_id if product.is_seat_product else product.attr.UUID,
+                query=query
+            )
+            response = cache.get(cache_key)
+            if response is False:
+                lines.remove(line)
```
jaebradley: Manipulating the underlying `lines` array might not be an obvious side-effect to the method caller. You could consider creating a `cached_lines` (there must exist a better name) array that you push to when `response is True` and then return that instead of `lines`.

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,75 @@ def clean_percentage(self):
...
+                    course_uuids=','.join([course_uuid for (course_uuid, cache_key, line) in course_uuids]),
+                    query=query,
+                    partner=partner_code
+                )
+            except:  # pylint: disable=bare-except
+                raise Exception('Could not contact Discovery Service.')
+
+            # Cache range-state individually for each course or run identifier and remove lines not in the range.
+            for identifier, cache_key, line in course_run_ids + course_uuids:
+                in_range = response[str(identifier)]
```
jaebradley: It's probably just my lack of context, but it's not terribly obvious what `in_range` is supposed to represent?

If it is and I'm just being ignorant, plz ign0r3.

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,75 @@ def clean_percentage(self):
...
+            # Hit Discovery Service to determine if remaining courses and runs are in the range.
+            try:
+                response = site.siteconfiguration.discovery_api_client.catalog_query.contains.get(
+                    course_run_ids=','.join([course_run_id for (course_run_id, cache_key, line) in course_run_ids]),
+                    course_uuids=','.join([course_uuid for (course_uuid, cache_key, line) in course_uuids]),
+                    query=query,
+                    partner=partner_code
+                )
+            except:  # pylint: disable=bare-except
+                raise Exception('Could not contact Discovery Service.')
```
jaebradley: Would it be useful to log / monitor anything about the Discovery Service error?

ecommerce/extensions/offer/models.py
McKenzieW: would `in_catalog_range` be clearer?

McKenzieW:
@jaebradley @staubina @Mjloturco Could you take another look and share any other comments you have?

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,73 @@ def clean_percentage(self):
...
+            site = basket.site
+            partner_code = site.siteconfiguration.partner.short_code
+            course_run_ids, course_uuids, applicable_lines = self._identify_uncached_product_identifiers(
+                applicable_lines, site.domain, partner_code, query
+            )
+            if course_run_ids or course_uuids:
+                # Hit Discovery Service to determine if remaining courses and runs are in the range.
+                try:
+                    response = site.siteconfiguration.discovery_api_client.catalog_query.contains.get(
+                        course_run_ids=','.join([course_run_id for (course_run_id, cache_key, line) in course_run_ids]),
```
jaebradley: Nit: The verbosity here is valuable but if you were looking for something more terse you could also do

```python
','.join([metadata[0] for metadata in course_run_ids])
```

image

If you used `dict`s instead of `tuple`s you could use the key instead of the index

```python
','.join([metadata['id'] for metadata in course_run_ids])
```

Not a big deal either way.

McKenzieW: I definitely like the idea of using a dict to preserve the detail from the verbosity while also being a bit more terse

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,73 @@ def clean_percentage(self):
...
+                        course_uuids=','.join([course_uuid for (course_uuid, cache_key, line) in course_uuids]),
+                        query=query,
+                        partner=partner_code
+                    )
+                except:  # pylint: disable=bare-except
+                    raise Exception('Could not contact Discovery Service.')
```
jaebradley: I know this is copied code and I know this error message is only used in one place across all edX repos

image

but should there be a slightly more descriptive error message like `Failed to fetch course information from the Discovery Service when calculating applicable lines`? 

Also, I think it would be helpful to at least log the Discovery exception since I think it's going to get lost in the stack trace when you raise the blank `Exception`.

[Example](https://repl.it/repls/RottenWateryCylinders)

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +42,73 @@ def clean_percentage(self):
...
+                resource='catalog_query.contains',
+                course_id=product.course_id if product.is_seat_product else product.attr.UUID,
+                query=query
+            )
+            response = cache.get(cache_key)
+            if response is False:
+                lines.remove(line)
+            elif response is None:
+                if product.is_seat_product:
+                    course_run_ids.append((product.course.id, cache_key, line))
```
jaebradley: This needs to be a `dict` as well?

McKenzieW: it sure does. Thanks for catching that

ecommerce/extensions/offer/models.py
McKenzieW: Yeah that makes sense. I'll add a bit more details to the error message and log the failure.

jaebradley: I don't know why that blank `Exception` exists but another thing to consider is whether or not a more specific `Exception` class is more useful.

[This is a good explanation of various error-handling strategies](https://stackoverflow.com/a/24065533/5225575).

ecommerce/extensions/offer/models.py
staubina: Per our conversation @McKenzieW Please take another look here.

staubina (changes requested):
The change to the conditional should do the trick for me. Along with the changes recommended by others.

staubina (approved):
:shipit:

staubina:
Just a quick reminder to please squash commits before merging.

ecommerce/extensions/offer/models.py
```diff
@@ -43,6 +46,76 @@ def clean_percentage(self):
...
+            applicable_lines = self._filter_for_paid_course_products(basket.all_lines(), applicable_range)
+            site = basket.site
+            partner_code = site.siteconfiguration.partner.short_code
+            course_run_ids, course_uuids, applicable_lines = self._identify_uncached_product_identifiers(
+                applicable_lines, site.domain, partner_code, query
+            )
+            if course_run_ids or course_uuids:
+                # Hit Discovery Service to determine if remaining courses and runs are in the range.
+                try:
+                    response = site.siteconfiguration.discovery_api_client.catalog_query.contains.get(
```
Mjloturco: Sorry I didn't think of this earlier, does discovery_api_client.catalog_query not do some smart looking up of results against the cache? it seems like '_identify_uncached_product_identifiers' might be something that the catalog query should be handling internally. thoughts?

Mjloturco (commented):
Just one last comment, Also I believe merging this is blocked until the discovery side is in place, is that correct?

ecommerce/extensions/offer/models.py
McKenzieW: As far as I know the rest of our API hits (in ecommerce at least) are preceded by an explicit cache check, suggesting that the client doesn't have a smart lookup. For this endpoint in particular it would be even less safe to assume that the smart lookup happens, because it would be dependent, I believe, on the request body being the same (i.e the catalog_query as well as the list of run ids and course uuids). Because we want to be able to cache data about the individual courses and runs so that future requests are more likely to hit the cache rather than pinging discovery, I don't believe relying on a smart lookup would be sufficient in this case.

McKenzieW:
@Mjloturco that is definitely right. We absolutely don't want to make our coupon logic dependent on an API endpoint that does not yet exist.

ecommerce/extensions/offer/models.py
Mjloturco: Cool if thats our general practice for now I say lets :shipit: (when unblocked). Might be worth an architecture chat at somepoint, It seems like in general dealing with caching should be handled in the api_client, which would help us ensure we are always using caching on endpoints when available and would abstract a lot of this work out of the individual views.

McKenzieW: Yeah, very good point. Moving the caching work into the API client would be incredible.

Mjloturco (approved):

ecommerce/extensions/offer/models.py
Mjloturco: would you mind making a ticket for that as a techdebt/ecommerce improvement , and tagging this as an example of where it would be helpful?

ecommerce/extensions/offer/tests/test_models.py
```diff
@@ -396,6 +337,36 @@ def test_condition_not_satisfied(self):
...
+    def test_is_range_condition_satisfied(self):
+        """
+        Verify that a basket satisfies a condition only when all of its products are in its range's catalog queryset.
+        """
+        valid_user_email = 'valid@{domain}'.format(domain=self.valid_sub_domain)
+        basket = factories.BasketFactory(site=self.site, owner=factories.UserFactory(email=valid_user_email))
+        product = self.create_entitlement_product()
+        another_product = self.create_entitlement_product()
+
+        _range = factories.RangeFactory()
```
staubina, line 349: Small nit, Why are we naming range starting with an underscore?

ecommerce/extensions/offer/models.py
staubina: Modifying an array in a method can be a dangerous habit as the caller of the method might not be expecting it to change in size.  I second @jaebradley here. We should make a copy and return the copy.

staubina (changes requested):
Some minor comments and a requested change, overall looks good.

ecommerce/extensions/offer/tests/test_models.py
McKenzieW: mostly just because range is an existing keyword, but I can change it

McKenzieW: It does, however, follow the convention that already existed for the other tests in the file, so for consistency I'd either like to remove the underscore from all instances of the variable `_range` or keep them on my newly added instances.

staubina: Ok, if it is consistent that is fine with me.

staubina: I do not think it is hurting anything.

staubina (approved):
Squash and :shipit:

PR #1666, opened 2018-03-05
make JS requirements as part of static make process

iloveagent57:
If e.g. bootstrap isn't currently installed, it's gonna be tough to run `make static`.


mikix (approved):
Makes sense.

PR #1668, opened 2018-03-07
Updated travis validate translations

Ayub-Khan:
Updated travis to use the i18n-tools validate
to validate translations in pullrequests.
LEARNER-2556

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,8 @@ install:
 
 script:
   - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/docs && make html'
-  - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ && DJANGO_SETTINGS_MODULE=ecommerce.settings.test make validate_translations'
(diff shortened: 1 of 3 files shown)


amangano-edx:
@Ayub-Khan have you validated that the build actually breaks when bad translations are committed? If not, try creating a test branch from this one and committing a translation with mismatched HTML-tags. If the build fails due to the `i18n_tool validate` call, then the changes here should be fine.

.travis.yml
```diff
@@ -25,7 +25,7 @@ install:
 
 script:
   - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/docs && make html'
-  - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ && DJANGO_SETTINGS_MODULE=ecommerce.settings.test make validate_translations'
+  - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ecommerce && DJANGO_SETTINGS_MODULE=ecommerce.settings.test i18n_tool validate'
```
amangano-edx: Just want to double check that `ecommerce.settings.test` is the correct settings module to use for this.

amangano-edx:
Also, please confirm that the correct version of edx-i18n-tools is installed (>= 0.4.0) in the travis environment.

Ayub-Khan:
Following pr will fix the translation issues (https://github.com/edx/ecommerce/pull/1697)

ahsan-ul-haq (approved):

PR #1670, opened 2018-03-08
don't access course id unless the product is a seat

McKenzieW:
Saw NoneType errors in stage when I tried to apply new coupon I made to an entitlement product

[LEARNER-4166](https://openedx.atlassian.net/browse/LEARNER-4166), [Ecommerce](https://openedx.atlassian.net/browse/LEARNER-4359)

### DIFF
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -64,22 +64,25 @@ class Benefit(AbstractBenefit):
         course_uuids = []
         applicable_lines = lines
         for line in applicable_lines:
-            product = line.product
+            if line.product.is_seat_product:
+                product_id = line.product.course.id
+            else:  # All lines passed to this method should either have a seat or an entitlement product
+                product_id = line.product.attr.UUID
             cache_key = get_cache_key(
                 site_domain=domain,
                 partner_code=partner_code,
                 resource='catalog_query.contains',
-                course_id=product.course_id if product.is_seat_product else product.attr.UUID,
+                course_id=product_id,
                 query=query
             )
             response = cache.get(cache_key)
             if response is False:
                 applicable_lines.remove(line)
             elif response is None:
-                if product.is_seat_product:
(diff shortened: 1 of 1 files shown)


staubina (approved):

ecommerce/extensions/offer/models.py
```diff
@@ -64,22 +64,25 @@ def _identify_uncached_product_identifiers(self, lines, domain, partner_code, qu
         course_uuids = []
         applicable_lines = lines
         for line in applicable_lines:
-            product = line.product
+            if line.product.is_seat_product:
+                product_id = line.product.course.id
+            else:
```
Mjloturco: maybe worth giving this an explicit elif? again trying to think forward for if we have additional product types at somepoint.

Mjloturco (changes requested):
Might want to add a bit more defensiveness if we are in here anyway.

ecommerce/extensions/offer/models.py
McKenzieW: good call

Mjloturco:
Nice, do we want to do something in the case that it is not a Seat or an Entitlement? we could throw an error or log or maybe doing nothing is the correct call for that case, I'm unsure.

McKenzieW:
@Mjloturco yes we definitely do, the product should be removed from applicable lines. Such a keen eye you have.

McKenzieW:
So the `_filter_for_paid_course_products` method already removes any products that aren't seats or or course entitlements from the applicable lines. @Mjloturco do you think it's still important to handle products falling outside of that category in `_identify_uncached_product_identifiers`, or will it be sufficient to just more verbosely describe the `else` block with an `elif...`?

Mjloturco (approved):

Mjloturco:
@McKenzieW nice, I didn't have the context that other products would already be filtered out, Looks good to me (once you can convince codecov to stop complaining)

PR #1671, opened 2018-03-09
stop using deprecated basket single item view

tasawernawaz:
[LEARNER-4428](https://openedx.atlassian.net/browse/LEARNER-4428)

* The [`BasketSingleItemView`](https://github.com/edx/ecommerce/blob/4cfb8c902b84af23584f7a535a421781fb5decd7/ecommerce/extensions/basket/views.py#L46) is deprecated and causing database deadlock in ecommerce.
* Remove [`BasketSingleItemView`](https://github.com/edx/ecommerce/blob/4cfb8c902b84af23584f7a535a421781fb5decd7/ecommerce/extensions/basket/views.py#L46) from code base

**Sandbox:** https://tasawernawaz.sandbox.edx.org

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -14,7 +14,6 @@ from opaque_keys.edx.keys import CourseKey
 from oscar.apps.basket.views import VoucherAddView as BaseVoucherAddView
 from oscar.apps.basket.views import VoucherRemoveView as BaseVoucherRemoveView
 from oscar.apps.basket.views import *  # pylint: disable=wildcard-import, unused-wildcard-import
-from oscar.core.decorators import deprecated
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import SlumberBaseException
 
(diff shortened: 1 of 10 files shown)


tasawernawaz:
@McKenzieW Please review this PR as well, this is for removing usage of deprecated view from ecommerce.

McKenzieW (approved):

tasawernawaz:
@McKenzieW Do you know why codcov is failing on the PR ? I don't think there is any change which require test.

McKenzieW:
@tasawernawaz it looks like because the line you updated was in close vicinity to / acted upon by a couple of other lines of code that were already uncovered it's identifying those uncovered lines in the diff
https://codecov.io/gh/edx/ecommerce/compare/bb6f52ac1879fa3027509f259228dd5ce7fe02aa...8c79e6f6100952ce96ea4c5f39bbcdf99ecd3e1f/diff

tasawernawaz:
@McKenzieW Could you please review this again, I have rmeoved the unused class [`BasketSingleItemView`](https://github.com/edx/ecommerce/blob/4cfb8c902b84af23584f7a535a421781fb5decd7/ecommerce/extensions/basket/views.py#L46).

McKenzieW (approved):

tasawernawaz:
@edx/learner-spartans Please review it.

Ayub-Khan:
I have some concerns that why all we removing all code in this pull request should not we follow a strategy of adding a waffle flag or config model to stop using the single-item basket. and then if its working file remove all the related code. If we move this kinda change to prod and it fails it will cause a lot of problems for learners. Plus It will be a lot to revert if it causes problems. second. with waffle flag or config model aproch we can simply turn the change off if its not working.

Ayub-Khan:
As we have already removed the usage from platform i think we no longer have any need to use the waffle flag.

ecommerce/extensions/basket/views.py
```diff
@@ -42,61 +41,6 @@
...
-                request,
-                [product],
-                [sku],
-                'basket:single-item'
-            )
-            if code_redemption_redirect:
-                return code_redemption_redirect
-
-        # If the product isn't available then there's no reason to continue with the basket addition
-        purchase_info = request.strategy.fetch_for_product(product)
```
Ayub-Khan, line 84: We want to include this check for all products in multiple item basket.

Ayub-Khan (changes requested):

ecommerce/extensions/basket/views.py
Ayub-Khan: You will need to check all the products availability. if some product is not available drop that product and start payment process for remaining. if all products are not available log error and return HttpResponseBadRequest(error message). 
Possible Cases could be:
1 unavailable product (Log error and return bad request)
1 available product (No action needed)
many available and many unavailable procucts (Log warnings for unavailable products and move forward with available products)
all available products  (no action needed)
all unavailable products (log error and return bad request)

Ayub-Khan:
also have a look at prepare_basket method you might have to update that too .

ecommerce/extensions/basket/views.py
tasawernawaz: Updated

staubina:
I think we can ignore this codecov failure.  Once you get approval @tasawernawaz from @Ayub-Khan I think we can merge this PR.

staubina:
It looks like the Codecov is reporting an overall decrease in the % of covered code because you are deleting code, but your changes are 100% covered.

McKenzieW:
Although it is not a newly uncovered line of code, it might be useful to add a test for [this line](https://codecov.io/gh/edx/ecommerce/compare/5ff8e61795627e2f1700a2aad24f4b4281364273...74ded71bafa5099fdc8e739fd2387cbfc726d256/diff#D7-21) so as to increase coverage and simplify the process of merging this PR.

ecommerce/extensions/basket/views.py
```diff
@@ -42,61 +41,6 @@
...
-        try:
-            prepare_basket(request, [product], voucher)
-        except AlreadyPlacedOrderException:
-            msg = _('You have already purchased {course} seat.').format(course=product.title)
-            return render(request, 'edx/error.html', {'error': msg})
-        url = add_utm_params_to_url(reverse('basket:summary'), self.request.GET.items())
-        return HttpResponseRedirect(url, status=303)
-
-
 class BasketMultipleItemsView(View):
```
Ayub-Khan: As this class now handles single items baskets too we need to update its name and description.

ecommerce/extensions/basket/views.py
```diff
@@ -133,8 +77,21 @@ def get(self, request):
             if code_redemption_redirect:
                 return code_redemption_redirect
 
+        # check availability of products
```
Ayub-Khan, line 80: *Optional* double check this functionality using sandbox as we have recently seen some students placing orders for expired course seats.
If you think it is out of the scope just ignore this we can come back to it after this is done.

tasawernawaz: @Ayub-Khan Do we have ticket to investigate "orders for expired course seats." ? If not please create one, after this got merged. I will look into that.

tasawernawaz:
@Ayub-Khan I have added jasmine test to fix code coverage issue. Also refactored as you requested. Please review again when you have time.

ecommerce/extensions/basket/views.py
Ayub-Khan: ok

Ayub-Khan (approved):

Ayub-Khan:
:+1:

PR #1672, opened 2018-03-09
ENT-898: Fixed Bulk enrollment code creation.

saleem-latif:
**JIRA Ticket:** [ENT-898](https://openedx.atlassian.net/browse/ENT-898)
__Description:__
This PR fixes the issue where bulk enrollment codes were being created without correct parameters.
There were two problems present with enrollment codes

1. Coupon code had 10,000 redemptions created for enrollment codes
2. Single course/multi-course and cert type fields were both blank for enrollment codes

We could not reproduce `10,000 redemptions` part as all new coupons had only one redemption. and with the merge of https://github.com/edx/ecommerce/pull/1655 I think this problems mostly likely already resolved.
Whenever, a enrollment is created via CAT tool or some other way it will have an associated range, benefit, coupon code product etc. This range contains fields like catalog, catalog query, seat types etc. that define the course and seat types for the coupon application. Previously, the Range created for enrollment codes did not have any of these fields populated which was causing problems. This PR fixes that issue by assigning a catalog for the course whose enrollment codes are being created.

**Testing Instructions:**

1. Create a could with bulk enrollment codes enabled, https://mitxpro-sb.sandbox.edx.org/courses/course-v1:MITxPRO+MPTC101+2018_T1/about can be used for testing.
2. Click on "Purchase for a Group", you will be redirected to ecommerce basket page where you will need to add enrollment codes quantity and payment information.
3. After submitting the form, make sure there is a new Range product with catalog field populated. Look for the newest Range product.

### DIFF
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -25,6 +25,7 @@ from ecommerce.courses.models import Course
 from ecommerce.courses.utils import mode_for_product
 from ecommerce.enterprise.utils import get_or_create_enterprise_customer_user
 from ecommerce.extensions.analytics.utils import audit_log, parse_tracking_context
+from ecommerce.extensions.api.v2.views.coupons import CouponViewSet
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.fulfillment.status import LINE
 from ecommerce.extensions.voucher.models import OrderLineVouchers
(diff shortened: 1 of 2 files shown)


zubair-arbi (approved):
Changes looks good to me 👍 
Please update the test for `fulfill_product` to verify that range object now contains coupon catalog.

brittneyexline (commented):
As far as I can tell, this seems like a reasonable thing to do. It isn't apparent from reading the code alone how this addresses the bug - if you could provide a little more context in the description on how this fixes the bug, as well as verification instructions, that would be helpful.

saleem-latif:
@brittneyexline I have updated the PR description, please take a look.

brittneyexline (approved):

PR #1674, opened 2018-03-12
fix return type of get_applicable_lines

McKenzieW:
Get applicable lines is expected to return a list of (price, line) tuples, not just a list of lines.

[LEARNER-4166](https://openedx.atlassian.net/browse/LEARNER-4166), [Ecommerce subtask](https://openedx.atlassian.net/browse/LEARNER-4166)

### DIFF
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -116,7 +116,7 @@ class Benefit(AbstractBenefit):
                     cache.set(metadata['cache_key'], in_range, settings.COURSES_API_CACHE_TIMEOUT)
                     if not in_range:
                         applicable_lines.remove(metadata['line'])
-            return applicable_lines
+            return [(line.product.stockrecords.first().price_excl_tax, line) for line in applicable_lines]
         else:
             return super(Benefit, self).get_applicable_lines(offer, basket, range=range)  # pylint: disable=bad-super-call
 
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/offer/models.py
```diff
@@ -116,7 +116,7 @@ def get_applicable_lines(self, offer, basket, range=None):  # pylint: disable=re
                     cache.set(metadata['cache_key'], in_range, settings.COURSES_API_CACHE_TIMEOUT)
                     if not in_range:
                         applicable_lines.remove(metadata['line'])
-            return applicable_lines
+            return [(line.product.stockrecords.first().price_excl_tax, line) for line in applicable_lines]
```
Mjloturco, line 119: Nit: purely style/readability, it might be more readable to shift ```line.product.stockrecords.first().price_excl_tax``` into its own line and declare it as a variable.

```python
price = line.product.stockrecords.first().price_excl_tax
return [(price, line) for line in applicable_lines]

Mjloturco (approved):
ignore my comment :shipit:  👍

ecommerce/extensions/offer/models.py
Mjloturco: oh wait,
that wouldn't work, never mind!

staubina (approved):
:shipit: Seems like a straightforward change.

PR #1676, opened 2018-03-14
Skip receipt page if placing free order

xitij2000:
This PR changes the behaviour of ecommerce to skip the receipt page when a coupon redemption doesn't require any payment. When placing a free order the receipt page will be skipped and users will be directed to the LMS dashboard page on redemption instead. 

**Sandbox URL**: TBD - sandbox is being provisioned.

**Merge deadline**: None

**Testing instructions**:
The regular test suite can be used for testing this change by running the tests for  ``ecommerce.extensions.checkout``

**Reviewers**
- [ ] @kaizoku
- [ ] edX reviewer[s] TBD

### DIFF
--- a/ecommerce/extensions/checkout/views.py
+++ b/ecommerce/extensions/checkout/views.py
@@ -21,7 +21,6 @@ from ecommerce.core.url_utils import (
 from ecommerce.enterprise.utils import has_enterprise_offer
 from ecommerce.extensions.checkout.exceptions import BasketNotFreeError
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
-from ecommerce.extensions.checkout.utils import get_receipt_page_url
 
 Applicator = get_class('offer.applicator', 'Applicator')
 Basket = get_model('basket', 'Basket')
@@ -52,7 +51,7 @@ class FreeCheckoutView(EdxOrderPlacementMixin, RedirectView):
     """ View to handle free checkouts.
 
     Retrieves the user's basket and checks to see if the basket is free in which case
-    the user is redirected to the receipt page. Otherwise the user is redirected back
+    the user is redirected to the LMS dashboard. Otherwise the user is redirected back
     to the basket summary page.
     """
 
@@ -90,11 +89,7 @@ class FreeCheckoutView(EdxOrderPlacementMixin, RedirectView):
                     course_run_id = order.lines.all()[:1].get().product.course.id
                     url = get_lms_courseware_url(course_run_id)
             else:
-                receipt_path = get_receipt_page_url(
-                    order_number=order.number,
-                    site_configuration=order.site.siteconfiguration
-                )
-                url = site.siteconfiguration.build_lms_url(receipt_path)
+                url = get_lms_dashboard_url()
         else:
             # If a user's basket is empty redirect the user to the basket summary
             # page which displays the appropriate message for empty baskets.
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -12,7 +12,7 @@ from factory.fuzzy import FuzzyText
 from oscar.core.loading import get_class, get_model
 from oscar.test.factories import OrderFactory, OrderLineFactory, RangeFactory, VoucherFactory
 
-from ecommerce.core.url_utils import get_lms_url
+from ecommerce.core.url_utils import get_lms_dashboard_url, get_lms_url
 from ecommerce.coupons.tests.mixins import CouponMixin, DiscoveryMockMixin
 from ecommerce.coupons.views import voucher_is_valid
 from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
@@ -430,12 +430,15 @@ class CouponRedeemViewTests(CouponMixin, DiscoveryTestMixin, LmsApiMockMixin, En
 
     @httpretty.activate
     def test_basket_redirect_enrollment_code(self):
-        """ Verify the view redirects to the receipt page when an enrollment code is provided. """
+        """ Verify the view redirects to the LMS dashboard when an enrollment code is provided. """
         code = self.create_coupon_and_get_code(benefit_value=100, code='', catalog=self.catalog)
         self.mock_account_api(self.request, self.user.username, data={'is_active': True})
         self.mock_access_token_response()
 
-        self.assert_redirects_to_receipt_page(code=code)
+        self.assert_redemption_page_redirects(
+            get_lms_dashboard_url(),
+            code=code,
+        )
 
     @httpretty.activate
     @mock.patch.object(EdxOrderPlacementMixin, 'place_free_order')
@@ -444,7 +447,7 @@ class CouponRedeemViewTests(CouponMixin, DiscoveryTestMixin, LmsApiMockMixin, En
         code = self.create_coupon_and_get_code(benefit_value=100, code='', catalog=self.catalog)
         self.mock_account_api(self.request, self.user.username, data={'is_active': True})
         self.mock_access_token_response()
-        place_free_order.return_value = Exception
+        place_free_order.side_effect = Exception
 
         with mock.patch('ecommerce.coupons.views.logger.exception') as mock_logger:
             self.assert_redemption_page_redirects(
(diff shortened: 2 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @xitij2000! I've created [OSPR-2295](https://openedx.atlassian.net/browse/OSPR-2295) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

If you like, you can add yourself to the [AUTHORS](https://github.com/edx/ecommerce/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

nedbat:
@edx/learner Is this the right way to do this?

nedbat:
@xitij2000 Ping me when this is ready for review.

Ayub-Khan:
a better way to do that would be using waffle flag. As in ecommerce all the such things are done using waffle flags.

xitij2000:
@Ayub-Khan Thanks for pointing that out! I have updated the code to use a waffle flag.

nedbat:
@xitij2000 is this still a work-in-progress, or is it ready for review?

xitij2000:
@nedbat  Sorry for the delay! I will try to get this finished today and get it reviewed internally and ping you as soon as possible.

kaizoku:
:+1:

- [✓] I tested this: (deployed in devstack, enabled waffle switch, and checked out 100% discount coupon)
- [✓] I read through the code

xitij2000:
@nedbat Sorry for the delay! This is ready for review now.

mduboseedx:
@edx/learner Could this please be reviewed?

mikix:
I know we told you that waffle flags are decent, but actually, we are trying to reduce our dependence on them (they are not as easy to document and make our code more spaghetti like than we want).

Instead of waffle flags, we are hoping to eventually move optional features to plugins (that can be installed or not) or maybe config values if we do have to have tweakability.

However in this case, my engineering brain thinks maybe this is just a change we always want? Though I admit I don't have much context here. I'll kick this to the Product team to see if they see value in the old way (receipt for free orders). <- @edx/learner-product

xitij2000:
@mikix Thaks for the additional context! 
If it makes more sense to skip the receipt for free orders then perhapes the condition of this flag can be reversed. So we can have a ``SHOW_RECEIPT_FOR_FREE_ORDERS`` flag/setting. 
Either way seems fine to me. I look forward to hearing from the Product team.

mikix:
@xitij2000 to be clear, I was proposing dropping the flag altogether and just hard-coding this behavior.

marcotuts:
One question here I have is what is the current behavior with & without the flag for showing the payment in the order history tab of the account settings area. 

Thanks for flagging the product questions here. I will defer to Deen on this one, I can connect with him offline and update this flag with a question. Independent of his feedback, an answer to the question above would be helpful context. Thanks!

xitij2000:
@marcotuts  Sorry for the incredibly late reply, but I think after this change you will still be able to see details in order history.

marcotuts:
No flag and defaulting to skipping receipt for free orders makes sense to me. This can move to engineering review next.

mikix (approved):
I didn't run the PR, but the code looks good to me. Looks like you and kaizoku have run it though.

Once tests pass, this should be fine. Thank you!

xitij2000:
@mikix Thanks! I have rebased the code on the latest master, and the tests are passing now. I also tested it manually to see if it still works after the rebase.

mikix:
@robrap can someone from the ecommerce guild look at this & push if it's good?

ecommerce/extensions/checkout/views.py
```diff
@@ -90,11 +89,7 @@ def get_redirect_url(self, *args, **kwargs):
                     course_run_id = order.lines.all()[:1].get().product.course.id
                     url = get_lms_courseware_url(course_run_id)
             else:
-                receipt_path = get_receipt_page_url(
-                    order_number=order.number,
-                    site_configuration=order.site.siteconfiguration
-                )
-                url = site.siteconfiguration.build_lms_url(receipt_path)
+                url = get_lms_dashboard_url()
```
robrap, line 92: The docstring for this class needs to be updated appropriately.  

Also, is it safe to assume that this method is only called once it is determined the basket was free?  If so, maybe add the following to this method:
```
Returns the redirect url for the free basket.
```

robrap (changes requested):
Please make the minor docstring changes.  Thank you.  Also, ecommerce can only be merged if it is up to date, so please squash/rebase as well with your change.

ecommerce/extensions/checkout/views.py
xitij2000: @robrap Earlier in this method it checks if the basket total is 0, and raises an error if it isn't, so by the time we reach this code I think it's safe to assume it's a free order. 

I've updated the docstring for the class, however since the method is called ``get_redirect_url`` so not sure adding that docstring will really help!

robrap: @xitij2000: I don't know why I didn't see that code before.  I thought I looked at the whole method.  You can ignore this comment.  Thanks.

robrap (approved):

robrap:
@xitij2000: You can squash and rebase.  However, I don't plan on merging until Thursday morning after the holiday here, so you may want to wait and rebase then.

Also, have you ensured that there are no side-effects on the receipt page other than displaying the receipt?  Tracking log events or anything else?  Just want to ensure that there are no unintended consequences.  Thanks.

xitij2000:
@robrap  I had a closer look at the replaced code, and it's just code for building URLs, nothing that would create any kind of log output of any kind. 

I've also rebased and squashed the commits.

openedx-webhooks:
@xitij2000 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/ecommerce&entry.867055334=https://github.com/edx/ecommerce/pull/1676&entry.1484655318=https://github.com/xitij2000&entry.752974735=2018-03-14+12:34&entry.1917517419=2018-07-06+13:02&entry.2133058324=Yes) so we can improve your experience in the future.

xitij2000:
@robrap  Thanks!

robrap:
@xitij2000: Unfortunately, I'm not sure if I will need to revert this or not.  We have some e2e tests that relied on the receipt page being there that are now failing, and not sure when someone will be able to pick up the work to fix them and/or if you would be able to.  More next week...

xitij2000:
@robrap Sure will try to help in whatever way I can.

robrap:
Looks like I will need to revert on Monday. Beyond the end-to-end tests failing, I think the user is getting to the dashboard with no message and no course, presumably beating out fulfillment.

xitij2000:
@robrap OK. Would it be more acceptable if I put this behaviour behind a waffle flag/switch or setting like it was originally?

robrap:
My personal opinion is that this requires some thought, but it’s not my call. As written, the user sometimes lands on the dashboard without the newly enrolled course being available and with no message. The user has to refresh the page for it to appear. This isn’t a great user experience. It seems that a consequence of the receipt page is adding enough delay to generally avoid this problem.

A more complete solution would be to show a success message that could tell the user what they need to do, or automates the appropriate reload, but as you could imagine this could get complicated, especially because it is across services. A related alternative is to embed this into the receipt page with a timer countdown that automatically redirects, and includes a link to manually go to dashboard before the timer. Probably there are other ideas as well. I just don’t think the current issue is good for users as-is, whether it is behind a flag or not.

xitij2000:
I see. That does sound like a more complex issue to solve! Thanks for all your help. I'll see if there is any way to take this forward in a way that works, or if it's even worth it.

robrap:
@xitij2000: I don't plan on putting any more time into this as-is.  If you want to consider approaches and discuss alternatives, let me know.  Thanks.

xitij2000:
@robrap I checked up on this a bit, and the main requirement of our client is that for users enrolling via an enrollment code should not be confronted with a price point + 100% discount. So simply hiding those details in case of a free enrollment would be sufficient. 

Would it be acceptable to add the ability to hide the pricepoint in case of free orders behind a flag?

robrap:
@marcotuts: Can you get the proper Product review for the above comment?  Again, it would be great to avoid a flag.

xitij2000:
I figured a flag would be best because I can understand a genuine need for showing the price in the receipt even for a 100% discount.

xitij2000:
@marcotuts Does the approach in my previous comment seem reasonable? 

Essentially, if a user is joining a course using a code that reduces the price to 0, then don't show them the price at all. This could behind a flag, or (as @robrap mentioned above) could be the new default behaviour. 

Would love to know if this approach is acceptable/feasible so we can start work on it.

PR #1678, opened 2018-03-20
Ensure single-use coupons are not applied to multi-product baskets

McKenzieW:
Add voucher usage verification to overridden `is_condition_satisfied` method in `ConditionalOffer` class to ensure that single-use coupons having catalog_queries to determine their applicable range cannot be applied to multi-product baskets.

[LEARNER-4166](https://openedx.atlassian.net/browse/LEARNER-4166), [Ecommerce subtask](https://openedx.atlassian.net/browse/LEARNER-4359)


Mjloturco (approved):

Mjloturco:
Nice

staubina (approved):

PR #1679, opened 2018-03-20
[REV-23] add course id to receipt page

MatthewPiatetsky:
I have the other necessary information set up to post the relevant data to the key-value store once the user hits the receipt page 
(key - course id, value - doesn't matter yet, but we may find a use for it later)  

The plan was that when a user tries to unenroll, we check if they are in the experiment for the course, and if they are, we allow audit mode refunds.

P.S. how do we want code reviews to work? Lately on the growth team we had the PR submitter pick one code reviewer to assign to review the PR and this was meant to be high priority for that specific person (this worked pretty fast). At one point in the past we tagged the whole team with an alias in the PR description. I'll include everyone until we've decided this.

### DIFF
--- a/ecommerce/templates/edx/checkout/receipt.html
+++ b/ecommerce/templates/edx/checkout/receipt.html
@@ -89,7 +89,9 @@
                 <dd class="course-description">
                   <span>{{ line.title }}</span>
                   {% if line.product.course %}
-                    <p>{{ line.product.course.id|course_organization }}</p>
+                    <p class="course-description-subtitle" data-course-id="{{ line.product.course.id }}">
+                      {{ line.product.course.id|course_organization }}
+                    </p>
                   {% endif %}
                 </dd>
                 <dt class="line-price sr">{% trans "Item Price:" %}</dt>


ecommerce/templates/edx/checkout/receipt.html
```diff
@@ -89,7 +89,9 @@ <h2>{% trans "Order Information" %}</h2>
                 <dd class="course-description">
                   <span>{{ line.title }}</span>
                   {% if line.product.course %}
-                    <p>{{ line.product.course.id|course_organization }}</p>
+                    <p class="course-description-subtitle" data-course-id="{{ line.product.course.id }}">
```
mulby, line 92: Not a blocker, but out of curiosity...

Is this always a course ID or can it be a catalog course ID for bundled purchases of entitlements?

mulby (approved):
LGTM

ecommerce/templates/edx/checkout/receipt.html
MatthewPiatetsky: I'm guessing you might not see it for entitlements, but I'm assuming we're going to exclude bundle purchases from the experiment, since they're part of separate flow and they just confuse the revenue metrics

McKenzieW: If the product type is an entitlement `line.product.course` will be null. We use the course uuid to identify the course associated with an entitlement product, which is accessed from the order line as follows `line.product.attr.UUID`.

Mjloturco: Line 91 should filter out entitlement products, so it should just be course seats.

Mjloturco:
Looks fine to me. Just for clarity, is this necessary because optimizely needs selectors in order to grab data from the page?

Mjloturco (approved):

MatthewPiatetsky:
I'm adding new data the ID is consumed by the org filter

PR #1680, opened 2018-03-21
Make ecommerce_sessionid cookie secure

ahsan-ul-haq:
LEARNER-4657
FYI @edx/learner-spartans

### DIFF
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -470,6 +470,7 @@ TEST_RUNNER = 'django.test.runner.DiscoverRunner'
 SESSION_COOKIE_NAME = 'ecommerce_sessionid'
 CSRF_COOKIE_NAME = 'ecommerce_csrftoken'
 LANGUAGE_COOKIE_NAME = 'ecommerce_language'
+SESSION_COOKIE_SECURE = False
 # END COOKIE CONFIGURATION
 
 


ecommerce/settings/base.py
```diff
@@ -470,6 +470,7 @@
 SESSION_COOKIE_NAME = 'ecommerce_sessionid'
 CSRF_COOKIE_NAME = 'ecommerce_csrftoken'
 LANGUAGE_COOKIE_NAME = 'ecommerce_language'
+SESSION_COOKIE_SECURE = True
```
robrap: Please update this to something like the following:
```
SESSION_COOKIE_SECURE = False  # Use True in Production to meet various ecommerce requirements.
```
NOTE: See the ticket for where and how to set to True.

ecommerce/settings/test.py
```diff
@@ -17,6 +17,8 @@
 
 TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
 
+SESSION_COOKIE_SECURE = False
```
robrap: Did it work without this?  If some tests broke, do we have an explanation, and are we sure we don't have actual bug?

It would be better if our tests could include this config change, otherwise it is untested, and you'll need to test manually.  Also, can you confirm how much manual testing you did?  Thank you.

robrap (commented):
Looks pretty simple and close.  Thank you.  Just some questions to ensure things are all good.

ecommerce/settings/test.py
robrap: This line can be deleted because we will continue to default to False.

ahsan-ul-haq:
@robrap made the suggested change

ecommerce/settings/production.py
```diff
@@ -84,3 +84,5 @@ def get_env_setting(setting):
 # END PAYMENT PROCESSOR OVERRIDES
 
 ENTERPRISE_API_URL = urljoin(ENTERPRISE_SERVICE_URL, 'api/v1/')
+
+SESSION_COOKIE_SECURE = True
```
robrap: @edx/devops: I'm not sure whether 'settings/production.py' would give the same results as the following recommendation, or if this change should be removed and we should this as it was done in edx-platform as described below? 

I'm copying in the comment from [LEARNER-4657](https://openedx.atlassian.net/browse/LEARNER-4657):

> I just spoke to devops. It looks like there will be changes needed in a bunch of repos to get this wired up properly.
> This includes configuration, edx-internals, edge-internals, and sandbox-internals.
> 
> Please see the following github search for [EDXAPP_SESSION_COOKIE_SECURE](https://github.com/search?q=org%3Aedx+EDXAPP_SESSION_COOKIE_SECURE&type=Code).

fredsmith (approved):

PR #1683, opened 2018-03-23
Switch to segment analytics-python from edx fork

BenjiLee:
@bderusha Is this the fork change you are looking for?

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -3,6 +3,7 @@ import hashlib
 import logging
 from urlparse import urljoin
 
+from analytics import Client as SegmentClient
 from dateutil.parser import parse
 from django.conf import settings
 from django.contrib.auth.models import AbstractUser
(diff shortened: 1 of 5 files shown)


feanil (approved):

feanil:
For reference: we're going to ship this along with https://github.com/edx/configuration/pull/4415 and https://github.com/edx/edx-internal/pull/332

BenjiLee:
There seems to be an isort issue on master? No idea why it's surfacing in this PR. 

The fix for that is here https://github.com/edx/ecommerce/pull/1684. Once merge, I will rebase this PR.

bderusha (approved):

bderusha:
@BenjiLee I believe the isort issue is because it's treated differently as a package install VS github install.  So isort wants it in a different group

bderusha:
:shipit: 👍

PR #1686, opened 2018-03-26
Update message when bulk purchasing just a single code

georgebabey:
Change messaging on the basket page: `Purchasing access just for yourself` -> `Purchasing just for yourself`

Minor change that came up during demo


asadiqbal08 (approved):
👍

PR #1688, opened 2018-03-27
Revert "Switch to segment analytics-python from edx fork"

robrap:
LEARNER-4675: This reverts commit 75909f6

Things have gotten worse in Production.  It isn't clear what combo of moving away from the fork and dropping use of gevent is the cause.  I'm not sure the sum total of changes in the fork that we lost.

This is just a straight up revert to get back to earlier bad steady-state.

The following is a graph of both healthy and unhealthy hosts from the ELB for ecommerce:
screen shot 2018-03-27 at 12 04 43 pm

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -3,7 +3,6 @@ import hashlib
 import logging
 from urlparse import urljoin
 
-from analytics import Client as SegmentClient
 from dateutil.parser import parse
 from django.conf import settings
 from django.contrib.auth.models import AbstractUser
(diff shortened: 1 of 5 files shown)


robrap:
@feanil: I'm not sure where the other changes were done for this work.  Could you help me locate the PR(s)?
FYI: @BenjiLee

robrap:
The build failed.  I didn't see the error.  Guessing it might be the isort issue again, so trying again with that change reverted.

BenjiLee:
@robrap 
https://github.com/edx/configuration/pull/4415 https://github.com/edx/edx-internal/pull/332

BenjiLee (approved):

BenjiLee:
https://github.com/edx/edx-internal/pull/339/files This is the PR we need to ship with https://github.com/edx/edx-internal/pull/339/files

PR #1689, opened 2018-03-27
Fix styles and pricing on basket page

georgebabey:
When bulk enrollment changes went out, discounted baskets as well as programs
were negatively impacted. This fixes those pages as well as cleans up some minor
styling changes and typo correction

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -206,6 +206,7 @@ class BasketSummaryView(BasketView):
         display_verification_message = False
         lines_data = []
         show_voucher_form = True
+        is_enrollment_code_purchase = False
         switch_link_text = partner_sku = order_details_msg = None
 
         for line in lines:
@@ -234,6 +235,7 @@ class BasketSummaryView(BasketView):
                         )
             elif line.product.is_enrollment_code_product:
                 line_data = self._get_course_data(line.product)
+                is_enrollment_code_purchase = True
                 show_voucher_form = False
                 order_details_msg = _(
                     '{paragraph_start}By purchasing, you and your organization agree to the following terms:'
@@ -286,7 +288,8 @@ class BasketSummaryView(BasketView):
             'order_details_msg': order_details_msg,
             'partner_sku': partner_sku,
             'show_voucher_form': show_voucher_form,
-            'switch_link_text': switch_link_text
+            'switch_link_text': switch_link_text,
+            'is_enrollment_code_purchase': is_enrollment_code_purchase
         }
 
         return context_updates, lines_data
--- a/ecommerce/static/sass/partials/views/_basket.scss
+++ b/ecommerce/static/sass/partials/views/_basket.scss
@@ -177,6 +177,10 @@
                 color: gray;
             }
 
+            span {
+                padding: 0;
+            }
+
             .remove-voucher {
                 border: none;
                 background: none;
@@ -532,6 +536,7 @@
 
             .price {
                 text-align: right;
+                padding-right: 20px;
             }
 
             .row {
@@ -539,31 +544,14 @@
             }
 
             .amount {
-                padding: 0;
+                padding-bottom: 20px;
                 color: $gray-color;
                 font-size: 18px;
-                
-                span {
-                  float: none !important;
-                  display: block;
-                  text-align: unset;
-
-
-                  &:first-child {
-                    margin: 0 0 20px;
-                  }
-                }
             }
 
             #basket-total {
                 font-size: 18px;
                 font-weight: bold;
-
-                .price {
-                  font-weight: bold;
-                  text-align: unset;
-                  padding: 0;
-                }
             }
 
             #voucher-information {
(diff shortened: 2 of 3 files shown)


ecommerce/static/sass/partials/views/_basket.scss
```diff
@@ -177,6 +177,10 @@
                 color: gray;
             }
 
+            span {
+                padding: 0;
```
georgebabey, line 181: I don't love this, but it gets the line-item spans aligned by removing the bootstrap padding gutters

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -45,24 +45,29 @@ <h2 id="cart-region" class="title">{% trans "in your cart" %}</h2>
     </form>
     <div role="region" aria-labelledby="summary-region" aria-live="polite">
         <h2 id="summary-region" class="title">{% trans "summary" %}</h2>
-        <div class="row">
-            <div id="line-price" class="amount col-xs-4 col-sm-4">
-                <span class="description">{% trans "Price" %}</span>
-                <span class="price">{{line_price|currency:basket.currency}}</span>
+        {% if enrollment_code_purchase %}
+            <div id="summary-header" class="row">
```
georgebabey: Now 2 rows, one for header, one for line item.

ecommerce/extensions/basket/views.py
```diff
@@ -286,7 +288,8 @@ def _process_basket_lines(self, lines):
             'order_details_msg': order_details_msg,
             'partner_sku': partner_sku,
             'show_voucher_form': show_voucher_form,
-            'switch_link_text': switch_link_text
+            'switch_link_text': switch_link_text,
+            'enrollment_code_purchase': is_enrollment_code_purchase
```
jaebradley: Nit: I don't think it really matters, and I don't know if it's Django convention, but should the key value be `is_enrollment_code_purchase`?

georgebabey: I haven't seen a single convention yet, but I'll update to at least make it consistent

georgebabey:
This has been updated now so that the old view of a single line for Price / Discount is shown when a non-enrollment code purchase is made.

Deployed to enterprise sandbox:
- [Bulk purchase](https://ecommerce-business.sandbox.edx.org/basket/add/?sku=C3F8E7A
)
- [Single purchase (offer code `10OFF`)](https://ecommerce-business.sandbox.edx.org/basket/add/?sku=8CF08E5)
- [Program Purchase](https://ecommerce-business.sandbox.edx.org/basket/add/?sku=DD6D2C7&sku=8CF08E5&bundle=52ad909b-c57d-4ff1-bab3-999813a2479b)



**Bulk purchase of enrollment codes**
screen shot 2018-03-27 at 3 55 22 pm

**Single purchase, no offers**
screen shot 2018-03-27 at 3 55 46 pm

**Single purchase w/ 10% off coupon**
screen shot 2018-03-27 at 4 01 08 pm


**Program purchase (2 courses - 25% discount)**
screen shot 2018-03-27 at 10 01 01 pm

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
```diff
@@ -45,24 +45,29 @@ <h2 id="cart-region" class="title">{% trans "in your cart" %}</h2>
...
         <h2 id="summary-region" class="title">{% trans "summary" %}</h2>
-        <div class="row">
-            <div id="line-price" class="amount col-xs-4 col-sm-4">
-                <span class="description">{% trans "Price" %}</span>
-                <span class="price">{{line_price|currency:basket.currency}}</span>
+        {% if is_enrollment_code_purchase %}
+            <div id="summary-header" class="row">
+                <span class="col-xs-4 col-sm-4">{% trans "Price" %}</span>
+                <span class="col-xs-4 col-sm-4">{% trans "Quantity" %}</span>
+                <span class="price col-xs-4 col-sm-4" >{% trans "Subtotal" %}</span>
```
jaebradley, line 52: FYI the `price` class is used for [client-side currency conversion](https://github.com/edx/ecommerce/blob/cb5cc595820635f02ad6b7e8459340a74d2ceafa/ecommerce/static/js/pages/basket_page.js#L326) (yeah, I wrote that, sorry 😞 ).

I *think* this is fine, but you should check by changing [the `edx-price-l10n` cookie](https://github.com/edx/ecommerce/blob/cb5cc595820635f02ad6b7e8459340a74d2ceafa/ecommerce/static/js/pages/basket_page.js#L265) value.

I created [a JIRA ticket](https://openedx.atlassian.net/browse/ENT-939) to track the rewrite work.

rrakhshan (approved):

ecommerce/templates/oscar/basket/partials/client_side_checkout_basket.html
adamstankiewicz: Yeah, I think this is fine since the logic in the client-side currency conversion that @jaebradley linked to uses the [regex](https://github.com/edx/ecommerce/blob/cb5cc595820635f02ad6b7e8459340a74d2ceafa/ecommerce/static/js/pages/basket_page.js#L297) for `prefixMatch` wouldn't match for "Subtotal".

adamstankiewicz (approved):

PR #1691, opened 2018-03-28
Fix unable to download coupon report in non-english languages.

waheedahmed:
field_names list have translated values but upon removing we are using untranslated strings.

LEARNER-4648


waheedahmed:

@edx/learner FYI.

tasawernawaz (approved):
LGTM

PR #1694, opened 2018-03-29
Allow transaction on commit for segment calls

BenjiLee:
@robrap This is ready for review

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -14,8 +14,12 @@ from oscar.core.loading import get_class, get_model
 from rest_framework import serializers
 from rest_framework.reverse import reverse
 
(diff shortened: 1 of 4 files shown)


jmbowman (approved):
Repeating the logic to check the switch and call track_segment_event accordingly slightly triggers my code duplication aversion, but it's probably not worth abstracting it out if this it's just these 2 spots and/or this switch won't be in use for long.  Otherwise, LGTM.

robrap:
@jmbowman: Thanks. The switch is very temporary, in case there are any unforeseen issues.

BenjiLee:
Ticket: https://openedx.atlassian.net/browse/LEARNER-4815
Clean up ticket: https://openedx.atlassian.net/browse/LEARNER-4816

PR #1697, opened 2018-03-30
Fix translations

Ayub-Khan:
-removed i18n generate merge options

-updated header

-updated-translations

### DIFF
--- a/ecommerce/conf/locale/config.yaml
+++ b/ecommerce/conf/locale/config.yaml
@@ -94,9 +94,3 @@ ignore_dirs:
     - node_modules
     - static/bower_components
     - static/build
-
-generate_merge:
-    django.po:
-        - django-partial.po
-    djangojs.po:
-        - djangojs-partial.po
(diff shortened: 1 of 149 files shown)


ecommerce/conf/locale/config.yaml
```diff
@@ -94,9 +94,3 @@ ignore_dirs:
     - node_modules
     - static/bower_components
     - static/build
-
-generate_merge:
```
amangano-edx, line 98: Why does this need to be removed?

amangano-edx (approved):
This seems ok. Do you plan on adding the make target for `i18n_tool validate` in a later commit?

Ayub-Khan:
here is the pr for validate command https://github.com/edx/ecommerce/pull/1668

ecommerce/conf/locale/config.yaml
Ayub-Khan: As it is not used and causes problems for i18n-tool validate command

PR #1698, opened 2018-03-30
Temporary disable of Django Debug toolbar for performance issues

staubina:
By disabling this Debug toolbar we are able to remove the performance pressure on the ecommerce service in the development environment, this is only meant to be a temporary change until we can resolve other ecommerce performance issues.


BenjiLee (dismissed):
Docs: 
See `SHOW_TOOLBAR_CALLBACK` https://django-debug-toolbar.readthedocs.io/en/1.8/configuration.html

Example:
https://coderwall.com/p/-tikrw/disable-django-debug-toolbar-when-debug-true

This would make this not temporary and up to the user to enable it or not.

Edit: This is more a nit, but I think setting it to false by default without commenting it out would make it less temporary.

staubina:
@BenjiLee I want this to be temporary as I believe it is indicative of a greater performance issue.  A canary in a coal mine in a way.

staubina:
@BenjiLee @robrap Should we merge this until we are able to correct the other issues?

PR #1699, opened 2018-03-30
Implement Fulfillment for Journal Products

bessiesteinberg:
- Create Journal Fulfillment module, which is called when a user
  purchases a journal product
- This fulfillment module calls the journal access api in journals
  service which grants access to a specific journal for a specific user
- This commit also includes an additional option field on the
  SiteConfiguration table, "journals_api_url".
- Work remaining:
  - Configure fulfullment module and journals api so that it will only
    accept traffic with appropriate access token


BenjiLee:
Why am I on your commit!!!

bill-filler (approved):
looks great!

PR #1711, opened 2018-04-11
LEARNER-4913: Remove call to enrollments and entitlements when it is an anonymous

robrap:
LEARNER-4913

Skips calls to get enrollments and entitlements for an anonymous basket calculate.

NOTE: The implementation for this is to create a basket with the user=None.  Several other changes were needed to make this work.

This change is protected by a waffle flag.

### DIFF
--- a/ecommerce/enterprise/conditions.py
+++ b/ecommerce/enterprise/conditions.py
@@ -36,6 +36,10 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
         Returns:
             bool
         """
+        if not basket.owner:
+            # An anonymous user is never linked to any EnterpriseCustomer.
+            return False
+
         try:
             learner_data = fetch_enterprise_learner_data(basket.site, basket.owner)['results'][0]
         except (ConnectionError, KeyError, SlumberHttpBaseException, Timeout):
@@ -46,7 +50,6 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
             )
             return False
         except IndexError:
-            # Learner is not linked to any EnterpriseCustomer.
             return False
 
         enterprise_customer = learner_data['enterprise_customer']
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -134,6 +134,9 @@ def track_segment_event(site, user, event, properties):
         (success, msg): Tuple indicating the success of enqueuing the event on the message queue.
             This can be safely ignored unless needed for debugging purposes.
     """
+    if not user:
+        return False, 'Event is not fired for anonymous user.'
+
     site_configuration = site.siteconfiguration
     if not site_configuration.segment_key:
         msg = 'Event [{event}] was NOT fired because no Segment key is set for site configuration [{site_id}]'
--- a/ecommerce/extensions/api/v2/tests/views/test_baskets.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_baskets.py
@@ -452,7 +452,8 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
         self.assertEqual(response.status_code, 200)
 
     @httpretty.activate
-    def test_basket_calculate_by_staff_user_other_username(self):
+    @mock.patch('ecommerce.programs.conditions.ProgramCourseRunSeatsCondition._get_lms_resource_for_user')
+    def test_basket_calculate_by_staff_user_other_username(self, mock_get_lms_resource_for_user):
         """Verify a staff user passing a valid username gets a response about the other user"""
         self.site_configuration.enable_partial_program = True
         self.site_configuration.save()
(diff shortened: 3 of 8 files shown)


robrap:
@BenjiLee @staubina: FYI: I have not tested this locally (outside of unit tests).  I need to either get Programs working locally and/or have someone else test this and/or test on stage. :)

ecommerce/programs/conditions.py
```diff
@@ -67,6 +67,10 @@ def get_user_ownership_data(self, basket, retrieve_entitlements=False):
         """
         enrollments = []
         entitlements = []
+
+        if not basket.owner:
```
robrap: Reviewers: This will fix most of the newly created bug [LEARNER-5001](https://openedx.atlassian.net/browse/LEARNER-5001) by never calling `get_lms_resource`.   However, it is still possible that `get_lms_resource` caching will fail for real learners, so it should be fixed none-the-less.

staubina: This seems absolutely possible for learners when they first start with us and have 0 Enrollments and Entitlements.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -440,12 +436,18 @@ def get(self, request):
             else:
                 return HttpResponseForbidden('Unauthorized user credentials')
         elif user.username == self.MARKETING_USER:
```
staubina, line 438: We plan to change this check once we get the call from the Marketing Site and else where to announce they are requesting for an anonymous user right?

staubina (commented):
Overall looks good to me.  We should test locally before merging though. 

Notes on setting up local programs can be found here: https://openedx.atlassian.net/wiki/spaces/LEARNER/pages/171180253/Program+Bundling

robrap:
@robrap: Remember to search for basket.owner.

ecommerce/extensions/analytics/utils.py
```diff
@@ -134,6 +134,9 @@ def track_segment_event(site, user, event, properties):
         (success, msg): Tuple indicating the success of enqueuing the event on the message queue.
             This can be safely ignored unless needed for debugging purposes.
     """
+    if not user:
```
BenjiLee: Are we ignoring all tracking events for anon users?

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -478,13 +479,89 @@ def test_basket_calculate_by_staff_user_other_username(self):
         }
 
         response = self.client.get(url)
+
+        # Note: This possibly should be in a test specifically around entitlements/enrollments.
+        self.assertTrue(mock_get_lms_resource.called, msg='LMS calls should be made for non-anonymous case.')
```
BenjiLee: Where is this message being generated from? I can't seem to find this string

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -478,13 +479,89 @@ def test_basket_calculate_by_staff_user_other_username(self):
...
+        }
+
+        response = self.client.get(url)
+
+        self.assertTrue(mock_get_lms_resource.called, msg='LMS calls should be skipped for anonymous case.')
+
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.data, expected)
 
+    def setup_anonymous_basket_calculate(self):
```
BenjiLee, line 534: Could we move this method with the other help method?

ecommerce/extensions/api/v2/views/baskets.py
robrap: Yes, just not as part of this PR.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
robrap: Sure. Please open a follow-up PR.

ecommerce/enterprise/conditions.py
```diff
@@ -36,6 +36,10 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
         Returns:
             bool
         """
+        if not basket.owner:
```
robrap, line 39: Can anyone from @edx/business-enterprise-team confirm that only a logged in user could ever satisfy this condition?  
FYI: This is blocking this performance fix from making it to Production.

adamstankiewicz: @robrap Replied in HipChat as well. As far as I know, I don't think that condition would ever be true for an anonymous user, aka would need to be logged in.

robrap: Thanks @adamstankiewicz.

ecommerce/programs/conditions.py
```diff
@@ -55,24 +67,25 @@ def get_lms_resource(self, basket, resource_name, endpoint):
         if not data_list:
             user = basket.owner.username
             try:
-                data_list = endpoint.get(user=user)
-                cache.set(cache_key, data_list, settings.ENROLLMENT_API_CACHE_TIMEOUT)
+                data_list = self._get_lms_resource(endpoint, user)
```
robrap: @BenjiLee: This fails the test because it can't cache (pickle) the mocked object. I need to find another way to write/test this. Argh.

ecommerce/programs/conditions.py
robrap: @BenjiLee: I decided to add this back to try to land this since it is already tested.  I also left the added protection in _get_lms_resource.

ecommerce/programs/conditions.py
robrap: @BenjiLee: Something got messed up here.  This is old code.  I'll look at it in the morning.

BenjiLee (approved):

PR #1713, opened 2018-04-13
small change to allow for Apple Pay verification

staubina:
[LEARNER-4981]

Updating a URL that Apple uses to verify our applications.


jlajoie (approved):
:+1:

PR #1714, opened 2018-04-13
 LEARNER-4968: Do not call any segment events when Calculating the basket

BenjiLee:


ecommerce/coupons/tests/mixins.py
```diff
@@ -461,7 +461,7 @@ def apply_voucher(self, user, site, voucher):
         """ Apply the voucher to a basket. """
         basket = factories.BasketFactory(owner=user, site=site)
         product = voucher.offers.first().benefit.range.all_products()[0]
-        basket.add_product(product)
+        basket.add_product_with_tracking(product)
```
robrap, line 464: Feels risky to stop overriding unless we are sure that is what we want.  There could be other callers in other places.

I think we may need to read some of the oscar docs to determine the right way to do this. If you already have, and determined this is the right way to go, do you mind pointing me to the docs? Thanks.

jlajoie: @robrap this didn't feel like it was related to Oscar specifically but rather just how we were using inheritance, I'm not sure how anything internal to Oscar would be explicitly using our Subclassed Basket, but I could be wrong. Though I agree, it feels a bit of a code smell to modify > 30 files worth of function calls. 

Alternatively, we can solve this by making the default option to track segments on `add_product`, and then explicitly call a `add_product_without_tracking` method, or even the options approach I proposed earlier, which looked like

```python
        # Treat this as a non-temporary basket unless provided in options
        temporary_basket = False
        if options:
            temporary_basket = options.pop('temporary_basket', False)  # Default to non-temporary if not provided

        line, created = super(Basket, self).add_product(product, quantity, options)  # pylint: disable=bad-super-call

        # Do not fire events for free items. The volume we see for edX.org leads to a dramatic increase in CPU
        # usage. Given that orders for free items are ignored, there is no need for these events.
        if line.stockrecord.price_excl_tax > 0 and not temporary_basket:
            properties['cart_id'] = self.id
            track_segment_event(self.site, self.owner, 'Product Added', properties)

        return line, created
```

Where we would then just need to add in 
```python
basket.add_product(product, options={'temporary_basket': true)
```

But @BenjiLee wanted us to be more explicit about the use of including segment tracking which is why we opted for this approach.

ecommerce/extensions/basket/models.py
```diff
@@ -59,15 +59,16 @@ def flush(self):
...
         super(Basket, self).flush()  # pylint: disable=bad-super-call
 
-    def add_product(self, product, quantity=1, options=None):
+    def add_product_with_tracking(self, product, quantity=1, options=None):
         """ Add the indicated product to basket.
 
         Performs AbstractBasket add_product method and fires Google Analytics 'Product Added' event.
         """
-        line, created = super(Basket, self).add_product(product, quantity, options)  # pylint: disable=bad-super-call
+        line, created = self.add_product(product, quantity, options)  # pylint: disable=bad-super-call
```
staubina, line 67: Can we remove the pylint disable?

ecommerce/extensions/basket/models.py
```diff
@@ -59,15 +59,16 @@ def flush(self):
 
         super(Basket, self).flush()  # pylint: disable=bad-super-call
 
-    def add_product(self, product, quantity=1, options=None):
+    def add_product_with_tracking(self, product, quantity=1, options=None):
         """ Add the indicated product to basket.
 
         Performs AbstractBasket add_product method and fires Google Analytics 'Product Added' event.
```
staubina, line 65: Update the comment please.

ecommerce/extensions/basket/models.py
```diff
@@ -59,15 +59,16 @@ def flush(self):
 
         super(Basket, self).flush()  # pylint: disable=bad-super-call
 
-    def add_product(self, product, quantity=1, options=None):
+    def add_product_with_tracking(self, product, quantity=1, options=None):
```
staubina, line 62: Judging by the number of changes in the PR I feel like it would be more efficient to create the method add_product _with_no_tracking.

staubina (commented):
Some questions and small comments.

BenjiLee:
Closed in favor of https://github.com/edx/ecommerce/pull/1715

PR #1715, opened 2018-04-13
LEARNER-4968: Do not call any segment events when Calculating the basket

jlajoie:


ecommerce/extensions/basket/models.py
```diff
@@ -64,11 +64,16 @@ def add_product(self, product, quantity=1, options=None):
...
+
         line, created = super(Basket, self).add_product(product, quantity, options)  # pylint: disable=bad-super-call
 
         # Do not fire events for free items. The volume we see for edX.org leads to a dramatic increase in CPU
         # usage. Given that orders for free items are ignored, there is no need for these events.
-        if line.stockrecord.price_excl_tax > 0:
+        if line.stockrecord.price_excl_tax > 0 and not temporary_basket:
             properties = translate_basket_line_for_segment(line)
             properties['cart_id'] = self.id
             track_segment_event(self.site, self.owner, 'Product Added', properties)
```
robrap, line 79: This all looks great.  However, I had discussed with @BenjiLee adding a required field to `track_segment_event` for skipping the event tracking.  He had that code started locally.  What happened to it?

Also, what about the rest of the calls to `track_segment_event`?

ecommerce/extensions/basket/models.py
```diff
@@ -64,11 +64,16 @@ def add_product(self, product, quantity=1, options=None):
...
+        temporary_basket = False
+        if options:
+            temporary_basket = options.pop('temporary_basket', False)  # Default to non-temporary if not provided
+
         line, created = super(Basket, self).add_product(product, quantity, options)  # pylint: disable=bad-super-call
 
         # Do not fire events for free items. The volume we see for edX.org leads to a dramatic increase in CPU
         # usage. Given that orders for free items are ignored, there is no need for these events.
-        if line.stockrecord.price_excl_tax > 0:
+        if line.stockrecord.price_excl_tax > 0 and not temporary_basket:
```
staubina, line 76: 👍

staubina (approved):
I like this approach a lot better. Thanks for taking another swing at this.

ecommerce/extensions/basket/models.py
jlajoie: I'm not sure I like that approach, it seems a bit on the code smell side of things and we could just not call it if we don't want it, instead of adding the line and then saying "no"?

robrap: @jlajoie: Agreed that this isn't ideal, but I think it solves the following problem.

I add some code and I want to add an event so I copy the line:
```
track_segment_event(self.site, self.owner, 'Product Added', properties)
```
Then I update the event, and I never realize that there are certain odd circumstances in which I don't want to track the event and I have introduced difficult to find bugs.

Or, I copy a line like:
```
track_segment_event(self.site, self.owner, 'Product Added', properties, is_temporary_basket)
```
And my new code doesn't yet have `is_temporary_basket` so I need to read up on what it is and how it gets used and I don't introduce a difficult to find bug.

If you have another proposal that solves this problem, that would be great.  Note that is why I wanted to revive the code that @BenjiLee and I wrote together, because we added `is_temporary_basked` (or something like that) as a required parameter with a doc string that explained the reasoning above.

staubina: @robrap I believe that the above change would mix concerns. track_segment_event is a clear method that should be left as it is.  

The approach here seems cleaner because it does not cross concerns.

staubina: We should potentially talk about this in person.

staubina: @robrap Also `is_temporary_basket` is added to the options within this PR when the method add_product is called.

https://github.com/edx/ecommerce/pull/1715/files#diff-ae237e82e578ba8c25c0917a33381619R355

robrap: @staubina @jlajoie: I don't love my solution, but I haven't heard an alternate solution to the problem I stated.

The main problem is that it is and will be way too easy to introduce a bug.

It is also possible to rename `is_temporary_basket` to `skip_event_tracking` or `mute_event_tracking`.  In either case, the docs will need to clear up why that parameter exists and how it should be used.

I am open to an alternate solution...since as I said...I don't love this one.

jlajoie: @robrap I guess maybe I'm missing what problem we're trying to solve here? Following the example above, we'd add the line, pass in false for `is_temporary_basket` because we want to get events persisted, and then we'd be in the same boat as today. The developer would have to ask themselves "Do I want to add an event here", but instead of deciding whether or not to add the call at all, you now need to decide whether or not to set a boolean.

My understanding is that we ran into this issue with the segment events because we were always emitting these events in add_products when we actually didn't want to be. So this isn't really a problem with the event_tracking, but rather us calling a method when we shouldn't have, and us missing that case. 

I just have a hard time advocating for explicitly calling a method, but then passing in a parameter for that method to both return nothing and do nothing.

I think before you mentioned potentially using the request cache to mute events. If it would be something to suppress ALL events, despite feeling a bit of a hack, I think that approach of setting a context global makes more sense than specifying whether they're muted at the method call level, if that's the route we'd want to go.

BenjiLee: So, it looks like I forgot to also handle flush() which also sends a tracking event that we dont want to be sending in basket calculate. https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/basket/models.py#L50

Unfortunately, we will not be able to use this approach currently in this PR since flush doesn't have an options parameter like add_product does.

BenjiLee:
Attempting to implement request caching in a separate PR which should make handling the flush call easier.

BenjiLee (changes requested):
Blocked on self

robrap:
@BenjiLee: I'm going to close this.  We can reopen or create a new one when we have something that will work.

BenjiLee:
Closed in favor of https://github.com/edx/ecommerce/pull/1786

PR #1718, opened 2018-04-17
WL-1541 | Test cases for whitelabel journal.

hasnain-naveed:


hasnain-naveed:
@ziafazal

ecommerce/journal/api/tests/test_views.py
```diff
@@ -0,0 +1,112 @@
...
+    StockRecordFactory
+)
+
+from ecommerce.tests.testcases import TestCase
+
+Product = get_model('catalogue', 'Product')
+
+
+class JournalProductViewTests(TestCase):
+    path = None
```
ziafazal: nit:Do need these variables here assuming they are initialized in `setUp` of test class.?

ecommerce/journal/api/tests/test_views.py
```diff
@@ -0,0 +1,112 @@
...
+        Asserts the updated product values.
+        """
+        expected_data = self._get_data_for_put_end_point()
+        product = Product.objects.get(id=self.product.id)
+        self.assertEqual(product.title, expected_data['title'])
+        for index, stockrecord in enumerate(product.stockrecords.all()):
+            self.assertEqual(stockrecord.price_excl_tax, expected_data['stockrecords'][index]['price_excl_tax'])
+            self.assertEqual(stockrecord.price_currency, expected_data['stockrecords'][index]['price_currency'])
+
+    def test_update_journal_product(self):
```
ziafazal: I think we should have a test case for product create with valid and invalid data.

ecommerce/journal/api/tests/test_views.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def setUp(self):
+        """
+        Creates the sample data for test cases.
+        """
+        super(JournalProductViewTests, self).setUp()
+        self.value_text = "dummy_text"
+        self.path = reverse("journal:api:v1:journal-list")
+
+        product_class = ProductClassFactory(name="Journal")
```
ziafazal: I think instead of using these Factory classes for product, productclass, StockRecord and productattribute we should use `journal:api:v1` endpoint to create journal.

ecommerce/journal/api/tests/test_serializers.py
```diff
@@ -0,0 +1,191 @@
...
+    StockRecordSerializer,
+    StockRecordSerializerForUpdate
+)
+from ecommerce.tests.factories import PartnerFactory, StockRecordFactory
+from ecommerce.tests.testcases import TestCase
+
+StockRecord = get_model('partner', 'StockRecord')
+
+
+class AttributesSerializerTest(TestCase):
```
ziafazal, line 25: I'm wondering do we really need test cases for `Serializers`?. i think those should be covered with test cases written for views.

ecommerce/journal/tests/test_views.py
```diff
@@ -0,0 +1,36 @@
...
+        self.path = reverse('journal:offers:list')
+
+    def _get_mocked_endpoint(self):
+        journal_bundle = mock.Mock()
+        journal_bundle.uuid = str(uuid.uuid4())
+        journal_bundle.title = "mocked_title"
+        return journal_bundle
+
+    @mock.patch('ecommerce.journal.views.fetch_journal_bundle')
+    def test_get(self, mock_journal_endpoint):
```
ziafazal: Should have get test with a scenario where we don't have any offers.

ecommerce/journal/tests/test_views.py
```diff
@@ -0,0 +1,36 @@
...
+from oscar.core.loading import get_model
+
+from ecommerce.extensions.test import factories
+from ecommerce.tests.testcases import TestCase, ViewTestMixin
+
+Benefit = get_model('offer', 'Benefit')
+ConditionalOffer = get_model('offer', 'ConditionalOffer')
+
+
+class JournalOfferListViewTests(ViewTestMixin, TestCase):
```
ziafazal: Why we don't have offer create/update related test cases?

ziafazal (changes requested):
@hasnain-naveed completed first round of review. See inline comments.

ecommerce/journal/api/tests/test_views.py
hasnain-naveed: removed.

ecommerce/journal/api/tests/test_views.py
hasnain-naveed: added

ecommerce/journal/tests/test_views.py
hasnain-naveed: added

ecommerce/journal/api/tests/test_serializers.py
hasnain-naveed: I think, the main purpose of unit tests is to test each module of code separately. If you still think, that's not the case then I can remove these test cases.

ecommerce/journal/tests/test_views.py
hasnain-naveed: added

ecommerce/journal/api/tests/test_views.py
hasnain-naveed: used the endpoint

ecommerce/journal/api/tests/test_views.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        self.client.logout()
+        response = self.client.get(self.path)
+        self.assertEqual(response.status_code, 401)
+
+    def test_journal_product_view(self):
+        """
+        Tests 'GET' end point for JournalProductViewSet.
+        """
+        self._login_staff_user()
```
ziafazal: It appears you need a user session in all of these test cases. I think we should add these 
```
user = self.create_user(is_staff=True)
self.client.login(username=user.username, password=self.password)
```
in `setUp` and remove this method `_login_staff_user `

ecommerce/journal/api/tests/test_views.py
```diff
@@ -0,0 +1,138 @@
...
+                }
+            ]
+        }
+
+    def _assert_product_update(self):
+        """
+        Asserts the updated product values.
+        """
+        expected_data = self._get_data_for_put_endpoint()
+        product = Product.objects.get(id=self.product.id)
```
ziafazal: I think you should use journal list view to get journal products instead of getting them directly from database

ecommerce/journal/api/tests/test_views.py
```diff
@@ -0,0 +1,138 @@
...
+                {
+                    "dummy-key1": "dummy-value1",
+                    "dummy-key2": "dummy-value2"
+                }
+            ),
+            "application/json"
+        )
+
+        # product's title remains unchanged.
+        self.assertEqual(Product.objects.get(id=self.product.id).title, product_title)
```
ziafazal: I think we should assert an HTTP invalid request status 400 instead of comparing product titles

ziafazal (changes requested):
@hasnain-naveed thanks for the update there is still some room for improvement.

ziafazal (approved):
👍

bill-filler (approved):
nice work @hasnain-naveed

PR #1720, opened 2018-04-18
LEARNER-4913/LEARNER-5061: Added is_anonymous parameter to the calculate URL

staubina:
https://openedx.atlassian.net/browse/LEARNER-4913
https://openedx.atlassian.net/browse/LEARNER-5061

Add in check for the is_anonymous URL parameter.  If it is not present we will treat the request as if there were a username.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_baskets.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_baskets.py
@@ -490,7 +490,6 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
     @httpretty.activate
     @mock.patch('ecommerce.programs.conditions.ProgramCourseRunSeatsCondition._get_lms_resource_for_user')
     @override_flag("use_basket_calculate_none_user", active=True)
-    @override_switch("debug_logging_for_excessive_lms_calls", active=True)
     def test_basket_calculate_anonymous_skip_lms(self, mock_get_lms_resource_for_user):
         """Verify a call for an anonymous user skips calls to LMS for entitlements and enrollments"""
         products, url = self.setup_anonymous_basket_calculate()
--- a/ecommerce/extensions/api/v2/tests/views/test_baskets.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_baskets.py
@@ -319,11 +319,11 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
         super(BasketCalculateViewTests, self).setUp()
         self.products = ProductFactory.create_batch(3, stockrecords__partner=self.partner, categories=[])
         self.path = reverse('api:v2:baskets:calculate')
-        self.url = self._generate_sku_url(self.products)
         self.range = factories.RangeFactory(includes_all_products=True)
         self.product_total = sum(product.stockrecords.first().price_excl_tax for product in self.products)
 
         self.user = self._login_as_user(is_staff=True)
+        self.url = self._generate_sku_url(self.products, username=self.user.username)
 
     def test_no_sku(self):
         """ Verify bad response when not providing sku(s) """
@@ -481,7 +481,6 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
 
         response = self.client.get(url)
 
-        # Note: This possibly should be in a test specifically around entitlements/enrollments.
         self.assertTrue(mock_get_lms_resource_for_user.called, msg='LMS calls should be made for non-anonymous case.')
 
         self.assertEqual(response.status_code, 200)
@@ -536,14 +535,10 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
         """
         Sets up anonymous basket calculate.
 
-        Side Effects:
-            Logs in the Marketing User, required for anonymous baskets at this time.
-
         Returns:
             products, url: The product list and the url for the anonymous basket
                 calculate.
         """
-        self._login_as_user(BasketCalculateView.MARKETING_USER, is_staff=True)
         self.site_configuration.enable_partial_program = True
         self.site_configuration.save()
         offer = ProgramOfferFactory(
@@ -560,17 +555,17 @@ class BasketCalculateViewTests(ProgramTestMixin, TestCase):
         return products, url
 
     @httpretty.activate
+    @mock.patch('ecommerce.extensions.api.v2.views.baskets.logger.warning')
     @mock.patch('ecommerce.extensions.api.v2.views.baskets.BasketCalculateView._calculate_temporary_basket')
-    @override_flag("use_cached_basket_calculate_for_marketing_user", active=True)
-    def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
-        """Verify a request made by the Marketing Staff user is cached"""
-        self._login_as_user(BasketCalculateView.MARKETING_USER, is_staff=True)
-
+    @override_flag('use_cached_basket_calculate_for_marketing_user', active=True)
+    @override_switch('debug_logging_predates_is_anonymous', active=True)
+    def test_basket_calculate_anonymous_caching(self, mock_calculate_basket, mocked_logger):
+        """Verify a request made with the is_anonymous parameter is cached"""
         url_with_one_sku = self._generate_sku_url(self.products, number_of_products=1, username=None)
         url_with_two_skus = self._generate_sku_url(self.products, number_of_products=2, username=None)
 
         expected = {'Test Succeeded': True}
-        mock_calculate_basket.return_value = {'Test Succeeded': True}
+        mock_calculate_basket.return_value = expected
 
         # Call BasketCalculate. The cache should not be hit.
         response = self.client.get(url_with_one_sku)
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -562,14 +564,14 @@ def setup_anonymous_basket_calculate(self):
     @mock.patch('ecommerce.extensions.api.v2.views.baskets.BasketCalculateView._calculate_temporary_basket')
     @override_flag("use_cached_basket_calculate_for_marketing_user", active=True)
     def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
-        """Verify a request made by the Marketing Staff user is cached"""
-        self._login_as_user(BasketCalculateView.MARKETING_USER, is_staff=True)
+        """Verify a request made with the is_anonymous parameter is cached"""
+        self._login_as_user(MARKETING_USER, is_staff=True)
```
robrap: I'd no longer include the MARKETING_USER for this test.  We want this to work for any user passing the is_anonymous param.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -562,14 +564,14 @@ def setup_anonymous_basket_calculate(self):
...
 
         url_with_one_sku = self._generate_sku_url(self.products, number_of_products=1, username=None)
         url_with_two_skus = self._generate_sku_url(self.products, number_of_products=2, username=None)
 
         expected = {'Test Succeeded': True}
-        mock_calculate_basket.return_value = {'Test Succeeded': True}
+        mock_calculate_basket.return_value = expected
```
robrap, line 568: Oops.  Nice.  :)

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -607,18 +609,60 @@ def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
 
         # Check that a being logged in as a different user doesn't hit the cache
```
robrap: Instead of updating this, I'd delete this whole case.  I don't think it applies any more.  It used to be checking that the caching only worked for marketing_site_worker, but not other users.  However, that is no longer the case.  Now it should work for any user with the is_anonymous flag.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -607,18 +609,60 @@ def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
...
+        response = self.client.get(different_user_url)
         self.assertEqual(response.status_code, 200)
         self.assertTrue(mock_calculate_basket.called, msg='The cache should be missed.')
         self.assertEqual(response.data, expected)
 
+    @httpretty.activate
+    @mock.patch('ecommerce.extensions.api.v2.views.baskets.logger.warning')
+    @mock.patch('ecommerce.extensions.api.v2.views.baskets.BasketCalculateView._calculate_temporary_basket')
+    @override_flag("use_cached_basket_calculate_for_marketing_user", active=True)
+    def test_no_user_anonymous_params_log(self, mock_calculate_basket, mocked_logger):
```
robrap: Rename to `test_no_query_params_log`

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -607,18 +609,60 @@ def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
...
+        """
+        Verify that when the request contains neither a username parameter or is_anonymous a Warning is logged and
+        response is not invalid.
```
robrap: Can you add something like the following:
```
NOTE: This is temporary until we no longer have these calls and can ultimately return a 400 error.
Due to backward incompatibility, we should make the switch to a 400 error once we need a version change.
```

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -607,18 +609,60 @@ def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
...
+        """
+        self._login_as_user(MARKETING_USER, is_staff=True)
```
robrap: I think we want something different in the following two cases:
- For marketing_site_worker, the code should continue to work as-is, i.e., acting as if you sent is_anonymous.
- For all other users, the code should continue to work as-is, i.e., acting as if you sent  username=user.username? 

Given this, I think we still temporarily want to support the `MARKETING_USER` special case in the code for now.

Once everything is working as planned, we can remove this code.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -607,18 +609,60 @@ def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
...
+        self.assertTrue(mock_calculate_basket.called)
+        self.assertTrue(mocked_logger.called)
+        self.assertEqual(response.data, expected)
+
+    @httpretty.activate
+    @mock.patch('ecommerce.extensions.api.v2.views.baskets.BasketCalculateView._calculate_temporary_basket')
+    @override_flag("use_cached_basket_calculate_for_marketing_user", active=True)
+    def test_conflicting_user_anonymous_params(self, mock_calculate_basket):
+        """
+        Verify that when the request contains both a username parameter and a is_anonymous a Bad Request response
```
robrap: I made some minor tweaks:
```
Verify that when the request contains both a username and an is_anonymous parameter, a Bad Request response
is returned.
```

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -340,7 +340,6 @@ class BasketDestroyView(generics.DestroyAPIView):
 
 class BasketCalculateView(generics.GenericAPIView):
     permission_classes = (IsAuthenticated,)
-    MARKETING_USER = 'marketing_site_worker'
```
robrap: I think we may still want this.  See my test related comment.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -435,14 +437,14 @@ def get(self, request):
                     logger.debug('Request username: [%s] does not exist', username)
             else:
                 return HttpResponseForbidden('Unauthorized user credentials')
-        elif user.username == self.MARKETING_USER:
-            # A request made by the marketing user without a username query param,
-            # means this is calculating the non-logged in (anonymous) price.
-            # TODO: LEARNER-4993: Switch to a query param rather than hardcoding to
-            # the marketing user.
+        elif is_anonymous:
```
robrap: May want this to be something like:
```
elif is_anonymous OR user.username == self.MARKETING_USER:
    if not is_anonymous:
        logger.warning('Request to Basket Calculate by marketing_site_worker without any parameter.')
    ...
```

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -435,14 +437,14 @@ def get(self, request):
...
             if waffle.flag_is_active(request, "use_basket_calculate_none_user"):
                 user = None
-            is_anonymous = True
+        if not username and not is_anonymous:
+            logger.warning(
+                'Request to Basket Calculate must supply either username or is_anonymous query param [%s]',
```
robrap: `Request` => `Requests`.  Also, I'd fix the message a little like:
```
        logger.warning(
                'Request to Basket Calculate must supply either username or is_anonymous query param. Requesting user=[%s]',
                user.username
        )
```

robrap (changes requested):
Thanks @staubina.  Some of my requests assume a 2 phase roll-out:
1. Add new functionality for parameter.
2. In a future PR, clean-up old code for MARKETING_USER.

We can discuss if you don't agree or have questions.  Thanks.

staubina:
@robrap Please take another look when you have a moment.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -536,13 +536,13 @@ def setup_anonymous_basket_calculate(self):
         Sets up anonymous basket calculate.
 
         Side Effects:
-            Logs in the Marketing User, required for anonymous baskets at this time.
+            Logs in the Marketing User, required along with the is_anonymous parameter for anonymous
```
robrap: Fix message.  No longer logs in the marketing user.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -603,23 +602,66 @@ def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
         self.assertTrue(mock_calculate_basket.called, msg='The cache should be missed.')
         self.assertEqual(response.data, expected)
 
+        # Check that cache works and that log message is sent for Marketing user
+        # Note: This and the following checks related to is should be removed once we remove
```
robrap: Typo: `related to is should`
Can we change this to a `TODO: LEARNER-XXXX: ..." comment after creating a new ticket for this clean-up?

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -603,23 +602,66 @@ def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
...
         self.assertEqual(response.data, expected)
 
+    @httpretty.activate
+    @mock.patch('ecommerce.extensions.api.v2.views.baskets.logger.warning')
+    @mock.patch('ecommerce.extensions.api.v2.views.baskets.BasketCalculateView._calculate_temporary_basket')
+    @override_flag("use_cached_basket_calculate_for_marketing_user", active=True)
+    def test_no_query_params_log(self, mock_calculate_basket, mocked_logger):
+        """
+        Verify that when the request contains neither a username parameter or is_anonymous a Warning is logged and
+        response is not invalid.
```
robrap: `response is valid`

ecommerce/extensions/api/v2/tests/views/test_baskets.py
```diff
@@ -603,23 +602,66 @@ def test_basket_calculate_anonymous_caching(self, mock_calculate_basket):
...
+
+        NOTE: This is temporary until we no longer have these calls and can ultimately return a 400 error.
+        Due to backward incompatibility, we should make the switch to a 400 error once we need a version change.
+        """
+        expected = {'Test Succeeded': True}
+        mock_calculate_basket.return_value = expected
+
+        url_with_one_sku = self._generate_sku_url(self.products, number_of_products=1)
```
robrap, line 635: This is where we have two cases, right?
- No parameters and Marketing user => anonymous
- No parameters and a different user => username=user.username

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -435,14 +438,22 @@ def get(self, request):
                     logger.debug('Request username: [%s] does not exist', username)
             else:
                 return HttpResponseForbidden('Unauthorized user credentials')
-        elif user.username == self.MARKETING_USER:
+        elif is_anonymous or user.username == self.MARKETING_USER:
             # A request made by the marketing user without a username query param,
             # means this is calculating the non-logged in (anonymous) price.
-            # TODO: LEARNER-4993: Switch to a query param rather than hardcoding to
-            # the marketing user.
+            # the marketing user. The check for the marketing user is being left until usage
```
robrap: Delete `the marketing user.`.
Maybe add `TODO: LEARNER-XXXX: ...` for clean-up story.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -435,14 +438,22 @@ def get(self, request):
...
+            # of the is_anonymous flag is improved.
+            if not is_anonymous:
+                logger.warning('Request to Basket Calculate by marketing_site_worker without any parameter.')
+                is_anonymous = True
             if waffle.flag_is_active(request, "use_basket_calculate_none_user"):
                 user = None
-            is_anonymous = True
```
robrap: This line should have been left alone.  This whole case is about making it anonymous.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -435,14 +438,22 @@ def get(self, request):
...
-        elif user.username == self.MARKETING_USER:
+        elif is_anonymous or user.username == self.MARKETING_USER:
             # A request made by the marketing user without a username query param,
             # means this is calculating the non-logged in (anonymous) price.
-            # TODO: LEARNER-4993: Switch to a query param rather than hardcoding to
-            # the marketing user.
+            # the marketing user. The check for the marketing user is being left until usage
+            # of the is_anonymous flag is improved.
+            if not is_anonymous:
+                logger.warning('Request to Basket Calculate by marketing_site_worker without any parameter.')
```
robrap: FYI: We _might_ need logger.error.  Let's see what we determine based on the PR we are trying to get to Prod.  I wasn't able to see log messages, and not sure why yet.

robrap (changes requested):
Thanks @staubina.

ecommerce/extensions/api/v2/views/baskets.py
staubina: It is more explicit the way it is now.  I set is_anonymous to True if it is not already true on line 448 right above here. Right after logging that we are checking for the marketing_site_worker.

ecommerce/extensions/api/v2/tests/views/test_baskets.py
staubina: This was added to trigger this code path explicitly

```
        if not username and not is_anonymous:
            logger.warning(
                ('Request to Basket Calculate must supply either username or '
                 'is_anonymous query param. Requesting user=[%s]'),
                user.username
            )
```

staubina:
@robrap Please take another look.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -421,29 +421,50 @@ def get(self, request):
...
-            # the marketing user.
-            if waffle.flag_is_active(request, "use_basket_calculate_none_user"):
-                user = None
+
+        if basket_owner.username == self.MARKETING_USER and not is_anonymous:
+            # For legacy requests that predate is_anonymous parameter, we will calculate
+            # an anonymous basket if the calculated user is the marketing user.
+            # TODO: LEARNER-5057: Remove this special case for the marketing user
+            # once logs show no more requests with no parameters.
             is_anonymous = True
```
robrap: @staubina: I understand your earlier comment now about this.  I made this its own if statement to make it even more clear.

Also, I removed the log message here, because this case will already have a log message from one of the cases above.

robrap (approved):

PR #1725, opened 2018-04-20
Fix for https://openedx.atlassian.net/browse/LEARNER-4765

raeeschachar:
https://openedx.atlassian.net/browse/LEARNER-4765

### DIFF
--- a/e2e/test_payment.py
+++ b/e2e/test_payment.py
@@ -8,8 +8,9 @@ from selenium.webdriver.support.select import Select
 from selenium.webdriver.support.wait import WebDriverWait
 
 from e2e.api import DiscoveryApi, EcommerceApi, EnrollmentApi
-from e2e.config import LMS_USERNAME, PAYPAL_EMAIL, PAYPAL_PASSWORD
-from e2e.helpers import EcommerceHelpers
+from e2e.config import LMS_USERNAME
+from e2e.constants import ADDRESS_FR, ADDRESS_US
+from e2e.helpers import EcommerceHelpers, LmsHelpers
 
 log = logging.getLogger(__name__)
 
@@ -60,28 +61,6 @@ class TestSeatPayment(object):
         # Click the payment button
         selenium.find_element_by_id('payment-button').click()
 
-    def checkout_with_paypal(self, selenium):
-        selenium.find_element_by_css_selector('button.payment-button[data-processor-name=paypal]').click()
-
-        # Wait for login form to load. PayPal's test environment is slow.
-        login_iframe = selenium.find_element_by_css_selector('#injectedUnifiedLogin > iframe')
-        selenium.switch_to.frame(login_iframe)
-
-        # Log into PayPal
-        email = selenium.find_element_by_id('email')
-        password = selenium.find_element_by_id('password')
-        email.send_keys(PAYPAL_EMAIL)
-        password.send_keys(PAYPAL_PASSWORD)
-
-        selenium.find_element_by_id('btnLogin').click()
-        selenium.switch_to.default_content()
-
-        # Wait for the checkout form to load, and for the loading spinner to disappear.
-        wait = WebDriverWait(selenium, 10)
-        spinner_invisibility = EC.invisibility_of_element_located((By.ID, 'spinner'))
-        wait.until(spinner_invisibility)
-        selenium.find_element_by_id('confirmButtonTop').click()
-
     def assert_browser_on_receipt_page(self, selenium):
         selenium.find_element_by_id('receipt-container')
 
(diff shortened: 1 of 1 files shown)


raeeschachar:
@robrap @staubina 
Please review. This fixes the Credit Card payment issue. Will be pushing another PR for Paypal fix later. Idea here is to merge this in as soon for payment coverage, as discussed. 
results: https://tools-edx-jenkins.edx.org/job/e2e/job/stage-ecommerce/395/console

ecommerce/extensions/order/migrations/0015_create_disable_repeat_order_check_switch.py
```diff
@@ -9,7 +9,7 @@
 
 def create_switch(apps, schema_editor):
     Switch = apps.get_model('waffle', 'Switch')
-    Switch.objects.get_or_create(name=DISABLE_REPEAT_ORDER_CHECK_SWITCH_NAME, defaults={'active': False})
+    Switch.objects.get_or_create(name=DISABLE_REPEAT_ORDER_CHECK_SWITCH_NAME, defaults={'active': True})
```
robrap: Migrations should be considered semi-permanent. Assume many environments have already run an existing migration. I think you want to add a new migration. You can use a similar name, but call it “update” instead of “create”. 

A couple of other thoughts about this:
1. We need to ensure we won’t break any environments by turning this on, especially prod, and including open source. If this flag is still required, we may need to first check that someone didn’t set it to False in an environment before changing its default. I don’t think we can get this from the model. Changing defaults is complex.
2. Is it possible that this was meant to be a temporary flag for rollout? I do not know its history. If it was meant to be temporary, the much simpler and better solution would be cleaning up this flag and all of its code.

robrap (changes requested):
Thanks Raees. Don’t forget to take a break as well. 🙂

ecommerce/extensions/order/migrations/0015_create_disable_repeat_order_check_switch.py
robrap: Actually, we should leave this decision and code to a separate PR. Since you manually turned this on in Stage, I don’t think you need this for Stage e2e. We will want to figure this out, but I’d separate it out to a separate PR and/or story.

raeeschachar: @robrap 
Makes sense. I will remove the flag part from this PR and keep the flag active for stage. 
Resting too after this :smile:

raeeschachar:
@robrap Done. 
https://tools-edx-jenkins.edx.org/job/e2e/job/stage-ecommerce/396/console
Feel free to review on Monday. Happy weekend :)

e2e/test_payment.py
```diff
@@ -60,28 +61,6 @@ def checkout_with_credit_card(self, selenium, address):
         # Click the payment button
         selenium.find_element_by_id('payment-button').click()
 
-    def checkout_with_paypal(self, selenium):
```
robrap, line 63: Is this a function we should have deleted earlier with the initial delete?

e2e/test_payment.py
```diff
@@ -136,3 +119,20 @@ def get_verified_seat(self, course_run):
...
+        verified_seat = self.get_verified_seat(course_run)
+
+        for address in (ADDRESS_US, ADDRESS_FR,):
+            self.add_item_to_basket(selenium, verified_seat['sku'])
+            self.checkout_with_credit_card(selenium, address)
+            self.assert_browser_on_receipt_page(selenium)
+
+            course_run_key = course_run['key']
+            self.assert_user_enrolled_in_course_run(LMS_USERNAME, course_run_key)
+            self.refund_orders_for_course_run(course_run_key)
```
robrap: Does refund need an assertion.

robrap (approved):
When you have a green build AND the Jenkins run against this branch passes, feel free to merge.

e2e/test_payment.py
```diff
@@ -136,3 +119,20 @@ def get_verified_seat(self, course_run):
                 verified_seat = seat
                 break
         return verified_seat
+
+    def test_verified_seat_payment_with_credit_card(self, selenium):
```
robrap, line 124: One last thing. It may make sense to add a comment that states that this test requires that flag to be turned on.

e2e/test_payment.py
raeeschachar: Yes, I was trying to fix this so both tests can be merged together that is why it was here.

e2e/test_payment.py
raeeschachar: Done

Ayub-Khan (changes requested):
As this run through the jenkins job and jenkins job creates a new ecommerce environment and after that injects the stage credentials and variable to configure it with stage. Its not actual copy of stage. Which means the switches have the default value set by the code. This will fail. To correct the behavior we will have to set the default value of 'disable_repeat_order_check' True.

ecommerce/extensions/order/migrations/0015_create_disable_repeat_order_check_switch.py
Ayub-Khan: @robrap this flag was created for the e2e tests. To enable e2e buy a single item multiple times.

Ayub-Khan (approved):
It seems my assumption was not correct. Activating the stage switch passing the test.

BenjiLee:
https://tools-edx-jenkins.edx.org/job/e2e/job/stage-ecommerce/407
Build is still failing

Reverting commit https://github.com/edx/ecommerce/pull/1733

PR #1726, opened 2018-04-20
WIP: LEARNER-5061: Use anonymous calculation for additional cases.

robrap:
Use anonymous calculation for additional cases.
Also cleans up some logging.

LEARNER-5061

Notes:
- Still needs test clean-up.
- Still left in one case for debug_logging_for_excessive_lms_calls in case this doesn't solve the rest of the problem.

### DIFF
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -429,10 +429,6 @@ class BasketCalculateView(generics.GenericAPIView):
         user_does_not_exist = False
         # If a username is passed in, validate that the user has staff access or is the same user.
         if username:
-            if waffle.switch_is_active("debug_logging_for_excessive_lms_calls"):
-                if username == self.MARKETING_USER:
-                    logger.warning("BasketCalculateView called with query parameter username=marketing_site_worker. "
-                                   "user=[%s]", user.username)  # pragma: no cover
             if user.is_staff or (user.username.lower() == username.lower()):
                 try:
                     user = User.objects.get(username=username)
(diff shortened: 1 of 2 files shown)


BenjiLee:
~Looks good. I'm not entirely convinced that it will resolve https://openedx.atlassian.net/browse/LEARNER-5064 but we have the reproduction steps to test it.~

edit: I guess that depends on whether a newly registered user is still considered user=marketing_stie_worker. If it is, this approach would work.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -441,20 +437,20 @@ def get(self, request):
...
-                                   'user_does_not_exist=[%s].', username, user_does_not_exist)  # pragma: no cover
-        elif user.username == self.MARKETING_USER:
+
+        # This covers the following cases:
+        # 1. The MARKETING_USER is doing an anonymous basket calculate.
+        # 2. Someone passed a query parameter with username=MARKETING_USER
+        # 3. The username parameter user did not exist and this call was made by
+        #    the MARKETING_USER.
+        # For all cases, we will provide the anonymous user calculation.
+        if user.username == self.MARKETING_USER:
```
BenjiLee, line 445: Should we also add 
`if user_does_not_exist or user.username == self.MARKETING_USER:`

If a user doesn't exist, there are no enrollments/entitlements to get.

Edit: Nevermind, then we would be defaulting to the original user.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -429,10 +429,6 @@ def get(self, request):
         user_does_not_exist = False
         # If a username is passed in, validate that the user has staff access or is the same user.
         if username:
```
BenjiLee: I find myself getting confused with `username` and `user`. `user` also has a `user.username`

Maybe we can do 
`username` -> `requested_user`?
`user` -> `logged_in_user`?

Thoughts?

ecommerce/extensions/api/v2/views/baskets.py
staubina: Should this just be merged with my changes?

robrap:
Closing in favor of https://github.com/edx/ecommerce/pull/1720

PR #1727, opened 2018-04-20
LEARNER-5061: Partial revert of additional logging.

robrap:
Partial revert of additional logging.

NOTE: I left in the logging near the LMS calls just in case we need to see where LMS calls are coming from in the future.

LEARNER-5061


robrap:
@staubina @bderusha @BenjiLee: I'm not sure if I am blocked on merging due to code cov or lack of a review.  I'd like to get a thumb to see.

If it is due to code coverage, it is because I deleted some covered code which brought down the percent by 0.01%.  Do we need a different threshold?  Or do we want to leave it as it is, and force me to increase coverage elsewhere?

PR #1730, opened 2018-04-23
Added paypal test

raeeschachar:


raeeschachar:
@robrap @staubina 
Added paypal test, please review.
https://tools-edx-jenkins.edx.org/job/e2e/job/stage-ecommerce/434/console

robrap:
@raeeschachar: Did you see the comments about the credit card test from @BenjiLee elsewhere?
- The test is reliant on a switch that is off in Prod, so the proper fix should be to update the test rather than turn the switch on in Stage.
- The test continued to fail in Jenkins on master so it was reverted.
I don't know how much of the above will apply to this as well?
Thanks and good luck.

raeeschachar:
@robrap 
Yup I did. I am rechecking everything after de activating the flag. Will re push once everything is working as expected, for review. Thank you.

raeeschachar:
@robrap @staubina 
Paypal is working now without disable_repeat_order_check flag as on prod. Please review. Once this is in I will try to fix credit card test with same flags. 
https://tools-edx-jenkins.edx.org/job/e2e/job/stage-ecommerce/445/console

tasawernawaz (approved):
Don't have much idea about `selenium`, but LGTM.

robrap (approved):
Thanks for this. 

I'm still not sure: 
- why these tests aren't using bok_choy helpers, and
- why these tests don't use page objects for encapsulation.

We should keep this in mind when and if we wish to adjust/refactor in the future.

robrap:
@raeeschachar: Once this merges, please ensure through Jenkins/GoCD that Stage looks ok and the e2e tests have passed.  If we see failures again, please revert and get it green again.  Hopefully that doesn't happen, but just in case.

staubina (approved):

PR #1731, opened 2018-04-23
logical grouping fix

McKenzieW:
Being an entitlement product was lumped with all remaining necessary conditions for coupon applicability instead of first checking for an applicable product type, then also for all other necessary conditions.

### DIFF
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -53,7 +53,7 @@ class Benefit(AbstractBenefit):
         """" Filters out products that aren't seats or entitlements or that don't have a paid certificate type. """
         return [
             line for line in lines
-            if line.product.is_seat_product or line.product.is_course_entitlement_product and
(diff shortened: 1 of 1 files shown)


McKenzieW:
@cpennington @MatthewPiatetsky @mulby

McKenzieW:
@staubina it looks like this change make a test you added for the new anonymous basketcalculate logic fail. Any thoughts on why that might be?

ecommerce/extensions/offer/models.py
```diff
@@ -235,9 +235,8 @@ def is_condition_satisfied(self, basket):
         In addition to Oscar's check to see if the condition is satisfied,
         a check for if basket owners email domain is within the allowed email domains.
         """
-        if basket.owner and not self.is_email_valid(basket.owner.email):
```
staubina: The removal of `basket.owner and `

PR #1735, opened 2018-04-26
Credit Card test fix https://openedx.atlassian.net/browse/LEARNER-4765

raeeschachar:


raeeschachar:
@robrap @staubina 
Please review. 
Tested it multiple times with 'disable repeat order check' flag turned to off as on Prod. Hopefully GoCD pipeline runs will be stable now. 
https://tools-edx-jenkins.edx.org/job/e2e/job/stage-ecommerce/453/
https://tools-edx-jenkins.edx.org/job/e2e/job/stage-ecommerce/452/
https://tools-edx-jenkins.edx.org/job/e2e/job/stage-ecommerce/451/ 

@robrap for your valid comment on why aren't we using bokchoy. We can do that too later on once we have a local environment fix because it would require rewriting the tests. Currently I had to print page_source every time to add/verify selectors.

ahsan-ul-haq (approved):

robrap (approved):
Thanks for this.

In regards to you comment: 
> Currently I had to print page_source every time to add/verify selectors.

In this situation, I've typically used debugging tools inside the browser for breakpoints and/or using the console to write/verify selectors.

PR #1736, opened 2018-04-26
compare enrollment run key to program data course run list

McKenzieW:
Like on the marketing site, we were comparing course run key strings to course key strings in the offer satisfaction logic in the ecommerce service instead of comparing them to the list of course run keys for a given course in the program data retrieved from discovery. This caused issues with the Supply Chain Micromasters due to the state of course keys and course run keys for the program.

[LEARNER-5030](https://openedx.atlassian.net/browse/LEARNER-5030)


robrap (approved):
I approve, assuming the following is true:
1. The old test data was wrong and this just fixes it to align with real data, 
2. The updated test fails without the new code change.
Thanks!

PR #1741, opened 2018-05-01
LEARNER-5001: Introduce CACHE_MISS utility object.

robrap:
[LEARNER-5088](https://openedx.atlassian.net/browse/LEARNER-5088)

@ormsbee: FYI: I'd like to try some of these ideas out and give them some burn in before moving them to their own repo.

### DIFF
--- a/ecommerce/programs/conditions.py
+++ b/ecommerce/programs/conditions.py
@@ -11,6 +11,7 @@ from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
 
+from ecommerce.core.cache_utils import CACHE_MISS, safe_cache_get
 from ecommerce.core.utils import get_cache_key, traverse_pagination
 from ecommerce.extensions.offer.decorators import check_condition_applicability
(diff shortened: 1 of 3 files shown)


ecommerce/programs/conditions.py
```diff
@@ -55,7 +53,7 @@ def _get_lms_resource_for_user(self, basket, resource_name, endpoint):
             resource=resource_name,
             username=basket.owner.username,
         )
-        data_list = cache.get(cache_key, CACHE_MISS)
+        data_list = safe_cache_get(cache_key)
         if data_list is CACHE_MISS:
```
BenjiLee, line 57: CACHE_MISS as a variable does not seem inherently related to the safe cache util.

Just an idea:

If we created another method in the util like below...
```
def is_safe_cache_miss(cache_result):
    if cache_result is CACHE_MISS:
        return True
```

It would look something like...
```
data_list = safe_cache_get(cache_key)
    if is_safe_cache_miss(data_list):
        ...            
```

BenjiLee (approved):

ecommerce/programs/conditions.py
robrap: As discussed, we'll go with the original style for now.

PR #1742, opened 2018-05-02
Attribute error in Receipt Response View

ahsan-ul-haq:
LEARNER-5157


staubina (approved):
This seems straight forward and should not need another review.

staubina:
FYI @robrap 
I do not think you need to review, but wanted to give you a heads up.

robrap:
Thanks @staubina.  FYI: @mikedikan, this should resolve one of the existing alerts.

PR #1751, opened 2018-05-08
Allow pytest to be run locally

BenjiLee:
Allows pytest to be run locally. 

Pytest is much quicker to run locally than `./manage.py test`. Not all tests currently work with pytest.


robrap (approved):
\o/

PR #1753, opened 2018-05-09
ARCH-92 Use django-rest-swagger for API doc generation.

douglashall:
https://openedx.atlassian.net/browse/ARCH-92

**Sandbox:** https://ecommerce-swagger.sandbox.edx.org/api-docs/


ecommerce/templates/rest_framework_swagger/index.html
```diff
@@ -1,44 +0,0 @@
-{% extends 'rest_framework_swagger/base.html' %}
```
nasthagiri, line 1: Did this template provide any overrides or customizations that is now no longer there?

nasthagiri (approved):
approved, though with a question.

bderusha (approved):

ecommerce/templates/rest_framework_swagger/index.html
douglashall: As far as I could tell, this template was not being used anywhere.

PR #1756, opened 2018-05-10
Allow multiple sites for single partner.

waheedahmed:
Allow multiple site configuration objects against same partner for different sites.

LEARNER-5052

PR #1757, opened 2018-05-10
Add default site to partner.

waheedahmed:
LEARNER-5052


waheedahmed:

staubina (approved):
The code changes seem straight forward.  We should test around this according to @feanil's suggestions on the ticket though before merging.

waheedahmed:
@staubina I have added my findings on the ticket.

ahsan-ul-haq (approved):

ecommerce/extensions/partner/models.py
```diff
@@ -8,6 +8,7 @@ class Partner(AbstractPartner):
     short_code = models.CharField(max_length=8, unique=True, null=False, blank=False)
     enable_sailthru = models.BooleanField(default=True, verbose_name=_('Enable Sailthru Reporting'),
                                           help_text='DEPRECATED: Use SiteConfiguration!')
+    default_site = models.OneToOneField('sites.Site', null=True, blank=True, on_delete=models.PROTECT)
```
waheedahmed, line 11: @staubina Should we make this field required after deploying to production?

staubina: It is only required if there is more than on site correct?

waheedahmed: @staubina In actual yes but then we have to check if there is a single site configuration object then return site from it otherwise return default site. By making this required we don't have to check that and we can just use the default site.

staubina: 👍 I like that plan then.

waheedahmed: Okay, I'll add the migration in https://github.com/edx/ecommerce/pull/1756.

PR #1408, opened 2017-07-22
Mfrey/learner 2014

mjfrey:
Do not send GA order events for Coupons or Enrollment Codes.

LEARNER-2014

PR #1759, opened 2018-05-11
ENT-967: track enrollment code purchases on ecommerce basket page.

saleem-latif:
**JIRA Ticket:** [ENT-967](https://openedx.atlassian.net/browse/ENT-967)

**Description:**
This PR adds tracking for bulk enrollment code purchases.

### DIFF
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -23,26 +23,18 @@ ORDER_LINE_COUNT = 1
 @receiver(post_checkout, dispatch_uid='tracking.post_checkout_callback')
 @silence_exceptions('Failed to emit tracking event upon order completion.')
 def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unused-argument
-    """Emit a tracking event when an order is placed."""
+    """
+    Emit a tracking event when
+    1. An order is placed OR
+    2. An enrollment code purchase order is placed.
+    """
     if order.total_excl_tax <= 0:
         return
-
-    for line in order.lines.all():
-        if line.product.is_coupon_product or line.product.is_enrollment_code_product:
-            return
-
-    voucher = order.basket_discounts.filter(voucher_id__isnull=False).first()
-    coupon = voucher.voucher_code if voucher else None
-    try:
-        bundle_id = BasketAttribute.objects.get(basket=order.basket, attribute_type__name=BUNDLE).value_text
-    except BasketAttribute.DoesNotExist:
-        bundle_id = None
 
     properties = {
         'orderId': order.number,
         'total': str(order.total_excl_tax),
         'currency': order.currency,
-        'coupon': coupon,
         'discount': str(order.total_discount_incl_tax),
         'products': [
             {
@@ -59,6 +51,25 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
             } for line in order.lines.all()
         ],
     }
+
+    for line in order.lines.all():
+        if line.product.is_enrollment_code_product:
+            # Send analytics events to track bulk enrollment code purchases.
+            track_segment_event(order.site, order.user, 'Bulk Enrollment Codes Order Completed', properties)
+            return
+
+        if line.product.is_coupon_product:
+            return
+
+    voucher = order.basket_discounts.filter(voucher_id__isnull=False).first()
+    coupon = voucher.voucher_code if voucher else None
+
+    try:
+        bundle_id = BasketAttribute.objects.get(basket=order.basket, attribute_type__name=BUNDLE).value_text
+    except BasketAttribute.DoesNotExist:
+        bundle_id = None
+
+    properties['coupon'] = coupon
 
     try:
         bundle_id = BasketAttribute.objects.get(basket=order.basket, attribute_type__name=BUNDLE).value_text
--- a/ecommerce/extensions/checkout/tests/test_signals.py
+++ b/ecommerce/extensions/checkout/tests/test_signals.py
@@ -286,7 +286,7 @@ class SignalTests(ProgramTestMixin, CouponMixin, TestCase):
             assert not mock_track.called
 
     def test_track_completed_enrollment_order(self):
-        """ Make sure we do not send GA events for Enrollment Code orders """
+        """ Make sure we are sending GA events for Enrollment Code orders """
         with mock.patch('ecommerce.extensions.checkout.signals.track_segment_event') as mock_track:
 
             course = CourseFactory()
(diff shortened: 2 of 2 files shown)


ecommerce/extensions/checkout/signals.py
```diff
@@ -60,6 +48,23 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
         ],
     }
 
+    for line in order.lines.all():
+        if line.product.is_enrollment_code_product:
+            track_segment_event(order.site, order.user, 'Bulk Enrollment Codes Purchase Completed', properties)
```
robrap: Some thoughts...
1. If the line.product.is_coupon_product appears earlier than a line.product.is_enrollment_code_product, this event will never happen.  If it appears afterward, it will happen.  I'm not sure which is correct, but we may want to document the relationship more clearly.
2. It would be cleaner to return here, rather than in a separate if that comes later.
3. The doc string of this method may want to detail the two cases?
4. Maybe the name ''Bulk Enrollment Codes Order Completed' would be more consistent with 'Order Completed'?
Thanks.

saleem-latif: @robrap I have updated the docstring and the event message, please take another look. And for

> If the line.product.is_coupon_product appears earlier than a line.product.is_enrollment_code_product, this event will never happen. If it appears afterward, it will happen. I'm not sure which is correct, but we may want to document the relationship more clearly.

We only want to track events with enrollment code purchase, that is why I have placed `line.product.is_enrollment_code_product` first to ensure that these events are logged.

> It would be cleaner to return here, rather than in a separate if that comes later.

Most part of the changes in the PR is just a refactor to avoid code duplication, only addition here is this conditions.
```
        if line.product.is_enrollment_code_product:
            # Send analytics events to track bulk enrollment code purchases.
            track_segment_event(order.site, order.user, 'Bulk Enrollment Codes Order Completed', properties)
```
I could not merge the following condition with this as the next condition contains an `OR` and not an `AND` for these to be combined somehow.

robrap: You have the following code:
```
if line.product.is_enrollment_code_product:
    track_segment_event(order.site, order.user, 'Bulk Enrollment Codes Purchase Completed', properties)

if line.product.is_coupon_product or line.product.is_enrollment_code_product:
    return
```

I was recommending that the following simplification would make it more obvious that the function is done after tracking the event:
```
if line.product.is_enrollment_code_product:
    track_segment_event(order.site, order.user, 'Bulk Enrollment Codes Purchase Completed', properties)
    return

if line.product.is_coupon_product:
    return
```

As mentioned, it is still unclear to me what assumptions we are making about all the lines/products.  For example, it is unclear why it would be true that any basket with a `line.product.is_coupon_product` does not need a `Order Completed` event, but only if it is not a `line.product.is_enrollment_code_product`.

saleem-latif: Done, and regarding

>As mentioned, it is still unclear to me what assumptions we are making about all the lines/products. For example, it is unclear why it would be true that any basket with a line.product.is_coupon_product does not need a Order Completed event, but only if it is not a line.product.is_enrollment_code_product.

This logic was added as part of https://github.com/edx/ecommerce/pull/1408 , I am not sure why it was needed that is why I did not alter it.

robrap: I understand that you did not write the is_coupon_product code. However, that code was written to return and never trigger an Order Complete event. 

The way this is coded, it may now trigger the new Bundle Order Complete event, even though their was an earlier decision not to for coupon products.

Additionally, because this occurs within a list of products, if a coupon product is found first it won’t trigger the new event, but if an enrollment code product his hit first, then it will trigger your new event.

I recommend getting someone from Product who can clarify the requirements so the code can more clearly reflect whatever we actually want to happen. For example, It may be that in the loop we just capture whether there is any coupon product and separately whether there is any enrollment product. Then, after the loop, you could more clearly code what you want to happen when and if both are true.

saleem-latif: @robrap let me know if https://openedx.atlassian.net/browse/ENT-967?focusedCommentId=320387&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-320387 answers yout question.

saleem-latif:
@robrap please take another look.

ecommerce/extensions/checkout/signals.py
robrap: @georgebabey @adamstankiewicz: Can either of you provide direction here?  The ticket provided by @saleem-latif does not answer my questions, and I don't have the full context to be a final reviewer here.  I was just giving my opinion on what was clear and what was not clear.  At this point, it needs someone with more context to help clarify what needs to be done here.  I'm not sure if that would be one of you or someone else.

Thank you.

adamstankiewicz: @robrap If an order contains an enrollment code product, it will be the only product in the list of products when iterating through `order.lines.all()`. There won't be a situation in which there is both a coupon product and an enrollment code product in the basket, so it should be fine if the iteration is ended after finding an enrollment code product and tracking the event. Does that help clarify your question?

Would something like this make things clearer (based on your suggestion above)?:

```python
for line in order.lines.all():
    if line.product.is_enrollment_code_product:
        has_enrollment_code_product = True
    if line.product.is_coupon_product:
        has_coupon_product = True

if has_enrollment_code_product:
    # Send analytics event to track bulk enrollment code purchase
    track_segment_event(order.site, order.user, 'Bulk Enrollment Codes Order Completed', properties)
    return

if has_coupon_product:
    # Don't send analytics event for coupon purchase
    return
```

zubair-arbi (approved):
LGTM 👍

ecommerce/extensions/checkout/signals.py
georgebabey: @saleem-latif @robrap Coming late to this one but its clear there are 2 confusing (and potentially conflicting) requirements at play.

- Based on https://openedx.atlassian.net/browse/LEARNER-2014 it looks like the coupon redemptions were throwing off GA numbers. Unfortunately we don't have any more context than that as the ticket lacks details. Therefore we should keep that same logic, which this PR does.
- We have new requirements from product for bulk purchase to track these in GA through the sales funnel. 

As Adam mentions, current support for bulk enrollment codes is one-per-order only. It is a good call-out however and its interesting that even prior to bulk enrollment it was somewhat indeterministic in that a coupon code line item first would immediately return. This PR doesn't change that behavior as its unclear how that tracking is used.

georgebabey (approved):
Needs a rebase but looks good

PR #1760, opened 2018-05-11
fix credit card test

raeeschachar:


raeeschachar:
@robrap @staubina 
The sku that test was hitting https://ecommerce.stage.edx.org/basket/add/?sku=A9213AB was displaying 'User has already purchased the course', well that was a valid issue. Checked it in ecommerce, refund had a fulfilment error. Fixed that and tests are now passing. This PR is to fix the doc string as @robrap mentioned.

robrap (approved):
Thanks. Does anyone know if this is a flag that can be retired?

Also, for simple changes like this, feel free to just tag me as FYI, but no need to wait on my review.  Thank you!

raeeschachar:
@robrap 
I am not sure, I think @Ayub-Khan has a better idea on this but he is on vacations. I can discuss with him as soon. 
Acknowledged on the second part of the comment.

PR #1761, opened 2018-05-11
LEARNER-5182: Refactor tiered cache.

robrap:
[LEARNER-5182](https://openedx.atlassian.net/browse/LEARNER-5182)

Refactor of tiered cache based on learnings from applying to the rest of ecommerce (see upcoming PR).

Also, includes removal of waffle switch `use_request_cache_for_getting_lms_resource`.

NOTE: @nasthagiri came upon https://github.com/dimagi/quickcache which is a possible replacement for most of this, with the following caveats:
1. We still need to implement a RequestCache, but it would need to be a complete Django Cache.
2. Quick cache documents being automatically versioned by the source it wraps.  Could be good in many cases.  May not be great if you want to do something like add logging to a call that is expensive.  Something to keep an eye on.
3. Are we ok with the key generation code in quick cache?  The current implementation punts on this issue by just accepting keys.
4. Would we want to enforce that the Django caches are never called directly?  This implementation fixes a common bug that appears when caching Falsey values like `None`.  The quick cache gets around this by avoiding calls to get/set, by instead using decorators only.

### DIFF
--- a/ecommerce/cache_utils/utils.py
+++ b/ecommerce/cache_utils/utils.py
@@ -6,6 +6,11 @@ import threading
 from django.core.cache import cache as django_cache
 from django.core.cache.backends.base import DEFAULT_TIMEOUT
 
+SHOULD_FORCE_CACHE_MISS_KEY = 'cache_utils.should_force_cache_miss'
+FORCE_CACHE_MISS_PARAM = 'force_cache_miss'
+
+_CACHE_MISS = object()
+
 
 class RequestCache(threading.local):
     """
@@ -22,38 +27,20 @@ class RequestCache(threading.local):
         cls._data = {}
 
     @classmethod
-    def get_value_or_cache_miss(cls, key):
+    def get_cache_response(cls, key):
         """
-        Retrieves a cached value from the request cache or CACHE_MISS.
-
-        See README for more details.
+        Retrieves a CacheResponse for the provided key.
 
         Args:
             key (string)
 
         Returns:
-            The value associated with key, or the CACHE_MISS object.
+            A CacheResponse with hit/miss status and value.
 
         """
-        return cls._data.get(key, CACHE_MISS)
-
-    @classmethod
-    def get_value_or_default(cls, key, default):
-        """
-        Retrieves a cached value from the request cache or the default argument
-
-        Note: If you need to know if you got a cache miss, use
-        get_value_or_cache_miss instead.
-
-        Args:
-            key (string)
-            default (object)
-
-        Returns:
-            The value associated with key, or the default argument.
-
-        """
-        return cls._data.get(key, default)
+        cached_value = cls._data.get(key, _CACHE_MISS)
+        is_miss = cached_value is _CACHE_MISS
+        return CacheResponse(is_miss, cached_value)
 
     @classmethod
     def set(cls, key, value):
@@ -63,8 +50,21 @@ class RequestCache(threading.local):
         Args:
             key (string)
             value (object)
+
         """
         cls._data[key] = value
+
+    @classmethod
+    def delete(cls, key):
+        """
+        Deletes the cached value for the provided key.
+
+        Args:
+            key (string)
+
+        """
+        if key in cls._data:
+            del cls._data[key]
 
 
 class TieredCache(object):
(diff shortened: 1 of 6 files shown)


ecommerce/programs/conditions.py
```diff
@@ -53,23 +51,19 @@ def _get_lms_resource_for_user(self, basket, resource_name, endpoint):
...
-            except (ConnectionError, SlumberBaseException, Timeout) as exc:
-                logger.error('Failed to retrieve %s : %s', resource_name, str(exc))
-        return data_list if data_list else []
+        cache_response = TieredCache.get_cache_response(cache_key)
+        if cache_response.is_hit:
+            return cache_response.value
+
+        user = basket.owner.username
+        try:
+            data_list = endpoint.get(user=user)
```
robrap: `data_list = endpoint.get(user=user) or []`

ecommerce/cache_utils/utils.py
```diff
@@ -136,96 +110,139 @@ def set_all_tiers(cls, key, value, django_cache_timeout=DEFAULT_TIMEOUT):
...
-    def _get_from_django_cache_and_set_request_cache(cls, key):
+    def _get_cache_response_from_django_cache(cls, key):
         """
-        Retrieves a value from the django cache and sets it in the request
-        cache (if not a CACHE_MISS).
+        Retrieves a CacheResponse for the given key from the django cache.
 
         If the request was set to force cache misses, then this will always
-        return CACHE_MISS.
+        return a cache miss: (True, None).
```
robrap: Fix docstring.

ecommerce/cache_utils/utils.py
```diff
@@ -136,96 +110,139 @@ def set_all_tiers(cls, key, value, django_cache_timeout=DEFAULT_TIMEOUT):
...
-        value = django_cache.get(key, CACHE_MISS)
-        if value is not CACHE_MISS:
-            RequestCache.set(key, value)
-        return value
+        cached_value = django_cache.get(key, _CACHE_MISS)
+        is_miss = cached_value is _CACHE_MISS
+        return CacheResponse(is_miss, cached_value)
+
+    @classmethod
+    def _set_request_cache_for_django_cache_hit(cls, key, django_cache_response):
```
robrap: if_hit

ecommerce/cache_utils/utils.py
```diff
@@ -136,96 +110,139 @@ def set_all_tiers(cls, key, value, django_cache_timeout=DEFAULT_TIMEOUT):
...
+            http://clobert.com/api/v1/resource?force_django_cache_miss=true
+
+        """
+        force_django_cache_miss = request.GET.get(FORCE_DJANGO_CACHE_MISS_KEY, 'false').lower() == 'true'
+        RequestCache.set(FORCE_DJANGO_CACHE_MISS_KEY, force_django_cache_miss)
 
     @classmethod
     def _force_django_cache_miss(cls):
-        return RequestCache.get_value_or_default('force_django_cache_miss', False)
+        cache_response = RequestCache.get_cache_response(FORCE_DJANGO_CACHE_MISS_KEY)
```
robrap: function name: `_should_force_cache_miss`
query param: `force_cache_miss`
Key: `SHOULD_FORCE_CACHE_MISS_KEY`

ecommerce/cache_utils/utils.py
```diff
@@ -136,96 +110,139 @@ def set_all_tiers(cls, key, value, django_cache_timeout=DEFAULT_TIMEOUT):
...
 
-class CacheMissError(Exception):
+class CacheResponseError(Exception):
     """
-    An error used when the CACHE_MISS object is misused in any context other
-    than checking if it is the CACHE_MISS object.
+    Error used when CacheResponse is misused.
     """
-    USAGE_MESSAGE = 'Proper Usage: "if value is CACHE_MISS: value = DEFAULT; ...".'
+    USAGE_MESSAGE = 'CacheResponse was misused. Only the use the attributes is_hit (or is_miss) and value.'
```
robrap: Oops.

robrap (commented):
Comments from pairing review with @nasthagiri.

PR #1764, opened 2018-05-15
LEARNER-5182: Reuse TieredCache.

robrap:
LEARNER-5182

@nasthagiri: FYI: This is the larger refactor of Ecommerce.  Let me know if you think we want to dedicate my initial time on Arch Squad to trying to use quickcache and skipping this step.  Or, if as discussed, we should land this and return to this when and if we wish to prioritize.

### DIFF
--- a/ecommerce/coupons/utils.py
+++ b/ecommerce/coupons/utils.py
@@ -3,10 +3,10 @@ import hashlib
 import logging
 
 from django.conf import settings
-from django.core.cache import cache
 from oscar.core.loading import get_model
 from slumber.exceptions import HttpNotFoundError
 
+from ecommerce.cache_utils.utils import TieredCache
 from ecommerce.core.utils import get_cache_key
 
 Product = get_model('catalogue', 'Product')
@@ -66,18 +66,20 @@ def get_catalog_course_runs(site, query, limit=None, offset=None):
     )
     cache_key = hashlib.md5(cache_key).hexdigest()
 
-    response = cache.get(cache_key)
-    if not response:
-        api = site.siteconfiguration.discovery_api_client
-        endpoint = getattr(api, api_resource_name)
-        response = endpoint().get(
-            partner=partner_code,
-            q=query,
-            limit=limit,
-            offset=offset
-        )
-        cache.set(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
+    cached_response = TieredCache.get_cached_response(cache_key)
+    if cached_response.is_hit:
+        return cached_response.value
 
+    api = site.siteconfiguration.discovery_api_client
+    endpoint = getattr(api, api_resource_name)
+
+    response = endpoint().get(
+        partner=partner_code,
+        q=query,
+        limit=limit,
+        offset=offset
+    )
+    TieredCache.set_all_tiers(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
     return response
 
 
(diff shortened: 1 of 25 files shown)


ecommerce/coupons/utils.py
```diff
@@ -161,8 +162,10 @@ def fetch_course_catalog(site, catalog_id):
             logger.exception("Catalog '%s' not found.", catalog_id)
             raise
 
-    cache.set(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
```
robrap: Note: No longer setting cache on cache hits.

ecommerce/courses/tests/test_utils.py
```diff
@@ -73,8 +73,8 @@ def test_get_course_run_info_from_catalog(self, course_run):
         else:
             self.assertEqual(response['title'], product.title)
 
-        cached_course = cache.get(cache_key)
-        self.assertEqual(cached_course, response)
+        cached_course = TieredCache.get_cache_response(cache_key)
```
BenjiLee: `get_cache_response`-> `get_cached_response`

BenjiLee: `cached_course` -> `course_cached_response`

BenjiLee:
@robrap Currently looking at the enterprise/api codcov

ecommerce/enterprise/tests/test_api.py
```diff
@@ -145,6 +146,20 @@ def test_catalog_contains_course_runs(self, expected, enterprise_customer_catalo
...
+        Verify `catalog_contains_course_runs` returns a cached response
+        """
+        self.mock_catalog_contains_course_runs(
+            [self.course_run.id],
+            'fake-uuid',
+            enterprise_customer_catalog_uuid=None,
+            contains_content=True,
+        )
+        self._assert_contains_course_runs(True, [self.course_run.id], 'fake-uuid', None)
+        self._assert_contains_course_runs(True, [self.course_run.id], 'fake-uuid', None)
```
BenjiLee: @robrap Do you know of any good way to explicitly confirm a cache hit? 

This is the line this test is covering https://github.com/edx/ecommerce/blob/b621a2dc8317acdb0a1c8ea8ab68afec17aca162/ecommerce/enterprise/api.py#L192-L193

robrap: Not sure of the best way.  Some ideas:
1. Split method into _get_cache_X() and _get_X().  Then mock and check if _get_X() is called.
2. Spy on the method and see what it is doing.  Spying is when you let it do its thing, but you watch it do its thing.  At least in Jasmine that is what it is called I think.  Here is an example of spying I found with mocks:
http://wesmckinney.com/blog/spying-with-python-mocks/
You can see that it is setting "wraps" for the same thing you are patching.  Hopefully that would both make the real call and allow you to spy on its return value (somehow) and see that it returned is_hit.

ecommerce/enterprise/tests/test_api.py
```diff
@@ -145,6 +146,19 @@ def test_catalog_contains_course_runs(self, expected, enterprise_customer_catalo
...
```
BenjiLee: Use the inner function `catalog_contains_course_runs`

ecommerce/courses/tests/test_views.py
```diff
@@ -128,6 +130,27 @@ def test_credit_providers_in_context(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.context['credit_providers'], provider_json)
 
+    @httpretty.activate
+    def test_credit_providers_in_context_cached(self):
+
```
jmyatt: TODO: remove this extra whitespace line

BenjiLee (approved):

PR #1765, opened 2018-05-16
LEARNER-5323: Add logging for merge basket.

robrap:
LEARNER-5323

### DIFF
--- a/ecommerce/extensions/basket/middleware.py
+++ b/ecommerce/extensions/basket/middleware.py
@@ -1,7 +1,11 @@
+import logging
+
 from oscar.apps.basket.middleware import BasketMiddleware as OscarBasketMiddleware
 from oscar.core.loading import get_model
 
 Basket = get_model('basket', 'basket')
+
+logger = logging.getLogger(__name__)
 
 
 class BasketMiddleware(OscarBasketMiddleware):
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/basket/middleware.py
```diff
@@ -42,6 +46,10 @@ def get_basket(self, request):
                 basket = old_baskets[0]
                 for other_basket in old_baskets[1:]:
                     self.merge_baskets(basket, other_basket)
+                logger.warning("Merging multiple baskets for user={username} and site={site}.".format(
+                    username=request.user.username,
+                    site=request.site,
+                ))
```
robrap, line 52: @BenjiLee: What is a "cookie basket", and do we need to be concerned that it is also merged below?

BenjiLee: wgx772

robrap: On second thought, maybe we shouldn't throw away the basket.

BenjiLee (approved):

robrap:
@BenjiLee: I am going to close this for now.  I'd like to think this through a bit more with you to determine what we need, and why the test failed.  We can reopen when we get to this.

PR #1772, opened 2018-05-21
Remove unecessary logic in coupon utils

BenjiLee:
This PR unblocks codecov requirements in https://github.com/edx/ecommerce/pull/1764

Remove unnecessary handling of the case where limit == False. DRF already handles all different inputs for limit. 

For each example below, DRF will just use the default limit. 
`www.thing.com?limit=` 
`www.thing.com?limit=gwesgw`
`www.thing.com?limit=None`
`www.thing.com?limit=False`
`www.thing.com?limit=0`

The other possible use case is that someone would want a response that has all of the data in one giant call. For example, instead of 
```
{
    'count': 100,
    'next': 'www.page2of5.com',
    'previous': 'None',
    'results': [
        1,
        2,
        ...
        20
    ]
}
```
A caller may want
```
{
    'count': 100,
    'next': 'None',
    'previous': 'None',
    'results': [
        1,
        2,
        ...
        100
    ]
}
```

This seems like an odd and dangerous use case since the latter response traverses the entire course_runs catalog. There are only two places that this method is called and both callers set the limit with...
`limit = request.GET.get('limit', DEFAULT_CATALOG_PAGE_SIZE)` where `DEFAULT_CATALOG_PAGE_SIZE` is set to 100.

### DIFF
--- a/ecommerce/coupons/utils.py
+++ b/ecommerce/coupons/utils.py
@@ -7,7 +7,7 @@ from django.core.cache import cache
 from oscar.core.loading import get_model
 from slumber.exceptions import HttpNotFoundError
 
-from ecommerce.core.utils import get_cache_key, traverse_pagination
+from ecommerce.core.utils import get_cache_key
 
 Product = get_model('catalogue', 'Product')
 
@@ -70,27 +70,12 @@ def get_catalog_course_runs(site, query, limit=None, offset=None):
     if not response:
         api = site.siteconfiguration.discovery_api_client
         endpoint = getattr(api, api_resource_name)
-
-        if limit:
-            response = endpoint().get(
-                partner=partner_code,
-                q=query,
-                limit=limit,
-                offset=offset
-            )
-        else:
-            response = endpoint().get(
-                partner=partner_code,
-                q=query
-            )
-            all_response_results = traverse_pagination(response, endpoint)
-            response = {
-                'count': len(all_response_results),
-                'next': 'None',
-                'previous': 'None',
-                'results': all_response_results,
-            }
-
+        response = endpoint().get(
+            partner=partner_code,
+            q=query,
+            limit=limit,
+            offset=offset
+        )
         cache.set(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
 
     return response


robrap (commented):
@BenjiLee: Let's see if we can get @mattdrayer to review since he helped review (the PR that added this)[https://github.com/edx/ecommerce/pull/1104/files].

I agree that the method `traverse_pagination` should probably be retired, because it basically negates the point of pagination.  It may sometimes be used in places where we don't have a better filter for an API, and we should probably add that where needed rather than just paginating.

NOTE: If we think it should be retired should we mark it as deprecated with a note like the one above?

BenjiLee:
@robrap Traverse pagination is still being used in other places so I'm don't think we should deprecate it. In `get_catalog_course_runs`, traverse_pagination isn't being reached at all for all cases.

robrap:
Is `traverse_pagination` ever the proper solution in any of the cases it is used? Is it just a way of saying: “hey endpoint, you don’t have the options I need so I’m just going to abuse you and blast through all your pages.” If it has proper uses we may want to document when to use it and how to avoid it. If it has no proper uses, we deprecate it so people don’t keep using it and we state what they should do instead.

robrap:
Note: this isn’t a blocker for this PR, but highly related.

BenjiLee:
:+1: for out of scope. 
:+1: for deprecating traverse pagination. I agree we shouldn't be using it like this. I wonder if there are any excessive calls being made for this. Maybe we could add some logging to see how often this is abused.

mattdrayer:
Pulling in @georgebabey @zubair-arbi and @saleem-latif to comment on the need for `traverse_pagination` -- I believe the system is using this function in order to combine chunks of metadata into singular sets for comparison and reporting purposes.

mattdrayer:
Note that I'm alright marking the function as deprecated, as long as it's not removed until an alternative solution is implemented.

robrap:
@BenjiLee: Does this need any logging ahead of time to ensure the deleted code isn't used?

zubair-arbi:
@mattdrayer `traverse_pagination` is used for getting all results in one call, i.e. showing all catalog on creating a catalog coupon, getting offers from a catalog query other then these 2 cases I don't think we are using `traverse_pagination`.
screen shot 2018-05-23 at 1 45 42 pm

robrap (approved):
@BenjiLee: Thank you.  Feel free to ignore all my deprecation comments as part of this PR.  However, I do want to know if we should ensure this is safe to remove via logging, or if you feel fully confident this can be removed.

Thanks @zubair-arbi and @mattdrayer.  I don't think `traverse_pagination` is a pattern we should follow, so I think we should deprecate it (not remove it) until we have another solution and then remove it.  I don't want to see it continue to get used, and I think it was used again recently for entitlements.  See usages here: https://github.com/edx/ecommerce/search?utf8=%E2%9C%93&q=traverse_pagination&type=

I imagine renaming it to `deprecated_traverse_pagination` with a clear comment.  However, more discovery is needed to be sure and to craft a clear comment, so this would be a separate effort.

PR #1775, opened 2018-05-22
Refactored code and added test for default_site.

waheedahmed:
Refactored code for create or update site command and added test case for default site.

### DIFF
--- a/ecommerce/core/management/commands/create_or_update_site.py
+++ b/ecommerce/core/management/commands/create_or_update_site.py
@@ -164,9 +164,10 @@ class Command(BaseCommand):
         if partner_created:
             partner.name = partner_name
             partner.short_code = partner_code
-            partner.default_site = site
-            partner.save()
             logger.info('Partner created with code %s', partner_code)
+
+        partner.default_site = site
+        partner.save()
 
         site_configuration_defaults = {
             'partner': partner,
--- a/ecommerce/core/tests/test_commands.py
+++ b/ecommerce/core/tests/test_commands.py
@@ -35,6 +35,7 @@ class CreateOrUpdateSiteCommandTests(TestCase):
         site_configuration = site.siteconfiguration
         self.assertEqual(site_configuration.site, site)
         self.assertEqual(site_configuration.partner, partner)
+        self.assertEqual(site_configuration.partner.default_site, site)
         self.assertEqual(site_configuration.lms_url_root, self.lms_url_root)
         self.assertEqual(site_configuration.payment_processors, self.payment_processors)
         self.assertEqual(site_configuration.client_side_payment_processor, self.client_side_payment_processor)


waheedahmed:
@robrap Added the assertion kindly review.

CC: @staubina

ecommerce/core/management/commands/create_or_update_site.py
```diff
@@ -164,10 +164,13 @@ def handle(self, *args, **options):
         if partner_created:
             partner.name = partner_name
             partner.short_code = partner_code
-            partner.default_site = site
             partner.save()
             logger.info('Partner created with code %s', partner_code)
 
+        if partner.default_site != site:
+            partner.default_site = site
```
robrap: When `partner_created` is True, does this still need to set the site?  Is this going to result in two saves?  I'm torn, because two saves seems less ideal but I like that this code is very clean.

Options seem to be:
1. Leave as-is with repeated save().
2. Replace `partner.save()` with `is_save_needed = True` and then later save if `is_save_needed`.
3. Change the second if to elif and repeat the line `partner.default_site = site` twice (inside the if and inside the elif).

To be clear: I'm not sure what is best just just giving some options for you to consider.

ecommerce/core/tests/test_commands.py
```diff
@@ -35,6 +35,7 @@ def _check_site_configuration(self, site, partner):
         site_configuration = site.siteconfiguration
         self.assertEqual(site_configuration.site, site)
         self.assertEqual(site_configuration.partner, partner)
+        self.assertEqual(site_configuration.partner.default_site, site)
```
robrap, line 38: Great.  Thanks.  Did this turn up a bug when `partner_created` was False?

robrap (approved):
I gave some potential changes, but not sure what is best.  So I'm approving and letting you decide.  Let me know if any of the options aren't clear.

Thank you!

robrap:
It looks like codecov wants to see a test for `if partner.default_site != site:` when it is False.

ecommerce/core/tests/test_commands.py
waheedahmed: @robrap Yes.

waheedahmed:
@robrap removing this `if partner.default_site != site:` and moving save after `partner.default_site = site` outside if statement because it is always going to be null and we have to assign it a site.

robrap:
@waheedahmed: that’s for create, but not necessarily update, right? I’m on phone, so I’m just going by the function name.

waheedahmed:
@robrap Yes, but we have to update default site every time even it's created or not.

waheedahmed:
@robrap Can I merge it now?

robrap (commented):
Thank you!

robrap:
@waheedahmed: I spoke to @feanil, and I now have a better understanding of why default_site was added, but I don't think it is the right way to go. Sorry I didn't really get it beforehand.  Maybe I don't get it now?  Anyway, I will explain and you can help me understand.

The major change that was needed was to make it so that a Partner could be reused for many Site Configurations.  When it was one-to-one, there were lots of simplified assumptions that could be made that no longer can be made, and I think this may require a little thought and digging to get to a more complete solution.

One issue is that with a 1-to-1 mapping, it might have allowed us to add what really should have been "Partner Configuration" into the "Site Configuration" by mistake.  So the first question would be whether there are fields in the Site Configuration that really should live in the Partner, and that should equally apply to ALL site configurations associated with that Partner.  Each of these could be detailed and moved to its rightful place.

Next, once this is done, for any management command that only took a Partner, we need to determine the following:
1. Does it now have everything it needs once all partner specific fields are moved from Site Configuration to Partner?
2. If it doesn't, does the command simply need more arguments to target the correct relationship and objects?

Note: @feanil stated that he would provide guidance on the right way to move data around and retire unused columns.

I will work with @staubina separately to determine resourcing for the above work.  At this time, I'm not asking for your to do anything.  I just want you to be aware of my thinking.  Please feel free to let me know if you think I am missing anything big here.

Thank you again for all of your work on this (and everything).  I've got a great deal of appreciation for all your experience and efforts.

waheedahmed:
@robrap Moving the fields to partner which can be applied to multiple site configurations associated with that partner is good idea, the main reason we added `default_site` in partner was for management commands because after changing the relationship of partner and site configuration we cannot get the site object directly from partner.

I think we can remove `default_site` from partner and pass it as an argument to those commands which require it.

robrap:
Thanks @waheedahmed. When you get to it, maybe you could document the fields that should be partner fields? Also, we can document the affected management commands, and see if any only require the partner at that point and which will require both. Thank you.

PR #1777, opened 2018-05-24
WL-1575: Fixed management command to handle a scenario when dns is changed

ziafazal:
This PR has changes to handle a scenario when site domain of and existing partner is changed. These changes would allow us to update site instead of creating new one and then causing `IntegrityError` when we created site configuration.
@hasnain-naveed could you please review.

### DIFF
--- a/ecommerce/theming/management/commands/create_sites_and_partners.py
+++ b/ecommerce/theming/management/commands/create_sites_and_partners.py
@@ -59,18 +59,7 @@ class Command(BaseCommand):
         """
         Create Sites, SiteThemes, SiteConfigurations, and Courses (if requested)
         """
-        site, _ = Site.objects.get_or_create(
-            domain=site_domain,
-            defaults={"name": theme_dir_name}
-        )
-
-        logger.info('Creating %s SiteTheme', site_domain)
-        SiteTheme.objects.get_or_create(
-            site=site,
-            theme_dir_name=theme_dir_name
-        )
-
-        logger.info('Creating %s Partner', site_domain)
+        logger.info('Creating or updating %s Partner', partner_code)
         partner, _ = Partner.objects.get_or_create(
             short_code=partner_code,
             defaults={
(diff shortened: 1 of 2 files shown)


hasnain-naveed:
@ziafazal can you paste the link of sandbox here? With this fix.

ziafazal:
@hasnain-naveed here are the links of sandbox
https://wl-1572.sandbox.edx.org/
https://mitxpro-wl-1572.sandbox.edx.org/

ecommerce/theming/management/commands/create_sites_and_partners.py
```diff
@@ -59,26 +59,33 @@ def _create_sites(self, site_domain, theme_dir_name, site_configuration, partner
         """
         Create Sites, SiteThemes, SiteConfigurations, and Courses (if requested)
```
robrap, line 60: Should include Partner.

ecommerce/theming/management/commands/create_sites_and_partners.py
```diff
@@ -59,26 +59,33 @@ def _create_sites(self, site_domain, theme_dir_name, site_configuration, partner
...
         """
-        site, _ = Site.objects.get_or_create(
-            domain=site_domain,
-            defaults={"name": theme_dir_name}
+        logger.info('Creating or updating %s Partner', partner_code)
```
robrap, line 62: `Creating or retrieving...` (not updating).

ecommerce/theming/management/commands/create_sites_and_partners.py
```diff
@@ -59,26 +59,33 @@ def _create_sites(self, site_domain, theme_dir_name, site_configuration, partner
...
+        partner, _ = Partner.objects.get_or_create(
+            short_code=partner_code,
+            defaults={
+                "name": partner_code
+            }
         )
 
+        logger.info('Finding if site already exists for %s partner', partner_code)
```
robrap, line 70: Imagine above we returned `was_partner_created`. We would have the following cases:
- `was_partner_created is True`: This is a new partner that would have no site, right?
- `was_partner_created is False`: This is an existing partner that may or may not have one or more associated sites, right? Do we need to find if there are any associated sites, and if one of them is the site with `site_domain`?

robrap, line 70: Additionally, if this try block survives, I'd probably remove the log message immediately above, and instead add one to the except case below.

ecommerce/theming/management/commands/create_sites_and_partners.py
```diff
@@ -59,26 +59,33 @@ def _create_sites(self, site_domain, theme_dir_name, site_configuration, partner
...
-        logger.info('Creating %s Partner', site_domain)
-        partner, _ = Partner.objects.get_or_create(
-            short_code=partner_code,
-            defaults={
-                "name": partner_code
-            }
-        )
-
-        logger.info('Creating %s SiteConfiguration', site_domain)
+        logger.info('Creating or updating %s SiteConfiguration', site_domain)
```
robrap, line 88: Can you add the partner to this log messages?

ecommerce/theming/management/commands/tests/test_create_sites_and_partners.py
```diff
@@ -160,3 +160,12 @@ def test_create_site_and_partner(self):
...
         # if we run command with same dns then it will not duplicates the sites and partners.
         self._assert_sites_data_is_valid()
+
+        self.dns_name = "new-dns"
+        call_command(
+            "create_sites_and_partners",
+            "--dns-name", self.dns_name,
+            "--theme-path", self.theme_path
+        )
+        # if we run command with same partner but different dns then it should update sites.
```
robrap, line 170: Is this really what we want?  How would you know when you want to update a site vs adding a new site?  Is this management command trying to do too much?

ziafazal:
@hasnain-naveed @robrap we don't need this work now. I'm close this in favor of #1756.

PR #1783, opened 2018-05-29
Add Make command to run vs check isort

BenjiLee:


Makefile
```diff
@@ -57,6 +57,9 @@ clean:
 clean_static:
 	rm -rf assets/* ecommerce/static/build/*
 
+run_check_isort:
+	VIRTUAL_ENV=/edx/app/ecommerce/ecommerce_env isort --check-only --recursive e2e/ ecommerce/
```
robrap: What's the difference between this and `run_isort` below?

BenjiLee: :D Great catch

robrap (commented):
I think the names are more clear, but you are changing the meaning of `run_isort` without changing its name.  That seems risky.  Also, what about other repos with isort integration (including the cookie cutter app)?  Should we have some consistency?

BenjiLee:
@robrap I always thought that the original `make isort` was misleading. `isort` by itself is supposed to actually sort things. Our case of `make isort` is only a check that is used by travis. This PR give Travis it's own make command for a non-default behaviour. I don't know if the other repos do anything similar but I'll poke around.

BenjiLee:
Discovery uses a `make quality` which makes sense to only run the checking of `isort`. Discovery also runs into the issue where if one of the tests fail (pylint, pep8, isort), it stops and fails the check without guaranteeing that all three checks will run.

robrap (approved):
Your names make sense.  I just wanted to make sure nothing would break and we'd keep consistency (which may have meant making your change in other places).  Thanks for checking.

robrap (approved):
Your names make sense.  I just wanted to make sure nothing would break and we'd keep consistency (which may have meant making your change in other places).  Thanks for checking.

PR #1784, opened 2018-05-30
WL-1558: Support Journal Images on Basket Page

bessiesteinberg:
- Journal products now show card image on basket page
- Created method to call journal discovery api and cache results

Remaining work:
- Handle journal bundle case (WL-1619)


ecommerce/core/models.py
```diff
@@ -398,6 +398,27 @@ def discovery_api_client(self):
...
+            split_url.scheme,
+            split_url.netloc,
+            JOURNALS_API_PATH,
+            split_url.query,
+            split_url.fragment
+        ])
+
+        return EdxRestApiClient(journal_discovery_url, jwt=self.access_token)
+
+
```
bill-filler: this seems to be duplicated? method already exists in this file

bill-filler:
@BessieSteinberg 
closing this PR as it's against our old branch. Needs to be re-created against master. Will add reference to it in https://openedx.atlassian.net/browse/WL-1558

PR #1785, opened 2018-05-31
Fix debugging text to give a string instead of Object

BenjiLee:
@jmyatt I'm accessing a protected class since `endpoint`, a Slumber Object, doesn't give me any good way to know what endpoint is being used. The alternative is to have any method that uses deprecate_traverse_pagination pass in an url identifying argument to be passed to the debugger.


jmyatt:
Yea I think this approach requiring the pylint disable is the best/simplest thing here, considering this code within the waffle switch block will likely be removed shortly after it has served its purpose

jmyatt (approved):

robrap (approved):

PR #1786, opened 2018-05-31
Remove analytics calls for temporary basket calculation

BenjiLee:
https://openedx.atlassian.net/browse/LEARNER-4968

### DIFF
--- a/ecommerce/extensions/basket/constants.py
+++ b/ecommerce/extensions/basket/constants.py
@@ -1,0 +1,1 @@
+is_calculate_temporary_basket = "Calculate temporary basket"
--- a/ecommerce/extensions/basket/models.py
+++ b/ecommerce/extensions/basket/models.py
@@ -3,7 +3,9 @@ from django.utils.translation import ugettext_lazy as _
 from oscar.apps.basket.abstract_models import AbstractBasket
 from oscar.core.loading import get_class
 
+from ecommerce.cache_utils.utils import RequestCache
 from ecommerce.extensions.analytics.utils import track_segment_event, translate_basket_line_for_segment
+from ecommerce.extensions.basket.constants import is_calculate_temporary_basket
 
 OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
 Selector = get_class('partner.strategy', 'Selector')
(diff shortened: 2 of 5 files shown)


ecommerce/extensions/basket/models.py
```diff
@@ -49,13 +51,14 @@ def get_basket(cls, user, site):
...
+        cached_response = RequestCache.get_cached_response(TEMPORARY_BASKET_CACHE_KEY)
+        if cached_response.is_miss:
+            for line in self.all_lines():
+                # Do not fire events for free items. The volume we see for edX.org leads to a dramatic increase in CPU
+                # usage. Given that orders for free items are ignored, there is no need for these events.
+                if line.stockrecord.price_excl_tax > 0:
+                    properties = translate_basket_line_for_segment(line)
+                    track_segment_event(self.site, self.owner, 'Product Removed', properties)
 
         super(Basket, self).flush()  # pylint: disable=bad-super-call
```
BenjiLee: I'd love to know why this was pylint ignored and at the bottom :/

ecommerce/extensions/basket/constants.py
```diff
@@ -0,0 +1 @@
+IS_TEMPORARY_BASKET_CACHE_KEY = "Calculate temporary basket"
```
robrap: Nit: Can we name this something more like: `ecommerce.is_calculate_temporary_basket`?

1. Namespace is not important for request cache, but it’s a good habit.
2. I’d rather the key look more like a key than a description.

ecommerce/extensions/basket/models.py
robrap: On your pylint note, are there any clues? Who wrote it? (Reviewing by phone.)

robrap (approved):
1. Can you add a note to JIRA to clean this up if we can stop using the basket for calculate?
2. Can you add an assertion that track_segment_event is never called? It should fail without your changes (temporary) and pas with your changes.

Approval conditional on minor requests.

ecommerce/extensions/basket/models.py
BenjiLee: https://github.com/edx/ecommerce/pull/1386 Everyone is gone >.< I think its safe to move it back up. It looks like she someone commented to put the super method back in, but never specified where, only to put it in.

robrap: Ok.

BenjiLee: nvm. we need this down there

BenjiLee:
This is the note for cleanup https://openedx.atlassian.net/browse/LEARNER-5463

ecommerce/extensions/basket/models.py
robrap: If you understand now, and you didn't before, does it warrant a comment to explain?

ecommerce/extensions/basket/constants.py
```diff
@@ -0,0 +1 @@
+is_calculate_temporary_basket = "Calculate temporary basket"
```
robrap, line 1: I'm sad this didn't change. :(

BenjiLee: Ah! I thought I made the change you wanted. I must have misunderstood

BenjiLee: > 2. I’d rather the key look more like a key than a description.

Looks like I didn't do this part.

> Nit: Can we name this something more like: ecommerce.is_calculate_temporary_basket?
 
and did this instead.

robrap: I meant:
```
is_calculate_temporary_basket = "ecommerce.is_calculate_temporary_basket"
```

PR #1787, opened 2018-06-01
Fix add voucher view to use multi use coupon for bundles.

waheedahmed:
LEARNER-5436

### DIFF
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -32,7 +32,7 @@ from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
 from ecommerce.entitlements.utils import create_or_update_course_entitlement
 from ecommerce.extensions.analytics.utils import translate_basket_line_for_segment
-from ecommerce.extensions.basket.utils import get_basket_switch_data
+from ecommerce.extensions.basket.utils import _set_basket_bundle_status, get_basket_switch_data
 from ecommerce.extensions.catalogue.tests.mixins import DiscoveryTestMixin
 from ecommerce.extensions.offer.utils import format_benefit_value
 from ecommerce.extensions.order.utils import UserAlreadyPlacedOrder
@@ -786,6 +786,17 @@ class VoucherAddViewTests(LmsApiMockMixin, TestCase):
             value_text='test_bundle'
         )
         self.assert_form_valid_message("Coupon code '{code}' is not valid for this basket.".format(code=voucher.code))
+
+    def test_multi_use_voucher_valid_for_bundle(self):
+        """ Verify multi use coupon works when voucher is used against a bundle. """
+        self.mock_access_token_response()
+        self.mock_account_api(self.request, self.user.username, data={'is_active': True})
+        voucher, product = prepare_voucher(code=COUPON_CODE, benefit_value=10, usage=Voucher.MULTI_USE)
+        new_product = factories.ProductFactory(categories=[], stockrecords__partner__short_code='second')
+        self.basket.add_product(product)
+        self.basket.add_product(new_product)
+        _set_basket_bundle_status('test-bundle', self.basket)
+        self.assert_form_valid_message("Coupon code '{code}' added to basket.".format(code=voucher.code))
 
     def test_form_valid_without_basket_id(self):
         """ Verify the view redirects to the basket summary view if the basket has no ID.  """
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -456,7 +456,10 @@ class VoucherAddView(BaseVoucherAddView):  # pylint: disable=function-redefined
             basket=self.request.basket,
             attribute_type=BasketAttributeType.objects.get(name=BUNDLE)
         )
-        if len(bundle_attribute) > 0 and not voucher.offers.first().condition.program_uuid:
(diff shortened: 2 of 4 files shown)


waheedahmed:
@edx/learner-spartans @staubina Please review.

staubina:
@robrap Please review this ticket.

staubina (approved):
The code changes seem straightforward and I remember these being made before and reverted until the other work was done.  

Please manually test the existing seat coupon types if you have not already.

ecommerce/extensions/basket/views.py
```diff
@@ -456,7 +456,8 @@ def apply_voucher_to_basket(self, voucher):
             basket=self.request.basket,
             attribute_type=BasketAttributeType.objects.get(name=BUNDLE)
         )
-        if len(bundle_attribute) > 0 and not voucher.offers.first().condition.program_uuid:
+        if (len(bundle_attribute) > 0 and
```
robrap: It took me a little time to decipher this.  Is the following refactor accurate?:
```
is_bundle_purchase = len(bundle_attribute) > 0
is_voucher_valid_for_bundle = voucher.offers.first().condition.program_uuid or voucher.usage == Voucher.MULTI_USE
is_voucher_invalid_for_bundle = not is_voucher_valid_for_bundle
if (is_bundle_purchase and is_voucher_invalid_for_bundle):
    ...
```

I'm not sure if these are used on other places, and if we may want to ultimately refactor to use functions.  That doesn't need to happen in this PR.

ecommerce/extensions/basket/tests/test_views.py
```diff
@@ -787,6 +787,22 @@ def test_voucher_not_valid_for_bundle(self):
...
+        new_product = factories.ProductFactory(categories=[], stockrecords__partner__short_code='second')
+        self.basket.add_product(product)
+        self.basket.add_product(new_product)
+        BasketAttributeType.objects.get_or_create(name=BUNDLE)
+        BasketAttribute.objects.update_or_create(
+            basket=self.basket,
+            attribute_type=BasketAttributeType.objects.get(name=BUNDLE),
+            value_text='test_bundle'
+        )
+        self.assert_form_valid_message("Coupon code '{code}' added to basket.".format(code=voucher.code))
```
robrap: Do we have separate tests to ensure that the coupon is handled properly (i.e., will be applied correctly to all courses and provide the correct final price)?

robrap (approved):
Potential changes noted, including a minor refactor.  We could refactor after and/or you could include some as part of this PR to help clarify the code.

Let me know if there is anything wrong with my suggested code.

robrap:
@waheedahmed @staubina: I just saw an email from marketing about 20% off coupon with fine print that it isn't applicable to bundles.  
1. Did they determine this isn't a blocker?
2. Did we ensure (through automated or manual test) that a 20% coupon against a bundle with this fix will apply the 20%, but not apply the bundle offer?

As noted above, although this is approved in order to unblock whatever was blocked, it would be great to get the minor code refactor in (assuming I didn't miscode anything).  It would also be good to determine the proper unit testing around multi-use coupons and bundles, so we don't need a manual test to determine that the bundle offer won't be applied when the multi-use coupon is applied.

waheedahmed:
@robrap I did the requested refactoring. I will try to test that locally if I'm unable to test @staubina can you please help me testing this?

ecommerce/extensions/basket/views.py
```diff
@@ -456,7 +456,10 @@ def apply_voucher_to_basket(self, voucher):
             basket=self.request.basket,
             attribute_type=BasketAttributeType.objects.get(name=BUNDLE)
         )
-        if len(bundle_attribute) > 0 and not voucher.offers.first().condition.program_uuid:
+        is_bundle_purchase = len(bundle_attribute) > 0
+        voucher_program_uuid = voucher.offers.first().condition.program_uuid
+        is_voucher_valid_for_bundle = voucher_program_uuid or voucher.usage == Voucher.MULTI_USE
+        if is_bundle_purchase and not is_voucher_valid_for_bundle:
```
robrap, line 462: Thanks.  This is much easier to understand.

waheedahmed:
@robrap I have added another test case and tried to cover your test scenario in it, please take a look and let me know if this looks good or not?

CC: @staubina

ecommerce/programs/tests/test_offer.py
```diff
@@ -68,3 +73,28 @@ def test_offer(self):
         self.assertEqual(basket.total_discount, Decimal(50) * len(lines))
         for line in lines:
             self.assertEqual(line.line_price_incl_tax_incl_discounts, 50)
+
```
robrap: I appreciate you writing this, but I’m thinking the assertions right before this code might be enough. I know it is testing Single Use (vs Multi Use), but I think it may be more of a voucher vs offer precedence question. If you are ok with dropping this test, I would do that. If not, let’s discuss.

Chris and I added a minor refactor that came from us trying to understand/clarify some other code.

robrap (approved):
After dropping duplicate test. See comment. Thank you.

ecommerce/programs/tests/test_offer.py
waheedahmed: @robrap First I thought same but then I thought we changed the view for multi use vouchers, that's why I added this test. I'm okay with dropping and adding the assertions above L73.

Thanks for the refactor.

waheedahmed:
@robrap I have removed the duplicate test and added the assertions above, also did some refactoring with the original test to add voucher using the API call instead of directly adding it into basket.

PR #1788, opened 2018-06-01
Fix logging for deprecated_traverse_pagination

BenjiLee:
deprecated_traverse_pagination is being called even if there are no pages to paginate to. Moved logger to log only when an external call is made.

### DIFF
--- a/ecommerce/core/utils.py
+++ b/ecommerce/core/utils.py
@@ -67,18 +67,17 @@ def deprecated_traverse_pagination(response, endpoint):
         list of dict.
 
     """
-    if waffle.switch_is_active("debug_logging_for_deprecated_traverse_pagination"):  # pragma: no cover
-        base_url = ""
-        try:
-            base_url = endpoint._store['base_url']  # pylint: disable=protected-access
-        except:  # pylint: disable=bare-except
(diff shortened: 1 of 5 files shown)


ecommerce/core/utils.py
```diff
@@ -67,18 +67,17 @@ def deprecated_traverse_pagination(response, endpoint):
...
-            base_url = endpoint._store['base_url']  # pylint: disable=protected-access
-        except:  # pylint: disable=bare-except
-            pass
-        logger.info("deprecated_traverse_pagination method is called for endpoint %s", base_url)
-
     results = response.get('results', [])
 
     next_page = response.get('next')
     while next_page:
+        if waffle.switch_is_active("debug_logging_for_deprecated_traverse_pagination"):  # pragma: no cover
```
robrap, line 74: Snuck this in too?

BenjiLee: I moved it. This PR was meant to fix the location of this logging code.

robrap: I guess it was the other change that snuck in. Either way, I was just joking. All is good. Thank you.

PR #1790, opened 2018-06-04
Journal Integration For Ecommerce

bessiesteinberg:
Journal Service Integration for Ecommerce.
For further information checkout the docs: ecommerce/ecommerce/journal/docs/journals.rst


robrap:
@BessieSteinberg: I understand this isn't ready fro review yet.  I want to preemptively ask that documentation in the form of .rst files (as needed) are added as part of this PR.  Thank you.

ecommerce/journal/forms.py
```diff
@@ -59,14 +59,14 @@ def __init__(self, data=None, files=None, auto_id='id_%s', prefix=None, initial=
                 'benefit_type': instance.benefit.proxy().benefit_class_type,
                 'benefit_value': instance.benefit.value,
             })
-        super(JournalOfferForm, self).__init__(data, files, auto_id, prefix, initial, error_class, label_suffix,
-                                               empty_permitted, instance)
+        super(JournalBundleOfferForm, self).__init__(data, files, auto_id, prefix, initial, error_class, label_suffix,
+                                                     empty_permitted, instance)
```
nedbat: If you don't indent to the opening paren, then lines like this don't have to move when you change the names of things:
```
super(JournalBundleOfferForm, self).__init__(
    data, files, auto_id, prefix, initial, error_class, label_suffix,
    empty_permitted, instance,
)
```

bessiesteinberg: I'll clean that up - Thanks!

ecommerce/core/management/commands/create_or_update_site.py
```diff
@@ -147,6 +154,8 @@ def handle(self, *args, **options):
         payment_support_url = options.get('payment_support_url', '')
         base_cookie_domain = options.get('base_cookie_domain', '')
         discovery_api_url = options.get('discovery_api_url')
+        # TODO: journal dependency
+        journals_api_url = options.get('journals_api_url', '')
```
jmyatt: it might make sense to leave out the default empty string value here, to keep consistent with how discovery_api_url and lms_url_root are handled

ecommerce/core/migrations/0046_merge_20180525_1418.py
```diff
@@ -0,0 +1,17 @@
...
+# TODO: journal dependency
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0045_siteconfiguration_journals_api_url'),
```
jmyatt: Instead of having two # 45 migrations, I think siteconfiguration_journals_api_url should be 0046 and merge_20180525_1418 should be 0047

ecommerce/core/tests/test_models.py
```diff
@@ -316,6 +317,25 @@ def test_enterprise_api_client(self):
...
+        REST API client for journal discovery service API
+        """
+        token = self.mock_access_token_response()
+        client = self.site.siteconfiguration.journal_discovery_api_client
+        client_store = client._store  # pylint: disable=protected-access
+        client_auth = client_store['session'].auth
+
+        self.assertEqual(
+            client_store['base_url'],
+            urljoin(self.site.siteconfiguration.discovery_api_url, '/journal/api/v1/')
```
jmyatt: better to import this api path from ecommerce.journal.constants

ecommerce/extensions/catalogue/migrations/0031_journal_product_class.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
```
jmyatt: it looks like there's another 0031 now, so after rebasing from master you'll need to regenerate or rename this one

ecommerce/extensions/catalogue/migrations/0031_journal_product_class.py
```diff
@@ -0,0 +1,62 @@
...
+    """ Reverse function. """
+    Product.objects.filter(product_class=ProductClass.objects.get(name=JOURNAL_PRODUCT_CLASS_NAME)).delete()
+    Category.objects.filter(slug=JOURNAL_SLUG_NAME).delete()
+    ProductClass.objects.filter(name=JOURNAL_PRODUCT_CLASS_NAME).delete()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('catalogue', '0001_initial'),
+        ('catalogue', '0024_fix_enrollment_code_slug')
```
jmyatt: for dependencies from this same app, you should just need the one (the migration prior to this one)

ecommerce/extensions/catalogue/migrations/0032_merge_20180321_2215.py
```diff
@@ -0,0 +1,17 @@
...
+# TODO: journal dependency
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0030_auto_20180124_1131'),
```
jmyatt: same as above - 0030 is not needed here since you have 0031 listed

ecommerce/extensions/catalogue/migrations/0033_merge_20180620_1551.py
```diff
@@ -0,0 +1,16 @@
...
+# Generated by Django 1.11.11 on 2018-06-20 15:51
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0031_course_entitlement_idverifyreq_attr'),
```
jmyatt: same comment

ecommerce/extensions/catalogue/tests/test_models.py
```diff
@@ -69,6 +71,13 @@ def test_deactivated_enrollment_code_update(self):
         enrollment_code.refresh_from_db()
         self.assertNotEqual(enrollment_code.expires, expiration_datetime)
 
+    # TODO: journal dependency
+    def test_journal_product_attribute(self):
+        """Verify journal product class."""
+        note = 'Some other test note.'
+        coupon = self._create_coupon_product_with_note_attribute(note)
+        self.assertNotEqual(coupon.is_journal_product, JOURNAL_PRODUCT_CLASS_NAME)
```
jmyatt: looks like this is comparing a bool to a string.  Did you mean to say:
    
    self.assertNotEqual(coupon.get_product_class().name, JOURNAL_PRODUCT_CLASS_NAME)

or
    
    self.assertFalse(coupon.is_journal_product)

ecommerce/extensions/payment/migrations/0019_auto_20180326_1545.py
```diff
@@ -0,0 +1,27 @@
...
+        migrations.AlterField(
+            model_name='sdncheckfailure',
+            name='created',
+            field=django_extensions.db.fields.CreationDateTimeField(auto_now_add=True, verbose_name='created'),
+        ),
+        migrations.AlterField(
+            model_name='sdncheckfailure',
+            name='modified',
+            field=django_extensions.db.fields.ModificationDateTimeField(auto_now=True, verbose_name='modified'),
+        ),
```
jmyatt: These 2 changes look like no-ops judging from [django-extensions](https://github.com/django-extensions/django-extensions/blob/3410f6a819a2ac9373d0387ca62c3255260a5e18/django_extensions/db/fields/__init__.py#L339-L371) - is this needed?  There are more instances of these same AlterField statements in your some of your other migrations.  

I do see other instances of similar changes in migrations in ecommerce, so maybe this is fine as is:
* [invoice 0006](https://github.com/edx/ecommerce/blob/master/ecommerce/invoice/migrations/0006_auto_20180228_1057.py#L19-L24)
* [payment 0013](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/payment/migrations/0013_sdncheckfailure.py#L21-L22)

ecommerce/extensions/refund/migrations/0005_merge_20180418_1605.py
```diff
@@ -0,0 +1,16 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('refund', '0004_auto_20180403_1120'),
+        ('refund', '0004_auto_20180326_1545'),
```
jmyatt: looks like you just need the second one here

ecommerce/journal/api/serializers.py
```diff
@@ -0,0 +1,116 @@
...
+    """ Serializer for ProductAttributeValue objects. """
+    name = serializers.SerializerMethodField()
+    code = serializers.SerializerMethodField()
+    value = serializers.CharField(max_length=256)
+
+    def get_name(self, instance):
+        return instance.attribute.name
+
+    def get_code(self, instance):
+        return instance.attribute.code
```
jmyatt: I *think* that instead of using SerializerMethodField() for these, you can simply declare them like this:

    name = serializers.CharField(source='attribute.name')
    code = serializers.CharField(source='attribute.code')

ecommerce/journal/api/serializers.py
```diff
@@ -0,0 +1,116 @@
...
+    attribute_values = AttributesSerializer(many=True)
+    stockrecords = StockRecordSerializer(many=True)
+
+    def create(self, validated_data):
+        attributes_data = validated_data.pop('attribute_values')
+        stockrecord_data = validated_data.pop('stockrecords')
+        product = Product.objects.create(**validated_data)
+
+        # Create the AttributeValues
+        product_class = ProductClass.objects.get(id=product.product_class_id)
```
jmyatt: doesn't look like you actually need to fetch the ProductClass object - you can just use `product.product_class_id` where you need it in the ProductAttribute query below

ecommerce/journal/api/serializers.py
```diff
@@ -0,0 +1,116 @@
...
+        # TODO - get attribute name and code from serializer and use in lookup of ProductAttribute
+        product_attribute = ProductAttribute.objects.get(product_class__id=product_class.id)
```
jmyatt: this assumes you'll get exactly one item for the filter terms.   Also, you can use `product_class_id` instead of `product_class__id` to avoid a join in the resulting query (since product_class_id exists as a field on the ProductAttribute model). So to avoid that join and to avoid hitting `MultipleObjectsReturned` or `DoesNotExist` exceptions, you should instead use something like:

    product_attributes = ProductAttribute.objects.filter(product_class_id=product.product_class_id)

then you'll need to handle that resulting list in the ProductAttributeValue.objects.create logic below

ecommerce/journal/api/v1/views.py
```diff
@@ -0,0 +1,28 @@
...
+Product = get_model('catalogue', 'Product')
+
+
+class JournalProductViewSet(viewsets.ModelViewSet):
+    """
+    API endpoint that allows journals to be viewed or edited.
+    """
+    # this does lookup based on UUID value of productattributevalue table
+    lookup_field = 'attribute_values__value_text'
+    queryset = Product.objects.filter(product_class__name='Journal')
```
jmyatt: you should import 

    from ecommerce.journal.constants import JOURNAL_PRODUCT_CLASS_NAME

and use the constant here, like you do where you create the ProductClass in migrations/0031_journal_product_class.py

ecommerce/journal/client.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def fetch_journal_bundle(site, journal_bundle_uuid):
+    """
+    Retrieve journal bundle for given uuid.
+    Retrieve it from cache if present, otherwise send GET request to journal bundle
+        discovery api and store in cache.
+
+    Args:
+        site
```
jmyatt: pls include data type and description

ecommerce/journal/conditions.py
```diff
@@ -0,0 +1,180 @@
...
+from slumber.exceptions import HttpNotFoundError, SlumberBaseException
+
+from ecommerce.extensions.offer.decorators import check_condition_applicability
+from ecommerce.extensions.offer.mixins import SingleItemConsumptionConditionMixin
+from ecommerce.journal.client import fetch_journal_bundle
+
+Condition = get_model('offer', 'Condition')
+
+
+class JournalBundleCondition(SingleItemConsumptionConditionMixin, Condition):
```
jmyatt: pls add class docstring

ecommerce/journal/conditions.py
```diff
@@ -0,0 +1,180 @@
...
+            if diff == self.basket_skus:
+                return False
+
+            # If there is a difference between the basket SKUs and the course SKUs and course SKUs,
+            # it represents the basket SKUs minus the SKUs for the current course.  Since we have already
+            # verified the course is represented, its SKUs can be safely removed from the set of SKUs
+            # in the basket being checked. (It does not affect the actual basket, just our copy)
+            self.basket_skus = diff
+
+        return True
```
jmyatt: nice job inline-commenting this logic.  Instead of using a class attribute for `self.basket_skus`, it might be clearer to scope that set locally in `is_satisfied`, then pass it to each of `basket_contains_all_required_courses` and `basket_contains_all_journals`, with those methods returning the altered set. So is_satisfied would have something like:

    basket_skus = set([line.stockrecord.partner_sku for line in basket.all_lines()])
    basket_skus, condition1 = self._basket_contains_all_required_courses(basket_skus)
    if not condition1:
        return False
    basket_skus, condition2 = self._basket_contains_all_journals(basket_skus)
    if not condition2:
        return False

but I dunno - that's kind of clunky.  Maybe simply marking the 2 'contains' functions private (with a leading underscore in the name) and adding a comment in each one (and/or at the top of this class where basket_skus is defined) to note that it is populated within `is_satisfied` would be best.

ecommerce/journal/conditions.py
```diff
@@ -0,0 +1,180 @@
...
+        Determines if a user is eligible for a journal bundle offer based on the products in their basket.
+
+        Args:
+            basket: contains information on line items for order, associated siteconfiguration
+                and associated user
+
+        Returns:
+            bool: True if condition is met
+        """
+        self.basket_skus = set([line.stockrecord.partner_sku for line in basket.all_lines()])
```
jmyatt: this can be moved to line 134 below, since it's not used until then

ecommerce/journal/forms.py
```diff
@@ -0,0 +1,132 @@
...
+            self.add_error('start_datetime', _('The start date must occur before the end date'))
+
+        return cleaned_data
+
+    def save(self, commit=True):
+        journal_bundle_uuid = self.cleaned_data['journal_bundle_uuid']
+        site = self.request.site
+
+        journal_bundle = fetch_journal_bundle(
+            site=self.request.site,
```
jmyatt: this can just be

    site=site

ecommerce/journal/fulfillment/modules.py
```diff
@@ -0,0 +1,111 @@
...
+
+class JournalFulfillmentModule(BaseFulfillmentModule):
+    """ Fulfillment Module for granting learner access to a Journal """
+
+    def supports_line(self, line):
+        logger.debug('Line order: [%s], is journal: [%s]', line, line.product.is_journal_product)
+        return line.product.is_journal_product
+
+    def get_supported_lines(self, lines):
+        """ Return a list of lines that can be fulfilled.
```
jmyatt: use an empty first line since it's a multi-line docstring

ecommerce/journal/fulfillment/modules.py
```diff
@@ -0,0 +1,111 @@
...
+        Args:
+            lines (List of Lines): Order Lines, associated with purchased products in an Order.
+        Returns:
+            A supported list of unmodified lines associated with "Journal" products.
+        """
+
+        return [line for line in lines if self.supports_line(line)]
+
+    def fulfill_product(self, order, lines):
+        """ Fulfills the purchase of a 'Journal'
```
jmyatt: empty first docstring line

ecommerce/journal/fulfillment/modules.py
```diff
@@ -0,0 +1,111 @@
...
+                    'Unable to fulfill line [%d] of order [%s]',
+                    line.id,
+                    order.number
+                )
+                line.set_status(LINE.FULFILLMENT_SERVER_ERROR)
+
+        logger.info('Finished fulfilling "Journal" product types for order [%s]', order.number)
+        return order, lines
+
+    def revoke_line(self, line):
```
jmyatt: docstring please

ecommerce/journal/tests/test_client.py
```diff
@@ -0,0 +1,253 @@
...
+        self.assertEqual(len(responses.calls), 2, "Incorrect number of API calls")
+        request = responses.calls[1].request
+        self.assertEqual(json.loads(request.body), data)
+        self.assertEqual(request.method, responses.POST)
+        response = responses.calls[1].response
+        self.assertEqual(response.status_code, 200)
+
+    @responses.activate
+    def test_fetch_journal_bundle(self):
+        """ Test 'fetch_journal_bundle'
```
jmyatt: use an empty first docstring line

ecommerce/journal/tests/test_views.py
```diff
@@ -0,0 +1,149 @@
...
+
+        response = self.assert_get_response_status(200)
+        self.assertContains(response, "mocked_title")
+        self.assertContains(response, "Journal Bundle Offers")
+        self.assertEqual(list(response.context['object_list']), journal_bundle_offers)
+
+    @mock.patch('ecommerce.journal.views.fetch_journal_bundle')
+    def test_get_filter_non_journal_bundle_offers(self, mock_journal_endpoint):
+        """ Test that if there are non journal bundle offers in the conditional offers table, this class will not try
+         to apply them """
```
jmyatt: docstring format

ecommerce/referrals/migrations/0005_auto_20180326_1545.py
```diff
@@ -0,0 +1,27 @@
...
+
+    dependencies = [
+        ('referrals', '0004_auto_20170215_2234'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='referral',
+            name='created',
+            field=django_extensions.db.fields.CreationDateTimeField(auto_now_add=True, verbose_name='created'),
```
jmyatt: same question from elsewhere on whether this migration is necessary or is a no-op

jmyatt (changes requested):

ecommerce/core/management/commands/create_or_update_site.py
bessiesteinberg: Yeah that is a good point - I'll make that change

ecommerce/core/migrations/0046_merge_20180525_1418.py
BenjiLee: I believe these numbers are automatically generated by manage.py makemigrations and are numbered by django app

ecommerce/extensions/test/factories.py
```diff
@@ -10,6 +10,9 @@
 from ecommerce.enterprise.benefits import EnterpriseAbsoluteDiscountBenefit, EnterprisePercentageDiscountBenefit
 from ecommerce.enterprise.conditions import EnterpriseCustomerCondition
 from ecommerce.extensions.offer.models import OFFER_PRIORITY_ENTERPRISE, OFFER_PRIORITY_VOUCHER
+# TODO: journal dependency
+from ecommerce.journal.benefits import JournalBundleAbsoluteDiscountBenefit, JournalBundlePercentageDiscountBenefit
+from ecommerce.journal.conditions import JournalBundleCondition
```
BenjiLee: I'm confused by the mix of the singular/plural use of Journal/s. I think JournalBundle makes sense, but I see `journals` referenced as plural, but the app name is `journal`.

ecommerce/journal/admin.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+# Register your models here.
```
BenjiLee: Do we want any Django Admin entries for Journal/s?

ecommerce/journal/api/serializers.py
```diff
@@ -0,0 +1,116 @@
...
+        fields = ('name', 'code', 'value',)
+
+
+class StockRecordSerializer(serializers.ModelSerializer):
+    """ Serializer for stock record objects. """
+    partner = serializers.SlugRelatedField(slug_field='short_code', queryset=Partner.objects.all())
+
+    class Meta(object):
+        model = StockRecord
+        fields = ('partner', 'partner_sku', 'price_currency', 'price_excl_tax',)
```
BenjiLee: I noticed in extensions/api/serializers.py we use the same subclassed name for this serializer.

```
class StockRecordSerializer(serializers.ModelSerializer):
    """ Serializer for stock record objects. """

    class Meta(object):
        model = StockRecord
        fields = ('id', 'product', 'partner', 'partner_sku', 'price_currency', 'price_excl_tax',)
```

We should either...
a. Separate journal/s from the main code and keep this serializer if we cannot use the other one. Additionally, we should rename this serializer to be Journal specific
b. Use the already existing serializer and possibly move it to a utils(?)

ecommerce/journal/api/serializers.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class StockRecordSerializerForUpdate(StockRecordSerializer):
```
BenjiLee: Same comment as StockRecordSerializer above

ecommerce/core/migrations/0046_merge_20180525_1418.py
bessiesteinberg: With the exception of the `0031_journal_product_class.py`, all of the migrations are auto generated by django.  Would you like me to squash the migrations within each app?

ecommerce/journal/api/serializers.py
bessiesteinberg: The `StockRecordSerializer` in extensions expects different fields and expects the `partner` field to be in a different format. So I will rename the one for journals

ecommerce/journal/api/serializers.py
bessiesteinberg: I'll rename it (see comment about `StockRecordSerializer`

ecommerce/core/migrations/0046_merge_20180525_1418.py
jmyatt: Hmm I think the only app's migrations you could safely squash would be `journal` since that's the only new app on this feature branch, but there's not much value in doing that since that app has just the 2 migrations.  I suspect that these issues I'm noticing are due to multiple rebases against master that you've probably done over the life of this branch.  

So although django has created these migrations, we may still have issues like redundant dependencies and (more importantly) incorrect/out-of-order dependencies, which I believe could create migration breakages as soon as this branch is merged.

I think one course to take on this would be:
* remove all migrations from this branch
  * on a clean environment representing master (with the db in the state represnted by master's migrations), switch to this branch and run makemigrations
  * also drop in the 00##_journal_product_class.py migration (at the appropriate number), and make sure its dependency is correct
  * run migrate to apply the new migrations
  * ensure everything looks right runtime, in terms of schema / data
  * then the branch is ready to merge once the rest of the code review is done (assuming there are no new migrations added master in the interim)

Another option would be to manually fix some subset of these issues I've raised to make sure the resulting dependency chain is correct.

I'd welcome input from others on my comments here and existing best practices with migration conflicts (@robrap?)

bessiesteinberg: Thank you for explaining that process!  I'm happy to do that.  I'll deal with these migration inconsistencies last then incase more come up as I'm working on this PR.

ecommerce/journal/admin.py
bessiesteinberg: There aren't any for 'programs' or 'entitlements' so I took that to assume that we haven't really found a need for Django Admin entries for products.  We can add them later if we need them.

Would you like me to delete this file?  It was autogenerated when we created a new Django app, but I can delete it if that would be cleaner.

ecommerce/journal/conditions.py
bessiesteinberg: Yeah this is a good point, I opted for the second option because I felt it was more readable.  It's still a little clunky but at least it won't be confusing now :)

ecommerce/journal/admin.py
BenjiLee: I think the custom interface that's been created for Enterprise Offers and Programs Offers replaces the need for the admin interface. This is fine the way it is.

BenjiLee:
@BessieSteinberg , Just a heads up Jason and I are on PTO for 7/4-7/6

ecommerce/journal/admin.py
robrap: I'd delete the file if it isn't serving any purpose.

jmyatt (approved):

ecommerce/journals/api/tests/test_serializers.py
```diff
@@ -0,0 +1,186 @@
...
+class StockRecordSerializerForUpdateTest(TestCase):
+
+    def setUp(self):
+        super(StockRecordSerializerForUpdateTest, self).setUp()
+        self.stock_record = StockRecordFactory(
+            partner=PartnerFactory(
+                short_code="dummy-partner"
+            ),
+            product=ProductFactory(
+                categories=""
```
BenjiLee, line 88: @BessieSteinberg This is completely up to you whether you want to look into this or not.

Looks like we are setting `categories=""` in quite a few places. Should we be subclassing `ProductFactory()` and add `categoriers=""` as a default?

ecommerce/journals/api/tests/test_views.py
```diff
@@ -0,0 +1,113 @@
+from decimal import Decimal as D
+
+import simplejson as json
+from django.urls import reverse
+from oscar.core.loading import get_model
+
+from ecommerce.journals.tests.mixins import JournalMixin     # pylint: disable=no-name-in-module
```
BenjiLee, line 7: Why is this pylint violation occuring? 
Are there ways to resolve this without disabling it?

bessiesteinberg: For some reason pylint fails with this error, even though there is a class JournalMixin defined in mixins.py - I am not really sure why but i am looking into it

BenjiLee (approved):

BenjiLee:
Note: @BessieSteinberg @jmyatt  and I verified the forwards and backwards migrations offline.  We also set up a journals product locally to confirm this feature's behaviour.

BenjiLee (changes requested):
@robrap Mentioned we should have asynchronous calls from e-commerce to LMS. This means calls such as the fulfillment call after purchased should be pushed on to a ecom worker. 

Synchronous calls have been a boon to the ecommerce service and we want to stop introducing more of these into ecommerce. Although the merging of this code won't immediately create synchronous calls, we will need to pass these calls to the ecom_worker before enabling journals. To ensure that we have no synchronous calls to other services from ecommerce, the action item here is to create a commited JIRA ticket to pass the synchronous calls (such as fulfillment) to the ecom_worker before turning Journals on.

bill-filler (approved):
looks good

robrap:
@BessieSteinberg: I know you were working hard with everyone to get this merged.  I don't think we should get in the habit of merging PRs when a reviewer has a red X "requested changes" outstanding on a PR.  Sounds like it was nearly there, but not sure why the preemptive merge.
FYI: @BenjiLee

bessiesteinberg:
@robrap My understanding was that the only reason for the X was because I needed to comment everywhere an async call is currently being made and create an associated JIRA task.  Every where that there is an asyc call I have included a comment in the code: ```# TODO: WL-1680: All calls from ecommerce to other services should be async``` 
I was under the impression that it would be fine to merge once that had been done and everything else was passing.

bill-filler:
I was under the same impression as @BessieSteinberg based on our guild meeting last week. Sorry if we jumped the gun, but thought we'd met the requirements and were hesitant to wait any longer with this huge PR.

robrap:
@BessieSteinberg @bill-filler: I think as a general rule it is good to get a thumb from someone who requests changes.  If someones gives an Approval with a final comment saying just do X...then I would do what you did.

Also, note that I don't think WL-1680 is what you think it is.
https://openedx.atlassian.net/browse/WL-1680

PR #1794, opened 2018-06-05
Add waffle to force all users to get the anonymous BasketCalculateView price

BenjiLee:
https://openedx.atlassian.net/browse/LEARNER-5497

### DIFF
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -427,7 +427,7 @@ class BasketCalculateView(generics.GenericAPIView):
         return response
 
     @transaction.non_atomic_requests
-    def get(self, request):
+    def get(self, request):  # pylint: disable=too-many-statements
         """ Calculate basket totals given a list of sku's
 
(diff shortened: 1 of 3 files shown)


jmyatt (approved):
LGTM

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -522,7 +528,9 @@ def get(self, request):
             if cached_response.is_hit:
                 return Response(cached_response.value)
 
-        if waffle.flag_is_active(request, "disable_calculate_temporary_basket_atomic_transaction"):
+        # Only disable the atomic transaction if we are also forcing the anonymous response for all users
```
robrap: Provide a little more detail.  Something like the following?
```
# There are too many open questions around dropping the atomic transaction for a user's basket,
# including how user basket merges was coded.  For now, only allow disabling the atomic
# transaction if we are also forcing the anonymous basket response for all users.
```

robrap (approved):

BenjiLee:
Closed in favor of https://github.com/edx/ecommerce/pull/1795

PR #1796, opened 2018-06-08
Initial fix work for missing id_verification_required attr

staubina:
These changes will add the id_verification_required attribute to entitlements so that they can be fulfilled correctly.

LEARNER-5456

### DIFF
--- a/ecommerce/extensions/catalogue/migrations/0031_course_entitlement_idverifyreq_attr.py
+++ b/ecommerce/extensions/catalogue/migrations/0031_course_entitlement_idverifyreq_attr.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from oscar.core.loading import get_model
+
+from ecommerce.core.constants import COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME
+
+ProductAttribute = get_model("catalogue", "ProductAttribute")
+ProductClass = get_model("catalogue", "ProductClass")
+
+
+def create_idverifyreq_attribute(apps, schema_editor):
+    """Create entitlement code 'id_verification_required' attribute."""
+    entitlement_code_class = ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME)
+    ProductAttribute.objects.create(
+        product_class=entitlement_code_class,
+        name='id_verification_required',
+        code='id_verification_required',
+        type='boolean',
+        required=False
+    )
+
+
+def remove_idverifyreq_attribute(apps, schema_editor):
+    """Remove enrollment code 'id_verification_required' attribute."""
+    enrollment_code_class = ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME)
+    ProductAttribute.objects.get(
+        product_class=enrollment_code_class,
+        name='id_verification_required'
+    ).delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0001_initial'),
+        ('catalogue', '0025_course_entitlement'),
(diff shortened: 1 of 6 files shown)


ecommerce/extensions/catalogue/migrations/0031_course_entitlement_idverifyreq_attr.py
```diff
@@ -0,0 +1,33 @@
...
+    """Create entitlement code 'id_verification_required' attribute."""
+    entitlement_code_class = ProductClass.objects.get(name=COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME)
+    ProductAttribute.objects.create(
+        product_class=entitlement_code_class,
+        name='id_verification_required',
+        code='id_verification_required',
+        type='boolean',
+        required=False
+    )
+
```
BenjiLee, line 23: Include the reverse migration.

https://github.com/edx/ecommerce/blob/2d2760cd94118e18529306b5cb08933d3c2b3821/ecommerce/extensions/catalogue/migrations/0019_enrollment_code_idverifyreq_attribute.py#L25-L28

ecommerce/extensions/catalogue/migrations/0031_course_entitlement_idverifyreq_attr.py
```diff
@@ -0,0 +1,33 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0001_initial'),
+        ('catalogue', '0025_course_entitlement'),
+        ('catalogue', '0030_auto_20180124_1131')
+    ]
+    operations = [
+        migrations.RunPython(create_idverifyreq_attribute)
```
BenjiLee: Add the reverse migration function here.
https://github.com/edx/ecommerce/blob/2d2760cd94118e18529306b5cb08933d3c2b3821/ecommerce/extensions/catalogue/migrations/0019_enrollment_code_idverifyreq_attribute.py#L38

BenjiLee (changes requested):

ecommerce/extensions/checkout/views.py
```diff
@@ -240,7 +240,7 @@ def get_order_dashboard_context(self, order):
 
     def get_order_verification_context(self, order):
```
BenjiLee, line 241: @robrap  and I have a suggested refactor.
```
    def get_and_maybe_update_order_verification_context(self, order):
        context = {}
        request = self.request
        site = request.site

        if request.user != order.user:
            return context
    
        for line in order.lines.all():
            product = line.product

            user_verification_required = \
                getattr(product.attr, 'id_verification_required', False) and \
                (getattr(product.attr, 'course_key', False) or getattr(product.attr, 'UUID', False))
            
            if user_verification_required:
                context.update({
                    'verification_url': site.siteconfiguration.build_lms_url('verify_student/reverify'),
                    'user_verified': request.user.is_verified(site),
                })
                break

        return context
```

BenjiLee, line 241: We renamed `get_and_maybe_update_order_verification_context`

staubina: I am not a fan of this method name change.  I prefer what it currently is named.

BenjiLee: `get_order_verification_context` is misleading as there is a possible update in addition to a get. Do you have a better suggestion to more closely reflect what this method is doing?

staubina: This is more or an update than a get so does one of the following work for you?

get_updated_order_verification_context

update_order_verification_context

BenjiLee: Talked offline with @staubina. This method is ONLY doing a get for a context.

bessiesteinberg (approved):

BenjiLee (approved):

PR #1808, opened 2018-06-18
Debug logging for get_basket_switch_data

BenjiLee:
Adds debug logging for the case where course_id is None. This will allow us to confirm what partner_sku is being set to when course_id is None.

### DIFF
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -5,6 +5,7 @@ from urllib import unquote, urlencode
 
 import newrelic.agent
 import pytz
+import waffle
 from django.conf import settings
 from django.contrib import messages
 from django.db import transaction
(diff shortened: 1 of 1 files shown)


ecommerce/extensions/basket/utils.py
```diff
@@ -144,6 +145,30 @@ def get_basket_switch_data(product):
...
             partner_sku = stock_record.partner_sku
             break
+
+    if waffle.switch_is_active("debug_logging_for_get_basket_switch_data"):  # pragma: no cover
+        if product.course_id is None:
+            if partner_sku is not None:
+                logger.info("get_basket_switch_data: product.course_id is None. partner_sku has been set")
+
+                courseless_stock_records = StockRecord.objects.filter(
+                    product__course_id=product.id,
```
jmyatt: why are you using course_id == product.id ?  Shouldn't this be:

    courseless_stock_records = StockRecord.objects.filter(
        product__course_id=None
    )

BenjiLee: Good catch.
It's supposed to be
```
                courseless_stock_records = StockRecord.objects.filter(
                    product__id=product.id,
                )
```

jmyatt (approved):
LGTM

PR #1810, opened 2018-06-19
Add more details to debug_logging_for_get_basket_switch_data

BenjiLee:
Add more details to debug_logging_for_get_basket_switch_data

### DIFF
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -151,6 +151,16 @@ def get_basket_switch_data(product):
             break
 
     if waffle.switch_is_active("debug_logging_for_get_basket_switch_data"):  # pragma: no cover
+        msg = "get_basket_switch_data: product.course_id={}, product.product_class.name={}, structure={}, " \
+            "partner_sku={}, product_cert_type={}, product_seat_type={}".format(
+                product.course_id,
+                product.product_class.name,
+                structure,
+                partner_sku,
+                product_cert_type,
+                product_seat_type)
+        logger.info(msg)
+
         if product.course_id is None:
             if partner_sku is not None:
                 logger.info("get_basket_switch_data: product.course_id is None. partner_sku has been set")


ecommerce/extensions/basket/utils.py
```diff
@@ -151,6 +151,11 @@ def get_basket_switch_data(product):
             break
 
     if waffle.switch_is_active("debug_logging_for_get_basket_switch_data"):  # pragma: no cover
+        msg = "get_basket_switch_data: product.course_id={} , structure={}, partner_sku={}".format(
+            product.course_id, structure, partner_sku
+        )
+        logger.info(msg)
+
```
jmyatt: as discussed, let's log more information about the products passed in to learn more about the various states / paths this function can take.  for instance:
* product.course_id (esp if None)
* product.product_class.name (or just product.product_class_id to avoid adding more sql queries)
* product_cert_type
* product_seat_type
* structure

jmyatt (changes requested):

BenjiLee:
@jmyatt I thought about adding some labels, but we should be able to filter in splunk so I didn't end up creating anything fancy.

ecommerce/extensions/basket/utils.py
```diff
@@ -151,6 +151,17 @@ def get_basket_switch_data(product):
...
+        msg = "get_basket_switch_data: product.course_id={}, product.product_class.name={}, structure={}, " \
+              "partner_sku={}, product_cert_type={}, product_seat_type={}".format(
+                product.course_id,
+                product.product_class.name,
+                structure,
+                partner_sku,
+                product_cert_type,
+                product_seat_type
+                )
+        logger.info(msg)
```
jmyatt: looks good.  what about adding another logger call in addition to this one, for only the cases where product_course_id=None?  That will make it easier to target only those cases in splunk

BenjiLee: We can use a splunk query to single those out

jmyatt: ok sounds good

jmyatt (approved):

PR #1813, opened 2018-06-19
Add more details to debug_logging_for_get_basket_switch_data 3

BenjiLee:


ecommerce/extensions/basket/utils.py
```diff
@@ -151,10 +151,11 @@ def get_basket_switch_data(product):
             break
 
     if waffle.switch_is_active("debug_logging_for_get_basket_switch_data"):  # pragma: no cover
-        msg = "get_basket_switch_data: product.course_id={}, product.product_class.name={}, structure={}, " \
-            "partner_sku={}, product_cert_type={}, product_seat_type={}".format(
+        msg = "get_basket_switch_data: product.course_id={}, product.get_product_class().name={}, structure={}, " \
+            "product.structure={}, partner_sku={}, product_cert_type={}, product_seat_type={}".format(
```
jmyatt, line 155: pls swap the 2 structures - the labels in the log message don't line up with the data

BenjiLee: good catch

BenjiLee: updated

jmyatt (approved):

PR #1817, opened 2018-06-21
check if enterprise catalog id exists

muhammad-ammar:
ENT-1055

### DIFF
--- a/ecommerce/enterprise/entitlements.py
+++ b/ecommerce/enterprise/entitlements.py
@@ -116,6 +116,10 @@ def get_course_entitlements_for_learner(site, user, course_id):
     try:
         enterprise_catalog_id = enterprise_learner_data[0]['enterprise_customer']['catalog']
         learner_id = enterprise_learner_data[0]['id']
+
+        if not enterprise_catalog_id:
+            logger.info('Invalid enterprise catalog id "[%s]"', enterprise_catalog_id)
+            return None
     except KeyError:
         logger.exception('Invalid structure for enterprise learner API response for the learner [%s]', user.username)
         return None
--- a/ecommerce/enterprise/tests/test_entitlements.py
+++ b/ecommerce/enterprise/tests/test_entitlements.py
@@ -484,3 +484,21 @@ class EntitlementsTests(EnterpriseServiceMockMixin, DiscoveryTestMixin, Discover
 
         log_message = 'Unable to connect to Discovery Service for catalog contains endpoint.'
         self._assert_is_course_in_enterprise_catalog_for_failure(2, log_message)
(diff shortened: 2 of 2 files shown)


ecommerce/enterprise/entitlements.py
```diff
@@ -116,9 +117,15 @@ def get_course_entitlements_for_learner(site, user, course_id):
     try:
         enterprise_catalog_id = enterprise_learner_data[0]['enterprise_customer']['catalog']
         learner_id = enterprise_learner_data[0]['id']
+
+        if not enterprise_catalog_id:
+            raise InvalidEnterpriseCatalogId
     except KeyError:
         logger.exception('Invalid structure for enterprise learner API response for the learner [%s]', user.username)
         return None
+    except InvalidEnterpriseCatalogId:
```
robrap: Hi.  Some questions about this:
1. Why are you raising an exception if you don't intend to use it?  If you just want to log something and return None, seems that could be done without the exception.
2. The ecommerce logs are filled with messages that are being ignored.  Is this really an exception or error?  Is it just a warning?  Who plans on looking at these messages and acting upon them?
Thanks @muhammad-ammar.

zubair-arbi: @robrap Good points.
1. Lets remove this exception and return above instead of raising this exception.
2. Lets add an info log here.

robrap: Just curious.  When does this happen?  Is this an expected case?

zubair-arbi (approved):
LGTM 👍

ecommerce/enterprise/entitlements.py
georgebabey: @robrap Theres some context in the JIRA - its a bit of an edge case but not entirely unexpected: "There are some customers that have no catalog (codes only or just set up for reporting)"

If there was no catalog, we were still making a call to discovery (which actually lead to 404 warnings in discovery which this will clean up as well)

I agree with @zubair-arbi there is still some value in the info level logging here if there is no catalog on an enterprise customer, but it doesn't need to be an exception.

robrap: Sounds great.  Just making sure that:
1. We aren't logging Warnings, Errors, or Exceptions for things that are not a problem, and
2. That we aren't sweeping real problems under the rug by logging as info.
Sounds like neither are the case any more.  Thank you!

zubair-arbi:
@georgebabey please approve the PR so that we can merge it.

saleem-latif (approved):
Looks Good 👍

georgebabey (approved):

PR #1822, opened 2018-06-21
Call get_basket_switch_data only when relevant

jmyatt:
From discovery work on performance of the Basket Summary page, we found that the util function get_basket_switch_data was being called for product types that are not relevant to the original functionality.  This is causing very slow load times for this view in cases where Entitlement products are in the users basket, because the code loops over a large and ever-growing number of StockRecords in search of a toggle product it will never find.

This change addresses the problem by ensuring this potentially expensive function gets called only when it will be efficient and when it can produce meaningful output.  I also added a doc string to the function to help developers understand its purpose.

Fixes https://openedx.atlassian.net/browse/LEARNER-5610

### DIFF
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -13,7 +13,14 @@ from oscar.test.factories import BasketFactory, ProductFactory, RangeFactory, Vo
 from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME
 from ecommerce.core.tests import toggle_switch
 from ecommerce.courses.tests.factories import CourseFactory
-from ecommerce.extensions.basket.utils import add_utm_params_to_url, attribute_cookie_data, prepare_basket
+from ecommerce.entitlements.utils import create_or_update_course_entitlement
+from ecommerce.extensions.basket.tests.mixins import BasketMixin
+from ecommerce.extensions.basket.utils import (
+    add_utm_params_to_url,
+    attribute_cookie_data,
+    get_basket_switch_data,
+    prepare_basket
+)
 from ecommerce.extensions.catalogue.tests.mixins import DiscoveryTestMixin
 from ecommerce.extensions.order.constants import DISABLE_REPEAT_ORDER_CHECK_SWITCH_NAME
 from ecommerce.extensions.order.exceptions import AlreadyPlacedOrderException
@@ -30,6 +37,7 @@ BasketAttributeType = get_model('basket', 'BasketAttributeType')
 BUNDLE = 'bundle_identifier'
 Option = get_model('catalogue', 'Option')
 Product = get_model('catalogue', 'Product')
+ProductClass = get_model('catalogue', 'ProductClass')
 
 
 def timeoutException():
@@ -37,7 +45,7 @@ def timeoutException():
 
 
 @ddt.ddt
-class BasketUtilsTests(DiscoveryTestMixin, TestCase):
+class BasketUtilsTests(DiscoveryTestMixin, BasketMixin, TestCase):
     """ Tests for basket utility functions. """
 
     def setUp(self):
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/basket/utils.py
```diff
@@ -117,49 +117,36 @@ def prepare_basket(request, products, voucher=None):
...
 
-    # Determine the proper partner SKU to embed in the single/multiple basket switch link
-    # The logic here is a little confusing.  "Seat" products have "certificate_type" attributes, and
-    # "Enrollment Code" products have "seat_type" attributes.  If the basket is in single-purchase
-    # mode, we are working with a Seat product and must present the 'buy multiple' switch link and
-    # SKU from the corresponding Enrollment Code product.  If the basket is in multi-purchase mode,
-    # we are working with an Enrollment Code product and must present the 'buy single' switch link
-    # and SKU from the corresponding Seat product.
+    Returns:
+        tuple containing the link display text and associated sku
```
BenjiLee: `tuple (str, str): containing the link display text and associated sku`

ecommerce/extensions/basket/utils.py
```diff
@@ -117,49 +117,36 @@ def prepare_basket(request, products, voucher=None):
...
-    product_seat_type = getattr(product.attr, 'seat_type', None)
-    for stock_record in stock_records:
-        stock_record_cert_type = getattr(stock_record.product.attr, 'certificate_type', None)
-        stock_record_seat_type = getattr(stock_record.product.attr, 'seat_type', None)
-        if (product_seat_type and product_seat_type == stock_record_cert_type) or \
-           (product_cert_type and product_cert_type == stock_record_seat_type):
-            partner_sku = stock_record.partner_sku
-            break
+
+    if product.is_enrollment_code_product or product.is_seat_product:
```
BenjiLee: opinion: I think we can remove the nesting and 2 duplicated product property calls (this might not matter)
```
    if product.is_enrollment_code_product:
        partner_sku = _find_seat_enrollment_toggle_sku(product)
        switch_link_text = _('Click here to just purchase an enrollment for yourself')
    elif product.is_seat_product:
        partner_sku = _find_seat_enrollment_toggle_sku(product)
        switch_link_text = _('Click here to purchase multiple seats in this course')
```

ecommerce/extensions/basket/utils.py
```diff
@@ -188,6 +175,47 @@ def get_basket_switch_data(product):
...
+    Returns:
+        sku (str): The sku of the associated Seat or Enrollment Code product.
+    """
+    target_structure = None
+    if product.is_seat_product:
+        target_structure = 'standalone'
+    elif product.is_enrollment_code_product:
+        target_structure = 'child'
+
+    stock_records = StockRecord.objects.filter(
```
BenjiLee: I think a useful note here could be to add 
`# Note: This query filter will not perform well with product types that do not have a course_id.`

BenjiLee: Or maybe
`# Note: This query filter is not selective for product types that do not have a course_id.`

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -409,6 +413,48 @@ def test_prepare_basket_attribute_delete(self):
...
+        do not relate to course_run instances.
+        """
+        COURSE_RUN_PRODUCT_CLASSES = ['Seat', 'Enrollment Code']
+
+        counter = 0
+        for product_class in ProductClass.objects.all():
+            product = ProductFactory(
+                product_class=product_class,
+                title='{} product title'.format(product_class.name),
+                upc=counter,
```
BenjiLee: This can be blank. The tests should pass on travis.

ecommerce/extensions/basket/utils.py
```diff
@@ -188,6 +175,43 @@ def get_basket_switch_data(product):
...
+
+    Arguments:
+        product (Product): Product of type Seat or Enrollment Code
+        target_structure (str): Structure of the related product we're seeking
+
+    Returns:
+        sku (str): The sku of the associated Seat or Enrollment Code product.
+    """
+
+    stock_records = StockRecord.objects.filter(
```
BenjiLee: We can either add a note here or write defensive code

`# Note: This query filter is not selective for product types that do not have a course_id.`

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -409,6 +417,46 @@ def test_prepare_basket_attribute_delete(self):
...
+        Verify that no basket switch data is retrieved for product classes that
+        do not relate to course_run instances.
+        """
+        COURSE_RUN_PRODUCT_CLASSES = ['Seat', 'Enrollment Code']
+
+        for product_class in ProductClass.objects.all():
+            product = ProductFactory(
+                product_class=product_class,
+                title='{} product title'.format(product_class.name),
+                upc=None,
```
BenjiLee: This need to be gone so that the super's upc entry can use the lambda.

jmyatt: hmm but removing this causes the IntegrityError.  are you saying that error only occurs locally (not on travis)?

ecommerce/extensions/basket/utils.py
jmyatt: I did both - defensive plus a comment as to why

ecommerce/extensions/basket/tests/test_utils.py
BenjiLee: Correct

BenjiLee:
Last thing is to update the commit message with the ticket number

BenjiLee (approved):

BenjiLee:
SHIP!

BenjiLee:
SHIP

PR #1824, opened 2018-06-25
Added update order line partner management command.

waheedahmed:
LEARNER-5608

### DIFF
--- a/ecommerce/extensions/order/management/commands/prompt.py
+++ b/ecommerce/extensions/order/management/commands/prompt.py
@@ -1,0 +1,33 @@
+import sys
+
+
+def query_yes_no(question, default="yes"):
+    """Ask a yes/no question via raw_input() and return their answer.
+
+    "question" is a string that is presented to the user.
+    "default" is the presumed answer if the user just hits <Enter>.
+        It must be "yes" (the default), "no" or None (meaning
+        an answer is required of the user).
+
+    The "answer" return value is one of "yes" or "no".
+    """
+    valid = {
+        "yes": True,
+        "y": True,
+        "no": False,
+        "n": False,
+    }
+    if default is None or default in valid.keys():
+        prompt = " [y/n] "
+    else:
+        raise ValueError("Invalid default answer: '%s'" % default)
+
+    while True:
+        sys.stdout.write(question + prompt)
+        choice = raw_input().lower()
+        if default is not None and choice == '':
+            return valid[default]
+        elif choice in valid:
+            return valid[choice]
+        else:
+            sys.stdout.write("Please respond with one of the following ({}).\n".format(', '.join(valid.keys())))
--- a/ecommerce/extensions/order/management/commands/tests/test_update_order_lines_partner.py
+++ b/ecommerce/extensions/order/management/commands/tests/test_update_order_lines_partner.py
@@ -1,0 +1,67 @@
+import ddt
+from django.core.management import call_command
+from django.core.management.base import CommandError
+from mock import patch
+from oscar.core.loading import get_model
+
+from ecommerce.extensions.test.factories import create_order
+from ecommerce.tests.factories import PartnerFactory
+from ecommerce.tests.testcases import TestCase
+
+LOGGER_NAME = 'ecommerce.extensions.order.management.commands.update_order_lines_partner'
+OrderLine = get_model('order', 'Line')
+
+
+@ddt.ddt
+class UpdateOrderLinePartnerTests(TestCase):
+    """Tests for update_order_lines_partner management command."""
+
+    PARTNER_CODE = 'testX'
+    YES_NO_PATCH_LOCATION = 'ecommerce.extensions.order.management.commands.update_order_lines_partner.query_yes_no'
+
+    def assert_error_log(self, error_msg, *args):
+        """Helper to call command and assert error log."""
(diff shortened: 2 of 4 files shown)


waheedahmed:
@edx/learner-spartans, @staubina Please review.

ecommerce/extensions/order/management/commands/update_order_lines_partner.py
```diff
@@ -0,0 +1,53 @@
...
+            raise CommandError(msg)
+
+        try:
+            partner = Partner.objects.get(short_code__iexact=partner_code)
+        except Partner.DoesNotExist:
+            msg = 'No Partner exists for code {}.'.format(partner_code)
+            logger.exception(msg)
+            raise CommandError(msg)
+
+        OrderLine.objects.filter(partner_sku__in=skus).update(partner=partner, partner_name=partner.name)
```
staubina: Can you add to this filter to exclude order lines that already have the correct partner?

staubina (changes requested):
Can we also add a step that will output how many lines will be updated and ask the user to confirm the operation?  

I could imagine the user running this command without realizing the impact and potential performance implications.

waheedahmed:
@staubina I have added the requested changes kindly take another look.

raeeschachar (approved):

ecommerce/extensions/order/management/commands/prompt.py
```diff
@@ -0,0 +1,38 @@
...
+    The "answer" return value is one of "yes" or "no".
+    """
+    valid = {
+        "yes": True,
+        "y": True,
+        "ye": True,
+        "no": False,
+        "n": False,
+    }
+    if default is None:
```
staubina: I feel like this conditional block could be reduced to something like 

```
if default is None or default in valid.keys():
  prompt = " [y/n] "
else:
  raise ValueError("invalid default answer: '%s'" % default)
```

ecommerce/extensions/order/management/commands/prompt.py
```diff
@@ -0,0 +1,38 @@
...
+
+    while True:
+        sys.stdout.write(question + prompt)
+        choice = raw_input().lower()
+        if default is not None and choice == '':
+            return valid[default]
+        elif choice in valid:
+            return valid[choice]
+        else:
+            sys.stdout.write("Please respond with 'yes' or 'no' (or 'y' or 'n').\n")
```
staubina: I would consider using the valid.keys() here to set the valid answers to protect against copy and paste errors.

ecommerce/extensions/order/management/commands/prompt.py
```diff
@@ -0,0 +1,38 @@
...
+    "default" is the presumed answer if the user just hits <Enter>.
+        It must be "yes" (the default), "no" or None (meaning
+        an answer is required of the user).
+
+    The "answer" return value is one of "yes" or "no".
+    """
+    valid = {
+        "yes": True,
+        "y": True,
+        "ye": True,
```
staubina: Is ye a typo?

staubina (changes requested):
Some recommendations on the new code.

Thank you for taking my requested changes into account.

ecommerce/extensions/order/management/commands/tests/test_update_order_lines_partner.py
```diff
@@ -0,0 +1,69 @@
...
+        order = create_order()
+        order_line = order.lines.first()
+        self.assertNotEqual(order_line.partner, new_partner)
+        with patch(self.YES_NO_PATCH_LOCATION) as mocked_yes_no:
+            mocked_yes_no.return_value = False
+            call_command('update_order_lines_partner', order_line.partner_sku, '--partner={}'.format(self.PARTNER_CODE))
+            # Verify that partner is not updated
+            self.assertNotEqual(order_line.partner, new_partner)
+
+    def test_update_order_lines_partner(self):
```
robrap: You could use ddt and pass in `yes_no_value` as True/False to remove duplicate code with test above.

Not a blocker, but please clean up while here.  Thanks.

ecommerce/extensions/order/management/commands/prompt.py
waheedahmed: @staubina This conditional block is setting capital Y or N based on default value so that user can know the default value and respond accordingly.

ecommerce/extensions/order/management/commands/tests/test_update_order_lines_partner.py
waheedahmed: Done.

waheedahmed:
@staubina @robrap I have addressed your comments, kindly take another look.

ecommerce/extensions/order/management/commands/prompt.py
staubina: I would consider adjusting this then and only having the defaulted value in the brackets rather than the capitalized letter.

staubina (approved):
One minor comment that is not a blocker.  This PR is ready to merge.

robrap:
FYI: I don't plan on doing a full review.  Don't wait on me. :)

raeeschachar:
Thanks for the heads up. A fyi, Waheed will be out till Monday, I wouldn't expect a reply back till Tuesday.

waheedahmed:
@staubina I have updated the PR, please merge it.

PR #1827, opened 2018-06-26
Added backoff decorator

irfanuddinahmad:
Description: 
There were 8876 errors in the logs from April18 to June18 due to 
File "/edx/app/ecommerce/ecommerce/ecommerce/enterprise/conditions.py", line 44, in is_satisfied
The exponential backoff and retry for exceptions in this function should reduce the connection failures

Test instructions:
Monitor the splunk logs and verify reduction in monthly error rate

### DIFF
--- a/ecommerce/core/constants.py
+++ b/ecommerce/core/constants.py
@@ -31,6 +31,9 @@ COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME = 'Course Entitlement'
 # Discovery Service constants
 DEFAULT_CATALOG_PAGE_SIZE = 100
 
+# Backoff for api calls switch
+BACKOFF_FOR_API_CALLS_SWITCH = u'enable_backoff_for_enterprise_api_calls'
+
 
 class Status(object):
     """Health statuses."""
--- a/ecommerce/enterprise/api.py
+++ b/ecommerce/enterprise/api.py
@@ -4,11 +4,14 @@ Methods for fetching enterprise API data.
 import logging
 from urllib import urlencode
 
+import backoff
+import waffle
 from django.conf import settings
 from requests.exceptions import ConnectionError, Timeout
-from slumber.exceptions import SlumberHttpBaseException
+from slumber.exceptions import SlumberBaseException, SlumberHttpBaseException
 
 from ecommerce.cache_utils.utils import TieredCache
+from ecommerce.core.constants import BACKOFF_FOR_API_CALLS_SWITCH
 from ecommerce.core.utils import get_cache_key
 
 logger = logging.getLogger(__name__)
@@ -162,10 +165,42 @@ def fetch_enterprise_learner_data(site, user):
     api = site.siteconfiguration.enterprise_api_client
     endpoint = getattr(api, api_resource_name)
     querystring = {'username': user.username}
-    response = endpoint().get(**querystring)
 
+    if waffle.switch_is_active(BACKOFF_FOR_API_CALLS_SWITCH):
+        response = _get_with_backoff(endpoint, **querystring)
+    else:
+        response = endpoint().get(**querystring)
     TieredCache.set_all_tiers(cache_key, response, settings.ENTERPRISE_API_CACHE_TIMEOUT)
     return response
+
+
+@backoff.on_exception(backoff.expo,
+                      (ConnectionError,
+                       SlumberBaseException),
+                      max_tries=3,
+                      max_time=30)
+def _get_with_backoff(endpoint, **querystring):
+    """
+    Resilient helper function for fetch_enterprise_learner_data(site, user)
+    In case of the below mentioned exceptions, the endpoint request is sent again after
+    an exponential backoff interval
+
+    Arguments:
+        endpoint: api endpoint
+        querystring: Dict containing query parameters
+
+    Returns:
+        dict: see docstring for fetch_enterprise_learner_data(site, user)
+
+    Raises:
+        ConnectionError: requests exception "ConnectionError", raised if if ecommerce is unable to connect
+            to enterprise api server.
+        SlumberBaseException: base slumber exception "SlumberBaseException", raised if API response contains
+            http error status like 4xx, 5xx etc.
+        Timeout: requests exception "Timeout", raised if enterprise API is taking too long for returning
+            a response. This exception is raised for both connection timeout and read timeout.
+    """
+    return endpoint().get(**querystring)
 
 
 def catalog_contains_course_runs(site, course_run_ids, enterprise_customer_uuid, enterprise_customer_catalog_uuid=None):
--- a/ecommerce/core/migrations/0046_backoff_for_api_calls_switch.py
+++ b/ecommerce/core/migrations/0046_backoff_for_api_calls_switch.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import migrations
+
+from ecommerce.core.constants import BACKOFF_FOR_API_CALLS_SWITCH
+
+
+def create_switch(apps, schema_editor):
+    """Create a switch for backoff and retry of api calls."""
+    Switch = apps.get_model('waffle', 'Switch')
+    Switch.objects.get_or_create(name=BACKOFF_FOR_API_CALLS_SWITCH, defaults={'active': False})
+
+
+def remove_switch(apps, schema_editor):
+    """Remove backoff for api calls switch."""
+    Switch = apps.get_model('waffle', 'Switch')
+    Switch.objects.filter(name=BACKOFF_FOR_API_CALLS_SWITCH).delete()
+
+
+class Migration(migrations.Migration):
+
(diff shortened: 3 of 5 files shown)


ecommerce/enterprise/conditions.py
```diff
@@ -24,6 +24,10 @@ class Meta(object):
     def name(self):
         return "Basket contains a seat from {}'s catalog".format(self.enterprise_customer_name)
 
+    @backoff.on_exception(backoff.expo,
```
zubair-arbi: use this decorator specifically on line 
```python
learner_data = fetch_enterprise_learner_data(basket.site, basket.owner)['results'][0] 
```
as this method catches the exceptions `except (ConnectionError, KeyError, SlumberHttpBaseException, Timeout):` also this will avoid use logging exception each time when error occurs. We have to log just once for after using all retries.

Also I think 60 sec time is too much, maybe we can set it to 30 sec as this method is called on basket page to determines if a user is eligible for an enterprise customer offer based on their association with the enterprise customer. @georgebabey what is your suggestion.

Also if possible we can specify the number of calls made during this `max_time` by using `max_retires` decorator argument.

muhammad-ammar (commented):
@irfanuddinahmad don't we need to add the tests?

ecommerce/enterprise/api.py
```diff
@@ -67,6 +66,10 @@ def fetch_enterprise_learner_entitlements(site, learner_id):
     return entitlements
 
 
+@backoff.on_exception(backoff.expo,
+                      RequestException,
```
robrap: 1. Why is RequestException not mentioned in the docstring?  Which does this apply to?
2. We should consider what it might look like to add more features into the api client to get this behavior by default.
3. At this time, these calls are happening inside an atomic transaction.  Anything that we do to increase time to failure will increase the likelihood of lock wait timeouts.  We may need to consider precomputing/caching external calls early in the transaction, outside of the atomic transaction.

FYI: @georgebabey

ecommerce/enterprise/api.py
```diff
@@ -162,12 +163,38 @@ def fetch_enterprise_learner_data(site, user):
...
+    response = _get_response(endpoint, **querystring)
 
     TieredCache.set_all_tiers(cache_key, response, settings.ENTERPRISE_API_CACHE_TIMEOUT)
     return response
 
 
+@backoff.on_exception(backoff.expo,
+                      (ConnectionError,
+                       SlumberBaseException),
+                      max_tries=4)
```
zubair-arbi: please set `max_tries=3` and `max_time=30`

zubair-arbi (approved):
Good effort 👍

ecommerce/enterprise/api.py
georgebabey: @robrap 
2. Agreed - it would be good to get some intelligent defaults for backoff into edx-rest-api-client, however I'm not sure what that would look like as it is likely use-case dependent.
3. Hmm... Good call - I'd hate to increase lock timeouts when trying to fix other exceptions

georgebabey: @robrap Any prior art for (3) - prefetching the data outside of the atomic transaction?

ecommerce/enterprise/api.py
```diff
@@ -162,12 +163,39 @@ def fetch_enterprise_learner_data(site, user):
...
 
 
+@backoff.on_exception(backoff.expo,
+                      (ConnectionError,
+                       SlumberBaseException),
+                      max_tries=3,
+                      max_time=30)
+def _get_response(endpoint, **querystring):
+    """
+        Resilient helper function for fetch_enterprise_learner_data(site, user)
```
zubair-arbi: nit: dedent this docsting

ecommerce/enterprise/api.py
robrap: 2. Agreed that this may take some thought. I do want this thinking to happen, but not a blocker for this PR.
3. Only discussion And a ticket so far. This may be simple, but there are some details that may mess up the idea.
General idea:
a. Use decorator to remove atomic transaction from the view, and manually add back the transaction via code.
b. Make the same call that you know you’ll make in the transaction using the same TieredCache.
c. The second time you go to process, the result will simply be pulled from the Request Cache in the TieredCache.
Here’s the tricky caveat: Oscar recommends protecting all transactions as atomic by default, as we’ve done, with little detail about what matters most. We have middleware that messes with the basket that I am guessing is part of the overall atomic transaction as-is, and I presume would no longer be if the decorator is used to disable it for the view. We need some way to determine if this is safe.
I’ve got other alternative hack ideas as well.
There’s also the idea of redundant data to keep from having to make external calls.
Food for thought. I’m not sure if this makes sense without some solution to that problem. Or, we might need load testing where we force the issue and track the lock wait timeouts.

Lastly, an additional thought is that I’ve discussed using Circuit Breakers with Feanil. We found some libraries that showed promise. It would need to play nice with the backoff library, but I could see backoff without a Circuit Breaker creating a new problem for us of lots of threads waiting the full timeout.

I’m not certain what the recommendation should be. One possibility is to have two methods, one with and one without backoff, and use a waffle flag to choose which to call so we can turn this feature on and off temporarily. That way, whether we experiment in load test or prod we can try either without a redeploy.

georgebabey:
I like @robrap idea of using a waffle flag as circuit breaker to turn this on/off as mentioned here: https://github.com/edx/ecommerce/pull/1827#discussion_r199145150. I worry we'll see an increase in deadlocks which can be tricky to reproduce

ecommerce/core/constants.py
```diff
@@ -31,6 +31,8 @@
 # Discovery Service constants
 DEFAULT_CATALOG_PAGE_SIZE = 100
 
+#Backoff for api calls switch
+BACKOFF_FOR_API_CALLS_SWITCH = u'enable_backoff_for_api_calls'
```
georgebabey: Can we update the name to `enable_backoff_for_enterprise_api_calls` just to avoid confusion if someone thinks the backoff applies to all api calls.

ecommerce/enterprise/api.py
```diff
@@ -162,12 +165,42 @@ def fetch_enterprise_learner_data(site, user):
...
 
 
+@backoff.on_exception(backoff.expo,
+                      (ConnectionError,
+                       SlumberBaseException),
+                      max_tries=3,
+                      max_time=30)
+def _get_response(endpoint, **querystring):
+    """
+    Resilient helper function for fetch_enterprise_learner_data(site, user)
```
georgebabey: Would be good to mention the backoff in the comments in case someone is just looking at the docstrings/documentation

ecommerce/enterprise/api.py
```diff
@@ -162,12 +165,42 @@ def fetch_enterprise_learner_data(site, user):
...
     TieredCache.set_all_tiers(cache_key, response, settings.ENTERPRISE_API_CACHE_TIMEOUT)
     return response
 
 
+@backoff.on_exception(backoff.expo,
+                      (ConnectionError,
+                       SlumberBaseException),
+                      max_tries=3,
+                      max_time=30)
+def _get_response(endpoint, **querystring):
```
georgebabey: could rename this as `get_with_backoff` to be more explicit on what this method is doing

georgebabey (approved):
A few small comments but overall looks good!

irfanuddinahmad:
@georgebabey I have incorporated your suggestions. Kindly review and merge. Thanks.

ecommerce/enterprise/api.py
```diff
@@ -162,12 +165,44 @@ def fetch_enterprise_learner_data(site, user):
...
+        response = endpoint().get(**querystring)
     TieredCache.set_all_tiers(cache_key, response, settings.ENTERPRISE_API_CACHE_TIMEOUT)
     return response
 
 
+@backoff.on_exception(backoff.expo,
+                      (ConnectionError,
+                       SlumberBaseException),
+                      max_tries=3,
+                      max_time=30)
```
robrap, line 181: @georgebabey: I'm having second thoughts about us putting this through. I'm not convinced that this isn't trading off one problem for a different and worse problem. The problem with the Lock Wait Timeouts is that they are so hard to solve and/or know which transactions/situations are the root cause, and they affect the entire system due to the locking.

For this, and all of our back-end synchronous calls, we need a better solution. 30 seconds is just asking for locking issues any time the LMS is having trouble, which happens every LMS deployment (as well as other times).  Let's chat in person.

Sorry about this...but it feels wrong to introduce this at the same time that Chris and Jason are working so hard to shrink these kinds of problems.

georgebabey: @robrap No problem and totally understand the concern. 👍  Lets circle back and come up with a good game plan. I think we do need some resiliency for external calls and exponential backoff is likely the right way, but we need to do so without introducing additional lock waits.

georgebabey (changes requested):
Requesting changes until we have a better strategy for the locking

robrap:
@georgebabey: Does it make sense to close this PR and just refer to it from some ticket related to the ongoing research?

PR #1828, opened 2018-06-26
Remove BasketSummaryView debug logging and traces

jmyatt:
For BasketSummaryView performance analysis, we temporarily added debug logging and granular New Relic function traces.  These have served their purpose and are no longer needed so this PR removes them.  

https://openedx.atlassian.net/browse/LEARNER-5611

### DIFF
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -3,9 +3,7 @@ import json
 import logging
 from urllib import unquote, urlencode
 
-import newrelic.agent
 import pytz
-import waffle
 from django.conf import settings
 from django.contrib import messages
 from django.db import transaction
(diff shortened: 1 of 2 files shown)


BenjiLee (approved):

BenjiLee:
Looks like .1 wasn't good enough :(

jmyatt:
This guy hit the same problem but the issue page he points to does not exist:
https://github.com/codecov/codecov-bash/issues/54 

I'm guessing codecov deleted their 'support' repo when switching to an email support model: https://codecov.io/support

According to documentation, codecov is supposed to detect and avoid this exact situation:

`Codecov will set the Project Status to success ✅ when it detects that the coverage was decreased due to the offset of the removed lines. Clearly, the coverage is lower, but not due to a developer fault.`

https://docs.codecov.io/docs/coverage-offset

jmyatt:
closing this - changes were incorporated into #1839

PR #1832, opened 2018-06-27
override codecov precision

jmyatt:
This change adds a CodeCov configuration file to the repo to enable us to override the default settings.  The setting being overridden is coverage precision changing from 2 decimal places to 1.  This should avoid many instances where a PR with 100% coverage on the patch being blocked from merging due to a resulting slightly lower overall coverage percentage for the repo.  The tolerance for that will now be up to 0.09% instead of 0.01%.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,3 @@
+# Override team (default) settings
+coverage:
+  precision: 1


BenjiLee (approved):

BenjiLee:
@jmyatt What other settings are there?

jmyatt:
for coverage config it's just these few things:

`range`: for color coding - how low a percentage is considered red (bad).  default 70-100
`round`: up, down, or nearest. default down
`precision`: changed in this PR

then there is `status` for whether to track the project and/or patch

https://docs.codecov.io/docs/coverage-configuration

robrap:
@jmyatt: If you create a new small PR with the following code:
```
never_true = None
if never_true:
    print('hello')
```
Will this fail or pass codecov?  It would not be good for this to pass.

jmyatt:
@robrap testing that now: https://github.com/edx/ecommerce/pull/1840

jmyatt:
@robrap so, a change like that fails coverage percentage on both the patch and the project (by 0.1%).  My guess is that the rounding does not allow a small percentage change to ever be represented as zero / no change.

robrap:
Great.  Thanks for this fix!

PR #1833, opened 2018-06-28
Updated logging for order creation exception

Ayub-Khan:
Updated logging for order creation exception. Now it shows a bit more about the exception. what caused the failure previously it was not providing any context What really happened.

### DIFF
--- a/ecommerce/extensions/checkout/mixins.py
+++ b/ecommerce/extensions/checkout/mixins.py
@@ -38,7 +38,8 @@ class EdxOrderPlacementMixin(OrderPlacementMixin):
     # Instance of a payment processor with which to handle payment. Subclasses should set this value.
     payment_processor = None
 
-    order_placement_failure_msg = 'Payment was received, but an order for basket [%d] could not be placed.'
+    order_placement_failure_msg = 'Order Failure: Payment was received, but an order for basket [%d] ' \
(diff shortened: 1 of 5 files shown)


robrap (commented):
Only tests have changed. Was that the intention?

tasawernawaz (approved):

Ayub-Khan:
@robrap  yeah improved logs a bit. because of LEARNER-5603

ecommerce/extensions/checkout/mixins.py
```diff
@@ -38,7 +38,8 @@ class EdxOrderPlacementMixin(OrderPlacementMixin):
     # Instance of a payment processor with which to handle payment. Subclasses should set this value.
     payment_processor = None
 
-    order_placement_failure_msg = 'Payment was received, but an order for basket [%d] could not be placed.'
+    order_placement_failure_msg = 'Payment was received, but an order for basket [%d] could not be placed' \
```
robrap: The tests might be less brittle if we had something like:
```
order_placement_failure_msg = 'Order Failure: Payment was received, ... '`
```
Then, for testing, we could just check if "Order Failure" was in the message, rather than trying to get the exact message.

robrap (approved):
The following is clean-up that I'd like to see, but I won't make it a blocker.

PR #1835, opened 2018-06-28
Remove BasketSummaryView debug logging and traces

jmyatt:
For BasketSummaryView performance analysis, we temporarily added debug logging and granular New Relic function traces. These have served their purpose and are no longer needed so this PR removes them.

https://openedx.atlassian.net/browse/LEARNER-5611


jmyatt:
closing in favor of #1828

PR #1837, opened 2018-06-29
Bessiesteinberg/new migrations

bessiesteinberg:
Refactor Journal to Plural and Clean Migrations

- The django app name 'journal' has been changed to 'journals' to be
  in line with other product apps like 'entitlements' and 'programs'
- The ecommerce journal api path has been changed from 'journal/api/v1'
  to 'journals/api/v1'

- Migrations were in an odd state because this feature branch had been
  rebased from master many times over the course of months. Clean up
  included:
  - Changing name of 0031_journal_product_class.py migration to
    0032_journal_product_class.py so there wouldn't be two migrations
    sharing the same number
  - Regenerating all the auto generated migrations


bill-filler (approved):
Please make sure to file PR's against the following repos to change /journal/api/v1 to /journals/api/v1 for ecommerce  journals api url:
Themes: (whitelabel/journal branch) https://github.com/edx/edx-themes
Configuration: (master)
https://github.com/edx/configuration
edx-internal: (master)
https://github.com/edx/edx-internal

PR #1839, opened 2018-06-29
Cleanup of BasketSummary debug hooks

jmyatt:
We will no longer be using the waffle switch `force_anonymous_user_response_for_basket_calculate` since we changed the approach to the original problem that flag was addressing.  The LMS calls to this API now send in the is_anonymous parameter so that all the this API's functionality remains intact, as it is required for the Enterprise use case.  This change removes usage of the flag including the override in the associated tests.

Also, we temporarily added debug logging and granular New Relic function traces. These have served their purpose and are no longer needed so this PR removes them.

[LEARNER-5611](https://openedx.atlassian.net/browse/LEARNER-5611)
[LEARNER-5664](https://openedx.atlassian.net/browse/LEARNER-5664)

### DIFF
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -471,13 +471,8 @@ class BasketCalculateView(generics.GenericAPIView):
 
         basket_owner = request.user
 
-        if waffle.switch_is_active("force_anonymous_user_response_for_basket_calculate"):
-            # Use the anonymous user program price for all users
-            requested_username = ''
-            is_anonymous = True
-        else:
-            requested_username = request.GET.get('username', default='')
-            is_anonymous = request.GET.get('is_anonymous', 'false').lower() == 'true'
+        requested_username = request.GET.get('username', default='')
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -528,11 +523,7 @@ def get(self, request):  # pylint: disable=too-many-statements
             if cached_response.is_hit:
                 return Response(cached_response.value)
 
-        # There are too many open questions around dropping the atomic transaction for a user's basket,
-        # including how user basket merges was coded.  For now, only allow disabling the atomic
-        # transaction if we are also forcing the anonymous basket response for all users.
```
BenjiLee, line 533: Just nothing this here: This comment needs to be updated, but since we will be removing the atomic transaction conditional here in a future PR, this comment won't be needed anyway.

BenjiLee (approved):

ecommerce/extensions/api/v2/views/baskets.py
jmyatt: yea, that matches my thinking on removing this comment

PR #1843, opened 2018-07-02
LEARNER-5603 Hot Fix

Ayub-Khan:
[LEARNER-5603](https://openedx.atlassian.net/browse/LEARNER-5603)
Students are skipping the basket views which have all the voucher checks if its valid or not. Students are able to submit payments but after payments oscar checks if voucher is valid or not and breaks the system as vouchers used for those baskets is not valid. On average we are receiving around 20 such cases every day which is causing a lot of problem for support. Also we have to fix all the previous paid baskets which have expired vouchers. 
This will prevent students from paying using a expired coupon. I do know this is not a good solution. We will be reverting this after we figure out how students are able to bypass basket views and direct submit payment using submit payment view.

### DIFF
--- a/ecommerce/extensions/payment/tests/views/test_cybersource.py
+++ b/ecommerce/extensions/payment/tests/views/test_cybersource.py
@@ -1,6 +1,7 @@
 """ Tests of the Payment Views. """
 from __future__ import unicode_literals
 
+import datetime
 import json
 
 import ddt
@@ -8,10 +9,12 @@ import mock
 import responses
 from django.conf import settings
 from django.urls import reverse
+from django.utils.timezone import now
 from freezegun import freeze_time
 from oscar.apps.payment.exceptions import TransactionDeclined
 from oscar.core.loading import get_class, get_model
 from oscar.test import factories
+from waffle.testutils import override_switch
 
 from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME, ENROLLMENT_CODE_SWITCH
 from ecommerce.core.models import BusinessClient
@@ -21,11 +24,12 @@ from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.extensions.api.serializers import OrderSerializer
 from ecommerce.extensions.basket.utils import basket_add_organization_attribute
 from ecommerce.extensions.order.constants import PaymentEventTypeName
+from ecommerce.extensions.payment.constants import VOUCHER_VALIDATION_BEFORE_PAYMENT
 from ecommerce.extensions.payment.exceptions import InvalidBasketError, InvalidSignatureError
 from ecommerce.extensions.payment.processors.cybersource import Cybersource
 from ecommerce.extensions.payment.tests.mixins import CybersourceMixin, CybersourceNotificationTestsMixin
 from ecommerce.extensions.payment.views.cybersource import CybersourceInterstitialView
-from ecommerce.extensions.test.factories import create_basket
+from ecommerce.extensions.test.factories import create_basket, prepare_voucher
 from ecommerce.invoice.models import Invoice
 from ecommerce.tests.testcases import TestCase
 
@@ -36,9 +40,10 @@ Order = get_model('order', 'Order')
 OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
 PaymentEvent = get_model('order', 'PaymentEvent')
 PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
+Product = get_model('catalogue', 'Product')
 Selector = get_class('partner.strategy', 'Selector')
 Source = get_model('payment', 'Source')
-Product = get_model('catalogue', 'Product')
+Voucher = get_model('voucher', 'Voucher')
 
 post_checkout = get_class('checkout.signals', 'post_checkout')
 
@@ -76,6 +81,16 @@ class CybersourceSubmitViewTests(CybersourceMixin, TestCase):
         """ Creates a Basket ready for checkout. """
         basket = create_basket(owner=self.user, site=self.site)
         basket.strategy = Selector().strategy()
+        basket.thaw()
+        return basket
+
+    def _prepare_basket_for_voucher_validation_tests(self, voucher_start_date, voucher_end_date):
+        """ Prepares basket for voucher validation """
+        basket = Basket.objects.create(site=self.site, owner=self.user)
+        voucher, product = prepare_voucher(start_datetime=voucher_start_date, end_datetime=voucher_end_date)
+        basket.strategy = Selector().strategy()
+        basket.add_product(product)
+        basket.vouchers.add(voucher)
         basket.thaw()
         return basket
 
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/payment/views/__init__.py
```diff
@@ -77,6 +77,15 @@ def get_form_kwargs(self):
             'request': self.request,
         }
 
+    def check_valid_voucher(self):
+        """LEARNER-5603 Hot fix"""
```
staubina: Add more detail here as the ticket ID could potentially change.

jmyatt: yes, please point to the code in place_order() that you're copying here and briefly explain why we need that

jmyatt:
LGTM as a temporary fix

robrap (commented):
If we think this change is risky, because we aren't sure about it, some ways to mitigate risk:
1. Add a unit test to at least ensure we aren't introducing any issues.  Codecov says we are never going into this code.
2. Add appropriate logging to potentially help us track down when this is happening?
3. Could add a temporary waffle switch so you could turn this on/off if it doesn't do what you want it to do.

ecommerce/extensions/payment/migrations/0019_create_voucher_validation_switch.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
```
jmyatt: hmm not sure a migration is the best way to add a waffle switch, esp since it would require a future migration to remove the switch, which we expect to do in a future, more comprehensive fix for this issue.  Is there a compelling reason to do this instead of simply creating the switch in django admin?

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -179,6 +183,54 @@ def test_field_error(self):
...
+        self.assertEqual(response['content-type'], JSON)
+
+    @mock.patch(
+        'ecommerce.extensions.voucher.models.Voucher.is_available_to_user',
+        return_value=(False, None)
+    )
+    def test_submit_view_fails_if_voucher_not_available(self, mock_is_available_to_user):
+        """ Verify SubmitPaymentView fails if basket voucher not available to student"""
+        # If switch is disable enable it for the test
+        toggle_switch(VOUCHER_VALIDATION_BEFORE_PAYMENT, True)
```
robrap: Nit: Use the following decorator:
```
@override_switch(VOUCHER_VALIDATION_BEFORE_PAYMENT, active=True)
```
This isn't strictly required, but this is how we've been doing this.

ecommerce/extensions/payment/migrations/0019_create_voucher_validation_switch.py
robrap: Agreed.  Let's drop this migration.

ecommerce/extensions/payment/views/__init__.py
```diff
@@ -77,6 +79,26 @@ def get_form_kwargs(self):
...
         }
 
+    def check_valid_voucher(self):
+        """
+        LEARNER-5603 Hot fix
+        Learners are able to bypass the basket views and pay using the
+        expired vouchers. This will disable students from doing so.
+        (https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/order/utils.py#L68-L70)
+        """
+        if waffle.switch_is_active(VOUCHER_VALIDATION_BEFORE_PAYMENT):
```
robrap: Please add a JIRA ticket for removing this switch in the future and add a comment like the following above this line:
```
# TODO: LEARNER-XXXX: Clean-up validation code as needed after further investigation.
```
Note: I imagine that LEARNER-5603 will be closed.  If it will remain open, then maybe this is just LEARNER-5603.  It just needs to be linked to an open ticket.  Thanks!

robrap (changes requested):
Minor changes requested.  Thank you!

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -179,6 +183,54 @@ def test_field_error(self):
         errors = json.loads(response.content)['field_errors']
         self.assertIn(field, errors)
 
+    @ddt.data(
+        (now() - datetime.timedelta(days=3), 400),
+        (now() + datetime.timedelta(days=3), 200))
+    @ddt.unpack
+    def test_submit_view_fails_for_invalid_voucher(self, end_time, status_code):
```
robrap: Also, can you add one more test (or manually verify) that if the switch is False then it succeeds without an error.  Basically, I want to make sure that this test proves that the bug exists with the switch disabled, and is fixed with it enabled.  Thank you!

robrap:
@Ayub-Khan: Thank you for adding both the switch and the test. :)

ecommerce/extensions/payment/tests/views/test_cybersource.py
Ayub-Khan: @robrap In ecommerce we have been using toggle_switch, decorator override_switch was only used once before. I have updated the test tho decorator is a better choice. May be create a ticket to remove toggle_switch altogether and only use decorator only.

ecommerce/extensions/payment/tests/views/test_cybersource.py
Ayub-Khan: done.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -179,6 +194,52 @@ def test_field_error(self):
...
+        response = self.client.post(self.path, data)
+
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response['content-type'], JSON)
+        self.assertEqual(mock_is_available_to_user.call_count, 3)
+
+    @override_switch(VOUCHER_VALIDATION_BEFORE_PAYMENT, active=False)
+    def test_successful_submit_view_with_voucher_switch_disabled(self):
+        """
+        Verify SubmitPaymentView call is successful if voucher waffle switch disabled
```
robrap: Great. Thank you.
Some minor updates.
1. Could you update this docstring as follows:
```
“””
Temporary test to confirm the problem with SubmitPaymentView
Accepting an invalid voucher when the waffle switch is False.
This will be cleaned up in LEARNER-5719.
“””
```
2. Nit: these tests would be a little cleaner to have both `start_time` and `end_time` as variables. I’m this test, you could name it `expired_end_time’.

robrap (approved):
Minor comment if you don’t mind updating as part of squashing. We can merge and deploy tomorrow morning when back from our holiday. Thank you!

Ayub-Khan:
sure

staubina (approved):

staubina:
@jmyatt Do you mind finalizing your review?

robrap:
@staubina: This is ready for you to merge/deploy.  @jmyatt made one comment that I already had the team see through.  Note: after deployment, you'll want to turn on the switch and monitor for issues.

PR #1845, opened 2018-07-03
dummy commit

estute:


robrap:
@estute: FYI: Most of our issues are when when non-test code is being deleted.

PR #1846, opened 2018-07-05
LEARNER-5681

Ayub-Khan:
Increased timeout to fix drive timeout issues.

### DIFF
--- a/e2e/pytest.ini
+++ b/e2e/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
 addopts = --driver=Chrome
-# Each test must complete in less than 45 seconds
-timeout = 60
+# Each test must complete in less than 90 seconds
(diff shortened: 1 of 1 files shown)


raeeschachar (approved):

robrap (approved):
We can see if this improves things, but this doesn't seem like the right way to go.  No successful call should be taking over 60 seconds.  The timeout should probably be a lot lower.

Do we know what call is taking this long?  Is it an external call to the payment provider?  We may need to find some way to mock things so that we don't have flakiness tied to an external provider.  Needs some thought.

In other words, I'm not blocking this if you want to see if this helps resolve.  But, I'm planting a seed in case this doesn't succeed and just makes real failures take longer.  We will want to go to a lower timeout and resolve in another way.

Thanks.

PR #1848, opened 2018-07-05
ENT-1085 Revert changes related to ENT-1052

brittneyexline:
We are currently seeing no enterprise offers being applied in the basket page in production. To be sure that we will use the previous offers logic, I am reverting the feature changes in ecommerce. The lms will still send the catalog param but the intent of this revert is that ecommerce will do nothing with the param.

### DIFF
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -3,7 +3,6 @@ import json
 import logging
 from urllib import unquote, urlencode
 
-import newrelic.agent
 import pytz
 from django.conf import settings
 from django.contrib import messages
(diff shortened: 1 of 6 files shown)


ecommerce/extensions/basket/utils.py
```diff
@@ -271,31 +269,6 @@ def basket_add_organization_attribute(basket, request_data):
 
 
 @newrelic.agent.function_trace()
```
robrap: Did you mean to move this trace and not remove it

brittneyexline: ah, looks like i should have removed it. thanks!

bill-filler (approved):
approved

brittneyexline:
@bill-filler @robrap is there a reason quality checks would start failing for ecommerce/static/js/payment_processors/cybersource.js (especially since that file didn't change)?
that appears to be why the checks are failing...

PR #1850, opened 2018-07-06
[ENT-1081] Fixed the Enterprise Offer Name Issues

asadiqbal08:
[ENT-1081](https://openedx.atlassian.net/browse/ENT-1081)
@georgebabey can you look into the change set?


ecommerce/enterprise/forms.py
```diff
@@ -96,11 +96,15 @@ def save(self, commit=True):
...
-        self.instance.name = _(u'Discount of type {} provided by {} for {}.'.format(
+        offer_name = _(u'Discount of type {} provided by {}.'.format(
             ConditionalOffer.SITE,
-            enterprise_customer_name,
-            enterprise_customer_catalog_uuid,
+            enterprise_customer_name
         ))
+
+        # Truncate offer_names down to 128 characters, as Oscar's AbstractConditionalOffer name is max_length 128
+        offer_name = (offer_name[:125] + '...') if len(offer_name) > 128 else offer_name  # pylint: disable=unsubscriptable-object
```
asadiqbal08: @georgebabey I can see, in programs form, A [similar approach](https://github.com/edx/ecommerce/blob/master/ecommerce/programs/forms.py#L92) is adopted to fix that issue. Thoughts ?

ecommerce/enterprise/forms.py
```diff
@@ -96,11 +96,15 @@ def save(self, commit=True):
...
+
+        self.instance.name = offer_name
```
georgebabey: The instance name still needs to be unique - so we need the catalog_uuid in the name. Thats also why I recommend truncating just the enterprise_name down, otherwise we risk the catalog_uuid getting cut off and it failing to save because of the unique constraint.

Note when we do change the name to include the UUID, we need to update the receipt to not show the offer name directly, but instead show `{{ benefit }} discount provided by {{ enterprise_customer_name }}.` since the offer name would then show the catalog uuid, which doesn't make sense to present to an end user.

georgebabey (changes requested):
We still need the catalog_uuid in the name, as there is a unique constraint on the offer name. That is what caused this issue in the first place.

I'd recommend

asadiqbal08:
@georgebabey can you review the change set ? I will fix the test failure.

ecommerce/enterprise/forms.py
```diff
@@ -96,11 +96,15 @@ def save(self, commit=True):
 
         # Note: the actual name is not displayed like this in the template, so it's safe to use the UUID here.
         # And in fact we have to, because otherwise we face integrity errors since Oscar forces this name to be unique.
-        self.instance.name = _(u'Discount of type {} provided by {} for {}.'.format(
+        # Truncate enterprise_customer_name down to 68 characters, as Oscar's AbstractConditionalOffer name is
```
georgebabey: Looks like you used 48 characters, which is fine, just comment needs updating

ecommerce/enterprise/forms.py
```diff
@@ -116,14 +120,14 @@ def save(self, commit=True):
 
             if hasattr(self.instance, 'condition'):
                 self.instance.condition.enterprise_customer_uuid = enterprise_customer_uuid
-                self.instance.condition.enterprise_customer_name = enterprise_customer_name
+                self.instance.condition.enterprise_customer_name = enterprise_customer['name']
```
georgebabey: Why did enterprise_customer_name not work here if its what we're using on line 103? Doesn't look like you're modifying it...

georgebabey (approved):
This looks good, just one minor question for my own understanding

ecommerce/enterprise/forms.py
georgebabey: Nevermind, you fixed this before I submitted 👍

ecommerce/enterprise/forms.py
asadiqbal08: ah yes. It is just for testing. I am going to revert these lines.

ecommerce/enterprise/forms.py
```diff
@@ -96,11 +96,15 @@ def save(self, commit=True):
 
         # Note: the actual name is not displayed like this in the template, so it's safe to use the UUID here.
         # And in fact we have to, because otherwise we face integrity errors since Oscar forces this name to be unique.
-        self.instance.name = _(u'Discount of type {} provided by {} for {}.'.format(
+        # Truncate enterprise_customer_name down to 48 characters, as Oscar's AbstractConditionalOffer name is
```
zubair-arbi:
```
Truncate `enterprise_customer_name` to 48 characters so that our complete name with 
format `Discount of type {site} provided by {enterprise_name} for {catalog_uuid}.` does
not exceed the limit of 128 characters for Oscar's `AbstractConditionalOffer` name.
```

ecommerce/enterprise/forms.py
```diff
@@ -96,11 +96,15 @@ def save(self, commit=True):
...
+        # max_length 128
+        offer_name = _(u'Discount of type {} provided by {} for {}.'.format(
             ConditionalOffer.SITE,
-            enterprise_customer_name,
-            enterprise_customer_catalog_uuid,
+            enterprise_customer_name[:48] if len(enterprise_customer_name) > 48 else enterprise_customer_name,  # pylint: disable=unsubscriptable-object,
```
zubair-arbi: We can just do:
```python
enterprise_customer_name[:48]
```

ecommerce/templates/edx/checkout/receipt.html
```diff
@@ -121,7 +121,19 @@ <h2>{% trans "Order Information" %}</h2>
                         </span>
                       {% endwith %}
                     {% else %}
-                      <span class="discount">{{ discount.offer_name }}</span>
+                       {% if offer.condition.enterprise_customer_name %}
+                        <span class="discount">
```
zubair-arbi: nit: It looks like we can move `<span>` tag outside of `if-else` condition.

zubair-arbi (changes requested):

zubair-arbi (approved):
LGTM 👍 🚢

PR #1851, opened 2018-07-06
ENT-1052 apply enterprise offers by provided catalog uuid

zubair-arbi:
**Description:** Add/Show discount for enterprise learners by provided valid enterprise catalog `UUID`. This enterprise catalog `UUID` can be provided by passing the queryparam `catalog={enterprise_catalog_uuid}` to enterprise course enrollment urls. If catalog `UUID` is provided in course enrollment url then the default enterprise coupon (first enterprise offer) will be applied.

**JIRA:** [ENT-1052](https://openedx.atlassian.net/browse/ENT-1052)

**Revert PR:** https://github.com/edx/ecommerce/pull/1848/files

**Dependencies:**

1. edx-enterprise: https://github.com/edx/edx-enterprise/pull/343

**Installation instructions:**  
1. Checkout edx-platform to branch `master`.
2. Install `edx-enterprise` from branch `zub/ENT-1052-pass-catalog-input-only-when-provided` in `edx-platform`.
3. Checkout ecommerce to branch `zub/ENT-1052-apply-enterprise-offer-by-catalog-id`.
4. In ecommerce django admin (admin/waffle/switch/) verify following waffle flags:
```
force_anonymous_user_response_for_basket_calculate = False
enable_enterprise_offers = True
enable_enterprise_on_runtime = True
```
**Testing instructions:**

1. Create an enterprise with 2 catalogs with a verified course (both catalogs should have this verified course)
2. For catalog A, create enterprise offer with 100% discount.
3. For catalog B, create enterprise offer with 15% discount.
4. Now access the enterprise course enrollment URL with the `UUID` of catalog A, i.e. `http://zubair.localhost:8000/enterprise/72416e52-8c77-4860-9584-15e5b06220fb/course/course-v1:edX+TEST_01+2016_R1/enroll/?catalog=a113c7e2-13ac-46ae-9e24-b75bf7c0058d`
5. Verify that learner sees 100 discount on verified seat for verified course
6. Clicking on 'Continue' button (after consenting) user is redirect to courseware page after automatic enrollment (100% discount).
7. Now access the enterprise course enrollment URL with the `UUID` of catalog B, i.e. `http://zubair.localhost:8000/enterprise/72416e52-8c77-4860-9584-15e5b06220fb/course/course-v1:edX+TEST_01+2016_R1/enroll/?catalog=6eca3efb-f3a0-4c08-806f-c6e6b65d61cb`
8. Verify that learner sees 15 discount on verified seat for verified course
9. Clicking on 'Continue' button (after consenting) user is redirect to basket page with 15% discount applied by the learner's enterprise.
10. Now access the enterprise course enrollment URL without the catalog `UUID`, i.e. `http://zubair.localhost:8000/enterprise/72416e52-8c77-4860-9584-15e5b06220fb/course/course-v1:edX+TEST_01+2016_R1/enroll/`
11. Now verify that course enrollment page shows the 100% discount offered by the first enterprise offer.
12. Also verify the clicking on  button (after consenting) user is redirect to courseware page after automatic enrollment (100% discount).

### DIFF
--- a/ecommerce/enterprise/conditions.py
+++ b/ecommerce/enterprise/conditions.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import logging
+from uuid import UUID
 
 from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
@@ -8,9 +9,12 @@ from slumber.exceptions import SlumberHttpBaseException
 
 from ecommerce.enterprise.api import catalog_contains_course_runs, fetch_enterprise_learner_data
 from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_SWITCH
+from ecommerce.extensions.basket.utils import ENTERPRISE_CATALOG_ATTRIBUTE_TYPE
 from ecommerce.extensions.offer.decorators import check_condition_applicability
 from ecommerce.extensions.offer.mixins import ConditionWithoutRangeMixin, SingleItemConsumptionConditionMixin
 
+BasketAttribute = get_model('basket', 'BasketAttribute')
+BasketAttributeType = get_model('basket', 'BasketAttributeType')
 Condition = get_model('offer', 'Condition')
 logger = logging.getLogger(__name__)
 
@@ -30,7 +34,15 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
         Determines if a user is eligible for an enterprise customer offer
         based on their association with the enterprise customer.
 
-        Args:
+        It also filter out the offer if the `enterprise_customer_catalog_uuid`
+        value set on the offer condition does not match with the basket catalog
+        value when explicitly provided by the enterprise learner.
+
+        Note: Currently there is no mechanism to prioritize or apply multiple
+        offers that may apply as opposed to disqualifying offers if the
+        catalog doesn't explicitly match.
+
+        Arguments:
             basket (Basket): Contains information about order line items, the current site,
                              and the user attempting to make the purchase.
         Returns:
@@ -66,6 +78,14 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
 
             course_run_ids.append(course.id)
 
+        # Verify that the current conditional offer is related to the provided
+        # enterprise catalog, this will also filter out offers which don't
+        # have `enterprise_customer_catalog_uuid` value set on the condition.
+        catalog = self._get_enterprise_catalog_uuid_from_basket(basket)
+        if catalog:
+            if offer.condition.enterprise_customer_catalog_uuid != catalog:
+                return False
+
         if not catalog_contains_course_runs(basket.site, course_run_ids, self.enterprise_customer_uuid,
                                             enterprise_customer_catalog_uuid=self.enterprise_customer_catalog_uuid):
             # Basket contains course runs that do not exist in the EnterpriseCustomerCatalogs
@@ -73,3 +93,37 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
             return False
 
         return True
+
+    @staticmethod
+    def _get_enterprise_catalog_uuid_from_basket(basket):
+        """
+        Helper method for fetching valid enterprise catalog UUID from basket.
+
+        Arguments:
+             basket (Basket): The provided basket can be either temporary (just
+             for calculating discounts) or an actual one to buy a product.
+        """
+        # For temporary basket try to get `catalog` from request
+        catalog = basket.strategy.request.GET.get(
+            'catalog'
+        ) if basket.strategy.request else None
+
+        if not catalog:
+            # For actual baskets get `catalog` from basket attribute
+            enterprise_catalog_attribute, __ = BasketAttributeType.objects.get_or_create(
+                name=ENTERPRISE_CATALOG_ATTRIBUTE_TYPE
+            )
+            enterprise_customer_catalog = BasketAttribute.objects.filter(
+                basket=basket,
+                attribute_type=enterprise_catalog_attribute,
+            ).first()
+            if enterprise_customer_catalog:
+                catalog = enterprise_customer_catalog.value_text
+
+        # Return only valid UUID
+        try:
+            catalog = UUID(catalog) if catalog else None
+        except ValueError:
+            catalog = None
+
+        return catalog
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -3,6 +3,7 @@ import json
 import logging
 from urllib import unquote, urlencode
 
+import newrelic.agent
 import pytz
 from django.conf import settings
 from django.contrib import messages
@@ -22,6 +23,7 @@ BasketAttribute = get_model('basket', 'BasketAttribute')
 BasketAttributeType = get_model('basket', 'BasketAttributeType')
 BUNDLE = 'bundle_identifier'
 ORGANIZATION_ATTRIBUTE_TYPE = 'organization'
+ENTERPRISE_CATALOG_ATTRIBUTE_TYPE = 'enterprise_catalog_uuid'
 StockRecord = get_model('partner', 'StockRecord')
 OrderLine = get_model('order', 'Line')
 Refund = get_model('refund', 'Refund')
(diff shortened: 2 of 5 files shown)


ecommerce/extensions/basket/utils.py
```diff
@@ -267,6 +270,35 @@ def basket_add_organization_attribute(basket, request_data):
...
+    is provided in the request.
+
+    Arguments:
+        basket(Basket): order basket
+        request_data (dict): HttpRequest data
+
+    """
+    # Value of enterprise catalog UUID is being passed as `catalog` from
+    # basket page
+    enterprise_catalog_uuid = request_data.get('catalog') if request_data else None
```
brittneyexline, line 286: i think we need to do type checking around this to make sure it is a uuid, and if it isn't, we shouldn't save it as a basket attribute

ecommerce/extensions/basket/utils.py
```diff
@@ -267,6 +270,35 @@ def basket_add_organization_attribute(basket, request_data):
...
+    enterprise_catalog_attribute, __ = BasketAttributeType.objects.get_or_create(
+        name=ENTERPRISE_CATALOG_ATTRIBUTE_TYPE
+    )
+    if enterprise_catalog_uuid:
+        BasketAttribute.objects.update_or_create(
+            basket=basket,
+            attribute_type=enterprise_catalog_attribute,
+            defaults={
+                'value_text': enterprise_catalog_uuid.strip()}
```
brittneyexline: nit: the closing bracket should be on a new line or this entire line should be put with the opening bracket.

ecommerce/enterprise/conditions.py
```diff
@@ -30,6 +33,9 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
         Determines if a user is eligible for an enterprise customer offer
         based on their association with the enterprise customer.
 
+        It also verifies the catalog `catalog` on the
```
brittneyexline: Something I am trying to wrap my head around is how we are handling edge cases in this function.
My understanding of how this ticket was intended to handle the catalog parameter was as follows (and I could be mistaken):

If no catalog parameter is sent, select the first offer that applies in the following priority order: enterprise level, catalog level. If there are multiple catalog level offers that apply, choose the "first" one.

If a catalog parameter is sent, select the first offer that applies in the following priority order: a catalog level offer with a matching catalog, the enterprise level offer, a catalog level offer with a non-matching catalog.

The tricky part here is choosing between multiple offers that may apply, as opposed to disqualifying offers if the catalog doesn't explicitly match. In order to do this, however, I think we would need to override the Applicator in django oscar: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/applicator.py#L59

My guess is that we don't want to do this since this is time sensitive, but I want to call out that there might be edge cases with unexpected behavior, such as if catalogs are sent that don't exist with an offer anymore.

zubair-arbi: > If no catalog parameter is sent, select the first offer that applies in the following priority order: enterprise level, catalog level. If there are multiple catalog level offers that apply, choose the "first" one.

I agree.
> If a catalog parameter is sent, select the first offer that applies in the following priority order: a catalog level offer with a matching catalog, the enterprise level offer, a catalog level offer with a non-matching catalog.

Since we are explicitly providing catalog `UUID` then we will find all enterprise offers related to the provided catalog `UUID` and apply the first one and if there is no catalog or invalid `UUID`  then we don't need to provide any offer otherwise learners can follow the first flow for default offer by not providing `catalog` argument on enterprise course enrollment page. For now our changes are only for enterprise offers.

Also since the catalog `UUID` will be automatically added to enterprise course enrollment urls so we can assume that the urls will always have a catalog with latest offer.

ecommerce/extensions/basket/utils.py
zubair-arbi: @brittneyexline  @georgebabey  So If a learner provides invalid `UUID` then we will be applying the default enterprise offer? (since we will not save the invalid `UUID` as basket attribute and at time of filtering we cannot tell which enterprise catalog user provided)

zubair-arbi: I have update the method `_get_enterprise_catalog_uuid_from_basket` in `ecommerce/enterprise/conditions.py` so the we can validate UUID for both temporary and actual baskets.

ecommerce/enterprise/conditions.py
robrap: I haven't followed all of this, but it sounds like the cases are complicated and not intuitive, so once we have an appropriate summary can we get it into a comment in the code (assuming it is difficult to determine otherwise).  Thanks.

BenjiLee: We've played around with the applicator. This is what it would sort of look like. 
https://github.com/edx/ecommerce/commit/46081e63fba061e7941ea28b04e61a585d0a3f82

Today, for a basket that has any kind of coupon/voucher being added to it, it will run through every single offer on the site. This is not very selective and most likely why there are excessive memcache calls for anything that applies the is_satisfied condition. Some more context https://openedx.atlassian.net/browse/LEARNER-5756

Unless you override the applicator, the code will run through every single offer.

BenjiLee: An example
```
class EnterpriseApplicator(Applicator):

    def get_offers(self, basket, user=None, request=None):
        # These are the defaults
        site_offers = self.get_site_offers()
        basket_offers = self.get_basket_offers(basket, user)
        user_offers = self.get_user_offers(user)
        session_offers = self.get_session_offers(request)

        # This is the enterprise offer selector
        enterprise_offer = self.get_enterprise_offers(basket)

        # This is where you add the offers you want to iterate through.
        return list(sorted(chain(
            session_offers, basket_offers, user_offers, site_offers),
            key=lambda o: o.priority, reverse=True))

        # but I think you only want a single enterprise offer?
        return list(sorted(chain(enterprise_offer),
                           key=lambda o: o.priority, reverse=True))

    @staticmethod
    def get_enterprise_offers(basket):
        # Add logic here that finds the offers based on the input
```

is_satisfied will only be run for the offers found. There are some things we have not sorted out. We need to ensure that this offers list includes every offer that should and can be applied. For example, if an enterprise customer has a coupon for a course that is $15 off, and the user specifically gets another $10 off, then we need to apply both offers for a total of $25. With the way the code is written now, I believe it will only take $15 dollars off.

georgebabey: @BenjiLee Thanks - the Applicator code and WIP PR is very helpful. To be clear is this a recommendation, an idea or a plan moving forward?

BenjiLee: https://openedx.atlassian.net/browse/LEARNER-5756 is going to be moved to the GECOM backlog we would like it to be the plan going forward. 

I think the important context for this PR is that valid multiple offers cannot be applied with the current code in the PR. Performance won't be any worse than it is today.

What doesn't make sense to me about ENT-1052 is why both offers (0% discount and 15% discount) are not being applied. It should be iterating over every offer unless I'm mistaken.

zubair-arbi: @BenjiLee According to my understanding, we loop on all enterprise offers and filter out all the offers which doesn't fulfill the criterion and apply the first one from the selected offers. 

In ENT-1052 we narrowed this filtering criterion by adding the catalog check without changing the other implementation of how coupons apply on basket items.

zubair-arbi: @robrap @georgebabey @brittneyexline @BenjiLee I have updated the description of method `is_satisfied` reflecting that 

> Currently there is no mechanism to prioritize or apply multiple
offers that may apply as opposed to disqualifying offers if the
catalog doesn't explicitly match.

zubair-arbi: FYI @mattdrayer

ecommerce/extensions/basket/utils.py
```diff
@@ -267,6 +270,36 @@ def basket_add_organization_attribute(basket, request_data):
...
+
+    """
+    # Value of enterprise catalog UUID is being passed as `catalog` from
+    # basket page
+    enterprise_catalog_uuid = request_data.get('catalog') if request_data else None
+    enterprise_catalog_attribute, __ = BasketAttributeType.objects.get_or_create(
+        name=ENTERPRISE_CATALOG_ATTRIBUTE_TYPE
+    )
+    if enterprise_catalog_uuid:
+        BasketAttribute.objects.update_or_create(
```
BenjiLee, line 291: ~~Note: It looks like this function is executing a write operation most of the time. If this function is running for every offer, there may be an excessive number of SQL write operations.~~

Edit: nvm. I thought this was the is_satisfied called. Prepare basket is only called once.

BenjiLee (dismissed):

ecommerce/enterprise/conditions.py
BenjiLee: Perhaps a good test to add for this function is  to check if other valid non enterprise offer are applied.

georgebabey (approved):
This looks good, but I do want to address the custom applicator as @BenjiLee  mentions in the near term. Created https://openedx.atlassian.net/browse/ENT-1095 to track that

PR #1853, opened 2018-07-09
Add partner in course model.

waheedahmed:
LEARNER-5052


waheedahmed:
@feanil @robrap Please review, I have tried to remove all the usages of `Course.site` I don't think we need to wrap this in a function.

ecommerce/extensions/api/v2/views/products.py
```diff
@@ -25,7 +25,8 @@ def get_queryset(self):
         # Products are then filtered by:
         #   - stockrecord partner: for products that have stockrecords (seats, coupons, ...)
         #   - course site: for products that don't have a stockrecord (parent course)
```
robrap: course site => course partner

ecommerce/courses/tests/test_create_enrollment_codes_command.py
```diff
@@ -50,8 +50,8 @@ def create_course(self, seat_type='verified'):
         Returns:
             The created course.
         """
-        course = CourseFactory(site=self.site)
-        course.create_or_update_seat(seat_type, False, Decimal(10.0), self.site.siteconfiguration.partner)
+        course = CourseFactory(partner=self.site.siteconfiguration.partner)
```
robrap: Can this just be `self.partner`?

ecommerce/courses/management/commands/publish_to_lms.py
```diff
@@ -41,7 +41,7 @@ def handle(self, *args, **options):
                 try:
                     course_id = course_id.strip()
                     course = Course.objects.get(id=course_id)
-                    publishing_error = course.publish_to_lms()
+                    publishing_error = course.publish_to_lms(course.partner.default_site)
```
robrap: 1. I thought `partner.default_site` was something we were considering dropping.  I need to better understand it.
2. I'm not clear on what the proper relationships should be between a course and its partner, and a course publish to lms, and the site that is used for publishing.
a. Will all sites represent a single lms, and does it not matter which is used?
b. Should `publish_to_lms` just use the course.partner.default_site internally, if that is something that will always exist?
c. Below, we call `publish_to_lms(request.site)`? Do we care if the request.site has no relationship with the course partner's sites?  Would we need an assertion somewhere?

robrap (commented):
@feanil: Please don't take this as a fully comprehensive review.  I didn't look at migrations or other things, and used search to check certain specific questions.

Thanks @waheedahmed.  Sorry about the delay.  I had to keep coming back to this.

ecommerce/courses/migrations/0008_course_partner.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
```
feanil, line 1: - Separate the 3 migraitons here to be separate.
  - One for the addition of the field,
  - One for making the site optional
  - One for making doing the data migration.
- The data migration should have a reverse that updates the course.site.siteconfiguration.partner with the current value of the course.partner

ecommerce/courses/publishers.py
```diff
@@ -97,7 +98,7 @@ def publish(self, course):
                 'modes': modes,
             }
 
-            commerce_api_client = course.site.siteconfiguration.commerce_api_client
+            commerce_api_client = site.siteconfiguration.commerce_api_client
```
feanil, line 100: Why not just use course.partner.default_site?  Then you don't need to pass the extra parameter.

feanil (changes requested):
Generaly looks good, just a few questions and some changes needed in the migrations.

ecommerce/courses/publishers.py
robrap: @feanil: Does this mean `partner.default_site` is here to stay?  Also, does it need to be a specific site, or is that equivalent to something like `partner.any_one_of_my_sites`?

feanil: I don't remember if we came to any decision on that.  I could see arguments for just removing it and always being explicit as well.

ecommerce/courses/management/commands/publish_to_lms.py
waheedahmed: @robrap Yes we are, I'm just trying to use the current logic here. We will think about it later with a better resolution.

waheedahmed:
@robrap @feanil Addressed all comments, kindly take another look.

ecommerce/courses/publishers.py
waheedahmed: @robrap No, I'm just using the available logic for now, it will be cleaned up in follow up ticket with a better logic.

ecommerce/courses/management/commands/publish_to_lms.py
robrap: Sounds good @waheedahmed.  When we get to that ticket, let's ensure we agree on a design/plan before you implement.  It's not clear to me what is the right thing to do you on this front.  Thanks!

ecommerce/courses/migrations/0009_auto_20180727_0751.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
```
feanil: give this migration a better name, maybe `0009_allow_site_to_be_nullable.py`

ecommerce/courses/migrations/0010_auto_20180727_0757.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
```
feanil: Give this migration a better name, maybe `0010_migrate_partner_data_to_courses.py`

feanil (commented):
Two file renames and then I think this is good to go.  Don't forget to update the dependencies when you rename the auto migrations.

ahsan-ul-haq (approved):
LGTM

waheedahmed:
@feanil Renamed the migrations also tested locally with multiple site objects against same partner and course api returning same data for both sites. Kindly take another look.

feanil (approved):

PR #1855, opened 2018-07-09
Revert "Skip receipt page for free order"

robrap:
This reverts commit 979dc88

See the discussion about why this is being reverted here:
https://github.com/edx/ecommerce/pull/1676#issuecomment-403252390


robrap:
FYI: @coryleeio

robrap:
@edx/business-enterprise-team: I imagine that the enterprise code that does something similar probably has the same flaky problem for learners, which is probably not a great experience.  You may want to look at the discussion here: https://github.com/edx/ecommerce/pull/1676#issuecomment-403252390

Also, if you already have a solution to this and/or add a solution, maybe you could let us know and we could reuse it for this reverted PR.  Thanks.

PR #1859, opened 2018-07-13
Add Bundle ID to event tracking

BenjiLee:
Is it as simple as this?

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -31,6 +31,7 @@ from ecommerce.extensions.analytics.utils import (
 )
 from ecommerce.extensions.basket.constants import EMAIL_OPT_IN_ATTRIBUTE
 from ecommerce.extensions.basket.utils import (
+    BUNDLE,
     add_utm_params_to_url,
     apply_voucher_on_basket_and_check_discount,
     get_basket_switch_data,
@@ -367,6 +368,13 @@ class BasketSummaryView(BasketView):
                 'cart_id': basket.id,
                 'products': [translate_basket_line_for_segment(line) for line in basket.all_lines()],
             }
+            attributes = BasketAttribute.objects.filter(
+                basket=basket,
+                attribute_type__name=BUNDLE
+            )
+            if len(attributes) == 1:
+                properties['bundle_id'] = attributes[0].value_text
+
             track_segment_event(request.site, request.user, 'Cart Viewed', properties)
 
             properties = {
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/basket/views.py
```diff
@@ -336,9 +336,11 @@ def _get_payment_processors_data(self, payment_processors):
 
     def get(self, request, *args, **kwargs):
         basket = request.basket
+        bundle_id = request.GET.get('bundle', None)
```
BenjiLee: Note to self: Only the UTM params are passed as query parameters to BasketSummary from BasketAdd. Perhaps I can store the bundle ID in the request cache?

BenjiLee: Note to self: Only the UTM params are passed as query parameters to BasketSummary from BasketAdd. Perhaps I can store the bundle ID in the request cache?

robrap: Isn't the bundle id stored in a basket attribute?

BenjiLee: yes. yes it is.
```
SELECT value_text from ecommerce.basket_basketattribute WHERE attribute_type_id=2;
```
returns the bundle-id

robrap:
@BenjiLee: Giving up? :(

robrap:
@BenjiLee: Can you remind me what you couldn't do and what you were trying to fix with this?  I spoke to the Data Science team, and they think they have ways, so just curious.  Thank you!

BenjiLee:
7 years later... 

I think the original goal was to see how our 20s to 2s performance improvement for programs impacted the bounce rate of programs. After we made the improvement, it was found that we do not track programs separately from courses so we couldn't answer that question. This PR was to add the bundle-id so that we can see just programs in an analytics filter/query.

robrap:
Thank you @BenjiLee. I see that this was for ‘Cart Viewed’ and ‘Checkout Step Viewed’. Lily said they have generally been determining this based on the product type. Is that what you were told, but the joins were too complicated, or were you trying to figure it out another way? In any case, I was seeing if we could have the Data Science team support getting this through if it would be useful to them.

BenjiLee:
I was told that it was possible to get the information I was looking for, I just needed to get access to the data warehouse pieces and learn how to do the query. I thought this was a a high barrier to entry for finding out something like this, thus this PR.

robrap:
Thanks @BenjiLee.
Note: @cpennington may try to actually land some of your old PRs, including the apply CustomApplicator everywhere, which still hasn't landed.  If you have any other PRs you closed that @cpennington should know about, please feel free to share. :)

cpennington:
I just chatted with @lxedx, and I don't think we need to add this to segment. We can analyze the downstream data and know in the data warehouse whether a user was purchasing a bundle or not.

@BenjiLee (or @robrap), when did the performance change land? I'd be interested in actually doing the before/after analysis, because the question of whether improving perf improves revenue is important to us.

robrap:
@cpennington: I forwarded you an email thread titled "Learner Performance Team Update for Ecommerce".

PR #1862, opened 2018-07-18
Fix for LEARNER-5603, LEARNER-5630, LEARNER-5876

Ayub-Khan:
[LEARNER-5603](https://openedx.atlassian.net/browse/LEARNER-5603)
[LEARNER-5630](https://openedx.atlassian.net/browse/LEARNER-5630)
[LEARNER-5876](https://openedx.atlassian.net/browse/LEARNER-5876)

Kindly check tickets for more details. 

Tests coming soon.

Optional Review : @edx/learner-spartans

### DIFF
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -8,8 +8,9 @@ import mock
 import pytz
 import requests
 from django.db import transaction
+from django.utils.timezone import now
 from oscar.core.loading import get_model
-from oscar.test.factories import BasketFactory, ProductFactory, RangeFactory, VoucherFactory
+from oscar.test.factories import BasketFactory, ProductFactory, RangeFactory
 
 from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME
 from ecommerce.core.tests import toggle_switch
@@ -108,13 +109,14 @@ class BasketUtilsTests(DiscoveryTestMixin, BasketMixin, TestCase):
 
     def test_multiple_vouchers(self):
         """ Verify only the last entered voucher is contained in the basket. """
-        product = ProductFactory()
-        voucher1 = VoucherFactory(code='FIRST')
+        product = ProductFactory(stockrecords__price_excl_tax=100)
+        new_range = RangeFactory(products=[product, ])
+        voucher1, __ = prepare_voucher(code='TEST1', _range=new_range, benefit_value=10)
         basket = prepare_basket(self.request, [product], voucher1)
         self.assertEqual(basket.vouchers.count(), 1)
         self.assertEqual(basket.vouchers.first(), voucher1)
 
-        voucher2 = VoucherFactory(code='SECOND')
+        voucher2, __ = prepare_voucher(code='TEST2', _range=new_range, benefit_value=20)
         new_basket = prepare_basket(self.request, [product], voucher2)
         self.assertEqual(basket, new_basket)
         self.assertEqual(new_basket.vouchers.count(), 1)
@@ -388,8 +390,10 @@ class BasketUtilsTests(DiscoveryTestMixin, BasketMixin, TestCase):
         """
         Test prepare_basket clears vouchers for a bundle
         """
-        product = ProductFactory()
-        voucher = VoucherFactory(code='FIRST')
+        product = ProductFactory(stockrecords__price_excl_tax=100)
+        new_range = RangeFactory(products=[product, ])
+        voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
+
         request = self.request
         basket = prepare_basket(request, [product], voucher)
         self.assertTrue(basket.vouchers.all())
@@ -488,6 +492,84 @@ class BasketUtilsTests(DiscoveryTestMixin, BasketMixin, TestCase):
                     self.assertEqual(mock_track.call_count, 1)
                 else:
                     mock_track.assert_not_called()
+
+    def test_prepare_basket_ignores_invalid_voucher(self):
+        """
+        Tests that prepare_basket validates provided voucher and
+        does not applies it if invalid.
+        """
+        voucher_start_time = now() - datetime.timedelta(days=5)
+        voucher_end_time = now() - datetime.timedelta(days=3)
+        product = ProductFactory(stockrecords__partner__short_code='test1', stockrecords__price_excl_tax=100)
+        expired_voucher, __ = prepare_voucher(start_datetime=voucher_start_time, end_datetime=voucher_end_time)
+
+        basket = prepare_basket(self.request, [product], expired_voucher)
+        self.assertIsNotNone(basket)
+        self.assertEqual(basket.status, Basket.OPEN)
+        self.assertEqual(basket.lines.count(), 1)
+        self.assertEqual(basket.lines.first().product, product)
+        self.assertEqual(basket.vouchers.count(), 0)
+
+    def test_prepare_basket_applies_valid_voucher_argument(self):
+        """
+            Tests that prepare_basket applies a valid voucher passed as an
+            an argument, even when there is also a valid voucher already on
+            the basket.
+        """
+        product = ProductFactory(stockrecords__partner__short_code='test1', stockrecords__price_excl_tax=100)
+        new_range = RangeFactory(products=[product])
+        new_voucher, __ = prepare_voucher(code='xyz', _range=new_range, benefit_value=10)
+        existing_voucher, __ = prepare_voucher(code='test', _range=new_range, benefit_value=50)
+
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        basket.vouchers.add(existing_voucher)
+        self.assertEqual(basket.vouchers.count(), 1)
+
+        basket = prepare_basket(self.request, [product], new_voucher)
+        self.assertIsNotNone(basket)
+        self.assertEqual(basket.vouchers.count(), 1)
+        self.assertEqual(basket.vouchers.first().code, 'XYZ')
+        self.assertEqual(basket.total_discount, 10.00)
+
+    def test_prepare_basket_removes_existing_basket_invalid_voucher(self):
+        """
+        Tests that prepare_basket removes an existing basket voucher that is invalid
+        for multiple products when used to purchase any of those products.
+        """
+        voucher_start_time = now() - datetime.timedelta(days=5)
+        voucher_end_time = now() - datetime.timedelta(days=3)
+        product = ProductFactory(stockrecords__partner__short_code='xyz', stockrecords__price_excl_tax=100)
+        expired_voucher, __ = prepare_voucher(start_datetime=voucher_start_time, end_datetime=voucher_end_time)
+
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        basket.vouchers.add(expired_voucher)
+
+        self.assertEqual(basket.vouchers.count(), 1)
+        basket = prepare_basket(self.request, [product])
+        self.assertIsNotNone(basket)
+        self.assertEqual(basket.lines.first().product, product)
+        self.assertEqual(basket.vouchers.count(), 0)
+
+    def test_prepare_basket_applies_existing_basket_valid_voucher(self):
+        """
+        Tests that prepare_basket applies an existing basket voucher that is valid
+        for multiple products when used to purchase any of those products.
+        """
+        product = ProductFactory(stockrecords__partner__short_code='test1', stockrecords__price_excl_tax=100)
+
+        new_range = RangeFactory(products=[product])
+        voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
+
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        basket.vouchers.add(voucher)
+        self.assertEqual(basket.vouchers.count(), 1)
+
+        basket = prepare_basket(self.request, [product])
+        self.assertIsNotNone(basket)
+        self.assertEqual(basket.vouchers.count(), 1)
+        self.assertEqual(basket.lines.first().product, product)
+        self.assertIsNotNone(basket.applied_offers())
+        self.assertEqual(basket.total_discount, 10.00)
 
 
 class BasketUtilsTransactionTests(TransactionTestCase):
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -27,6 +27,7 @@ ENTERPRISE_CATALOG_ATTRIBUTE_TYPE = 'enterprise_catalog_uuid'
 StockRecord = get_model('partner', 'StockRecord')
 OrderLine = get_model('order', 'Line')
 Refund = get_model('refund', 'Refund')
+Voucher = get_model('voucher', 'Voucher')
 
 logger = logging.getLogger(__name__)
 
@@ -104,11 +105,14 @@ def prepare_basket(request, products, voucher=None):
 
     if len(products) == 1 and products[0].is_enrollment_code_product:
         basket.clear_vouchers()
-    elif voucher:
+    elif voucher or basket.vouchers.exists():
+        voucher = voucher or basket.vouchers.first()
         basket.clear_vouchers()
-        basket.vouchers.add(voucher)
-        Applicator().apply(basket, request.user, request)
-        logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
+        is_valid, message = validate_voucher(voucher, request.user, basket, request.site)
+        if is_valid:
+            _apply_voucher_on_basket(voucher, request, basket)
+        else:
+            logger.info(message)
 
     attribute_cookie_data(basket, request)
     return basket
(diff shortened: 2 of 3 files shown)


tasawernawaz:
Change make sense to me. It would be great if you can add tests too, also if it is feasible please create sandbox to test this.

ecommerce/extensions/basket/models.py
```diff
@@ -62,6 +62,9 @@ def flush(self):
                 properties = translate_basket_line_for_segment(line)
                 track_segment_event(self.site, self.owner, 'Product Removed', properties)
 
+        # Remove all vouchers
+        self.clear_vouchers()
```
robrap: This doesn't feel right.  Some questions:
1. From your description, it sounds like prepare_basket calls the flush and could also clear the vouchers.
2. I'm unclear if this temporarily clears the vouchers (just like the basket being emptied), and then it adds it back with the proper discount?  Or, does the user have to add the voucher again?  What would Product prefer?
3. Agreed.  Please add a test for whatever we decide.
Thank you!

jmyatt: Adding to @robrap 's comments, I agree that this should live in prepare_basket along with the other code there that does the following:
  * flushes products via this Basket.flush method and conditionally adds them back
  * conditionally clears / adds vouchers

Specifically, maybe [this code block](https://github.com/edx/ecommerce/blob/d7f5586b9b3c1ece5d7423563be37a1a5f41e4ef/ecommerce/extensions/basket/utils.py#L105-L111) could be augmented to something like:

    if len(products) == 1 and products[0].is_enrollment_code_product:
        basket.clear_vouchers()
    elif voucher:
        basket.clear_vouchers()
        basket.vouchers.add(voucher)
        logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
    elif basket.vouchers.exists():
        for voucher in basket.vouchers.all():
            if not voucher.is_active():
                basket.vouchers.remove(voucher)

(the last 4 lines being the new logic).  But if we do that it probably makes sense to include all the logic you recently added to check_valid_voucher

Ayub-Khan: @robrap 
2. This does not clear the vouchers added by user. It clears the vouchers already existing in the basket obj. 
3. @jmyatt  Agree but we will need to reapply the voucher too if we want to keep the valid voucher added in the basket.

ecommerce/extensions/basket/utils.py
```diff
@@ -106,9 +107,21 @@ def prepare_basket(request, products, voucher=None):
         basket.clear_vouchers()
     elif voucher:
         basket.clear_vouchers()
-        basket.vouchers.add(voucher)
-        Applicator().apply(basket, request.user, request)
-        logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
+        is_valid, message = validate_voucher(voucher, request.user, basket, request.site)
```
Ayub-Khan, line 111: Fixes the issue LEARNER-5876

ecommerce/extensions/basket/utils.py
```diff
@@ -106,9 +107,21 @@ def prepare_basket(request, products, voucher=None):
...
+        if is_valid:
+            _apply_voucher_on_basket(voucher, request, basket)
+        else:
+            logger.info(message)
+    elif basket.vouchers.exists():
+        valid_vouchers = []
```
Ayub-Khan: Just in case if we start using Multiple voucher functionality for single basket.

ecommerce/extensions/basket/utils.py
```diff
@@ -106,9 +107,21 @@ def prepare_basket(request, products, voucher=None):
...
         basket.clear_vouchers()
-        basket.vouchers.add(voucher)
-        Applicator().apply(basket, request.user, request)
-        logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
+        is_valid, message = validate_voucher(voucher, request.user, basket, request.site)
+        if is_valid:
+            _apply_voucher_on_basket(voucher, request, basket)
+        else:
+            logger.info(message)
+    elif basket.vouchers.exists():
```
Ayub-Khan: Checks if there is a existing voucher which was added before in the basket. Validate it and apply again if its valid now and active now and can be used for new products. LEARNER-5603, LEARNER-5630

ecommerce/extensions/basket/utils.py
```diff
@@ -326,3 +339,73 @@ def _set_basket_bundle_status(bundle, basket):
         basket.clear_vouchers()
     else:
         BasketAttribute.objects.filter(basket=basket, attribute_type__name=BUNDLE).delete()
+
+
+def validate_voucher(voucher_code, user, basket, request_site):
```
Ayub-Khan: Moved Voucher validate for basket code to this new method to avoid code duplication in VoucherAddView and prepare_basket.

ecommerce/extensions/basket/utils.py
```diff
@@ -326,3 +339,73 @@ def _set_basket_bundle_status(bundle, basket):
...
+    is_voucher_valid_for_bundle = voucher_program_uuid or voucher.usage == Voucher.MULTI_USE
+
+    if is_bundle_purchase and not is_voucher_valid_for_bundle:
+        message = _("Coupon code '{code}' is not valid for this basket.").format(code=voucher_code)
+        return False, message
+
+    return True, ''
+
+
+def _apply_voucher_on_basket(voucher, request, basket):
```
Ayub-Khan, line 389: Created this to avoid code duplication in prepare basket method.

robrap:
This is probably already planned, but can you add at least one test per ticket that confirms that the issue is fixed.  All new tests should fail against master and pass for this PR.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -489,6 +493,66 @@ def test_basket_switch_data_for_non_course_run_products(self):
                 else:
                     mock_track.assert_not_called()
 
+    def test_prepare_basket_ignores_invalid_voucher(self):
+        """
+        Tests that prepare_basket validates provided voucher and
+        not applies it if invalid.
```
robrap: Should be “does not apply it if invalid.”

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -489,6 +493,66 @@ def test_basket_switch_data_for_non_course_run_products(self):
...
+        """
+        voucher_start_time = now() - datetime.timedelta(days=5)
+        voucher_end_time = now() - datetime.timedelta(days=3)
+        product = ProductFactory(stockrecords__partner__short_code='test1', stockrecords__price_excl_tax=100)
+        voucher, __ = prepare_voucher(start_datetime=voucher_start_time, end_datetime=voucher_end_time)
```
robrap: Can you rename `voucher` to `expired_voucher` for clarity?

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -489,6 +493,66 @@ def test_basket_switch_data_for_non_course_run_products(self):
...
+
+    def test_prepare_basket_removes_old_basket_voucher_if_invalid(self):
+        """
+        Tests that prepare_basket removes existing voucher from basket
+        if not valid for new product.
+        """
+        voucher_start_time = now() - datetime.timedelta(days=5)
+        voucher_end_time = now() - datetime.timedelta(days=3)
+        product = ProductFactory(stockrecords__partner__short_code='test1', stockrecords__price_excl_tax=100)
+        voucher, voucher_product = prepare_voucher(start_datetime=voucher_start_time, end_datetime=voucher_end_time)
```
robrap: expired_voucher, _ = ...

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -489,6 +493,66 @@ def test_basket_switch_data_for_non_course_run_products(self):
...
+        
+        basket = prepare_basket(self.request, [product])
+        self.assertIsNotNone(basket)
+        self.assertEqual(basket.lines.first().product, product)
+        self.assertEqual(basket.vouchers.count(), 0)
+
+    def test_prepare_basket_applies_old_basket_voucher_if_valid(self):
+        """
+        Tests that prepare_basket applies the old voucher on new product
+        if its valid for new product.
```
robrap: What does new or old mean here? Is this a test that a voucher made for two products is valid for either one?  Can you better explain this test?

ecommerce/extensions/basket/utils.py
robrap: The doc string for this method states that we will only allow up to one voucher, but this code is not written as such.  Is the docstring accurate?  

Also, these two `elif` sections are extremely similar and not DRY.  Could we have a single `elif voucher or basket.vouchers.exists():`.  In this new `elif`, I would create a list with the first element as the passed in voucher, and the rest of the list being the basket vouchers.  Once you have this list, you can clear the basket vouchers. Then loop over the list in order until you find a valid voucher (or reach the end), and break after the first valid voucher (if we only want a max of one voucher list the docstring says).  Note: If we are not sure about the business rule, we can ask.

ecommerce/extensions/basket/utils.py
```diff
@@ -326,3 +340,72 @@ def _set_basket_bundle_status(bundle, basket):
...
+
+    if not voucher.is_active():
+        message = _("Coupon code '{code}' is not active.").format(code=voucher.code)
+        return False, message
+
+    is_available, message = voucher.is_available_to_user(user)
+
+    if not is_available:
+        if voucher.usage == Voucher.SINGLE_USE:
+            message = _("Coupon code '{code}' has already been redeemed.").format(code=voucher.code)
```
robrap: Does this custom message belong in `voucher.is_available_to_user()`?

ecommerce/extensions/basket/utils.py
```diff
@@ -326,3 +340,72 @@ def _set_basket_bundle_status(bundle, basket):
...
+        return False, message
+
+    # Do not allow coupons for one site to be used on another site
+    voucher_site = voucher.offers.first().site
+    if request_site and voucher_site and request_site != voucher_site:
+        message = _("Coupon code '{code}' is not valid for this basket.").format(code=voucher.code)
```
robrap, line 370: If `voucher_site` exists, are we ok if `request_site` is None, or would that still be invalid? Note: I know that this code did not originate in this PR, so this isn't a blocker.  But, I would like to get this answered.

ecommerce/extensions/basket/utils.py
```diff
@@ -326,3 +340,72 @@ def _set_basket_bundle_status(bundle, basket):
...
+
+    # Do not allow single course run coupons used on bundles.
+    bundle_attribute = BasketAttribute.objects.filter(
+        basket=basket,
+        attribute_type=BasketAttributeType.objects.get(name=BUNDLE)
+    )
+    is_bundle_purchase = len(bundle_attribute) > 0
+    voucher_program_uuid = voucher.offers.first().condition.program_uuid
+    is_voucher_valid_for_bundle = voucher_program_uuid or voucher.usage == Voucher.MULTI_USE
+
```
robrap, line 381: @georgebabey @BessieSteinberg @bill-filler: FYI: I'm not sure how Journal and Enterprise vouchers are handling this case, and whether it is handled elsewhere, or if we need more expanded logic around this here, since it is only looking at programs?

ecommerce/extensions/basket/utils.py
```diff
@@ -326,3 +340,72 @@ def _set_basket_bundle_status(bundle, basket):
...
+    if is_bundle_purchase and not is_voucher_valid_for_bundle:
+        message = _("Coupon code '{code}' is not valid for this basket.").format(code=voucher.code)
```
robrap, line 383: Please update the message to provide more detail about the problem.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -108,13 +109,14 @@ def test_prepare_basket_enrollment_with_voucher(self):
...
-        voucher1 = VoucherFactory(code='FIRST')
+        product = ProductFactory(stockrecords__price_excl_tax=100)
+        new_range = RangeFactory(products=[product, ])
+        voucher1, __ = prepare_voucher(code='TEST1', _range=new_range, benefit_value=10)
         basket = prepare_basket(self.request, [product], voucher1)
         self.assertEqual(basket.vouchers.count(), 1)
         self.assertEqual(basket.vouchers.first(), voucher1)
 
-        voucher2 = VoucherFactory(code='SECOND')
+        voucher2, __ = prepare_voucher(code='TEST2', _range=new_range, benefit_value=20)
```
robrap, line 119: 1. Why does `_range` have an underscore prefix? (If that code is outside the PR, I'll leave it to you to update.)
2. Can you explain why the range and benefit value need to be added for some of these tests?

I may recommend wrapping this in a helper with a name that clarifies what kind of voucher this is, but that depends on the answer to the second question.

robrap (changes requested):
It will be great to have these fixes.  I've got a lot of comments, but they are mostly minor.  Let me know if you think otherwise and we can discuss.  Thanks.

ecommerce/extensions/basket/tests/test_utils.py
Ayub-Khan: 1. its the % by default its 100% so in case of valid coupons tests we check if the basket price is updated or not.

ecommerce/extensions/basket/utils.py
Ayub-Khan: yes we do only allow one voucher. I wrote it bcz of forward compatibility. To be more specific we only count the last added voucher. We have a test for that too. i have updated the code too.

ecommerce/extensions/basket/utils.py
Ayub-Khan: Im not sure if i understand what you mean. But we can move this there. Here its also kinda replacing the error msg returned by is_available_to_user.

ecommerce/extensions/basket/utils.py
Ayub-Khan: Im not sure if we will reach this point if request.site is none. so even if some how we reach this point this should be invalid.

ecommerce/extensions/basket/utils.py
Ayub-Khan: I think we should get opinion from bundling people if its ok for us to update those error messages. Also this message appears to students too.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -489,6 +493,67 @@ def test_basket_switch_data_for_non_course_run_products(self):
...
+        basket = prepare_basket(self.request, [product])
+        self.assertIsNotNone(basket)
+        self.assertEqual(basket.lines.first().product, product)
+        self.assertEqual(basket.vouchers.count(), 0)
+
+    def test_prepare_basket_applies_old_basket_valid_voucher(self):
+        """
+        Tests that prepare_basket applies the old voucher on new product
+        if its valid for new product.
+        """
```
robrap: I'm not sure if it is the name of this test that is incorrect, or the docstring, or the code.  It depends what we wish to test.  

The test before this is named `test_prepare_basket_removes_old_basket_invalid_voucher` and it uses `basket.vouchers.add(expired_voucher)`.  I imagine that `old_basket_invalid_voucher` means `existing_basket_invalid_voucher`.  In this test, we use the name `test_prepare_basket_applies_old_basket_valid_voucher`, but we don't have the line `basket.vouchers.add(voucher)`, and we probably should, rather than sending it into `prepare_basket`.  If we need a separate test for sending it in to `prepare_basket`, we could also have that.

In the docstring, I think `old voucher` just means `existing basket voucher`, right?  Also, I don't think `new` in `new product` signifies anything, unless I am missing something.  It looks like the following would be a better docstring:
```
Tests that prepare_basket applies an existing basket voucher that is valid
for multiple products when used to purchase any of those products.
```
NOTE: The phrase `existing basket voucher` only applies if you add the line `basket.vouchers.add(voucher)`.  If we don't, or we need an additional test for passing the voucher, we could use something like:
```
Tests that prepare_basket applies that passed voucher that is valid
for multiple products when used to purchase any of those products.
```

Let me know if this all makes sense.  Thanks.

ecommerce/extensions/basket/utils.py
```diff
@@ -104,11 +105,14 @@ def prepare_basket(request, products, voucher=None):
 
     if len(products) == 1 and products[0].is_enrollment_code_product:
         basket.clear_vouchers()
-    elif voucher:
+    elif voucher or basket.vouchers.exists():
+        voucher = voucher or basket.vouchers.last()
```
robrap: If voucher exists, but is expired, wouldn't we want the basket voucher?  Or, do we only want the voucher that was passed in if it was passed in?  Also, if the basket had 2 vouchers (not sure if it could), and the last one is expired, but the other one is not, wouldn't we want the unexpired one?

Your code may or may not be ok depending on the answers to the above question.  

If we do need to handle more complicated rules, we could have something like:
```
voucher_list = []
voucher_list.append(voucher)
voucher_list.extend(basket.vouchers)
voucher_list = filter(None, voucher_list)
basket.clear_vouchers()
for voucher in voucher_list:
    is_valid, message = validate_voucher(voucher, request.user, basket, request.site)
    if is_valid:
        _apply_voucher_on_basket(voucher, request, basket)
        break
    else:
        logger.info(message)
```
NOTE: Do not update your code unless we need to support the more complicated behavior.  Your code is simpler to write and test. :)

ecommerce/extensions/basket/utils.py
robrap: You do not have to do this.  I feel like it belongs in an overridden method somewhere so that anyone calling this would get the more complete message, but you were just moving this code.  You can leave this as-is.

ecommerce/extensions/basket/utils.py
robrap: Not a blocker.  You can leave as-is for now.

ecommerce/extensions/basket/tests/test_utils.py
robrap: I still don't see where the added `benefit_value` is used in this particular test.  But, if it is necessary for some reason, you can leave it.

robrap (commented):
I wrote a lot of text, but I'm not asking for much.  This is close.  Thank you!

ecommerce/extensions/basket/utils.py
robrap: ok

ecommerce/extensions/basket/tests/test_utils.py
Ayub-Khan: its in the last assert to check the discount value.

ecommerce/extensions/basket/tests/test_utils.py
Ayub-Khan: Updated.

ecommerce/extensions/basket/utils.py
Ayub-Khan: @robrap we prefer the last added voucher over already existing voucher. Also we do not usually pass voucher code directly into baskets. So it will be a rare case. i think we pass voucher directly to basket add page from coupon redeem pages so there is a more less chance its going to be invalid. I do think we will need to use the above mentioned code when we start using multiple vouchers for same basket.

robrap: @Ayub-Khan: Please change from `vouchers.last` to `vouchers.first` for consistency with the rest of the code.  I'm not sure which is more accurate, or if there could only be 1 so it doesn't matter, so I'd like to retain consistency.

FYI: I'll follow-up with Product around the current and future status of handling multiple vouchers.  I want to document in ecommerce if we are committed to always handling a max of 1 voucher, or if we should handle multiple.  We can always add another PR if needed, but don't want to block this one.

Note: In the future, if we wanted to handle multiple here, a simpler version of what I wrote would be: `vouchers = [voucher] or basket.vouchers`, which is a mix of yours and mine.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -489,6 +493,63 @@ def test_basket_switch_data_for_non_course_run_products(self):
...
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        basket.vouchers.add(expired_voucher)
+
+        self.assertEqual(basket.vouchers.count(), 1)
+        basket = prepare_basket(self.request, [product])
+        self.assertIsNotNone(basket)
+        self.assertEqual(basket.lines.first().product, product)
+        self.assertEqual(basket.vouchers.count(), 0)
+
+    def test_prepare_basket_applies_existing_basket_valid_voucher(self):
```
robrap, line 553: I think we are missing a test, but let me know if I am wrong.  Here would be the new test:
```
def test_prepare_basket_applies_valid_voucher_argument(self):
    """
    Tests that prepare_basket applies a valid voucher passed as an
    an argument, even when there is also a valid voucher already on
    the basket.
    """
```
Note: You would need to create two different valid vouchers, add one to the basket and pass one to `prepare_basket`, and ensure it was the `prepare_basket` one that gets used.
However, I'm basing this on you stating that this is what should happen.  I don't really know. :)

robrap (dismissed):
Very close.  Just a missing test and a very minor change.

robrap (commented):
@Ayub-Khan: This is approved if you add the missing test as asked and make the other minor change from `last` to `first`.  Just squash and rebase and you can get @waheedahmed to merge and you can test on Stage and then deploy.  I don't want you to have to wait another day for me.
Thanks!

ecommerce/extensions/basket/utils.py
Ayub-Khan: Agree

ecommerce/extensions/basket/tests/test_utils.py
Ayub-Khan: done.

waheedahmed (approved):

PR #1863, opened 2018-07-18
Fix UnicodeError when getting enrolment codes - ENT-1079

mushtaqak:
This PR fixes unicode error when user tries to get enrolment codes through order completion email.

[ENT-1079](https://openedx.atlassian.net/browse/ENT-1079)

### DIFF
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -1,8 +1,8 @@
 from __future__ import unicode_literals
 
-import csv
 import logging
 
+import unicodecsv as csv
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.core.exceptions import PermissionDenied
--- a/ecommerce/coupons/tests/test_views.py
+++ b/ecommerce/coupons/tests/test_views.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import datetime
 import urllib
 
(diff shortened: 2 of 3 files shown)


mushtaqak:
@zubair-arbi @muhammad-ammar This is ready for review.

ecommerce/coupons/views.py
```diff
@@ -319,7 +319,7 @@ def get(self, request, number):
 
         order_line_vouchers = OrderLineVouchers.objects.filter(line__order=order)
         for order_line_voucher in order_line_vouchers:
-            writer.writerow([order_line_voucher.line.product.title])
+            writer.writerow([order_line_voucher.line.product.title.encode('utf-8')])
```
Qubad786: We should use [unicodecsv](https://github.com/jdunck/python-unicodecsv):

At L316, `writer` would be:
```bash
import unicodecsv as csv
writer = csv.writer(response, encoding='utf-8')
```

Qubad786 (commented):
Minor concerns otherwise LGTM 👍

muhammad-ammar (approved):

ecommerce/coupons/views.py
mushtaqak: We aren't using `unicodecsv` in ecommerce, so I guess I am not inclined to add it for now at-least.

zubair-arbi (approved):
Nice 👍

ecommerce/coupons/views.py
robrap: Good call @Qubad786.  Based on [usage across the platform](https://github.com/search?p=3&q=org%3Aedx+unicodecsv&type=Code), it seems like this is the way to go.  Can you add `unicodecsv` to avoid future issues?

mushtaqak:
@georgebabey Can you review and merge this (if all is good :P) ?

ecommerce/coupons/views.py
georgebabey: Agreed, we should switch to using the unicodecsv as @Qubad786 indicates. This seems like a small lift to bring this repo in line with the other edx projects.  Also seems more robust in that this just fixes encoding for the product title, but if another field was added we'd have to remember to encode there as well.

georgebabey (changes requested):

georgebabey (approved):

robrap:
Great.  Thanks @mushtaqak!

PR #1864, opened 2018-07-19
LEARNER-5864

Ayub-Khan:
[LEARNER-5864](https://openedx.atlassian.net/browse/LEARNER-5864)
Fixed missing order issue in case of duplicate reference number exception.
Check ticket description for more details. 
@edx/ecommerce , @edx/learner-spartans  Kindly Review.

### DIFF
--- a/ecommerce/extensions/payment/views/cybersource.py
+++ b/ecommerce/extensions/payment/views/cybersource.py
@@ -249,21 +249,12 @@ class CybersourceNotificationMixin(CyberSourceProcessorMixin, OrderCreationMixin
                 )
                 raise
             except DuplicateReferenceNumber:
-                if Order.objects.filter(number=order_number).exists() or PaymentProcessorResponse.objects.filter(
-                        basket=basket).exclude(transaction_id__isnull=True).exclude(transaction_id='').exists():
-                    logger.info(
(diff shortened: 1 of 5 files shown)


staubina (approved):
This seems like a nice way to tie up a loose end.

Please clear up testing issues also.

robrap (commented):
Agreed this looks good, but please add a test or assertions for this case. Thank you.

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -249,21 +249,13 @@ def validate_notification(self, notification):
...
-                    )
-                else:
-                    logger.info(
-                        'Received duplicate CyberSource payment notification for basket [%d] which is not associated '
-                        'with any existing order (Missing Order Issue)',
-                        basket.id,
-                    )
+                logger.info(
+                    'Received CyberSource payment notification for basket [%d] which is associated '
+                    'with existing order [%s] or had an existing valid payment notification. '
```
jmyatt: please remove the text 'or had an existing valid payment notification' - since with this PR we will be (correctly IMO) raising DuplicateReferenceNumber *only* when an order matching this basket id exists

jmyatt (changes requested):

jmyatt (approved):

PR #1865, opened 2018-07-19
log duplicate order attempts accurately

jmyatt:
There are some problems with the way we log order placement failures after payment has been successfully processed:

1. in some cases an order for the basket id / order number in question already exists
2. in other cases an inactive voucher was somehow in place on the basket up through this post-payment pre-order point
3. there may be other cases where some other problem caused this condition

Other work is in progress to handle item 2 (PR #1862).  This change addresses item 1 by logging duplicate order cases separately from any other (item 3) cases that may occur.  This will enable us to make the alerting for the duplicate order case more precise and establish a straightforward runbook for Support to manually address those cases.

https://openedx.atlassian.net/browse/LEARNER-5466

### DIFF
--- a/ecommerce/extensions/checkout/mixins.py
+++ b/ecommerce/extensions/checkout/mixins.py
@@ -38,8 +38,10 @@ class EdxOrderPlacementMixin(OrderPlacementMixin):
     # Instance of a payment processor with which to handle payment. Subclasses should set this value.
     payment_processor = None
 
-    order_placement_failure_msg = 'Order Failure: Payment was received, but an order for basket [%d] ' \
-                                  'could not be placed because of exception [%s].'
+    duplicate_order_attempt_msg = 'Duplicate Order Attempt: %s payment was received, but an order with number [%s] ' \
+                                  'already exists. Basket id: [%d].'
+    order_placement_failure_msg = 'Order Failure: %s payment was received, but an order for basket [%d] ' \
+                                  'could not be placed.'
 
     __metaclass__ = abc.ABCMeta
 
@@ -256,3 +258,14 @@ class EdxOrderPlacementMixin(OrderPlacementMixin):
             Invoice.objects.create(
                 order=order, business_client=client, type=Invoice.BULK_PURCHASE, state=Invoice.PAID
             )
+
+    def log_order_placement_exception(self, order_number, basket_id):
+        payment_processor = self.payment_processor.NAME.title() if self.payment_processor else None
+        if Order.objects.filter(number=order_number).exists():
+            # One cause of this is Cybersource sending us duplicate notifications for a single payment.
+            # See Jira ticket EG-15
+            logger.exception(
+                self.duplicate_order_attempt_msg, payment_processor, order_number, basket_id
+            )
+        else:
+            logger.exception(self.order_placement_failure_msg, payment_processor, basket_id)
--- a/ecommerce/extensions/payment/tests/mixins.py
+++ b/ecommerce/extensions/payment/tests/mixins.py
@@ -39,9 +39,18 @@ post_checkout = get_class('checkout.signals', 'post_checkout')
 
 
 class PaymentEventsMixin(object):
+
+    DUPLICATE_ORDER_LOGGER_NAME = 'ecommerce.extensions.checkout.mixins'
+
     def get_order(self, basket):
         """ Return the order associated with a basket. """
         return Order.objects.get(basket=basket)
+
+    def get_duplicate_order_error_message(self, order, payment_processor):
+        """ Return the expected error message for a duplicate order attempt. """
+        return 'Duplicate Order Attempt: {payment_processor} payment was received, but an order with number ' \
+               '[{order_number}] already exists. Basket id: [{basket_id}].' \
+            .format(payment_processor=payment_processor, order_number=order.number, basket_id=order.basket.id)
 
     def assert_processor_response_recorded(self, processor_name, transaction_id, response, basket=None):
         """ Ensures a PaymentProcessorResponse exists for the corresponding processor and response. """
@@ -508,9 +517,18 @@ class CybersourceNotificationTestsMixin(CybersourceMixin):
             'handle_order_placement',
             side_effect=exception
         ) as fake_handle_order_placement:
-            error_message = 'Order Failure: Payment was received, but an order for basket [{basket_id}]' \
-                            ' could not be placed because of exception [].'.format(basket_id=self.basket.id)
-            self._assert_processing_failure(notification, error_message)
+            error_message = \
+                'Order Failure: {payment_processor} payment was received, but an order for basket [{basket_id}] ' \
+                'could not be placed.'.format(payment_processor='Cybersource', basket_id=self.basket.id)
+
+            logger_name = 'ecommerce.extensions.checkout.mixins'
+            with LogCapture(logger_name) as lc:
+                self.client.post(self.path, notification)
+
+                lc.check(
+                    (logger_name, 'ERROR', error_message)
+                )
+
             self.assertTrue(fake_handle_order_placement.called)
 
     def test_invalid_basket(self):
--- a/ecommerce/extensions/payment/tests/views/test_cybersource.py
+++ b/ecommerce/extensions/payment/tests/views/test_cybersource.py
@@ -8,12 +8,14 @@ import ddt
 import mock
 import responses
 from django.conf import settings
+from django.test.client import RequestFactory
 from django.urls import reverse
 from django.utils.timezone import now
 from freezegun import freeze_time
 from oscar.apps.payment.exceptions import TransactionDeclined
 from oscar.core.loading import get_class, get_model
 from oscar.test import factories
+from testfixtures import LogCapture
 from waffle.testutils import override_switch
 
 from ecommerce.core.constants import ENROLLMENT_CODE_PRODUCT_CLASS_NAME, ENROLLMENT_CODE_SWITCH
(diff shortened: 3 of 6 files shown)


ecommerce/extensions/checkout/mixins.py
```diff
@@ -38,7 +38,9 @@ class EdxOrderPlacementMixin(OrderPlacementMixin):
     # Instance of a payment processor with which to handle payment. Subclasses should set this value.
     payment_processor = None
 
-    order_placement_failure_msg = 'Order Failure: Payment was received, but an order for basket [%d] ' \
+    duplicate_order_number_msg = 'Duplicate Order Attempt: %s payment was received, but an order with number [%s] ' \
+                                 'already exists. Basket id: [%d], exception: [%s].'
+    order_placement_failure_msg = 'Order Failure: %s payment was received, but an order for basket [%d] ' \
                                   'could not be placed because of exception [%s].'
```
Ayub-Khan: We can remove the exception part from both logs. As both are logged using logger.exception it includes traceback too in logs.

jmyatt: thanks, all set now

ecommerce/extensions/payment/views/paypal.py
```diff
@@ -128,10 +134,12 @@ def get(self, request):
             )
             self.handle_post_order(order)
 
-            return redirect(receipt_url)
-        except Exception as e:  # pylint: disable=broad-except
-            logger.exception(self.order_placement_failure_msg, basket.id, e)
-            return redirect(receipt_url)
+        except Exception:  # pylint: disable=broad-except
+            if Order.objects.filter(number=order_number).exists():
+                logger.exception(
```
robrap: 1. Do we understand why and when this happens, and whether or not it is a problem?  Did we mistakenly place the order at some point?  Did we get a second payment call?  Other?  
2. Based on answer above, do we have a different problem that someone should be looking into?  Or, do we think this is not a problem, and should  just be a warning?  Note: if we need the exception info for some reason, you could add `exc_info=True` as a parameter.
NOTE: Based on the above, we might want to add a relevant comment to the code, like:
```
# We do not understand why this happens and further investigation is needed.  See EG-XXX.
```
Or whatever is appropriate. :)

ecommerce/extensions/payment/views/paypal.py
```diff
@@ -128,10 +134,12 @@ def get(self, request):
...
+                    self.duplicate_order_number_msg, self.payment_processor.NAME.title(), order_number, basket.id)
```
robrap: I think `self.duplicate_order_attempt_msg`, which reuses "attempt" from the actual message, is more accurate.

ecommerce/extensions/payment/views/paypal.py
```diff
@@ -128,10 +134,12 @@ def get(self, request):
             )
             self.handle_post_order(order)
 
-            return redirect(receipt_url)
-        except Exception as e:  # pylint: disable=broad-except
-            logger.exception(self.order_placement_failure_msg, basket.id, e)
-            return redirect(receipt_url)
+        except Exception:  # pylint: disable=broad-except
+            if Order.objects.filter(number=order_number).exists():
```
robrap: Did you consider a shared `log_order_placement_exception` method?

ecommerce/extensions/payment/tests/mixins.py
```diff
@@ -508,8 +508,10 @@ def test_unable_to_place_order(self, exception):
             'handle_order_placement',
             side_effect=exception
         ) as fake_handle_order_placement:
-            error_message = 'Order Failure: Payment was received, but an order for basket [{basket_id}]' \
-                            ' could not be placed because of exception [].'.format(basket_id=self.basket.id)
+            error_message = \
+                'Order Failure: {payment_processor} payment was received, but an order for basket [{basket_id}] ' \
+                'could not be placed.'.format(
+                    payment_processor='Cybersource', basket_id=self.basket.id)
```
robrap: Please move closing paren to the next line.  It should line up with the start of the line that contains the open paren.

robrap (commented):
Thanks.  Next steps will be based on your responses.

ecommerce/extensions/payment/views/paypal.py
jmyatt: ah yes much DRYer - I'll do that

ecommerce/extensions/payment/views/paypal.py
jmyatt: sounds good

ecommerce/extensions/payment/views/paypal.py
jmyatt: 1. It's still unclear whether the blame for the duplicate order case is on us for "mistakenly placing the order at some point" or on the payment gateway for accepting payment twice when our end user re-submits quickly.  So more investigation is necessary.
2. This happens infrequently (3 times in the past 30 days) but when it does we need to alert on it and the user needs to be refunded.  I'll add the further investigation ticket and add a comment here about that.  About exc_info=True, that's redundant when using logger.exception (stack trace is sent automatically).

**UPDATE:** it looks like this is being caused by Cybersource sending us duplicate notifications for a single transaction.  For each of the 3 cases in the past 30 days (orders EDX-33667445, EDX-33777616, and EDX-33709348) we logged an extra call to CybersourceInterstitialView - examination of those 3 in the Cybersource admin UI shows that we only charged the user once.  So the Support runbook for this alert is going to describe (as a during-work-hours task) verifying (in Cybersource) that the user was charged once (or if not, refunding the dup payment), and marking the instance as 'handled'/'verified' somehow.  Separately, we will pursue (with Cybersource) the question of how these dup notifications are happening.

robrap: Thanks. It sounds like they might need to be refunded, but only if we charged them twice. Is there a way for us to know this up front? Is there a way to autorefund when this happens and two payments were found for the order?

Ayub-Khan: @jmyatt  I would like to add some thing here. Multiple payment issue does not happen if user re submits quickly. We have a cybersource setting enabled which do not accept payments and sends a notification with error code '104'. 
https://support.cybersource.com/s/article/What-does-this-response-code-mean
Which mean in every multiple payment issue second request was made after 15 min of first payment. 
@robrap A simple suggestion could be that we create a zendesk ticket through splunk and support refunds the payment from cybersource whenever this issue occurs. (As we do for fulfillment issues).

ecommerce/extensions/payment/views/paypal.py
Ayub-Khan: +1

ecommerce/extensions/payment/views/paypal.py
jmyatt: I agree that an actionable ticket to handle this is what we want, esp since the 3 I examined closely were not in fact dup payments, only dup notifcations.  So we don't want to try to auto-refund, and this is infrequent and delicate enough that it justifies the manual steps.

ecommerce/extensions/payment/tests/views/test_paypal.py
```diff
@@ -259,20 +251,47 @@ def test_unable_to_place_order(self):
         """
         with mock.patch.object(PaypalPaymentExecutionView, 'handle_order_placement',
                                side_effect=UnableToPlaceOrder) as fake_handle_order_placement:
-            error_message = 'Order Failure: Payment was received, but an order for basket [{basket_id}] ' \
-                            'could not be placed because of exception [].'.format(basket_id=self.basket.id)
+            error_message = \
```
robrap: Couldn't the message be moved to `_assert_order_placement_failure` so it appears in only one place?

ecommerce/extensions/payment/tests/views/test_paypal.py
```diff
@@ -259,20 +251,47 @@ def test_unable_to_place_order(self):
...
                                side_effect=KeyError) as fake_handle_order_placement:
-            error_message = 'Order Failure: Payment was received, but an order for basket [{basket_id}] ' \
-                            'could not be placed because of exception [].'.format(basket_id=self.basket.id)
+            error_message = \
+                'Order Failure: {payment_processor} payment was received, but an order for basket [{basket_id}] ' \
+                'could not be placed.'.format(payment_processor='Paypal', basket_id=self.basket.id)
             self._assert_order_placement_failure(error_message)
             self.assertTrue(fake_handle_order_placement.called)
 
+    def test_duplicate_order_attempt_logging(self):
```
robrap: Couldn't most of this been moved to the mixin as well?

robrap (changes requested):
Hopefully minor refactors.  Let me know.  Thanks.

ecommerce/extensions/payment/tests/views/test_cybersource.py
```diff
@@ -366,6 +368,27 @@ def test_order_creation_error(self):
...
+        """
+        Verify that attempts at creation of a duplicate order are logged correctly
+        """
+        prior_order = create_order()
+        dummy_request = RequestFactory(SERVER_NAME='testserver.fake').get('')
+        dummy_mixin = OrderCreationMixin()
+        dummy_mixin.payment_processor = Cybersource(self.site)
+
+        with LogCapture(self.DUPLICATE_ORDER_LOGGER_NAME) as lc:
+            try:
```
robrap: If an exception isn't raised, this test would also pass, and I am guessing it shouldn't. This should probably be:
```
with LogCapture(self.DUPLICATE_ORDER_LOGGER_NAME) as lc:
    with self.assertRaises(ValueError):
        dummy_mixin.create_order(dummy_request, prior_order.basket, None)
    lc.check( ... )
```

robrap (approved):
I was originally thinking the whole LogCapture context code could be a helper method, but looking more closely, I see they are just different enough.  Not sure why one processing method raises an exception and the other does not, but that is a separate question.

Looking more closely, I made one minor comment.  Feel free to fix locally when you squash, or you could leave it.

ecommerce/extensions/payment/tests/views/test_cybersource.py
jmyatt: good point - I'll fix that

PR #1866, opened 2018-07-20
Improve Performance of Journal Bundle Offer Page

bessiesteinberg:
- The Problem: The Journal Bundle Offer page had a very high response
  time and was often timing out.  This was because it was looping over
  every row in the conditional_offer table.  This was unecessary because
  the Journal Bundle Offer page does not need knowledge of Program
  Offers or Enterprise Offers.  It was also a huge performance hit
  because there are a lot of conditional offers.
- The Solution: Implemented 'get_queryset' for
  'JournalBundleOfferViewMixin' so it only has to loop through the
  relevant offers.  Previously, we were looping through every offer and
  checking that it was a Journal Bundle Offer, I removed that check as
  it is no longer needed.

### DIFF
--- a/ecommerce/journals/tests/test_views.py
+++ b/ecommerce/journals/tests/test_views.py
@@ -29,7 +29,7 @@ class JournalBundleOfferListViewTests(ViewTestMixin, TestCase):
         """ Test the "GET" endpoint with offers and assert the rendered template """
         mock_journal_endpoint.return_value = _get_mocked_endpoint()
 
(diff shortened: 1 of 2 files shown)


robrap (approved):
Nice. I assume there are already tests.  Once this green of course.

Should this be done for other offer types (like enterprise?), or did those already have this code?

bill-filler (approved):
approved, good to merge after tests pass

bessiesteinberg:
@robrap This is already being done for enterprise and program offers

PR #1868, opened 2018-07-23
ARCH-177: Add RequestCache namespacing.

robrap:
The following changes were added in preparation of pulling cache_utils
out to another repo:
1. Added namespacing to the RequestCache.
2. Query param 'force_django_cache_miss' now requires staff permissions.

[ARCH-177](https://openedx.atlassian.net/browse/ARCH-177)

### DIFF
--- a/ecommerce/cache_utils/README.rst
+++ b/ecommerce/cache_utils/README.rst
@@ -6,13 +6,16 @@ RequestCache
 
 A thread-local for storing request scoped cache values.
 
+An optional namespace can be used with the RequestCache, or you can use
+the `DEFAULT_REQUEST_CACHE`.
+
 
 TieredCache
 -----------
 
-The first tier is a request cache that is tied to the life of a
-given request. The second tier is the Django cache -- e.g. the
-"default" entry in settings.CACHES, typically backed by memcached.
+The first tier is the default request cache that is tied to the life of a
+given request. The second tier is the Django cache -- e.g. the "default"
+entry in settings.CACHES, typically backed by memcached.
 
 Some baseline rules:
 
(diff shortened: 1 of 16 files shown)


ecommerce/cache_utils/README.rst
```diff
@@ -42,12 +44,17 @@ Sample Usage using is_miss::
     return x_cached_response.value
 
 
+You must include 'ecommerce.cache_utils.middleware.CacheUtilsMiddleware'
```
robrap, line 48: Reminder: this will need to change once this moves to the new repo.

ecommerce/cache_utils/utils.py
```diff
@@ -12,22 +12,69 @@
...
+    """
+
+    def __init__(self, namespace=None):
+        """
+        Creates a request cache with the provided namespace.
+
+        Args:
+            namespace (string): (optional) uses 'default' if not provided.
+        """
+        self.namespace = namespace or 'default'
```
nasthagiri: To make sure a client's own namespace doesn't conflict with the default one, add an assertion that `namespace` doesn't equal "default"?

Or rename "default" to prevent its usage.

ecommerce/cache_utils/utils.py
```diff
@@ -52,19 +98,25 @@ def set(cls, key, value):
...
         Args:
             key (string)
 
         """
-        if key in cls._data:
-            del cls._data[key]
+        if key in self._data():
+            del self._data()[key]
+
+    def _data(self):
```
nasthagiri: Nit: can we rename this to something like `_get_data`?  I initially got confused thinking this was related to the `_data` member declared in the top of the file.

ecommerce/cache_utils/utils.py
```diff
@@ -165,24 +219,34 @@ def _set_request_cache_if_django_cache_hit(cls, key, django_cached_response):
...
+
+        This functionality is only available for staff.
 
         Example:
             http://clobert.com/api/v1/resource?force_cache_miss=true
 
         """
-        force_cache_miss = request.GET.get(FORCE_CACHE_MISS_PARAM, 'false').lower() == 'true'
-        RequestCache.set(SHOULD_FORCE_CACHE_MISS_KEY, force_cache_miss)
+        if not (request.user.is_active and request.user.is_staff):
```
nasthagiri: Do you also need to add `request.user and ..`?  Jus to prevent a 500 in case it's called by an Anonymous user?

nasthagiri (approved):

PR #1871, opened 2018-07-30
Reverted hot fix for LEARNER-5603

Ayub-Khan:
Reverted hot fix for LEARNER-5603
As we fixed the root cause in https://github.com/edx/ecommerce/pull/1862
We do not need this code any more. 

This is a revert of https://github.com/edx/ecommerce/pull/1843

### DIFF
--- a/ecommerce/extensions/payment/constants.py
+++ b/ecommerce/extensions/payment/constants.py
@@ -55,5 +55,3 @@ APPLE_PAY_CYBERSOURCE_CARD_TYPE_MAP = {
 STRIPE_CARD_TYPE_MAP = {
     value['stripe_brand']: key for key, value in six.iteritems(CARD_TYPES) if 'stripe_brand' in value
 }
-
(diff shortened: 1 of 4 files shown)


tasawernawaz (approved):

ahsan-ul-haq (approved):

robrap (approved):
Seems ok.  Please test on Stage that you can't recreate the issue once this is reverted.  Also, remember to remove the waffle flag from all environments once this is out there.
@waheedahmed can merge for you when it is ready.
Thanks.

Ayub-Khan:
@robrap this can not be tested on stage as the root cause was fixed so voucher will be removed from the basket view if its already expired.

robrap:
I think that is what I meant. Ensuring that the root cause fix still works as it did, to ensure that the original hot fix is no longer relevant. But, maybe you are saying something else.

Ayub-Khan:
got it.

waheedahmed:
@Ayub-Khan I'm assuming all comments are addressed here.

PR #1874, opened 2018-08-07
Proposal: Skip displaying prices in the receipt page for free orders

lgp171188:
This is a follow-up to the https://github.com/edx/ecommerce/pull/1676 PR which was merged and then reverted a while later. The goal of that PR was to skip displaying the receipt page when it is a free order. The implementation that was merged initially, skipped the receipt page by default. It was reverted later because the receipt page added sufficient delay for the enrolled course to show up in the dashboard and without that, the users were redirected to the dashboard which didn't show the enrolled course till they refresh the page. Also, there were some e2e tests which were failing and had to be rewritten.

**Proposed approach**
Instead of skipping the receipt page for free orders, the receipt page will be shown without the price. This could behind a flag, or could be the new default behaviour.

In case there is a need to display the price for free orders, that could be made configurable at a per-course level or at a per-coupon level if it makes sense.

CC @mduboseedx @kaizoku

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,4 @@ docs/_build
 # Certificates
 *.cer
 *.pem
+


openedx-webhooks:
Thanks for the pull request, @lgp171188! I've created [OSPR-2547](https://openedx.atlassian.net/browse/OSPR-2547) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

mduboseedx:
@edx/learner-product @marcotuts Could you look over this proposal about a change to the receipt page?

lgp171188:
@edx/learner-product @marcotuts did you get a chance to review this and are there any updates? @mduboseedx

mduboseedx:
@lgp171188 I'm sorry for the delay, as I was actually out of the office this past week. I'm checking with our Product team about your proposal.

lgp171188:
@mduboseedx, did you get a chance to check with the product team about this?

mduboseedx:
@lgp171188 Sorry for the delay. Our Product folks are looking into your proposal and will then give a response on the PR. Thanks!

marcotuts:
I have a question that goes back to the earlier proposal, and then a comment on the current proposal.

> It was reverted later because the receipt page added sufficient delay for the enrolled course to show up in the dashboard and without that, the users were redirected to the dashboard which didn't show the enrolled course till they refresh the page.

Is this something we could solve for knowing that you are coming from a receipt page for a course / set of courses? That is, what would a "processing enrollment" area / message at the top of the dashboard that would eventually load in the course look like? Would the dashboard need to be rewritten in more modern react patterns for us to have this kind of dynamic page option / path? It is ok in my mind if this is a large technical scope and not a reasonable next step but I'd like to understand the implications of this. Not opting for what I believe is the correct path (taking learners to their dashboard or even directly into the course if it is open), seems like something we should lean into instead of seeking alternate proposals.

As for the current proposal, I think there are reasons why altering the receipt display might impact other more common use cases (enrollment / coupon codes, enterprise, etc). Ultimately if a learner has to see the receipt page, whether they see a $0 or no dollar value doesn't seem to make much of an improvement to the learner workflow. For that reason I would suggest we close this second proposal, discuss what would need to change to make the original proposal work, even if that isn't technically feasible in the short term. 

FYI - @robrap

lgp171188:
@marcotuts, thank you for the review!

The code changes which were initially merged in the previous proposal skipped the display of the receipt page in case of free orders. But the learner could still go to the receipt page in the order history where the price information is displayed.

As mentioned in a [comment](https://github.com/edx/ecommerce/pull/1676#issuecomment-405367232) in the original PR by its author, skipping the receipt page doesn't fully satisfy the requirement.

>I checked up on this a bit, and the main requirement of our client is that for users enrolling via an enrollment code should not be confronted with a price point + 100% discount. So simply hiding those details in case of a free enrollment would be sufficient.

So the information displayed in the receipt still needs to be changed if the pricing information has to be hidden. It could be done by overriding the template in an ecommerce theme and that wouldn't require any code changes to the ecommerce repository. In such a scenario, skipping the receipt page is not really needed.

We are currently discussing this internally and will let you know soon about whether we want to take the processing area/dashboard changes forward.

marcotuts:
Sounds great thanks for the heads up. The kind of change we are discussing to the receipt page is one that I would encourage you to use your own template overrides as opposed to adding a configuration flag / option here. The dashboard changes are likely much larger, and would be easier to build once we move the dashboard to being a react application, I would expect. I don't anticipate we would want to merge dashboard changes here unless it helps us move down the split front / back end architectural pattern for the dashboard. If your team is able to make progress on that, then great! (I anticipate though that the path of a client specific template override is the best step here.) 

For now I'm going to mark this closed and we can evaluate another proposal depending on what your team decides. Thanks again for creating this in advance of code changes!

openedx-webhooks:
@lgp171188 Even though your pull request wasn’t merged, please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/ecommerce&entry.867055334=https://github.com/edx/ecommerce/pull/1874&entry.1484655318=https://github.com/lgp171188&entry.752974735=2018-08-07+14:59&entry.1917517419=2018-10-02+14:06&entry.2133058324=No) so we can improve your experience in the future.

lgp171188:
Thanks for your review @marcotuts. We will come up with another proposal depending on what we decide to do for this. :+1: CC @kaizoku

PR #1875, opened 2018-08-09
LEARNER-5754 disabled django debugtoolbar template panel

Ayub-Khan:
LEARNER-5754
Django debug toolbar causes a lot of delays when number of templates to render is a lot. This will just uncheck the template option from the debug toolbar panel by default to avoid the delays. If some one is debugging the templates can check the templates option and use it. 

@edx/learner-spartans  kindly review.


robrap (approved):
I thought we removed the toolbar temporarily? Does the PR also need to bring it back by default?

waheedahmed (approved):

Ayub-Khan:
following will bring back the django debug toolbar https://github.com/edx/devstack/pull/330.

PR #1881, opened 2018-08-17
Added verify transactions management command to ensure orders and pay…

calistenson:
…ments match.

https://openedx.atlassian.net/browse/LEARNER-5763
https://openedx.atlassian.net/browse/LEARNER-6012

### DIFF
--- a/ecommerce/core/management/commands/tests/factories.py
+++ b/ecommerce/core/management/commands/tests/factories.py
@@ -1,0 +1,11 @@
+import factory
+from factory.fuzzy import FuzzyInteger
+from oscar.core.loading import get_model
+
+
+class PaymentEventFactory(factory.DjangoModelFactory):
+
+    id = FuzzyInteger(1000, 9999)
+
+    class Meta(object):
+        model = get_model('order', 'PaymentEvent')
--- a/ecommerce/core/management/commands/tests/test_verify_transactions.py
+++ b/ecommerce/core/management/commands/tests/test_verify_transactions.py
@@ -1,0 +1,137 @@
+"""
+Tests for Django management command to verify ecommerce transactions.
+"""
+import datetime
+
+import pytz
+from django.core.management import call_command
+from django.core.management.base import CommandError
+from django.test import TestCase
+from oscar.core.loading import get_class, get_model
+from oscar.test.factories import OrderFactory
+
+from ecommerce.core.management.commands.tests.factories import PaymentEventFactory
+from ecommerce.core.management.commands.verify_transactions import DEFAULT_END_DELTA_TIME, DEFAULT_START_DELTA_TIME
+
+Order = get_model('order', 'Order')
+PaymentEvent = get_model('order', 'PaymentEvent')
+PaymentEventType = get_model('order', 'PaymentEventType')
+PaymentEventTypeName = get_class('order.constants', 'PaymentEventTypeName')
+
+
+class VerifyTransactionsTest(TestCase):
+
+    def setUp(self):
+        # Timestamp in the middle of the time window
+        time_delta = (DEFAULT_START_DELTA_TIME + DEFAULT_END_DELTA_TIME) / 2
+        self.timestamp = datetime.datetime.now(pytz.utc) - datetime.timedelta(minutes=time_delta)
+
+        self.payevent = PaymentEventType.objects.get_or_create(name=PaymentEventTypeName.PAID)
+        self.refundevent = PaymentEventType.objects.get_or_create(name=PaymentEventTypeName.REFUNDED)
+        self.order = OrderFactory(total_incl_tax=90, date_placed=self.timestamp)
+        self.order.save()
+
+    def test_time_window(self):
+        """Test verify_transactions only examines the correct time window"""
+        with self.assertRaises(CommandError) as cm:
+            call_command('verify_transactions')
+        exception = cm.exception
+        self.assertIn("The following orders are without payments", exception.message)
+        self.assertIn(str(self.order.id), exception.message)
+
+        time_outside_window = DEFAULT_START_DELTA_TIME + DEFAULT_END_DELTA_TIME + 1
+        time_outside_window_datetime = datetime.datetime.now(pytz.utc) - datetime.timedelta(minutes=time_outside_window)
+        self.order.date_placed = time_outside_window_datetime
+        self.order.save()
+
+        try:
+            call_command('verify_transactions')
+        except CommandError:
+            self.fail("Failed to verify transactions when no errors were expected.")
+
+    def test_no_errors(self):
+        """Test verify_transactions with order and payment of same amount."""
+        payment = PaymentEventFactory(order=self.order,
+                                      amount=90,
+                                      event_type_id=self.payevent[0].id,
+                                      date_created=self.timestamp)
+        payment.save()
+        refund = PaymentEventFactory(order=self.order,
+                                     amount=90,
+                                     event_type_id=self.refundevent[0].id,
+                                     date_created=self.timestamp)
+        refund.save()
+        try:
+            call_command('verify_transactions')
+        except CommandError:
+            self.fail("Failed to verify transactions when no errors were expected.")
+
+    def test_no_payment_for_order(self):
+        """Verify errors are thrown when there are orders without payments."""
+        with self.assertRaises(CommandError) as cm:
+            call_command('verify_transactions')
+        exception = cm.exception
+        self.assertIn("The following orders are without payments", exception.message)
+        self.assertIn(str(self.order.id), exception.message)
+
+    def test_two_same_payments_for_order(self):
+        """ Verify that errors are thrown when their are multiple payments on an order."""
+        payment1 = PaymentEventFactory(order=self.order,
+                                       amount=90,
+                                       event_type_id=self.payevent[0].id,
+                                       date_created=self.timestamp)
+        payment2 = PaymentEventFactory(order=self.order,
+                                       amount=90,
+                                       event_type_id=self.payevent[0].id,
+                                       date_created=self.timestamp)
+        payment1.save()
+        payment2.save()
+        with self.assertRaises(CommandError) as cm:
+            call_command('verify_transactions')
+        exception = cm.exception
+        self.assertIn("The following orders had multiple payments ", exception.message)
+        self.assertIn(str(self.order.id), exception.message)
+        self.assertIn(str(payment1.id), exception.message)
+        self.assertIn(str(payment2.id), exception.message)
+
+    def test_multiple_payments_for_order(self):
+        """ Verify that errors are thrown when their are multiple payments on an order."""
+        payment1 = PaymentEventFactory(order=self.order,
+                                       amount=90,
+                                       event_type_id=self.payevent[0].id,
+                                       date_created=self.timestamp)
+        payment2 = PaymentEventFactory(order=self.order,
+                                       amount=90,
+                                       event_type_id=self.payevent[0].id,
+                                       date_created=self.timestamp)
+        payment3 = PaymentEventFactory(order=self.order,
+                                       amount=90,
+                                       event_type_id=self.payevent[0].id,
+                                       date_created=self.timestamp)
+        payment1.save()
+        payment2.save()
+        payment3.save()
+        with self.assertRaises(CommandError) as cm:
+            call_command('verify_transactions')
+        exception = cm.exception
+        self.assertIn("The following orders had multiple payments ", exception.message)
+        self.assertIn(str(self.order.id), exception.message)
+        self.assertIn(str(payment1.id), exception.message)
+        self.assertIn(str(payment2.id), exception.message)
+        self.assertIn(str(payment3.id), exception.message)
+
+    def test_totals_mismatch(self):
+        """ Verify errors thrown when payment and order totals don't match."""
+        payment = PaymentEventFactory(order=self.order,
+                                      amount=100,
+                                      event_type_id=self.payevent[0].id,
+                                      date_created=self.timestamp)
+        payment.save()
+        with self.assertRaises(CommandError) as cm:
+            call_command('verify_transactions')
+        exception = cm.exception
+        self.assertIn("Order totals mismatch with payments received", exception.message)
+        self.assertIn(str(self.order.id), exception.message)
+        self.assertIn(str(payment.id), exception.message)
+        self.assertIn("Amount: 90.00", exception.message)
+        self.assertIn("Amount: 100.00", exception.message)
--- a/ecommerce/core/management/commands/verify_transactions.py
+++ b/ecommerce/core/management/commands/verify_transactions.py
@@ -1,0 +1,113 @@
+"""
+Django management command to verify ecommerce transactions.
+
+Command is run by Jenkins job each hour. By default the job runs on a three
+hour time window starting one hour in the past.
+
+For each order in the time window the command verifies exactly one payment of
+the expected value exists in the database.
+
+If a PaymentEvent does not exist, multiple PaymentEvents exist, or the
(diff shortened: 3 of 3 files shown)


ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,52 @@
...
+Order = get_model('order', 'Order')
+PaymentEvent = get_model('order', 'PaymentEvent')
+
+
+class Command(BaseCommand):
+
+    help = 'Management command to verify ecommerce transactions and trigger alerts if there is any imbalance.'
+
+    def handle(self, *args, **options):
+        exit_errors = {}
```
BenjiLee: Add a docstring.

- An example or structure of the output.
- What exceptions are handled

All of this might be better off in the main docstring for the file, but it should be noted here or at the top of the file.

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,52 @@
+"""
+Django management command to verify ecommerce transactions.
```
BenjiLee, line 2: Add more details about the management command as a whole.

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,52 @@
...
+from oscar.core.loading import get_model
+
+logger = logging.getLogger(__name__)
+Order = get_model('order', 'Order')
+PaymentEvent = get_model('order', 'PaymentEvent')
+
+
+class Command(BaseCommand):
+
+    help = 'Management command to verify ecommerce transactions and trigger alerts if there is any imbalance.'
```
BenjiLee: Since we are not explicitly triggering alerts, we should clarify that we are just outputting logs.

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,52 @@
...
+        order_sum = 0
+        payment_sum = 0
+
+        for order in orders:
+            order_sum += order.total_incl_tax
+            pays = PaymentEvent.objects.filter(order_id=order.id)
+            if pays.count() == 0 and order.total_incl_tax > 0:
+                orders_without_payments.append(order.id)
+            elif pays.count() > 1:
+                exit_errors["multi_payment"] = "Multiple payments on order."
```
BenjiLee: Add more details to this log message such as which orders and what payments.

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,52 @@
...
+        payments = PaymentEvent.objects.all().filter(date_created__gte=start, date_created__lt=end)
+
+        orders_without_payments = []
+
+        # Sum the total order costs and payment costs then verify they're equal.
+        order_sum = 0
+        payment_sum = 0
+
+        for order in orders:
+            order_sum += order.total_incl_tax
```
BenjiLee: I think every order after the first erroneous order will fail because we are not resetting the sums.

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,52 @@
...
+            pays = PaymentEvent.objects.filter(order_id=order.id)
```
BenjiLee: We should use a better variable name here such as `payment_events` or maybe `payment_events_for_order`

BenjiLee: This filters against the full database. I think you were trying to further filter against your queryset that you filtered in the previous lines. (`payments`)

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,52 @@
...
+            if pays.count() == 0 and order.total_incl_tax > 0:
```
BenjiLee: if an order.total_incl_tax == 0, do we want to verify that there is no payment event?

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,52 @@
...
+                orders_without_payments.append(order.id)
+            elif pays.count() > 1:
+                exit_errors["multi_payment"] = "Multiple payments on order."
+
+        for payment in payments:
+            payment_sum += payment.amount
+
+        if order_sum != payment_sum:
+            exit_errors["totals_mismatch"] = "Order totals mismatch with payments received." \
```
BenjiLee: When there is a mismatch, how can we identify which orders/paymentevents are wrong if  all orders have payments.

BenjiLee (changes requested):

ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -0,0 +1,92 @@
...
+PaymentEventTypeName = get_class('order.constants', 'PaymentEventTypeName')
+
+
+class VerifyTransactionsTest(TestCase):
+
+    def setUp(self):
+        self.payevent = PaymentEventType.objects.get_or_create(name=PaymentEventTypeName.PAID)
+        self.timestamp = datetime.datetime.now(pytz.utc) - datetime.timedelta(hours=1, minutes=30)
+
+    def tearDown(self):
```
BenjiLee: We should be able to get rid of this. All changes are rolled back after the test.

https://docs.djangoproject.com/en/1.9/topics/testing/tools/#transactiontestcase

ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -0,0 +1,92 @@
...
+        Order.objects.all().delete()
+
+    def test_no_errors(self):
+        """Test verify_transactions with order and payment of same amount."""
+        order = OrderFactory()
```
BenjiLee: I believe we can shorten this to...

`OrderFactory(total_incl_tax = 100, date_placed = self.timestamp)`
```
order = OrderFactory()
order.total_incl_tax = 100
order.date_placed = self.timestamp
```

ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -0,0 +1,92 @@
...
+        order.total_incl_tax = 100
+        order.date_placed = self.timestamp
+        order.save()
+        payment = PaymentEventFactory(order=order, amount=100, event_type_id=self.payevent[0].id)
```
BenjiLee: Same refactoring comment as Order() above.

ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -0,0 +1,92 @@
...
+        payment.date_created = self.timestamp
+        payment.save()
+        try:
+            call_command('verify_transactions')
+        except CommandError:
+            self.fail("Failed to verify transactions when no errors were expected.")
+
+    def test_no_payment_for_order(self):
+        """Verify errors are thrown when there are orders without payments."""
+        order = OrderFactory()
```
BenjiLee: Looks like we are using a lot of the same code for the OrderFactory(). We can put this code in setUp().

BenjiLee:
Git suggests using present-tense and imperative language. https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?id=HEAD#n133
`Added verify transactions management command to ensure orders and payments match.` -> `Add verify_transactions management command to ensure orders and payments match.`

Shorten the first line of the commit message. If it need more details, begin the paragraph on the 3rd line.
```
Add verify_transactions management command

The verify_transactions management command ensures that
orders amounts and paymentevent amounts match.

LEARNER-5763
LEARNER-6012
```

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,65 @@
+"""
+Django management command to verify ecommerce transactions.
+
+Command is run by Jenkins job each hour on the previous hour and 10 minutes of orders in ecommerce.
+
+For each order in the time window the command verifies exactly one payment of the expected value exists in the database.
```
BenjiLee: Nit and FYI
https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#id4

Docstrings should be 80 chars in width.

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,65 @@
...
+
+If a PaymentEvent does not exist, multiple payments exist, or the payment amount is different from the order amount
```
BenjiLee:
```
If a PaymentEvent does not exist, multiple PaymentEvents exist, or the
PaymentEvent amount is different from the order amount, then the order 
id and relevant payment information is logged in a list associated with
each of these scenarios.
```

BenjiLee:
@calistenson Could you output an example of this management command where all 3 scenarios fail at the same time?

ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -0,0 +1,75 @@
...
+Order = get_model('order', 'Order')
+PaymentEvent = get_model('order', 'PaymentEvent')
+PaymentEventType = get_model('order', 'PaymentEventType')
+PaymentEventTypeName = get_class('order.constants', 'PaymentEventTypeName')
+
+
+class VerifyTransactionsTest(TestCase):
+
+    def setUp(self):
+        self.payevent = PaymentEventType.objects.get_or_create(name=PaymentEventTypeName.PAID)
```
BenjiLee: nit: a `payevent` is not a `paymenteventtype`.`self.payevent` -> `self.payment_event_type`

ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -0,0 +1,75 @@
...
+        self.timestamp = datetime.datetime.now(pytz.utc) - datetime.timedelta(hours=1, minutes=30)
```
BenjiLee: nit: self.timestamp is not a random timestamp. This might be easier to code when we have parser arguments being passed in for the management command. As it is right now, changing the management command time window could potentially break tests. Having a way to pass in a time window will make our tests more flexible.

ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -0,0 +1,75 @@
...
+        self.order = OrderFactory(total_incl_tax=90, date_placed=self.timestamp)
+        self.order.save()
+
+    def test_no_errors(self):
+        """Test verify_transactions with order and payment of same amount."""
+        payment = PaymentEventFactory(order=self.order, amount=90, event_type_id=self.payevent[0].id)
+        payment.date_created = self.timestamp
```
BenjiLee: Nit: Should be able to shove this into the factory initialization.
```
payment = PaymentEventFactory(
            order=self.order, 
            amount=90, 
            event_type_id=self.payevent[0].id, 
            date_created=self.timestamp
        )
```
this can also be done to the payment events below

BenjiLee:
@calistenson  Is the same process going to be followed for a refunded order?

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,88 @@
+"""
+Django management command to verify ecommerce transactions.
+
+Command is run by Jenkins job each hour on the previous hour and 10 minutes of orders in ecommerce.
```
BenjiLee: This is no longer true. We can make this more resistant to changes.

`Command will be run as a cron job in Jenkins that looks at a time window for Order/Payment discrepancies.`

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,110 @@
...
+
+        orders = Order.objects.all().filter(date_placed__gte=start, date_placed__lt=end)
+
+        orders_without_payments = []
+        multi_payment_on_order = []
+        order_payment_totals_mismatch = []
+
+        for order in orders:
+            payment_events = PaymentEvent.objects.filter(order_id=order.id)
+            if payment_events.count() == 0 and order.total_incl_tax > 0:
```
BenjiLee, line 82: Add a comment to why this conditional is here. 
`# If a coupon is used to purchase a product for the full price, there will be no PaymentEvent`

calistenson: Done

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,113 @@
...
+            if payment_events.count() > 1:
+                if payment_events.count() == 2:
+                    event1 = payment_events[0]
+                    event2 = payment_events[1]
+                    if event1.event_type_id == event2.event_type_id:
+                        # If PaymentEvents are the same type then there were multiple payments.
+                        multi_payment_on_order.append((order.id, [event.id for event in payment_events]))
+                else:
+                    multi_payment_on_order.append((order.id, [event.id for event in payment_events]))
+            for event in payment_events:
```
BenjiLee, line 94: nit: I believe we should be guaranteed only 1 payment event here. We are using this for loop to extract a single item from a list and rename it.

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -0,0 +1,113 @@
...
+                print("event: ", event)
+                if event.amount != order.total_incl_tax:
+                    order_payment_totals_mismatch.append(("Order: " + str(order.id) + " Amount: " +
+                                                          str(order.total_incl_tax), "Payment: " +
+                                                          str(event.id) + " Amount: " +
+                                                          str(event.amount)))
+
+        if len(order_payment_totals_mismatch):
```
BenjiLee: fun python! empty lists are falsy

ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -0,0 +1,137 @@
...
+                                       event_type_id=self.payevent[0].id,
+                                       date_created=self.timestamp)
+        payment1.save()
+        payment2.save()
+        with self.assertRaises(CommandError) as cm:
+            call_command('verify_transactions')
+        exception = cm.exception
+        self.assertIn("The following orders had multiple payments ", exception.message)
+        self.assertIn(str(self.order.id), exception.message)
+        self.assertIn(str(payment1.id), exception.message)
```
BenjiLee, line 94: There's a potential for order.id == payment.id since they are not unique with each other. A way to resolve this is to customize the id generation field in the PaymentFactory. Example: https://github.com/edx/ecommerce/blob/af0190314929a482421f52162ccad3eaf296e982/ecommerce/courses/tests/factories.py#L14

Since we are running tests in a fresh test database, the likelyhood of order.id == payment.id is high.

ecommerce/core/management/commands/verify_transactions.py
calistenson: We still want to catch mismatched totals in the cases where there are multiple payments. So I will leave the loop here.

ecommerce/core/management/commands/tests/test_verify_transactions.py
calistenson: Fixed this issue in the test by setting the order id manually to "1000". Should I add a comment in the test about why I did this?

ecommerce/core/management/commands/tests/factories.py
```diff
@@ -0,0 +1,8 @@
+import factory
+from oscar.core.loading import get_model
+
+
+class PaymentEventFactory(factory.DjangoModelFactory):
+
```
BenjiLee, line 7: `id = random.randint(1000,9999)`

This it basically an int version of FuzzyText for our purposes

BenjiLee (approved):

BenjiLee:
shipit

PR #1883, opened 2018-08-23
Add read replica util.

calistenson:
Add read replica util and update queries in verify_transactions to first try to point at read replica.

https://openedx.atlassian.net/browse/LEARNER-6235


BenjiLee (approved):
Approval pending on DEVOPs input

PR #1887, opened 2018-08-28
Fix program and enterprise offers page for multiple sites.

waheedahmed:
Offers pages are not showing data for multiple sites associated
with same partner

LEARNER-6249


waheedahmed:
@robrap @feanil @staubina Please review.

ecommerce/enterprise/tests/test_views.py
```diff
@@ -52,17 +52,17 @@ def test_get_queryset(self):
         """ Should return only Conditional Offers with Site offer type. """
```
bderusha, line 52: Should this docstring be updated too?

ecommerce/enterprise/tests/test_views.py
```diff
@@ -52,17 +52,17 @@ def test_get_queryset(self):
...
 
         # Conditional Offer should contain a condition with enterprise customer uuid set in order to be returned
-        site_conditional_offer = factories.EnterpriseOfferFactory(site=self.site)
+        site_conditional_offer = factories.EnterpriseOfferFactory(partner=self.partner)
```
bderusha: should this variable be `partner_conditional_offer`?

bderusha (approved):
2 small changes, otherwise this looks good to me.

@georgebabey may want to also take a look since this is enterprise related.

ecommerce/enterprise/tests/test_views.py
waheedahmed: @bderusha Actually, the ConditionalOffer type is still `SITE`, we are just replacing the site object with partner. Please refer to following line https://github.com/edx/ecommerce/blob/29f672bedeaafa329e4ae6495a2705f10175cc68/ecommerce/enterprise/tests/test_views.py#L63

waheedahmed:
@georgebabey Could you please take a look on this PR?

georgebabey:
This looks good to me but is probably more in @bill-filler wheel-house

bill-filler (commented):
@waheedahmed I'm a bit confused what problem is trying to be solved here. What exactly is broken?

Offers are currently linked to a Site and a Site has an associated Partner. This change seems to add a direct relationship from Offer to Partner. What is the benefit of that?

waheedahmed:
@bill-filler This is blocking the work needs to be done under https://openedx.atlassian.net/browse/OPS-2235, maybe @feanil can explain better.

staubina:
@waheedahmed I have talked with @bill-filler and he is going to talk with @robrap and @feanil

waheedahmed:
@bill-filler, @feanil Any update on this?

CC: @robrap, @staubina

robrap:
@bill-filler: The higher level change that was needed to accomplish the OPS work was to replace the 1-to-1 mapping between Partner and Site with a 1-to-many mapping between Partner and SiteConfig (which itself has a 1-to-1 mapping with a Site).

In the past, because of the 1-to-1 mapping between Partner and Site, some things that should be associated with the Partner and some things that should be associated with the Site may have landed anywhere, because you could always get directly from one to another.

I did not look at the code yet, but you had written: "Offers are currently linked to a Site and a Site has an associated Partner."  Now my question would be what should the offers be associated with?  Should the offers be associated with any Site that the Partner is associated with?  Only a select set of sites?  Maybe the public/private Site is a special type of grouping that requires a different kind of designation?

This is not to replace your meeting, but to provide a little additional background for when we meet.

bill-filler (approved):
approved

waheedahmed:
Thank you @bill-filler!

PR #1893, opened 2018-08-31
Add support for filtering products out of the verify-transaction mgmt cmd

bderusha:


ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -53,12 +58,12 @@ def test_no_errors(self):
         """Test verify_transactions with order and payment of same amount."""
         payment = PaymentEventFactory(order=self.order,
                                       amount=90,
-                                      event_type_id=self.payevent[0].id,
```
jmyatt, line 56: good call fixing these!

ecommerce/core/management/commands/tests/test_verify_transactions.py
```diff
@@ -147,25 +167,22 @@ def test_totals_mismatch(self):
         self.assertIn(str(payment.id), exception.message)
         self.assertIn("Amount: 90.00", exception.message)
         self.assertIn("Amount: 100.00", exception.message)
+        self.assertIn("Amount: 100.00", exception.message)
 
     def test_refund_exceeded(self):
-        """ Verify errors thrown when payment and order totals don't match."""
+        """Test verify_transactions with order and payment of same amount."""
```
jmyatt: is this comment correct?  Isn't this still testing the order/payment mismatch?

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -176,8 +179,21 @@ def error_msg(self, errors):
...
+        # We only expect immediate payments for Seats and Entitlements.
+        # Filter out orders that were flagged as being without payment for other product types
+        for order, payments in self.ORDERS_WITHOUT_PAYMENTS:
+            for line in order.lines.all():
+                if not self.verifiable_product(line.product):
+                    self.ORDERS_WITHOUT_PAYMENTS.remove((order, payments))
+                    break
+
+    def verifiable_product(self, product):
+        valid_product_class_names = [SEAT_PRODUCT_CLASS_NAME, COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME]
```
jmyatt: I think this should be a constant (at module scope like the 'DEFAULT_' ones)

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -176,8 +179,21 @@ def error_msg(self, errors):
...
         return msg
+
+    def clean_orders(self):
+        # We only expect immediate payments for Seats and Entitlements.
+        # Filter out orders that were flagged as being without payment for other product types
+        for order, payments in self.ORDERS_WITHOUT_PAYMENTS:
+            for line in order.lines.all():
+                if not self.verifiable_product(line.product):
+                    self.ORDERS_WITHOUT_PAYMENTS.remove((order, payments))
+                    break
```
jmyatt: I see two potential problems with this loop:
* deleting from a list (`self.ORDERS_WITHOUT_PAYMENTS`) that you are iterating over
* this doesn't handle the case where a single order could contain both a verifiable and non-verifiable product, in different order lines

for the deleting-while-iterating issue you could reverse the list using reversed() or use list comprehension with an if expression to call a function that would determine whether to include the particular tuple 

for the multi-product case, we would still want payment for such orders, correct?  If the order contains a $100 Seat and a $50 Coupon, we would still expect to see a payment of $100.  So for the purpose of pruning the `self.ORDERS_WITHOUT_PAYMENTS` list, I think the logic here should be "remove the tuple if there are no order lines containing a verifiable product".  And then you can remove the `break`

jmyatt (changes requested):

bderusha:
@jmyatt ready for another pass

jmyatt (approved):
🚢

ecommerce/core/management/commands/verify_transactions.py
```diff
@@ -44,6 +46,7 @@
 
 DEFAULT_START_DELTA_TIME = 240
 DEFAULT_END_DELTA_TIME = 60
+VALID_PRODUCT_CLASS_NAMES = [SEAT_PRODUCT_CLASS_NAME, COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME]
```
bderusha, line 49: @BessieSteinberg this is the part you would update for Journals when ready

PR #1896, opened 2018-09-03
Add management command to remove partners conditional offers.

waheedahmed:
To help remove partner completely, currently raising `Condition matching query doesn't exist.` due to duplicate conditional offers.

LEARNER-6265

### DIFF
--- a/ecommerce/extensions/offer/management/commands/remove_partner_offers.py
+++ b/ecommerce/extensions/offer/management/commands/remove_partner_offers.py
@@ -1,0 +1,73 @@
+"""
+This command removes partner conditional offers.
+"""
+from __future__ import unicode_literals
+
+import logging
+
+from django.core.management import BaseCommand
+from django.db.models import signals
+from django.template.defaultfilters import pluralize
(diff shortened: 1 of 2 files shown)


waheedahmed:
@edx/learner-spartans @jibsheet Please review.

jibsheet (approved):
Makes sense to me.
Sometimes we prefer to list everything and then delete everything, but I think prompting for each one is ok here, since I'd rather carefully delete things in ecommerce.

Thanks for putting this together.

waheedahmed:
@jibsheet Did you mean, not prompting for each one? Because this management command is prompting for only once and deleting all offers.

I have added the code to list down all offers to be deleted while prompting, kindly take another look. It will list down the offers like this:
```
2018-09-05 07:27:19,383 WARNING 1366 - Conditional offers to be deleted for partner [edx] 
 1. Test Offer 
 2. Test Offer [1] 
You're going to remove 2 conditional offers. Do you want to continue? [y/n]
```

jibsheet:
@waheedahmed I misread (probably because it was printing one originally).  Printing them all is perfect, thank you.

tasawernawaz (approved):

PR #1898, opened 2018-09-07
ENT-1163 Add Enterprise Customer Catalog field to coupon form

brittneyexline:
This PR adds an Enterprise Customer Catalog field to the coupon. When the coupon is redeemed, the enterprise catalog will be used to look up the products that are available for purchase with the coupon. This will help ensure that enterprise coupons do not need to use the catalog query field in order to specify the set of courses for the coupon, and eliminate the risk that a user redeems a coupon for a course that is not actually part of the enterprise catalog.

This PR is dependent on a change to the discovery service search endpoint to return course runs for course objects, since the enterprise catalog endpoint is a query to the discovery search endpoint, and many enterprise catalogs return courses. This is necessary because ecommerce products are keyed off of course runs ids, and in order to look up the products available from an enterprise catalog with courses, we will need to know the course runs.

**Testing**
When creating or editing a coupon in the coupon administration tool, you should see an opitonal field for Enterprise Customer Catalog (ECC) appear for the single course and multi-course options, below the Enterprise Customer field. _Note that the field does not perform validation on the input beyond it being a valid uuid._  If you fill out the ECC field, the Enterprise Customer field will be required.

The expected behavior upon redeeming a coupon, if the ECC is set for the coupon, is that it will take precedence over the course(s) fields specified in the coupon form. This means that the user will see the list of products for the coupon corresponding to the items in the ECC, for any ECC with course or courserun content. When the user selects a product, the redemption logic will call the consent API which will perform the check that the corresponding course run is in (any of) the Enterprise's catalogs. _Note that this was existing logic and not changed as part of this PR._

**Coupon configured for a query catalog attached to an enterprise:**
- Coupon: https://ecommerce-business.sandbox.edx.org/coupons/4/
- Redemption url: https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=RNBL737ZAUJXUM6E

**Coupon configured with an ECC with all course runs:**
- Coupon: https://ecommerce-business.sandbox.edx.org/coupons/321/
- Redemption url: https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=AVMKCC5UXUS6EZB5

**Couopn configured with an ECC with all courses:**
- Coupon: https://ecommerce-business.sandbox.edx.org/coupons/319/
- Redemption url: https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=5QOTPNISCJ7C2DJQ

**Coupon configured with an ECC with 1 course (DemoX):**
- Coupon: https://ecommerce-business.sandbox.edx.org/coupons/320/
- Redemption url: https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=VM2J6EXIYPFE4KR5

Open Question: If the course is not in the catalog, we do not prompt for consent and the user can still apply the coupon. This seems like a bug, but what should the experience be? Filed bug: https://openedx.atlassian.net/browse/ENT-1223


ecommerce/enterprise/utils.py
```diff
@@ -373,3 +374,36 @@ def has_enterprise_offer(basket):
...
+
+    client = get_enterprise_api_client(site)
+    path = [resource, str(enterprise_catalog)]
+    client = reduce(getattr, path, client)
+
+    response = client.get(
+        limit=limit,
+        page=page,
+    )
+    TieredCache.set_all_tiers(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
```
robrap: Note: I am not doing an overall review, just some quick comments:
1. Is `COURSES_API_CACHE_TIMEOUT` just an hour?  Can we start using a day or more in more places?
- FYI: If needed, the cache can always be refreshed for a particular request.  See https://github.com/edx/edx-django-utils/tree/master/edx_django_utils/cache#force-django-cache-miss
2. Based on Dave's talk the other day, do you also want to investigate if the `limit` is set well, or if it could be larger and reduce paging?

Food for thought.  The first is more important, because if the catalog is rarely changing, it would be great to get better caching.  Thanks.

brittneyexline: regarding number 1, I did not look up what that value was - I took it from the similar call made to discovery to get the course runs associated with the coupon. it seems like they should be the same value since the nature of the calls is this same, so we can certainly look into raising it.

regarding number 2, I have not fully investigated that, because I found out from testing this that it will definitely require some modification of our enterprise api endpoints (which don't support offset), and that would probably blow up the scope of this task. I will file a follow-up ticket for our team.

robrap: For paging, I’d discuss with our non-existent Discovery Guild. According to Dave, we should avoid paging. If paging is necessary, we should be using CursorPagination or possibly paging off a dump of the results. My main point is that Discovery may not be doing what is best here, so for the follow-up ticket, you may want to question it in addition to just wrapping it. Also, is this a place where ecommerce is just looping and paging over the whole catalog? If so, we may want discussions on the Ecommerce side as well about what makes sense. Ultimately, I think we will want a local cache of relevant data.

For cache timeouts...yes...please investigate what the timeout is and whether we can increase it. This is a quick mitigation to our lack of local copy for this data. If you think we need a larger conversation between product and engineers for this, just provide the relevant info and we can have a meeting if need be.

Thank you.

brittneyexline: I believe the discovery service does support offset, and it is being used for coupons linked to a discovery catalog or query. Here, I am introducing a new catalog configuration option for coupons to be linked to enterprise catalogs, which means calling the enterprise catalog api instead of discovery. This api endpoint does not support offset, so I've filed this ticket: https://openedx.atlassian.net/browse/ENT-1217

I've looked into where the COURSES_API_CACHE_TIMEOUT setting is set, and it seems to be set to one hour here: https://github.com/edx/ecommerce/blob/master/ecommerce/settings/base.py#L249

However, there are many calls to the discovery api throughout ecommerce that are using this cache setting, so do you think it would be good to release the configuration change separately so that we can see the effects?

```
$ git grep -n COURSES_API_CACHE_TIMEOUT
ecommerce/coupons/utils.py:82:    TieredCache.set_all_tiers(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
ecommerce/coupons/utils.py:153:    TieredCache.set_all_tiers(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
ecommerce/courses/utils.py:47:    TieredCache.set_all_tiers(cache_key, course, settings.COURSES_API_CACHE_TIMEOUT)
ecommerce/courses/utils.py:87:    TieredCache.set_all_tiers(cache_key, results, settings.COURSES_API_CACHE_TIMEOUT)
ecommerce/enterprise/entitlements.py:183:            TieredCache.set_all_tiers(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
ecommerce/enterprise/utils.py:407:    TieredCache.set_all_tiers(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
ecommerce/extensions/offer/models.py:119:                    TieredCache.set_all_tiers(metadata['cache_key'], in_range, settings.COURSES_API_CACHE_TIMEOUT)
ecommerce/extensions/offer/models.py:356:            TieredCache.set_all_tiers(cache_key, response, settings.COURSES_API_CACHE_TIMEOUT)
ecommerce/extensions/order/utils.py:158:            TieredCache.set_all_tiers(key, entitlement, settings.COURSES_API_CACHE_TIMEOUT)
```

robrap: On the pagination front, I got clarification from Dave.  Please ignore my earlier pagination comments.

Regarding the timeout, you can consider introducing a newly named longer timeout, and we could slowly migrate to the new setting as it makes sense.  Maybe some things can only be cache for an hour, maybe some for a day, maybe some for a week or month?  I don't know, but I don't want a one-size-fits-all timeout of an hour to hold us back from reducing our server-to-server calls in transactions.

Unless you have questions for me, I'm going to leave the rest of this to you and @georgebabey and your team to review.

Thank you.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -326,6 +328,9 @@ def update_range_data(self, request, vouchers):
         else:
             range_data['enterprise_customer'] = None
 
+        if 'enterprise_customer_catalog' in request.data:
+            range_data['enterprise_customer_catalog'] = request.data.get('enterprise_customer_catalog') or None
```
christopappas, line 332: Do we specify the `or None` here because we want to return `None` if `request.data.get('enterprise_customer_catalog')` returns any falsy value?

christopappas (commented):
I have a lot of studying to do, just the 1 question I had when reviewing

ecommerce/extensions/api/v2/views/coupons.py
brittneyexline: i think i added this because empty string was being sent along and then attempting to save in the model, which caused an error because it was expecting a uuid

ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -106,26 +109,50 @@ def retrieve_course_objects(self, results, course_seat_types):
...
+                        course_run_metadata[course_run['key']]['title'] = result['title']
+                        course_run_metadata[course_run['key']]['card_image_url'] = result['card_image_url']
+            elif is_course_run_enrollable(result):
+                course_run_metadata[result['key']] = result
 
         products = []
         for seat_type in course_seat_types.split(','):
             products.extend(Product.objects.filter(
-                course_id__in=nonexpired_course_ids if seat_type == 'professional' else all_course_ids,
+                course_id__in=course_run_metadata.keys(),
```
georgebabey, line 147: Do we know why this was previously only looking at non-expired courses for professional seat?

georgebabey (approved):
Just the one question around expired courses, everything else looks good!

ecommerce/extensions/api/v2/views/vouchers.py
brittneyexline: I haven't gotten any reply except for Lauren from support citing the ticket we groomed. I think we should just move forward with this because it makes sense as expected behavior.

PR #1899, opened 2018-09-08
Upgrade edx-django-utils.

robrap:
ARCH-223


robrap:
@BenjiLee: FYI: We had a RequestCache bug.  The middleware was clearing the request cache for all requests.  We weren't properly using thread.local, because we set up a class variable.  This has been fixed here: https://github.com/edx/edx-django-utils/pull/11

nasthagiri (approved):

BenjiLee:
@robrap  This seems directly related to https://openedx.atlassian.net/browse/EG-5 

Thoughts?

PR #1901, opened 2018-09-12
LEARNER-6347: Fix: Ecommerce Basket is Allowing Coupon Codes to Apply to Incorrect Courses

Ayub-Khan:
LEARNER-6347
The response returns the integer representation of boolean. Because of which check does not work. 
this will fix the issue. 

Have tested the fix locally.

@edx/learner-spartans kindly review.

### DIFF
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -63,14 +63,16 @@ class Benefit(AbstractBenefit):
         Checks the cache to see if each line is in the catalog range specified by the given query
         and tracks identifiers for which discovery service data is still needed.
         """
-        course_run_ids = []
-        course_uuids = []
+        uncached_course_run_ids = []
+        uncached_course_uuids = []
+
         applicable_lines = lines
         for line in applicable_lines:
             if line.product.is_seat_product:
                 product_id = line.product.course.id
             else:  # All lines passed to this method should either have a seat or an entitlement product
                 product_id = line.product.attr.UUID
+
             cache_key = get_cache_key(
                 site_domain=domain,
                 partner_code=partner_code,
@@ -78,26 +80,35 @@ class Benefit(AbstractBenefit):
                 course_id=product_id,
                 query=query
             )
-            cached_response = TieredCache.get_cached_response(cache_key)
-            if not cached_response.is_found:
+            in_catalog_range_cached_response = TieredCache.get_cached_response(cache_key)
+
+            if not in_catalog_range_cached_response.is_found:
                 if line.product.is_seat_product:
-                    course_run_ids.append({'id': product_id, 'cache_key': cache_key, 'line': line})
+                    uncached_course_run_ids.append({'id': product_id, 'cache_key': cache_key, 'line': line})
                 else:
-                    course_uuids.append({'id': product_id, 'cache_key': cache_key, 'line': line})
-            elif cached_response.value is False:
+                    uncached_course_uuids.append({'id': product_id, 'cache_key': cache_key, 'line': line})
+            elif not in_catalog_range_cached_response.value:
                 applicable_lines.remove(line)
-        return course_run_ids, course_uuids, applicable_lines
+
+        return uncached_course_run_ids, uncached_course_uuids, applicable_lines
 
     def get_applicable_lines(self, offer, basket, range=None):  # pylint: disable=redefined-builtin
+        """
+        Returns the basket lines for which the benefit is applicable.
+        """
         applicable_range = range if range else self.range
+
         if applicable_range and applicable_range.catalog_query is not None:
+
             query = applicable_range.catalog_query
             applicable_lines = self._filter_for_paid_course_products(basket.all_lines(), applicable_range)
+
             site = basket.site
             partner_code = site.siteconfiguration.partner.short_code
             course_run_ids, course_uuids, applicable_lines = self._identify_uncached_product_identifiers(
                 applicable_lines, site.domain, partner_code, query
             )
+
             if course_run_ids or course_uuids:
                 # Hit Discovery Service to determine if remaining courses and runs are in the range.
                 try:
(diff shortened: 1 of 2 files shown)


hamzamunir7300 (approved):
:+1: LGTM

Ayub-Khan:
Some tests are failing 
ERROR: /edx/app/ecommerce/ecommerce/ecommerce/static/build/bower_components/bootstrapaccessibilityplugin/_site/bs3/test-infra/s3_cache.py Imports are incorrectly sorted.
Those files do not exist on my local.

ecommerce/extensions/offer/models.py
```diff
@@ -84,7 +84,7 @@ def _identify_uncached_product_identifiers(self, lines, domain, partner_code, qu
                     course_run_ids.append({'id': product_id, 'cache_key': cache_key, 'line': line})
                 else:
                     course_uuids.append({'id': product_id, 'cache_key': cache_key, 'line': line})
-            elif cached_response.value is False:
+            elif not cached_response.value:
```
robrap: This may be correct (I can’t tell yet), but it seems like there is a larger issue that makes this code brittle.

Note: I’m reviewing from a phone, so if I misread the code and my comments below don’t make sense, just let me know. Also, I know you didn’t write the code that I will be commenting on, but I believe these suggestions will make bugs less likely in the future.

Typically, caching code looks as follows:
```
def get_thing(self, stuff):
     # making the key doesn’t need to be a separate method if it is only used in one method
    thing_cache_key = make_thing_cache_key(stuff)
    thing_cached_response = cache.get_cached_response(thing_cache_key)
    if thing_cached_response.is_found:
        return thing_cached_response.value

    # calculate_thing may not be a separate method if it is only needed here
    thing = calculate_thing(stuff)
    cache.set(thing_cache_key, thing)
    return thing
```

The main differences between my example and the code as I’ve Read it are as follows:
1. The cache key is computed in one place only so you know the same key is used to set and get from the cache. In this code, I don’t see the set, so I presume that is happening elsewhere.
2. In the example above, the name ‘thing’ is used all over to let you know what is being retrieved (and set) from the cache.
3. Even if ‘thing’ wasn’t used in the names, because the calculation of the value and setting it in the cache is all in a single function, you can see what is added to the cache and how a cached value should be used.

If in this case you really only need to know if something is in the cache, you might need the following 3 types of methods:
- get_thing (like the pseudo code I wrote)
- _is_thing_in_cache (returns cached_response.is_found. Do not use the result of get_thing to determine if something was cached.)
- _make_thing_cache_key (used by the above two methods to ensure the same key is used for gets and sets.

I would imagine that some type of refactor would be possible to make this code more clear and less likely to have a bug.

Also, please add a test that fails before any changes were added and passes once you have your fix. This will help ensure we avoid regressions.

Thank you!

robrap (changes requested):
Please let me know if you have questions. The main requests are:
1. Refactor cache key code so we know we are setting/getting with the same key, and we are naming whatever is getting cached. I’m still not clear what the cached value is.
2. Add the test as I described.
Thank you.

robrap:
Now that I saw the new comments in the ticket, here are two additional things that it would be great to see:
1. The value set in the cache should be consistent if possible.  If this is meant to be a boolean (what is a name for what is being cached), then can it always be a boolean?  Or is the cached value used differently in different places?  Needs some discovery and then probably clean-up.
2. The ticket states this values is getting set with `metadata['cache_key']`.  Do we have access to `metadata['cache_key']` when we are reading from the cache?  If so, that could be the consistent cache key.  If not, then `metadata['cache_key']` would need to be set with the shared `make_XXX_cache_key` method to ensure we always are using the same cache key.

Thank you.

package-lock.json
```diff
@@ -163,7 +163,7 @@
     "arr-flatten": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/arr-flatten/-/arr-flatten-1.1.0.tgz",
-      "integrity": "sha1-NgSLv/TntH4TZkQxbJlmnqWukfE=",
+      "integrity": "sha512-L3hKV5R/p5o81R7O02IGnwpDmkp6E982XhtbuwSe3O4qOtMMMtodicASA1Cny2U+aCXcNpml+m4dPsvsJ3jatg==",
```
robrap: I don't think these changes should be needed as part of this PR.

ecommerce/extensions/offer/models.py
```diff
@@ -58,46 +58,69 @@ def _filter_for_paid_course_products(self, lines, applicable_range):
...
                 query=query
             )
-            cached_response = TieredCache.get_cached_response(cache_key)
-            if not cached_response.is_found:
+            is_found, contains = self._check_catalog_query_contains(cache_key=cache_key)
+
+            if not is_found:
                 if line.product.is_seat_product:
-                    course_run_ids.append({'id': product_id, 'cache_key': cache_key, 'line': line})
+                    uncached_course_run_ids.append({'id': product_id, 'cache_key': cache_key, 'line': line})
```
robrap: Thanks for the name change.

ecommerce/extensions/offer/models.py
```diff
@@ -58,46 +58,69 @@ def _filter_for_paid_course_products(self, lines, applicable_range):
...
             cache_key = get_cache_key(
                 site_domain=domain,
                 partner_code=partner_code,
                 resource='catalog_query.contains',
                 course_id=product_id,
                 query=query
             )
-            cached_response = TieredCache.get_cached_response(cache_key)
-            if not cached_response.is_found:
+            is_found, contains = self._check_catalog_query_contains(cache_key=cache_key)
```
robrap: I think it would be cleaner to do the following:

1. Can you add the following lines before the TieredCache.set() line in the other method:
```
# Convert to int, because this is what memcached will return, and the request cache should return the same value.
# Note: once the TieredCache is fixed to handle this case, we could remove this line.
in_range = int(in_range)
```
This way, get_cached_response() will return the same value every time.

2. I'd go back to the previous TieredCache.get_cache_response call directly from this method, but I'd rename `cached_response` to `in_catalog_range_cached_response`.

robrap (commented):
Thanks again for the find.

Ayub-Khan:
Done.

robrap (approved):

PR #1902, opened 2018-09-12
Add custom program applicator

bbaker6225:
Custom program applicator aims to reduce queries for program offers by
matching offers with the bundle identifier.

[EG-3](https://openedx.atlassian.net/browse/EG-3)

### DIFF
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -5,6 +5,7 @@ from urllib import unquote, urlencode
 
 import newrelic.agent
 import pytz
+import waffle
 from django.conf import settings
 from django.contrib import messages
 from django.db import transaction
@@ -12,12 +13,14 @@ from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_class, get_model
 
 from ecommerce.courses.utils import mode_for_product
+from ecommerce.extensions.offer.constants import PROGRAM_APPLICATOR_USE_FLAG
 from ecommerce.extensions.order.exceptions import AlreadyPlacedOrderException
 from ecommerce.extensions.order.utils import UserAlreadyPlacedOrder
 from ecommerce.extensions.payment.utils import embargo_check
 from ecommerce.referrals.models import Referral
 
 Applicator = get_class('offer.applicator', 'Applicator')
+ProgramApplicator = get_class('offer.applicator', 'ProgramApplicator')
 Basket = get_model('basket', 'Basket')
 BasketAttribute = get_model('basket', 'BasketAttribute')
 BasketAttributeType = get_model('basket', 'BasketAttributeType')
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/offer/applicator.py
```diff
@@ -0,0 +1,71 @@
...
+
+from oscar.apps.offer.applicator import Applicator
+from oscar.core.loading import get_model
+
+BasketAttribute = get_model('basket', 'BasketAttribute')
+BasketAttributeType = get_model('basket', 'BasketAttributeType')
+BUNDLE = 'bundle_identifier'
+
+
+class ProgramApplicator(Applicator):
```
robrap, line 16: It would be great if this could be written more generically to work for the other offer types.  Is the idea to start with this and refactor later?  Thanks!
FYI: @BenjiLee

bbaker6225: We originally considered starting with a base applicator to try and make it more generic/work for the other offers, but I felt that I didn't have enough information on how the other offers worked to start that generic.  

Long story short, my idea was to start with this, and then refactor into a base applicator later, but we can definitely change plans if necessary.

robrap: Sounds like a fine plan.  Just wanted to ensure you knew where we eventually wanted to land.  Thanks.

robrap:
Note: the following comment may not have any reality, but are there marketing email links or any other ways in which we might come here and not have the bundle id?  Or is it guaranteed to be there?

bbaker6225:
@robrap I'm not aware of anything that might lead here without having the bundle, but if something did arrive without it, the program applicator would just get all the site offers, basket offers, etc. like oscar's default applicator does.

robrap:
@bbaker6225: Great.  I guess we just need to: 
1. Ensure that if it happens it does result in all offers loaded, rather than no offers.
2. Maybe we could add logging or something to see if this actually is happening in the wild, so we could plug that hole if necessary.  NOTE: Sometimes when we add temporary logging we hide it behind a flag to ensure we can turn it off if it is noisy.

bbaker6225:
Note: The flag migration didn't seem appropriate since this will most likely be a temporary flag, so that can be created in the admin page when this is live.

BenjiLee:
note: Codcov is probably failing because of the waffle flag path.

bbaker6225:
Yep, adding a log test now, thanks

ecommerce/extensions/basket/utils.py
```diff
@@ -396,5 +403,24 @@ def _apply_voucher_on_basket(voucher, request, basket):
         request (Request): Request object
     """
     basket.vouchers.add(voucher)
-    Applicator().apply(basket, request.user, request)
+    if (
+            waffle.flag_is_active(request, PROGRAM_APPLICATOR_COMPARE_FLAG) and
+            waffle.flag_is_active(request, PROGRAM_APPLICATOR_LOG_FLAG)
+    ):  # pragma: no cover
+        program_applicator_offers = ProgramApplicator().get_offers(basket, request.user, request)
```
BenjiLee: What we want to compare here is the final price of the basket after the offer is applied.

bbaker6225: Once we call Applicator().apply() the basket will be changed, so that will change how the second one behaves.  Don't we want to compare the offers they return since that is the only thing that will affect the price?

BenjiLee (approved):

PR #1903, opened 2018-09-13
Disable SSI test

coryleeio:


robrap:
Does the test ensure it is never mistakenly enabled? Or is that not a possibility?

BenjiLee:
@coryleeio Could you add more context in the PR?

robrap (approved):
@coryleeio explained offline why this test is causing us issues.  All set.

robrap:
@coryleeio: We do need this to be green before merging.

coryleeio:
https://openedx.atlassian.net/browse/OPS-3301

has more context if needed for anyone later

BenjiLee (approved):

PR #1904, opened 2018-09-14
Learner 6375

Ayub-Khan:
Now this will check for the discounts and offers validity on the coupons carried on by the previous basket. 
Updated coupon add view to remove the invalid coupon.

### DIFF
--- a/ecommerce/extensions/basket/tests/test_utils.py
+++ b/ecommerce/extensions/basket/tests/test_utils.py
@@ -20,6 +20,7 @@ from ecommerce.extensions.basket.tests.mixins import BasketMixin
 from ecommerce.extensions.basket.utils import (
     ENTERPRISE_CATALOG_ATTRIBUTE_TYPE,
     add_utm_params_to_url,
+    apply_voucher_on_basket_and_check_discount,
     attribute_cookie_data,
     get_basket_switch_data,
     prepare_basket
@@ -74,8 +75,8 @@ class BasketUtilsTests(DiscoveryTestMixin, BasketMixin, TestCase):
         """ Verify a basket is returned and contains a voucher and the voucher is applied. """
         # Prepare a product with price of 100 and a voucher with 10% discount for that product.
         product = ProductFactory(stockrecords__price_excl_tax=100)
-        new_range = RangeFactory(products=[product, ])
-        voucher, product = prepare_voucher(_range=new_range, benefit_value=10)
+        new_range = RangeFactory(products=[product])
+        voucher, __ = prepare_voucher(_range=new_range, benefit_value=10)
 
         stock_record = StockRecord.objects.get(product=product)
         self.assertEqual(stock_record.price_excl_tax, 100.00)
@@ -550,6 +551,22 @@ class BasketUtilsTests(DiscoveryTestMixin, BasketMixin, TestCase):
         self.assertEqual(basket.lines.first().product, product)
         self.assertEqual(basket.vouchers.count(), 0)
 
+    def test_prepare_basket_removes_existing_basket_invalid_range_voucher(self):
+        """
+        Tests that prepare_basket removes an existing basket voucher that is not
+        valid for the product and used to purchase that product.
+        """
+        product = ProductFactory(stockrecords__partner__short_code='xyz', stockrecords__price_excl_tax=100)
+        invalid_range_voucher, __ = prepare_voucher()
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+
+        basket.vouchers.add(invalid_range_voucher)
+        basket = prepare_basket(self.request, [product])
+
+        self.assertIsNotNone(basket)
+        self.assertEqual(basket.lines.first().product, product)
+        self.assertEqual(basket.vouchers.count(), 0)
+
     def test_prepare_basket_applies_existing_basket_valid_voucher(self):
         """
         Tests that prepare_basket applies an existing basket voucher that is valid
@@ -570,6 +587,62 @@ class BasketUtilsTests(DiscoveryTestMixin, BasketMixin, TestCase):
         self.assertEqual(basket.lines.first().product, product)
         self.assertIsNotNone(basket.applied_offers())
         self.assertEqual(basket.total_discount, 10.00)
+
+    def test_apply_voucher_on_basket_and_check_discount_with_valid_voucher(self):
+        """
+        Tests apply_voucher_on_basket_and_check_discount when called with valid voucher
+        applies voucher and returns the correct values.
+        """
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        voucher, product = prepare_voucher()
+        basket.add_product(product, 1)
+        applied, msg = apply_voucher_on_basket_and_check_discount(voucher, self.request, basket)
+        self.assertEqual(applied, True)
+        self.assertIsNotNone(basket.applied_offers())
+        self.assertEqual(msg, "Coupon code '{code}' added to basket.".format(code=voucher.code))
+
+    def test_apply_voucher_on_basket_and_check_discount_with_invalid_voucher(self):
+        """
+        Tests apply_voucher_on_basket_and_check_discount when called with invalid voucher
+        does not apply voucher and returns the correct values.
+        """
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        product = ProductFactory(stockrecords__partner__short_code='test1', stockrecords__price_excl_tax=100)
+        voucher, __ = prepare_voucher()
+        basket.add_product(product, 1)
+        applied, msg = apply_voucher_on_basket_and_check_discount(voucher, self.request, basket)
+        self.assertEqual(applied, False)
+        self.assertEqual(basket.applied_offers(), {})
+        self.assertEqual(msg, 'Basket does not qualify for coupon code {code}.'.format(code=voucher.code))
+
+    def test_apply_voucher_on_basket_and_check_discount_with_invalid_product(self):
+        """
+        Tests apply_voucher_on_basket_and_check_discount when called with invalid product
+        does not apply voucher and returns the correct values.
+        """
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        product = ProductFactory(stockrecords__partner__short_code='test1', stockrecords__price_excl_tax=0)
+        voucher, __ = prepare_voucher(_range=RangeFactory(products=[product]))
+        basket.add_product(product, 1)
+        applied, msg = apply_voucher_on_basket_and_check_discount(voucher, self.request, basket)
+        self.assertEqual(applied, False)
+        self.assertEqual(basket.applied_offers(), {})
+        self.assertEqual(msg, 'Basket does not qualify for coupon code {code}.'.format(code=voucher.code))
+
+    def test_apply_voucher_on_basket_and_check_discount_with_multiple_vouchers(self):
+        """
+        Tests apply_voucher_on_basket_and_check_discount when called with a basket already
+        containing a valid voucher it only checks the new voucher.
+        """
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        product = ProductFactory(stockrecords__partner__short_code='test1', stockrecords__price_excl_tax=10)
+        invalid_voucher, __ = prepare_voucher(code='TEST1')
+        valid_voucher, __ = prepare_voucher(code='TEST2', _range=RangeFactory(products=[product]))
+        basket.add_product(product, 1)
+        basket.vouchers.add(valid_voucher)
+        applied, msg = apply_voucher_on_basket_and_check_discount(invalid_voucher, self.request, basket)
+        self.assertEqual(applied, False)
+        self.assertEqual(msg, 'Basket does not qualify for coupon code {code}.'.format(code=invalid_voucher.code))
 
 
 class BasketUtilsTransactionTests(TransactionTestCase):
--- a/ecommerce/extensions/basket/utils.py
+++ b/ecommerce/extensions/basket/utils.py
@@ -10,6 +10,7 @@ from django.conf import settings
 from django.contrib import messages
 from django.db import transaction
 from django.utils.translation import ugettext_lazy as _
+from oscar.apps.basket.signals import voucher_addition
 from oscar.core.loading import get_class, get_model
 
 from ecommerce.courses.utils import mode_for_product
@@ -115,7 +116,7 @@ def prepare_basket(request, products, voucher=None):
         basket.clear_vouchers()
         is_valid, message = validate_voucher(voucher, request.user, basket, request.site)
         if is_valid:
-            _apply_voucher_on_basket(voucher, request, basket)
+            apply_voucher_on_basket_and_check_discount(voucher, request, basket)
         else:
             logger.info(message)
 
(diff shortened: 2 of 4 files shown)


robrap:
I'll look at this more deeply, but first please add a test that fails without your change and passes with your change.  Thank you.

robrap:
@Ayub-Khan: I'm planning on getting another reviewer from @edx/ecommerce, and/or you can have @waheedahmed (who is a member) review.  Thank you for fixing these issues.

staubina:
@waheedahmed  Please review this ticket when you are available.

robrap:
As noted above, please do not merge any coupon related fixes without tests.  We are running into many coupon related issues.  Thank you.

ecommerce/extensions/basket/utils.py
```diff
@@ -401,19 +402,41 @@ def validate_voucher(voucher, user, basket, request_site):
...
+        if discount['voucher']:
+            found_discount = True
+            break
+
+    if found_discount:
+        logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
+        msg = _("Coupon code '{code}' added to basket.").format(code=voucher.code)
+        return True, msg
+    else:
+        msg = _('Your basket does not qualify for a coupon code discount.')
```
waheedahmed: Can we log warning here for investigation purpose?

waheedahmed (approved):
Looks good to me, just one minor comment. Also please fix code coverage and let me know.

ecommerce/extensions/basket/utils.py
Ayub-Khan: done

Ayub-Khan:
@robrap  Kindly review.

waheedahmed (approved):

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -550,6 +551,22 @@ def test_prepare_basket_removes_existing_basket_invalid_voucher(self):
         self.assertEqual(basket.lines.first().product, product)
         self.assertEqual(basket.vouchers.count(), 0)
 
+    def test_prepare_basket_removes_existing_basket_invalid_range_voucher(self):
+        """
+        Tests that prepare_basket removes an existing basket voucher that is not
+        valid for the product and used to purchase that product.
+        """
+        product = ProductFactory(stockrecords__partner__short_code='xyz', stockrecords__price_excl_tax=100)
+        invalid_range_vocuher, __ = prepare_voucher()
```
robrap: Typo in `invalid_range_vocuher`

ecommerce/extensions/basket/utils.py
```diff
@@ -115,7 +116,7 @@ def prepare_basket(request, products, voucher=None):
         basket.clear_vouchers()
         is_valid, message = validate_voucher(voucher, request.user, basket, request.site)
         if is_valid:
-            _apply_voucher_on_basket(voucher, request, basket)
+            apply_voucher_on_basket_and_check_discount(voucher_addition, voucher, request, basket)
```
robrap: We are sending in `voucher_addition` as the caller, to later call:
```
voucher_addition.send(sender=voucher_addition, ...)
```
Seems odd.  There are also no tests around this signal addition.  Can you explain what it is all about?

ecommerce/extensions/basket/utils.py
```diff
@@ -401,19 +402,42 @@ def validate_voucher(voucher, user, basket, request_site):
...
     if waffle.flag_is_active(request, PROGRAM_APPLICATOR_USE_FLAG):  # pragma: no cover
         ProgramApplicator().apply(basket, request.user, request)
     else:
         Applicator().apply(basket, request.user, request)
 
-    logger.info('Applied Voucher [%s] to basket [%s].', voucher.code, basket.id)
+    # Recalculate discounts to see if the voucher gives any
+    discounts_after = basket.offer_applications
+
+    # Look for discounts from this new voucher
```
robrap, line 430: It feels very odd that this logic would be needed.  I haven't gone back to the django-oscar docs, but doesn't it feel strange to you all that we would need to do this calculation to determine if it was applied to determine if it should be clear out?
@BenjiLee, @jmyatt: Do you have any experience from the work you were doing?

robrap (commented):
I'd like to better understand the situation and why this wouldn't be better built into oscar's voucher validation.  Can you someone help clarify why this is the right solution?

FYI: @BenjiLee, @jmyatt

ecommerce/extensions/basket/utils.py
Ayub-Khan: just used as a counter 
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/voucher/receivers.py

ecommerce/extensions/basket/tests/test_utils.py
Ayub-Khan: ok

ecommerce/extensions/basket/utils.py
Ayub-Khan: Only validation oscar provides us is the is_available_to_user. (https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/voucher/abstract_models.py#L195) It does not provide any validation methods for the product. We can move all that to a new method there called is_valid_for_basket.

robrap: @georgebabey: Can you help provide insight on the best way to provide the validation that @Ayub-Khan is trying to achieve?  Based on his last comment, I'm not sure if we are missing `can_apply_condition` logic, but I have very little voucher/offer experience at this point.

Also: Feel free to reach out to Product to determine if we need/want to be carrying the coupons over from basket to basket, rather than having them re-enter each time.

Ayub-Khan: @robrap i think we can move forward on this if @georgebabey  do not have any thing to add on this. Deen gave the product opinion that we want to carry the coupons over from basket to basket.

georgebabey: @Ayub-Khan Apologies for the delay here. I don't have anything to add - this seems to work given that we want to carry the voucher forward across baskets.

ecommerce/extensions/basket/utils.py
robrap: Note: It feels like it would be more accurate to state that the 'basket does not qualify for coupon code `{code}`'.
However:
1. I see you are just moving this code, so this doesn't need to happen in this PR.
2. I see this code originally comes from Django Oscar code, so maybe I am missing something and my suggestion is off?

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -571,6 +588,62 @@ def test_prepare_basket_applies_existing_basket_valid_voucher(self):
...
+        """
+        basket = BasketFactory(owner=self.request.user, site=self.request.site)
+        product = ProductFactory(stockrecords__partner__short_code='test1', stockrecords__price_excl_tax=10)
+        invalid_voucher, __ = prepare_voucher(code='TEST1')
+        valid_voucher, __ = prepare_voucher(code='TEST2', _range=RangeFactory(products=[product]))
+        basket.add_product(product, 1)
+        basket.vouchers.add(valid_voucher)
+        applied, msg = apply_voucher_on_basket_and_check_discount(self, invalid_voucher, self.request, basket)
+        self.assertEqual(applied, False)
+        self.assertEqual(msg, 'Your basket does not qualify for a coupon code discount.')
```
robrap: Note: This is an odd test since I _think_ we always clear the basket vouchers before making this call.

ecommerce/extensions/basket/tests/test_utils.py
```diff
@@ -571,6 +588,62 @@ def test_prepare_basket_applies_existing_basket_valid_voucher(self):
...
+        voucher, product = prepare_voucher()
+        basket.add_product(product, 1)
+        applied, msg = apply_voucher_on_basket_and_check_discount(self, voucher, self.request, basket)
+        self.assertEqual(applied, True)
+        self.assertIsNotNone(basket.applied_offers())
+        self.assertEqual(msg, "Coupon code '{code}' added to basket.".format(code=voucher.code))
+
+    def test_apply_voucher_on_basket_and_check_discount_with_invalid_voucher(self):
+        """
+        Tests apply_voucher_on_basket_and_check_discount when valled with invalid voucher
```
robrap: Typo: 'valled' => 'called'

ecommerce/extensions/basket/utils.py
```diff
@@ -401,19 +402,42 @@ def validate_voucher(voucher, user, basket, request_site):
...
     Args:
+        caller (Class): Used to send signals
         voucher (Voucher): voucher to be applied
         basket (Basket): basket object on which voucher is going to be applied
         request (Request): Request object
     """
+    # Reset any site offers that are applied so that only one offer is active.
+    basket.reset_offer_applications()
     basket.vouchers.add(voucher)
+    voucher_addition.send(sender=caller, basket=basket, voucher=voucher)
```
robrap: Since we aren't calling from a class any more, can you just use:
`voucher_addition.send(sender=None, basket=basket, voucher=voucher)`
Then remove caller from the arguments and docstring.

ecommerce/extensions/basket/utils.py
```diff
@@ -401,19 +402,42 @@ def validate_voucher(voucher, user, basket, request_site):
 
 
 @newrelic.agent.function_trace()
-def _apply_voucher_on_basket(voucher, request, basket):
+def apply_voucher_on_basket_and_check_discount(caller, voucher, request, basket):
     """
     Applies voucher on a product.
```
robrap, line 408: Now that I found the original code, would you mind add the following comment:
```
Note: some of this code was adapted from `apply_voucher_to_basket` in Oscar's `BasketAddView`.
See https://github.com/django-oscar/django-oscar/blob/1.5.1/src/oscar/apps/basket/views.py#L310-L356
```

ecommerce/extensions/basket/tests/test_utils.py
robrap: Reminder.

ecommerce/extensions/basket/utils.py
robrap: **tl;dr**: all good.  Thanks.  Sorry about the confusion.

**Longer:**
Sorry @Ayub-Khan and @georgebabey.
1. I have been in and out of this PR too many times, and I either missed or eventually forgot that this was just moving code from below.
2. I still felt like there should have been a different way, but you both are correct.  I didn't realize that this **IS** django-oscar code just duplicated here.  Note: I asked to have a comment above to reference the source so that would be more clear for me the next time I ask you all. :)

ecommerce/extensions/basket/views.py
```diff
@@ -430,41 +431,24 @@ def get_context_data(self, **kwargs):
...
-        found_discount = False
-        for discount in discounts_after:
-            if discount['voucher'] and discount['voucher'] == voucher:
-                found_discount = True
-                break
-        if not found_discount:
-            messages.warning(
-                self.request,
-                _('Your basket does not qualify for a coupon code discount.'))
+        valid, msg = apply_voucher_on_basket_and_check_discount(self, voucher, self.request, self.request.basket)
```
robrap: Future: If every call to `apply_voucher_on_basket_and_check_discount` is preceded by a call to `basket.clear_vouchers()` and `validate_voucher`, maybe we could create a new method like the following pseudo-code:
```
def clear_and_apply_voucher():
    self.request.basket.clear_vouchers()
    _pre_validate_voucher()
    _apply_voucher_and_post_validate()
```
NOTE: If we could get `_` prefixes and make these functions private, then we can lessen the chances that someone could call any of these functions without the others.

I'd love to see a ticket for this final clean-up, but it should not block this PR.

robrap (commented):
@Ayub-Khan: This is nearly ready, after you add some minor fixes (typos, fix signal, add comments as asked).  When you complete the above, you can have @waheedahmed merge without further review.

However, I would like a follow-up PR to take this refactor one step further as noted in my comment regarding the pseudo-code for `clear_and_apply_voucher`.

Let me know if you have any questions.  Thank you for your work and your patience.

ecommerce/extensions/basket/utils.py
Ayub-Khan: done

ecommerce/extensions/basket/utils.py
Ayub-Khan: @robrap We are not just only moving code we are introducing discount check on the basket add item view too if some voucher is present.
We can move the validate code to a new method in voucher model "is_available_for_basket". it will look some thing like that (https://github.com/django-oscar/django-oscar/blob/a76b0af31efc8cab284658a8b5a9982b4f454a2b/src/oscar/apps/voucher/abstract_models.py#L222-L238).

ecommerce/extensions/basket/tests/test_utils.py
Ayub-Khan: We do clear the voucher in prepare_basket. This is just for to cover the condition that only discounted voucher is checked. not the other valid voucher.

ecommerce/extensions/basket/tests/test_utils.py
Ayub-Khan: done

ecommerce/extensions/basket/utils.py
Ayub-Khan: Done

ecommerce/extensions/basket/utils.py
Ayub-Khan: done

ecommerce/extensions/basket/views.py
Ayub-Khan: As we are gonna hold this bcz of master lauch i think we can move forward with this if you say.

ecommerce/extensions/basket/utils.py
robrap: Thanks @Ayub-Khan.  A couple of things.
1. I meant that I had missed that the discount check was code that was already written, and even originally came from Oscar.  I understand that it was being moved in order to re-use it.
2. I'm still not clear on why Django Oscar has this discount check when their also exists `is_available_for_basket` on the Voucher.  I haven't had enough time to spend with the Django Oscar code, and there is something I still don't get about why the discount check is needed and what it does that a check to `is_available_for_basket` doesn't do.  I also didn't have a chance to look for where it does call `is_available_for_basket`.
3. I wouldn't advocate for a switch to `is_available_for_basket` before better understanding the purpose of these in Django Oscar.

As a general rule, I'd love it if our code was doing things as Django Oscar would do it, and that we are extending it as it expects to be extended.  When we go our own way, I'd love it if we could leave better comments regarding why we had to go our own way, and what Oscar code we based our code on.  Without this knowledge, we will continue to have bugs we don't understand and we will have a nightmare when and if we need to ever upgrade.

I very much appreciate your help on all of this.  I completely understand that this is not code you wrote, nor anyone that is left here. :)  So, as a team, I'm just trying to make it simpler for all of us to come back to this code.  Thanks again.

FYI: @waheedahmed

Ayub-Khan: @robrap It is not a part of django-oscar yet. https://github.com/django-oscar/django-oscar/pull/2844.  I made this as this was a better way to do it. Lets see how it goes. 
I completely agree with you.

robrap: Sorry @Ayub-Khan.  I thought you were pointing me toward existing Oscar code.  I didn't realize that was your code.  Yes - your code is more what I would have expected to be there from the start.  Thanks.  And I understand that this may or may not make it into Oscar any time soon and I don't know when and if we'll be able to upgrade, but it makes more sense to me.

PR #1905, opened 2018-09-14
ENT-1163 Extend COURSES_API_CACHE_TIMEOUT from 1 hr to 24 hrs

brittneyexline:
Relating to https://github.com/edx/ecommerce/pull/1898, @robrap requested that we raise the value of COURSE_API_CACHE_TIMEOUT. From what I can tell, this is where the value gets set. I am making this change separately since various calls throughout ecommerce to discovery use this value, so we can monitor it separately from my changeset.


brittneyexline:
I didn't see your comment until after I opened this @robrap... sorry about that!
I will create a new setting and use that for the call I introduced in my PR.

PR #1906, opened 2018-09-14
Add NewRelic tracing to utils and BasketSummary

BenjiLee:
We are working on the program applicator and would like to get more data for the before and after.


jmyatt (approved):

robrap:
@BenjiLee: Don't forget about the following for refactor testing:
https://github.com/joealcorn/laboratory

nedbat:
Is this code permanent? We should use NewRelic conditionally: try importing it, and if it fails, don't use it.

BenjiLee:
@nedbat This code is not permanent. 

We enable this for before and afters when deploying performance related code and then remove it afterwards.

PR #1907, opened 2018-09-14
Fix duplicate Condition bug

tuchfarber:
If a condition with the program_uuid that you are trying to update to already exists, don't update it so we don't get duplicates.

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -421,7 +421,8 @@ class CouponViewSet(EdxOrderPlacementMixin, viewsets.ModelViewSet):
         voucher_offers = vouchers.first().offers
         voucher_offer = voucher_offers.first()
 
-        if program_uuid:
+        # Only update Conditions if one doesn't already exist with the new program_uuid
+        if program_uuid and not Condition.objects.filter(program_uuid=program_uuid).exists():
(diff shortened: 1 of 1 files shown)


robrap (commented):
Could you add a test? Thanks.

robrap:
@tuchfarber: Is this tied to a ticket that we can follow?  Is this a bug that should be fixed?  Other?  Thanks.

tuchfarber:
Sorry about this. It's a bug that should be fixed. I backburnered it for a couple days then completely forgot about it. 

The bug is: 
`_get_or_create_offer` assumes 1 coupon per program.

This method (`update_coupon_offer`) allows you to change a program offer to a program_uuid that already exists, thus breaking the _get_or_create_offer call later in the function. Unsure if these transactions are atomic or not. If not, it would leave the offers in a broken state. 

I can put in an ecomm-guild ticket if you want.

robrap:
Ecommerce transactions are atomic by default. Not sure if you saw a problem occur, or just saw a potential problem in the code. Either way, would you mind creating a ticket so we can better track this? Thank you.

feanil:
This code was updated differently by other work and this line no longer exists.  It looks like it moved over here: https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/voucher/utils.py

and now does a get or create.

PR #1911, opened 2018-09-18
Fix coupon offers not being set with correct partner.

waheedahmed:
LEARNER-6393

### DIFF
--- a/ecommerce/extensions/test/factories.py
+++ b/ecommerce/extensions/test/factories.py
@@ -122,7 +122,7 @@ def prepare_voucher(code='COUPONTEST', _range=None, start_datetime=None, end_dat
             benefit=benefit,
             condition=condition,
             email_domains=email_domains,
-            site=site,
+            partner=site.siteconfiguration.partner if site else None,
             priority=OFFER_PRIORITY_VOUCHER
         )
     voucher.offers.add(offer)
--- a/ecommerce/extensions/voucher/tests/test_utils.py
+++ b/ecommerce/extensions/voucher/tests/test_utils.py
@@ -184,7 +184,8 @@ class UtilTests(CouponMixin, DiscoveryMockMixin, DiscoveryTestMixin, LmsApiMockM
         email_domains = 'edx.org,example.com'
         self.data.update({
             'email_domains': email_domains,
-            'name': 'Tešt voučher'
+            'name': 'Tešt voučher',
+            'site': self.site
         })
         vouchers = create_vouchers(**self.data)
 
@@ -200,6 +201,7 @@ class UtilTests(CouponMixin, DiscoveryMockMixin, DiscoveryTestMixin, LmsApiMockM
         self.assertEqual(voucher_offer.benefit.range.catalog, self.catalog)
         self.assertEqual(voucher_offer.email_domains, email_domains)
         self.assertEqual(voucher_offer.priority, OFFER_PRIORITY_VOUCHER)
+        self.assertEqual(voucher_offer.partner, self.partner)
         self.assertEqual(len(coupon_voucher.vouchers.all()), 11)
         self.assertEqual(voucher.end_datetime, self.data['end_datetime'])
         self.assertEqual(voucher.start_datetime, self.data['start_datetime'])
--- a/ecommerce/extensions/voucher/utils.py
+++ b/ecommerce/extensions/voucher/utils.py
@@ -374,6 +374,7 @@ def _get_or_create_offer(
         max_global_applications=max_uses,
         email_domains=email_domains,
         site=site,
+        partner=site.siteconfiguration.partner if site else None,
         priority=OFFER_PRIORITY_VOUCHER,
     )
 
--- a/ecommerce/extensions/basket/tests/test_views.py
+++ b/ecommerce/extensions/basket/tests/test_views.py
@@ -40,7 +40,7 @@ from ecommerce.extensions.payment.constants import CLIENT_SIDE_CHECKOUT_FLAG_NAM
 from ecommerce.extensions.payment.forms import PaymentForm
 from ecommerce.extensions.payment.tests.processors import DummyProcessor
 from ecommerce.extensions.test.factories import create_order, prepare_voucher
-from ecommerce.tests.factories import ProductFactory, SiteFactory, StockRecordFactory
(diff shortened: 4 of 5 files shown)


ecommerce/extensions/voucher/tests/test_utils.py
```diff
@@ -200,6 +201,7 @@ def test_create_vouchers(self):
         self.assertEqual(voucher_offer.benefit.range.catalog, self.catalog)
         self.assertEqual(voucher_offer.email_domains, email_domains)
         self.assertEqual(voucher_offer.priority, OFFER_PRIORITY_VOUCHER)
+        self.assertEqual(voucher_offer.partner, self.partner)
```
robrap, line 204: Can you confirm that this test covers the actual change, and not just the factory change? If not, is there a way to get a real test?

robrap (commented):
I guess we missed this. Looks good, but one question. Also, can you confirm that you were able to manually test this?

ecommerce/extensions/voucher/tests/test_utils.py
waheedahmed: @robrap Let me try with a real test. Yes, I tested it manually by creating a coupon on my local machine and partner was successfully added to offers.

ecommerce/extensions/test/factories.py
```diff
@@ -122,7 +122,7 @@ def prepare_voucher(code='COUPONTEST', _range=None, start_datetime=None, end_dat
             benefit=benefit,
             condition=condition,
             email_domains=email_domains,
-            site=site,
```
staubina, line 125: Can we set the site too for safety? Or would that cause other issues?

waheedahmed: @staubina We are going to remove site object completely and using partner instead.

robrap: Agreed. Since we switched from site offers to partner offers, I think leaving the site could just lead to other future bugs and misunderstanding.

robrap:
I can’t yet see the Travis failures. I generally approve, but want a green build first. We may leave an additional test to a follow-up PR.

bill-filler:
@waheedahmed can you grep the code to make sure there are not any other places where we are creating the ConditionalOffer and not correctly setting the partner? I want to make sure we haven't missed any other cases.

ecommerce/extensions/test/factories.py
staubina: Ok, we should update the white label team as well.

@bill-filler FYI

ecommerce/extensions/voucher/tests/test_utils.py
waheedahmed: @robrap `create_vouchers ` is not a factory function it is the actual function which is used to create vouchers and offers while creating coupons.

waheedahmed:
@bill-filler Yes, I did grep the code and found this place only and a test case where conditional offer is being created.

ecommerce/extensions/voucher/tests/test_utils.py
robrap: Great. I wasn’t sure from just the PR code. 

Still may want to consider a deeper integration test of creating an offer and ensuring it applies appropriately, akin to what you tested manually.

ecommerce/extensions/voucher/utils.py
```diff
@@ -373,7 +373,7 @@ def _get_or_create_offer(
         benefit=offer_benefit,
         max_global_applications=max_uses,
         email_domains=email_domains,
-        site=site,
+        partner=site.siteconfiguration.partner if site else None,
```
robrap: What would happen if Partner were None? Is that ever valid, or do we need to log an error?

ecommerce/extensions/voucher/tests/test_utils.py
waheedahmed: @robrap Sure thing, I'll try to add a test case in follow-up PR.

ecommerce/extensions/test/factories.py
waheedahmed: @robrap So, should I set the site as well for now?

bill-filler:
Also wondering if we need a migration similar to the previous PR which updates the partner ID field for all the conditional offers. The use case would be to fix all of the offers that have been created since the original PR landed all have null partner ID currently

ecommerce/extensions/voucher/utils.py
```diff
@@ -374,6 +374,7 @@ def _get_or_create_offer(
         max_global_applications=max_uses,
         email_domains=email_domains,
         site=site,
```
waheedahmed, line 376: @staubina @robrap Added the site back.

staubina: I thought you an @robrap  wanted to remove it.

staubina:
> Also wondering if we need a migration similar to the previous PR which updates the partner ID field for all the conditional offers. The use case would be to fix all of the offers that have been created since the original PR landed all have null partner ID currently

@waheedahmed Please consider this request as well.  We have a number of items without partner right now that we might want to fix.

waheedahmed:
@staubina working on it.

waheedahmed:
@bill-filler, @robrap, @staubina Added the migration to update partner for offers created in Sep, 2018. Kindly take a look.

bill-filler (approved):
looks good, thanks

ecommerce/extensions/voucher/utils.py
robrap: @staubina @waheedahmed: I’ll Let you both decide what is safest in the short term. Once the dust has settled, there should be a follow-up clean up if you’ve decided it is safer to leave it for now.

robrap (commented):
Does the migration need to go separately, or do we just need to ensure no test coupons come in on the old cluster during deployment? At a minimum, we should check the db afterwards to ensure all is well.

waheedahmed:
@robrap Actually that is a valid concern, let me make a separate PR for the migration which lands on production after this PR. Does this sounds good?

waheedahmed:
@robrap Created the separate PR for the migration https://github.com/edx/ecommerce/pull/1912.

CC: @bill-filler @staubina

PR #1914, opened 2018-09-21
Add option to use cybersource email for notifications

kaizoku:
This PR will use the email address returned from the CyberSource checkout process to send ecommerce notifications, rather than the user's email address from edx-platform.

**Sandbox URL**: TBD - sandbox is being provisioned.

**Testing instructions**:

1. Complete the purchase of a seat in the sandbox providing a different email address from the one configured in edX during the CyberSource checkout sequence.
2. Check to see that the seat fulfillment notification is sent to the address used on CyberSource.

**Author notes and concerns**:

1. Should this be behind a waffle flag?

**Reviewers**
- [ ] (@xirdneh)
- [ ] edX reviewer[s] TBD

### DIFF
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -94,7 +94,7 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
 
 @receiver(post_checkout, dispatch_uid='send_completed_order_email')
 @silence_exceptions("Failed to send order completion email.")
-def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable=unused-argument
+def send_course_purchase_email(sender, order=None, request=None, **kwargs):  # pylint: disable=unused-argument
     """Send course purchase notification email when a course is purchased."""
     if waffle.switch_is_active('ENABLE_NOTIFICATIONS'):
         # We do not currently support email sending for orders with more than one item.
@@ -116,6 +116,8 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
                     order_number=order.number,
                     site_configuration=order.site.siteconfiguration
                 )
+
+                recipient = request.POST.get('req_bill_to_email', order.user.email)
 
                 if provider_data:
                     send_notification(
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -129,7 +129,8 @@ def send_course_purchase_email(sender, order=None, request=None, **kwargs):  # p
                             'credit_hours': product.attr.credit_hours,
                             'credit_provider': provider_data['display_name'],
                         },
-                        order.site
+                        order.site,
+                        recipient
                     )
 
         else:
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -117,7 +117,7 @@ def send_course_purchase_email(sender, order=None, request=None, **kwargs):  # p
                     site_configuration=order.site.siteconfiguration
                 )
 
-                recipient = request.POST.get('req_bill_to_email', order.user.email)
+                recipient = request.POST.get('req_bill_to_email', order.user.email) if request else order.user.email
 
                 if provider_data:
                     send_notification(
--- a/ecommerce/extensions/customer/utils.py
+++ b/ecommerce/extensions/customer/utils.py
@@ -25,26 +25,27 @@ class Dispatcher(Dispatcher):
             # pylint: disable=protected-access
             CommunicationEvent._default_manager.create(order=order, event_type=event_type)
 
-    def dispatch_user_messages(self, user, messages, site=None):  # pylint: disable=arguments-differ
+    def dispatch_user_messages(self, user, messages, site=None, recipient=None):  # pylint: disable=arguments-differ
         """
         Send messages to a site user
         """
         if messages['subject'] and (messages['body'] or messages['html']):
-            self.send_user_email_messages(user, messages, site)
+            self.send_user_email_messages(user, messages, site, recipient)
         if messages['sms']:
             self.send_text_message(user, messages['sms'])
 
-    def send_user_email_messages(self, user, messages, site=None):  # pylint: disable=arguments-differ
+    def send_user_email_messages(self, user, messages, site=None, recipient=None):  # pylint: disable=arguments-differ
         """
         Sends message to the registered user / customer and collects data in
         database
         """
-        if not user.email:
+        if not (recipient or user.email):
             msg = "Unable to send email messages: No email address for '{username}'.".format(username=user.username)
             self.logger.warning(msg)
             return
 
-        email = self.send_email_messages(user.email, messages, site)
+        recipient = recipient if recipient else user.email
+        email = self.send_email_messages(recipient, messages, site)
 
         # Is user is signed in, record the event for audit
         if email and user.is_authenticated():
--- a/ecommerce/notifications/notifications.py
+++ b/ecommerce/notifications/notifications.py
@@ -10,7 +10,7 @@ CommunicationEventType = get_model('customer', 'CommunicationEventType')
 Dispatcher = get_class('customer.utils', 'Dispatcher')
 
 
-def send_notification(user, commtype_code, context, site):
+def send_notification(user, commtype_code, context, site, recipient):
     """Send different notification mail to the user based on the triggering event.
 
     Args:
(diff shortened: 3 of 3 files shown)


openedx-webhooks:
Thanks for the pull request, @kaizoku! I've created [OSPR-2627](https://openedx.atlassian.net/browse/OSPR-2627) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

natabene:
Let me know when this is ready for review, @kaizoku .

kaizoku:
@natabene I can fix the migration issues if we're going to use this approach, but we're wondering if this approach is likely to be upstreamed, or if another method would be preferrable.

Currently this code stores the email provided from CyberSource as a `CharField` on the `ecommerce.core.models.User` model, but we're unsure if edX would want us to extend that model, or perhaps create another table elsewhere. Could you or someone at edX advise on our approach and what would be suitable for upstreaming?

natabene:
@marcotuts Can you have a look at this?

UmanShahzad:
@marcotuts @natabene Anyone able to look at this? Note we're mostly trying to see what edX would consider acceptable for what this PR is trying to achieve. If that means the approach needs to change a bit, we can talk about it.

marcotuts:
I'm not clear on the intent here, so I had a few questions - 
1.) Today we send this notification / email through edx-platform and the proposal is to use some details within cybersource to send this instead? Was cybersource a "channel" / email delivery mechanism already or is this introducing that? 

2.) @nasthagiri may be able to chime in here in case I'm misrepresenting this architectural direction but I believe we'd like to shift to having ACE be aware of all notifications from the platform, regardless of which "channel" (aka Sailthru, Amazon SES, Firebase Cloud Messaging, etc) sends the notification. This will let us eventually provide notification settings and controls for learners. How can we help move closer to this edx-ace usage as part of this PR?

robrap:
@marcotuts: I think the issue is that the CyberSource account that someone uses to pay (this applies to Paypal and other providers as well), might be a very different account with a different email address.  It might even be a different user, if a parent or someone else's account is used to pay.

I did not look at this PR deeply, but I imagine it is trying to determine where and when we could send notifications to the payer email address that is returned back to us, rather than the email address of the Open edX account being payed for.  I don't know if there are any legal questions as well about usage of this other email address?

Hope this helps.

UmanShahzad:
@robrap Your interpretation is correct - we have a use case to send an email to the address used on the payment gateway side (in this case, specifically Cybersource, but we'd want to make it generic if possible), not the Open edX side.

cc @kaizoku to expand if necessary.

kaizoku:
That's correct @robrap and @UmanShahzad, this will sent the receipt email to the email address returned from cybersource instead of the one associated to the user's edX account.

Cybersource itself isn't used to deliver messages, from edX, but this makes it a source for the address used to deliver, in case a different address is provided there, than the one from the user's account.

UmanShahzad:
With that said, is there any idea how to approach this problem in a way edX would accept?

natabene:
@edx/arch-review Team, can you have a look at this?

robrap:
@UmanShahzad @kaizoku: Product needs to weigh-in, but this idea would require extra review because of the legal territory with consequences on PII (Personally Identifiable Information) and GDPR.  It also gets into questions around our permissions to use that email.

macdiesel:
@stroilova FYI.

UmanShahzad:
@robrap Makes sense. Just let us know what the result of the legal investigation turns out to be, if it's not too difficult.

natabene:
@stroilova  Do you want to have a look at this?

marcotuts:
@macdiesel - Would you mind giving this a review and suggest next steps in terms of approach? This seems related to student privacy and general PII. Initial review can be conceptual and not a code level review. Thanks!

ecommerce/core/models.py
```diff
@@ -473,6 +473,7 @@ class User(AbstractUser):
     """Custom user model for use with OIDC."""
 
     full_name = models.CharField(_('Full Name'), max_length=255, blank=True, null=True)
+    cybersource_email = models.CharField(_('Cybersource Email'), max_length=255, blank=True, null=True)
```
macdiesel: I would prefer this be named something more vendor neutral. Perhaps purchaser email? Also thinking about this, I'm not sure that this email address should be attached to the user object. 

Couldn't this email address change per transaction (Basket? I'm not very familiar with the ecommerce code base)? It would be better to store this with the transaction I would think.

ecommerce/extensions/customer/utils.py
```diff
@@ -39,12 +39,13 @@ def send_user_email_messages(self, user, messages, site=None):  # pylint: disabl
         Sends message to the registered user / customer and collects data in
         database
         """
-        if not user.email:
+        if not user.email or user.cybersource_email:
```
macdiesel: I would prefer to see the refactor on this pass the email address to this method. The caller of this method should own the logic of which email address to user, not this method.

macdiesel (changes requested):
In light of GDPR my overall suggestion here would be that we should not store this data if we can avoid doing so. In addition I think there is some re-work that should be done here, please see my comments. 

One suggestion I do have is this: If you want to send the email to the cybersource email address, why store it at all? Would it be possible to take the email address cybersource returns, fire the email (perhaps log or fire an event that the email was sent) and then just not save the information? 

I would prefer a route here that would not store this users email address in the system at all if we can avoid it.

natabene:
@kaizoku  Please let us know once you have addressed our comments.

kaizoku:
Thanks @macdiesel, It does make more sense not to store this email.
I've pushed up a WIP new approach which adds a `recipient` argument to the message dispatching methods, and only uses the email returned from cybersource to send off the email, not storing it in the database.

I haven't fully tested this yet, since I wanted to check that this approach was acceptable before making sure it's fully fleshed out. Does this method seem alright?

One current oddity is that the email is extracted from the request in the signal handler, which is not cybersource specific, as it's in ` ecommerce/extensions/checkout/signals.py`.

Instead we could change the `post_checkout` signal to also send an email address to be used for the receipt, and then choose the email in the cybersource views instead. Is that preferrable to checking the `request.POST` from the signal handler?

marcotuts:
@macdiesel, @kaizoku  - I'm going to push this past product review and leave this in your hands to find a path that makes this configurable and eliminates PII concerns. It sounds like we are making progress toward this, nothing else on my side from a product standpoint. If @macdiesel you'd like to bring me back into this conversation for whatever reason lmk. Thanks!

kaizoku:
@macdiesel any chance you've been able to look at this WIP? Just wanted to make sure it's on the right track before I finalize and make sure it's all tested and working well.

macdiesel:
@kaizoku I'm ok with this. I however am unable to merge in to the ecommerce repo. @cpennington can the rev team take over this PR?

cpennington:
Yup, we'll take it.

ecommerce/extensions/checkout/signals.py
```diff
@@ -118,6 +118,8 @@ def send_course_purchase_email(sender, order=None, **kwargs):  # pylint: disable
                     site_configuration=order.site.siteconfiguration
                 )
 
+                recipient = request.POST.get('req_bill_to_email', order.user.email)
```
cpennington: This isn't being used. Is it left over from a previous version of the code?

cpennington (approved):
Except for the currently unused changes in `checkout/signals.py`, this looks good to me.

macdiesel (approved):

ecommerce/extensions/checkout/signals.py
cpennington: It looks like this needs to be passed to `send_notification` below.

cpennington (changes requested):
Reading this PR again more closely, I noticed that currently, the *only* notification being affected is purchase receipt for a *credit* purchase. Was this the intention?

ecommerce/extensions/checkout/signals.py
kaizoku: Ah yep, thanks. This was still WIP so I hadn't tested everything yet.

kaizoku:
@cpennington, right, we're not intending to change notifications for enrollment coupons for instance, if that's what you mean.

I addressed the test issues and this is ready for another review.

cpennington (approved):

openedx-webhooks:
@kaizoku 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/ecommerce&entry.867055334=https://github.com/edx/ecommerce/pull/1914&entry.1484655318=https://github.com/kaizoku&entry.752974735=2018-09-21+09:30&entry.1917517419=2019-04-25+14:35&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1917, opened 2018-09-25
ENT-1254. Fix for vouchers applied to courses with null enrollment_start

georgebabey:


ecommerce/extensions/api/v2/views/vouchers.py
```diff
@@ -125,7 +125,7 @@ def is_course_run_enrollable(course_run):
 
             return (
                 (not end or end > current_time) and
-                (enrollment_start and enrollment_start <= current_time) and
+                (not enrollment_start or enrollment_start <= current_time) and
```
brittneyexline, line 128: if you could update the comment above that describes the logic, that would be nice :-)

brittneyexline (approved):

ecommerce/extensions/api/v2/views/vouchers.py
robrap: Nit: You could use variable names as comments like:
```
enrollment_start_not_set_or_in_past = not enrollment_start or enrollment_start <= current_time
```
Just food for thought from Clean Code.  You don't have to make this change.

PR #1919, opened 2018-09-27
Add partner in Order model.

waheedahmed:
LEARNER-6474


robrap:
@waheedahmed: I have not looked at this in detail, but please use the RequestCache for request caching.  Thanks.

waheedahmed:
@robrap Actually I don't need it, it's already done in SiteMixin.

waheedahmed:

CC: @feanil @robrap

ecommerce/extensions/api/v2/tests/views/test_orders.py
```diff
@@ -156,6 +157,46 @@ def assert_list_with_username_filter(self, user, order):
...
+        )
+
+        self.request.site = site_configuration.site
+        self.client = self.client_class(SERVER_NAME=domain)
+
+        response = self.client.get(self.path, HTTP_AUTHORIZATION=self.token)
+        self.assertEqual(response.status_code, 200)
+        content = json.loads(response.content)
+
+        self.assertEqual(content['count'], 2)
```
waheedahmed: This test is failing without this change.
https://github.com/edx/ecommerce/blob/18d80c696ec08e8d279b46762f60d8051d2aa2e9/ecommerce/extensions/order/utils.py#L98

waheedahmed:
@robrap @feanil Please review.

robrap (commented):
I don't feel equipped to review without discussing with @feanil. 

This is intended to allow us to switch back to the internal url?  I'm not clear on how the internal url will have the correct partner to allow this to work, since the orders could be for various partners, but the internal url would only have one partner.  I'm clearly missing something. :)

ecommerce/extensions/order/migrations/0018_migrate_partner_to_orders.py
```diff
@@ -0,0 +1,35 @@
...
+# Generated by Django 1.11.15 on 2018-09-27 11:13
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def add_partner_to_orders(apps, schema_editor):
+    Order = apps.get_model('order', 'Order')
+
+    for order in Order.objects.all():
```
feanil: There are currently 28703378 orders in the production orders table.  This migration will not work and will probably be detrimental to the database.  What we probably is a management command that back fills this information instead.

waheedahmed: @feanil Will that command back fills data at once or in batches?

waheedahmed:
@feanil I have removed the data migration and will create a new PR with management command but until the order site partner is migrated to orders we cannot fix the order API to work for multiple sites.

This PR is now just adding the partner field in Order model and will create new orders with partner and site both.

CC: @robrap

feanil (approved):

Ayub-Khan (approved):

PR #1920, opened 2018-09-27
LEARNER-6478: Ideas for fixing FulfillFrozenBaskets.

robrap:
LEARNER-6478 is a case of a particular user that could use this to have an order created.  They are already in the verified track, and we should confirm they stay in it if this works.

See the ticket for more details.

### DIFF
--- a/ecommerce/extensions/payment/processors/paypal.py
+++ b/ecommerce/extensions/payment/processors/paypal.py
@@ -278,8 +278,9 @@ class Paypal(BasePaymentProcessor):
             # Raise an exception for payments that were not successfully executed. Consuming code is
             # responsible for handling the exception
             error = self._get_error(payment)
-            # pylint: disable=unsubscriptable-object
-            entry = self.record_processor_response(error, transaction_id=error['debug_id'], basket=basket)
+            # Do not set the transaction_id.  Errors could just be Duplicate or Already Paid, and better to not
+            # confuse them with actual PAY transaction_ids, which may have also come back.
+            entry = self.record_processor_response(error, basket=basket)  # pylint: disable=unsubscriptable-object
 
             logger.warning(
                 "Failed to execute PayPal payment on attempt [%d]. "
--- a/ecommerce/management/utils.py
+++ b/ecommerce/management/utils.py
@@ -67,6 +67,12 @@ def refund_basket_transactions(site, basket_ids):
 
 
 class FulfillFrozenBaskets(EdxOrderPlacementMixin):
+    """
+    TODO: Add Validation for this:
+        1. Check basket is actually frozen.
+        2. Ensure the processorresponse from paypal has state="approved", cybersource has "decision":"ACCEPT".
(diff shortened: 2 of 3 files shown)


ecommerce/extensions/payment/processors/paypal.py
```diff
@@ -278,8 +278,9 @@ def handle_processor_response(self, response, basket=None):
             # Raise an exception for payments that were not successfully executed. Consuming code is
             # responsible for handling the exception
             error = self._get_error(payment)
-            # pylint: disable=unsubscriptable-object
-            entry = self.record_processor_response(error, transaction_id=error['debug_id'], basket=basket)
+            # Do not set the transaction_id.  Errors could just be Duplicate or Already Paid, and better to not
```
robrap, line 281: NOTE: For this specific user, the following records would block this management command from finding a unique transaction id:
```
select id, transaction_id from payment_paymentprocessorresponse where basket_id = 34162004 and response like '%"debug_id"%';
```
It would be great if we could clear them out to try this out.

Ayub-Khan: I think we can use this 
select id, transaction_id from payment_paymentprocessorresponse where basket_id = 34162004 and response like u'state': u'completed';
to check if there is a completed transaction response.

ecommerce/management/utils.py
```diff
@@ -67,6 +67,12 @@ def refund_basket_transactions(site, basket_ids):
...
 
 class FulfillFrozenBaskets(EdxOrderPlacementMixin):
+    """
+    TODO: Add Validation for this:
+        1. Check basket is actually frozen.
+        2. Ensure the processorresponse from paypal has state="approved", cybersource has "decision":"ACCEPT".
+        3. Ensure there is not already an order before trying to create one and the payment info.
+    """
 
     def fulfill_basket(self, basket_id, site):
```
Ayub-Khan, line 77: Those are very rare cases that will involve order failure too. So i think for the usual cases where fulfillment is failed and we have a existing order this code will break. We need to check the order if there is an order we just need to call the fulfillment again.

robrap: Do we have an existing error/run book for fulfillment failing?  What does it say to do?
I agree that this would hopefully be a rare case, but we still need some way to recover and correct the data, right?

Ayub-Khan: @robrap Actually its quite common. Happens every day. I have a splunk alert which creates the zendesk tickets when this happens. Support team resolves those. I created a run book for this 1 year ago for support team to process those issues. They might still have that. that is related to order fulfillment. I dnt think there is a existing run book for basket fulfillment. But this code gives a good idea what steps are there .

robrap: Thanks @Ayub-Khan.  
FYI: @jmyatt - this is the PR we created during the hackathon to capture our thoughts.

robrap:
I'm going to close this ticket and delete the branch.  I think all work on LEARNER-6478 is happening in other PRs.

PR #1921, opened 2018-09-28
Add id endpoint to allow for Segment retirement

bmedx:
When ecommerce cannot find an LMS ID to use as a user identifier for Segment, it uses one in the format of `ecommerce-{ecommerce user id}`. When we retire a user from Segment we need to retire this ID as well, but it is unavailable to other systems currently. This PR creates an endpoint to expose both the ecommerce user ID and the formatted tracking id.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_retirement.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_retirement.py
@@ -1,0 +1,26 @@
+import ddt
+from rest_framework import status
+
+from ecommerce.extensions.analytics.utils import ECOM_TRACKING_ID_FMT
+from ecommerce.extensions.api.v2.views.retirement import EcommerceIdView
+from ecommerce.tests.testcases import TestCase
+
+
+@ddt.ddt
+class EcommerceIdViewTest(TestCase):
+    def test_successful_get(self):
+        user = self.create_user()
+        response = EcommerceIdView().get(None, username=user.username)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertDictEqual(
+            response.data,
+            {
+                'id': user.pk,
+                'ecommerce_tracking_id': ECOM_TRACKING_ID_FMT.format(user.pk)
+            }
+        )
+
+    @ddt.data('does_not_exist', None)
+    def test_unknown_user(self, username):
+        response = EcommerceIdView().get(None, username=username)
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
--- a/ecommerce/extensions/api/v2/views/retirement.py
+++ b/ecommerce/extensions/api/v2/views/retirement.py
@@ -1,0 +1,43 @@
+"""
+Endpoints to facilitate retirement actions
+"""
+
+from edx_rest_framework_extensions.authentication import JwtAuthentication
+from rest_framework import permissions, status
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from ecommerce.core.models import User
+from ecommerce.extensions.analytics.utils import ECOM_TRACKING_ID_FMT
+
+
+class EcommerceIdView(APIView):
+    """
+    Allows synchronization of the ecommerce user id and tracking id with
+    other systems. Specifically this is used to retire users identified
+    by "ecommerce-{id}" from Segment.
+    """
+    authentication_classes = (JwtAuthentication, )
+    permission_classes = (permissions.IsAuthenticated, permissions.IsAdminUser)
+
+    def get(self, _, username):
+        """
+        Returns the ecommerce tracking id of the given LMS user, identified
+        by username.
+        """
+        try:
+            if not username:
+                raise User.DoesNotExist()
+
+            user = User.objects.get(username=username)
+            return Response(
+                {
+                    'id': user.pk,
+                    'ecommerce_tracking_id': ECOM_TRACKING_ID_FMT.format(user.pk)
+                }
+            )
+        except User.DoesNotExist:
+            return Response(
+                status=status.HTTP_404_NOT_FOUND,
+                data={'message': 'Invalid user.'}
+            )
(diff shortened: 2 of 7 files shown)


ecommerce/extensions/api/v2/views/retirement.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class EcommerceIdView(APIView):
+    """
+    Allows synchronization of the ecommerce user id and tracking id with
+    other systems. Specifically this is used to retire users identified
+    by "ecommerce-{id}" from Segment.
+    """
+    authentication_classes = (JwtAuthentication, )
+    permission_classes = (permissions.IsAuthenticated, permissions.IsAdminUser)
```
bmedx, line 21: Reviewer question- these seem to be the permissions used for service API requests coming from  outside the system. Our use case would be for a Tubular script running on Jenkins to authenticate and make this call as part of the retirement process. Currently we do this against LMS with a particular user. Is this the correct set of permissions for that kind of call? Would an existing user in app-permissions be able to SSO against LMS and make this call?

ecommerce/extensions/api/v2/tests/views/test_retirement.py
```diff
@@ -0,0 +1,19 @@
...
+@ddt.ddt
+class EcommerceIdViewTest(TestCase):
+    def test_successful_get(self):
+        user = self.create_user()
+        response = EcommerceIdView().get(None, username=user.username)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertDictEqual(response.data, {'id': user.pk, 'ecommerce_tracking_id': 'ecommerce-{}'.format(user.pk)})
+
+    @ddt.data('does_not_exist', None)
+    def test_unkown_user(self, username):
```
robrap: Typo in `unknown`.

ecommerce/extensions/api/v2/views/retirement.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        try:
+            if not username:
+                raise User.DoesNotExist()
+
+            user = User.objects.get(username=username)
+            return Response(
+                {
+                    'id': user.pk,
+                    'ecommerce_tracking_id': 'ecommerce-{}'.format(user.pk)
```
robrap: Is there a method that exists to convert into a tracking id that could be used here?

ecommerce/extensions/api/v2/views/retirement.py
robrap: If you've got SSO against the LMS, this seems right.

ecommerce/extensions/api/v2/views/retirement.py
```diff
@@ -0,0 +1,42 @@
...
+from ecommerce.core.models import User
+
+
+class EcommerceIdView(APIView):
+    """
+    Allows synchronization of the ecommerce user id and tracking id with
+    other systems. Specifically this is used to retire users identified
+    by "ecommerce-{id}" from Segment.
+    """
+    authentication_classes = (JwtAuthentication, )
```
robrap, line 20: `JWT_AUTHENTICATION` is a default, so this is only needed if you intentionally are trying to remove the other defaults.

robrap (commented):
Thanks @bmedx.

robrap (commented):
Thanks @bmedx.

ecommerce/extensions/api/v2/views/retirement.py
bmedx: That's probably for the best in this case, I can't imagine another use for this call.

bmedx:
@robrap I moved all of the uses of the format string to one constant and fixed the typo. If it looks good I can squash if need be.

ecommerce/extensions/refund/tests/test_signals.py
```diff
@@ -21,7 +22,10 @@ def assert_refund_event_fired(self, mock_track, refund, tracking_context=None):
         (event_user_id, event_name, event_payload), kwargs = mock_track.call_args
 
         self.assertTrue(mock_track.called)
-        self.assertEqual(event_user_id, tracking_context.get('lms_user_id', 'ecommerce-{}'.format(refund.user.id)))
+        self.assertEqual(
+            event_user_id,
+            tracking_context.get('lms_user_id', ECOM_TRACKING_ID_FMT.format(refund.user.id))
```
robrap: On phone, so I may be misreading this, but it looks like we are providing a default. Seems wrong for a test assertion.
I know this wasn’t originally your code, but what do you think?

robrap (approved):
Looks good. Please consider my question, but you aren’t blocked on it. 
Also, did you test on a sandbox?

ecommerce/extensions/refund/tests/test_signals.py
bmedx: This is a test utility method that is called with `tracking_context` as `None` in `test_successful_refund_tracking_without_context`, so the default is necessary to handle that case. Not my favorite pattern, but probably not as bad as what you were thinking.

I've now tested on devstack and a sandbox from the Tubular scripts that would call this and was able to exercise the call successfully. Just squashed and added the url, which I had forgotten to do initially.

robrap: Still seems a bit odd when you actually want to assert it is there.  Could you have:
```
if tracking_context:
    self.assertEqual(...)
```

robrap: Note: I'm not blocking this on this, but it feels like it would be a more accurate test.

bmedx: There were some other usages of `.get` in there, latest commit attempts to be explicit about the assumptions made in that method. Needed to re-kick the PR anyways since statuses weren't reported.

robrap (approved):
Feel free to squash and rebase.  
However, I think we are holding up on releasing during the Masters announcement, and I can't merge until it is up to date with master.  So, feel free to wait until tomorrow and when I can merge and you can deploy.   We try not to let changes sit on master without deploying.
Thanks.

bmedx:
@robrap this is rebased as of this morning and squashed to the 2 commits since they have different purposes.

PR #1922, opened 2018-09-28
Unpin and upgrade requirements.

robrap:


robrap:
@georgebabey: I dropped the DRF related upins/upgrades.  Everything passes so far. :)
We've got e2e tests (ecommerce + WL).  If it passes that, I can just monitor errors in Production.

georgebabey (approved):
Nice! Glad you were able to unpin all those other ones! I'm a little concerned we're leaving DRF pinned but some like DRF-JWT unpinned. That might bite us down the road so we should try to figure out what is needed to get DRF updated in the near future.

robrap:
@georgebabey: I could pin djangorestframework-jwt again.  But, I'm hoping our tests will let us know if we have a problem, and that we can work to unpin more, or at least upgrade DRF.

georgebabey:
@robrap No no, I agree keeping it unpin and letting the tests know when something breaks is the way to go. Just calling out the the fact this is likely to happen with DRF pinned.

PR #1923, opened 2018-09-28
Apply Program Applicator to middleware

BenjiLee:


robrap:
Is the ProgramApplicator really a Program only applicator, or is it more like a Smart Applicator that is smarter about programs when applicable?

PR #1926, opened 2018-10-03
update basket add view error message

tasawernawaz:
[LEARNER-6489](https://openedx.atlassian.net/browse/LEARNER-6489)

Tested at https://ecommerce-tasawernawaz.sandbox.edx.org/basket/add/?sku=%3Cimg%20src=%22https://cdn.pixabay.com/photo/2016/06/18/17/42/image-1465348_960_720.jpg%22%3E

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -10,6 +10,7 @@ import newrelic.agent
 import waffle
 from django.http import HttpResponseBadRequest, HttpResponseRedirect
 from django.shortcuts import redirect, render
+from django.utils.html import escape
 from django.utils.translation import ugettext as _
 from opaque_keys.edx.keys import CourseKey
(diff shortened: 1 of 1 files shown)


tasawernawaz:
@inventhouse please review this.
FYI @edx/learner-spartans

robrap (approved):

BenjiLee:
Rather than escaping the sku, I think we should not return this information at all. It seems like this is meant more for debugging which shouldn't be on production. We should add a log statement in the case of missed SKUs and create an alert in splunk to catch these (possibly with a thresholds to avoid one off alerts).
https://github.com/edx/ecommerce/blob/275f0dff43ea710a638cc51afd5c1ebfc365e625/ecommerce/extensions/basket/views.py#L68


Pros:
* We don't have to worry about escaping anything
* We will have a log history of SKU's that are not working
* We do not expose any product information in our database to users

Cons:
* We will have another alert to potentially calibrate

PR #1928, opened 2018-10-03
Do not render bad SKUs when SKUs are not found

BenjiLee:
Rather than escaping the sku, I think we should not return this information at all. It seems like this is meant more for debugging which shouldn't be on production. We should add a log statement in the case of missed SKUs and create an alert in splunk to catch these (possibly with a thresholds to avoid one off alerts).
https://github.com/edx/ecommerce/blob/275f0dff43ea710a638cc51afd5c1ebfc365e625/ecommerce/extensions/basket/views.py#L68

Pros:
* We don't have to worry about escaping anything
* We will have a log history of SKU's that are not working
* We do not expose any product information in our database to users

Cons:
* We will have another alert to potentially calibrate

### DIFF
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -418,7 +418,8 @@ class BasketCalculateView(generics.GenericAPIView):
 
         products = Product.objects.filter(stockrecords__partner=partner, stockrecords__partner_sku__in=skus)
         if not products:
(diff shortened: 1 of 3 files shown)


robrap:
@BenjiLee: This is another way to implement the safety for the HttpResponseBadRequest, but my concern is that until we add validation for the SKUs, if products are returned for any of the SKUs, won't it continue with the full list of SKUs and other potentially vulnerable code could be run?

PR #1931, opened 2018-10-05
LEARNER-6478: Fix FulfillFrozenBaskets 

Ayub-Khan:
This is to fix the FulfillFrozenBaskets issues and make it such that it could work with the baskets not have been fulfilled(orders created) or order which have not fulfilled (student not in correct corse mode) and correct those issues. 

https://openedx.atlassian.net/browse/LEARNER-6478

### DIFF
--- a/ecommerce/management/utils.py
+++ b/ecommerce/management/utils.py
@@ -1,5 +1,7 @@
+import json
 import logging
 
+from django.db.models import Q
 from oscar.apps.partner import strategy
 from oscar.core.loading import get_class, get_model
 
@@ -12,8 +14,13 @@ logger = logging.getLogger(__name__)
 
 Applicator = get_class('offer.applicator', 'Applicator')
 Basket = get_model('basket', 'Basket')
+EventHandler = get_class('order.processing', 'EventHandler')
 NoShippingRequired = get_class('shipping.methods', 'NoShippingRequired')
+Order = get_model('order', 'Order')
 OrderTotalCalculator = get_class('checkout.calculators', 'OrderTotalCalculator')
+ShippingEventType = get_model('order', 'ShippingEventType')
+
+SHIPPING_EVENT_NAME = 'Shipped'
 
 _payment_processors = {}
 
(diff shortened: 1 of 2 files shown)


robrap:
@Ayub-Khan: I added more notes on the ticket about potential approach for this issue.  Thank you for your work on this.

robrap:
@jmyatt: Can you review this?

ecommerce/management/utils.py
```diff
@@ -70,67 +76,103 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
...
+        # Start order fulfillment if order exists but is not fulfilled.
+        elif order.is_fulfillable:
+            order_lines = order.lines.all()
+            line_quantities = [line.quantity for line in order_lines]
+
+            shipping_event, __ = ShippingEventType.objects.get_or_create(name=SHIPPING_EVENT_NAME)
+            EventHandler().handle_shipping_event(order, shipping_event, order_lines, line_quantities)
+
+            if order.is_fulfillable:
+                return False
```
jmyatt, line 184: why not log a warning here, noting the order id that we attempted to fulfill?

ecommerce/management/utils.py
```diff
@@ -70,67 +76,103 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
...
+            # run the fulfillment to update course mode.
+            try:
+                Applicator().apply(basket, user=basket.owner)
+            except ValueError:
+                basket.clear_vouchers()
+
+            # Filter the successful payment processor response which in case
+            # of Cybersource includes "u'decision': u'ACCEPT'" and in case of
+            # Paypal includes "u'state': u'approved'".
+            successful_transaction = basket.paymentprocessorresponse_set.filer(
```
jmyatt: filer --> filter

ecommerce/management/utils.py
```diff
@@ -70,67 +76,103 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
...
+        if not order:
+            basket.strategy = strategy.Default()
+
+            # Need to handle the case that applied voucher has been expired.
+            # This will create the order  with out discount but subsequently
+            # run the fulfillment to update course mode.
+            try:
+                Applicator().apply(basket, user=basket.owner)
+            except ValueError:
+                basket.clear_vouchers()
```
jmyatt, line 110: so will this result in the payment amount and order amount being out of sync?  I agree that is the best way to handle the issue, so that the order still goes through, just want to understand the expected outcome.  Also the prod-edx-verify_transactions mgmt command will flag these so we may need to manually ignore those instances after running this job

ecommerce/management/utils.py
```diff
@@ -70,67 +76,103 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
...
+            # Paypal includes "u'state': u'approved'".
+            successful_transaction = basket.paymentprocessorresponse_set.filer(
+                Q(response__contains="u'decision': u'ACCEPT'")
+                |
+                Q(response__contains="u'state': u'approved'")
+            )
+
+            # In case of no successful transactions log and stop the process.
+            if not successful_transaction:
+                logger.info('Basket %d do not have any successful payment response', basket_id)
```
jmyatt: "do" --> "does"

ecommerce/management/utils.py
```diff
@@ -70,67 +76,103 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
...
+                return False
+
+            successful_payment_notification = successful_transaction[0]
```
jmyatt, line 128: I think continuing with the fulfillment in the case where there are multiple successful transactions logged is the right thing to do (as opposed to short-circuiting as the prior code did).  I'm just not sure about arbitrarily selecting the first transaction.  Maybe leave this as is (selecting the 0-index transaction), but log a warning if we have more than one transaction id among the successful transactions found?  Borrowing from the prior code, this would be something like:

```
unique_transaction_ids = set([response.transaction_id for response in successful_transaction])
if len(unique_transaction_ids) > 1:
    logger.warning('Basket %d has more than one successful transaction id, using the first one', basket_id)
```

ecommerce/management/utils.py
```diff
@@ -70,67 +76,103 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
...
+            self.payment_processor = _get_payment_processor(successful_payment_notification.processor_name)
+
+            # Create handled response
+            handled_response = HandledProcessorResponse(
+                transaction_id=successful_payment_notification.transaction_id,
+                total=basket.total_excl_tax,
+                currency=basket.currency,
+                card_number=successful_payment_notification.response.get('req_card_number', None),
+                card_type=successful_payment_notification.response.get('req_card_type', 'Paypal Account'),
```
jmyatt: for Cybersource, I think the value we want is not the raw `req_card_type` but rather `CYBERSOURCE_CARD_TYPE_MAP.get(req_card_type)`

ecommerce/management/utils.py
```diff
@@ -70,67 +76,103 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
...
+                    order_number=order_number,
+                    user=user,
+                    basket=basket,
+                    shipping_address=None,
+                    shipping_method=shipping_method,
+                    shipping_charge=shipping_charge,
+                    billing_address=None,
+                    order_total=order_total,
+                )
+                logger.info('Successfully create order for basket %d'. basket.id)
```
jmyatt: 'create' --> 'created'

jmyatt (changes requested):

ecommerce/management/utils.py
Ayub-Khan: Yes. Payment amount and order amount will be out of sync.

ecommerce/management/utils.py
Ayub-Khan: done

ecommerce/management/utils.py
Ayub-Khan: done

ecommerce/management/utils.py
Ayub-Khan: done

ecommerce/management/utils.py
Ayub-Khan: @jmyatt above line runs the following code which have extensive logging 
https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/fulfillment/api.py#L21
thats why i did not log any thing here. should i add a log here too.

ecommerce/management/utils.py
Ayub-Khan: done

ecommerce/management/utils.py
Ayub-Khan: done

ecommerce/management/utils.py
jmyatt: ah right - no, you're right, the existing logging should be enough

ecommerce/management/utils.py
jmyatt: ok makes sense

Ayub-Khan:
kindly ignore the i18n changes i will remove them.

ecommerce/management/utils.py
```diff
@@ -95,7 +95,7 @@ def fulfill_basket(self, basket_id, site):
             if order:
                 logger.info('Basket %d does have a existing order %s', basket.id, basket.order_number)
         except Order.DoesNotExist:
-            pass
+            order = None
```
jmyatt, line 98: ah good catch

ecommerce/management/utils.py
jmyatt: I think you still want to de-dup these on `response.transaction_id` - that's actually the case we want to warn about.

ecommerce/management/utils.py
```diff
@@ -70,67 +76,106 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
...
+
+            # Create handled response
+            handled_response = HandledProcessorResponse(
+                transaction_id=successful_payment_notification.transaction_id,
+                total=basket.total_excl_tax,
+                currency=basket.currency,
+                card_number=successful_payment_notification.response.get('req_card_number', None),
+                card_type=CYBERSOURCE_CARD_TYPE_MAP.get(
+                    successful_payment_notification.response.get('req_card_type', 'Paypal Account'),
+                    'Paypal Account')
```
jmyatt: looking at this again, it seems you have the Paypal values for card_number and card_type swapped - the existing code uses 'Paypal Account' for card_number and None for card_type.  So I think these lines should be:

```
    card_number=successful_payment_notification.response.get('req_card_number', 'Paypal Account'),
    card_type=CYBERSOURCE_CARD_TYPE_MAP.get(
        successful_payment_notification.response.get('req_card_type'), None)
```

jmyatt (approved):

PR #1925, opened 2018-10-01
ENT-1248/ENT-1266 Migrate enterprise coupons to use enterprise conditional offers

brittneyexline:
**Description**
The rationale of this set of changes for (ENT-1248) is to consolidate enterprise discounting behavior so that there are fewer code paths to follow for any enterprise related discount code, and to make it easier to add the user assignment functionality that we plan to add as part of our enterprise code management tool. This PR is the first step in this refactoring.
The goal of this PR is to create enterprise conditional offers for each voucher that is specific to an enterprise customer. There is a migration that creates the new conditional offer based on the original one for the voucher and attaches it to the same voucher. The same logic has been added to the coupon creation flow so that any enterprise coupons created or updated in the meantime have the appropriate data.
Additionally, there are some modifications to ensure that we are only applying the original conditional offer (not the new enterprise offer). In the next PR, we will change the implementation of enterprise coupons to use only enterprise conditional offers when being created or redeemed.
There should be no change in product behavior - we are merely creating the data we will need to modify behavior in the future.

**Test Plan**
This is currently deployed on the business sandbox.
Run the management command - it should succeed and create enterprise conditional offers in the database attached to vouchers for enterprise coupons only. The offers, conditions, and benefits created should have equivalent data but not have a range object.
For example:
```
-- find a voucher in the database for an enterprise customer
mysql> select * from voucher_voucher limit 1;
+----+---------------------+------------------+-----------+----------------------------+----------------------------+----------------------+------------+----------------+----------------------------+
| id | name                | code             | usage     | start_datetime             | end_datetime               | num_basket_additions | num_orders | total_discount | date_created               |
+----+---------------------+------------------+-----------+----------------------------+----------------------------+----------------------+------------+----------------+----------------------------+
|  1 | Pied Piper Discount | RNBL737ZAUJXUM6E | Multi-use | 2017-06-01 00:00:00.000000 | 2025-08-31 00:00:00.000000 |                    0 |         68 |         682.00 | 2017-06-08 00:00:00.000000 |
+----+---------------------+------------------+-----------+----------------------------+----------------------------+----------------------+------------+----------------+----------------------------+
1 row in set (0.00 sec)

-- find the conditional offers attached to the voucher
mysql> select * from voucher_voucher_offers where voucher_id = 1;
+------+------------+---------------------+
| id   | voucher_id | conditionaloffer_id |
+------+------------+---------------------+
|    1 |          1 |                   1 |
| 1817 |          1 |                 499 |
+------+------------+---------------------+
[...]

PR #1936, opened 2018-10-16
ENT-1248/ENT-1267 Introduce waffle flag for using enterprise conditions

brittneyexline:
**Description**
This PR depends on https://github.com/edx/ecommerce/pull/1925
Now that all coupons associated with an enterprise have both regular offer and enterprise offer data available, we need to introduce the ability to use the enterprise offer data when redeeming enterprise coupon vouchers.
This PR introduces a waffle switch that controls which offer to use for redemption logic when a voucher contains enterprise offer data. If the switch is off, the system will behave as it has, using the existing offer and ignoring any enterprise offers. If the switch is on, the system will use the enterprise conditional offer if it is available.
We should flip this switch when we have migrated the enterprise coupon data to be fully supported for this feature set. In particular, we will start to enforce that enterprise catalog is required for enterprise coupons.

**Test Plan**
This is currently deployed on the business sandbox, and the migration introducing the waffle switch has already been run.

With the waffle switch off:

- Test that non-enterprise coupons can be redeemed as expected. (https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=10OFF)
- Test that enterprise offers (applied for enterprise customers with SSO integration) are applied as expected. (https://business.sandbox.edx.org/enterprise/d583b097-7bd9-4a7e-a98d-c15dea220354/course/course-v1:NYIF+CM2.x+3T2017/enroll/?catalog=e1a27c18-b612-4c67-a5f5-9ad0cba993d5&utm_medium=enterprise&utm_source=successfactors)
- Test that the appropriate courses are displayed on the offers page for an enterprise coupon (https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=PIPNJSUK33P7PTZH)
, regardless of if an enterprise catalog is configured. If one isn't configured (https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=RNBL737ZAUJXUM6E), the catalog query or catalog id option should be used to populate the list of offers.
- Test that enterprise coupons can be redeemed as expected through the coupon redemption link, and that the offer applications are being made against the regular conditional offer.
- Test that enterprise coupons can be redeemed as expected by applying a code on the basket page, and that the offer applications are being made against the regular conditional offer.

With the waffle switch on:

- Test that non-enterprise coupons can be redeemed as expected.
- Test that enterprise offers are applied as expected.
- Test that the appropriate courses are displayed on the offers page for an enterprise coupon, only if an enterprise catalog is configured. If one isn't configured, there should be no available courses for the coupon.
- Test that enterprise coupons can be redeemed as expected through the coupon redemption link, and that the offer applications are being made against the enterprise conditional offer.
- Test that enterprise coupons can be redeemed as expected by applying a code on the basket page, and that the offer applications are being made against the enterprise conditional offer.

### DIFF
--- a/ecommerce/enterprise/conditions.py
+++ b/ecommerce/enterprise/conditions.py
@@ -3,12 +3,13 @@ from __future__ import unicode_literals
 import logging
 from uuid import UUID
 
+import waffle
 from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import SlumberHttpBaseException
 
 from ecommerce.enterprise.api import catalog_contains_course_runs, fetch_enterprise_learner_data
-from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_SWITCH
+from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH, ENTERPRISE_OFFERS_SWITCH
 from ecommerce.extensions.basket.utils import ENTERPRISE_CATALOG_ATTRIBUTE_TYPE
 from ecommerce.extensions.offer.decorators import check_condition_applicability
 from ecommerce.extensions.offer.mixins import ConditionWithoutRangeMixin, SingleItemConsumptionConditionMixin
@@ -53,10 +54,12 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
             # An anonymous user is never linked to any EnterpriseCustomer.
             return False
 
-        if offer.offer_type == ConditionalOffer.VOUCHER:
+        if (offer.offer_type == ConditionalOffer.VOUCHER and
+                not waffle.switch_is_active(ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH)):
             logger.info('Skipping Voucher type enterprise conditional offer until we are ready to support it.')
             return False
 
+        learner_data = {}
         try:
             learner_data = fetch_enterprise_learner_data(basket.site, basket.owner)['results'][0]
         except (ConnectionError, KeyError, SlumberHttpBaseException, Timeout):
@@ -67,11 +70,14 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
             )
             return False
         except IndexError:
-            return False
+            if offer.offer_type == ConditionalOffer.SITE:
+                logger.info('No learner data returned for user %s', basket.owner)
+                return False
 
-        enterprise_customer = learner_data['enterprise_customer']
-        if str(self.enterprise_customer_uuid) != enterprise_customer['uuid']:
+        if (learner_data and 'enterprise_customer' in learner_data and
+                str(self.enterprise_customer_uuid) != learner_data['enterprise_customer']['uuid']):
             # Learner is not linked to the EnterpriseCustomer associated with this condition.
+            logger.info('Learner\'s enterprise does not match this conditions\'s enterprise.')
             return False
 
         course_run_ids = []
@@ -79,6 +85,7 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
             course = line.product.course
             if not course:
                 # Basket contains products not related to a course_run.
+                logger.info('Basket contains products not related to a course_run.')
                 return False
 
             course_run_ids.append(course.id)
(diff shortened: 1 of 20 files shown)


ecommerce/enterprise/conditions.py
```diff
@@ -67,18 +70,22 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
             )
             return False
         except IndexError:
-            return False
+            if offer.offer_type == ConditionalOffer.SITE:
```
christopappas, line 73: Why do we only return here if the offer_type is `ConditionalOffer.SITE`?

ecommerce/enterprise/conditions.py
```diff
@@ -89,12 +96,14 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
...
         if catalog:
             if offer.condition.enterprise_customer_catalog_uuid != catalog:
+                logger.info('Enterprise catalog id on the basket does not match the catalog for this condition.')
                 return False
 
         if not catalog_contains_course_runs(basket.site, course_run_ids, self.enterprise_customer_uuid,
                                             enterprise_customer_catalog_uuid=self.enterprise_customer_catalog_uuid):
             # Basket contains course runs that do not exist in the EnterpriseCustomerCatalogs
             # associated with the EnterpriseCustomer.
+            logger.info('Enterprise catalog does not contain the course(s) in this basket.')
```
christopappas, line 106: +1 to all these log entries

ecommerce/extensions/voucher/models.py
```diff
@@ -62,9 +65,27 @@ def does_exist(cls, code):
...
+        if not waffle.switch_is_active(ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH):
+            return self.original_offer
+        # If the switch is enabled, return the enterprise offer if it exists.
+        try:
+            return self.offers.get(condition__enterprise_customer_uuid__isnull=False)
+        except ObjectDoesNotExist:
+            # If no enterprise offer is found, return the first available offer.
+            return self.original_offer
+        except MultipleObjectsReturned:
+            logger.exception('There is more than one enterprise offer associated with voucher %s!', self.id)
```
christopappas, line 87: Does this reraise the error or does it actually return `self.original_offer` and keep on going?

christopappas (approved):
Even having looked at this stuff with you, this is super dense. I had a few questions about small stuff, but nothing major.

Also, I like that you outlined the steps you are going to take to test this change 👍

ecommerce/enterprise/conditions.py
```diff
@@ -67,18 +70,22 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
...
+            if offer.offer_type == ConditionalOffer.SITE:
+                logger.info('No learner data returned for user %s', basket.owner)
+                return False
 
-        enterprise_customer = learner_data['enterprise_customer']
-        if str(self.enterprise_customer_uuid) != enterprise_customer['uuid']:
+        if (learner_data and 'enterprise_customer' in learner_data and
+                str(self.enterprise_customer_uuid) != learner_data['enterprise_customer']['uuid']):
             # Learner is not linked to the EnterpriseCustomer associated with this condition.
+            logger.info('Learner\'s enterprise does not match this conditions\'s enterprise.')
```
georgebabey, line 80: might want to bump that up to a warning. I assume we shouldn't get here?

ecommerce/enterprise/conditions.py
```diff
@@ -89,12 +96,14 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
         catalog = self._get_enterprise_catalog_uuid_from_basket(basket)
         if catalog:
             if offer.condition.enterprise_customer_catalog_uuid != catalog:
+                logger.info('Enterprise catalog id on the basket does not match the catalog for this condition.')
```
georgebabey, line 99: Should this also be a warning?

ecommerce/enterprise/conditions.py
brittneyexline: because users redeeming enterprise coupons (type VOUCHER) for the first time will not have enterprise learner data created until after they redeem the coupon.

ecommerce/extensions/voucher/models.py
brittneyexline: it only logs the exception, so it will keep going.

brittneyexline:
@georgebabey @christopappas i am going to close this PR and open a new one in order to get the coverage checks to run, just fyi

PR #1938, opened 2018-10-17
Learner 6478 basket api

Ayub-Khan:
Added basket api.
Added tests.

Tested Locally.

basket-api

Review @robrap @waheedahmed

### DIFF
--- a/ecommerce/extensions/api/v2/urls.py
+++ b/ecommerce/extensions/api/v2/urls.py
@@ -100,6 +100,7 @@ urlpatterns = [
 ]
 
 router = ExtendedSimpleRouter()
+router.register(r'basket-details', basket_views.BasketViewSet, base_name='basket')
 router.register(r'catalogs', catalog_views.CatalogViewSet, base_name='catalog') \
     .register(r'products', product_views.ProductViewSet, base_name='catalog-product',
               parents_query_lookups=['stockrecords__catalogs'])
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -8,6 +8,7 @@ import waffle
 from dateutil.parser import parse
 from django.contrib.auth import get_user_model
 from django.db import transaction
+from django.db.models import Q
 from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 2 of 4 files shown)


ecommerce/extensions/api/v2/urls.py
```diff
@@ -78,7 +78,7 @@
 ]
 
 urlpatterns = [
-    url(r'^baskets/', include(BASKET_URLS, namespace='baskets')),
+    url(r'^basket-edit/', include(BASKET_URLS, namespace='baskets')),
```
Ayub-Khan: @robrap Im not sure if this is the right thing to do. Do we use those api views.

robrap:
@Ayub-Khan: Ultimately, can you provide more details/screenshots on how you expect Customer Support to use this?  
Also, let us know when this is ready and we'll find a reviewer.  Thanks.

Ayub-Khan:
@robrap  The idea is to give this api to support team so that they could check what went wrong against a basket. In case if we are not able to create an order for a basket we will redirect the alert to support team. They will use this to check if we received payment for the order other related information to upate the student in right track and use the management view to create the order for user.

robrap:
@jmyatt: This PR is a continuation of the work for [LEARNER-6478](https://openedx.atlassian.net/browse/LEARNER-6478).

ecommerce/extensions/api/v2/urls.py
```diff
@@ -100,6 +100,7 @@
 ]
 
 router = ExtendedSimpleRouter()
+router.register(r'basket-details', basket_views.BasketViewSet, base_name='basket')
```
waheedahmed, line 103: This API endpoint suggests that we are viewing a single basket details but I believe it's not the case, right? I believe we should update it to more precise name, `basket-list` maybe?

waheedahmed (changes requested):

ecommerce/extensions/api/v2/urls.py
Ayub-Khan: yes. i agree as this shows basket list it would be appropriate. But as the basket includes some other details. and we want to use it for single baskets. basket-details will be more appropriate.

waheedahmed (approved):

PR #1941, opened 2018-10-22
Fix for LEARNER-6529 Can't edit coupons with fixed prices

raeeschachar:
https://openedx.atlassian.net/browse/LEARNER-6529

PR #1945, opened 2018-10-24
Remove broken selenium tests

jmyatt:
Remove the 8 tests that rely on WebDriver due to a sudden failure we seem to have inherited from the Firefox implementation of WebDriver.  Details on what these tests were doing, in case we want to test these things manually, are captured in the following ticket:
https://openedx.atlassian.net/projects/EG/issues/EG-38

### DIFF
--- a/ecommerce/extensions/dashboard/refunds/tests/test_acceptance.py
+++ b/ecommerce/extensions/dashboard/refunds/tests/test_acceptance.py
@@ -12,7 +12,6 @@ from selenium.webdriver.support.wait import WebDriverWait
 
 from ecommerce.extensions.refund.status import REFUND
 from ecommerce.extensions.refund.tests.mixins import RefundTestMixin
-from ecommerce.tests.testcases import LiveServerTestCase
 
 Refund = get_model('refund', 'Refund')
 
@@ -211,19 +210,3 @@ class RefundAcceptanceTestMixin(RefundTestMixin):
                 self.selenium.find_element_by_css_selector,
                 self.deny_button_selector
             )
-
-
-class RefundListViewTests(RefundAcceptanceTestMixin, LiveServerTestCase):
-    """Acceptance tests of the refund list view."""
-
-    def setUp(self):
-        super(RefundListViewTests, self).setUp()
-        self.path = reverse('dashboard:refunds:list')
-
-
-class RefundDetailViewTests(RefundAcceptanceTestMixin, LiveServerTestCase):
-    """Acceptance tests of the refund detail view."""
-
-    def setUp(self):
-        super(RefundDetailViewTests, self).setUp()
-        self.path = reverse('dashboard:refunds:detail', args=[self.refund.id])
--- a/ecommerce/extensions/dashboard/orders/tests.py
+++ b/ecommerce/extensions/dashboard/orders/tests.py
@@ -1,15 +1,10 @@
 import os
-from unittest import skipIf
 
 from django.contrib.messages import constants as MSG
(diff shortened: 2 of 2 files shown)


robrap:
Is there an associated ticket? What are the implications of this? Is it temporary? Why not just delete them like in the Flaky Yest Process?

jmyatt:
@robrap yes I think we need to follow the Flaky Test process here to avoid these broken tests blocking deployment of time sensitive work, since a fix for this will take some digging

ecommerce/extensions/dashboard/refunds/tests/test_acceptance.py
```diff
@@ -1,5 +1,5 @@
 import os
-from unittest import skip
+from unittest import skip, skipIf
```
robrap: Do you need to remove this skipIf?

robrap (changes requested):
If I understand correctly, you are removing some older skipIf calls + the 4 new failing tests?

jmyatt:
@robrap this commit removed the 4 test classes that inherit the failing call to WebDriver() - those are in the base classes `RefundAcceptanceTestMixin` and `OrderViewBrowserTestBase`.  not sure this change is the right thing to do, just wanted to confirm that it would avoid the travis failure.  Will remove the unused import to avoid the pylint failure now, just to get this to green but I'm thinking it might also be worth trying a version upgrade on selenium or pytest-selenium.  But that's just a guess as I don't see any info on this in the wild

robrap:
@jmyatt: This also removes `test_retry_fulfillment` which was never really running anyway, right?
1. I think it is ok to remove these while we figure out the real fix, if we first learn what these tests are doing, whether or not we have e2e coverage, and/or what the steps are for someone to manually test these.
2. Ultimately, the real fix is probably to get these working, and if that is faster than #1, we can do that instead.  But, I'm guessing not.

jmyatt:
@robrap by my count, this removes:
6 Order tests, 2 of which were being skipped
8 Refund tests, 4 of which were being skipped

robrap:
Are 2 of these unnecessarily getting removed?
In this PR: https://github.com/edx/ecommerce/pull/1941
In Travis: https://travis-ci.org/edx/ecommerce/jobs/446169438
I see: "FAILED (SKIP=2, errors=4)"

jmyatt:
I think that error count is inaccurate (it's counting classes as opposed to tests) because there's a class level block that's failing for 4 different classes - 2 each extending from the 2 problem base classes.  I also don't understand the skip count of 2 - there are 2 affected tests with the `skip` decorator and 2 with `skipIf` (travis)

robrap (approved):
This is ready to merge as soon as we have recommendations for manual testing required.

I can merge with the codecov issue if you can't merge.

jmyatt:
Codecov is upset for a non browser test related loss of coverage.  I'll try to rework these tests so that we leave the non browser related tests alone.  Then I'll try to assess the impact of removing these tests as you described @robrap

jmyatt:
@robrap this passed tests and checks - I just re-pushed in order to squash.  Updating the Jira ticket now with manual testing steps in the absence of these tests

jmyatt:
@robrap let's go with this approach.  TestEng has not seen this issue but has supplied some ideas on digging deeper to try to fix it.  I added those notes plus detail on how we can manually test these aspects (if we think that's necessary) in the Jira ticket.  I also attempted switching to chrome, safari, and ie webdrivers, all of which had their own OS related complications (probably why we were using Firefox in the first place).

PR #1946, opened 2018-10-25
Run browser tests on chrome instead of firefox

jmyatt:


jmyatt:
Closing this - non-FF implementations of WebDriver incur OS specific dependencies so swapping this out does not provide the quick fix I was looking for

PR #1947, opened 2018-10-26
LEARNER-6578 Removed useless code

Ayub-Khan:
Removed code we are not using any more.
From looking into templates, it seems, we do not use this code. This will not fix the actual issue, we are facing, but this will help us debug it better !


Ayub-Khan:
blocking this in favor of https://github.com/edx/ecommerce/pull/1951

PR #1952, opened 2018-10-30
ENT-1248/ENT-1267 Introduce waffle flag for using enterprise conditions

brittneyexline:
**Description**
This PR depends on https://github.com/edx/ecommerce/pull/1925
Now that all coupons associated with an enterprise have both regular offer and enterprise offer data available, we need to introduce the ability to use the enterprise offer data when redeeming enterprise coupon vouchers.
This PR introduces a waffle switch that controls which offer to use for redemption logic when a voucher contains enterprise offer data. If the switch is off, the system will behave as it has, using the existing offer and ignoring any enterprise offers. If the switch is on, the system will use the enterprise conditional offer if it is available.
We should flip this switch when we have migrated the enterprise coupon data to be fully supported for this feature set. In particular, we will start to enforce that enterprise catalog is required for enterprise coupons.

**Test Plan**
This is currently deployed on the business sandbox, and the migration introducing the waffle switch has already been run.

With the waffle switch off:

- Test that non-enterprise coupons can be redeemed as expected. (https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=10OFF)
- Test that enterprise offers (applied for enterprise customers with SSO integration) are applied as expected. (https://business.sandbox.edx.org/enterprise/d583b097-7bd9-4a7e-a98d-c15dea220354/course/course-v1:NYIF+CM2.x+3T2017/enroll/?catalog=e1a27c18-b612-4c67-a5f5-9ad0cba993d5&utm_medium=enterprise&utm_source=successfactors)
- Test that the appropriate courses are displayed on the offers page for an enterprise coupon (https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=PIPNJSUK33P7PTZH)
, regardless of if an enterprise catalog is configured. If one isn't configured (https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=RNBL737ZAUJXUM6E), the catalog query or catalog id option should be used to populate the list of offers.
- Test that enterprise coupons can be redeemed as expected through the coupon redemption link, and that the offer applications are being made against the regular conditional offer.
- Test that enterprise coupons can be redeemed as expected by applying a code on the basket page, and that the offer applications are being made against the regular conditional offer.

With the waffle switch on:

- Test that non-enterprise coupons can be redeemed as expected.
- Test that enterprise offers are applied as expected.
- Test that the appropriate courses are displayed on the offers page for an enterprise coupon, only if an enterprise catalog is configured. If one isn't configured, there should be no available courses for the coupon.
- Test that enterprise coupons can be redeemed as expected through the coupon redemption link, and that the offer applications are being made against the enterprise conditional offer.
- Test that enterprise coupons can be redeemed as expected by applying a code on the basket page, and that the offer applications are being made against the enterprise conditional offer.

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -60,7 +60,12 @@ def is_enrollment_code(obj):
 
 def retrieve_benefit(obj):
     """Helper method to retrieve the benefit from voucher. """
-    return retrieve_voucher(obj).benefit
+    return retrieve_offer(obj).benefit
+
+
+def retrieve_condition(obj):
+    """Helper method to retrieve the benefit from voucher. """
+    return retrieve_offer(obj).condition
 
 
 def retrieve_end_date(obj):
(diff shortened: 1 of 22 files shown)


ecommerce/extensions/api/serializers.py
```diff
@@ -60,7 +60,12 @@ def is_enrollment_code(obj):
 
 def retrieve_benefit(obj):
     """Helper method to retrieve the benefit from voucher. """
-    return retrieve_voucher(obj).benefit
+    return retrieve_offer(obj).benefit
```
georgebabey, line 63: Will this impact any existing flows that are expecting retrieve_benefits to return the first voucher's benefit as retrieve_voucher was doing?

georgebabey (approved):
This looks great! Just had one small question on changing the existing retrieve_benefit method

ecommerce/extensions/api/serializers.py
brittneyexline: as far as i can tell, this is only used in the coupon serializer for the coupon admin screen. the way this was implemented was wrong/broken, since benefits are not linked to vouchers.

PR #1956, opened 2018-11-01
LEARNER-6578 Added logs around program applicator

Ayub-Khan:
Added logs around program applicator


waheedahmed (approved):

robrap (commented):
Note: when adding test logging I often wrap it in a waffle switch in case the logging is very noisy. It let’s us shut it off quickly.

PR #1960, opened 2018-11-02
LEARNER-6578 Removed logs

Ayub-Khan:
Removed logs we no longer need.


ecommerce/extensions/basket/views.py
```diff
@@ -305,11 +304,6 @@ def _process_basket_lines(self, lines):
             except IndexError:
                 logger.exception('Basket {%d} line{%d} has discount value {%s} ', self.request.basket.id, line.id,
                                  line.discount_value)
-                Applicator().apply(self.request.basket, self.request.user, self.request)
```
Ayub-Khan, line 308: We no longer need to call applicator again too. Issue was not with the applicator.

waheedahmed (approved):

PR #1962, opened 2018-11-02
Fix orders api to use partner for filter.

waheedahmed:
We have successfully added/migrated partner into orders, now we can use partner to filter the results instead of site, so that the endpoint can work for multiple sites against same partner.

LEARNER-6474


waheedahmed:
@feanil @edx/learner-spartans Please review.

CC: @robrap

ahsan-ul-haq (approved):

Ayub-Khan (approved):

PR #1963, opened 2018-11-02
LEARNER-6578 Fix

Ayub-Khan:
Program condition is_satisfied will only return true if program is Active.
If courses have multiple programs. Applicators get_offers returns offers for retired programs too. We do not update program_offers to expire. Because of which retired program offers are applied to baskets. In some cases when pcs change those offers to 0 discount. We get in a situation where after first retired program offer we have applied offer with 0 discount but actual offer was skipped. 
ie. this was happening for 
"uuid": "688dbadb-7a4c-4311-b5f3-27948bc2c41b", Expired Program
https://ecommerce.edx.org/programs/offers/19493/edit/

"uuid": "a11c408f-0986-4393-8268-8bc16500cdf3", Active Program 
https://ecommerce.edx.org/programs/offers/20270/edit/

i had to set start date and end data in expired program offer to fix the issue for active program.

### DIFF
--- a/ecommerce/programs/conditions.py
+++ b/ecommerce/programs/conditions.py
@@ -118,7 +118,7 @@ class ProgramCourseRunSeatsCondition(SingleItemConsumptionConditionMixin, Condit
         except (HttpNotFoundError, SlumberBaseException, Timeout):
             return False
 
-        if program:
+        if program and program['status'] == 'active':
             applicable_seat_types = program['applicable_seat_types']
         else:
             return False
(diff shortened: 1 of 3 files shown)


robrap (approved):
Looks good. As a potential follow-up, it seems we are still calculating potential program matches, rather than just using the bundle id to declare the program. Do you know where the code lives that determined that multiple programs apply? I’m hoping we can save processing by not trying to figure this out, and thus never having multiple programs.
FYI: @BenjiLee

Ayub-Khan:
@robrap https://github.com/django-oscar/django-oscar/blob/1.5.1/src/oscar/apps/offer/applicator.py#L63
We have a cutom applicator (https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/offer/applicator.py#L16) after looking a bit at code this would have saved us from that issue. This is under waffle flag which is not active yet.

robrap:
@BenjiLee Why isn't the flag active?

Ayub-Khan:
@robrap @BenjiLee Im going to merge this as i think Custom Program Applicator still needs more work.

BenjiLee:
@robrap We turned the flag off because there was another issue present at the time we turned the flag on. It was unrelated to the flag but we turned it off so that others could rule out the program applicator as the cause.

PR #1965, opened 2018-11-02
ARCH-274: Enhance jwt_decode_handler to handle transition.

robrap:
As part of ARCH-261, the ecommerce service and ecom-workers will be
transitioning to use the LMS as the sole issuer of JWT tokens. This
update temporarily make the jwt_decode_handler work with several
steps of the final rollout. Ultimately, once the jwt_decode_handler
is only being used to delegate the the edx-drf-extensions version of
the jwt_decode_handler, this custom one will be retired.  See ARCH-261
for the detailed plan.

ARCH-274

### DIFF
--- a/ecommerce/extensions/api/handlers.py
+++ b/ecommerce/extensions/api/handlers.py
@@ -2,25 +2,66 @@
 import logging
 
 import jwt
+import waffle
 from django.conf import settings
+from edx_django_utils import monitoring as monitoring_utils
+from edx_rest_framework_extensions.auth.jwt.decoder import jwt_decode_handler as edx_drf_extensions_jwt_decode_handler
 from rest_framework_jwt.settings import api_settings
 
 logger = logging.getLogger(__name__)
 
+JWT_DECODE_HANDLER_METRIC_KEY = 'ecom_jwt_decode_handler'
 
-def jwt_decode_handler(token):
-    """Attempt to decode the given token with each of the configured JWT issuers.
 
-    Args:
-        token (str): The JWT to decode.
+def _ecommerce_jwt_decode_handler_updated_configs(token):
+    """
+    Same as original with minor modifications to expect
+    configuration format matching the expectations of
+    edx-drf-extensions decoder.
 
-    Returns:
-        dict: The JWT's payload.
+      JWT_AUTH:
+        JWT_ISSUERS:
+          - AUDIENCE: '{{ COMMON_JWT_AUDIENCE }}'
+            ISSUER: '{{ COMMON_JWT_ISSUER }}'
+            SECRET_KEY: '{{ COMMON_JWT_SECRET_KEY }}'
 
-    Raises:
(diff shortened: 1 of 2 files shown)


robrap:
@nasthagiri: This is ready.  I have a small follow-up PR that I am working on that I'm not yet sure whether or not it needs to be a follow-up. :)

ecommerce/extensions/api/handlers.py
```diff
@@ -2,25 +2,66 @@
...
+                algorithms=[api_settings.JWT_ALGORITHM]
+            )
+        except jwt.InvalidIssuerError:
+            # Ignore these errors since we have multiple issuers
+            pass
+        except jwt.DecodeError:
+            # Ignore these errors since we have multiple issuers
+            pass
+        except jwt.InvalidTokenError:
+            logger.exception('Custom config JWT decode failed!')
```
nasthagiri, line 59: Why not put all of the above exceptional paths behind the waffle switch?  And log for each one - just in case?

ecommerce/extensions/api/handlers.py
```diff
@@ -53,6 +94,57 @@ def jwt_decode_handler(token):
...
+
+    """
+
+    # The following versions of decoding are part of a larger rollout plan that
+    # will ultimately end in retiring this ecommerce jwt_decode_handler
+    # altogether.  See ARCH-261 for detailed plan.
+
+    # first, try jwt_decode_handler from edx_drf_extensions
+    try:
+        jwt_payload = edx_drf_extensions_jwt_decode_handler(token)
```
nasthagiri, line 123: Are there any `log.exception` or `log.warning` messages in the `edx_drf_extensions` decoder that may over-populate splunk?

ecommerce/extensions/api/handlers.py
```diff
@@ -53,6 +94,57 @@ def jwt_decode_handler(token):
                 # Ignore these errors since we have multiple signing keys
                 pass
             except jwt.InvalidTokenError:
-                logger.exception('JWT decode failed!')
+                logger.exception('Original JWT decode failed!')
```
nasthagiri, line 97: Should this also be behind the waffle switch?

ecommerce/extensions/api/handlers.py
robrap: Nice catch. Will double-check warn and info.

nasthagiri (approved):

ecommerce/extensions/api/handlers.py
nasthagiri: Note: @robrap clarified that this exceptional case only happens when there's something wrong with the token itself - like an expiration date. For the above exceptional paths, it can be noisy in the multiple issuer case - and we'll add logging afterwards if we really need to debug those cases.

ecommerce/extensions/api/handlers.py
nasthagiri: See note above regarding this exception being specific to invalid tokens.

PR #1967, opened 2018-11-05
Adds Middlewares to debug Request/Session objects

adeelkhan:
These are temporary middlewares used to debug
issue of no state available inside session by
inspecting & logging request/session loaded
during authentication on ecommerce via lms.

[LEARNER-6226](https://openedx.atlassian.net/browse/LEARNER-6226)

### DIFF
--- a/ecommerce/social_auth/middleware.py
+++ b/ecommerce/social_auth/middleware.py
@@ -1,0 +1,209 @@
+
+import hashlib
+import logging
+import waffle
+from social_django.middleware import SocialAuthExceptionMiddleware
+from social_core.exceptions import AuthStateMissing
+log = logging.getLogger('ISM')
+
+
+ENABLE_SESSION_INSPECT = 'enable_session_inspect'
+
+
+class LogInspectRequestSessionMiddleware(object):
+    """Middleware to inspect session for different values
+        and see what key and values are being created on Request.
+    """
+
+    def process_request(self, request):
+
+        active = waffle.switch_is_active(ENABLE_SESSION_INSPECT)
+
+        request_info = {}
+        if active:
+            request_logs = ''
+            request_logs += '\n=======================================================================\n'
+            request_logs += 'Request:' + _get_client_ip(request) + ' Logs for Sessions/Request object'
+            request_logs += '\n=======================================================================\n'
+            request_info['log_title'] = request_logs
+
+            # Ensuring that this middleware is only hit when both session and
+            # user object is already populated by upper levels middlewares
+            # for session and user object
+            if hasattr(request, 'session') and hasattr(request, 'user'):
+
+                _inspecting_request(request, request_info)
+                _inspecting_session(request, request_info)
+                _generate_logs(request_info)
+
+            else:
+                log.info('No session and authentication middleware called.')
+
+
+class LogInspectResponseSessionMiddleware(object):
+    """Middleware to inspect session for different values
+        and see what key and values are being created on Response.
+    """
+
+    def process_response(self, request, response):
+
+        active = waffle.switch_is_active(ENABLE_SESSION_INSPECT)
+
+        if active:
+            response_info = {}
+            response_logs = ''
+            response_logs += '\n=======================================================================\n'
+            response_logs += 'Response:' + _get_client_ip(request) + ' Logs for Sessions/Request object'
+            response_logs += '\n=======================================================================\n'
+            response_info['log_title'] = response_logs
+
+            # Ensuring that this middleware is only hit when both session and
+            # user object is already populated by lower levels middlewares
+            # for session and user object.
+            if hasattr(request, 'session') and hasattr(request, 'user'):
+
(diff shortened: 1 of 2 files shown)


robrap:
Before adding additional logging, I just want to make sure you are aware of what was added to NewRelic.

See:

- [Authentication Related Metrics](https://openedx.atlassian.net/wiki/spaces/AC/pages/807371420/Authentication+Related+Metrics)
  - Note the info about `tripId` which can be useful to follow a transaction across services.
- [More on all Custom Metrics (including other Authentication Metrics)](https://openedx.atlassian.net/wiki/spaces/PERF/pages/54362736/Custom+Metrics+in+New+Relic)

adeelkhan:
@robrap Going through the above documentation i couldn't find anything related to sso that is done via openid. The authentication section in the doc is related to DOT/DOP based access. As per my understanding DOT/DOP are used for api request and not for sso and the existing metrics isn't logging anything related to it. I want to log the flow when user is redirected to ecommerce via lms during authentication.

robrap:
Thanks @adeelkhan.  I don't want you to have to over-invest if this is just temporary debugging help.  Here's what I'd like to see:
1. Please add a waffle switch to turn on/off logging.  This will be useful if we find that you get way too many logs.
2. Add try blocks to protect again any coding issues you may have missed.
3. Manually test on stage once this is ready.  (This may require coordination, because I may need to merge it due to the codecov issue.)

adeelkhan:
@robrap I have made changes for pt#1 and pt#2. Please review it when possible. thanks

robrap (commented):
Is the session_key code printing the session id? Is this a security concern? Do you have sample output?

Also, the NewRelic request metrics added were for all requests, so you may have some data available already, unless possibly it is for errors.

Again, because this is temporary, I’d do whatever you need as long as there isn’t a security concern.

ecommerce/social_auth/middleware.py
```diff
@@ -0,0 +1,120 @@
...
+    try:
+        if request.user.is_authenticated():
+            log.info("Session Type: Authenticated")
+        else:
+            log.info("Session Type: Unauthenticated")
+        if request.user.is_authenticated():
+            log.info(request.user.email)
+
+        if request.session:
+            session_key = str(request.session.session_key)
```
robrap: I confirmed the security concern here, and we definitely don't want to log anything that could enable someone to steal a session.  If this info is absolutely needed, you could use a one-way hash to encrypt this info, and someone couldn't reverse it.

robrap (changes requested):
Note the security concern.  Thanks.

ecommerce/social_auth/middleware.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _inspecting_request(request):
+    """
+    Logging information from request object
+    :param request:
+    :return: None
+    """
+
+    log.info('-----------------Inspecting Request:-------------')
```
robrap: FYI: I'll leave it to you to determine what is useful, but I would imagine you would want a single log message containing everything you want to output, otherwise these messages could get intermingled with other calls. Maybe Splunk makes it simple for you and you plan on using IP or something to correlate?

ecommerce/social_auth/middleware.py
adeelkhan: Yes that was a concern in my mind too, that is why i am logging just a part of session key.  I think hashing is good idea, that would remove possibility of exposing session key info all together.

ecommerce/social_auth/middleware.py
adeelkhan: thanks for raising this. Ideally i think a single log message is the right solution and yes there must be a way to correlate request/response logs with IP.  I'll implement that too.

ecommerce/social_auth/middleware.py
robrap: Yes. Please either hash this if you think it is necessary, or remove it. It was also pointed out that the IP might be a good enough proxy for the user. So, you could try landing this more quickly without hashed session ID and try using the IP and see if you have enough info that way, and if not, you can add more.

In all cases, I think you have enough info to review and land this without further review from Cambridge. 

Good luck getting to the bottom of things.

robrap (commented):
Feel free to complete the review process locally and have Waheed merge when ready. Thanks.

robrap:
@adeelkhan: Are you still pursuing this or should it be closed?

adeelkhan:
@robrap Yup. Was stuck in some other more important tickets so had to switch my attentions to them, but now have time to look into this. Secondly exploring a possibility of not using this merge at all as some one was facing the same problem and thought about first knowing their resolutions and fixing things at social-auth package level as the person is claiming. For now i am not able to reproduce above issue by any configuration of package. I have no intention to merge this or next week, until i am dead sure that we really need to merge to know more.

robrap:
Thanks.  We were trying to do some PR clean-up.  Feel free to close and reopen when and if you need this if you want.  Good luck.

PR #1968, opened 2018-11-05
ENT-1281/ENT-1288 Refactoring Coupon create/update api backend

brittneyexline:
**Description**
In this PR, the backend api for creating and updating Coupons is refactored to better support the new coupon screen. There are no functional changes in how the code works, all changes are related to how the code is structured and organized.

**Test Instructions**
Most of what we need to test for is finding regressions.
- Create a coupon without enterprise data.
- Create a coupon with enterprise data. Each voucher should be linked to two conditional offers in the database.
- Update a non-enterprise related field on a previously existing coupon that does not have enterprise data.
- Update a coupon without an enterprise to have an enterprise. This should create an additional enterprise conditional offer for each voucher.
- Update a coupon set to have multiple codes that can be used once by multiple customers. Download the csv and proceed with redeeming a code. It should be successful. [Specifically, we are testing that https://openedx.atlassian.net/browse/ENT-1288 is fixed]

**Special Notes**
This PR stops short of being able to have a new enterprise conditional offer created when updating previously existing coupons with an enterprise set but no enterprise catalog. It has no adverse effects, and it should be possible with the next PR.

### DIFF
--- a/ecommerce/extensions/api/v2/views/coupons.py
+++ b/ecommerce/extensions/api/v2/views/coupons.py
@@ -24,7 +24,9 @@ from ecommerce.extensions.catalogue.utils import create_coupon_product, get_or_c
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
 from ecommerce.extensions.payment.processors.invoice import InvoicePayment
 from ecommerce.extensions.voucher.models import CouponVouchers
-from ecommerce.extensions.voucher.utils import update_voucher_offer, update_voucher_with_enterprise_offer
+from ecommerce.extensions.voucher.utils import (
+    get_or_create_enterprise_offer, update_voucher_offer, update_voucher_with_enterprise_offer
+)
 from ecommerce.invoice.models import Invoice
 
 Basket = get_model('basket', 'Basket')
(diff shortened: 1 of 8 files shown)


ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -134,6 +113,32 @@ def create(self, request, *args, **kwargs):
         except ValidationError as e:
             raise serializers.ValidationError(e.message)
 
+    def create_coupon_product(self, cleaned_voucher_data):
```
georgebabey: I like moving this out, but do you think its confusing having this local method the same name as the one imported from catalogue.utils?

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -290,18 +295,18 @@ def create_order_for_invoice(self, basket, coupon_id, client, invoice_data=None)
 
         return response_data
 
-    def update_range_data(self, request, vouchers):
+    def update_range_data(self, request_data, vouchers):
```
georgebabey: We should probably make the same change to `clean_voucher_request_data` above as it also just uses request.data (the name even hints at this)

georgebabey (approved):

ecommerce/extensions/api/v2/views/coupons.py
brittneyexline: yea, i could change it

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -437,95 +428,80 @@ def update_coupon_offer(self, coupon, vouchers, site, benefit_value=None, progra
...
-            client_username (str): Client username
-        """
-        client, __ = BusinessClient.objects.get_or_create(name=client_username)
-        Invoice.objects.filter(order__basket=baskets.first()).update(business_client=client)
+            voucher.offers.add(updated_original_offer)
+            if updated_enterprise_offer:
+                voucher.offers.add(updated_enterprise_offer)
 
-    def update_invoice_data(self, coupon, data):
+    def update_invoice_data(self, data, coupon):
```
adamstankiewicz: Should `data` be named `request_data` to be consistent with the other methods?

adamstankiewicz (approved):

PR #1969, opened 2018-11-06
ENT-1281 Add enterprise coupon backend api

brittneyexline:
**Description**
This PR introduces new api endpoints for creating and updating Enterprise Coupons, which inherit from the existing Coupon endpoints. We also ensure that based on the enterprise coupon switch, that the original coupon endpoints will not accept or return enterprise data. In order to aid with the transition, we have disabled the patch option for the coupon screen so that all fields get sent for updates instead of just changed fields.

**Test Instructions**
Deployed on the business sandbox.

_With the ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH off:_
- Creating and updating coupons without enterprise data should be unaffected.
- Creating a coupon with enterprise data will result in two conditional offers created per voucher, with the enterprise conditional offer having a condition with proxy_class of AssignableEnterpriseCustomerCondition. The client name will be the Enterprise name even if a client name is supplied. The Business Client model will have the enterprise uuid field populated.
- Updating a coupon with enterprise data should result in an additional enterprise conditional offer created or updated, and the client name will be the enterprise name. Even updates with no data changes should trigger these updates. This will allow us to migrate coupons no matter what enterprise data has already been set.

_With the ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH on:_
- Creating and updating coupons without enterprise data should be unaffected.
- All "migrated" enterprise coupons will not show in the list view. A "migrated" enterprise coupon is one that has the attached business client model with the enterprise uuid set, as well as enterprise conditional offers for each voucher available.
- Non "migrated" enterprise coupons can still be updated, but once updated, they will not show up in the list view.
- New coupons with enterprise data cannot be created from the coupons screen.

_Note:_ The backend apis for CRUD operations on enterprise coupons are created in this PR, but the new coupon screen UI that utilizes those apis is not included in this change set, so test manual testing instructions for exercising those endpoints will be included in the PR that has the coupon screen UI.


ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -12,3 +41,114 @@ class EnterpriseCustomerViewSet(generics.GenericAPIView):
     def get(self, request):
         site = request.site
         return Response(data={'results': get_enterprise_customers(site)})
+
+
+class EnterpriseCouponViewSet(CouponViewSet):
```
mushtaqak, line 46: @brittneyexline so this view only returns a list of enterprise coupons ... I am just wondering what I would need for listing coupons for specific enterprise (https://openedx.atlassian.net/browse/ENT-1260) ... Pardon me if I got something wrong...

brittneyexline: the data needed for the portal frontend is not the same as what is returned for this endpoint, which is intended to be used for the edX operator creating coupons on behalf of customers. For example, the endpoint you are building will return summary data about how many codes have been redeemed (relevant to the portal frontend, but not the coupon edit screen), while this one will return details about the invoice (relevant to the coupon edit screen, but not to the portal frontend). 
However, you will likely need to use the same queryset filtering that this viewset does. The simplest thing to do would to make a new detail endpoint (called summary?) in this viewset and use a different serializer to return the relevant data, and that way you get the queryset for free.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -88,6 +102,7 @@ def create(self, request, *args, **kwargs):
         try:
             with transaction.atomic():
                 try:
+                    self.validate_acess_for_enterprise_switch(request.data)
```
mushtaqak: nit: typo
```suggestion
                    self.validate_access_for_enterprise_switch(request.data)
```

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -348,6 +374,7 @@ def update(self, request, *args, **kwargs):
         """Update coupon depending on request data sent."""
         try:
             super(CouponViewSet, self).update(request, *args, **kwargs)
+            self.validate_acess_for_enterprise_switch(request.data)
```
mushtaqak: nit: typo
```suggestion
            self.validate_access_for_enterprise_switch(request.data)
```

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -12,3 +42,118 @@ class EnterpriseCustomerViewSet(generics.GenericAPIView):
...
+                enterprise_catalog=enterprise_catalog,
+                email_domains=email_domains,
+                site=site,
+            )
+            updated_orginal_offer = None
+            if voucher.original_offer != voucher.enterprise_offer:
+                updated_orginal_offer = update_voucher_offer(
+                    offer=voucher.original_offer,
+                    benefit_value=benefit_value,
+                    max_uses=max_uses,
```
georgebabey, line 156: What happens if the new max_uses is less than the already redeemed vouchers?

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -63,3 +86,130 @@ def test_get_customers(self, mock_client):
...
+            if method == 'GET':
+                return self.client.get(path)
+            elif method == 'POST':
+                return self.client.post(path, json.dumps(data), 'application/json')
+            elif method == 'PUT':
+                return self.client.put(path, json.dumps(data), 'application/json')
+        return None
+
+    def test_list_enterprise_coupons(self):
+        Switch.objects.update_or_create(name=ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH, defaults={'active': True})
```
mushtaqak, line 129: Suggestion:
We can probably add a decorator like:
```python
from functools import wraps
def override_switch(switch, active):
    """
    Overrides the given waffle switch to `active` boolean.

    Arguments:
        switch(str): switch name
        active(bool): A boolean representing (to be overridden) value
    """
    def decorate(function):
        @wraps(function)
        def inner(*args, **kwargs):
            Switch.objects.update_or_create(name=switch, defaults={'active': active})
            return function(*args, **kwargs)
        return inner

    return decorate
```
And use before the tests.
```python
@override_switch(ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH, True)
def test_list_enterprise_coupons(...)
```

ecommerce/extensions/api/v2/views/enterprise.py
brittneyexline: i don't think any validation exists right now, so as it is it probably means that the voucher would become unredeemable.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
brittneyexline: there are a couple test cases that require switching to on or off at different points in the test, so I am going to leave it as is. eventually, we will have to clean up the flag everywhere anyway, including the tests, once the feature is released and stable.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -141,6 +159,12 @@ def create_coupon_and_vouchers(self, cleaned_voucher_data):
             site=self.request.site
         )
 
+    def validate_access_for_enterprise_switch(self, request_data):
+        enterprise_customer_data = request_data.get('enterprise_customer')
+        if (enterprise_customer_data and enterprise_customer_data.get('id') and
+                waffle.switch_is_active(ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH)):
```
muhammad-ammar, line 162: I think we can do this.
```python
enterprise_customer_data = request_data.get('enterprise_customer', {})
if enterprise_customer_data.get('id') and waffle.switch_is_active(ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH):
    raise ValidationError('Enterprise coupons can no longer be created or updated from this endpoint.')
```

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -195,6 +219,7 @@ def clean_voucher_request_data(cls, request_data, partner):
 
         try:
             enterprise_customer = enterprise_customer_data['id'] if enterprise_customer_data else None
+            enterprise_customer_name = enterprise_customer_data.get('name') if enterprise_customer_data else None
```
muhammad-ammar, line 219: we can follow the same pattern as mentioned in the above comment.

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -402,8 +428,14 @@ def update_coupon_product_data(self, request_data, coupon):
             ProductCategory.objects.filter(product=coupon).update(category=category)
 
         client_username = request_data.get('client')
-        if client_username:
-            client, __ = BusinessClient.objects.get_or_create(name=client_username)
+        enterprise_customer_data = request_data.get('enterprise_customer')
+        enterprise_customer = enterprise_customer_data.get('id', None) if enterprise_customer_data else None
+        enterprise_customer_name = enterprise_customer_data.get('name', None) if enterprise_customer_data else None
```
muhammad-ammar, line 430: same here.
```python
enterprise_customer_data = request_data.get('enterprise_customer', {})
enterprise_customer = enterprise_customer_data.get('id')
enterprise_customer_name = enterprise_customer_data.get('name')
```

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -429,7 +461,8 @@ def update_offer_data(self, request_data, vouchers, site):
         """
         program_uuid = request_data.get('program_uuid')
         benefit_value = request_data.get('benefit_value')
-        enterprise_customer = request_data.get('enterprise_customer', {}).get('id', None)
+        enterprise_customer_data = request_data.get('enterprise_customer')
+        enterprise_customer = enterprise_customer_data.get('id', None) if enterprise_customer_data else None
```
muhammad-ammar, line 462:
```python
enterprise_customer_data = request_data.get('enterprise_customer', {})
enterprise_customer = enterprise_customer_data.get('id')
```

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -12,3 +41,128 @@ class EnterpriseCustomerViewSet(generics.GenericAPIView):
...
+        """Update coupon depending on request data sent."""
+        try:
+            self.validate_access_for_enterprise_switch(request.data)
+        except ValidationError as error:
+            logger.exception(error.message)
+            raise serializers.ValidationError(error.message)
+        return super(EnterpriseCouponViewSet, self).update(request, *args, **kwargs)
+
+    def update_range_data(self, request_data, vouchers):
+        pass
```
muhammad-ammar, line 108: Will this be addressed in a future PR?

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -12,3 +41,128 @@ class EnterpriseCustomerViewSet(generics.GenericAPIView):
...
+        """
+        Remove all offers from the vouchers and add a new offer
+        Arguments:
+            coupon (Product): Coupon product associated with vouchers
+            vouchers (ManyRelatedManager): Vouchers associated with the coupon to be updated
+            site (Site): The Site associated with this offer
+            benefit_value (Decimal): Benefit value associated with a new offer
+            program_uuid (str): Program UUID
+            enterprise_customer (str): Enterprise Customer UUID
+            enterprise_catalog (str): Enterprise Catalog UUID
```
muhammad-ammar: we need to update the docstring according to the three received params.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -12,3 +41,128 @@ class EnterpriseCustomerViewSet(generics.GenericAPIView):
...
+                    'Failed to update Coupon. '
+                    'max_global_applications field cannot be set for voucher type [{voucher_type}].'.format(
+                        voucher_type=Voucher.SINGLE_USE
+                    ))
+            try:
+                max_uses = int(max_uses)
+                if max_uses < 1:
+                    raise ValueError
+            except ValueError:
+                raise ValidationError('max_global_applications field must be a positive number.')
```
muhammad-ammar, line 137: should the error message say `max_uses` or `max_global_applications` or some other message like `Maximum Number of Uses field be a positive number`?

ecommerce/extensions/voucher/utils.py
```diff
@@ -416,7 +416,7 @@ def get_or_create_enterprise_offer(
     enterprise_customer_name = enterprise_customer_object.get('name', '')
 
     condition, __ = Condition.objects.get_or_create(
-        proxy_class=class_path(EnterpriseCustomerCondition),
+        proxy_class=class_path(AssignableEnterpriseCustomerCondition),
```
muhammad-ammar, line 419: Is this the pre-req code for code assignment feature?

muhammad-ammar (commented):
@brittneyexline Looks good to me. Left some minor comments.

ecommerce/extensions/api/v2/views/coupons.py
brittneyexline: when testing on the sandbox, i ran into issues with code like this because enterprise_customer was sent explicitly as null, so it didn't fall through to being set to {}, which caused issues when get is called.

ecommerce/extensions/api/v2/views/enterprise.py
brittneyexline: no - enterprise coupons do not have ranges. though perhaps a comment is warranted to explain why we're doing this.

ecommerce/extensions/api/v2/views/enterprise.py
brittneyexline: sure, this logic was moved from somewhere else in the code and i didn't think to change the message.

ecommerce/extensions/voucher/utils.py
brittneyexline: yes, so that we are writing the correct proxy class from the start and we don't have to update these coupons again.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -12,3 +41,126 @@ class EnterpriseCustomerViewSet(generics.GenericAPIView):
...
+    def get_queryset(self):
+        invoices = Invoice.objects.filter(business_client__enterprise_customer_uuid__isnull=False)
+        orders = Order.objects.filter(id__in=[invoice.order_id for invoice in invoices])
+        basket_lines = Line.objects.filter(basket_id__in=[order.basket_id for order in orders])
+        return Product.objects.filter(
+            product_class__name=COUPON_PRODUCT_CLASS_NAME,
+            stockrecords__partner=self.request.site.siteconfiguration.partner,
+            id__in=[line.product_id for line in basket_lines],
+            coupon_vouchers__vouchers__offers__condition__enterprise_customer_uuid__isnull=False,
+        )
```
muhammad-ammar: @brittneyexline Tests are failing on this PR and mine because now multiple Product objects are returned. This is also happening when I hit the `/api/v2/enterprise/coupons/{coupon_id}/codes` endpoint. Is this expected?

brittneyexline: no, I am still working on resolving this.

brittneyexline: @muhammad-ammar you should be good to rebase now!

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -547,54 +371,180 @@ def test_update(self):
         self.assertEqual(response_data['title'], 'New title')
         self.assertIsNone(response_data['email_domains'])
 
-    def _test_update_enterprise_coupon(self):
+    def test_update_multi_offer_coupon(self):
```
christopappas, line 374: missing a doctring here I believe

ecommerce/extensions/api/serializers.py
```diff
@@ -70,6 +70,12 @@ def retrieve_condition(obj):
     return retrieve_offer(obj).condition
 
 
+def retrieve_enterprise_condition(obj):
+    """Helper method to retrieve the benefit from voucher. """
+    enterprise_offer = retrieve_enterprise_offer(obj)
+    return enterprise_offer and enterprise_offer.condition
```
christopappas, line 76: i've never seen `and` used like this before. interesting!

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -12,3 +41,126 @@ class EnterpriseCustomerViewSet(generics.GenericAPIView):
...
+            product_class__name=COUPON_PRODUCT_CLASS_NAME,
+            stockrecords__partner=self.request.site.siteconfiguration.partner,
+            id__in=[line.product_id for line in basket_lines],
+            coupon_vouchers__vouchers__offers__condition__enterprise_customer_uuid__isnull=False,
+        ).distinct()
+
+    def get_serializer_class(self):
+        if self.action == 'list':
+            return EnterpriseCouponListSerializer
+        return CouponSerializer
```
christopappas, line 63: I think I might understand, but, why do we use the regular Coupon Serializer for the EnterpriseCouponViewSet when not dealing with the list view?

christopappas (commented):
Made a few assorted comments/questions, but nothing major! 👍

ecommerce/extensions/voucher/utils.py
```diff
@@ -555,6 +555,63 @@ def validate_voucher_fields(
...
+        max_uses,
+        voucher_type,
+        benefit_type,
+        benefit_value,
+        code,
+        end_datetime,
+        start_datetime)
+
+    offers = []
+    num_of_offers = quantity if voucher_type in (Voucher.MULTI_USE, Voucher.ONCE_PER_CUSTOMER) else 1
```
muhammad-ammar, line 586: @brittneyexline we need to cleanup the `quantity` and convert it `int`. Currently it is a string here and we are getting an exception on the `range` method. This is reproducible on sandbox

Here is the stacktrace.
```python
b/python2.7/site-packages/rest_framework/views.py", line 489, in dispatch
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/views.py", line 449, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/rest_framework/views.py", line 486, in dispatch
  File "/edx/app/ecommerce/ecommerce/ecommerce/extensions/api/v2/views/coupons.py", line 113, in create
    coupon_product = self.create_coupon_and_vouchers(cleaned_voucher_data)
  File "/edx/app/ecommerce/ecommerce/ecommerce/extensions/api/v2/views/enterprise.py", line 123, in create_coupon_and_vouchers
    name=cleaned_voucher_data['title']
  File "/edx/app/ecommerce/ecommerce/ecommerce/extensions/voucher/utils.py", line 586, in create_enterprise_vouchers
    for num in range(num_of_offers):
TypeError: range() integer end argument expected, got unicode.
```

ecommerce/extensions/api/v2/views/enterprise.py
brittneyexline: CouponSerializer is what is used in the parent view for the detailed views for Coupons, and there was no need for it to be different for this view.

ecommerce/extensions/voucher/utils.py
brittneyexline: thanks for finding this!

muhammad-ammar: I think we need to store the converted value into `quantity` because we have a second loop`for i in range(quantity):` in this function which iterates over `quantity` so it will give the same crash

brittneyexline: should be taken care of now 👍

brittneyexline:
@georgebabey - need your final approval on this, and this is the first PR that should be merged to unblock the others.

georgebabey (approved):

PR #1971, opened 2018-11-07
LEARNER-6679 Updated ProgramApplicator to CustomApplicator.

Ayub-Khan:
-Override Applicator get_site_offers to only get offers
valid for baket without bundle.
-Override BasketMiddleware apply_offers_to_basket to
use CustomApplicator.
-Used CustomApplicator in basketmiddleware and utils
-Updated constants
-Updated tests

### DIFF
--- a/ecommerce/extensions/basket/middleware.py
+++ b/ecommerce/extensions/basket/middleware.py
@@ -1,7 +1,14 @@
-from oscar.apps.basket.middleware import BasketMiddleware as OscarBasketMiddleware
-from oscar.core.loading import get_model
+import newrelic.agent
+import waffle
 
+from oscar.apps.basket.middleware import BasketMiddleware as OscarBasketMiddleware
+from oscar.core.loading import get_class, get_model
+
+from ecommerce.extensions.offer.constants import CUSTOM_APPLICATOR_USE_FLAG
+
+Applicator = get_class('offer.applicator', 'Applicator')
 Basket = get_model('basket', 'basket')
+CustomApplicator = get_class('offer.applicator', 'CustomApplicator')
 
 
 class BasketMiddleware(OscarBasketMiddleware):
@@ -62,3 +69,11 @@ class BasketMiddleware(OscarBasketMiddleware):
         request._basket_cache = basket
 
         return basket
+
+    @newrelic.agent.function_trace()
+    def apply_offers_to_basket(self, request, basket):
+        if not basket.is_empty:
+            if waffle.flag_is_active(request, CUSTOM_APPLICATOR_USE_FLAG):  # pragma: no cover
+                CustomApplicator().apply(basket, request.user, request)
+            else:
+                Applicator().apply(basket, request.user, request)
--- a/ecommerce/extensions/offer/applicator.py
+++ b/ecommerce/extensions/offer/applicator.py
@@ -5,7 +5,7 @@ import waffle
 from oscar.apps.offer.applicator import Applicator
 from oscar.core.loading import get_model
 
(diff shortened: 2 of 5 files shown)


ecommerce/extensions/offer/constants.py
```diff
@@ -1,2 +1,2 @@
-PROGRAM_APPLICATOR_LOG_FLAG = 'enable_program_applicator_logging'
-PROGRAM_APPLICATOR_USE_FLAG = 'enable_program_applicator_use'
+CUSTOM_APPLICATOR_LOG_FLAG = 'enable_program_applicator_logging'
+CUSTOM_APPLICATOR_USE_FLAG = 'enable_program_applicator_use'
```
Ayub-Khan: Only updating Variables as we already have flags.

ecommerce/extensions/basket/middleware.py
```diff
@@ -62,3 +69,11 @@ def get_basket(self, request):
         request._basket_cache = basket
 
         return basket
+
+    @newrelic.agent.function_trace()
+    def apply_offers_to_basket(self, request, basket):
```
Ayub-Khan, line 74: This is the place from where applicator is actually called for baskets without vouchers(bundle baskets) so we need to override this. 
https://github.com/django-oscar/django-oscar/blob/1.5.1/src/oscar/apps/basket/middleware.py#L202

ecommerce/extensions/offer/applicator.py
```diff
@@ -67,6 +72,14 @@ def get_offers(self, basket, user=None, request=None):
             )
         )
 
+    def get_site_offers(self):
```
Ayub-Khan, line 75: Overrides Applicator get_site_offers. We do not want to check bundle offers against voucher baskets.
https://github.com/django-oscar/django-oscar/blob/1.5.1/src/oscar/apps/offer/applicator.py#L63

Ayub-Khan:
Previously if there are 4 products in program and student have bought 3 he would have got the bundle discount on the last basket in case of direct purchases(no bundle id in basket). this pr will disable that. After that student will only get discount if purchases the last product or remaining products using the bundle purchase option. If he directly purchases the product wont get any discount. 
But this will save us from 150+ program api calls to course discovery and 150-300+ enrollment and entitlement api calls to platform for every basket/add call.

ecommerce/extensions/offer/constants.py
BenjiLee: We should also update the flags. The flags can always be renamed in the django admin.

robrap:
@BenjiLee @Ayub-Khan @brittneyexline @georgebabey: Can you document the plan for our intention around the applicator(s)?  Given that we call the applicator in multiple places, do we need just one that knows how to respond accordingly?  Was a special one being added for Enterprise, and will that break-down when the/an applicator is used in later parts of the payment process?  If this has already been thought through and it all just works, can you help clarify to the team how it works?  If not, can you make it work and then clarify? ;)

BenjiLee:
This PR looks fine to me. I'm not sure what the enterprise workflow looks like and whether or not it includes a bundle id in the request or if program offer can be applied WITH an enterprise offer. @brittneyexline and @georgebabey may have a better idea of if the new workflow here will affect enterprise.

BenjiLee (approved):

brittneyexline:
@BenjiLee @robrap - we don't yet have plans to change the Applicator logic, so these changes should not impact our plans. If or when we do need to change the Applicator, we will update you on the plans.

robrap:
Thanks  @brittneyexline.

PR #1972, opened 2018-11-07
Update JWT_ISSUERS to the edx-drf-extensions format.

robrap:
Update ecommerce default configs to use the updated format of
JWT_ISSUERS which is a dict including audience, secret-key, and
issuer (name), to match the format used by the edx-drf-extensions
jwt_decode_handler.

ARCH-275

### DIFF
--- a/ecommerce/extensions/api/handlers.py
+++ b/ecommerce/extensions/api/handlers.py
@@ -7,6 +7,7 @@ from django.conf import settings
 from edx_django_utils import monitoring as monitoring_utils
 from edx_rest_framework_extensions.auth.jwt.decoder import jwt_decode_handler as edx_drf_extensions_jwt_decode_handler
 from rest_framework_jwt.settings import api_settings
+from six import string_types
 
 logger = logging.getLogger(__name__)
 
@@ -30,6 +31,7 @@ def _ecommerce_jwt_decode_handler_updated_configs(token):
         'verify_exp': api_settings.JWT_VERIFY_EXPIRATION,
         'verify_aud': settings.JWT_AUTH['JWT_VERIFY_AUDIENCE'],
     }
+    error_msg = ''
 
     # JWT_ISSUERS is not one of DRF-JWT's default settings, and cannot be accessed
     # using the `api_settings` object without overriding DRF-JWT's defaults.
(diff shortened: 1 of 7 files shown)


ecommerce/extensions/api/tests/test_handlers.py
```diff
@@ -97,8 +141,11 @@ def test_decode_success_edx_drf_extensions(self, mock_set_custom_metric):
         the way that edx-drf-extensions is expected, and using the secret-key
         of the first configured issuer.
         """
-        payload = generate_jwt_payload(self.user, issuer='test-issuer')
-        token = generate_jwt_token(payload, 'test-secret-key')
+        # import nose.tools as nosepdb;
+        # nosepdb.set_trace()
```
nasthagiri: Reminder to remove debugging code here.

nasthagiri (approved):
So do you plan to update these configuration settings before deploying the previous PR?

robrap:
@nasthagiri: I added [a commit](https://github.com/edx/ecommerce/pull/1972/commits/af387f10a64c3605d948976e929a7f3dde0b575c) with two major changes:
1. Rather than trying all 3 jwt_decode_handlers, I check if we have a the old or new config for the `JWT_ISSUERS`.  If we have the old, I call the original.  If we have the new, I call the edx-drf-extensions with fallback to the multiple-issuer implementation for the updated config.
2. I now build up an error message in case of failure that provides additional details that will hopefully make it simpler to debug if we do end up having a config issue.

ecommerce/extensions/api/handlers.py
```diff
@@ -118,33 +123,37 @@ def jwt_decode_handler(token):
...
-        raise
+    use_original_jwt_decode_handler = isinstance(settings.JWT_AUTH['JWT_ISSUERS'][0], string_types)
+    if use_original_jwt_decode_handler:
+
+        try:
+            jwt_payload = _ecommerce_jwt_decode_handler_original(token)
+            monitoring_utils.set_custom_metric(JWT_DECODE_HANDLER_METRIC_KEY, 'ecommerce-original')
+            return jwt_payload
+        except Exception:  # pylint: disable=broad-except
+            if waffle.switch_is_active('jwt_decode_handler.log_exception.ecommerce-original'):  # pragma: no cover
```
robrap, line 134: REVIEWER: FYI: I do have a test for this turned on, but didn't want to duplicate the test or introduce DTD just to test the logger turned off, especially since this whole method should be removed imminently.

nasthagiri (approved):

PR #1974, opened 2018-11-09
Enterprise coupons overview

mushtaqak:
The purpose of this PR is to add backend support for listing coupons for an enterprise needed for edx-portal.

From jira story:
> Returns a list of json objects with: Coupon ID (coupon product id), Coupon Name, Valid From and To, # of Codes, Usage Limitation (Maximum # of Usages per code), % of available Code redemptions available that have been redeemed.

Sample endpoint URL: https://ecommerce-entcoupon.sandbox.edx.org/api/v2/coupons/enterprise/cf246b88d5f64908a522fc307e0b0c60/overview/

Sample response:
```python
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "end_date": "2019-03-12T00:00:00Z",
            "has_error": false,
            "id": 6,
            "max_uses": 2,
            "num_codes": 100,
            "num_unassigned": 0,
            "num_uses": 0,
            "start_date": "2018-11-19T00:00:00Z",
            "title": "Test edX coupon2",
            "usage_limitation": "Once per customer"
        },
        {
            "end_date": "2019-02-21T00:00:00Z",
            "has_error": false,
            "id": 5,
            "max_uses": null,
            "num_codes": 10,
            "num_unassigned": 0,
            "num_uses": 0,
            "start_date": "2018-11-19T00:00:00Z",
            "title": "Test coupon1",
            "usage_limitation": "Single use"
        }
    ]
}
```

[Backend API for Coupon Overview - ENT-1260](https://openedx.atlassian.net/browse/ENT-1260)

[Sandbox](https://ecommerce-entcoupon.sandbox.edx.org/coupons/)

**Test Instructions**
- [Create some coupons](https://ecommerce-entcoupon.sandbox.edx.org/coupons/new/) and associate it with some client (eg: entX).
> After you have created a couple of coupons with `entX` client, You might want to update `client`'s `enterprise_customer_uuid` (eg: cf246b88d5f64908a522fc307e0b0c65) from [django admin](https://ecommerce-entcoupon.sandbox.edx.org/admin/core/businessclient/), since sandbox is not linked with LMS.
- [Observe all coupons](https://ecommerce-entcoupon.sandbox.edx.org/coupons/) list in ecommerce.
- Observe the list of coupons associated with <enterprise_customer_id> at 
https://ecommerce-business.sandbox.edx.org/api/v2/coupons/enterprise/cf246b88d5f64908a522fc307e0b0c65/overview/

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -714,6 +714,59 @@ class CouponListSerializer(serializers.ModelSerializer):
         fields = ('category', 'client', 'code', 'id', 'title', 'date_created')
 
 
+class EnterpriseCouponOverviewListSerializer(serializers.ModelSerializer):
+    """
+    Serializer for Enterprise Coupons list overview.
+    """
+    end_date = serializers.SerializerMethodField()
+    has_error = serializers.SerializerMethodField()
+    max_uses = serializers.SerializerMethodField()
+    num_codes = serializers.SerializerMethodField()
+    num_unassigned = serializers.SerializerMethodField()
+    num_uses = serializers.SerializerMethodField()
+    start_date = serializers.SerializerMethodField()
+    usage_limitation = serializers.SerializerMethodField()
+
+    # TODO: ENT-1184
+    def get_num_unassigned(self, obj):  # pylint: disable=unused-argument
+        return 0
+
+    # TODO: ENT-1184
+    def get_has_error(self, obj):   # pylint: disable=unused-argument
+        return False
+
+    # Max number of codes available (Maximum Coupon Usage).
+    def get_max_uses(self, obj):
+        offer = retrieve_offer(obj)
+        return offer.max_global_applications
+
+    # Redemption count.
+    def get_num_uses(self, obj):
+        voucher = retrieve_voucher(obj)
+        return voucher.num_orders
+
+    # Number of codes.
+    def get_num_codes(self, obj):
+        return retrieve_quantity(obj)
+
+    # Usage Limitation (Maximum # of usages per code).
+    def get_usage_limitation(self, obj):
+        return retrieve_voucher_usage(obj)
+
+    def get_start_date(self, obj):
+        return retrieve_start_date(obj)
+
+    def get_end_date(self, obj):
+        return retrieve_end_date(obj)
+
+    class Meta(object):
+        model = Product
+        fields = (
+            'end_date', 'has_error', 'id', 'max_uses', 'num_codes', 'num_unassigned',
+            'num_uses', 'start_date', 'title', 'usage_limitation'
+        )
+
+
 class EnterpriseCouponListSerializer(serializers.ModelSerializer):
     client = serializers.SerializerMethodField()
     enterprise_customer = serializers.SerializerMethodField()
--- a/ecommerce/extensions/api/v2/tests/views/test_enterprise.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_enterprise.py
@@ -68,14 +68,14 @@ class TestEnterpriseCustomerView(EnterpriseServiceMockMixin, TestCase):
         )
         url = reverse('api:v2:enterprise:enterprise_customers')
         result = self.client.get(url)
-        self.assertEqual(result.status_code, 401)
+        self.assertEqual(result.status_code, status.HTTP_401_UNAUTHORIZED)
 
         user = self.create_user(is_staff=True)
 
         self.client.login(username=user.username, password=self.password)
 
         result = self.client.get(url)
-        self.assertEqual(result.status_code, 200)
+        self.assertEqual(result.status_code, status.HTTP_200_OK)
         self.assertJSONEqual(
             result.content.decode('utf-8'),
             {
@@ -95,7 +95,9 @@ class TestEnterpriseCustomerView(EnterpriseServiceMockMixin, TestCase):
 
 @ddt.ddt
 class EnterpriseCouponViewSetTest(CouponMixin, DiscoveryTestMixin, DiscoveryMockMixin, ThrottlingMixin, TestCase):
-    """Test the enterprise coupon order creation functionality."""
+    """
+    Test the enterprise coupon order functionality.
+    """
     def setUp(self):
         super(EnterpriseCouponViewSetTest, self).setUp()
         self.user = self.create_user(is_staff=True)
@@ -119,11 +121,58 @@ class EnterpriseCouponViewSetTest(CouponMixin, DiscoveryTestMixin, DiscoveryMock
         self.course = CourseFactory(id='course-v1:test-org+course+run', partner=self.partner)
         self.verified_seat = self.course.create_or_update_seat('verified', False, 100)
 
+    def get_coupon_voucher(self, coupon):
+        """
+        Helper method to get coupon voucher.
+        """
+        return coupon.attr.coupon_vouchers.vouchers.first()
+
+    def get_coupon_data(self, coupon_title):
+        """
+        Helper method to return coupon data by coupon title.
+        """
+        coupon = Product.objects.get(title=coupon_title)
+        return {
+            'end_date': self.get_coupon_voucher_end_date(coupon),
+            'has_error': False,
+            'id': coupon.id,
+            'max_uses': None,
+            'num_codes': 2,
+            'num_unassigned': 0,
+            'num_uses': 0,
+            'start_date': self.get_coupon_voucher_start_date(coupon),
+            'title': coupon.title,
+            'usage_limitation': 'Single use'
+        }
+
+    def get_coupon_voucher_start_date(self, coupon):
+        """
+        Helper method to return coupon voucher start date.
+        """
+        return self.get_coupon_voucher(coupon).start_datetime.isoformat().replace('+00:00', 'Z')
+
+    def get_coupon_voucher_end_date(self, coupon):
+        """
+        Helper method to return coupon voucher end date.
+        """
+        return self.get_coupon_voucher(coupon).end_datetime.isoformat().replace('+00:00', 'Z')
+
     def get_response(self, method, path, data=None):
-        """Helper method for sending requests and returning the response."""
+        """
+        Helper method for sending requests and returning the response.
+        """
+        enterprise_id = ''
+        enterprise_name = 'ToyX'
+        if data and data.get('enterprise_customer'):
+            enterprise_id = data['enterprise_customer']['id']
+            enterprise_name = data['enterprise_customer']['name']
+
         with mock.patch(
-            "ecommerce.extensions.voucher.utils.get_enterprise_customer",
-            mock.Mock(return_value={'name': 'Fake enterprise'})
+            'ecommerce.extensions.voucher.utils.get_enterprise_customer',
+            mock.Mock(return_value={
+                'name': enterprise_name,
+                'enterprise_customer_uuid': enterprise_id
+            })
         ):
             if method == 'GET':
                 return self.client.get(path)
@@ -131,6 +180,15 @@ class EnterpriseCouponViewSetTest(CouponMixin, DiscoveryTestMixin, DiscoveryMock
                 return self.client.post(path, json.dumps(data), 'application/json')
             elif method == 'PUT':
                 return self.client.put(path, json.dumps(data), 'application/json')
+        return None
+
+    def get_response_json(self, method, path, data=None):
+        """
+        Helper method for sending requests and returning JSON response content.
+        """
+        response = self.get_response(method, path, data)
+        if response:
+            return json.loads(response.content)
         return None
 
     def test_list_enterprise_coupons(self):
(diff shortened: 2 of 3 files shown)


mushtaqak:
@brittneyexline Can you have a very quick look at the following sample API response and see if are going in correct direction.
```python
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "end_date": "2019-03-08T00:00:00Z",
            "id": 2,
            "max_uses": null,
            "num_uses": 0,
            "quantity": 1,
            "start_date": "2018-11-12T00:00:00Z",
            "title": "TESt2",
            "voucher_type": "Single use"
        },
        {
            "end_date": "2019-02-01T00:00:00Z",
            "id": 1,
            "max_uses": null,
            "num_uses": 0,
            "quantity": 100,
            "start_date": "2018-11-12T00:00:00Z",
            "title": "TEST_COUPON",
            "voucher_type": "Multi-use"
        }
    ]
}
```
Esp. about `% of available Code redemptions available that have been redeemed.` part mentioned in the PR description. I mean, we don't need to calculate percentage here, we just need to send the the data to front-end, right?

adamstankiewicz:
> Esp. about `% of available Code redemptions available that have been redeemed.` part mentioned in the PR description. I mean, we don't need to calculate percentage here, we just need to send the the data to front-end, right?

Yeah, the front-end will calculate the percentage of redemptions used/available.

adamstankiewicz:
@mushtaqak @brittneyexline Will we be able to know through this API whether any of the codes have an error (e.g., undeliverable email)? The front-end has an assumption that the API response for the coupons will contain a boolean field like `has_error` to determine whether we should show an error state for a coupon.

adamstankiewicz:
@mushtaqak In the above example API response, do any of those fields align with the number of unassigned codes remaining? Is it the `quantity` field?

See the below screenshot (i.e., "Unassigned Codes") where this data will be used:

image

brittneyexline:
@adamstankiewicz i think including the error at this level is a question of what the ui needs. however, it probably won't be implemented with this change since we haven't implemented notifying or assigning users yet.

robrap:
@adamstankiewicz @mushtaqak: It would be great if names can be updated to be more readily understandable.  So if it is `quantity`, could we rename to `num_unassigned` (or something like that)?

adamstankiewicz:
> @adamstankiewicz i think including the error at this level is a question of what the ui needs. however, it probably won't be implemented with this change since we haven't implemented notifying or assigning users yet.

@brittneyexline Yeah, that makes sense. Just making sure it can eventually be supported in this API 😄

adamstankiewicz:
> @adamstankiewicz @mushtaqak: It would be great if names can be updated to be more readily understandable. So if it is `quantity`, could we rename to `num_unassigned` (or something like that)?

Agreed 👍

mushtaqak:
@adamstankiewicz `quantity` is `# of codes` mentioned in the ticket. However, I think the mockups has changed, so we can either remove it or UI won't use it (atleast not in the overview screen) 

About, `Unassigned Codes`, it's functionality part is not implemented yet (scoped in some other JIRA story). So, I will add just the placeholder in this PR.

Regarding error status, can you explain the a bit more, what info front end would need? copoun status? Sorry, I don't have much of info on coupon/code terms :P

adamstankiewicz:
@mushtaqak Sounds good about the placeholder for number of unassigned codes. If we do keep the `quantity` field, can we rename it to be more explicit about what it's referring to?

As for the error status, the UI needs to know whether a code(s) within a coupon contains an error to properly prompt the user that an error exists. The coupon doesn't need to know what the actual error is; just that one exists. It could just be a boolean field in the API response, like `has_error: true`. 

For reference, this is the coupon error state in the UI:

image

mushtaqak:
@brittneyexline @adamstankiewicz Can you guys review this?

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -48,7 +53,10 @@ class EnterpriseCouponViewSet(CouponViewSet):
     """ Coupon resource. """
 
     def get_queryset(self):
+        enterprise_id = self.kwargs.get('enterprise_id')
         invoices = Invoice.objects.filter(business_client__enterprise_customer_uuid__isnull=False)
+        if enterprise_id:
```
brittneyexline, line 58: nit: more readable if this is an if/else statement

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -60,8 +68,31 @@ def get_queryset(self):
...
+        elif self.action == 'overview':
+            return EnterpriseCouponOverviewListSerializer
         return CouponSerializer
 
+    @list_route(methods=['get'])
+    def overview(self, request, enterprise_id):
+        """
+        Overview of Enterprise coupons.
+        Returns the following data:
+            - Copoun ID
```
brittneyexline: nit: typo (Coupon)

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -63,3 +81,83 @@ def test_get_customers(self, mock_client):
...
         )
+
+
+
+class EnterpriseCouponViewSetTest(CouponMixin, DiscoveryTestMixin, DiscoveryMockMixin, TestCase):
+    """
+    Tests EnterpriseCouponViewSet works as expected.
+    """
+
+    def setUp(self):
```
brittneyexline, line 101: you'll have to rebase on my branch again - i have just added a bunch of tests so some of this is redundant

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -114,8 +115,40 @@ def setUp(self):
...
         }
-        self.response = self.get_response('POST', COUPONS_LINK, self.data)
-        self.coupon = Product.objects.get(title=self.data['title'])
+        data2 = dict(self.data, **{'title': 'Test coupon 2'})
+        data3 = dict(self.data, **{
+            'enterprise_customer': {'id': self.enterprise_id2},
+            'title': 'Test coupon 3'
+        })
+
+        self.coupon3 = self.create_get_coupon(data3)
```
brittneyexline: i'd prefer if we move these coupon creations to the test method that needs them.

ecommerce/extensions/api/serializers.py
```diff
@@ -618,6 +618,48 @@ class Meta(object):
...
+    voucher_type = serializers.SerializerMethodField()
+
+    # Max num of codes available (Maximum Coupon Usage).
+    def get_max_uses(self, obj):
+        offer = retrieve_offer(obj)
+        return offer.max_global_applications
+
+    # Redemption count.
+    def get_num_uses(self, obj):
+        offer = retrieve_offer(obj)
```
brittneyexline: we will need to revisit this because many vouchers will have two offers, and we'll probably want to add together the num_applications on each offer to get the actual total number of applications.

brittneyexline (commented):
On the right track!

ecommerce/extensions/api/serializers.py
mushtaqak: @brittneyexline Do you think, this will be covered in this PR? What do we have to do? I looked at a couple of places in codebase, do you think something like this would be needed?

```python
offer = voucher.best_offer
redemption_count = offer.num_applications

if voucher.usage == Voucher.SINGLE_USE:
    max_coupon_usage = 1
    redemption_count = voucher.num_orders
elif voucher.usage != Voucher.SINGLE_USE and offer.max_global_applications is None:
    max_coupon_usage = 10000
else:
    max_coupon_usage = offer.max_global_applications
```

mushtaqak:
@brittneyexline I have addressed all of your feedback except the `num_uses` where I have asked the question. Rest this is ready for your review again.

mushtaqak:
@georgebabey Can you review this as well ?

ecommerce/extensions/api/serializers.py
brittneyexline: for now let's leave it as is... i haven't quite figured out how we're going to manage these things when we have to switch which offer we use

brittneyexline (commented):
looking good! the next thing i would like to see is if you could deploy this on the business sandbox and write out some test instructions with different test cases.

mushtaqak:
@brittneyexline [Sandbox](https://ecommerce-entcoupon.sandbox.edx.org/api/v2/coupons/enterprise/cf246b88d5f64908a522fc307e0b0c60/overview/) is ready

adamstankiewicz:
@mushtaqak @brittneyexline Just taking a look at the API response, and all the fields the frontend needs seems to be there! Nice 👍

adamstankiewicz:
It looks like the API response is paginated (e.g., has `prev` and `next` fields), but it doesn't have a `num_pages` field like API responses in our other services. Is this expected?

mushtaqak:
@adamstankiewicz I guess REST APIs have a `count` field (not the`num_pages`). Below is the response containing fields returned ...
```python
{
    "count": ...,
    "next": ...,
    "previous": ...,
    "results": [...]
}

File: ecommerce/extensions/api/v2/tests/views/test_enterprise.py [hunk 6]
```diff
@@ -280,3 +331,64 @@ def test_update_max_uses_invalid_value(self):
...
+        ),
+        (
+            'f5c9149f-8dce-4410-bb0f-85c0f2dda860',
+            []
+        ),
+    )
+    @ddt.unpack
+    def test_get_enterprise_coupon_overview_data(self, enterprise_id, expected_coupons):
+        """
+        Test if we get correct enterprise coupoun overview data.
```
brittneyexline: nit: coupoun should be coupon

ecommerce/extensions/api/v2/urls.py
```diff
@@ -105,6 +105,8 @@
     .register(r'products', product_views.ProductViewSet, base_name='catalog-product',
               parents_query_lookups=['stockrecords__catalogs'])
 router.register(r'coupons', coupon_views.CouponViewSet, base_name='coupons')
+router.register(r'coupons/enterprise/(?P<enterprise_id>.+)', enterprise_views.EnterpriseCouponViewSet,
```
brittneyexline: should we add /overview to the pattern so that we match exactly?

ecommerce/extensions/api/serializers.py
brittneyexline: after pulling some data for an audit of the coupons yesterday, i remembered that there is a `num_orders` field on the voucher - i think using this value will give us the best value for num_uses without having to account for the multiple conditional offers.

brittneyexline (approved):
i think this looks good overall! i have a couple of small suggestions, and in particular the num_uses field should be updated before merging. this is also blocked on merge/release by https://github.com/edx/ecommerce/pull/1969 getting merged/released, so i will try to work with @georgebabey to get that released soon.

ecommerce/extensions/api/v2/urls.py
brittneyexline: i noticed that the paths are slightly different between yours and Ammar's PR - here it is `coupons/enterprise` and on his and for the other endpoints it is `enterprise/coupons`. i think you should make sure the path is consistent with the other endpoints.

georgebabey:
@mushtaqak Looks like this is ready to go but theres some conflicts to resolve. Can you update this and I'll get it merged in.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -63,3 +88,309 @@ def test_get_customers(self, mock_client):
...
+            self.get_response('POST', ENTERPRISE_COUPONS_LINK, dict(self.data, **coupon_data))
+
+        # Build expected results.
+        expected_results = []
+        for coupon_title in expected_coupons:
+            expected_results.append(self.get_coupon_data(coupon_title))
+
+        overview_response = self.get_response_json(
+            'GET',
+            reverse('api:v2:enterprise-coupons-(?P<enterprise-id>.+)/overview-list',
```
georgebabey: Are you sure this is correct? It looks like it should be `/overview` instead of `/overview-list` https://github.com/edx/ecommerce/pull/1974/commits/91fcfa83978643ebb6d3dd20bf87016f2746750b#diff-141fa9c110c00aba0d7f70839c4a77bbR179

georgebabey (changes requested):

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
mushtaqak: Yes, actually ViewSet creates url route for the endpoint. While the URL pattern name is `api:v2:enterprise-coupons-(?P<enterprise-id>.+)/overview-list`, the actual url pattern is `api/v2/enterprise/coupons/<enterprise-id>/overview/`

mushtaqak:
@georgebabey This is ready to merge!

georgebabey (approved):

PR #1976, opened 2018-11-13
LEARNER-6534 Create order notes in case of fulfillment Error

Ayub-Khan:
Create order notes in case of fulfillment Error. So that non technical people know why fulfillment of order was failed. As support team handles those orders later on. They should have a way to find out why fulfillment of order failed without using splunk. 

@edx/learner-spartans  Kindly Review.

### DIFF
--- a/ecommerce/extensions/fulfillment/modules.py
+++ b/ecommerce/extensions/fulfillment/modules.py
@@ -340,16 +340,19 @@ class EnrollmentFulfillmentModule(BaseFulfillmentModule):
                         "Fulfillment of line [%d] on order [%s] failed with status code [%d]: %s",
                         line.id, order.number, response.status_code, reason
                     )
+                    order.notes.create(message=reason, note_type='Error')
                     line.set_status(LINE.FULFILLMENT_SERVER_ERROR)
             except ConnectionError:
                 logger.error(
                     "Unable to fulfill line [%d] of order [%s] due to a network problem", line.id, order.number
                 )
+                order.notes.create(message='Fulfillment of order failed due to a network problem.', note_type='Error')
                 line.set_status(LINE.FULFILLMENT_NETWORK_ERROR)
             except Timeout:
                 logger.error(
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/fulfillment/modules.py
```diff
@@ -340,16 +340,19 @@ def fulfill_product(self, order, lines, email_opt_in=False):
                         "Fulfillment of line [%d] on order [%s] failed with status code [%d]: %s",
                         line.id, order.number, response.status_code, reason
                     )
+                    order.notes.create(message=reason, note_type='Error')
```
Ayub-Khan, line 343: records the reason why lms did not upgrade student enrollment.

ecommerce/extensions/fulfillment/modules.py
```diff
@@ -340,16 +340,19 @@ def fulfill_product(self, order, lines, email_opt_in=False):
...
                     line.set_status(LINE.FULFILLMENT_SERVER_ERROR)
             except ConnectionError:
                 logger.error(
                     "Unable to fulfill line [%d] of order [%s] due to a network problem", line.id, order.number
                 )
+                order.notes.create(message='Fulfillment of order failed due to a network problem.', note_type='Error')
```
Ayub-Khan, line 349: not sure if we want to use gettext hooks around those.

ecommerce/extensions/fulfillment/modules.py
waheedahmed, line 343: I think we can move this to `line.set_status()` and pass the message and type to it , to avoid repetaion.

waheedahmed (commented):
LGTM, just one comment.

waheedahmed (approved):

ecommerce/extensions/fulfillment/modules.py
Ayub-Khan: we will have to override the model and property of orderline model from oscar. could have been easy in case of refund as that is done using mixin and we handle those ourself.

PR #1977, opened 2018-11-13
Reducing severity of log messages for common cases for enterprise conditions

brittneyexline:


ecommerce/enterprise/conditions.py
```diff
@@ -71,7 +71,7 @@ def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
             return False
         except IndexError:
             if offer.offer_type == ConditionalOffer.SITE:
-                logger.warning(
+                logger.info(
```
brittneyexline: should we consider going down to debug?

georgebabey: Debug level makes sense since the majority of learners are non-enterprise but still go through this flow. If/when we create our own applicator that only applies to enterprise learners then we could bounce this back up to a higher level.

georgebabey (approved):

Ayub-Khan:
You missed one : https://github.com/edx/ecommerce/blob/master/ecommerce/enterprise/conditions.py#L83

PR #1978, opened 2018-11-13
ENT-1281/ENT-1288 Refactoring Coupon create/update api backend

brittneyexline:
**Description**
Copy of https://github.com/edx/ecommerce/pull/1968
In this PR, the backend api for creating and updating Coupons is refactored to better support the new coupon screen. There are no functional changes in how the code works, all changes are related to how the code is structured and organized.

**Test Instructions**
Most of what we need to test for is finding regressions.
- Create a coupon without enterprise data.
- Create a coupon with enterprise data. Each voucher should be linked to two conditional offers in the database.
- Update a non-enterprise related field on a previously existing coupon that does not have enterprise data.
- Update a coupon without an enterprise to have an enterprise. This should create an additional enterprise conditional offer for each voucher.
- Update a coupon set to have multiple codes that can be used once by multiple customers. Download the csv and proceed with redeeming a code. It should be successful. [Specifically, we are testing that https://openedx.atlassian.net/browse/ENT-1288 is fixed]

**Special Notes**
This PR stops short of being able to have a new enterprise conditional offer created when updating previously existing coupons with an enterprise set but no enterprise catalog. It has no adverse effects, and it should be possible with the next PR.

### DIFF
--- a/ecommerce/enterprise/conditions.py
+++ b/ecommerce/enterprise/conditions.py
@@ -80,11 +80,11 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
         if (learner_data and 'enterprise_customer' in learner_data and
                 str(self.enterprise_customer_uuid) != learner_data['enterprise_customer']['uuid']):
             # Learner is not linked to the EnterpriseCustomer associated with this condition.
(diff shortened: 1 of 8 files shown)


georgebabey (approved):
Looks good - review done in https://github.com/edx/ecommerce/pull/1968

PR #1979, opened 2018-11-14
ENT-1339 coupon codes detail backend

muhammad-ammar:
[ENT-1339](https://openedx.atlassian.net/browse/ENT-1339)

This adds an endpoint to get usage details of codes belong to a coupon.

**Test Instructions:**
Deployed on the business sandbox.

* Replace the `coupon_id` in the url [https://ecommerce-business.sandbox.edx.org/api/v2/enterprise/coupons/<coupon_id>/codes/](https://ecommerce-business.sandbox.edx.org/api/v2/enterprise/coupons/<coupon_id>/codes/) with a valid coupon id and you should see the detailed usage of codes belong to that coupon.
* For example, https://ecommerce-business.sandbox.edx.org/api/v2/enterprise/coupons/124/codes/

**Sample responses below:**

**once by one customer with all codes redeemed**
**Number of Codes: 2**
```python
{
    "results": [
        {
            "code": "2KA4LB2JJTN3JGFD",
            "redemptions": {
                "used": 1,
                "available": 1,
            }
            "assigned_to": "",
        },
        {
            "code": "FEANQA6ZY2TSKA27",
            "redemptions": {
                "used": 1,
                "available": 1,
            }
            "assigned_to": "",
        },
    ]
}
```

**once by one customer with one code redeemed**
**Number of Codes: 2**
```python
{
    "results": [
        {
            "code": "2KA4LB2JJTN3JGFD",
            "redemptions": {
                "used": 0,
                "available": 1,
            }
            "assigned_to": "",
        },
        {
            "code": "FEANQA6ZY2TSKA27",
            "redemptions": {
                "used": 1,
                "available": 1,
            }
            "assigned_to": "",
        },
    ]
}
```

**once by multiple customers**
**Number of Codes: 1
Maximum Number of Uses: 3**
```python
{
    "results": [
        {
            "code": "SCXWY374JZ4P3G4N",
            "redemptions": {
                "used": 2,
                "available": 3,
            }
            "assigned_to": "",
        },
        {
            "code": "SCXWY374JZ4P3G4N",
            "redemptions": {
                "used": 2,
                "available": 3,
            }
            "assigned_to": "",
        },
    ]
}
```

**multiple times by multiple customers
Number of Codes: 3
Maximum Number of Uses: 3**
```python
{
    "results": [
        {
            "code": "3ZXRIYEKAC6K62RN",
            "redemptions": {
                "used": 3,
                "available": 3,
            }
            "assigned_to": "",
        },
        {
            "code": "3ZXRIYEKAC6K62RN",
            "redemptions": {
                "used": 3,
                "available": 3,
            }
            "assigned_to": "",
        },
        {
            "code": "3ZXRIYEKAC6K62RN",
            "redemptions": {
                "used": 3,
                "available": 3,
            }
            "assigned_to": "",
        },
        {
            "code": "WPVAYMVZ4PPR7RDP",
            "redemptions": {
                "used": 3,
                "available": 3,
            }
            "assigned_to": "",
        },
        {
            "code": "WPVAYMVZ4PPR7RDP",
            "redemptions": {
                "used": 3,
                "available": 3,
            }
            "assigned_to": "",
        },
        {
            "code": "WPVAYMVZ4PPR7RDP",
            "redemptions": {
                "used": 3,
                "available": 3,
            }
            "assigned_to": "",
        },
        {
            "code": "NJIL6VWO446PLXQ7",
            "redemptions": {
                "used": 2,
                "available": 3,
            }
            "assigned_to": "",
        },
        {
            "code": "NJIL6VWO446PLXQ7",
            "redemptions": {
                "used": 2,
                "available": 3,
            }
            "assigned_to": "",
        },
    ]
}
```

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -656,6 +656,34 @@ class VoucherSerializer(serializers.ModelSerializer):
         )
 
 
+class CouponVoucherSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    code = serializers.SerializerMethodField()
+    assigned_to = serializers.SerializerMethodField()
+    redemptions = serializers.SerializerMethodField()
+
+    def get_code(self, voucher):
+        return voucher.code
+
+    def get_assigned_to(self, obj):  # pylint: disable=unused-argument
+        return ''
+
+    def get_redemptions(self, voucher):
+        offer = voucher.best_offer
+        redemption_count = voucher.num_orders
+
+        if voucher.usage == Voucher.SINGLE_USE:
+            max_coupon_usage = 1
+        elif voucher.usage != Voucher.SINGLE_USE and offer.max_global_applications is None:
+            max_coupon_usage = 10000
+        else:
+            max_coupon_usage = offer.max_global_applications
+
+        return {
+            'used': redemption_count,
+            'available': max_coupon_usage,
+        }
+
+
 class CategorySerializer(serializers.ModelSerializer):
     # NOTE (CCB): We are explicitly ignoring child categories. They are not relevant to our current needs. Support
     # should be added later, if needed.
--- a/ecommerce/extensions/api/v2/views/enterprise.py
+++ b/ecommerce/extensions/api/v2/views/enterprise.py
@@ -1,10 +1,12 @@
 from __future__ import unicode_literals
 
 import logging
+
 import waffle
 from django.core.exceptions import ValidationError
 from oscar.core.loading import get_model
 from rest_framework import generics, serializers
+from rest_framework.decorators import detail_route
 from rest_framework.permissions import IsAdminUser, IsAuthenticated
 from rest_framework.response import Response
 
@@ -12,7 +14,11 @@ from ecommerce.core.constants import COUPON_PRODUCT_CLASS_NAME
 from ecommerce.core.utils import log_message_and_raise_validation_error
 from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH
 from ecommerce.enterprise.utils import get_enterprise_customers
-from ecommerce.extensions.api.serializers import CouponSerializer, EnterpriseCouponListSerializer
+from ecommerce.extensions.api.serializers import (
+    CouponSerializer,
+    CouponVoucherSerializer,
+    EnterpriseCouponListSerializer
+)
 from ecommerce.extensions.api.v2.views.coupons import CouponViewSet
 from ecommerce.extensions.catalogue.utils import (
     attach_vouchers_to_coupon_product,
@@ -164,3 +170,42 @@ class EnterpriseCouponViewSet(CouponViewSet):
 
         if coupon_was_migrated:
             super(EnterpriseCouponViewSet, self).update_range_data(request_data, vouchers)
+
+    @detail_route(url_path='codes')
+    def codes(self, request, pk):  # pylint: disable=unused-argument
+        """
+        GET codes belong to a `coupon`.
+
+        Response will looks like
+        {
+            results: [
+                {
+                    code: '1234-5678-90',
+                    assigned_to: 'Barry Allen',
+                    redemptions: {
+                        used: 1,
+                        available: 5,
+                    },
+                },
+            ]
+        }
+        """
+        coupon = self.get_object()
+
+        # this will give us vouchers against each voucher application, so if a
+        # voucher has two applications than this queryset will give 2 vouchers
+        coupon_vouchers_with_applications = Voucher.objects.filter(
+            applications__voucher_id__in=coupon.attr.coupon_vouchers.vouchers.all()
+        )
+        # this will give us only those vouchers having no application
+        coupon_vouchers_wo_applications = Voucher.objects.filter(
+            coupon_vouchers__coupon__id=coupon.id,
+            applications__isnull=True
+        )
+
+        # we need a combined querset so that pagination works as expected
+        all_coupon_vouchers = coupon_vouchers_with_applications | coupon_vouchers_wo_applications
+
+        page = self.paginate_queryset(all_coupon_vouchers)
+        serializer = CouponVoucherSerializer(page, many=True)
+        return self.get_paginated_response(serializer.data)
--- a/ecommerce/extensions/api/v2/tests/views/test_enterprise.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_enterprise.py
@@ -5,15 +5,18 @@ import datetime
 import json
 from uuid import uuid4
 
+import ddt
 import httpretty
 import mock
 from django.urls import reverse
 from django.utils.timezone import now
 from oscar.core.loading import get_model
+from oscar.test import factories
 from rest_framework import status
 from waffle.models import Switch
 
 from ecommerce.coupons.tests.mixins import CouponMixin, DiscoveryMockMixin
+from ecommerce.courses.tests.factories import CourseFactory
 from ecommerce.enterprise.benefits import BENEFIT_MAP as ENTERPRISE_BENEFIT_MAP
 from ecommerce.enterprise.conditions import AssignableEnterpriseCustomerCondition
 from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH
@@ -28,6 +31,7 @@ Basket = get_model('basket', 'Basket')
 Benefit = get_model('offer', 'Benefit')
 Product = get_model('catalogue', 'Product')
 Voucher = get_model('voucher', 'Voucher')
+VoucherApplication = get_model('voucher', 'VoucherApplication')
 
 ENTERPRISE_COUPONS_LINK = reverse('api:v2:enterprise-coupons-list')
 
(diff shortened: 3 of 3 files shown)


muhammad-ammar:
@georgebabey I am working on adding tests. In the mean time can you give it a quick look.

muhammad-ammar:
@georgebabey @brittneyexline This is ready for review.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -166,3 +172,42 @@ def update_offer_data(self, request_data, vouchers, site):
 
         if coupon_was_migrated:
             super(EnterpriseCouponViewSet, self).update_range_data(request_data, vouchers)
+
+    @list_route(url_path=r'(?P<coupon_id>\d+)/codes')
```
brittneyexline: i don't think list route makes sense for this endpoint, because it implies we are returning a list of coupons. i would change this to detail route, indicating details for a particular coupon, and then i'm pretty sure DRF will automatically get the coupon with the provided id, so you won't need the queryset logic you have below.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -166,3 +172,42 @@ def update_offer_data(self, request_data, vouchers, site):
...
+
+        coupon_vouchers_with_applications = Voucher.objects.filter(
+            applications__voucher_id__in=coupon.attr.coupon_vouchers.vouchers.all()
+        )
+        coupon_vouchers_wo_applications = Voucher.objects.filter(
+            coupon_vouchers__coupon__id=coupon_id,
+            applications__isnull=True
+        )
+
+        all_coupon_vouchers = coupon_vouchers_with_applications | coupon_vouchers_wo_applications
```
brittneyexline, line 207: why make the two different calls if we are just merging them back together? this could just be `coupon.attr.coupon_vouchers.vouchers.all()`, right?

ecommerce/extensions/api/serializers.py
```diff
@@ -656,6 +656,35 @@ class Meta(object):
...
+    def get_assigned_to(self, obj):  # pylint: disable=unused-argument
+        return ''
+
+    def get_redemptions(self, voucher):
+        offer = voucher.best_offer
+        redemption_count = offer.num_applications
+
+        if voucher.usage == Voucher.SINGLE_USE:
+            max_coupon_usage = 1
+            redemption_count = voucher.num_orders
```
brittneyexline: is there a reason not to use voucher.num_orders as redemption count for all cases?

ecommerce/extensions/api/serializers.py
```diff
@@ -656,6 +656,35 @@ class Meta(object):
...
+class CouponVoucherSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    code = serializers.SerializerMethodField()
+    assigned_to = serializers.SerializerMethodField()
+    redemptions = serializers.SerializerMethodField()
+
+    def get_code(self, voucher):
+        return voucher.code
+
+    def get_assigned_to(self, obj):  # pylint: disable=unused-argument
+        return ''
```
brittneyexline, line 668: this should probably be a list - multiple people will be assigned to a single code in some cases.

ecommerce/extensions/api/serializers.py
```diff
@@ -656,6 +656,35 @@ class Meta(object):
         )
 
 
+class CouponVoucherSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    code = serializers.SerializerMethodField()
+    assigned_to = serializers.SerializerMethodField()
+    redemptions = serializers.SerializerMethodField()
```
brittneyexline, line 662: my understanding was that the detail api would need to return who has redeemed the code, not just the total number of redemptions. i'm also wondering if we need details about the redemptions beyond user, such as when they redeemed and what course they got. perhaps the ui mocks no longer include this info?

brittneyexline (commented):
I am a little confused about where we landed with what the api call and response would look like. I was under the impression that now that we are toggling between showing "open" slots, assigned, and redeemed, we would have three separate endpoints so that we can properly paginate - is this no longer the case? I'm also not sure if more detailed data is needed than what is here.

ecommerce/extensions/api/serializers.py
muhammad-ammar: @brittneyexline as per mockups we only need the above three things.

ecommerce/extensions/api/serializers.py
muhammad-ammar: I was unsure about that...this will be changed once we handle the assignment functionality.

ecommerce/extensions/api/v2/views/enterprise.py
muhammad-ammar: Lets say a voucher has 3 applications than we need 3 voucher instances of same voucher so that pagination works as expected. We want to expand the results at the queryset level. The below queryset will give us that

```python
coupon_vouchers_with_applications = Voucher.objects.filter(
    applications__voucher_id__in=coupon.attr.coupon_vouchers.vouchers.all()
)
```

The combined queryset gives the both vouchers with and without applications whereas the `coupon.attr.coupon_vouchers.vouchers.all()` only gives us all the vouchers irrespective of applications.

For example, we have 1 voucher and it can be used 2 times. And lets say we have used the voucher once. Then the `all_coupon_vouchers` queryset will give us 2 voucher instances but `coupon.attr.coupon_vouchers.vouchers.all()` will only give us a queryset with 1 voucher instance.

ecommerce/extensions/api/serializers.py
muhammad-ammar: I am not sure about this. This is how the [csv download](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/voucher/utils.py#L191-L198) implemented it. I will check this. Any suggestion?

muhammad-ammar:
> I am a little confused about where we landed with what the api call and response would look like. I was under the impression that now that we are toggling between showing "open" slots, assigned, and redeemed, we would have three separate endpoints so that we can properly paginate - is this no longer the case? I'm also not sure if more detailed data is needed than what is here.

@brittneyexline As per mockups we only need `code`, `redemptions` and `assigned to`. Also `assignment functionality` and `Multiple times by one customer` coupon type is not yet implemented so after discussion with @georgebabey we decided that for now just return redeemed and unredeemed detail about vouchers. But as per my understanding atleast we will always show both the redeemed and unredeemed details combined like we are doing in this PR.

@georgebabey Your thoughts?

muhammad-ammar:
@brittneyexline @georgebabey This is ready for next round of review.

muhammad-ammar:
@georgebabey @brittneyexline This is waiting for review.

ecommerce/extensions/api/v2/views/enterprise.py
brittneyexline: thanks for the explanation - could you add a comment above this line summarizing this?

brittneyexline (commented):
looks good overall, i just have a question about the response format for the multiple uses by multiple users case. it appears that for each redemption, we are returning that code with its full data. for example, this coupon on the sandbox is returning 18 identical objects for the same code: https://ecommerce-business.sandbox.edx.org/api/v2/enterprise/coupons/5/codes/
perhaps it was already discussed/decided, but why not just have the code there once in the response, since the metadata indicates that it was redeemed x number of times?

muhammad-ammar:
>but why not just have the code there once in the response, since the metadata indicates that it was redeemed x number of times?

@ Regarding your above question, [ENT-1171](https://openedx.atlassian.net/browse/ENT-1171) says the below. may be @georgebabey can add more on this?

> **Multiple times by Multiple Customers: show same code in multiple rows equal to the usage limitation**
> When a code can be used multiple times by multiple people: the UI should show an individual row for each usage. i.e. If 20 codes can be used 3 times each by multiple people , there should be 60 rows of codes.

georgebabey:
I believe I was the one that suggested the backend explode the response for multiple times / multiple customers. My rationale:
- If we required the frontend to do this logic it would make pagination extremely difficult. Consider https://ecommerce-business.sandbox.edx.org/api/v2/enterprise/coupons/5/codes/?page_size=5  I've limited `page_size` just to illustrate the code spanning multiple pages. For simplicity lets also imagine the frontend was only showing 5 codes per screen. You can imagine if we required the frontend to handle the expansion we would need send an varying amount of responses to the frontend. Code A may have 2 redemptions already, so in that case we'd send Code A (3/5 remaining) and Code B (5/5 remaining) as response while if Code A had 0 redemptions, we'd only send Code A since it would expand to take the entire screen. Because the expansion of codes directly impacts pagination, having the consumer of the API have to handle that seemed overly complicated.

- Lesser concern, but the fact we only store these as a single code seemed like a backend implementation detail. You could imagine an alternate implementation that creates a table with a row for every "code usage" rather than our single row for each code.

This is not without tradeoffs. This does increase size across the network, but the screen is paginated and admin functionality so I don't think that is a concern. You could argue this is actually the frontend implementation leaking into our backend, and a different UI may prefer not to see this coupon usage case exploded which I can agree with to an extent but this is where we landed.

brittneyexline (approved):
thanks for the explanation about the response format @muhammad-ammar and @georgebabey!

robrap:
@georgebabey: Sounds a lot like an ADR.

georgebabey:
@robrap Sounds good, thanks for offering to capture it

muhammad-ammar:
@brittneyexline are we good to merge this into master?

PR #1980, opened 2018-11-14
ARCH-279: Remove jwt_decode_handler for old config format.

robrap:
The original jwt_decode_handler code for ecommerce had JWT_ISSUERS
configured as a list of issuer names, and the secret keys were in a
separate config JWT_SECRET_KEYS. The newer config that consolidates
these into an array of issuer objects under JWT_ISSUERS is now complete,
so we are removing all the code that relied on the original
configuration that is no longer in use.

NOTE: Top-level JWT_SECRET_KEY config in JWT_AUTH could probably also
be cleaned up, but leaving that to a separate effort.

ARCH-279

### DIFF
--- a/ecommerce/extensions/api/handlers.py
+++ b/ecommerce/extensions/api/handlers.py
@@ -1,4 +1,10 @@
-"""Handler overrides for JWT authentication."""
+"""
+Handler overrides for JWT authentication.
+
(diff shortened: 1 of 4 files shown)


nasthagiri (approved):
Looks good.  Thanks!

PR #1981, opened 2018-11-15
ARCH-283: Use suppress logging in jwt_decode_handler.

robrap:
Temporarily, ecommerce is delegating to the edx-drf-extensions
jwt_decode_handler, with its own fallback to allow for multiple issuers.
Logging suppression inside the edx-drf-extensions jwt_decode_handler is
required until ecommerce is down to a single issuer and can use the
edx-drf-extensions jwt_decode_handler directly.

Note: this should be reverted as part of ARCH-276.

ARCH-283

### DIFF
--- a/ecommerce/extensions/api/tests/test_handlers.py
+++ b/ecommerce/extensions/api/tests/test_handlers.py
@@ -85,8 +85,9 @@ class JWTDecodeHandlerTests(TestCase):
     )
     @override_switch('jwt_decode_handler.log_exception.edx-drf-extensions', active=True)
     @mock.patch('edx_django_utils.monitoring.set_custom_metric')
+    @mock.patch('edx_rest_framework_extensions.auth.jwt.decoder.logger')
     @mock.patch('ecommerce.extensions.api.handlers.logger')
-    def test_decode_success_multiple_issuers(self, mock_logger, mock_set_custom_metric):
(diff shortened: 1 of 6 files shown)


ecommerce/extensions/api/tests/test_handlers.py
```diff
@@ -85,8 +85,9 @@ def test_decode_success_edx_drf_extensions(self, mock_set_custom_metric):
     )
     @override_switch('jwt_decode_handler.log_exception.edx-drf-extensions', active=True)
     @mock.patch('edx_django_utils.monitoring.set_custom_metric')
+    @mock.patch('edx_rest_framework_extensions.auth.jwt.decoder.logger')
```
robrap, line 88: Reviewers: The original unit test was checking the wrong logger.  I ensured that this test fails with the old edx-drf-extensions and passes when requesting to suppress logs.

robrap (commented):
Reviewers: Ignore the quality issue.  It is a temporary isort problem because edx-drf-extensions is temporarily pointing to github.

douglashall:
Is it not possible to modify the ecommerce alerting to ignore these info logs?

douglashall (approved):

robrap:
Closing in favor of https://github.com/edx/ecommerce/pull/1983 which switches the order of the jwt_decode_handlers.

PR #1983, opened 2018-11-16
ARCH-283: Fix logging issue in jwt_decode_handler.

robrap:
Now use ecommerce jwt_decode_handler for multiple issuers first, since
we have more control over the logging, and we can avoid extra logging
from edx-drf-extensions jwt_decode_handler.

Also, added JWT_PUBLIC_SIGNING_JWK_SET to devstack.

ARCH-283

### DIFF
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -406,6 +406,7 @@ JWT_AUTH = {
     'JWT_ISSUERS': (),
     # NOTE (CCB): This is temporarily set to False until we decide what values are acceptable.
     'JWT_VERIFY_AUDIENCE': False,
+    'JWT_PUBLIC_SIGNING_JWK_SET': None,
 }
 
 # Service user for worker processes.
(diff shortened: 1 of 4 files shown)


robrap:
FYI: Tests still need to be updated.  I did test manually with an asymmetric key and it worked.  I added JWT_PUBLIC_SIGNING_JWK_SET as part of this PR to devstack, but other PRs would be needed for other environments.
Note: "ENT-1364: Enable JWT Cookies in ecommerce" is planned for next sprint.

ecommerce/settings/base.py
```diff
@@ -406,6 +406,7 @@
     'JWT_ISSUERS': (),
     # NOTE (CCB): This is temporarily set to False until we decide what values are acceptable.
     'JWT_VERIFY_AUDIENCE': False,
+    'JWT_PUBLIC_SIGNING_JWK_SET': None,
```
nasthagiri, line 409: Was this needed by any code change - or for local development?

nasthagiri (approved):

ecommerce/settings/base.py
robrap: It wasn't required for either.  I was just trying to be consistent, and didn't know if this was forming documentation of sorts of what would ultimately be needed.

PR #1990, opened 2018-11-26
EG-43: Cache catalog query contains call

BenjiLee:
This PR will cache the response from CatalogQueryContainsViewSet from the course-discovery IDA.

https://openedx.atlassian.net/browse/EG-43

https://rpm.newrelic.com/accounts/88178/key_transactions/35300/x_rays/4669#id=75ca5033-f190-11e8-8af5-0242ac11000a_36532_39981&tab-detail_75ca5033-f190-11e8-8af5-0242ac11000a_36532_39981=database_queries

### DIFF
--- a/ecommerce/enterprise/migrations/0004_add_enterprise_offers_for_coupons.py
+++ b/ecommerce/enterprise/migrations/0004_add_enterprise_offers_for_coupons.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 from django.db import migrations
+
 from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH
 
 
--- a/ecommerce/extensions/offer/models.py
+++ b/ecommerce/extensions/offer/models.py
@@ -21,6 +21,7 @@ from threadlocals.threadlocals import get_current_request
 
 from ecommerce.core.utils import get_cache_key, log_message_and_raise_validation_error
 from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH
+from ecommerce.settings.base import CATALOG_QUERY_CONTAINS_TIMEOUT
 
 OFFER_PRIORITY_ENTERPRISE = 10
 OFFER_PRIORITY_VOUCHER = 20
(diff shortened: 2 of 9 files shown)


ecommerce/enterprise/migrations/0004_add_enterprise_offers_for_coupons.py
```diff
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 from django.db import migrations
+
```
robrap, line 5: Why are you messing with migrations?  Was this isort?

ecommerce/extensions/offer/models.py
```diff
@@ -100,48 +101,56 @@ def get_applicable_lines(self, offer, basket, range=None):  # pylint: disable=re
         """
         applicable_range = range if range else self.range
 
-        if applicable_range and applicable_range.catalog_query is not None:
+        if not applicable_range and applicable_range.catalog_query is None:
```
robrap: `NOT (X AND Y) => NOT X OR NOT Y`
Should this be?:
```
if not applicable_range or (applicable_range and applicable_range.catalog_query is None):
```
Were tests failing?  If not, do we have a coverage issue?

ecommerce/extensions/offer/models.py
```diff
@@ -100,48 +101,56 @@ def get_applicable_lines(self, offer, basket, range=None):  # pylint: disable=re
...
+        site = basket.site
+        partner_code = site.siteconfiguration.partner.short_code
+        course_run_ids, course_uuids, applicable_lines = self._identify_uncached_product_identifiers(
+            applicable_lines, site.domain, partner_code, query
+        )
 
-            if course_run_ids or course_uuids:
-                # Hit Discovery Service to determine if remaining courses and runs are in the range.
+        if course_run_ids or course_uuids:
+            cache_key = get_cache_key(course_run_ids=','.join([metadata['id'] for metadata in sorted(course_run_ids)]),
```
robrap: 1. Could this be a separate method with a docstring?
2. You dropped the comment (which wasn't really great anyway).  I'm not clear what this is really doing, and what these course run ids and course uuids would be.  Maybe I need to review more of this code?
3. Consider using local variable like `course_run_id_list=','.join([metadata['id'] for metadata in sorted(course_run_ids)]` in the two calls below.

ecommerce/settings/base.py
```diff
@@ -244,6 +244,7 @@
 # Cache course info from course API.
 COURSES_API_CACHE_TIMEOUT = 3600  # Value is in seconds
 PROGRAM_CACHE_TIMEOUT = 3600  # Value is in seconds.
+CATALOG_QUERY_CONTAINS_TIMEOUT = 3600  # Value is in seconds.
```
robrap, line 247: Can you rename to `CATALOG_QUERY_CONTAINS_CACHE_TIMEOUT`?

robrap (changes requested):

ecommerce/enterprise/migrations/0004_add_enterprise_offers_for_coupons.py
BenjiLee: This was isort :/

ecommerce/extensions/offer/models.py
BenjiLee: Not sure. I think this changed the behaviour. currently looking into this now.

ecommerce/extensions/offer/models.py
BenjiLee: 1. I consider that as well. I think it will be helpful to get context of what the api is doing
2. #1 should clear this up. 
3. will do

ecommerce/settings/base.py
BenjiLee: will do

ecommerce/extensions/offer/models.py
robrap: It changed it because you had an `and` instead of an `or`.

BenjiLee: Tests were failing which is why I wasn't sure if my change was correct.

BenjiLee:
Closing PR. Looks like my cache isn't working because there is a different cache thats working correctly. 

https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/offer/models.py#L63

BenjiLee:
@robrap This PR might still be relevant. I think there's something weird with the current way this caching works today.

PR #1993, opened 2018-11-27
EG-42: Use Custom Applicator for offers everywhere

BenjiLee:
https://openedx.atlassian.net/browse/EG-42

### DIFF
--- a/ecommerce/extensions/api/v2/views/baskets.py
+++ b/ecommerce/extensions/api/v2/views/baskets.py
@@ -33,8 +33,8 @@ from ecommerce.extensions.partner.shortcuts import get_partner_for_site
 from ecommerce.extensions.payment import exceptions as payment_exceptions
 from ecommerce.extensions.payment.helpers import get_default_processor_class, get_processor_class_by_name
 
-Applicator = get_class('offer.applicator', 'Applicator')
 Basket = get_model('basket', 'Basket')
+CustomApplicator = get_class('offer.applicator', 'CustomApplicator')
 logger = logging.getLogger(__name__)
 Order = get_model('order', 'Order')
 OrderNumberGenerator = get_class('order.utils', 'OrderNumberGenerator')
@@ -377,7 +377,7 @@ class BasketCalculateView(generics.GenericAPIView):
                     basket.vouchers.add(voucher)
 
                 # Calculate any discounts on the basket.
-                Applicator().apply(basket, user=user, request=request)
+                CustomApplicator().apply(basket, user=user, request=request)
 
                 discounts = []
                 if basket.offer_discounts:
--- a/ecommerce/extensions/offer/applicator.py
+++ b/ecommerce/extensions/offer/applicator.py
@@ -1,12 +1,10 @@
 import logging
 from itertools import chain
 
-import waffle
 from oscar.apps.offer.applicator import Applicator
 from oscar.core.loading import get_model
 
 from ecommerce.enterprise.utils import get_enterprise_id_for_user
-from ecommerce.extensions.offer.constants import CUSTOM_APPLICATOR_LOG_FLAG
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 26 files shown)


ecommerce/extensions/offer/applicator.py
```diff
@@ -77,7 +63,7 @@ def get_site_offers(self):
         Return site offers that are available to baskets without bundle ids.
         """
         ConditionalOffer = get_model('offer', 'ConditionalOffer')
-        qs = ConditionalOffer.active.filter(offer_type=ConditionalOffer.SITE, condition__program_uuid__isnull=True)
+        qs = ConditionalOffer.active.filter(offer_type=ConditionalOffer.SITE)
```
BenjiLee, line 72: @Ayub-Khan @waheedahmed Is there a specific reason why `condition__program_uuid__isnull=True` needed to be added in here? One of the tests did not pass because of this. Could we delete this method altogether (since it would be same as the override)?

Ayub-Khan: @BenjiLee this is called when we are getting offers for a basket without bundle. If we remove this it means we are checking every basket for around 200 open offers.

robrap: @Ayub-Khan: This makes sense to me generally, but we will need to ensure we aren't breaking anything.  We'll need to see why we get the test failure, and see if it is an impossible situation.  Can we guarantee that we never have a program offer apply now without the bundle id?  Is there logging or anything we want to add before committing to this everywhere?

Ayub-Khan: @robrap  We use to apply program offers on baskets without bundle ids. When user have bought all other program courses(without bundle id) he would have automatically got bundle discount on last course. Or If a program had only one course we would have given it discount even without bundle id(because we were checking all program offers for every basket).

robrap:
Ownership for completing this task could move to the sustaining team.  Revenue Squad: Consider creating a sustaining ticket that gets linked to EG-42.

robrap:
@cpennington: If you are working on this, can you see if you want to bring in some ideas from this PR: https://github.com/edx/ecommerce/pull/2235/files . It is not a working PR, but should be considered.  Thank you.

ecommerce/extensions/api/v2/views/baskets.py
```diff
@@ -33,8 +33,8 @@
 from ecommerce.extensions.payment import exceptions as payment_exceptions
 from ecommerce.extensions.payment.helpers import get_default_processor_class, get_processor_class_by_name
 
-Applicator = get_class('offer.applicator', 'Applicator')
```
robrap, line 36: @cpennington: When you get back to this, I wonder if these lines should be?:
```
Applicator = get_class('offer.applicator', 'CustomApplicator')
```
I'm not sure how django-oscar allows you to override, and it makes calls to the Applicator (in various places), and we would want it to also use our custom applicator.

cpennington: This code only affects this file. Only the places in this file that use `Applicator(...)` would be affected, and in this case, there's only one use in the file, and it's already been converted to `CustomApplicator(...)`. We'd need a more invasive change to force all instances of `Applicator` to be overridden (and I'm not even sure where that change would go).

robrap: The default Applicator is very slow at this point and will only get worse as our data grows.  I highly recommend it never get used, and only the CustomApplicator get used.

Here are details for overriding core classes in django-oscar, and where we do it for other classes:

https://django-oscar.readthedocs.io/en/releases-0.5/howto/how_to_override_a_core_class.html#installed-apps-tweak

https://github.com/edx/ecommerce/blob/204f53edc5147bfa4ae5ed9f3062351ba71001e0/ecommerce/settings/_oscar.py#L21-L36

PR #1994, opened 2018-11-28
LEARNER-6787

Ayub-Khan:
Handle cybersource response status review
handle Error Reason code 104
modified logs to include notification status(so we do not have to check notification from db to understand error on splunk).

### DIFF
--- a/ecommerce/extensions/payment/processors/cybersource.py
+++ b/ecommerce/extensions/payment/processors/cybersource.py
@@ -21,6 +21,7 @@ from ecommerce.core.url_utils import get_ecommerce_url
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.payment.constants import APPLE_PAY_CYBERSOURCE_CARD_TYPE_MAP, CYBERSOURCE_CARD_TYPE_MAP
 from ecommerce.extensions.payment.exceptions import (
+    AuthorizationError,
     DuplicateReferenceNumber,
     InvalidCybersourceDecision,
     InvalidSignatureError,
@@ -234,11 +235,12 @@ class Cybersource(ApplePayMixin, BaseClientSidePaymentProcessor):
             basket (Basket): Basket being purchased via the payment processor.
 
         Raises:
+            AuthorizationError: Authorization was declined.
             UserCancelled: Indicates the user cancelled payment.
             TransactionDeclined: Indicates the payment was declined by the processor.
             GatewayError: Indicates a general error on the part of the processor.
             InvalidCyberSourceDecision: Indicates an unknown decision value.
-                Known values are ACCEPT, CANCEL, DECLINE, ERROR.
+                Known values are ACCEPT, CANCEL, DECLINE, ERROR, REVIEW.
             PartialAuthorizationError: Indicates only a portion of the requested amount was authorized.
 
         Returns:
@@ -272,19 +274,20 @@ class Cybersource(ApplePayMixin, BaseClientSidePaymentProcessor):
                 raise {
                     'cancel': UserCancelled,
                     'decline': TransactionDeclined,
-                    'error': GatewayError
+                    'error': GatewayError,
+                    'review': AuthorizationError,
                 }.get(decision, InvalidCybersourceDecision)
 
         # Raise an exception if the authorized amount differs from the requested amount.
         # Note (CCB): We should never reach this point in production since partial authorization is disabled
         # for our account, and should remain that way until we have a proper solution to allowing users to
-        # complete authorization for the entire order.
-        if response['auth_amount'] != response['req_amount']:
+        # complete authorization for the entire order
+        if response['auth_amount'] and response['auth_amount'] != response['req_amount']:
             raise PartialAuthorizationError
 
         currency = response['req_currency']
         total = Decimal(response['req_amount'])
-        transaction_id = response['transaction_id']
+        transaction_id = response.get('transaction_id', None)  # Error Notifications does not include a transaction id.
         card_number = response['req_card_number']
         card_type = CYBERSOURCE_CARD_TYPE_MAP.get(response['req_card_type'])
 
--- a/ecommerce/extensions/payment/views/cybersource.py
+++ b/ecommerce/extensions/payment/views/cybersource.py
@@ -29,7 +29,12 @@ from ecommerce.extensions.api.serializers import OrderSerializer
 from ecommerce.extensions.basket.utils import basket_add_organization_attribute
 from ecommerce.extensions.checkout.mixins import EdxOrderPlacementMixin
 from ecommerce.extensions.checkout.utils import get_receipt_page_url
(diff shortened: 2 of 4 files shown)


ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -272,19 +274,16 @@ def handle_processor_response(self, response, basket=None):
                 raise {
                     'cancel': UserCancelled,
                     'decline': TransactionDeclined,
-                    'error': GatewayError
+                    'error': GatewayError,
+                    'review': AuthorizationError,
```
robrap, line 278: Why ‘review’? Isn’t that one of the values? Should this be ‘decision’?

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -272,19 +274,16 @@ def handle_processor_response(self, response, basket=None):
...
-        # for our account, and should remain that way until we have a proper solution to allowing users to
-        # complete authorization for the entire order.
-        if response['auth_amount'] != response['req_amount']:
+        if response['auth_amount'] and response['auth_amount'] != response['req_amount']:
             raise PartialAuthorizationError
 
         currency = response['req_currency']
         total = Decimal(response['req_amount'])
-        transaction_id = response['transaction_id']
+        transaction_id = response.get('transaction_id', None)  # Error Notifications does not include a transaction id.
```
robrap, line 290: Minor typo: ‘does’ to ‘do’
When there is no transaction id, do we still have enough info to associate this properly? Also, did this result in missing responses?

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -256,15 +262,29 @@ def validate_notification(self, notification):
                     order_number
                 )
                 raise
+            except AuthorizationError:
+                logger.info(
+                    'Payment Authorization was declined for basket [%d]. The payment response was '
+                    'recorded in entry [%d].',
+                    basket.id,
```
robrap, line 269: Is there a reason why the response was not added to this message like for the others?

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -256,15 +262,29 @@ def validate_notification(self, notification):
...
+                    ppr.id,
+                )
             except PaymentError:
                 logger.exception(
```
robrap, line 273: Curious why this is treated as exception log and not info log. Is this more unexpected than the others?

robrap (commented):
Thanks.

ecommerce/extensions/payment/processors/cybersource.py
Ayub-Khan: @robrap http://apps.cybersource.com/library/documentation/dev_guides/Secure_Acceptance_Checkout_API/Secure_Acceptance_Checkout_API.pdf
This is a decision value sent by cybersource we do not handle.

ecommerce/extensions/payment/processors/cybersource.py
Ayub-Khan: Yes and yes.

ecommerce/extensions/payment/views/cybersource.py
Ayub-Khan: This happens very rarely. Response was not added because we do not need it. We can not do any thing regarding this. thats why.

ecommerce/extensions/payment/views/cybersource.py
Ayub-Khan: My opinion is that this represents gatewayerror and InvalidCybersourceDecision thats why.

ecommerce/extensions/payment/processors/cybersource.py
robrap: Oops.  I was reading the code wrong.  Ignore my comment.

However, in the CyberSource doc, they have the following text regarding their error codes:
```
... reviewed the transaction status either directly in the Business
Center or programmatically through the Single Transaction Query.
```
What is the Single Transaction Query, and is that what we could use to learn more?  Or, does that only work if we have a response and know what transaction to refer to?

ecommerce/extensions/payment/processors/cybersource.py
robrap: Cool.  Hopefully we will have fewer issues.  Thanks.

ecommerce/extensions/payment/views/cybersource.py
robrap: Ok.

ecommerce/extensions/payment/views/cybersource.py
robrap: Ok.

ecommerce/extensions/payment/processors/cybersource.py
Ayub-Khan: Single Transaction Query is like a form view which will submit some information to cybersource and cybersource will return transaction details regarding it. Kinda like reports.

robrap: @Ayub-Khan: They do mention using this "programmatically".  Does it accept the Order ID as a way to query?

Ayub-Khan: @robrap yes it does.

Ayub-Khan: Yes.

Ayub-Khan:
@uzairr  will be working on this now.

robrap:
@uzairr: Any status on this?

adeelkhan:
@robrap  Iam looking into this one. Need to know is it good to be merge, as i don't find any issue with existing implementation. Would merge it as you approve this. thanks

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -272,19 +274,16 @@ def handle_processor_response(self, response, basket=None):
                 raise {
                     'cancel': UserCancelled,
                     'decline': TransactionDeclined,
-                    'error': GatewayError
+                    'error': GatewayError,
+                    'review': AuthorizationError,
                 }.get(decision, InvalidCybersourceDecision)
 
-        # Raise an exception if the authorized amount differs from the requested amount.
-        # Note (CCB): We should never reach this point in production since partial authorization is disabled
```
robrap: Should we leave in this "Note"?

robrap:
@adeelkhan: I think this is ok to go, except the comment that was removed (starting with "Note...") seems reasonable if it is accurate.

ecommerce/extensions/payment/processors/cybersource.py
adeelkhan: This code changes doesn't alters the intent of this check. I don't know much about partial authorization, but if it was already there and with no other code changes, we should keep this intact for now.

adeelkhan:
@robrap Please review. thanks

PR #1995, opened 2018-11-28
LEARNER-6513 find frozen baskets missing payment response

zainab-amir:
Parent Ticket : https://openedx.atlassian.net/browse/LEARNER-6513
Sub task: https://openedx.atlassian.net/browse/LEARNER-6781

Description:
This management command is responsible for checking the baskets and finding out the baskets for which we submitted the payment form.

### DIFF
--- a/ecommerce/core/management/commands/find_unfulfilled_baskets.py
+++ b/ecommerce/core/management/commands/find_unfulfilled_baskets.py
@@ -1,0 +1,83 @@
+""" Django management command to find unfulfilled baskets. """
+
+import logging
+from datetime import datetime
+
+import pytz
+from django.core.management.base import BaseCommand, CommandError
+from django.db.models import Q, Subquery
+from oscar.core.loading import get_model
+
+logger = logging.getLogger(__name__)
+Basket = get_model('basket', 'Basket')
+PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
+
+
+class IncorrectRange(Exception):
+    """
+    Exception raised explicitly when end date is prior to start date
+    """
+    pass
+
+
+class Command(BaseCommand):
+    help = """
+    Management command to find unfulfilled baskets.
+
+    This management command is responsible for checking the baskets
+    and finding out the baskets for which the payment form was submitted.
+
+    Date format: yyyy-mm-dd
+    Start date should be prior to End date
+
+    Example:
+        $ ... find_unfulfilled_baskets -s 2018-11-01 -e 2018-11-02
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument('-s', '--start-date',
+                            dest='start_date',
+                            required=True,
+                            help='start date (yyyy-mm-dd)')
+        parser.add_argument('-e', '--end-date',
+                            dest='end_date',
+                            required=True,
+                            help='end date (yyyy-mm-dd)')
+
+    def handle(self, *args, **options):
+        """
+        Handler for the command
+
+        It checks for date format and range validity and then
+        calls find_unfulfilled_baskets for the given date range
+        """
+        try:
+            start_date = datetime.strptime(options['start_date'], '%Y-%m-%d')
+            end_date = datetime.strptime(options['end_date'], '%Y-%m-%d')
+            if end_date < start_date:
+                raise IncorrectRange('Incorrect date range')
+        except (ValueError, IncorrectRange):
+            logger.exception('Incorrect date format or Range')
+            raise
+
+        start_date = pytz.utc.localize(start_date)
+        end_date = pytz.utc.localize(end_date)
+
+        self.find_unfulfilled_baskets(start_date, end_date)
+
+    def find_unfulfilled_baskets(self, start_date, end_date):
+        """ Find baskets that are Frozen and unfulfilled """
+        frozen_baskets = Basket.objects.filter(status='Frozen', date_submitted=None)
+        frozen_baskets = frozen_baskets.filter(Q(date_created__range=(start_date, end_date)) |
+                                               Q(date_merged__range=(start_date, end_date)))
+        unfulfilled_baskets = frozen_baskets.exclude(id__in=Subquery(
+            PaymentProcessorResponse.objects.values_list('basket_id')))
+
+        if not unfulfilled_baskets:
+            logger.info("No unfulfilled baskets found")
+        else:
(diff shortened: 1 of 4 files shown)


ecommerce/core/management/commands/find_unfulfilled_baskets.py
```diff
@@ -0,0 +1,83 @@
+""" Django management command to find unfulfilled baskets. """
```
robrap: “Unfulfilled baskets” should be “Frozen baskets missing payment response” here and where applicable. See comments below.

ecommerce/core/management/commands/find_unfulfilled_baskets.py
```diff
@@ -0,0 +1,83 @@
...
+from django.core.management.base import BaseCommand, CommandError
+from django.db.models import Q, Subquery
+from oscar.core.loading import get_model
+
+logger = logging.getLogger(__name__)
+Basket = get_model('basket', 'Basket')
+PaymentProcessorResponse = get_model('payment', 'PaymentProcessorResponse')
+
+
+class IncorrectRange(Exception):
```
robrap: Can you switch to “InvalidDateRange”?

ecommerce/core/management/commands/find_unfulfilled_baskets.py
```diff
@@ -0,0 +1,83 @@
...
+
+    This management command is responsible for checking the baskets
+    and finding out the baskets for which the payment form was submitted.
+
+    Date format: yyyy-mm-dd
+    Start date should be prior to End date
+
+    Example:
+        $ ... find_unfulfilled_baskets -s 2018-11-01 -e 2018-11-02
+    """
```
robrap: Could you add Sample Output as well?

ecommerce/core/management/commands/find_unfulfilled_baskets.py
```diff
@@ -0,0 +1,83 @@
...
+        Handler for the command
+
+        It checks for date format and range validity and then
+        calls find_unfulfilled_baskets for the given date range
+        """
+        try:
+            start_date = datetime.strptime(options['start_date'], '%Y-%m-%d')
+            end_date = datetime.strptime(options['end_date'], '%Y-%m-%d')
+            if end_date < start_date:
+                raise IncorrectRange('Incorrect date range')
```
robrap: Please update message to match new name. Thanks.

ecommerce/core/management/commands/find_unfulfilled_baskets.py
```diff
@@ -0,0 +1,83 @@
...
+        end_date = pytz.utc.localize(end_date)
+
+        self.find_unfulfilled_baskets(start_date, end_date)
+
+    def find_unfulfilled_baskets(self, start_date, end_date):
+        """ Find baskets that are Frozen and unfulfilled """
+        frozen_baskets = Basket.objects.filter(status='Frozen', date_submitted=None)
+        frozen_baskets = frozen_baskets.filter(Q(date_created__range=(start_date, end_date)) |
+                                               Q(date_merged__range=(start_date, end_date)))
+        unfulfilled_baskets = frozen_baskets.exclude(id__in=Subquery(
```
robrap: I don’t think unfulfilled is the right term here. It’s more like frozen_baskets_missing_response, or something like that.

ecommerce/core/management/commands/find_unfulfilled_baskets.py
```diff
@@ -0,0 +1,83 @@
...
+            PaymentProcessorResponse.objects.values_list('basket_id')))
+
+        if not unfulfilled_baskets:
+            logger.info("No unfulfilled baskets found")
+        else:
+            basket_ids = "Basket ids : "
+            for unfulfilled_basket in unfulfilled_baskets:
+                basket_ids += str(unfulfilled_basket.id) + " ,"
+            logger.info("Unfulfilled baskets found : " + basket_ids)
```
robrap: How do we think this will be used? If this text should just be copied into a ticket for customer support, it would be great if we had a row for each basket with additional info (eg order id, payment provider name, and possibly the time stamp of the initiated payment.

robrap (commented):
Thanks. Please add some clarity to the ticket about the intended use.

ecommerce/core/management/commands/find_unfulfilled_baskets.py
zainab-amir: sure

ecommerce/core/management/commands/find_unfulfilled_baskets.py
Ayub-Khan: @robrap  yes, this will be copied into a customer support ticket. 
We can add order_id, Payment provider is only cybersource so i think there is no need to add it. We have no payment response so we can not figure out when the payment was made if payment was made, we can not use basket created_at as there are cases where a new basket is not created.

robrap: @Ayub-Khan:
1. We support multiple providers, and we can't guarantee that this could never occur for any provider than CyberSource.  So, even if we expect it always to be CyberSource, I think it would be good to include it.
2. I think we have a Splunk log for when the payment is initiated.  Do we store that data anywhere, or is it only in log messages?  Also, wouldn't we need to use this date for the date range of the management command?  I'm unclear why we think the dates we are using are useful enough for filtering but not reporting.
Thanks.

Ayub-Khan: @robrap We have multiple providers but we freeze basket only in case of cybersource. because for paypal user is redirected to paypal so after payment basket is updated to submited. Can you explain more how to include it(in out put?).
Our Goal is to run the job every two hours to check the orders for two hours. So this will take custom date input which will decide for how much hours we want to include. Current implementation(default behaviour) which are following example : if this runs on 4pm it will check the data between 1pm and 3pm data. which can be modified using the edx-internal command arguments.

zainab-amir:
@robrap Are we keeping this command or is it on hold like the Jenkins job too ?

robrap:
@zainab-amir: I'd put this on hold as well until the dust settles on the other pending work, like your discovery work on moving up order creation.  I think this will ultimately look different, if it is needed at all.

robrap:
@zainab-amir: I'm closing this PR for now.  If there is anything you want to document in the ticket for this PR and its status, feel free.  Thank you.

PR #1996, opened 2018-11-28
PROD-1274: Custom Applicator ADR

BenjiLee:


docs/decisions/0003-custom-applicator.rst
```diff
@@ -0,0 +1,31 @@
+2. Use Custom Applicator for Applying Offers
+------------------------------------------------------------
+
+Status
+------
+
+Ran away
```
robrap: Maybe "Draft" would work. ;)

docs/decisions/0003-custom-applicator.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Context
+-------
+
+When a learner has wants to purchase a specific set of courses known as a program, the learner should get a discount.
```
robrap: I think we need to do this for several types of offers (e.g. enterprise, journals, etc.), and not have this just be about programs.  Program offers is just the first one that was tackled.

There are business requirements that need to be worked out, like:
1. Can we state a decision around supporting a maximum of one offer, or do we need to support multiple offers?
2. Can we clearly state the priority of different offers (e.g. enterprise before program, etc.) so that we can appropriately implement the Custom Applicator.

docs/decisions/0003-custom-applicator.rst
```diff
@@ -0,0 +1,31 @@
...
+Decision
+--------
+
+Rather than attempting to apply every offer to a basket, we should override the `get_site_offers` method in the
+Applicator class as recommended by Oscar.
+
+Consequences
+------------
+
+
```
robrap, line 32: Section needs to be completed, detailing the performance gains of not processing every program offer for each user, and instead looking at the single possibly applicable offer.

robrap (commented):
These are starter comments for the Revenue Squad to take over.

docs/decisions/0003-custom-applicator.rst
robrap: New update: It looks like there could be bugs related to not using the Custom Applicator (differences between the two), so this ADR should probably be updated to reflect that we should only use the Custom Applicator.

docs/decisions/0003-custom-applicator.rst
robrap: UPDATE: This ADR should clarify:
1. That the Custom Applicator will encapsulate business logic, like the order of precedence of certain competing offers.  It should mention that the CustomApplicator comments should be kept up to date with the full set of business logic decisions as needed.
2. Because the CustomApplicator encapsulates business logic, it should be the only applicator used throughout ecommerce, including code in django-oscar that uses the configured applicator.  Otherwise, we will have bugs as different offers are applied by different code.

waheedahmed:
Hi @robrap, I have updated the ADR with suggested changes. Could you please take a look?

docs/decisions/0003-custom-applicator.rst
```diff
@@ -0,0 +1,46 @@
...
+Decision
+--------
+
+Rather than attempting to apply every offer to a basket, we should override the `get_site_offers` method in the
+Applicator class as recommended by Oscar and use it throughout Ecommerce and django-oscar as default applicator.
+
+Consequences
+------------
+
+Use `CustomApplicator`_ as default applicator throughout Ecommerce and django-oscar, otherwise we will have bugs as
```
robrap: Replace 
```
`CustomApplicator`_ 
```
with `the custom applicator`.

docs/decisions/0003-custom-applicator.rst
```diff
@@ -0,0 +1,46 @@
...
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+This custome applicator will encapsulate bussiness logic to prefilter the programs and enterprise offers first.
```
robrap: The "Context" section is currently part Context and part Decision.  What do you think of the following suggestions?:
1. Move the "This custome applicator will encapsulate bussiness logic..." and "This should be the only applicator used throughout Ecommerce..." paragraphs to the "Decision" section.  Maybe put the paragraph that is currently in the "Decision" section between these two paragraphs?
2. Try the following for the "Context" section:
```
Oscar uses an Applicator object to apply offers to a basket.  The `default behaviour of Oscar's Applicator`_ is to look for all offers available in the system, and then attempt to apply each offer that fulfills the requirements of the basket.

This default behavior was causing performance issues by looking at many Program offers, when it was clear ahead of time that only 0 or 1 of them could possibly apply.

Additionally, after creating a custom applicator with our own logic, it provided different results than the default applicator.  When we did not use it consistently throughout the system, we saw bugs as different offers were applied by different part of the code base.

.. _default behaviour of Oscar's Applicator: https://github.com/django-oscar/django-oscar/blob/40a4cacc27223ac675f5e859e7568b632e3f304c/src/oscar/apps/offer/applicator.py#L46-L61
```

robrap: For at least one use of the text "custom applicator", change it to:
```
`custom applicator`_
```
And add:
```
.. _customer applicator: https://<ADD-LINK-TO-OUR-CUSTOM-APPLICATOR-FILE>

File: docs/decisions/0003-custom-applicator.rst [hunk 6]
```diff
@@ -0,0 +1,46 @@
...
+different offers are applied by different code.
+
+We noticed some bugs in which learners were seeing different prices and discounts on program about pages from actual
+basket page because we were using django-oscar applicator on calculate basket endpoint but the basket enpoint was using
+our custom applicator. We also recieved some reports where no offer was applied on basket but the payment processor code
+applied an offer on the actual payment due to django-oscar applicator and learners were chanrged less then the actual
+price.
+
+
+References
```
robrap: You can delete this section if you aren't using it.

docs/decisions/0003-custom-applicator.rst
```diff
@@ -0,0 +1,46 @@
...
+Accepted
+
+Context
+-------
+
+This custome applicator will encapsulate bussiness logic to prefilter the programs and enterprise offers first.
+Otherwise, gets the site offers not associated with any bundle or enterprise customer rather than blindly returning
+every offer, including ones that could never apply.
+
+This should be the only applicator used throughout Ecommerce including django-oscar that uses the configured applicator.
```
robrap: According to OEP-19, we no longer use hard line breaks.  Lines 16-18 should all be on one long line (16) and your editor should just wrap for you.  Please fix this throughout.

robrap (approved):
1. Looks good.  Thank you.  Some minor changes requested.
2. There are some spelling error typos like `custome` and `bussiness`.  Can you copy into a spellchecker?
Feel free to merge after adjusting, unless you want a rereview.  Thank you!

ecommerce/extensions/offer/applicator.py
```diff
@@ -40,15 +40,14 @@ def get_offers(self, basket, user=None, request=None, bundle_id=None):  # pylint
...
         Returns all offers to apply to the basket.
 
-        Does prefiltering to filter out offers that could never apply to a
-        particular basket. As an example, if the basket has a bundle,
-        i.e. a program, gets only the site offers associated with that specific bundle,
-        rather than all site offers. Otherwise, gets the site offers not associated
+        Does prefiltering to filter out offers that could never apply to a particular
+        basket. As an example, if the basket has a bundle ID or an enterprise customer
+        UUID, gets only the site offers associated with that specific bundle or enterprise
+        customer, rather than all site offers. Otherwise, gets the site offers not associated
```
robrap, line 50: In the class docstring, does it make sense to add something like?:
```
This extension Applicator will be used when you use the following::
 
 Applicator = get_class('offer.applicator', 'Applicator')
```
I know this is no different than any other class (like Basket), so maybe not needed, but may not hurt. :)

robrap (commented):
One more thought.

waheedahmed:
Thank you @robrap!

I'll address all the suggested changes and let you know if needed another review.

PR #1997, opened 2018-11-29
ENT-1287. Additional coupon usage field for multi times/one customer

muhammad-ammar:
[ENT-1287](https://openedx.atlassian.net/browse/ENT-1287) 

**Description:**
This updates the coupon usage field with fourth choice(`Can be used multiple times by one customer`) only for enterprise coupons. 

**Testing Instructions:**
Deployed to Sandbox

0. Enable the `enable_enterprise_offers_for_coupons` switch
1. Goto https://ecommerce-business.sandbox.edx.org/enterprise/coupons/new/ 
2. You should see the `Can be used multiple times by one customer` option in usage dropdown list. This option should not be available for original coupons screen [here](https://ecommerce-business.sandbox.edx.org/coupons/new/)
3. Create an enterprise coupon and it should be created without any issue.
4. You should see the newly created enterprise coupon on coupon list page.


ecommerce/extensions/voucher/utils.py
```diff
@@ -581,8 +583,11 @@ def create_enterprise_vouchers(
         end_datetime,
         start_datetime)
 
+    voucher_types = (Voucher.MULTI_USE, Voucher.ONCE_PER_CUSTOMER, MULTI_USE_PER_CUSTOMER_VOUCHER)
+    quantity = int(quantity)
```
muhammad-ammar: Note to reviewers: This conversion to `int` will be handled in https://github.com/edx/ecommerce/pull/1969. Doing here just to avoid the crash.

muhammad-ammar:
@georgebabey @brittneyexline This is ready for review. Frontend side is completed but I am not sure about what else we need to do on the backend side?

ecommerce/extensions/voucher/utils.py
brittneyexline: nice catch, i have updated my PR 👍

brittneyexline (commented):
looks good! there are no backend changes at this point since it depends on the assignment logic. we are backed up with our PRs right now (as you probably know) so we'll have to hold off on merging for now.

ecommerce/extensions/voucher/utils.py
```diff
@@ -46,6 +46,8 @@
 Voucher = get_model('voucher', 'Voucher')
 VoucherApplication = get_model('voucher', 'VoucherApplication')
 
+MULTI_USE_PER_CUSTOMER_VOUCHER = 'Multi-use-per-Customer'
```
brittneyexline: now that i have started the code assignment, it occurs to me that if this value is sent through the UI, it will fail to be stored in the database because the field is an enum on voucher defined by oscar (https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/voucher/abstract_models.py#L133).
i am wondering if we should add some sort of flag of our own to the voucher model to indicate this instead of trying to add the enum option explicitly to the model, because i suspect it will break in unexpected ways with django oscar's default behavior.
all this is to say that i was wrong when i said there shouldn't be backend changes, but i think the goal for the changes would be that saving a coupon with this option doesn't break, it saves some data indicating the option, and for now behaves the same as the multi use voucher. does that make sense?

muhammad-ammar: @brittneyexline I have updated the `Voucher` model with new choice and also updated the `is_available_to_user` method. Please have a look.

ecommerce/extensions/voucher/models.py
```diff
@@ -28,6 +29,26 @@ class OrderLineVouchers(models.Model):
...
+    USAGE_CHOICES = (
+        (SINGLE_USE, _("Can be used once by one customer")),
+        (MULTI_USE, _("Can be used multiple times by multiple customers")),
+        (ONCE_PER_CUSTOMER, _("Can only be used once per customer")),
+        (MULTI_USE_PER_CUSTOMER, _("Can be used multiple times by one customer")),
+    )
+    usage = models.CharField(_("Usage"), max_length=128,
+                             choices=USAGE_CHOICES, default=MULTI_USE)
+
+    def is_available_to_user(self, user=None):
```
brittneyexline, line 43: because this field is an oscar default, i think adding a new option will probably break things in places where we are relying on oscar's default behavior. i see that you are accounting for that here, but do you know if this is the only place where we need to modify oscar's default behavior to avoid breaking things? we'll also have to ensure it doesn't break if we upgrade oscar in the future...

otherwise, it might be better to capture this data by adding a new nullable boolean field on the model that in conjunction with usage being multi use, would indicate that it should be assigned/redeemed by one user.

robrap: Great point @brittneyexline.  Before move forward, please point to Oscar docs regarding how they recommend extending like this.  Oscar is meant to be extendable in many different ways, and sometimes it isn't clear from the docs.  If it isn't, please feel free to ask that community questions.  I recently learned in another area that there are sometimes common patterns that are not well documented.

muhammad-ammar: @brittneyexline Yes, I checked that and `is_available_to_user` method is the only place that needs to be updated.

> because this field is an oscar default, i think adding a new option will probably break things in places where we are relying on oscar's default behavior.

Regarding the above, we are using `is_available_to_user` in many places and as per my understanding it should be updated to reflect the new usage type. And if we add a new nullable boolean field, even than we must need to update the `is_available_to_user` method so that it can return the correct result when usage type is `Can be used multiple times by one customer`

I have also looked into oscar docs for [model customization](http://docs.oscarcommerce.com/en/latest/howto/how_to_customise_models.html) and we are doing things like they recommend.

robrap: Thanks @muhammad-ammar.  @brittneyexline: I'll leave this to your review, but could you add the above to any decision records that you or @georgebabey can share with the guild?  Thank you.

brittneyexline: thanks @muhammad-ammar for the confirmation! @robrap - we don't have an ADR yet but we'll include this when we write it.

brittneyexline (commented):
looks good so far, next thing that would be good to see is some test instructions and an environment to see it in action!

muhammad-ammar:
@brittneyexline I have updated the business sandbox with changes in this PR and also updated the PR description.

brittneyexline (approved):
nice work!

ecommerce/extensions/voucher/migrations/0006_auto_20181205_1017.py
```diff
@@ -0,0 +1,20 @@
...
+        ('voucher', '0005_auto_20180124_1131'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='voucher',
+            name='usage',
+            field=models.CharField(choices=[(b'Single use', 'Can be used once by one customer'), (b'Multi-use', 'Can be used multiple times by multiple customers'), (b'Once per customer', 'Can only be used once per customer'), (b'Multi-use-per-Customer', 'Can be used multiple times by one customer')], default=b'Multi-use', max_length=128, verbose_name='Usage'),
+        ),
+    ]
```
muhammad-ammar, line 20: @brittneyexline Yesterday I forgot to update this PR with the latest changes which includes this migration file. But the sandbox was updated with the latest code including the migration file.

muhammad-ammar:
@georgebabey Can you please merge this?

PR #1998, opened 2018-11-29
ENT-1362 api for sending emails to assigned learners

irfanuddinahmad:
Description: This PR provides three functions briefly described below. The Jira link below provides more details.
Jira: https://openedx.atlassian.net/browse/ENT-1362

1: Utility function to send an asynchronous coupon code assignment email via ecommerce-worker to learners (ecommerce.extensions.offer.utils.send_assigned_offer_email)
2: API for receiving the status of the email from ecommerce-worker. The API handler updates the offer_assignment model with the email status. 
(POST /ecommerce/api/v2/assignment-email/status)
3: API for receiving email bounce status from Sailthru. The API handler updates the offer_assignment model with the email status
(POST /ecommerce/api/v2/assignment-email/bounce)

***Message formats***

The message signatures for the API are described below:
```
 POST /ecommerce/api/v2/assignment-email/status
        Requires a JSON object of the following format:
       {
            'offer_assignment_id': 555,
            'send_id': 'XBEn85WnoQJsIhk6'
            'status': 'success'
        }
        Returns a JSON object of the following format:
       {
            'offer_assignment_id': 555,
            'send_id': 'XBEn85WnoQJsIhk6'
            'status': 'updated'
            'error': ''
        }
        Keys:
        *offer_assignment_id*
            Primary key of the entry in the offer_assignment model.
        *status*
            The offer_assignment model update status
        *error*
            Error detail. Empty on a successful update.



 POST /ecommerce/api/v2/assignment-email/bounce
        Requires a JSON object of the following format:
        {
            'email': blashsdsd@dfsdf.com
            'send_id': WFQKQW5K3LBgi0mk
            'action': hardbounce
            'api_key': 6b805755ce5a23e3c0459aaa598efc56
            'sig': 13b3987b656c7771fcac92982fdfb331
        }
        Returns HTTP_200_OK

        Keys:
        *email*
            Email of the customer whose email bounced.
        *send_id*
            The unique identifier of the bounced message.
        *action*
            hardbounce: Email bounced.
        *api_key*
            Key provided by Sailthru.
        *sig*
            Hash of API key and all parameter values for the postback call

### CODE REVIEW

**Comment by robrap** [comment]
1. Please update names and/or comments to make it more clear what this is for.  What is an "assignment email"?
2. Please ensure to get this reviewed by @macdiesel or someone regarding GDPR and handling of PII data (e.g. email).

File: ecommerce/extensions/api/v2/urls.py [hunk 1]
```diff
@@ -86,6 +87,12 @@
     url(r'^customers$', enterprise_views.EnterpriseCustomerViewSet.as_view(), name='enterprise_customers'),
 ]
 
+ASSIGNMENT_EMAIL_URLS = [
+    url(r'^template$', assignment_email.AssignmentEmail.as_view(), name='get_template'),
+    url(r'^updatestatus$', assignment_email.AssignmentEmailStatus.as_view(), name='update_status'),
```
georgebabey: I think this should just be `status`, including the verb doesn't make it feel very REST

ecommerce/extensions/api/v2/urls.py
```diff
@@ -86,6 +87,12 @@
...
+    url(r'^receivebounce$', assignment_email.AssignmentEmailBounce.as_view(), name='receive_bounce')
```
georgebabey: This could also just be `bounce`

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,157 @@
...
+        }
+        Returns a JSON object of the following format:
+       {
+            'offer_id': '555',
+            'send_id': 'XBEn85WnoQJsIhk6'
+            'status': 'updated'
+            'error': ''
+        }
+        Keys:
+        *offer_id*
```
brittneyexline: i would call this offer_assignment_id to avoid confusion with potentially referring to the offer object itself

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,157 @@
...
+                    'status': 'updated',
+                    'error': ''
+                }
+            except OfferAssignment.DoesNotExist as exc:
+                logger.exception('[Offer Assignment] AssignmentEmailStatus update raised: %r', exc)
+                extra_status = {
+                    'status': 'failed',
+                    'error': str(exc)
+                }
+        update_status.update(extra_status)
```
brittneyexline, line 107: what is `extra_status` if `email_status != 'success'`?

ecommerce/extensions/offer/models.py
```diff
@@ -481,4 +481,9 @@ class OfferAssignment(TimeStampedModel):
     )
 
 
+class OfferAssignmentEmail(models.Model):
```
brittneyexline: i would call this `OfferAssignmentEmailAttempt` to be a little clearer. also not sure if we should attempt to think about other email mechanisms besides sailthru and make this an abstract model that is inherited by a Sailthru specific model? could be overkill right now though...

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,157 @@
...
+from rest_framework.views import APIView
+
+from ecommerce.extensions.offer.constants import OFFER_ASSIGNED, OFFER_ASSIGNMENT_EMAIL_BOUNCED
+from ecommerce.extensions.offer.models import OfferAssignment, OfferAssignmentEmail
+
+logger = logging.getLogger(__name__)
+
+
+class AssignmentEmail(APIView):
+    """Sends assignment email(s) to Learners."""
```
brittneyexline: might be useful to explain for each of these Views who is expected to call these apis

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,157 @@
...
+    permission_classes = (IsAuthenticated,)
```
brittneyexline: should this also include `IsAdminUser`?

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,157 @@
...
+                extra_status = {
+                    'status': 'failed',
+                    'error': str(exc)
+                }
+        update_status.update(extra_status)
+        return Response(update_status, status=status.HTTP_200_OK)
+
+
+class AssignmentEmailBounce(APIView):
+    """Receive Sailthru bounce-api POST"""
```
brittneyexline: if this is going to be received through Sailthru, will we need to do anything special to make sure this endpoint is accessible from outside our usual services?

brittneyexline (commented):
looking good so far!

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: @georgebabey This would require changes to the ecommerce-worker. Please advise.

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: The API is only called when email is a success, otherwise the status in offer_assignment model remains EMAIL_PENDING. Should we add an EMAIL_FAILED status to the offer_assignment model?

ecommerce/extensions/offer/models.py
irfanuddinahmad: The plan is to enhance edx-ace https://github.com/edx/edx-ace 
We could not use it here since it does not have any bounce functions at this time.

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Not sure. @georgebabey Please advise.
ecommerce-worker uses the following user:
'''
service_username = get_configuration('ECOMMERCE_SERVICE_USERNAME', site_code=site_code)
'''
One of the ecommerce apis used by ecommerce-worker /api/v2/orders has the following:
permission_classes = (IsAuthenticated, IsStaffOrOwner, DjangoModelPermissions,)
throttle_classes = (ServiceUserThrottle,)

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: @georgebabey You mentioned that the Amazon API Gateway will proxy this url. Should I add a DevOps ticket for that?

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Fixed! ... good catch!

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/urls.py
```diff
@@ -97,6 +104,7 @@
     url(r'^refunds/', include(REFUND_URLS, namespace='refunds')),
     url(r'^retirement/', include(RETIREMENT_URLS, namespace='retirement')),
     url(r'^sdn/', include(SDN_URLS, namespace='sdn')),
+    url(r'^assignmentemail/', include(ASSIGNMENT_EMAIL_URLS, namespace='assignmentemail')),
```
georgebabey: I think `assignment-email` would be more readable. hyphens for urls

ecommerce/extensions/api/v2/views/assignmentemail.py
georgebabey: I dont know if we need `IsAdminUser` but I do think adding `IsStaffOrOwner` would protect this endpoint from abuse

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,163 @@
...
+                     'Enrollment url: {enrollment_url}'
+                     'Access Code: {code}'
+                     'Expiration date: {code_expiration_date}'
+
+                     'You may go directly to the Enrollment URL to view courses that are available for this code'
+                     ' or you can insert the access code at check out under "coupon code" for applicable courses.'
+
+                     'For any questions, please reach out to your Learning Manager.')}
+        """
+        email_template = ('Your learning manager has provided you with a new access code to take a course at edX.'
```
georgebabey: Thinking we may want to pull this into config so we could potentially have a different email for stage vs prod. Also would allow Open edX installs to more easily configure this default

ecommerce/extensions/api/v2/views/assignmentemail.py
georgebabey: Agree with @brittneyexline here. Given the similarity of model names and likely confusion let's update this and ecomm-worker. You even call it `offer_assignment_id` on the first line below

ecommerce/extensions/api/v2/views/assignmentemail.py
georgebabey: We don't need to go through the API Gateway, we have many examples where we do not and it has not been widely adopted yet.  This should be accessible as our frontends are also hitting the backend APIs, Sailthru should be able to.

We should think through what would happen if a bad actor hit this endpoint. I think they'd be able to muck with the data since the IDs are auto-incrementing.  Therefore we should think how we can limit the access. Does Sailthru publish IP blocks we could whitelist? Lets chat about this with devops before this lands.

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,163 @@
...
+                    'status': 'failed',
+                    'error': str(exc)
+                }
+        update_status.update(extra_status)
+        return Response(update_status, status=status.HTTP_200_OK)
+
+
+class AssignmentEmailBounce(APIView):
+    """Receive Sailthru bounce-api POST"""
+    # Note: Will use a proxy api to secure this endpoint
```
georgebabey: We need to figure out how to secure this - whether through a proxy or some other way.

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,163 @@
...
+    permission_classes = ()
```
georgebabey, line 117: Lets make sure we we add a big comment here why we're removing base permission class.

requirements/test.txt
```diff
@@ -5,6 +5,7 @@
 #    make upgrade
 #
 -e git+https://github.com/edx/analytics-python.git@1.2.11#egg=analytics-python==1.2.11
+-e git+https://github.com/edx/ecommerce-worker.git@master#egg=ecommerce-worker==master
```
georgebabey: Need to publish a version of the worker so we don't pull from master.

georgebabey (changes requested):

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: @georgebabey Are you referring to the edx-configuration repo?

https://github.com/edx/configuration/blob/0a413f7ba6a8b61e9d170f9fa9892aff3967f08e/playbooks/roles/ecommerce/defaults/main.yml#L121

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done. The companion PR in ecommerce-worker is at:
https://github.com/edx/ecommerce-worker/pull/39

ecommerce/extensions/offer/models.py
irfanuddinahmad: Name changed.

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: IsStaffOrOwner permission added.

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Unfortunately, I could not find any IP white-list in the docs. The endpoint receives an already created send_id which identifies the message to be updated. The auto increment ID is already created by the ecommerce-worker's call to /api/v2/assignment-email/status

ecommerce/extensions/api/v2/urls.py
irfanuddinahmad: Done.

ecommerce/extensions/api/v2/urls.py
irfanuddinahmad: Done.

ecommerce/extensions/api/v2/urls.py
irfanuddinahmad: Done.

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Added.

requirements/test.txt
irfanuddinahmad: Waiting for https://github.com/edx/ecommerce-worker/pull/39 to be pulled into master.

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        offer_assignment_id = int(request.data.get('offer_assignment_id'))
+        send_id = request.data.get('send_id')
+        email_status = request.data.get('status')
+        update_status = {
+            'offer_assignment_id': offer_assignment_id,
+            'send_id': send_id,
+        }
+        extra_status = {}
+        if email_status == 'success':
```
brittneyexline: should we consider logging a warning if email status is not success? it seems like you are expecting only this value to come back

irfanuddinahmad: @brittneyexline You are right ... the ecommerce-worker only calls the api when the Sailthru call is successful, otherwise it logs an error. Since our default status in the offer_assignment model is 'EMAIL_PENDING', we can do away with the status check altogether. The only reason for keeping it would be if we were to introduce another status in offer_assignment e.g. EMAIL_FAILED ...  ... what do you think?

brittneyexline: if something changes, we'll have to update this anyway, so if it makes sense to assume this only gets called upon success, then we can just log a warning if the format of the payload is unexpected.

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,166 @@
...
+       {
+            'offer_assignment_id': '555',
+            'send_id': 'XBEn85WnoQJsIhk6'
+            'status': 'updated'
+            'error': ''
+        }
+        Keys:
+        *offer_assignment_id*
+            Primary key of the entry in the offer_assignment model.
+        *status*
```
muhammad-ammar, line 69: `send_id` needs to be added to docstring

brittneyexline (approved):
from a code perspective, i think this is ready to go. the only small thing i'd like to see is more clarity on the AssignmentEmailStatus view when email_status is not 'success' (add a test?).
so, once you get the tests resolved, which i assume is due to ecommerce-worker, i think we can merge this.

ecommerce/extensions/api/v2/views/assignmentemail.py
georgebabey: Local project django settings would be fine

irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+
+from ecommerce.extensions.api.permissions import IsStaffOrOwner
+from ecommerce.extensions.offer.constants import OFFER_ASSIGNED, OFFER_ASSIGNMENT_EMAIL_BOUNCED
+from ecommerce.extensions.offer.models import OfferAssignment, OfferAssignmentEmailAttempt
+
+logger = logging.getLogger(__name__)
+
+
+class AssignmentEmail(APIView):
+    """Coupon code assignment email related functions"""
```
muhammad-ammar: This view only has one endpoint to get the template so can update the docstring according to the endpoint?

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+        return Response(
+            status=status.HTTP_200_OK,
+            data={'template': email_template}
+        )
+
+
+class AssignmentEmailStatus(APIView):
+    """
+    This api is called from ecommerce-worker to update assignment email status
+    in offer_assignment and offer_assignment_email model.
```
muhammad-ammar: Can we update the model names in docstring like  `offer_assignment` -> `OfferAssignment`?

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+            similar_offers = OfferAssignment.objects.select_for_update().filter(
+                user_email=assigned_offer.user_email,
+                code=assigned_offer.code
+            )
+            similar_offers.update(status=OFFER_ASSIGNED)
+            OfferAssignmentEmailAttempt.objects.create(offer_assignment=assigned_offer, send_id=send_id)
+
+    def post(self, request):
+        """
+        POST /ecommerce/api/v2/assignment-email/status
```
muhammad-ammar, line 50: Can we add bit more explanation about what this endpoint does?

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def update_email_status(self, offer_assignment_id, send_id):
+        """Update the OfferAssignment and OfferAssignmentEmailAttempt model"""
+        assigned_offer = OfferAssignment.objects.get(id=offer_assignment_id)
+        with transaction.atomic():
+            similar_offers = OfferAssignment.objects.select_for_update().filter(
+                user_email=assigned_offer.user_email,
+                code=assigned_offer.code
+            )
+            similar_offers.update(status=OFFER_ASSIGNED)
```
muhammad-ammar: this can be re-written as 
```python
OfferAssignment.objects.select_for_update().filter(
    user_email=assigned_offer.user_email,
    code=assigned_offer.code
).update(status=OFFER_ASSIGNED)
```

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+        }
+        Keys:
+        *offer_assignment_id*
+            Primary key of the entry in the offer_assignment model.
+        *status*
+            The offer_assignment model update status
+        *error*
+            Error detail. Empty on a successful update.
+        """
+        offer_assignment_id = int(request.data.get('offer_assignment_id'))
```
muhammad-ammar: `offer_assignment_id` is an integer, so why we need to convert it to `int`?

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+        send_id = request.data.get('send_id')
+        email_status = request.data.get('status')
+        update_status = {
+            'offer_assignment_id': offer_assignment_id,
+            'send_id': send_id,
+        }
+        extra_status = {}
+        if email_status == 'success':
```
muhammad-ammar: Ideally we should have enum for these('success', `updated`) constants something like

```python
class OfferAssignmentEmailStatus(object):
    SUCCESS = 'success'
    UPDATED = 'updated'
```
And then we should use this enum everywhere we need the statuses. 
```python
if email_status == OfferAssignmentEmailStatus.SUCCESS:
```

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+        else:
+            logger.error(
+                '[Offer Assignment] AssignmentEmailStatus could not update status for OfferAssignmentId: %d',
+                offer_assignment_id)
+            extra_status = {
+                'status': 'failed',
+                'error': 'Incorrect status'
+            }
+        update_status.update(extra_status)
+        return Response(update_status, status=status.HTTP_200_OK)
```
muhammad-ammar, line 108: We are always sending the `200` even in case of error, this means that endpoint consumer always need to check the response to see if there is an error or not. Ideally endpoint should send 4XX in case of an error.

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+    def update_email_status(self, send_id):
+        """Update the OfferAssignment model"""
+        offer = OfferAssignmentEmailAttempt.objects.get(send_id=send_id)
+        assigned_offer = OfferAssignment.objects.get(id=offer.offer_assignment.id)
+        with transaction.atomic():
+            similar_offers = OfferAssignment.objects.select_for_update().filter(
+                user_email=assigned_offer.user_email,
+                code=assigned_offer.code
+            )
+            similar_offers.update(status=OFFER_ASSIGNMENT_EMAIL_BOUNCED)
```
muhammad-ammar: the above can be re-written as 
```python
OfferAssignment.objects.select_for_update().filter(
    user_email=assigned_offer.user_email,
    code=assigned_offer.code
).update(status= OFFER_ASSIGNMENT_EMAIL_BOUNCED)
```

ecommerce/extensions/offer/utils.py
```diff
@@ -99,3 +100,41 @@ def render_email_confirmation_if_required(request, offer, product):
...
+        email_body = template.format(
+            code_usage_count=code_usage_count,
+            user_email=learner_email,
+            enrollment_url=enrollment_url,
+            code=code,
+            code_expiration_date=code_expiration_date
+        )
+        send_offer_assignment_email.delay(learner_email, offer_assignment_id, email_subject, email_body)
+    except Exception:  # pylint: disable=broad-except
+        return False
```
muhammad-ammar, line 151: I think we should log the exception, this will help in case of an error.

ecommerce/extensions/offer/utils.py
```diff
@@ -99,3 +100,41 @@ def render_email_confirmation_if_required(request, offer, product):
...
+        *code_usage_count*
+            Number of times the code can be redeemed.
+        *code_expiration_date*
+            Date till code is valid.
+
+    Returns:
+         True when successful or False in case of any exception
+    """
+
+    email_subject = 'New edX course assignment'
```
muhammad-ammar: Should we make this configurable?

ecommerce/extensions/offer/models.py
```diff
@@ -481,4 +481,9 @@ class OfferAssignment(TimeStampedModel):
     )
 
 
+class OfferAssignmentEmailAttempt(models.Model):
```
muhammad-ammar, line 484: Can we add a comment to explain why we need this model?

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+        return Response(update_status, status=status.HTTP_200_OK)
+
+
+class AssignmentEmailBounce(APIView):
+    """
+    Receive Sailthru bounce-api POST
+    Note: The endpoint is configured without security since Sailthru does not have any configuration options for
+    adding security tokens. Please refer to https://getstarted.sailthru.com/developers/api-basics/postbacks/
+    """
+    permission_classes = ()
```
muhammad-ammar, line 117: Currently this endpoint has no authentication and verification so anyone can make a hit to this endpoint and try to update the database.

It would be great if we can add post verification as mentioned [here](https://getstarted.sailthru.com/developers/api-basics/postbacks/). We don't need to do this now as a part of this PR but we should definitely look this in future.

ecommerce/extensions/api/v2/tests/views/test_assignmentemail.py
```diff
@@ -0,0 +1,240 @@
...
+from testfixtures import LogCapture
+
+from ecommerce.extensions.offer.constants import OFFER_ASSIGNED, OFFER_ASSIGNMENT_EMAIL_BOUNCED
+from ecommerce.extensions.offer.models import OfferAssignment, OfferAssignmentEmailAttempt
+from ecommerce.extensions.test import factories
+from ecommerce.tests.testcases import TestCase
+
+
+@ddt.ddt
+class AssignmentEmailTests(TestCase):
```
muhammad-ammar: `AssignmentEmailTests` -> `AssignmentEmailTemplateTests`

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,173 @@
...
+from rest_framework.views import APIView
+
+from ecommerce.extensions.api.permissions import IsStaffOrOwner
+from ecommerce.extensions.offer.constants import OFFER_ASSIGNED, OFFER_ASSIGNMENT_EMAIL_BOUNCED
+from ecommerce.extensions.offer.models import OfferAssignment, OfferAssignmentEmailAttempt
+
+logger = logging.getLogger(__name__)
+
+
+class AssignmentEmail(APIView):
```
muhammad-ammar: `AssignmentEmail` -> `AssignmentEmailTemplate` because this endpoint only handles template thing.

ecommerce/extensions/api/v2/tests/views/test_assignmentemail.py
```diff
@@ -0,0 +1,240 @@
...
+    """ Tests for AssignmentEmail API view. """
```
muhammad-ammar: `Tests for AssignmentEmail API view.` -> `Tests for AssignmentEmailTemplate API view.`

muhammad-ammar (changes requested):
@irfanuddinahmad Done with first pass of the review. Let me know once it is ready for next review.

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/offer/models.py
irfanuddinahmad: Done

ecommerce/extensions/offer/utils.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Added message md5 hash verification with sailthru key and secret

requirements/test.txt
irfanuddinahmad: Done

irfanuddinahmad:
@muhammad-ammar I have incorporated most of your suggestions. Please review.

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/tests/views/test_assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/tests/views/test_assignmentemail.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        email_template = settings.OFFER_ASSIGNMENT_EMAIL_DEFAULT_TEMPLATE
+
+        return Response(
+            status=status.HTTP_200_OK,
+            data={'template': email_template}
+        )
+
+
+class OfferAssignmentStatus(object):
```
muhammad-ammar: `OfferAssignmentStatus` -> `OfferAssignmentEmailStatus `

muhammad-ammar (approved):
@irfanuddinahmad Looks good to me.

Currently AWS SES does not work on sandbox so we can't test changes in this PR. My only concern is that it would be nice if we can test the changes on sandbox before merging.

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,210 @@
...
+        received_data = {
+            'email': request.data.get('email'),
+            'send_id': send_id,
+            'action': request.data.get('action')
+        }
+        received_hash = request.data.get('sig')
+        sailthru_secret = settings.SAILTHRU_SECRET
+        sailthru_key = settings.SAILTHRU_KEY
+        if sailthru_secret and sailthru_key:
+            hash_string = sailthru_secret + sailthru_key + 'json' + json.dumps(received_data)
```
feanil, line 165: If the sailthru keys need to get rotated, all the hashes will mismatch, so signatures won't match for anything that is being sent to sailthru during the time when the rotation is occuring.  It seems like we wouldn't want that.  Should you be using a different token to salt the hash?  Like a new sailthru salt setting?

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,210 @@
...
+            computed_hash = hashlib.md5(hash_string).hexdigest()
```
feanil, line 166: There are lots of places that use openEdx which have stringent security requirements about not using MD5 for anything.  I think you should use sha256 instead as this makes the hash more secure.

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,210 @@
...
+        """Update the OfferAssignment model"""
+        offer = OfferAssignmentEmailAttempt.objects.get(send_id=send_id)
+        assigned_offer = OfferAssignment.objects.get(id=offer.offer_assignment.id)
+        with transaction.atomic():
+            OfferAssignment.objects.select_for_update().filter(
+                user_email=assigned_offer.user_email,
+                code=assigned_offer.code
+            ).update(status=OFFER_ASSIGNMENT_EMAIL_BOUNCED)
+
+    def post(self, request):
```
feanil, line 130: Can we make this endpoint rate limited?  Do we know the rate at which we expect sailthru to hit this endpoint?

irfanuddinahmad: @feanil Assuming that we send out ~100000 emails in 24 hours and again assuming that 1% bounce, that would entail approx. 1000 calls to this api in those 24 hours. However, we should get the exact stats by monitoring the api in AWS for a month.

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: @feanil This api handles postbacks 'from' Sailthru. Messages are 'sent' via edx-ecommerce-worker which already has the following two settings:
ECOMMERCE_WORKER_SAILTHRU_KEY
ECOMMERCE_WORKER_SAILTHRU_SECRET

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: @feanil You are correct, however, Sailthru is sending us the postback message hash in MD5 format

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: UserRateThrottle is globally defined. Default rate in edx-ecommerce is 50 requests per minute ... which should be more than enough for this endpoint

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Keys added as part of https://openedx.atlassian.net/servicedesk/customer/portal/3/DEVOPS-8165 ...

ecommerce/extensions/api/v2/views/assignmentemail.py
coryleeio: Marking as resolved since endpoint is rate limited globally on a per endpoint basis.

coryleeio (approved):

PR #1999, opened 2018-11-30
LEARNER-6788 fix: management view failed to process basket_id

zainab-amir:
JIRA Ticket: https://openedx.atlassian.net/browse/LEARNER-6788

#### Problem : 
When basket id was used in management view to create the order
https://ecommerce.edx.org/management/
the view was unable to create the order

#### Reason :
This was happening because the coupon applied to basket expired after the payment and before fulfillment of basket time.

### DIFF
--- a/ecommerce/management/utils.py
+++ b/ecommerce/management/utils.py
@@ -143,10 +143,10 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
             # Need to handle the case that applied voucher has been expired.
             # This will create the order  with out discount but subsequently
             # run the fulfillment to update course mode.
-            try:
-                Applicator().apply(basket, user=basket.owner)
-            except ValueError:
+            voucher = basket.vouchers.first()
+            if voucher and not (voucher.is_active() or voucher.is_available_to_user(user=basket.owner)):
                 basket.clear_vouchers()
+            Applicator().apply(basket, user=basket.owner)
 
             payment_notification = self.get_payment_notification(basket)
             if not payment_notification:
@@ -183,7 +183,7 @@ class FulfillFrozenBaskets(EdxOrderPlacementMixin):
                 # than to retrieve an invoice number from PayPal.
                 order_number = basket.order_number
 
-                self.handle_order_placement(
+                order = self.handle_order_placement(
                     order_number=order_number,
                     user=user,
                     basket=basket,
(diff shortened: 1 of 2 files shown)


ecommerce/management/utils.py
```diff
@@ -143,10 +143,10 @@ def fulfill_basket(self, basket_id, site):
             # Need to handle the case that applied voucher has been expired.
             # This will create the order  with out discount but subsequently
             # run the fulfillment to update course mode.
```
robrap, line 145: @zainab-amir: Thank you for the changes.  I am unable to do a full review at this time.  However, I do want to better understand whether or not this is the best approach.  

Something seems wrong about clearing vouchers off a basket that was valid during purchase.  Is this really the only way?  Also, if this is the solution, to we have proper audit data to know what voucher was used during purchase? 

This may take reviewing django-oscar documentation, and mailing list, and possibly asking the mailing list as well.  Maybe you already have docs you can point to?

robrap: @georgebabey: Not sure if you or @brittneyexline could be a part of this, but there are questions around proper handling of voucher expiration that happens after purchase and before fulfillment that I'd like us to be clear about as a guild.

zainab-amir: @robrap If the expired voucher is kept in the basket and we call [handle order placement](https://github.com/edx/ecommerce/blob/master/ecommerce/management/utils.py#L186) with it, it breaks the code at  [this](https://github.com/django-oscar/django-oscar/blob/1.5.1/src/oscar/apps/order/utils.py#L71). 

Maybe after the order has been placed we can update the order.basket again and add the expired voucher for the record?

brittneyexline: @robrap - I haven't had to dig into what happens yet on the purchase/fulfillment side regarding coupons as it hasn't had to change so far in our work, but I'm happy to be part of whatever conversation around improving coupons that you want to have more broadly.

robrap: Thanks @brittneyexline.
@zainab-amir: Agreed.  Not sure what the right way to handle this would be, because the order should match what the user paid for.  I created an issue in oscar: https://github.com/django-oscar/django-oscar/issues/2891

robrap:
@zainab-amir: The recommended solution in the issue is creating the order as pending at the time the payment is initiated. This sounds reasonable to me, but I’m not sure of all of the ramifications. See https://github.com/django-oscar/django-oscar/issues/2891#issuecomment-443442916.

zainab-amir:
@robrap I will look into it.

robrap:
@zainab-amir: I'm closing this PR while we do the other discovery.  Please make sure this PR is properly documented in the ticket if necessary.

PR #2001, opened 2018-12-04
ENT-1253 Introducing OfferAssignment model

brittneyexline:
**Description**
In preparation of building the code assignment capability, we need this data model for tracking assignments and their current state. Adding it separately here from the implementations so that more work can be done in parallel.

**Testing**
Not needed for now - testing will be done as part of the implementation changes. This PR simply adds the model and the corresponding migration without using it.


irfanuddinahmad (approved):
LGTM!

ecommerce/extensions/offer/models.py
```diff
@@ -449,4 +449,33 @@ class Condition(AbstractCondition):
...
+        (REDEEMED, _("Code has been redeemed by user.")),
+        (EMAIL_BOUNCED, _("Email to user bounced.")),
+        (REVOKED, _("Code has been revoked for this user.")),
+    )
+
+    offer = models.ForeignKey('offer.ConditionalOffer')
+    code = models.CharField(max_length=128)
+    user_email = models.EmailField()
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(auto_now=True)
```
muhammad-ammar: What about inheriting from `TimeStampedModel`? It will give us `created` and `modified` fields and we are already using it at many places.

muhammad-ammar (approved):
@brittneyexline Looks good to me 👍. Just one minor comment.

ecommerce/extensions/offer/models.py
brittneyexline: good point! I'll update this.

ecommerce/extensions/offer/models.py
```diff
@@ -449,4 +450,31 @@ class Condition(AbstractCondition):
...
+    STATUS_CHOICES = (
+        (EMAIL_PENDING, _("Email to user pending.")),
+        (ASSIGNED, _("Code successfully assigned to user.")),
+        (REDEEMED, _("Code has been redeemed by user.")),
+        (EMAIL_BOUNCED, _("Email to user bounced.")),
+        (REVOKED, _("Code has been revoked for this user.")),
+    )
+
+    offer = models.ForeignKey('offer.ConditionalOffer')
+    code = models.CharField(max_length=128)
```
mushtaqak, line 470: Should we make `code` a ForeignKey here? In this case, we may not need `voucher_application`. Thoughts?

mushtaqak (approved):
@brittneyexline Just one question, otherwise looks good.

ecommerce/extensions/offer/models.py
brittneyexline: we need voucher application because it indicates the moment when a user actually redeems the voucher and offer. i agree that it is a little weird here... the reason i added it was because for single use coupons, where every voucher points to the same conditional offer. because we assign at the code level, we need to indicate that here. the reason we are tying this to conditional offer and not voucher is because logically, that is the object in oscar that indicates that a discount is applied based on some condition; in our case the condition is contingent on assignment.

irfanuddinahmad:
@brittneyexline @georgebabey We need the (user_email, code) pair to uniquely identify an assignment message in this model.
As it turns out, the Sailthru bounce api only POSTs a message identifier for the bounced email. Using the api, we can get the bounced message details, however, in the response only the user_email is available. In order to update the offer_assignment model, I propose that we add a column for the unique message identifier sent by Sailthru (to be saved at send time). However, this would tightly couple the implementation to Sailthru ... 
Thoughts ... ?
image

brittneyexline:
@irfanuddinahmad - this morning at standup we discussed filing a separate ticket to track bounced emails, so i think whatever we decide should come after we merge this (which i'm hoping to do today).

as such, we'll have more time to discuss a solution, but i would agree that adding a field on this model would be tightly coupling to sailthru, and i think that is best avoided. i don't know the exact answer, but i would say tracking attempted emails should be in a separate model. also, emails are not one to one with offer assignments because multiple offer assignments will exist if someone is assigned to the same code more than once (this happens for multiple use for one customer codes), so user_email/code will not be unique.

georgebabey (approved):

PR #2003, opened 2018-12-04
ENT-1253 Adding api for assigning codes

brittneyexline:
**Description**
This PR introduces an API endpoint to assign "slots" in a coupon to user emails, with the option to specify which codes those slots should be assigned to. This endpoint is able to support both bulk and individual assign use cases for the coupon management admin portal.

The endpoint can be called by sending a POST request to: `{ecommerce_hostname}/api/v2/enterprise/coupons/{coupon_id}/assign/` with a json payload containing `emails`, an array of email addresses, which is required, and an optional `codes`, which is an array of codes that should correspond to vouchers for that coupon that assignment should be limited to.

When it is called, we verify that the vouchers attached to the coupon (filtered by the codes param, if sent) contain enough open slots for the number of emails sent in the request. A voucher has 0 or more open slots depending on the number of times it has been redeemed, the number of active (ie not revoked or redeemed) assignments that exist, and the number of max_uses for the associated enterprise offer. If there are not enough slots, the request fails. If there are, we go on to write the appropriate OfferAssignments for each email sent.

**Testing instructions**
Deployed to the business sandbox.

Hitting the endpoint for a non-enterprise coupon should fail with a 404.
Example request: 
```
POST /api/v2/enterprise/coupons/284/assign/ HTTP/1.1
Host: ecommerce-business.sandbox.edx.org
Authorization: JWT {jwt_token}
Content-Type: application/json
{"emails": ["bexline@edx.org"]}
```

_SINGLE_USE Coupons_
Hit the endpoint for a SINGLE_USE enterprise coupon with one code that has not been redeemed:
Example request:
```
POST /api/v2/enterprise/coupons/395/assign/ HTTP/1.1
Host: ecommerce-business.sandbox.edx.org
Authorization: JWT {jwt_token}
Content-Type: application/json
{"emails": ["bexline@edx.org"]}
```
This should successfully return a response like so, and create a new row in the offer_offerassignment table in the ecommerce db:
```
{
    "offer_assignments": [
        {
            "user_email": "bexline@edx.org",
            "code": "7OIZO5GRFT2D7C4R"
        }
    ]
}
```
If you try again with a different email, the request should fail because the code only has one slot available and it has already been assigned; the failure message will look like this:
```
{
    "non_field_errors": [
        "Not enough available codes for assignment!"
    ]
}
```
You would get the same result if the code had already been redeemed.

Hit the endpoint for a SINGLE_USE coupon that has many codes and assign to multiple emails.
Example Request:
```
POST /api/v2/enterprise/coupons/386/assign/ HTTP/1.1
Host: ecommerce-business.sandbox.edx.org
Authorization: JWT {jwt_token}
Content-Type: application/json
{"emails": ["bexline+1@edx.org","bexline+2@edx.org","bexline+3@edx.org"]}
```
Example Response:
```
{
    "offer_assignments": [
        {
            "user_email": "bexline+1@edx.org",
            "code": "ZKZMKBL6ROY4YA7R"
        },
        {
            "user_email": "bexline+2@edx.org",
            "code": "PO73EDONFDRJIYL5"
        },
        {
            "user_email": "bexline+3@edx.org",
            "code": "LQUKUCDDVZZWM4VD"
        }
    ]
}
```
Now try hitting again by adding a code filter with unused codes from this coupon. The response should contain only the codes that were sent in the request.
Example Request:
```
POST /api/v2/enterprise/coupons/386/assign/ HTTP/1.1
Host: ecommerce-business.sandbox.edx.org
Authorization: JWT {jwt_token}
Content-Type: application/json
{"emails": ["bexline+1@edx.org","bexline+2@edx.org","bexline+3@edx.org"],"codes":["GHGSVMQE6XKNZE2V","CYOFOH6XTEEBMKVN","IFXGPNTP6SKRENTC"]}
```
Response:
```
{
    "offer_assignments": [
        {
            "user_email": "bexline+1@edx.org",
            "code": "CYOFOH6XTEEBMKVN"
        },
        {
            "user_email": "bexline+2@edx.org",
            "code": "IFXGPNTP6SKRENTC"
        },
        {
            "user_email": "bexline+3@edx.org",
            "code": "GHGSVMQE6XKNZE2V"
        }
    ]
}
```

_MULTI_USE and ONCE_PER_CUSTOMER Coupons_
These two coupon types behave more or less the same way. The coupons can have codes that can be assigned multiple users. ONCE_PER_CUSTOMER Coupons have the additional restriction that a user can only have one assignment per code; requests that contain users who have already redeemed or been assigned to a code will remove those codes from consideration for that assignment request, increasing the chance that the request might be invalid. The number of assignments possible is contingent on the number of redemptions already made on a voucher, the max_global_applications of the offer, and the number of existing active assignments. If max_global_applications is not set, it is assumed to be 10000 (this is an oscar default). Code filtering works the same way as stated for single use coupons.

MULTI_USE Example request:
```
POST /api/v2/enterprise/coupons/5/assign/ HTTP/1.1
Host: ecommerce-business.sandbox.edx.org
Authorization: JWT {jwt_token}
Content-Type: application/json
{"emails": ["bexline+1@edx.org","bexline+2@edx.org","bexline+3@edx.org"]}
```
Response:
```
{
    "offer_assignments": [
        {
            "user_email": "bexline+1@edx.org",
            "code": "PIPNJSUK33P7PTZH"
        },
        {
            "user_email": "bexline+2@edx.org",
            "code": "PIPNJSUK33P7PTZH"
        },
        {
            "user_email": "bexline+3@edx.org",
            "code": "PIPNJSUK33P7PTZH"
        }
    ]
}
```

ONCE_PER_CUSTOMER Example Request:
```
POST /api/v2/enterprise/coupons/384/assign/ HTTP/1.1
Host: ecommerce-business.sandbox.edx.org
Authorization: JWT {jwt_token}
Content-Type: application/json
{"emails": ["bexline@edx.org"]}
```

Response:
```
{
    "offer_assignments": [
        {
            "user_email": "bexline@edx.org",
            "code": "IMKJFFOZMS2BXH5Y"
        }
    ]
}
```

If we try the request again, passing in the codes param with the assigned code, it should fail since the user has already been assigned to that code.

_MULTI_USE_PER_CUSTOMER Coupons_
This is a new coupon type that we have introduced. In order to enforce the fact that a code can be used multiple times, but is restricted to one user, we are going to write n number of offer assignments for that email/code, where n is whatever max_global_applications is set to.

Example request for a coupon with 5 uses per code:
```
POST /api/v2/enterprise/coupons/399/assign/ HTTP/1.1
Host: ecommerce-business.sandbox.edx.org
Authorization: JWT {jwt_token}
Content-Type: application/json
{"emails": ["bexline+1@edx.org"]}
```
Response:
```
{
    "offer_assignments": [
        {
            "user_email": "bexline+1@edx.org",
            "code": "33BRHEELIR7OMXMW"
        },
        {
            "user_email": "bexline+1@edx.org",
            "code": "33BRHEELIR7OMXMW"
        },
        {
            "user_email": "bexline+1@edx.org",
            "code": "33BRHEELIR7OMXMW"
        },
        {
            "user_email": "bexline+1@edx.org",
            "code": "33BRHEELIR7OMXMW"
        },
        {
            "user_email": "bexline+1@edx.org",
            "code": "33BRHEELIR7OMXMW"
        }
    ]
}
```


ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -164,3 +167,28 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        emails = request.data.get('emails')
+
+        try:
+            offers_for_assignment = get_available_offers_for_assignment(coupon, len(emails), codes)
+        except Exception as e:
+            raise ValidationError(e.message)
+
+        email_iterator = iter(emails)
+        response_data = []
+        for code in offers_for_assignment:
```
muhammad-ammar: @brittneyexline It would be great if we can add comments here and in `get_available_offers_for_assignment` and `slots_available_for_assignment` so that others can understand why we are doing what we are doing :smile:

brittneyexline: agreed - this is my "rough draft" so i will be sure to add comments and continue restructuring for clarity as i test.

ecommerce/extensions/api/serializers.py
```diff
@@ -943,3 +944,94 @@ class ProviderSerializer(serializers.Serializer):  # pylint: disable=abstract-me
...
+        child=serializers.EmailField(), required=True, write_only=True
+    )
+    offer_assignments = serializers.ListField(
+        child=OfferAssignmentSerializer(), read_only=True
+    )
+
+    def create(self, validated_data):
+        """Create OfferAssignment objects for each email and the available_assignments determined from validation."""
+        emails = validated_data.get('emails')
+        voucher_usage_type = validated_data.get('voucher_usage_type')
```
christopappas: .pop might be useful here if you are going to delete later

ecommerce/extensions/api/serializers.py
```diff
@@ -943,3 +944,94 @@ class ProviderSerializer(serializers.Serializer):  # pylint: disable=abstract-me
...
+        available_assignments = validated_data.get('available_assignments')
+        email_iterator = iter(emails)
+        offer_assignments = []
+
+        for code in available_assignments:
+            offer = available_assignments[code]['offer']
+            email = next(email_iterator) if voucher_usage_type == Voucher.MULTI_USE_PER_CUSTOMER else None
```
christopappas, line 1031: this is clever use of this iterator 👍

ecommerce/extensions/voucher/models.py
```diff
@@ -90,4 +112,33 @@ def best_offer(self):
...
+        ).count()
+
+        # If this a Single use or Multi use per customer voucher,
+        # it must have no orders or existing assignments to be assigned.
+        if self.usage in (self.SINGLE_USE, self.MULTI_USE_PER_CUSTOMER):
+            if self.num_orders or num_assignments:
+                return 0
+            return enterprise_offer.max_global_applications or 1
+        else:
+            offer_max_uses = enterprise_offer.max_global_applications or 10000
```
christopappas: where does the 10000 come from?

christopappas (commented):
I know its a WIP, it looks good. Just a few minor comments

ecommerce/extensions/voucher/models.py
brittneyexline: it's the default limit on an offer when none is specified: https://django-oscar.readthedocs.io/en/releases-1.1/_modules/oscar/apps/offer/abstract_models.html#AbstractConditionalOffer.get_max_applications

robrap: If it is a number that can be pulled from Oscar, that would be ideal.  If not, please use a constant and avoid magic numbers.  If you haven't heard the term before, see: https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad
Thanks both.  Sorry for butting in.  This might have been Chris's reply anyway, but I didn't want to have to remember to come back later in case it wasn't. :)

georgebabey (approved):
Code looks good and thanks for the thorough testing instructions!

PR #2005, opened 2018-12-06
Add host information to Segment events.

brianhw:
Addressing DE-1140.

Passes url context to Segment, so that Segment can pass hostname (and path) to GA.

### DIFF
--- a/ecommerce/extensions/analytics/utils.py
+++ b/ecommerce/extensions/analytics/utils.py
@@ -1,6 +1,7 @@
 import json
 import logging
 from functools import wraps
+from urlparse import urlunsplit
 
 from ecommerce.courses.utils import mode_for_product
 
@@ -144,10 +145,24 @@ def track_segment_event(site, user, event, properties):
         return False, msg
 
     user_tracking_id, ga_client_id, lms_ip = parse_tracking_context(user)
+    # construct a URL, so that hostname can be sent to GA.
+    # For now, send a dummy value for path.  Segment parses the URL and sends
+    # the host and path separately. When needed, the path can be fetched by adding:
+    # request = crum.get_current_request()
+    # if request:
+    #     path = request.META.get('PATH_INFO')
+    hostname = site.domain
+    path = '/'
+    parts = ("https", hostname, path, "", "")
+    page = urlunsplit(parts)
+
     context = {
         'ip': lms_ip,
         'Google Analytics': {
-            'clientId': ga_client_id
+            'clientId': ga_client_id,
+        },
+        'page': {
+            'url': page,
         }
     }
     return site.siteconfiguration.segment_client.track(user_tracking_id, event, properties, context=context)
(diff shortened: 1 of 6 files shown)


robrap:
@brianhw 
1. Do you have any concerns someone should be aware of, or is this as straight-forward as it appears?
2. It probably makes sense to just have someone on your team review.

brianhw:
Thanks for the reply, @robrap .   
1. It is as straight-forward as it appears.  I've made this as minimal as necessary to get the immediate GA problems side-stepped.   There may be more GA issues that will require other interventions. 
2. I could have someone on my team review it, but none of us in data engineering have permission to merge.

ecommerce/extensions/analytics/utils.py
```diff
@@ -144,10 +145,19 @@ def track_segment_event(site, user, event, properties):
...
+    page = urlunsplit(parts)
+
     context = {
         'ip': lms_ip,
         'Google Analytics': {
-            'clientId': ga_client_id
+            'clientId': ga_client_id,
+        },
+        'page': {
+            'url': page,
```
macdiesel, line 165: As long as we are aware that the URL will not be the actual URL and only the hostname with the root path and this is the intended behavior I'm ok with this.

macdiesel (approved):

ecommerce/extensions/analytics/utils.py
brianhw: Yup, this is the intended behavior for now.   Segment parses the URL and sends the host and path separately.  So when we need GA to have the correct path, we can fix this then.   The path would require introducing crum.get_current_request().  It would work, but it is more invasive, and I don't want to incur unnecessary risk in this repo.

robrap: @brianhw: Maybe this deserves a code comment regarding your decision?

robrap (commented):
One minor request and this is good to go.  Thanks.

ecommerce/extensions/analytics/utils.py
brianhw: Excellent point, thanks.  I'll add that (after my interview/lunch).

brianhw: Done.

brianhw:
@robrap This is ready for your review and merge.

robrap (approved):
This needs to be rebased before merge.  I'll either take care of that myself, or coordinate at a time I can merge with you.

brianhw:
I did a diff of the branch against master, and I don't see any conflict.  I could do that rebase on master if that would help.   Or does it need to be rebased to a different branch.

robrap:
Rebase only helps if I merge before someone else does first. That’s why timing is important. 🙂

brianhw:
I rebased on master as it currently exists, in case that simplifies your other rebase/merge.

PR #2006, opened 2018-12-07
Filter coupon details

mushtaqak:
The purpose of this PR is to add filtering for coupon details.

f/e PR: https://github.com/edx/edx-portal/pull/129

[ENT-1345](https://openedx.atlassian.net/browse/ENT-1345)


ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -192,22 +225,18 @@ def codes(self, request, pk):  # pylint: disable=unused-argument
...
-        # this will give us only those vouchers having no application
-        coupon_vouchers_wo_applications = Voucher.objects.filter(
-            coupon_vouchers__coupon__id=coupon.id,
-            applications__isnull=True
-        )
-
-        # we need a combined querset so that pagination works as expected
-        all_coupon_vouchers = coupon_vouchers_with_applications | coupon_vouchers_wo_applications
+        coupon_code_filter = request.query_params.get('code_filter')
+        if coupon_code_filter == 'not-redeemed':
```
brittneyexline: it would be good to make `not-redeemed` and `not-assigned` constants

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -192,22 +225,18 @@ def codes(self, request, pk):  # pylint: disable=unused-argument
         }
         """
         coupon = self.get_object()
+        coupon_vouchers = self.get_coupon_vouchers(coupon)
```
brittneyexline: why not `coupon_vouchers = coupon.attr.coupon_vouchers.all()`?

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -172,6 +174,37 @@ def update_offer_data(self, request_data, vouchers, site):
...
+            coupon_vouchers__coupon__id=coupon.id
+        )
+
+    def get_not_assigned_coupon_vouchers(self, queryset):
+        """
+        Returns vouchers which have not been redeemed yet with no offer assigment.
+        """
+        not_redeemed_vouchers = self.get_not_redeemed_coupon_vouchers(queryset)
+        assigned_codes = OfferAssignment.objects.filter(voucher_application__isnull=True, status__in=[OFFER_ASSIGNED, OFFER_ASSIGNMENT_EMAIL_PENDING]).values_list('code', flat=True)
+        return not_redeemed_vouchers.exclude(code__in=assigned_codes)
```
brittneyexline: did we ever decide if not assigned meant not ever assigned (as you have here) or has assignments left?

brittneyexline (commented):
just a few questions, but looks good so far!

ecommerce/extensions/api/v2/views/enterprise.py
mushtaqak: So you are thinking about a case where a voucher was assigned once and then got revoked, now it is assignable (not-assigned currently), right? In this case when voucher is revoked, it's status should be changed from `OFFER_ASSIGNED` to `OFFER_ASSIGNMENT_REVOKED` and it should be available for assignment.

Also, I am assuming that if a voucher is redeemed, it's not assignable.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -179,6 +181,34 @@ def update_offer_data(self, request_data, vouchers, site):
         if coupon_was_migrated:
             super(EnterpriseCouponViewSet, self).update_range_data(request_data, vouchers)
 
+    def get_not_assigned_coupon_vouchers(self, queryset):
+        """
+        Returns vouchers which have not been redeemed yet with no offer assigment.
+        """
+        # First we filter to get coupon not-redeemed vouchers which will give us assigned and not-assigned vouchers
```
mushtaqak: @brittneyexline Just changed the logic to exclude `OFFER_REDEEMED` status offerassignements. And added the explainaing comment above. Does this make sense to you ?

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -179,6 +181,34 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        Returns vouchers having no voucher application.
+        """
+        return queryset.filter(applications__isnull=True)
+
+    def get_redeemed_coupon_vouchers(self, coupon):
+        """
+        Returns vouchers against each voucher application, so if a
+        voucher has two applications than this queryset will give 2 vouchers.
+        """
+        return Voucher.objects.filter(
```
brittneyexline, line 220: this is the only method that doesn't take the queryset, any reason this can't be `return queryset.filter(applications__isnull=False)`?

brittneyexline (commented):
my only concern is making sure that we all have the same understanding of "not redeemed" and "not assigned" - i don't think anyone has a clear idea about it.

This is especially a problem for MULTI_USE coupons, or really any coupon type where a voucher could be redeemed or assigned multiple times.

For example, if I have a MULTI_USE code 'ABCDEF' that has max_uses = 10, has been redeemed once by user1 and assigned once to user2, which toggle(s) should it show up in? It has been redeemed, but has 9 redemptions left. It has been assigned, but has 8 assignments left.

In short, do 'not redeemed' and 'not assigned' mean never have been redeemed/assigned (as you have implemented it), or has redemptions/assignments left?

I am going to start a conversation in slack so we can clarify.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -199,22 +250,26 @@ def codes(self, request, pk):  # pylint: disable=unused-argument
...
-        )
-        # this will give us only those vouchers having no application
-        coupon_vouchers_wo_applications = Voucher.objects.filter(
-            coupon_vouchers__coupon__id=coupon.id,
-            applications__isnull=True
-        )
-
-        # we need a combined querset so that pagination works as expected
-        all_coupon_vouchers = coupon_vouchers_with_applications | coupon_vouchers_wo_applications
+        coupon_code_filter = request.query_params.get('code_filter')
```
mushtaqak, line 299: @brittneyexline Can you have quick look at `get_not_assigned_coupon_vouchers `, `get_not_redeemed_coupon_vouchers` and `get_redeemed_coupon_vouchers` methods ... and if there is anything missing ? or If I am going in right direction ?

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -179,6 +187,49 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        """
+        return Voucher.objects.filter(
+            applications__voucher_id__in=coupon_vouchers
+        )
+
+    def get_not_assigned_coupon_vouchers(self, coupon_vouchers):
+        """
+        Returns vouchers which have not been redeemed yet with no offer assigment.
+        """
+        vouchers_without_applications = self.get_vouchers_with_no_applications(coupon_vouchers)
```
brittneyexline: if we are trying to get codes with assignments left, that means we need to include ones that have been redeemed but still have assignments left. right now, this method is only returning codes that have never been redeemed or assigned, it should also return codes that have been redeemed or assigned but still have assignments left.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -179,6 +187,49 @@ def update_offer_data(self, request_data, vouchers, site):
         if coupon_was_migrated:
             super(EnterpriseCouponViewSet, self).update_range_data(request_data, vouchers)
 
+    def get_assigned_codes(self):
+        """
+        Returns the list of all assigned codes.
+        """
+        return OfferAssignment.objects.exclude(status__in=[OFFER_REDEEMED]).values_list('code', flat=True)
```
brittneyexline: we also need to filter out revoked codes.

ecommerce/extensions/api/v2/views/enterprise.py
brittneyexline: looks like this is currently returning codes that have ever been redeemed, not codes that are fully redeemed (ie no usages left), which i believe is the intention.
once you have the fully redeemed codes, you could move this to the partial redeemed function and then exclude fully redeemed coupons for the implementation of partial redeemed.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -199,22 +250,26 @@ def codes(self, request, pk):  # pylint: disable=unused-argument
...
+            coupon_vouchers = self.get_not_redeemed_coupon_vouchers(coupon_vouchers)
+        elif coupon_code_filter == VOUCHER_PARTIAL_REDEEMED:
+            # Assigned vouchers having at-least one redemption.
+            coupon_vouchers = self.get_partial_redeemed_coupon_vouchers(coupon_vouchers)
+        else coupon_code_filter == VOUCHER_REDEEMED:
+            # Fully redeemed vouchers.
+            coupon_vouchers = self.get_redeemed_coupon_vouchers(coupon_vouchers)
+        else:
+            # Return all vouchers for coupon (redeemed + not-redeemed).
+            coupon_vouchers = self.get_redeemed_coupon_vouchers(coupon_vouchers) | self.get_not_redeemed_coupon_vouchers(coupon_vouchers)
```
brittneyexline: with the change in definitions, this might need to be redeemed + partial-redeemed + not-redeemed.

keep in mind that codes will only appear in one of the three toggles at any given time, so the union of all of those should be all of the codes.

brittneyexline (commented):
take another look at your wiki page and the test/release plan document to see if your definitions are aligned - right now it doesn't seem to match up quite, based on my understanding.

mushtaqak:
Being addressed in https://github.com/edx/ecommerce/pull/2099

PR #2007, opened 2018-12-07
implement custom condition for user assignment

muhammad-ammar:
[ENT-1251](https://openedx.atlassian.net/browse/ENT-1251)

**Description**
This adds a custom condition to enforce two things, A user can redeem the voucher 
* if voucher is assigned to that user, OR
* If there are slots available 

**Testing instructions**
Deployed to the business sandbox.

1. Create [enterprise coupons](https://ecommerce-business.sandbox.edx.org/enterprise/coupons/) of different types
2. Download the csv report of a coupon. This report contains all the vouchers belong to a coupon and information like` redeem URL`, `max_global_applications` etc
3. Create voucher assignments as mentioned in description of https://github.com/edx/ecommerce/pull/2003. You need to use the codes from `step 2`.
4. You can see the created voucher assignments [here](https://ecommerce-business.sandbox.edx.org/admin/offer/offerassignment/).
5. Login as some assigned user.
6. Use the redeem URL from `step 2`. You should be able to redeem the voucher if the any of the above 2 conditions satisfies.

 **Detailed Instructions**

**SINGLE_USE** coupon
a. After completion of `step 1` and `step 2`, we will have have something like below for a

| code   |      Max Usage      |  Redeem URL |
|----------|:-------------:|------:|
| 5NZMPU3DJRS7YCQB |  1 |  https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=5NZMPU3DJRS7YCQB |
| RAFGT7MKPXGOKLGU |  1 |  https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=RAFGT7MKPXGOKLGU |

b. And below are voucher assignments created 

```python
{
    "offer_assignments": [
        {
            "user_email": "mammar0@edx.org",
            "code": "5NZMPU3DJRS7YCQB"
        },
        {
            "user_email": "mammar1@edx.org",
            "code": "RAFGT7MKPXGOKLGU"
        }
    ]
}
```

c. Login as `mammar0@edx.org` and redeem the voucher using https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=5NZMPU3DJRS7YCQB and `mammar0@edx.org` should be able to redeem the code. `5NZMPU3DJRS7YCQB` voucher can only be used once by assigned user only.

d. Login as `other@edx.org` and try to redeem the voucher using https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=RAFGT7MKPXGOKLGU and you should not able to redeem voucher because `RAFGT7MKPXGOKLGU` can only be redeemed by `mammar1@edx.org`

**MULT_USE_PER_CUSTOMER** coupon
a. After completion of `step 1` and `step 2`, we will have have something like below for a

| code   |      Max Usage      |  Redeem URL |
|----------|:-------------:|------:|
| Y4BK3HXW4WQAXYYC |  3 |  https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=Y4BK3HXW4WQAXYYC |
| BNSSYNWWXRUI7MFI |  3 |  https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=BNSSYNWWXRUI7MFI |

b. And below are voucher assignments created 

```python
{
    "offer_assignments": [
        {
            "user_email": "mammar21@edx.org",
            "code": "Y4BK3HXW4WQAXYYC"
        },
        {
            "user_email": "mammar21@edx.org",
            "code": "Y4BK3HXW4WQAXYYC"
        },
        {
            "user_email": "mammar21@edx.org",
            "code": "Y4BK3HXW4WQAXYYC"
        },
        {
            "user_email": "mammar32@edx.org",
            "code": "BNSSYNWWXRUI7MFI"
        },
        {
            "user_email": "mammar32@edx.org",
            "code": "BNSSYNWWXRUI7MFI"
        },
        {
            "user_email": "mammar32@edx.org",
            "code": "BNSSYNWWXRUI7MFI"
        }
    ]
}
```

c. Login as `mammar21@edx.org` and redeem the voucher using https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=Y4BK3HXW4WQAXYYC and `mammar21@edx.org` should be able to redeem the code. `Y4BK3HXW4WQAXYYC` voucher can be used 3 times by `mammar21@edx.org` only.

d. Login as `other@edx.org` and try to redeem the voucher using https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=Y4BK3HXW4WQAXYYC and you should see `This voucher is only available to another user` message because `MULT_USE_PER_CUSTOMER` coupon can only be used by the user it is assigned to `plus` once this coupon is used by a user then after that only that user can further use it.

**MULT_USE** coupon
a. After completion of `step 1` and `step 2`, we will have have something like below for a

| code   |      Max Usage      |  Redeem URL |
|----------|:-------------:|------:|
| SEDVNXYPKUJ24CNL |  3 |  https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=SEDVNXYPKUJ24CNL |
| KGENOZIU4AGTZFS3 |  3 |  https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=KGENOZIU4AGTZFS3 |

b. And below are voucher assignments created 

```python
{
    "offer_assignments": [
        {
            "user_email": "mammar12@edx.org",
            "code": "SEDVNXYPKUJ24CNL"
        },
        {
            "user_email": "mammar13@edx.org",
            "code": "SEDVNXYPKUJ24CNL"
        },
        {
            "user_email": "mammar14@edx.org",
            "code": "SEDVNXYPKUJ24CNL"
        },
        {
            "user_email": "mammar11@edx.org",
            "code": "KGENOZIU4AGTZFS3"
        }
    ]
}
```

c. Login as `mammar12@edx.org` and redeem the voucher using https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=SEDVNXYPKUJ24CNL and `mammar21@edx.org` should be able to redeem the code. `SEDVNXYPKUJ24CNL` voucher is can only be used 3 times by assigned users only.

d. Login as `mammar11@edx.org` and redeem the voucher using https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=KGENOZIU4AGTZFS3 and `mammar21@edx.org` should be able to redeem the code. `KGENOZIU4AGTZFS3` voucher can be used 3 times by 3 users, 1 time by `mammar11@edx.org` and plus 2 other users. 

**ONCE_PER_CUSTOMER** coupon
a. After completion of `step 1` and `step 2`, we will have have something like below for a

| code   |      Max Usage      |  Redeem URL |
|----------|:-------------:|------:|
| VD7KMNPHLGIC3UPT |  3 |  https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=VD7KMNPHLGIC3UPT |
| GIDVKHA44QPPI5EM |  3 |  https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=GIDVKHA44QPPI5EM |

b. And below are voucher assignments created 

```python
{
    "offer_assignments": [
        {
            "user_email": "mammar7@edx.org",
            "code": "VD7KMNPHLGIC3UPT"
        },
        {
            "user_email": "mammar8@edx.org",
            "code": "VD7KMNPHLGIC3UPT"
        },
        {
            "user_email": "mammar9@edx.org",
            "code": "VD7KMNPHLGIC3UPT"
        },
        {
            "user_email": "mammar10@edx.org",
            "code": "GIDVKHA44QPPI5EM"
        }
    ]
}
```

c. Login as `mammar7@edx.org` and redeem the voucher using https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=VD7KMNPHLGIC3UPT and `mammar7@edx.org` should be able to redeem the code. `VD7KMNPHLGIC3UPT` voucher can only be used 3 times by assigned users only.

d. Login as `mammar10@edx.org` and redeem the voucher using https://ecommerce-business.sandbox.edx.org/coupons/offer/?code=GIDVKHA44QPPI5EM and `mammar21@edx.org` should be able to redeem the code. `GIDVKHA44QPPI5EM` voucher is can be used 3 times by 3 users, 1 time by `mammar10@edx.org` and plus 2 other users.

### DIFF
--- a/ecommerce/enterprise/conditions.py
+++ b/ecommerce/enterprise/conditions.py
@@ -11,6 +11,7 @@ from slumber.exceptions import SlumberHttpBaseException
 from ecommerce.enterprise.api import catalog_contains_course_runs, fetch_enterprise_learner_data
 from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH, ENTERPRISE_OFFERS_SWITCH
 from ecommerce.extensions.basket.utils import ENTERPRISE_CATALOG_ATTRIBUTE_TYPE
+from ecommerce.extensions.offer.constants import OFFER_ASSIGNMENT_REVOKED, OFFER_REDEEMED
 from ecommerce.extensions.offer.decorators import check_condition_applicability
 from ecommerce.extensions.offer.mixins import ConditionWithoutRangeMixin, SingleItemConsumptionConditionMixin
 
@@ -18,6 +19,8 @@ BasketAttribute = get_model('basket', 'BasketAttribute')
 BasketAttributeType = get_model('basket', 'BasketAttributeType')
 Condition = get_model('offer', 'Condition')
 ConditionalOffer = get_model('offer', 'ConditionalOffer')
+OfferAssignment = get_model('offer', 'OfferAssignment')
+Voucher = get_model('voucher', 'Voucher')
 logger = logging.getLogger(__name__)
 
 
@@ -162,3 +165,37 @@ class AssignableEnterpriseCustomerCondition(EnterpriseCustomerCondition):
     class Meta(object):
         app_label = 'enterprise'
         proxy = True
+
+    def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
+        """
+        Determines that if user has assigned a voucher and is eligible for redeem it.
+
+        Arguments:
+            offer (ConditionalOffer): The offer to be redeemed.
+            basket (Basket): The basket of products being purchased.
+
+        Returns:
+            bool
+        """
+        condition_satisfied = super(AssignableEnterpriseCustomerCondition, self).is_satisfied(offer, basket)
+        if condition_satisfied is False:
+            return False
+
+        voucher = basket.vouchers.first()
+
+        # get assignments for the basket owner and basket voucher
+        user_with_code_assignments = OfferAssignment.objects.filter(
+            code=voucher.code, user_email=basket.owner.email
+        ).exclude(
+            status__in=[OFFER_REDEEMED, OFFER_ASSIGNMENT_REVOKED]
+        )
+
+        # user has assignments available
+        if user_with_code_assignments.exists():
+            return True
+
+        # basket owner can redeem the voucher if free slots are avialable
+        if voucher.slots_available_for_assignment:
+            return True
+
+        return False
--- a/ecommerce/enterprise/tests/test_conditions.py
+++ b/ecommerce/enterprise/tests/test_conditions.py
@@ -3,20 +3,31 @@ from uuid import uuid4
 
 import ddt
 import httpretty
+import mock
 from oscar.core.loading import get_model
 from waffle.models import Switch
 
+from ecommerce.coupons.tests.mixins import CouponMixin
 from ecommerce.courses.tests.factories import CourseFactory
+from ecommerce.enterprise.conditions import EnterpriseCustomerCondition
 from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH, ENTERPRISE_OFFERS_SWITCH
 from ecommerce.enterprise.tests.mixins import EnterpriseServiceMockMixin
+from ecommerce.extensions.api.serializers import CouponCodeAssignmentSerializer
 from ecommerce.extensions.basket.utils import basket_add_enterprise_catalog_attribute
 from ecommerce.extensions.catalogue.tests.mixins import DiscoveryTestMixin
+from ecommerce.extensions.offer.constants import (
+    OFFER_ASSIGNMENT_EMAIL_PENDING,
+    OFFER_ASSIGNMENT_REVOKED,
+    OFFER_REDEEMED
+)
 from ecommerce.extensions.test import factories
 from ecommerce.tests.factories import ProductFactory, SiteConfigurationFactory
 from ecommerce.tests.testcases import TestCase
 
 ConditionalOffer = get_model('offer', 'ConditionalOffer')
+OfferAssignment = get_model('offer', 'OfferAssignment')
 Product = get_model('catalogue', 'Product')
+Voucher = get_model('voucher', 'Voucher')
 LOGGER_NAME = 'ecommerce.programs.conditions'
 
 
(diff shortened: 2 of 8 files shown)


ecommerce/enterprise/conditions.py
```diff
@@ -162,3 +168,37 @@ class AssignableEnterpriseCustomerCondition(EnterpriseCustomerCondition):
     class Meta(object):
         app_label = 'enterprise'
         proxy = True
+
+    def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
```
muhammad-ammar, line 169: @brittneyexline This is still work in progress but can you please give it a quick look to see if I am on correct path?

ecommerce/enterprise/conditions.py
```diff
@@ -162,3 +168,37 @@ class AssignableEnterpriseCustomerCondition(EnterpriseCustomerCondition):
...
+
+        Returns:
+            bool
+        """
+        condition_satisfied = super(AssignableEnterpriseCustomerCondition, self).is_satisfied(offer, basket)
+
+        if condition_satisfied is False:
+            return False
+
+        voucher = basket.vouchers.first()
```
brittneyexline, line 184: this wasn't spelled out in the ticket, but we will need some additional logic to avoid vouchers having more redemptions available than they should. when i introduced the second offer specifically for enterprise, i copied over the max_global_applications value. however, the voucher might have been redeemed some number of times already with the original offer. i think this is the best place to check that the voucher.num_orders is less than max_global_applications, since we have introduced this weird situation.

ecommerce/enterprise/conditions.py
```diff
@@ -162,3 +168,37 @@ class AssignableEnterpriseCustomerCondition(EnterpriseCustomerCondition):
...
+
+        # check if user has any assigned offer
+        offer_assignments = OfferAssignment.objects.filter(
+            offer=offer,
+            user_email=basket.owner.email,
+            code=voucher.code
+        )
+        if not offer_assignments.exists():
```
brittneyexline: we will have to account for if the assignments have taken up all the available slots or not - i have created a property on the voucher with my PR (https://github.com/edx/ecommerce/pull/2003) that will tell you if the voucher has any slots available, accounting for assignments taking up slots.

we should also enforce that for MULTI_USE_PER_CUSTOMER vouchers, that anyone that does not have an offer assignment with a matching email cannot redeem it.

brittneyexline (commented):
on the right track! besides the comments i left inline, i believe the ticket also covers updating the OfferAssignment record with the VoucherApplication and redeemed state, which has to happen after purchase, so probably in the same place as orders and voucher applications are recorded.

ecommerce/enterprise/conditions.py
muhammad-ammar: I have added additional using `voucher.num_orders` and `offer.max_global_applications`.

ecommerce/enterprise/conditions.py
muhammad-ammar: I don't think that we can use `slots_available_for_assignment` because let's say voucher is of `SINGLE_USE` type and we have offer assignments for it. Then `slots_available_for_assignment` will return `0` but actually slots are available. 

Regarding `MULTI_USE_PER_CUSTOMER`, As per my understanding the current conditions will handle all the coupon types. 

`if not offer_assignments.exists():` is specifically handles the scenario where a voucher is not assigned to any user, in that case as per my understanding we can't move further and condition is satisfied.

muhammad-ammar:
@brittneyexline Thanks for the initial review. I have made more changes and replied to your comments. Also I will update the code so that records in the OfferAssignment model are updated once an assigned voucher is redeemed

ecommerce/enterprise/conditions.py
muhammad-ammar: Please help me understand if some scenario is not covered with the current logic in `is_satisfied` method.

ecommerce/enterprise/conditions.py
```diff
@@ -162,3 +170,44 @@ class AssignableEnterpriseCustomerCondition(EnterpriseCustomerCondition):
...
+        condition_satisfied = super(AssignableEnterpriseCustomerCondition, self).is_satisfied(offer, basket)
+        if condition_satisfied is False:
+            return False
+
+        voucher = basket.vouchers.first()
+
+        # check if voucher has any assignments
+        voucher_offer_assignments = OfferAssignment.objects.filter(code=voucher.code)
+        # basket owner can redeem the voucher if there were no offer assignments created for voucher
+        if not voucher_offer_assignments.exists():
```
brittneyexline: i think the logic needs to be expanded to say that the user can redeem the voucher when the voucher might have assignments (but not for this user), and there are fewer assignments than remaining slots.

ecommerce/enterprise/tests/test_conditions.py
```diff
@@ -260,3 +271,181 @@ def test_is_satisfied_course_run_not_in_catalog(self):
...
+
+            voucher, __ = Voucher.objects.get_or_create(
+                usage=voucher_type,
+                code=code,
+                defaults={
+                    'start_datetime': now() - datetime.timedelta(days=10),
+                    'end_datetime': now() + datetime.timedelta(days=10),
+                }
+            )
+            voucher.offers.add(enterprise_offer)
```
brittneyexline: not sure if it matters for testing purposes, but the only time that multiple vouchers point to the same offer is when the vouchers are SINGLE_USE.

ecommerce/enterprise/tests/test_conditions.py
```diff
@@ -260,3 +271,181 @@ def test_is_satisfied_course_run_not_in_catalog(self):
...
+
+    @mock.patch.object(EnterpriseCustomerCondition, 'is_satisfied', mock.Mock(return_value=True))
+    @ddt.data(
+        (0, 'test1@example.com', 'test1@example.com', OFFER_ASSIGNMENT_EMAIL_PENDING, True),
+        (0, 'test1@example.com', 'test1@example.com', OFFER_REDEEMED, False),
+        (0, 'test1@example.com', 'test1@example.com', OFFER_ASSIGNMENT_REVOKED, False),
+        (1, 'test1@example.com', 'test1@example.com', OFFER_ASSIGNMENT_EMAIL_PENDING, False),
+    )
+    @ddt.unpack
+    def test_is_satisfied(self, num_orders, assgined_email, owner_email, offer_status, condition_result):
```
brittneyexline: nit: assgined_email -> assigned_email
also, it appears that you have the same values for assigned_email and owner_email in all of the cases you have?

ecommerce/enterprise/tests/test_conditions.py
```diff
@@ -260,3 +271,181 @@ def test_is_satisfied_course_run_not_in_catalog(self):
...
+            is_condition_satisfied = self.condition.is_satisfied(enterprise_offer, basket)
+            assert is_condition_satisfied == expected_condition_result
+
+    @mock.patch.object(EnterpriseCustomerCondition, 'is_satisfied', mock.Mock(return_value=True))
+    @ddt.data(
+        (0, 'test1@example.com', 'test1@example.com', OFFER_ASSIGNMENT_EMAIL_PENDING, True),
+        (0, 'test1@example.com', 'test1@example.com', OFFER_REDEEMED, False),
+        (0, 'test1@example.com', 'test1@example.com', OFFER_ASSIGNMENT_REVOKED, False),
+        (1, 'test1@example.com', 'test1@example.com', OFFER_ASSIGNMENT_EMAIL_PENDING, False),
+    )
```
brittneyexline, line 346: with the cases you have here so far for a SINGLE_USE voucher, i don't think the 2nd and 4th cases here should be possible states. that being said, i don't know if or how we should handle unexpected states if they occur...
i would say for the third case, i'm wouldn't expect the condition to return false if no other assignments have been created, since the code would essentially be unassigned at that point; my understanding is that revoking doesn't necessarily mean they can't redeem the code, it's just that their slot isn't reserved anymore. perhaps a question for product?

ecommerce/extensions/checkout/mixins.py
```diff
@@ -283,3 +288,23 @@ def log_order_placement_exception(self, order_number, basket_id):
             )
         else:
             logger.exception(self.order_placement_failure_msg, payment_processor, basket_id)
+
+    def update_assigned_voucher_offer_assignment(self, order):
```
brittneyexline, line 292: where does this function get called?

ecommerce/extensions/checkout/mixins.py
```diff
@@ -283,3 +288,23 @@ def log_order_placement_exception(self, order_number, basket_id):
...
+        """
+        Update `OfferAssignment` when an assigned voucher is redeeemed.
+        """
+        basket = order.basket
+        voucher = basket.vouchers.first()
+        if voucher:
+            offer = voucher.enterprise_offer
```
brittneyexline: what if there is no enterprise offer for this voucher? not all coupons are enterprise coupons...

ecommerce/extensions/checkout/mixins.py
```diff
@@ -283,3 +288,23 @@ def log_order_placement_exception(self, order_number, basket_id):
...
+            assignment = offer.offerassignment_set.filter(code=voucher.code, user_email=basket.owner.email).exclude(
+                status__in=[OFFER_REDEEMED, OFFER_ASSIGNMENT_REVOKED]
+            ).first()
+
+            if assignment:
```
brittneyexline: i'm wondering if we should create an offer assignment with redeemed state for redemptions that were valid without having been assigned? what do you think?

brittneyexline (commented):
this is some complicated logic so thanks for ploughing through!

ecommerce/extensions/checkout/mixins.py
muhammad-ammar: added a call to this method in `handle_successful_order` above in the same class.

ecommerce/enterprise/conditions.py
muhammad-ammar: Unfortunately I am unable to understand this, perhaps an example may be helpful to understand what needs to be done?

ecommerce/enterprise/tests/test_conditions.py
muhammad-ammar: Regarding revoked, lets say if a user has an assignment and it was revoked later. Below is code assignment before and after

initial code assignment record

| code   |      user_email      |  status |
|----------|:-------------:|------:|
| ASD   |  u1@example.com | ASSIGNED |

code assignment record after revoked

| code   |      user_email      |  status |
|----------|:-------------:|------:|
| ASD   |  u1@example.com | REVOKED |

So should we allow the user(`u1@example.com`) to redeem the revoked code if the code was assigned earlier?

ecommerce/enterprise/tests/test_conditions.py
muhammad-ammar: I have updated the code and now offers are correctly created and linked to vouchers.

ecommerce/extensions/checkout/mixins.py
muhammad-ammar: updated the code, we will do nothing if voucher has no `enterprise_offer `

ecommerce/extensions/checkout/mixins.py
muhammad-ammar: I think offer assignments are specifically tracking which code is assigned to whom and its voucher application info. So I am not sure about an offer assignment without the assigned info?

ecommerce/enterprise/conditions.py
brittneyexline: so i would think about assignments more like reservations at a restaurant. if you go to a restaurant with a reservation, you automatically have seats available to you in the restaurant. but if you go to a restaurant without a reservation, you can still get seated as long as there are some available that are not in use or haven't been reserved.

now for a case that fits this description for code assignments, imagine we have a single multi use code that has 100 uses. someone in the admin tool has assigned it to 10 users who have not yet redeemed the code, and there are another 10 existing redemptions. that means the code has 80 open "slots" that any other user could use, so they should be let through. your current logic restricts this so that the code could not have been assigned at all in order to be redeemed by someone not otherwise assigned to it.

ecommerce/enterprise/tests/test_conditions.py
brittneyexline: i believe so. see my restaurant analogy above. this would be like if i made a reservation at a restaurant, then cancelled it, then showed up to the restaurant anyway. as long as the restaurant has seats available, i can get seated, but it's not guaranteed since i had cancelled the reservation.

this has been my understanding of how assignments are supposed to work in general, but i am going to check with product to clarify the behavior of revoked codes.

ecommerce/extensions/checkout/mixins.py
brittneyexline: yea, i kind of agree. i do wonder about the implications for MULTI_USE_PER_CUSTOMER codes though - what if one such code was never assigned, wouldn't we need a way to enforce that whoever first redeemed the code is the only one who can redeem it later? i will follow up with product on this as well.

muhammad-ammar:
@brittneyexline I have made more changes and handled the scenario discussed in https://github.com/edx/ecommerce/pull/2007#discussion_r241190375, added more tests.

I have more questions:
1. `MULTI_USE_PER_CUSTOMER`: Should this must be assigned before redeemed as discussed in enterprise room https://edx-internal.slack.com/archives/CD3L7U2SZ/p1544650817105100
2. Now what to do with `OFFER_ASSIGNMENT_REVOKED`? Should this be handled now or in the revoke ticket?

muhammad-ammar:
@brittneyexline I think now all the things are handled except `Revoked assignments`. I have also updated the `is_available_to_user` in `Voucher` model to handle the `MULTI_USE_PER_CUSTOMER` so that if a user has used the voucher then no other user can redeem it. Can you review please the changes? I will update the sandbox as soon as possible.

muhammad-ammar:
@brittneyexline I have updated the sandbox. Please have a look.

ecommerce/enterprise/tests/test_conditions.py
```diff
@@ -260,3 +271,350 @@ def test_is_satisfied_course_run_not_in_catalog(self):
             contains_content=False
         )
         self.assertFalse(self.condition.is_satisfied(offer, basket))
+
+
+@ddt.ddt
+class AssignableEnterpriseCustomerCondition(EnterpriseServiceMockMixin, DiscoveryTestMixin, CouponMixin, TestCase):
```
brittneyexline: nit: the class name should include Test at the end to not be confused with the class it is testing

.travis.yml
```diff
@@ -2,10 +2,6 @@ language: python
 python:
   - "2.7"
 
-branches:
```
brittneyexline: should this be part of this PR?

brittneyexline (approved):
nice work, especially with capturing the complex logic and adding the detailed test instructions!

.travis.yml
muhammad-ammar: no, deleted this so that travis build can run against my PR. will remove changes in this file before merge

muhammad-ammar:
@georgebabey This is ready to merge.

PR #2075, opened 2018-12-17
Add revenue field to order completed event

rao-abdul-mannan:
For Rockerbox integration, we need to add some fields in events being emitted from platform & ecommerce. This PR adds revenue field to order completed event.

### DIFF
--- a/ecommerce/extensions/checkout/signals.py
+++ b/ecommerce/extensions/checkout/signals.py
@@ -34,6 +34,9 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
     properties = {
         'orderId': order.number,
         'total': str(order.total_excl_tax),
+        # For Rockerbox integration, we need a field named revenue since they cannot parse a field named total.
+        # TODO: DE-1188: Remove / move Rockerbox integration code.
+        'revenue': str(order.total_excl_tax),
         'currency': order.currency,
         'discount': str(order.total_discount_incl_tax),
         'products': [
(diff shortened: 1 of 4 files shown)


robrap:
I added a comment to the ticket: https://openedx.atlassian.net/browse/DE-1171?focusedCommentId=353404&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-353404

brianhw (approved):
LGTM.   👍

HassanJaveed84 (approved):

ecommerce/extensions/checkout/signals.py
```diff
@@ -34,6 +34,8 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
     properties = {
         'orderId': order.number,
         'total': str(order.total_excl_tax),
+        # For Rockerbox integration, we need a field named revenue since they cannot parse a field named total.
```
robrap, line 37: @brianhw @macdiesel: It doesn't feel right to embed a specific integration here.  It's easy, so I get why you would like to do it.

Maybe you could determine some  proposal for being able to plug-in transformations required for specific integrations?  Or any other ideas you have? You could then create a ticket and add an additional comment like: `TODO: DE-XXXX: Remove/move Rockerbox integration code.`

Do you have any other ideas to make it clear that this is not a great pattern, but we don't want to block things, and we have a plan for how to iterate?

@rao-abdul-mannan: Thank you for adding the current comment.

macdiesel: @robrap We would be happy to put in the ticketing to remove these items. @rao-abdul-mannan could you create these tickets and mark the code with the TODO item please?

Lets talk about the plan to iterate on this offline.

robrap:
Please squash when you get a chance.  Thanks.

PR #2079, opened 2018-12-19
Add email in course purchase event

rao-abdul-mannan:


ecommerce/extensions/checkout/signals.py
```diff
@@ -36,6 +36,7 @@ def track_completed_order(sender, order=None, **kwargs):  # pylint: disable=unus
         'total': str(order.total_excl_tax),
         'currency': order.currency,
         'discount': str(order.total_discount_incl_tax),
+        'email': order.user.email,
```
robrap:
```
File "/edx/app/ecommerce/ecommerce/ecommerce/extensions/checkout/tests/test_signals.py", line 137, in _generate_event_properties
    'email': order.user.email,
AttributeError: 'NoneType' object has no attribute 'email'
```
Looks like:
1. The code should be more resilient in case email is missing, and
2. The test code also needs to have email available.

macdiesel: Thanks for the tip @robrap.

brianhw (approved):
LGTM.  👍    Though it seems annoying to me that Rockerbox requires emails, just sayin'.

macdiesel:
@robrap Could you merge this?

PR #2080, opened 2018-12-20
Add asynchronous backbone calls

zainab-amir:
[JIRA Ticket: https://openedx.atlassian.net/browse/LEARNER-6826]

#### description
- The Synchronous calls are deprecated and has detrimental effects to the end user's experience
- Add asynchronous backbone calls and then call couponApp.start() using javascript $.when()


ecommerce/static/js/apps/coupon_admin_app.js
```diff
@@ -17,27 +17,30 @@ require([
...
-            couponApp.start();
-
-            // Handle navbar clicks.
-            $('a.navbar-brand').on('click', navigate);
 
-            // Handle internal clicks
-            $app.on('click', 'a', navigate);
+            $.when(ecommerce.coupons.categories.fetch(),
+                ecommerce.coupons.catalogs.fetch(),
+                ecommerce.coupons.enterprise_customers.fetch()).always(
```
georgebabey, line 42: Although always starting the app follows the existing logic, it seems strange to me we would just carry on if there was an error fetching the data.

The goal of this was to remove the synchronous calls though, and this accomplishes that so looks good here.

georgebabey (approved):
Looks good, thanks for cleaning this up!

PR #2082, opened 2018-12-21
ENT-1417: Update assignment endpoint to trigger emails

irfanuddinahmad:
Description: The PR updates the assignment endpoint to take an additional parameter (email template) and triggers the emails to be fired after successful assignments.

The new message formats are:

Request:
```
{
    "template": "Customized template",
    "emails": ["bexline+1@edx.org","bexline+2@edx.org","bexline+3@edx.org"],
    "codes":["GHGSVMQE6XKNZE2V","CYOFOH6XTEEBMKVN","IFXGPNTP6SKRENTC"]
}
```
Response:
```
{
    "offer_assignments": [
        {
            "id": 555 
            "user_email": "bexline+1@edx.org",
            "code": "CYOFOH6XTEEBMKVN",
        },
        {
            "id": 556
            "user_email": "bexline+2@edx.org",
            "code": "IFXGPNTP6SKRENTC"
        },
        {
            "id": 557
            "user_email": "bexline+3@edx.org",
            "code": "GHGSVMQE6XKNZE2V"
        }
    ]
}
```

Dependency: This PR is dependant on https://github.com/edx/ecommerce/pull/1998

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 import logging
+from datetime import timedelta
 from decimal import Decimal
 
 import waffle
@@ -25,6 +26,7 @@ from ecommerce.core.url_utils import get_ecommerce_url
 from ecommerce.courses.models import Course
 from ecommerce.entitlements.utils import create_or_update_course_entitlement
 from ecommerce.extensions.offer.constants import OFFER_ASSIGNMENT_REVOKED, OFFER_MAX_USES_DEFAULT
+from ecommerce.extensions.offer.utils import send_assigned_offer_email
 from ecommerce.invoice.models import Invoice
 
 logger = logging.getLogger(__name__)
@@ -1016,7 +1018,7 @@ class ProviderSerializer(serializers.Serializer):  # pylint: disable=abstract-me
 class OfferAssignmentSerializer(serializers.ModelSerializer):
     class Meta(object):
         model = OfferAssignment
-        fields = ('user_email', 'code')
+        fields = ('id', 'user_email', 'code')
 
 
 class CouponCodeAssignmentSerializer(serializers.Serializer):  # pylint: disable=abstract-method
@@ -1034,21 +1036,27 @@ class CouponCodeAssignmentSerializer(serializers.Serializer):  # pylint: disable
         """Create OfferAssignment objects for each email and the available_assignments determined from validation."""
         emails = validated_data.get('emails')
         voucher_usage_type = validated_data.pop('voucher_usage_type')
+        email_template = validated_data.pop('template')
         available_assignments = validated_data.pop('available_assignments')
         email_iterator = iter(emails)
         offer_assignments = []
+        emails_already_sent = set()
 
         for code in available_assignments:
             offer = available_assignments[code]['offer']
             email = next(email_iterator) if voucher_usage_type == Voucher.MULTI_USE_PER_CUSTOMER else None
             for _ in range(available_assignments[code]['num_slots']):
-                offer_assignments.append(
-                    OfferAssignment.objects.create(
-                        offer=offer,
-                        code=code,
-                        user_email=email or next(email_iterator),
-                    )
+                new_offer_assignment = OfferAssignment.objects.create(
+                    offer=offer,
+                    code=code,
+                    user_email=email or next(email_iterator),
                 )
+                offer_assignments.append(new_offer_assignment)
+                # Start async email task. For MULTI_USE_PER_CUSTOMER, a single email is sent
+                email_code_pair = frozenset((new_offer_assignment.user_email, new_offer_assignment.code))
+                if email_code_pair not in emails_already_sent:
+                    self._trigger_email_sending_task(email_template, new_offer_assignment, voucher_usage_type)
+                    emails_already_sent.add(email_code_pair)
 
         validated_data['offer_assignments'] = offer_assignments
         return validated_data
@@ -1061,6 +1069,7 @@ class CouponCodeAssignmentSerializer(serializers.Serializer):  # pylint: disable
         codes = data.get('codes')
         emails = data.get('emails')
         coupon = self.context.get('coupon')
+        template = self.context.get('template')
         available_assignments = {}
         vouchers = coupon.attr.coupon_vouchers.vouchers
 
@@ -1123,4 +1132,29 @@ class CouponCodeAssignmentSerializer(serializers.Serializer):  # pylint: disable
         # Add available_assignments to the validated data so that we can perform the assignments in create.
         data['voucher_usage_type'] = voucher_usage_type
         data['available_assignments'] = available_assignments
+        data['template'] = template
         return data
+
+    def _trigger_email_sending_task(self, template, assigned_offer, voucher_usage_type):
+        """
+        Schedule async task to send email to the learner who has been assigned the code.
+        """
+        url = get_ecommerce_url('/coupons/offer/')
+        enrollment_url = '{url}?code={code}'.format(url=url, code=assigned_offer.code)
+        code_expiration_date = assigned_offer.created + timedelta(days=365)
+        redemptions_remaining = (
+            assigned_offer.offer.max_global_applications if voucher_usage_type == Voucher.MULTI_USE_PER_CUSTOMER else 1
+        )
+        try:
+            send_assigned_offer_email(
+                template=template,
+                offer_assignment_id=assigned_offer.id,
+                learner_email=assigned_offer.user_email,
+                code=assigned_offer.code,
+                enrollment_url=enrollment_url,
+                redemptions_remaining=redemptions_remaining,
+                code_expiration_date=code_expiration_date.strftime('%d %B,%Y')
+            )
+        except Exception as exc:  # pylint: disable=broad-except
+            logger.exception(
+                '[Offer Assignment] Email for offer_assignment_id: %d raised exception: %r', assigned_offer.id, exc)
--- a/ecommerce/extensions/api/v2/views/enterprise.py
+++ b/ecommerce/extensions/api/v2/views/enterprise.py
@@ -1,7 +1,6 @@
 from __future__ import unicode_literals
 
 import logging
-
 import waffle
 from django.core.exceptions import ValidationError
 from oscar.core.loading import get_model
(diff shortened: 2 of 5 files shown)


irfanuddinahmad:
@brittneyexline @georgebabey The PR contains commits from ENT-1362's branch as well. The last commit contains the work for ENT-1417

ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -0,0 +1,163 @@
...
+            'status': 'success'
+        }
+        Returns a JSON object of the following format:
+       {
+            'offer_id': '555',
+            'send_id': 'XBEn85WnoQJsIhk6'
+            'status': 'updated'
+            'error': ''
+        }
+        Keys:
```
christopappas: What is the `send_id` here?

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -242,12 +245,47 @@ def assign(self, request, pk):  # pylint: disable=unused-argument
...
+        """
+        Schedule async tasks to send emails to learners who have been assigned codes.
+        """
+        emails_status = []
+        for offer_assignment in offer_assignments:
+            assigned_offer = OfferAssignment.objects.get(id=offer_assignment['id'])
+            url = get_ecommerce_url('/coupons/offer/')
+            enrollment_url = '{url}?code={code}'.format(url=url, code=assigned_offer.code)
+            code_expiration_date = assigned_offer.created + timedelta(days=365)
+            email_send_status = send_assigned_offer_email(
```
christopappas: ~If a task is successfully kicked off, this will return true. But what if the task that is successfully spun off fails to actually finish running? We still return True.~

~Imagine there are 4 workers, and 1 of them accepts tasks, but, for some reason, never actually processes them, such that 1 out of 4 email tasks never actually is run. I need to look at this code a little longer to make sure I understand it fully, but I've seen things like this happen before, and wanted to make sure we don't get bitten by this~

~Keeping track of what emails actually get sent seems tricky. If there's something I'm overlooking presently, definitely let me know, but I want to continue thinking about this one!~

~EDIT: we might not actually care too much about this, but wanted to drop the thoughts here just in case~

EDIT 2: I just learned about sailthru and what it actually does, you can ignore this

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -242,12 +245,47 @@ def assign(self, request, pk):  # pylint: disable=unused-argument
...
-            return Response(serializer.data, status=status.HTTP_200_OK)
-
+            tasks_status = self._trigger_email_sending_task(
+                email_template,
+                serializer.data.get('offer_assignments')
+            )
+            return Response({'offer_assignments': tasks_status}, status=status.HTTP_200_OK)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    def _trigger_email_sending_task(self, template, offer_assignments):
```
brittneyexline: why not do this in the serializer? then you wouldn't have to go through the serialized statuses to add the email, you could do it within the loop that they're generated.

ecommerce/extensions/api/v2/views/enterprise.py
brittneyexline: another thing to consider here is that we are creating multiple OfferAssignment objects in the MULTI_USE_PER_CUSTOMER case, but I doubt we want to send an email for each object since it's the same user and code... I'm not sure what to do about it though 😕

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: send_id is the message identifier from Sailthru.

christopappas: ah, okay. thanks :)

ecommerce/extensions/api/v2/views/enterprise.py
irfanuddinahmad: Thanks. Done.

ecommerce/extensions/api/v2/views/enterprise.py
irfanuddinahmad: Added support for suppressing emails for the MULTI_USE_PER_CUSTOMER case.

ecommerce/extensions/api/serializers.py
```diff
@@ -1046,9 +1054,11 @@ def validate(self, data):
         Validate that the given emails can be assigned to a slot in the coupon, to the codes if specified.
         A slot is a potential redemption of a voucher contained within the top level Coupon.
         """
+
```
mushtaqak: nit: unintended extra line ?

ecommerce/extensions/api/serializers.py
```diff
@@ -1111,4 +1121,22 @@ def validate(self, data):
...
+        data['template'] = template
         return data
+
+    def _trigger_email_sending_task(self, template, assigned_offer):
+        """
+        Schedule async task to send email to the learner who has been assigned the code.
+        """
+        url = get_ecommerce_url('/coupons/offer/')
+        enrollment_url = '{url}?code={code}'.format(url=url, code=assigned_offer.code)
+        code_expiration_date = assigned_offer.created + timedelta(days=365)
```
mushtaqak, line 1144: Does this count leap year ? or do we need to care about it? Just bringing it up ...

mushtaqak (approved):
Looks good 👍

brittneyexline (approved):

ecommerce/extensions/api/v2/views/enterprise.py
irfanuddinahmad: @brittneyexline Added support for bulk status update (async) of OfferAssignment objects in https://github.com/edx/ecommerce/pull/1998
based on (user_email, code) ... can we assume that a particular (user_email, code) combination will bring in all the relevant rows whose status needs to be updated?
https://github.com/edx/ecommerce/pull/1998/commits/40017161270f256dab4bcb477cff24af0cfe6c31

ecommerce/extensions/offer/utils.py
```diff
@@ -99,3 +104,49 @@ def render_email_confirmation_if_required(request, offer, product):
...
+    try:
+        email_body = template.format(
+            REDEMPTIONS_REMAINING=redemptions_remaining,
+            USER_EMAIL=learner_email,
+            ENROLLMENT_URL=enrollment_url,
+            CODE=code,
+            EXPIRATION_DATE=code_expiration_date
+        )
+        send_offer_assignment_email.delay(learner_email, offer_assignment_id, email_subject, email_body)
+    except Exception as exc:  # pylint: disable=broad-except
```
brittneyexline: in my PR for revoke i moved this to the serializer so we could return the exception message to the caller (https://github.com/edx/ecommerce/pull/2087/files#diff-852a6b45e4fa1255461cf3fa6e5ceda7R1202), which i think is preferable. what do you think? i'd like these email util methods to be consistent with one another

irfanuddinahmad: Changed as suggested.

ecommerce/extensions/api/serializers.py
irfanuddinahmad: Removed

ecommerce/extensions/api/serializers.py
irfanuddinahmad: @jshin-edx @georgebabey We are calculating the code expiration date by adding 365 days without any leap year checks. Is this ok?

georgebabey: @irfanuddinahmad Okay in my book

ecommerce/extensions/offer/utils.py
irfanuddinahmad: @brittneyexline I have moved the exception handling to the serializer. Do you think the 'detail' key value as used in CouponCodeRevokeBulkSerializer is required for the CouponCodeAssignmentSerializer as well?

irfanuddinahmad: For email, an exception can be raised while we are initiating the ecommerce-worker celery task and also within ecommerce worker. In both cases, the offer assignment status remains at the initial default status of 'EMAIL_PENDING'. Hence, if we send back the 'detail', it would only cover the errors raised while initiating the ecommerce-worker celery task.

ecommerce/extensions/api/serializers.py
```diff
@@ -1123,4 +1132,30 @@ def validate(self, data):
...
+    def _trigger_email_sending_task(self, template, assigned_offer, voucher_usage_type):
+        """
+        Schedule async task to send email to the learner who has been assigned the code.
+        """
+        url = get_ecommerce_url('/coupons/offer/')
+        enrollment_url = '{url}?code={code}'.format(url=url, code=assigned_offer.code)
+        code_expiration_date = assigned_offer.created + timedelta(days=365)
+        multi_use_voucher_types = [Voucher.MULTI_USE_PER_CUSTOMER, Voucher.MULTI_USE]
+        redemptions_remaining = (
+            assigned_offer.offer.max_global_applications if voucher_usage_type in multi_use_voucher_types else 1
```
irfanuddinahmad: @brittneyexline is this method of computing initial available redemptions ok? ... is there a better way?

brittneyexline: i think MULTI_USE_PER_CUSTOMER is the only case when redemptions remaining should be max_global_applications and not 1.

ecommerce/extensions/offer/utils.py
brittneyexline: in my PR, i wrapped the database changes in the try/except as well, in case there is an error there too. i guess it depends on how important it is for the caller to have that info bubbled up - in the case of the email task failing, it is more of a warning than a failure, if the offer assignment creation succeeded. it's also tricky here because there aren't explicit independent operations like there is for revoke. for now we can leave this as is, but we may need to revisit at some point.

ecommerce/extensions/api/serializers.py
```diff
@@ -1123,4 +1132,30 @@ def validate(self, data):
...
+                offer_assignment_id=assigned_offer.id,
+                learner_email=assigned_offer.user_email,
+                code=assigned_offer.code,
+                enrollment_url=enrollment_url,
+                redemptions_remaining=redemptions_remaining,
+                code_expiration_date=code_expiration_date.strftime('%d %B,%Y')
+            )
+        except Exception as exc:  # pylint: disable=broad-except
+            logger.exception(
+                '[Offer Assignment] Offer assignment email async task raised: %r', exc)
```
brittneyexline: it would be good to put more detail in this log message, like the offer assignment id, so we can more easily debug.

irfanuddinahmad: Done

ecommerce/extensions/api/serializers.py
irfanuddinahmad: Updated.

PR #2084, opened 2018-12-24
Coupon codes CSV

mushtaqak:
[ENT-1346](https://openedx.atlassian.net/browse/ENT-1346)

The purpose of this PR is to add the functionality for `/codes` endpoint to return csv on demand.

### DIFF
--- a/ecommerce/extensions/api/v2/urls.py
+++ b/ecommerce/extensions/api/v2/urls.py
@@ -1,4 +1,5 @@
 from django.conf.urls import include, url
+from rest_framework.urlpatterns import format_suffix_patterns
 from rest_framework_extensions.routers import ExtendedSimpleRouter
 
 from ecommerce.core.constants import COURSE_ID_PATTERN
@@ -121,3 +122,4 @@ router.register(r'vouchers', voucher_views.VoucherViewSet, base_name='vouchers')
 router.register(r'stockrecords', stockrecords_views.StockRecordViewSet, base_name='stockrecords')
 
 urlpatterns += router.urls
+urlpatterns = format_suffix_patterns(urlpatterns)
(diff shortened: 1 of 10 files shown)


mushtaqak:
@irfanuddinahmad Can you have a look at this?

brittneyexline (approved):
looks good! just need the quality checks to pass

ecommerce/extensions/api/v2/urls.py
```diff
@@ -121,3 +122,4 @@
 router.register(r'stockrecords', stockrecords_views.StockRecordViewSet, base_name='stockrecords')
 
 urlpatterns += router.urls
+urlpatterns = format_suffix_patterns(urlpatterns)
```
muhammad-ammar, line 125: This will add format suffix support for all the urls, is this expected?

muhammad-ammar (changes requested):
@mushtaqak According to the changes it is mandatory that `no_page` param must be present in query params and appending the format alone is not enough. As per my understanding this is unintuitive.

`https://ecommerce-business.sandbox.edx.org/api/v2/enterprise/coupons/124/codes.csv/` is more intuitive than
`https://ecommerce-business.sandbox.edx.org/api/v2/enterprise/coupons/124/codes.csv/?no_page`

As per my understanding we can make the decision on `format` keyword argument in codes view. If format is other than `None` than do not paginate else paginate.

muhammad-ammar (approved):
@mushtaqak Looks good to me.

ecommerce/extensions/api/v2/urls.py
mushtaqak: I guess this is OK, since this is used across many edX repositories.

mushtaqak:
>As per my understanding we can make the decision on format keyword argument in codes view. If format is other than None than do not paginate else paginate.

@muhammad-ammar I have removed `no_page` query param as per above feedback :)

mushtaqak:
@georgebabey Can do a quick review and merge - I don't merge rights :P ?

robrap:
FYI: I'm merging on behalf of @georgebabey due to codecov issues.

PR #2085, opened 2018-12-31
ENT-1172 Add notification email for enterprise coupons

irfanuddinahmad:
**Description:** 
This PR adds a notification email field to https://ecommerce-business.sandbox.edx.org/enterprise/coupons/new/ and sends the email to that address whenever an enterprise coupon is created.

**Test Instructions:**
Deployed to sandbox

1: Open https://ecommerce-business.sandbox.edx.org/enterprise/coupons/new/ in a browser
2: Create a new coupon and supply a notification email **[This is failing currently because AWS SES is not working for sandbox]**
3: Verify that the notification email is received on the supplied notification email address

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_enterprise.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_enterprise.py
@@ -8,6 +8,7 @@ from uuid import uuid4
 import ddt
 import httpretty
 import mock
+from django.conf import settings
 from django.urls import reverse
 from django.utils.timezone import now
 from oscar.core.loading import get_model
@@ -117,10 +118,16 @@ class EnterpriseCouponViewSetTest(CouponMixin, DiscoveryTestMixin, DiscoveryMock
             'voucher_type': Voucher.SINGLE_USE,
             'enterprise_customer': {'name': 'test enterprise', 'id': str(uuid4()).decode('utf-8')},
             'enterprise_customer_catalog': str(uuid4()).decode('utf-8'),
+            'notify_email': 'batman@gotham.comics',
         }
 
         self.course = CourseFactory(id='course-v1:test-org+course+run', partner=self.partner)
         self.verified_seat = self.course.create_or_update_seat('verified', False, 100)
+        self.enterprise_slug = 'batman'
+
+        patcher = mock.patch('ecommerce.extensions.api.v2.utils.send_mail')
+        self.send_mail_patcher = patcher.start()
+        self.addCleanup(patcher.stop)
 
     def get_coupon_voucher(self, coupon):
         """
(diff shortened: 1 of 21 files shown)


brittneyexline (approved):
looks good, nice work!

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -125,11 +126,21 @@ def create(self, request, *args, **kwargs):
...
+                        cleaned_voucher_data['enterprise_customer'],
+                        coupon_product.id
+                    )
                 return Response(response_data, status=status.HTTP_200_OK)
         except ValidationError as e:
             raise serializers.ValidationError(e.message)
 
+    @staticmethod
+    def send_codes_availability_email(site, email_address, enterprise_id, coupon_id):
+        pass
```
mushtaqak, line 142: This is the case for old Coupon screen (https://ecommerce-business.sandbox.edx.org/coupons/new/) ? we just want to do nothing in this case. right ?

ecommerce/extensions/api/v2/views/coupons.py
```diff
@@ -444,6 +463,11 @@ def update_coupon_product_data(self, request_data, coupon):
             coupon.attr.note = note
             coupon.save()
 
+        notify_email = request_data.get('notify_email')
+        if notify_email is not None:
```
mushtaqak: I think this should be just `if notify_email:` otherwise we are getting an error when we add a notification email and then remove it ... thus form will make it "" - empty string and this will be bypassed here and model save does not handle empty string notification email ... and it gives error, making us unable to remove a notification email.

```suggestion
        if notify_email:
```
screen shot 2019-01-10 at 1 11 12 pm

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -117,10 +118,16 @@ def setUp(self):
...
             'enterprise_customer_catalog': str(uuid4()).decode('utf-8'),
+            'notify_email': 'batman@gotham.comics',
         }
 
         self.course = CourseFactory(id='course-v1:test-org+course+run', partner=self.partner)
         self.verified_seat = self.course.create_or_update_seat('verified', False, 100)
+        self.enterprise_slug = 'batman'
+
+        patcher = mock.patch('ecommerce.extensions.api.v2.utils.send_mail')
+        self.send_mail_patcher = patcher.start()
```
mushtaqak, line 129: Nice :)

mushtaqak (changes requested):
Great work 👍 , just one miner change for removing a notification email.

ecommerce/extensions/api/v2/views/coupons.py
muhammad-ammar: @mushtaqak I have fixed the issue. Can you please test it again?

mushtaqak (approved):
👍

PR #2088, opened 2019-01-04
Enterprise Coupon ADR

brittneyexline:


christopappas (approved):
I'm not sure if there is a specific rubric for approving an ADR, but it was an informative read, and looks to be a great summary of all the work done, so 👍 from me

muhammad-ammar (approved):
Looks good to me.

@georgebabey It would be great if you can also take a look this.

docs/decisions/0003-enterprise-coupons.rst
```diff
@@ -0,0 +1,151 @@
+3. Enterprise Coupons
+---------------------
+
+Status
+------
+
+Accepted?
```
georgebabey: `Accepted` once it merges!

georgebabey (approved):
Looks great, thanks for getting this written up!

PR #2098, opened 2019-01-21
ENT-1385 Offer assignment reminder email api

irfanuddinahmad:
Description: This PR provides api for sending reminder emails to learners regarding their un-redeemed offers.

JIRA: https://openedx.atlassian.net/browse/ENT-1385

The endpoint accepts an email template, and a list of email/code combinations. Each of these is evaluated independently, so if one email/code combination is not valid, others specified will still go through if they are valid. The response will indicate the success of each email/code pair sent.

**Testing Instructions**

Hitting the endpoint with a code that is not part of the Coupon should fail.
Request:
```POST /api/v2/enterprise/coupons/419/remind/ HTTP/1.1
Authorization: JWT {jwt_token}
Content-Type: application/json
{"template":"", "assignments":[{"email":"johndoe@edx.org", "code":"abcdef"}]}
```

Response:
```[
    {
        "non_field_errors": [
            "Code abcdef is not associated with this Coupon"
        ]
    }
]
```

Hitting the endpoint with a code/email that do not have an active assignment should fail.
Request:
```POST /api/v2/enterprise/coupons/419/remind/ HTTP/1.1
Authorization: JWT {jwt_token}
Content-Type: application/json
{"template":"", "assignments":[{"email":"johndoe@edx.org", "code":"TDXSDBTU3YSVCK3X"}]}
```

Response:
```[
    {
        "non_field_errors": [
            "No assignments exist for user johndoe@edx.org and code TDXSDBTU3YSVCK3X"
        ]
    }
]
```

Reminder email for the code:
``` POST /api/v2/enterprise/coupons/419/remind/ HTTP/1.1
Authorization: JWT {JWT_TOKEN}
Content-Type: application/json
{"template":"
    This is a reminder email that your learning manager has provided you with a access code to take a course at edX.
    You have redeemed this code {REDEEMED_OFFER_COUNT} of times out of {TOTAL_OFFER_COUNT} number of available course redemptions.
    
    edX login: {USER_EMAIL}
    Enrollment url: {ENROLLMENT_URL}
    Access Code: {CODE}
    Expiration date: {EXPIRATION_DATE}

    You may go directly to the Enrollment URL to view courses that are available for this code
    or you can insert the access code at check out under "coupon code" for applicable courses.

    For any questions, please reach out to your Learning Manager.
", "assignments": [{"email":"johndoe@edx.org", "code": "WYZIOEZT2N7VYJ5K"}]}
[
    {
        "code": "WYZIOEZT2N7VYJ5K",
        "email": "johndoe@edx.org",
        "detail": "success"
    }
]
```

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -30,9 +30,14 @@ from ecommerce.extensions.offer.constants import (
     OFFER_ASSIGNED,
     OFFER_ASSIGNMENT_EMAIL_PENDING,
     OFFER_ASSIGNMENT_REVOKED,
-    OFFER_MAX_USES_DEFAULT
+    OFFER_MAX_USES_DEFAULT,
+    OFFER_REDEEMED
 )
-from ecommerce.extensions.offer.utils import send_assigned_offer_email, send_revoked_offer_email
+from ecommerce.extensions.offer.utils import (
+    send_assigned_offer_email,
+    send_assigned_offer_reminder_email,
+    send_revoked_offer_email
+)
 from ecommerce.invoice.models import Invoice
 
 logger = logging.getLogger(__name__)
@@ -1166,7 +1171,7 @@ class CouponCodeAssignmentSerializer(serializers.Serializer):  # pylint: disable
                 '[Offer Assignment] Email for offer_assignment_id: %d raised exception: %r', assigned_offer.id, exc)
 
 
-class CouponCodeRevokeBulkSerializer(serializers.ListSerializer):  # pylint: disable=abstract-method
+class CouponCodeRevokeRemindBulkSerializer(serializers.ListSerializer):  # pylint: disable=abstract-method
 
     def to_internal_value(self, data):
         """
@@ -1216,10 +1221,36 @@ class CouponCodeRevokeBulkSerializer(serializers.ListSerializer):  # pylint: dis
         ]
 
 
-class CouponCodeRevokeSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+class CouponCodeMixin(object):
+
+    def validate_coupon_has_code(self, coupon, code):
+        """
+        Validate that the code is associated with the coupon
+        :param coupon: (Product): Coupon product associated with vouchers
+        :param code: (str): Code associated with the voucher
+        :raises rest_framework.exceptions.ValidationError in case code is not associated with the coupon
+        """
+        if not coupon.attr.coupon_vouchers.vouchers.filter(code=code).exists():
+            raise serializers.ValidationError('Code {} is not associated with this Coupon'.format(code))
+
+    def get_unredeemed_offer_assignments(self, code, email):
+        """
+        Returns offer assignments associated with the code and email
+        :param code: (str): Code associated with the voucher
+        :param email: (str): Learner email
+        :return: QuerySet containing offer assignments associated with the code and email
+        """
+        return OfferAssignment.objects.filter(
+            code=code,
+            user_email=email,
+            status__in=[OFFER_ASSIGNED, OFFER_ASSIGNMENT_EMAIL_PENDING]
+        )
+
+
+class CouponCodeRevokeSerializer(CouponCodeMixin, serializers.Serializer):  # pylint: disable=abstract-method
 
     class Meta:  # pylint: disable=old-style-class
-        list_serializer_class = CouponCodeRevokeBulkSerializer
+        list_serializer_class = CouponCodeRevokeRemindBulkSerializer
 
     code = serializers.CharField(required=True)
     email = serializers.EmailField(required=True)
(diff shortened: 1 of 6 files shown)


irfanuddinahmad:
Reviewers: Please review the last commit since the earlier commits contain work from ENT-1384 and ENT-1417

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -283,3 +284,21 @@ def revoke(self, request, pk):  # pylint: disable=unused-argument
...
 
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    @detail_route(methods=['post'])
+    def remind(self, request, pk):  # pylint: disable=unused-argument
+        """
+        Remind users of pending offer assignments by email.
+        """
+        coupon = self.get_object()
+        email_template = request.data.pop('template', None)
```
muhammad-ammar, line 294: shouldn't the template always present in the request data? as per my understanding template is mandatory so we should raise error if template is not present.

ecommerce/extensions/api/serializers.py
```diff
@@ -1275,3 +1280,87 @@ def validate(self, data):
...
+    email = serializers.EmailField(required=True)
+    detail = serializers.CharField(read_only=True)
+
+    def create(self, validated_data):
+        """
+        Send remind email(s) for pending OfferAssignments.
+        """
+        offer_assignments = validated_data.get('offer_assignments')
+        code = validated_data.get('redeemed_offer_count')
+        email = validated_data.get('total_offer_count')
```
muhammad-ammar: looks like the above two lines are left by mistake?

ecommerce/extensions/api/serializers.py
```diff
@@ -1275,3 +1280,87 @@ def validate(self, data):
...
+        redeemed_offer_count = validated_data.get('redeemed_offer_count')
+        total_offer_count = validated_data.get('total_offer_count')
+        template = self.context.get('template')
+        detail = 'success'
+
+        try:
+            if template and offer_assignments:
```
muhammad-ammar: I think we don't need this `if condition` because if `template` is not present than control should not get here, we should validate the template in the view and raise error. Also we don't need to check for `offer_assignments` because if there are no `offer_assignments` than `validate` method will raise error and control will not get here.

ecommerce/settings/base.py
```diff
@@ -652,6 +652,22 @@
...
+    Enrollment url: {ENROLLMENT_URL}
+    Access Code: {CODE}
+    Expiration date: {EXPIRATION_DATE}
+
+    You may go directly to the Enrollment URL to view courses that are available for this code
+    or you can insert the access code at check out under "coupon code" for applicable courses.
+
+    For any questions, please reach out to your Learning Manager.
+'''
+OFFER_ASSIGNMENT_EMAIL_REMINDER_DEFAULT_SUBJECT = 'Reminder on edX course assignment'
```
muhammad-ammar, line 669: email subject is not clear enough to convey the message? What about `Reminder to redeem your course coupon` or something better?

ecommerce/settings/base.py
```diff
@@ -652,6 +652,22 @@
 OFFER_ASSIGNMENT_EMAIL_DEFAULT_SUBJECT = 'New edX course assignment'
 OFFER_REVOKE_EMAIL_DEFAULT_SUBJECT = 'edX Course Assignment Revoked'
 
+OFFER_ASSIGNMENT_EMAIL_REMINDER_DEFAULT_TEMPLATE = '''
+    This is a reminder email that your learning manager has provided you with a access code to take a course at edX.
+    You have redeemed this code {REDEEMED_OFFER_COUNT} of times out of {TOTAL_OFFER_COUNT} number of available course redemptions.
+    
+    edX login: {USER_EMAIL}
+    Enrollment url: {ENROLLMENT_URL}
```
muhammad-ammar, line 660: nit: `url` -> `URL`

ecommerce/settings/base.py
```diff
@@ -652,6 +652,22 @@
...
+    Access Code: {CODE}
+    Expiration date: {EXPIRATION_DATE}
+
+    You may go directly to the Enrollment URL to view courses that are available for this code
+    or you can insert the access code at check out under "coupon code" for applicable courses.
```
muhammad-ammar, line 665: Is this expected that a leaner knows the flow mention in `you can insert the access code at check out under "coupon code" for applicable courses`?

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -283,3 +284,21 @@ def revoke(self, request, pk):  # pylint: disable=unused-argument
...
+        serializer = CouponCodeRemindSerializer(
+            data=request.data.get('assignments'),
+            many=True,
+            context={'coupon': coupon, 'template': email_template}
+        )
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
muhammad-ammar, line 304: we are always sending `200` even if there are errors. Is this OK?

mushtaqak:
I did end-to-end testing on my local docker machine with `CELERY_ALWAYS_EAGER` set `True` in `devstack.py`. Things are looking good 👍  (although we might need to think over the response message in case of some of the reminders failed)

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -1054,3 +1054,136 @@ def test_coupon_codes_revoke_bulk(self):
...
+                '/api/v2/enterprise/coupons/{}/remind/'.format(coupon_id),
+                {'template': 'Test template', 'assignments': [
+                    {'email': offer_assignment.user_email, 'code': offer_assignment.code},
+                    {'email': 'test3@example.com', 'code': 'RANDOMCODE'},
+                ]}
+            )
+        response = response.json()
+        assert response == [
+            {'email': offer_assignment.user_email, 'code': offer_assignment.code, 'detail': 'success'},
+            {'non_field_errors': ['Code RANDOMCODE is not associated with this Coupon']},
```
muhammad-ammar, line 1187: Though not needed now but it would be nice if response format for success and failure is same, instead of `non_field_errors` what about 
```python
{'email': offer_assignment.user_email, 'code': offer_assignment.code, 'detail': 'failure', 'error_message': 'Code RANDOMCODE is not associated with this Coupon'}
```
we need to do something like below in `to_internal_value` method in `CouponCodeRevokeRemindBulkSerializer` 

```python
ret = []
for item in data:
    try:
        validated = self.child.run_validation(item)
    except serializers.ValidationError as exc:
        ret.append(dict(item, **{'detail': 'failure', 'message': exc.detail['non_field_errors'][0]}))
    else:
        ret.append(validated)
```
This will help us on frontend side to attach the error with correct row if we ever want to show the individual errors.

muhammad-ammar (changes requested):
@irfanuddinahmad I am done with first pass of the review.

ecommerce/settings/base.py
irfanuddinahmad: @georgebabey This was provided by @jshin-edx

ecommerce/settings/base.py
irfanuddinahmad: @georgebabey This was provided by @jshin-edx

ecommerce/extensions/api/serializers.py
irfanuddinahmad: the perils of copy-paste!

ecommerce/settings/base.py
irfanuddinahmad: @georgebabey This was provided by @jshin-edx

ecommerce/extensions/api/v2/views/enterprise.py
irfanuddinahmad: @georgebabey We are sending a per record status ... what should be the overall status?

ecommerce/extensions/api/v2/views/enterprise.py
irfanuddinahmad: Done

ecommerce/extensions/api/serializers.py
irfanuddinahmad: Done

ecommerce/extensions/api/serializers.py
irfanuddinahmad: Removed

muhammad-ammar (approved):

ecommerce/extensions/api/serializers.py
```diff
@@ -1275,3 +1280,86 @@ def validate(self, data):
...
+
+    def validate(self, data):
+        """
+        Validate that the code is part of the Coupon the code and email provided have an active OfferAssignment.
+        """
+        code = data.get('code')
+        email = data.get('email')
+
+        coupon = self.context.get('coupon')
+        if not coupon.attr.coupon_vouchers.vouchers.filter(code=code).exists():
```
brittneyexline: i wonder if there is an opportunity to DRY out some of this validation logic, since it is quite similar to the revoke serializer's logic?

ecommerce/extensions/api/v2/views/enterprise.py
brittneyexline: given i'm the one who introduced this pattern with revoke, i think it's ok that the status is 200 even if individual requests within the larger request fail - the overall request still succeeded by going through each individual one. if the overall request format is one that we cannot process, then it is a bad request.

brittneyexline (approved):
let me know what you think about consolidating the serializer validation code, and then i think we're good to merge/release this!

ecommerce/extensions/api/serializers.py
irfanuddinahmad: Done. Still a bit WET though! :)

ecommerce/extensions/api/serializers.py
```diff
@@ -1259,19 +1265,93 @@ def validate(self, data):
         """
         code = data.get('code')
         email = data.get('email')
-
         coupon = self.context.get('coupon')
-        if not coupon.attr.coupon_vouchers.vouchers.filter(code=code).exists():
-            raise serializers.ValidationError('Code {} is not associated with this Coupon'.format(code))
-
+        validate_coupon_has_code(coupon, code)
```
brittneyexline: i think it would be better to create a serializer class or mixin that had this method as well as one that checks if there are active offer assignments that exist (since both validate methods perform this as well). then both serializers could inherit from it, and possibly the list serializer could be specified at this level too.

irfanuddinahmad: Done!

PR #2099, opened 2019-01-22
Filter coupon details

mushtaqak:
Filter coupon details - [ENT-1345](https://openedx.atlassian.net/browse/ENT-1345)


ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +199,102 @@ def update_offer_data(self, request_data, vouchers, site):
...
+            SINGLE_USE, 3 quantiy, 0 assignment, 3 redemptions --> 0 slot available.
+            SINGLE_USE, 3 quantiy, 1 assignment, no redemption --> 2 slots available.
+            SINGLE_USE, 3 quantiy, no assignment, 1 redemption --> 2 slots available.
+            SINGLE_USE, 3 quantiy, 1 assignment, 1 assigned redemption --> 2 slots available.
+            SINGLE_USE, 3 quantiy, 1 assignment, 1 assigned + other redemption --> 1 slots available.
+            SINGLE_USE, 3 quantiy, 1 assignment, 1 assigned + 2 other redemptions --> 0 slots available.
+            SINGLE_USE, 3 quantiy, 3 assignment, 3 assigned redemptions --> 0 slots available.
+
+
+            MULTI_USE, 3 quantiy, 10 max_uses, no assignment, no redemption --> 3 slots available.
```
muhammad-ammar: I think in this scenario we will have `10 slots available` per code

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +199,102 @@ def update_offer_data(self, request_data, vouchers, site):
...
+    def get_partial_redeemed_coupon_vouchers(self, coupon_vouchers):
+        """
+        Returns assigned vouchers having at-least one redemption.
+        """
+        # TODO: FAQ - Single use has only one redemption? what to do in this case? return empty queryset?
+        return coupon_vouchers.filter(num_orders_gte=1, offers__max_global_applications__gt=F('num_orders'))
+
+    def get_fully_redeemed_coupon_vouchers(self, coupon_vouchers):
+        """
+        Returns vouchers that are fully redeemed, these vouchers do not have any assignment or redemption left.
```
muhammad-ammar: nit: `Return vouchers ...`

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +199,102 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        ).values_list('code', flat=True)
+
+    def get_redeemable_vouchers(self, coupon_vouchers):
+        """
+        Return vouchers which are redeemable. These vouchers are not fully redeemed.
+        """
+        vouchers_with_full_redemptions = self.get_fully_redeemed_coupon_vouchers(coupon_vouchers)
+        redeemable_vouchers =  coupon_vouchers.exclude(pk__in=vouchers_with_full_redemptions)
+
+        assigne_codes = self.get_assigned_codes()
```
muhammad-ammar: `assigne_codes` -> `assigned_codes `

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +199,102 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        redeemable_vouchers = self.get_redeemable_vouchers(coupon_vouchers)
+        unredeemed_vouchers = redeemable_vouchers.filter(code__in=assigned_codes)
+        return unredeemed_vouchers
+
+    def get_partial_redeemed_coupon_vouchers(self, coupon_vouchers):
+        """
+        Returns assigned vouchers having at-least one redemption.
+        """
+        # TODO: FAQ - Single use has only one redemption? what to do in this case? return empty queryset?
+        return coupon_vouchers.filter(num_orders_gte=1, offers__max_global_applications__gt=F('num_orders'))
```
muhammad-ammar: I think empty queryset in case of `SINGLE_USE` make sense but I am not sure? `max_global_applications` 
 is set to `None` for `SINGLE_USE` so we need to add a test with `SINGLE_USE` type to specifically check the above filter condition.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +199,102 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        assigned_codes = self.get_assigned_codes()
+        redeemable_vouchers = self.get_redeemable_vouchers(coupon_vouchers)
+        unassigned_vouchers = redeemable_vouchers.exclude(code__in=assigned_codes)
+        return unassigned_vouchers
+
+    def get_unredeemed_coupon_vouchers(self, coupon_vouchers):
+        """
+        Returns assigned vouchers but never redeemed.
+        """
+        assigned_codes = self.get_assigned_codes()
```
muhammad-ammar: `assigned_codes` -> `assigned_and_unredeemed_codes`

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +199,102 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        redeemable_vouchers = self.get_redeemable_vouchers(coupon_vouchers)
+        unredeemed_vouchers = redeemable_vouchers.filter(code__in=assigned_codes)
+        return unredeemed_vouchers
```
muhammad-ammar, line 283: I think we don't need to call `self.get_redeemable_vouchers` in this method. We just need to filter the `coupon_vouchers` with `assigned_and_unredeemed_codes` like below

```python
def get_unredeemed_coupon_vouchers(self, coupon_vouchers):
    assigned_and_unredeemed_codes = self.get_assigned_codes()
    return coupon_vouchers.filter(code__in= assigned_and_unredeemed_codes)
```

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +199,102 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        Returns vouchers against each voucher application, so if a
+        voucher has two applications than this queryset will give 2 vouchers.
+        """
+        return Voucher.objects.filter(
+            applications__voucher_id__in=coupon_vouchers
+        )
+
+    def get_unassigned_coupon_vouchers(self, coupon_vouchers):
+        """
+        Returns only those vouchers which have free slots available.
```
muhammad-ammar: nit:`Return only ...`

ecommerce/extensions/api/serializers.py
```diff
@@ -695,8 +698,8 @@ def get_redemptions(self, voucher):
             max_coupon_usage = offer.max_global_applications
 
         return {
-            'used': redemption_count,
-            'available': max_coupon_usage,
+            'used': redemption_count + num_offerassignments,
+            'available': voucher.slots_available_for_assignment,
```
muhammad-ammar: `available` should be set to `max_coupon_usage`. It would be great if we can rename `available` to `total` on both the backend and frontend side

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +199,102 @@ def update_offer_data(self, request_data, vouchers, site):
...
+            SINGLE_USE, 3 quantiy, 1 assignment, 1 assigned + 2 other redemptions --> 0 slots available.
+            SINGLE_USE, 3 quantiy, 3 assignment, 3 assigned redemptions --> 0 slots available.
+
+
+            MULTI_USE, 3 quantiy, 10 max_uses, no assignment, no redemption --> 3 slots available.
+        """
+        assigned_codes = self.get_assigned_codes()
+        redeemable_vouchers = self.get_redeemable_vouchers(coupon_vouchers)
+        unassigned_vouchers = redeemable_vouchers.exclude(code__in=assigned_codes)
+        return unassigned_vouchers
```
muhammad-ammar:
```python
from django.db.models import OuterRef, Subquery
from django.db.models.functions import Coalesce

# subquery to count number of unredeemed offer assignments of a code
# the way this subquery is written is suggested in [django docs](https://docs.djangoproject.com/en/1.11/ref/models/expressions/#using-aggregates-within-a-subquery-expression)  
offerassignments = OfferAssignment.objects.filter(
    code=OuterRef('code')
).exclude(
    status__in=[OFFER_REDEEMED, OFFER_ASSIGNMENT_REVOKED]
).values('code').annotate(assignment_count=Count('pk', distinct=True)).values('assignment_count')

# Coalesce is used so that if there are no assigments than return 0
offerassignments_subquery = Coalesce(Subquery(offerassignments, output_field=IntegerField()), 0)

if SINGLE_USE:
    qs_may_have_free_slots = coupon_vouchers.filter(num_orders=0)
    vouchers_with_free_slots = qs_may_have_free_slots.annotate(
        assignment_count=offerassignments_subquery
    ).filter(
        assignment_count=0,
        num_orders=0
    )
else:
    qs_may_have_free_slots = coupon_vouchers.filter(offers__max_global_applications__gt=F('num_orders'))
    vouchers_with_free_slots = qs_may_have_free_slots.annotate(
        assignment_count=offerassignments_subquery
    ).filter(
        offers__max_global_applications__gt=F(assignment_count) + F('num_orders')
    )
```

  max = 10
  redeemed = 2
  assigned = 3
  redeemed and assigned are different 
  free slots = 5
  
  max = 10
  redeemed = 2
  assigned = 3
  redeemed and assigned are same 
  free slots = 7

ecommerce/extensions/api/v2/views/enterprise.py
muhammad-ammar: [This](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#none) is how we can return an empty queryset.

muhammad-ammar (changes requested):
@mushtaqak I am done with review. Please see my suggestions.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +200,143 @@ def update_offer_data(self, request_data, vouchers, site):
...
             super(EnterpriseCouponViewSet, self).update_range_data(request_data, vouchers)
 
+    def get_voucher_type(self):
+        """
+        Returns voucher type of the coupon object.
+        """
+        coupon = self.get_object()
+        return coupon.attr.coupon_vouchers.vouchers.first().usage
+
+    def get_assigned_uredeemed_codes(self):
```
adamstankiewicz: typo: `get_assigned_unredeemed_codes`

ecommerce/extensions/api/serializers.py
```diff
@@ -694,9 +697,14 @@ def get_redemptions(self, voucher):
...
+        slots_used = redemption_count + num_unredeemed_offerassignments
+        if voucher.usage in (Voucher.SINGLE_USE, Voucher.MULTI_USE_PER_CUSTOMER):
+            if redemption_count or num_unredeemed_offerassignments:
+                slots_used = max_coupon_usage
+
         return {
-            'used': redemption_count,
-            'available': max_coupon_usage,
+            'used': slots_used,
+            'total': max_coupon_usage,
```
adamstankiewicz, line 706: I agree that we should change this field name to `total` as it's a little more descriptive than `available`. However, we'll have to change the frontend code to reflect this API change, just FYI as a note.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -189,6 +200,143 @@ def update_offer_data(self, request_data, vouchers, site):
...
+    def get_partial_redeemed_coupon_vouchers(self, coupon_vouchers):
+        """
+        Return assigned vouchers having at-least one redemption.
+        """
+        # Single use has only one redemption, so there would be no partial redemption in this case.
+        voucher_type = self.get_voucher_type()
+        if voucher_type == Voucher.SINGLE_USE:
+            return coupon_vouchers.none()
+
+        assigned_uredeemed_codes = self.get_assigned_uredeemed_codes()
```
adamstankiewicz: typo: `assigned_unredeemed_codes`

mushtaqak:
@muhammad-ammar This is finally ready for your review :P

ecommerce/extensions/api/serializers.py
mushtaqak: Yep, addressed in https://github.com/edx/edx-portal/pull/164 :)

ecommerce/extensions/api/serializers.py
```diff
@@ -700,8 +702,8 @@ def get_redemptions(self, voucher):
             max_coupon_usage = offer.max_global_applications
 
         return {
-            'used': redemption_count,
-            'available': max_coupon_usage,
+            'used': redemption_count + num_unredeemed_offerassignments,
```
brittneyexline, line 705: i thought that assignments that have not redeemed would not go towards used, as used is for the number of redemptions. however, i do feel that we need a way to represent the number of assignments that have not been redeemed.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -190,6 +202,117 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        Returns the list of all those assigned codes which are not redeemed yet.
+        """
+        return OfferAssignment.objects.exclude(
+            status__in=[OFFER_REDEEMED, OFFER_ASSIGNMENT_REVOKED]
+        ).values_list('code', flat=True)
+
+    def get_vouchers_with_free_slots(self, coupon_vouchers):
+        """
+        Return vouchers with free slots available.
+        """
```
brittneyexline, line 223: i have to admit, i find this function hard to understand, so it's hard for me to know if it does the same thing as i did. this was my implementation: https://github.com/edx/ecommerce/pull/2109/files#diff-141fa9c110c00aba0d7f70839c4a77bbR260

brittneyexline (commented):
the biggest difference between what you've done here and what i've done in my PR is the conceptual model of the type of data we are returning. this implementation focuses on the status of an individual code (which is the way we originally thought it would work). my implementation focuses on the status of unique user/code pairs. this happens to be the same for SINGLE_USE and MULTI_USE_PER_CUSTOMER codes where only one user is assigned to one code, but it starts to get weird with MULTI_USE and ONCE_PER_CUSTOMER.

based on my conversation with Adam, we shifted to thinking about the status of the code and user together. consider the following scenario:

code A is a MULTI_USE code that has 4 usages. one usage has been redeemed by user 1, one usage is assigned to user 1 but not redeemed, one usage is assigned to user 2, the last usage is not assigned.

your implementation might show something like this (although you are not dealing with assigned to in the PR):
not-assigned - Redemptions: 3 of 4 | Code: A
not-redeemed - Assigned to: user 2 | Redemptions: 3 of 4 | Code: A
partial-redeemed - Assigned to: user 1 | Redemptions: 3 of 4 | Code: A
redeemed: nothing

my implementation would show this:
not-assigned - Redemptions: 1 of 4 | Code: A
not-redeemed - Assigned to: user 2 | Redemptions: 0 of 1 | Code: A
partial-redeemed - Assigned to: user 1 | Redemptions: 1 of 2 | Code: A
redeemed: nothing

when user 1 redeems their second assignment, my implementation would show a row in redeemed with redemptions: 2 of 2, while yours would remain the same in partial-redeemed.

i think we have to consider the user in order for the rows returned to make sense, which is why redemptions ends up scoped to the user (except for the not-assigned filter). as such, i don't think we can reconcile our PRs. i would encourage you to review mine closely to make sure i haven't missed anything that you are covering here.

mushtaqak:
Closed in favour of https://github.com/edx/ecommerce/pull/2109

PR #2103, opened 2019-01-28
Adding in create entitlement API

Dillon-Dumesnil:


Dillon-Dumesnil:
Related PR: https://github.com/edx/course-discovery/pull/1755

ecommerce/extensions/api/v2/views/products.py
```diff
@@ -30,3 +33,43 @@ def get_queryset(self):
...
             Q(course__partner=partner)
         )
+
+    def create(self, request, *args, **kwargs):
+        """
+        Create an Entitlement
+        """
+        entitlement_data = request.data
+        product_class = entitlement_data.get('product_class')
+        if product_class == COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME:
```
mikix: nit: I would use `data` instead of `entitlement_data` -- you don't know what class it is yet, and makes more sense in case we ever add support for other products to this method.

Likewise for the docstring, I'd make it a little more generic.

ecommerce/extensions/api/v2/views/products.py
```diff
@@ -30,3 +33,43 @@ def get_queryset(self):
...
+                    'These fields are required for Entitlement creation.'
+                return Response(bad_request_message, status=status.HTTP_400_BAD_REQUEST)
+
+            entitlement = create_or_update_course_entitlement(
+                product_creation_fields['certificate_type'],
+                product_creation_fields['price'],
+                product_creation_fields['partner'],
+                product_creation_fields['uuid'],
+                product_creation_fields['name']
+            )
```
mikix: You can just do:
```
entitlement = create_or_update_course_entitlement(**product_creation_fields)
```

ecommerce/extensions/api/v2/views/products.py
```diff
@@ -30,3 +33,43 @@ def get_queryset(self):
...
+                'uuid': entitlement_data.get('uuid'),
+                'name': entitlement_data.get('title'),
+            }
+
+            missing_values = []
+            for attribute_name, attribute_value in product_creation_fields.items():
+                if attribute_value is None:
+                    missing_values.append(attribute_name)
+            if missing_values:
+                bad_request_message = 'Missing or bad value(s) for: {}. '.format(', '.join(missing_values)) + \
```
mikix: I don't love the `(s)` - red flag for me, who is used to translating and being careful with plurals. It's also inconsistent with the phrase "These fields" just a moment later. is there a way to rephrase? Or always use the plural? I dunno. Your call.

Additionally, I think it's an interesting question of whether this error (and the one later in this file) should be translated? I understand the argument that these should basically never be seen. But that's more an argument for not writing them at all. If we expect a human to ever look at these, might be nice to translate. But just a nit. These are rare enough that I don't really care. If they are translated, the correct thing would probably be using ngettext to handle the plural correctly.

ecommerce/extensions/api/v2/views/products.py
```diff
@@ -30,3 +33,43 @@ def get_queryset(self):
...
+                'price': entitlement_data.get('price'),
+                'partner': request.site.partner,
+                'uuid': entitlement_data.get('uuid'),
+                'name': entitlement_data.get('title'),
+            }
+
+            missing_values = []
+            for attribute_name, attribute_value in product_creation_fields.items():
+                if attribute_value is None:
+                    missing_values.append(attribute_name)
```
mikix: Could do:
```
missing_values = [k for k, v in product_creation_fields.items() if v is not None]
```

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -135,6 +135,38 @@ def test_no_partner_product(self):
...
+            "price": 50,
+            "certificate_type": "verified",
+            "uuid": "f9044e15-133f-4a4f-b587-99530e8a8e88",
+        }
+        super(ProductViewSetCourseEntitlementTests, self).setUp()
+
+    def test_entitlement_post(self):
+        """ Verify the view allows individual Course Entitlement products to be made via post"""
+        response = self.client.post('/api/v2/products/', json.dumps(self.entitlement_data), JSON_CONTENT_TYPE)
+        self.assertEqual(response.status_code, 201)
```
mikix, line 157: Might also want to test that the database object got created correctly? Or at least mock create_or_update_course_entitlement and confirm it got called correctly.

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -135,6 +135,38 @@ def test_no_partner_product(self):
...
+
+    def test_entitlement_post_bad_request(self):
+        """ Verify the view allows individual Course Entitlement products to be made via post"""
```
mikix: docstring

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -135,6 +135,38 @@ def test_no_partner_product(self):
...
+        del self.entitlement_data['title']
+        del self.entitlement_data['price']
+        response = self.client.post('/api/v2/products/', json.dumps(self.entitlement_data), JSON_CONTENT_TYPE)
+        error_message = 'Missing or bad value(s) for: price, name. These fields are required for Entitlement creation.'
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, error_message)
+
+    def test_non_entitlement_post(self):
+        """ Verify the view allows individual Course Entitlement products to be made via post"""
```
mikix: docstring

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -135,6 +135,38 @@ def test_no_partner_product(self):
...
+        self.entitlement_data['product_class'] = 'Seat'
+        response = self.client.post('/api/v2/products/', json.dumps(self.entitlement_data), JSON_CONTENT_TYPE)
+        self.assertEqual(response.status_code, 400)
```
mikix, line 178: Might be worth testing message here. For all you know, it's complaining about missing price & name like the test above, rather than class type.

ecommerce/extensions/api/v2/tests/views/test_products.py
Dillon-Dumesnil: Disagree here. This docstring is under the class of tests called `ProductViewSetCourseEntitlementTests`. I feel comfortable specifying that it is a Course Entitlement in the docstring.

ecommerce/extensions/api/v2/tests/views/test_products.py
Dillon-Dumesnil: Disagree here. This docstring is under the class of tests called `ProductViewSetCourseEntitlementTests`. I feel comfortable specifying that it is a Course Entitlement in the docstring.

ecommerce/courses/tests/test_utils.py
```diff
@@ -57,7 +57,7 @@ def test_get_course_run_info_from_catalog(self, course_run):
                 course, discovery_api_url=self.site_configuration.discovery_api_url
             )
         else:
-            product = create_or_update_course_entitlement(
+            product, __ = create_or_update_course_entitlement(
```
mikix: Two underscores? That's fine too, but I usually see one. Do not change, super not worth it. Another style would be "_sku" to label what you're tossing away so the reader isn't wondering. But again, not worth changing.

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -135,6 +137,115 @@ def test_no_partner_product(self):
...
+
+        entitlement = get_entitlement(self.entitlement_data['uuid'], self.entitlement_data['certificate_type'])
+        self.assertEqual('Course ' + self.entitlement_data['title'], entitlement.title)
+
+        patch_data = {
+            'product_class': COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME,
+            'uuid': 'f9044e15-133f-4a4f-b587-99530e8a8e88',
+            'certificate_type': 'verified',
+            'title': title or self.entitlement_data['title'],
+            'price': price or self.entitlement_data['price'],
```
mikix: Why the `or` here - we probably want to test passing in partial data, no?

ecommerce/programs/tests/mixins.py
```diff
@@ -27,15 +27,15 @@ def mock_program_detail_endpoint(self, program_uuid, discovery_api_url, empty=Fa
             courses = []
             for i in range(1, 5):
                 uuid = '268afbfc-cc1e-415b-a5d8-c58d955bcfc' + str(i)
-                entitlement = create_or_update_course_entitlement('verified', 10, partner, uuid, uuid)
+                __, sku = create_or_update_course_entitlement('verified', 10, partner, uuid, uuid)
```
mikix: Heh nice

mikix (approved):
Thank you :pray: for doing the partial update stuff for me.

mikix (approved):
Nice title/name cleanup

ecommerce/extensions/api/v2/tests/views/test_products.py
```diff
@@ -135,6 +137,115 @@ def test_no_partner_product(self):
...
 
+@ddt.ddt
+class ProductViewSetCourseEntitlementTests(ProductViewSetBase):
+    def setUp(self):
+        self.entitlement_data = {
+            'product_class': COURSE_ENTITLEMENT_PRODUCT_CLASS_NAME,
+            'title': 'Test Course',
+            'price': 50,
+            'certificate_type': 'verified',
+            'uuid': 'f9044e15-133f-4a4f-b587-99530e8a8e88',
```
jmyatt: why not generate a fresh uuid instead of hard coding it?  More importantly, you could reference this value where it's used in the test code (`put_data` and `patch_data`) using `self.entitlement_data['uuid']` instead of also hard coding it there

jmyatt (changes requested):

jmyatt (approved):

PR #2104, opened 2019-01-29
Handle AuthStateMissing exception on oauth redirect

adeelkhan:
This patch would enable ecommerce to handle
AuthStateMissing exception raised on oauth
redirection from lms.

[LEARNER-6993](https://openedx.atlassian.net/browse/LEARNER-6993)

### DIFF
--- a/ecommerce/social_auth/middleware.py
+++ b/ecommerce/social_auth/middleware.py
@@ -1,0 +1,36 @@
+"""Middleware classes for social_auth."""
+
+import logging
+
+from django.shortcuts import redirect
+from django.urls import reverse
+from social_core.exceptions import AuthStateMissing
+from social_django.middleware import SocialAuthExceptionMiddleware
+
+log = logging.getLogger(__name__)
+
+
+class ExceptionMiddleware(SocialAuthExceptionMiddleware):
+    """Custom middleware that handles conditional redirection."""
+
+    def process_exception(self, request, exception):
+        """Handles specific exception raised by Python Social Auth eg AuthStateMissing."""
+
+        request_path = request.path
+        if request_path and isinstance(exception, AuthStateMissing):
+            if request_path == reverse('social:complete', args=['edx-oidc']):
+
+                from django.conf import settings
+                redirect_url = settings.SOCIAL_AUTH_EDX_OIDC_LOGOUT_URL
+                exception_logs = 'AuthStateMissing exception'
+
+                redirect_url = redirect_url if redirect_url else '/'
+                if redirect_url != '/':
+                    exception_logs += ', redirecting learner to lms logout url.'
+                else:
+                    exception_logs += ', redirecting learner to ecommerce index page.'
+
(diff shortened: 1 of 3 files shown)


adeelkhan:
I have tested this on local setup. I don't know any social auth configuration to test the generate similar error on sandbox, as what is applicable on local is not on sandbox.

fysheets (approved):
Looks good. For my own curiosity, was this pattern taken from somewhere else in our system? Also, due to your testing comments, let's just keep an eye on this and any errors that may occur once this is released.

adeelkhan:
@fysheets Some context here based on the question you asked.
1) In other IDA's i didn't find any such implementation of handling `AuthStateMissing`, as this issue is quite unique to ecommerce. 
2) Credentials and ecommerce for now are using this [https://github.com/edx/ecommerce/blob/0f1ced4ceac479ffe9c6a43b4c8c12795771c23c/ecommerce/settings/base.py#L229](url) to log social auth exception. 
3) Here just handling AuthStateMissing and letting social_django builtin middleware pick rest of errors eg `AuthCanceled`, `AuthParamMissing` etc.

4) The only issues i would be worried about after the merge are; 
- As we know the frequency of this exception are on avg 100/week, we would be calling LMS `/logout` a lot.
- As we are not sure of the exact flow/call which is causing learner to hit ecommerce on oauth and we are deliberately logging out that learner from LMS too,  we might see user logging out of LMS unexpectedly as somehow ecommerce failed to create/find the session.

fysheets:
Hm okay before you merge, let's quickly catch up on the two bullet point concerns that you have mentioned here.

ecommerce/social_auth/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+
+        request_path = request.path
+        if request_path and isinstance(exception, AuthStateMissing):
+            if request_path == reverse('social:complete', args=['edx-oidc']):
+
+                from django.conf import settings
+                redirect_url = settings.SOCIAL_AUTH_EDX_OIDC_LOGOUT_URL
+                exception_logs = 'AuthStateMissing exception'
+
+                redirect_url = redirect_url if redirect_url else '/'
```
waheedahmed, line 27: More pythonic way of doing this is `redirect_url = redirect_url or '/'`

waheedahmed (approved):
One minor comment, otherwise 👍

ecommerce/social_auth/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+                if redirect_url != '/':
+                    exception_logs += ', redirecting learner to lms logout url.'
+                else:
+                    exception_logs += ', redirecting learner to ecommerce index page.'
+
+                log.info(exception_logs)
```
cpennington, line 33: It might be worth logging the original stack trace from the exception, if possible.

cpennington (approved):

PR #2105, opened 2019-01-31
ENT-1505 Protect against errors from duplicate Enterprise Conditions.

brittneyexline:
**Description**
This PR addresses a current production bug where duplicate enterprise conditions exist and as a result, we are unable to update a particular Coupon.

It's unclear how duplicate enterprise conditions were created - my best guess is a race condition when two similar coupons were updated nearly simultaneously which triggered two conditions with the same data to get created.

In any case, protecting against errors from this is way easier than trying to understand the reasons why this happened.

**Testing Instructions**

_Recreating the bug_
Before deploying to the business sandbox, I turned off ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH, and created a new coupon with just an enterprise attached. I then went into the database and created a duplicate Condition matching the one used by the Coupon I just created. Then I tried to save the Coupon I created and received an error.

screen shot 2019-01-31 at 11 25 35 am

_Testing the bug fix_
I deployed the changes to the sandbox. I tried again to save the Coupon in question(https://ecommerce-business.sandbox.edx.org/coupons/425/edit/) and it was successful.

### DIFF
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -529,6 +529,44 @@ class CouponViewSetFunctionalTest(CouponMixin, DiscoveryTestMixin, DiscoveryMock
         new_coupon = Product.objects.get(id=self.coupon.id)
         self._check_enterprise_fields(new_coupon, enterprise_customer_id, enterprise_catalog_id, enterprise_name)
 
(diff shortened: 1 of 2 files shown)


brittneyexline:
Test Coverage: https://codecov.io/github/edx/ecommerce/commit/a704357352ef72a90edd308b966f5d806671e5b1

georgebabey (approved):

PR #2106, opened 2019-02-01
ENT-1394 update coupon codes detail endpoint

irfanuddinahmad:
Description: This PR updates the CouponVoucherSerializer so that the assigned_to field sends data as a dictionary of learner email and respective count.
For example:
```
{u'example_33@example.com': 2, u'example_32@example.com': 4, u'example_34@example.com': 1}
```

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 import logging
+from collections import Counter
 from datetime import timedelta
 from decimal import Decimal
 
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/api/serializers.py
```diff
@@ -685,8 +686,12 @@ def get_redeem_url(self, obj):
         url = get_ecommerce_url('/coupons/offer/')
         return '{url}?code={code}'.format(url=url, code=obj.code)
 
-    def get_assigned_to(self, obj):  # pylint: disable=unused-argument
-        return ''
+    def get_assigned_to(self, voucher):
+        offer = voucher.best_offer
```
brittneyexline, line 690: i would use voucher.enterprise_offer here instead

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -1187,3 +1187,43 @@ def test_coupon_codes_remind_bulk(self):
...
+            'expected_assigned_user_count': 4
+        },
+        {
+            'voucher_type': Voucher.MULTI_USE,
+            'quantity': 2,
+            'max_uses': 3,
+            'expected_assigned_user_count': 6
+        },
+    )
+    def test_coupon_codes_assigned_to_detail(self, data):
```
brittneyexline, line 1211: nit: any reason not to have the items in the dictionary listed explicitly as args here? that tends to be the convention in tests that i've seen.

ecommerce/extensions/api/serializers.py
```diff
@@ -685,8 +686,12 @@ def get_redeem_url(self, obj):
...
+        user_email_from_applications = list(voucher.applications.values_list('user__email', flat=True))
+        user_email_from_offer_assignments = list(offer.offerassignment_set.filter(code=voucher.code).exclude(
+            status__in=[OFFER_ASSIGNMENT_REVOKED, OFFER_REDEEMED]).values_list('user_email', flat=True))
+        return Counter(user_email_from_applications + user_email_from_offer_assignments)
```
brittneyexline, line 694: just for clarification, the expectation is that assigned_to will return both redemptions and assignments? if a user has one redemption and one assignment, does that mean it will be collapsed into {'user_email': 2}, without distinguishing the redemption from the assignment?

mushtaqak:
Closing in favour of https://github.com/edx/ecommerce/pull/2109

PR #2109, opened 2019-02-04
Updating code details endpoint logic

brittneyexline:
I noticed we've had some confusion around the code details endpoint and what should be returned. This is my attempt to lay out the logic I discussed with Adam today. Below are my notes outlining what the toggles will display for different usage types.


ecommerce/extensions/api/serializers.py
```diff
@@ -705,6 +710,64 @@ def get_redemptions(self, voucher):
...
+                voucher=obj,
+                user__email=obj.user_email
+            ).count()
+            return {'used': num_applications, 'available': num_assignments + num_applications}
+
+
+class RedeemedCodeUsageSerializer(CodeUsageSerializer):  # pylint: disable=abstract-method
+
+    def get_code(self, obj):
+        return obj.voucher__code
```
christopappas: should this be `obj.code`?

ecommerce/extensions/api/serializers.py
```diff
@@ -705,6 +710,64 @@ def get_redemptions(self, voucher):
...
+
+    def get_assigned_to(self, obj):
+        return obj.user__email
```
christopappas: and `obj.user_email` here?

ecommerce/extensions/api/serializers.py
```diff
@@ -705,6 +710,64 @@ def get_redemptions(self, voucher):
...
+
+    def get_redemptions(self, obj):
+        num_applications = VoucherApplication.object.filter(
+            voucher__code=obj.voucher__code,
```
christopappas: same question here (`voucher__code=obj.code`?)
and `user__email=obj.user_email`?

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -211,28 +224,137 @@ def codes(self, request, pk, format=None):  # pylint: disable=unused-argument, r
...
+                queryset = queryset.union(id=voucher.id, all=True)
+            elif voucher.usage in (Voucher.MULTI_USE, Voucher.ONCE_PER_CUSTOMER):
+                for _ in range(0, slots_available):
+                    queryset = queryset.union(id=voucher.id, all=True)
+
+        return queryset
+
+    def _get_not_redeemed_usages(self, vouchers):
+        """
+        Returns a queryset containing unique code and user_email pairs from OfferAssignments.
```
christopappas, line 278: more of nitpick than anything cause the code makes it clear what's going on here, but I had to read this and the following sentence a few times to get what type of objects the queryset contained.

Maybe something like:

```
Returns a queryset of OfferAssignments containing unique code and user_email pairs.
Only OfferAssignments with code and user_email pairs that have no corresponding VoucherApplication are returned
```

Feel free to disregard though

christopappas: maybe disregard this, i dunno ><'

christopappas (commented):
Overall looks good! I like the clear if logic determining what serializer gets called.

I think the one thing I noticed that isn't an issue, but made me stop and think, was in the detail-view for codes, those helper methods return querysets of different types of objects that are then serialized, but those different serializers all inherit from the same CodeUsageSerializer. Not that there's anything necessarily wrong with that, I guess I'm just used to a stronger connection between serializers and the models of the objects they serialize

Only other thing I would mention is adding tests! 👍

muhammad-ammar:
@brittneyexline @mushtaqak is already working on [code filters](https://github.com/edx/ecommerce/pull/2099) and his work is almost done. Any particular reason to open this PR?

muhammad-ammar:
@brittneyexline Just had a discussion with @irfanuddinahmad and found that this work is for demo purposes.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -437,123 +443,177 @@ def create_coupon_with_applications(self, coupon_data, voucher_type, quantity, m
...
+            '/api/v2/enterprise/coupons/{}/codes/?code_filter={}'.format(coupon_id, VOUCHER_NOT_ASSIGNED)
+        ).json()
+        self.assertEqual(len(response['results']), 1)
+        self.assert_code_detail_response(response['results'][0], codes[0], '', 0, 2)
+
+        response = self.get_response(
+            'GET',
+            '/api/v2/enterprise/coupons/{}/codes/?code_filter={}'.format(coupon_id, VOUCHER_NOT_REDEEMED)
+        ).json()
+        print response['results']
```
christopappas: 🖨 :)

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -437,123 +444,153 @@ def create_coupon_with_applications(self, coupon_data, voucher_type, quantity, m
...
-            coupon_data['max_uses']
-        )
+    def test_coupon_codes_detail(
+            self,
+            voucher_type,
+            quantity,
+            max_uses,
+            code_assignments,
+            code_redemptions,
+            expected_responses):
```
christopappas, line 594: i think i'd ask that you add in a doctring here. the ddt data is really big so it's hard (for me) to tell what's intended on being tested without really needing to dig in

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -437,123 +444,163 @@ def create_coupon_with_applications(self, coupon_data, voucher_type, quantity, m
...
         """
-        Verify that `/api/v2/enterprise/coupons/{coupon_id}/codes/` endpoint pagination works
+        Tests the expected response for the code details endpoint based on various types and usage states for a coupon.
```
christopappas, line 596: Would it be better in this case to then have 4 separate tests, each one being its own test case? I will leave it to your judgement, but usually I'm for more, smaller tests that focus on a smaller scope

brittneyexline: perhaps - in the interest of time, i am not going to do it for this PR.

christopappas: 👌

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -211,28 +225,119 @@ def codes(self, request, pk, format=None):  # pylint: disable=unused-argument, r
...
-        # we need a combined querset so that pagination works as expected
-        all_coupon_vouchers = coupon_vouchers_with_applications | coupon_vouchers_wo_applications
+        coupon_vouchers = coupon.attr.coupon_vouchers.vouchers.all()
+        usage_type = coupon_vouchers.first().usage
+        code_filter = request.query_params.get('code_filter')
+        queryset = None
+        serializer_class = None
+
+        if not code_filter:
+            raise serializers.ValidationError('code_filter must be specified')
```
mushtaqak, line 235: **Just as an FYI** that this will prevent csv downloading at-least for now (until csv is fetched against code filters)

brittneyexline: yep - i've talked to adam and we are going to update the frontend to pass the code filter for the csv call.

christopappas (approved):
🚢 👍

adamstankiewicz (approved):

PR #2110, opened 2019-02-05
ENT-1511 | Allowing for CORS for this service

christopappas:


ecommerce/settings/production.py
```diff
@@ -84,3 +84,12 @@ def get_env_setting(setting):
...
 
 ENTERPRISE_API_URL = urljoin(ENTERPRISE_SERVICE_URL, 'api/v1/')
+
+CORS_ORIGIN_WHITELIST = environ.get('CORS_ORIGIN_WHITELIST', ())
+CORS_URLS_REGEX = r'{regex}'.format(regex=environ.get('CORS_URLS_REGEX', ''))
+CORS_ALLOW_HEADERS = corsheaders_default_headers + (
+    'use-jwt-cookie',
+)
+CORS_ALLOW_CREDENTIALS = True
+CORS_ORIGIN_ALLOW_ALL = environ.get('CORS_ORIGIN_ALLOW_ALL', False)
```
georgebabey: Is this and `CORS_ALLOW_INSECURE ` needed at all? I can't imagine a scenario where we'd be using production.py and want to allow any origin.

georgebabey (approved):
Just one question but this looks good!

ecommerce/settings/production.py
christopappas: can remove, np

ecommerce/settings/production.py
```diff
@@ -84,3 +84,10 @@ def get_env_setting(setting):
 # END PAYMENT PROCESSOR OVERRIDES
 
 ENTERPRISE_API_URL = urljoin(ENTERPRISE_SERVICE_URL, 'api/v1/')
+
+CORS_ORIGIN_WHITELIST = environ.get('CORS_ORIGIN_WHITELIST', ())
+CORS_URLS_REGEX = r'{regex}'.format(regex=environ.get('CORS_URLS_REGEX', ''))
+CORS_ALLOW_HEADERS = corsheaders_default_headers + (
```
mushtaqak, line 91: We need to import this as well.
```python
from corsheaders.defaults import default_headers as corsheaders_default_headers
```

mushtaqak (changes requested):
@christopappas Looks good, one thing, we need to import `corsheaders_default_headers`

ecommerce/settings/production.py
christopappas: goood catch. thank you mushtaq

mushtaqak (approved):
Looks good :shipit:

PR #2115, opened 2019-02-07
Fix enrollments redeemed calculation from coupon overview endpoint

brittneyexline:
Updated max_uses and num_uses to reflect the value across all vouchers in the coupon.


ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -852,7 +878,9 @@ def test_coupon_num_unassigned(
         )
 
         # Verify that we get correct results.
-        self.assertEqual(coupon_overview_response['results'][0]['num_unassigned'], expected_num_unassigned)
+        print coupon_overview_response['results']
```
christopappas: 🖨

christopappas (approved):
makes sense to me. 👍

PR #2116, opened 2019-02-07
Rewrite unredeemed and partially-redeemed to not use unions for queryset

brittneyexline:
The unions intermittently cause sql errors, this ensures we don't rely on unions to build the queryset.


christopappas (approved):
👍

PR #2121, opened 2019-02-08
create assignments for unassigned multi-use-per-customer coupon

muhammad-ammar:
[ENT-1536](https://openedx.atlassian.net/browse/ENT-1536)


ecommerce/extensions/checkout/mixins.py
```diff
@@ -311,3 +314,22 @@ def update_assigned_voucher_offer_assignment(self, order):
...
+        if not offer:
+            return None
+
+        if voucher.usage == voucher.MULTI_USE_PER_CUSTOMER:
+            if not OfferAssignment.objects.filter(code=voucher.code, user_email=basket.owner.email).exists():
+                assignments = [
+                    OfferAssignment(offer=offer, code=voucher.code, user_email=basket.owner.email)
+                    for __ in range(offer.max_global_applications)
+                ]
+                OfferAssignment.objects.bulk_create(assignments)
```
mushtaqak, line 336: Nice!

mushtaqak (approved):
Looks good to me 👍

ecommerce/extensions/checkout/mixins.py
```diff
@@ -311,3 +314,22 @@ def update_assigned_voucher_offer_assignment(self, order):
...
+        voucher = basket.vouchers.first()
+        offer = voucher and voucher.enterprise_offer
+        # can't entertain non enterprise offers
+        if not offer:
+            return None
+
+        if voucher.usage == voucher.MULTI_USE_PER_CUSTOMER:
+            if not OfferAssignment.objects.filter(code=voucher.code, user_email=basket.owner.email).exists():
+                assignments = [
+                    OfferAssignment(offer=offer, code=voucher.code, user_email=basket.owner.email)
```
brittneyexline: as this is, i think the status will be email pending... and we're not kicking off an email task. i think it would be better to set status to assigned from the outset, because we are not emailing.

brittneyexline (approved):
just one small update i'd like to see before this merges!

ecommerce/extensions/checkout/mixins.py
muhammad-ammar: @brittneyexline I have addressed your feedback. Can you please merge if it is good to go?

brittneyexline (approved):
I will merge/release this Thursday morning @muhammad-ammar

PR #2122, opened 2019-02-08
Fix Oscar dashboard's login

nasthagiri:


ecommerce/extensions/dashboard/app.py
```diff
@@ -1,14 +1,11 @@
-from auth_backends.urls import auth_urlpatterns
+from auth_backends.urls import oauth2_urlpatterns
+from auth_backends.views import EdxOAuth2LoginView
```
doctoryes: Seems like this import isn't needed?

nasthagiri: good catch.

doctoryes (approved):

PR #2127, opened 2019-02-12
Hasnain naveed/wl 1887

hasnain-naveed:


ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -0,0 +1,333 @@
...
+        Returns list of dicts, each dict represents hubspot DEAL.
+        """
+        hubspot_orders = []
+        for order in orders:
+            hubspot_orders.append({
+                'integratorObjectId': str(order.id),
+                'action': 'UPSERT',
+                'changeOccurredTimestamp': int(time.time()),
+                'propertyNameToValues': {
+                    'total_incl_tax': int(order.total_incl_tax),
```
ziafazal: this should be float.

ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -0,0 +1,333 @@
...
+        hubspot_line_items = []
+        for line in lines:
+            hubspot_line_items.append({
+                'integratorObjectId': str(line.id),
+                'action': 'UPSERT',
+                'changeOccurredTimestamp': int(time.time()),
+                'propertyNameToValues': {
+                    'order_id': str(line.order.id),
+                    'price_currency': str(line.order.currency),
+                    'tax': int(line.line_price_incl_tax) - int(line.line_price_excl_tax),
```
ziafazal: this should be float.

ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -0,0 +1,333 @@
...
+                    'product_id': str(line.product.id),
+                    'price_incl_tax': int(line.line_price_incl_tax),
```
ziafazal: this should be float.

ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -0,0 +1,333 @@
...
+                    'price_excl_tax': int(line.line_price_excl_tax),
```
ziafazal: this should be float.

ziafazal (changes requested):
@hasnain-naveed 
@HammadAhmadWaqas overall looks good few minor things to fix. However, I think we should improve verbosity of command by add info at the end of every action it performs.

ecommerce/core/management/commands/sync_hubspot.py
hasnain-naveed: so you want me to convert float value into a string then store it? Because hubspot only support the following data types.
`["STRING", "NUMBER", "DATETIME", "AVATAR_IMAGE"]`

ziafazal: I'm talking about `total_incl_tax`. I think `NUMBER` should accept floats.

ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -0,0 +1,380 @@
...
+            message = 'Starting to sync data of site: {site} to Hubspot.'.format(
+                site=site_configuration.site.domain
+            )
+
+            for order in unsynced_orders:
+                lines = OrderLine.objects.filter(order=order)
+                for line in lines:
+                    unsynced_order_lines.append(line)
+                    unsynced_products.append(line.product)
+
```
HammadAhmadWaqas: @hasnain-naveed 
I think we can save loop queries and nested for loop using these two lines
```
unsynced_order_lines = OrderLine.objects.filter(order__in=unsynced_orders)
unsynced_products = Product.objects.filter(line__in=unsynced_order_lines)
```

HammadAhmadWaqas (commented):
added a suggestion

PR #2129, opened 2019-02-12
move ecommerce task to thier own queue

syed-awais-ali:


ecommerce/settings/base.py
```diff
@@ -539,6 +540,10 @@
 CELERY_ALWAYS_EAGER = False
 # END CELERY
 
+CELERY_DEFAULT_QUEUE = 'ecommerce'
+CELERY_QUEUES = (
+    Queue('ecommerce', Exchange('ecommerce'), routing_key='ecommerce'),
```
coryleeio: Instead of just setting this to ecommerce, I would default it to celery(I think this is what it is now), and load it from an environment variable.  That away people running this code already dont have their queue changed out from underneath them, because if the old queue is not empty that work wont be completed.

I would read in a config in production.py, default that value to celery deploy that to production, then make the config change in internal.

When we roll the config change out you will need to make sure the old default queue is clear before we allow the ecommerce service to terminate.

coryleeio (changes requested):
load from env, similar to notifier PR

ecommerce/settings/base.py
```diff
@@ -536,6 +537,10 @@
...
```
coryleeio: You have a routing key variable below, but ROUTING_KEY = environ.get('ECOMMERCE_ROUTING_KEY')


but it is still hardcoded here to 'ecommerce'

coryleeio (dismissed):

coryleeio:
I've removed my review since you appear to be redoing this

coryleeio:
Please file a new review ticket once this is ready to go

ecommerce/settings/production.py
```diff
@@ -89,3 +89,5 @@ def get_env_setting(setting):
 CORS_ALLOW_HEADERS = corsheaders_default_headers + (
     'use-jwt-cookie',
 )
+
+ROUTING_KEY = environ.get('ECOMMERCE_ROUTING_KEY')
```
coryleeio: same pattern here as https://github.com/edx/ecommerce-worker/pull/52#pullrequestreview-225196290

coryleeio (changes requested):

ecommerce/settings/production.py
syed-awais-ali: @coryleeio should this be moved to base.py ?
i am moving it to base.py

coryleeio (approved):

PR #2132, opened 2019-02-13
Remind all coupon assignments

mushtaqak:
This PR adds the capability to send reminders to all assignments associated with a coupon.

Related frontend PR: https://github.com/edx/edx-portal/pull/177

[ENT-1540](https://openedx.atlassian.net/browse/ENT-1540)

### DIFF
--- a/ecommerce/extensions/api/v2/views/enterprise.py
+++ b/ecommerce/extensions/api/v2/views/enterprise.py
@@ -407,8 +407,34 @@ class EnterpriseCouponViewSet(CouponViewSet):
         email_template = request.data.pop('template', None)
         if not email_template:
             log_message_and_raise_validation_error(str('Template is required.'))
+
+        if request.data.get('assignments'):
+            assignments = request.data.get('assignments')
+        else:
+            # If no assignment is passed, send reminder to all assignments associated with the coupon.
+            vouchers = coupon.attr.coupon_vouchers.vouchers.all()
+            code_filter = request.data.get('code_filter')
+
+            if not code_filter:
+                raise serializers.ValidationError('code_filter must be specified')
+
+            if code_filter == VOUCHER_NOT_REDEEMED:
+                assignment_usages = self._get_not_redeemed_usages(vouchers)
+            elif code_filter == VOUCHER_PARTIAL_REDEEMED:
+                assignment_usages = self._get_partial_redeemed_usages(vouchers)
+            else:
+                raise serializers.ValidationError('Invalid code_filter specified: {}'.format(code_filter))
+
+            assignments = [
+                {
+                    'code': assignment['code'],
+                    'email': assignment['user_email']
+                }
+                for assignment in assignment_usages
+            ]
+
         serializer = CouponCodeRemindSerializer(
-            data=request.data.get('assignments'),
+            data=assignments,
             many=True,
             context={'coupon': coupon, 'template': email_template}
         )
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -398,8 +398,22 @@ def remind(self, request, pk):  # pylint: disable=unused-argument
         email_template = request.data.pop('template', None)
         if not email_template:
             log_message_and_raise_validation_error(str('Template is required.'))
+
+        # For remind_all, get all un-redemeed usages associated with the coupon.
+        if request.data.get('remind_all'):
```
adamstankiewicz: Do we need the `remind_all` property to be passed to this endpoint to support reminding all? Couldn't we achieve the same result if we just look for whether `assignments` is passed in? If `assignments` is not provided, then we would know the request is for reminding all and can create the `assignments` list at that point.

```python
if request.data.get('assignments'):
    assignments = request.data.get('assignments')
else:
    vouchers = coupon.attr.coupon_vouchers.vouchers.all()
    assignments = [
        {
            'code': assignment['code'],
            'email': assignment['user_email']
        }
        for assignment in self._get_not_redeemed_usages(vouchers)
    ]

**Review by muhammad-ammar — APPROVED** [review]
LGTM 👍

File: ecommerce/extensions/api/v2/views/enterprise.py (hunk 1 above)
Comment (mushtaqak) [reply ecommerce/extensions/api/v2/views/enterprise.py]: Done, makes sense ( addressed on frontend side as well) 👍

**Comment by mushtaqak** [comment]
@georgebabey / @brittneyexline This is ready to merge. Can you guys have a look at this and merge ?

File: ecommerce/extensions/api/v2/views/enterprise.py [hunk 2]
```diff
@@ -398,8 +398,22 @@ def remind(self, request, pk):  # pylint: disable=unused-argument
...
+            assignments = request.data.get('assignments')
+        else:
+            # If no assignment is passed, send reminder to all assignments associated with the coupon.
+            vouchers = coupon.attr.coupon_vouchers.vouchers.all()
+            assignments = [
+                {
+                    'code': assignment['code'],
+                    'email': assignment['user_email']
+                }
+                for assignment in self._get_not_redeemed_usages(vouchers)
```
brittneyexline: should we also get partial redeemed usages? they have the option of getting reminded too. however, i suppose it depends on the view you are on in the UI, so maybe that needs to be indicated in the api? what do you think @adamstankiewicz?

mushtaqak: I guess, this makes sense 👍

adamstankiewicz: @brittneyexline @mushtaqak hmm, yeah. I feel like the bulk actions should only apply to the codes in the current view. If the admin is reminding all for the unredeemed view, we should probably only send reminders for unredeemed codes; similarly, only send reminders for partial redeemed codes if the admin is in the partially-redeemed view.

The current table view would need to be passed to the API to make this happen.

mushtaqak: I have added the capability to support partial-redeemed toggle as well.

mushtaqak:
@brittneyexline This is good to get your final review.

brittneyexline (approved):

adamstankiewicz (approved):

PR #2141, opened 2019-02-15
(WIP) Add bulk_change_username command.

tuchfarber:
Command accepts a csv of current usernames and new usernames and renames them across all relevant models.

Note: This should not be used alone. It will be part of a larger job that will run across all services.

All similar PRs:
edx/edx-platform#19815
edx/ecommerce#2141
edx/credentials#635
edx/course-discovery#1786
edx/edx-notes-api#90


tuchfarber:
Closing as we've decided to go with an API instead of a management command. See #2151

PR #2144, opened 2019-02-18
ENT-1566 fix issues with offer assignment email bounce api

irfanuddinahmad:
Description: This PR replaces the message hash verification with Sailthru client library based message verification


mushtaqak (approved):
Looks good but I am not sure this would solve our issue (email bounce handling) but this definitely improves the code since we are now implementing Sailthru's hashing mechanism and also improving logging/ removing hash which is good.

georgebabey (approved):

PR #2118, opened 2019-02-08
WL-1880 | Added hubspot secret key field in site configuration.

hasnain-naveed:

PR #2146, opened 2019-02-19
 WL-1881 | Added a management command to sync the ecommerce data to hubspot.

hasnain-naveed:


hasnain-naveed:
@HammadAhmadWaqas this is final PR.

ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -0,0 +1,439 @@
...
+class Command(BaseCommand):
+    help = 'Sync Product, Orders and Lines to partners Hubspot server.'
+
+    def _get_hubspot_enable_sites(self):
+        """
+        Returns all SiteConfigurations which have hubspot enabled
+        """
+        return SiteConfiguration.objects.filter(hubspot_secret_key__isnull=False)
+
+    def _hubspot_endpoint(self, api, api_url, api_type, body=None, **kwargs):
```
ziafazal: Can we make signature of this method a little more readable.
i.e `def _hubspot_endpoint(self, hubspot_object, api_url, method, body=None, **kwargs):`

ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -0,0 +1,439 @@
...
+            if orders:
+                self.stdout.write(
+                    'Pulled unsynced orders for site {site} from {last_sync}'.format(
+                        site=site_configuration.site.domain, last_sync=last_synced_order.date_placed
+                    )
+                )
+        else:
+            orders = orders.filter(date_placed__lte=datetime.now())
+            self.stdout.write(
+                'It is first syncing call, it will pull all the orders for syncing for site {site}'.format(
```
ziafazal: May be we can change the wording `No last synced order found...`

ziafazal (approved):
@hasnain-naveed overall looks good just few nits.

ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -0,0 +1,455 @@
...
+            self._upsert_hubspot_objects(
+                LINE_ITEM,
+                self._get_hubspot_line_item_structure(unsynced_order_lines),
+                site_configuration
+            )
+        else:
+            self.stdout.write('No data found to sync for site {site}'.format(site=site_configuration.site.domain))
+
+    def add_arguments(self, parser):
+
```
georgebabey: nit: space

ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -0,0 +1,455 @@
...
+    ORDER.COMPLETE: "processed"
+}
+PRODUCT = "PRODUCT"
+LINE_ITEM = "LINE_ITEM"
+DEAL = "DEAL"
+BATCH_SIZE = 200
+
+
+class Command(BaseCommand):
+    help = 'Sync Product, Orders and Lines to partners Hubspot server.'
```
georgebabey: I would take `partner` out of here as that is an edx specific term. Maybe `Sync Products, Orders and Line Items to a Hubspot server.`

ecommerce/core/migrations/0048_siteconfiguration_hubspot_secret_key.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('core', '0047_businessclient_enterprise_customer_uuid'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='siteconfiguration',
+            name='hubspot_secret_key',
+            field=models.CharField(blank=True, help_text='Secret key for Hubspot portal authentication', max_length=255, null=True, verbose_name='Hubspot Portal Secret Key'),
```
georgebabey: General django best practice is to not allow null for CharField (see https://docs.djangoproject.com/en/dev/ref/models/fields/#null) 

Right now you are checking for non-`null` to determine if hubspot is enabled, but new configurations will have this set to empty string.  https://github.com/edx/ecommerce/pull/2146/files#diff-344c0546bfd4c3a65a58efc931f82f3aR147

I'd recommend changing this to `null=False`

georgebabey (changes requested):
Looks great overall but I believe there is one issue with null vs empty_string on the configuration

ecommerce/core/migrations/0048_siteconfiguration_hubspot_secret_key.py
hasnain-naveed: @georgebabey WL-1880 is pre-requisite of WL-1881. WL-1880 covers this model change. I have updated my [PR](https://github.com/edx/ecommerce/pull/2118) according to your suggestion. Waiting for your +1. Thanks.

georgebabey: Merged WL-1880 in, so this will need to be rebased now

georgebabey (approved):

georgebabey (approved):

PR #2147, opened 2019-02-20
ENT-1566 Updated logging for offer assignment when send_id is not found in model

irfanuddinahmad:
Description: This PR updates the bounce endpoint (POST /ecommerce/api/v2/assignment-email/bounce).
It removes logging for the case when the Sailthru send_id is not found in the OfferAssignmentEmailAttempt model. This log was generating a lot of splunk entries since Sailthru sends a bounce callback for all bounces, irrespective of the origin email template. Consequently, most of the send_ids are expected to be not present in the OfferAssignmentEmailAttempt model and hence should not generate a log entry.


ecommerce/extensions/api/v2/views/assignmentemail.py
```diff
@@ -161,9 +161,8 @@ def post(self, request):
             if sailthru_client.receive_hardbounce_post(request.data):
                 try:
                     self.update_email_status(send_id)
-                except (OfferAssignment.DoesNotExist, OfferAssignmentEmailAttempt.DoesNotExist) as exc:
-                    logger.exception(
-                        '[Offer Assignment] AssignmentEmailBounce could not update status and raised: %r', exc)
+                except (OfferAssignment.DoesNotExist, OfferAssignmentEmailAttempt.DoesNotExist):
+                    pass
```
georgebabey, line 167: Can you add a comment here indicating why we're passing. Something like: Marketing email bounces also come through this code path and its expected they would not have a corresponding OfferAssignment

georgebabey (changes requested):

ecommerce/extensions/api/v2/views/assignmentemail.py
irfanuddinahmad: Done

georgebabey (approved):

muhammad-ammar (approved):
LGTM 👍

PR #2150, opened 2019-02-21
for multi-use, only revoke a single offer assignment

muhammad-ammar:
[ENT-1574 
](https://openedx.atlassian.net/browse/ENT-1574)[ENT-1586](https://openedx.atlassian.net/browse/ENT-1586)

**Description:**
In case of `MULTI_USE` coupon, for multiple assignments for a user for a same code should only revoke a single assignment instead of all the assignments.

Consider the below sample assignments, Without the changes in this PR, a revoke request for (`ABC101`, `batman@abc.com`) will revoke all the assignments which should no happen.
 ```python
code = ABC101 email = batman@abc.com
code = ABC101 email = batman@abc.com
code = ABC101 email = batman@abc.com
```

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -1399,6 +1399,15 @@ class CouponCodeRevokeSerializer(CouponCodeMixin, serializers.Serializer):  # py
         offer_assignments = self.get_unredeemed_offer_assignments(code, email)
         if not offer_assignments.exists():
             raise serializers.ValidationError('No assignments exist for user {} and code {}'.format(email, code))
+
+        # Only revoke a single assignment, Consider the below sample assignments,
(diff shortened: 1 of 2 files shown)


mushtaqak (approved):
Looks good 👍

brittneyexline (changes requested):
i've talked with jennifer about this case because this implementation was my best guess of what should happen at the time. she confirmed that when we revoke a code for a user with multiple assignments for a multi use code, all of the assignments should be revoked.

i'm not sure what the implication is of not doing this for the bugs you were trying to address, but i think we need to preserve this particular behavior and find a different way to address those bugs.

muhammad-ammar:
@brittneyexline Thanks for the details. Closing this in favour of https://github.com/edx/ecommerce/pull/2155

PR #2151, opened 2019-02-22
Adds username replacement API

tuchfarber:
New API replaces all copies of username in LMS a desired username. 
Requires user be in username_replacement_admin group. Should only be ran as a larger job  to update usernames across all services,  otherwise the system will be left in a broken state for those users.

### DIFF
--- a/ecommerce/extensions/api/permissions.py
+++ b/ecommerce/extensions/api/permissions.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.conf import settings
 from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
 from rest_framework import permissions
@@ -11,6 +12,8 @@ from ecommerce.extensions.api.serializers import retrieve_enterprise_condition
 Product = get_model('catalogue', 'Product')
 
 LOGGER = logging.getLogger(__name__)
+
+USERNAME_REPLACEMENT_GROUP = "username_replacement_admin"
 
 
 class CanActForUser(permissions.IsAdminUser):
@@ -100,5 +103,12 @@ class HasDataAPIDjangoGroupAccess(permissions.BasePermission):
                 return False
             enterprise_condition = retrieve_enterprise_condition(coupon)
             enterprise_id = enterprise_condition and enterprise_condition.enterprise_customer_uuid
-
         return self._request_is_permitted_for_enterprise(request, enterprise_id)
+
+
+class CanReplaceUsername(permissions.BasePermission):
+    """
+    Grants access to the Username Replacement API for a the service user.
+    """
+    def has_permission(self, request, view):
+        return request.user.username == getattr(settings, 'USERNAME_REPLACEMENT_WORKER')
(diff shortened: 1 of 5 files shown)


ecommerce/extensions/api/v2/views/user_management.py
```diff
@@ -0,0 +1,147 @@
...
+        {
+            "successful_replacements": [
+                {"old_username_1": "new_username_1"}
+            ],
+            "failed_replacements": [
+                {"old_username_2": "new_username_2"}
+            ]
+        }
+
+        TODO: Determine if we need an audit trail outside of logging and API response.
```
bessiesteinberg: remove/resolve TODO

ecommerce/extensions/api/v2/views/user_management.py
```diff
@@ -0,0 +1,147 @@
...
+        for obj in post_data:
+            if not (isinstance(obj, dict) and len(obj) == 1):
+                return False
+        return True
+
+    def _replace_username_for_all_models(self, current_username, new_username, replacement_locations):
+        """
+        Replaces current_username with new_username for all (model, column) pairs in replacement locations.
+        Returns if it was successful or not. Will return successful even if no matching
+
```
bessiesteinberg: not sure what it means by "will return successful even if no matching" does that mean: will return true even if no records were found for 'current_username'.  In what case does it return false?

ecommerce/extensions/api/v2/views/user_management.py
```diff
@@ -0,0 +1,147 @@
...
+        TODO: Determine if logs of username are a PII issue.
```
bessiesteinberg: remove TODO

bessiesteinberg (approved):
Looks good! There are two TODOs that should be removed or resolved and one comment I thought could use clarification, otherwise the code is really clear and well done. Nice!

ecommerce/extensions/api/v2/views/user_management.py
tuchfarber: Removed

ecommerce/extensions/api/v2/views/user_management.py
tuchfarber: Fixed with text from other PRs. I think I quit typing in the middle of a sentence :). It returns false only if there is an exception while replacing, which should only happen if there's an issue with the DB or the models changes.

ecommerce/extensions/api/v2/views/user_management.py
tuchfarber: Done

ecommerce/extensions/api/permissions.py
```diff
@@ -100,5 +103,13 @@ def has_permission(self, request, view):
                 return False
             enterprise_condition = retrieve_enterprise_condition(coupon)
             enterprise_id = enterprise_condition and enterprise_condition.enterprise_customer_uuid
-
         return self._request_is_permitted_for_enterprise(request, enterprise_id)
+
+
+class CanReplaceUsername(permissions.BasePermission):
+    """
+    Grants access to the Username Replacement API for anyone in the group,
```
georgebabey: Same comment as platform - comment mentions group but code just does username check

georgebabey (approved):
Looks good-  just same comments as the edx-platform PR

PR #2152, opened 2019-02-22
Fix redeemed filter

mushtaqak:
Fix redeem filter to not exclude those voucher applications with user having offerassignment with assigned status.

[ENT-1569](https://openedx.atlassian.net/browse/ENT-1569)


brittneyexline (approved):

muhammad-ammar (approved):
LGTM 👍

PR #2160, opened 2019-02-27
ENT-1606 New permission -  HasDataAPIDjangoGroupAccess

irfanuddinahmad:
Description: This PR adds a new permission HasDataAPIDjangoGroupAccess which is applied on the enterprise api endpoints.


ecommerce/enterprise/clients.py
```diff
@@ -0,0 +1,101 @@
...
+from edx_rest_api_client.exceptions import HttpClientError, HttpServerError
+from rest_framework.exceptions import NotFound, ParseError
+
+from ecommerce.core.utils import get_cache_key
+
+DEFAULT_REPORTING_CACHE_TIMEOUT = 60 * 60 * 6  # 6 hours (Value is in seconds)
+LOGGER = logging.getLogger(__name__)
+
+
+class EnterpriseApiClient(EdxRestApiClient):
```
brittneyexline: I'm pretty sure we already have an api client for hitting enterprise apis: https://github.com/edx/ecommerce/blob/master/ecommerce/enterprise/api.py
you should probably add the with_access_to call to this file.

ecommerce/enterprise/clients.py
```diff
@@ -0,0 +1,101 @@
...
+            return None
+
+        return response['results'][0]
+
+    def get_with_access_to(self, user, enterprise_id):
+        """
+        Get the enterprises that this user has access to for the data api permission django group.
+        """
+        cache_key = get_cache_key(
+            resource='enterprise-customer',
```
brittneyexline: maybe 'enterprise-customer-access' would reflect the resource better?

ecommerce/extensions/api/permissions.py
```diff
@@ -43,3 +47,57 @@ class IsStaffOrModelPermissionsOrAnonReadOnly(permissions.DjangoModelPermissions
...
+class IsStaffOrEnterpriseUser(permissions.BasePermission):
+    """
+    Permission that checks to see if the request user is staff or is associated with the enterprise in the request.
+
+    NOTE: This permission check may make a request to the LMS to get the enterprise association if it is not already in
+    the session. This fetch should go away when JWT Scopes are fully implemented and the association is stored on
+    the token.
+    """
+
+    ENTERPRISE_DATA_API_GROUP = 'enterprise_data_api_access'
```
brittneyexline: seems weird that we are defining the same constant twice?

ecommerce/extensions/api/permissions.py
```diff
@@ -43,3 +47,57 @@ class IsStaffOrModelPermissionsOrAnonReadOnly(permissions.DjangoModelPermissions
     def has_permission(self, request, view):
         user = request.user
         return user.is_staff or super(IsStaffOrModelPermissionsOrAnonReadOnly, self).has_permission(request, view)
+
+
+class IsStaffOrEnterpriseUser(permissions.BasePermission):
```
brittneyexline: so, i think i may have pointed you to the wrong permission class to use in the data api code. we should be mimicking HasDataAPIDjangoGroupAccess (https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/permissions.py#L70).
staff users will also need to go through the with_access_to check.

brittneyexline (commented):
you're on the right track, but there are some things that will need to be adjusted. i'm also wondering if you are going to add the permission check to the endpoints in this PR or in a separate PR?

ecommerce/extensions/api/permissions.py
```diff
@@ -43,3 +48,35 @@ class IsStaffOrModelPermissionsOrAnonReadOnly(permissions.DjangoModelPermissions
...
+        if not enterprise_data:
+            return None
+
+        return enterprise_data
+
+    def has_permission(self, request, view):
+        """
+        Verify the user is staff or the associated enterprise matches the requested enterprise.
+        """
+        enterprise_in_url = request.parser_context.get('kwargs', {}).get('enterprise_id', '')
```
brittneyexline: not all of the urls to our views will have the enterprise_id in it - you will have to add logic to derive it from the coupon product if that is what is available instead

ecommerce/extensions/api/permissions.py
```diff
@@ -43,3 +48,35 @@ class IsStaffOrModelPermissionsOrAnonReadOnly(permissions.DjangoModelPermissions
...
+
+        permitted = self.get_enterprise_with_access_to(request.site, request.user, enterprise_in_url)
```
brittneyexline: just calling out that you are planning on relying on the cache instead of the user's session to prevent further redundant calls to the endpoint?

ecommerce/enterprise/api.py
```diff
@@ -209,3 +210,43 @@ def catalog_contains_course_runs(site, course_run_ids, enterprise_customer_uuid,
...
+    """
+    Get the enterprises that this user has access to for the data api permission django group.
+    """
+    enterprise_data_api_group = 'enterprise_data_api_access' #TODO add to settings
+    api_resource_name = 'enterprise-customer'
+    api = site.siteconfiguration.enterprise_api_client
+    endpoint = getattr(api, api_resource_name)
+
+    cache_key = get_cache_key(
+        resource=api_resource_name,
```
brittneyexline: can we add a little more detail to this cache key name to differentiate it from other calls to the enterprise-customer set of endpoints? see https://github.com/edx/ecommerce/blob/410496eee46e6b3c4b80a54ed32f12168d12e70c/ecommerce/enterprise/api.py#L185

ecommerce/enterprise/clients.py
irfanuddinahmad: api client updated.

ecommerce/enterprise/clients.py
irfanuddinahmad: Updated key

ecommerce/extensions/api/permissions.py
irfanuddinahmad: added settings value

ecommerce/extensions/api/permissions.py
irfanuddinahmad: added the new permission

ecommerce/extensions/api/permissions.py
irfanuddinahmad: added coupon based logic

ecommerce/extensions/api/permissions.py
irfanuddinahmad: Yes. This is the case with other permissions also.

ecommerce/enterprise/api.py
irfanuddinahmad: Done

ecommerce/extensions/api/permissions.py
```diff
@@ -43,3 +48,36 @@ class IsStaffOrModelPermissionsOrAnonReadOnly(permissions.DjangoModelPermissions
...
+
+    def has_permission(self, request, view):
+        """
+        Verify the user is staff or the associated enterprise matches the requested enterprise.
+        """
+        enterprise_id = request.parser_context.get('kwargs', {}).get('enterprise_id', '')
+        if not enterprise_id:
+            coupon = view.get_object()
+            enterprise_id = (
+                coupon.attr.coupon_vouchers.vouchers.first().enterprise_offer.condition.enterprise_customer_uuid)
```
brittneyexline: nit: end parentheses should be on the next line

ecommerce/enterprise/api.py
```diff
@@ -209,3 +209,38 @@ def catalog_contains_course_runs(site, course_run_ids, enterprise_customer_uuid,
...
+    except (ConnectionError, SlumberHttpBaseException, Timeout) as exc:
+        logger.warning('Unable to retrieve Enterprise Customer with_access_to details for user %s: %r',
+                       user.username, exc)
+        return None
+    if response.get('results', None) is None or response['count'] == 0:
+        logger.warning('Unable to process Enterprise Customer with_access_to details for user %s, enterprise %s:'
+                       ' No Results Found', user.username, enterprise_id)
+        return None
+
+    TieredCache.set_all_tiers(cache_key, response['results'][0], settings.ENTERPRISE_API_CACHE_TIMEOUT)
```
brittneyexline, line 247: this setting is 300s, which is only 5 min. i don't have a very strong feeling about this, but i think this points towards using the user session as a better point of storing this data to avoid an api call. since we'll be ripping this out at some point soon, i don't think you need to change it, unless the number of calls to this endpoint become problematic before the other authz stuff is ready.

brittneyexline (approved):

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -71,6 +72,7 @@ def get(self, request):
 
 class EnterpriseCouponViewSet(CouponViewSet):
     """ Coupon resource. """
+    permission_classes = (IsAuthenticated, HasDataAPIDjangoGroupAccess,)
```
brittneyexline: we need to separate the code management endpoints into their own view set class so that we don't apply this permission check to the endpoints used by the coupon admin pages.

brittneyexline (changes requested):
we'll need to adjust how the permission_class gets applied by moving the code management endpoints into their own viewset. also, it appears that you are also just below the coverage threshold

ecommerce/extensions/api/permissions.py
irfanuddinahmad: Changed the method for getting the enterprise_id

ecommerce/enterprise/api.py
irfanuddinahmad: Noted.

ecommerce/extensions/api/v2/views/enterprise.py
irfanuddinahmad: I have added the permission in the detail_route decorator. Please review.

brittneyexline (approved):

PR #2163, opened 2019-03-04
EG-60 EG-61 Run make upgrade

crice100:
Running _make upgrade_ without unpinning anything results in enough changes that I thought it might be good to make a PR with just those changes, before unpinning (or upgrading the pin on) _django-oscar_.

Upgrading _django-oscar_ will follow in a later PR.

Edit:
@cpennington pointed out that _oauthlib_ gets a major version bump in this PR (from 2.1.0 to 3.0.1). The release notes for 3.0.0 mention some breaking changes:

```
OAuth2.0 Provider - API/Breaking Changes

Add "request" to confirm_redirect_uri #504
confirm_redirect_uri/get_default_redirect_uri has a bit changed #445
invalid_client is now a FatalError #606
Changed errors status code from 401 to 400:
invalid_grant: #264
invalid_scope: #620
access_denied/unauthorized_client/consent_required/login_required #623
401 must have WWW-Authenticate HTTP Header set. #623
```
https://github.com/oauthlib/oauthlib/releases 

_pyopenssl_ also gets a major version bump (from 18.0.0 to 19.0.0). Its breaking changes are:

```
Backward-incompatible changes:
X509Store.add_cert no longer raises an error if you add a duplicate cert. #787
```
https://pyopenssl.org/en/stable/changelog.html

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -14,17 +14,17 @@ billiard==3.3.0.23        # via celery
 cached-property==1.5.1    # via zeep
 celery==3.1.26.post2      # via edx-ecommerce-worker
 certifi==2018.11.29       # via requests
-cffi==1.11.5              # via cryptography
+cffi==1.12.2              # via cryptography
 chardet==3.0.4            # via requests
 coreapi==2.3.3
 coreschema==0.0.4         # via coreapi
-cryptography==2.4.2       # via pyopenssl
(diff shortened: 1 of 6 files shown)


cpennington:
It looks like we use `login_required` and `consent_required` in the codebase, but at first glance a change to a 400 instead of a 401 seems ok.

cpennington (approved):

PR #2164, opened 2019-03-05
ENT-1627: Bounced codes do not have remind/revoke buttons.

saleem-latif:
**JIRA Ticket:** [ENT-1627](https://openedx.atlassian.net/browse/ENT-1627)

**Description:**
Unredeemed codes do not have a "Remind / Revoke" option next. This PR adds `OFFER_ASSIGNMENT_EMAIL_BOUNCED` status to the filter so that bounced codes are included in the API response and are handled accordingly on the edx-portal side.


irfanuddinahmad (approved):
LGTM!

ecommerce/extensions/api/serializers.py
```diff
@@ -750,7 +750,7 @@ def get_redemptions(self, obj):
             num_assignments = OfferAssignment.objects.filter(
                 code=self.get_code(obj),
                 user_email=self.get_assigned_to(obj),
-                status__in=[OFFER_ASSIGNED, OFFER_ASSIGNMENT_EMAIL_PENDING],
+                status__in=[OFFER_ASSIGNED, OFFER_ASSIGNMENT_EMAIL_PENDING, OFFER_ASSIGNMENT_EMAIL_BOUNCED],
```
mushtaqak, line 753: I don't think bounced codes would ever appear in partial-redeemed since there would not be any email sent in the first place to redeem the code, hence no partial-redemption.

mushtaqak (changes requested):

ecommerce/extensions/api/serializers.py
mushtaqak: @brittneyexline what do you think?

mushtaqak: @adamstankiewicz Can you have a look at this? I will get this merged if this needs fix.

brittneyexline: i think you are right @mushtaqak, but there is really no impact from this change if we merge it.

brittneyexline (approved):

adamstankiewicz (approved):

PR #2166, opened 2019-03-07
Remove duplicate code for initializing API clients.

waheedahmed:
Use JWT access token initialized clients instead of OAuth access token clients, also remove duplicate code for API client initialization.

Oauth access token calls to LMS returning 401 Unauthenticated response for some users.

LEARNER-7145

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -19,7 +19,6 @@ from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
 
 from analytics import Client as SegmentClient
-from ecommerce.core.url_utils import get_lms_url
 from ecommerce.core.utils import log_message_and_raise_validation_error
 from ecommerce.extensions.payment.exceptions import ProcessorNotFoundError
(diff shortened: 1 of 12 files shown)


ecommerce/tests/mixins.py
```diff
@@ -370,7 +370,7 @@ def mock_account_api(self, request, username, data):
             username (string): The username of the user.
             data (dict): Dictionary of data the account API should return.
         """
-        url = '{host}/accounts/{username}'.format(
+        url = '{host}/accounts/{username}/'.format(
```
nasthagiri: What change in this PR triggered the need to add a trailing slash here?  Is this indicative of a possible regression or change in behavior?

nasthagiri (approved):
Looks good overall.  I had just one question - to make sure we're not causing an unexpected change in behavior with trailing slashes.

ecommerce/tests/mixins.py
waheedahmed: @nasthagiri Sorry forgot to remove this one, just added for testing purpose during test fixing. Removed now, thanks for notifying.

uzairr (approved):

uzairr (approved):

PR #2176, opened 2019-03-13
ENT-1558 Django models for authz logic on api endpoints

irfanuddinahmad:
Description: This PR adds Django models for authz logic on api endpoints

JIRA: https://openedx.atlassian.net/browse/ENT-1558


ecommerce/core/models.py
```diff
@@ -642,3 +643,46 @@ def save(self, *args, **kwargs):
                 'Failed to create BusinessClient. BusinessClient name may not be empty.'
             )
         super(BusinessClient, self).save(*args, **kwargs)
+
+
+class EnterpriseCouponAdmin(UserRole):
```
brittneyexline: i would rename this to something like EcommerceFeatureRole - this class is for defining what different roles are available for the feature, and EnterpriseCouponAdmin is one of those roles. we will want to add this new role like so: https://github.com/edx/edx-enterprise/pull/431/files#diff-1d927fb937a4e4e005fb9b0969764aceR9

ecommerce/core/models.py
```diff
@@ -642,3 +643,46 @@ def save(self, *args, **kwargs):
...
+        return "<EnterpriseCouponAdmin {role}>".format(role=self.name)
+
+    def __repr__(self):
+        """
+        Return uniquely identifying string representation.
+        """
+        return self.__str__()
+
+
+class EnterpriseCouponAdminRoleAssignment(UserRoleAssignment):
```
brittneyexline: i would rename this to EcommerceFeatureRoleAssignment, as per above.

ecommerce/core/decorators.py
```diff
@@ -0,0 +1,30 @@
+"""
+Taken from https://github.com/escodebar/django-rest-framework-rules/blob/master/rest_framework_rules/decorators.py.
+"""
+from __future__ import absolute_import
+
+
+def permission_required(*permissions, **decorator_kwargs):
```
brittneyexline: this needs to be moved to edx-rbac so that all services can take advantage of this.

ecommerce/core/admin.py
```diff
@@ -41,3 +47,16 @@ def get_queryset(self, request):
 @admin.register(BusinessClient)
 class BusinessClientAdmin(admin.ModelAdmin):
     pass
+
+
+@admin.register(EcommerceFeatureRole)
+class EcommerceFeatureRoleAdmin(admin.ModelAdmin):
```
brittneyexline: i don't think we should expose this in django admin at this point.

ecommerce/core/admin.py
```diff
@@ -41,3 +47,16 @@ def get_queryset(self, request):
...
+    list_display = ('name', 'created', 'modified')
+    search_fields = ['name']
+
+
+@admin.register(EcommerceFeatureRoleAssignment)
+class EcommerceFeatureRoleAssignmentAdmin(admin.ModelAdmin):
```
brittneyexline: i would hold off on this until https://github.com/edx/edx-rbac/pull/8 has merged - then you can take advantage of the form in implementing this.

brittneyexline (commented):
let's break this up into smaller PRs so we can merge things independently:
* introducing the models can be one PR
* introducing the django admin can be another PR
* put the decorator in edx-rbac
* implement the django rules based permission checks in another PR

ecommerce/core/models.py
irfanuddinahmad: Done

ecommerce/core/models.py
irfanuddinahmad: Done

ecommerce/core/decorators.py
irfanuddinahmad: Done ... https://github.com/edx/edx-rbac/pull/9

ecommerce/core/admin.py
irfanuddinahmad: Ok

ecommerce/core/admin.py
irfanuddinahmad: Ok

PR #2177, opened 2019-03-13
ENT-1673 Fix permissions logic for enterprise coupons

brittneyexline:
**Description**
The permission logic was expecting users to be linked to an enterprise, which may not be the case for staff users. We should let the with_access_to check decide whether the user has access to a particular enterprise. Also, we were not grabbing the enterprise id associated with the request, either directly or from the coupon object. The net result was that anyone linked to an enterprise would be able to access any of these endpoints, and anyone not linked to an enterprise would be denied access, even if they should have access.

### DIFF
--- a/ecommerce/extensions/api/permissions.py
+++ b/ecommerce/extensions/api/permissions.py
@@ -1,12 +1,14 @@
 import logging
 
+from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
 from rest_framework import permissions
-
 from slumber.exceptions import SlumberHttpBaseException
 
-from ecommerce.enterprise.api import fetch_enterprise_learner_data, get_with_access_to
+from ecommerce.enterprise.api import get_with_access_to
+from ecommerce.extensions.api.serializers import retrieve_enterprise_condition
 
+Product = get_model('catalogue', 'Product')
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


ecommerce/extensions/api/permissions.py
```diff
@@ -68,28 +70,35 @@ def get_enterprise_with_access_to(self, site, user, jwt, enterprise_id):
...
+            pk = request.parser_context.get('kwargs', {}).get('pk', '')
             try:
-                learner_data = fetch_enterprise_learner_data(request.site, request.user)['results'][0]
-                if learner_data and 'enterprise_customer' in learner_data:
-                    enterprise_id = learner_data['enterprise_customer']['uuid']
-            except (ConnectionError, IndexError, KeyError, SlumberHttpBaseException, Timeout):
-                LOGGER.exception(
-                    'Failed to retrieve enterprise id for site [%s] and user [%s].', request.site, request.user)
+                coupon = Product.objects.get(pk=pk)
+            except Product.DoesNotExist:
```
irfanuddinahmad, line 99: Do we need to take care of the case when the enterprise associated with the user is different from the enterprise associated with the coupon?

irfanuddinahmad (commented):
I think the access should be based on the user credentials (e.g. enterprise_id associated with the user). Getting the enterprise_id associated with a product that the user has purchased may be a basis for access for a subset of cases ... 
Also ... I am not sure how anyone associated with an enterprise and not having the 'enterprise_data_api_access' permission would be able to access these endpoints ...

ecommerce/extensions/api/permissions.py
brittneyexline: the with_access_to endpoint handles that logic.

brittneyexline:
@irfanuddinahmad - the with_access_to endpoint performs the check that we want - does this user have the enterprise_data_api_access group, and are they associated with the enterprise or staff.

keep in mind that "Coupons" are Products, and because the endpoints for everything except overview are handing us coupon id (aka product id), we need to get the enterprise associated with the coupon to determine which enterprise to send in the with_access_to call. you are right that at this point, this permission class will only be applicable to this subset of endpoints - this is intentional, and will soon be  replaced by what you are working on.

irfanuddinahmad (approved):
LGTM!

PR #2180, opened 2019-03-18
Added hubspot deal's description.

hasnain-naveed:


ecommerce/core/management/commands/sync_hubspot.py
```diff
@@ -72,6 +73,11 @@
                 'dataType': 'DATETIME',
                 'targetHubspotProperty': 'closedate'
             },
+            {
```
ziafazal, line 76: Do we really need to define `description` property here?

ziafazal (approved):
👍

ecommerce/core/management/commands/sync_hubspot.py
hasnain-naveed: yes, we need to define this property in settings. I have tested it.

georgebabey (commented):
Code looks good but I'm wondering if we need to remove LineItems and Products from the `HUBSPOT_ECOMMERCE_SETTINGS`? I worry the call may fail if we try posting and the customer does not have access to these objects in Hubspot.

PR #2181, opened 2019-03-19
ENT-1697 Django admin for EcommerceFeatureRoleAssignment model

irfanuddinahmad:
Description: This PR adds a django admin form for EcommerceFeatureRoleAssignment model

Jira: https://openedx.atlassian.net/browse/ENT-1558
        https://openedx.atlassian.net/browse/ENT-1697


requirements/base.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    make upgrade
+#    pip-compile --upgrade -o requirements/base.txt requirements/base.in
```
saleem-latif, line 5: Did you use `pip-compile --upgrade -o requirements/base.txt requirements/base.in` instead of `make upgrade`? If so, I think it would be better to use `make upgrade` for upgrading requirements. This would make sure all requirements files are updated correctly.

saleem-latif (approved):
One question otherwise looks good 👍

christopappas (approved):

requirements/base.txt
irfanuddinahmad: make upgrade was used

ecommerce/core/forms.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+
+from edx_rbac.admin.forms import UserRoleAssignmentAdminForm
+
+from ecommerce.core.models import EcommerceFeatureRoleAssignment
+
+
+class EcommerceFeatureRoleAssignmentAdminForm(UserRoleAssignmentAdminForm):
```
christopappas, line 8: actually, question. I see in Ammar's PR he actually added an admin to create the roles themselves, in addition to an admin for creating RoleAssignments. Is that something we want here too?

brittneyexline: i chose to only create an admin form for the role assignments because i don't think there is a need to expose adding new roles to django admin. this is because creating those roles will do nothing without code changes, so it seems safer to have people add them via migration.

brittneyexline (approved):

ecommerce/core/forms.py
christopappas: okie doke

PR #2182, opened 2019-03-21
ENT-1698 django rules based permission checks for enterprise api endpoints

irfanuddinahmad:
Description: Updates authorization checks on all API endpoints, that are currently performing custom authorization logic related to enterprise, to use the django-rules based permission checks. Uses 'use_role_based_access_control' waffle switch

JIRA: https://openedx.atlassian.net/browse/ENT-1558

Test Instructions:

High level tests (thanks @brittneyexline ):

Waffle switch off:

- staff users with the enterprise_data_api_access group who are may or may not be linked to an enterprise get access.
- non-staff users linked to the same enterprise as the request object who have the enterprise_data_api_access group get access.
- all other users are denied (staff without the group, matching enterprise users without the group, non matching enterprise users with the group)

Waffle switch on:

- users with the system wide role in their jwt and a matching enterprise get access.
- users with a feature specific role assignment with a matching enterprise get access.
- all other users are denied (this includes staff)


ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,24 @@
...
+
+import rules
+from edx_rbac.utils import user_has_access_via_database
+
+from ecommerce.core.constants import ENTERPRISE_COUPON_ADMIN_ROLE
+from ecommerce.core.models import EcommerceFeatureRoleAssignment
+
+
+@rules.predicate
+def is_enterprise_coupon_admin(user, obj):  # pylint: disable=unused-argument
```
brittneyexline: i think we will need two predicates, similar to what is in the data api PR: https://github.com/edx/edx-enterprise-data/pull/112/files#diff-6c49a377b4da1d1d150c51be9f4afffeR19

brittneyexline (commented):
good start! i would take a look at the data api PR to see how the predicates should be set up. the comment about checking enterprise id applies here as well. also, please add testing instructions :-)

brittneyexline:
@irfanuddinahmad @mushtaqak - based on our discussion and a quick look at the code, here are a few options:
1. make the enabled flag passed into the decorator take a function instead. as you said, it's not pretty, but there is already a lot going on in the decorator arguments.
2. try mocking the waffle call. this might be tricky to get right considering there are other places we call waffle.
3. move the waffle switch check to the predicate function directly. also not pretty, but considering this is being used as a implementation detail of how we're rolling it out, and we would want to remove the conditional logic anyway, this might be the best option in terms of both clarity and simplicity. i think it might require changing the decorator to allow a return value from the predicates that says skip this check, so that it doesn't return true or throw PermissionDenied.

my vote is for option 3, but i will leave it to you.

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,51 @@
...
+    request = get_request_or_stub()
+    decoded_jwt = get_decoded_jwt_from_request(request)
+    if decoded_jwt:
+        return request_user_has_implicit_access_via_jwt(decoded_jwt, ENTERPRISE_COUPON_ADMIN_ROLE)
+    else:
+        return False
+
+
+@rules.predicate
+def request_user_has_explicit_access(*args, **kwargs):  # pylint: disable=unused-argument
```
christopappas: This is a nitpicky thing, but if we're only using `user` from the args, should the function definition just accept `user` and not `*args` and `**kwargs`?

christopappas: You can ignore the above

ecommerce/extensions/api/v2/utils.py
```diff
@@ -38,3 +41,14 @@ def send_new_codes_notification_email(site, email_address, enterprise_id, coupon
...
     logger.info('New codes email sent to enterprise customer [%s] for coupon [%s]', enterprise_id, coupon_id)
+
+
+def rbac_enabled():
+    """
+    Check if role based access control is enabled or not.
+    """
+    try:
+        return waffle.switch_is_active(USE_ROLE_BASED_ACCESS_CONTROL)
+    except:
```
christopappas: Nitpick-ish, but, I'm always a fan of specifying the exact Exceptions that could be raised, instead of a bare `except`

ecommerce/extensions/api/permissions.py
```diff
@@ -94,6 +97,8 @@ def has_permission(self, request, view):
         """
         Verify the user is staff or the associated enterprise matches the requested enterprise.
         """
+        if waffle.switch_is_active(USE_ROLE_BASED_ACCESS_CONTROL):
```
christopappas, line 100: So if our switch is active, then we are letting the user simply pass this permission check, eh?

I think my personal preference would be to not even have this check happen at all (kind of what we're doing in the data api); however, this seems to be functionally equivalent, so I'm cool with it. 👍  (Especially where it's likely we'll be dropping this code altogether soon enough)

christopappas:
Overall this looks good to me! Just gotta make codecov happy :)

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,50 @@
...
+def request_user_has_implicit_access(*args, **kwargs):  # pylint: disable=unused-argument
+    """
+    Check that if request user has implicit access to `ENTERPRISE_COUPON_ADMIN_ROLE` feature role.
+     Returns:
+        boolean: whether the request user has access or not
+    """
+    if not waffle.switch_is_active(USE_ROLE_BASED_ACCESS_CONTROL):
+        return True
+    request = get_request_or_stub()
+    decoded_jwt = get_decoded_jwt_from_request(request)
```
brittneyexline, line 30: we are going to need to take into account the context (enterprise id) and make sure it matches the request object, both here and below. for the explicit access check, we should add a nullable field to capture the enterprise id on the role assignment.

ecommerce/extensions/api/permissions.py
brittneyexline: my concern with returning true here and in the predicates is that we will unintentionally give access all the time when the switch is on or off. we should consider what return value will result in skipping the check and relying on the correct one to evaluate - i don't know if returning true will result in that behavior.

brittneyexline (commented):
i think as you are nearing the end of this implementation, it's time to write out test instructions for devstack for both the waffle switch on and off. at a high level, we need to test:
with the switch off:
* staff users with the enterprise_data_api_access group who are may or may not be linked to an enterprise get access.
* non-staff users linked to the same enterprise as the request object who have the enterprise_data_api_access group get access.
* all other users are denied (staff without the group, matching enterprise users without the group, non matching enterprise users with the group)

with the waffle switch on:
* users with the system wide role in their jwt and a matching enterprise get access.
* users with a feature specific role assignment with a matching enterprise get access.
* all other users are denied (this includes staff, for now, since we are relying on a future story to implement this)

ecommerce/enterprise/rules.py
irfanuddinahmad: @brittneyexline updated context validation ... is this what you had in mind?

ecommerce/extensions/api/permissions.py
irfanuddinahmad: ... the return true here is using an opposite condition as compared to the one in the predicates

ecommerce/enterprise/rules.py
irfanuddinahmad: Added

ecommerce/extensions/api/v2/utils.py
irfanuddinahmad: Changed

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,116 @@
...
+                if basket_lines.exists():
+                    return True
+    return False
+
+
+def has_correct_context_for_explicit_access(user, obj):
+    """
+    Check if request has correct role assignment context.
+    """
+    if isinstance(obj, unicode):
```
christopappas: do we want to use six.string_types here? 1 less thing to change for the impending py3 upgrade

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,116 @@
...
+
+def has_correct_context_for_implicit_access(obj, decoded_jwt, system_wide_role_name):
+    """
+    Check if request has correct role assignment context.
+    """
+    jwt_roles_claim = decoded_jwt.get('roles', []) if decoded_jwt else []
+    for role_data in jwt_roles_claim:
+        role_in_jwt, enterprise_id_in_jwt = role_data.split(':')
+        if role_in_jwt == system_wide_role_name:
+            if isinstance(obj, unicode):
```
christopappas: same question here about six.string_types

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,113 @@
...
+
+def has_correct_context_for_implicit_access(obj, decoded_jwt, system_wide_role_name):
+    """
+    Check if request has correct role assignment context.
+    """
+    jwt_roles_claim = decoded_jwt.get('roles', []) if decoded_jwt else []
+    for role_data in jwt_roles_claim:
+        role_in_jwt, enterprise_id_in_jwt = role_data.split(':')
+        if role_in_jwt == system_wide_role_name:
+            if isinstance(obj, six.string_types):
```
brittneyexline: i think it would be cleaner to either separate out this string or Product check into a separate function and just return a related enterprise id and pass it in for comparison. it looks kind of messy here and below - feels like it could by DRYer.

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,113 @@
...
+    decoded_jwt = get_decoded_jwt_from_request(request)
+
+    implicit_access = (
+        request_user_has_implicit_access_via_jwt(decoded_jwt, ENTERPRISE_COUPON_ADMIN_ROLE) if decoded_jwt else False
+    )
+
+    if not implicit_access:
+        return False
+
+    return has_correct_context_for_implicit_access(obj, decoded_jwt, SYSTEM_ENTERPRISE_ADMIN_ROLE)
```
brittneyexline: maybe once i merge https://github.com/edx/edx-rbac/pull/14/files we can refactor this to one call to the util function.

ecommerce/extensions/api/permissions.py
brittneyexline: i think you are right @irfanuddinahmad - this should work.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -210,6 +211,9 @@ def update_offer_data(self, request_data, vouchers, site):
             super(EnterpriseCouponViewSet, self).update_range_data(request_data, vouchers)
 
     @detail_route(url_path='codes', permission_classes=[IsAuthenticated, HasDataAPIDjangoGroupAccess])
+    @permission_required(
+        'enterprise.can_view_coupon', fn=lambda request, pk, format=None: Product.objects.filter(pk=pk).first(),
```
brittneyexline: another option here would be to create a function that handles grabbing the enterprise id from a given product id and call it in this lambda function. that way the predicates are only concerned with enterprise ids being passed in.

ecommerce/settings/base.py
```diff
@@ -232,6 +232,7 @@
     'edx_django_utils.cache.middleware.TieredCacheMiddleware',
     'edx_rest_framework_extensions.middleware.RequestMetricsMiddleware',
     'edx_rest_framework_extensions.auth.jwt.middleware.EnsureJWTAuthSettingsMiddleware',
+    'edx_rest_framework_extensions.auth.jwt.middleware.JwtAuthCookieMiddleware',
```
brittneyexline: can you explain why we had to reorder this middleware setting for my own understanding?

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,113 @@
...
+    if not waffle.switch_is_active(USE_ROLE_BASED_ACCESS_CONTROL):
+        return True
+    explicit_access = EcommerceFeatureRoleAssignment.objects.filter(
+        user=user,
+        role__name=ENTERPRISE_COUPON_ADMIN_ROLE
+    ).first()
+    if not explicit_access:
+        return False
+
+    return has_correct_context_for_explicit_access(user, obj)
```
brittneyexline: here also, we can consolidate to one call after the edx-rbac PR is merged.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -210,6 +211,9 @@ def update_offer_data(self, request_data, vouchers, site):
...
+        enabled=True)
```
brittneyexline: do you think we should remove this enabled flag functionality?

ecommerce/enterprise/rules.py
irfanuddinahmad: Done

ecommerce/enterprise/rules.py
irfanuddinahmad: Done

ecommerce/settings/base.py
irfanuddinahmad: This was updated in master and I merely fixed a conflict. :)

ecommerce/extensions/api/v2/views/enterprise.py
irfanuddinahmad: Yes. This was added to support rollout with waffle ... however, it seems we cannot pass anything which uses the database ... some nose issue. Hence, it is quite redundant as far as ecommerce is concerned.

ecommerce/enterprise/rules.py
irfanuddinahmad: Done

ecommerce/enterprise/rules.py
irfanuddinahmad: Done

ecommerce/extensions/api/v2/views/enterprise.py
irfanuddinahmad: Done

ecommerce/enterprise/rules.py
irfanuddinahmad: Done

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Check that if request user has implicit access to `ENTERPRISE_COUPON_ADMIN_ROLE` feature role.
+     Returns:
+        boolean: whether the request user has access or not
+    """
+    if not waffle.switch_is_active(USE_ROLE_BASED_ACCESS_CONTROL):
+        return True
+    request = get_request_or_stub()
+    decoded_jwt = get_decoded_jwt_from_request(request)
+    if not context:
```
mushtaqak, line 31: I think we should add a comment why this is neccessary or this should be done in rbac [here](https://github.com/edx/edx-rbac/blob/daade8dd8711f30087e5ed1ead9a562fbe275bea/edx_rbac/utils.py#L88)

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,58 @@
...
+def request_user_has_explicit_access(user, context):
+    """
+    Check that if request user has explicit access to `ENTERPRISE_COUPON_ADMIN_ROLE` feature role.
+    Returns:
+        boolean: whether the request user has access or not
+    """
+    if not waffle.switch_is_active(USE_ROLE_BASED_ACCESS_CONTROL):
+        return True
+    if not context:
+        return False
```
mushtaqak, line 48: same as above.

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@rules.predicate
+def request_user_has_implicit_access(user, context):  # pylint: disable=unused-argument
+    """
+    Check that if request user has implicit access to `ENTERPRISE_COUPON_ADMIN_ROLE` feature role.
+     Returns:
+        boolean: whether the request user has access or not
+    """
+    if not waffle.switch_is_active(USE_ROLE_BASED_ACCESS_CONTROL):
```
mushtaqak, line 27: it would be great if we could add a new predicate for this switch  that can be placed in all rules?

ecommerce/enterprise/rules.py
```diff
@@ -0,0 +1,58 @@
...
+        return True
+    request = get_request_or_stub()
+    decoded_jwt = get_decoded_jwt_from_request(request)
+    if not context:
+        return False
+    return request_user_has_implicit_access_via_jwt(
+        decoded_jwt, ENTERPRISE_COUPON_ADMIN_ROLE, context
+    ) if decoded_jwt else False
```
mushtaqak, line 35: Should this be handled in rbac ?

ecommerce/enterprise/constants.py
```diff
@@ -3,3 +3,6 @@
 
 # Waffle switch used to enable/disable using Enterprise Offers for Coupons.
 ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH = 'enable_enterprise_offers_for_coupons'
+
+# Waffle switch used to enable/disable using role based access control.
+USE_ROLE_BASED_ACCESS_CONTROL = 'use_role_based_access_control'
```
mushtaqak, line 8: This can be renamed to `ENTERPRISE_ROLE_BASED_ACCESS_CONTROL_SWITCH` for consistency.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -135,7 +149,6 @@ def setUp(self):
         self.course = CourseFactory(id='course-v1:test-org+course+run', partner=self.partner)
         self.verified_seat = self.course.create_or_update_seat('verified', False, 100)
         self.enterprise_slug = 'batman'
-
```
mushtaqak, line 138: unintentional ?

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -1781,3 +1794,1806 @@ def test_coupon_codes_remind_all_with_invalid_code_filter(self):
...
+            'enterprise_customer': {'name': 'test enterprise', 'id': str(uuid4()).decode('utf-8')},
+            'enterprise_customer_catalog': str(uuid4()).decode('utf-8'),
+            'notify_email': 'batman@gotham.comics',
+        }
+
+        self.course = CourseFactory(id='course-v1:test-org+course+run', partner=self.partner)
+        self.verified_seat = self.course.create_or_update_seat('verified', False, 100)
+        self.enterprise_slug = 'batman'
+        self.role = EcommerceFeatureRole.objects.get(name=ENTERPRISE_COUPON_ADMIN_ROLE)
+        if self.role:
```
mushtaqak, line 1837: I don't think we need to add if statement here, since we are using `.get` which will throw `DoesNotExist` if not found.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -1781,3 +1794,1806 @@ def test_coupon_codes_remind_all_with_invalid_code_filter(self):
...
+
+    def test_implicit_permission_coupon_overview(self):
+        """
+        Test that we get implicit access via role assignment
+        """
+        Switch.objects.update_or_create(name=ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH, defaults={'active': True})
+
+        response = self.get_response('POST', ENTERPRISE_COUPONS_LINK, self.data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        EcommerceFeatureRoleAssignment.objects.all().delete()
```
mushtaqak, line 2546: Why are we deleting this almost in every second test ?

mushtaqak (changes requested):
This is looking almost ready. I had a couple comments regarding rbac ...

ecommerce/enterprise/rules.py
irfanuddinahmad: Agreed. I would suggest a separate ticket for the rbac work.

ecommerce/enterprise/rules.py
irfanuddinahmad: Agreed. This could be added to the rbac refactoring.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
irfanuddinahmad: This will disable the explicit permission so that we can effectively test the implicit ones.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
irfanuddinahmad: This should be fine.

ecommerce/enterprise/constants.py
irfanuddinahmad: That seems too long. Also, this code will be removed pretty soon.

ecommerce/enterprise/rules.py
irfanuddinahmad: Since we are fetching it in this function, we should not pass on the validation to rbac.

ecommerce/enterprise/rules.py
irfanuddinahmad: This is temporary code that will be removed very soon.

ecommerce/enterprise/rules.py
irfanuddinahmad: @christopappas @brittneyexline your thoughts on returning False when context is absent ... as would be the case if the enterprise_id cannot be found in Invoice.business_client

brittneyexline (approved):

mushtaqak (approved):

PR #2189, opened 2019-03-26
ENT-1716 Safely pull enterprise id from the voucher during order fulfillment

brittneyexline:
For new enterprise coupons and some migrated ones, we were not properly creating the enterprise user, enrollment, or consent records when users redeemed the coupon. This PR addresses this bug.


christopappas:
Is it worth writing a regression test for this?

christopappas (approved):

PR #2191, opened 2019-03-26
ARCH-603: (part 1) update lms_user_id via social-auth for analytics

robrap:
When social-auth makes user_id available to ecommerce, uses this as the backup lms_user_id for later analytics calls.

This is part 1 of a two part change.  This code will use the user_id from social-auth when it is available, but it won't yet be available in part 1.  It will start writing some NewRelic metrics that can be tested.

Part 2 will require the auth-backends upgrade to actually make the user_id available via social-auth.  But this upgrade requires the following:
- Ensure edx-platform change https://github.com/edx/edx-platform/pull/20057 is in Production.
- Add application access in django admin for ecommerce-sso that includes `user_id` scope.
- Ensure this upgrades to 2.0.0 to auth-backends.

ARCH-603

### DIFF
--- a/ecommerce/settings/base.py
+++ b/ecommerce/settings/base.py
@@ -219,7 +219,6 @@ MIDDLEWARE_CLASSES = (
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.contrib.sites.middleware.CurrentSiteMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
-    'edx_rest_framework_extensions.auth.jwt.middleware.EnsureJWTAuthSettingsMiddleware',
     'waffle.middleware.WaffleMiddleware',
     # NOTE: The overridden BasketMiddleware relies on request.site. This middleware
     # MUST appear AFTER CurrentSiteMiddleware.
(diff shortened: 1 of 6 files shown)


ecommerce/settings/base.py
```diff
@@ -451,6 +451,11 @@
 SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT = None
 SOCIAL_AUTH_EDX_OAUTH2_LOGOUT_URL = None
 
+# TODO: Determine where this actually should live.
+# This adds the user_id to the social auth extra data in the database.
+# Note: the third param keeps this value from being overwritten when it is not found.
+SOCIAL_AUTH_EDX_OAUTH2_EXTRA_DATA = [('user_id', 'lms_user_id', True)]
```
robrap: Reviewer: As stated here, I don't know where this should live.  The other settings live in the Site Configurations, but that seems more painful.  Worst case, I think this will add `{'lms_user_id': 'None'}` to the social-auth extra data if this gets used in places where `user_id` is not available.

robrap: FYI: Moving this into auth-backends.  Updates coming.

robrap:
@nasthagiri @douglashall: This is ready for review with the caveat that tests may not be passing.  I was having difficulty running them locally.

douglashall (approved):

PR #2196, opened 2019-03-29
ARCH-603: clean-up lms_user_id code

robrap:
- sets custom metric and returns as soon as an lms_user_id found.
- rename custom metric 'ecommerce_user_id' to
  'ecommerce_user_missing_lms_user_id'.

ARCH-603

PR #2198, opened 2019-04-01
ARCH-603: send user_id from jwt to newrelic

robrap:
Pulls `user_id` from JWT and sets it in NewRelic.  Does not use it yet.

A follow-up PR would add a test and actually make use of this user_id.  Just want to ensure that all is well.

Includes approved clean-up code from https://github.com/edx/ecommerce/pull/2196, with additional refactor into different private functions.

ARCH-603 (Oops.  Branch name has incorrect JIRA ticket number.)


ecommerce/core/models.py
```diff
@@ -491,31 +499,77 @@ def access_token(self):
...
+        Returns the LMS user_id, or None if not found.
         """
-        # get lms_user_id passed through tracking_context.
-        tracking_context = self.tracking_context or {}
-        lms_user_id_tracking_context = tracking_context.get('lms_user_id')
-        monitoring_utils.set_custom_metric('lms_user_id_tracking_context', lms_user_id_tracking_context)
+        # TODO: If found, return user_id from JWT first
+        # - To start, being careful and just adding the metric
+        # - Are there any cases where the JWT could be for the wrong user, like ecommerce-worker?
+        self._get_lms_user_id_from_jwt()
```
robrap: @douglashall: The idea is to push this out asap to get logging of this value, and then to follow-up soon to actually return this value if found.

douglashall (approved):

ecommerce/core/models.py
```diff
@@ -491,31 +499,77 @@ def access_token(self):
...
+        Returns None if not found.
+
+        Side effect:
+            If found, writes custom metric: 'lms_user_id_jwt_cookie'
+        """
+        request = crum.get_current_request()
+        if not request:
+            return None
+
+        decoded_jwt = get_decoded_jwt_from_jwt_cookie(request)
```
robrap, line 532: @nasthagiri: FYI:
1. I renamed on import because I find this name more clear when not appearing next to the import, since there are different possibilities for what JWT one might be referring to.
2. I also understand that `from_jwt_cookie` is different from the argument `request`, but I think that `jwt_cookie` is more precise and helpful, and the name `get_decoded_jwt_from_jwt_cookie_in_request` seems too long.

If this persuades you at all, I'm happy to update the library.  If not, no worries, and I can let my opinion shine on more local PRs. :)

PR #2200, opened 2019-04-02
add enterprise-openedx-operator permission checks

mushtaqak:
Description: Add permission checks for enterprise_openedx_operator role and update the logic for existing implicit and explicit access checks.

[ENT-1737](https://openedx.atlassian.net/browse/ENT-1737)

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -23,6 +23,7 @@ from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
 
 from analytics import Client as SegmentClient
+from ecommerce.core.constants import ALL_ACCESS_CONTEXT
 from ecommerce.core.utils import log_message_and_raise_validation_error
 from ecommerce.extensions.payment.exceptions import ProcessorNotFoundError
 from ecommerce.extensions.payment.helpers import get_processor_class, get_processor_class_by_name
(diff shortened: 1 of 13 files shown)


mushtaqak:
build is not being triggered, closing and re-opening ...

ecommerce/enterprise/rules.py
```diff
@@ -12,20 +12,35 @@
...
 )
 
-from ecommerce.core.constants import ENTERPRISE_COUPON_ADMIN_ROLE
+from ecommerce.core.constants import ENTERPRISE_COUPON_ADMIN_ROLE, SYSTEM_ENTERPRISE_OPERATOR_ROLE
 from ecommerce.core.models import EcommerceFeatureRoleAssignment
 from ecommerce.enterprise.constants import USE_ROLE_BASED_ACCESS_CONTROL
 
 
+@rules.predicate
+def request_user_is_edx_operator(user, context):    # pylint: disable=unused-argument
```
brittneyexline: i think as long as the system wide role gets mapped to the coupon admin role in the settings, this check won't be needed. the context for the system wide role will be `*`, which is now supported to mean access to all enterprises based on Ammar's PR. so the existing check for the single feature specific role of enterprise_coupon_admin should be sufficient.

christopappas (approved):
Waffle switch refactor stuff looks good. I think we can remove the new predicate here because the same logic is covered in ammar's edx-rbac PR, other than that we're 👍

ecommerce/core/models.py
```diff
@@ -753,9 +754,12 @@ class EcommerceFeatureRoleAssignment(UserRoleAssignment):
 
     def get_context(self):
         """
-        Return the context for this role assignment class.
+        Return the enterprise customer id or `*` if the user has access to all resources.
         """
-        return self.enterprise_id and str(self.enterprise_id)
+        enterprise_id = ALL_ACCESS_CONTEXT
+        if self.enterprise_id:
+            enterprise_id = self.enterprise_id and str(self.enterprise_id)
```
muhammad-ammar: I think we only need `enterprise_id = str(self.enterprise_id)`

muhammad-ammar (changes requested):
LGTM There is one minor comment.

muhammad-ammar (approved):

mushtaqak:
@brittneyexline Can you please merge this? :shipit:

PR #2203, opened 2019-04-09
ENT-1714: Update Overview API endpoint to Include Valid Assignments Remaining

saleem-latif:
**Jira Ticket:** [ENT-1714](https://openedx.atlassian.net/browse/ENT-1714)

**Description:**
Currently, the overview endpoint returns "unassigned codes" but this doesn't actually show the # of un-assignments remaining i.e. if 1 code has 500 potential assignments, it will always only show 1. It would be more useful to the admin to know the # of assignments there are left for them to take action on.


brittneyexline (commented):
looks good, just need to add/fix tests!

christopappas (approved):
👍

mushtaqak (approved):
Looks good 👍

PR #2204, opened 2019-04-10
Paginate courses and coupons view

uzairr:
To improve user experience jerks while loading courses and coupons
are improved so that they can load seamlessly.I created 300 dummy courses 
to verify its behaviour.

It uses djangorestframework-datatables to implement server side processing.
https://django-rest-framework-datatables.readthedocs.io/en/latest/quickstart.html

Sandbox: https://datatable.sandbox.edx.org

Steps to observe changes:

Go to https://ecommerce-datatable.sandbox.edx.org/admin
enter username/password: edx/edx
then go to https://ecommerce-datatable.sandbox.edx.org/courses

Observe seamless datatable of courses

LEARNER-6314


cpennington (approved):
Looks good to me, although it looks like some of the tests need to be updated with the new url format.

waheedahmed (changes requested):
@uzairr I think this ticket was meant for both courses and coupons tabs, are you going to fix that in a separate PR?

There are some other tabs as well (e.g. Enterprise Coupons, Journal Offers etc), @cpennington should we fix them as well or just courses and coupons?

cpennington:
I'd love to see any existing page that is currently fetching every item from the backend up-front (by walking down the pages of the api query) converted to this new pattern. I'm not sure what the scope of that is, though, and we don't have to do all of them all at once. I'd happily have each one merge as soon as it's ready, to reduce load on the system.

waheedahmed (approved):

waheedahmed:
Looks like we need to fix the Enterprise Coupon list view as well in this PR, because it's extending `CouponListView`.

waheedahmed:
@cpennington Please take another look on this PR, I have also updated the coupon and Enterprise coupon tabs.

cpennington (approved):
Looks good to me. Have you tried the Coupon and Enterprise Coupon views with data in them (there's no data at the moment on the sandbox)

waheedahmed:
@cpennington Yes, I have tested those locally.

PR #2208, opened 2019-04-15
WL-1911: Set priority of new enterprise offers same as old enterprise offers

ziafazal:
This PR sets priority of new enterprise offers same as old enterprise offers.

### DIFF
--- a/ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
+++ b/ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
@@ -1,0 +1,60 @@
+"""
+This command change priority of conditional offers.
+"""
+from __future__ import unicode_literals
+
+import logging
+
+from django.core.management import BaseCommand
+from django.template.defaultfilters import pluralize
+from oscar.core.loading import get_model
+
+from ecommerce.extensions.offer.models import OFFER_PRIORITY_VOUCHER
+from ecommerce.extensions.order.management.commands.prompt import query_yes_no
+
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
```diff
@@ -0,0 +1,60 @@
...
+            offer_names = '{names} {index}. {name} {line_feed}'.format(
+                names=offer_names, index=i + 1, name=conditional_offers[i].name, line_feed=line_feed
+            )
+
+        # List down all conditional which needs to be updated.
+        logger.warning(offer_names)
+
+        pluralized = pluralize(count)
+        if query_yes_no(self.CONFIRMATION_PROMPT.format(count=count, pluralized=pluralized), default="no"):
+            conditional_offers.update(priority=OFFER_PRIORITY_VOUCHER)
```
brittneyexline, line 56: do we know approximately how many offers will be updated at once with this query? we should be careful to avoid large updates that might lock the database.

brittneyexline (commented):
looks good overall, but i am concerned about how large that update query might be.

ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
ziafazal: @brittneyexline DB query suggests it is going to update `1264 rows`

brittneyexline (approved):

PR #2209, opened 2019-04-17
Added batch update support to change_priority_of_offers command

ziafazal:
This PR has changes to add batch update support to `change_priority_of_offers` command.
We can update batch of offers by trigger this command with `offset` and `limit` parameters. If we don't provide `limit` it is going to update 100 offers in one go.
@brittneyexline could you please review?

### DIFF
--- a/ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
+++ b/ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
@@ -29,32 +29,65 @@ class Command(BaseCommand):
     CONFIRMATION_PROMPT = u"You're going to change priority of {count} conditional offer{pluralized}. " \
                           u"Do you want to continue?"
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--offset',
+            dest='offset',
+            default=0,
+            help='index to start from.',
+            type=int,
+        )
+        parser.add_argument(
+            '--limit',
+            dest='limit',
+            default=1000,
+            help='Number of offers to update.',
+            type=int,
+        )
+        parser.add_argument(
+            '--priority',
+            dest='priority',
+            default=5,
+            help='Priority of the Offers witch needs to be updated',
+            type=int,
+        )
+
     def handle(self, *args, **options):
-        conditional_offers = ConditionalOffer.objects.filter(
-            offer_type=ConditionalOffer.VOUCHER
-        ).exclude(priority=OFFER_PRIORITY_VOUCHER)
+        limit = options['limit']
+        offset = options['offset']
+        priority = options['priority']
 
-        count = len(conditional_offers)
-        if count == 0:
-            logger.info('No offer found which needs a priority fix')
-            return
+        try:
+            conditional_offers = ConditionalOffer.objects.filter(
+                offer_type=ConditionalOffer.VOUCHER,
+                priority=priority
+            )[offset:offset + limit]
 
-        line_feed = '\n'
-        offer_names = 'Conditional offers to be updated {line_feed}'.format(line_feed=line_feed)
-        for i in range(count):
-            if i == count - 1:
-                line_feed = ''
-            offer_names = '{names} {index}. {name} {line_feed}'.format(
-                names=offer_names, index=i + 1, name=conditional_offers[i].name, line_feed=line_feed
-            )
+            count = len(conditional_offers)
+            if count == 0:
+                logger.info('No offer found which needs a priority fix')
+                return
 
-        # List down all conditional which needs to be updated.
-        logger.warning(offer_names)
+            line_feed = '\n'
+            offer_names = 'Conditional offers to be updated{line_feed}'.format(line_feed=line_feed)
+            for i in range(count):
+                if i == count - 1:
+                    line_feed = ''
+                offer_names = '{names}{index}. {name}{line_feed}'.format(
+                    names=offer_names, index=i + 1, name=conditional_offers[i].name, line_feed=line_feed
+                )
 
-        pluralized = pluralize(count)
-        if query_yes_no(self.CONFIRMATION_PROMPT.format(count=count, pluralized=pluralized), default="no"):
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
```diff
@@ -29,32 +29,69 @@ class Command(BaseCommand):
...
-                line_feed = ''
-            offer_names = '{names} {index}. {name} {line_feed}'.format(
-                names=offer_names, index=i + 1, name=conditional_offers[i].name, line_feed=line_feed
-            )
+        limit = options['limit']
+        offset = options['offset']
+        priority = options['priority']
+
+        try:
+            conditional_offers = ConditionalOffer.objects.filter(
```
brittneyexline, line 61: doesn't seem like we would loop over all of the conditional offers the way this is written, right?

ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
```diff
@@ -29,32 +29,69 @@ class Command(BaseCommand):
...
-        offer_names = 'Conditional offers to be updated {line_feed}'.format(line_feed=line_feed)
-        for i in range(count):
-            if i == count - 1:
-                line_feed = ''
-            offer_names = '{names} {index}. {name} {line_feed}'.format(
-                names=offer_names, index=i + 1, name=conditional_offers[i].name, line_feed=line_feed
-            )
+        limit = options['limit']
+        offset = options['offset']
+        priority = options['priority']
```
brittneyexline, line 58: i've also seen that a common option for these types of scripts is to add an option to specify a sleep in between updates of a couple seconds or so.

ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
```diff
@@ -29,32 +29,69 @@ class Command(BaseCommand):
...
-            conditional_offers.update(priority=OFFER_PRIORITY_VOUCHER)
-            logger.info('Operation completed. %d conditional offer%s updated successfully.', count, pluralized)
-        else:
-            logger.info('Operation canceled.')
-            return
+            # List down all conditional which needs to be updated.
+            logger.warning(offer_names)
+
+            pluralized = pluralize(count)
+            if query_yes_no(self.CONFIRMATION_PROMPT.format(count=count, pluralized=pluralized), default="no"):
```
brittneyexline, line 84: would the result of this be that we query yes or no for each batch of 100? that seems like it would be tedious for devops to run...

ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
ziafazal: yeah. we are puling batch of the offers having there type set to `Voucher` and priority set to that value of `priority` argument which defaults to `5`.

ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
ziafazal: 100 is a default value for limit let me increase it to 1000.

ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
ziafazal: I don't think we need to add sleep option since this command does not auto update next batch.

ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
```diff
@@ -29,32 +29,69 @@ class Command(BaseCommand):
...
+                for offer in conditional_offers:
+                    offer.priority = OFFER_PRIORITY_VOUCHER
+                    offer.save()
```
ziafazal, line 87: @brittneyexline we updating offers one at a time so chance of database deadlock are minimal.

ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
brittneyexline: right, but we aren't incrementing offset to get to the next batch of conditional offers automatically, or am i missing where that is done?

ecommerce/extensions/offer/management/commands/change_priority_of_offers.py
brittneyexline: i think you can still do a batch here of updating up to 100 at once, the problem is when you try to do thousands at once.

ziafazal: we are updating one offer at a time. We are selecting a batch of offers from pool and then updating those offers in that batch one by one.

ziafazal:
@brittneyexline could you please review and merge this one if it looks good.

brittneyexline:
@ziafazal i'm concerned that if you ran the script, it would only process the first 1000 offers and we would have to run it again and again to get through all 20k. Maybe I'm missing something, but it looks like the only iteration is over the 1000 offers in a batch.

ziafazal:
@brittneyexline I have changed this command to update offers one by  one instead of updating them in one go to avoid database deadlock. This command has `limit` argument to select size of the batch which we want to update if provider `limit` argument we'll select a batch of offer and still updating them one by one. This command has one more modification it accepts `priority` argument now we can update only those offers where we have a certain priority value. By default we are passing 5 as priority which mean it would update only those offers where priority is set to 5 and our query suggested we have around 6K such offers so even if we decide to run this command with default arguments we'll have to run it 6 times.

brittneyexline:
ok. please update the devops ticket to outline this, as i don't think it would be obvious that the script would behave that way.

brittneyexline (approved):

feanil:
There is a lot of prior art in the edx-platform for batched managemnet command, a good best practice is to have a sleep parameter between batches so that operators don't have to manually choose which offsets to start at if they just want to run over everything.

PR #2210, opened 2019-04-17
WL-1911 Add better logging to the enterprise condition so we can better understand the errors.

brittneyexline:
Based on my investigations, I suspect that there are a variety of failures, and they are likely coming from this particular code block where we evaluate if a code is applicable to the basket.

However, the information in the logs hasn't been sufficient or consistent between different types of failures. This PR introduces some consistency on what data is logged for the errors impacting enterprise codes and amplifies the exception logging from api calls which was lacking when I looked in splunk.


mushtaqak (approved):
Once build is green 👍

irfanuddinahmad (approved):
Okay once build is green!

PR #2212, opened 2019-04-18
ENT-1838 Filter out enterprise site offers that don't match a user's enterprise

brittneyexline:
This PR changes the CustomApplicator to filter out enterprise site offers when fetching site offers, and grabs them separately by matching on the user's enterprise customer id, if it exists.

Currently we are looping over all of the Enterprise site wide offers which results in us going through redundant logic, and creates noise in the logs. We will cut down on api calls as well with this change.

**Testing Instructions**
This change should have no impact to the user experience beyond possibly faster basket page loads.
As such, the following should be verified:

1. If an enterprise offer is configured for a given enterprise, an enterprise user previously linked to the enterprise should get the offer automatically applied to their basket.
	- Find (or create) a user that is linked to an enterprise.
	- Ensure there is a valid, non-expired enterprise offer for that enterprise: https://ecommerce.stage.edx.org/enterprise/offers
	- Find an enroll-able course in that enterprise’s catalog which has a verified track.
	- Enroll in the verified track for the course.
	- If the enterprise offer is for 100% off, you should bypass the basket page. Otherwise, you should see the offer discount applied in the basket page.

2. If enterprise offer is not configured for a given enterprise, an enterprise user linked to that enterprise should not have any enterprise offers evaluated for them.
	- Find (or create) a user that is linked to an enterprise.
	- Ensure there is either no enterprise offers configured for the given enterprise or, if there is one, ensure it is expired (end date in the past).
	- Find an enroll-able course in that enterprise’s catalog which has a verified track.
	- Enroll in the verified track for the course.
	- The user should end up on basket page with no offer discount applied.

3. If a non-enterprise user purchases a course, they should not have any enterprise offers evaluated for them.
	- Find (or create) a user that is NOT linked to an enterprise.
	- Ensure there is a valid, non-expired enterprise offer for that enterprise: https://ecommerce.stage.edx.org/enterprise/offers
	- Find an enroll-able course in that enterprise’s catalog which has a verified track.
	- Enroll in the verified track for course.
	- The user should end up on basket page with no offer discount applied.

4. If a user (linked or not to an enterprise) attempts to redeem a valid enterprise coupon, they should be able to successfully do so.
	- Find (or create) a user that is NOT linked to an enterprise.
	- Find an enroll-able course in that enterprise’s catalog which has a verified track.
	- Enroll in the verified track for the course.
	- The user should end up on basket page with no offer discount applied.
	- Enter in a valid coupon code for an enterprise. I found this by viewing the Code Management page of edx-portal to get a coupon code for Pied Piper.
	- If the coupon code is for an offer with 100% off, you should bypass the checkout process and be enrolled in the verified track. Otherwise, you should see the offer discount applied in the basket page.

Also, just FYI: Finding courses on stage that will work is kind of a pain. For your convenience, here's the course I was testing with: https://stage.edx.org/course/next-generation-infrastructure-delftx-ngix-1

### DIFF
--- a/ecommerce/enterprise/utils.py
+++ b/ecommerce/enterprise/utils.py
@@ -7,17 +7,21 @@ import logging
 from collections import OrderedDict
 from urllib import urlencode
 
+import crum
 import waffle
 from django.conf import settings
 from django.urls import reverse
 from django.utils.translation import ugettext as _
 from edx_django_utils.cache import TieredCache
 from edx_rest_api_client.client import EdxRestApiClient
+from edx_rest_framework_extensions.auth.jwt.cookies import get_decoded_jwt
 from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import SlumberHttpBaseException
 
+from ecommerce.core.constants import SYSTEM_ENTERPRISE_LEARNER_ROLE
 from ecommerce.core.utils import deprecated_traverse_pagination
+from ecommerce.enterprise.api import fetch_enterprise_learner_data
 from ecommerce.enterprise.exceptions import EnterpriseDoesNotExist
 from ecommerce.extensions.offer.models import OFFER_PRIORITY_ENTERPRISE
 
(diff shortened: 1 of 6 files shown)


ecommerce/enterprise/utils.py
```diff
@@ -444,3 +448,36 @@ def get_enterprise_catalog(site, enterprise_catalog, limit, page):
...
+
+
+def get_enterprise_id_for_current_request_user_from_jwt():
+    request = crum.get_current_request()
+    decoded_jwt = get_decoded_jwt(request)
+    if decoded_jwt:
+        roles_claim = decoded_jwt.get('roles', [])
+        for role_data in roles_claim:
+            role_in_jwt, __, context_in_jwt = role_data.partition(':')
+            if role_in_jwt == SYSTEM_ENTERPRISE_LEARNER_ROLE and context_in_jwt:
```
christopappas, line 460: Only question here is just to verify that we don't want this check for enterprise admin role too. If we do want to add this, there is a minor unit test to tweak

christopappas (approved):
👍 once the unit tests merge

PR #2215, opened 2019-04-22
Revert "move ecommerce task to thier own queue"

syed-awais-ali:
Reverts edx/ecommerce#2129


syed-awais-ali:
Reverting this as It was deployed for stage, but the stage ecom-worker pipeline was stuck on an error which we will now look into.

nadeemshahzad (approved):

PR #2217, opened 2019-04-23
WL-1916 | Added extra check for entitled course in redemption workflow.

hasnain-naveed:


ecommerce/coupons/views.py
```diff
@@ -196,6 +196,9 @@ def get(self, request):
                 {'error': _('Couldn\'t find a matching Enterprise Customer for this coupon.')}
             )
 
+        if enterprise_customer and product.is_course_entitlement_product:
+            return render(request, template_name, {'error': _('Trying to redeem entitle course run.')})
```
waheedahmed: Use more descriptive error message like: `This coupon code is not valid for entitlement course product. Try a different course.`

ecommerce/coupons/tests/test_views.py
```diff
@@ -423,6 +423,30 @@ def test_expired_voucher(self):
...
         self.assertEqual(response.context['error'], 'This coupon code has expired.')
 
+    @httpretty.activate
+    def test_invalid_entitlement_course(self):
+        """ Verify an error is returned when you apply enterprise coupon on entitled product. """
+        entitle_product = ProductFactory(
+            product_class=self.entitlement_product_class, stockrecords__partner=self.partner,
+            stockrecords__price_currency='USD', categories=[]
+        )
+        self.catalog.stock_records.remove(StockRecord.objects.get(product=self.seat))
```
ziafazal: Why we have to remove seat product stock record for this test case?

ziafazal (approved):
👍

ecommerce/coupons/views.py
```diff
@@ -196,6 +196,13 @@ def get(self, request):
                 {'error': _('Couldn\'t find a matching Enterprise Customer for this coupon.')}
             )
 
+        if enterprise_customer and product.is_course_entitlement_product:
```
tuchfarber, line 199: I was going to suggest we take off the `enterprise_customer` check if we shouldn't be able to purchase entitlements at all, but I was unable to get a clear answer from anyone regarding that.. so I think it's fine to limit scope for now.

tuchfarber (approved):

PR #2218, opened 2019-04-23
Added military address in CC

adeelehsan:
military state options added in cyber Source

### DIFF
--- a/ecommerce/extensions/payment/forms.py
+++ b/ecommerce/extensions/payment/forms.py
@@ -11,6 +11,8 @@ from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_class, get_model
 
+from ecommerce.extensions.payment.constants import US_MILITARY_ADDRESS
+
 logger = logging.getLogger(__name__)
 
 Applicator = get_class('offer.applicator', 'Applicator')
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/payment/forms.py
```diff
@@ -167,16 +169,16 @@ def clean(self):
...
-                    # is fixed.
-                    if not pycountry.subdivisions.get(code=code):
-                        raise KeyError
-                except KeyError:
-                    msg = _('{state} is not a valid state/province in {country}.').format(state=state, country=country)
-                    logger.debug(msg)
-                    raise ValidationError({'state': msg})
+                if country != 'US' and state not in US_MILITARY_ADDRESS:
+                    try:
+                        # TODO: Remove the if statement once https://bitbucket.org/flyingcircus/pycountry/issues/13394/
```
waheedahmed, line 175: This issue has been resolved, after verifying can you please also remove the following if statement?

waheedahmed (commented):
Add a test case otherwise LGTM 👍

ecommerce/extensions/payment/tests/test_forms.py
```diff
@@ -183,6 +183,9 @@ def test_organization_field_in_form(self):
         form = PaymentForm(user=self.user, data=data, request=self.request)
         self.assertTrue('organization' in form.fields)
 
+    def test_pycountry_validation_skipped_on_us_military_address(self):
```
waheedahmed: nit: `test_form_validation_with_military_address`

waheedahmed (approved):
One minor comment otherwise 👍

abutterworth:
Do we need a similar change in the frontend-app too? 
https://github.com/edx/frontend-app-payment/blob/master/src/payment/checkout/payment-form/utils/countryStatesMap.js#L17

adeelehsan:
@abutterworth yes you need these changes in the frontend as well. Are we using the same backend view for payment MFE ?

PR #2222, opened 2019-04-24
Fix and refactor num_assignments

mushtaqak:
Fix `num_assignments` and refactor so that we have one single place to get `num_assignments` for all `CodeUsageSerializer` child serializers.

[ENT-1706](https://openedx.atlassian.net/browse/ENT-1706)

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -706,6 +706,16 @@ class CodeUsageSerializer(serializers.Serializer):  # pylint: disable=abstract-m
             'total': max_coupon_usage,
         }
 
+    def num_assignments(self, code, user_email=None):
+        offer_assignments = OfferAssignment.objects.filter(
+            code=code,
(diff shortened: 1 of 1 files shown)


ecommerce/extensions/api/serializers.py
```diff
@@ -714,13 +724,7 @@ def get_assigned_to(self, obj):
 
     def get_redemptions(self, obj):
         redemptions = super(NotAssignedCodeUsageSerializer, self).get_redemptions(obj)
-
-        num_assignments = OfferAssignment.objects.filter(
-            code=self.get_code(obj),
-            status__in=[OFFER_ASSIGNED, OFFER_ASSIGNMENT_EMAIL_PENDING],
```
mushtaqak, line 720: **Note to reviewers:** FYI this was the bug, `OFFER_ASSIGNMENT_EMAIL_BOUNCED` was missing from the `status__in` filter.

saleem-latif (approved):
Looks Good 👍

christopappas (approved):
makes sense to me

PR #2225, opened 2019-04-25
ask lms for a discount

emma-green:


emma-green:
Things I need help with: 
 - why is the build failing?
 - did I put the js in the right place?

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -17,6 +17,19 @@
...
+    if(!window.location.href.includes('discount_jwt')){
+        $.ajax({
+            url: "{{ lms_url_root }}/api/discounts/{{ formset_lines_data.0.1.course_key }}",
+            xhrFields: {
+                withCredentials: true
+            }
+        }).done(function(discount){
+            if(discount.discount_applicable)
+            {
+                window.location.replace(window.location.href + '?discount_jwt='  + discount.jwt)
```
aekao: I thought the point of using an AJAX call from the basket page was to avoid basket page refreshes whenever possible. However, this line refreshes the basket page, I think unnecessarily. 

Why does the discount_jwt have to be passed via GET request/URL? Why not just act on the returned discount data within the .done() section? 

Unless there is something particular about this page, like the back end can't access data from the front unless it's passed as a GET request? (this sounds very unlikely to me, but hey you never know!)

emma-green: If there's a discount, we need to display that information to the user, and then also pass that information to the backend so that it can correctly record the info and charge the user.
^^All of that is *MUCH* easier to do once we have a micro-frontend. And we plan to have one soon.

If we reload (only if there's a discount), we can just send all of the information without needing to do any of the additional work to edit the page and send info to the backend.

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -17,6 +17,19 @@
 
   <script>
     window.paymentErrorPath = "{% url 'payment_error' %}";
+    if(!window.location.href.includes('discount_jwt')){
```
cpennington: Personally, I'd feel more comfortable if this actually parsed the url and pull the `discount_jwt` out of the parameters, rather than just string-matching on the whole URL.

ecommerce/templates/oscar/basket/basket.html
cpennington: I think it might be better to include data on the page itself specifying whether it has already had the discount jwt applied, rather than using the state of the url as that flag.

cpennington:
It looks to me like the tests are failing because your inline script tag isn't being handled correctly by django-compressor. Better practice would be to move that script into a separate file, so that django-compressor can correctly optimize it, I think. Although, to be fair, I'm not super familiar w/ how django-compressor works.

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -17,6 +17,19 @@
...
+        $.ajax({
+            url: "{{ lms_url_root }}/api/discounts/{{ formset_lines_data.0.1.course_key }}",
+            xhrFields: {
+                withCredentials: true
+            }
+        }).done(function(discount){
+            if(discount.discount_applicable)
```
crice100: Did we decide if we're going to verify the jwt as step one?

ecommerce/templates/oscar/basket/basket.html
emma-green: If the discount jwt doesn't apply properly, it might end up in an infinite reloading loop? Unless you have thoughts about how to mitigate that?

emma-green:
@cpennington 
> It looks to me like the tests are failing because your inline script tag isn't being handled correctly by django-compressor. Better practice would be to move that script into a separate file, so that django-compressor can correctly optimize it, I think. Although, to be fair, I'm not super familiar w/ how django-compressor works.

Any suggestions about where to put it?

ecommerce/templates/oscar/basket/basket.html
emma-green: do you have a parsing library suggestion? It looks like URL is not supported by IE. @cpennington

ecommerce/templates/oscar/basket/basket.html
emma-green: Yep. We don't need to verify it in the javascript to decide if we want to reload.

ecommerce/templates/oscar/basket/basket.html
cpennington: I guess I was more suggesting that the data be embedded on the page, rather than as a query parameter. (The same data can be embedded either way).

ecommerce/templates/oscar/basket/basket.html
cpennington: https://www.npmjs.com/package/url-polyfill, maybe? Or, if the data is embedded in the page instead, then you sidestep this issue.

cpennington:
I would just put the code that you added to the inline script tag into a js file instead, and put it w/ the rest of the static js.

ecommerce/templates/oscar/basket/basket.html
emma-green: I think I'm misunderstanding you. The data needs to be sent in the page load request to the ecommerce service. Once it's received there, it can be embeded on the page. However, checking for whether it's embeded on the page, means that you are assuming the computation will always work. If it doesn't work, and retains the parameter (as it will currently, because there's nothing listening for it), it will go in an endless loop.

cpennington: Yeah, I see what you're saying now. I still don't love the fuzziness of checking for a particular string in the URL, but it does make sense to just use that as a flag of "did we check this basket or not".

ecommerce/templates/oscar/basket/basket.html
emma-green: After chatting about this with Cale, I've decided to leave it as it is. It will be ripped when the MFE gets put on, so the overhead of adding another package is not necessary.

ecommerce/templates/edx/base.html
```diff
@@ -68,6 +68,9 @@
   <script src="{{ optimizely_snippet_src }}"></script>
 {% endif %}
 
+{% block pre_app_js %}
+{% endblock %}
```
cpennington, line 77: Please add a comment indicating why you might want to use this block.

ecommerce/static/js/apps/basket_app.js
```diff
@@ -7,6 +7,20 @@ require([
...
+            // TODO: Change to responsive javascript after this site turns into a MFE
+            if (!window.location.href.split('discount_jwt=')[1]) {
+                $.ajax({
+                    url: window.lmsUrlRoot + '/api/discounts/course/' + window.courseKey,
+                    xhrFields: {
+                        withCredentials: true
+                    }
+                }).done(function(discount) {
+                    if (discount.discount_applicable) {
+                        window.location.replace(window.location.href + '?discount_jwt=' + discount.jwt);
```
cpennington: Do we need to worry about this losing any other parameters being passed to this view? If so, you could maybe solve it by checking for the presence of `?` in the current href, and if it exists, using `&` instead? Also, `#` fragments would throw this off.

On the other hand, if we don't need to worry about those things, this should work fine.

cpennington (approved):

ecommerce/templates/edx/base.html
emma-green: Yep

ecommerce/static/js/apps/basket_app.js
emma-green: No reason to not add checks for this..

cpennington (approved):

cpennington:
Don't forget to squash before you merge this

PR #2228, opened 2019-04-30
Add polyfill for Number.inInteger()

zainab-amir:
Number.isInteger() is used in the validation step on basket page. isInteger() is not supported in IE11. To the end user, this results in the validation error not being shown when a non-number is entered in the CVN or expiry fields but the form not being submitted which is quite confusing. 

##### Notes:
* Added a polyfill for isInteger()

[LEARNER-6821](https://openedx.atlassian.net/browse/LEARNER-6821)


cpennington:
I confirmed that the code being copied is licensed CC0.

cpennington (approved):

PR #2231, opened 2019-04-30
ENT-1853 Part 1: Populate enterprise id in a product attribute for the coupon

brittneyexline:
This PR introduces a new attribute on the Coupon Product for storing enterprise id. It also migrates existing coupon products with enterprise ids to set this attribute, and updates the update and create paths related to enterprise id on coupons to set the attribute.

The next part, which will come in a separate PR, will switch the query sets filtering coupons for the regular and enterprise coupon screens to use this field instead of a combination of the condition and the business client. The way the querysets are for the two screens are inconsistent with each other in a way that is leading to this bug. Additionally, we'll remove the business client enterprise id updating logic, which is causing the data condition that makes the query inconsistency lead to some coupons not displaying on either screen.


ziafazal (approved):
👍

PR #2232, opened 2019-05-01
Add history to more models

pwnage101:
Added history to the following models:

Oscar's Offer app:

 - Benefit
 - Condition
 - ConditionalOffer
 - OfferAssignment
 - Range
 - RangeProduct

This PR skips tracking history in any AutoSlugFields.

DE-1424


ecommerce/core/models.py
```diff
@@ -486,6 +486,10 @@ class User(AbstractUser):
 
     tracking_context = JSONField(blank=True, null=True)
 
+    # Adding a history model for this model would conflict with 0006_add_service_user which updates
+    # rows in this model.
+    # history = HistoricalRecords()
```
bmedx: Maybe instead of this add a comment to the model about how migrations on this model may break?

bmedx (approved):
Just one comment. My PR has merged, so you should be able to rebase off master. I had to delete and regenerate all of my migrations as well.

brittneyexline (commented):
just a note that Benefits, Conditions, and Ranges are typically not updated; usually they are immutable and new ones are created, so i don't know the value of having the history tables for these. but if you feel it is necessary, i don't see any problem with it.

pwnage101:
@brittneyexline since the addition of history would have minimal impact (performance, storage, or otherwise) on those append-only tables, we would rather have them put in now and yanked later according to observed behavior if that seems appropriate to you.

bmedx:
We also are adding them because even the rare changes can have big downstream impacts on financial reporting, so even if something only changes very rarely we need to know fairly precisely when and have a way to map back to the old value. This bit us with a retired whitelabel site so we're trying to be extra cautious now. :)

brittneyexline (approved):

PR #2233, opened 2019-05-01
ENT-1853 Part 2: Change querysets for filtering coupons for each admin screen

brittneyexline:
Depends on https://github.com/edx/ecommerce/pull/2231, has to be released separately.

This PR switchES the query sets filtering coupons for the regular and enterprise coupon screens to use the new enterprise id product attribute instead of a combination of the condition and the business client. The way the querysets are for the two screens are inconsistent with each other in a way that is leading to this bug. Additionally, we'll remove the business client enterprise id updating logic, which is causing the data condition that makes the query inconsistency lead to some coupons not displaying on either screen.

### DIFF
--- a/ecommerce/extensions/api/v2/utils.py
+++ b/ecommerce/extensions/api/v2/utils.py
@@ -6,11 +6,9 @@ from django.core.mail import send_mail
 from oscar.core.loading import get_model
 
 from ecommerce.enterprise.utils import get_enterprise_customer
-from ecommerce.invoice.models import Invoice
 
 logger = logging.getLogger(__name__)
 
-Basket = get_model('basket', 'Basket')
 Product = get_model('catalogue', 'Product')
 
 
@@ -54,8 +52,6 @@ def get_enterprise_from_product(product_id):
     """
     try:
         product = Product.objects.get(pk=product_id)
-        basket = Basket.objects.filter(lines__product_id=product.id).first()
-        invoice = Invoice.objects.get(order__basket=basket)
-        return str(invoice.business_client.enterprise_customer_uuid)
-    except (Product.DoesNotExist, Invoice.DoesNotExist):
+        return product.attr.enterprise_customer_uuid
+    except Product.DoesNotExist:
         return None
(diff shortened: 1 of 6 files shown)


ecommerce/extensions/api/v2/utils.py
```diff
@@ -54,8 +52,6 @@ def get_enterprise_from_product(product_id):
     """
     try:
         product = Product.objects.get(pk=product_id)
-        basket = Basket.objects.filter(lines__product_id=product.id).first()
-        invoice = Invoice.objects.get(order__basket=basket)
-        return str(invoice.business_client.enterprise_customer_uuid)
-    except (Product.DoesNotExist, Invoice.DoesNotExist):
+        return product.attr.enterprise_customer_uuid
+    except Product.DoesNotExist:
```
muhammad-ammar, line 56: Awesome...Less code

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -77,19 +76,14 @@ class EnterpriseCouponViewSet(CouponViewSet):
...
-        orders = Order.objects.filter(id__in=[invoice.order_id for invoice in invoices])
-        basket_lines = Line.objects.filter(basket_id__in=[order.basket_id for order in orders])
-        return Product.objects.filter(
-            product_class__name=COUPON_PRODUCT_CLASS_NAME,
-            stockrecords__partner=self.request.site.siteconfiguration.partner,
-            id__in=[line.product_id for line in basket_lines],
-            coupon_vouchers__vouchers__offers__condition__enterprise_customer_uuid__isnull=False,
-        ).distinct()
+            filter_kwargs['attribute_values__value_text'] = enterprise_id
+        return Product.objects.filter(**filter_kwargs).distinct()
```
muhammad-ammar, line 86: More Awesomeness....

muhammad-ammar (approved):
👍 LGTM

ziafazal (commented):
👍

PR #2237, opened 2019-05-06
Fix enterprise coupons tab pagination.

waheedahmed:
Before we were using custom DefaultPagination class from `edx-drf-extensions` which wasn't allowing datatables pagination to work correctly, created a custom pagination class by inheriting `DefaultPagination` and `DatatablesPageNumberPagination`.

### DIFF
--- a/ecommerce/extensions/api/pagination.py
+++ b/ecommerce/extensions/api/pagination.py
@@ -1,9 +1,14 @@
-from rest_framework import pagination
+from edx_rest_framework_extensions.paginators import DefaultPagination
+from rest_framework_datatables.pagination import DatatablesPageNumberPagination
 
 
-class PageNumberPagination(pagination.PageNumberPagination):
+class PageNumberPagination(DatatablesPageNumberPagination):
     page_size_query_param = 'page_size'
 
     # NOTE (CCB): This is a hack, necessary until the frontend
     # can properly follow our paginated lists.
     max_page_size = 10000
+
+
+class DatatablesDefaultPagination(DefaultPagination, PageNumberPagination):
+    pass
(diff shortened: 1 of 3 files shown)


ecommerce/extensions/api/pagination.py
```diff
@@ -1,9 +1,6 @@
...
+from rest_framework_datatables.pagination import DatatablesPageNumberPagination
 
 
-class PageNumberPagination(pagination.PageNumberPagination):
+class PageNumberPagination(DefaultPagination, DatatablesPageNumberPagination):
     page_size_query_param = 'page_size'
-
-    # NOTE (CCB): This is a hack, necessary until the frontend
-    # can properly follow our paginated lists.
-    max_page_size = 10000
```
brittneyexline: should this be deleted?

brittneyexline (commented):
i think it's ok because our tests are passing, but i wanted to note that our use of the DefaultPagination class was introduced for the code management endpoints, which are called from a different place than the enterprise coupon management screen endpoints. unless we separate the two sets of endpoints into different viewset classes, it's a little murky in that the pagination needs are now different for endpoints within the same viewset. It appears that is handled by the `DatatablesPageNumberPagination` class, but i want to call it out as a risk.

ecommerce/extensions/api/pagination.py
waheedahmed: Sorry about that, removed mistakenly.

waheedahmed:
@brittneyexline Your concerns are valid but the `DatatablesPageNumberPagination` comes into play only when we pass `?format=datatables` in query param during API calls otherwise all APIs will perform exactly same as they were working previously. It was done to make our backend compatible with already implemented frontend using datatables.

I have made some changes along with the ENT coupon viewset pagination class to make our APIs exactly same as it was before this change https://github.com/edx/ecommerce/pull/2204.

brittneyexline (approved):

adeelkhan (approved):

PR #2238, opened 2019-05-06
Adding historical Catalog and Core models

bmedx:
Adding more historical models to support financial reports.


ecommerce/core/tests/test_models.py
```diff
@@ -258,6 +258,25 @@ def test_payment_processor_field_parsing(self, payment_processors_str, expected_
         site_config = _make_site_config(payment_processors_str)
         self.assertEqual(site_config.payment_processors_set, expected_result)
 
+    @ddt.data(
```
bmedx, line 261: I ran into some Codecov issues with this PR and added this test in the process of trying to fix it. It has nothing to do with this PR, and as near as I can tell the test target isn't even used (unless Oscar uses it behind the scenes). Even so, it bumps up the coverage a bit so I figured I'd leave it. Happy to remove if reviewers feel otherwise.

ecommerce/extensions/catalogue/migrations/0002_auto_20150223_1052.py
```diff
@@ -19,39 +19,48 @@ def create_catalog(apps, schema_editor):
     ProductAttribute = apps.get_model("catalogue", "ProductAttribute")
     ProductClass = apps.get_model("catalogue", "ProductClass")
 
+    for klass in (Category, ProductAttribute, ProductClass):
```
bmedx: The various migration changes are to support adding django-simple-history to the impacted models. Because the migrations are doing inserts before migrations are complete, the `post_save` history handler gets triggered before the new history tables are created, which causes the migrations to fail. These change just turn off history during migrations to allow them to succeed.

ecommerce/extensions/catalogue/models.py
```diff
@@ -17,6 +25,20 @@
...
 
+class CreateSafeHistoricalRecords(HistoricalRecords):
+    """
+    Overrides default HistoricalRecords so that newly created rows can avoid being saved to history.
+
+    This prevents errors during migrations with models that have HistoricalRecords and also
+    try to create rows as part of their migrations (before the history table is created).
+    """
+    def post_save(self, instance, created, using=None, **kwargs):
+        if hasattr(instance, "skip_history_when_saving"):
```
bmedx, line 41: The only difference between this method and the original is this line, which omits a check to see if the object is newly created:
https://github.com/treyhunner/django-simple-history/blob/2.7.2/simple_history/models.py#L456

ecommerce/extensions/catalogue/migrations/0002_auto_20150223_1052.py
```diff
@@ -20,38 +20,44 @@ def create_catalog(apps, schema_editor):
     ProductClass = apps.get_model("catalogue", "ProductClass")
 
     # Create a new product class for course seats
-    seat = ProductClass.objects.create(
+    seat = ProductClass(
```
bmedx, line 23: These various migration changes are to support adding django-simple-history to the impacted models. Because the migrations are doing inserts / updates / deletes before migrations are complete, the `post_save` history handler gets triggered before the history tables are created, which causes the migrations to fail. These change just turn off history during migrations to allow them to succeed.

bessiesteinberg: thanks for explaining that!

ecommerce/core/tests/test_models.py
doctoryes: If it passes and raises coverage, then 👍

ecommerce/extensions/catalogue/migrations/0036_coupon_notify_email_attribute.py
```diff
@@ -13,18 +13,21 @@
 def create_notify_email_attribute(apps, schema_editor):
     """Create coupon notify_email attribute."""
     coupon = ProductClass.objects.get(name=COUPON_PRODUCT_CLASS_NAME)
-    ProductAttribute.objects.create(
+    pa = ProductAttribute.objects.create(
```
doctoryes: Shouldn't this line be `pa = ProductAttribute(` instead?

doctoryes: On that note, will you be testing a forward/backward progression/regression of these migration changes?

bmedx: This should be fixed. I did run migrations forward and back (as far as they would go)

doctoryes (approved):

bessiesteinberg (approved):

bmedx:
Thanks @BessieSteinberg can we get a merge and release at your convenience?

ecommerce/extensions/catalogue/models.py
pwnage101: In other words, deletes and updates can't have history skipped in vanilla django-simple-history?

pwnage101 (approved):
It is alarming to me that so many of the migrations insert/update/delete rows in the database, even as recent as 13 days ago.  Is this required as a workaround for some reason?

ecommerce/extensions/catalogue/models.py
bmedx: Other way around. The original code explicitly excludes inserts, I removed that.

bmedx:
@pwnage101 I can't speak to the reasons behind it, though it is a known Django anti-pattern. Maybe @BessieSteinberg or someone familiar with Ecomm history could explain. If it helps I don't think new data-touching migrations would need these types of changes since the history tables will actually exist before they get run.

bmedx:
@BessieSteinberg any idea when you might get to this?

PR #2239, opened 2019-05-09
clean up rbac authorization related waffle switches and logic

muhammad-ammar:
**Description:** Removed all the role based accessed control switch OFF logic as now switch is permanently enabled. This PR only removes the code usages of switch.

**JIRA:** [ENT-1885](https://openedx.atlassian.net/browse/ENT-1885)


ecommerce/extensions/api/permissions.py
```diff
@@ -96,8 +94,6 @@ def has_permission(self, request, view):
         """
         Verify the user is staff or the associated enterprise matches the requested enterprise.
         """
-        if waffle.switch_is_active(USE_ROLE_BASED_ACCESS_CONTROL):
```
brittneyexline, line 99: i think instead of doing this, which would effectively re-enable the old logic, we should not be using this permission class at all anymore, meaning this whole class should be deleted and not used.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -288,7 +291,6 @@ def test_create_ent_offers_switch_on(self):
         enterprise_customer_id = self.data['enterprise_customer']['id']
         enterprise_name = self.data['enterprise_customer']['name']
         enterprise_catalog_id = self.data['enterprise_customer_catalog']
-        self.assertEqual(coupon.attr.enterprise_customer_uuid, enterprise_customer_id)
```
brittneyexline: we need to add this line back in. i didn't realize that to test rbac, we copy pasted the all the tests, so this is not up to date in the rbac-based tests.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -366,15 +368,13 @@ def test_update_migrated_ent_coupon(self):
             'PUT',
             reverse('api:v2:enterprise-coupons-detail', kwargs={'pk': coupon.id}),
             data={
-                'enterprise_customer': self.data['enterprise_customer'],
```
brittneyexline: same here, add this back into the test.

ecommerce/extensions/api/v2/tests/views/test_enterprise.py
```diff
@@ -366,15 +368,13 @@ def test_update_migrated_ent_coupon(self):
...
                 'enterprise_customer_catalog': new_catalog,
                 'benefit_value': 50,
                 'title': 'Updated Enterprise Coupon',
             }
         )
         updated_coupon = Product.objects.get(title='Updated Enterprise Coupon')
         self.assertEqual(coupon.id, updated_coupon.id)
-        self.assertEqual(updated_coupon.attr.enterprise_customer_uuid, self.data['enterprise_customer']['id'])
```
brittneyexline: same here, add this back into the test.

brittneyexline (approved):
nice work @muhammad-ammar - i will merge/release this on monday.

PR #2245, opened 2019-05-13
Raise the default per-user throttle rate to reduce the instances of t…

cpennington:
…he drupal service user hitting it in service of anonymous users on the marketing site

This should help https://openedx.atlassian.net/browse/REVMI-185


MatthewPiatetsky (approved):
assuming we don't know if this value will solve the issue but we're just trying it out?

cpennington:
Yeah, basically.

PR #2248, opened 2019-05-14
Adding more thorough logging of VoucherAddView

brittneyexline:
This PR adds more robust logging to the user-facing error paths related to users redeeming codes, which occur in VoucherAddView and CouponRedeemView.

Each message is prefixed with `[Code Redemption Failure]` so that they can be easily filtered in splunk. Each message also includes relevant metadata, if available, in a standard order that can also be parsed by a splunk query. These fields are, in order: User, Offer, Product, Basket, Code, Message, Enterprise, Catalog, Courses, Range.

### DIFF
--- a/ecommerce/coupons/views.py
+++ b/ecommerce/coupons/views.py
@@ -144,7 +144,7 @@ class CouponRedeemView(EdxOrderPlacementMixin, View):
 
     @method_decorator(set_enterprise_cookie)
     @method_decorator(login_required)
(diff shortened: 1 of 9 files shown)


christopappas (approved):
LGTM once tests pass 👍

mushtaqak (approved):
Once build is green 👍

muhammad-ammar (approved):
👍 once build is green

PR #2250, opened 2019-05-15
Cleanup enterprise offer switches

mushtaqak:
[ENT-1713](https://openedx.atlassian.net/browse/ENT-1713)

### DIFF
--- a/ecommerce/enterprise/conditions.py
+++ b/ecommerce/enterprise/conditions.py
@@ -3,16 +3,13 @@ from __future__ import unicode_literals
 import logging
 from uuid import UUID
 
-import waffle
 from oscar.core.loading import get_model
 from requests.exceptions import ConnectionError, Timeout
 from slumber.exceptions import SlumberHttpBaseException
 
 from ecommerce.enterprise.api import catalog_contains_course_runs, fetch_enterprise_learner_data
-from ecommerce.enterprise.constants import ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH, ENTERPRISE_OFFERS_SWITCH
 from ecommerce.extensions.basket.utils import ENTERPRISE_CATALOG_ATTRIBUTE_TYPE
 from ecommerce.extensions.offer.constants import OFFER_ASSIGNMENT_REVOKED, OFFER_REDEEMED
-from ecommerce.extensions.offer.decorators import check_condition_applicability
 from ecommerce.extensions.offer.mixins import ConditionWithoutRangeMixin, SingleItemConsumptionConditionMixin
 
 BasketAttribute = get_model('basket', 'BasketAttribute')
@@ -33,7 +30,6 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
     def name(self):
         return "Basket contains a seat from {}'s catalog".format(self.enterprise_customer_name)
 
-    @check_condition_applicability([ENTERPRISE_OFFERS_SWITCH])
     def is_satisfied(self, offer, basket):  # pylint: disable=unused-argument
         """
         Determines if a user is eligible for an enterprise customer offer
@@ -55,11 +51,6 @@ class EnterpriseCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumpt
         """
         if not basket.owner:
             # An anonymous user is never linked to any EnterpriseCustomer.
-            return False
-
-        if (offer.offer_type == ConditionalOffer.VOUCHER and
-                not waffle.switch_is_active(ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH)):
-            logger.info('Skipping Voucher type enterprise conditional offer until we are ready to support it.')
             return False
 
         enterprise_customer = str(self.enterprise_customer_uuid)
--- a/ecommerce/extensions/api/v2/tests/views/test_coupons.py
+++ b/ecommerce/extensions/api/v2/tests/views/test_coupons.py
@@ -18,12 +18,10 @@ from oscar.core.loading import get_model
 from oscar.test import factories
 from rest_framework import status
 from testfixtures import LogCapture
-from waffle.models import Switch
 
 from ecommerce.coupons.tests.mixins import CouponMixin, DiscoveryMockMixin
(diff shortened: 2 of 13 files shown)


ecommerce/extensions/offer/models.py
```diff
@@ -281,11 +279,6 @@ def is_condition_satisfied(self, basket):
                            'User: %s, Offer: %s, Basket: %s', basket.owner.username, self.id, basket.id)
             return False
 
-        if (self.benefit.range and self.benefit.range.enterprise_customer and
```
mushtaqak, line 284: @brittneyexline Do you think we should keep this if condition; I am not able to correctly understand this code?

mushtaqak:
@brittneyexline @muhammad-ammar @christopappas This is ready for initial review :)

ecommerce/enterprise/conditions.py
```diff
@@ -33,7 +31,7 @@ class Meta(object):
     def name(self):
         return "Basket contains a seat from {}'s catalog".format(self.enterprise_customer_name)
 
-    @check_condition_applicability([ENTERPRISE_OFFERS_SWITCH])
+    @check_condition_applicability()
```
brittneyexline: i would delete this decorator entirely

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -464,122 +462,16 @@ def _check_enterprise_fields(
...
-        response = self.client.get(COUPONS_LINK)
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-        coupon_data = json.loads(response.content)['results']
-        self.assertEqual(len(coupon_data), 2)
-        self.assertEqual(coupon_data[0]['title'], 'Test Create Enterprise Coupon')
-        self.assertEqual(coupon_data[0]['client'], 'test enterprise')
-        self.assertEqual(coupon_data[1]['title'], original_title)
-        self.assertEqual(coupon_data[1]['client'], original_client)
-
-        # Now with the switch on, see that the enterprise coupon gets filtered out
```
brittneyexline, line 489: we should maintain this part of the test, since it has logic for testing with the switch on

ecommerce/extensions/api/v2/tests/views/test_coupons.py
```diff
@@ -594,25 +486,6 @@ def test_update_enterprise_offers_regular_coupon_switch_on(self):
         self.assertEqual(new_coupon.attr.enterprise_customer_uuid, enterprise_customer_id)
         self._check_enterprise_fields(new_coupon, enterprise_customer_id, enterprise_catalog_id, 'test enterprise')
 
-    def test_update_enterprise_offers_enterprise_coupon_switch_on(self):
-        """Test updating an enterprise coupon with the enterprise offers switch on."""
-        Switch.objects.update_or_create(name=ENTERPRISE_OFFERS_FOR_COUPONS_SWITCH, defaults={'active': False})
-        enterprise_customer_id = str(uuid4()).decode('utf-8')
-        enterprise_catalog_id = str(uuid4()).decode('utf-8')
-        enterprise_name = 'test enterprise'
-        self._create_enterprise_coupon(enterprise_customer_id, enterprise_catalog_id, enterprise_name)
```
brittneyexline, line 603: i would keep this test and update this function to use the `ENTERPRISE_COUPONS_LINK` to create an enterprise coupon. this way, we still test that this endpoint does not support updating enterprise coupons.

ecommerce/extensions/offer/models.py
brittneyexline: we need to keep this if statement, and remove the part that checks the waffle switch. this code skips the old offer data so that it does not get evaluated.

mushtaqak: @brittneyexline when I added `if self.benefit.range and self.benefit.range.enterprise_customer` condition, `test_enterprise_customer_successful_redemption_message` in `ecommerce/coupons/tests/test_views.py` fails because [is_voucher_applied(basket, voucher)](https://github.com/edx/ecommerce/blob/b1ce6ef6f151486bab6286ea4b3d45e3c1a071aa/ecommerce/coupons/views.py#L266) returns false. After some debugging, found out that debuggers hits  this method (`is_condtion_satisified`) twice, where `benifit.range` is not empty and has an enterprise_customer, on second hit, range is undefined. See screenshot below:

Screenshot 2019-05-23 at 6 16 12 PM

I don't know why this is happening and failing now, since switch was turned on previously as well. @brittneyexline Do you think we need to do something more to update range in [test_views.py](https://github.com/edx/ecommerce/blob/b1ce6ef6f151486bab6286ea4b3d45e3c1a071aa/ecommerce/coupons/tests/test_views.py#L591) or something

brittneyexline: you'll see that the test doesn't turn the switch on, so this test was written to work with the switch off. so the fact that it is breaking does make sense, and we should update the test accordingly.

if i had to guess, the root problem to fix is that the enterprise coupon data it is creating here (https://github.com/edx/ecommerce/blob/b1ce6ef6f151486bab6286ea4b3d45e3c1a071aa/ecommerce/coupons/tests/test_views.py#L586) isn't creating the new enterprise data using enterprise offers properly.

mushtaqak:
@irfanuddinahmad @muhammad-ammar Can you also take a look :)

brittneyexline (approved):

brittneyexline:
@mushtaqak - can you write down the testing instructions you went through, and also squash your commits before i merge?

mushtaqak:
@brittneyexline  I did some testing around code management and saw that things were working fine.

mushtaqak:
@brittneyexline  I did some testing around code management and saw that things were working fine.

mushtaqak:
@brittneyexline Just saw that you had pushed a commit https://github.com/edx/ecommerce/commit/a59be2700ff10e47b927dd8d5f74999a57f8af79 on my branch, but I rebased my local changes. Can you push your changes again? sorry to bother you.

brittneyexline:
@mushtaqak - looks like the tests are failing after updating your branch...

mushtaqak:
@brittneyexline this is ready to merge!

PR #2254, opened 2019-05-22
Disable apply voucher button during submit.

waheedahmed:
Stop concurrent requests by disabling submit button during submission.

PROD-301

### DIFF
--- a/ecommerce/static/js/pages/basket_page.js
+++ b/ecommerce/static/js/pages/basket_page.js
@@ -399,6 +399,12 @@ define([
                     BasketPage.hideVoucherForm();
                 });
 
+                $('#voucher_form').on('submit', function() {
+                    $('#apply-voucher-button').attr('disabled', true);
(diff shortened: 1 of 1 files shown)


cpennington:
Does the voucher button need to be re-enabled when the call to apply it returns?

Also, it would be great to disable the payment submit button while applying a voucher as well, not having that has caused some order discrepancies recently. Is that something you'd be willing to add on to this PR?

waheedahmed:
@cpennington No, it doesn't need to be re enabled because whole page reloaded again upon this request which automatically enables the apply button.

Sure, I'll add the disable functionality for payment button as well.

robrap:
Note that there are two ways to pay, including PayPal, and both would need to be disabled.

waheedahmed:
@cpennington @robrap I have updated the PR kindly take another look.

cpennington (approved):

adeelkhan (approved):

PR #2256, opened 2019-05-22
todo for clean-up of LMS user_id from JWT

robrap:
Adds a todo comment for later.

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -511,6 +511,10 @@ class User(AbstractUser):
         Returns the LMS user_id, or None if not found.
         """
         # JWT cookie is used with API calls from new microfrontends. This is not persisted.
+        # TODO: Rename ``_get_lms_user_id_from_jwt_cookie`` to ``_get_lms_user_id_from_jwt``
+        #   and update to use new method to be added to JwtAuthentication in edx-drf-extensions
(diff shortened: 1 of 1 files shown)


ecommerce/core/models.py
```diff
@@ -511,6 +511,13 @@ def lms_user_id(self):
         Returns the LMS user_id, or None if not found.
         """
         # JWT cookie is used with API calls from new microfrontends. This is not persisted.
+        # TODO: Rename ``_get_lms_user_id_from_jwt_cookie`` to ``_get_lms_user_id_from_jwt`` 
+        #   and update to use new method to be added to JwtAuthentication in edx-drf-extensions
+        #   to get the decoded JWT used for authentication, no matter where it came from.
+        #   See https://github.com/edx/edx-drf-extensions/pull/69#discussion_r286618922
+        #   Additionally, this method could also replace
```
crice100: Could you clarify this last sentence?

robrap: After remembering the details of this method, I would just delete this sentence. :) . I think that method should stay as-is until we decide it is no longer needed. :)

robrap: I deleted this part of the message.

crice100 (approved):

PR #2257, opened 2019-05-23
Adds management command to fix Cybersource payment response failure.

adeelkhan:
This command would check Cybersource for payment responses for all
frozen baskets and would create orders for successfull transactions.

LEARNER-7184


ecommerce/core/management/commands/find_frozen_baskets_missing_payment_response.py
```diff
@@ -0,0 +1,483 @@
...
+
+        hashobj = hashlib.sha256()
+        hashobj.update(message_body.encode('utf-8'))
+        hash_data = hashobj.digest()
+        hashdigest = base64.b64encode(hash_data)
+
+        self.digest = 'SHA-256=' + hashdigest.decode("utf-8")
+
+    def generate_post_signature(self):
+        signature_header = \
```
waheedahmed: nit: Can we break it in following way?
```
signature_header = "host: {host}\ndate: {date}\n(request-target): post /tss/v2/searches\ndigest: " \
                   "{digest}\nv-c-merchant-id: {merchant_id}"
```

ecommerce/core/management/commands/find_frozen_baskets_missing_payment_response.py
```diff
@@ -0,0 +1,483 @@
...
+        "host: {host}\ndate: {date}\n(request-target): post /tss/v2/searches\ndigest: {digest}\nv-c-merchant-id: {merchant_id}"
```
waheedahmed: Can we put this endpoint `/tss/v2/searches` in configuration instead of hardcoding?

ecommerce/core/management/commands/find_frozen_baskets_missing_payment_response.py
```diff
@@ -0,0 +1,483 @@
...
+            'Signature': 'keyid="{keyid}", '
+                         'algorithm="HmacSHA256", headers="host date (request-target) '
+                         'digest v-c-merchant-id", '
+                         'signature="{signature}"'.format(signature=self.post_signature, keyid=self.key)
+        }
+        return headers
+
+    def generate_get_signature(self, path):
+        signature_header = \
+        "host: {host}\ndate: {date}\n(request-target): get {path}\nv-c-merchant-id: {merchant_id}"
```
waheedahmed: Nit: I believe we can move this to single line.

ecommerce/core/management/commands/find_frozen_baskets_missing_payment_response.py
```diff
@@ -0,0 +1,483 @@
...
+
+
+def check_cs_search_transaction(client, basket, message_body):
+    """
+    Hit Cybersource transaction search api to get meta details
+    regarding a transaction.
+    """
+
+    try:
+        response = requests.post('https://{host}/tss/v2/searches'.format(host=client.host),
```
waheedahmed: Same here move this `/tss/v2/searches` into configuration if we can.

ecommerce/core/management/commands/find_frozen_baskets_missing_payment_response.py
```diff
@@ -0,0 +1,483 @@
...
+        except InvalidTimeRange:
+            logger.exception(u"Incorrect time range given.")
+            return
+
+        start = datetime.now(pytz.utc) - timedelta(hours=start_delta)
+        end = datetime.now(pytz.utc) - timedelta(hours=end_delta)
+
+        self.find_frozen_baskets_missing_payment_response(start, end)
+
+    def find_frozen_baskets_missing_payment_response(self, start, end):
```
waheedahmed: nit: Rename it to `find_frozen_baskets_and_fulfill`.

ecommerce/core/management/commands/find_frozen_baskets_missing_payment_response.py
```diff
@@ -0,0 +1,483 @@
+""" Django management command to find frozen baskets missing payment response. """
```
waheedahmed: Rename this file to `find_frozen_baskets_and_fulfill`.

waheedahmed (commented):
Can we also move all helper functions to a separate file?

First pass is complete let me know when you're done with the requested changes.

ecommerce/core/management/commands/find_frozen_baskets_missing_payment_response.py
```diff
@@ -0,0 +1,483 @@
...
+            basket.submit()
+
+            logger.info(u"Order Number: {order_number} created successfully."
+                        .format(order_number=basket.order_number))
+
+            logger.info(u'Requesting fulfillment of order [%s].', order.number)
+            fulfill_order.delay(
+                order.number,
+                site_code=order.site.siteconfiguration.partner.short_code,
+                email_opt_in=False
```
waheedahmed: I believe we should check if user's email preferences are set in basket attributes before defaulting to `False`: https://github.com/edx/ecommerce/blob/fb8ef4c196d1eecc959a68e350715bd7a7e86ca7/ecommerce/extensions/checkout/mixins.py#L163-L170

ecommerce/core/management/commands/find_frozen_baskets_and_fulfill.py
```diff
@@ -0,0 +1,484 @@
...
+
+        It checks for date format and range validity and then
+        calls find_frozen_baskets_missing_payment_response for
+        the given date range
+        """
+        start_delta = options['start_delta']
+        end_delta = options['end_delta']
+
+        try:
+            if end_delta > start_delta:
```
waheedahmed, line 230: Mostly we are going to run this command on default values or maybe hardcoding it in jenkins jobs if we are changing them, I don't think this check is necessary here. And if someone gives a wrong value the command will not fetch any data from DB, thoughts?

ecommerce/core/management/commands/find_frozen_baskets_and_fulfill.py
```diff
@@ -0,0 +1,484 @@
...
+        cs_config = settings.CS_API_CONFIG
+
+        self.key = cs_config.get('api_key_id', None)
+        self.secret_key = cs_config.get('api_secret_key', None)
+        self.host = cs_config.get('host', None)
+        self.merchant_id = cs_config.get('merchant_id', None)
+        self.search_uri = cs_config.get('search_uri', None)
+
+
+class CybersourceAPIClient(object):
```
waheedahmed, line 55: Can we use https://github.com/CyberSource/cybersource-rest-client-python instead of making our own, eventually we are moving to Python 3?

ecommerce/settings/base.py
```diff
@@ -660,6 +660,16 @@
 
 # END ENTERPRISE APP CONFIGURATION
 
+# CYBERSOURCE API CONFIGURATION
+CS_API_CONFIG = {
+    'host': os.environ.get('CYBSERSOURCE_API_ROOT', None),
```
waheedahmed, line 665: Instead of getting single value from configuration, can we get whole config dict like we do for payment processors: https://github.com/edx/ecommerce/blob/edf7b378b86f62875772ce92c4d2b40378ee7739/ecommerce/settings/base.py#L744-L777

ecommerce/settings/test.py
```diff
@@ -172,3 +172,11 @@
...
 SAILTHRU_KEY = 'abc123'
 SAILTHRU_SECRET = 'top_secret'
+
+CS_API_CONFIG={
+    'host': 'apitest.cybersource.com',
+    'search_uri': '/tss/v2/searches',
+    'merchant_id': 'test_merchant',
+    'api_key_id': '11abcd11-d11e-123c-test-111a11a1d1f1',
+    'api_secret_key': 'AAAAAAA+2222222222222/nx1+Obf/BBBBBBBBBBBBB='
+}
```
waheedahmed, line 182: Add a new line.

waheedahmed (commented):
My main concern here is about writing our own API client instead of using already provided by CyberSourse https://github.com/CyberSource/cybersource-rest-client-python.

ecommerce/core/management/commands/find_frozen_baskets_and_fulfill.py
adeelkhan: This being already discussed way back in May on the ticket. I had tested their api but it requires python 3.4+ and in May we had little clue that we would be moving code base to python 3. For now we are not sure of the timelines of when the transition to python 3 would be complete. We have already spent considerable time in implementation and i find no reasons to block this. The current implementation is such that we can easily move existing code to cybersource api.

ecommerce/core/management/commands/find_frozen_baskets_and_fulfill.py
adeelkhan: Its part of general defensive programming and is a good thing to sanitize functions inputs before doing anything. Yes in this case a user might never invoke this him/herself but i think an params checking should be context agnostic.  I don't think this would put any extra performance cost too.

PR #2258, opened 2019-05-23
REVMI-197 Add LMS user id to ecommerce_user table

crice100:
This PR adds a new column for the LMS user id. Future PRs will save data to and read from this column.

Suggestions on alternate names/locations welcome.

Let me know if _IntegerField_ doesn't seem like the correct type.


ecommerce/core/models.py
```diff
@@ -491,6 +491,11 @@ class User(AbstractUser):
     """
 
     full_name = models.CharField(_('Full Name'), max_length=255, blank=True, null=True)
+    user_id = models.IntegerField(
```
cpennington: For clarity, I think I'd prefer `lms_user_id` (to keep it distinct from `user.id`, which is the id of the user in ecommerce)

ecommerce/core/models.py
```diff
@@ -511,7 +510,7 @@ def access_token(self):
             return None
 
     @property
-    def lms_user_id(self):
```
cpennington, line 509: Hrm... I don't like calling a property `get_foo`. I'd prefer `get_foo()` (that is, not a property anymore). Or, we could call it `lms_user_id_from_request`. Or, we could make the field name be something else, and leave this as is.

Separately, do you think that this method should return the value of `lms_user_id` first, and then fall back to pulling the data out of the request? I haven't looked all the contexts it's used from.

crice100: Yeah, I can see the confusion. I renamed the property and added a TODO.

I'd like to keep this PR small and have it just add the new db column. After this PR is merged successfully, I'll come back and add code to start saving the id to the db as well as read the id from the db in this method.

cpennington:
Maybe it would be better to modify the management command so that it dumps the timestamp of the last processed record into memcache, and then to use that in preference to the commandline argument, if it exists? That would prevent double counting, while also removing the risk of missing entries unless we lost the data from memcache and went more than 3 hours between instances of the command running.

cpennington (approved):

PR #2261, opened 2019-05-24
ARCH-752: Create WIP basket api endpoint.

robrap:
Note: there are follow-up tickets to flesh this out.  This does need to be ready for Production though.

ARCH-752

Reviewers: because we don't want to chance breaking the existing view, and tests are failing, we should probably start the api in the mock story.  In this story, we will also want to remove some processing that never gets used, like the payment provider list.

### DIFF
--- a/ecommerce/extensions/api/v2/urls.py
+++ b/ecommerce/extensions/api/v2/urls.py
@@ -22,6 +22,7 @@ from ecommerce.extensions.api.v2.views import sdn as sdn_views
 from ecommerce.extensions.api.v2.views import stockrecords as stockrecords_views
 from ecommerce.extensions.api.v2.views import user_management as user_management_views
 from ecommerce.extensions.api.v2.views import vouchers as voucher_views
+from ecommerce.extensions.basket.views import WIPBasketApiView
 from ecommerce.extensions.voucher.views import CouponReportCSVView
 
 ORDER_NUMBER_PATTERN = r'(?P<number>[-\w]+)'
@@ -33,6 +34,8 @@ USERNAME_PATTERN = r'(?P<username>{regex})'.format(regex=USERNAME_REGEX_PARTIAL)
 
 BASKET_URLS = [
     url(r'^$', basket_views.BasketCreateView.as_view(), name='create'),
+    # TODO: ARCH-867: See ticket for details of final clean-up.
+    url(r'^wip-api/', WIPBasketApiView.as_view(), name='baskets'),
     url(
         r'^{basket_id}/$'.format(basket_id=BASKET_ID_PATTERN),
         basket_views.BasketDestroyView.as_view(),
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -16,7 +16,10 @@ from opaque_keys.edx.keys import CourseKey
 from oscar.apps.basket.views import VoucherAddView as BaseVoucherAddView
 from oscar.apps.basket.views import VoucherRemoveView as BaseVoucherRemoveView
 from oscar.apps.basket.views import *  # pylint: disable=wildcard-import, unused-wildcard-import
+from oscar.core.prices import Price
 from requests.exceptions import ConnectionError, Timeout
+from rest_framework.response import Response
+from rest_framework.views import APIView
 from slumber.exceptions import SlumberBaseException
 
 from ecommerce.core.exceptions import SiteConfigurationError
@@ -117,9 +120,9 @@ class BasketAddItemsView(View):
         return HttpResponseRedirect(url, status=303)
 
 
-class BasketSummaryView(BasketView):
+class BasketLogicMixin(object):
     """
-    Display basket contents and checkout/payment options.
+    Business logic for determining basket contents and checkout/payment options.
     """
 
     @newrelic.agent.function_trace()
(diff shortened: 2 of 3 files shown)


ecommerce/extensions/basket/views.py
```diff
@@ -442,6 +420,94 @@ def get_context_data(self, **kwargs):
...
+        See https://github.com/django-oscar/django-oscar/blob/1.5.4/src/oscar/apps/basket/views.py#L92-L132
+        """
+        context = {}
+        shipping_charge = Price('USD', Decimal(0))
+        context['order_total'] = OrderTotalCalculator().calculate(self.request.basket, shipping_charge)
+        return context
+
+    def get(self, request):  # pylint: disable=unused-argument
+        context = self._get_initial_context()
+        context = self.get_basket_context_data(context)
```
robrap, line 496: @nasthagiri: Do you want to help me come up with better names for these two functions?  
1. `_get_initial_context` is basically the replacement for `super(BasketSummaryView, self).get_context_data(**kwargs)`.
2. `.get_basket_context_data` now comes from the mixin, and is the replacement for what was `get_context_data` on the view.

I don't like the current names, but I was getting caught up on them and it would be helpful to brainstorm for a minute.

robrap: After discussion, plan is to create context here, and change `_get_initial_context` to `_get_order_total`.

ecommerce/extensions/basket/views.py
```diff
@@ -442,6 +420,94 @@ def get_context_data(self, **kwargs):
...
+
+            properties = {
+                'checkout_id': basket.order_number,
+                'step': 1
+            }
+            track_segment_event(request.site, request.user, 'Checkout Step Viewed', properties)
+        except Exception:  # pylint: disable=broad-except
+            logger.exception('Failed to fire Cart Viewed event for basket [%d]', basket.id)
+
+        if has_enterprise_offer(basket) and basket.total_incl_tax == Decimal(0):
```
nasthagiri, line 448: Will this line of business logic eventually be moved into the Mixin?

ecommerce/extensions/api/v2/urls.py
```diff
@@ -33,6 +34,10 @@
 
 BASKET_URLS = [
     url(r'^$', basket_views.BasketCreateView.as_view(), name='create'),
+    # TODO: This wip-api is a WIP. It will likely need the following changes:
+    # 1. It should be rolled into the existing /baskets endpoint, providing the GET
+    # 2. Although the user should only have a single basket, possibly provide as a list of one.
```
nasthagiri: Nit: Do you want to maintain 2 separate TODO lists or have this TODO refer to the main TODO in the API view?

ecommerce/extensions/basket/views.py
```diff
@@ -442,6 +420,94 @@ def get_context_data(self, **kwargs):
...
+        context = self.get_basket_context_data(context)
+
+        # TODO: Properly serialize Price
+        # - django-oscar-api has serializer for Price
+        # - from oscarapi.serializers.fields import TaxIncludedDecimalField
+        context['order_total'] = {
+            'currency': context['order_total'].currency,
+            'excl_tax': context['order_total'].excl_tax,
+            'incl_tax': context['order_total'].incl_tax,
+        }
```
nasthagiri: Nit: consider moving this serialization logic into a sub-helper method.

ecommerce/extensions/basket/views.py
```diff
@@ -442,6 +420,94 @@ def get_context_data(self, **kwargs):
...
+        # - If so, properly serialize.
+        # - If not, avoid computation or leave as-is for now?
+        # Remove objects that fail default serialization
+        # TypeError: <class 'ecommerce.extensions.payment.processors.cybersource.Cybersource'> is not JSON serializable
+        context.pop('payment_processors', None)
+        context.pop('client_side_payment_processor', None)
+        # TypeError: <PaymentForm bound=False, valid=Unknown,
+        #   fields=(basket;first_name;last_name;address_line1;address_line2;city;state;postal_code;country)>
+        #   is not JSON serializable
+        context.pop('payment_form', None)
```
nasthagiri: Will these fields that are `pop`ed eventually be extracted out of the common Mixin?

nasthagiri (approved):

ecommerce/extensions/basket/views.py
robrap: Yes.  This was one of the follow-up tickets that was needed/documented in the ticket.  Thanks.

ecommerce/extensions/basket/views.py
robrap: Yes.

ecommerce/extensions/basket/views.py
robrap: Agreed.  Lots of clean-up needed.  I'm going to land this and give us a base to iterate on.

ecommerce/extensions/api/v2/urls.py
robrap: Sorry.  I know this isn't very DRY.  It's just a back-up to help the reviewer as I finish out ticketing.  Ultimately, I'll try to update these to refer to an appropriate ticket (or tickets).  Thanks!

robrap: FYI: I updated this comment before squashing.

ecommerce/extensions/basket/views.py
robrap: Added ARCH-865.

ecommerce/extensions/basket/views.py
robrap: This change is coming.

ecommerce/extensions/basket/views.py
robrap: I added a note to ARCH-687, updated the comment, and switched away from the pop.

PR #2276, opened 2019-06-03
Move ecommerce defaults from configration repo

syedimranhassan:


ecommerce/settings/base.py
```diff
@@ -693,7 +696,52 @@
...
+        'paypal': {
+            'cancel_url': '/checkout/cancel-checkout/',
+            'client_id': 'SET-ME-PLEASE',
+            'client_secret': 'SET-ME-PLEASE',
+            'error_url': '/checkout/error/',
+            'mode': 'sandbox',
+            'receipt_url': '/checkout/receipt/'
+        }
+    }
+}
```
coryleeio, line 762: You need to include all these values as well:
https://github.com/edx/configuration/blob/master/playbooks/roles/edx_django_service/defaults/main.yml#L182-L222

as they get merged into the edx_django_service_config here:
https://github.com/edx/configuration/blob/master/playbooks/roles/edx_django_service/defaults/main.yml#L226

coryleeio (changes requested):

coryleeio (approved):

PR #2278, opened 2019-06-04
REVMI-196 Save lms_user_id to db

crice100:
Update the existing middleware, which is already updating the user, to also save the _lms_user_id_. Also update the _JWT_PAYLOAD_USER_ATTRIBUTE_MAPPING_ to start saving the _lms_user_id_, if it is present in the JWT.

A future PR start reading the _lms_user_id_ from this column.

Note: I removed the comment in _models.py_ because that ticket has been closed.

### DIFF
--- a/ecommerce/extensions/analytics/middleware.py
+++ b/ecommerce/extensions/analytics/middleware.py
@@ -1,23 +1,58 @@
 """
-Middleware for analytics app to parse GA cookie.
+Middleware for analytics app to parse Google Analytics (GA) cookie and the lms_user_id.
 """
-
+import logging
 from ecommerce.extensions.analytics.utils import get_google_analytics_client_id
+
+logger = logging.getLogger(__name__)
 
 
 class TrackingMiddleware(object):
     """
-    Middleware that parse `_ga` cookie and save/update in user tracking context.
+    Middleware that parse `_ga` cookie for the user tracking context to obtain the GA client id, extracts the
+    lms_user_id, and updates the user if necessary.
     """
 
     def process_request(self, request):
         user = request.user
         if user.is_authenticated():
+            save_user = False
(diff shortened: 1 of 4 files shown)


ecommerce/extensions/analytics/middleware.py
```diff
@@ -1,23 +1,60 @@
...
+    def _get_lms_user_id_from_social_auth(self, user):
+        """
+        Return LMS user_id passed through social auth, if found.
+        Returns None if not found.
+        """
+        lms_user_id_social_auth = None
+        try:
+            lms_user_id_social_auth = user.social_auth.first().extra_data[u'user_id']  # pylint: disable=no-member
+        except Exception:  # pylint: disable=broad-except
+            pass
```
cpennington, line 63: It would be good to capture the set of exceptions you know are going to happen here, I think. Off the top of my head, one exception is that `user_id` isn't in `extra_data`... I think that would be better handled by using `extra_data.get('user_id')`. I could also see `social_auth` not being set on `user`. You could catch that `AttributeError`, or check using `hasattr`. I suspect `first()` will always work, if `social_auth` is set on `user`.

cpennington (approved):

ecommerce/extensions/analytics/middleware.py
```diff
@@ -1,23 +1,60 @@
...
             old_client_id = tracking_context.get('ga_client_id')
             ga_client_id = get_google_analytics_client_id(request)
-
             if ga_client_id and ga_client_id != old_client_id:
                 tracking_context['ga_client_id'] = ga_client_id
                 user.tracking_context = tracking_context
+                save_user = True
+
+            if not user.lms_user_id:
+                # Check for the lms_user_id in social auth
```
crice100, line 36: Add comment: could also come from jwt

PR #2280, opened 2019-06-04
REVMI-157: Add a management command to import LMS user ids from a temporary table

cpennington:


ecommerce/core/management/commands/import_user_ids.py
```diff
@@ -0,0 +1,109 @@
...
+        FIELDS TERMINATED BY ','
+        OPTIONALLY ENCLOSED BY '"'
+        LINES TERMINATED BY '\n'
+        FROM (
+            SELECT
+                username,
+                id as user_id
+            FROM auth_user
+            WHERE id NOT IN (
+                SELECT user_id FROM user_api_userretirementstatus
```
crice100, line 31: If a user is retired, are they still in ecomm's user table?

ecommerce/core/management/commands/import_user_ids.py
```diff
@@ -0,0 +1,109 @@
...
+            FROM user_api_userretirementstatus
+        ) u
+
+    Import data into ecommerce using:
+
+    CREATE TABLE temp_username_userid
+    (
+    username VARCHAR(150),
+    lms_user_id INT(11),
+    PRIMARY KEY(lms_user_id)
```
crice100, line 45: For sanity's sake, do we want to insert them in ascending order by this id or is that overkill?

cpennington: I don't think that matters? We can always sort by that when we query (and, in fact, that might end up being the default sort order just because it's the primary key)

cpennington: I also don't really want to have to sort the data at query time in the source services.

ecommerce/core/management/commands/import_user_ids.py
cpennington: I believe so, but I'll verify.

ecommerce/core/management/commands/import_user_ids.py
```diff
@@ -0,0 +1,109 @@
...
+
+        while True:
+            with transaction.atomic():
+                with connection.cursor() as cursor:
+                    update_count = cursor.execute(
+                        """
+                        UPDATE ecommerce_user eu
+                        JOIN temp_username_userid tuu
+                        ON eu.username = tuu.username
+                        SET eu.lms_user_id = tuu.lms_user_id
```
brianhw, line 95: Not sure what behavior we want when there are two LMS users for a given ecommerce user.   In testing this out, I came across only one where this seemed to matter.   LMS user was retired, and before the hole was patched, the user created an LMS user differing in case and email, and continued using that one.   The new lms_user_id field currently has the newer lms_user_id value, while the tracking context still has the old value.  The username matches the old LMS username, so this code would overwrite the new LMS user_id with the old one.   Will this break anything in live code?   (Specifics can be provided on request.)

cpennington: Yeah, the backfill query should probably just not update any rows that already have data.

cpennington: OK, updated to only update null `lms_user_id` rows.

ecommerce/core/management/commands/import_user_ids.py
crice100: Am I reading this correctly that it will not include the ids from retired users? If ecomm still has those users, don't we want to backfill ids for them?

cpennington: The `UNION` clause directly below will get the retired users' ids.

crice100 (approved):

PR #2281, opened 2019-06-04
ARCH-857: Enhance WIP basket api endpoint.

robrap:
- adds additional mock data.
- fixes permissions

ARCH-857

Sample output:
```
{
  "show_voucher_form": true,
  "payment_providers": [
    {
      "type": "cybersource"
    },
    {
      "type": "paypal"
    }
  ],
  "order_total": 149,
  "line_discount": 12,
  "sdn_check": true,
  "line_total": 161,
  "products": [
    {
      "img_url": "https://prod-discovery.edx-cdn.org/media/course/image/21be6203-b140-422c-9233-a1dc278d7266-941abf27df4d.small.jpg",
      "name": "Introduction to Happiness",
      "seat_type": "verified-certificate"
    }
  ],
  "voucher": {
    "benefit": {
      "type": "Percentage",
      "value": 20
    },
    "code": "SUMMER20"
  }
}
```

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -18,6 +18,7 @@ from oscar.apps.basket.views import VoucherRemoveView as BaseVoucherRemoveView
 from oscar.apps.basket.views import *  # pylint: disable=wildcard-import, unused-wildcard-import
 from oscar.core.prices import Price
 from requests.exceptions import ConnectionError, Timeout
+from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.views import APIView
 from slumber.exceptions import SlumberBaseException
@@ -451,7 +452,7 @@ class BasketSummaryView(BasketLogicMixin, BasketView):
             return super(BasketSummaryView, self).get(request, *args, **kwargs)
 
 
-class WIPBasketApiView(BasketLogicMixin, APIView):
+class PaymentApiView(BasketLogicMixin, APIView):
     """
     Api for retrieving basket contents and checkout/payment options.
 
@@ -460,6 +461,8 @@ class WIPBasketApiView(BasketLogicMixin, APIView):
     GET:
     Retrieves basket contents and checkout/payment options.
     """
+    permission_classes = (IsAuthenticated,)
+
     def _get_order_total(self, context):
         """
         Add the order total to the context in preparation for the get_basket_context_data call.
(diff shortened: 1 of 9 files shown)


ecommerce/extensions/basket/views.py
```diff
@@ -473,21 +477,43 @@ def _get_order_total(self, context):
...
-        # - django-oscar-api has serializer for Price
-        # - from oscarapi.serializers.fields import TaxIncludedDecimalField
         return {
-            'currency': price.currency,
-            'excl_tax': price.excl_tax,
-            'incl_tax': price.incl_tax,
+            'id': basket.id,
+            'line_total': context['order_total'].excl_tax + 12,
+            'line_discount': 12,
+            'order_total': context['order_total'].excl_tax,
```
robrap, line 490: Reviewers: Prices in django-oscar include currency ('US'), and price with and without tax.  I erred toward dropping stuff we aren't using to make things more clear for future.

ecommerce/extensions/basket/views.py
```diff
@@ -498,8 +527,12 @@ def get(self, request):  # pylint: disable=unused-argument
         # TODO: ARCH-867: Remove unnecessary processing of anything added to context (e.g. payment_processors) that
         # isn't ultimately passed on to the response.
 
-        response = {}
-        response['order_total'] = self._get_serialized_price(context['order_total'])
+        response = {
+            'count': 1,
```
albemarle: What does this field mean?

robrap: See sample output in description. This is junk that is added to make it more like our generic rest apis that could have multiple results. May yank this if we go for an mfe specific endpoint, and not concern ourselves with best practices. To be discussed.

ecommerce/extensions/basket/views.py
```diff
@@ -473,21 +477,46 @@ def _get_order_total(self, context):
...
+            'detail_message_type': 'single-course',  # could also be: bundle or group
+            'show_voucher_form': True,
+            'voucher': {
+                'code': 'SUMMER20',
+                "benefit": {
+                    "type": "Percentage",
+                    "value": 20
+                },
+            },
+            'payment_providers': [
```
robrap, line 507: Note to self: Does it make sense to move this up and out of "basket" for a Payment page api?

ecommerce/extensions/basket/views.py
```diff
@@ -473,21 +477,46 @@ def _get_order_total(self, context):
...
+            'line_discount': 12,
+            'order_total': context['order_total'].excl_tax,
+            'products': [
+                {
+                    'name': 'Introduction to Happiness',
+                    'seat_type': 'verified-certificate',
+                    'img_url': 'https://prod-discovery.edx-cdn.org/media/course/image/21be6203-b140-422c-9233-a1dc278d7266-941abf27df4d.small.jpg',
+                },
+            ],
+            'detail_message_type': 'single-course',  # could also be: bundle or group
```
robrap: Post discussion: will drop this for now, and add group purchase details later.

ecommerce/extensions/basket/views.py
```diff
@@ -473,21 +477,46 @@ def _get_order_total(self, context):
...
         return {
-            'currency': price.currency,
-            'excl_tax': price.excl_tax,
-            'incl_tax': price.incl_tax,
+            'id': basket.id,
+            'line_total': context['order_total'].excl_tax + 12,
+            'line_discount': 12,
+            'order_total': context['order_total'].excl_tax,
+            'products': [
+                {
```
robrap, line 493: Post discussion: provide a bundle product example as-well, which will have additional depth.

robrap (commented):
Notes from discussion with @abutterworth.

ecommerce/extensions/basket/views.py
robrap: I removed this.

ecommerce/extensions/basket/views.py
robrap: I removed this.

ecommerce/payment_bff/urls.py
```diff
@@ -0,0 +1,18 @@
+from django.conf.urls import include, url
+from rest_framework.urlpatterns import format_suffix_patterns
+
+from ecommerce.extensions.basket.views import PaymentApiView
+
+PAYMENT_URLS = [
+    url(r'^payment/', PaymentApiView.as_view(), name='payment'),
```
robrap, line 7: FYI: This is a straw-man proposal:
```
GET payment-bff/api/v0/payment
POST payment-bff/api/v0/payment/voucher
- use POST over PUT because it returns results of /payment?
DELETE payment-bff/api/v0/payment/voucher
```

ecommerce/extensions/basket/views.py
robrap: No change.

ecommerce/extensions/basket/views.py
robrap: Added a comment, but won't do this until we get to bundles.

ecommerce/payment_bff/urls.py
douglashall: LOVE IT!

ecommerce/extensions/basket/views.py
```diff
@@ -473,21 +477,45 @@ def _get_order_total(self, context):
...
+                    # I think we will need something else here, but not sure.  Otherwise, this could just be an enabled/disabled flag.
+                },
+                {
+                    'type': 'paypal',
+                }
+            ],
+            'sdn_check': True,
         }
 
     def get(self, request):  # pylint: disable=unused-argument
```
douglashall, line 520: Maybe drop `post` and `delete` in here as well returning the same mock response so that the frontend can start making those calls as well.

douglashall (approved):

ecommerce/extensions/basket/views.py
robrap: Those are coming.  Possibly as a separate PR.  Thanks.

PR #2282, opened 2019-06-04
Set celery exchange and routing key; prefix queues

jdmulloy:


bderusha (approved):
Do we also want to add `CELERY_DEFAULT_QUEUE` here and set it to `ecommerce.default`?

bderusha (approved):

PR #2283, opened 2019-06-05
install code-annotations

estute:
install edx/code-annotations so that we can generate feature toggle data. This will be consumed by the feature toggle reporter.


openedx-webhooks:
Although this pull request is already closed,  I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #2284, opened 2019-06-06
REVMI-198 Read lms_user_id from db

crice100:


ecommerce/core/models.py
```diff
@@ -509,13 +509,40 @@ def access_token(self):
...
+        """
+        # Check the cache for the id
+        cache_key = 'lms_user_id_{user_id}'.format(user_id=self.id)
+        cached_response = DEFAULT_REQUEST_CACHE.get_cached_response(cache_key)
+        if cached_response.is_found:
+            return cached_response.value
+
+        # Calculate the id
+        lms_user_id = self._lms_user_id_from_request()
+        if lms_user_id is not None:
```
robrap: If we don't expect it to be updated during the request, we shouldn't do this check and just cache it, even if we are caching/returning None.

ecommerce/core/models.py
```diff
@@ -509,13 +509,40 @@ def access_token(self):
         except Exception:  # pylint: disable=broad-except
             return None
 
-    # TODO Update this to first check the lms_user_id (see REVMI-198). Also consider making this not a property and
-    # renaming it.
-    @property
     def lms_user_id_from_request(self):
```
robrap, line 507: This no longer feels like an appropriate name.

ecommerce/core/models.py
crice100: If we do that, I was concerned that in the future we might add new middleware that runs very early in the request and sees the id is None. Then our existing TrackingMiddleware (which runs later) updates the id, but the new id won't be used until subsequent requests due to the caching. 

I suppose we could catch this situation with logging, but the current approach seemed safer. Thoughts?

ecommerce/core/models.py
crice100: Maybe _get_cached_lms_user_id()_?

ecommerce/core/models.py
robrap: Right.  Should all this code be in the middleware?

ecommerce/core/models.py
robrap: `get_and_cache_lms_user_id()`?

PR #83, opened 2015-05-05
Report course_key and certificate_type to cybersource.

jimabramson:
XCOM-274

@rlucioni or @clintonb please review.

PR #2295, opened 2019-06-10
Attach course_id/program_id to Cybersource transactions.

pwnage101:
https://openedx.atlassian.net/browse/DE-1470


ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -209,6 +210,9 @@ def _generate_parameters(self, basket, use_sop_profile, **kwargs):
                 parameters['item_{}_unit_of_measure'.format(index)] = 'ITM'
                 parameters['item_{}_unit_price'.format(index)] = str(line.unit_price_incl_tax)
 
+                if line.product.course:
+                    extra_data.append(line.product.course.id)
```
pwnage101: Unanswered questions:
* Does this need to be modified to account for child products?
* How can we make this handle bundles? Should we store each course ID in separate `merchant_defined_data#` fields, or collected into a single `merchant_defined_data#` field?
  * Something to keep in mind is that we can store up to 100 `merchant_defined_data#` fields, but \# 5-100 are handled differently: "Merchant-defined data fields 1 to 4 are stored against the payment token and are used for subsequent token based transactions. Merchant defined data fields 5 to 100 are passed trough to Decision Manager as part of the initial payment request and are not stored against the payment token." http://apps.cybersource.com/library/documentation/dev_guides/Secure_Acceptance_Hosted_Checkout/html/wwhelp/wwhimpl/js/html/wwhelp.htm#href=app_fields_WM.10.2.html
* Even if we store a single course_id per field, should it be standalone or have some sort of descriptive decoration?  Hassan suggests possibly using a format like this: `course_id:<COURSE_ID>`
  * Note that AlphaNumericPunctuation is defined as: Alphanumeric including `! "#$%&'()*+,-./:;=?@^_~` (and unicode alpha), but note that curly brackets are not included, so things like JSON would not be possible.

Additional work:
* Also store the program_id in a separate `merchant_defined_data#` field.  This can be inferred via the BasketAttribute model corresponding to the current basket.

doctoryes: A single `merchant_defined_data#` can only store 100 characters. Maybe each `merchant_defined_data#` should contain a course_id/course_type combo?

doctoryes: BTW, I looked through Github history and found the following:

merchant_defined_data# fields were added to processors.py in this PR: https://github.com/edx/ecommerce/pull/83
Moved to cybersource.py in this PR: https://github.com/edx/ecommerce/pull/105
And removed in this PR: https://github.com/edx/ecommerce/pull/983

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -209,6 +210,18 @@ def _generate_parameters(self, basket, use_sop_profile, **kwargs):
                 parameters['item_{}_unit_of_measure'.format(index)] = 'ITM'
                 parameters['item_{}_unit_price'.format(index)] = str(line.unit_price_incl_tax)
 
+                # For each line, add the pair of course_run_id,type as extra CSV data sent to Cybersource.
+                line_course = line.product.course
+                extra_data.append("course,{course_run_id},{course_type}".format(
+                    course_run_id=line_course.id if line_course else "",
+                    course_type=line_course.type if line_course else ""
+                ))
```
doctoryes: The maximum length allowed of a `merchant_defined_data#` field is 100 characters - and the number of these fields that get stored for later retrieval is only four. I decided to:
- prioritize sending the basket course_ids over the program_ids by adding the program_id only after adding all course_ids
- use a CSV format for each course_id line

brianhw: My initial reaction would be to prioritize the program_id over the course_ids.  Seems like there's a chance that not all course_ids would fit, and in theory it should be possible to recover the courses given the program_id as of a given date.   But my next thought is that the program_id should go in a different merchant-defined-data field.   E.g. putting as many course_id values into merchant-defined-data-1, and put the program_uuid into merchant-defined-data-2. 

Also, do these extra fields have an impact on the hashing of the fields being sent?  I wonder if that was why they were removed in the first place, since that seems to be what the PR was implementing where the fields were being removed.

bmedx: I'm +1 on making sure the program IDs get in there.

doctoryes: Thanks for the feedback on program_id, @bmedx and @brianhw - I'll add it first before course_ids (if present) in order to make ensure it can be accessed with the Cybersource transaction.

@brianhw All fields are signed with a private key before being communicated so that Cybersource can detect if any parameters were changed in transit. Adding extra fields will definitely change the signature - but the signature can still be verified as before.

doctoryes: @brianhw For the ordering of the `merchant_defined_data#` parameters, adding all the course_ids to a single field that only supports 100 characters isn't my first choice, as they might be cut-off in the middle - potentially causing confusion. I'm thinking:
- `merchant_defined_data1: program,<program_id>` (if a program_id exists, else course_ids start here...)
- `merchant_defined_data2: course, <course_id1>, <course_type1>`
- `merchant_defined_data3: course, <course_id2>, <course_type2>`
- `merchant_defined_data4: course, <course_id3>, <course_type3>`
- `merchant_defined_data5: course, <course_id4>, <course_type4>`
`...`

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -209,6 +215,13 @@ def _generate_parameters(self, basket, use_sop_profile, **kwargs):
                 parameters['item_{}_unit_of_measure'.format(index)] = 'ITM'
                 parameters['item_{}_unit_price'.format(index)] = str(line.unit_price_incl_tax)
 
+                # For each line, add the pair of course_run_id,type as an extra CSV-formatted param sent to Cybersource.
+                line_course = line.product.course
+                extra_data.append("course,{course_run_id},{course_type}".format(
```
pwnage101: I'm not 100% on whether every order line is always a course.  Does anybody know?

brianhw: Good point.   For bundles, the individual order_line entries should be for entitlements, so there is certainly no concept there of a course_run_id, at any rate.   But the "catalog course" (or whatever we call a "course" that is not a run) may still be located in line.product.course.

ecommerce/extensions/payment/processors/cybersource.py
brianhw: That sounds good.  I might suggest that data1 would be dedicated to program_id, so that it's easier to interpret on the other end.   I believe the only times we would be expecting to get multiple courses is when there's a bundle with a program, so we wouldn't need to work as hard in the course seat case to fit a bunch of courses in.   So if we always start courses at data2, then mostly we would have data1 empty and data2 with course_id, except for programs where we have program_id in data1 and courses in data2 onward.    What ends up getting stored, though, would only be data2, data3, data4, because the others are ignored by cybersource? 

Oh, and I think I like the lack of spaces between commas as you've implemented it.   Slightly different from the example above for courses (but not program).

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -222,6 +235,14 @@ def _generate_parameters(self, basket, use_sop_profile, **kwargs):
             raise PCIViolation('One or more PCI-related fields is contained in the payment parameters. '
                                'This service is NOT PCI-compliant! Deactivate this service immediately!')
 
+        if extra_data:
+            # CyberSource allows us to send additional data in merchant_defined_data# fields.
+            # Only the 1st-4th merchant_defined_data# fields are stored against the payment token.
+            # The 5th-100th nerchant_defined_data# fields are passed through but not stored by Cybersource.
+            for num, item in enumerate(extra_data, start=1):
+                key = u"merchant_defined_data{num}".format(num=num)
+                parameters[key] = item
```
brianhw: If we wanted to skip data1 when there is no program, we could just write None into extra_data, and check here for item is not None.

doctoryes: OK - I've made the changes to have the program UUID always in `merchant_defined_data1` if it exists. And the basket lines which have course products will start at `merchant_defined_data2` and will continue up from there.

brianhw (approved):
LGTM!  👍

pwnage101 (commented):
LGTM

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -339,6 +340,11 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
 
             return self.redirect_to_receipt_page(notification)
         except:  # pylint: disable=bare-except
+            logger.exception(
+                "Failed order creation - order number: %s - notification: %s",
+                request.POST.get('req_reference_number'),
+                notification
+            )
```
doctoryes: I checked in Splunk for prod-edx - and there were only 15 Cybersource payment errors that happened last week and only 94 Cybersource payment errors in the last month. So these two new logs won't spam Splunk, while providing ecommerce developers an idea of why a Cybersource transaction failed.

doctoryes:
@crice100 Can you merge ecommerce PRs? If so, this one is ready for review and merge.

crice100:
@doctoryes This branch was out of date, so I merged master into it. Once the test pass, I'll merge and deploy it

doctoryes:
@robrap You had mentioned in Slack that the Arch Squad was working on an MFE for the shopping cart. Are you aware of this change that has recently shipped to production? Does it affect your dev work?

robrap:
@doctoryes: sorry I forgot to respond. I did appreciate you tagging me.

PR #2297, opened 2019-06-11
Automate CyberSource duplicate payment runbook.

waheedahmed:
If transaction ID for payment notification is different and an order already exists for that order number raise DuplicatePaymentNotification.

LEARNER-6492

### DIFF
--- a/ecommerce/extensions/payment/exceptions.py
+++ b/ecommerce/extensions/payment/exceptions.py
@@ -59,3 +59,13 @@ class InvalidBasketError(PaymentError):
 class AuthorizationError(PaymentError):
     """ Authorization was declined. """
     pass
+
+
+class MultiplePaymentNotification(PaymentError):
+    """ Raised when duplicate payment notification is detected with same transaction ID. """
+    pass
+
+
+class DuplicatePaymentNotification(PaymentError):
+    """ Raised when duplicate payment notification is detected with different transaction ID. """
+    pass
--- a/ecommerce/extensions/payment/exceptions.py
+++ b/ecommerce/extensions/payment/exceptions.py
@@ -61,11 +61,11 @@ class AuthorizationError(PaymentError):
     pass
 
 
-class MultiplePaymentNotification(PaymentError):
+class RedundantPaymentNotificationError(PaymentError):
     """ Raised when duplicate payment notification is detected with same transaction ID. """
     pass
 
 
-class DuplicatePaymentNotification(PaymentError):
+class ExcessivePaymentForOrderError(PaymentError):
     """ Raised when duplicate payment notification is detected with different transaction ID. """
     pass
--- a/ecommerce/extensions/payment/processors/cybersource.py
+++ b/ecommerce/extensions/payment/processors/cybersource.py
@@ -22,9 +22,11 @@ from ecommerce.extensions.checkout.utils import get_receipt_page_url
 from ecommerce.extensions.payment.constants import APPLE_PAY_CYBERSOURCE_CARD_TYPE_MAP, CYBERSOURCE_CARD_TYPE_MAP
 from ecommerce.extensions.payment.exceptions import (
     AuthorizationError,
+    DuplicatePaymentNotification,
     DuplicateReferenceNumber,
     InvalidCybersourceDecision,
     InvalidSignatureError,
+    MultiplePaymentNotification,
     PartialAuthorizationError,
     PCIViolation,
     ProcessorMisconfiguredError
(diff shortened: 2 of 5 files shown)


waheedahmed:
@edx/ecommerce Can someone please review?

ecommerce/extensions/payment/tests/mixins.py
```diff
@@ -488,7 +494,14 @@ def test_not_accepted(self, decision):
         (KeyError, 'ERROR', 'Attempts to handle payment for basket [{basket_id}] failed. The payment response '
                             '[Unknown Error] was recorded in entry [{response_id}].'),
         (AuthorizationError, 'INFO', 'Payment Authorization was declined for basket [{basket_id}]. The payment '
-                                     'response was recorded in entry [{response_id}].')
+                                     'response was recorded in entry [{response_id}].'),
+        (DuplicatePaymentNotification, 'INFO', 'Received duplicate CyberSource payment notification with different '
```
cpennington: This message suggests that we're requesting a refund, but the code doesn't automatically request a refund. Is the plan to automatically request the refund via the cybersource api?

waheedahmed: As suggested in the ticket we will create new splunk alert for this error and will send appropriate details to customer support so that they can initiate a refund.

ecommerce/extensions/payment/exceptions.py
```diff
@@ -59,3 +59,13 @@ class InvalidBasketError(PaymentError):
...
     """ Authorization was declined. """
     pass
+
+
+class MultiplePaymentNotification(PaymentError):
+    """ Raised when duplicate payment notification is detected with same transaction ID. """
+    pass
+
+
+class DuplicatePaymentNotification(PaymentError):
```
mulby: nit: It's difficult for me to tell the difference between these error types just from their names.

I would suggest:

"RedundantPaymentNotificationError" (for the case when we receive multiple notifications for the same transaction)

and

"ExcessivePaymentForOrderError" (for the case where we receive multiple payment notifications with different transaction IDs for the same order)

ecommerce/extensions/payment/processors/cybersource.py
```diff
@@ -278,6 +281,15 @@ def handle_processor_response(self, response, basket=None):
                     'review': AuthorizationError,
                 }.get(decision, InvalidCybersourceDecision)
 
+        transaction_id = response.get('transaction_id', None)  # Error Notifications does not include a transaction id.
+        if transaction_id and decision == 'accept':
+            if PaymentProcessorResponse.objects.filter(transaction_id=transaction_id).exists():
+                if Order.objects.filter(number=response['req_reference_number']).exists():
```
mulby: nit: since this check happens in both branches of the conditional on the previous line, it might be more clear to do something like this:

```python
if Order.objects.filter(number=response['req_reference_number']).exists():
    if PaymentProcessorResponse.objects.filter(transaction_id=transaction_id).exists():
        raise MultiplePaymentNotification
    else:
        raise DuplicatePaymentNotification
```

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -262,6 +264,22 @@ def validate_notification(self, notification):
...
+                    'Received multiple CyberSource payment notification with same transaction ID for basket [%d] '
+                    'which is associated with an existing order [%s]. No payment was collected.',
+                    basket.id,
+                    order_number
+                )
+                raise
+            except DuplicatePaymentNotification:
+                logger.info(
+                    'Received duplicate CyberSource payment notification with different transaction ID for basket [%d] '
+                    'which is associated with an existing order [%s]. Payment collected twice, requesting a refund.',
```
mulby: Given the conversation above, this message leads me to believe a refund is being automatically requested.

We also should run this message by support to ensure it contains all of the data they need to issue the refund.

mulby (commented):
Thanks for doing this, will help a great deal with the pager!

I put in a few comments.

ecommerce/extensions/payment/views/cybersource.py
waheedahmed: I believe all they need is order id and basket id as described in the run book:

> send an email to info@edx.org to open a support ticket, asking them to check CyberSource (or Paypal as appropriate) for a duplicate payment for the transaction. You can send the order id from the Splunk alert message, and include the basket id just in case.

waheedahmed:
Thank you @mulby and @cpennington, I have addressed your comments kindly take another look.

cpennington (approved):

PR #2290, opened 2019-06-10
Remove enterprise related fields from old coupon tool UI

muhammad-ammar:
**Description:** This will remove the enterprise customer and enterprise customer catalog fields from old coupon tool. On coupon create screen, both of the the fields will not be shown. On coupon edit screen, enterprise customer catalog field will not be shown at all and enterprise customer field will visible if there is an enterprise customer attached with the coupon.

**JIRA:** [ENT-1985](https://openedx.atlassian.net/browse/ENT-1985)

**Sandbox Login:** Authentication on sandbox is broken but one can login through django admin. Below are the steps I followed to login on Sandbox

0. Sandbox basic auth credentials
   username: **margaret**
   password: **hamilton**
1. Login to https://ecommerce-mammar.sandbox.edx.org/admin using your superuser credentials
2. After successful login in step 1, Go to https://ecommerce-mammar.sandbox.edx.org/coupons and you should be able to explore the coupons :)

**Testing:** Sandbox is updated with changes in this PR.
- Edit: [Old coupon without enterprise customer doesn't show the enterprise customer field](https://ecommerce-mammar.sandbox.edx.org/coupons/5/edit/)
- Create: [Enterprise Customer Catalog and Enterprise Customer Catalog fields are not shown on old coupon screen](https://ecommerce-mammar.sandbox.edx.org/coupons/new/)

PR #2298, opened 2019-06-11
ENT-1987: Added catalog preview link in enterprise coupon screen

saleem-latif:
**Description:**
Now that enterprise catalogs are required for enterprise coupons, we can make it easier to see the courses associated with that coupon.

**Acceptance Criteria:**
Add a link below the enterprise catalog field when a catalog is filled in. This link should open in a new tab and show the JSON results of the catalog.

### DIFF
--- a/ecommerce/extensions/api/serializers.py
+++ b/ecommerce/extensions/api/serializers.py
@@ -7,12 +7,14 @@ from decimal import Decimal
 
 import waffle
 from dateutil.parser import parse
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db import transaction
 from django.db.models import Q
 from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
 from oscar.core.loading import get_class, get_model
+from path import Path
 from rest_framework import serializers
 from rest_framework.reverse import reverse
 from rest_framework.settings import api_settings
@@ -957,6 +959,7 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     seats = serializers.SerializerMethodField()
     start_date = serializers.SerializerMethodField()
     voucher_type = serializers.SerializerMethodField()
+    enterprise_catalog_url_template = serializers.SerializerMethodField()
 
     def get_benefit_type(self, obj):
         return retrieve_benefit(obj).type or getattr(retrieve_benefit(obj).proxy(), 'benefit_class_type', None)
@@ -1092,6 +1095,9 @@ class CouponSerializer(ProductPaymentInfoMixin, serializers.ModelSerializer):
     def get_voucher_type(self, obj):
         return retrieve_voucher_usage(obj)
 
+    def get_enterprise_catalog_url_template(self, obj):  # pylint: disable=unused-argument
+        return Path(settings.ENTERPRISE_API_URL) / "enterprise_catalogs/"
+
     class Meta(object):
         model = Product
         fields = (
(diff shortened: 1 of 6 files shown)


irfanuddinahmad (commented):
LGTM! ... except for 
image

ecommerce/extensions/api/serializers.py
```diff
@@ -1092,14 +1095,18 @@ def get_seats(self, obj):
     def get_voucher_type(self, obj):
         return retrieve_voucher_usage(obj)
 
+    def get_enterprise_catalog_url_template(self, obj):  # pylint: disable=unused-argument
+        return Path(settings.ENTERPRISE_API_URL) / "enterprise_catalogs/"
```
muhammad-ammar, line 1099: Can you please explain what is happening here? Why we are using `Path` and why function is named as `_template`?

ecommerce/static/js/views/enterprise_coupon_form_view.js
```diff
@@ -84,7 +88,22 @@ define([
...
+                var enterpriseAPIURL = this.model.get('enterprise_catalog_url_template');
+                if (enterpriseCoupon && enterpriseAPIURL) {
+                    this.$(
+                        '#enterprise-catalog-details'
+                    ).attr('href', enterpriseAPIURL + enterpriseCoupon).addClass('external-link').removeClass('hidden');
+                } else {
+                    this.$(
+                        '#enterprise-catalog-details'
+                    ).removeAttr('href').removeClass('external-link').addClass('hidden');
+                }
```
muhammad-ammar, line 106: Changes in the file will only work on coupon edit screen, right?

ecommerce/static/templates/enterprise_coupon_form.html
```diff
@@ -160,6 +160,7 @@
         <div class="form-group enterprise-customer-catalog">
             <label for="enterprise-customer-catalog"><%= gettext('Enterprise Customer Catalog') %></label>
             <input id="enterprise-customer-catalog" type="text" class="form-control" name="enterprise_customer_catalog">
+            <a id="enterprise-catalog-details" aria-label="View enterprise catalog details.">Enterprise Catalog Details</a>
```
muhammad-ammar, line 163: Do we need a product review for this link text?

muhammad-ammar (approved):
I have testing the changes locally on my devstack and they are looking good. Left some minor questions though.

ecommerce/extensions/api/serializers.py
saleem-latif: Here we are constructing a path to enterprise catalog API endpoint from the base API URL, Path is used to handle trailing/missing `\` in the base URL.  I used this name as it was suggested @ https://openedx.atlassian.net/browse/ENT-1987?focusedCommentId=384275&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-384275

ecommerce/static/js/views/enterprise_coupon_form_view.js
saleem-latif: The code will execute in case of Coupon creation screen, but since [enterpriseAPIURL](https://github.com/edx/ecommerce/pull/2298/files/df0ff9d3d8522f9678d0bcc7f4e09f186b33ec7e#diff-fc4f7da766f57eb9ac803e5b5cab61b1R97) will be null for coupon creation screen. This link will be hidden.

ecommerce/static/templates/enterprise_coupon_form.html
saleem-latif: Yes, it would be good if they review as well, I think once the changes are deployed somewhere, we can ask for their review as well.

christopappas (commented):
Makes sense at a high level to me 👍

ecommerce/static/templates/enterprise_coupon_form.html
brittneyexline: since this an internal tool, i think this wording makes sense and doesn't need a review from product.

brittneyexline (approved):
i will release this and https://github.com/edx/ecommerce/pull/2290 together once some issues with the ecommerce pipeline are resolved.
fyi @muhammad-ammar

PR #2299, opened 2019-06-11
Added tox to run unit tests under Python 2.7 and 3.6

shafqatfarhan:
This is the first step in the migration of e-commerce to python 3.6. I have added tox that will make unit tests run under Python 2.7 and 3.6. This PR contains tox configurations only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: python
-python:
-  - "2.7"
 
 branches:
   only:
@@ -26,11 +24,10 @@ before_install:
 
 matrix:
   include:
-    - env: TESTNAME=quality-and-js
-
+    - python: 2.7
+      env: TESTNAME=quality-and-js
       install:
         - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ && make requirements'
-
       script:
         - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/docs && make html'
         - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ && DJANGO_SETTINGS_MODULE=ecommerce.settings.test make check_translations_up_to_date'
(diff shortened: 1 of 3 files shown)


cpennington:
Can you add 3.6 testing in travis as well? I should be able to make it not required to merge, but that way we can track progress/regressions on PRs.

cpennington:
It looks like travis python 3.6 is passing tests, which makes me think it's not actually running python 3.6 when it runs the tests.

cpennington:
My guess would be that `make validate_python` uses the system-wide python in the docker container, and that that is python 2.7. You likely need to make it possible to use tox, and then explicitly use either the 2.7 tox environment or the 3.6 tox environment.

shafqatfarhan:
Thanks for the suggestions @cpennington

Yes you were right that docker container's python 2.7 is being used as docker commands in Travis script  activate ecommerce_env that uses python 2.7

I have updated the Travis script and now python 3.6 is being used to execute tox to run tests. The script is modified to run python3.6 Travis job outside of ecommerce-test docker container because the container does not have python 3.6 installed in it. Maybe we have to update container to have python 3.6 installed in it. Let me know your suggestions.

.travis.yml
```diff
@@ -42,19 +39,25 @@ matrix:
...
         - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ && DJANGO_SETTINGS_MODULE=ecommerce.settings.test make clean_static'
         - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ && DJANGO_SETTINGS_MODULE=ecommerce.settings.test make static'
         - docker exec -t ecommerce_testing bash -c 'source /edx/app/ecommerce/ecommerce_env && cd /edx/app/ecommerce/ecommerce/ && DJANGO_SETTINGS_MODULE=ecommerce.settings.test TRAVIS=1 xvfb-run make validate_python'
-
       after_success:
         - pip install -U codecov
         - docker exec ecommerce_testing /edx/app/ecommerce/ecommerce/.travis/run_coverage.sh
         - codecov
+    
+    - python: 3.6
```
cpennington, line 56: We should use https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail to allow this build to fail

shafqatfarhan: I have added `allow_failures` in the script. Travis CI build is green now.

tox.ini
```diff
@@ -0,0 +1,11 @@
+[tox]
+skipsdist=True
+envlist = {py27,py36}-master
+
+[testenv]
+setenv =
+    PYTHONPATH = {toxinidir}
+whitelist_externals =
+    make
+commands = make requirements
```
cpennington, line 10: I think that tox has code that will only rebuild an environment if dependencies have changed, but to use it I think you'd need to reference the python requirements specifically inside the `depends` (I think that's the name) block in tox, rather than using `make requirements`. That said, that could be a future improvement, rather than something added right now.

cpennington:
Can you squash this branch, and then I'll merge it?

shafqatfarhan:
@cpennington I have squashed the commits. Let me know if you want me to update tox script to use `deps` in the same PR.

cpennington:
No, we can fix `deps` as a follow on task. Merging this unblocks you all from starting to make incremental improvements.

PR #2301, opened 2019-06-13
make enterprise catalog field a dropdown

muhammad-ammar:
**Description:** This converts the `Enterprise Customer Catalog` field on Enterprise Create/Edit screen to a dropdown. Contents of the dropdown will be names of all the catalogs belong to the enterprise customer selected in the `Enterprise Customer` field.  

**JIRA:** [ENT-1986](https://openedx.atlassian.net/browse/ENT-1986)

**Testing:** PR changes are deployed to sandbox.

**Credentials:**
username: edx 
password: edx

1. Login to https://ecommerce-mammar.sandbox.edx.org/admin with the above credentials
2. Then go to https://ecommerce-mammar.sandbox.edx.org/enterprise/coupons/new/
3. Creating/Editing an Enterprise Coupon should work

### DIFF
--- a/ecommerce/static/js/test/specs/collections/course_collection_spec.js
+++ b/ecommerce/static/js/test/specs/collections/course_collection_spec.js
@@ -37,8 +37,7 @@ define([
                     response.next = '/api/v2/courses/?page=2';
 
                     collection.parse(response);
-                    expect(collection.url).toEqual(response.next);
-                    expect(collection.fetch).toHaveBeenCalledWith({remove: false});
+                    expect(collection.fetch).toHaveBeenCalledWith({remove: false, url: '/api/v2/courses/?page=2'});
                 });
             });
         });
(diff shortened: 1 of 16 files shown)


ecommerce/static/js/test/specs/collections/course_collection_spec.js
```diff
@@ -37,8 +37,7 @@ define([
                     response.next = '/api/v2/courses/?page=2';
 
                     collection.parse(response);
-                    expect(collection.url).toEqual(response.next);
-                    expect(collection.fetch).toHaveBeenCalledWith({remove: false});
```
brittneyexline, line 41: any reason why we are changing this?

brittneyexline (approved):
looks good overall, just one question about the tests

ecommerce/static/js/test/specs/collections/course_collection_spec.js
muhammad-ammar: @brittneyexline This is due to the change in `paginated_collection.js`. In that file, we removed `this.url = response.next;` and passed the `response.next` to `fetch` method. We did this because setting `this.url` in `parse` method will replace the original url which is not the correct behaviour and will cause problems when we try to fetch the collection again.

PR #2303, opened 2019-06-13
ARCH-853: implement add voucher api

robrap:
ARCH-853


ecommerce/extensions/basket/views.py
```diff
@@ -640,19 +687,35 @@ class AddVoucherApiView(APIView):
...
     """
     permission_classes = (IsAuthenticated,)
+    voucher_model = get_model('voucher', 'voucher')
 
     def post(self, request):  # pylint: disable=unused-argument
+        # TODO: Does code need to be trimmed to match original?
         code = request.data.get('code')
-        # TODO: ARCH-853: Replace mock implementation with real implementation
+
+        error_message = self.check_for_empty_basket_error(code)
```
davidjoy, line 701: Reminder: Change status code!

robrap: Done.

ecommerce/extensions/basket/views.py
```diff
@@ -558,123 +562,196 @@ def apply_voucher_to_basket(self, voucher):
...
+                }
+            )
+            return HttpResponseRedirect(
+                '{path}?{params}'.format(
+                    path=reverse('coupons:redeem'),
+                    params=params
+                )
+            )
+
+        self.apply_voucher_to_basket(voucher)
```
douglashall, line 673: It's hard to tell from the diff. Is this just a refactor of existing code or is there something new in here? Is the refactor necessary to accomplish the current task at hand?

douglashall (approved):

ecommerce/extensions/basket/views.py
```diff
@@ -558,123 +562,196 @@ def apply_voucher_to_basket(self, voucher):
...
 class RemoveVoucherApiView(APIView):
     """
     Api for removing voucher from a basket.
 
-    DELETE:
-    Will return 200 and the relevant basket updates as json if successful.
-    If unsuccessful, will return 406 with the error.
-
-    NOTE: Because this is a backend-for-frontend api, we are ok with returning content for a DELETE api.
+    DELETE /payment-bff/v0/payment/vouchers/{voucherid}
```
abutterworth: This url should now be /bff/payment/...?

abutterworth (approved):

ecommerce/extensions/basket/views.py
robrap: `VoucherAddView` is old.  `AddVoucherApiView` is new.  (Note: This probably should be `VoucherAddApiView` for consistency.)

The same code is reused in both to ensure they are doing the same thing and they can't drift, rather than copying/pasting the old code (which is also a possibility).  The same question will arise with the unit tests (refactor/reuse vs copy/paste).  Although the refactor isn't strictly necessary, it ensures the old and new are calling the same code.

robrap: Note: for unit tests, I may end up copying/pasting, but there is a separate ticket for that later.

PR #2305, opened 2019-06-14
REVMI-276 Log warning if refund is missing lms_user_id

crice100:
Log warning if refund is missing lms_user_id and add new relic metrics

### DIFF
--- a/ecommerce/extensions/api/v2/views/refunds.py
+++ b/ecommerce/extensions/api/v2/views/refunds.py
@@ -1,5 +1,8 @@
 """HTTP endpoints for interacting with refunds."""
+import logging
+
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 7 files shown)


ecommerce/extensions/api/v2/views/refunds.py
```diff
@@ -81,6 +89,16 @@ def create(self, request, *args, **kwargs):
         except User.DoesNotExist:
             raise BadRequestException('User "{}" does not exist.'.format(username))
 
+        if not user.lms_user_id:
```
cpennington: Should this just be `lms_user_id_with_metric`?

crice100: Probably. :-) I'll update it

PR #2317, opened 2019-06-18
ARCH-853: refactor and remove voucher api

robrap:
- full payment page results are returned by payment and voucher apis
- implemented remove voucher api

ARCH-853

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -453,16 +453,10 @@ class BasketSummaryView(BasketLogicMixin, BasketView):
             return super(BasketSummaryView, self).get(request, *args, **kwargs)
 
 
-class PaymentApiView(BasketLogicMixin, APIView):
+class PaymentApiLogicMixin(BasketLogicMixin):
     """
-    Api for retrieving basket contents and checkout/payment options.
-
-    # TODO: ARCH-867: See ticket for clean-up and refactoring details for this API code.
-
-    GET:
-    Retrieves basket contents and checkout/payment options.
+    Business logic for the various Payment APIs.
     """
-    permission_classes = (IsAuthenticated,)
 
     def _get_order_total(self, context):
         """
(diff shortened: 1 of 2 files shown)


ecommerce/extensions/basket/views.py
```diff
@@ -713,45 +739,51 @@ def post(self, request):  # pylint: disable=unused-argument
...
+# TODO: ARCH-960: Remove "pragma: no cover"
+class VoucherRemoveApiView(PaymentApiLogicMixin, APIView):  # pragma: no cover
     """
     Api for removing voucher from a basket.
 
     DELETE /bff/payment/v0/vouchers/{voucherid}
     """
     permission_classes = (IsAuthenticated,)
+    voucher_model = get_model('voucher', 'voucher')
+    remove_signal = signals.voucher_removal
```
albemarle, line 755: I don't know if this is obvious to others, but I wouldn't know what remove_signal is meant to be.

ecommerce/extensions/basket/views.py
```diff
@@ -713,45 +739,51 @@ def post(self, request):  # pylint: disable=unused-argument
...
-        # TODO: ARCH-853: Replace mock implementation with real implementation
-        return Response(
-            {
-                'total_excl_discount': 112,
-                'total_discount': 0,
-                'order_total': 112,
+        if not request.basket.id:
+            # Hacking attempt - the basket must be saved for it to have
+            # a voucher in it.
+            # Note: original django-oscar code had no error message.
```
albemarle, line 769: Is that important?

albemarle (approved):

ecommerce/extensions/basket/views.py
robrap: 1. Not sure if you are familiar with Django Signal's: https://docs.djangoproject.com/en/2.2/topics/signals/
2. I don't know if we are using this signal, but keeping as close to the django-oscar code (which was copy/pasted) as possible to not unintentionally break things.

ecommerce/extensions/basket/views.py
robrap: Very. ;)
Maybe not, but since this code was copy/pasted from django-oscar, I was just trying to call out differences in case a django-oscar upgrade ever warrants an updated to the copied code.

PR #2339, opened 2019-06-18
INCR-508 - Run python-modernize on ./ecommerce/extensions/checkout

ihtram:


ecommerce/extensions/checkout/mixins.py
```diff
@@ -20,6 +21,7 @@
 from ecommerce.extensions.offer.constants import OFFER_ASSIGNED, OFFER_ASSIGNMENT_REVOKED, OFFER_REDEEMED
 from ecommerce.extensions.order.constants import PaymentEventTypeName
 from ecommerce.invoice.models import Invoice
+from six.moves import range  # pylint: disable=import-error,ungrouped-imports
```
crice100: Could we group the imports instead of disabling the _ungrouped-imports_ check?

ihtram: @crice100 In that case we have to skip isort check on this group. Let me commit this change.

ihtram: @crice100 I have updated the code. Please review again

ecommerce/extensions/checkout/mixins.py
```diff
@@ -21,6 +21,10 @@
 from ecommerce.extensions.order.constants import PaymentEventTypeName
 from ecommerce.invoice.models import Invoice
 
+import six  # isort:skip
```
crice100: @cpennington Could you look at this to see if ecommerce's isort or pylint configs should be updated? It seems odd to me that we need to skip sorting to have a clean build.

cpennington: After rebasing onto master w/ the latest isort, I think this is fixed.

cpennington:
Unfortunately, even upgrading both isort and pylint doesn't fix this: See https://github.com/PyCQA/pylint/pull/2824, which has been merged to pylint master, but not yet released.

I'm going to continue w/ the pylint upgrade anyway, though.

crice100 (approved):

PR #2342, opened 2019-06-18
REVMI-286 Get most recent access token

crice100:
Get the most recently saved access token and LMS user id.

Background: when we initially added code to get the user_id from the user's social auth, we only checked in the first (oldest) social auth row. This had a high rate of misses. When we switched to looping over all of the social auth rows for a user, the rate of misses plummeted.

Since the access token code also only checked the first (oldest) row, it seemed like a good idea to update it to account for users can having multiple social auth db rows (see REVMI-285 for queries I ran while looking into this).

### DIFF
--- a/ecommerce/core/models.py
+++ b/ecommerce/core/models.py
@@ -497,8 +497,17 @@ class User(AbstractUser):
 
     @property
     def access_token(self):
+        """
+        Returns the access token from the extra data in the user's social auth.
+
+        Note that a single user_id can be associated with multiple provider/uid combinations. For example:
+            provider    uid             user_id
+            edx-oidc    person          123
+            edx-oauth2  person          123
+            edx-oauth2  person@edx.org  123
+        """
         try:
-            return self.social_auth.first().extra_data[u'access_token']  # pylint: disable=no-member
+            return self.social_auth.order_by('-id').first().extra_data[u'access_token']  # pylint: disable=no-member
         except Exception:  # pylint: disable=broad-except
             return None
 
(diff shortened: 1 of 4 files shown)


ecommerce/core/models.py
```diff
@@ -497,8 +497,17 @@ class Meta(object):
...
+
+        Note that a single user_id can be associated with multiple provider/uid combinations. For example:
+            provider    uid             user_id
+            edx-oidc    person          123
+            edx-oauth2  person          123
+            edx-oauth2  person@edx.org  123
+        """
         try:
-            return self.social_auth.first().extra_data[u'access_token']  # pylint: disable=no-member
+            return self.social_auth.last().extra_data[u'access_token']  # pylint: disable=no-member
```
cpennington: I would feel better if you used an explicit sort order here (`.order_by(-id).first()`, maybe?)

ecommerce/extensions/analytics/middleware.py
```diff
@@ -57,15 +58,20 @@ def process_request(self, request):
...
+        Find the LMS user_id passed through social auth. Because a single user_id can be associated with multiple
+        provider/uid combinations, start by checking the most recently saved social auth entry.
+
+        Returns:
+            (lms_user_id, social_auth_id): a tuple containing the LMS user id and the id of the social auth entry
+                where the LMS user id was found. Returns None, None if the LMS user id was not found.
         """
         try:
-            auth_entries = user.social_auth.all()
+            auth_entries = user.social_auth.all().order_by('-id')
```
cpennington: I don't think you need the `all()` anymore if you have the `order_by`, but I'm not 100% sure.

crice100: Updated

ecommerce/core/models.py
crice100: Works for me. I made the change.

brittneyexline (approved):
lgtm! i am curious what the implications of doing this are/what the intended impact is - ie, will this reduce the auth related errors we see in ecommerce because we effectively won't be using the old style tokens anymore?

PR #2343, opened 2019-06-19
ARCH-969: handle more voucher errors

robrap:
I still need to manually test the new errors, but I think this all works.

ARCH-969

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -536,8 +536,7 @@ class PaymentApiLogicMixin(BasketLogicMixin):
         response = self.get_serialized_basket(context)
         if errors:
             response['errors'] = errors if isinstance(errors, list) else [errors]
-        # TODO: Check flash messages to see if there are any additional errors or warnings?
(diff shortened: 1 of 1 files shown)


ecommerce/extensions/basket/views.py
```diff
@@ -536,6 +536,8 @@ def get_payment_api_response(self, errors=None):
         response = self.get_serialized_basket(context)
         if errors:
             response['errors'] = errors if isinstance(errors, list) else [errors]
+        # TODO: Check flash messages to see if there are any additional errors or warnings?
```
robrap: Reviewers: I need help deciding whether we should:
1. Implement this TODO as part of this ticket, or
2. Add an ARCH-XXX comment and add this to a different ticket, or
3. Forget about this, and if there are any, we just lose that info.  This is a safety net, and I don't know whether it will catch anything or not.

robrap: I decided this isn't strictly needed, so not doing it now.  I chose to add a new ticket, and we can decide to implement or not later.

albemarle (approved):

PR #2344, opened 2019-06-20
Emma green/ecom frontend

emma-green:
This is for this ticket: https://openedx.atlassian.net/browse/REVEM-281


ecommerce/enterprise/tests/test_migrate_enterprise_conditional_offers_command.py
```diff
@@ -174,7 +175,7 @@ def test_handle(self):
         handle should create new conditional offers for all voucher objects
         that ultimately relate to a range that has an enterprise_customer
         """
-        offers = ConditionalOffer.objects.all()
+        offers = ConditionalOffer.objects.filter(~Q(name='dynamic_conditional_offer'))
```
cpennington: I think this could just be `exclude(name='dynamic_conditional_offer')`

ecommerce/enterprise/tests/test_views.py
```diff
@@ -130,8 +131,9 @@ def test_post(self):
             'benefit_type': Benefit.PERCENTAGE,
             'benefit_value': expected_benefit_value,
         }
+        existing_offer_ids = [c.id for c in ConditionalOffer.objects.all()]
         response = self.client.post(self.path, data, follow=False)
-        enterprise_offer = ConditionalOffer.objects.get()
+        enterprise_offer = ConditionalOffer.objects.get(~Q(id__in=existing_offer_ids))
```
cpennington: This assumes that there's only one offer that's not in `existing_offer_ids`. Is that because you're trying to test creating a new one?

cpennington: It might be clearer to query for `exclude(id_in=existing_offer_ids)`, and then assert that a) there's only one, and b) that that single offer has the requisite parameters.

ecommerce/extensions/offer/dynamic_conditional_offer.py
```diff
@@ -0,0 +1,79 @@
...
+PercentageDiscountBenefit = get_model('offer', 'PercentageDiscountBenefit')
+Condition = get_model('offer', 'Condition')
+
+
+def get_decoded_jwt_discount_from_request():
+    request = crum.get_current_request()
+    if request.method == 'GET':
+        discount_jwt = request.GET.get('discount_jwt')
+    else:
+        discount_jwt = request.POST.get('discount_jwt')
```
cpennington, line 30: It seems odd to me that the `discount_jwt` might be a get parameter, or might be a post. Maybe worth a comment?

ecommerce/extensions/offer/tests/test_applicator.py
```diff
@@ -64,8 +66,9 @@ def test_get_offers_with_bundle(self):
 
     def test_get_offers_without_bundle(self):
         """ Verify that all non bundle offers are returned if no bundle id is given. """
+        co_in_db = [co for co in ConditionalOffer.active.filter(offer_type=ConditionalOffer.SITE).all()]
```
cpennington: You don't need `all()` after `filter(...)`.

cpennington: Also, you can convert to a list by just calling `list` on a `QuerySet`, rather than using a list comprehension.

ecommerce/enterprise/tests/test_migrate_enterprise_conditional_offers_command.py
MatthewPiatetsky: perhaps a comment for why this is necessary?

ecommerce/extensions/offer/dynamic_conditional_offer.py
```diff
@@ -0,0 +1,79 @@
+
```
MatthewPiatetsky: can you add a docstring?  
I don't know what a dynamic conditional offer.  
Could you also explain in the comment how this differs from other offers currently being used like enterprise?

ecommerce/extensions/offer/dynamic_conditional_offer.py
```diff
@@ -0,0 +1,79 @@
...
+                basket,
+                condition,
+                offer,
+                discount_percent=percent,
+                max_total_discount=max_total_discount)
+            return application_result
+        return None
+
+
+class DynamicCustomerCondition(ConditionWithoutRangeMixin, SingleItemConsumptionConditionMixin, Condition):
```
MatthewPiatetsky: is this a DynamicCustomerCondition or dynamic_discount_condition like mentioned below?  
Could you add a docstring to the class explaining what this is

ecommerce/extensions/offer/migrations/0025_auto_20190624_1747.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
```
MatthewPiatetsky, line 1: is there any reason this migration would need a reverse?

ecommerce/extensions/offer/tests/test_applicator.py
MatthewPiatetsky: personally i prefer longer but clearer variable names for clarity.  
don't actually know what this variable means but maybe it could be offers_in_db_list?
though i think variable naming matters much less in tests

ecommerce/extensions/offer/tests/test_dynamic_conditional_offer.py
```diff
@@ -0,0 +1,88 @@
+
```
MatthewPiatetsky, line 1: extra spaces?  
maybe a docstring for each class describing what they're testing? (though if it is not possible to make it clear what is tested in a couple sentences then not worth it)

ecommerce/extensions/payment/views/cybersource.py
```diff
@@ -193,6 +197,22 @@ def _get_basket(self, basket_id):
             basket_id = int(basket_id)
             basket = Basket.objects.get(id=basket_id)
             basket.strategy = strategy.Default()
+            if waffle.flag_is_active(self.request, DYNAMIC_DISCOUNT_FLAG):
+                # TODO: Remove as a part of REVMI-124
+                # This is a hacky solution to the problem of orders being created after payment processing
+                lms_discount_client = EdxRestApiClient(get_lms_url('/api/discounts/'), jwt=basket.owner.access_token)
```
MatthewPiatetsky: could you describe here what the solution is?  
I'm assuming this was the issue you discussed where you no longer had access to jwt and discount wasn't saved with basket so you're making a second request?

ecommerce/programs/tests/test_views.py
```diff
@@ -125,8 +126,9 @@ def test_post(self):
             'benefit_type': Benefit.PERCENTAGE,
             'benefit_value': expected_benefit_value,
         }
+        existing_offer_ids = [c.id for c in ConditionalOffer.objects.all()]
         response = self.client.post(self.path, data, follow=False)
-        program_offer = ConditionalOffer.objects.get()
+        program_offer = ConditionalOffer.objects.get(~Q(id__in=existing_offer_ids))
```
MatthewPiatetsky: what is the ~Q doing here vs not having it?

ecommerce/static/js/apps/basket_app.js
```diff
@@ -9,7 +9,7 @@ require([
         $(document).ready(function() {
             // TODO: Change to responsive javascript after this site turns into a MFE
             var oldUrl, indexOfFragment, discountString, newUrl;
-            if (!window.location.href.split('discount_jwt=')[1]) {
+            if (window.courseKey && !window.location.href.split('discount_jwt=')[1]) {
```
MatthewPiatetsky: could you explain this change?

ecommerce/templates/edx/checkout/receipt.html
```diff
@@ -126,6 +126,8 @@ <h2>{% trans "Order Information" %}</h2>
                                 {% blocktrans trimmed with type=discount.offer.offer_type enterprise_name=discount.offer.condition.enterprise_customer_name %}
                                     Discount of type {{ type }} provided by {{ enterprise_name }}
                                 {% endblocktrans %}
+                            {% elif discount.offer.condition.name == 'dynamic_discount_condition' %}
+                                Discount for your first upgrade
```
MatthewPiatetsky: i think this is missing translation tags

ecommerce/templates/oscar/basket/basket.html
```diff
@@ -13,10 +13,12 @@
 {% endblock navbar %}
 
 {% block pre_app_js %}
+  {% if formset_lines_data|length == 1 %}
```
MatthewPiatetsky, line 17: what is this doing?

MatthewPiatetsky (commented):
i'm generally more interested in whether the code works, which i'll be able to QA better during a QA session, but left a couple comments about clarity  
this PR is probably not suitable for screenshots, but if there are things you think you could screenshot and add to the PR that would be nice
also, if you had a writeup somewhere of how all of these things work, perhaps linking to it in the docstrings of a bunch of the new files could be helpful

ecommerce/extensions/offer/dynamic_conditional_offer.py
emma-green: Good Catch. Yes, this is a dynamic discount condition, I forgot to rename...

ecommerce/extensions/offer/tests/test_applicator.py
emma-green: oooh. That's cleaner. Done.

emma-green: done.

ecommerce/enterprise/tests/test_migrate_enterprise_conditional_offers_command.py
emma-green: yep!

ecommerce/enterprise/tests/test_migrate_enterprise_conditional_offers_command.py
```diff
@@ -6,6 +6,7 @@
 import logging
 
 from django.core.management import call_command
+from django.db.models import Q
```
cpennington: Not sure if pylint will catch this in ecom (I think it will?), but this import is unnecessary now

ecommerce/enterprise/tests/test_views.py
```diff
@@ -130,8 +131,9 @@ def test_post(self):
             'benefit_type': Benefit.PERCENTAGE,
             'benefit_value': expected_benefit_value,
         }
+        existing_offer_ids = [c.id for c in ConditionalOffer.objects.all()]
```
cpennington: This could be `ConditionalOffer.objects.all().values_list('id')`, and it would fetch less data from the database.

ecommerce/static/js/apps/basket_app.js
emma-green: if the user is not purchasing a course, there will not be a course key and therefore nothing to ask the lms for...

ecommerce/extensions/offer/tests/test_applicator.py
```diff
@@ -94,7 +97,8 @@ def test_log_is_fired_when_get_offers_with_bundle(self):
     @override_flag(CUSTOM_APPLICATOR_LOG_FLAG, active=True)
     def test_log_is_fired_when_get_offers_without_bundle(self):
         """ Verify that logs are fired when no bundle id is given but offers are being applied"""
-        site_offers = ConditionalOfferFactory.create_batch(3)
+        co_in_db = [co for co in ConditionalOffer.active.filter(offer_type=ConditionalOffer.SITE).all()]
```
cpennington: `all()` is superfluous when you use `filter` (and this is another place you can just use `list`).

ecommerce/templates/oscar/basket/basket.html
emma-green: only run the js if the user is only purchasing one course.

ecommerce/extensions/test/factories.py
```diff
@@ -311,3 +312,11 @@ class JournalBundleOfferFactory(ConditionalOfferFactory):
     offer_type = ConditionalOffer.SITE
     priority = OFFER_PRIORITY_ENTERPRISE
     status = ConditionalOffer.OPEN
+    status = ConditionalOffer.OPEN
```
cpennington: Was this an accidental duplication?

ecommerce/templates/edx/checkout/receipt.html
emma-green: What do I need to do to add the translation? Just add the tags?

ecommerce/enterprise/tests/test_migrate_enterprise_conditional_offers_command.py
emma-green: yep

ecommerce/extensions/test/factories.py
emma-green: yep. Good catch. Not sure how this happened...

ecommerce/extensions/payment/views/cybersource.py
emma-green: Yes, the real solution is to create the order prior to payment processing. This (hacky) solution asks for the information again, and adds it to the request. Which part of that do you feel needs more explanation here?

MatthewPiatetsky: Just a brief description like what you said I think could be useful in a comment

ecommerce/static/js/apps/basket_app.js
MatthewPiatetsky: in what situations is window.courseKey falsy?

ecommerce/templates/edx/checkout/receipt.html
MatthewPiatetsky: you can probably just use the trans tag https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#std:templatetag-trans
or the blocktrans tag like above

ecommerce/templates/oscar/basket/basket.html
MatthewPiatetsky: is this because the discount this js is for does not apply to bundles/only applies to single verified certs? 
if so, could we add a comment to that effect?

ecommerce/enterprise/tests/test_views.py
emma-green: yep

ecommerce/enterprise/tests/test_views.py
emma-green: yep

ecommerce/enterprise/tests/test_migrate_enterprise_conditional_offers_command.py
emma-green: sure

ecommerce/extensions/offer/migrations/0025_auto_20190624_1747.py
emma-green: I don't think so? It will be very easy to remove from the admin panel if needed? Feel free to disagree.

ecommerce/extensions/offer/tests/test_dynamic_conditional_offer.py
emma-green: sure

ecommerce/extensions/payment/views/cybersource.py
emma-green: k

ecommerce/extensions/offer/dynamic_conditional_offer.py
emma-green: Sure

ecommerce/templates/oscar/basket/basket.html
emma-green: sure

ecommerce/programs/tests/test_views.py
emma-green: It's an exclude. I've changed it to reflect that.

ecommerce/extensions/offer/tests/test_applicator.py
emma-green: fair.

ecommerce/static/js/apps/basket_app.js
emma-green: If it doesn't exist. I'll change this to window.courseKey !== undefined to make that more clear...

ecommerce/templates/edx/checkout/receipt.html
emma-green: done

MatthewPiatetsky (approved):
don't forget to squash commits
if this is merged by sharing time maybe a demo would be useful

ecommerce/extensions/offer/migrations/0025_auto_20190624_1747.py
cpennington: reverse isn't for the admin panel, it's for rollbacks.

emma-green: I added it. :)

cpennington (approved):

PR #2345, opened 2019-06-20
implement product details in payment apis

robrap:
ARCH-856

### DIFF
--- a/ecommerce/extensions/basket/views.py
+++ b/ecommerce/extensions/basket/views.py
@@ -410,6 +410,8 @@ class BasketLogicMixin(object):
         num_of_items = self.request.basket.num_items
         context.update({
             'formset_lines_data': zip(formset, lines_data),
+            # the new payment api is not using the formset, so the zip above doesn't work
+            'lines_data': lines_data,
             'free_basket': context['order_total'].incl_tax == 0,
             'homepage_url': get_lms_url(''),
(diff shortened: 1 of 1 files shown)


ecommerce/extensions/basket/views.py
```diff
@@ -505,6 +498,14 @@ def get_serialized_basket(self, context):
             ],
         }
 
+        for line_data in context['lines_data']:
+            serialized_basket['products'].append({
+                'sku': line_data['sku'],
+                'title': line_data['product_title'],
```
robrap, line 504: Reviewers: I renamed `name` to `title` to better match the original names, but this is breaking the frontend.  Can someone please take this over and make that frontend PR as well?  Thank you!

albemarle (approved):

ecommerce/extensions/basket/views.py
robrap: FYI: @julianajlk will be taking care of the frontend change.

robrap:
@edx/ecommerce: FYI: This is also basically just messing with the new api endpoints, and should be safe to merge/deploy.  I'm going OOO for a couple of days, so someone else on my team may take care of it, but they may not push it to Prod.

albemarle:
The front-end PR is https://github.com/edx/frontend-app-payment/pull/40 .

PR #2349, opened 2019-06-21
ENT-1987 Take 2

muhammad-ammar:
**Description:** This updates the `Enterprise Customer Details` link below `Enterprise Customer Catalog` field on Enterprise Coupon Detail/Edit screen. 

**JIRA:** [ENT-1987](https://openedx.atlassian.net/browse/ENT-1987)

**Testing:** PR changes are deployed to sandbox.

**Admin Credentials:**
username: edx 
password: edx

1. Login to https://ecommerce-mammar.sandbox.edx.org/login
2. Coupon Edit Screen: [Enterprise Catalog Details link](https://ecommerce-mammar.sandbox.edx.org/enterprise/coupons/7/edit/) should work
3. Coupon Detail Screen: [Enterprise Catalog Details link](https://ecommerce-mammar.sandbox.edx.org/enterprise/coupons/7/) should work

### DIFF
--- a/ecommerce/extensions/api/v2/views/enterprise.py
+++ b/ecommerce/extensions/api/v2/views/enterprise.py
@@ -6,16 +6,22 @@ from django.core.exceptions import ValidationError
 from django.shortcuts import get_object_or_404
 from edx_rbac.decorators import permission_required
 from oscar.core.loading import get_model
+from requests.exceptions import ConnectionError, Timeout
 from rest_framework import generics, serializers, status
 from rest_framework.decorators import detail_route, list_route
 from rest_framework.permissions import IsAdminUser, IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.viewsets import ViewSet
 from six.moves.urllib.parse import urlparse  # pylint: disable=import-error, relative-import
+from slumber.exceptions import SlumberHttpBaseException
 
-from ecommerce.core.constants import COUPON_PRODUCT_CLASS_NAME
+from ecommerce.core.constants import COUPON_PRODUCT_CLASS_NAME, DEFAULT_CATALOG_PAGE_SIZE
 from ecommerce.core.utils import log_message_and_raise_validation_error
-from ecommerce.enterprise.utils import get_enterprise_customer_catalogs, get_enterprise_customers
+from ecommerce.enterprise.utils import (
+    get_enterprise_catalog,
+    get_enterprise_customer_catalogs,
+    get_enterprise_customers
+)
 from ecommerce.extensions.api.pagination import DatatablesDefaultPagination
 from ecommerce.extensions.api.serializers import (
     CouponCodeAssignmentSerializer,
@@ -64,6 +70,7 @@ DEPRECATED_COUPON_CATEGORIES = ['Bulk Enrollment']
 class EnterpriseCustomerViewSet(generics.GenericAPIView):
 
     permission_classes = (IsAuthenticated, IsAdminUser,)
+    queryset = ''
 
     def get(self, request):
         site = request.site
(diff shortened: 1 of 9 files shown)


ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -84,6 +89,17 @@ def get(self, request):
         )
         return Response(data=enterprise_catalogs)
 
+    def retrieve(self, request, **kwargs):
+        endpoint_request_url = urlparse(request.build_absolute_uri())._replace(query=None).geturl()
+        catalog = get_enterprise_catalog(
```
brittneyexline: you should catch potential exceptions here and return a formatted error message so that there is some insight into any failing requests

brittneyexline (commented):
one comment for you, and a few observations from seeing what you have on the sandbox.

i noticed that all of the catalog links are returning empty results, and this is true even when i removed some filters from the default query. i think it would be good to double check why.
i also noticed that i get a 500 for the catalog link when i tried with a coupon that had a new enterprise customer (ie, the pied piper coupon/customer). the ecommerce worker user appears to be linked to the original customer. if the ecommerce worker is going to have the openedx operator role, they should never be linked to an enterprise customer, and the catalog links should work for all enterprise customers and catalogs.

muhammad-ammar:
@brittneyexline The 500 is coming because implicit access is not working on sandbox because oauth authentication is not working. Only the explicit access is currently working on my sandbox and linking the ecommerce_worker user with multiple enterprises will not work because we are explicitly checking the multiple enterprises [here](https://github.com/edx/edx-enterprise/blob/master/enterprise/models.py#L1805-L1828)

I will add the exception handling and see why empty results are returned?

brittneyexline:
@muhammad-ammar - the original reason we needed to change our approach on this is because this case needs to work for implicit access specifically; this page is used by the ECS team who are internal edx admins. this page would never be exposed to a non-edx employee on our instance. if i manage to make progress on the business sandbox today, then we can test there first. otherwise, if we're having trouble testing, we can release to stage and test there.

muhammad-ammar:
@brittneyexline Authentication is fixed on sandbox and all the catalog related issues are fixed now. Implicit access now works. Please have a look again.

ecommerce/extensions/api/v2/views/enterprise.py
```diff
@@ -84,6 +91,29 @@ def get(self, request):
...
+                endpoint_request_url=endpoint_request_url
+            )
+        except (ConnectionError, SlumberHttpBaseException, Timeout) as exc:
+            logger.exception(
+                'Unable to retrieve catalog for enterprise customer! customer: %s, Exception: %s',
+                kwargs.get('enterprise_catalog_uuid'),
+                exc
+            )
+            return Response(
+                {'error': 'Unable to retrieve enterprise catalog'},
```
brittneyexline: i think we can surface the exception message here as well

brittneyexline (approved):
just one small thing i'd like to see before i merge/release this, otherwise looks good!

muhammad-ammar:
@brittneyexline Addressed your feedback. Any idea how to render indented catalog response when  Enterprise Catalog Details link is opened in the new tab? We already have `JSONRenderer` as our default renderer in ecommerce but json is not indented. I have also tried the `indent` query param but no luck.

brittneyexline:
@muhammad-ammar - not sure if there is much you can do from the code side, i think rendering json is the right thing to do. i have a chrome extension that formats json so that would be my recommendation to those utilizing this link.

muhammad-ammar:
@brittneyexline BTW firefox implicitly beautify the json response without any extension.

PR #2351, opened 2019-06-21
install code-annotations

estute:
As part of the feature toggle report generator, we need to have code-annotations installed in each IDA that we want to report on. This will be used to gather data on the in-code definitions of feature toggles.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -102,7 +102,6 @@ pymongo==3.8.0            # via edx-opaque-keys
 pyopenssl==19.0.0         # via ndg-httpsclient, paypalrestsdk
 python-dateutil==2.8.0
 python-openid==2.2.5 ; python_version == "2.7"
-python3-openid==3.1.0 ; python_version >= "3"  # via social-auth-core
 pytz==2016.10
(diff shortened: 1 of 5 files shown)


requirements/base.txt
```diff
@@ -102,7 +102,6 @@ pymongo==3.8.0            # via edx-opaque-keys
 pyopenssl==19.0.0         # via ndg-httpsclient, paypalrestsdk
 python-dateutil==2.8.0
 python-openid==2.2.5 ; python_version == "2.7"
-python3-openid==3.1.0 ; python_version >= "3"  # via social-auth-core
```
jmbowman, line 105: This removal looks suspicious given that social-auth-core wasn't upgraded or removed.  Could you try adding `--rebuild` to each `pip-compile` command and re-run `make upgrade`?

estute: Yeah, I was suspicious as well. Will do.

openedx-webhooks:
Although this pull request is already closed,  I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #2378, opened 2019-06-24
install code annotations

estute:
As part of the feature toggle report generator, we need to have code-annotations installed in each IDA that we want to report on. This will be used to gather data on the in-code definitions of feature toggles.

Previous branch got messed up (https://github.com/edx/ecommerce/pull/2351), so I reran it clean

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -102,7 +102,6 @@ pymongo==3.8.0            # via edx-opaque-keys
 pyopenssl==19.0.0         # via ndg-httpsclient, paypalrestsdk
 python-dateutil==2.8.0
 python-openid==2.2.5 ; python_version == "2.7"
-python3-openid==3.1.0 ; python_version >= "3"  # via social-auth-core
 pytz==2016.10
 pyyaml==5.1.1             # via edx-django-release-util
 rcssmin==1.0.6            # via django-compressor
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -46,8 +46,8 @@ premailer==2.9.2
 pycountry==17.1.8
 pygments
 python-dateutil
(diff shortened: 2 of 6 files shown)


requirements/base.txt
```diff
@@ -102,7 +102,6 @@ pymongo==3.8.0            # via edx-opaque-keys
 pyopenssl==19.0.0         # via ndg-httpsclient, paypalrestsdk
 python-dateutil==2.8.0
 python-openid==2.2.5 ; python_version == "2.7"
-python3-openid==3.1.0 ; python_version >= "3"  # via social-auth-core
```
jmbowman, line 105: social-auth-core has some logic in setup.py that pip-compile can't follow for Python 2/3 compatibility, we may have to add this explicitly to the base requirements for now: https://github.com/python-social-auth/social-core/blob/c607873521eb1efe180631e842d94a134359f8b7/setup.py

estute: I don't think pip-tools/pip-compile has a great story for compiling a requirements file for multiple Python versions at once. If we had the following:
```
python-openid ; python_version == "2.7"
python3-openid ; python_version >= "3"
```
We could explicitly pass in a version to the invocation of the command, as explained here:
```
If you use multiple Python versions, you can run pip-compile as py -X.Y -m piptools compile ... on Windows and pythonX.Y -m piptools compile ... on other systems.
```
from (https://pypi.org/project/pip-tools/), but I believe this would overwrite the files created from previous runs.

Do you think that it is ok as it stands?

jmbowman: I guess, it looks like we have the same problem in edx-platform.  We'll tackle that when we're further along with the Python 3 upgrade.

PR #2382, opened 2019-06-25
Unpin isort and upgrade packages

cpennington:


requirements/base.txt
```diff
@@ -102,7 +102,6 @@ pymongo==3.8.0            # via edx-opaque-keys
 pyopenssl==19.0.0         # via ndg-httpsclient, paypalrestsdk
 python-dateutil==2.8.0
 python-openid==2.2.5 ; python_version == "2.7"
-python3-openid==3.1.0 ; python_version >= "3"  # via social-auth-core
```
crice100: Should this have been removed?

crice100 (approved):
See question about _python3-openid_

requirements/base.txt
cpennington: I'm not sure. Given that the social-auth-core version number didn't change, probably not, I'll put it back.

requirements/production.txt
```diff
@@ -107,7 +107,6 @@ pyopenssl==19.0.0         # via ndg-httpsclient, paypalrestsdk
 python-dateutil==2.8.0
 python-memcached==1.58
 python-openid==2.2.5 ; python_version == "2.7"
-python3-openid==3.1.0 ; python_version >= "3"  # via social-auth-core
```
crice100: Did we want to put this back, too?

cpennington: oops. yup.

PR #2384, opened 2019-06-25
Upgrade test/dev requirements

cpennington:


.pycodestyle
```diff
@@ -0,0 +1,8 @@
+[pycodestyle]
+# Disabled:
+# E501: line length
```
crice100, line 3: Just curious--are these checks enabled or disabled? I'm not sure how to read the initial #

crice100 (approved):

.pycodestyle
cpennington: I'm commenting on what the checks being ignored actually are.

PR #2385, opened 2019-06-26
Toggle for Microfrontend for Basket Page

nasthagiri:
https://openedx.atlassian.net/browse/ARCH-758


ecommerce/extensions/basket/views.py
```diff
@@ -56,6 +59,20 @@
 Voucher = get_model('voucher', 'Voucher')
 
 
+def _redirect_to_payment_microfrontend_if_configured(request):
+    if waffle.flag_is_active(request, ENABLE_MICROFRONTEND_FOR_BASKET_PAGE_FLAG_NAME):
+        if (
+                request.site.siteconfiguration.enable_microfrontend_for_basket_page and
+                request.site.siteconfiguration.payment_microfrontend_url
```
douglashall, line 66: `enable_microfrontend_for_basket_page` seems a bit redundant. Maybe just have the presence of the `payment_microfrontend_url` value be the trigger here.

douglashall (approved):

ecommerce/extensions/basket/views.py
nasthagiri: Yeah, I was thinking about that.  I went this way for the following reasons:

1. To be consistent with the presence of this boolean field in edx-platform: https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/user_api/accounts/toggles.py

2. In case we want to have multiple pages redirecting to microfrontend(s) and we want to keep the enablement for a WL site decoupled from the set of URLs.

3. Coding wise, it "smelled" a bit to have a `None` value be indicative of an intention.

robrap (commented):
👍

ecommerce/extensions/payment/constants.py
```diff
@@ -38,6 +38,19 @@
 
 CLIENT_SIDE_CHECKOUT_FLAG_NAME = 'enable_client_side_checkout'
 
+# .. toggle_name: ENABLE_MICROFRONTEND_FOR_BASKET_PAGE_FLAG_NAME
```
robrap, line 41: Just noticed this, but shouldn't toggle_name be `enable_microfrontend_for_basket_page`?  I think the report should output what one would find in admin.

robrap: @nasthagiri: I am going to make this change in the e2e PR which is relevant.

PR #2389, opened 2019-06-26
TE-2337: WIP: run e2e tests against firefox container

robrap:
FYI: This is a starter PR for discussion.  It does not actually fix the e2e tests yet.

This currently gives the following error when trying to use the browser:
```
WebDriverException: Message: First match w3c capabilities is zero length
```
The following old ticket with the same error may or may not be related: https://openedx.atlassian.net/browse/TE-2337

### DIFF
--- a/docs/test_ecommerce.rst
+++ b/docs/test_ecommerce.rst
@@ -150,6 +150,8 @@ the following procedures.
    :depth: 1
    :local:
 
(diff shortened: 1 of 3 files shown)


robrap:
This is only meant to be a reference for TE-2337.  I am going to close this PR.

PR #2392, opened 2019-06-27
INCR-537 - Run python-modernize on ./ecommerce/enterprise

ihtram:


ecommerce/enterprise/utils.py
```diff
@@ -1,11 +1,13 @@
...
 Helper methods for enterprise app.
 """
+from __future__ import absolute_import
+
 import hashlib
 import hmac
 import logging
 from collections import OrderedDict
-from urllib import urlencode
+from functools import reduce  # pylint: disable=redefined-builtin
```
crice100, line 10: @cpennington Is disabling _redefined-builtin_ ok here?

cpennington: Yes. `reduce` is removed in python3, so you need to pull it from functools, but that shadows the builtin in python2.

cpennington (approved):

PR #2397, opened 2019-06-27
INCR-542 - Run python-modernize on ./ecommerce/extensions/payment/tests

ihtram:


ecommerce/extensions/payment/tests/mixins.py
```diff
@@ -1,13 +1,13 @@
...
 
 import datetime
 import os
 from decimal import Decimal
-from urlparse import urljoin
 
 import ddt
 import mock
 import responses
+import six  # pylint: disable=ungrouped-imports
```
crice100, line 10: @cpennington Do we want to disable ungrouped imports?

cpennington: Yup. See the other PR for an explanation.

cpennington (approved):

PR #2398, opened 2019-06-27
INCR-543 - Run python-modernize on ./ecommerce/extensions/payment

ihtram:


ihtram:
we need to create this PR again after merging rest of the PRs as this runs modernize on payments base directory. This will result in code conflicts.

PR #2403, opened 2019-06-27
INCR-548 - Run python-modernize on ./ecommerce/extensions/api

ihtram:


ihtram:
we need to create this PR again after merging rest of the PRs as this runs modernize on api base directory. This will result in code conflicts.

PR #2404, opened 2019-06-27
INCR-549 - Run python-modernize on ./ecommerce/extensions/catalogue

ihtram:


ecommerce/extensions/catalogue/management/commands/convert_course.py
```diff
@@ -1,11 +1,13 @@
-from __future__ import unicode_literals
+from __future__ import absolute_import, unicode_literals
 
 import logging
 
+import six  # pylint: disable=ungrouped-imports
```
crice100, line 5: @cpennington Should we be disabling _ungrouped-imports_?

cpennington: Yes, pylint and isort are incompatible until pylint releases a new release, so disabling `ungrouped-imports` on this import when we both `import six` and `from six.foo import bar` has been the way I've been handling it.

cpennington (approved):

PR #2406, opened 2019-06-27
INCR-551 - Run python-modernize on ./ecommerce/extensions

ihtram:


ihtram:
we need to create this PR again after merging rest of the PRs as this runs modernize on extensions base directory. This will result in code conflicts.

